PR #92, opened 2015-08-17
Accept QoS profiles via an XML file

esteve:
Connect to ros2/ros2#86

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -97,11 +97,11 @@ namespace utilities
 {
 
 void
-init(int argc, char * argv[])
+init(int argc, char * argv[], const std::string & qos_xml_filename = "")
 {
   (void)argc;
   (void)argv;
-  rmw_ret_t status = rmw_init();
+  rmw_ret_t status = rmw_init(qos_xml_filename.c_str());
   if (status != RMW_RET_OK) {
     // *INDENT-OFF* (prevent uncrustify from making unecessary indents here)
     throw std::runtime_error(


dirk-thomas:
This this exposes vendor specific configuration should we let the user call vendor specific API instead? The problem is otherwise that the code uses only rclcpp / rmw but is still not portable.

esteve:
@dirk-thomas this is tricky. The QoSProvider API which standardizes how loading an XML file should happen, is supposed to be now part of the DDS spec, but it's not implemented by all of the vendors. And to top it we currently don't use the new C++ API of either Connext or OpenSplice.

For now, I'd expose a vendor specific API, but I wanted to see what the feedback would be before implementing that.

wjwwood:
Maybe I missed the discussion about this, but why do we need to be able to load from an xml file?

dirk-thomas:
@esteve It is fine to add vendor specific API but then it should be really vendor specific and not provided by rclcpp / rmw. The user should call something like `connext::set_qos_xml_filename()` instead since the argument passes is not vendor agnostic.

esteve:
@wjwwood the spec for QoS is huge, so adding all of the possible QoS settings as part of our API seems a bit infeasible. If a user who has DDS already deployed and has experience with a predefined set of QoS profiles, they may want to reuse that knowledge. The DDS defines an entry point for loading QoS profiles via XML files (the QosProvider API), but it requires further changes in our code (migrate to the new C++ API)

esteve:
@dirk-thomas yeah, that's what I meant, before adding that I wanted to discuss such approach. This PR does the opposite, though, less invasive in terms of API changes, but with different behaviors, which I think is not the most appropriate approach.

wjwwood:
In that case I agree that this should not be passed through our API, but through a DDS specific or even vendor specific API. As long as the value of the parameter is vendor specific or even just DDS specific we should avoid it. Imagine what the eprosima implementation or Morgan's implementation would have to do in order to accommodate this in our API. If we make it optional, then we need to have a way to check its availability and that complicates our interface as well as the user's code which might want to use it.

I understand why someone might want to do this at some point, but my question is do we need it now? In my opinion, anything we're configuring through "our" QoS profiles should be accessible through our API and shouldn't require additional DDS QoS settings in order to work correctly. I think it would be ok if we need a custom XML file for all of ROS, which could be owned by the rmw implementation of that vendor and applied under our API's.

Do we currently have a use case for using the DDS XML configuration file where our interface is insufficient or expanding it to be sufficient is overly burdensome? I could imagine it would be convenient for your testing of the various QoS settings, for example.

esteve:
@wjwwood things like buffers for the UDP transport or the discovery timeouts may be easier to configure via the XML file instead of making our API more complex.

wjwwood:
I see, but do you expect users to be changing that on a per node basis? (what is the scope of these changes, per process, per node, per topic?)

What I was imagining is that we define profiles, which users can select one of, then modify QoS settings we provide, like history length, durability, etc... These named profiles have a description and the rmw implementations would have the opportunity to have implementation settings to help achieve that.

For example, if you have a lossy wifi profile, then the user selects that from our API, but under the hood the Connext implementation may do one thing to try and best achieve that and OpenSplice might do something else. The idea is that this profile is well enough described that the behavior can remain quantitively consistent between implementations, but my vary qualitatively.

Maybe that's not realistic, but that's what I had imagined. What I think we should really try to avoid is users feeling as if they must set vendor specific settings for common use cases. There will always be people who need to use the implementation directly, but I'm not sure we're trying to meet their needs yet. Once we start considering those use cases more invasive access to the implementation may be warranted.

esteve:
@wjwwood QoS profiles in a XML can configure settings on every entity in a DDS network (topics, writers, readers, etc.) at any level (presentation, durability, etc.). I don't the majority of users to delve into every detail of the QoS settings, but it makes tweaking easier in order to find the most appropriate values for certain entities (buffer and max message sizes, memory allocation, etc.)

An example of many (but not all) the settings that can be configure can be found in:

https://github.com/rticommunity/rticonnextdds-src/blob/master/dds_c.1.0/resource/qos_profiles_5.2.0/xml/NDDS_QOS_PROFILES.xml

wjwwood:
> https://github.com/rticommunity/rticonnextdds-src/blob/master/dds_c.1.0/resource/qos_profiles_5.2.0/xml/BuiltinProfiles.documentationONLY.xml

That link appears to be a 404. (I'm logged in, but maybe only you have access to it. If it's private we shouldn't link it here.)

The question remains, do we need to expose that for users? I understand that it's useful when tuning a profile, but do we expect users to do that? What about my suggestion of having user level profiles, which can be tied to an xml file by way of an implementation detail, but importantly the user does not specify an XML file directly?

esteve:
@wjwwood that's private, but it's part of the community source distribution of RTI Connext. Anyway, I removed your link and mine.

I don't think we should expose so much detail to users. Those who are interested in configuring such variables, have a deep understanding of DDS and would probably feel more comfortable using the vendor's tools.

jacquelinekay:
@esteve identifies two kinds of users when it comes to QoS policies:

1. The users who prioritize convenience over control. These users probably came from ROS 1 and want to see an API similar to ROS 1. They care less about the complexities of the underlying transport layer and want to save as much time as possible so they can work on the layer above ROS 2. This probably applies to a lot of people in research.

2. Veteran DDS users/systems experts. They might be coming to ROS 2 because they want a convenient API specialized for robotics that's compatible at the wire level with their existing DDS system, and/or they might be interested in migrating an existing DDS system to ROS 2. They are interested in ROS 2 because they are already familiar with the strengths of DDS, and they already own a license to a vendor implementation.

I think there's another group we should consider:

3. ROS "power users" who do not already use DDS, but do care deeply about the performance of the transport layer of their system. This may include engineers at small companies trying to deploy industrial robots, autonomous vehicles, quadcopters, real-time systems, etc. Many users in this category may currently use `ros_control` or other tools to use ROS 1 in a high-performance context, or they use ROS introspection tools like `rviz` and use something else entirely for transport.

Group 3 would probably be willing to spend the time to figure out the complexities of DDS QoS options and tune the XML file with all their settings, but I agree with @wjwwood that this should not be the only entry point into the QoS API, for these reasons:

1. XML parsing is not implemented by all vendors (from what I've gathered).
2. Users may want to changing QoS settings during runtime.
3. William's argument: even if XML parsing was implemented by all vendors, QoS settings may not have the same behavior across different implementations. Thus the same ROS 2 code with the same QoS settings might have different behaviors across implementations. Creating "tuned" QoS profiles would compensate for this failure of the spec to create consistent behavior, but then it would be on us to guarantee that consistency. I'm actually less concerned about this point than I am about the first two.

I think it's a fine idea to expose the option for xml configuration, but I also agree with @dirk-thomas that vendor-specific behavior should not be in `rclcpp`.

esteve:
Thank you all for your feedback.

I agree with @jacquelinekay about a third set of users, which are a bridge between the DDS experts and the ROS users who don't want/need to so much control.

However, the QoS API is humongous (really!), adding counterparts of all the classes and settings would effectively turn our API into a mirror of the DDS QoS API, which doesn't provide much value. I believe a small set of settings should be provided via our API (like what we already do, possibly with more settings), but more specialized usecases should just use the mechanisms provided by DDS.

The standardized way of accessing the full set of QoS settings is via the `QosProvider` API (included in Connext and OpenSplice) [1][2], and the most convenient way of doing so is via XML files. XML files can be loaded at any time, and the changes become effective immediately.

The reason the `QosProvider` is not used is because we're not using the modern C++ API defined by DDS. In OpenSplice's case, we can't mix `dcps` and `isocpp`, so even if the `QosProvider` API is available in OpenSplice, we can't use it.

The code in this PR assumes an XML file with RTI's schema, so it shouldn't belong in `rclcpp`, however accepting the XML files with the standard `QosProvider` schema is fine in my opinion, as it's not vendor specific and it's a schema approved by the DDS spec. However, not all vendors implement the `QosProvider` API, so in those cases we should raise an exception.

1 - https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/api/connext_dds/api_cpp2/classdds_1_1core_1_1QosProvider.html
2 - http://dev.opensplice.org/releases/downloads/docs/isocpp/html/spec_2dds_2core_2_qos_provider_8hpp.html

wjwwood:
> However, the QoS API is humongous (really!), adding counterparts of all the classes and settings would effectively turn our API into a mirror of the DDS QoS API, which doesn't provide much value. 

I think we need to discuss this in our meeting, because as far as I can see no one is suggesting we do this or that we shouldn't use the XML files. The question, which we've been circling but cannot seem to get clear opinions on, is can we allow usage of the XML files by us and by users without having them pass a path to one through the rclcpp API. There have been a few suggestions:

- Use a separate vendor or DDS specific API.
 - Allows arbitrary manipulation of the underlying DDS API.
 - Using this locks the user's application into DDS or a specific vendor, which is not desirable.
- Have described profiles who's goal is to achieve the described behavior and let the rmw implementation developers craft xml files which best achieve that goal.
 - This means users will not ever have to pass an XML file.
 - But it puts the pressure of meeting the described behavior of the profile on the rmw implementation developers.
 - It might be possible to allow user to create new profiles, which they could implement in vendor specific ways. Then if someone else wanted to use their profile with a new vendor they would potentially have to customize the profile for their rmw implementation.

> accepting the XML files with the standard QosProvider schema is fine in my opinion, as it's not vendor specific and it's a schema approved by the DDS spec

As far as I'm aware, our aim is to not only be vendor agnostic but also DDS agnostic? So until we re-evaluate this goal, I'd have to disagree that we want to accept any DDS specific XML file, regardless of its portability between vendors.

> However, not all vendors implement the QosProvider API, so in those cases we should raise an exception.

This would apply to eProsima and Morgan. I don't think we must constrain ourselves to support them explicitly, but it would be nice to not make it harder for them and I'm very wary of having partially support parts of the API.

esteve:
> As far as I'm aware, our aim is to not only be vendor agnostic but also DDS agnostic? So until we re-evaluate this goal, I'd have to disagree that we want to accept any DDS specific XML file, regardless of its portability between vendors.

I think that's the root of all these questions. Although the `rmw` layer is agnostic, we've put all our efforts into DDS, we didn't attempt to implement the `rmw` in any other transport. And I think that's fine, so at certain point we'll accept that DDS is a requirement for ROS2 or decide that it's just one of the many transports that can be used to implement the ROS2 spec/API.

> This would apply to eProsima and Morgan. I don't think we must constrain ourselves to support them explicitly, but it would be nice to not make it harder for them and I'm very wary of having partially support parts of the API.

Except that we're not forbidding these implementations from being compatible with ROS2, loading DDS-approved XML files that define QoS profiles is a supplement and a convenience for our users to tune even the smallest entity in a DDS graph, but not necessarily a requirement.

gerkey:
Result of discussion:

* Put the XML-loading support (the DDS XML spec, not the RTI-specific XML spec) into a package that is RTI-specific.  Over time, revisit it and consider promoting it to a common API as more implementations come to support it.
* Consider putting this in the ros_to_dds package, which has the extra benefit of making it clear in the dependency graph what's happening.

mikaelarguedas:
Decided not to provide this feature because of conflicting vendor definition. We may offer a design spec to do that in vendor agnostic way in the future.

gbiggs:
Is this a case of the vendors not complying with the spec? The DDS4CCM specification defines the XML to be used to set QoS properties.

wjwwood:
Yes, in part that's true, but for me at least the bigger thing is that we're passing a DDS specific (and in some cases vendor specific) thing through our ROS API which is not the best IMO.

PR #93, opened 2015-08-22
Prevent malloc(0) in memory strategy

jacquelinekay:
Check before we try to malloc something of size 0 and short-circuit the call.

Thanks to @codebot for discovering this and doing the work to make it actually compile.


tfoote:
+1

dirk-thomas:
+1

PR #96, opened 2015-08-25
check intra-process subscription handle before dereference

jacquelinekay:
This was causing a segfault in the realtime demo, which does not use intra-process messaging.


wjwwood:
+1

dirk-thomas:
+1

PR #97, opened 2015-08-27
Document user-facing code in rclcpp

jacquelinekay:
With a focus on functions and members used in the demos, such as executors, publish/subscribe and memory strategy classes.

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -136,10 +136,11 @@ public:
     }
   }
 
-  /// Add a node to executor, complete the next available unit of work, and remove the node.
+  /// Add a node to executor, execute the next available unit of work, and remove the node.
   /* \param[in] node Shared pointer to the node to add.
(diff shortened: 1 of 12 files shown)


rclcpp/include/rclcpp/executor.hpp
```diff
@@ -36,6 +36,15 @@ namespace rclcpp
 namespace executor
 {
 
+/// Coordinates the order and timing of available communication tasks.
```
dirk-thomas: Verb without `s`: `Coordinate ...`

dirk-thomas:
Currently the detailed descriptions are not extracted by doxygen at all. I assume because they need to start with `/**` instead of `/*` (http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#cppblock)?

esteve:
+1

gerkey:
I'm resolving conflicts on this now.

gerkey:
This PR only adds/changes comments, so I'm merging it.

PR #98, opened 2015-08-27
Add missing API for the intra process demo

wjwwood:
This does a few things:

- Adds a `publish(std::unique_ptr<MessageT>)` like method.
- Allows the user's subscription callback to look like any of these:
 - `void cb(const std::shared_ptr<MessageT>)`
 - `void cb(const std::shared_ptr<MessageT>, const rmw_message_info_t &)`
 - `void cb(const std::unique_ptr<MessageT>)`
 - `void cb(const std::unique_ptr<MessageT>, const rmw_message_info_t &)`

It also adds (along with ros2/rmw#36) a `from_intra_process` bool to the `rmw_message_info_t` struct. This allows better tests and demonstrating intra process is being used in demos.

Connects to ros2/demos#12

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -29,6 +29,7 @@
 #include <rclcpp/callback_group.hpp>
 #include <rclcpp/client.hpp>
 #include <rclcpp/context.hpp>
+#include <rclcpp/function_traits.hpp>
 #include <rclcpp/macros.hpp>
 #include <rclcpp/message_memory_strategy.hpp>
 #include <rclcpp/parameter.hpp>
@@ -55,42 +56,6 @@ class Executor;
 
 namespace node
 {
-
-/* NOTE(esteve):
- * We support service callbacks that can optionally take the request id,
- * which should be possible with two overloaded create_service methods,
- * but unfortunately std::function's constructor on VS2015 is too greedy,
- * so we need a mechanism for checking the arity and the type of each argument
- * in a callback function.
- */
-template<typename FunctionT>
-struct function_traits
-{
-  static constexpr std::size_t arity =
-    function_traits<decltype( & FunctionT::operator())>::arity - 1;
-
-  template<std::size_t N>
-  using argument_type =
-      typename function_traits<decltype( & FunctionT::operator())>::template argument_type<N + 1>;
-};
-
-template<typename ReturnTypeT, typename ... Args>
-struct function_traits<ReturnTypeT(Args ...)>
-{
-  static constexpr std::size_t arity = sizeof ... (Args);
-
-  template<std::size_t N>
-  using argument_type = typename std::tuple_element<N, std::tuple<Args ...>>::type;
-};
-
-template<typename ReturnTypeT, typename ... Args>
-struct function_traits<ReturnTypeT (*)(Args ...)>: public function_traits<ReturnTypeT(Args ...)>
-{};
-
-template<typename ClassT, typename ReturnTypeT, typename ... Args>
-struct function_traits<ReturnTypeT (ClassT::*)(Args ...) const>
-  : public function_traits<ReturnTypeT(ClassT &, Args ...)>
-{};
 
 /* ROS Node Interface.
  *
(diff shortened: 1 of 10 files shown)


wjwwood:
It's worth note that the subscription callbacks which receive unique pointers will cause memory allocation, since our memory strategy has no way of handling unique pointers as storage.

jacquelinekay:
also fixes https://github.com/ros2/ros2/issues/94, correct?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -135,7 +135,53 @@ class Node
...
+    bool ignore_local_publications = false,
+    typename rclcpp::message_memory_strategy::MessageMemoryStrategy<MessageT>::SharedPtr
+    msg_mem_strat = nullptr);
+
+  template<typename MessageT>
+  typename rclcpp::subscription::Subscription<MessageT>::SharedPtr
+  create_subscription(
+    const std::string & topic_name,
+    const rmw_qos_profile_t & qos_profile,
+    typename rclcpp::subscription::AnySubscriptionCallback<MessageT>::UniquePtrWithInfoCallback
```
jacquelinekay: Perhaps we could template create_subscription on the callback type? It might save a lot of code.

wjwwood: We could do so for the prototype's listed here, but we still have to specialize them in the `node_impl.hpp` file. I'll do that.

wjwwood: Actually, I don't think it's possible since we need to partially specialize them and that doesn't seem to be allowed for function templates and/or member function templates.

jacquelinekay: Got it. Even if it is possible, it's probably not worth spending time on right now.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -148,11 +194,12 @@ class Node
     rclcpp::timer::CallbackType callback,
     rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr);
 
-  rclcpp::timer::WallTimer::SharedPtr
-  create_wall_timer(
-    std::chrono::duration<long double, std::nano> period,
-    rclcpp::timer::CallbackType callback,
-    rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr);
+  // TODO(wjwwood): reenable this once I figure out why the demo doesn't build with it.
```
jacquelinekay: I got the demo to build with this function uncommented by changing https://github.com/ros2/demos/pull/12/files#diff-1c605fc584a4b2a7d84651a92fed1390R26 to:

```
timer_ = this->create_wall_timer(1.0_s, [this]() { ...
```

The reason it doesn't compile is because `1_s` is now ambiguous; the 1 could be an int or a double.

Even if you remove the definition of operator `"" _s` that takes a double, it doesn't compile with `1_s` specified to `create_wall_timer`, because there's no `"" _s` operator that returns nanoseconds, which is expected by `create_wall_timer`. And adding a function that returns nanoseconds and takes an integer would be ambiguous and also wouldn't compile (as you probably know from writing the functions added to `rclcpp.hpp` in this PR).

But, overall, I actually think it's fine to disable the `create_wall_timer` function that's commented out here, because it allows the user to specify times in sub-nanosecond precision, which is hard to achieve on most computing platforms...

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -98,7 +99,21 @@ class Publisher
       // *INDENT-ON*
     }
     if (store_intra_process_message_) {
-      uint64_t message_seq = store_intra_process_message_(intra_process_publisher_id_, msg);
+      // Take the pointer from the unique_msg, release it and pass as a void *
```
jacquelinekay: This seemed weird to me at first but it makes sense that it's necessary to transfer the ownership of the unique_ptr.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -87,29 +88,65 @@ class Publisher
...
+    //   subscriptions. For now call the other publish().
+    std::unique_ptr<MessageT> unique_msg(new MessageT(*msg.get()));
+    return this->publish(unique_msg);
+  }
+
+  template<typename MessageT>
+  void
+  publish(const MessageT & msg)
+  {
+    std::unique_ptr<MessageT> unique_msg(new MessageT(msg));
```
jacquelinekay: one last nitpick: Can you do the same thing you did for the shared pointer publish here (with `do_inter_process_publish(&msg)` instead)?

wjwwood: Ah thanks, I missed that: 8464d96

PR #99, opened 2015-08-27
declare dependency

dirk-thomas:
It must be declared correctly in order to not break isolated builds.


jacquelinekay:
+1, invoking ament with `--isolated` with the `fix_missing_dependency` branches works for me now

tfoote:
+1

PR #102, opened 2015-09-04
always ensure that the ROS_DOMAIN_ID is set

wjwwood:
Needed to support https://github.com/ros2/rmw_opensplice/pull/81

Connects to ros2/rmw_opensplice#81


wjwwood:
This is now just a minor improvement. I can squash before merging.

dirk-thomas:
+1

PR #106, opened 2015-09-09
StaticMemoryStrategy -> HeapPoolMemoryStrategy, create StackPoolMemoryStrategy

jacquelinekay:
What was previously known as "StaticMemoryStrategy" is now "HeapPoolMemoryStrategy".

"StackPoolMemoryStrategy" is a new memory strategy that only allocates static memory.

Changes due to API breakage in forthcoming PRs to `examples` and `demos`.


dirk-thomas:
I am not sure I get the difference between the two classes. One seems to set the maximum sizes as template arguments while the other receives them as constructor arguments. But both seem to allocate the memory once and never reallocate it. Is there any difference beside them? What makes one "static" and the other "realtime"?

In general: when moving / renaming a file please do that in a separate commit since it is right now not possible to determine if the new file `rt_memory_strategy.hpp` has been modified in the process.

jacquelinekay:
StaticMemoryStrategy only allocates memory on the stack (never uses malloc, new, free, or delete). RtMemoryStrategy allocates memory on the heap. Neither strategy claims memory or releases while the executor is executing (i.e. calls to a `spin` function).  Both are trying to be "safe" for real-time. One might be more suitable based on your use case (e.g. static might be better for embedded systems).

dirk-thomas:
Since as you said both of them are usable for real-time I think the class names could be improved since they currently suggest otherwise.

What about `StaticHeapMemoryStrategy` and `StaticStackMemoryStrategy`?

jacquelinekay:
The usage of "static" originally referred to the fact that heap memory allocation is also called "dynamic" memory allocation and stack memory allocation is "static" memory allocation. I think instead I should refer to the usage of memory "pools" in both of these classes. How about HeapPoolMemoryStrategy and StackPoolMemoryStrategy?

dirk-thomas:
`HeapPoolMemoryStrategy` and `StackPoolMemoryStrategy` definitely sound better to me.

jacquelinekay:
CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/345/
http://ci.ros2.org/job/ros2_batch_ci_osx/264/
http://ci.ros2.org/job/ros2_batch_ci_windows/399/

jacquelinekay:
New CI:
http://ci.ros2.org/job/ros2_batch_ci_linux/348/
http://ci.ros2.org/job/ros2_batch_ci_osx/266/
http://ci.ros2.org/job/ros2_batch_ci_windows/403/

jacquelinekay:
Windows is fixed, moving this back to In Review.

dirk-thomas:
Can you please update the commits to first rename the file (I think `static_memory_strategy.hpp` become `heap_pool_memory_strategy.hpp`) and then perform the modifications and create the new strategy(`stack_pool_memory_strategy.hpp`).

jacquelinekay:
Done.

dirk-thomas:
Thanks, that makes the diff much easier to read.

One question based on that. I assume you have removed several comments from `heap_pool_memory_strategy.hpp` since the methods are already documented in the parent class. Should the docblocks then also be removed from the `stack_pool_memory_strategy.hpp`?

rclcpp/include/rclcpp/strategies/stack_pool_memory_strategy.hpp
```diff
@@ -0,0 +1,232 @@
...
+    if (memory_map_.count(ptr) == 0) {
+      // We expect to have the state for all blocks pre-mapped into memory_map_
+      throw std::runtime_error("Unexpected pointer in StackPoolMemoryStrategy::free.");
+    }
+
+    memset(ptr, 0, memory_map_[ptr]);
+  }
+
+private:
+  //ObjectPoolBounds bounds_;
```
dirk-thomas: This comment could be removed.

dirk-thomas:
+1

PR #112, opened 2015-09-15
Remove references to shared pointers

esteve:
This PR removes references to shared pointers.

Connects to ros2/ros2#120

### DIFF
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -78,8 +78,8 @@ public:
   virtual std::shared_ptr<void> create_request() = 0;
   virtual std::shared_ptr<void> create_request_header() = 0;
   virtual void handle_request(
-    std::shared_ptr<void> & request_header,
-    std::shared_ptr<void> & request) = 0;
+    std::shared_ptr<void> request_header,
+    std::shared_ptr<void> request) = 0;
 
 private:
   RCLCPP_DISABLE_COPY(ServiceBase);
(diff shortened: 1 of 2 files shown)


tfoote:
+1

jacquelinekay:
State of my workspace on master:
all parameter examples crash with:
```
  what():  Invalid argument
Aborted (core dumped)
```

State of my workspace with this PR:
All parameter examples work

+1

rclcpp/include/rclcpp/service.hpp
```diff
@@ -137,7 +137,7 @@ class Service : public ServiceBase
     return std::shared_ptr<void>(new rmw_request_id_t);
   }
 
-  void handle_request(std::shared_ptr<void> & request_header, std::shared_ptr<void> & request)
+  void handle_request(std::shared_ptr<void> request_header, std::shared_ptr<void> request)
```
dirk-thomas: Question unrelated to this PR: since this method is defined in the templated class why can't the signature use `shared_ptr` of the specific types (`ServiceT::Request`, `rmw_request_id_t`) instead of `void`?

dirk-thomas:
LGTM.

rclcpp/include/rclcpp/service.hpp
esteve: `handle_request` is called internally by the executor which passes `void *` versions of `ServiceT::Request` and `rmw_request_id_t`. This could be changed to strongly type them and only recast them as `void *` when needed, though.

Issue #111, opened 2015-09-14
Add tests for spin_until_future_complete


jacquelinekay:
A recent PR ros2/rclcpp/pull/107 added a return code and a timeout to this function. We should add tests for this function to test_rclcpp (and possibly for other Executor::spin functions).

PR #113, opened 2015-09-16
Pass timeout to spin_until_future_complete

jacquelinekay:
I made the API change to pass a timeout to `spin_node_until_future_complete` but forgot to pass it to `spin_until_future_complete`.

Connects to #111


esteve:
+1

PR #116, opened 2015-09-30
move vectors instantiated in wait_for_work to MemoryStrategy

jacquelinekay:
This was done in my attempt to reduce pagefaults in Executor::spin. Several vectors that were being instantiated on the stack in Executor::spin have been moved into the MemoryStrategy, so that they can be preallocated (with `vector::reserve`) on construction.

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -342,11 +342,11 @@ protected:
   void
   wait_for_work(std::chrono::duration<int64_t, T> timeout = std::chrono::duration<int64_t, T>(-1))
   {
+    memory_strategy_->subs.clear();
+    memory_strategy_->services.clear();
+    memory_strategy_->clients.clear();
     // Collect the subscriptions and timers to be waited on
     bool has_invalid_weak_nodes = false;
-    std::vector<rclcpp::subscription::SubscriptionBase::SharedPtr> subs;
-    std::vector<rclcpp::service::ServiceBase::SharedPtr> services;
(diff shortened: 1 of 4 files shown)


wjwwood:
Seems reasonable to me, but I don't have time to try it right now. I'd recommend CI before merging.

dirk-thomas:
Is this necessary for the demo at ROSCon? If not I would suggest postponing it until Monday.

jacquelinekay:
CI:
http://ci.ros2.org/job/ros2_batch_ci_linux/403/
http://ci.ros2.org/job/ros2_batch_ci_osx/300/
http://ci.ros2.org/job/ros2_batch_ci_windows/448/

Windows is already done and looks good besides the connext_dynamic tests, so that's a good sign.

esteve:
The `test_requester_replier_cpp__empty__rmw_connext_dynamic_cpp.test_requester_replier` test fails, but I don't think it's caused by this. That test seems flaky and fails with other PRs randomly.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -361,14 +361,14 @@ class Executor
         for (auto & weak_subscription : group->subscription_ptrs_) {
           auto subscription = weak_subscription.lock();
           if (subscription) {
-            subs.push_back(subscription);
+            memory_strategy_->subs.push_back(subscription);
```
esteve: I think the `MemoryStrategy` API would look clear if, instead of exposing the `std::vector` directly, provide `add_subscription`, `add_service`, etc. and hide the underlying storage.

Issue #118, opened 2015-10-09
Subscribing with a `ConstSharedPtr` in the callback signature/publishing a `ConstSharedPtr`


jacquelinekay:
Nowhere on the master branches of ROS 2 repositories do we subscribe to a topic with a callback signature receiving a `MessageT::ConstSharedPtr` (`shared_ptr<const MessageT`>, or publish a `ConstSharedPtr`.

I tried to do this and there two problems:

1. `AnySubscriptionCallback` doesn't have definitions for the `ConstSharedPtr` type. This is easy to fix and I have an upcoming pull request for it.

2. The `IntraProcessManager` receives the published message as a `void*` in its `store_intra_process` callback. The compile error is because the publisher template tries to copy the const message into a unique pointer of a const message and pass that to `store_intra_proces`, but `const` message cannot be cast to void. Fixing this requires a bit more work to ensure the copy of the message is not const.

jacquelinekay:
TODO for me: investigate if 2 is still a problem (maybe add a test for publishing a ConstSharedPtr)

PR #119, opened 2015-10-09
Add any subscription callback types for shared_ptr<const T>

jacquelinekay:
I noticed this was missing in #118. This should enable us to set up a subscription callback with `ConstSharedPtr`.

### DIFF
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -32,20 +32,26 @@ namespace any_subscription_callback
 template<typename MessageT>
 struct AnySubscriptionCallback
 {
-  using SharedPtrCallback = std::function<void(const std::shared_ptr<MessageT> &)>;
+  using SharedPtrCallback = std::function<void(const std::shared_ptr<MessageT>)>;
   using SharedPtrWithInfoCallback =
-      std::function<void(const std::shared_ptr<MessageT> &, const rmw_message_info_t &)>;
-  using UniquePtrCallback = std::function<void(std::unique_ptr<MessageT> &)>;
+      std::function<void(const std::shared_ptr<MessageT>, const rmw_message_info_t &)>;
+  using ConstSharedPtrCallback = std::function<void(const std::shared_ptr<const MessageT>)>;
+  using ConstSharedPtrWithInfoCallback =
+      std::function<void(const std::shared_ptr<const MessageT>, const rmw_message_info_t &)>;
+  using UniquePtrCallback = std::function<void(std::unique_ptr<MessageT>)>;
   using UniquePtrWithInfoCallback =
-      std::function<void(std::unique_ptr<MessageT> &, const rmw_message_info_t &)>;
+      std::function<void(std::unique_ptr<MessageT>, const rmw_message_info_t &)>;
 
   SharedPtrCallback shared_ptr_callback;
   SharedPtrWithInfoCallback shared_ptr_with_info_callback;
+  ConstSharedPtrCallback const_shared_ptr_callback;
+  ConstSharedPtrWithInfoCallback const_shared_ptr_with_info_callback;
   UniquePtrCallback unique_ptr_callback;
   UniquePtrWithInfoCallback unique_ptr_with_info_callback;
 
   AnySubscriptionCallback()
   : shared_ptr_callback(nullptr), shared_ptr_with_info_callback(nullptr),
+    const_shared_ptr_callback(nullptr), const_shared_ptr_with_info_callback(nullptr),
     unique_ptr_callback(nullptr), unique_ptr_with_info_callback(nullptr)
   {}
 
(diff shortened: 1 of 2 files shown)


rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -35,17 +35,23 @@ struct AnySubscriptionCallback
   using SharedPtrCallback = std::function<void(const std::shared_ptr<MessageT> &)>;
   using SharedPtrWithInfoCallback =
       std::function<void(const std::shared_ptr<MessageT> &, const rmw_message_info_t &)>;
+  using ConstSharedPtrCallback = std::function<void(const std::shared_ptr<const MessageT> &)>;
```
dirk-thomas: The signature should not use a reference to a shared_ptr. The some for all the other signatures not touched in this PR.

wjwwood:
+1 with CI

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -82,6 +88,38 @@ struct AnySubscriptionCallback
...
+  >::type * = nullptr,
+  typename std::enable_if<
+    std::is_same<
+      typename function_traits<CallbackT>::template argument_type<0>,
+      typename std::shared_ptr<const MessageT>
+    >::value
+  >::type * = nullptr
+  >
+  void set(CallbackT callback)
+  {
```
esteve: Could you add a `static_assert` here to check for the second argument of the callback? It'd be better to add it as part of the `std::enable_if`, but... Windows.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacquelinekay: I will remove references to shared_ptrs but not to unique_ptrs as this seems to mess up how unique_ptrs are being copied for intra-process.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacquelinekay: As you can see, I blindly copied this from the specializations of `set` for `shared_ptr`, so I don't really know what's going on. Is having multiple specializations of `set` like this not going to compile on Windows because of the greedy template resolution in VS2015? Should I move the arity check from the std::enable if and into the body of the function with `static_assert`?

jacquelinekay:
CI, let's see if Windows builds:
http://ci.ros2.org/job/ros2_batch_ci_linux/411
http://ci.ros2.org/job/ros2_batch_ci_osx/308
http://ci.ros2.org/job/ros2_batch_ci_windows/457

rclcpp/include/rclcpp/any_subscription_callback.hpp
dirk-thomas: But keeping the reference for unique pointers implies that the caller as well as the callee have a shared handle. If the caller lets the variable go out of scope while the callee is still holding the reference any access will access invalid memory.

rclcpp/include/rclcpp/any_subscription_callback.hpp
esteve: It'll work on Windows as is. The only thing is that the type of the second argument of the callback is not being checked. Unfortunately, we can't use `std::is_same...` for the second argument because of the lack of expression SFINAE in VS2015, so instead we can use `static_assert` in the body of the function to check the type.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacquelinekay: indeed. Well, I'll do some refactoring to see if I can make the correct callback signature work in rclcpp in this branch.

rclcpp/include/rclcpp/any_subscription_callback.hpp
esteve: This is what we current do for service callbacks:

https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node.hpp#L322

I'm thinking that maybe it'd make sense to move the `is_same` check from the `enable_if` to a `static_assert` in our code, since we only care about the arity of the callbacks. It'd also make our code more uniform.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacquelinekay: ok, that wasn't as bad as I thought, I just changed a few places where the `unique_ptr` copy constructor was being used. I'll restart the CI jobs.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacquelinekay: Still confused about this. I think we do care about the type as well as the arity of the callback here. If Windows were only looking at the arity and not the type, then I would expect any code that exercises the new callback type to fail, because the subscriber would end up trying to call the `shared_ptr` callback instead of the `const_shared_ptr` callback...

In fact, I'm added a test for this since the new callback type to make sure:

ros2/system_tests#53

http://ci.ros2.org/job/ros2_batch_ci_windows/460/

esteve: That's true, I didn't the other version of set, my bad. Still, it'd be good to have a `static_assert` for the second argument, since we can't add the check in `enable_if` because of Windows.

jacquelinekay: The new test passes: http://ci.ros2.org/job/ros2_batch_ci_windows/460/testReport/%28root%29/test_subscription__rmw_connext_cpp/subscription_shared_ptr_const/

I'm still not convinced that we need a static assert, maybe we should chat about it offline?

jacquelinekay: alright, static asserts have been added. I will rerun CI just to make sure.

jacquelinekay:
Latest CI:
http://ci.ros2.org/job/ros2_batch_ci_linux/416/
http://ci.ros2.org/job/ros2_batch_ci_osx/312/
http://ci.ros2.org/job/ros2_batch_ci_windows/460/

William said +1 with CI, and the PR making the changes in `demos` has two +1s, so I'll merge this soon if there are no more objections.

esteve:
@jacquelinekay I've solved/worked around the issues on Windows with callbacks that have variable number of arguments in https://github.com/ros2/rclcpp/pull/122

The changes in that PR may conflict with yours, let me know if you'd rather merge your PR first and I can fix the conflicts in mine.

PR #120, opened 2015-10-09
Use variadic macros

jacquelinekay:
While exposing the allocator template in rclcpp, I noticed our macro definitions have trouble when there are commas in a type (i.e. when there are multiple template arguments). I decided to change all the macro definitions to variadic macros, which will allow commas in the type names.

### DIFF
--- a/rclcpp/include/rclcpp/macros.hpp
+++ b/rclcpp/include/rclcpp/macros.hpp
@@ -19,19 +19,19 @@
  *
  * Use in the private section of the class.
  */
-#define RCLCPP_DISABLE_COPY(Class) \
-  Class(const Class &) = delete; \
-  Class & operator=(const Class &) = delete;
+#define RCLCPP_DISABLE_COPY(...) \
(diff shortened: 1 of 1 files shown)


esteve:
+1

rclcpp/include/rclcpp/macros.hpp
```diff
@@ -19,19 +19,19 @@
  *
  * Use in the private section of the class.
  */
-#define RCLCPP_DISABLE_COPY(Class) \
-  Class(const Class &) = delete; \
-  Class & operator=(const Class &) = delete;
+#define RCLCPP_DISABLE_COPY(...) \
+  __VA_ARGS__(const __VA_ARGS__ &) = delete; \
```
dirk-thomas: This looks weird to me. The template accepts multiple arguments now (e.g. `Foo, Bar`) which then ends up as `const Foo, Bar &`?

jacquelinekay: yes. I wasn't sure how else to allow passing a class with multiple template arguments to these macros.

PR #123, opened 2015-10-13
add publish specialization for shared_ptr<const MessageT>

### LINKED ISSUE #118: Subscribing with a `ConstSharedPtr` in the callback signature/publishing a `ConstSharedPtr`

jacquelinekay:
Fixes #118 

CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/431/
http://ci.ros2.org/job/ros2_batch_ci_osx/341/
http://ci.ros2.org/job/ros2_batch_ci_windows/483/

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -159,6 +159,24 @@ public:
 
   template<typename MessageT>
   void
+  publish(std::shared_ptr<const MessageT> msg)
+  {
+    // Avoid allocating when not using intra process.
+    if (!store_intra_process_message_) {
+      // In this case we're not using intra process.
+      return this->do_inter_process_publish(msg.get());
+    }
+    // Otherwise we have to allocate memory in a unique_ptr and pass it along.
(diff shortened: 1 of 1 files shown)


wjwwood:
lgtm

dirk-thomas:
+1

esteve:
+1

PR #128, opened 2015-10-14
add version of create_subscription which just takes queue size and not a qos profile

wjwwood:
Connects to ros2/design#56.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -117,6 +117,31 @@ public:
   /// Create and return a Subscription.
   /**
    * \param[in] topic_name The topic to subscribe on.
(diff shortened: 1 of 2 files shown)


wjwwood:
CI:
- http://ci.ros2.org/job/ros2_batch_ci_linux/437/
- http://ci.ros2.org/job/ros2_batch_ci_osx/346/
- http://ci.ros2.org/job/ros2_batch_ci_windows/488/

wjwwood:
CI is good, up for review.

wjwwood:
Also up for review is this one: https://github.com/ros2/system_tests/pull/56

dirk-thomas:
+1

In the future I think we should have a one set of method (which is as small as possible) providing the core functionality. And beside that a separate set of methods which are only syntactic sugar and reusing the core functionality internally but expose a different interface.

PR #129, opened 2015-10-14
Avoid redundant allocation in wait_for_work

jacquelinekay:
Connects to #5

### DIFF
--- a/rclcpp/include/rclcpp/memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/memory_strategy.hpp
@@ -46,7 +46,8 @@ public:
 
   /// Borrow memory for storing data for subscriptions, services, clients, or guard conditions.
   /**
-   * The default implementation ignores the handle type and dynamically allocates the memory.
+   * The default implementation stores std::vectors for each handle type and resizes the vectors
+   * as necessary based on the requested number of handles.
    * \param[in] The type of entity that this function is requesting for.
    * \param[in] The number of handles to borrow.
    * \return Pointer to the allocated handles.
--- a/rclcpp/include/rclcpp/memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/memory_strategy.hpp
@@ -53,8 +53,31 @@ public:
    */
   virtual void ** borrow_handles(HandleType type, size_t number_of_handles)
   {
-    (void)type;
-    return static_cast<void **>(alloc(sizeof(void *) * number_of_handles));
+    switch (type) {
+      case HandleType::subscription_handle:
+        if (subscription_handles.size() < number_of_handles) {
+          subscription_handles.resize(number_of_handles, 0);
+        }
+        return static_cast<void **>(subscription_handles.data());
+      case HandleType::service_handle:
+        if (service_handles.size() < number_of_handles) {
+          service_handles.resize(number_of_handles, 0);
+        }
+        return static_cast<void **>(service_handles.data());
+      case HandleType::client_handle:
+        if (client_handles.size() < number_of_handles) {
+          client_handles.resize(number_of_handles, 0);
+        }
+        return static_cast<void **>(client_handles.data());
+      case HandleType::guard_condition_handle:
+        if (number_of_handles > 2) {
+          throw std::runtime_error("Too many guard condition handles requested!");
+        }
+        return guard_cond_handles.data();
+      default:
+        throw std::runtime_error("Unknown HandleType " + std::to_string(static_cast<int>(type)) +
+                ", could not borrow handle memory.");
+    }
   }
 
   /// Return the memory borrowed in borrow_handles.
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -65,8 +90,22 @@ class MemoryStrategy
...
+        break;
+      case HandleType::client_handle:
+        memset(handles, 0, client_handles.size());
+        break;
+      case HandleType::guard_condition_handle:
+        guard_cond_handles.fill(0);
+        break;
+      default:
+        throw std::runtime_error("Unrecognized enum, could not borrow handle memory.");
+    }
```
wjwwood: Is it worth checking that the value in handles matches the address of the pointer returned by `.data()` of the corresponding vector?

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -53,8 +53,33 @@ class MemoryStrategy
...
   {
-    (void)type;
-    return static_cast<void **>(alloc(sizeof(void *) * number_of_handles));
+    switch (type) {
+      case HandleType::subscription_handle:
+        if (subscription_handles.size() < number_of_handles) {
+          subscription_handles.resize(number_of_handles, 0);
+        }
+        return static_cast<void **>(subscription_handles.data());
+        break;
```
dirk-thomas: Unreachable.

Same below.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -53,8 +53,33 @@ class MemoryStrategy
...
+        }
+        return static_cast<void **>(client_handles.data());
+        break;
+      case HandleType::guard_condition_handle:
+        if (number_of_handles > 2) {
+          throw std::runtime_error("Too many guard condition handles requested!");
+        }
+        return guard_cond_handles.data();
+      default:
+        throw std::runtime_error("Unrecognized enum, could not borrow handle memory.");
```
wjwwood: Rather than `Unrecognized enum` what about `Unknown HandleType`?

dirk-thomas: The message could also include the actual value which would help debugging the problem if it happens.

wjwwood:
Other than minor comments, lgtm.

rclcpp/include/rclcpp/memory_strategy.hpp
jacquelinekay: Since HandleType is an enum, how would you suggest I print its representation? As an integer?

dirk-thomas: Yes.

jacquelinekay:
I started CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/439/
http://ci.ros2.org/job/ros2_batch_ci_osx/348/
http://ci.ros2.org/job/ros2_batch_ci_windows/491/

esteve:
+1

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -53,8 +53,31 @@ class MemoryStrategy
    */
   virtual void ** borrow_handles(HandleType type, size_t number_of_handles)
   {
-    (void)type;
-    return static_cast<void **>(alloc(sizeof(void *) * number_of_handles));
+    switch (type) {
```
tfoote: The doc block above looks out of date with this change.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -101,6 +155,11 @@ class MemoryStrategy
   std::vector<rclcpp::subscription::SubscriptionBase::SharedPtr> subs;
   std::vector<rclcpp::service::ServiceBase::SharedPtr> services;
   std::vector<rclcpp::client::ClientBase::SharedPtr> clients;
+
+  std::vector<void *> subscription_handles;
```
tfoote: Is there a max number of handles supported which could be reserved in the constructor? Using resize may or may not allocate unless it's been previously allocated.

wjwwood: For this memory strategy it doesn't matter if it allocates on `.resize()`.

jacquelinekay: I think it's fine for the vectors to be initialized to empty. Specifying a number of handles reserve beforehand seems unnecessarily complicated for the default case. The alternative MemoryStrategy implementations take max numbers of handles as their arguments, however.

When `borrow_handles` is called, if the vector size is smaller than the requested size `n`, then the vector will call `resize(n)`.

http://en.cppreference.com/w/cpp/container/vector/resize

My interpretation is that calling resize for a size greater than the vector's current size will insert elements into the vector to make it grow to the required size.

tfoote: Yes resize is safe it will allocate if not enough is reserved. The title was about reducing allocations I didn't know if we wanted to do the preallocation. If this is just the default it will work fine.

rclcpp/include/rclcpp/memory_strategy.hpp
jacquelinekay: I've updated the doc block now.

rclcpp/include/rclcpp/memory_strategy.hpp
jacquelinekay: @dirk-thomas clarified that if the vector's capacity is smaller than the requested size, `std::vector.resize()` will allocate a block of new entries, often to the next power of two. So resize won't necessarily allocate every time it's called, but it may incur a large penalty when the capacity needs to grow.

But for our purposes this is fine since these vectors will usually store small numbers (<16).

PR #133, opened 2015-10-19
Made rmw_qos_profile argument optional

esteve:
Moved `rmw_qos_profile` argument around to make it optional.

Connects to ros2/ros2#80

### DIFF
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -209,8 +209,8 @@ template<typename MessageT, typename CallbackT>
 typename rclcpp::subscription::Subscription<MessageT>::SharedPtr
 Node::create_subscription(
   const std::string & topic_name,
-  const rmw_qos_profile_t & qos_profile,
   CallbackT callback,
+  const rmw_qos_profile_t & qos_profile,
   rclcpp::callback_group::CallbackGroup::SharedPtr group,
   bool ignore_local_publications,
   typename rclcpp::message_memory_strategy::MessageMemoryStrategy<MessageT>::SharedPtr
@@ -321,8 +321,8 @@ Node::create_subscription(
   qos.depth = qos_history_depth;
   return this->create_subscription<MessageT, CallbackT>(
     topic_name,
-    qos,
     callback,
+    qos,
     group,
     ignore_local_publications,
     msg_mem_strat);
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -112,13 +112,14 @@ public:
   template<typename MessageT>
   typename rclcpp::publisher::Publisher<MessageT>::SharedPtr
   create_publisher(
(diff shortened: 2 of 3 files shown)


dirk-thomas:
Why is the order of the topic name and callback changes too?

esteve:
I tried to make it consistent so that the `rmw_qos_profile` argument always comes after the topic name, but I don't really have a strong opinion on where the callback argument should go.

wjwwood:
In order to keep in line with the order from ROS 1 (topic, queue size, callback), I think at least we should keep callback after the topic. At most I'm not sure we want to allow users to create publishers and subscriptions without specifying at least queue size. Maybe we do, but I think we should have some rational as to why or why not.

tfoote:
Since that API is a convenience for similar backwards compatibility it would be good to keep it consistent with the old signature.

wjwwood:
> Since that API is a convenience for similar backwards compatibility it would be good to keep it consistent with the old signature.

How is that a convenience for backwards compatibility? There is no similar signature in ROS 1 that I'm aware of.

esteve:
I'm reordering the arguments so that the callback comes after the topic name.

tfoote:
create_subscription with (topic, queue size, callback) instead of the new qos_profile I consider a convenience method for backwards compatibility. We only expose the one qos dimention previously in ROS1. It's the analog to [NodeHandle::subscribe](http://docs.ros.org/api/roscpp/html/classros_1_1NodeHandle.html#aad44a59900a86f6d24f175b28e20c8a3)

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -309,9 +309,9 @@ Node::create_subscription(
 template<typename MessageT, typename CallbackT>
 typename rclcpp::subscription::Subscription<MessageT>::SharedPtr
 Node::create_subscription(
+  CallbackT callback,
   const std::string & topic_name,
   size_t qos_history_depth,
```
tfoote: @wjwwood this is the one i'm talking about

wjwwood:
@tfoote is talking about this one: https://github.com/ros2/rclcpp/pull/133/files#diff-809285dba1df7726afedd63625b6a6ceR311

Which I agree should end up as (topic, queue size, callback) and not (topic, callback, queue size).

esteve:
I've changed it so that the callback comes after the topic name for the signature of `create_signature` where the `rmw_qos_profile` is optional. And for the other signature, I moved it after the queue size.

esteve:
CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_windows/504/
http://ci.ros2.org/job/ros2_batch_ci_linux/447/
http://ci.ros2.org/job/ros2_batch_ci_osx/359/

wjwwood:
lgtm

esteve:
Since this is just a change in the `create_subscription` signature and the Linux and OSX jobs pass, I've cancelled the Windows CI job.

PR #137, opened 2015-10-22
Allocator template for publish/subscribe pipeline

jacquelinekay:
This pull request exposes template arguments to Publisher, Subscription, AnySubscriptionCallback, and MappedRingBuffer so that a custom allocator can be passed to these entities.

- [x] Memory leaks in classes with Allocators. Should change raw pointer to shared pointer to handle ownership.
- [x] Style cleanup to make templated names less scary-looking. (I would appreciate more feedback on this; I'm considering writing macros since I repeat similar type aliases in several files.)
- [x] Pass allocator to shared_ptrs instantiated during runtime so that their copies and control blocks use the custom allocator.
- [x] Figure out how to pass allocator to shared pointers and other stdlib structures in: CallbackGroup, Executor, IntraProcessManager (add allocator to Context?)

After this PR is merged, both the intra-process and inter-process pipelines should be set up for real-time safe allocation.


jacquelinekay:
Contrary to my expectations, the branch in its current state built on the Windows Jenkins job!

http://ci.ros2.org/job/ros2_batch_ci_windows/518/

wjwwood:
Well that's exciting.

jacquelinekay:
I tried to implement a change to Executor where I got rid of MemoryStrategy and replaced with ExecutorRuntimeState, a templated class that keeps track of all the stdlib structures that dynamically change size during the execution path, and inherits from a pure virtual non-templated base class. I ripped out a lot of the functionality of Executor (basically anything that interfaces with the lists and vectors in Executor) and put it in ExecutorRuntimeState. But then I realized it wouldn't compile because a lot of the other entities in rclcpp declare themselves as `friend` of Executor, and they can't declare themselves as `friend` of ExecutorRuntimeState<T> without knowing about T (I think).

Back to the drawing board... I'm going to look at how to template the IntraProcessManager next.

wjwwood:
We may be able to get rid of the friend relationships. Don't be afraid to suggest changes in the relationships between objects.

jacquelinekay:
I now consider this ready for review.

CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/488/
http://ci.ros2.org/job/ros2_batch_ci_osx/387/
http://ci.ros2.org/job/ros2_batch_ci_windows/538/

dirk-thomas:
New code should pass cpplint.

Btw. the jobs are failing to compile.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <memory>
+
+#include <rclcpp/allocator/allocator_deleter.hpp>
+
+namespace rclcpp
+{
+namespace allocator
+{
+
+template<typename U, typename Alloc>
```
dirk-thomas: What does the `U` stand for? If nothing specific I would recommend to use the common variable `T` for template typenames.

Same below.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -28,6 +28,7 @@
 #include <rclcpp/any_executable.hpp>
 #include <rclcpp/macros.hpp>
 #include <rclcpp/memory_strategy.hpp>
+#include <rclcpp/memory_strategies.hpp>
```
dirk-thomas: Wrong include order.

rclcpp/include/rclcpp/allocator/allocator_deleter.hpp
```diff
@@ -0,0 +1,106 @@
...
+    allocator_ = a.get_allocator();
+  }
+
+  template<typename U>
+  void operator()(U * ptr)
+  {
+    std::allocator_traits<AllocRebind<U>>::destroy(*allocator_, ptr);
+    // TODO: figure out if we're guaranteed to be destroying only 1 item here
+    std::allocator_traits<AllocRebind<U>>::deallocate(*allocator_, ptr, 1);
+    ptr = NULL;
```
wjwwood: I'd use `nullptr` here.

rclcpp/include/rclcpp/intra_process_manager.hpp
```diff
@@ -122,7 +122,11 @@ class IntraProcessManager
 public:
   RCLCPP_SMART_PTR_DEFINITIONS(IntraProcessManager);
 
-  IntraProcessManager() = default;
+  //IntraProcessManager() = default;
```
wjwwood: Should this be removed or uncommented?

rclcpp/include/rclcpp/intra_process_manager_state.hpp
```diff
@@ -0,0 +1,277 @@
...
+#include <map>
+#include <unordered_map>
+#include <set>
+
+namespace rclcpp
+{
+namespace intra_process_manager
+{
+
+class IntraProcessManagerStateBase
```
wjwwood: We talked about this off-line, but the name off this class isn't as descriptive as I'd like. I don't have a strong candidate for the alternative name, but maybe `IntraProcessManagerDataBase` or `IntraProcessManagerDetailBase`?

rclcpp/include/rclcpp/intra_process_manager_state.hpp
```diff
@@ -0,0 +1,277 @@
...
+{
+public:
+  //RCLCPP_SMART_PTR_DEFINITIONS(IntraProcessManagerStateBase);
```
wjwwood: Should this be removed?

rclcpp/include/rclcpp/intra_process_manager_state.hpp
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  void
+  remove_publisher(uint64_t intra_process_publisher_id)
+  {
+    publishers_.erase(intra_process_publisher_id);
+  }
+
+  // TODO
+  // return message_seq and mrb
```
wjwwood: Should this TODO be addressed before merging? It's not clear to me what needs to be done here and if it has been done yet or not.

rclcpp/include/rclcpp/intra_process_manager_state.hpp
```diff
@@ -0,0 +1,277 @@
...
+      }
+    }
+    return false;
+  }
+
+private:
+  template<typename T>
+  using RebindAlloc = typename std::allocator_traits<Allocator>::template rebind_alloc<T>;
+
+  //using AllocString = std::basic_string<char, std::char_traits<char>, RebindAlloc<char>>;
```
wjwwood: Should this be removed or uncommented?

rclcpp/include/rclcpp/memory_strategies.hpp
```diff
@@ -15,16 +15,24 @@
...
 namespace rclcpp
 {
 namespace memory_strategies
 {
 
-using rclcpp::memory_strategies::heap_pool_memory_strategy::HeapPoolMemoryStrategy;
-using rclcpp::memory_strategies::stack_pool_memory_strategy::StackPoolMemoryStrategy;
+using rclcpp::memory_strategies::allocator_memory_strategy::AllocatorMemoryStrategy;
+//using rclcpp::memory_strategies::heap_pool_memory_strategy::HeapPoolMemoryStrategy;
+//using rclcpp::memory_strategies::stack_pool_memory_strategy::StackPoolMemoryStrategy;
```
wjwwood: It looks like you later delete the files, so you might as well remove the `using` and `#include` statements.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -38,137 +36,171 @@ namespace memory_strategy
...
+  // return the new number of clients
+  virtual size_t fill_client_handles(void ** & ptr) = 0;
+
+  virtual void clear_active_entities() = 0;
+
+  virtual void clear_handles() = 0;
+  virtual bool collect_entities(const WeakNodeVector & weak_nodes) = 0;
+
+  /// Provide a newly initialized AnyExecutable object.
+  // \return Shared pointer to the fresh executable.
```
wjwwood: If you want doxygen to pick this up I think you need the `\return` to be within a `/* */`. But I'm not 100% sure.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -38,137 +36,171 @@ namespace memory_strategy
...
-  /// Return the memory borrowed in borrow_handles.
-  /**
-   * return_handles should always mirror the way memory was borrowed in borrow_handles.
-   * \param[in] The type of entity that this function is returning.
-   * \param[in] Pointer to the handles returned.
-   */
-  virtual void return_handles(HandleType type, void ** handles)
+  static rclcpp::service::ServiceBase::SharedPtr
+  get_service_by_handle(void * service_handle,
+    const WeakNodeVector & weak_nodes)
```
wjwwood: Looks like all of the parameters could be on the same line without going over 100 characters.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -38,137 +36,171 @@ namespace memory_strategy
...
+          }
         }
-        guard_cond_handles.fill(0);
-        break;
-      default:
-        throw std::runtime_error("Unknown HandleType " + std::to_string(static_cast<int>(type)) +
-                ", could not borrow handle memory.");
+      }
     }
+    return rclcpp::service::ServiceBase::SharedPtr();
```
wjwwood: I think this can simply be `return nullptr` (even if we were not doing that before). I think that also better conveys the intention.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -38,137 +36,171 @@ namespace memory_strategy
...
   }
 
-  /// Provide a newly initialized AnyExecutable object.
-  // \return Shared pointer to the fresh executable.
-  virtual executor::AnyExecutable::SharedPtr instantiate_next_executable()
+  static rclcpp::client::ClientBase::SharedPtr
+  get_client_by_handle(void * client_handle,
+    const WeakNodeVector & weak_nodes)
```
wjwwood: Same unnecessary multi-line declaration as above.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -38,137 +36,171 @@ namespace memory_strategy
...
+          continue;
+        }
+        for (auto & client : group->get_client_ptrs()) {
+          if (client->get_client_handle()->data == client_handle) {
+            return client;
+          }
+        }
+      }
+    }
+    return rclcpp::client::ClientBase::SharedPtr();
```
wjwwood: The same comment applies here as above about returning a `nullptr` instead. I think there are more cases below too.

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -0,0 +1,359 @@
...
+    for (auto & subscription : subscriptions_) {
+      subscriber_handles_[count++] = subscription->get_subscription_handle()->data;
+      if (subscription->get_intra_process_subscription_handle()) {
+        subscriber_handles_[count++] =
+          subscription->get_intra_process_subscription_handle()->data;
+      }
+    }
+    ptr = subscriber_handles_.data();
+    return count;
+  }
```
wjwwood: We also talked about this offline, but I think these functions sort of don't belong in a memory strategy class. I agree that they probably don't belong in the Executor class either, so maybe they need another container class which is responsible for handling data within it. The executor can make requests of the container and the container can delegate allocation to the memory strategy or an allocator where appropriate.

I don't think we have to solve this design problem in this pull request, but we should consider doing this refactoring soon. Perhaps I'll do it in the second part of rclcpp library work, after the alpha 2 release.

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -0,0 +1,359 @@
...
+        } else {
+          any_exec->subscription = subscription;
+        }
+        any_exec->callback_group = group;
+        any_exec->node = get_node_by_group(group, weak_nodes);
+        to_erase.push_back(it++);
+        return;
+      }
+      // Else, the subscription is no longer valid, remove it and continue
+      // TODO
```
wjwwood: Does this still need to be resolved?

jacquelinekay: this is on an outdated diff, the TODO is gone in the latest version

jacquelinekay:
Another batch of CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/491/
http://ci.ros2.org/job/ros2_batch_ci_osx/388/
http://ci.ros2.org/job/ros2_batch_ci_windows/540/

wjwwood:
Code review looks ok to me.

It was a bit confusing at first as to why some of the changes were being made, but after talking with @jacquelinekay I think they're all necessary. The recurring pattern in use is that you have a user facing class that shouldn't be templated, but does need to be configurable with an allocator and that requires a template argument. Therefore you have the pattern where the user facing class is not templated, but instead takes a templated parameter which is an implementation class for the class being constructed. This implementation class has a non-template base class which the user facing class stores the pointer as, and a templated child class which is templated on the allocator.

This pattern can be observed between the `IntraProcessManager` class and the `IntraProcessManagerStateBase` and `IntraProcessManagerState<template Allocator>` classes.

With CI and some comments addressed I think I'd give this a +1.

jacquelinekay:
Trying to fix the OSX and Windows builds: I changed the `get_next_subscription/client/service` functions to use an erase/remove-if idiom on the void* handles vectors, where the remove if lambda function has side effects on the AnyExecutable object passed to the `get_next...` function.

https://github.com/ros2/rclcpp/commit/2b95066ead4e728ad9bf181d2fa7fae2e5c659d2

http://ci.ros2.org/job/ros2_batch_ci_osx/390/

rclcpp/include/rclcpp/intra_process_manager.hpp
```diff
@@ -321,48 +301,31 @@ class IntraProcessManager
...
-      auto it = std::find(
-        target_subs->begin(), target_subs->end(),
-        requesting_subscriptions_intra_process_id);
-      if (it == target_subs->end()) {
-        // This publisher id/message seq pair was not intended for this subscription.
-        return;
-      }
-      target_subs->erase(it);
+    using MRBMessageAlloc = typename std::allocator_traits<Alloc>::template rebind_alloc<MessageT>;
+    typedef typename mapped_ring_buffer::MappedRingBuffer<MessageT, MRBMessageAlloc> TypedMRB;
```
esteve: I'd change this to use `using`

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -244,7 +251,7 @@ class Subscription : public SubscriptionBase
 private:
   typedef
```
esteve: `using` instead of `typedef`

jacquelinekay:
Another set of CI jobs.

I'm going to wait on a clean set of CI before squashing, just in case:

http://ci.ros2.org/job/ros2_batch_ci_linux/506 (The linux CI job appears to be broken)
http://ci.ros2.org/job/ros2_batch_ci_osx/392/
http://ci.ros2.org/job/ros2_batch_ci_windows/543/

dirk-thomas:
Why do you think the Linux CI job is broken?

jacquelinekay:
there are two jobs pending that haven't started yet, and they've both been pending for a few minutes.

dirk-thomas:
This job is currently running on the Linux slave: http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/

rclcpp/include/rclcpp/subscription.hpp
jacquelinekay: this is unrelated to this PR, so I'm not going to take care of it for now

jacquelinekay:
and again:

http://ci.ros2.org/job/ros2_batch_ci_linux/513
http://ci.ros2.org/job/ros2_batch_ci_osx/393/
http://ci.ros2.org/job/ros2_batch_ci_windows/544/

jacquelinekay:
There are a lot of regressions. But many of them exist in the nightly build:

http://ci.ros2.org/view/nightly/job/ros2_batch_ci_osx_nightly/94/testReport/
vs.
http://ci.ros2.org/job/ros2_batch_ci_osx/393/testReport/

On Linux, the allocator test doesn't work, probably because the backtrace trick only works when the code is compiled in Debug mode. I will change the test so that it fails if the number of allocations during execution per call to "spin" exceeds a set number, which is not as great.

I'm not sure where many of the regressions in Windows are coming from: the nightly build has 86 failures and the CI for this branch has 214 failures. I am rerunning Windows CI on default just to compare.

dirk-thomas:
For some of the failing tests their are PRs which are only waiting for reviews...

esteve:
+1 I just ran this on Windows and the tests pass

PR #138, opened 2015-10-26
Remove declaration of friendship for Executor

jacquelinekay:
This is to make adding allocator templates to stdlib structures in the executor easier.

### DIFF
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -32,10 +32,6 @@ namespace node
 {
 class Node;
 } // namespace node
-namespace executor
-{
-class Executor;
-} // namespace executor
 
 namespace callback_group
 {
@@ -49,7 +45,6 @@ enum class CallbackGroupType
 class CallbackGroup
 {
   friend class rclcpp::node::Node;
-  friend class rclcpp::executor::Executor;
 
 public:
   RCLCPP_SMART_PTR_DEFINITIONS(CallbackGroup);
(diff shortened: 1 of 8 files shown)


rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -58,6 +53,38 @@ class CallbackGroup
   : type_(group_type), can_be_taken_from_(true)
   {}
 
+  std::vector<subscription::SubscriptionBase::WeakPtr> &
+  get_subscription_ptrs()
```
dirk-thomas: Can these methods be `const` (and return a const reference)?

Same below.

wjwwood:
That looks like a clean change. The reason I originally did the friend pattern was that I felt the extra functions crowded the user facing API, but I think the result from the pr is still pretty clean. I may try to continue to hide these interfaces when I move `rclcpp` to a library, but in a way where the interfaces are still public (no need for friend) but they are somehow separate from the more normal user functions like `publish`, `get_topic_name`, `get_type`, etc. This also has to be done (in some related fashion) for `Node` being a friend of several classes.

But for now, this gets a +1 from me.

jacquelinekay:
Jenkins after fixing style and an issue with timers that was causing test failures:

http://ci.ros2.org/job/ros2_batch_ci_linux/476/
http://ci.ros2.org/job/ros2_batch_ci_osx/378/
http://ci.ros2.org/job/ros2_batch_ci_windows/527/

esteve:
+1

PR #139, opened 2015-10-28
Replace unsigned long with uint32_t

esteve:
Replace unsigned long with uint32_t

Connects to ament/ament_lint#34


tfoote:
+1

wjwwood:
It would be good if we could get at least this pr merged so I could work on the rclcpp library pr. @esteve are we waiting on CI for these pr's or something else?

Issue #48, opened 2015-06-24
refactor into library


dirk-thomas:
Currently all functions are defined in the header files directly. This makes it impossible to create executables from multiple C++ files since each will include its own version of the symbol and then collide at link time.

wjwwood:
To be changed: https://github.com/ros2/ros1_bridge/commit/b41b472b7b4f18257196e16de342ccc4e8f870ba#diff-d1066de058a30ec5a5edd4febc16451cR23

dirk-thomas:
Also to be changed: https://github.com/ros2/ros1_bridge/commit/587ec6ede433a36ef159b2d4a95253082d9e3aee#diff-af3b638bc2a3e6c650974192a53c7291R98

wjwwood:
I'm going to work on this, since my prototyping is now stalled on being able to build things into separate object files and then combine them later into a library or executable.

wjwwood:
Ok, I've been iterating on things in a simplified example in https://github.com/wjwwood/weak_linking.git.

I have one pr (https://github.com/wjwwood/weak_linking/pull/3) which is (as near as I can tell) functionally the same as what we have now with the implementation moved into a cpp file from the header. It has Linux passing but OS X and Windows failing.

The crux of the issue is that the cpp files in `rclcpp` will use symbols declared in the headers in `rmw`, which are defined in an rmw implementation, but the implementation is neither available when building `librclcpp` nor could we pick which one to use if they were. On Linux this is ok, because the default behavior when linking is to ignore missing symbols. This means that the `librclcpp` library is incomplete, because it references symbols which have not been resolved yet and instead will try to at runtime. The default on OS X is to require all symbols be resolved at link time by way of providing the libraries which define them as link arguments. So by default this fails with OS X, however, this can be changed.

The next pr (https://github.com/wjwwood/weak_linking/pull/2) tells the OS X linker to handle undefined symbols by looking them up dynamically at runtime. This allows Linux and OS X to pass, but Windows is still lagging behind.

I've tried a lot of things to get this to work on Windows. I'll try to summarize them:

- Use the `/FORCE` option for `link.exe`.
 - This allows `librclcpp` to link with "warnings" about the missing symbols and for the user's executables to compile and link.
 - When the user executable is run though I get an access violation when trying to call a function from the rmw implementation library.
 - Looking at the executable with [Dependency Walker](http://www.dependencywalker.com/) it is linked against both `librclcpp` and `librmw_impl{1 or 2}`.
[...]

PR #140, opened 2015-10-29
Make rclcpp in to a library

wjwwood:
Connects to ros2/rclcpp#48

This pr currently builds on https://github.com/ros2/rclcpp/pull/137, but I'll rebase it against master once that is merged. In the mean time you can review changes specific to this pr with this link: https://github.com/ros2/rclcpp/compare/allocator_template...rclcpp_library

I have not reenabled the intra process manager tests because it was using a somewhat peculiar method of mocking the classes within rclcpp and I'll have to come up with an alternative approach.

On Windows this relies on the unresolved library symbols workaround. The Windows CI job will need to be updated before it can handle the new limitations this puts on Windows, see: https://github.com/ros2/rclcpp/issues/48#issuecomment-150714947


rclcpp/CMakeLists.txt
```diff
@@ -5,31 +5,66 @@ project(rclcpp)
...
+  src/rclcpp/parameter_service.cpp
+  src/rclcpp/publisher.cpp
+  src/rclcpp/node.cpp
+  src/rclcpp/service.cpp
+  src/rclcpp/subscription.cpp
+  src/rclcpp/timer.cpp
+  src/rclcpp/utilities.cpp
+)
+add_library(${PROJECT_NAME} SHARED ${${PROJECT_NAME}_SRCS})
+ament_target_dependencies(${PROJECT_NAME} "rmw" "rcl_interfaces")
```
dirk-thomas: Dependencies in alphabetical order.

rclcpp/CMakeLists.txt
```diff
@@ -5,31 +5,66 @@ project(rclcpp)
...
-  if(TARGET test_intra_process_manager)
-    target_include_directories(test_intra_process_manager PUBLIC
-      "${rcl_interfaces_INCLUDE_DIRS}"
-      "${rmw_INCLUDE_DIRS}")
-  endif()
+  # ament_add_gtest(test_intra_process_manager test/test_intra_process_manager.cpp)
+  # if(TARGET test_intra_process_manager)
+  #   target_include_directories(test_intra_process_manager PUBLIC
+  #     "${rcl_interfaces_INCLUDE_DIRS}"
+  #     "${rmw_INCLUDE_DIRS}")
```
dirk-thomas: These two lines should not be quotes in the first place since they can contain more than a single directory.

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -49,77 +49,46 @@ class CallbackGroup
...
-    return can_be_taken_from_;
-  }
-
-  const CallbackGroupType & type() const
-  {
-    return type_;
-  }
+  CallbackGroup(CallbackGroupType group_type);
+
+  const std::vector<rclcpp::subscription::SubscriptionBase::WeakPtr> &
```
dirk-thomas: Is there a need to prepend the local namespace `rclcpp`?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+  // TODO: determine if posix's nanosleep(2) is more efficient here
+  std::unique_lock<std::mutex> lock(::g_interrupt_mutex);
+  auto cvs = ::g_interrupt_condition_variable.wait_for(lock, nanoseconds);
+  // Return true if the timeout elapsed successfully, otherwise false.
+  return cvs != std::cv_status::no_timeout;
+}
+
+} /* namespace utilities */
+} /* namespace rclcpp */
```
dirk-thomas: New code should follow C++ style guide / pass ament_cpplint.

dirk-thomas:
For a more detailed review I will wait until the PR is rebased since the referenced diff is still difficult to read. Can I also assume that all the created .cpp files contain the exact code extracted from the .hpp files and don't need to be reviewed?

wjwwood:
> For a more detailed review I will wait until the PR is rebased since the referenced diff is still difficult to read.

I don't expect it will be easier to read, after the rebase, than this: https://github.com/ros2/rclcpp/compare/allocator_template...rclcpp_library because the only thing the rebase will do is remove @jacquelinekay's commits from this pr.

> Can I also assume that all the created .cpp files contain the exact code extracted from the .hpp files and don't need to be reviewed?

For the most part that is true. I can summarize the handful of cases where I had to do more:

- I changed two functions in `executor.hpp` to no longer be templated so I could put them in `.cpp`. The `wait_for_work` and `get_next_executable` used to be templated on a chrono duration ratio, but now they just take chrono nanoseconds and the calling functions do the duration cast before passing the timeout.

- I moved a static class member into a static global in a .cpp file in the `IntraProcessManager` class.

- I also moved some functions from `rclcpp.hpp` into `executors.hpp` so they could be put into a .cpp file. The `rclcpp.hpp` header cannot be included by the rclcpp library, but must be included by the user, see: https://github.com/ros2/rclcpp/compare/allocator_template...rclcpp_library#diff-690567624ee368716009118718985564R29 and the corresponding: https://github.com/ros2/rclcpp/compare/allocator_template...rclcpp_library#diff-809285dba1df7726afedd63625b6a6ceR34

wjwwood:
Rebased, but I'm still working on comments and fixing the intra process tests.

jacquelinekay:
Linux build was broken, I added a missing include.

wjwwood:
Thanks @jacquelinekay, I've updated it with fixes for `uncrustify` and I fixed the intra process tests.

rclcpp/include/rclcpp/callback_group.hpp
wjwwood: I think the headers should us the fully qualified names so people reading it can easily track down what the full namespace of parameters are. Plus I had to do this in the .cpp file or do `using namespace rclcpp` which I was trying to avoid and this made the function prototypes in the hpp and cpp files consistent with each other.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: I'm going to do a follow up pr to make it cpplint compliant. I'll do that at the same time as rcl and rclc. A more salient point is that I shouldn't use the namespace bracket like that in the cpp file and this file is inconsistent with the other cpp file's style. Fixed in 0b59eba.

wjwwood:
Ok, I think I've addressed all of the comments, so here is the Linux and OS X CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/522/
- http://ci.ros2.org/job/ros2_batch_ci_osx/395/

The biggest thing left to do before I think this is ready for merge is to update the Windows CI farm to do a matrix build of the different DDS vendors rather than all in one workspace. The matrix can either be just running N builds back to back in the same job or really using Jenkin's matrix build pattern. I don't know what the best option is in that case. I'll look at it tomorrow morning, but if others have ideas on how that should work, I'd be happy to have some help on that.

wjwwood:
I had to fix a problem that only showed up in isolated builds, where rclcpp had an undeclared dependency on `rosidl_generator_cpp`.

However, there is still and issue with how we are using some of the message related symbols within rclcpp which I haven't sorted out yet. I'll try to figure that out in the morning.

jacquelinekay:
Why do node.hpp and node_impl.hpp need to be in different files?

rclcpp/test/test_intra_process_manager.cpp
```diff
@@ -122,7 +122,8 @@ class SubscriptionBase
 #define Publisher mock::Publisher
 #define PublisherBase mock::PublisherBase
 #define SubscriptionBase mock::SubscriptionBase
-#include <rclcpp/intra_process_manager.hpp>
+#include "../src/rclcpp/intra_process_manager.cpp"
```
jacquelinekay: Ah. Interesting workaround.

wjwwood:
I think I have fixed the type support problems.

jacquelinekay:
It's interesting that the Linux CI job failed, because the isolated build actually works on my desktop... maybe I have fpermissive turned off?

wjwwood:
Thanks @jacquelinekay, I'm kicking off new CI jobs:

- http://ci.ros2.org/job/ros2_batch_ci_linux/524/
- http://ci.ros2.org/job/ros2_batch_ci_osx/397/
- http://ci.ros2.org/job/ros2_batch_ci_windows/552/

wjwwood:
I spent some time working on the Windows issues, but I've been unable to finish those changes. Instead I've just pushed some changes that should fix Linux's build.

The root of the issue on Linux was that sometimes no template functions which use `rmw_*` functions where used from `rclcpp`'s headers. This resulted in a situation where librclcpp was missing rmw symbols but the user's executable or library was not. So when linked against the rmw implementation shared library, the linker discarded it even though librclcpp would have used them. By using an rmw symbol directly in `rclcpp.hpp` I've forced all users of the header to depend on at least one rmw symbol, and therefore avoiding the aforementioned scenario.

There's still a problem that I introduced while working on Windows which introduces an infinite recursion that I need to track down, and so I won't start new CI. I'll have to tackle that tomorrow.

wjwwood:
I believe I have fixed the infinite recursion issue (it was a consequence of a Windows fix), so I'll do CI for Linux and OS X:

- http://ci.ros2.org/job/ros2_batch_ci_linux/529/
- http://ci.ros2.org/job/ros2_batch_ci_osx/400/

wjwwood:
It might be that the code I conditioned on WIN32 in d6769db5cfbbb535e4b8f094a64dfef0efda6210 maybe related to the SEH errors in Windows that we were seeing (they could also be caused by infinite recursion calls). In that case we would need to find another way to do 9cad948.

wjwwood:
Rebased after https://github.com/ros2/rclcpp/pull/142

wjwwood:
I have this branch building and tests running on all three platforms now, but there are still many warnings that need to be looked at on Windows, but in the mean time new CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/530/
- http://ci.ros2.org/job/ros2_batch_ci_osx/401/
- http://ci.ros2.org/job/ros2_batch_ci_windows/556/

wjwwood:
Ok I've fixed the linkage warnings on Windows. Now the only thing remaining is figuring out what to do about the dependency structure on Windows. That last CI job failed because `rmw_opensplice_cpp` was not built and installed before `rclcpp`:

http://ci.ros2.org/job/ros2_batch_ci_windows/556/

Strangely `rmw_connext_cpp` is built before, but I'm not sure why that might happen just yet. We'll probably end up having to put a dependency on the default rmw_implementations from some package that rclcpp depends on. We'll have to see what affect that has on Linux/OS X. Currently some of them build before rclcpp so I guess it shouldn't be that big of a change, e.g.:

```
# Topological order
[ ... ]
 - connext_cmake_module
 - opensplice_cmake_module
 - rmw
 - rmw_implementation
 - rmw_implementation_cmake
 - rosidl_cmake
 - rosidl_default_generators
 - rosidl_generator_c
 - rosidl_generator_dds_idl
 - rosidl_parser
 - rcl
 - rclc
 - rmw_connext_shared_cpp
 - rosidl_generator_cpp
 - rosidl_typesupport_connext_cpp
 - rmw_connext_cpp
 - rosidl_typesupport_introspection_cpp
 - rmw_connext_dynamic_cpp
 - rosidl_typesupport_opensplice_cpp
 - builtin_interfaces
 - example_interfaces
 - pendulum_msgs
 - rcl_interfaces
 - rclcpp
 - rmw_opensplice_cpp
[ ... ]
```

As you can see, both connext rmw implementations come before `rclcpp`, but `rmw_opensplice_cpp` comes just after for some reason.

wjwwood:
Ok, I added a build dependency on `rmw_implementation` in `rclcpp`. I'm not convinced this is correct, but I don't have a better idea and I think it will allow the branches to build and pass on all platforms, so here is CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/531/
- http://ci.ros2.org/job/ros2_batch_ci_osx/402/
- http://ci.ros2.org/job/ros2_batch_ci_windows/557/

If the CI looks ok, then we can start to discuss the content of the pull request and I can make any adjustments that we think are necessary.

wjwwood:
With https://github.com/ros2/rmw_opensplice/pull/90 I can build Windows completely. New CI for Windows (the Linux and OS X jobs look good):

- http://ci.ros2.org/job/ros2_batch_ci_windows/560/

I'll start a Connext one if that one looks good (and it doesn't interfere with ros2 packaging jobs).

wjwwood:
The Windows Connext job (http://ci.ros2.org/job/ros2_batch_ci_windows/561/) fails because it finds two rmw implementations: connext and connext dynamic.

Based on that I think we'll need to have options not just for connext vs opensplice, but also for connext vs connext static.

dirk-thomas:
That will be more difficult since both Connext rmw packages use the same underlying infrastructure. There is currently no way to select only one of them.

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -466,211 +349,8 @@ Node::create_service(
...
-}
-
-
-const Node::CallbackGroupWeakPtrList &
-Node::get_callback_groups() const
-{
-  return callback_groups_;
-}
+// #if defined _WIN32 || defined __CYGWIN__
+// RCLCPP_EXPIMP_TEMPLATE template class RCLCPP_PUBLIC std::shared_ptr<rmw_node_t>;
```
jacquelinekay: this is commented out, is it still needed?

rclcpp/include/rclcpp/visibility_control.hpp
```diff
@@ -0,0 +1,57 @@
...
+  #else
+    #define RCLCPP_PUBLIC
+    #define RCLCPP_LOCAL
+  #endif
+  #define RCLCPP_PUBLIC_TYPE
+#endif
+
+#if defined _WIN32 || defined __CYGWIN__
+  #ifdef RCLCPP_BUILDING_DLL
+    #define RCLCPP_EXPIMP_TEMPLATE
```
jacquelinekay: Can you provide a comment about why this macro is needed on Windows?

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+SyncParametersClient::SyncParametersClient(
+  rclcpp::executor::Executor::SharedPtr executor,
+  rclcpp::node::Node::SharedPtr node)
+: executor_(executor), node_(node)
+{
+  async_parameters_client_ = std::make_shared<AsyncParametersClient>(node);
+}
+
+std::vector<rclcpp::parameter::ParameterVariant>
```
jacquelinekay: @esteve made some changes to convert many of the vectors in these files to const references:

421f6b8b5e502fb28b7568baa26366a413db3691

it looks like they haven't been incorporated into the new cpp file, since that commit got merged after this PR was opened.

rclcpp/include/rclcpp/node_impl.hpp
wjwwood: I saw this and removed it when going over the cpplint stuff in 94e4f7e.

rclcpp/include/rclcpp/visibility_control.hpp
wjwwood: That's not needed anymore, I removed it in d3e5548.

rclcpp/src/rclcpp/parameter_client.cpp
wjwwood: I may have missed that one, but I definitely merged those manually and intended to take the new changes. @esteve can I ask you to go over those signatures again and make sure they're right?

The header looks ok, maybe the cpp file is out of sync: https://github.com/ros2/rclcpp/compare/master...rclcpp_library#diff-3d5a5b02e888be31f79e8c934aab2847

esteve: The function signatures look good to me, the const references are there. :+1:

jacquelinekay: my bad, I wasn't even looking at the input parameters

rclcpp/test/test_function_traits.cpp
```diff
@@ -142,15 +143,15 @@ TEST(TestFunctionTraits, arity) {
       return 0;
     };
 
-  auto lambda_one_int = [](int) {
+  auto lambda_one_int = [](int /*one*/) {
```
jacquelinekay: Is this the preferred style for unused function input variables?

dirk-thomas: I would suggest leaving out the variable name.
Only if an unused argument should be documented the body should contain something like this:

    (void)arg_name;

wjwwood: Two problems with this particular line. One is that cpplint think that is a C-Style cast, which it isn't, but the other issue is that cpplint asserts that you should put the parameters into function prototypes. Examples:

```
/Users/william/ros2/src/ros2/rclcpp/rclcpp/test/test_function_traits.cpp:24:  Using C-style cast.  Use static_cast<int>(...) instead  [readability/casting] [4]
```

```
/Users/william/ros2/src/ros2/rclcpp/rclcpp/test/test_function_traits.cpp:431:  All parameters should be named in a function  [readability/function] [3]
```

So unless I suppress those warnings, it isn't possible to leave out the parameter name.

wjwwood: Is there a suggestion for how to address this differently and avoid the cpplint warnings?

esteve: I always use `(void)unused_variable;`

dirk-thomas: I didn't know that you changed it because of cpplint. You can keep it as-is then.

wjwwood:
Ok, I made the final change (dcbdf86) I wanted to do before proposing this for squash and merge. I think it's ready for review and an up/down vote.

I'm happy to discuss the rational behind any questionable changes in person to speed up the process.

wjwwood:
In the mean time, I'll be looking into https://github.com/ros2/ros2/issues/147

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -81,7 +95,18 @@ spin_node_until_future_complete(
   return FutureReturnCode::INTERRUPTED;
 }
 
-} // namespace executors
-} // namespace rclcpp
+}  // namespace executors
+
+template<typename FutureT, typename TimeT = std::milli>
```
jacquelinekay: this is now out of the executors namespace and in the rclcpp namespace?

wjwwood: If I read the diff correctly it was always in the rclcpp namespace. I think I organized it like this because this function uses the above function which is in the executors namespace.

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -40,67 +36,36 @@ class MultiThreadedExecutor : public executor::Executor
...
   size_t number_of_threads_;
+  std::unordered_map<std::thread::id, size_t> thread_number_by_thread_id_;
 };
 
-}  // namespace multi_threaded_executor
-}  // namespace executors
-}  // namespace rclcpp
+} /* namespace multi_threaded_executor */
+} /* namespace executors */
+} /* namespace rclcpp */
```
dirk-thomas: Why change these three lines?

dirk-thomas:
lgtm

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
jacquelinekay: indeed, isn't this a regression for cpplint?

esteve:
+1

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
dirk-thomas: I would expect ament_cpplint to complain about these lines...

wjwwood: I didn't intentionally do that, I was making the alternative transformation everywhere else, but yet:

```
william@farl ~/ros2/src/ros2/rclcpp (git:rclcpp_library:737dba6)
% ament_cpplint /Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
Using '--root=rclcpp/include' argument

Done processing /Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp

No errors found

william@farl ~/ros2/src/ros2/rclcpp (git:rclcpp_library:737dba6)
% cat /Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp | grep namespace
namespace rclcpp
namespace executors
namespace multi_threaded_executor
} /* namespace multi_threaded_executor */
} /* namespace executors */
} /* namespace rclcpp */
```

dirk-thomas: Does the same happen with plain `cpplint`? If yes, can you please fill this upstream with a minimal example.

wjwwood: The `cpplint` from pip doesn't detect it either:

```
% cpplint /Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:18:  <mutex> is an unapproved C++11 header.  [build/c++11] [5]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:19:  <thread> is an unapproved C++11 header.  [build/c++11] [5]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:28:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:30:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:32:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:35:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:36:  public: should be indented +1 space inside class MultiThreadedExecutor  [whitespace/indent] [3]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:41:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:54:  protected: should be indented +1 space inside class MultiThreadedExecutor  [whitespace/indent] [3]
/Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp:59:  private: should be indented +1 space inside class MultiThreadedExecutor  [whitespace/indent] [3]
Done processing /Users/william/ros2/src/ros2/rclcpp/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
Total errors found: 10
```

wjwwood:
Ok, I'm going to squash.

wjwwood:
Ok I squashed and I'll try to get the connext on Windows tests going right after lunch and then it should be good to merge.

wjwwood:
Ok, hopefully final CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/566/
- http://ci.ros2.org/job/ros2_batch_ci_osx/439/
- http://ci.ros2.org/job/ros2_batch_ci_windows/584/ (opensplice)
- http://ci.ros2.org/job/ros2_batch_ci_windows/585/ (connext static)
- http://ci.ros2.org/job/ros2_batch_ci_windows/586/ (connext dynamic)

wjwwood:
Ok, they're all green except the connext versions of the Windows jobs. I'm gonna merge.

esteve:
:clap: :clap: :clap: :fireworks:

tfoote:
:tada:

dirk-thomas:
This PR results in the following compiler warning: http://ci.ros2.org/view/packaging/job/packaging_linux/51/warnings17Result/package.-1174361695/

esteve:
I've pushed this branch (https://github.com/ros2/rclcpp/tree/avoid_compiler_warning) to hopefully fix the compiler warning. CI jobs:

http://ci.ros2.org/job/ci_linux/659/
http://ci.ros2.org/job/ci_osx/542/
http://ci.ros2.org/job/ci_windows_opensplice/735/
http://ci.ros2.org/job/ci_windows_connext_dynamic/22/
http://ci.ros2.org/job/ci_windows_connext_static/21/

dirk-thomas:
Since the warning is only present in the packaging job but not in the CI job we should run the packaging job if we want to actually check if the patch makes a difference.

wjwwood:
Thanks for looking into it @esteve, you beat me to it.

esteve:
@wjwwood no problem, it's been fun to learn about (yet) another GNU extension :smile:

PR #143, opened 2015-11-03
Remove reference so that lambdas are supported

esteve:
This removes a reference, lambdas can't be passed otherwise.


tfoote:
+1

PR #144, opened 2015-11-03
account for spurious wake ups from std::condition_variable::wait_for

wjwwood:
On Windows this issue is most prenounced, but apparently it is possible on all platforms that `std::condition_variable::wait_for` can be spuriously interrupted and return `std::cv_status::no_timeout` even when not notified.

This logic attempts to deal with this by adding an additional bool to indicate when the system has been interrupted either by ctrl-c or `rclcpp::shutdown` and will recurse if the specified duration has not yet been exceeded.

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -45,7 +45,7 @@ rmw_guard_condition_t * g_sigint_guard_cond_handle = \
   rmw_create_guard_condition();
 /// Condition variable for timed sleep (see sleep_for).
 std::condition_variable g_interrupt_condition_variable;
-std::atomic<bool> g_is_interrupted = false;
+std::atomic<bool> g_is_interrupted(false);
 /// Mutex for protecting the global condition variable.
 std::mutex g_interrupt_mutex;
 
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -45,6 +45,7 @@ rmw_guard_condition_t * g_sigint_guard_cond_handle = \
   rmw_create_guard_condition();
 /// Condition variable for timed sleep (see sleep_for).
 std::condition_variable g_interrupt_condition_variable;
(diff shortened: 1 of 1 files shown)


wjwwood:
This bug affects the `cyclic_pipeline` demo.

wjwwood:
From: http://en.cppreference.com/w/cpp/thread/condition_variable/wait_for
> 1) [ ... ] It may also be unblocked spuriously. [ ... ]

esteve:
+1

jacquelinekay:
that doesn't seem like a very useful quality of `std::condition_variable`.

So I assume you're discarding the timeout status of the condition variable because you no longer consider it useful (e.g. we could spuriously wake up without having timed out)?

wjwwood:
Yeah, basically the return code is somewhat useless. There is a discussion about this on the Boost mailing list as well:

- https://svn.boost.org/trac/boost/ticket/9856
- http://comments.gmane.org/gmane.comp.lib.boost.devel/257246

And the conclusion of the mailing list seems to be that if you write code which doesn't expect spurious wake ups from `wait_for`/`wait_until` then you're gonna have a bad time. Apparently this can happen on *nix systems, but is uncommon. On Windows it happens like 30% of the time for me.

wjwwood:
I tested this manually on Windows, but I'll do CI for Linux and OS X:

- http://ci.ros2.org/job/ros2_batch_ci_linux/542/
- http://ci.ros2.org/job/ros2_batch_ci_osx/411/

jacquelinekay:
This is amazing: "In essence, if you now ask Windows to go sleep for X milliseconds, 
Windows Vista onwards will in fact sleep for anywhere between zero 
and X+N milliseconds where N is some arbitrarily long value."

Code looks good, waiting on CI

jacquelinekay:
Build failure on Linux because `std::atomic` is not copy constructible.

wjwwood:
Beat me to it :smile:.

wjwwood:
New CI: http://ci.ros2.org/job/ros2_batch_ci_linux/543/

tfoote:
+1

PR #145, opened 2015-11-03
fix warning about ignored return value

dirk-thomas:
Before http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/41/warnings17Result/package.-340390346/
After http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/42/


wjwwood:
lgtm

dirk-thomas:
For the record: the warning did not occur in our CI jobs since we don't specify a CMake build type. In the packaging jobs the warnings appear since we build with RelWithDebInfo.

PR #146, opened 2015-11-05
Change behavior of timer on construction

jacquelinekay:
Fixes ros2/system_tests#36

Based on the discussion in ros2/system_tests#66

Previously, timers would fire on the very first call to spin, no matter how much time passed between their construction and the call to spin.

This PR will change the behavior of timers so that the timer will fire one period after being created, or on the first call to spin IF that call happens after one period.

### DIFF
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -97,9 +97,8 @@ public:
   GenericTimer(std::chrono::nanoseconds period, CallbackType callback)
   : TimerBase(period, callback), loop_rate_(period)
   {
-    /* Subtracting the loop rate period ensures that the callback gets triggered
-       on the first call to check_and_trigger. */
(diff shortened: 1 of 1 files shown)


wjwwood:
I'm working on getting https://github.com/ros2/rclcpp/pull/140 merged today, can we defer this fix until then?

tfoote:
+1

jacquelinekay:
Sure. I'll run tests with this branch to check its correctness but I won't
merge until #140 is in.

wjwwood:
It looks good to me too.

dirk-thomas:
+1

esteve:
+1

jacquelinekay:
This appears to have caused a regression in test_rclcpp:

http://ci.ros2.org/job/ros2_batch_ci_linux/567/testReport/

I'll follow up with a fix.

PR #147, opened 2015-11-05
Pass callbacks as universal references

esteve:
This PR uses universal references for callbacks.

### DIFF
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -125,7 +125,7 @@ public:
 
   SharedFuture async_send_request(
     typename ServiceT::Request::SharedPtr request,
-    CallbackType cb)
+    CallbackType && cb)
   {
     int64_t sequence_number;
     if (RMW_RET_OK != rmw_send_request(get_client_handle(), request.get(), &sequence_number)) {
@@ -137,7 +137,8 @@ public:
 
     SharedPromise call_promise = std::make_shared<Promise>();
     SharedFuture f(call_promise->get_future());
-    pending_requests_[sequence_number] = std::make_tuple(call_promise, cb, f);
+    pending_requests_[sequence_number] =
+      std::make_tuple(call_promise, std::forward<CallbackType>(cb), f);
     return f;
   }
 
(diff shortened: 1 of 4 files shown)


esteve:
CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/556/
http://ci.ros2.org/job/ros2_batch_ci_osx/427/
http://ci.ros2.org/job/ros2_batch_ci_windows/573/

wjwwood:
Again, I appreciate it if this could be deferred until https://github.com/ros2/rclcpp/pull/140 is merged.

esteve:
Sure, no problem, it's not marked for review yet, though.

esteve:
I've cancelled the CI jobs.

esteve:
CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/568/
http://ci.ros2.org/job/ros2_batch_ci_osx/442/
http://ci.ros2.org/job/ros2_batch_ci_windows/589/

dirk-thomas:
The Windows build will fail since you can only build with a single rmw impl. enabled.

esteve:
I triggered two more jobs, one for OpenSplice and another for Connext:

http://ci.ros2.org/job/ros2_batch_ci_windows/590/
http://ci.ros2.org/job/ros2_batch_ci_windows/591/

dirk-thomas:
The "Connext" job will fail too since you must select a single rmw implementation.

esteve:
http://ci.ros2.org/job/ros2_batch_ci_windows/591/ for Connext static
http://ci.ros2.org/job/ros2_batch_ci_windows/592/ for Connext dynamic

jacquelinekay:
Can you give a summary of why a universal reference is needed, instead of the current behavior, or instead of an lvalue reference?

esteve:
This way we avoid extra copies and passing normal references breaks lambdas.

jacquelinekay:
so you can't pass a lambda to a function by lvalue reference? I couldn't find any documentation on the matter.

I recall you made this PR because of that problem: https://github.com/ros2/rclcpp/pull/143

was it because somewhere the lambdas were being passed as temporaries? Then I can see why perfect forwarding would be needed.

Anyway, regardless of the details, the change is a good idea; I'm trying to learn more about C++11...

esteve:
Yes, the result of a lambda expression is a temporary, which can't be bound to lvalue references, hence the need for universal ones.

esteve:
In general, any function that takes a template argument should use a universal references, unless we really need copies.

jacquelinekay:
+1

PR #153, opened 2015-11-13
Implement Executor::cancel

jacquelinekay:
implement state and logic for cancelling out of Executor::spin or Executor::spin_some.

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -20,8 +20,7 @@ using rclcpp::executor::AnyExecutable;
 using rclcpp::executor::Executor;
 
 Executor::Executor(rclcpp::memory_strategy::MemoryStrategy::SharedPtr ms)
-: interrupt_guard_condition_(rmw_create_guard_condition()),
-  canceled(false),
+: spinning(false), interrupt_guard_condition_(rmw_create_guard_condition()),
   memory_strategy_(ms)
 {
 }
@@ -109,26 +108,33 @@ Executor::spin_node_some(rclcpp::node::Node::SharedPtr node)
 void
 Executor::spin_some()
 {
-  canceled = false;
+  if (spinning.exchange(true)) {
+    throw std::runtime_error("spin_some() called while already spinning");
+  }
   AnyExecutable::SharedPtr any_exec;
-  while ((any_exec = get_next_executable(std::chrono::milliseconds::zero())) && !canceled) {
+  while ((any_exec = get_next_executable(std::chrono::milliseconds::zero())) && spinning.load()) {
     execute_any_executable(any_exec);
   }
+  spinning.store(false);
 }
 
 void
 Executor::spin_once(std::chrono::nanoseconds timeout)
 {
+  if (spinning.exchange(true)) {
+    throw std::runtime_error("spin_once() called while already spinning");
+  }
   auto any_exec = get_next_executable(timeout);
   if (any_exec) {
     execute_any_executable(any_exec);
   }
+  spinning.store(false);
 }
 
 void
 Executor::cancel()
 {
-  canceled = true;
+  spinning.store(false);
   rmw_ret_t status = rmw_trigger_guard_condition(interrupt_guard_condition_);
   if (status != RMW_RET_OK) {
     throw std::runtime_error(rmw_get_error_string_safe());
@@ -147,7 +153,7 @@ Executor::set_memory_strategy(rclcpp::memory_strategy::MemoryStrategy::SharedPtr
 void
 Executor::execute_any_executable(AnyExecutable::SharedPtr any_exec)
 {
-  if (!any_exec) {
+  if (!any_exec || !spinning.load()) {
     return;
   }
   if (any_exec->timer) {
@@ -514,14 +520,11 @@ Executor::get_next_executable(std::chrono::nanoseconds timeout)
   if (!any_exec) {
     // Wait for subscriptions or timers to work on
     wait_for_work(timeout);
-    if (canceled) {
+    if (!spinning.load()) {
       return nullptr;
     }
     // Try again
     any_exec = get_next_ready_executable();
-  }
-  if (canceled) {
-    return nullptr;
   }
   // At this point any_exec should be valid with either a valid subscription
   // or a valid timer, or it should be a null shared_ptr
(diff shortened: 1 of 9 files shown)


tfoote:
Seeing this do we have race conditions related to calling spin* from multiple threads? It lookslike I could call spin_some in one thread then cancel, and spin_some in another thread and if the timing is right both will continue to spin, since the cancel has been reset by the spin_some?

wjwwood:
The spin functions are not thread safe, so I'd say that's unsupported behavior.

jacquelinekay:
If users want parallelized execution they should use the `MultiThreadedExecutor` (for which I will implement `spin_some` soon). I think that the entry point to spin should always be from a single thread, however.

jacquelinekay:
CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/590/
http://ci.ros2.org/job/ros2_batch_ci_osx/
http://ci.ros2.org/job/ros2_batch_ci_windows/650/

jacquelinekay:
Rerunning CI with changes from @wjwwood to ensure spin* cannot be entered by multiple threads at the same time.

http://ci.ros2.org/job/ros2_batch_ci_linux/591/
http://ci.ros2.org/job/ros2_batch_ci_osx/477/
http://ci.ros2.org/job/ros2_batch_ci_windows/652/

wjwwood:
Lgtm, but someone else should review it as well.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -108,20 +108,37 @@ Executor::spin_node_some(rclcpp::node::Node::SharedPtr node)
...
-    get_next_executable(std::chrono::milliseconds::zero()))
-  {
+  if (spinning.exchange(true)) {
+    throw std::runtime_error("spin_some() called while already spinning");
+  }
+  AnyExecutable::SharedPtr any_exec;
+  while ((any_exec = get_next_executable(std::chrono::milliseconds::zero())) && spinning.load()) {
     execute_any_executable(any_exec);
   }
+  spinning.store(false);
```
dirk-thomas: Manually resetting this is not safe when exceptions happen. Since it can't be done in a finally block this should use RAII to ensure that value gets reset to false.

Same below.

jacquelinekay: Do you suggest setting it to false in the destructor, then?

dirk-thomas: Let me move the comment to the next occurrence instead.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -108,20 +108,37 @@ Executor::spin_node_some(rclcpp::node::Node::SharedPtr node)
...
 Executor::spin_once(std::chrono::nanoseconds timeout)
 {
+  if (spinning.exchange(true)) {
+    throw std::runtime_error("spin_once() called while already spinning");
+  }
   auto any_exec = get_next_executable(timeout);
   if (any_exec) {
     execute_any_executable(any_exec);
   }
+  spinning.store(false);
```
dirk-thomas: Manually resetting this is not safe when exceptions happen. Since it can't be done in a finally block this should use RAII to ensure that value gets reset to false.

Same below.

rclcpp/src/rclcpp/executor.cpp
dirk-thomas: I refer to calling `spinning.exchange(true)` before in this function. If an exception happens afterwards it doesn't get set back to false.

jacquelinekay: Got it.

I was trying to avoid adding mutexes to Executor where they're not needed. But `mutex` and `lock_guard` are the most suited to RAII-style management of this state. Maybe this would work with `try_lock` instead of `lock` so that we throw an exception if the Executor is already spinning instead of blocking.

@wjwwood, since you implemented about half of this PR, any thoughts?

wjwwood: Yeah I'll send another pr asap.

jacquelinekay: I am working on something using mutexes right now, let me know if I should stop since it sounds like you're working on it too...

wjwwood: Well you can do it if you like, but I would recommend not using a mutex. Atomic bool can be more efficient. I was just going to wrap it in a small class so it got reset on destruction when leaving that scope.

jacquelinekay: Yeah, my first pass at the strategy I suggested was not successful (deadlock) and it makes the code a lot more confusing.

wjwwood: Sorry, I just realized you were expecting me to do this :smile:. I totally misread it, but anyways I just pushed my fix. I did it by implementing what's called a scope exit or scope guard. Boost has a scope exit and so does the D language. There's some interesting discussion about including something like this in the c++ std library at some point. Anyways this should work us and might be useful else where too.

tfoote: With this implemented should we remove the redundant setting to false at the end of the scope?

wjwwood: That's a fair point, I'll push another commit.

wjwwood: f1e7ea5

wjwwood:
New CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/595/
- http://ci.ros2.org/job/ros2_batch_ci_osx/482/
- http://ci.ros2.org/job/ros2_batch_ci_windows/659/ (opensplice)
- http://ci.ros2.org/job/ros2_batch_ci_windows/660/ (connext static)
- http://ci.ros2.org/job/ros2_batch_ci_windows/661/ (connext dynamic)

jacquelinekay:
Lots of Executor test failures. I'll take a look.

wjwwood:
Sorry, I actually ran the tests locally and there was a failing test, I just missed it (I am not a smart man). I fixed the issue, see 1c9eb0b.

dirk-thomas:
lgtm

wjwwood:
CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/596/
- http://ci.ros2.org/job/ros2_batch_ci_osx/483/
- http://ci.ros2.org/job/ros2_batch_ci_windows/662/ (opensplice)
- http://ci.ros2.org/job/ros2_batch_ci_windows/663/ (connext)
- http://ci.ros2.org/job/ros2_batch_ci_windows/664/ (connext dynamic)

wjwwood:
CI looks good.

PR #156, opened 2015-11-18
Fixes for multithreaded execution

jacquelinekay:
Connects to ros2/ros2#92

Known issues:

to reproduce:

./build/test_rclcpp/gtest_multithreaded__rmw_opensplice_cpp --gtest_filter=*publisher_intra_process --gtest_repeat=-1

Issues only seem to occur in Opensplice in the intra process test cases, and much more frequently in the multi_access_publisher_intra_process case.

- AnyExecutable occasionally gets garbage in a few of its shared_ptr fields.

```
timer: {_M_ptr = 0x106cbf8, _M_refcount = {_M_pi = 0x106cbe0}}, <No data fields>}
subscription: {_M_ptr = 0x0, _M_refcount = {_M_pi = 0x2}}, <No data fields>}
subscription_intra_process: {_M_ptr = 0x100003b31, _M_refcount = {_M_pi = 0x0}}
```

It looks to me like:
timer is a valid shared_ptr. The pointer value changes between runs but it always looks like a valid shared_ptr.
subscription is a null pointer with a refcount of 2 (not good). These values stay the same in every occurrence of the bug.
subscription_intra_process is an invalid pointer (value evaluates to 2^32+15153 in decimal) with no ref count. The ref count is always 0, the pointer value changes between runs (but it's always over 2^32 aka 0x100000000).

Timer is not always set, sometimes the error occurs when the any_executable already has null ptr fields and it's trying to set `subscription`. However, it segfaults because it sees that `subscription` has a non-zero refcount and tries to release already null memory.

- Random deadlock.
- Random segfaults due to ownership issues of an rmw_guard_condition (probably have to extend locking of the interrupt_guard_condition in Executor).

### DIFF
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -48,9 +48,9 @@ struct AnyExecutable
   rclcpp::callback_group::CallbackGroup::SharedPtr get_callback_group() const;
   rclcpp::node::Node::SharedPtr get_node() const;
 
-  void set_subscription(const rclcpp::subscription::SubscriptionBase::SharedPtr subscription);
-  void set_subscription_intra_process(const rclcpp::subscription::SubscriptionBase::SharedPtr subscription);
-  void set_timer(const rclcpp::timer::TimerBase::SharedPtr timer);
+  void set_subscription(const rclcpp::subscription::SubscriptionBase::ConstSharedPtr subscription);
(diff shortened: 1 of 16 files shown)


esteve:
> in progress. don't look here!

Too late! What should I do now?

jacquelinekay:
Now that you've looked at this issue, you have to help me fix it! Bwahahaha!

Issue #35, opened 2015-05-26
Add support for a service response callback with request payload embedded


tfoote:
This was identified as potentially valuable during the parameter API review. 

The use case is that there are a sequence of queries in a request, the response objects can be sorted in the same order as the requests, instead of requiring the response to be entirely self describing.

garyservin:
@tfoote can you add more info for this issue? I'm feeling a bit out out of context here. Thanks!

tfoote:
This was mostly in the context of an async service. Where at the time of response the initial request may be out of scope. In that case the response could be coupled with the request. 

For example [here]( https://github.com/ros2/examples/blob/master/rclcpp_examples/src/services/add_two_ints_client_async.cpp#L35) you could print both the request and the result instead of just the result. This saves the user from having to book keep which request is associated with which response if they're running multiple in parallel, which is not represented in the simple example.

PR #157, opened 2015-11-18
Added async_send_request_return_request to return the originating request

esteve:
Connects to #35

### DIFF
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -135,7 +135,7 @@ public:
   template<
     typename CallbackT,
     typename std::enable_if<
-      rclcpp::same_arguments<
+      rclcpp::function_traits::same_arguments<
         CallbackT,
         CallbackType
       >::value
@@ -161,7 +161,7 @@ public:
   template<
     typename CallbackT,
     typename std::enable_if<
-      rclcpp::same_arguments<
+      rclcpp::function_traits::same_arguments<
         CallbackT,
         CallbackWithRequestType
       >::value
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -23,6 +23,7 @@
 #include <tuple>
 #include <utility>
 
+#include "rclcpp/function_traits.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/utilities.hpp"
 #include "rclcpp/visibility_control.hpp"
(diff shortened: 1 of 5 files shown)


esteve:
I'm not sure about the name of the method, so I'd appreciate if anyone comes with a better one :-)

esteve:
@tfoote let me know if this is what you had in mind or if it needs to be reworked.

esteve:
CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/604/
http://ci.ros2.org/job/ros2_batch_ci_osx/490/
http://ci.ros2.org/job/ros2_batch_ci_windows/672/

tfoote:
That's what we talked about during the 

Is it not possible to overload the existing method which would avoid the very verbose method name?

rclcpp/include/rclcpp/client.hpp
```diff
@@ -142,10 +163,39 @@ class Client : public ClientBase
     return f;
   }
 
+  SharedFutureWithRequest async_send_request_return_request(SharedRequest request)
```
tfoote: Do we need this method? A noop callback doesn't need the request data.

esteve:
New CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/605/
http://ci.ros2.org/job/ros2_batch_ci_osx/491/
http://ci.ros2.org/job/ros2_batch_ci_windows/673/

esteve:
I ended up using `function_traits` to fix a compilation bug on Windows because of the recurring issue with `std::function`. I've also added specializations for references and universal references to `function_traits` so that it can be used with lambdas that are not wrapped in `std::function` objects.

New CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/605/
http://ci.ros2.org/job/ros2_batch_ci_osx/492/
http://ci.ros2.org/job/ros2_batch_ci_windows_opensplice/674/

tfoote:
+1 

The connext_dynamic job failed on the filename length. @wjwwood is that fix still pending?

esteve:
I pushed one last change to move the `function_traits` templates to a separate namespace. I could open a separate PR for it. In any case, I triggered new CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/607/
http://ci.ros2.org/job/ros2_batch_ci_osx/493/
http://ci.ros2.org/job/ros2_batch_ci_windows_connext_dynamic/3/
http://ci.ros2.org/job/ros2_batch_ci_windows_connext_static/3/
http://ci.ros2.org/job/ros2_batch_ci_windows_opensplice/675/

tfoote:
+1 the new name is much clearer. There probably will be usages in other repos of the old names?

esteve:
`function_traits` is only used in `rclcpp` itself. I updated the test in `system_tests` to use `async_send_request` instead of the longer name.

esteve:
I also updated the code to pass uncrustify.

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_linux/611/testReport/junit/rclcpp.uncrustify/xunit/include_rclcpp_function_traits_hpp/

esteve:
https://github.com/ros2/rclcpp/commit/03697d149a8f89337c2da3acd385c0945547f96c

PR #159, opened 2015-11-19
Add alternative callback signature to timer for accepting a Timer reference

jacquelinekay:
~~As discussed with @wjwwood offline. This PR changes the Timer callback signature to return a bool, which signals the Executor to cancel the timer if true.~~

Create an alternative callback signature to Timer that accepts a Timer object. This enables the user to implement timers that fire any number greater than 1 times by capturing a variable to keep track of the number of callbacks triggered and conditionally cancelling the timer. It could also allow the user to change the loop rate in the future.

### DIFF
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -20,7 +20,9 @@
 #include <memory>
 #include <sstream>
 #include <thread>
+#include <type_traits>
 
+#include "rclcpp/function_traits.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/rate.hpp"
 #include "rclcpp/utilities.hpp"
@@ -33,15 +35,13 @@ namespace rclcpp
 namespace timer
 {
 
-using CallbackType = std::function<void()>;
-
 class TimerBase
 {
 public:
   RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(TimerBase);
 
   RCLCPP_PUBLIC
-  TimerBase(std::chrono::nanoseconds period, CallbackType callback);
+  TimerBase(std::chrono::nanoseconds period);
 
   RCLCPP_PUBLIC
   virtual ~TimerBase();
@@ -51,12 +51,8 @@ public:
   cancel();
 
   RCLCPP_PUBLIC
-  void
-  execute_callback() const;
-
-  RCLCPP_PUBLIC
-  const CallbackType &
-  get_callback() const;
+  virtual void
+  execute_callback() = 0;
 
   /// Check how long the timer has until its next scheduled callback.
   // \return A std::chrono::duration representing the relative time until the next callback.
@@ -77,13 +73,23 @@ public:
 
 protected:
   std::chrono::nanoseconds period_;
-  CallbackType callback_;
 
   bool canceled_;
 };
 
+
+using VoidCallbackType = std::function<void()>;
+using TimerCallbackType = std::function<void(TimerBase &)>;
+
 /// Generic timer templated on the clock type. Periodically executes a user-specified callback.
-template<class Clock = std::chrono::high_resolution_clock>
+template<
+  typename FunctorT,
+  class Clock = std::chrono::high_resolution_clock,
+  typename std::enable_if<
+    rclcpp::function_traits::same_arguments<FunctorT, VoidCallbackType>::value ||
+    rclcpp::function_traits::same_arguments<FunctorT, TimerCallbackType>::value
+  >::type * = nullptr
+>
 class GenericTimer : public TimerBase
 {
 public:
@@ -94,8 +100,8 @@ public:
    * \param[in] period The interval at which the timer fires.
    * \param[in] callback User-specified callback function.
    */
-  GenericTimer(std::chrono::nanoseconds period, CallbackType callback)
-  : TimerBase(period, callback), loop_rate_(period)
+  GenericTimer(std::chrono::nanoseconds period, FunctorT && callback)
+  : TimerBase(period), callback_(callback), loop_rate_(period)
   {
     /* Set last_triggered_time_ so that the timer fires at least one period after being created. */
     last_triggered_time_ = Clock::now();
@@ -106,6 +112,38 @@ public:
   {
     // Stop the timer from running.
     cancel();
+  }
+
+  void
+  execute_callback()
+  {
+    execute_callback_delegate<>();
+  }
+
+  // void specialization
+  template<
+    typename CallbackT = FunctorT,
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<CallbackT, VoidCallbackType>::value
+    >::type * = nullptr
+  >
+  void
+  execute_callback_delegate()
+  {
+    callback_();
+  }
+
+  template<
+    typename CallbackT = FunctorT,
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<CallbackT, TimerCallbackType>::value
+    >::type * = nullptr
+  >
+  void
+  execute_callback_delegate()
+  {
+    //callback_(std::move(std::shared_ptr<TimerBase>(this)));
+    callback_(*this);
   }
 
   bool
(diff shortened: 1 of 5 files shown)


wjwwood:
+1, I was originally thinking that we'd add an alternative callback signature (so you could continue to use `void()`), but this is probably better.

jacquelinekay:
I was wondering about implementing this as an alternative callback signature instead of replacing it, but this was the shortest change to make. It is a bit ugly that "infinite" timers are crowded with "return false".

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -33,7 +33,9 @@ namespace rclcpp
 namespace timer
 {
 
-using CallbackType = std::function<void()>;
+/// The timer callback signature returns a boolean.
+// If this callback returns true, then the timer should cancel.
```
wjwwood: This surprised me when looking at the changes to the demos and tests. I was originally thinking that returning false would cancel the timer. But I don't have a good reason for that. To me neither "return false to indicate the timer should not continue firing" nor "return true to cancel the timer" are particularly intuitive. Perhaps we should return an typed enum so it's either `return timer::continue;` or `return timer::cancel;`. Not sure it's worth the effort, but see a bare `return false;` at the end of the timer callbacks left me guessing what that meant.

wjwwood:
> It is a bit ugly that "infinite" timers are crowded with "return false".

I was just thinking the same thing: https://github.com/ros2/rclcpp/pull/159#discussion_r45417160

tfoote:
+1 for enum return type

rclcpp/include/rclcpp/timer.hpp
jacquelinekay: alright, can you think of a good synonym for `continue` (since it's a reserved C++ keyword)? I'm thinking `resume`?

Also, should I elevate the namespace of `timer::ReturnCode::resume` so that we can use `timer::resume`

dirk-thomas:
Another alternative would be a `bool & cancel` argument for the timer callback. Then a callback which doesn't care about cancel has no arguments and no return. While a callback which would like to do so has an extra argument which it can use to cancel.

jacquelinekay:
Can you pass a function of signature `void()`  to a constructor that expects a `std::function<void(bool &)>`? It would be convenient, but I'm not sure it would compile.

dirk-thomas:
No, this would require two different callback signatures.

jacquelinekay:
If we're going to have two different callback signatures, I think I prefer the `void()` signature and `timer::ReturnCode()` signature.

wjwwood:
Another option is to pass the Timer object into the callback, so the signature would be `void(TimerSharedPtr timer)`. Then the callback could just do `timer.cancel();` before returning. In the future you even see the callback using this handle to the timer to get information like when was the last time it was called or even change the timer until the next call. Perhaps instead of handle to the Timer object a TimerEvent object which could be used to get this info and to cancel, but without exposing the original Timer's entire interface. Just a thought, I'd be happy with any of the options. Unfortunately I don't have a good alternative to continue, but maybe something like `run_again`. I dunno.

dirk-thomas:
@jacquelinekay I would choose the argument over the return value because of the zero-overhead principle: "what you don't use, you don't pay for".

@wjwwood Optionally passing the timer would be even more powerful. The callback can not only cancel the timer but also change its interval etc.

jacquelinekay:
Example with input argument:

```
auto callback = [](bool & cancel) {
  if (...) {
    cancel = true;
    return;
  }
  ...
};
```

---
Executor code:
```
void execute_timer(Timer::SharedPtr timer) {
  bool cancel = false;
  timer->execute_callback(cancel);
  if (cancel) {
    timer->cancel();
  }
}
```

---
Example with return code:
```
auto callback = []() -> rclcpp::timer::ReturnCode {
  if (...) {
    return rclcpp::timer::cancel;
  }
  ...
  return rclcpp::timer::return;
};
```

---
Executor code:
```
void execute_timer(Timer::SharedPtr timer) {
  if (timer->execute_callback()) {
    timer.cancel();
  }
}
```

Comparing these two examples, it's not entirely clear to me where the zero-overhead principle applies. I still slightly prefer the second one because it's slightly less verbose and the callback declaration in user code is less confusing. I can understand that as an argument to keep the `void()` callback type as an alternative.

dirk-thomas:
If you apply the zero-overhead principle from the users point of view you could argue that if the users wants a plain callback he neither wants a function argument nor a return statement. "The simple case should be simple."

esteve:
@jacquelinekay I just pushed https://github.com/ros2/rclcpp/pull/160 in case you want to SFINAE callbacks on their return type.

jacquelinekay:
thanks @esteve, that's exactly what I'll need!

gbiggs:
I like the pass-a-timer-object option. It has lots of potential to do powerful things.

If there are concerns about needing to pass something that isn't used, then perhaps there is a way for the user to provide a timer callback that doesn't take an argument, in which case they will not receive the timer object. Function traits are not my strong point, but they should be able to deduce something like that.

esteve:
@gbiggs we had to implement `function_traits` to work around an issue in VS2015 related to `std::function`. We use `function_traits` for supporting service callbacks that may or may not have a request header, and also for supporting intraprocess communication, see:

https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/function_traits.hpp#L48
https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/any_service_callback.hpp#L57
https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/any_service_callback.hpp#L71

this is a similar usecase to the one you describe.

jacquelinekay:
Ok, sounds like passing a pointer to a timer is the way to go.

I also want to support multiple callback signatures, though it will require some shuffling around to hide the CallbackType from the Timer base class.

jacquelinekay:
I've finished implementing the improvements we discussed and added a new test case in test_timer I'm currently passing the timer to the new callback signature as a reference, it seemed like the safest thing to do.

ci.ros2.org/job/ros2_batch_ci_linux/619
ci.ros2.org/job/ros2_batch_ci_osx/510
ci.ros2.org/job/ros2_batch_ci_windows_opensplice/686

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -77,13 +73,21 @@ class TimerBase
...
+using VoidCallbackType = std::function<void()>;
+using TimerCallbackType = std::function<void(TimerBase &)>;
+
 /// Generic timer templated on the clock type. Periodically executes a user-specified callback.
-template<class Clock = std::chrono::high_resolution_clock>
+template<typename FunctorT, class Clock = std::chrono::high_resolution_clock /*,
+  typename std::enable_if<
+    rclcpp::function_traits::same_arguments<FunctorT, VoidCallbackType>::value ||
+    rclcpp::function_traits::same_arguments<FunctorT, TimerCallbackType>::value
+  >::type * = nullptr */
```
dirk-thomas: Should this comment be removed or does it need to be used to check the types?

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -108,6 +112,36 @@ class GenericTimer : public TimerBase
...
+  execute_callback()
+  {
+    execute_callback_delegate<>();
+  }
+
+  // void specialization
+  template<typename CallbackT = FunctorT,
+  typename std::enable_if<
+    rclcpp::function_traits::same_arguments<CallbackT, VoidCallbackType>::value
+  >::type * = nullptr
```
esteve: Does this pass `uncrustify` ?

esteve:
+1

rclcpp/include/rclcpp/timer.hpp
jacquelinekay: It does; in fact, I think uncrustify changed my indentation to this from something like

```
  template<typename CallbackT = FunctorT,
    typename std::enable_if<
      rclcpp::function_traits::same_arguments<CallbackT, VoidCallbackType>::value
    >::type * = nullptr
  >
```

which I find more readable.

esteve: Your original indentation looks more readable. Maybe moving the first `typename` to a new line would prevent `uncrustify` from reformatting this code further.

jacquelinekay: Looks like that worked.

esteve: Awesome!

dirk-thomas:
Linter warnings: http://ci.ros2.org/job/ci_linux/628/

dirk-thomas:
Fails to build on Windows: http://ci.ros2.org/job/ci_windows_opensplice/700/

PR #160, opened 2015-11-20
Added support for deducing the return type of a functor

esteve:
This adds `return_type` as a member of the `function_traits` structure.

### DIFF
--- a/rclcpp/test/test_function_traits.cpp
+++ b/rclcpp/test/test_function_traits.cpp
@@ -14,7 +14,9 @@
 
 #include <gtest/gtest.h>
 
-#include <rclcpp/function_traits.hpp>
+#include <string>
+
+#include "rclcpp/function_traits.hpp"
 
 int func_no_args()
 {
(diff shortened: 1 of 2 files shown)


esteve:
@jacquelinekay if you don't need this, I can close this PR without merging the code. But if you think this may be of use to you, could you give it a look? Thanks.

jacquelinekay:
I don't need it right now since I decided to only use callbacks with a void return type in the timer callback pull request, but it may be useful in the future.

jacquelinekay:
The code looks okay to me. I started some jobs but they are all pending except for Linux:

http://ci.ros2.org/job/ci_linux/627/

rclcpp/test/test_function_traits.cpp
```diff
@@ -15,6 +15,7 @@
 #include <gtest/gtest.h>
 
 #include <rclcpp/function_traits.hpp>
+#include <string>
```
dirk-thomas: System headers like `string` should be included before user land header like `rclcpp`.

esteve: Actually, I had to put the headers in this order because `cpplint` if `<string>` was before `<rclcpp/function_traits.hpp>`:

```
/home/esteve/Projects/ros2/ros2/rclcpp/rclcpp/test/test_function_traits.cpp:18:  Found C system header after C++ system header. Should be: test_function_traits.h, c system, c++ system, other.  [build/include_order] [4]
```

jacquelinekay: I believe the convention here is the put `rclcpp` headers in quotes.

dirk-thomas: Because you need to include rclcpp with quotes instead.

esteve: Fixed.

esteve:
This compiles and runs fine on Windows too:

http://ci.ros2.org/job/ci_windows_opensplice/698/testReport/%28root%29/TestFunctionTraits/return_type/

however there's a new failing test, which I think is unrelated to this PR:

http://ci.ros2.org/job/ci_windows_opensplice/698/testReport/nose.failure/Failure/runTest/

jacquelinekay:
I've seen that nose test fail intermittently on OSX and Windows on other branches, too.

+1

dirk-thomas:
@esteve The failing test in osrf_pycommon is fixed with osrf/osrf_pycommon#17 which hasn't been reviewed / merged yet.

@jacquelinekay The PR only fixes a Windows related problem though. Can you please point me to the failing test on OS X.

esteve:
@dirk-thomas I just approved the PR you linked.

Going to merge this.

jacquelinekay:
This is one instance of the failure on OSX. The error is different from the failure Esteve linked, however (can't find `launch` vs. can't find `termios`).

http://ci.ros2.org/job/ci_osx_nightly/120/testReport/nose.failure/Failure/runTest/

esteve:
@jacquelinekay that error was fixed in `launch` https://github.com/ros2/launch/commit/7857c91eedb5ba10b083ead0ae977e00f3822836

PR #162, opened 2015-11-25
use std::shared_ptr type directly instead of using message specific type

dirk-thomas:
The current code requires the type `MessageT` to have a type `SharedPtr` / `ConstSharedPtr` defined within.

With the patch this requirement i gone and the code uses std::shared_ptr directly (that is also explicitly embedded in the signature of the function being invoked next).

### DIFF
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -183,16 +183,16 @@ public:
   {
     (void)message_info;
     if (shared_ptr_callback_) {
-      typename MessageT::SharedPtr shared_message = std::move(message);
+      typename std::shared_ptr<MessageT> shared_message = std::move(message);
       shared_ptr_callback_(shared_message);
     } else if (shared_ptr_with_info_callback_) {
-      typename MessageT::SharedPtr shared_message = std::move(message);
+      typename std::shared_ptr<MessageT> shared_message = std::move(message);
       shared_ptr_with_info_callback_(shared_message, message_info);
     } else if (const_shared_ptr_callback_) {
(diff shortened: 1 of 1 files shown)


wjwwood:
Why?

dirk-thomas:
To not require to have that internal type being defined. It is simply not necessary to make that assumption.

With this patch I can pass whatever `MessageT` I like - as long as there is template specialization providing the necessary type support.

wjwwood:
I get that but what are you doing which lead you to need to do so? We would need to do this for type masquerading for example, but I can't imagine what you're doing where this is a necessary change.

The changes lgtm.

esteve:
I don't see why either, I like the `::SharedPtr` idiom. Is the current behavior preventing anything on your side to work properly?

gbiggs:
Admittedly, I haven't been able to do any real programming for about four years now [sad face, lots of tears], but I thought that the ::SharedPtr idiom was considered good practice.

wjwwood:
Well he's not removing the option to use that idiom, but you can't pass C structs through the pipeline, for example, if you require an internal typedef like that.

esteve:
I see, if it's needed for C then +1

gbiggs:
Makes sense. So using std::shared_ptr would make the code usable with a wider range of types.

dirk-thomas:
Yes, I want to pass a type which doesn't have that.

And since we already define the callback with the plain std::shared_ptr type (https://github.com/ros2/rclcpp/blob/12859aca2cab6b3f264e38e597ea99b7f40b0c94/rclcpp/include/rclcpp/any_subscription_callback.hpp#L43) using the subtype here is unnecessary restricting.

PR #163, opened 2015-11-30
Move scope exit to include folder

jacquelinekay:
In order to write an Executor in a library outside of rclcpp, scope_exit needs to be installed to the include directory.

The fact that it wasn't already doing this suggests to me that there is a problem with this approach (possibly on Windows)... @wjwwood, any comments?


dirk-thomas:
The `scope_exit.hpp` header seems to be removed instead of being moved.

wjwwood:
That should be fine, I did it this way to limit the files scope to the source folder because I didn't need it in any headers at the time.

jacquelinekay:
CI failures are unrelated:

http://ci.ros2.org/job/ci_linux/640/
http://ci.ros2.org/job/ci_osx/522/
http://ci.ros2.org/job/ci_windows_connext_dynamic/17/
http://ci.ros2.org/job/ci_windows_connext_static/17/
http://ci.ros2.org/job/ci_windows_opensplice/705/

tfoote:
+1

dirk-thomas:
Please make sure to squash the first and third commit to keep the history.

PR #164, opened 2015-12-01
Replace shared pointers in AnyExecutable with raw pointers

jacquelinekay:
In progress.

Potential fix for the weird pointer corruption bugs in ros2/system_tests#72 (still testing).

I broke AnyExecutable into a State class so that I can use a pure POD type for the lock-free queue in LockFreeExecutor. I can revert this part of the PR and create the class in LockFreeExecutor if need be.

### DIFF
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -26,8 +26,22 @@ namespace rclcpp
 namespace executor
 {
 
-struct AnyExecutable
+struct AnyExecutableState
 {
+  // Only one of the following pointers will be set.
+  rclcpp::subscription::SubscriptionBase * subscription;
+  rclcpp::subscription::SubscriptionBase * subscription_intra_process;
+  rclcpp::timer::TimerBase * timer;
+  rclcpp::service::ServiceBase * service;
+  rclcpp::client::ClientBase * client;
+  // These are used to keep the scope on the containing items
+  rclcpp::callback_group::CallbackGroup * callback_group;
+  rclcpp::node::Node * node;
+};
+
+class AnyExecutable
+{
+public:
   RCLCPP_SMART_PTR_DEFINITIONS(AnyExecutable);
 
   RCLCPP_PUBLIC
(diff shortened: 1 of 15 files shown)


jacquelinekay:
In the future, execute_client, execute_service, and execute_timer will either need to accept a const shared pointer to be truly thread safe, or we need to add locking mechanisms to Client, Service and Timer to make sure the functions called in execution are thread safe.

rclcpp/include/rclcpp/any_executable.hpp
```diff
@@ -26,8 +26,22 @@ namespace rclcpp
...
 
-struct AnyExecutable
+struct AnyExecutableState
 {
+  // Only one of the following pointers will be set.
+  rclcpp::subscription::SubscriptionBase * subscription;
+  rclcpp::subscription::SubscriptionBase * subscription_intra_process;
+  rclcpp::timer::TimerBase * timer;
+  rclcpp::service::ServiceBase * service;
+  rclcpp::client::ClientBase * client;
```
dirk-thomas: How can we ensure that the pointers stored in here are still valid when the executable is actually being processed? The instance might have been deleted by then.

tfoote:
Ideas discussed today:

* We can only put POD structures into lock-free queues.  That's how we got to using raw pointers here.
* Storing raw pointers is dangerous; we could consider some kind of mock object instead, but still have a problem of managing ownership and storage of the underlying data.
* We're currently not queuing work, but rather relying on DDS to manage it.  It's not ideal to introduce a queue on our side, irrespective of whether it's lock-free or not.

To make a decision on how to proceed here, we need to get a better idea of the requirements.  @jaquelinekay, please write up a proposal for what is and what is not supported in a real-time-safe / lock-free manner, e.g.:

* are message-based callbacks supported?
 * vs. just timers, similar to how OROCOS RTT works: a real-time-safe component can only ask to be invoked at a given frequency; when it wakes up, it is able to check for work
* are multiple threads operating on a single subscriber supported?
 * vs. just single-threaded executors for real-time-safe work (could have implications on design of executor to allow, for example, a node with one real-time subscription and one non-real-time subscription)

jacquelinekay:
closing this set of changes, going to put the comments in a different issue

PR #166, opened 2015-12-02
Store handle of rmw_waitset_t in Executor

jacquelinekay:
Connects to ros2/rmw#49

TODO:

- [ ] Expose `max_conditions` argument to the user-facing API in a graceful way.
- [ ] Synchronization.

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -44,6 +44,24 @@ namespace executor
  */
 enum FutureReturnCode {SUCCESS, INTERRUPTED, TIMEOUT};
 
+///
+/**
+ * Options to be passed to the executor constructor.
+ */
+struct ExecutorArgs
+{
+  memory_strategy::MemoryStrategy::SharedPtr memory_strategy;
+  size_t max_conditions = 0;
+};
+
+static inline ExecutorArgs create_default_executor_arguments()
+{
+  ExecutorArgs args;
+  args.memory_strategy = memory_strategies::create_default_strategy();
+  args.max_conditions = 0;
+  return args;
+}
+
 /// Coordinate the order and timing of available communication tasks.
 /**
  * Executor provides spin functions (including spin_node_once and spin_some).
@@ -62,8 +80,7 @@ public:
   /// Default constructor.
   // \param[in] ms The memory strategy to be used with this executor.
   RCLCPP_PUBLIC
-  explicit Executor(
-    memory_strategy::MemoryStrategy::SharedPtr ms = memory_strategies::create_default_strategy());
+  explicit Executor(const ExecutorArgs & args = create_default_executor_arguments());
 
   /// Default destructor.
   RCLCPP_PUBLIC
@@ -262,8 +279,13 @@ protected:
   /// Spinning state, used to prevent multi threaded calls to spin and to cancel blocking spins.
   std::atomic_bool spinning;
 
+  rmw_guard_conditions_t fixed_guard_conditions_;
+
   /// Guard condition for signaling the rmw layer to wake up for special events.
   rmw_guard_condition_t * interrupt_guard_condition_;
+
+  /// Waitset for managing entities that the rmw layer waits on.
+  rmw_waitset_t * waitset_;
 
   /// The memory strategy: an interface for handling user-defined memory allocation strategies.
   memory_strategy::MemoryStrategy::SharedPtr memory_strategy_;
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -19,15 +19,57 @@
 
 using rclcpp::executor::AnyExecutable;
 using rclcpp::executor::Executor;
+using rclcpp::executor::ExecutorArgs;
 
-Executor::Executor(rclcpp::memory_strategy::MemoryStrategy::SharedPtr ms)
-: spinning(false), interrupt_guard_condition_(rmw_create_guard_condition()),
-  memory_strategy_(ms)
+Executor::Executor(const ExecutorArgs & args)
+: spinning(false),
+  memory_strategy_(args.memory_strategy)
 {
+  interrupt_guard_condition_ = rmw_create_guard_condition();
+  if (!interrupt_guard_condition_) {
+    throw std::runtime_error("Failed to create interrupt guard condition in Executor constructor");
+  }
+
+  // The number of guard conditions is fixed at 2: 1 for the ctrl-c guard cond,
+  // and one for the executor's guard cond (interrupt_guard_condition_)
+  // These guard conditions are permanently attached to the waitset.
+  const size_t number_of_guard_conds = 2;
+  fixed_guard_conditions_.guard_condition_count = number_of_guard_conds;
+  fixed_guard_conditions_.guard_conditions = static_cast<void **>(guard_cond_handles_.data());
+
+  // Put the global ctrl-c guard condition in
+  assert(fixed_guard_conditions_.guard_condition_count > 1);
+  fixed_guard_conditions_.guard_conditions[0] = \
+    rclcpp::utilities::get_global_sigint_guard_condition()->data;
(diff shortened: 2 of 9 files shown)


rclcpp/src/rclcpp/executor.cpp
```diff
@@ -22,12 +22,21 @@ using rclcpp::executor::Executor;
 
 Executor::Executor(rclcpp::memory_strategy::MemoryStrategy::SharedPtr ms)
 : spinning(false), interrupt_guard_condition_(rmw_create_guard_condition()),
+  waitset_(rmw_create_waitset(0)),
```
wjwwood: Can this function fail (return `nullptr`)? Should its value be checked in the constructor?

jacquelinekay: rmw_create_guard_condition can also return nullptr.

I will change the constructor to check both values and throw an error if it failed.

wjwwood: Yup, that's a good point.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -22,12 +22,27 @@ using rclcpp::executor::Executor;
...
   memory_strategy_(ms)
 {
+  if (!waitset_) {
+    throw std::runtime_error("Failed to create waitset in Executor constructor");
+  }
+  if (!interrupt_guard_condition_) {
+    throw std::runtime_error("Failed to create interrupt gaurd condition in Executor constructor");
+  }
```
dirk-thomas: This will leak the successfully allocated memory.

jacquelinekay: how does 99c3ca4cebc2703fd6cd9ba19aadbc9fabf4b816 look?

dirk-thomas: Better, but when the `waitset` is valid but not the `interrupt_guard_condition_` it still leaks the `waitset`.

wjwwood: I believe you also need to conditionally destroy the `waitset_` if the `interrupt_guard_condition_` member ends up being `nullptr`.

jacquelinekay: ok, I decided to not use the initializer list for the rmw objects to keep this logic simpler and shorter.

dirk-thomas:
I just committed a simpler condition - same as a few lines above.

wjwwood:
I know it's not in review yet, but the changes so far lgtm now.

jacquelinekay:
with regards to exposing the `max_conditions` variable, I'm thinking this is a good time to add an ExecutorArgs struct to pass as the single argument to the constructor of Executor (and a default factory function).

Alternatively I could make the max_conditions variable an argument to MemoryStrategy and read it out of the MemoryStrategy when constructing the rmw_waitset, which makes more sense conceptually (max_conditions specifies a limit on allocation) but is more indirect in the implementation.

jacquelinekay:
On synchronization:

I made a test case for these changes in branch `waitset_handle` of `system_tests` to test 2 executors spinning in separate threads. Predictably, the test case segfaults occasionally (after ~50-100 iterations) on detach_condition, because the two threads share the global interrupt guard condition handle.

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff5d445a0 in c_iterTake () from /usr/lib/libddskernel.so
(gdb) bt
#0  0x00007ffff5d445a0 in c_iterTake () from /usr/lib/libddskernel.so
#1  0x00007ffff5cbc3e5 in _ConditionRemoveWaitset () from /usr/lib/libddskernel.so
#2  0x00007ffff5c9ca1d in gapi_waitSet_detach_condition () from /usr/lib/libddskernel.so
#3  0x00007ffff7bc0ab6 in atexit_t::~atexit_t (this=0x7fffffffd150, __in_chrg=<optimized out>)
    at /home/jackie/code/ros2_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_wait.cpp:94
#4  0x00007ffff7bc175a in rmw_wait (subscriptions=0x7fffffffd2f0, guard_conditions=0x7fffffffd320, 
    services=0x7fffffffd300, clients=0x7fffffffd310, waitset=0x7e19fa0, wait_timeout=0x7fffffffd330)
    at /home/jackie/code/ros2_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_wait.cpp:325
#5  0x00007ffff6f25614 in rclcpp::executor::Executor::wait_for_work (this=0x7fffffffd520, timeout=...)
    at /home/jackie/code/ros2_ws/src/ros2/rclcpp/rclcpp/src/rclcpp/executor.cpp:397
```

There are a few possible ways of fixing this:

1. Make rmw_wait single-threaded using a global mutex. This is really inefficient and I'd prefer not to do it.
2. Make attaching/detaching conditions to the waitset globally single-threaded using atomics in rmw, which is slightly less inefficient and more complicated to implement.
3. Only attach/detach the global interrupt guard condition to the waitset once per Executor (in constructor and destructor). We would also have to disallow concurrent construction/destruction of Executors, and we would also have to enforce that a node can only belong to one Executor. We would also have to figure out a way to ignore detaching the global interrupt guard condition in rmw_wait, which could be tricky.

jacquelinekay:
I tried implementing approach 3 as described above to synchronizing access to the shared guard condition and it fixes the segfault in the multiple executors test case. I still haven't done anything to synchronize Executor construction/destruction.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -44,6 +44,24 @@ namespace executor
  */
 enum FutureReturnCode {SUCCESS, INTERRUPTED, TIMEOUT};
 
+///
+/**
+ *
+ */
```
dirk-thomas: Comment or no comment :wink:

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -19,15 +19,56 @@
...
+  fixed_guard_conditions_.guard_conditions = static_cast<void **>(guard_cond_handles_.data());
+
+  // Put the global ctrl-c guard condition in
+  assert(fixed_guard_conditions_.guard_condition_count > 1);
+  fixed_guard_conditions_.guard_conditions[0] = \
+    rclcpp::utilities::get_global_sigint_guard_condition()->data;
+  // Put the executor's guard condition in
+  fixed_guard_conditions_.guard_conditions[1] = interrupt_guard_condition_->data;
+
+  // Copy the guard condition handles object to keep it in scope.
```
dirk-thomas: I don't understand the comment. The guard conditions are stored in the member variable `fixed_guard_conditions_`. Nothing gets copied here?

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -19,15 +19,56 @@
...
+
+  if (!waitset_) {
+    fprintf(stderr,
+      "[rclcpp::error] failed to create waitset: %s\n", rmw_get_error_string_safe());
+    rmw_ret_t status = rmw_destroy_guard_condition(interrupt_guard_condition_);
+    if (status != RMW_RET_OK) {
+      fprintf(stderr,
+        "[rclcpp::error] failed to destroy guard condition: %s\n", rmw_get_error_string_safe());
+    }
+    throw std::runtime_error("Failed to create waitset in Executor constructor");
```
dirk-thomas: This seems to leak already allocated memory?

jacquelinekay: Which memory gets leaked? The `interrupt_guard_condition` gets deallocated if `rmw_destroy_guard_condition` is successful in line 53, and if `waitset_` is null, then `rmw_create_waitset` did not allocate any memory, or deallocated all memory it allocated (assuming I implemented it correctly).

dirk-thomas: You are right. I misunderstood where the memory from `guard_cond_handles_.fixed_guard_conditions_.guard_conditions` comes from. All good.

PR #167, opened 2015-12-03
Add RCLCPP_PUBLIC to Context constructor

jacquelinekay:
Potential fix for currently broken Windows build.

I use a local Context in the multithreaded test, to make sure the IntraProcessManager gets cleaned up between test cases.

The Context constructor does not use the RCLCPP_PUBLIC macro.

I believe we previously didn't test this constructor anywhere in user code.

### DIFF
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -37,6 +37,7 @@ class Context
 public:
   RCLCPP_SMART_PTR_DEFINITIONS(Context);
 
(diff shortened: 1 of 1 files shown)


jacquelinekay:
http://ci.ros2.org/job/ci_windows_opensplice/734/

wjwwood:
+1 for the approach of the fix.

esteve:
+1 if the CI job passes.

dirk-thomas:
The build passed.

PR #170, opened 2015-12-08
Remove extraneous mutex

jacquelinekay:
I've determined this mutex is unnecessary because the shared state accessed by the `store_intra_process_message_` callback is locked by mutexes internal to `IntraProcessManager`.

I've run the relevant multithreaded pub/sub intra-process test case over 600 times with Opensplice without seeing a segfault.


wjwwood:
:tada: lgtm

esteve:
:hamburger: +1

jacquelinekay:
yum, burgers.

jacquelinekay:
Take the burgers back. I have tested this some more and seen deadlock after a few hundred iterations. I will investigate to see if this was a problem before this commit and revert it if it was.

PR #171, opened 2015-12-08
Lock mapped_ring_buffer

jacquelinekay:
Possible fix for deadlock introduced by deleting a mutex in Publisher.

some code complexity is introduced, but the mutex is delegated to a subclass of MappedRingBufferBase. An alternative lock-free implementation could use a lock-free data structure instead of the default MappedRingBuffer implementation.

### DIFF
--- a/rclcpp/include/rclcpp/mapped_ring_buffer.hpp
+++ b/rclcpp/include/rclcpp/mapped_ring_buffer.hpp
@@ -19,6 +19,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <mutex>
 #include <vector>
 
 #include "rclcpp/allocator/allocator_common.hpp"
@@ -97,6 +98,7 @@ public:
   void
   get_copy_at_key(uint64_t key, ElemUniquePtr & value)
   {
+    std::lock_guard<std::mutex> lock(data_mutex_);
     auto it = get_iterator_of_key(key);
     value = nullptr;
     if (it != elements_.end() && it->in_use) {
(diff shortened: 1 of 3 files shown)


jacquelinekay:
I'm launching two sets of CI for this changeset, one normal and one that will repeat only the multithreaded tests until failure.

jacquelinekay:
On Linux and OSX, the multithreaded jobs pass:
http://ci.ros2.org/job/ci_linux/692/
http://ci.ros2.org/job/ci_linux/693/
http://ci.ros2.org/job/ci_osx/568/
http://ci.ros2.org/job/ci_osx/569/

There is a warning due to an unused variable on OSX that I fixed in the last commit.

I had to add an additional header for Windows to compile, so I've relaunched the Windows jobs.

dirk-thomas:
Were the tests previously reliably failing when running the tests 20 times?

jacquelinekay:
On OSX, the multithreaded test for Connext (static and dynamic) has failed in the nightly jobs for the past few days:

http://ci.ros2.org/view/nightly/job/nightly_osx/138/testReport/
http://ci.ros2.org/view/nightly/job/nightly_osx/137/testReport/
http://ci.ros2.org/view/nightly/job/nightly_osx/136/testReport/
http://ci.ros2.org/view/nightly/job/nightly_osx/135/testReport/

The multithreaded test for Opensplice on Windows always fails. I don't have the report from Jenkins for this build yet, however.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -196,6 +201,9 @@ class MappedRingBuffer : public MappedRingBufferBase
   bool
   has_key(uint64_t key)
   {
+    std::lock_guard<std::mutex> lock(data_mutex_);
+    bool did_replace = elements_[head_].in_use;
+    (void)did_replace;
```
dirk-thomas: I see that you added the (void) line to avoid the compiler warning. But why not remove the variable if it is not used? Or does the above line have any side effects?

dirk-thomas:
This looks like a good improvement then: +1

PR #172, opened 2015-12-08
refactor typesupport

dirk-thomas:
Connect to ros2/rmw_implementation#9

PR #174, opened 2015-12-11
inline methods to avoid multiple definition linking problems

tfoote:
These were causing multiple definitions when building a library and an executable which used rclcpp.


dirk-thomas:
Please don't merge this since it collides with #172.

tfoote:
After talking with @dirk-thomas this is only a partial solution and it breaks other use cases. It will need the full work in #172

Issue #173, opened 2015-12-11
subscriber callbacks bound to class methods do not compile


tfoote:
It fails due to an inability to resolve an overloaded function. 

```
/home/tfoote/work/ros2/tf3/install/include/rclcpp/function_traits.hpp:51:85: error: decltype cannot resolve address of overloaded function
       typename function_traits<decltype( & FunctionT::operator())>::arguments>::type;
```

The full error is here: https://gist.github.com/tfoote/17222805a01e7c05c3f1

jacquelinekay:
Can you make this a test case in system_tests/test_rclcpp instead?

jacquelinekay:
Are you working on a fix for this right now? I will take a look if you're not actively working on it.

tfoote:
@esteve found a workaround to use a lambda function. I'll add a unit test for it too.

jacquelinekay:
You mean using a lambda function in the example? Or to make bound functions work as the input to create_subscription?

None of our tested rclcpp code so far uses `std::bind`. Any code that captures arguments does so using lambdas.

Playing around with this, I found another "workaround" that compiles:
```
std::function<void(const std_msgs::msg::String::SharedPtr)> callback = std::bind(&Chatter::chatterCallback, &chatter, std::placeholders::_1);
  auto sub = node->create_subscription<std_msgs::msg::String>(
    "chatter", callback, rmw_qos_profile_default);
```

Ideally we would support any valid callback signature for create_subscription, so I think we should do the necessary work to make your example compile.

esteve:
@tfoote could you have a look at this branch and see if it works? Thanks.

https://github.com/ros2/rclcpp/tree/std_bind_function_traits

esteve:
@tfoote I got your example compiling with my branch, I'll clean up the changes here and submit them for review.

tfoote:
@esteve I can confirm that your branch works for me, but I see it's not a cross platform solution unfortunately :-(

esteve:
@tfoote I got the workaround to compile on all three platforms in #183, I'll see if I can compile the `std::bind` test in ros2/system#88

tfoote:
fixed in #183

PR #175, opened 2015-12-11
Add support for std::bind

esteve:
This PR adds support for `std::bind` in `function_traits`

`std::_Bind<std::_Mem_fn<...>>` should be replace with `decltype(std::bind(...))`, but I couldn't get that to compile.

Connects to #173

### DIFF
--- a/rclcpp/include/rclcpp/function_traits.hpp
+++ b/rclcpp/include/rclcpp/function_traits.hpp
@@ -77,6 +77,12 @@ template<typename ReturnTypeT, typename ... Args>
 struct function_traits<ReturnTypeT (*)(Args ...)>: function_traits<ReturnTypeT(Args ...)>
 {};
 
+// std::bind
+template<typename ClassT, typename ReturnTypeT, typename ... Args, typename ... FArgs>
+struct function_traits<std::_Bind<std::_Mem_fn<ReturnTypeT (ClassT::*)(Args ...)>(FArgs ...)>>
+: function_traits<ReturnTypeT(Args ...)>
+{};
+
 // Lambdas
 template<typename ClassT, typename ReturnTypeT, typename ... Args>
 struct function_traits<ReturnTypeT (ClassT::*)(Args ...) const>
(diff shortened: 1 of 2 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/708/
http://ci.ros2.org/job/ci_osx/586/
http://ci.ros2.org/job/ci_windows_opensplice/777/

esteve:
I've only triggered the jobs for OpenSplice because this is specific to the rclcpp API.

esteve:
As I expected, this didn't compile on OSX because it's using `std::_Bind<std::_Mem_fn>>`, which is the internal implementation of `std::bind` in GCC.

esteve:
Can't get a generalized version of this working, I think some combination of `decltype` with `declval` may work, but I agree with @jacquelinekay 's approach of wrapping `std::bind` in `std::function` as a workaround, or use a lambda.

jacquelinekay:
Did you try playing around with `std::is_bind_expression`? http://en.cppreference.com/w/cpp/utility/functional/is_bind_expression

esteve:
Yeah, but the problem with `std::bind` is that it's polymorphic, so you don't know what `std::bind()` produces, each compiler returns a different type (`std::_Bind` on GCC, `std::__1:__bind` on Clang) because it's not specified in the C++ standard. In general, it seems that lambdas are preferred over `std::bind`, Scott Meyer recommends using the former:

http://scottmeyers.blogspot.cz/2013/01/effective-c11-content-and-status.html

> Lambda Expressions
>
>    Prefer lambdas to std::bind.
>    Prefer lambdas to variadic arguments for threading functions
>    Beware default captures in member functions.

jacquelinekay:
Hmm. It would still be nice to support `bind` for migration purposes, since ROS 1 package authors will expect bound functions to work for callback like Tully did. An alternative solution would be to change the templating for `create_subscription` such that there is a specialization for a callback of type `std::function<T>` and a specialization for function pointers (void*(Args...) or whatever), such that the compiler would be forced to convert the bound function to std::function (basically automating the workaround I suggested above).

esteve:
Good idea, that'd be a very useful feature.

esteve:
I gave it some thought and I found a solution to make this work across different compilers/c++ standard libraries.

esteve:
I can't seem to reopen this PR, I'll submit a new one.

PR #177, opened 2015-12-13
remove vestigial file and cmake logic

wjwwood:
This is left over from the early node component demos and is no longer used.


dirk-thomas:
We will hopefully get towards "nodelets" in the next spring. Should we keep it for that or better readd out then?

tfoote:
+1 this is just a small mostly boilerplate file. I don't think that it has anything in it that we cannot reproduce easily.

wjwwood:
I think we should remove it since some of it has already been removed (the macro that nodes are supposed to use, as well as the examples which use it).

PR #176, opened 2015-12-12
inline helper functions in rclcpp.hpp to avoid multiple definitions

tfoote:
There are a few functions which need inlining from #174 not caught in the refactor.  This adds them to the refactor branch. This extends #172

PR #178, opened 2015-12-14
inline helper functions in rclcpp.hpp to avoid multiple definitions

tfoote:
Retargeting #176 against master
CI:

http://ci.ros2.org/job/ci_linux/716/
http://ci.ros2.org/job/ci_osx/593/
http://ci.ros2.org/job/ci_windows_opensplice/782/
http://ci.ros2.org/job/ci_windows_connext_static/57
http://ci.ros2.org/job/ci_windows_connext_dynamic/47

### DIFF
--- a/rclcpp/include/rclcpp/rclcpp.hpp
+++ b/rclcpp/include/rclcpp/rclcpp.hpp
@@ -28,33 +28,33 @@
 #include "rclcpp/visibility_control.hpp"
 
 // NOLINTNEXTLINE(runtime/int)
-const std::chrono::seconds operator"" _s(unsigned long long s)
+inline const std::chrono::seconds operator"" _s(unsigned long long s)
 {
   return std::chrono::seconds(s);
 }
-const std::chrono::nanoseconds operator"" _s(long double s)
+inline const std::chrono::nanoseconds operator"" _s(long double s)
 {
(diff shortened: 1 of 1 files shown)


dirk-thomas:
My previous comment still applies: why making the functions inline instead of moving the implementation to a .cpp file?

tfoote:
This is the simplest solution to fix the problem of over linking which is blocking my other progress.

jacquelinekay:
It's a bit tricky because these functions aren't namespaced. I started a branch called helper_functions_cpp that follows Dirk's suggestion and I'm trying to build it on Windows right now. Will let you know how it goes:

https://github.com/ros2/rclcpp/compare/helper_functions_cpp

gerkey:
@tfoote Double-check why the return values are const. Otherwise, +1

wjwwood:
I'm not sure if they need to be const, but they are returning a literal converted into different type, so it's really a different view of a literal, which should always be const. In the C++14 implementation of the time literals they actually use `constexpr`:

http://en.cppreference.com/w/cpp/chrono/operator%22%22s

I believe I was doing that until I realized it didn't work right on Windows, but it might now.

tfoote:
Thanks for explaining the reason for const @wjwwood This is something we should plan to replace once c++14 is available with the full implementation so I'll leave it as const for now.

PR #183, opened 2015-12-15
Added support for std::bind (redux)

esteve:
This PR adds support for std::bind in function_traits

Extends #175 

Connects to #173

PR #184, opened 2015-12-16
Added support for free functions and std::bind

esteve:
This adds support for free functions, which were missing in the previous `std::bind` PR (https://github.com/ros2/rclcpp/pull/183)

### DIFF
--- a/rclcpp/include/rclcpp/function_traits.hpp
+++ b/rclcpp/include/rclcpp/function_traits.hpp
@@ -78,7 +78,7 @@ template<typename ReturnTypeT, typename ... Args>
 struct function_traits<ReturnTypeT (*)(Args ...)>: function_traits<ReturnTypeT(Args ...)>
 {};
 
-// std::bind
+// std::bind for object methods
 template<typename ClassT, typename ReturnTypeT, typename ... Args, typename ... FArgs>
 #if defined _LIBCPP_VERSION  // libc++ (Clang)
 struct function_traits<std::__1::__bind<ReturnTypeT (ClassT::*)(Args ...), FArgs ...>>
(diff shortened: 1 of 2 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/740/
http://ci.ros2.org/job/ci_osx/621/
http://ci.ros2.org/job/ci_windows/811/

esteve:
The new tests in `test_function_traits.cpp` pass:

http://ci.ros2.org/job/ci_linux/740/testReport/%28root%29/TestFunctionTraits/
http://ci.ros2.org/job/ci_osx/621/testReport/%28root%29/TestFunctionTraits/
http://ci.ros2.org/job/ci_windows/812/testReport/%28root%29/TestFunctionTraits/

rclcpp/include/rclcpp/function_traits.hpp
```diff
@@ -92,6 +92,18 @@ struct function_traits<
   : function_traits<ReturnTypeT(Args ...)>
 {};
 
+// std::bind for free functions
+template<typename ReturnTypeT, typename ... Args, typename ... FArgs>
+#if defined _LIBCPP_VERSION  // libc++ (Clang)
```
jacquelinekay: I'm not so sure about this approach because it restricts our portability between compilers. The return type of std::bind doesn't have the same name in the implementation for these three compilers because it's not specified in the C++11 standard, right? That implies to me that we should try to come up with an approach that conforms better to the standard. I know we said we're only supporting these three major compilers, but it still feels wrong to me.

rclcpp/include/rclcpp/function_traits.hpp
```diff
@@ -92,6 +92,18 @@ struct function_traits<
...
+struct function_traits<std::__1::__bind<ReturnTypeT( &)(Args ...), FArgs ...>>
+#elif defined __GLIBCXX__  // glibc++ (GNU C++)
+struct function_traits<std::_Bind<ReturnTypeT(*(FArgs ...))(Args ...)>>
+#elif defined _MSC_VER  // MS Visual Studio
+struct function_traits<std::_Binder<std::_Unforced, ReturnTypeT(__cdecl &)(Args ...), FArgs ...>>
+#endif
```
dirk-thomas: This should also have an `#else` which contains `#error`. Otherwise this is invalid syntax and will result in a very difficult error message.

rclcpp/include/rclcpp/function_traits.hpp
esteve: Yeah, the C++11 spec doesn't define the return value of a `std::bind` call:

http://en.cppreference.com/w/cpp/utility/functional/bind

I agree with you, but I couldn't come up with a different approach. I tried using `decltype` and `declval` so that it was compiler-independent, but couldn't get it to work. This is the least intrusive way I can think of supporting `std::bind` across compilers.

rclcpp/include/rclcpp/function_traits.hpp
esteve: Done.

jacquelinekay:
Well, I thought about it about std::bind for a while and couldn't come up with anything better.

+1

PR #182, opened 2015-12-15
Add unit tests for rclcpp Rate objects

tfoote:
Connects to #181 

First adding unit tests for coverage. Next debug problem. 

The problem from 181 is reproduced in these unit tests: 

CI: http://ci.ros2.org/job/ci_osx/602/

```
08:42:55 In file included from /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/test/test_rate.cpp:19:
08:42:55 /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/rate.hpp:74:20: error: no viable overloaded '+='
08:42:55     last_interval_ += period_;
08:42:55     ~~~~~~~~~~~~~~ ^  ~~~~~~~
08:42:55 /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/test/test_rate.cpp:32:5: note: in instantiation of member function 'rclcpp::rate::GenericRate<std::__1::chrono::system_clock>::sleep' requested here
08:42:55   r.sleep();
08:42:55     ^
08:42:55 /Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/chrono:782:79: note: candidate function not viable: no known conversion from 'duration<[...], ratio<[...], 1000000000>>' to 'const duration<[...], ratio<[...], 1000000>>' for 1st argument
08:42:55     __attribute__ ((__visibility__("hidden"), __always_inline__)) time_point& operator+=(const duration& __d) {__d_ += __d; return *this;}
08:42:55                                                                               ^
08:42:55 In file included from /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/test/test_rate.cpp:19:
08:42:55 /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/rate.hpp:81:24: error: no viable overloaded '='
08:42:55         last_interval_ = now + period_;
08:42:55         ~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~
08:42:55 /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/test/test_rate.cpp:32:5: note: in instantiation of member function 'rclcpp::rate::GenericRate<std::__1::chrono::system_clock>::sleep' requested here
08:42:55   r.sleep();
08:42:55     ^
[...]

PR #185, opened 2015-12-16
different fix for the rate failing

wjwwood:
@tfoote


tfoote:
Thanks that looks good. I'll pull this in and squash it into #182 replacing my fix.

PR #186, opened 2015-12-17
remove excess sleep

tfoote:
CI: 
- http://ci.ros2.org/job/ci_linux/760 (expected)
- http://ci.ros2.org/job/ci_osx/640/
- http://ci.ros2.org/job/ci_windows/825/

 identified in #169

PR #187, opened 2015-12-17
Avoid reinitializing allocator

jacquelinekay:
Connects to ros2/realtime_support#16

Correct an error message and pass the same allocator to the `AllocSet` (a set of uint64_t) that is stored in a map in IntraProcessManager. This is to avoid reinitializing an allocator on each call to store_intra_process_message. This is useful if the default constructor of an allocator has to preallocate memory, but the copy constructor of an allocator does not.

### DIFF
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -142,7 +142,7 @@ public:
     std::lock_guard<std::mutex> lock(runtime_mutex_);
     auto it = publishers_.find(intra_process_publisher_id);
     if (it == publishers_.end()) {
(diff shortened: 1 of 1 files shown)


jacquelinekay:
I'm also sneaking in the removal of an extraneous sleep in MultiThreadedExecutor pointed out by @tfoote earlier.

wjwwood:
@jacquelinekay does that not overlap with https://github.com/ros2/rclcpp/pull/186?

jacquelinekay:
sorry, I didn't see that!

wjwwood:
It's ok.

tumblr_n2i13gz4ts1r70ozfo4_500

jacquelinekay:
I'll remove the commit that overlaps with #186

gerkey:
+1

PR #191, opened 2015-12-31
fix rate test

gerkey:
Fixes failures like this:

http://ci.ros2.org/view/nightly/job/nightly_osx/163/testReport/%28root%29/projectroot/test_rate/

The problem was that we were requiring the time elapsed between the first and second calls to sleep to be greater than or equal to the period. That's not guaranteed, in particular if the first call to sleep ran long. Instead we should check that the time elapsed between the creation of the rate object and the nth call to sleep is greater than or equal to n periods.

With that fix, we no longer need to allow `epsilon` underrun (because underrun should **never** occur on a properly working timer).

While I was in there, I updated the `WallRate` tests to use the `steady_clock` instead of the `system_clock`, because that's what's being used in the underlying classes (in fact, it's the only difference between `Rate` and `WallRate`).

### DIFF
--- a/rclcpp/test/test_rate.cpp
+++ b/rclcpp/test/test_rate.cpp
@@ -39,9 +39,9 @@ TEST(TestRate, rate_basics) {
   rclcpp::utilities::sleep_for(offset);
   ASSERT_TRUE(r.sleep());
   auto two = std::chrono::system_clock::now();
-  delta = two - one;
-  ASSERT_TRUE(period < delta + epsilon);
-  ASSERT_TRUE(period * overrun_ratio > delta);
+  delta = two - start;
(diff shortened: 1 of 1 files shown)


gerkey:
CI:

http://ci.ros2.org/job/ci_linux/834/
http://ci.ros2.org/job/ci_osx/694/
http://ci.ros2.org/job/ci_windows/887/

wjwwood:
lgtm

tfoote:
+1

gerkey:
New OSX job (original interrupted by apparent reboot):

http://ci.ros2.org/job/ci_osx/696/

And I fixed style in 8188e48.

Issue #196, opened 2016-02-17
Feature / Question - List all parameters of a parameter server


firesurfer:
Is it possible to list all parameters of a parameter server ?
I created a workaround in https://github.com/firesurfer/ros2_components/blob/master/src/ros2_components/Robot.h#L119 by requesting about 20.000 possible parameter names. It would be nicer if it would be possible to get a list of all parameters

dirk-thomas:
The list parameters service call (https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/srv/ListParameters.srv) should provide that functionality. The service call is exposed as API in rclcpp as https://github.com/ros2/rclcpp/blob/8c47f6151bd5c68bd7b2a31e375c114ad8ba087e/rclcpp/include/rclcpp/parameter_client.hpp#L146-L150

firesurfer:
Yeah. But what to pass for as prefix if I want to list all parameters? I already tried a empty list or * but it didn't work as far as I remember.

dirk-thomas:
From looking at the implementation (https://github.com/ros2/rclcpp/blob/69f7bca85d523ee8647c2051651184023ad1c6ad/rclcpp/src/rclcpp/node.cpp#L236) I can't figure out how this is currently possible. Maybe @esteve can comment on this?

It would be good to also document the necessary arguments of this common use case in the .srv file.

gerkey:
ros2/rclcpp#197 adds the ability to list all parameters by supplying an empty prefix list. ros2/system_tests#107 adds a test for it.

dirk-thomas:
@gerkey Can you please run a CI job for it.

gerkey:
I'd like to get feedback on the approach. Is an empty prefix list and ignored depth a good way to list all parameters?

dirk-thomas:
A think returning all parameters when an empty prefix list is passes is good.

But the depth parameter should still be honored in that case.

gerkey:
Proposal: when the prefix list is empty, a depth of `D` means: return all parameters whose names contain `<=D` separators (currently the `.` character). Does that make sense?

dirk-thomas:
I don't think `<=D` makes sense. `D=1` should return everything *without* a separator - basically the first level. And of course also the parent group name for everything with a separator.

And I assume `depth=0` has a similar semantic as an empty prefix list. So return all parameters with an arbitrary number of separators.

firesurfer:
[...]

PR #197, opened 2016-02-18
List all parameters

gerkey:
Connects to ros2/rclcpp#196 and ros2/system_tests#107.

Add the ability to retrieve the list of all parameters by passing an empty list of prefixes. In that case, the depth argument is ignored. This is one way to do it; other ideas for calling semantics to get the desired behavior?

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -75,7 +75,7 @@ ament_export_include_directories(include)
 
 ament_export_libraries(${PROJECT_NAME})
 
-if(AMENT_ENABLE_TESTING)
+if(BUILD_TESTING)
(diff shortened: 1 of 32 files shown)


jacquelinekay:
The code looks fine to me.

I launched CI jobs:

http://ci.ros2.org/job/ci_linux/1042/
http://ci.ros2.org/job/ci_osx/859
http://ci.ros2.org/job/ci_windows/1103

gerkey:
I'm still intending to respond to the comments from #196 about handling the depth argument.

jacquelinekay:
I see, I haven't been following that discussion.

gerkey:
Replaced by the same change in #230.

PR #198, opened 2016-02-19
added remote node name to sync parameter client

firesurfer:
I added the remote node name to the constructor of the SyncParametersClient and passed it to the internal used AsyncParametersClient

### DIFF
--- a/rclcpp/src/rclcpp/parameter_client.cpp
+++ b/rclcpp/src/rclcpp/parameter_client.cpp
@@ -228,11 +228,11 @@ AsyncParametersClient::list_parameters(
 }
 
 SyncParametersClient::SyncParametersClient(
-  rclcpp::node::Node::SharedPtr node)
+  rclcpp::node::Node::SharedPtr node,const std::string & remote_node_name)
 : node_(node)
 {
   executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
-  async_parameters_client_ = std::make_shared<AsyncParametersClient>(node);
+  async_parameters_client_ = std::make_shared<AsyncParametersClient>(node, remote_node_name);
 }
 
 SyncParametersClient::SyncParametersClient(
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -120,7 +120,8 @@ public:
 
   RCLCPP_PUBLIC
   SyncParametersClient(
-    rclcpp::node::Node::SharedPtr node);
+    rclcpp::node::Node::SharedPtr node,
+    const std::string & remote_node_name = "");
 
   RCLCPP_PUBLIC
   SyncParametersClient(


rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -228,11 +228,11 @@ AsyncParametersClient::list_parameters(
 }
 
 SyncParametersClient::SyncParametersClient(
-  rclcpp::node::Node::SharedPtr node)
+  rclcpp::node::Node::SharedPtr node,const std::string & remote_node_name)
```
wjwwood: You need to have a single space after the `,`.

In the future you can run out linter tools by running the tests for `rclcpp` or by running `ament_cpplint path/to/rclcpp`, `ament_uncrustify path/to/rclcpp`, etc... manually. You have to do this in a terminal where you sourced the setup file after building.

wjwwood:
Thanks for the PR @firesurfer! That definitely looks like an oversight from our perspective.

I had one nitpick comment (which would fail our CI style checker). If you can fix that up then I'll trigger a CI job to test your changes.

tfoote:
This was a decision when we put this together to not support remote parameter server queries with the synchronous API since remote calls are inherently unreliable and possibly block indefinitely. For example it could block [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/parameter_client.cpp#L250) indefinitely if you lost connection or mistyped the node name. 

To add this support we would need to extend the synchronous API to add things like timeouts, to provide recovery behaviors that then get passed through to the asynchronous implementation under the hood.

We can revisit that decision, but our goal was to keep the simple case as simple as possible. And if you want to be querying for a remote parameter you can just use the async API and take care of your timeouts and other implications as you see fit.

wjwwood:
I didn't consider that @tfoote, maybe it should be called a `LocalSyncParameterClient` then? Is there a TODO somewhere to capture this? Is the idea that you manage your own "timeout" with the asynchronous version, i.e. canceling the request after a period of time? (there doesn't seem to be a way to cancel the asynchronous requests either, other than ignoring the result in your given callback)

Based on what @tfoote said, there may be more work involved before this could be merged.

tfoote:
I would expect that when we add documentation to the API that the local only nature of the API would be documented and clear. 

For the async API we just return the future and it's the developers job to hold it and wait for it to be complete. If they just want to block they can spin_until_future_complete. Otherwise we should probably add some generic multi future handling methods, but hopefully they can be generic and work for any future.

tfoote:
After discussing, the very minimum we need to make this change is a timeout parameter to the synchronous client calls. This timeout would be passed to the various `spin*` functions used to implement each of the synchronous calls.

For example, consider this function:

https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/parameter_client.cpp#L246-L257

You would need to pass a timeout as a parameter to that function and make sure that timeout was used in the spin call within that function.

PR #202, opened 2016-03-04
Fix flaky subscription and spinning test

wjwwood:
This pr does two things, first it adds a way to print out a meaningful string for a given FutureReturnCode (which I needed when debugging). You can now get an output like this:

```
/Users/william/ros2_ws/src/ros2/system_tests/test_rclcpp/test/test_subscription.cpp:80: Failure
Value of: future_ret
  Actual: INTERRUPTED (1)
Expected: FutureReturnCode::SUCCESS
Which is: SUCCESS (0)
```

Second it refactors the `executor.spin_until_future_complete` which was, imo, not working correctly before and which I needed when addressing some flaky-ness of one of the tests in `test_rclcpp`. Before this patch, I believe it would pass the full `timeout` value to `spin_once` even if it had spent some time spinning a previous time through the while loop. The changes to fix up the test (and use these changes) is forthcoming.

Connects to ros2/system_tests#108

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -22,6 +22,7 @@
 #include <iostream>
 #include <list>
 #include <memory>
+#include <string>
 #include <vector>
 
 #include "rclcpp/any_executable.hpp"
@@ -45,9 +46,11 @@ namespace executor
  */
 enum class FutureReturnCode {SUCCESS, INTERRUPTED, TIMEOUT};
 
+RCLCPP_PUBLIC
 std::ostream &
-operator << (std::ostream & os, const FutureReturnCode & future_return_code);
+operator<<(std::ostream & os, const FutureReturnCode & future_return_code);
 
+RCLCPP_PUBLIC
 std::string
 to_string(const FutureReturnCode & future_return_code);
 
@@ -166,7 +169,8 @@ public:
   /**
    * \param[in] executor The executor which will spin the node.
    * \param[in] node_ptr The node to spin.
-   * \param[in] future The future to wait on. If SUCCESS, the future is safe to access after this function
+   * \param[in] future The future to wait on. If SUCCESS, the future is safe to access after this
+     function.
    * \param[in] timeout Optional timeout parameter, which gets passed to Executor::spin_node_once.
      -1 is block forever, 0 is non-blocking.
      If the time spent inside the blocking loop exceeds this timeout, return a TIMEOUT return code.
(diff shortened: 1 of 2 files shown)


wjwwood:
This PR is attached to https://github.com/ros2/system_tests/pull/108, can I get some reviews of this?

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -177,23 +184,30 @@ class Executor
...
-        }
-      }
+      // Check if the future is set, return SUCCESS if it is.
       status = future.wait_for(std::chrono::seconds(0));
+      if (status == std::future_status::ready) {
+        return FutureReturnCode::SUCCESS;
+      }
+      // Otherwise check if we still have time to wait, return TIMEOUT if not.
+      auto elapsed_time = std::chrono::steady_clock::now() - start_time;
+      if (elapsed_time > timeout) {
```
dirk-thomas: Should we check for `>=` here?

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -177,23 +184,30 @@ class Executor
...
+        return FutureReturnCode::TIMEOUT;
+      }
+      // Subtract the elapsed time from the original timeout.
+      timeout -= std::chrono::duration_cast<std::chrono::duration<int64_t, TimeT>>(elapsed_time);
```
dirk-thomas: Why is `timeout` reduced here? Doesn't that make the loop compare wrongly on the next iteration of line 203?

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -569,3 +572,25 @@ Executor::get_next_executable(std::chrono::nanoseconds timeout)
...
+  using enum_type = std::underlying_type<FutureReturnCode>::type;
+  using std::string;
+  using std::to_string;
+  switch (future_return_code) {
+    case FutureReturnCode::SUCCESS:
+      return string("SUCCESS (" + to_string(static_cast<enum_type>(future_return_code)) + ")");
+    case FutureReturnCode::INTERRUPTED:
+      return string("INTERRUPTED (" + to_string(static_cast<enum_type>(future_return_code)) + ")");
+    case FutureReturnCode::TIMEOUT:
+      return string("TIMEOUT (" + to_string(static_cast<enum_type>(future_return_code)) + ")");
```
dirk-thomas: Needs `default`.

rclcpp/include/rclcpp/executor.hpp
jacquelinekay: For this comparison to work, don't you need to store the initial value of `elapsed_time`? Otherwise you could exit early.

rclcpp/include/rclcpp/executor.hpp
wjwwood: Yeah, it does. Let me refactor this real quick.

rclcpp/include/rclcpp/executor.hpp
wjwwood: Yeah, it's just wrong... I'm gonna refactor it again and write a test.

rclcpp/src/rclcpp/executor.cpp
wjwwood: Actually it doesn't since this is an Enum switch. The compiler will warn you if the enum is extended in the future that not all cases are covered.

wjwwood:
Ok, let's try this again!

I added a test for this particular function in which the last case tested failed with the code you guys tried to review before. I also refactored the function (again) and now it passes the tests. So this is ready for re-review.

dirk-thomas:
The algo lgtm now. +1

Issue #204, opened 2016-03-10
Refactor to use rcl


jacquelinekay:
rclcpp should be a wrapper around rcl (similar to rclpy).

I am going to start addressing this because I think it will help me understand how to design the rcl parameters API.

jacquelinekay:
I think I'm going to keep this open to ticket some things that weren't addressed by #207:

- [ ] Parameters. I'm going to start pushing parameters functionality into rcl, so another refactor will be needed in rclcpp once that is done.
- [ ] The beginnings of the graph API which are currently exposed through rmw (e.g. rmw_get_topic_names_and_types)

jacquelinekay:
nevermind, I'm going to close this in favor of more specific tickets

PR #207, opened 2016-03-13
Refactor to use rcl

jacquelinekay:
Connects to #204

TODO:

- [x] address problems linking to all rmw implementations
- [x] address intra process test failure
- [x] address multithreaded test failure
- [x] services hang (connext and connext dynamic only)

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -43,10 +43,13 @@ set(${PROJECT_NAME}_SRCS
 )
 
 macro(target)
+  if(NOT "${target_suffix} " STREQUAL " ")
+    get_rcl_information("${rmw_implementation}" "rcl${target_suffix}")
+  endif()
   add_library(${PROJECT_NAME}${target_suffix} SHARED
     ${${PROJECT_NAME}_SRCS})
   ament_target_dependencies(${PROJECT_NAME}${target_suffix}
-    "rcl"
+    "rcl${target_suffix}"
     "rcl_interfaces"
     "rmw"
     "rosidl_generator_cpp"
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -68,6 +68,7 @@ endmacro()
 call_for_each_rmw_implementation(target GENERATE_DEFAULT)
 
 ament_export_dependencies(ament_cmake)
+ament_export_dependencies(rcl)
 ament_export_dependencies(rcl_interfaces)
 ament_export_dependencies(rmw)
 ament_export_dependencies(rmw_implementation)
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -50,10 +50,7 @@ macro(target)
     ${${PROJECT_NAME}_SRCS})
   ament_target_dependencies(${PROJECT_NAME}${target_suffix}
     "rcl${target_suffix}"
-    "rcl_interfaces"
-    "rmw"
-    "rosidl_generator_cpp"
-    "${rmw_implementation}")
+    "rosidl_generator_cpp")
 
   # Causes the visibility macros to use dllexport rather than dllimport,
   # which is appropriate when building the dll but not consuming it.
@@ -72,9 +69,6 @@ call_for_each_rmw_implementation(target GENERATE_DEFAULT)
 
 ament_export_dependencies(ament_cmake)
 ament_export_dependencies(rcl)
-ament_export_dependencies(rcl_interfaces)
-ament_export_dependencies(rmw)
-ament_export_dependencies(rmw_implementation)
 ament_export_dependencies(rosidl_generator_cpp)
 
 ament_export_include_directories(include)
@@ -96,6 +90,7 @@ if(AMENT_ENABLE_TESTING)
   ament_add_gtest(test_mapped_ring_buffer test/test_mapped_ring_buffer.cpp)
   if(TARGET test_mapped_ring_buffer)
     target_include_directories(test_mapped_ring_buffer PUBLIC
+      ${rcl_INCLUDE_DIRS}
       ${rcl_interfaces_INCLUDE_DIRS}
       ${rmw_INCLUDE_DIRS}
       ${rosidl_generator_cpp_INCLUDE_DIRS}
@@ -104,6 +99,7 @@ if(AMENT_ENABLE_TESTING)
   ament_add_gtest(test_intra_process_manager test/test_intra_process_manager.cpp)
   if(TARGET test_intra_process_manager)
     target_include_directories(test_intra_process_manager PUBLIC
+      ${rcl_INCLUDE_DIRS}
       ${rcl_interfaces_INCLUDE_DIRS}
       ${rmw_INCLUDE_DIRS}
       ${rosidl_generator_cpp_INCLUDE_DIRS}
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(rclcpp)
 
 find_package(ament_cmake REQUIRED)
+find_package(rcl REQUIRED)
 find_package(rcl_interfaces REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rmw_implementation REQUIRED)
@@ -45,6 +46,7 @@ macro(target)
   add_library(${PROJECT_NAME}${target_suffix} SHARED
     ${${PROJECT_NAME}_SRCS})
   ament_target_dependencies(${PROJECT_NAME}${target_suffix}
+    "rcl"
     "rcl_interfaces"
     "rmw"
     "rosidl_generator_cpp"
--- a/rclcpp/cmake/get_rclcpp_information.cmake
+++ b/rclcpp/cmake/get_rclcpp_information.cmake
@@ -24,16 +24,22 @@
 # :type var_prefix: string
 #
 macro(get_rclcpp_information rmw_implementation var_prefix)
+  message(STATUS "calling get_rclcpp_information with var_prefix=${var_prefix}")
   # pretend to be a "package"
   # so that the variables can be used by various functions / macros
   set(${var_prefix}_FOUND TRUE)
+
+  # Get rcl using the existing macro
+  if(NOT "${target_suffix} " STREQUAL " ")
+    get_rcl_information("${rmw_implementation}" "rcl${target_suffix}")
+  endif()
 
   # include directories
   normalize_path(${var_prefix}_INCLUDE_DIRS
     "${rclcpp_DIR}/../../../include")
 
-  # libraries
-  set(_libs)
+  # The list of libraries was set in get_rcl_information
+  # set(_libs)
   # search for library relative to this CMake file
   set(_library_target "rclcpp")
   get_available_rmw_implementations(_rmw_impls)
@@ -61,11 +67,8 @@ macro(get_rclcpp_information rmw_implementation var_prefix)
   endif()
 
   # dependencies
-  set(_exported_dependencies
-    "rcl_interfaces"
-    "rmw"
+  list_append_unique(_exported_dependencies
     "rcl"
-    "${rmw_implementation}"
     "rosidl_generator_cpp")
   set(${var_prefix}_DEFINITIONS)
   foreach(_dep ${_exported_dependencies})
--- a/rclcpp/cmake/get_rclcpp_information.cmake
+++ b/rclcpp/cmake/get_rclcpp_information.cmake
@@ -67,6 +67,7 @@ macro(get_rclcpp_information rmw_implementation var_prefix)
 
   # dependencies
   set(_exported_dependencies
+    "rcl_interfaces"
     "rcl${target_suffix}"
     "rosidl_generator_cpp")
   set(${var_prefix}_DEFINITIONS)
--- a/rclcpp/cmake/get_rclcpp_information.cmake
+++ b/rclcpp/cmake/get_rclcpp_information.cmake
@@ -64,6 +64,7 @@ macro(get_rclcpp_information rmw_implementation var_prefix)
   set(_exported_dependencies
     "rcl_interfaces"
     "rmw"
+    "rcl"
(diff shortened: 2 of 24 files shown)


jacquelinekay:
@wjwwood I think f1625d4 fixes the linking issues we were looking at.

jacquelinekay:
OK, 88e2cd8 fixes the isolated build, though I'm still not sure if I'm following cmake/ament best practices (as usual). I also removed the commit that @wjwwood recently commented on pointing out that an isolated build probably won't work (f1625d4), sorry if that's confusing.

jacquelinekay:
nevermind. the commit does not fix the isolated build. I'm working on it

jacquelinekay:
this is (I think) the most recent CI job for this branch:

http://ci.ros2.org/job/ci_osx/934/

As you can see, the services test for Connext and Connext Dynamic times out because services hang indefinitely.

The failures in test_communication should be fixed by https://github.com/ros2/system_tests/pull/113 and are unrelated to the underlying problem of services hanging indefinitely.

jacquelinekay:
With some fixes to rcl services, I'm relaunching CI

http://ci.ros2.org/job/ci_linux/1195/
http://ci.ros2.org/job/ci_osx/971/
http://ci.ros2.org/job/ci_windows/1250/

rclcpp/CMakeLists.txt
```diff
@@ -90,6 +90,7 @@ if(AMENT_ENABLE_TESTING)
   ament_add_gtest(test_mapped_ring_buffer test/test_mapped_ring_buffer.cpp)
   if(TARGET test_mapped_ring_buffer)
     target_include_directories(test_mapped_ring_buffer PUBLIC
+      ${rcl_INCLUDE_DIRS}
       ${rcl_interfaces_INCLUDE_DIRS}
       ${rmw_INCLUDE_DIRS}
```
wjwwood: Is `rclcpp` still using `rmw` in places? Can/should references to `rmw` be removed from the `rclcpp` build files?

rclcpp/cmake/get_rclcpp_information.cmake
```diff
@@ -62,9 +67,7 @@ macro(get_rclcpp_information rmw_implementation var_prefix)
 
   # dependencies
   set(_exported_dependencies
-    "rcl_interfaces"
-    "rmw"
-    "${rmw_implementation}"
+    "rcl${target_suffix}"
```
wjwwood: Should `rcl_interfaces` be exported here? It was removed.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -23,9 +23,13 @@
 #include <tuple>
 #include <utility>
 
+#include "rcl/error_handling.h"
+#include "rcl/client.h"
```
wjwwood: nitpick: alpha-ordering

rclcpp/include/rclcpp/client.hpp
```diff
@@ -23,9 +23,13 @@
...
+
 #include "rclcpp/function_traits.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/utilities.hpp"
+#include "rclcpp/type_support_decl.hpp"
```
wjwwood: alpha

rclcpp/include/rclcpp/client.hpp
```diff
@@ -42,32 +46,28 @@ class ClientBase
 
   RCLCPP_PUBLIC
   ClientBase(
-    std::shared_ptr<rmw_node_t> node_handle,
-    rmw_client_t * client_handle,
+    std::shared_ptr<rcl_node_t> node_handle,
     const std::string & service_name);
 
   RCLCPP_PUBLIC
-  virtual ~ClientBase();
```
wjwwood: Why did you remove the destructor declaration? I think it might be important that it is declared `virtual`.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -42,32 +46,28 @@ class ClientBase
...
+  const rcl_client_t *
   get_client_handle() const;
 
   virtual std::shared_ptr<void> create_response() = 0;
   virtual std::shared_ptr<rmw_request_id_t> create_request_header() = 0;
   virtual void handle_response(
     std::shared_ptr<rmw_request_id_t> request_header, std::shared_ptr<void> response) = 0;
 
-private:
+protected:
```
wjwwood: Any rational here? making members protected is possibly the right thing to do, but it's not clear why you decided to do this.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -93,11 +93,32 @@ class Client : public ClientBase
...
 
   Client(
-    std::shared_ptr<rmw_node_t> node_handle,
-    rmw_client_t * client_handle,
-    const std::string & service_name)
-  : ClientBase(node_handle, client_handle, service_name)
-  {}
+    std::shared_ptr<rcl_node_t> node_handle,
+    const std::string & service_name,
+    rcl_client_options_t & client_options)
```
wjwwood: You changed the order of the arguments here, does this not need changes in the examples and stuff like that?

wjwwood: nvm, this is not used directly by the user.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -292,10 +295,6 @@ class Executor
   get_next_timer(AnyExecutable::SharedPtr any_exec);
 
   RCLCPP_PUBLIC
-  std::chrono::nanoseconds
-  get_earliest_timer();
```
wjwwood: Why was this removed?

rclcpp/include/rclcpp/client.hpp
jacquelinekay: I'm not sure. Previously the child templated class `Client` did not implement a custom destructor, and I add one in this pull request. Maybe I needed to delete this declaration to make it work. I can put this back in and see.

rclcpp/include/rclcpp/executor.hpp
wjwwood: nvm, I guess it's probably handled by `rcl_wait` now. Sorry for the noise.

rclcpp/include/rclcpp/client.hpp
jacquelinekay: I decided to push some member variables that are not templated but need to be accessible by the child class into the parent class.

rclcpp/cmake/get_rclcpp_information.cmake
jacquelinekay: I can add it back in, not sure why I removed it.

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -314,27 +295,17 @@ Node::create_client(
...
-  if (!client_handle) {
-    // *INDENT-OFF* (prevent uncrustify from making unecessary indents here)
-    throw std::runtime_error(
-      std::string("could not create client: ") +
-      rmw_get_error_string_safe());
-    // *INDENT-ON*
-  }
+  rcl_client_options_t options = rcl_client_get_default_options();
+  options.qos = qos_profile;
+  // options.allocator = rclcpp::allocator::get_rcl_allocator<ServiceT, Alloc>(*allocator.get());
```
wjwwood: uncomment or remove?

rclcpp/include/rclcpp/client.hpp
jacquelinekay: When I add it back I get:
```
librclcpp__rmw_opensplice_cpp.so: undefined reference to `rclcpp::client::ClientBase::~ClientBase()'
librclcpp__rmw_opensplice_cpp.so: undefined reference to `typeinfo for rclcpp::client::ClientBase'
librclcpp__rmw_opensplice_cpp.so: undefined reference to `vtable for rclcpp::client::ClientBase'
```

wjwwood: It needs to be implemented or made pure virtual. I'd say give it an empty implementation in the `.cpp` file.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -152,15 +151,54 @@ class Publisher : public PublisherBase
...
+    }
+    if (rmw_get_gid_for_publisher(publisher_rmw_handle, &rmw_gid_) != RMW_RET_OK) {
+      // *INDENT-OFF* (prevent uncrustify from making unecessary indents here)
+      throw std::runtime_error(
+        std::string("failed to get publisher gid: ") + rmw_get_error_string_safe());
+      // *INDENT-ON*
+    }
+  }
+
+  ~Publisher()
```
wjwwood: Please mark this as `virtual`.

rclcpp/include/rclcpp/service.hpp
```diff
@@ -39,19 +43,15 @@ class ServiceBase
 
   RCLCPP_PUBLIC
   ServiceBase(
-    std::shared_ptr<rmw_node_t> node_handle,
-    rmw_service_t * service_handle,
+    std::shared_ptr<rcl_node_t> node_handle,
     const std::string service_name);
 
   RCLCPP_PUBLIC
-  virtual ~ServiceBase();
```
wjwwood: Same thing here, I think you should implement it rather than remove it.

rclcpp/include/rclcpp/publisher.hpp
jacquelinekay: But this is the child class, not the parent class. And it's templated so it can't be inherited from.

rclcpp/include/rclcpp/service.hpp
wjwwood: Without this the child class's destructor will not be called when deleting a pointer of this type (which I think could happen in the future if it doesn't already): http://stackoverflow.com/a/461224/671658

rclcpp/include/rclcpp/service.hpp
```diff
@@ -88,15 +88,37 @@ class Service : public ServiceBase
...
+    {
+      throw std::runtime_error(
+              std::string("could not create service: ") +
+              rcl_get_error_string_safe());
+    }
+  }
 
   Service() = delete;
 
+  ~Service()
```
wjwwood: This should be `virtual` for the same reason, but in the case where the user makes a child class which inherits from this one.

rclcpp/include/rclcpp/publisher.hpp
jacquelinekay: Hmm, nevermind, I guess I'm misremembering how templates and inheritance interact in C++

wjwwood: I think you can inherit from a template class. My understanding is that you should almost always declare a destructor virtual, just in case someone later wants to inherit from it. I believe it applies to templates as well. According to this you can have a template or non-template class inherit from another template class:

http://www.prenhall.com/divisions/esm/app/kafura/secure/chapter7/html/7.5_inheritance.htm

jacquelinekay: ok so, the general pattern I'll follow is: parent class with empty non-virtual destructor implementation, child class with destructor declared as virtual

wjwwood: Yes.

Also if the parent class is already abstract, then you could make the destructor pure virtual and avoid the empty implementation. (which does not appear to be the case here)

jacquelinekay:
welp. Lots of regressions since the last time I tested this. Looking into it

wjwwood:
Well the code review looks ok to me (other than comments). I'll try to help you look into the regressions.

jacquelinekay:
One last round for good measure
http://ci.ros2.org/job/ci_linux/1207/
http://ci.ros2.org/job/ci_osx/976/
http://ci.ros2.org/job/ci_windows/1254

jacquelinekay:
I had to ifdef out the custom allocators in rcl because the way I did it was causing every Windows test to segfault. I also I had to fix some problems with inheritance in timers on Windows.

The multithreaded tests don't pass on Windows on my vm. I'm going to run CI to see what the status is but I'm tempted to just merge this so that I can unblock @wjwwood and then I can work on fixing up the technical debt in parallel.

http://ci.ros2.org/job/ci_linux/1214/
http://ci.ros2.org/job/ci_osx/981/
http://ci.ros2.org/job/ci_windows/1259/

jacquelinekay:
Another round with the atomics fix for Windows (this fixes the multithreaded tests locally for me): https://github.com/ros2/rcl/pull/49

http://ci.ros2.org/job/ci_windows/1264/

jacquelinekay:
OMG that last CI job has 18 failures, which is as many as the nightly Windows release job!! woohoo

wjwwood:
Are they the same ones :)?

wjwwood:
The changes lgtm. Anything else need to be done to merge this?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -89,14 +93,11 @@ signal_handler(int signal_value)
 void
 rclcpp::utilities::init(int argc, char * argv[])
 {
-  (void)argc;
-  (void)argv;
   g_is_interrupted.store(false);
-  rmw_ret_t status = rmw_init();
```
dirk-thomas: The call to `rmw_init` was removed but not replaces with any equivalent. As a result the rmw init function is not being called anymore. See discussion on the mailing list: https://groups.google.com/forum/#!topic/ros-sig-ng-ros/jr6x3vWy4ls

wjwwood: I believe `rcl_init` is being called one line below. I believe the problem is that `rcl_init` does not call `rmw_init`, not that `rclcpp` does not call `rmw_init`.

wjwwood: `rcl_init` does not call `rmw_init`, I'll open a pull request for this:

https://github.com/ros2/rcl/pull/69

However, there is no `rmw_shutdown` right now and `rmw_init` is generally not used, so I think I probably didn't use it because I knew we had to change some things about how init/shutdown worked with the rmw interface anyways. I'm not sure it should be merged in it's asymmetric state, or if we should address this when we fix the rmw init/shutdown code.

PR #208, opened 2016-03-15
Only enable Timer if clock type is steady

jacquelinekay:
rcl currently does not expose the clock type for timers; it assumes a steady clock.

Additionally, in our current implementation of rclcpp, specializing the GenericTimer type with a non-steady clock type results in undefined behavior.

Exposing the clock type for rcl and adding logic to handle non-monotonic timers will take a little more forethought and design (to be done when @tfoote continues pushing on ROS time features). In the meantime, this pull request disables the GenericTimer specialization if the Clock type is not steady (luckily `std::chrono` provides a constant boolean for Clock types for this kind of thing).

@wjwwood, we discussed this change but I'm now unsure how to write a test for this because I've disabled the specialization at compile time, let me know if you have any ideas.

### DIFF
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -84,10 +84,11 @@ using TimerCallbackType = std::function<void(TimerBase &)>;
 /// Generic timer templated on the clock type. Periodically executes a user-specified callback.
 template<
   typename FunctorT,
-  class Clock = std::chrono::high_resolution_clock,
+  class Clock,
   typename std::enable_if<
-    rclcpp::function_traits::same_arguments<FunctorT, VoidCallbackType>::value ||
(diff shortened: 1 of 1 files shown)


wjwwood:
I'm not sure this needs to be tested, or at least it's not worth it since it seems complicated to check a static assert. When we were talking before and I mentioned testing it, I meant if we support a particular clock type then we should document how it should behave and test that it does what we say. Since we're not adding support for system clock or anything like that then there's no need for a new test.

I just looked and couldn't find an easy way to test the SFINAE. If you find a way then consider it, but otherwise just test it locally.

+1

jacquelinekay:
I'll launch CI jobs just in case (since I don't remember exactly how Windows resolves SFINAE).

http://ci.ros2.org/job/ci_linux/1099/
http://ci.ros2.org/job/ci_osx/883/
http://ci.ros2.org/job/ci_windows/1164/

PR #210, opened 2016-03-17
Remove fixed guard conditions, add notify guard conditions

### LINKED ISSUE #209: wake up spin when new waitable things are added to a node
For example:

```c++
auto node = rclcpp::Node::make_shared("node_name");
executor.add_node(node);
std::thread spin_thread([&executor]() {executor.spin();});
auto timer = node->create_timer([]() {printf("hello world\n");});
spin_thread.join();
```

This code may never print "hello world" if the spin call starts before the timer is created. So to fix this, any time something that is affected by spinning is added to a node, spin should be woken up so that it can consider new work.

To accomplish this, each node will probably need a new guard condition which is triggered any time something new is added that needs to wake up spin. This would probably include subscriptions, timers, service servers or clients, and parameter servers or clients.

jacquelinekay:
Fixes #209 
also connects to ros2/rcl#33

### DIFF
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -134,6 +134,11 @@ Node::create_publisher(
       shared_publish_callback,
       intra_process_publisher_handle);
   }
+  if (rmw_trigger_guard_condition(notify_guard_condition_) != RMW_RET_OK) {
+    throw std::runtime_error(
+            std::string(
+              "Failed to notify waitset of change to node state: ") + rmw_get_error_string());
+  }
   return publisher;
 }
 
@@ -243,6 +248,11 @@ Node::create_subscription(
     default_callback_group_->add_subscription(sub_base_ptr);
   }
   number_of_subscriptions_++;
+  if (rmw_trigger_guard_condition(notify_guard_condition_) != RMW_RET_OK) {
+    throw std::runtime_error(
+            std::string(
+              "Failed to notify waitset of change to node state: ") + rmw_get_error_string());
+  }
   return sub;
 }
 
@@ -289,6 +299,11 @@ Node::create_wall_timer(
     default_callback_group_->add_timer(timer);
   }
   number_of_timers_++;
+  if (rmw_trigger_guard_condition(notify_guard_condition_) != RMW_RET_OK) {
+    throw std::runtime_error(
+            std::string(
+              "Failed to notify waitset of change to node state: ") + rmw_get_error_string());
+  }
   return timer;
 }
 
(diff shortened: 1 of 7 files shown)


rclcpp/src/rclcpp/node.cpp
```diff
@@ -36,7 +36,8 @@ Node::Node(
   bool use_intra_process_comms)
 : name_(node_name), context_(context),
   number_of_subscriptions_(0), number_of_timers_(0), number_of_services_(0),
-  use_intra_process_comms_(use_intra_process_comms)
+  use_intra_process_comms_(use_intra_process_comms),
+  notify_guard_condition_(rmw_create_guard_condition())
```
dirk-thomas: Where is this being deallocated?

jacquelinekay: ah, yeah, I have to add a custom destructor for Node now.

wjwwood: What if `rmw_create_guard_condition()` returns `nullptr` (fails)?

jacquelinekay: Does 9dc89bb answer these questions?

wjwwood: Lgtm.

wjwwood:
+1

jacquelinekay:
I think this is good now. I changed some stuff since @wjwwood gave a +1 so you may want to re-review.

http://ci.ros2.org/job/ci_linux/1126/
http://ci.ros2.org/job/ci_osx/912/ (ran with `repeat-until-fail 20` so it's taking a while, but I wanted to catch any flaky regressions)
http://ci.ros2.org/job/ci_windows/1194/

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -134,6 +134,11 @@ Node::create_publisher(
       shared_publish_callback,
       intra_process_publisher_handle);
   }
+  if (rmw_trigger_guard_condition(notify_guard_condition_) != RMW_RET_OK) {
+    throw std::runtime_error(
+            std::string(
+              "Failed to notify waitset of change to node state: ") + rmw_get_error_string());
```
dirk-thomas: Would it be helpful to customize the following error message to mention how the node state changed in order to distinguish them in case they happen?

jacquelinekay: That's a good idea. I changed the error messages to the format "Failed to notify waitset on X creation" in ddc6609, where X is the type of entity added to the node.

jacquelinekay:
this has sat for a while so I'm going to merge soon unless there are objections.

wjwwood:
It's in "in progress"?

I plus ones it before, so it should be fine. You were just fixing up tests right?

jacquelinekay:
Yeah... I think the label is a Waffle/Github sync bug. Not sure about that.

PR #221, opened 2016-05-20
add ability to wait for a service server to be available

wjwwood:
Connects to ros2/ros2#215

This is still a WIP.

### DIFF
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -25,6 +25,7 @@
 
 #include "rcl/client.h"
 #include "rcl/error_handling.h"
+#include "rcl/wait.h"
 
 #include "rclcpp/function_traits.hpp"
 #include "rclcpp/macros.hpp"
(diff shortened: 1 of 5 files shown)


jacquelinekay:
I know that this is still in progress, but I have mixed feelings about adding a rcl_wait_set to the Client. It seems to break the Executor pattern for rclcpp.

wjwwood:
> I know that this is still in progress, but I have mixed feelings about adding a rcl_wait_set to the Client. It seems to break the Executor pattern for rclcpp.

Can you explain why you feel that way? I don't see it that same way.

wjwwood:
I'm going to withdraw this pr and replace it with this one: https://github.com/ros2/rclcpp/pull/222

That new pr uses the branch name `wait_for_service2`. I'm going to bisect the changes related to `wait_for_service` into what can be merged now, and what needs to wait till after the alpha. The latter stuff will be pushed to a `wait_for_service2` branch.

PR #222, opened 2016-05-27
add wait_for_service() and service_is_ready() to Client

wjwwood:
Replacement for https://github.com/ros2/rclcpp/pull/221

Connects to ros2/ros2#215

### DIFF
--- a/rclcpp/include/rclcpp/graph_listener.hpp
+++ b/rclcpp/include/rclcpp/graph_listener.hpp
@@ -23,6 +23,7 @@
 
 #include "rcl/guard_condition.h"
 #include "rcl/wait.h"
+#include "rclcpp/visibility_control.hpp"
 
 namespace rclcpp
 {
@@ -63,8 +64,10 @@ public:
 class GraphListener
 {
 public:
+  RCLCPP_PUBLIC
   GraphListener();
 
+  RCLCPP_PUBLIC
   virtual ~GraphListener();
 
   /// Start the graph listener's listen thread if it hasn't been started.
@@ -72,6 +75,7 @@ public:
    *
    * \throws GraphListenerShutdownError if the GraphListener is shutdown
    */
+  RCLCPP_PUBLIC
   virtual
   void
   start_if_not_started();
@@ -83,6 +87,7 @@ public:
    * \throws std::invalid_argument if node is nullptr
    * \throws std::system_error anything std::mutex::lock() throws
    */
+  RCLCPP_PUBLIC
   virtual
   void
   add_node(rclcpp::node::Node * node);
@@ -92,6 +97,7 @@ public:
    *
    * \throws std::system_error anything std::mutex::lock() throws
    */
+  RCLCPP_PUBLIC
   virtual
   bool
   has_node(rclcpp::node::Node * node) const;
@@ -102,6 +108,7 @@ public:
    * \throws std::invalid_argument if node is nullptr
    * \throws std::system_error anything std::mutex::lock() throws
    */
+  RCLCPP_PUBLIC
   virtual
   void
   remove_node(rclcpp::node::Node * node);
@@ -112,6 +119,7 @@ public:
    * \throws rclcpp::execptions::RCLError from rcl_trigger_guard_condition()
    * \throws std::system_error anything std::mutex::lock() throws
    */
+  RCLCPP_PUBLIC
   virtual
   void
   interrupt();
@@ -130,16 +138,19 @@ public:
    * \throws rclcpp::execptions::RCLError from rcl_wait_set_fini()
    * \throws std::system_error anything std::mutex::lock() throws
    */
+  RCLCPP_PUBLIC
   virtual
   void
   shutdown();
 
 protected:
   /// Main function for the listening thread.
+  RCLCPP_PUBLIC
   virtual
   void
   run();
 
+  RCLCPP_PUBLIC
   virtual
   void
   run_loop();
--- a/rclcpp/include/rclcpp/graph_listener.hpp
+++ b/rclcpp/include/rclcpp/graph_listener.hpp
@@ -1,0 +1,163 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__GRAPH_LISTENER_HPP_
+#define RCLCPP__GRAPH_LISTENER_HPP_
+
+#include <atomic>
+#include <memory>
+#include <mutex>
+#include <thread>
+#include <vector>
+
+#include "rcl/guard_condition.h"
+#include "rcl/wait.h"
+
+namespace rclcpp
+{
+
+namespace node
+{
+class Node;
+}  // namespace node
+
+namespace graph_listener
+{
+
+/// Thrown when a function is called on a GraphListener that is already shutdown.
+class GraphListenerShutdownError : public std::runtime_error
+{
+public:
+  GraphListenerShutdownError()
+  : std::runtime_error("GraphListener already shutdown") {}
+};
+
+/// Thrown when a node has already been added to the GraphListener.
+class NodeAlreadyAddedError : public std::runtime_error
+{
+public:
+  NodeAlreadyAddedError()
+  : std::runtime_error("node already added") {}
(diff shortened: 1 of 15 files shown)


rclcpp/src/rclcpp/graph_listener.cpp
```diff
@@ -0,0 +1,251 @@
...
+        }
+        ret = rcl_wait_set_add_guard_condition(&wait_set_, graph_gc);
+        if (RCL_RET_OK != ret) {
+          throw_from_rcl_error(ret, "failed to add graph guard condition to wait set");
+        }
+      }
+    }  // release the nodes_ lock
+
+    // Wait for graph changes or interrupt.
+    ret = rcl_wait(&wait_set_, -1);  // block for ever until one of the
```
jacquelinekay: Incomplete comment here.

wjwwood: d5961e7

wjwwood:
This needs review. @jacquelinekay already reviewed in some while helping me find and fix a bug.

A few things I want to highlight and to "request comments" about:

- Exceptions:
  - I have been trying in this pull request to be better about throwing specific exceptions.
  - However, I'm not sure how granular to make them, so I'm open to suggestion on how to adjust those.
  - Also, I'm not sure how best to organize the exceptions: all in one header (`rclcpp/exceptions.hpp`) or all in the files they're related to (Node specific exceptions in `rclcpp/node.hpp`) or a mixture of both?
    - Currently I have a mixture of both, with some generic RCL exceptions and help functions in the `rclcpp/exceptions.hpp` header, and other exceptions in `node.hpp` and `graph_listener.hpp`.
- GraphListener:
  - The design is such that there is a single GraphListener with a single thread for all nodes in the given context.
  - This allows us to wait on all graph changes from all nodes in a single thread.
  - Waiting on graph changes in a thread allows functions like `wait_for_service` to work without an executor spinning.
  - It is setup to only create the thread and wake-up in that thread if there are nodes which need to be notified of graph changes, so it's lazy in that way to conserve resources when not being used.
- Testing:
  - All the new code is tested via integration tests in `test_communication`, `test_rclcpp`, and `rclcpp_examples`.
  - I wanted to test GraphListener in a unit test fashion, but I found it too difficult to do so since it takes a full Node into most of its arguments.
    - Mocking Node to simplify the GraphListener tests is too hard, or impossible without dirty hacks, see: https://github.com/ros2/rclcpp/blob/master/rclcpp/test/test_intra_process_manager.cpp#L23-L133
    - I have a plan on how to change the headers in `rclcpp` to make this easier, but I think that's out of scope for this pull request. However, adding narrower tests for these classes would hopefully be easier after the changes. I wanted to discuss them at the last meeting but we ran out of time.

I think that's all I wanted to highlight. Thanks in advance for reviews.

rclcpp/src/rclcpp/client.cpp
```diff
@@ -42,3 +47,58 @@ ClientBase::get_client_handle() const
...
+  do {
+    node_->wait_for_graph_change(event, time_to_wait);
+    if (event->check_and_clear()) {
+      if (this->service_is_ready()) {
+        return true;
+      }
+    }
+    // server is not ready, loop if there is time left
+    time_to_wait = timeout - (std::chrono::steady_clock::now() - start);
+  } while (timeout < std::chrono::nanoseconds(0) || time_to_wait > std::chrono::nanoseconds(0));
```
jacquelinekay: What happens in this loop when the program receives SIGINT/CTRL-C?

rclcpp/src/rclcpp/client.cpp
```diff
@@ -42,3 +47,58 @@ ClientBase::get_client_handle() const
...
+  // check to see if the server is ready immediately
+  if (this->service_is_ready()) {
+    return true;
+  }
+  if (timeout == std::chrono::nanoseconds(0)) {
+    // check was non-blocking, return immediately
+    return false;
+  }
+  // make an event to reuse, rather than create a new one each time
+  auto event = node_->get_graph_event();
```
jacquelinekay: Is it possible for the parent node to become null when this is called? Should the node_ member be a weak_ptr instead?

rclcpp/include/rclcpp/graph_listener.hpp
```diff
@@ -0,0 +1,174 @@
...
+   * \throws rclcpp::execptions::RCLError from rcl_trigger_guard_condition()
+   * \throws std::system_error anything std::mutex::lock() throws
+   */
+  RCLCPP_PUBLIC
+  virtual
+  void
+  interrupt();
+
+  /// Stop the listening thread.
+  /* The thread cannot be restarted, and the class is defunct after calling.
```
jacquelinekay: Hmm. If the class becomes "defunct" after calling, should there be a way of checking if the GraphListener has been shut down?

rclcpp/include/rclcpp/graph_listener.hpp
```diff
@@ -0,0 +1,163 @@
...
+/// Thrown when the given node is not in the GraphListener.
+class NodeNotFoundError : public std::runtime_error
+{
+public:
+  NodeNotFoundError()
+  : std::runtime_error("node not found") {}
+};
+
+/// Notifies many nodes of graph changes by listening in a thread.
+class GraphListener
```
dirk-thomas: Should this class prevent copy construction / assignment operator?

rclcpp/include/rclcpp/graph_listener.hpp
```diff
@@ -0,0 +1,174 @@
...
```
dirk-thomas: Should this class prevent copy construction / assignment operator?

rclcpp/src/rclcpp/event.cpp
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+
+#include "rclcpp/event.hpp"
+
+namespace rclcpp
+{
+namespace event
+{
+
+Event::Event()
```
jacquelinekay: nitpick: the name "Event" is a little vague, maybe "GraphListenerEvent" would be better?

wjwwood: I originally called it something like that, but I figured it could be used for other stuff too, since it really is just a generic "event" class. For example, it is currently returned from `node->get_graph_event` but there might be a `node->get_parameter_event` in the future. I don't know if that's the case, but I didn't see any reason it should be less generic, but I'm open to changing it if you like.

rclcpp/include/rclcpp/graph_listener.hpp
wjwwood: All the functions will throw a `GraphListenerShutdownError` if it is shutdown, but I also added an `is_shutdown` method to the class in 86dd574.

rclcpp/include/rclcpp/graph_listener.hpp
wjwwood: Yeah, I disabled it in 42e20cf.

rclcpp/src/rclcpp/event.cpp
jacquelinekay: ¯\_(ツ)_/¯ Makes sense, I'm fine with keeping it this way.

jacquelinekay:
My major questions are about what happens in the do/while loop on CTRL-C, and the ownership of the Node pointer in Client. Otherwise it looks good.

rclcpp/src/rclcpp/client.cpp
wjwwood: Yes, I think so. I was avoiding this because basically we'll need to enable "make_shared_from_this" on Node. But I think that can't be avoided unless we rethink our ownership model a little bit.

I'll see what's involved to do that.

rclcpp/src/rclcpp/client.cpp
wjwwood: That's a good question... Currently I think it will not wake up, because neither `node_->wait_for_graph_change` nor this loop will consider `rclcpp::ok` or be notified by the shutdown call or signal processor.

What do you think the behavior of this function should be in that case? I'll need to adjust the documentation to be something like "returns true if graph changed or false if the timeout expired or rclcpp::shutdown occurred". But then you can't easily tell which happened timeout or shutdown, though I guess if you check `rclcpp::ok` after getting `false` back from this function and be reasonably sure that was the cause or not.

jacquelinekay: I think it's fine to check for `!rclcpp::ok()` in the while predicate and return false in that case. A return value of false means "we stopped waiting and the client was not ready." If you want a finer level of granularity we could add error codes (client_ready, timeout, interrupted) but I don't think that's necessary right now.

rclcpp/src/rclcpp/client.cpp
wjwwood: c7c4fa4 should do it. Let me know if that looks ok to you.

rclcpp/src/rclcpp/client.cpp
wjwwood: Ok, I added this behavior in 25e17b0 and I added a test in ros2/system_tests@268f54a.

jacquelinekay:
+1

dirk-thomas:
+1

PR #223, opened 2016-06-06
Add mutex to protect vectors of pointers in callbackgroup

gerkey:
Test failures such as:

* http://ci.ros2.org/view/nightly/job/nightly_osx_repeated/339/testReport/%28root%29/projectroot/gtest_executor__rmw_fastrtps_cpp/
* http://ci.ros2.org/view/nightly/job/nightly_win_rel/105/testReport/%28root%29/projectroot/gtest_executor__rmw_opensplice_cpp/

can be traced to concurrent access to the various vectors of pointers in `CallbackGroup`. For example, if you have a single-threaded executor spinning a node and you call `create_wall_timer()` on that node from another thread, then you can be pushing the new timer onto `CallbackGroup::timer_ptrs_` while the executor is reading the same vector.

This PR adds a single mutex to protect all of the vectors. If we observe contention, we could consider making the protection more granular.

* Linux:
 * [Build Status](http://ci.ros2.org/job/ci_linux/1409/)
* OSX:
 * [Build Status](http://ci.ros2.org/job/ci_osx/1093/)
* Windows:
 * [Build Status](http://ci.ros2.org/job/ci_windows/1417/)

### DIFF
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -16,6 +16,7 @@
 #define RCLCPP__CALLBACK_GROUP_HPP_
 
 #include <atomic>
+#include <mutex>
 #include <string>
 #include <vector>
 
@@ -97,6 +98,8 @@ private:
   add_client(const rclcpp::client::ClientBase::SharedPtr client_ptr);
 
   CallbackGroupType type_;
+  // Mutex to protect the subsequent vectors of pointers.
+  mutable std::mutex mutex_;
   std::vector<rclcpp::subscription::SubscriptionBase::WeakPtr> subscription_ptrs_;
   std::vector<rclcpp::timer::TimerBase::WeakPtr> timer_ptrs_;
   std::vector<rclcpp::service::ServiceBase::SharedPtr> service_ptrs_;
--- a/rclcpp/src/rclcpp/callback_group.cpp
+++ b/rclcpp/src/rclcpp/callback_group.cpp
@@ -26,24 +26,28 @@ CallbackGroup::CallbackGroup(CallbackGroupType group_type)
 const std::vector<rclcpp::subscription::SubscriptionBase::WeakPtr> &
 CallbackGroup::get_subscription_ptrs() const
 {
+  std::lock_guard<std::mutex> lock(mutex_);
   return subscription_ptrs_;
 }
 
 const std::vector<rclcpp::timer::TimerBase::WeakPtr> &
 CallbackGroup::get_timer_ptrs() const
 {
+  std::lock_guard<std::mutex> lock(mutex_);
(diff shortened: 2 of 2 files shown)


rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -55,19 +56,19 @@ class CallbackGroup
 
   RCLCPP_PUBLIC
   const std::vector<rclcpp::subscription::SubscriptionBase::WeakPtr> &
-  get_subscription_ptrs() const;
+  get_subscription_ptrs();
```
dirk-thomas: The signature of these methods should stay const.

gerkey: I don't think that that's possible when the functions are locking a mutex that is a member of the class.

gerkey: 8b6a417f8f405dbff4f0132dcc7ff6d9a05d73d1 goes back to `const`, with a `mutable` mutex. Thanks @dirk-thomas for the tip!

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -97,6 +98,8 @@ class CallbackGroup
   add_client(const rclcpp::client::ClientBase::SharedPtr client_ptr);
 
   CallbackGroupType type_;
+  // Mutex to protect the subsequent vectors of pointers.
+  mutable std::mutex lock_;
```
wjwwood: nitpick: this is a mutex not a lock, maybe it should be called `mutex_`? I don't consider this a blocker to merge.

wjwwood:
lgtm

rclcpp/include/rclcpp/callback_group.hpp
gerkey: Nitpicks welcome! 517232d

gerkey:
Linux and OSX look good. Windows showed the usual failures, plus this one, which is a timeout:

http://ci.ros2.org/job/ci_windows/1417/testReport/%28root%29/projectroot/test_demo_two_node_pipeline__rmw_fastrtps_cpp/

There's a chance that my addition of the mutex caused that timeout. Something to keep an eye on.

Also, I'll note that on Windows a different test in the same file failed, also with a crash:

http://ci.ros2.org/job/ci_windows/1417/testReport/%28root%29/projectroot/gtest_executor__rmw_opensplice_cpp/

jacquelinekay:
I'm not sure why this would cause the single-threaded two_node_pipeline to timeout, unless the overhead of acquiring a mutex on Windows is insane. Interesting.

wjwwood:
It would have to be deadlock or unrelated. I don't think acquiring the mutex would take that long.

PR #226, opened 2016-06-14
Give a different signal guard condition for each waitset

### LINKED ISSUE #225: Race condition on signal guard condition
As identified in https://github.com/ros2/rmw_opensplice/issues/99#issuecomment-224695313, there's a race condition on detaching the shared signal guard condition from multiple wait sets. A preliminary fix is in the `global_guard` branch.

gerkey:
Fix #225 by returning a different guard condition for each waitset, to avoid having one condition reused across waitsets.

@wjwwood, this isn't the implementation that you suggested, but I found it to be the most straightforward way to fix the problem. What do you think?

(I'll run CI after I get consensus on the implementation.)

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -52,7 +52,7 @@ shutdown();
 /// Get a handle to the rmw guard condition that manages the signal handler.
 RCLCPP_PUBLIC
 rcl_guard_condition_t *
-get_global_sigint_guard_condition();
+get_global_sigint_guard_condition(void*);
 
 /// Use the global condition variable to block for the specified amount of time.
 /**
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -20,6 +20,7 @@
 #include "rclcpp/visibility_control.hpp"
 
 #include "rcl/guard_condition.h"
+#include "rcl/wait.h"
 
 #include "rmw/macros.h"
 #include "rmw/rmw.h"
@@ -50,9 +51,20 @@ void
 shutdown();
 
 /// Get a handle to the rmw guard condition that manages the signal handler.
+/**
+ * The first time that this function is called for a given waitset a new guard
+ * condition will be created and returned; thereafter the same guard condition
+ * will be returned for the same waitset. This mechanism is designed to ensure
+ * that the same guard condition is not reused across waitsets (e.g., when
+ * using multiple executors in the same process). Can throw an exception if
+ * initialization of the guard condition fails.
+ * \param[waitset] waitset Pointer to the rcl_wait_set_t that will be using the
+ * resulting guard condition.
+ * \return Pointer to the guard condition.
+ */
 RCLCPP_PUBLIC
 rcl_guard_condition_t *
-get_global_sigint_guard_condition(void*);
+get_sigint_guard_condition(rcl_wait_set_t * waitset);
 
 /// Use the global condition variable to block for the specified amount of time.
 /**
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -56,7 +56,7 @@ shutdown();
  * condition will be created and returned; thereafter the same guard condition
  * will be returned for the same waitset. This mechanism is designed to ensure
  * that the same guard condition is not reused across waitsets (e.g., when
- * using multiple executors in the same process). Can throw an exception if
+ * using multiple executors in the same process). Will throw an exception if
  * initialization of the guard condition fails.
  * \param[waitset] waitset Pointer to the rcl_wait_set_t that will be using the
  * resulting guard condition.
@@ -65,6 +65,19 @@ shutdown();
 RCLCPP_PUBLIC
 rcl_guard_condition_t *
 get_sigint_guard_condition(rcl_wait_set_t * waitset);
+
+/// Release the previously allocated guard condition that manages the signal handler
+/**
+ * If you previously called get_sigint_guard_condition() for a given waitset
+ * to get a sigint guard condition, then you should call release_sigint_guard_condition()
+ * when you're done, to free that condition.  Will throw an exception if
+ * get_sigint_guard_condition() wasn't previously called for the given waitset.
+ * \param[waitset] waitset Pointer to the rcl_wait_set_t that was using the
+ * resulting guard condition.
+ */
+RCLCPP_PUBLIC
+void
+release_sigint_guard_condition(rcl_wait_set_t * waitset);
 
 /// Use the global condition variable to block for the specified amount of time.
 /**
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -66,7 +66,7 @@ RCLCPP_PUBLIC
 rcl_guard_condition_t *
 get_sigint_guard_condition(rcl_wait_set_t * waitset);
 
-/// Release the previously allocated guard condition that manages the signal handler
+/// Release the previously allocated guard condition that manages the signal handler.
 /**
  * If you previously called get_sigint_guard_condition() for a given waitset
  * to get a sigint guard condition, then you should call release_sigint_guard_condition()
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -77,6 +77,8 @@ Executor::~Executor()
     fprintf(stderr,
       "[rclcpp::error] failed to destroy guard condition: %s\n", rcl_get_error_string_safe());
   }
+  // Release the sigint guard condition
+  rclcpp::utilities::release_sigint_guard_condition(&waitset_);
 }
 
 void
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -46,7 +46,7 @@ Executor::Executor(const ExecutorArgs & args)
   // and one for the executor's guard cond (interrupt_guard_condition_)
 
   // Put the global ctrl-c guard condition in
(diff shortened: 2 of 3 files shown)


rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -156,18 +158,37 @@ void
...
   g_signal_status = SIGINT;
-  if (rcl_trigger_guard_condition(&g_sigint_guard_cond_handle) != RCL_RET_OK) {
-    fprintf(stderr,
-      "[rclcpp::error] failed to trigger guard condition: %s\n", rmw_get_error_string_safe());
+  {
+    std::lock_guard<std::mutex> lock(g_sigint_guard_cond_handles_mutex);
+    for (auto const & kv : g_sigint_guard_cond_handles) {
+      if (rcl_trigger_guard_condition(&(kv.second)) != RCL_RET_OK) {
+        fprintf(stderr,
+          "[rclcpp::error] failed to trigger guard condition: %s\n", rmw_get_error_string_safe());
```
wjwwood: I'd recommend mentioning that it is a sigint guard condition, and you should use `rcl_get_error_string_safe()`.

wjwwood:
My main concern with this approach is that you cannot remove sigint guard conditions without doing shutdown. It's a contrived use case, perhaps only useful for testing, but I wouldn't be able to create and destroy executors at will without also doing shutdown and init.

Also, I don't see any technical reason to tie the sigint guard condition to a wait set, though it does make sense as a defensive API design. Again a contrived example, but it might be useful to not have to create a new guard condition every time you create a wait set. Instead, you might want to get a sigint guard condition, and create and destroy wait sets repeatedly but reuse the same guard condition.

I think that this approach with a method for removing a sigint guard condition outside of shutdown would be ok, except that it requires some coordination between the executor and shutdown. Currently in this pr, if you have an executor that is spinning and then someone calls shutdown, then the shutdown function will fini guard conditions that may still be used in an `rcl_wait` call. So for this to be safe, you'd need to ensure all executors are no longer using the guard conditions before proceeding to destroy them in shutdown.

As an alternative approach you could have the function take no arguments but return a shared_ptr to a guard condition and keep a vector of weak_ptr's to the sigint guard conditions globally. This way you can remove unused sigint guard conditions by just letting the shared_ptr go out of scope and you never run the risk of shutdown destroying guard conditions that are still in use by the executor.

Hopefully that feedback is useful, thanks for working on it.

gerkey:
@wjwwood thanks for the feedback; very helpful.

You're right that, with this implementation, you can't create a waitset without also creating a sigint condition.

With regard to shutdown, I don't see any place that the sigint conditions are being destroyed. In `rclcpp::utilities::shutdown()`, they're only being triggered. Can you say more about the thread safety during shutdown?

Assuming that my understanding is correct, then this implementation does introduce an internal memory leak: each new executor will create a guard condition as a side-effect, but they won't be cleaned up until process exit.

wjwwood:
Oh, you're right. I misread the code. It's not destroying the guard conditions in the shutdown method, only triggering them.

So in that case, this method with a function to get rid of the sigint guard condition (probably in the Executor's destructor) would work fine.

rclcpp/src/rclcpp/utilities.cpp
gerkey: Sure, I'll update the message to be more descriptive.

The `fprintf` call is already appending the return from `rcl_get_error_string_safe()`. Is there something further I need to do?

wjwwood: `rcl_get_error_string_safe` not `rmw_get_error_string_safe`?

gerkey: My mistake. Two letters make all the difference.

gerkey: 75e21ff

gerkey:
Added code to release sigint handler in 75e21ff.

* Linux:
 * [Build Status](http://ci.ros2.org/job/ci_linux/1445/)
* OSX:
 * [Build Status](http://ci.ros2.org/job/ci_osx/1129/)
* Windows:
 * [Build Status](http://ci.ros2.org/job/ci_windows/1462/)

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -50,9 +51,33 @@ void
...
+ * \param[waitset] waitset Pointer to the rcl_wait_set_t that will be using the
+ * resulting guard condition.
+ * \return Pointer to the guard condition.
+ */
 RCLCPP_PUBLIC
 rcl_guard_condition_t *
-get_global_sigint_guard_condition();
+get_sigint_guard_condition(rcl_wait_set_t * waitset);
+
+/// Release the previously allocated guard condition that manages the signal handler
```
wjwwood: nitpick: our "brief" descriptions usually end with punctuation.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -77,6 +77,8 @@ Executor::~Executor()
     fprintf(stderr,
       "[rclcpp::error] failed to destroy guard condition: %s\n", rcl_get_error_string_safe());
   }
+  // Release the sigint guard condition
+  rclcpp::utilities::release_sigint_guard_condition(&waitset_);
```
wjwwood: Does it need to be removed from the `memory_strategy_` or is that automatic in the destructors?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -156,18 +158,61 @@ void
...
+      fprintf(stderr,
+        "[rclcpp::error] failed to destroy sigint guard condition: %s\n",
+        rcl_get_error_string_safe());
+    }
+    g_sigint_guard_cond_handles.erase(kv);
+  } else {
+    // *INDENT-OFF* (prevent uncrustify from making unnecessary indents here)
+    throw std::runtime_error(std::string(
+      "Tried to release sigint guard condition for nonexistent waitset: %s") +
+      rcl_get_error_string_safe());
```
wjwwood: The rcl error string is probably not set at this point. Maybe you should put in the pointer to the wait set? Or nothing other than the message?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -156,18 +158,61 @@ void
...
+void
+rclcpp::utilities::release_sigint_guard_condition(rcl_wait_set_t * waitset)
+{
+  std::lock_guard<std::mutex> lock(g_sigint_guard_cond_handles_mutex);
+  auto kv = g_sigint_guard_cond_handles.find(waitset);
+  if (kv != g_sigint_guard_cond_handles.end()) {
+    if (rcl_guard_condition_fini(&kv->second) != RCL_RET_OK) {
+      fprintf(stderr,
+        "[rclcpp::error] failed to destroy sigint guard condition: %s\n",
+        rcl_get_error_string_safe());
```
wjwwood: This should probably throw. I have a new function that makes it easy and consistent to throw from an rcl ret code, but that's in my pending pr. If you want to leave this as-is or just throw a runtime_exception here I can clean it up in a follow on pr, which I planned to do later.

wjwwood: FYI https://github.com/ros2/rclcpp/pull/222/files#diff-5af14f62862bacbab35ea2de5215f716R30

wjwwood:
I had just a few small comments, but those shouldn't affect the CI results. Otherwise, code changes lgtm.

wjwwood:
One more thought, though I don't think it should block the pr. If I create a wait set and use the new function to get a sigint and then never call release, a guard condition will be leaked. In the case of guard conditions this is probably fine (since it's process shutdown). For other elements like subscribers it might be important to ensure fini is called for one reason or another. It would be possible to store the map in a struct, the destructor of which would call release on any straggling guard conditions still in the map. Again, not a blocker but something to consider.

gerkey:
Comments addressed in 3530655.

I considered the idea of arranging for automatic destruction of the guard condition map, and I see the point in general, but I decided against doing it here, for two reasons: (1) in this case the map is a global, which won't go out of scope until process exit, at which point we don't benefit much from explicitly destructing guard conditions; and (2) while it can happen, I don't expect this API to be commonly used outside of the executor, which calls it correctly in constructor and destructor.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -77,6 +77,9 @@ Executor::~Executor()
     fprintf(stderr,
       "[rclcpp::error] failed to destroy guard condition: %s\n", rcl_get_error_string_safe());
   }
+  // Remove and release the sigint guard condition
+  memory_strategy_->remove_guard_condition(rclcpp::utilities::get_sigint_guard_condition(&waitset_));
```
dirk-thomas: Fails linter: http://ci.ros2.org/job/ci_linux/1460/testReport/junit/(root)/projectroot/cpplint/

dirk-thomas:
Please run a CI job before merging - especially when not running the tests locally before.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -156,18 +158,60 @@ void
...
+void
+rclcpp::utilities::release_sigint_guard_condition(rcl_wait_set_t * waitset)
+{
+  std::lock_guard<std::mutex> lock(g_sigint_guard_cond_handles_mutex);
+  auto kv = g_sigint_guard_cond_handles.find(waitset);
+  if (kv != g_sigint_guard_cond_handles.end()) {
+    if (rcl_guard_condition_fini(&kv->second) != RCL_RET_OK) {
+      throw std::runtime_error(std::string(
+        "Failed to destroy sigint guard condition: ") +
+        rcl_get_error_string_safe());
```
dirk-thomas: Fails linter: http://ci.ros2.org/job/ci_linux/1460/testReport/junit/(root)/projectroot/uncrustify/

gerkey:
Oh, that rascally linter...#229.

PR #228, opened 2016-06-16
Only use _dupenv_s if it's defined

esteve:
This enables building on UWP, which does not implement `_dupenv_s`:

https://msdn.microsoft.com/en-us/library/windows/apps/jj606124.aspx

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -48,6 +48,9 @@ macro(target)
   endif()
   add_library(${PROJECT_NAME}${target_suffix} SHARED
     ${${PROJECT_NAME}_SRCS})
+  if("${CMAKE_SYSTEM_NAME} " STREQUAL "WindowsPhone " OR "${CMAKE_SYSTEM_NAME} " STREQUAL "WindowsStore ")
+    set_property(TARGET ${PROJECT_NAME}${target_suffix} PROPERTY VS_WINRT_COMPONENT 1)
+  endif()
   ament_target_dependencies(${PROJECT_NAME}${target_suffix}
     "rcl${target_suffix}"
     "rosidl_generator_cpp")
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -54,8 +54,10 @@ Node::Node(
 #ifndef _WIN32
   ros_domain_id = getenv(env_var);
 #else
+#ifndef __cplusplus_winrt
   size_t ros_domain_id_size;
   _dupenv_s(&ros_domain_id, &ros_domain_id_size, env_var);
+#endif
 #endif
   if (ros_domain_id) {
     uint32_t number = strtoul(ros_domain_id, NULL, 0);


dirk-thomas:
What should happen in the `else` case? Just nothing seems to be wrong since the string doesn't get copied at all.

esteve:
@dirk-thomas UWP applications don't have the concept of environment variables since they run in a sandbox, so unfortunately `domain_id` will always be 0 (taken from https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node.cpp#L51) because `ros_domain_id` is null (https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node.cpp#L60)

Unfortunately, I can't think of any other way (besides programmatically), to set the ROS domain.

esteve:
@dirk-thomas just something that occurred to me while replying is that the context in `Node` might be a place to store the domain id, what do you think?

dirk-thomas:
Anyway the patch can't be applied as-is. Simply because even if the function `_dupenv_s` is available there is no definition `_dupenv_s` so the block is always being removed which breaks the functionality on Windows.

esteve:
@dirk-thomas true, wrong patch. Anyway, can you comment on my proposal to store the domain id in the context?

dirk-thomas:
I am not sure if adding it to the context would fit the intention of the context object (https://github.com/ros2/rclcpp/blob/bf6394004c9bd324c81b6f6bcf1bbbe72240324f/rclcpp/include/rclcpp/node.hpp#L73). I will let @wjwwood comment on this if he think it should go in the context or be optionally passed as a constructor argument to the node.

esteve:
I've updated the patch to disable the use of `_dupenv_s` if `rclcpp` is built as a WinRT component.

esteve:
@dirk-thomas thanks for your comments.

wjwwood:
Probably should be a node argument, but the intra process manager shared between them would stop working if they're on different domains. That's to be expected but it also means it doesn't necessarily make sense for them to share a context. However I think a parameter to the node constructor still makes sense. 

Also programmatically setting the domain I'd won't work for opensplice on Windows because it has to match their config file.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -54,9 +54,11 @@ Node::Node(
 #ifndef _WIN32
   ros_domain_id = getenv(env_var);
 #else
+#ifndef __cplusplus_winrt
```
wjwwood: I think this would be better as a `#elif`.

wjwwood:
I would also advise conditioning out this line:

https://github.com/ros2/rclcpp/pull/228/files#diff-c80a17bb6743ece08c6f03fa06b1c5faR73

I know that the variable should always be nullptr and therefore this code should be run, but I think it would convey more intent if it was also excluded by the preprocessor.

rclcpp/CMakeLists.txt
```diff
@@ -48,6 +48,9 @@ macro(target)
   endif()
   add_library(${PROJECT_NAME}${target_suffix} SHARED
     ${${PROJECT_NAME}_SRCS})
+  if("${CMAKE_SYSTEM_NAME} " STREQUAL "WindowsPhone " OR "${CMAKE_SYSTEM_NAME} " STREQUAL "WindowsStore ")
```
dirk-thomas: Since we have moved to CMake 3.5 please avoid the trailing space as well as the quoting and explicit expansion of variables in the condition:

    if(CMAKE_SYSTEM_NAME STREQUAL "WindowsPhone" OR CMAKE_SYSTEM_NAME STREQUAL "WindowsStore")

mikaelarguedas:
@esteve what is the status on this ? should this be rebased and comments addressed ?

dirk-thomas:
I will close this ticket for now. Please feel free to comment and it can be reopened.

PR #229, opened 2016-06-16
Fix style

gerkey:
Fix style errors introduced in #226. I ran the linter tests locally with this change and they passed.


dirk-thomas:
+1

PR #230, opened 2016-06-20
Allow empty prefix list to get all params; check depth correctly.

gerkey:
Connects to https://github.com/ros2/rclcpp/issues/196

### DIFF
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -19,6 +19,7 @@
 #include <utility>
 #include <vector>
 
+#include "rcl_interfaces/srv/list_parameters.hpp"
 #include "rclcpp/node.hpp"
 
 using rclcpp::node::Node;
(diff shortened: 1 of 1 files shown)


wjwwood:
lgtm, but that if-statement is getting hairy looking 😄.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -271,17 +271,21 @@ Node::list_parameters(
 
   // TODO(esteve): define parameter separator, use "." for now
   for (auto & kv : parameters_) {
-    if (std::any_of(prefixes.cbegin(), prefixes.cend(), [&kv, &depth](const std::string & prefix) {
+    if (((prefixes.size() == 0) &&
+      ((depth == 0) ||
```
dirk-thomas: Please use the constant defined in the service rather then a plain `0` here.

gerkey: dfd3bf8

Matching change in https://github.com/ros2/system_tests/pull/140/commits/3d1d67219b039d5a0146cfa47f4d6ce2e8118401

PR #231, opened 2016-06-28
Implement std::to_string for Android

esteve:
This PR implements `std::to_string` for Android, which does not exist on the NDK.

Also, `bionic`, the C library for Android uses the standard signature for `strerror_r`, not the GNU extension, despite `GNU_SOURCE` being defined.

I don't know if `utilities.hpp` is the best place to put `std::to_string`, so I'm open to suggestions.

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -26,6 +26,22 @@
 #include "rmw/macros.h"
 #include "rmw/rmw.h"
 
+#ifdef ANDROID
+#include <string>
+#include <sstream>
+
+namespace std
+{
+template<typename T>
+std::string to_string(T value)
+{
+  std::ostringstream os;
+  os << value;
+  return os.str();
+}
+}
+#endif
+
 namespace rclcpp
 {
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Please run the tests locally and address the linter failures.

esteve:
@dirk-thomas fixed.

dirk-thomas:
Thanks. LGTM.

I will let @wjwwood decide if it should be in `utilities.hpp` or if he prefers a new / different file for it.

wjwwood:
It seems fine to me. If the number of "if ANDROID" things grow, then we can put them in a different header or even a compatibility package.

I'll just run CI (before / after) over it real quick, but it lgtm:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1582/) / [Build Status](http://ci.ros2.org/job/ci_linux/1583/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1208/) / [Build Status](http://ci.ros2.org/job/ci_osx/1209/)

wjwwood:
CI lgtm

PR #234, opened 2016-07-01
fix warnings on Windows

dirk-thomas:
Before: http://ci.ros2.org/job/ci_windows/1589/warnings35Result/new/package.-1592860977/
After: http://ci.ros2.org/job/ci_windows/1593/


gerkey:
Thanks!

PR #236, opened 2016-07-06
use the param profile for param services

gerkey:
We have a parameter qos profile (https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L33-L39), but we're not using it in the param service or client creation. This PR uses it.

A future extension to consider would be exposing the qos setting through the parameter service/client creation API, to allow the user to change it.

(This came up while working on eProsima/ROS-RMW-Fast-RTPS-cpp#46, wherein it was identified that the history depth was too shallow.)

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1640/) 
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1229/)

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -48,7 +48,8 @@ public:
   RCLCPP_PUBLIC
   AsyncParametersClient(
     const rclcpp::node::Node::SharedPtr node,
-    const std::string & remote_node_name = "");
+    const std::string & remote_node_name = "",
+    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
 
   RCLCPP_PUBLIC
   std::shared_future<std::vector<rclcpp::parameter::ParameterVariant>>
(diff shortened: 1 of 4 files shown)


dirk-thomas:
The patch lgtm.

It would be straight forward to expose the QoS profile as an optional argument. That could happen as part of this PR which would also avoid having to hard code this value in 11 locations.

gerkey:
Good idea: d91b4f1.

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1641/) 
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1232/)

gerkey:
The Linux failures are known problems. The OSX failures seem to be mostly known, but I'm not sure about the `test_parameter_server` failure. Anybody know anything about that?

Also, any comments on the new API that I exposed in  d91b4f1?

dirk-thomas:
The added optional parameter looks good to me.

gerkey:
After review, my opinion is that the failing tests are all the result of known issues, including flaky tests, so I'm going to merge.

PR #237, opened 2016-07-08
add parameter helpers (redo of #233)

gerkey:
This is a replacement for #233, with the same diff (because I once again went into the weeds while attempting to rebase; I'll get the hang of that eventually).

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1648/) 
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1237/)

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -135,6 +135,38 @@ public:
   get_parameters(const std::vector<std::string> & parameter_names);
 
   RCLCPP_PUBLIC
+  bool
+  has_parameter(const std::string & parameter_name);
+
+  template<typename T>
+  T
+  get_parameter(const std::string & parameter_name, const T & default_value)
+  {
+    std::vector<std::string> names;
+    names.push_back(parameter_name);
+    auto vars = get_parameters(names);
+    if ((vars.size() == 0) || (vars[0].get_type() == rclcpp::parameter::PARAMETER_NOT_SET)) {
+      return default_value;
+    } else {
+      return static_cast<T>(vars[0].get_value<T>());
+    }
+  }
+
+  template<typename T>
+  T
+  get_parameter(const std::string & parameter_name)
+  {
+    std::vector<std::string> names;
+    names.push_back(parameter_name);
+    auto vars = get_parameters(names);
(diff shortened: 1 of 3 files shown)


rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -135,6 +135,38 @@ class SyncParametersClient
...
+  }
+
+  template<typename T>
+  T
+  get_parameter(const std::string & parameter_name)
+  {
+    std::vector<std::string> names;
+    names.push_back(parameter_name);
+    auto vars = get_parameters(names);
+    if ((vars.size() == 0) || (vars[0].get_type() == rclcpp::parameter::PARAMETER_NOT_SET)) {
```
wjwwood: In what case is `vars.size() > 1` valid? Should that also be checked for?

Also, it seems as if these two functions could be de-duplicated a little bit, though that might just end up being more complicated. I could, however, imagine:

```c++
template<typename T>
T
get_parameter_impl(const std::string & parameter_name, std::function<T()> parameter_not_found_handler)
{
  std::vector<std::string> names;
  names.push_back(parameter_name);
  auto vars = get_parameters(names);
  if ((vars.size() == 0) || (vars[0].get_type() == rclcpp::parameter::PARAMETER_NOT_SET)) {
    return parameter_not_found_handler();
  } else {
    return static_cast<T>(vars[0].get_value<T>());
  } 
}

template<typename T>
T
get_parameter(const std::string & parameter_name, const T & default_value)
{
  return get_parameter_impl(parameter_name, [&default_value]() { return default_value; });
}

template<typename T>
T
get_parameter(const std::string & parameter_name)
{
  return get_parameter_impl(parameter_name, []() { throw std::runtime_error("Parameter not set"); });
}
```

wjwwood:
Other than a comment, the code changes lgtm.

gerkey:
Tests are in ros2/system_tests#155.

rclcpp/include/rclcpp/parameter_client.hpp
gerkey: Great suggestion, thanks! Done, with minor tweaks, in  	a1b4e15.

gerkey:
New CI, this time with the tests from ros2/system_tests#155 included:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1649/) 
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1238/)

PR #242, opened 2016-07-20
Added missing variable initialization in Node constructor

SantiagoMunoz:


dirk-thomas:
Thank you for the fix.

Issue #243, opened 2016-07-27
Parameter Events should be node specific


rohbotics:
Right now parameter events are published on a global topic `/parameter_events`.

This causes problems when there are multiple nodes listening to `/parameter_events`.
Either the message needs to be changed to contain the node name where the parameter is hosted,
or `/parameter_events` needs to become node local.

I think the local `~/parameter_events` is the cleaner way, as that way callbacks aren't triggered for all nodes every-time there is an update on one node.

wjwwood:
I think the purpose of the `/parameter_events` topic is so that a single node can efficiently monitor all changes. It shouldn't be subscribed to by every node that has parameters. However, I do agree that the parameter events should contain the node to which the parameter belongs and possibly also the node that requested the change originally. The former might be addressed if the canonical parameter addresses become `node + parameter namespace + parameter name` as has been discussed before in our meetings.

As for reacting to a parameter changing in a node, i.e. implementing `on_change('param name', function_to_call)`, I think that should be handled in the set parameter code, not by listening to the parameter events.

rohbotics:
I think that view of parameter events makes sense. For example a gui showing all the parameters shouldn't have to subscribe to individual node topics.

So for dynamic parameter changes, parameter_service should provide a hook for updating the parameters, as it receives the service call in the first place.
Should the hook take the from of `on_change('param_name', callback)` or `on_param_change(callback)`?

wjwwood:
That's correct. In fact further than just having a hook for notifying the node that it has been changed, the hook would allow the node to reject the change if it desired.

I suppose you could have either:

- `node.on_param_change(callback)`
- `node.on_param_change('param_foo', callback)`
- `node.on_param_change('gains/*', callback)`

I think we didn't implement this yet because it is so tied to how we end up formulating parameter names (currently just a variable style name with no namespaces).

rohbotics:
@wjwwood I made basic PR for `node.on_param_change(callback)`

sloretz:
What is the plan for this ticket? It appears the topic `parameter_events` is currently in the namespace given to the constructor of `Node`, not the global namespace.

https://github.com/ros2/rclcpp/blob/6129a12df5ec21b7ccbf6879036281b1374db0fb/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L41-L46

mikaelarguedas:
This will be addressed during the [parameter refactor](https://github.com/ros2/ros2/issues/432).
[...]

PR #244, opened 2016-07-28
Added basic hook for parameter changes

rohbotics:
See #243

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -353,8 +353,13 @@ public:
   size_t
   count_graph_users();
 
+  /// Register the callback for parameter changes
+  /**
+   * \param[in] User defined callback function, It is expected to atomically set parameters.
+   * \note Repeated invocations of this function will overwrite previous callbacks
+   */
   template<typename CallbackT>
-  void on_param_change(CallbackT && callback);
+  void register_param_change_callback(CallbackT && callback);
 
   std::atomic_bool has_executor;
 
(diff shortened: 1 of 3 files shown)


wjwwood:
lgtm, we should run CI over it before merging though.

wjwwood:
Here are some CI jobs for your pr:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1675/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1246/)

rohbotics:
The only failing test is from this issue:
https://github.com/osrf/osrf_pycommon/issues/25

rclcpp/include/rclcpp/node.hpp
```diff
@@ -353,6 +353,9 @@ class Node : public std::enable_shared_from_this<Node>
   size_t
   count_graph_users();
 
+  template<typename CallbackT>
+  void on_param_change(CallbackT && callback);
```
dirk-thomas: I would expect an actual callback function to have such a name (`on_param_change`). For this I think `register_param_change_callback` (or similar) would be better.

Also from the name itself it is not clear that repeated invocations will overwrite previous callback functions. That should at least be documented in a docblock. Maybe there should also be a way to unregister a callback again?

rohbotics:
@dirk-thomas Updated with your suggestions

dirk-thomas:
Thanks, lgtm.

wjwwood:
I think I fixed the `osrf_pycommon` issue. I started new CI since some changes were made here as well:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1677/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1247/)

wjwwood:
lgtm

Issue #54, opened 2015-07-10
allow to start a node without a ParameterEvent publisher


dirk-thomas:
For unit tests it would be good to start nodes without any additional overhead as it simplifies debugging.

https://github.com/ros2/rclcpp/blob/5baa5195dbad2ab04a571d3f51ca39d0df5a1bd1/rclcpp/include/rclcpp/node_impl.hpp#L60-L61

wjwwood:
I believe this is done in https://github.com/ros2/rclcpp/pull/614, though I don't think any tests are making use of this at the moment.

PR #245, opened 2016-07-29
Get single parameter

rohbotics:
Added an API for getting a single parameter.

Note: this PR is based on master, and might conflict with my other PR

### DIFF
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -373,12 +373,11 @@ bool Node::get_parameter(const std::string & name, ParameterT & parameter) const
 {
   std::lock_guard<std::mutex> lock(mutex_);
 
-  if (parameters_.count(name)) {
-    parameter = parameters_.at(name).get_value<ParameterT>();
-    return true;
-  } else {
-    return false;
-  }
+  rclcpp::parameter::ParameterVariant parameter_variant(parameter);
+  bool result = get_parameter_(name, parameter_variant);
+  parameter = parameter_variant.get_value<ParameterT>();
+
+  return result;
 }
 
 }  // namespace node
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -371,10 +371,8 @@ void Node::register_param_change_callback(CallbackT && callback)
 template<typename ParameterT>
 bool Node::get_parameter(const std::string & name, ParameterT & parameter) const
 {
-  std::lock_guard<std::mutex> lock(mutex_);
-
   rclcpp::parameter::ParameterVariant parameter_variant(name, parameter);
-  bool result = get_parameter_(name, parameter_variant);
+  bool result = get_parameter(name, parameter_variant);
   parameter = parameter_variant.get_value<ParameterT>();
 
   return result;
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -368,6 +368,19 @@ void Node::register_param_change_callback(CallbackT && callback)
   this->parameters_callback_ = callback;
 }
 
+template<typename ParameterT>
+bool Node::get_parameter(const std::string & name, ParameterT & parameter) const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  if (parameters_.count(name)) {
+    parameter = parameters_.at(name).get_value<ParameterT>();
+    return true;
+  } else {
+    return false;
+  }
+}
+
 }  // namespace node
 }  // namespace rclcpp
 
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -373,7 +373,7 @@ bool Node::get_parameter(const std::string & name, ParameterT & parameter) const
 {
   std::lock_guard<std::mutex> lock(mutex_);
 
-  rclcpp::parameter::ParameterVariant parameter_variant(parameter);
+  rclcpp::parameter::ParameterVariant parameter_variant(name, parameter);
   bool result = get_parameter_(name, parameter_variant);
   parameter = parameter_variant.get_value<ParameterT>();
 
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -245,14 +245,12 @@ Node::get_parameters(
 const rclcpp::parameter::ParameterVariant
 Node::get_parameter(const std::string & name) const
 {
-  std::lock_guard<std::mutex> lock(mutex_);
-
   rclcpp::parameter::ParameterVariant parameter;
 
-  if (get_parameter_(name, parameter)) {
+  if (get_parameter(name, parameter)) {
     return parameter;
   } else {
-    throw std::out_of_range("Parameter does not exist in map");
+    throw std::out_of_range("Parameter: " + name + " does not exist in map");
   }
 }
 
@@ -261,12 +259,6 @@ bool Node::get_parameter(const std::string & name,
 {
   std::lock_guard<std::mutex> lock(mutex_);
 
-  return get_parameter_(name, parameter);
-}
-
-bool Node::get_parameter_(const std::string & name,
-  rclcpp::parameter::ParameterVariant & parameter) const
-{
   if (parameters_.count(name)) {
     parameter = parameters_.at(name);
     return true;
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -250,6 +250,19 @@ Node::get_parameter(const std::string & name) const
   return parameters_.at(name);
 }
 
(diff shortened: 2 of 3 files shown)


wjwwood:
Should this function have an optional "default value" parameter? Or should we have a separate function that provides that syntax sugar?

wjwwood:
Sorry, looks like merging the other pr makes this one conflict. You'll need to rebase it.

rohbotics:
@wjwwood Rebased

Do you think the default value should be on all three versions of the function, or just on the one that returns a parameter?

dirk-thomas:
Adding more functions (in this case three different signatures) might be helpful for the user but imo it increases the long term maintenance burden since actual functionality and syntactic sugar is mixed (as I mentioned in https://github.com/ros2/rclcpp/pull/233#issuecomment-230037130).

From my point of view we should have one component with a minimal interface providing and implementing all the core functionality (call it a "micro kernel"). Any additional API for syntactic sugar should be separated.

rohbotics:
I thought that the minimal interface as supposed to be provided by rcl, and that rclcpp is the more user-friendly frontend.

If we want to keep the node class minimal, we could move all parameter functionality to the `parameter_service`, and have it have all the syntactic sugar.

That way any node that doesn't pull in the `parameter_service` isn't a parameter server, so doesn't have any internal map for storing parameters, which is currently in the node. We can also move the `parameter_events` publisher to the `parameter_service`, (Connects to #54). 

The downside to this approach would be that in order to pass a handle around for creating publishers/subscribers and getting/setting parameters, you would need a pointer to both the `parameter_service` and the node. We could mitigate this by having the node store a smart_ptr to a passed in `parameter_service`, or have a function that makes a `parameter_service` internal to the node.

rohbotics:
Would you say that maintainability would be increased if instead of the 3 signatures being completely independent/freestanding, they were composed on top of each-other?

rohbotics:
Sorry, I hit the close and comment instead of the comment button

wjwwood:
@rohbotics I have a different opinion from @dirk-thomas: this is the highest level interface and the appropriate place to have convenience functions. We can choose to back these functions with a minimal interface in the future if we choose. I do agree that you should try to have the functions call each other where possible. It looks like at least two of them are basically duplicates.

rohbotics:
How should I handle the locking between functions if they call each other?
Right now they all lock the same mutex, so if one of them locks it and then calls another, the callee will wait forever on the mutex, as the caller is holding it until it returns.

tfoote:
The higher level one does not need to lock if it's just calling the other one that locks.

rohbotics:
I made all the functions lock, then call a common non-locking private function.

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -368,6 +368,18 @@ void Node::register_param_change_callback(CallbackT && callback)
...
 }
 
+template<typename ParameterT>
+bool Node::get_parameter(const std::string & name, ParameterT & parameter) const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  rclcpp::parameter::ParameterVariant parameter_variant(parameter);
+  bool result = get_parameter_(name, parameter_variant);
+  parameter = parameter_variant.get_value<ParameterT>();
```
wjwwood: If `result` is `false` will this not raise?

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -368,6 +368,18 @@ void Node::register_param_change_callback(CallbackT && callback)
   this->parameters_callback_ = callback;
 }
 
+template<typename ParameterT>
+bool Node::get_parameter(const std::string & name, ParameterT & parameter) const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  rclcpp::parameter::ParameterVariant parameter_variant(parameter);
+  bool result = get_parameter_(name, parameter_variant);
```
wjwwood: Could you not call the `bool get_parameter(std::string, ParameterVariant)` version of the function here and avoid the mutex lock two lines above?

rclcpp/src/rclcpp/node.cpp
```diff
@@ -242,6 +242,39 @@ Node::get_parameters(
...
+bool Node::get_parameter(const std::string & name,
+  rclcpp::parameter::ParameterVariant & parameter) const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+
+  return get_parameter_(name, parameter);
+}
+
+bool Node::get_parameter_(const std::string & name,
+  rclcpp::parameter::ParameterVariant & parameter) const
```
wjwwood: Hmm, maybe you could expand on why you don't just lock this function as @tfoote suggested? Seems to me that you could move the lock in the other three functions to this one without any issues.

rclcpp/include/rclcpp/node_impl.hpp
rohbotics: This should not throw if `result == false`
That enables use like this
```
double foo = 0.1;
if (!get_parameter("foo", foo) {
  set_parameter("foo", foo);
}
```
We do throw if get_value fails because of differing types.

wjwwood:
Also, I'd like to clarify that I agree with @dirk-thomas in that we need to restructure the Node class and make a separation between high level interfaces and the foundation of those interfaces. However, I see it as a larger task that goes beyond these couple of new functions, and that it shouldn't be a pre-requisite for opportunistically taking some contributions like this one.

rclcpp/include/rclcpp/node_impl.hpp
wjwwood: But if `result` is `false`, then the `ParameterVariant` will be "not set" right? So `get_value` will raise I think (that was what I was referring to before).

wjwwood: After talking with @rohbotics I think I might be wrong about `get_value` throwing because the `ParameterVariant` is seeded with the given `parameter` value.

rohbotics: The seeding doesn't compile with anything other than a string, so looks like I need to rethink this a little bit.

rohbotics:
@wjwwood I cleaned up a little, and we don't need the non-locking private function

I added tests here https://github.com/ros2/system_tests/pull/158 as we discussed

rclcpp/src/rclcpp/node.cpp
```diff
@@ -242,6 +242,31 @@ Node::get_parameters(
...
 
+const rclcpp::parameter::ParameterVariant
+Node::get_parameter(const std::string & name) const
+{
+  rclcpp::parameter::ParameterVariant parameter;
+
+  if (get_parameter(name, parameter)) {
+    return parameter;
+  } else {
+    throw std::out_of_range("Parameter: " + name + " does not exist in map");
```
wjwwood: nitpick: would you change this to `"Parameter '" + name + "' not set"`? We usually wrap variables output into string inside of quotes in case `name` is an empty string, for example. Also, the fact that a map is used is an implementation detail.

wjwwood:
Other than my comment, this lgtm. I still need to review the tests you added (thanks for those btw), and then run CI.

rohbotics:
I fixed the exception message

rclcpp/src/rclcpp/node.cpp
```diff
@@ -242,6 +242,31 @@ Node::get_parameters(
...
 
+const rclcpp::parameter::ParameterVariant
+Node::get_parameter(const std::string & name) const
+{
+  rclcpp::parameter::ParameterVariant parameter;
+
+  if (get_parameter(name, parameter)) {
+    return parameter;
+  } else {
+    throw std::out_of_range("Parameter '" + name + "' not set");
```
wjwwood: Also, `out_of_range` may not be appropriate here. Instead maybe a custom exception should be used... I think if this were a method on a "ParameterSet" or "ParameterMap" object it might make sense because `out_of_range` is the functional reason the parameter cannot be returned (the given key resulted in a "key error"), but it's not really the non-functional reason. In the context of a user asking for a parameter, it might make sense to throw a more domain specific exception type.

I've not convinced myself enough to suggest that you should change it right now, but maybe others can give their opinions.

wjwwood:
I started CI for you:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1678/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1248/)

wjwwood:
@rohbotics looks like your new tests have some compiler warnings: http://ci.ros2.org/job/ci_linux/1678/warnings20Result/new/

wjwwood:
I kicked off a new Linux CI to check the warnings:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1679/)

dirk-thomas:
One new warning on Windows: http://ci.ros2.org/job/ci_windows/1609/warnings39Result/new/

rohbotics:
Fixed warning and test on Windows

@wjwwood, looks like on windows you have to use C++11's <cstdint> instead of C99's <stdint.h>, or you need a specific include order

wjwwood:
@rohbotics Yeah, makes sense. CI looks good, thanks for following up on this.

PR #253, opened 2016-09-09
removed extra semi-colon

geoffviola:
removed some warnings when compiling with -Wpedantic with g++ (Ubuntu 5.4.0-6ubuntu1~16.04.2) 5.4.0 20160609.

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -10,8 +10,8 @@ find_package(rmw_implementation REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
-if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+if(CMAKE_COMPILER_IS_GNUCXX)
+  set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11;-Wall;-Wextra;-Wpedantic")
 endif()
 
 include_directories(include)
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -10,7 +10,7 @@ find_package(rmw_implementation REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
-if(CMAKE_COMPILER_IS_GNUCXX)
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11;-Wall;-Wextra;-Wpedantic")
 endif()
 
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -11,7 +11,7 @@ find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11;-Wall;-Wextra;-Wpedantic")
+  add_compile_options(-std=c++11 -Wall -Wextra -Wpedantic)
 endif()
 
 include_directories(include)
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -28,7 +28,7 @@ namespace executor
 
 struct AnyExecutable
 {
-  RCLCPP_SMART_PTR_DEFINITIONS(AnyExecutable);
+  RCLCPP_SMART_PTR_DEFINITIONS(AnyExecutable)
 
   RCLCPP_PUBLIC
   AnyExecutable();
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -49,7 +49,7 @@ class CallbackGroup
   friend class rclcpp::node::Node;
 
 public:
-  RCLCPP_SMART_PTR_DEFINITIONS(CallbackGroup);
+  RCLCPP_SMART_PTR_DEFINITIONS(CallbackGroup)
 
   RCLCPP_PUBLIC
   explicit CallbackGroup(CallbackGroupType group_type);
@@ -79,7 +79,7 @@ public:
   type() const;
 
 private:
-  RCLCPP_DISABLE_COPY(CallbackGroup);
+  RCLCPP_DISABLE_COPY(CallbackGroup)
 
   RCLCPP_PUBLIC
   void
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -50,7 +50,7 @@ namespace client
 class ClientBase
 {
 public:
-  RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(ClientBase);
+  RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(ClientBase)
 
   RCLCPP_PUBLIC
   ClientBase(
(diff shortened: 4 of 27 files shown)


dirk-thomas:
Should we add `-Wpedantic` to the [flags](https://github.com/ros2/rclcpp/blob/6ea435f743e8ee54f367a8eb51cdbfa9910d9f17/rclcpp/CMakeLists.txt#L14) to prevent regressions like this in the future? That would probably be relevant for all packages.

mikaelarguedas:
+1 for adding it to the flags for all packages, it will be much easier to comply early on rather than cleaning the entire codebase afterwards

geoffviola:
I would appreciate that. I intend to compile the headers with "-Wall -Wextra -pedantic" on GCC, "/W4" in MSVC, and "-Weverything" on clang.

wjwwood:
@geoffviola can you add the flag to the `CMakeLists.txt` in this pr?

rclcpp/CMakeLists.txt
```diff
@@ -10,8 +10,8 @@ find_package(rmw_implementation REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
-if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+if(CMAKE_COMPILER_IS_GNUCXX)
```
dirk-thomas: Isn't this condition not matching `clang` anymore? The previous flags should still be applied in that case.

geoffviola: That's correct. The only common flag should be -std=c++11. Should support for clang be added? I can test that.

dirk-thomas: Before we were use all the options for both compilers. Why do you want to change the condition in the first place? Does `clang` have a problem with `-Wpedantic`?

geoffviola: Clang has included -Wall, -Wextra, -Wpedantic, and maybe a few other warnings into -Weverything

dirk-thomas: So we should either keep the original condition or if we want to use `Weverything` with `clang` add a separate block to set the clang specific flags.

geoffviola: Yes. I would have to test if -Weverything will introduce new warnings.

wjwwood:
Sorry for letting this slip, lgtm, but I cannot test it properly unless it is rebased on master, @geoffviola do you think you could do that? Otherwise I'll set up a branch on this repository with the rebased version of this for testing.

geoffviola:
I tried a rebase. Let me know if it is usable.

dirk-thomas:
The list of commits shows a lot of commits which are not related to your patch. It looks like something went wrong when you rebased.

geoffviola:
It looks like the final "git pull" command did something funny. I did another rebase. I think it looks better now.

rclcpp/CMakeLists.txt
```diff
@@ -10,8 +10,8 @@ find_package(rmw_implementation REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
-if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11;-Wall;-Wextra;-Wpedantic")
```
dirk-thomas: What is the reason to change the condition as well as the way to extend the c++ flags? Couldn't you just add ` -Wpedantic` to the existing line without changing it otherwise?

geoffviola: Changing the condition is meant to only add these flags to compilers that support these flags. Clang is included explicitly, because that is a targeted compiler that understands Wpedantic.
I used set_directory_properties, because I thought it was a little more precise. I switched over to using add_compile_options instead, because I thought it was clearer.

wjwwood:
Linux CI looks ok (failure is known issue):

[Build Status](http://ci.ros2.org/job/ci_linux/1952/)

Running OS X and Windows now, will merge if they are ok.

mikaelarguedas:
Should we propagate the pedantic flag to all the other c++ packages? (in a follow-up PR)

wjwwood:
Personally, I don't feel we need the pedantic flag, as it often points out harmless things or issues that are hard to resolve in a clean way. I'm ok with it here, since it is a relatively easy fix to comply to it, but if in the future it starts complaining at us over something frustrating, then I'd be ok removing it again. Along those lines, I don't see a reason to enable it in the other packages, but I wouldn't be opposed to it either as long as it doesn't require strange workarounds in our code to squelch the warnings.

geoffviola:
Pedantic seems to follow the lines of uncrustify and cpplint. @wjwwood, are there certain warnings from pedantic that you don't like in particular? I actually like the warning about extra semi colons, because it indicates a code smell.

wjwwood:
I don't have specific examples off hand, but in the case of multiple semi colons, I obviously don't want `;;` in the code, but you could imagine a case where a macro expands to something that needs a `;` in some cases and in others (like debug versus release) where it expands to something else it doesn't need it. It would potentially be burdensome to deal with that without turning off pedantic.

I guess my point is, we don't need it but so long as it doesn't add a burden to support it we can have it.

dirk-thomas:
I would be in favor of trying to enable it and look at the potential warnings and necessary updates. :+1:

wjwwood:
CI looks good:

OS X [Build Status](http://ci.ros2.org/job/ci_osx/1504/)
Windows [Build Status](http://ci.ros2.org/job/ci_windows/1910/)

As I said, I don't mind it, but I also don't think we need it. Adding more style checking is by far not the most important thing we could spend out time on right now, but I'm happy to take it as a contribution.

Karsten1987:
This PR has problems on OSX when generating messages at the same time. This option passes all definitions to every file; including C files. Clang complains with this:
```
01:02:17 error: invalid argument '-std=c++11' not allowed with 'C/ObjC'
```

I am reverting the c++11 flags to stay within the `CMAKE_CXX_FLAGS` scope:
https://github.com/ros2/rclcpp/pull/279/commits/0de1a957e0820da92cb7913c742f148128380a1c

geoffviola:
@Karsten1987,
Can you elaborate with some reproducible steps?
You're definitely right about clang not accepting C++ flags for C files, but I don't see C files in this package. Do the flags enter other projects with ament_target_dependencies or are you using the CMake commands in another CMakeLists.txt?

Karsten1987:
our CI fails on OSX with it:
http://ci.ros2.org/job/ci_osx/1509/

I am generating a msg for a test within this package.

geoffviola:
@Karsten1987, that change looks good.

It took me a while to see what was going on. That behavior is reproducible by using the latest code, setting the rclcpp project's flag to std=c++ for all targets again, and building with the tests.
When `--cmake-args -DCMAKE_VERBOSE_MAKEFILE=ON` is passed into ament build, the error is more obvious
```
[ 20%] Building C object CMakeFiles/mock_msgs__rosidl_generator_c.dir/rosidl_generator_c/rclcpp/srv/mock__request__functions.c.o
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc  -Dmock_msgs__rosidl_generator_c_EXPORTS -Iros2/src/ros2/rclcpp/rclcpp/include -Iros2/build_isolated/rclcpp/rosidl_generator_c -Iros2/install_isolated/rosidl_generator_c/include  -fPIC   -std=c11 -Wall -Wextra -std=c++11 -Wall -Wextra -Wpedantic -o CMakeFiles/mock_msgs__rosidl_generator_c.dir/rosidl_generator_c/rclcpp/srv/mock__request__functions.c.o   -c ros2/build_isolated/rclcpp/rosidl_generator_c/rclcpp/srv/mock__request__functions.c
error: invalid argument '-std=c++11' not allowed with 'C/ObjC'
```

The problem was that a .c file gets generated and compiled, but the -std=c++11 flag was still added  for both .c and .cpp files. There is a way to do the same thing with add_compile_options, but it wouldn't improve anything: http://stackoverflow.com/a/35361099. Ultimately, this functionality should probably be a CMake function with a flag for Windows to use more warnings as well.

PR #254, opened 2016-09-23
(dev) add simple sub/pub template

Karsten1987:
I added two simple templates for initializing a node with a single publisher or subscriber.
A node can easily be instantiated and added to the executor. The user only has to implement the constructor and the `publish` or `callback` function respectively. 

example publisher: https://github.com/Karsten1987/my_ros2_demos/blob/master/src/laser_publisher_node.hpp

example subscriber:
https://github.com/Karsten1987/my_ros2_demos/blob/master/src/laser_subscriber_node.hpp

If such an example would I make sense, I iterate more over it (test, style, etc.)

connects to https://github.com/ros2/examples/issues/129

### DIFF
--- a/rclcpp/include/rclcpp/simple_publisher_node.hpp
+++ b/rclcpp/include/rclcpp/simple_publisher_node.hpp
@@ -1,0 +1,49 @@
+#pragma once
+
+#include "rclcpp/node.hpp"
+#include "rclcpp/publisher.hpp"
+
+namespace rclcpp
+{
+namespace node
+{
+
+template<class MessageT>
+class SimplePublisherNode : public rclcpp::Node
+{
+public:
+  SimplePublisherNode( std::string node_name, std::string topic, size_t frequency ):
+    Node(std::move(node_name), true),
+    topic_(std::move(topic)),
+    frequency_(frequency)
+  {
+    pub_ = this->create_publisher<MessageT>(topic_, rmw_qos_profile_default);
+  }
+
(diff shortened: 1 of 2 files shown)


wjwwood:
If the intention is to have some syntactic sugar to make it easier to create a single publisher or single subscription node, then it would probably be appropriate to be in `rclcpp` or possibly in a small package that provides this template on top of `rclcpp` (like `rclcpp_helpers` or `rclcpp_pubsub_helpers`). One reason to consider putting it in a separate package from `rclcpp` is to minimize the API surface area of the `rclcpp` package (a shorter answer to "what is the rclcpp API?"). Also, the level of documentation and/or testing we will want to require for `rclcpp` may not make sense for helper patterns like this one. That being said, `rclcpp` is the place for nice C++ interfaces that we recommend people use, so I'd say the deciding factor should be based on how often we find ourselves using a pattern like this.

If you mean it to be an example for others to copy and paste and modify to become their code, then we have the `rclcpp_examples` package in the `ros2/examples` repository: https://github.com/ros2/examples. It's a more appropriate place for examples in my opinion, but it sounds like you mean it to be a helper pattern that is used by others and not copy-pasted.

gbiggs:
Having a separate package for various common node type templates would be nice. I think there would need to be more than two to make it worthwhile, though.

Karsten1987:
Can i close this PR in favor of the cookbook?
https://github.com/ros2/cookbook

PR #255, opened 2016-10-01
comply with stricter cpplint rules

dirk-thomas:
Update style to pass new cpplint check (ament/ament_lint#59).

http://ci.ros2.org/job/ci_linux/1763/


rclcpp/src/rclcpp/memory_strategies.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include "rclcpp/memory_strategies.hpp"
 
+#include <memory>
```
dirk-thomas: The same question as https://github.com/ros2/demos/pull/80#discussion_r81456795 applies here.

PR #258, opened 2016-10-26
open Node/Publisher API for allowing inheritance

Karsten1987:
change:
- `Node::create_publisher` function has a new template parameter for publisher type (`rclcpp::publisher::Publisher` by default)
- all `publish` functions are virtual for being able to inherit from

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -123,7 +123,8 @@ public:
    * \param[in] qos_history_depth The depth of the publisher message queue.
    * \return Shared pointer to the created publisher.
    */
-  template<typename MessageT, typename Alloc = std::allocator<void>, typename PublisherT = ::rclcpp::publisher::Publisher<MessageT, Alloc>>
+  template<typename MessageT, typename Alloc = std::allocator<void>,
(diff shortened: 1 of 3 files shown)


wjwwood:
> Node gets extended with one template parameter for publisher type 

Isn't it more accurate to say that the `Node::create_publisher` method has a new template parameter? Node is still not templated as far as I can see.

Karsten1987:
You're absolutely right. Corrected.

Karsten1987:
updated

Karsten1987:
build succesful http://ci.ros2.org/job/ci_linux/lastBuild/

AravindaDP:
Probably we should open client API also for inheritance?

I'm specifically interested in marking async_send_request call as virtual.

Note: What's I have in mind is for the unit testing purposes. (Mocking) Correct me if it is wrong way to handle things.

wjwwood:
@AravindaDP sorry it took so long to get back to you on this. If you'd like to see additional methods marked virtual, or a function refactored to take one of the arguments as a template, please just open a pr to that effect or open an issue for each case so we can evaluate them there.

AravindaDP:
@wjwwood no worries. Actually I wasn't sure about the correct place I should open the discussion.
I'll open an issue so we can evaluate the best approach to achieve the requirement there.

PR #265, opened 2016-11-01
Lifecyle node and C state_machine

Karsten1987:
up for design review! 
Not crustified yet. Not tested yet (except the demo). First prototype of a lifecycle node demo.

- `rcl_lifecycle` contains a C based state machine with primary states and transitions as depicted in [design.ros2.org](design.ros2.org). 

- `rclcpp_lifecycle` contains the higher level c++ implementation of a managed node. A managed node has a `rcl_state_machine_t` object as member for keeping track of the state and possible transitions. Further, it has the same API as a `rclcpp::Node` for creating publishers etc. However, each managed node has a handle to all pub/sub/srv/clients in order to activate/deactivate them.

- `lifecycle_talker.cpp` is a first demo application. Missing here is any kind of container which opens services for calling the transition functions such as `on_configure()` or `on_activate()`, etc.

First commits are equal to https://github.com/ros2/rclcpp/pull/258


rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+};
+}  // namespace lifecycle_interface
+
+#include <rclcpp_lifecycle/visibility_control.h>
+
+/**
+ * @brief LifecycleNode as child class of rclcpp Node
+ * has lifecycle nodeinterface for configuring this node.
+ */
+class LifecycleNode : public rclcpp::node::Node, public lifecycle_interface::NodeInterface
```
gbiggs: This is starting to show the problem of not having a `Node` interface. "Favor 'object composition' over 'class inheritance'" is often thrown around and abused, but C++ best practice generally says to not inherit implementations except for minor tweaks to behaviour. It may work as is for this case, but for a major entity like Node, having an abstract interface would be more flexible, and would make things more testable.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+  LIFECYCLE_EXPORT
+  explicit LifecycleNode(const std::string & node_name, bool use_intra_process_comms = false) :
+    Node(node_name, use_intra_process_comms)
+  {
+    setup_state_machine();
+  }
+
+  LIFECYCLE_EXPORT
+  ~LifecycleNode(){}
```
gbiggs: `= default`?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+        topic_name, qos_profile, allocator);
+
+    // keep weak handle for this publisher to enable/disable afterwards
+    weak_pub_ = pub;
+    return pub;
+  }
+
+  LIFECYCLE_EXPORT
+  void
+  print_state_machine()
```
gbiggs: Being able to get the transition map as a string that the caller can do with as they please might be more flexible.

A function in the same namespace could be added for actual printing. It would call the class method to get the transition map as a string and print it out.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+    if (weak_pub_.expired())
+    {
+      fprintf(stderr, "I have no publisher handle\n");
+      // TODO: go to error state here
+      return false;
+    }
+
+    // Placeholder print for all configuring work to be done
+    // with each pub/sub/srv/client
+    printf("I am doing a lot of activation work\n");
```
gbiggs: Buggy debug print!

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+  virtual bool
+  on_error()
+  {
+    return true;
+  }
+
+private:
+
+  // weak handle for the managing publisher
+  // TODO: Has to be a vector of weak publishers. Does on_deactivate deactivate every publisher?!
```
gbiggs: Or a map, indexed by topic?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -0,0 +1,171 @@
...
+{
+/**
+ * @brief base class with only
+ * pure virtual functions. A managed
+ * node can then deactivate or activate
+ * the publishing.
+ * It is more a convenient interface class
+ * than a necessary base class.
+ */
+class PublisherInterface
```
gbiggs: "ManagedLifecyclePublisherInterface" rather than using a namespace? Although as a pattern begins to emerge the use of a namespace doesn't look as bad, I don't think it is necessary nor as neat as naming the interface classes directly.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+      fprintf(stderr, "I have no publisher handle\n");
+      state_machine_.current_state = rcl_state_error;
+      return false;
+    }
+
+    // Placeholder print for all configuring work to be done
+    // with each pub/sub/srv/client
+    printf("I am doing a lot of activation work\n");
+    // TODO: does a return value make sense here?
+    weak_pub_.lock()->on_activate();
```
gbiggs: Race condition: weak_pub_ may have been deleted between the check on `weak_pub_.expired()` and this statement. Use something like:

    auto pub = weak_pub_.lock();
    if (pub) {
      pub->on_activate();
    }
    else {
      [Your error stuff]
    }

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+      fprintf(stderr, "I have no publisher handle\n");
+      // TODO: go to error state here
+      return false;
+    }
+
+    // Placeholder print for all configuring work to be done
+    // with each pub/sub/srv/client
+    printf("I am doing a lot of activation work\n");
+    // TODO: does a return value make sense here?
+    weak_pub_.lock()->on_deactivate();
```
gbiggs: See comment on the same race condition in `on_activate()`.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -0,0 +1,171 @@
...
+  LifecyclePublisher(
+      std::shared_ptr<rcl_node_t> node_handle,
+      std::string topic,
+      const rcl_publisher_options_t & publisher_options,
+      std::shared_ptr<MessageAlloc> allocator):
+    ::rclcpp::publisher::Publisher<MessageT, Alloc>(node_handle, topic, publisher_options, allocator),
+      enabled_(false)
+  {  }
+
+  ~LifecyclePublisher() {};
```
gbiggs: `= default`?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -0,0 +1,171 @@
...
+   * @brief briefly checks whether this publisher
+   * was enabled or disabled and forwards the message
+   * to the actual rclcp Publisher base class
+   */
+  virtual void
+  publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
+  {
+    if (!enabled_)
+    {
+      printf("I push every message to /dev/null\n");
```
gbiggs: The design document leaves undefined what to do with publications while the node is inactive. The choice made here is to make the message disappear. Given that subscriptions will determine what to do based on the configured QoS, is this code the right choice for publications?

rclcpp_lifecycle/include/rclcpp_lifecycle/type_traits/is_manageable_node.hpp
```diff
@@ -0,0 +1,54 @@
...
+{
+  static constexpr bool value = true;
+};
+
+template<class T, typename = void>
+struct is_manageable_node : std::false_type
+{};
+
+template<class T>
+struct is_manageable_node<T, typename std::enable_if< has_on_activate<T>::value
```
gbiggs: Big yes to doing things this way!

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -0,0 +1,220 @@
...
+  void
+  print_state_machine()
+  {
+    printf("current state is %s\n", state_machine_.current_state.label);
+    rcl_print_transition_map(&state_machine_.transition_map);
+  }
+
+  LIFECYCLE_EXPORT
+  virtual bool
+  on_configure()
```
gbiggs: This is not the right name for the external API for controlling the state machine.

This implementation mixes life cycle control (when to configure) and node-specific life cycle functionality (such as what to do in the `Configuring` transition state/`onConfigure()` callback) into a single function.

This function should be called `configure` and there should be a separate method, implemented by users who inherit from this class to define their own managed nodes, called `onConfigure`. In that function will go all their configuring stuff. In this function will go the life cycle management stuff (which is here already) and any configuration that should be done for every node.

This comment applies to the other transition functions as well.

rclcpp_lifecycle/src/lifecycle_talker.cpp
```diff
@@ -0,0 +1,87 @@
...
+ 
+  rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
+  custom_qos_profile.depth = 7;
+
+  std::shared_ptr<rclcpp::publisher::LifecyclePublisher<std_msgs::msg::String>> chatter_pub =
+    lc_node->create_publisher<std_msgs::msg::String>("chatter", custom_qos_profile);
+
+  lc_node->print_state_machine();
+
+  if (!lc_node->on_configure())
```
gbiggs: The external API of a managed lifecycle node for controlling transitions between these states is not the callbacks. There should be an API containing the 7 transitions mentioned in the design document (`create`, `configure`, etc.) When, for example, `configure()` is called on the node, it should execute the `on_configure()` callback and shift its state to the appropriate successor state based on the result.

See also my comment in the managed lifecycle node implementation.

gbiggs: It would be good if we can minimise how much the user has to manually control the node's life cycle. Possibly that will come through another API layered on top, which this talker example should be modified to use when it's done.

rclcpp_lifecycle/include/rcl_lifecycle/lifecycle_state.h
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * @brief: statemachine object holding
+ * a variable state object as current state
+ * of the complete machine.
+ * @param transition_map: a map object of all
+ * possible transitions registered with this
+ * state machine.
+ */
+typedef struct LIFECYCLE_EXPORT _rcl_state_machine_t
```
gbiggs: Do you intend to implement managed lifecycle nodes in rcl as well?

If not, what is the purpose of writing the state machine in C rather than C++, or rather than using something like Boost's state machine library?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
Karsten1987: I agree. The `node` implementation needs a refreshing. @wjwwood is currently working on a refactoring which addresses the things you mentioned such as composition vs. inheritance.

rclcpp_lifecycle/include/rcl_lifecycle/transition_map.h
```diff
@@ -0,0 +1,101 @@
...
+} rcl_transition_array_t;
+
+/**
+ * @brief helper index for keeping
+ * track of the map content
+ */
+typedef struct LIFECYCLE_EXPORT _index
+{
+  unsigned int index;
+  const char* label;
```
Karsten1987: `label` here was meant to be a human readable string such as "active" or "inactive".

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
Karsten1987: the idea here is going to be replicated for all pub/sub/srv/client. Generally, I would vote for cutting all communication when in `inactive` state and re-enable it once the node is in state `active`. But yes, this is up to discussion here.

rclcpp_lifecycle/include/rclcpp_lifecycle/type_traits/is_manageable_node.hpp
Karsten1987: That will work on compile time. However, a base class is needed at some point for dynamic runtime loading of nodes. See https://github.com/ros2/demos/pull/84 as an example.

rclcpp_lifecycle/src/lifecycle_talker.cpp
Karsten1987: There is no external API at this point. I put the callback calls in the talker example for demonstration. I see the talker code here as a lifecycle manager or similar which itself then has a public external API for triggering configure, activate, etc. (also via service calls)

rclcpp_lifecycle/include/rcl_lifecycle/lifecycle_state.h
Karsten1987: Good point. I indeed wrote this in C for multilanguage purposes such that every rcl* has the same set of states and valid transitions. 
I believe ROS2 generally tries not to rely on boost.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
dirk-thomas: We certainly need something more fine grain. E.g. in *inactive* state you might want to configure some node parameters which could happen through service calls. Maybe when creating a lifecycle publisher / subscriber / server / client you can optionally pass a policy kind of information to allow to bypass the restrictions which are in place by default?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
gbiggs: Yes, the external API is what is missing. Your prototype's example gives the impression that you are treating the external and internal APIs the same.

I think that there should be a node interface that defines the external API. When the node is registered with a `LIfecycleManager` or something like that, then it would work with that API. A user could also manage their node's lifecycle directly from their main function, if they wish. Simultaneously, there needs to be a base class with virtual void methods (`on_activate` etc) for the implementer of a specific node to override to provide their own functionality.

rclcpp_lifecycle/include/rcl_lifecycle/default_state_machine.h
```diff
@@ -0,0 +1,37 @@
...
+{
+#endif
+
+// primary states based on
+// design.ros2.org/articles/node_lifecycle.html
+/*static*/ const rcl_state_t LIFECYCLE_EXPORT rcl_state_unconfigured = {/*.state = */0, /*.label = */"unconfigured"};
+/*static*/ const rcl_state_t LIFECYCLE_EXPORT rcl_state_inactive     = {/*.state = */1, /*.label = */"inactive"};
+/*static*/ const rcl_state_t LIFECYCLE_EXPORT rcl_state_active       = {/*.state = */2, /*.label = */"active"};
+/*static*/ const rcl_state_t LIFECYCLE_EXPORT rcl_state_finalized    = {/*.state = */3, /*.label = */"finalized"};
+/*static*/ const rcl_state_t LIFECYCLE_EXPORT rcl_state_error        = {/*.state = */4, /*.label = */"error"};
```
gbiggs: I think you could argue that the transition states don't need to be actual states. The implementation of the state machine just needs to make sure the correct functions are called as it transitions between the "main" states.

However, having them as actual states would be useful for debugging, and it might make it easier to understand the execution time usage of a managed node that has real-time characteristics.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
gbiggs: I agree with @dirk-thomas. The life cycle is intended to still allow things to happen in the `inactive` state, such as changing the behaviour of the node via parameters, that could not be done while the node is running its normal behaviour. A particular use case is a real-time node, where only the `active` state needs to actually be real-time safe. You would put your nodes into the `inactive` state, re-configure them, then re-activate them and return to normal behaviour.

So the user needs some control over which bits of communication get cut off and which do not. How that should be done, I have only vague ideas and no real proposals. :)

rclcpp_lifecycle/include/rclcpp_lifecycle/type_traits/is_manageable_node.hpp
gbiggs: The two ideas are not incompatible.

rclcpp_lifecycle/src/lifecycle_talker.cpp
gbiggs: In that case, perhaps "talker" is not a good name for the demonstratation class.

rclcpp_lifecycle/include/rcl_lifecycle/lifecycle_state.h
gbiggs: I have no opinion on using or avoiding Boost, and I haven't actually looked at the Boost state machine library in quite a while. I just thought I would mention it as an option.

Anyway, I agree with having it in C so other languages can re-use the managed lifecycle as-is.

rclcpp_lifecycle/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+  SHARED
+  ${rcl_lifecycle_sources}
+)
+install(TARGETS rcl_lifecycle
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+)
+
+macro(targets)
+  if(NOT target_suffix STREQUAL "")
```
dirk-thomas: This condition is not necessary and doesn't do what you would expect. The condition is basically always true since `target_suffix` is never an empty string but undefined in that case. Just always calling `get_rclcpp_information` is fine.

rclcpp_lifecycle/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+)
+add_library(
+  rcl_lifecycle
+  SHARED
+  ${rcl_lifecycle_sources}
+)
+install(TARGETS rcl_lifecycle
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+)
```
dirk-thomas: Needs to install to `bin` on Windows / for `RUNTIME DESTINATION`.

dirk-thomas:
Maybe we should define the ROS-level interface of the managed node / the lifecycle manager. Similar to what is defined in [rcl_interfaces](https://github.com/ros2/rcl_interfaces) for the parameters.

rclcpp_lifecycle/CMakeLists.txt
Karsten1987: direct copy from https://github.com/ros2/examples/blob/master/rclcpp_examples/CMakeLists.txt#L30
we should generally define a CMakeLists.txt as an example.

Karsten1987:
@gbiggs I try to replicate basically what's depicted here: http://design.ros2.org/articles/node_lifecycle.html
I agree that these transition states are not completely defined. I just want to keep the vocabulary as is. But this may be open for change.

@dirk-thomas Do you propose to generate the c-style structs via message generation?

dirk-thomas:
I was thinking about the services to control the state machine as well as the events / messages when the state changes. Kind of the external API used for introspection / orchestration. I am not sure if it makes sense for the structs.

rclcpp_lifecycle/src/rclcpp_lifecycle/lifecycle_talker.cpp
```diff
@@ -0,0 +1,138 @@
...
+  {
+    return -1;
+  }
+
+  while (rclcpp::ok() && i <= 30) {
+    exe.spin_once(std::chrono::nanoseconds(0));
+    //loop_rate.sleep();
+    ++i;
+  }
+  */
```
dirk-thomas: What about these comments?

gbiggs:
The managed nodes design document only has an ambiguous description of the management interface. I had done some work on a more detailed description but it got put on hold when Tully said he was going to do the node lifecycle implementation soon (because I wanted to see if anything changed). So I dug that work up and finished it off: https://github.com/ros2/design/pull/99

Karsten1987:
@wjwwood I'd like to bring your attention to the last commit https://github.com/ros2/rclcpp/pull/265/commits/58ae12787a9e54979424fecd48359044efe72abb. Does this go along with what we discussed? Therefore, I'd separate the node into a `base_interface` and a `communication_interface`.

Then please note that in the current `talker.cpp` if have preprocessor #if/#else derivatives which indicate a possible user api - namely the decision on how strictly we take the DRY principle.

wjwwood:
e86473efbfc0b0bf6045ab58efd215d789d96a29 looks like the right direction to me.

Karsten1987:
See lifecycle_talker.cpp as a demo on how a lifecycle orchestration would look like.

Status so far:

* StateMachine in C
* basic support for Publisher/Server in RCL data structures
* LifecycleManager in CPP with two services `get_state` and `change_state`

TODO:

* refactor Pub/Sub/Srv/Clients for extending their constructors in taking rcl_data structures directly.
* write extensive tests
* clarify on whether to integrate this PR into `rcl/rclpp` or in a self containing package called `lifecycle`. 

Any thoughts on the latter?

Karsten1987:
This PR requires https://github.com/ros2/rclcpp/pull/279 to be merged.
And also a fix for ROSIDL message generation in C. @mikaelarguedas can you have a look at it?

```
diff --git a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
index 5fac344..f8ac1d6 100644
--- a/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
+++ b/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport.h
@@ -76,6 +76,7 @@ struct StringHelper<rosidl_typesupport_introspection_c__MessageMembers>
     std::string str;
     deser >> str;
     rosidl_generator_c__String * c_str = static_cast<rosidl_generator_c__String *>(field);
+    rosidl_generator_c__String__init(c_str);
     rosidl_generator_c__String__assign(c_str, str.c_str());
   }
 };

```

mikaelarguedas:
as discussed offline, we need to figure out why the `rosidl_generator_c__String` was not allocated properly before being passed to the fastrtps rmw_implementation. I proposed this quick fix just to get the thing unstuck but I think that reallocating memory in the vendor specific layers "in case it's not properly initialized" is not the proper fix. We should ensure at the `rcl/rclcpp` level that we are providing properly initialized data structures to **all** `rmw_implementation`s and not patch this in the vendor specific code if not necessary.

Could you run CI on this to see if the same behavior shows up on Connext?

Karsten1987:
I refactored this initial PR and extract 3 more dependent PRs on it. 

lifecycle_msgs: https://github.com/ros2/common_interfaces/pull/24
rcl_lifecycle: https://github.com/ros2/rcl/pull/91

further depends on two fixes:
rmw-fastrtps: https://github.com/ros2/rmw_fastrtps/pull/69
service in rclcpp: https://github.com/ros2/rclcpp/pull/279

wjwwood:
@Karsten1987 I pushed a fix for the import of the impl header which was causing your issue and I fixed one or two other small things. It is still not compiling because the signature for LifecyclePublusher's constructor is different from Publishers now, which means it cannot be interchanged with normal Publisher through the PublisherT template argument anymore. That's the next thing to fix, but I will look at it tomorrow if you're still having issues.

Karsten1987:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1562/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2022/)

Windows seems to be a unrelated flaky test

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -0,0 +1,401 @@
...
+  init()
+  {
+    state_machine_ = rcl_lifecycle_get_zero_initialized_state_machine();
+    rcl_ret_t ret = rcl_lifecycle_state_machine_init(
+      &state_machine_, node_base_interface_->get_rcl_node_handle(),
+      ROSIDL_GET_TYPE_SUPPORT(lifecycle_msgs, msg, TransitionEvent),
+      rosidl_typesupport_cpp::get_service_type_support_handle<ChangeStateSrv>(),
+      rosidl_typesupport_cpp::get_service_type_support_handle<GetStateSrv>(),
+      rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableStatesSrv>(),
+      rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableTransitionsSrv>(),
```
wjwwood: This is neat. 😄

It might be a good idea to explain why both C and C++ type supports are used, so that when read it again in the future it is less confusing.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -0,0 +1,401 @@
...
+      true);
+    if (ret != RCL_RET_OK) {
+      fprintf(stderr, "Error adding %s: %s\n",
+        node_base_interface_->get_name(), rcl_get_error_string_safe());
+      return;
```
wjwwood: Should this be an exception?

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -0,0 +1,401 @@
...
+    const std::shared_ptr<GetStateSrv::Request> req,
+    std::shared_ptr<GetStateSrv::Response> resp)
+  {
+    (void)header;
+    (void)req;
+    if (rcl_lifecycle_state_machine_is_initialized(&state_machine_) != RCL_RET_OK) {
+      resp->current_state.id =
+        static_cast<uint8_t>(lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN);
+      resp->current_state.label = "unknown";
+      return;
```
wjwwood: Since the state machine is initialized before the service, within the same function, is ever possible? Maybe instead of having an unknown state, this should just raise an exception if somehow it is called before the state machine is initialized.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -0,0 +1,401 @@
...
+    std::shared_ptr<GetAvailableStatesSrv::Response> resp)
+  {
+    (void)header;
+    (void)req;
+    if (rcl_lifecycle_state_machine_is_initialized(&state_machine_) != RCL_RET_OK) {
+      lifecycle_msgs::msg::State unknown_state;
+      unknown_state.id = static_cast<uint8_t>(lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN);
+      unknown_state.label = "unknown";
+      resp->available_states.push_back(unknown_state);
+      return;
```
wjwwood: Same comment as above.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -0,0 +1,401 @@
...
+    }
+    for (unsigned int i = 0; i < state_machine_.transition_map.size; ++i) {
```
wjwwood: nitpick: consider using `size_t`

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -0,0 +1,401 @@
...
+      } else {
+        // We call shutdown on the error state
+        rcl_lifecycle_trigger_transition(
+          &state_machine_, lifecycle_msgs__msg__Transition__TRANSITION_SHUTDOWN, true, true);
+      }
+    }
+    // This true holds in both cases where the actual callback
+    // was successful or not, since at this point we have a valid transistion
+    // to either a new primary state or error state
+    return true;
```
wjwwood: 👍

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -0,0 +1,401 @@
...
+        fprintf(stderr, "Caught exception in callback for transition %d\n",
+          it->first);
+        fprintf(stderr, "Original error msg: %s\n", e.what());
+        // maybe directly go for error handling here
+        // and pass exception along with it
+        cb_success = RCL_LIFECYCLE_RET_ERROR;
+      }
+    } else {
+      fprintf(stderr, "%s:%d, No callback is registered for transition %u.\n",
+        __FILE__, __LINE__, cb_id);
```
wjwwood: Is it appropriate to always print a (warning?) message to stderr when no callback is associated?

wjwwood: Perhaps it should throw?

PR #266, opened 2016-11-02
Fix has_invalid_weak_nodes

### LINKED ISSUE #264: has_invalid_weak_nodes can only ever be false
In allocator_memory_strategy.hpp in the function `collect_entities`, the variable `has_invalid_weak_nodes` looks to me like it can only ever be `false`.

https://github.com/ros2/rclcpp/blob/7494350ad24201bd0ef6f75b94760830948274cb/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp#L143

Is the correct behaviour for line 147 to set the value to `true`?

gbiggs:
Fixes #264/

My copy of the master for ROS 2 doesn't currently compile so I worked against Alpha8. The changes apply cleanly to master, however.

The need to call `rclcpp::init()` in the tests so that nodes can be created is a bit ugly. I'd welcome suggestions on how to improve that. (Making `Node` an interface so that mocks can be easily created would be a good approach, but also a significant change to the code base; I'm happy to do that work if there is interest.)

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -120,6 +120,13 @@ if(BUILD_TESTING)
       ${PROJECT_NAME}
     )
   endif()
+  ament_add_gtest(test_find_weak_nodes test/test_find_weak_nodes.cpp)
+  if(TARGET test_find_weak_nodes)
+    target_include_directories(test_find_weak_nodes PUBLIC
+      ${rcl_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_find_weak_nodes ${PROJECT_NAME})
+  endif()
 endif()
 
 ament_package(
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Ticket numbers in the title of PRs don't cross reference. Therefore I updated the title and you comment.

The current nightly builds of all default branches seems stable: http://ci.ros2.org/view/nightly/ Maybe you want to double check your local setup.

An interface for nodes definitely makes sense. @wjwwood is currently working on refactoring `rclcpp` and will likely do that too.

Regarding calling `init`: it is simply necessary to call this from a life cycle point of view. There might be various ways to call the function which might be shorter then what you used in this patch, e.g. https://github.com/ros2/system_tests/blob/07aa8fa2bf62ddefebf4c50f9a8c8348a6a860fb/test_rclcpp/test/test_executor.cpp#L227-L229

The patch looks good to me, but I will defer to @wjwwood to merge this.

dhood:
I kicked off CI with this branch:
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1793/)
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1878/)
OS X: [Build Status](http://ci.ros2.org/job/ci_osx/1413/)

dhood:
thanks for the fix and tests @gbiggs. We were delaying merging this branch while investigating an instability on OS X. It seems to have just been a race condition that has since resolved itself and this can be merged now.

PR #270, opened 2016-11-10
fix allocator type

dirk-thomas:
Otherwise this code fails the stricter static assertion added in libc++ 3.9.

```
In file included from /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/intra_process_manager_impl.hpp:22:
/usr/local/Cellar/llvm/3.9.0/bin/../include/c++/v1/map:820:5: error: static_assert failed "Allocator::value_type must be same type as value_type"
static_assert((is_same<typename allocator_type::value_type, value_type>::value),
```


dirk-thomas:
https://github.com/ros2/demos/pull/84#issuecomment-259836337 contains the jobs for this branch.

PR #275, opened 2016-11-17
Add functions to dump ParameterVariants to json encoded strings

tfoote:
This is a first step toward creating tests which can assert the expected values of parameters in json for easy debugging and comparison.

### DIFF
--- a/rclcpp/include/rclcpp/parameter.hpp
+++ b/rclcpp/include/rclcpp/parameter.hpp
@@ -265,6 +265,44 @@ private:
   rcl_interfaces::msg::ParameterValue value_;
 };
 
+/* Return a json encoded version of the parameter intended for a list. */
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@tfoote what is the status on this ?

tfoote:
These functions were a start toward being able to dump the parameter state to json so that we could easily write unit tests with the expected serialized json result in a file and assert they match. This was deprioritized so this implementation exists but the remainder of the testing capabilities were not implemented to test this.

ivanpauno:
@tfoote this sounds a bit outdated, can it be closed?

tfoote:
We can close this. There's nothing fundamental or ground breaking in this PR but it's a long way from being feature complete. Hopefully someone might find this and pick up the idea in the future. There are use cases but right now they're not pressing.

PR #277, opened 2016-11-17
break Node into several separate interfaces

wjwwood:
This is a work in progress, I've only converted some of the most basic functionality to what I'm calling the "NodeBaseInterface". This is a preview of what I'm doing for the other parts, for which I've sketched out "NodeTopicInterface", "NodeServiceInterface", "NodeGraphInterface", and "NodeParameterInterface".

I've run into a few issues, but specifically I have some work to do to resolve the ownership issues. This is demonstrated by this:

https://github.com/ros2/rclcpp/compare/node_interfaces?expand=1#diff-14f85fa16e78532d255620f1d3529debR52

Where the comment reads "These are used to keep the scope on the containing items" and I had to comment out the Node pointer. Had we gone with inheritance, the NodeBase pointer this class would have access to would have been enough, but given the composition nature, I cannot just hold that part of the interface and maintain the same guarantee. That's ok though, I never liked how we have so much "ownership spaghetti" (for lack of a better phrase) anyways. I just have to resolve how to handle this now. I've got the options and arguments for one way versus another partially written out, I'll try to post those soon. @Karsten1987 recently ran into this in a related situation and thinks (if I can speak for him a bit) that moving away from scoped objects would help a little bit because it could ensure the lifetime of the entities created by a node do not outlive it (using a proxy or handle pattern instead).

### DIFF
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -20,7 +20,7 @@
 #include "rclcpp/callback_group.hpp"
 #include "rclcpp/client.hpp"
 #include "rclcpp/macros.hpp"
-// #include "rclcpp/node.hpp"
+#include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/service.hpp"
 #include "rclcpp/subscription.hpp"
 #include "rclcpp/timer.hpp"
@@ -49,7 +49,7 @@ struct AnyExecutable
   rclcpp::client::ClientBase::SharedPtr client;
   // These are used to keep the scope on the containing items
   rclcpp::callback_group::CallbackGroup::SharedPtr callback_group;
-  // rclcpp::node::Node::SharedPtr node;
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base;
 };
 
 }  // namespace executor
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -17,8 +17,13 @@
 
 #include <memory>
 
+#include "rclcpp/callback_group.hpp"
+#include "rclcpp/client.hpp"
 #include "rclcpp/macros.hpp"
-#include "rclcpp/node.hpp"
+// #include "rclcpp/node.hpp"
+#include "rclcpp/service.hpp"
+#include "rclcpp/subscription.hpp"
+#include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 
 namespace rclcpp
@@ -44,7 +49,7 @@ struct AnyExecutable
   rclcpp::client::ClientBase::SharedPtr client;
   // These are used to keep the scope on the containing items
   rclcpp::callback_group::CallbackGroup::SharedPtr callback_group;
-  rclcpp::node::Node::SharedPtr node;
+  // rclcpp::node::Node::SharedPtr node;
 };
 
 }  // namespace executor
--- a/rclcpp/include/rclcpp/create_publisher.hpp
+++ b/rclcpp/include/rclcpp/create_publisher.hpp
@@ -1,0 +1,54 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__CREATE_PUBLISHER_HPP_
+#define RCLCPP__CREATE_PUBLISHER_HPP_
+
+#include <memory>
+#include <string>
+
+#include "rclcpp/node_interfaces/node_topics_interface.hpp"
+#include "rclcpp/publisher_factory.hpp"
+#include "rmw/qos_profiles.h"
+
+namespace rclcpp
+{
+
+template<typename MessageT, typename AllocatorT, typename PublisherT>
+std::shared_ptr<PublisherT>
+create_publisher(
+  rclcpp::node_interfaces::NodeTopicsInterface * node_topics,
+  const std::string & topic_name,
+  const rmw_qos_profile_t & qos_profile,
+  bool use_intra_process_comms,
+  std::shared_ptr<AllocatorT> allocator)
+{
+  auto publisher_options = rcl_publisher_get_default_options();
+  publisher_options.qos = qos_profile;
+
+  auto pub = node_topics->create_publisher(
+    topic_name,
+    rclcpp::create_publisher_factory<MessageT, AllocatorT, PublisherT>(allocator),
+    publisher_options,
+    use_intra_process_comms);
+  node_topics->add_publisher(
+    pub,
+    // this is the callback group, not currently used or exposed to the user for publishers
+    nullptr);
+  return std::dynamic_pointer_cast<PublisherT>(pub);
+}
+
+}  // namespace rclcpp
+
+#endif  // RCLCPP__CREATE_PUBLISHER_HPP_
--- a/rclcpp/include/rclcpp/create_publisher.hpp
+++ b/rclcpp/include/rclcpp/create_publisher.hpp
@@ -42,10 +42,7 @@ create_publisher(
     rclcpp::create_publisher_factory<MessageT, AllocatorT, PublisherT>(allocator),
     publisher_options,
     use_intra_process_comms);
-  node_topics->add_publisher(
-    pub,
-    // this is the callback group, not currently used or exposed to the user for publishers
-    nullptr);
+  node_topics->add_publisher(pub);
   return std::dynamic_pointer_cast<PublisherT>(pub);
 }
 
--- a/rclcpp/include/rclcpp/create_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_subscription.hpp
@@ -17,6 +17,7 @@
 
 #include <memory>
 #include <string>
+#include <utility>
 
 #include "rclcpp/node_interfaces/node_topics_interface.hpp"
 #include "rclcpp/subscription_factory.hpp"
@@ -45,7 +46,7 @@ create_subscription(
 
   auto factory =
     rclcpp::create_subscription_factory<MessageT, CallbackT, AllocatorT, SubscriptionT>(
-      std::forward<CallbackT>(callback), msg_mem_strat, allocator);
+    std::forward<CallbackT>(callback), msg_mem_strat, allocator);
 
   auto sub = node_topics->create_subscription(
     topic_name,
--- a/rclcpp/include/rclcpp/create_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_subscription.hpp
@@ -1,0 +1,61 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__CREATE_SUBSCRIPTION_HPP_
+#define RCLCPP__CREATE_SUBSCRIPTION_HPP_
+
+#include <memory>
+#include <string>
+
+#include "rclcpp/node_interfaces/node_topics_interface.hpp"
+#include "rclcpp/subscription_factory.hpp"
+#include "rmw/qos_profiles.h"
+
+namespace rclcpp
+{
+
+template<typename MessageT, typename CallbackT, typename AllocatorT, typename SubscriptionT>
+typename rclcpp::subscription::Subscription<MessageT, AllocatorT>::SharedPtr
+create_subscription(
+  rclcpp::node_interfaces::NodeTopicsInterface * node_topics,
+  const std::string & topic_name,
+  CallbackT && callback,
+  const rmw_qos_profile_t & qos_profile,
+  rclcpp::callback_group::CallbackGroup::SharedPtr group,
+  bool ignore_local_publications,
+  bool use_intra_process_comms,
+  typename rclcpp::message_memory_strategy::MessageMemoryStrategy<MessageT, AllocatorT>::SharedPtr
+  msg_mem_strat,
+  typename std::shared_ptr<AllocatorT> allocator)
+{
+  auto subscription_options = rcl_subscription_get_default_options();
+  subscription_options.qos = qos_profile;
+  subscription_options.ignore_local_publications = ignore_local_publications;
+
+  auto factory =
+    rclcpp::create_subscription_factory<MessageT, CallbackT, AllocatorT, SubscriptionT>(
+      std::forward<CallbackT>(callback), msg_mem_strat, allocator);
+
+  auto sub = node_topics->create_subscription(
+    topic_name,
+    factory,
+    subscription_options,
+    use_intra_process_comms);
+  node_topics->add_subscription(sub, group);
+  return std::dynamic_pointer_cast<SubscriptionT>(sub);
+}
+
+}  // namespace rclcpp
+
+#endif  // RCLCPP__CREATE_SUBSCRIPTION_HPP_
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -17,6 +17,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <cstring>
 #include <functional>
 #include <limits>
 #include <map>
@@ -92,6 +93,8 @@ public:
   add_subscription(uint64_t id, subscription::SubscriptionBase::SharedPtr subscription)
   {
     subscriptions_[id] = subscription;
+    // subscription->get_topic_name() -> const char * can be used as the key,
+    // since subscriptions_ shares the ownership of subscription
     subscription_ids_by_topic_[subscription->get_topic_name()].insert(id);
   }
 
@@ -253,8 +256,20 @@ private:
   using SubscriptionMap = std::unordered_map<uint64_t, subscription::SubscriptionBase::WeakPtr,
       std::hash<uint64_t>, std::equal_to<uint64_t>,
       RebindAlloc<std::pair<const uint64_t, subscription::SubscriptionBase::WeakPtr>>>;
-  using IDTopicMap = std::map<std::string, AllocSet,
-      std::less<std::string>, RebindAlloc<std::pair<const std::string, AllocSet>>>;
+
+  struct strcmp_wrapper : public std::binary_function<const char *, const char *, bool>
+  {
+     bool
+     operator()(const char * lhs, const char * rhs) const
+     {
+        return std::strcmp(lhs, rhs) < 0;
+     }
+  };
+  using IDTopicMap = std::map<
+      const char *,
+      AllocSet,
+      strcmp_wrapper,
+      RebindAlloc<std::pair<const std::string, AllocSet>>>;
 
   SubscriptionMap subscriptions_;
 
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -259,11 +259,11 @@ private:
 
   struct strcmp_wrapper : public std::binary_function<const char *, const char *, bool>
   {
-     bool
-     operator()(const char * lhs, const char * rhs) const
-     {
-        return std::strcmp(lhs, rhs) < 0;
-     }
+    bool
+    operator()(const char * lhs, const char * rhs) const
+    {
+      return std::strcmp(lhs, rhs) < 0;
+    }
   };
   using IDTopicMap = std::map<
       const char *,
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -39,6 +39,7 @@
 #include "rclcpp/event.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/message_memory_strategy.hpp"
+#include "rclcpp/node_interfaces/node_base.hpp"
 #include "rclcpp/parameter.hpp"
 #include "rclcpp/publisher.hpp"
 #include "rclcpp/service.hpp"
@@ -226,7 +227,6 @@ public:
 
   using CallbackGroup = rclcpp::callback_group::CallbackGroup;
   using CallbackGroupWeakPtr = std::weak_ptr<CallbackGroup>;
-  using CallbackGroupWeakPtrList = std::list<CallbackGroupWeakPtr>;
 
   /* Create and return a Client. */
   template<typename ServiceT>
@@ -294,7 +294,7 @@ public:
   count_subscribers(const std::string & topic_name) const;
 
   RCLCPP_PUBLIC
-  const CallbackGroupWeakPtrList &
+  const std::vector<CallbackGroupWeakPtr> &
   get_callback_groups() const;
 
   RCLCPP_PUBLIC
@@ -320,7 +320,7 @@ public:
    */
   RCLCPP_PUBLIC
   std::shared_ptr<rcl_node_t>
-  get_shared_node_handle();
+  get_shared_rcl_node_handle();
 
   /// Notify threads waiting on graph changes.
   /* Affects threads waiting on the notify guard condition, see:
@@ -377,7 +377,9 @@ public:
   template<typename CallbackT>
   void register_param_change_callback(CallbackT && callback);
 
-  std::atomic_bool has_executor;
+  RCLCPP_PUBLIC
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr
+  get_node_base_interface();
 
   RCLCPP_PUBLIC
   void
@@ -391,14 +393,7 @@ private:
   bool
   group_in_node(callback_group::CallbackGroup::SharedPtr group);
 
-  std::string name_;
-
-  std::shared_ptr<rcl_node_t> node_handle_;
-
-  rclcpp::context::Context::SharedPtr context_;
-
-  CallbackGroup::SharedPtr default_callback_group_;
-  CallbackGroupWeakPtrList callback_groups_;
+  rclcpp::node_interfaces::NodeBase::SharedPtr node_base_;
 
   size_t number_of_subscriptions_;
   size_t number_of_timers_;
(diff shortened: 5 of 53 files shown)


gbiggs:
I think I could do with a diagram. :bowtie:

wjwwood:
I justed pushed the publisher and subscription parts of the refactor, `rclcpp` is building for me locally, but the examples are broken because I'll have to add some convenience functions for things like `rclcpp::spin()` which take `Node *` and extract the `NodeBaseInterface *` from it.

@gbiggs I did some design work upfront and as part of it made this:

node_interfaces

I'm not sure it will end up looking exactly like that, but it will be similar.

wjwwood:
Things to point out in the pub/sub refactor:

- Node still has code in the `create_publisher<>()` like methods, but it's mostly variable checking.
- As much of the code that depends on the Message type or the Allocator type was put into a non-templated factory class, i.e. `PublisherFactory` and `SubscriptionFactory`.
- The non-template argument dependent code is in the `NodeTopics::create_*` functions which are also not templated and take a factory class as an argument.

---

The `Node::create_*<MessageT, ...>()` methods basically do this:

- Check arguments are valid.
- Create a `*Factory` class by calling the matching `create_*_factory<MessageT, ...>()` function.
- Pass the factory class to the `NodeTopics::create_*(..., factory, ...)` method, and get a `PublisherBase` or `SubscriptionBase` shared pointer back.
- Pass the `*Base` class to the corresponding `NodeTopics::add_*()` method.

We could consider putting the code that lives in `NodeTopics::create_*()` back into the templated `Node::create_*<MessageT, ...>()` methods, but by pushing the code into the `NodeTopics` class we minimize the code that is in the `Node::create_*<MessageT, ...>()` methods and by doing so gain:

- shorter compile times
- reduced amount of code that has to be duplicated if someone want to copy-paste-modify the Node class

Similarly, we could put the code that now lives in the factory classes (mostly code that depends on template variables in one way or another) back into the `Node::create_*<MessageT, ...>()` methods, but by putting it into a separate factory class it again improves the reusability of this code and would help whenever someone might want to copy-paste-modify the Node class. It especially reduces the amount of non-trivial, or non-boilerplate, code that needs to be copied when creating a new Node class.

----

It was not possible to have template functions on either the `NodeTopicsInterface` class or the `NodeTopics` class without also templating the `Node` class, which I wanted to avoid. That is why I didn't just move the templated code from the `Node` class into the `NodeTopics` class.

wjwwood:
@gbiggs just some follow up points to go with that diagram:

We wanted to accomplish a few things:

- Make it easier to mock up parts of the Node interface to more easily test the components that use them without instantiating a full Node implementation to do so.
- More clearly separate the various functionality that is aggregated by the Node class for the convenience of the user.
- Enable the creation of various other kinds of Node classes, e.g. `LifecylceNode` or `NodeWithoutParameters`.

We landed on a few things:

- Separate the current state into these discrete parts:
  - Base: core functionality and the rcl_node_t handle.
  - Topics: publisher and subscription related functionality
  - Services: service client and server related functionality
  - Parameters: parameter related functionality
  - Graph: ROS graph related functionality, i.e. wait_for_service, list_topics, etc.
- Use pure virtual "interface" classes for each of the parts, in order to make mocking them easier without worrying about side effects of the constructor.
- Use composition, rather than inheritance, to aggregate the functionality of each of parts into the Node class, i.e. `Node` has pointers to `NodeBaseInterface`, `NodeTopicsInterface`, etc. and provides forwarding methods (similar to the PIMPL pattern) rather than inheriting from them.

We discussed several of these points a lot internally and I have notes, but they're not organized at all, but I plan to put them somewhere as memoranda when I get a chance.

rclcpp/include/rclcpp/any_executable.hpp
```diff
@@ -17,8 +17,13 @@
 
 #include <memory>
 
+#include "rclcpp/callback_group.hpp"
+#include "rclcpp/client.hpp"
 #include "rclcpp/macros.hpp"
-#include "rclcpp/node.hpp"
+// #include "rclcpp/node.hpp"
```
dirk-thomas: Can this be removed?

rclcpp/include/rclcpp/any_executable.hpp
```diff
@@ -44,7 +49,7 @@ struct AnyExecutable
   rclcpp::client::ClientBase::SharedPtr client;
   // These are used to keep the scope on the containing items
   rclcpp::callback_group::CallbackGroup::SharedPtr callback_group;
-  rclcpp::node::Node::SharedPtr node;
+  // rclcpp::node::Node::SharedPtr node;
```
dirk-thomas: Can this be removed?

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  RCLCPP_PUBLIC
+  virtual
+  const std::string &
+  get_name() const;
+
+  RCLCPP_PUBLIC
+  virtual
+  rclcpp::context::Context::SharedPtr
+  get_context() const;
```
dirk-thomas: If the method is `const` shouldn't it return a const shared pointer? Otherwise the caller can modify the returned value. Maybe we need both: one const method with const return value and one non-const method with a non-const return value?

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -52,7 +52,7 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
   using VoidAllocTraits = typename allocator::AllocRebind<void *, Alloc>;
   using VoidAlloc = typename VoidAllocTraits::allocator_type;
 
-  using WeakNodeVector = std::vector<std::weak_ptr<node::Node>>;
+  // using WeakNodeVector = std::vector<std::weak_ptr<node::Node>>;
```
dirk-thomas: Can this be removed?

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -266,7 +266,7 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
           any_exec->subscription = subscription;
         }
         any_exec->callback_group = group;
-        any_exec->node = get_node_by_group(group, weak_nodes);
+        // any_exec->node = get_node_by_group(group, weak_nodes);
```
dirk-thomas: Can this be removed?

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -300,7 +300,7 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
         // Otherwise it is safe to set and return the any_exec
         any_exec->service = service;
         any_exec->callback_group = group;
-        any_exec->node = get_node_by_group(group, weak_nodes);
+        // any_exec->node = get_node_by_group(group, weak_nodes);
```
dirk-thomas: Can this be removed?

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -333,7 +333,7 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
         // Otherwise it is safe to set and return the any_exec
         any_exec->client = client;
         any_exec->callback_group = group;
-        any_exec->node = get_node_by_group(group, weak_nodes);
+        // any_exec->node = get_node_by_group(group, weak_nodes);
```
dirk-thomas: Can this be removed?

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  RCLCPP_PUBLIC
+  virtual
+  rcl_node_t *
+  get_rcl_node_handle() const;
```
dirk-thomas: Same here.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  RCLCPP_PUBLIC
+  virtual
+  std::shared_ptr<rcl_node_t>
+  get_shared_rcl_node_handle() const;
```
dirk-thomas: Same here.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  RCLCPP_PUBLIC
+  virtual
+  rcl_guard_condition_t *
+  get_notify_guard_condition();
+
+  RCLCPP_PUBLIC
+  virtual
+  std::unique_lock<std::recursive_mutex>
+  acquire_notify_guard_condition_lock() const;
```
dirk-thomas: With the ownership being changed this shouldn't be const?

rclcpp/src/rclcpp/any_executable.cpp
```diff
@@ -22,8 +22,8 @@ AnyExecutable::AnyExecutable()
   timer(nullptr),
   service(nullptr),
   client(nullptr),
-  callback_group(nullptr),
-  node(nullptr)
+  callback_group(nullptr)
+  // node(nullptr)
```
dirk-thomas: Can this be removed?

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -0,0 +1,195 @@
...
+  size_t ros_domain_id_size;
+  _dupenv_s(&ros_domain_id, &ros_domain_id_size, env_var);
+#endif
+  if (ros_domain_id) {
+    uint32_t number = strtoul(ros_domain_id, NULL, 0);
+    if (number == (std::numeric_limits<uint32_t>::max)()) {
+      // Finalize the interrupt guard condition.
+      finalize_notify_guard_condition();
+
+      throw std::runtime_error("failed to interpret ROS_DOMAIN_ID as integral number");
```
dirk-thomas: While this code block has only been moved this line leaks the memory of `ros_domain_id` on Windows.

wjwwood:
I just pushed the graph API refactor and rebased since @Karsten1987's pr was merged.

I'll address the commented out things and other comments once it is in review.

wjwwood:
I pushed the `NodeServices` API portion of the refactor and fixed some things added by @Karsten1987's pr which broke due to my changes. Only the `NodeParameters` part is left to go and to decide where to put the "create_timer" functions, either in a `NodeTimers` API part, or in the `NodeBase` part. I'm leaning towards giving timers their own part, mostly because it doesn't seem to fit in the `NodeBase` API very well (most things that use `NodeBaseInterface` don't need timers).

Karsten1987:
+1 for giving timers their own part

gbiggs:
I wonder if creating timers, etc. should be done via factory functions that are friends of the base node interface (or node implementation? I forget which is correct for friends). I think that this should allow any object that provides the necessary friend-accessible interface to allow the creation of timers, and would also keep the size of the node object APIs minimal. One of the popular C++ book authors wrote several articles and book chapters about keeping class APIs minimal by making good use of friend and non-friend functions when possible.

I haven't analysed this idea in any way, but I thought it's worth consideration.

gbiggs:
Thanks for the diagram and explanations, @wjwwood. It's all stuff I've been hoping for, so I'm really happy to see it happening.

wjwwood:
I just pushed Parameters and Timers API's. Now I just need to do some testing and address some existing comments.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
wjwwood: I don't know. I didn't make this `const` originally (this method was just moved not modified or added) because the shared ptr that is returned is a copy. From what I understand it is not required and not appropriate to return a const of a copy, e.g. `bool get_x() const` and not `const bool get_x() const`. Instead you return a const when you return a reference, e.g. `const MyClass & get_my_class() const` rather than `MyClass & get_my_class() const`.

So if you consider the shared pointer to be a copy, then it doesn't need to be const, i.e. if someone changes the shared ptr that was returned, or calls reset of on it, it would not change the class. However, since it is a pointer, you could modify the thing to which it points and so maybe it should be a `::ConstSharedPtr` as we call it (aka `std::shared_ptr<const Context>`). Getting rid of the shared pointer and just using a raw pointer, it is currently `Context * get_context() const` and maybe it should be `Context const * get_context() const` (reads "the const method get_context returns a pointer to a const Context").

I'm not sure there is a needs to be a const and non const version. I think a non const version is fine, and if the method could be const then that would be better, but not necessary.

So the question is whether or not we should leave it as is, or if we should have it return a `SharedPtr` but not be a const method, or if we should have it return a `ConstSharedPtr`and be a const method.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
wjwwood: Same thing applies here, despite the fact that it isn't a smart pointer.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
wjwwood: The returned lock is a temporary and the mutex is mutable, so I'd say no.

wjwwood:
Ok, I've addressed or continued discussion on the existing comments. I've also started some CI which will run tonight and I'll check on tomorrow. One thing I'm thinking of doing still is to remove the publisher and subscription factories I created since they don't make as much sense anymore since I've made the templated free functions create_publisher and create_subscription. It's not strictly necessary, but would be easier to read I think, as the factory breaks the flow of the code up considerably.

I put this in review.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
dirk-thomas: From my point of view if you get a `const Node` then you are not supposed to do anything non-const on the node. But if the class has const methods which expose non-const internals you can call that method and then manipulate the internal state. Therefore const methods should always expose internal state as const.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
dirk-thomas: I did't' know it is a temporary. That makes sense then.

wjwwood:
Ok, I think I've addressed the constness issues, now there is a tlsf test failing which may be something I broke in the allocator support that I need to look into. Still looking for more reviews in the meantime.

wjwwood:
Passing CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2013/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1545/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1987/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -39,43 +39,25 @@
...
-
-class EventNotRegisteredError : public std::runtime_error
-{
-public:
-  EventNotRegisteredError()
-  : std::runtime_error("event already registered") {}
-};
-
 /// Node is the single point of entry for creating publishers and subscribers.
 class Node : public std::enable_shared_from_this<Node>
```
Karsten1987: Does the new architecture allow to create Services inside the node constructor?

rclcpp/include/rclcpp/create_publisher.hpp
```diff
@@ -0,0 +1,54 @@
...
+  const rmw_qos_profile_t & qos_profile,
+  bool use_intra_process_comms,
+  std::shared_ptr<AllocatorT> allocator)
+{
+  auto publisher_options = rcl_publisher_get_default_options();
+  publisher_options.qos = qos_profile;
+
+  auto pub = node_topics->create_publisher(
+    topic_name,
+    rclcpp::create_publisher_factory<MessageT, AllocatorT, PublisherT>(allocator),
```
wjwwood: The factory captures all template code into non template functions so they can be interleaved with non template code in the cpp file. As I mentioned before, I think we could remove the factories now in favor of more linear code that lives in the template functions. The trade off being that there is more code in template functions which increase compile time. 

I'd rather do that refactor later though so that we can merge this and let other work build on top in the mean time.

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -0,0 +1,62 @@
...
+template<typename MessageT, typename CallbackT, typename AllocatorT, typename SubscriptionT>
+typename rclcpp::subscription::Subscription<MessageT, AllocatorT>::SharedPtr
+create_subscription(
+  rclcpp::node_interfaces::NodeTopicsInterface * node_topics,
+  const std::string & topic_name,
+  CallbackT && callback,
+  const rmw_qos_profile_t & qos_profile,
+  rclcpp::callback_group::CallbackGroup::SharedPtr group,
+  bool ignore_local_publications,
+  bool use_intra_process_comms,
```
wjwwood: We need an options structure in c++ for these classes, but I didn't want to tackle that in this pr.

rclcpp/include/rclcpp/node_interfaces/node_timers.hpp
```diff
@@ -0,0 +1,60 @@
...
+namespace rclcpp
+{
+namespace node_interfaces
+{
+
+/// Implementation of the NodeTimers part of the Node API.
+class NodeTimers : public NodeTimersInterface
+{
+public:
+  RCLCPP_SMART_PTR_ALIASES_ONLY(NodeTimers)
```
wjwwood: We discussed this before, and while I agree that the thing you ran into was a bad user experience, I don't think it should be addressed in this pr and should perhaps only be changed for classes from which people commonly inherit. Please make a new issue to discuss this.

rclcpp/include/rclcpp/node.hpp
wjwwood: Yes, but I haven't tested it.

rclcpp/include/rclcpp/create_publisher.hpp
```diff
@@ -0,0 +1,54 @@
...
+    publisher_options,
+    use_intra_process_comms);
+  node_topics->add_publisher(
+    pub,
+    // this is the callback group, not currently used or exposed to the user for publishers
+    nullptr);
```
wjwwood: I removed the argument in c18b6ab, I'll run CI on Linux only (due to the nature of the change that should be enough).

rclcpp/include/rclcpp/create_subscription.hpp
wjwwood: I'm creating an issue for this right now.

rclcpp/include/rclcpp/node_interfaces/node_timers.hpp
wjwwood: Can you make an issue for this, so it is tracked?

rclcpp/include/rclcpp/node.hpp
wjwwood: I'll add a test for this real quick.

rclcpp/include/rclcpp/intra_process_manager_impl.hpp
```diff
@@ -253,8 +256,20 @@ class IntraProcessManagerImpl : public IntraProcessManagerImplBase
   using SubscriptionMap = std::unordered_map<uint64_t, subscription::SubscriptionBase::WeakPtr,
       std::hash<uint64_t>, std::equal_to<uint64_t>,
       RebindAlloc<std::pair<const uint64_t, subscription::SubscriptionBase::WeakPtr>>>;
-  using IDTopicMap = std::map<std::string, AllocSet,
-      std::less<std::string>, RebindAlloc<std::pair<const std::string, AllocSet>>>;
+
+  struct strcmp_wrapper : public std::binary_function<const char *, const char *, bool>
```
Karsten1987: can we exclude this struct into an external header file? MIght be helpful for other pieces of code.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -117,6 +99,14 @@ class Node : public std::enable_shared_from_this<Node>
   rclcpp::callback_group::CallbackGroup::SharedPtr
   create_callback_group(rclcpp::callback_group::CallbackGroupType group_type);
 
+  using CallbackGroup = rclcpp::callback_group::CallbackGroup;
```
Karsten1987: can we shift that line before the first usage?

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -73,66 +75,15 @@ Node::create_publisher(
...
+    topic_name,
+    qos_profile,
+    use_intra_process_comms_,
+    allocator);
 }
 
-template<typename MessageT, typename CallbackT, typename Alloc>
+template<typename MessageT, typename CallbackT, typename Alloc, typename SubscriptionT>
 typename rclcpp::subscription::Subscription<MessageT, Alloc>::SharedPtr
 Node::create_subscription(
```
Karsten1987: out of consistency, I would also bring that forward function into the `create_subscription.cpp` file.

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -261,30 +140,17 @@ Node::create_subscription(
     allocator);
 }
 
-template<typename CallbackType>
-typename rclcpp::timer::WallTimer<CallbackType>::SharedPtr
+template<typename DurationT, typename CallbackT>
+typename rclcpp::timer::WallTimer<CallbackT>::SharedPtr
 Node::create_wall_timer(
```
Karsten1987: again, just to be consistent, I would also put the `create_wall_timer` into its own file.

rclcpp/include/rclcpp/create_subscription.hpp
wjwwood: https://github.com/ros2/rclcpp/issues/288

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+private:
+  RCLCPP_DISABLE_COPY(NodeBase)
+
+  std::string name_;
```
Karsten1987: why do we need this? Does it not make sense to hit directly to rcl for getting the name?

rclcpp/include/rclcpp/node_interfaces/node_services.hpp
```diff
@@ -0,0 +1,67 @@
...
+  virtual
+  void
+  add_client(
+    rclcpp::client::ClientBase::SharedPtr client_base_ptr,
+    rclcpp::callback_group::CallbackGroup::SharedPtr group);
+
+  RCLCPP_PUBLIC
+  virtual
+  void
+  add_service(
```
Karsten1987: I mean why is `add_client` in `NodeServices` and not in `NodeServiceClients` ? Or, if mixed, where is the constructor for `ServiceClient` ?

rclcpp/src/rclcpp/node_interfaces/node_timers.cpp
```diff
@@ -0,0 +1,47 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+
+#include "rclcpp/node_interfaces/node_timers.hpp"
```
Karsten1987: shouldn't that include be on first?

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if (rcl_trigger_guard_condition(node_base_->get_notify_guard_condition()) != RCL_RET_OK) {
+      throw std::runtime_error(
+              std::string(
+                "Failed to notify waitset on publisher creation: ") + rmw_get_error_string());
+    }
+  }
+}
+
+rclcpp::subscription::SubscriptionBase::SharedPtr
+NodeTopics::create_subscription(
```
Karsten1987: why is this not in `create_subscription.cpp` ?

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+NodeTopics::NodeTopics(rclcpp::node_interfaces::NodeBaseInterface * node_base)
+: node_base_(node_base)
+{}
+
+NodeTopics::~NodeTopics()
+{}
+
+rclcpp::publisher::PublisherBase::SharedPtr
+NodeTopics::create_publisher(
```
Karsten1987: why is this not in `create_publisher.cpp` ?

Karsten1987:
This PR looks good to me. I had a few comments/questions about file ordering and consistency. But no show-stoppers.

rclcpp/include/rclcpp/intra_process_manager_impl.hpp
wjwwood: I don't feel it needs it's own file, it's a pretty trivial class, I would have made it a lambda if I could have. I'd prefer not to separate it, but you can push here with that change if you like.

rclcpp/include/rclcpp/node_impl.hpp
wjwwood: This is (a) a template function which cannot be put in a cpp file and (b) is a method of the Node class, so it would go in the `node.cpp` file if it could go in a cpp file.

rclcpp/include/rclcpp/node_impl.hpp
wjwwood: Same as above.

rclcpp/include/rclcpp/node_interfaces/node_services.hpp
wjwwood: Clients and Services (Service servers) are two halves of the same part: "services". It is not separate for the same reason that publisher and subscription related functions are together on the `NodeTopics` interface: because they are related. I don't see any need to break up the `NodeServices` interface into `NodeServiceClients` and `NodeServiceServers`, but if you do, please make an argument for that case.

> where is the constructor for ServiceClient?

I don't understand the question, there is no class called `ServiceClient`.

rclcpp/include/rclcpp/node.hpp
wjwwood: I just removed it and the following line, not sure why I added that there. Might have been a copy-paste issue, thanks for pointing it out. 28a2cfc

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
wjwwood: No, I updated it in 83b610e. Thanks!

rclcpp/src/rclcpp/node_interfaces/node_timers.cpp
wjwwood: I tried that, but I think cpplint was not happy about it for some reason. I'll try again.

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
wjwwood: Because it is a member of `NodeTopics`?

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
wjwwood: Same.

rclcpp/src/rclcpp/node_interfaces/node_timers.cpp
wjwwood: I fixed it here and in other places in 23a7728.

rclcpp/include/rclcpp/node.hpp
wjwwood: I made a test in https://github.com/ros2/system_tests/pull/182 and fixed it in 	f0afcab 😛, good thing you got me to test it.

wjwwood:
Ok, @Karsten1987 I think I addressed your comments, can you check to make sure I did? Also, anyone else that wants to review, please do so now.

I'm starting new CI since I made some significant changes.

wjwwood:
The warnings are being addressed by @codebot and I cannot reproduce the OS X test failures on my machine, and I am debugging the Windows test timeouts and I will fix that up after merging this if this pr turns out to have caused it.

Issue #273, opened 2016-11-16
no std::bind with custom arguments for callbacks


Karsten1987:
I want to pass custom arguments to my callback function and use std::bind for it. However, I believe that function_traits are not allowing passing a functor with custom arguments.

This little demo shows the block which does not compile. 
https://github.com/ros2/examples/blob/callback_error/rclcpp_examples/src/services/add_two_ints_server.cpp#L58-L68

Karsten1987:
@esteve can you provide some help here? I guess that https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/function_traits.hpp doesn't allow additional arguments?

esteve:
@Karsten1987 sure, I just submitted #278 which should work with your example, it's just a quick hack. There's one caveat, though, you can't wrap the `std::bind` in a `std::function` (https://github.com/ros2/examples/blob/callback_error/rclcpp_examples/src/services/add_two_ints_server.cpp#L61), you have to pass the `std::bind` to `create_service` as is, like this:

```c++
auto fcn2 = std::bind(handle_add_two_ints2, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, i);
node->create_service<example_interfaces::srv::AddTwoInts>("add_two_ints", fcn2);
```

I haven't tested the pull request heavily, so I can't guarantee that it will work with all cases, but it should work with yours :-)

esteve:
@Karsten1987 I think this would work, without changing anything in `rclcpp`:

```cpp
std::function<void(const std::shared_ptr<rmw_request_id_t>,
  const std::shared_ptr<example_interfaces::srv::AddTwoInts::Request>,
  std::shared_ptr<example_interfaces::srv::AddTwoInts::Response>)> fcn2;
fcn2 = std::bind(handle_add_two_ints2, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, i);
node->create_service<example_interfaces::srv::AddTwoInts>("add_two_ints", fcn2);
```

note that I removed the fourth argument in the `std::function` definition. `std::bind` can be wrapped in a `std::function` even though their parameters don't match, as long as when you call it the call signature is compatible. When the wrapped `std::bind` gets called, it'll be compatible since it only has three placeholders, and `i` is already known at the time the `std::bind` was constructed.

Karsten1987:
@esteve thanks a bunch for iterating on this. Your solution(s) definitely do the trick.

[...]

PR #278, opened 2016-11-17
Added support for std::bind callbacks that have custom arguments

esteve:
This is a quick hack to allow `std::bind` functors that have custom arguments to be passed as callbacks. These changes involve relaxing the rules of `function_traits::same_arguments` to only check if the first N arguments of the `std::bind` functor match the callback signatures, so this has the problem that the `std::bind` might be invalid (e.g. the N+1 argument is a placeholder), the compiler won't complain, but there will be a runtime error.

I've also made a couple of cleanups and implemented two classes for concatenating and slicing `std::tuple`s at compile time.

PR is marked as open, so feel free to edit it to adapt it to your needs.

Addresses #273
Connects to #273

### DIFF
--- a/rclcpp/include/rclcpp/function_traits.hpp
+++ b/rclcpp/include/rclcpp/function_traits.hpp
@@ -15,6 +15,32 @@
 #ifndef RCLCPP__FUNCTION_TRAITS_HPP_
 #define RCLCPP__FUNCTION_TRAITS_HPP_
 
+#if defined _LIBCPP_VERSION  // libc++ (Clang)
+#define RCLCPP__FUNCTION_TRAITS_BIND_OBJECT_MEMBER_SIGNATURE \
+  std::__1::__bind<ReturnTypeT (ClassT::*)(Args ...), FArgs ...>
+#elif defined __GLIBCXX__  // glibc++ (GNU C++)
+#define RCLCPP__FUNCTION_TRAITS_BIND_OBJECT_MEMBER_SIGNATURE \
+  std::_Bind<std::_Mem_fn<ReturnTypeT (ClassT::*)(Args ...)>(FArgs ...)>
+#elif defined _MSC_VER  // MS Visual Studio
+#define RCLCPP__FUNCTION_TRAITS_BIND_OBJECT_MEMBER_SIGNATURE \
+  std::_Binder<std::_Unforced, ReturnTypeT(__cdecl ClassT::*)(Args ...), FArgs ...>
+#else
+#error "Unsupported C++ compiler / standard library"
+#endif
+
+#if defined _LIBCPP_VERSION  // libc++ (Clang)
+#define RCLCPP__FUNCTION_TRAITS_BIND_FREE_FUNCTION_SIGNATURE \
+  std::__1::__bind<ReturnTypeT( &)(Args ...), FArgs ...>
+#elif defined __GLIBCXX__  // glibc++ (GNU C++)
+#define RCLCPP__FUNCTION_TRAITS_BIND_FREE_FUNCTION_SIGNATURE \
+  std::_Bind<ReturnTypeT(*(FArgs ...))(Args ...)>
+#elif defined _MSC_VER  // MS Visual Studio
+#define RCLCPP__FUNCTION_TRAITS_BIND_FREE_FUNCTION_SIGNATURE \
+  std::_Binder<std::_Unforced, ReturnTypeT(__cdecl &)(Args ...), FArgs ...>
+#else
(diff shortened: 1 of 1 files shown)


wjwwood:
Wow, thanks @esteve! It's too bad this stuff isn't in the stdlib, it's pretty hairy stuff.

esteve:
@wjwwood any time! It's unfortunate that we had to implement all this just because of Visual Studio, I wish none of this were necessary :-)

dirk-thomas:
Since these traits have originally been added we have updated the minimum Visual Studio version which I think has a more capable compiler by now. Maybe we should try to go back to less custom traits which might now be supported.

wjwwood:
We can try, but I don't think our version has changed much. There is a vs2017 rc coming up though.

esteve:
According to [this](https://blogs.msdn.microsoft.com/vcblog/2016/06/07/expression-sfinae-improvements-in-vs-2015-update-3/), there were a bunch of improvements in VS2015 Update 3 which might solve the `std::function` ambiguity problem:

> Improvements since Visual Studio 2015 Update 2
> 
> Continued improvements in the quality of our expression SFINAE implementation enabled our Standard Template Library to begin using it in VS 2015 Update 2. Expression SFINAE is used in our implementations of std::function and result_of.

So I created the following example, and it compiles on VS2015 Update 3:

```cpp
#include <functional>

void f(std::function<void(int, int)> arg) {}
void f(std::function<void(int, std::string)> arg) {}

int main(void) {
  auto f1 = [](int, int) -> void {};
  auto f2 = [](int, std::string) -> void {};

  f(f1);
  f(f2);
  return 0;
}
```

the example might be a bit too synthetic, so there's a chance it doesn't reflect the issue here, but it's worth trying removing `function_traits.hpp` and require VS2015 Update 3.

dirk-thomas:
I have updated our Windows slaves so that all of them are using Visual Studio 2015 Update 3. So if anyone would provide a PR we can easily test it.

wjwwood:
@Karsten1987 what do you want to do with this one? Do you need it merged for your pr's to work? Or should we wait until someone has time to see about removing function traits all together?

Karsten1987:
@esteve 's workaround in https://github.com/ros2/rclcpp/issues/273 works for me as for now. The caveats of not using `auto` but explicitly declaring the `std::function` object does the job.

That being said, I am fine with putting this PR on hold and test the VS update at some point.

nkoenig:
This is blocked on seeing if this is fixed on the new VS compiler.

esteve:
@nkoenig unfortunately it's slightly more complex than originally thought. Although VS now properly compiles the snippet I posted, the following does not work on either Clang, GCC or VC++:

```c++
#include <functional>
#include <memory>

struct Message {};

void f(std::function<void(std::shared_ptr<Message>)> arg) {}
void f(std::function<void(std::unique_ptr<Message>)> arg) {}

int main(void) {
  auto f1 = [](std::shared_ptr<Message>) -> void {};
  auto f2 = [](std::unique_ptr<Message>) -> void {};

  f(f1);
  f(f2);
  return 0;
}
```

This produces an ambiguity error with both three compilers:

GCC:
http://rextester.com/COJ12179

Clang:
http://rextester.com/BWISR19029

VC++:
http://rextester.com/RKDX36843

We didn't see this issue when we first wrote the callback code because intraprocess communication used a different API, so there was no ambiguity between `std::unique_ptr` and `std::shared_ptr` (see https://github.com/ros2/rclcpp/commit/b0a2b7610d5b5a2955058774bfd8d36787634e7e). Once `function_traits` was able to deduce more information about a callback, @wjwwood and I unified the `create_subscription` and `create_subscription_with_unique_ptr_callback` methods to provide a cleaner API.

I don't know if there's a better way of achieving this without retorting to the `function_traits` voodoo, I can't seem to force `std::function` not to implicitly convert its arguments to avoid the ambiguity. The same ambiguity issue shows up if you have two functions, one that takes a `std::function<void(int)>` and another one that takes a `std::function<void(long)>`, for example.

mikaelarguedas:
try to review it soon, along https://github.com/ros2/rclcpp/pull/286

esteve:
Let me know if there's anything I can help with to get this and #286
merged. I tried to make the changes as clean as I could, but unfortunately
the code is not very readable anyway :-(

allenh1:
connect to #273

mikaelarguedas:
@allenh1 Tip: waffle parses only the first comment of each issue. So the 'connect to' needs to be added to the first comment

Karsten1987:
I just closed #273. Doing the same here.

PR #279, opened 2016-11-18
(dev) construct service from existing rcl_service_t

Karsten1987:
this PR enables a wrapping around a rcl_service_t which is externally initialized. The goal here is to instantiate a service on the rcl level and still be able to add it into a rclcpp executor. 

On the flipside, I have to have a boolean flag stating whether we own ownership over this service or not. This becomes important when trying to destroy the service.

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -127,6 +127,31 @@ if(BUILD_TESTING)
     )
     target_link_libraries(test_find_weak_nodes ${PROJECT_NAME})
   endif()
+
+  get_default_rmw_implementation(default_rmw)
+  find_package(${default_rmw})
+  get_rmw_typesupport(typesupport_impls_cpp "${default_rmw}" LANGUAGE "cpp")
+  get_rmw_typesupport(typesupport_impls_c "${default_rmw}" LANGUAGE "c")
+  set(mock_msg_files
+    "test/mock_msgs/srv/Mock.srv")
+  rosidl_generate_interfaces(mock_msgs
+    ${mock_msg_files})
+
+  ament_add_gtest(test_externally_defined_services test/test_externally_defined_services.cpp)
+  if(TARGET test_externally_defined_services)
+    target_include_directories(test_externally_defined_services PUBLIC
+      ${rcl_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_externally_defined_services ${PROJECT_NAME})
+    foreach(typesupport_impl_cpp ${typesupport_impls_cpp})
+      rosidl_target_interfaces(test_externally_defined_services
+        mock_msgs ${typesupport_impl_cpp})
+    endforeach()
+    foreach(typesupport_impl_c ${typesupport_impls_c})
+      rosidl_target_interfaces(test_externally_defined_services
+        mock_msgs ${typesupport_impl_c})
+    endforeach()
+  endif()
 endif()
 
 ament_package(
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -10,10 +10,8 @@ find_package(rmw_implementation REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
-if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-std=c++11")
-endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-std=c++11")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -10,8 +10,11 @@ find_package(rmw_implementation REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-std=c++11")
+endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-std=c++11 -Wall -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
 include_directories(include)
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -129,7 +129,7 @@ if(BUILD_TESTING)
   endif()
 
   get_default_rmw_implementation(default_rmw)
-  find_package(${default_rmw})
+  find_package(${default_rmw} REQUIRED)
   get_rmw_typesupport(typesupport_impls_cpp "${default_rmw}" LANGUAGE "cpp")
   get_rmw_typesupport(typesupport_impls_c "${default_rmw}" LANGUAGE "c")
   set(mock_msg_files
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -135,7 +135,8 @@ if(BUILD_TESTING)
   set(mock_msg_files
     "test/mock_msgs/srv/Mock.srv")
   rosidl_generate_interfaces(mock_msgs
-    ${mock_msg_files})
+    ${mock_msg_files}
+    SKIP_INSTALL)
 
   ament_add_gtest(test_externally_defined_services test/test_externally_defined_services.cpp)
   if(TARGET test_externally_defined_services)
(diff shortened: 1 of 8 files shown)


Karsten1987:
@dirk-thomas I added some dummy mock messages in order to create a type-support. Can you please verify the `CMakeLists.txt` that these are only used for the test and not being installed?

dirk-thomas:
If you don't want the generated messages to be installed you need to pass the `SKIP_INSTALL` flag (https://github.com/ros2/rosidl/blob/c9545c9d3ea750e288fda40d9cb458e8798a6106/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L34-L35).

Karsten1987:
http://ci.ros2.org/job/ci_linux/1947/badge/icon

Karsten1987:
[Build Status](http://ci.ros2.org/job/ci_linux/1948/)

Karsten1987:
@wjwwood any feedback about this?

rclcpp/include/rclcpp/service.hpp
```diff
@@ -66,10 +67,9 @@ class ServiceBase
 protected:
   RCLCPP_DISABLE_COPY(ServiceBase);
 
-  std::shared_ptr<rcl_node_t> node_handle_;
+  rcl_node_t * node_handle_;
```
wjwwood: Why did you change this to a pointer from a shared pointer?

rclcpp/src/rclcpp/service.cpp
```diff
@@ -39,11 +37,11 @@ ServiceBase::~ServiceBase()
 std::string
 ServiceBase::get_service_name()
 {
-  return this->service_name_;
+  return rcl_service_get_service_name(service_handle_);
```
wjwwood: I understand what you're doing here, but it would be better I think to leave it as is and fix all cases of this (subscription, client, etc.) at the same time. This change is not required to fulfill the main purpose of this pr I think.

wjwwood: Alternatively, update the other locations where this is happening in this pr. Basically, I don't want an inconsistent state.

rclcpp/test/test_externally_defined_services.cpp
```diff
@@ -0,0 +1,133 @@
...
+    rclcpp, srv, Mock);
+  if (rcl_service_init(&service_handle, node_handle->get_rcl_node_handle(),
+    ts, "base_node_service", &service_options) != RCL_RET_OK)
+  {
+    FAIL();
+    return;
+  }
+  rclcpp::any_service_callback::AnyServiceCallback<rclcpp::srv::Mock> cb;
+  // don't initialize the service
+  // expect fail
```
wjwwood: expect fail? The catch fails here.

rclcpp/include/rclcpp/service.hpp
```diff
@@ -95,30 +95,53 @@ class Service : public ServiceBase
...
 
+  Service(
+    rcl_node_t * node_handle,
+    rcl_service_t * service_handle,
+    AnyServiceCallback<ServiceT> any_callback)
+  : ServiceBase(node_handle), defined_extern(true),
+    any_callback_(any_callback)
+  {
+    // check if service handle was initialized
+    // TODO(Karsten1987): Can this go directly in RCL?
```
wjwwood: https://github.com/ros2/rcl/issues/81

rclcpp/include/rclcpp/service.hpp
```diff
@@ -160,6 +183,8 @@ class Service : public ServiceBase
 private:
   RCLCPP_DISABLE_COPY(Service);
 
+  bool defined_extern = false;
```
wjwwood: Consider a different name, or an unabbreviated name, e.g. `owns_rcl_handle` or `external_rcl_handle`.

rclcpp/include/rclcpp/service.hpp
Karsten1987: The second constructor takes raw `rcl` pointer. I am feeling a bit concerned about turning a raw pointer in a shared pointer since we'd miss 1 ref count. 
But generally, do we need this pointer as a shared component or would a `unique_ptr` help here as the class anyways is marked as non-copyable.

rclcpp/test/test_externally_defined_services.cpp
Karsten1987: that must be an erroneous copy-paste

rclcpp/include/rclcpp/node.hpp
```diff
@@ -379,6 +379,11 @@ class Node : public std::enable_shared_from_this<Node>
 
   std::atomic_bool has_executor;
 
+  RCLCPP_PUBLIC
+  void
+  add_service(const rclcpp::service::ServiceBase::SharedPtr & service,
```
wjwwood: If you actually read that post, it points out that copying shared pointers is more expensive compared to copying normal pointers or references, and that using a const reference only becomes a problem if you store the references in a vector or pass it to a function in another thread. Neither of which are happening here. I'd say that const reference is fine here, but you might still choose to not use const reference for other reasons (like the fact that it is a public interface which someone could misuse).

> In controlled circumstances you can pass the shared pointer by constant reference. Be sure that nobody is concurrently deleting the object, though this shouldn't be too hard if you're careful about to whom you give references.

I'd call this a controlled circumstance.

In this past we've solved issues before by removing the pass by reference of shared pointers, but in my opinion that's just using shared pointers to make the functions thread safe and generally doesn't have anything to do with shared pointers. We'd have the same problem if we were using raw pointers, i.e. someone calls this function in one thread and destroys the object (`Node`) in another thread. This is a general thread-safety issue, not a problem with how we're using shared pointers with const reference passing.

rclcpp/include/rclcpp/service.hpp
wjwwood: Why can the second constructor not take a shared pointer?

Karsten1987:
updated all comments.

firesurfer:
Is this functionality already possible for publisher and subscriptions or are there any special needs for services that make this necessary?

wjwwood:
@firesurfer there is a particular case (@Karsten1987's life cycle work) where we need this ability for Service's in particular. We plan to add support for this to the other classes for symmetry in the future, but we're not doing it here to avoid collisions with other ongoing prs, like my https://github.com/ros2/rclcpp/pull/277.

wjwwood:
Speaking of which, I would really like to get this one merged even if the other life cycle stuff isn't quite ready, so that I can build my pr on top of it.

rclcpp/include/rclcpp/service.hpp
```diff
@@ -100,25 +105,50 @@ class Service : public ServiceBase
...
 
+  Service(
+    std::shared_ptr<rcl_node_t> node_handle,
+    rcl_service_t * service_handle,
+    AnyServiceCallback<ServiceT> any_callback)
+  : ServiceBase(node_handle),
+    any_callback_(any_callback)
+  {
+    // check if service handle was initialized
+    // see: https://github.com/ros2/rcl/issues/81
```
wjwwood: I'd leave the todo, or link here from the that issue. Otherwise we'll never know to come back and fix this.

wjwwood:
Other than a small comment, I'd say merge this as soon as CI is passing for it.

rclcpp/CMakeLists.txt
```diff
@@ -127,6 +127,32 @@ if(BUILD_TESTING)
     )
     target_link_libraries(test_find_weak_nodes ${PROJECT_NAME})
   endif()
+
+  get_default_rmw_implementation(default_rmw)
+  find_package(${default_rmw} REQUIRED)
```
mikaelarguedas: is it a requirement that this test `find_package`s and uses another rmw than all the other tests of this package?

rclcpp/include/rclcpp/service.hpp
```diff
@@ -100,25 +105,50 @@ class Service : public ServiceBase
...
+    if (service_handle->impl == NULL) {
```
mikaelarguedas: we usually use `NULL == service_handle->impl` to avoid a missing `=` to cause the condition to pass everytime. `if (!service_handle->impl)` would also work

rclcpp/include/rclcpp/service.hpp
```diff
@@ -100,25 +105,50 @@ class Service : public ServiceBase
...
+      throw std::runtime_error(
+              std::string("rcl_service_t in constructor argument ") +
+              "has to be initialized beforehand");
```
mikaelarguedas: sorry for yet another OCD symptom: This looks a bit weird to me, is that for line length purposes that only half the string is passed to `std::string()`?

wjwwood: This must be a copy-paste issue, normally that pattern is used like this:

```c++
throw std::runtime_error(
  std::string("there was an error while trying to do something: ") +
  rcl_get_error_string_safe());
```

Where the first item needs to be a `std::string` so that the `+` operator works with the `const char *` returned by the error function.

mikaelarguedas:
looks good to me too, all my comments are cosmetic and are not blockers for merging this

wjwwood:
@Karsten1987 can you run CI for this or link some up-to-date successful runs if you already have?

Karsten1987:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1957/)

Karsten1987:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1957/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1917/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1510/)

wjwwood:
Awesome, I'm going to merge this so I can keep hacking on my pr.

Karsten1987:
I'd say this PR is ready to be merged.

PR #283, opened 2016-11-30
Parse command-line params

gerkey:
This is a work in progress from a while back.

### DIFF
--- a/rclcpp/src/rclcpp/parameter_service.cpp
+++ b/rclcpp/src/rclcpp/parameter_service.cpp
@@ -36,28 +36,28 @@ ParameterService::ParameterService(const rclcpp::node::Node::SharedPtr node)
       break;
     }
     int len = strlen(arg);
-    // Minimum expression is __n:=v
-    if (len > 5) {
-      if (arg[0] == '_' && arg[1] == '_') {
+    // Minimum expression is _n:=v
+    if (len > 4) {
+      if (arg[0] == '_') {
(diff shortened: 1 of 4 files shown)


rclcpp/src/rclcpp/parameter_service.cpp
```diff
@@ -17,11 +17,78 @@
...
+  char** argv;
+  argc = rcl_get_argv(&argv);
+  for (int i = 1; i < argc; ++i)
+  {
+    char * arg = strdup(argv[i]);
+    if (arg == NULL) {
+      break;
+    }
+    int len = strlen(arg);
+    // Minimum expression is _n:=v
```
sloretz: since [rcl_init()](http://docs.ros2.org/beta1/api/rcl/rcl_8h.html#a6abc5188c50b3a4a5fc41b357b49d6e5) requires the command line arguments to be passed to it, what do you think about parsing remapping rules in `rcl`? I think it would allow the same code to be reused across client libraries.

mikaelarguedas: sounds good to me. I think the general guideline is that all the logic that is neither rmw specific nor language specific should land in `rcl`. For now there is still a lot in the `rclcpp` layer because `rclpy` was/is lagging far behind and it was faster to prototype in c++ than in c

mikaelarguedas:
@gerkey can we close this ? (in favor of the c implementation in c_utilities?)

PR #293, opened 2016-12-15
Doxygen setup

wjwwood:
This pull request sets up Doxygen and fixes the documentation to look correct when rendered and to reference external data, like cppreference.com (where appropriate), rcl, and rmw.

I've pushed the current state up to [http://docs.ros2.org/beta1/api/rclcpp/index.html](http://docs.ros2.org/beta1/api/rclcpp/index.html).

This still needs a main page.

### DIFF
--- a/rclcpp/.gitignore
+++ b/rclcpp/.gitignore
@@ -1,0 +1,1 @@
+doc_output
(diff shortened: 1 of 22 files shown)


wjwwood:
Now that the beta is out, I'd like to get these basic configurations and documentation updates merged. Looking for a review.

PR #297, opened 2016-12-22
throw on second call of register_param_change_callback

rohbotics:
We should throw instead of silently overwriting the old callback

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -226,5 +226,9 @@ NodeParameters::list_parameters(const std::vector<std::string> & prefixes, uint6
 void
 NodeParameters::register_param_change_callback(ParametersCallbackFunction callback)
 {
+  if (parameters_callback_) {
+    fprintf(stderr, "Warning: param_change_callback already registered, "
+      "overwriting previous callback\n");
+  }
   parameters_callback_ = callback;
 }


rohbotics:
Linux CI: [Build Status](http://ci.ros2.org/job/ci_linux/2114/)

dirk-thomas:
If a user wants to overwrite the current callback this prevents him from doing so. Therefore I think a warning message would be more appropriate.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -226,9 +226,9 @@ NodeParameters::list_parameters(const std::vector<std::string> & prefixes, uint6
 void
 NodeParameters::register_param_change_callback(ParametersCallbackFunction callback)
 {
-  if (!parameters_callback_) {
-    parameters_callback_ = callback;
-  } else {
-    throw std::logic_error("param_change_callback already registered, can only register one");
+  if (parameters_callback_) {
+    fprintf(stderr, "Warning: param_change_callback already registered," 
+      "overwriting previous callback\n");
```
dirk-thomas: There should be a space between the `,` and the next character.

dirk-thomas:
Please also ensure to squash the commits.

rohbotics:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2116/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2123/)

rohbotics:
Fixed the whitespace issues that uncrustify reported.

PR #303, opened 2017-01-18
fix rcl_take ret comparison

Karsten1987:
I think there is a typo with the comparison of the return value from rcl_take

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -266,7 +266,7 @@ Executor::execute_subscription(
   if (ret == RCL_RET_OK) {
     message_info.from_intra_process = false;
     subscription->handle_message(message, message_info);
-  } else if (ret != RCL_RET_SUBSCRIPTION_TAKE_FAILED) {
+  } else if (ret == RCL_RET_SUBSCRIPTION_TAKE_FAILED) {
     fprintf(stderr,
       "[rclcpp::error] take failed for subscription on topic '%s': %s\n",
       subscription->get_topic_name(), rcl_get_error_string_safe());


rclcpp/src/rclcpp/executor.cpp
```diff
@@ -266,7 +266,7 @@ Executor::execute_subscription(
   if (ret == RCL_RET_OK) {
     message_info.from_intra_process = false;
     subscription->handle_message(message, message_info);
-  } else if (ret != RCL_RET_SUBSCRIPTION_TAKE_FAILED) {
+  } else if (ret == RCL_RET_SUBSCRIPTION_TAKE_FAILED) {
```
dirk-thomas: I think the previous comparison is correct and shouldn't be changed.

`rcl_take` returns `RCL_RET_SUBSCRIPTION_TAKE_FAILED` if there was no error but no message to take (see https://github.com/ros2/rcl/blob/88db558aa303ad29f24cc47da28f27f64a27eaa4/rcl/src/rcl/subscription.c#L146). In that case there should be no error message. Only in all other cases there should be a message printed to `stderr`. The constant would probably be more appropriately names `RCL_RET_SUBSCRIPTION_NOTHING_TAKEN` (or similar).

Karsten1987: Hmm, ok. I am confused by the comparison though.
It says "if there is no subscription take failure, then print error 'take failed ..'

dirk-thomas: Yes, in the case of the error message the return code is neither `RCL_RET_OK` nor `RCL_RET_SUBSCRIPTION_TAKE_FAILED` (aka `RCL_RET_SUBSCRIPTION_NOTHING_TAKEN`). Therefore something else which is a real problem hence the message.

Karsten1987:
gonna close this PR due to misunderstanding from my part.

PR #304, opened 2017-01-26
Rosnode list

Karsten1987:
Connect to ros2/cli_tools#1

### DIFF
--- a/rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
@@ -61,6 +61,11 @@ public:
 
   RCLCPP_PUBLIC
   virtual
+  std::vector<std::string>
+  get_node_names() const;
+
+  RCLCPP_PUBLIC
+  virtual
   size_t
   count_publishers(const std::string & topic_name) const;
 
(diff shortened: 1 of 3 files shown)


rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -75,6 +76,34 @@ NodeGraph::get_topic_names_and_types() const
...
+NodeGraph::get_node_names() const
+{
+  rcl_node_names_t node_names_c =
+    rcl_get_zero_initialized_node_names();
+
+  auto ret = rcl_get_node_names(node_base_->get_rcl_node_handle(),
+      &node_names_c);
+  if (ret != RMW_RET_OK) {
+    // *INDENT-OFF*
+    throw std::runtime_error(
```
dirk-thomas: This leaks the allocated `node_names_c`.

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
```diff
@@ -61,6 +61,11 @@ class NodeGraph : public NodeGraphInterface
 
   RCLCPP_PUBLIC
   virtual
+  std::vector<std::string>
```
dirk-thomas: I am not sure if the return type should be a vector. Since we don't need indexed access maybe a list would be more appropriate? A set would also be an option but since we are not clear if the names need to be unique that is probably too restrictive.

Karsten1987: Not sure about this. I guess in terms of performance, a vector or list don't differ that much and the underlying C-type is a `char **`. Depending on the application, we may actually want to have indexed access (such as exposing all topics attached to a individual rosnode).

dirk-thomas: We can keep it as-is for now. We can always revise it if necessary in the future.

PR #309, opened 2017-03-02
add a new Node::get_parameter() with a default value

wjwwood:
I ran into needing something like this when converting `map_server`. And I think it's useful, but I'm not sure about a few details.

First, If the parameter is not set, so it be set to the default value or not? Put another way:

```c++
int foo;
ASSERT_FALSE(node->get_parameter("foo", foo, 42));  // this is the new signature
bool result = node->get_parameter("foo", foo);  // Should result be true or false?
```

---

This new API is meant to replicate ROS 1's API (http://docs.ros.org/kinetic/api/roscpp/html/classros_1_1NodeHandle.html#a05941572641a453fd416c6603b91d39f):

```c++
// ROS 1
std::string frame_id;
node_handle.param("frame_id", frame_id, std::string("map"));
```

This is what it would look like with this pr:

```c++
// ROS 2
std::string frame_id;
node->get_parameter("frame_id", frame_id, std::string("map"));
```

---

Separately, we could choose to support the other API provided by ROS 1 which is similar but it returns `T` (http://docs.ros.org/kinetic/api/roscpp/html/classros_1_1NodeHandle.html#aa9b23d4206216ed13b5833fb1a090f1a):

```c++
// ROS 1 alternative
std::string frame_id = node_handle.param("frame_id", std::string("map"));
```

We could do this, but it begins to get a little confusing when you have one signature that takes `T &, const T &` and another that takes just `const T &`. The `const` should prevent you from getting into strange situations, but it might be the case that having fewer options would be better.

---

Finally, the Google style guide says you should pass `T *` rather than `T &` when you intend to modify the variable because it is easier to see what's going on from the calling code. In this case, I think that might be worth following. I consider this snippet from the `map_server` to be clearer with the rule applied:

```c++
// ROS 1 with pass by reference
private_nh.param("negate", negate, 0);
private_nh.param("occupied_thresh", occ_th, 0.65);
private_nh.param("free_thresh", free_th, 0.196);
```

Versus:

```c++
// ROS 1 with pass by pointer
private_nh.param("negate", &negate, 0);
private_nh.param("occupied_thresh", &occ_th, 0.65);
private_nh.param("free_thresh", &free_th, 0.196);
```

Novice users will probably lament that they have to add a `&` and they don't know why, but I think that it is easier to infer what is happening here if you know C++ well enough.

---

Anyways, I'd appreciate anyone who's interested in giving some feedback on this.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -253,6 +253,13 @@ public:
   bool
   get_parameter(const std::string & name, ParameterT & parameter) const;
 
+  template<typename ParameterT>
+  bool
+  get_parameter(
+    const std::string & name,
+    ParameterT & parameter,
+    const ParameterT & default_value) const;
+
   RCLCPP_PUBLIC
   std::vector<rcl_interfaces::msg::ParameterDescriptor>
   describe_parameters(const std::vector<std::string> & names) const;
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -235,6 +235,12 @@ public:
   rcl_interfaces::msg::SetParametersResult
   set_parameters_atomically(const std::vector<rclcpp::parameter::ParameterVariant> & parameters);
 
+  template<typename ParameterT>
+  void
+  set_parameter_if_not_set(
+    const std::string & name,
+    const ParameterT & default_value);
+
   RCLCPP_PUBLIC
   std::vector<rclcpp::parameter::ParameterVariant>
   get_parameters(const std::vector<std::string> & names) const;
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -239,7 +239,7 @@ public:
   void
   set_parameter_if_not_set(
     const std::string & name,
-    const ParameterT & default_value);
+    const ParameterT & value);
 
   RCLCPP_PUBLIC
   std::vector<rclcpp::parameter::ParameterVariant>
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -255,10 +255,10 @@ public:
 
   template<typename ParameterT>
   bool
-  get_parameter(
+  get_parameter_or(
     const std::string & name,
     ParameterT & value,
-    const ParameterT & default_value) const;
+    const ParameterT & alternative_value) const;
 
   RCLCPP_PUBLIC
   std::vector<rcl_interfaces::msg::ParameterDescriptor>
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -257,7 +257,7 @@ public:
   bool
   get_parameter(
     const std::string & name,
-    ParameterT & parameter,
+    ParameterT & value,
     const ParameterT & default_value) const;
 
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -255,10 +255,29 @@ public:
     const std::string & name,
     rclcpp::parameter::ParameterVariant & parameter) const;
 
+  /// Assign the value of the parameter if set into the parameter argument.
+  /**
+   * If the parameter was not set, then the "parameter" argument is never assigned a value.
+   *
+   * \param[in] name The name of the parameter to get.
+   * \param[out] parameter The output where the value of the parameter should be assigned.
+   * \returns true if the parameter was set, false otherwise
+   */
   template<typename ParameterT>
   bool
   get_parameter(const std::string & name, ParameterT & parameter) const;
 
+  /// Get the parameter value, or the "alternative value" if not set, and assign it to "value".
+  /**
+   * If the parameter was not set, then the "value" argument is assigned
+   * the "alternative_value".
+   * In all cases, the parameter remains not set after this function is called.
+   *
+   * \param[in] name The name of the parameter to get.
+   * \param[out] value The output where the value of the parameter should be assigned.
+   * \param[in] alternative_value Value to be stored in output if the parameter was not set.
+   * \returns true if the parameter was set, false otherwise
+   */
   template<typename ParameterT>
   bool
   get_parameter_or(
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -222,12 +222,12 @@ template<typename ParameterT>
 bool
 Node::get_parameter(
   const std::string & name,
-  ParameterT & parameter,
+  ParameterT & value,
   const ParameterT & default_value) const
 {
-  bool got_parameter = get_parameter(name, parameter);
+  bool got_parameter = get_parameter(name, value);
   if (!got_parameter) {
-    parameter = default_value;
+    value = default_value;
   }
   return got_parameter;
 }
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -220,14 +220,14 @@ Node::get_parameter(const std::string & name, ParameterT & parameter) const
 
 template<typename ParameterT>
 bool
-Node::get_parameter(
+Node::get_parameter_or(
   const std::string & name,
   ParameterT & value,
-  const ParameterT & default_value) const
+  const ParameterT & alternative_value) const
 {
   bool got_parameter = get_parameter(name, value);
   if (!got_parameter) {
-    value = default_value;
+    value = alternative_value;
   }
   return got_parameter;
 }
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -211,12 +211,12 @@ template<typename ParameterT>
 void
 Node::set_parameter_if_not_set(
   const std::string & name,
-  const ParameterT & default_value)
+  const ParameterT & value)
 {
   rclcpp::parameter::ParameterVariant parameter_variant;
   if (!this->get_parameter(name, parameter_variant)) {
     this->set_parameters({
-          rclcpp::parameter::ParameterVariant(name, default_value),
+          rclcpp::parameter::ParameterVariant(name, value),
         });
   }
 }
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -218,6 +218,20 @@ Node::get_parameter(const std::string & name, ParameterT & parameter) const
   return result;
 }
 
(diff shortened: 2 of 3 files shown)


wjwwood:
Looking at the ROS 1 code, it does not set the "set the parameter" with the default value if the parameter is not set:

http://docs.ros.org/jade/api/roscpp/html/node__handle_8h_source.html#l02039

wjwwood:
Note to self, if this is merged, revert/update this commit:

https://github.com/ros2/navigation/commit/d3d6c996c8f5bded76f083e608ad4a3b876df5ae

wjwwood:
Also, I would appreciate it if someone on @ros2/team could review this or just give some feedback on the things I put in the description.

Karsten1987:
I like the API in the way it is - meaning passing by ref. It's kind of clear to see what's happening here, ROS1 confirm and also since it's mainly primitives (int, floats, strings, etc), passing by pointer is kind of pointless :)

dhood:
> Looking at the ROS 1 code, it does not set the "set the parameter" with the default value if the parameter is not set

that matches what I would expect. if I want to change the stored value of the parameter, I should call `set_parameter`.

I think this would be less likely to trip people up (by having them wonder if it sets the parameter) if the function parameter weren't called `parameter`, but instead something that suggests that it is instead simply the "return value". (this is getting difficult to explain with three different meanings of the word parameter in one sentence!)

what I am suggesting is to rename the function parameter `ParameterT & parameter` to `ParameterT & value` since, in the end, what is returned in `parameter` does not necessarily represent the value of the parameter that is stored.

On the other hand, documentation that outlines the behaviour should be just as good, this was just food for thought.

wjwwood:
@dhood I suppose, reading this again, that you meant this prototype:

```c++
template<typename ParameterT>
bool
get_parameter(
  const std::string & name,
  ParameterT & value,
  const ParameterT & default_value) const;
```

dhood:
> I sort of agree. The only thing that caused me pause was that idea that if someone external from the node wants to know what the value of the parameter is (just a get), then they won't know what the node is actually using.

I might not have understood you, but isn't this what the `bool` return value is for? If it's true, then you know that the value that you got is the value that the node is using. if it's false, then you know that the node isn't using anything and that's why the provided default was used.

> And then you could later use these parameters without worrying about the default:
```ASSERT_TRUE(node->get_parameter("frame_id", frame_id));```

This seems to me like functionality that could be added _in addition_ to the signature proposed in this PR. Or do you see it as a competing alternative?


> I think you're doing it right, cppreference uses the term "function parameter". If you're talking about an item being passed into a function, that can be called an argument, imo:

yeah we just had parameter referring to the function parameter, the ROS parameter, and the variable name "`parameter`" all in one sentence heh. doesn't matter, was just an aside.

> @dhood I suppose, reading this again, that you meant this prototype:

yes

**Comment by wjwwood** [comment]
> I might not have understood you, but isn't this what the bool return value is for? If it's true, then you know that the value that you got is the value that the node is using. if it's false, then you know that the node isn't using anything and that's why the provided default was used.

I'm talking about remote nodes. They don't have this interface. They have the `/get_parameter` service call. If the parameter remains unset, then that service call cannot tell you if the node itself is using that parameter and if so what value it was using.

> This seems to me like functionality that could be added in addition to the signature proposed in this PR. Or do you see it as a competing alternative?

It can be in addition. However, right now it would need to throw if the parameter is unset, because it doesn't have the `bool` return code to indicate if it was set or not, nor does it have a default value to fall back on (without the "describe parameter" call preceding it).

> yes, just a tiny subtle change (I know you enjoy thinking carefully about naming)

Sure, I'll update it then: 7955c2e8a10993b4593b6baf918e349697eaf4aa and 0ee9f256430ea5aa28c082690de9b8784fd3756a

**Comment by wjwwood** [comment]
CI (since new change):

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2322/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1780/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2374/)

**Comment by wjwwood** [comment]
Ok, so after talking about it a bunch with @mikaelarguedas @dhood and @tfoote in person, we've changed our minds about this 😛:

- We should not set the parameter if it is unset, because it departs from the behavior of `dict.get(key, default)` in Python and `std::optional` in C++ does not have a method like this either ("get and set if not set").
- Instead I will rename this function to be `node->get_parameter_or(std::string name, T & value, const T & alternative)` to mimic the [`std::optional::value_or`](http://en.cppreference.com/w/cpp/utility/optional/value_or) method, which we think is clearer.
  - It will behave just as implemented right now, get the parameter value if set, otherwise return the default, but do not set the parameter value afterwards.
  - We could just use `std::optional` directly, but it's not widely available yet 😞.
- I'll also implement a `node->set_parameter_if_not_set(std::string name, const T & default_value)` method in this pr.
  - That way if you want the proposed new behavior (that we're declining) you can do it in two lines, with a `set_parameter_if_not_set(name, const T & default_value)` followed by a `get_parameter(name, T & value)`.
- In the future (not this pr), a user can use the "describe parameter" like method to set a default and then use `get_parameter(name, T & value)` freely afterwards without worrying about it being set, if that pattern is desired.
  - This is a better pattern for when you need to ensure a value is in the parameter before getting it, but avoids race conditions if this kind of "get with default" or "set if not set" function would be called from multiple threads (especially with different default values, which would not be advised but could happen accidentally).

**Comment by wjwwood** [comment]
Ok, changes made, and new tests in https://github.com/ros2/system_tests/pull/193.

File: rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp [hunk 1]
```diff
@@ -83,11 +83,18 @@ NodeParameters::set_parameters_atomically(
...
       if (p.get_type() != rclcpp::parameter::ParameterType::PARAMETER_NOT_SET) {
+        // case: parameter not set before, and input is something other than "NOT_SET"
         parameter_event->new_parameters.push_back(p.to_parameter());
       }
     } else if (p.get_type() != rclcpp::parameter::ParameterType::PARAMETER_NOT_SET) {
+      // case: parameter was set before, and input is something other than "NOT_SET"
       parameter_event->changed_parameters.push_back(p.to_parameter());
     } else {
+      // case: parameter was set before, and input is "NOT_SET"
+      // therefore we will "unset" the previous set parameter
```
mikaelarguedas: Nit " unset the previous**ly**  set parameter"

wjwwood:
Making use of this in https://github.com/ros2/navigation/pull/4.

wjwwood:
- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2323/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1781/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2376/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -235,6 +235,12 @@ class Node : public std::enable_shared_from_this<Node>
   rcl_interfaces::msg::SetParametersResult
   set_parameters_atomically(const std::vector<rclcpp::parameter::ParameterVariant> & parameters);
 
+  template<typename ParameterT>
+  void
+  set_parameter_if_not_set(
+    const std::string & name,
+    const ParameterT & default_value);
```
mikaelarguedas: NIt: "default_value" -> "value" or "new_value", default doesn't really carry meaning in this new approach. Same below

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -208,16 +208,46 @@ Node::register_param_change_callback(CallbackT && callback)
...
+void
+Node::set_parameter_if_not_set(
+  const std::string & name,
+  const ParameterT & default_value)
+{
+  rclcpp::parameter::ParameterVariant parameter_variant;
+  if (!this->get_parameter(name, parameter_variant)) {
+    this->set_parameters({
+          rclcpp::parameter::ParameterVariant(name, default_value),
+        });
```
mikaelarguedas: looks weird/over-indented but I guess it's what it takes to make uncrustify happy...

wjwwood: Yup, it even got it's own commit 👿: ff65663

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
wjwwood: a8f5d06

rclcpp/include/rclcpp/node.hpp
wjwwood: a8f5d06

wjwwood:
My changes in a8f5d06 compile locally, so I won't rerun CI for them.

mikaelarguedas:
> My changes in a8f5d06 compile locally, so I won't rerun CI for them.

:+1: Thanks for the quick iteration

wjwwood:
This is passed CI (unrelated test failures on macOS and Windows).

Any other comments before I merge these?

dirk-thomas:
If I read the function name `get_parameter_or` I immediately ask myself "or what?". I don't have a proposal at hand but `+-0` for using that function name.

wjwwood:
> If I read the function name get_parameter_or I immediately ask myself "or what?". I don't have a proposal at hand but +-0 for using that function name.

I agree it's a bit strange, and I'm open to alternative names.

I think it would be more natural if it were a member of a `ParameterVariant` for example, and it were called `value_or`, because that method would only have one parameter and the "or what" becomes more obvious. Though I'm uncertain as to whether or not there are other issues this approach brings up. If you're passionate about this as an alternative, speak up otherwise I'll just merge this one for lack of a better alternative.

On the positive side, this name would probably prompt you to try and understand it better. This happens for me with C++ all the time. I say "want the thing inside of that" so I go and look for methods and find `value()` and `value_or` (in this case `get_parameter` and `get_parameter_or`) and I think "oh what's the `value_or` option" and so I go and read about it. (and based on this realization I added some more docs to the pr)

wjwwood:
Since I haven't gotten anyone speaking up for the alternative in my previous comment, I'm going to merge this as-is and we can iterate on it again in the future.

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -208,16 +208,46 @@ Node::register_param_change_callback(CallbackT && callback)
...
 bool
-Node::get_parameter(const std::string & name, ParameterT & parameter) const
+Node::get_parameter(const std::string & name, ParameterT & value) const
 {
-  rclcpp::parameter::ParameterVariant parameter_variant(name, parameter);
+  rclcpp::parameter::ParameterVariant parameter_variant;
   bool result = get_parameter(name, parameter_variant);
-  parameter = parameter_variant.get_value<ParameterT>();
+  if (result) {
+    value = parameter_variant.get_value<ParameterT>();
```
dhood: I don't know what is configured differently on Portable but this line seems to have caused a warning: http://ci.ros2.org/job/ci_windows/2463/warnings41Result/

Issue #310, opened 2017-03-02
Implement rclcpp::time


Karsten1987:
I open an issue for sync between the @ros2/team in order to prevent double implementations.

mikaelarguedas:
Cross referencing with the original issue: https://github.com/ros2/ros2/issues/178

PR #311, opened 2017-03-15
Rclcpp time

Karsten1987:
connects to https://github.com/ros2/rclcpp/issues/310

prototype of `rclcpp::Time::now()`

For now it's mainly a convenience function, but may be object to further helper classes.
I introduced a new c++ class called Time, which by itself has a static member function `now` which returns basically a wrapper around `rcl_time_point_value_t`. In order to assign it to a message header, the same function can be called, but gets converter via the casting operator.

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.5)
 project(rclcpp)
 
 find_package(ament_cmake REQUIRED)
+find_package(builtin_interfaces REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rcl_interfaces REQUIRED)
 find_package(rmw REQUIRED)
@@ -17,7 +18,9 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wextra -Wpedantic")
 endif()
 
-include_directories(include)
+include_directories(
+  include
+  ${builtin_interfaces_INCLUDE_DIRS})
 
 set(${PROJECT_NAME}_SRCS
   src/rclcpp/any_executable.cpp
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -82,7 +82,9 @@ ament_export_dependencies(rosidl_generator_cpp)
 ament_export_dependencies(rosidl_typesupport_c)
 ament_export_dependencies(rosidl_typesupport_cpp)
 
-ament_export_include_directories(include)
+ament_export_include_directories(
+  include
+  ${builtin_interfaces_INCLUDE_DIRS})
 
 ament_export_libraries(${PROJECT_NAME})
 
@@ -173,6 +175,15 @@ if(BUILD_TESTING)
       rosidl_target_interfaces(test_externally_defined_services
         mock_msgs ${typesupport_impl_c})
     endforeach()
+  endif()
+
+  ament_add_gtest(test_time test/test_time.cpp
+    APPEND_LIBRARY_DIRS "${append_library_dirs}")
+  if(TARGET test_time)
+    target_include_directories(test_time PUBLIC
+      ${rcl_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_time ${PROJECT_NAME})
   endif()
 endif()
 
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -18,9 +18,7 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wextra -Wpedantic")
 endif()
 
-include_directories(
-  include
-  ${builtin_interfaces_INCLUDE_DIRS})
+include_directories(include)
 
 set(${PROJECT_NAME}_SRCS
   src/rclcpp/any_executable.cpp
@@ -60,6 +58,7 @@ set(${PROJECT_NAME}_SRCS
 add_library(${PROJECT_NAME} SHARED
   ${${PROJECT_NAME}_SRCS})
 ament_target_dependencies(${PROJECT_NAME}
+  "builtin_interfaces"
   "rcl"
   "rosidl_generator_cpp"
   "rosidl_typesupport_cpp")
@@ -77,14 +76,13 @@ install(
 )
 
 ament_export_dependencies(ament_cmake)
+ament_export_dependencies(builtin_interfaces)
 ament_export_dependencies(rcl)
 ament_export_dependencies(rosidl_generator_cpp)
 ament_export_dependencies(rosidl_typesupport_c)
 ament_export_dependencies(rosidl_typesupport_cpp)
 
-ament_export_include_directories(
-  include
-  ${builtin_interfaces_INCLUDE_DIRS})
+ament_export_include_directories(include)
 
 ament_export_libraries(${PROJECT_NAME})
 
@@ -180,9 +178,8 @@ if(BUILD_TESTING)
   ament_add_gtest(test_time test/test_time.cpp
     APPEND_LIBRARY_DIRS "${append_library_dirs}")
   if(TARGET test_time)
-    target_include_directories(test_time PUBLIC
-      ${rcl_INCLUDE_DIRS}
-    )
+    ament_target_dependencies(test_time
+      "rcl")
     target_link_libraries(test_time ${PROJECT_NAME})
   endif()
 endif()
--- a/rclcpp/include/rclcpp/rclcpp.hpp
+++ b/rclcpp/include/rclcpp/rclcpp.hpp
@@ -130,6 +130,7 @@
 #include "rclcpp/parameter_client.hpp"
 #include "rclcpp/parameter_service.hpp"
 #include "rclcpp/rate.hpp"
+#include "rclcpp/time.hpp"
 #include "rclcpp/utilities.hpp"
 #include "rclcpp/visibility_control.hpp"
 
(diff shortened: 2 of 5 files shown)


Karsten1987:
A few question, I'd like to have sorted out before merging:

- In case of error, I have a print. Is the idea to throw an exception at this point?
- Since `now` doesn't return the `uint64_t`, but rather first a class object and then a message object, how does that effect any realtime constraints?

mikaelarguedas:
FYI: added a connects to, assigned you on the corresponding issue and cross referenced this on the high level ticket

rclcpp/include/rclcpp/rclcpp.hpp
```diff
@@ -130,6 +130,7 @@
 #include "rclcpp/parameter_client.hpp"
 #include "rclcpp/parameter_service.hpp"
 #include "rclcpp/rate.hpp"
+//#include "rclcpp/time.hpp"
```
tfoote: Why added with a comment?

tfoote:
We've used time 0 in the past to indicate time uninitialized. Keeping that behavior seems reasonable for backwards compatibility. Though since we're starting from scratch an exception might be something we could consider. 

For realtime optimization might we want to simply recommend using the rcl api?

Karsten1987:
I changed the code to throw exceptions now, changed the test accordingly and also include this header directly into `rclcpp.hpp`

@tfoote if you could quickly verify these changes again, then i'd merge.

dirk-thomas:
If `rclcpp` is depending on `builtin_interfaces` it might be a good idea to move that message package out of `common_interfaces`. Otherwise all common interfaces need to be released before `rclcpp`. We can consider that when we get close to per-package releases though.

wjwwood:
> Otherwise all common interfaces need to be released before rclcpp. We can consider that when we get close to per-package releases though.

What's wrong with that? Do the packages in `common_interfaces` depend on `rclcpp`?

rclcpp/include/rclcpp/time.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include "rclcpp/exceptions.hpp"
+#include "rclcpp/visibility_control.hpp"
+
+namespace rclcpp
+{
+
+class Time
+{
+  rcl_time_point_value_t rcl_time_;
```
wjwwood: Please put member variables at the end of the class definition in its own private section.

If you want the constructors to be private, please mark them explicitly. See: https://google.github.io/styleguide/cppguide.html#Declaration_Order

rclcpp/test/test_time.cpp
```diff
@@ -0,0 +1,55 @@
...
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+};
+
+TEST(TestTime, rate_basics) {
+  using builtin_interfaces::msg::Time;
+  try {
+    rclcpp::Time::now<RCL_ROS_TIME>();
```
wjwwood: Should there be `RCLCPP_*_TIME` classes to avoid this mixing of rclcpp and rcl things?

dirk-thomas:
> > Otherwise all common interfaces need to be released before rclcpp. We can consider that when we get close to per-package releases though.

> What's wrong with that? Do the packages in `common_interfaces` depend on `rclcpp`?

No, but consider it valuable to keep dependencies which need to be released before a package can be released as small as possible.

rclcpp/test/test_time.cpp
```diff
@@ -0,0 +1,55 @@
...
+    FAIL();
+  } catch (const std::exception &) {
+    // test succeeded, ROStime not implemented yet
+    // TODO(karsten1987): Fix this test when ROStime is implemented
```
wjwwood: You can assert this case with these gtest macros:

https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#exception-assertions

mikaelarguedas:
I think that I agree with @dirk-thomas, low level interfaces used by the core would fit well in another repo. What do you guys think of `rcl_interfaces`?

dirk-thomas:
I haven't followed the conversations around time related stuff so likely this has been mentioned before. Why do we use a custom time type in rcl (`rcl_time_point_value_t`) and don't reuse the datatype defined in `builtin_interfaces`?

rclcpp/test/test_time.cpp
Karsten1987: well, as there is no real implementation, I could typedef them for now?

rclcpp/test/test_time.cpp
Karsten1987: oh that's great. Thanks for the link.

Karsten1987:
I am fine with whatever solution for the message packages. I have not enough insight about the build/package process to give qualified feedback on that one.

@dirk-thomas The idea here was to provide a real c++ object for time, which gives the opportunity to have handy helper functions or operators. That would leave the generated message struct a POD.

rclcpp/CMakeLists.txt
```diff
@@ -17,7 +18,9 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wextra -Wpedantic")
 endif()
 
-include_directories(include)
+include_directories(
+  include
+  ${builtin_interfaces_INCLUDE_DIRS})
```
Karsten1987: you mean on the packages which use this header file? As this is a header-only implementation, there is no target inside rclcpp which uses this header file.

Karsten1987:
@wjwwood I hope to have addressed all your points. But could you quickly verify that I did it the correct way?

rclcpp/CMakeLists.txt
wjwwood: > you mean on the packages which use this header file?

No I mean on the targets which use the header file.

> As this is a header-only implementation, there is no target inside rclcpp which uses this header file.

If no target needs these header files then it doesn't need to be in the `include_directories()`. I would have thought a test used this header file in a compilation unit in this package.

rclcpp/include/rclcpp/time.hpp
```diff
@@ -0,0 +1,76 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__TIME_HPP_
+#define RCLCPP__TIME_HPP_
+
+#include <chrono>
```
tfoote: I don't think this chrono include is necessary anymore. There are no calls to std::chrono visible in this diff.

wjwwood:
> I haven't followed the conversations around time related stuff so likely this has been mentioned before. Why do we use a custom time type in rcl (rcl_time_point_value_t) and don't reuse the datatype defined in builtin_interfaces?

`rcl_time_point_value_t` uses a single `uint64_t` rather than two 32-bit ints. I circulated this around at the time of the original rcl library implementation (https://github.com/ros2/rcl/pull/5), but no one commented on it. I specifically remember having in person discussions with people including @codebot, but I can't find any in depth discussion about it in the pull request.

My intention was to eventually write up some stuff about the time representation and advocate for a change in the structure of the `builtin_interfaces/Time.msg` file, but I never had the time to do so. However, that's a separate can of worms and introduces ROS 1 interoperability issues. In the meantime, having a single value for time in rcl greatly simplifies (in my opinion) the calculations done with it, making errors less likely.

In the future, if we decide to update the structure of the `Time.msg` to use a single value rather than two, then we could easily replace the `rcl_time_point_value_t` with the C struct generated for `builtin_interfaces/Time.msg`. Or we can discuss removing the type in rcl despite not changing the structure of `Time.msg`.

However, for the time being, I think it's functional enough to have this conversion constructor handle the conversion to the message generated type.

rclcpp/test/test_time.cpp
```diff
@@ -0,0 +1,49 @@
...
+protected:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+};
+
+TEST(TestTime, rate_basics) {
+  using builtin_interfaces::msg::Time;
+  EXPECT_ANY_THROW(rclcpp::Time::now<RCL_ROS_TIME>());
```
wjwwood: The `TODO(karsten1987)` that was here before still applies right?

tfoote:
RE: `rcl_time_point_value_t` The goal was that we could push the lower level time operations down to be shared between the rclc and rclcpp and rclpy implementations for now it's all in rclcpp but once we add the not implemented rostime datasource that would be great to share. And the single datatype will make that much easier. 

RE: moving the builtin_interfaces I think it makes sense to move it to rcl_interfaces repo makes sense at some point. Possibly into rcl_interfaces package too.

wjwwood:
I'm actually surprised that no messages in `rcl_interefaces` use `Time.msg` or `Duration.msg` yet.

mikaelarguedas:
While I think that putting these in the rcl_interfaces repo makes a lot of sense. I think it should stay a package on it's own. How I see it is that `rcl_interfaces` is defining core messages that most/all users won't use. `builtin_interfaces` is an hybrid package in the sense that it defines messages that we use in the core and we expect user to use them in most of their messages (either through Header or if they want to send Time/Duration information around). So I'd vote for keeping them in different packages but in the rcl_interfaces repo

Karsten1987:
So I'll merge this now if nobody has any outstanding comments. I propose we open another issue/PR for moving the `builtin_interfaces` to its appropriate place.

wjwwood:
Looks like there is a new warning on Linux from `test_time.cpp`:

http://ci.ros2.org/job/ci_linux/2341/warnings22Result/

PR #316, opened 2017-03-29
expose node namespace in API and pass to rcl

wjwwood:
Connects to ros2/rmw#95

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -67,22 +67,29 @@ public:
   /// Create a new node with the specified name.
   /**
    * \param[in] node_name Name of the node.
+   * \param[in] name_space Namespace of the node.
    * \param[in] use_intra_process_comms True to use the optimized intra-process communication
    * pipeline to pass messages between nodes in the same process using shared memory.
    */
   RCLCPP_PUBLIC
-  explicit Node(const std::string & node_name, bool use_intra_process_comms = false);
+  explicit Node(
+    const std::string & node_name,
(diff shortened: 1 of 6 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -67,22 +67,29 @@ class Node : public std::enable_shared_from_this<Node>
...
    * \param[in] node_name Name of the node.
+   * \param[in] name_space Namespace of the node.
    * \param[in] use_intra_process_comms True to use the optimized intra-process communication
    * pipeline to pass messages between nodes in the same process using shared memory.
    */
   RCLCPP_PUBLIC
-  explicit Node(const std::string & node_name, bool use_intra_process_comms = false);
+  explicit Node(
+    const std::string & node_name,
+    const std::string & name_space = "",
```
Karsten1987: do you have a specific reason for passing the name_space explicitly and don't capture it in a struct aka `node_options`?

we could also think of putting `use_intra_process_comms` into it.

wjwwood: I explained in the other pr that I feel that namespace is an important configurations that might ought to be directly in the signature, but I could be convinced otherwise.

In regard to the option structure, I neither like the idea of exposing the rcl structure directly nor do I like the idea of holding up this pr in order to implement a full option class for node. We do need to do the latter at some point, but I didn't want to do it right now.

The `use_intra_process_comms` is meant to be temporary. The idea is that in the future it would default to on and there would be an option in the options structure to disable them.

PR #320, opened 2017-04-11
use string_array_t from c_utilities package

Karsten1987:
connects to ros2/rmw#100

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -79,28 +79,28 @@ NodeGraph::get_topic_names_and_types() const
 std::vector<std::string>
 NodeGraph::get_node_names() const
 {
-  rcl_string_array_t node_names_c =
-    rcl_get_zero_initialized_string_array();
+  utilities_string_array_t node_names_c =
+    utilities_get_zero_initialized_string_array();
 
   auto ret = rcl_get_node_names(node_base_->get_rcl_node_handle(),
       &node_names_c);
-  if (ret != RMW_RET_OK) {
-    if (rmw_destroy_node_names(&node_names_c) != RMW_RET_OK) {
+  if (ret != RCL_RET_OK) {
+    if (utilities_string_array_fini(&node_names_c) != RCL_RET_OK) {
       RMW_SET_ERROR_MSG("Fatal: Leaking node_name memory.");
     }
     // *INDENT-OFF*
     throw std::runtime_error(
-      std::string("could not get node names: ") + rmw_get_error_string_safe());
+      std::string("could not get node names: ") + rcl_get_error_string_safe());
     // *INDENT-ON*
   }
 
   std::vector<std::string> node_names(&node_names_c.data[0],
     &node_names_c.data[0 + node_names_c.size]);
-  ret = rmw_destroy_node_names(&node_names_c);
+  ret = utilities_string_array_fini(&node_names_c);
   if (ret != RMW_RET_OK) {
     // *INDENT-OFF*
     throw std::runtime_error(
-      std::string("could not destroy node names: ") + rmw_get_error_string_safe());
+      std::string("could not destroy node names: ") + rcl_get_error_string_safe());
     // *INDENT-ON*
   }
 
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -54,10 +54,11 @@ NodeGraph::get_topic_names_and_types() const
   auto ret = rcl_get_topic_names_and_types(node_base_->get_rcl_node_handle(),
       &topic_names_and_types);
   if (ret != RMW_RET_OK) {
-    // *INDENT-OFF*
-    throw std::runtime_error(
-      std::string("could not get topic names and types: ") + rmw_get_error_string_safe());
-    // *INDENT-ON*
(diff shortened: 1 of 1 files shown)


rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -79,28 +79,28 @@ NodeGraph::get_topic_names_and_types() const
...
-    rcl_get_zero_initialized_string_array();
+  utilities_string_array_t node_names_c =
+    utilities_get_zero_initialized_string_array();
 
   auto ret = rcl_get_node_names(node_base_->get_rcl_node_handle(),
       &node_names_c);
-  if (ret != RMW_RET_OK) {
-    if (rmw_destroy_node_names(&node_names_c) != RMW_RET_OK) {
+  if (ret != RCL_RET_OK) {
+    if (utilities_string_array_fini(&node_names_c) != RCL_RET_OK) {
```
dirk-thomas: Shouldn't the return code be compared to one of these: https://github.com/ros2/c_utilities/blob/e6ff86450ddef392b18cb15a2bf5b165f77ca3d0/include/c_utilities/types/utilities_ret.h#L24-L26

Karsten1987: excellent catch! thanks

wjwwood: This still applies.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -79,28 +79,28 @@ NodeGraph::get_topic_names_and_types() const
...
     throw std::runtime_error(
-      std::string("could not get node names: ") + rmw_get_error_string_safe());
+      std::string("could not get node names: ") + rcl_get_error_string_safe());
     // *INDENT-ON*
   }
 
   std::vector<std::string> node_names(&node_names_c.data[0],
     &node_names_c.data[0 + node_names_c.size]);
-  ret = rmw_destroy_node_names(&node_names_c);
+  ret = utilities_string_array_fini(&node_names_c);
```
wjwwood: Same thing here, it's being compared to `RMW_RET_OK`, should be `UTILITIES_RET_OK`.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -79,28 +79,28 @@ NodeGraph::get_topic_names_and_types() const
...
   if (ret != RMW_RET_OK) {
     // *INDENT-OFF*
     throw std::runtime_error(
-      std::string("could not destroy node names: ") + rmw_get_error_string_safe());
+      std::string("could not destroy node names: ") + rcl_get_error_string_safe());
```
wjwwood: Is there actually an error string set here now that it is in `utilities_string_array_fini()`?

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -79,28 +79,30 @@ NodeGraph::get_topic_names_and_types() const
...
+  utilities_string_array_t node_names_c =
+    utilities_get_zero_initialized_string_array();
 
   auto ret = rcl_get_node_names(node_base_->get_rcl_node_handle(),
       &node_names_c);
-  if (ret != RMW_RET_OK) {
-    if (rmw_destroy_node_names(&node_names_c) != RMW_RET_OK) {
+  if (ret != RCL_RET_OK) {
+    if (utilities_string_array_fini(&node_names_c) != UTILITIES_RET_OK) {
       RMW_SET_ERROR_MSG("Fatal: Leaking node_name memory.");
```
wjwwood: Sorry, I know this isn't part of this pr, but this looks wrong. This should either print to stderr or throw an exception, probably the latter. No one is going to check the rmw error state after this function is called.

wjwwood: Can you please address this either in this pr or in an immediate follow up pr?

Karsten1987:
@wjwwood Would you mind reviewing the last commit again? I also think to have fixed a memory leak for the topic_names_and_types as this was not destroying memory before throwing.

I'll anyway run a new batch of CI before merging all this.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -54,10 +54,11 @@ NodeGraph::get_topic_names_and_types() const
   auto ret = rcl_get_topic_names_and_types(node_base_->get_rcl_node_handle(),
       &topic_names_and_types);
   if (ret != RMW_RET_OK) {
-    // *INDENT-OFF*
-    throw std::runtime_error(
-      std::string("could not get topic names and types: ") + rmw_get_error_string_safe());
-    // *INDENT-ON*
+    auto error_msg = std::string("Failed to get topic names and types.");
+    if (rmw_destroy_topic_names_and_types(&topic_names_and_types) != RMW_RET_OK) {
+      error_msg += "Fatal: Leaking topic_names_and_types memory.";
```
wjwwood: This would end up with a string like `Failed to get topic names and types.Fatal: Leaking topic_names_and_types memory.` which would be strange.

I think it's acceptable to throw from here, overruling the throw on the next line. Or you could throw here and reword it as `failed to cleanup after failing to get the topic names and types, leaking memory`.

Also, try to follow the existing style for error messages were possible. The message you're removing notably is not capitalized and lacks punctuation. This is also the case in other places in this file already.

PR #325, opened 2017-04-28
fix race conditions in guard condition handling

gerkey:
Starting from flaky requester-replier tests (e.g., http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/61/testReport/(root)/test_requester_replier__Empty__rclcpp__rclpy__rmw_fastrtps_cpp_Debug/test_requester_replier/),  @wjwwood and I debugged apparent race conditions in handling of guard conditions in rclcpp. There are three distinct but related changes here:

1. In [wait_for_service_nanoseconds](https://github.com/ros2/rclcpp/pull/325/files#diff-180df0c4318bfcfeab012a93ee002f71R73), create the graph event structure before checking whether the service is ready, to avoid the chance that it becomes ready between checking and event creation.
1. In [get_graph_event](https://github.com/ros2/rclcpp/pull/325/files#diff-ee33863795cb4ee8c58ca586c8a54d62R208), update the node's list of graph users before adding the node to the graph listener, to avoid missing the event because our node doesn't report having any graph users (https://github.com/ros2/rclcpp/blob/fix_guard_races/rclcpp/src/rclcpp/graph_listener.cpp#L151-L153).
1. In [wait_for_graph_change](https://github.com/ros2/rclcpp/pull/325/files#diff-ee33863795cb4ee8c58ca586c8a54d62R241), check the predicate condition that we want to have wake us up one time before going to sleep (the docs for `std::condition_variable::wait_until()` suggest that this isn't necessary, but it's not entirely clear).

Based on our testing, the second change is absolutely required, the first change is very likely required, and the third change is likely just defensive.

### DIFF
--- a/rclcpp/src/rclcpp/client.cpp
+++ b/rclcpp/src/rclcpp/client.cpp
@@ -70,6 +70,12 @@ bool
 ClientBase::wait_for_service_nanoseconds(std::chrono::nanoseconds timeout)
 {
   auto start = std::chrono::steady_clock::now();
+  // make an event to reuse, rather than create a new one each time
+  auto node_ptr = node_graph_.lock();
+  if (!node_ptr) {
+    throw InvalidNodeError();
+  }
+  auto event = node_ptr->get_graph_event();
   // check to see if the server is ready immediately
   if (this->service_is_ready()) {
     return true;
@@ -78,12 +84,6 @@ ClientBase::wait_for_service_nanoseconds(std::chrono::nanoseconds timeout)
     // check was non-blocking, return immediately
     return false;
   }
-  // make an event to reuse, rather than create a new one each time
-  auto node_ptr = node_graph_.lock();
-  if (!node_ptr) {
-    throw InvalidNodeError();
-  }
-  auto event = node_ptr->get_graph_event();
   // update the time even on the first loop to account for time spent in the first call
   // to this->server_is_ready()
   std::chrono::nanoseconds time_to_wait = timeout - (std::chrono::steady_clock::now() - start);
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -205,13 +205,13 @@ NodeGraph::get_graph_event()
 {
   auto event = rclcpp::event::Event::make_shared();
   std::lock_guard<std::mutex> graph_changed_lock(graph_mutex_);
+  graph_events_.push_back(event);
+  graph_users_count_++;
   // on first call, add node to graph_listener_
   if (should_add_to_graph_listener_.exchange(false)) {
     graph_listener_->add_node(this);
     graph_listener_->start_if_not_started();
   }
-  graph_events_.push_back(event);
-  graph_users_count_++;
   return event;
 }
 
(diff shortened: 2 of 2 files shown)


rclcpp/src/rclcpp/client.cpp
```diff
@@ -78,12 +84,6 @@ ClientBase::wait_for_service_nanoseconds(std::chrono::nanoseconds timeout)
     // check was non-blocking, return immediately
     return false;
   }
-  // make an event to reuse, rather than create a new one each time
-  auto node_ptr = node_graph_.lock();
-  if (!node_ptr) {
-    throw InvalidNodeError();
-  }
-  auto event = node_ptr->get_graph_event();
```
wjwwood: For time travelers, this block was moved above the `if (this->service_is_ready()) {return true;}` block on line 79. The reason being is that if the graph changed between that call on line 79 and here it could cause the state change of `this->service_is_ready()` to be missed.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -205,13 +205,13 @@ NodeGraph::get_graph_event()
...
   std::lock_guard<std::mutex> graph_changed_lock(graph_mutex_);
+  graph_events_.push_back(event);
+  graph_users_count_++;
   // on first call, add node to graph_listener_
   if (should_add_to_graph_listener_.exchange(false)) {
     graph_listener_->add_node(this);
     graph_listener_->start_if_not_started();
   }
-  graph_events_.push_back(event);
-  graph_users_count_++;
```
wjwwood: This block was moved because there was a data race between adding this node to the `graph_listener_` and updating the state internal to the node. This was just a silly mistake on my part.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -238,10 +238,13 @@ NodeGraph::wait_for_graph_change(
...
+  auto pred = [&event]() {
+      return event->check() || !rclcpp::utilities::ok();
+    };
   std::unique_lock<std::mutex> graph_lock(graph_mutex_);
-  graph_cv_.wait_for(graph_lock, timeout, [&event]() {
-    return event->check() || !rclcpp::utilities::ok();
-  });
+  if (!pred()) {
+    graph_cv_.wait_for(graph_lock, timeout, pred);
+  }
```
wjwwood: This reference says that, after a fashion, `wait_for` is equivalent to:

> Equivalent to
```c++
while (!pred()) {
    if (wait_until(lock, abs_time) == std::cv_status::timeout) {
        return pred();
    }
}
return true;
```
>  This overload may be used to ignore spurious wakeups.

-- http://en.cppreference.com/w/cpp/thread/condition_variable/wait_until

However, it does not appear to be guaranteed (or at least to my satisfaction) that the predicate is checked before entering to block. So this extra `if (!pred())` is really just defensive to changes in the std libraries implementation. On the other hand, it is a really cheap check and shouldn't do any harm.

gerkey:
CI:

* http://ci.ros2.org/job/ci_linux/2557/
* http://ci.ros2.org/job/ci_linux-aarch64/157/
* http://ci.ros2.org/job/ci_osx/2027/
* http://ci.ros2.org/job/ci_windows/2667/

Note that we're aiming to fix flaky tests here, so the real test will be in the nightlies, but the above jobs should at least demonstrate that we haven't made anything worse.

PR #331, opened 2017-05-31
Topic names

Karsten1987:
connects to ros2/rmw_fastrtps#114

will expand the topic name before asking `count_publisher` and `count_subscriber`. 
While doing, I noticed there is no cpp function for getting the node namespace, which is added in this PR as well. However, I am fine with removing that commit from this PR and open another one.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -102,7 +102,7 @@ public:
   get_name() const;
 
   /// Get the namespace of the node.
-  /** \return The namesoace of the node. */
+  /** \return The namespace of the node. */
   RCLCPP_PUBLIC
   const char *
   get_namespace() const;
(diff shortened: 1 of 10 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -101,6 +101,12 @@ class Node : public std::enable_shared_from_this<Node>
   const char *
   get_name() const;
 
+  /// Get the namespace of the node.
+  /** \return The namespace of the node. */
+  RCLCPP_PUBLIC
+  const char *
+  get_namespace() const;
```
wjwwood: I don't see an implementation for this. I would expect a change in `rclcpp/src/rclcpp/node.cpp` too.

wjwwood: Mabe add a simple test in this package too, since this would have been caught by that.

Karsten1987: done

PR #332, opened 2017-06-01
use time from rcutils

dirk-thomas:
Connect to ros2/rcutils#25.

Please review.

### DIFF
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -31,22 +31,22 @@ public:
   static Time
   now()
   {
-    rcl_time_point_value_t rcl_now = 0;
-    rcl_ret_t ret = RCL_RET_ERROR;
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/time.hpp
```diff
@@ -37,9 +37,9 @@ class Time
       throw std::runtime_error("RCL_ROS_TIME is currently not implemented.");
       ret = false;
     } else if (ClockT == RCL_SYSTEM_TIME) {
-      ret = rcl_system_time_now(&rcl_now);
+      ret = rcutils_system_time_now(&rcl_now);
```
wjwwood: Since you wish to keep the function name, the type and name of the rcl_now should be updated to use rcutils or this code changed to use the rcl time source.

dirk-thomas: Good point. Updated in bfa09fb78c5d82602c7e335325f79f920b124ec4

PR #333, opened 2017-06-05
include rcutils/time

mikaelarguedas:
follow up of #332 
Without this change rclcpp doesn't compile on my machine with a bunch of "not declared in this scope" errors

### DIFF
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -18,6 +18,7 @@
 #include <utility>
 
 #include "builtin_interfaces/msg/time.hpp"
+#include "rcl/time.h"
 #include "rcutils/time.h"
 
 #include "rclcpp/exceptions.hpp"
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -18,6 +18,7 @@
 #include <utility>
 
 #include "builtin_interfaces/msg/time.hpp"
+#include "rcutils/time.h"
 
 #include "rclcpp/exceptions.hpp"
 


dirk-thomas:
Since the current state seems to build successfully on all platforms in the nightly jobs I am wondering what the difference is for you. Are you sure you have the latest version of all repos?

mikaelarguedas:
Looks like I was missing some changes in rcl. given that both rcl and rcutils export their include directories, this PR should not be necessary. Still feels weird to me to not explicitly include the headers for the functions / types we use directly in this file though.

clalancette:
@mikaelarguedas Good you figured out the difference, but ultimately I agree with you that we should include the headers that we need.  So I'm +1 for this PR.

mikaelarguedas:
closing and reopening hoping that the second commit will show up

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2674/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/224/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2130/) (looks unrelated, @dirk-thomas are the logging tests known to be flaky ?)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2793/) (unrelated flaky test)

dirk-thomas:
The tests of the throttle macros rely on waking up within (I think currently) 20ms. I have seen OS X being unreliable in that area before. So that is likely flaky. We could increase the time intervals which should make it less flaky for the the cost of taking longer.

mikaelarguedas:
> So that is likely flaky. We could increase the time intervals which should make it less flaky for the the cost of taking longer.

I let you decide which solution you prefer. I just wanted to make sure you were aware of it.

Merging this then

PR #339, opened 2017-06-15
refactor to support multiple types and no_demangle

wjwwood:
Connects to ros2/ros2#361

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -321,7 +321,7 @@ public:
   register_param_change_callback(CallbackT && callback);
 
   RCLCPP_PUBLIC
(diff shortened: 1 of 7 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -321,10 +321,14 @@ class Node : public std::enable_shared_from_this<Node>
   register_param_change_callback(CallbackT && callback);
 
   RCLCPP_PUBLIC
-  std::map<std::string, std::string>
+  std::map<std::string, std::vector<std::string>>
```
dirk-thomas: Wouldn't a `set` make more sense here?

wjwwood: I'll ticket this.

Issue #347, opened 2017-08-02
rclcpp time should use rcl api


Karsten1987:
connects to https://github.com/ros2/rclcpp/pull/332#commitcomment-23439355

Karsten1987:
close with #348

PR #348, opened 2017-08-02
use rcl api for rclcpp time

Karsten1987:
connects to ros2/rclcpp#347

Not sure about the design of this, but it's using only rcl api. 
I put this in review directly to gather feedback.

### DIFF
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -36,8 +36,10 @@ public:
   now()
   {
     // TODO(karsten1987): This impl throws explicitely on RCL_ROS_TIME
-    // we have to do this, because rcl_time_source_init return RET_OK
-    // however sets system time under the hood
+    // we have to do this, because rcl_time_source_init returns RCL_RET_OK
+    // for RCL_ROS_TIME, however defaults to system time under the hood.
+    // ref: https://github.com/ros2/rcl/blob/master/rcl/src/rcl/time.c#L66-L77
+    // This section can be removed when rcl supports ROS_TIME correctly.
     if (ClockT == RCL_ROS_TIME) {
       throw std::runtime_error("RCL_ROS_TIME is currently not implemented.");
     }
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/time.hpp
```diff
@@ -33,38 +35,67 @@ class Time
   static Time
   now()
   {
-    rcutils_time_point_value_t rcutils_now = 0;
-    rcutils_ret_t ret = RCUTILS_RET_ERROR;
+    // TODO(karsten1987): This impl throws explicitely on RCL_ROS_TIME
+    // we have to do this, because rcl_time_source_init return RET_OK
+    // however sets system time under the hood
```
wjwwood: Can you add a sentence here about when we can remove this code, i.e. something like "this can be removed when the underlying implementation actually supports ROS time rather than implicitly defaulting to system time"?

rclcpp/include/rclcpp/time.hpp
```diff
@@ -33,38 +35,67 @@ class Time
...
+    ret = rcl_time_source_init(ClockT, &time_source);
+
+    if (ret == RCL_RET_ERROR) {
       rclcpp::exceptions::throw_from_rcl_error(
-        ret, "Could not get current time: ");
+        ret, "could not initialize time source: ");
     }
 
-    return Time(std::move(rcutils_now));
+    if (ret == RCL_RET_INVALID_ARGUMENT) {
```
wjwwood: What if ret is some other error like `RCL_RET_BAD_ALLOC` (just as an example)?

Wouldn't it better to move this check up and then do `if (ret != RCL_RET_OK) { rclcpp::exceptions::throw_from_rcl_error(...); }` after it?

See also: http://docs.ros2.org/beta2/api/rclcpp/namespacerclcpp_1_1exceptions.html#ad1d5563cfff9273f531a82c1dde13516

rclcpp/include/rclcpp/time.hpp
```diff
@@ -33,38 +35,67 @@ class Time
...
-  explicit Time(rcl_time_point_value_t && rcl_time)
+  explicit Time(rcl_time_point_t && rcl_time)
   : rcl_time_(std::forward<decltype(rcl_time)>(rcl_time))
   {}
+
+public:
+  virtual ~Time()
+  {
+    if (rcl_time_point_fini(&rcl_time_) != RCL_RET_OK) {
+      RCUTILS_LOG_FATAL("failed to destroy rclcpp::Time object");
```
wjwwood: I would suggest rewording this to something like `failed to reclaim rcl_time_point_t in destructor of rclcpp::Time`.

Karsten1987:
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2931/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/2375/)
windows: [Build Status](http://ci.ros2.org/job/ci_windows/3056/)

PR #351, opened 2017-08-06
time operators

Karsten1987:
connects to ros2/rclcpp#352

### DIFF
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -28,6 +28,31 @@
 namespace rclcpp
 {
 
+namespace
+{
+template<rcl_time_source_type_t ClockT = RCL_SYSTEM_TIME>
+rcl_time_point_t
+get_empty_time_point()
+{
+  rcl_time_source_t time_source;
+  auto ret = rcl_time_source_init(ClockT, &time_source);
+
+  if (ret != RCL_RET_OK) {
+    rclcpp::exceptions::throw_from_rcl_error(
+      ret, "could not initialize time source: ");
+  }
+  rcl_time_point_t time_point;
+  ret = rcl_time_point_init(&time_point, &time_source);
+
+  if (ret != RCL_RET_OK) {
+    rclcpp::exceptions::throw_from_rcl_error(
+      ret, "could not initialize time point: ");
+  }
+
+  return time_point;
+}
+}  // namespace
+
 class Time
 {
 public:
@@ -44,25 +69,9 @@ public:
       throw std::runtime_error("RCL_ROS_TIME is currently not implemented.");
     }
 
-    rcl_ret_t ret = RCL_RET_ERROR;
+    auto time_point = get_empty_time_point<ClockT>();
 
-    rcl_time_source_t time_source;
-    ret = rcl_time_source_init(ClockT, &time_source);
-
-    if (ret != RCL_RET_OK) {
-      rclcpp::exceptions::throw_from_rcl_error(
-        ret, "could not initialize time source: ");
-    }
-
-    rcl_time_point_t time_point;
-    ret = rcl_time_point_init(&time_point, &time_source);
-
-    if (ret != RCL_RET_OK) {
-      rclcpp::exceptions::throw_from_rcl_error(
-        ret, "could not initialize time point: ");
-    }
-
-    ret = rcl_time_point_get_now(&time_point);
+    auto ret = rcl_time_point_get_now(&time_point);
     if (ret != RCL_RET_OK) {
       rclcpp::exceptions::throw_from_rcl_error(
         ret, "could not get current time stamp: ");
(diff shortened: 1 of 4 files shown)


rclcpp/include/rclcpp/time.hpp
```diff
@@ -44,31 +69,38 @@ class Time
...
+  : rcl_time_(get_empty_time_point())
+  {
+    rcl_time_.nanoseconds = nanoseconds;
+  }
 
-    return Time(std::move(time_point));
+  Time(const builtin_interfaces::msg::Time & time_msg)  // NOLINT
+  : rcl_time_(get_empty_time_point())
+  {
+    rcl_time_.nanoseconds = RCL_S_TO_NS(time_msg.sec);
```
mikaelarguedas: this overlap the comments in https://github.com/ros2/rcutils/pull/44 but nothing prevents [`time_msg.sec`](https://github.com/ros2/common_interfaces/blob/master/builtin_interfaces/msg/Time.msg#L1) to be negative here. (same everywhere below)

rclcpp/include/rclcpp/time.hpp
```diff
@@ -44,31 +69,42 @@ class Time
...
 
-    if (ret != RCL_RET_OK) {
-      rclcpp::exceptions::throw_from_rcl_error(
-        ret, "could not initialize time point: ");
-    }
+  template<rcl_time_source_type_t ClockT = RCL_SYSTEM_TIME>
+  Time(int32_t seconds, uint32_t nanoseconds)
+  : rcl_time_(get_empty_time_point())
+  {
+    uint64_t ns = RCL_S_TO_NS(static_cast<uint64_t>(seconds));
```
dirk-thomas: Doesn't this need to check if `seconds` is negative?

Same below.

rclcpp/include/rclcpp/time.hpp
```diff
@@ -44,31 +69,42 @@ class Time
...
-      rclcpp::exceptions::throw_from_rcl_error(
-        ret, "could not get current time stamp: ");
+  template<rcl_time_source_type_t ClockT = RCL_SYSTEM_TIME>
+  explicit Time(uint64_t nanoseconds)
+  : rcl_time_(get_empty_time_point())
+  {
+    rcl_time_.nanoseconds = nanoseconds;
+  }
+
+  Time(const builtin_interfaces::msg::Time & time_msg)  // NOLINT
```
dirk-thomas: Why does this require `NOLINT`?

rclcpp/include/rclcpp/time.hpp
Karsten1987: should be corrected, however I only found one spot, where it was missing. Where's the other one?

dirk-thomas: Nvm, all other `static_cast`s have a check like `if (seconds < 0) {` before.

Karsten1987:
OK, so I think that's up for a final review.
I ended up splitting the file into a `.hpp` and `.cpp` and extended the tests for type and overflow checking. 

Given the underlying c API, I decided to write only one class `Time` which takes the respective timesource as parameter, defaulting to `RCL_SYSTEM_TIME`. I propose leaving that API as is for now until we have more concrete use cases (e.g. sim time) in which we might want to reconsider.

CI:
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2959/)
mac: [Build Status](http://ci.ros2.org/job/ci_osx/2388/)
win: [Build Status](http://ci.ros2.org/job/ci_windows/3074/)

rclcpp/src/rclcpp/time.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+rcl_time_source_t
+init_time_source(rcl_time_source_type_t clock = RCL_SYSTEM_TIME)
+{
+  rcl_time_source_t time_source;
+  auto ret = rcl_time_source_init(clock, &time_source);
+
+  if (ret != RCL_RET_OK) {
+    rclcpp::exceptions::throw_from_rcl_error(
+      ret, "could not initialize time source: ");
```
wjwwood: I think the `: ` is redundant:

https://github.com/ros2/rclcpp/blob/b90676871d397cbfdcfecdf265869efb0bd97d9f/rclcpp/src/rclcpp/exceptions.cpp#L59-L61

wjwwood: Here and below.

rclcpp/src/rclcpp/time.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+  return now;
+}
+
+Time::Time(int32_t seconds, uint32_t nanoseconds, rcl_time_source_type_t clock)
+: rcl_time_source_(init_time_source(clock)),
+  rcl_time_(init_time_point(rcl_time_source_))
+{
+  if (seconds < 0) {
+    throw std::runtime_error("can't convert a negative time msg to rclcpp::Time");
```
wjwwood: I would recommend this message instead, which I think is more to the point:

```
"cannot store a negative time point in rclcpp::Time"
```

rclcpp/src/rclcpp/time.cpp
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  uint64_t ns = RCL_S_TO_NS(static_cast<uint64_t>(seconds));
```
wjwwood: nitpick: you could assign this directly into `rcl_time_.nanoseconds` which is also a `uint64_t`.

rclcpp/src/rclcpp/time.cpp
wjwwood: Also below in at least one more place.

rclcpp/src/rclcpp/time.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+  rcl_time_.nanoseconds = RCL_S_TO_NS(static_cast<uint64_t>(time_msg.sec));
+  rcl_time_.nanoseconds += time_msg.nanosec;
+}
+
+bool
+Time::operator==(const rclcpp::Time & rhs) const
+{
+  if (rcl_time_.time_source->type != rhs.rcl_time_.time_source->type) {
+    throw std::runtime_error("can't compare times with different time sources");
```
wjwwood: I think we need to have a better overview of how this class should work. For example, I think that the fact that you cannot compare two Time objects with different sources is a runtime error is very inconvenient. Ideally it would be a compiler error when you try to compare a "system based" time point and a "steady based" time point. This pattern makes more sense when you're abstracting "system based" time point from "ROS based" time point because they have the same reference point. We might need to rethink the way it is setup in `rcl` as well to reflect this.

rclcpp/src/rclcpp/time.cpp
```diff
@@ -0,0 +1,233 @@
...
+Time
+Time::operator+(const rclcpp::Time & rhs) const
+{
+  if (rcl_time_.time_source->type != rhs.rcl_time_.time_source->type) {
+    throw std::runtime_error("can't add times with different time sources");
+  }
+
+  auto ns = rcl_time_.nanoseconds + rhs.rcl_time_.nanoseconds;
+  if (ns < rcl_time_.nanoseconds) {
+    throw std::runtime_error("addition leads to uint64_t overflow");
```
wjwwood: Probably http://en.cppreference.com/w/cpp/error/overflow_error instead of runtime.

rclcpp/src/rclcpp/time.cpp
```diff
@@ -0,0 +1,233 @@
...
+Time
+Time::operator-(const rclcpp::Time & rhs) const
+{
+  if (rcl_time_.time_source->type != rhs.rcl_time_.time_source->type) {
+    throw std::runtime_error("can't add times with different time sources");
+  }
+
+  auto ns = rcl_time_.nanoseconds - rhs.rcl_time_.nanoseconds;
+  if (ns > rcl_time_.nanoseconds) {
+    throw std::runtime_error("subtraction leads to uint64_t underflow");
```
wjwwood: http://en.cppreference.com/w/cpp/error/underflow_error too

rclcpp/test/test_time.cpp
```diff
@@ -48,3 +50,90 @@ TEST(TestTime, rate_basics) {
...
+
+  builtin_interfaces::msg::Time negative_time_msg;
+  negative_time_msg.sec = -1;
+  negative_time_msg.nanosec = 1;
+  try {
+    rclcpp::Time negative_time = negative_time_msg;
+    FAIL();
+  } catch (const std::exception &) {
+    SUCCEED();
+  }
```
wjwwood: https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#exception-assertions?

wjwwood: Below too.

Karsten1987: woooaaaah, that's so cool that you can actually put a complete block of code in it. thanks for that. didn't know it :)

rclcpp/src/rclcpp/time.cpp
Karsten1987: Agreed. I don't know if there should be three different time classes around. Open for opinions. I just basically need the default time class (and their operators) as for now.

rclcpp/test/test_time.cpp
dirk-thomas: If the invoked code is supposed to raise a specific exception I would suggest to use `EXPECT_THROW` rather than `EXPECT_ANY_THROW`.

mikaelarguedas:
typo 2: https://github.com/ros2/rclcpp/commit/2c3336510d4f109afddff5140550d97ffb094fe2

dhood:
@Karsten1987 these tests are failing on windows debug nightlies: http://ci.ros2.org/job/nightly_win_deb/568/testReport/junit/(root)/TestTime/operators/

PR #353, opened 2017-08-08
Restore old signal handler after shutdown

dhood:
connects to ros2/rmw_implementation#25

This PR makes two main changes:
1. the original signal handler is restored in an on_shutdown callback. This allows the original signal handler to be called even after rclcpp::shutdown has been called within a process
1. ~~during shutdown, ignore sigints from interrupting the shutdown process. otherwise deadlocks described in https://github.com/ros2/rmw_implementation/issues/25 can occur~~

This has been done to fix flaky tests caused by the deadlock described in https://github.com/ros2/rmw_implementation/issues/25. For it to take effect, it has to be combined with a change such as https://github.com/ros2/demos/commit/190d2f5e14dbcd84600b18e9356e53d254ebff68 so the processes call shutdown. ~~If we were to instead call shutdown in the rclcpp signal handler, that change wouldn't be necessary. Is it appropriate to replace [this block of code](https://github.com/ros2/rclcpp/blob/d090ddc35859901f9e07dac8eeaa1ce262e7a1fc/rclcpp/src/rclcpp/utilities.cpp#L86..L99) with a call to shutdown?~~

I have tried to maintain the preference for sigaction where available, following existing code, but please keep in mind that there may be subtleties of signal handlers that I'm likely to overlook

Standard CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2962/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/407/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2390/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3077/)

repeating the list_paramters* tests (usually very flaky): [Build Status](http://ci.ros2.org/job/ci_linux/2963/) (passed 60 times, failed on 60th because of [startup issue](http://ci.ros2.org/job/ci_linux/2963/testReport/junit/(root)/projectroot/test_tutorial_list_parameters__rmw_fastrtps_cpp/) that I understand to be unrelated)

### DIFF
--- a/rclcpp/src/rclcpp/utilities.cpp
+++ b/rclcpp/src/rclcpp/utilities.cpp
@@ -54,51 +54,6 @@ typedef void (* signal_handler_t)(int);
 static signal_handler_t old_signal_handler = 0;
 #endif
 
-void
-#ifdef HAS_SIGACTION
-signal_handler(int signal_value, siginfo_t * siginfo, void * context)
-#else
-signal_handler(int signal_value)
-#endif
-{
-  // TODO(wjwwood): remove? move to console logging at some point?
-  printf("signal_handler(%d)\n", signal_value);
-#ifdef HAS_SIGACTION
-  if (old_action.sa_flags & SA_SIGINFO) {
-    if (old_action.sa_sigaction != NULL) {
-      old_action.sa_sigaction(signal_value, siginfo, context);
-    }
-  } else {
-    // *INDENT-OFF*
-    if (
-      old_action.sa_handler != NULL &&  // Is set
-      old_action.sa_handler != SIG_DFL &&  // Is not default
-      old_action.sa_handler != SIG_IGN)  // Is not ignored
-    // *INDENT-ON*
-    {
-      old_action.sa_handler(signal_value);
-    }
-  }
-#else
-  if (old_signal_handler) {
-    old_signal_handler(signal_value);
-  }
-#endif
-  g_signal_status = signal_value;
-  {
-    std::lock_guard<std::mutex> lock(g_sigint_guard_cond_handles_mutex);
-    for (auto & kv : g_sigint_guard_cond_handles) {
-      rcl_ret_t status = rcl_trigger_guard_condition(&(kv.second));
-      if (status != RCL_RET_OK) {
-        fprintf(stderr,
-          "[rclcpp::error] failed to trigger guard condition: %s\n", rcl_get_error_string_safe());
-      }
-    }
-  }
-  g_is_interrupted.store(true);
-  g_interrupt_condition_variable.notify_all();
-}
-
 #ifdef HAS_SIGACTION
 struct sigaction
 set_sigaction(int signal_value, const struct sigaction & action)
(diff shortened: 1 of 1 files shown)


dhood:
@dirk-thomas has clarified that it is appropriate for the demo nodes to call shutdown themselves instead of having it called from rclcpp's signal handler. it doesn't necessarily make sense for the interrupt handler in rclcpp to call shutdown, since maybe users want to spin after sigint in some case.

https://github.com/ros2/system_tests/pull/215 adds some tests

dhood:
even if it's not appropriate to call shutdown from the signal handler I still think it's appropriate to ignore interrupts, because the deadlock occurs as a consequence of the guard condition triggering that happens in both places. as this PR is right now, `launch_testing` sending two interrupts in a row to a node can cause deadlock.

 I'm going to factorise out the logic of (ingoring interrupts + triggering the guard condition) and call it from both the signal handler and `rclcpp::shutdown`

dhood:
https://github.com/ros2/rclcpp/pull/353/commits/ec03f5f70e3c976cf4159aba14176496abebc66c factors the guard condition triggering logic out, and also changes from manually ignoring SIGINTS to just skipping responding to them if `g_is_interrupted` is true (but `signal_value != g_signal_status` might be more appropriate?).

dirk-thomas:
Why should a second SIGINT not notify the guard condition? I would expect a second signal to notify the condition again. Only in `shutdown` the signal handler is restored (symmetric to `init`). E.g. consider the following use case:

```
init()
...
spin()  // waiting for sigint to return from wait
// do some else
spin()  // a second SIGINT while waiting here should wake up the wait again
...
shutdown()
```

dhood:
Thanks for the clear example: I had the ideas of interrupt and shutdown conflated. This is because I thought that it was a double interrupt that was causing another deadlock to occur, since the first interrupt triggered some destruction. Looking closer, the example that I was double interrupting was exiting after the first interrupt (that's where the destruction was coming from), so it was just another instance of the same issue as in https://github.com/ros2/rmw_implementation/issues/25.

Just de-registering our signal handler should be sufficient to fix the deadlocks. I'll post back with CI to confirm

dhood:
ok finally got to 50 test passes without the other parameter flakiness issue interfering: [Build Status](http://ci.ros2.org/job/ci_linux/2988/) (this branch only)

the other flakiness issue is fixed in https://github.com/ros2/rclcpp/pull/356, so this job which includes commits from both branches was able to get the tests to pass 100 times in a row: [Build Status](http://ci.ros2.org/job/ci_linux/2991/)

conclusion is that ignoring sigints during shutdown is not necessary; restoring the state of the old signal handler is sufficient

dirk-thomas:
It would be good to separate the refatoring parts of the patch into one commit and the functional changes into a second commit.

dhood:
done; the refactoring in https://github.com/ros2/rclcpp/pull/353/commits/d7b7d7491f26b34fbc8e47bd81ea8ed84419f12f is optional, we can leave it out if it's easier

dhood:
CI after rebase
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2998/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/427/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2422/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3098/) (known flaky tests)

PR #354, opened 2017-08-10
Fix memory leaks in rclcpp

clalancette:
connects to ros2/ros2#387

CI is in the above issue

### DIFF
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -172,6 +172,7 @@ public:
         (std::cerr << ss.str()).flush();
         rcl_reset_error();
       }
+      delete service_handle_;
     }
   }
 
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -104,6 +104,7 @@ NodeBase::NodeBase(
         }
         throw_from_rcl_error(RCL_RET_ERROR, "failed to validate node name");
       }
+      delete rcl_node;
       throw rclcpp::exceptions::InvalidNodeNameError(
               node_name.c_str(),
               rmw_node_name_validation_result_string(validation_result),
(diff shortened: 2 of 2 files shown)


rclcpp/include/rclcpp/service.hpp
```diff
@@ -172,6 +172,7 @@ class Service : public ServiceBase
         (std::cerr << ss.str()).flush();
         rcl_reset_error();
       }
+      delete service_handle_;
```
dirk-thomas: If this class doesn't `owns_rcl_handle_` it can't call delete on the handle.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -104,6 +104,7 @@ NodeBase::NodeBase(
         }
         throw_from_rcl_error(RCL_RET_ERROR, "failed to validate node name");
       }
+      delete rcl_node;
```
dirk-thomas: What about the other locations where this function throws?

clalancette: Bah, of course.  I missed all of the throw_from_*() functions.  I'll fix it, thanks.

rclcpp/include/rclcpp/service.hpp
clalancette: Sorry, I'm not sure I understand.  This `delete service_handle_` is inside of the `if (owns_rcl_handle_)` block, so it will only be freed if we own the handle.  Did you mean something else?

dirk-thomas: Sorry, for the noise. I read the indentation wrong.

Karsten1987: do you actually need that? I'd assume that `rcl_service_fini` cleans up all heap allocations associated with that `service_handle_` and the destructor deletes this instances after going out of scope. But maybe I am missing something here.

clalancette: Certainly as the code stands we need it.  If you look at the (first) constructor for the Service() class, it does:

```
service_handle_ = new rcl_service_t;
```

Note that service_handle_ is not a shared_ptr<> or unique_ptr<>, and thus the programmer needs to explicitly call `delete service_handle_` somewhere (which was never being done).  `rcl_service_fini()` deals with freeing all of the *fields* of the rcl_service_t, but not the outer object itself.  This was all shown clearly when running the `test_executor` test from system_tests under valgrind.

Karsten1987: oh yeah, didn't see that it was actually a raw pointer. my bad.

PR #359, opened 2017-08-17
alias lifecycle_msgs with rclcpp_lifecycle constants

Karsten1987:
static constexpr aliases for lifecycle_msgs. This is mainly a convenience change.

ci:
[Build Status](http://ci.ros2.org/job/ci_linux/3032/)
[Build Status](http://ci.ros2.org/job/ci_osx/2438/)
[Build Status](http://ci.ros2.org/job/ci_windows/3110/)

### DIFF
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -17,6 +17,10 @@
 
 #include <string>
 
+#include "lifecycle_msgs/msg/state.hpp"
+
+#include "rcl_lifecycle/data_types.h"
+
 #include "rclcpp_lifecycle/visibility_control.h"
 
 // forward declare rcl_state_t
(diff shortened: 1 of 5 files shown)


dirk-thomas:
What is the reason for the aliases? Is it only to shorten the symbol names? I would like to point out that this adds a second location where these symbols are mainained (e.g. when adding something to the message) and might add to user confusion which one of the two to use.

Karsten1987:
I totally understand your point, and I was struggling with it myself. 
So the main problem I situation I have is that every callback returns a `rcl_lifecycle_transition_key_t` typedef'd to `uint8_t`. The constants defined inside the lifecycle_msgs are somehow not really obviously connected to these. Don't know how to cover that link correctly. The other alternative I see here, is that each callback natively returns a `uint8_t` and don't expose the `rcl_lifecycle_transition_key_t` towards the user.

Somewhat lost in translation here, it may only be vocabulary :)

dirk-thomas:
I am not sure if I get it. How do the new symbols help to make the connection from the return type `rcl_lifecycle_transition_key_t`?

Would it be better to extend the docblock of the functions returning `rcl_lifecycle_transition_key_t` to mention the message and its constants?

Karsten1987:
closing this because the approach in this PR is - as discussed - not the best.

Issue #357, opened 2017-08-15
Time operator tests failing on Windows Debug


mikaelarguedas:
Operator tests introduced in #351 are failing on the windows debug nightly:
example ci: http://ci.ros2.org/view/nightly/job/nightly_win_deb/573/testReport/junit/(root)/TestTime/operators/

@Karsten1987 will look into it soon, ticketing it here in the meantime to leave a paper trail

mikaelarguedas:
This has been fixed in #362 
Test passed a 100 times on this job http://ci.ros2.org/job/ci_windows/3167/

PR #362, opened 2017-08-23
freeing Time members in destructor, adding copy constructor / assignment operator

Karsten1987:
Connects to ros2/rclcpp#357

add a copy constructor to the rclcpp::Time class which is necessary to fix the windows debug tests.
I tested it locally on my windows machine.

CI:
[Build Status](http://ci.ros2.org/job/ci_windows/3141/)

### DIFF
--- a/rclcpp/src/rclcpp/time.cpp
+++ b/rclcpp/src/rclcpp/time.cpp
@@ -103,7 +103,7 @@ Time::Time(uint64_t nanoseconds, rcl_time_source_type_t clock)
 }
 
 Time::Time(const Time & rhs)
-: rcl_time_source_(rhs.rcl_time_source_),
+: rcl_time_source_(init_time_source(rhs.rcl_time_source_.type)),
   rcl_time_(init_time_point(rcl_time_source_))
 {
   rcl_time_.nanoseconds = rhs.rcl_time_.nanoseconds;
@@ -142,7 +142,7 @@ Time::operator builtin_interfaces::msg::Time() const
 void
 Time::operator=(const Time & rhs)
 {
-  rcl_time_source_ = rhs.rcl_time_source_;
+  rcl_time_source_ = init_time_source(rhs.rcl_time_source_.type);
   rcl_time_ = init_time_point(rcl_time_source_);
   rcl_time_.nanoseconds = rhs.rcl_time_.nanoseconds;
 }
--- a/rclcpp/src/rclcpp/time.cpp
+++ b/rclcpp/src/rclcpp/time.cpp
@@ -79,7 +79,7 @@ Time::now(rcl_time_source_type_t clock)
     rclcpp::exceptions::throw_from_rcl_error(
       ret, "could not get current time stamp");
   }
-
+  fprintf(stderr, "leaving now() with time source %d\n", now.rcl_time_.time_source->type);
   return now;
 }
 
@@ -100,6 +100,13 @@ Time::Time(uint64_t nanoseconds, rcl_time_source_type_t clock)
   rcl_time_(init_time_point(rcl_time_source_))
 {
   rcl_time_.nanoseconds = nanoseconds;
+}
+
+Time::Time(const Time & rhs)
+: rcl_time_source_(rhs.rcl_time_source_),
+  rcl_time_(init_time_point(rcl_time_source_))
+{
+  rcl_time_.nanoseconds = rhs.rcl_time_.nanoseconds;
 }
 
 Time::Time(const builtin_interfaces::msg::Time & time_msg)  // NOLINT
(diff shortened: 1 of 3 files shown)


Karsten1987:
Connects to ros2/rclcpp#357

dhood:
there's a `CI_CMAKE_BUILD_TYPE` option in the jobs that can be set to debug

rclcpp/src/rclcpp/time.cpp
```diff
@@ -79,7 +79,6 @@ Time::now(rcl_time_source_type_t clock)
     rclcpp::exceptions::throw_from_rcl_error(
       ret, "could not get current time stamp");
   }
-
```
wjwwood: nitpick: Why?

rclcpp/test/test_time.cpp
```diff
@@ -102,8 +102,8 @@ TEST(TestTime, operators) {
   EXPECT_EQ(sub.nanoseconds(), young.nanoseconds() - old.nanoseconds());
   EXPECT_EQ(sub, young - old);
 
-  rclcpp::Time system_time(1, 0, RCL_SYSTEM_TIME);
-  rclcpp::Time steady_time(2, 0, RCL_STEADY_TIME);
+  rclcpp::Time system_time(0, 0, RCL_SYSTEM_TIME);
+  rclcpp::Time steady_time(0, 0, RCL_STEADY_TIME);
```
wjwwood: What's the purpose of this change?

Karsten1987: to demonstrate that even though the timestamps are equal, the `operator==` fails because of the different time sources.

dirk-thomas:
I think there is something missing here. The patch tries to fix a memory ownership problem which is likely not specific to Windows Debug.

The underlying `rcl_time_source_t` stores a `data *` and `rcl_time_point_t` stores a `rcl_time_source_t *`. Without an explicit copy constructor / assignment operator the original `Time` instance as well as the copy will implicitly share the same underlying pointers which have been allocated by the original time instance.

It is certainly good to add the copy constructor / assignment operator to make the two `Time` instances independent from each other. But the fact that the lack of those results in a problem on Windows Debug indicates that something else is wrong with the memory.

While most of the time sources only set the data pointer to something "fixed" `RCL_ROS_TIME` will actually allocate memory for it. The fact that an example copy constructing a time with that source doesn't fail with a double free error indicates that the memory is never freed. That is easily demonstrated with a mini example in `valgrind` (unrelated to this patch):

```
rclcpp::Time t(1, 2, RCL_ROS_TIME);
```

```
16 bytes in 1 blocks are definitely lost in loss record 1 of 2
   at 0x4C2FB55: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x6801A46: rcl_ros_time_source_init (time.c:135)
   by 0x680193C: rcl_time_source_init (time.c:103)
   by 0x51C4205: (anonymous namespace)::init_time_source(rcl_time_source_type_t) (time.cpp:34)
   by 0x51C45D3: rclcpp::Time::Time(int, unsigned int, rcl_time_source_type_t) (time.cpp:<line>)
   ...
```

So as a conclusion the destructor of `Time` should be fixed first to actually `fini`ing the underlying rcl structures correctly. After that the above example should not leak anymore. After that I would expect "double free" errors due to sharing of the same underlying rcl resources. This is where this patch comes in handy which should create truely independent instance.

Karsten1987:
Having looked at the `rcl_time_` implementation, I indeed forgot to cleanup the resources correctly. The last commit should have fixed that. 

Running CI again:

win debug: [Build Status](http://ci.ros2.org/job/ci_windows/3142/)
win release: [Build Status](http://ci.ros2.org/job/ci_windows/3143/)

for comparison: 
linux debug: [Build Status](http://ci.ros2.org/job/ci_linux/3071/)
linux release: [Build Status](http://ci.ros2.org/job/ci_linux/3070/)

rclcpp/test/test_time.cpp
allenh1: Maybe this should just be added as an additional test? Sounds like a good distinction.

rclcpp/src/rclcpp/time.cpp
```diff
@@ -130,6 +140,14 @@ Time::operator builtin_interfaces::msg::Time() const
 }
 
 void
+Time::operator=(const Time & rhs)
+{
+  rcl_time_source_ = rhs.rcl_time_source_;
```
dirk-thomas: Can you try that with a `RCL_ROS_TIME`? Since it internally allocates `data` within the `rcl_time_source_t` I would expect that this can't be just copied here.

rclcpp/test/test_time.cpp
```diff
@@ -123,6 +123,16 @@ TEST(TestTime, operators) {
   EXPECT_ANY_THROW((void)(now > later));
   EXPECT_ANY_THROW((void)(now + later));
   EXPECT_ANY_THROW((void)(now - later));
+
+  for (auto time_source : {RCL_ROS_TIME, RCL_SYSTEM_TIME, RCL_STEADY_TIME}) {
+    rclcpp::Time time = rclcpp::Time(0, 0, time_source);
+    rclcpp::Time copy_constructor_time = time;
```
dirk-thomas: This should be `rclcpp::Time copy_constructor_time(time)` if you want to call the copy constructor.

Karsten1987: this line will also call the copy constructor.

dirk-thomas: True, maybe still update the line to make the intention clear?

Karsten1987:
Latest CI for windows debug:

[Build Status](http://ci.ros2.org/job/ci_windows/3155/)

Karsten1987:
MacOS memory sanitizer doesn't complain, so I'll take this as a valid outcome.

PR #364, opened 2017-08-31
use forward slashes instead of double underscores

Karsten1987:
The current parameter service and client implementation still use double underscores. I believe this has been forgotten to change when namespacing was introduced.

linux: [Build Status](http://ci.ros2.org/job/ci_linux/3119/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/2488/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/3177/) (test failures unrelated)

### DIFF
--- a/rclcpp/src/rclcpp/parameter_client.cpp
+++ b/rclcpp/src/rclcpp/parameter_client.cpp
@@ -19,6 +19,8 @@
 #include <string>
 #include <vector>
 
+#include "./parameter_service_names.hpp"
+
 using rclcpp::parameter_client::AsyncParametersClient;
 using rclcpp::parameter_client::SyncParametersClient;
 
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Should those service names share a common definition somewhere? At least in a header, maybe even in an .msg file?

dirk-thomas:
Should the header file be visible to the outside and therefore be in the include folder instead?

Karsten1987:
I don't think so, because in the end you'd call the ParameterClient/ParameterService to get the variants and don't want to actually call the service itself (IMO).
I further think that this needs more refactoring once we start implementing the parameters in C which is possibly the right place to also put these definitions into constants within a msg file.

Karsten1987:
waiting on CI:

[Build Status](http://ci.ros2.org/job/ci_osx/2495/)
[Build Status](http://ci.ros2.org/job/ci_linux/3134/)
[Build Status](http://ci.ros2.org/job/ci_windows/3191/)

PR #367, opened 2017-09-01
change parameter separator to forward slash

Karsten1987:
changes:

- change parameter separator from dot (`.`) to forward slash `/`
- exclude lambda condition for visibility
- use rcutils logging macros for warning

why?

this allows to correctly use the prefix when calling `list_parameters`

There are currently no unit-tests for rclcpp parameters (AFAICT). This should be ticketed separately.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -198,7 +198,7 @@ NodeParameters::list_parameters(const std::vector<std::string> & prefixes, uint6
   std::lock_guard<std::mutex> lock(mutex_);
   rcl_interfaces::msg::ListParametersResult result;
 
-  char separator = '/';
+  const char separator = '/';
   for (auto & kv : parameters_) {
     bool get_all = (prefixes.size() == 0) &&
       ((depth == rcl_interfaces::srv::ListParameters::Request::DEPTH_RECURSIVE) ||
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -22,6 +22,7 @@
 
 #include "rcl_interfaces/srv/list_parameters.hpp"
 #include "rclcpp/create_publisher.hpp"
+#include "rcutils/logging_macros.h"
 #include "rmw/qos_profiles.h"
 
 using rclcpp::node_interfaces::NodeParameters;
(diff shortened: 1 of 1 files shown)


dhood:
is this change only necessary because of how you're _setting_ the parameters (with a slash separating them instead of a dot)?
we have tests for list_parameters that work with the existing prefix, but only if you specify paramters [using the same prefix](https://github.com/ros2/system_tests/blob/b778786d74b4d5b9b0fbe938a0893a2ed60653b8/test_rclcpp/test/parameter_fixtures.hpp#L40).
If we make a change here it should be to make the prefix configurable, not switch from one arbitrary one to another IMO

Karsten1987:
I would advocate for a convention on how to set the parameters (analog to how they are set in ROS1). 
When loading a yaml file, e.g.

```
my_node:
  topics:
     sub: my_node_subscription
     pub: my_node_publication
```
I would set these parameters:
```
/my_node/topics/sub (= my_node_subsription)
/my_node/topics/pub (= my_node_publication)
```

dhood:
I don't have an issue with it myself; others may have more context if there was a reason for the dots over slashes, but from [the comment](https://github.com/ros2/rclcpp/pull/367/files#diff-43c1c52da4c73f0a25534075d66ece63L200) it seems like a placeholder.

Main point I wanted to make is that there _is_ a convention at the moment (it's just a bit secret..) and list_parameters will only work with that specific one being used symmetrically. So this PR isn't fixing a broken list_parameters, per se, but instead is changing the convention to what ROS 1 users might expect. If we decide to go in that direction we have tests and demos that would need to be updated accordingly

mikaelarguedas:
Looks like the place holder was chosen when we didn't want to use slashes anywhere, now that we do through namespaces I'm not sure that's a blocker anymore. I think the placeholder was chosen in a Java style and didnt have a strong rationale behind it

Karsten1987:
updated the related tests.

CI:
[Build Status](http://ci.ros2.org/job/ci_windows/3204/)
[Build Status](http://ci.ros2.org/job/ci_osx/2509/)
[Build Status](http://ci.ros2.org/job/ci_linux/3152/)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -197,26 +198,27 @@ NodeParameters::list_parameters(const std::vector<std::string> & prefixes, uint6
   std::lock_guard<std::mutex> lock(mutex_);
   rcl_interfaces::msg::ListParametersResult result;
 
-  // TODO(esteve): define parameter separator, use "." for now
+  char separator = '/';
```
dirk-thomas: Could this be `const`?

mikaelarguedas:
Follow up on this: It seems like the rationale for not using `/` is to avoid ambiguity (mentioned on the parameters design document as follow):

"
Address all parameters without ambiguity in the names
For example, one of the challenges of the current system is that there is a naming ambiguity between nodes and parameters `/foo/bar/baz` could be a node `/foo/bar/baz` or a private parameter `baz` on node `/foo/bar`.
"

PR #368, opened 2017-09-02
parameter client takes node interfaces

Karsten1987:
I had to refactor the constructor of the AsyncParameterClient in order to create an instance of it from a lifecycle node.

The current implementation keeps a handle on the original node pointer: 
https://github.com/ros2/rclcpp/blob/b1ed15ebc7e120b7e4e95df4caaf7f942fccbf15/rclcpp/src/rclcpp/parameter_service.cpp#L29
Why is that the case?

### DIFF
--- a/rclcpp/src/rclcpp/parameter_client.cpp
+++ b/rclcpp/src/rclcpp/parameter_client.cpp
@@ -34,7 +34,7 @@ AsyncParametersClient::AsyncParametersClient(
   if (remote_node_name != "") {
     remote_node_name_ = remote_node_name;
   } else {
-    remote_node_name_ = node_->get_name();
+    remote_node_name_ = node_base_interface->get_name();
   }
 
   rcl_client_options_t options = rcl_client_get_default_options();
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -24,29 +24,81 @@
 using rclcpp::parameter_client::AsyncParametersClient;
 using rclcpp::parameter_client::SyncParametersClient;
 
+RCLCPP_PUBLIC
```
wjwwood: This does not need to be in the cpp file.

Karsten1987:
After running CI, all tests are actually failing and I need help on this.
In particular, the internal spin never returns and the tests are timing out:

https://github.com/ros2/rclcpp/blob/interface_constructor_for_parameter_client/rclcpp/src/rclcpp/parameter_client.cpp#L419-L420

any ideas?

wjwwood:
I re-reviewed the code, but I don't see any obvious issue.

Karsten1987:
In the end, it was a stupid copy & paste mistake. Fixed the typos and rebased.

CI:
[Build Status](http://ci.ros2.org/job/ci_linux/3159/)
[Build Status](http://ci.ros2.org/job/ci_osx/2516/)
[Build Status](http://ci.ros2.org/job/ci_windows/3210/)

PR #371, opened 2017-09-08
update rclcpp to use the refactored TimeSource Clock logic

tfoote:
Connects to ros2/rcl#164
Continuing to follow up ros2/design#144

This implements a TimeSource in rclcpp, adds the Clock class. 

This still has many of the in progress commits, I'd suggest focusing on the overall diff. I'll clean up the history later.

@wjwwood Do you still have the Time/Duration updates pending?

Another question that is not resolved yet is how to make setting a Clock up easily with a Node. The current design is very modular but requires overhead to connect the TimeSource, node, and Clock all together. Suggestions for integrating them into the Node API would be appreciated, however I will probably hold off on that for this round and just say use the modular API. 
`Clock clock; TimeSource ts(node); ts.associateClock(clock);`

Before this passes CI a set of coordinated changes to all usages of rclcpp::Time::now() like above.

### DIFF
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -1,0 +1,56 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__CLOCK_HPP_
+#define RCLCPP__CLOCK_HPP_
+
+#include "rclcpp/time.hpp"
+#include "rclcpp/visibility_control.hpp"
+
+#include "rcl/time.h"
+
+namespace rclcpp
+{
+
+class TimeSource;
+
+class Clock
+{
+public:
+  RCLCPP_PUBLIC
+  explicit Clock(rcl_clock_type_t clock_type = RCL_SYSTEM_TIME);
+
+  RCLCPP_PUBLIC
+  ~Clock();
+
+  RCLCPP_PUBLIC
+  Time
+  now();
+
+  RCLCPP_PUBLIC
+  bool
+  isROSTimeActive();
+
+  RCLCPP_PUBLIC
+  rcl_clock_type_t
+  getClockType();
+
+private:
+  rcl_clock_t rcl_clock_;
+  friend TimeSource;  // Allow TimeSource to access the rcl_clock_ datatype.
+};
+
+}  // namespace rclcpp
+
+#endif  // RCLCPP__CLOCK_HPP_
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -102,6 +102,10 @@ public:
   get_clock_type();
 
   // Add a callback to invoke if the jump threshold is exceeded.
+  /**
+   * These callback functions must remain valid as long as the
+   * returned shared pointer is valid.
+   */
   RCLCPP_PUBLIC
   JumpHandler::SharedPtr
   create_jump_callback(
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -31,27 +31,32 @@ namespace rclcpp
 
 class TimeSource;
 
-class TimeJump
+/*  \brief A struct to represent a timejump.
+ *  It epresents the time jump duration and whether it changed clock type.
+ */
+struct TimeJump
 {
-public:
   typedef enum ClockChange_t
   {
-    ROS_TIME_NO_CHANGE,
-    ROS_TIME_ACTIVATED,
-    ROS_TIME_DEACTIVATED,
-    SYSTEM_TIME_NO_CHANGE
+    ROS_TIME_NO_CHANGE,    ///< The time type before and after the jump is ROS_TIME
+    ROS_TIME_ACTIVATED,    ///< The time type switched to ROS_TIME from SYSTEM_TIME
+    ROS_TIME_DEACTIVATED,  ///< The time type switched to SYSTEM_TIME from ROS_TIME
+    SYSTEM_TIME_NO_CHANGE  ///< The time type before and after the jump is SYSTEM_TIME
   } ClockChange_t;
 
-  ClockChange_t jump_type_;
-  rcl_duration_t delta_;
+  ClockChange_t jump_type_;  ///< The change in clock_type if there is one.
+  rcl_duration_t delta_;     ///< The change in time value.
 };
 
+/* \brief A class to store a threshold for a TimeJump
+ * This class can be used to evaluate a time jump's magnitude.
+*/
 class JumpThreshold
 {
 public:
-  uint64_t min_forward_;
-  uint64_t min_backward_;
-  bool on_clock_change_;
+  uint64_t min_forward_;   ///< The minimum jump forward to be considered exceeded..
+  uint64_t min_backward_;  ///< The minimum backwards jump to be considered exceeded.
+  bool on_clock_change_;   ///< Whether to trigger on any clock type change.
 
   // Test if the threshold is exceeded by a TimeJump
   RCLCPP_PUBLIC
@@ -59,11 +64,11 @@ public:
   is_exceeded(const TimeJump & jump);
 };
 
-class JumpCallback : public std::enable_shared_from_this<JumpCallback>
+class JumpHandler : public std::enable_shared_from_this<JumpHandler>
 {
 public:
-  RCLCPP_SMART_PTR_DEFINITIONS(JumpCallback)
-  JumpCallback(
+  RCLCPP_SMART_PTR_DEFINITIONS(JumpHandler)
+  JumpHandler(
     std::function<void()> pre_callback,
     std::function<void(TimeJump)> post_callback,
     JumpThreshold & threshold);
@@ -90,15 +95,15 @@ public:
 
   RCLCPP_PUBLIC
   bool
-  isROSTimeActive();
+  ros_time_is_active();
 
   RCLCPP_PUBLIC
   rcl_clock_type_t
-  getClockType();
+  get_clock_type();
 
   // Add a callback to invoke if the jump threshold is exceeded.
   RCLCPP_PUBLIC
-  JumpCallback::SharedPtr
+  JumpHandler::SharedPtr
   create_jump_callback(
     std::function<void()> pre_callback,
     std::function<void(const TimeJump &)> post_callback,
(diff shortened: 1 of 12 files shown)


wjwwood:
Sorry for missing the @ mention before. I still have some draft changes, but I haven't pushed them yet.

dirk-thomas:
@ros2/dev Request for API review. Not ready to be merged yet.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+  rclcpp::Time msg_time = rclcpp::Time(*msg);
+  rclcpp::Time now = clock->now();
+  auto diff = now - msg_time;
+  rclcpp::TimeJump jump;
+  jump.delta_.nanoseconds = diff.nanoseconds();
+  // TODO(tfoote) fill in time source change on jump
+
+
+  // TODO(tfoote) potential race condition with target going out of scope
```
wjwwood: What is `target` in this context? `clock` is a shared pointer and `jump` is local. What could go out of scope?

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -0,0 +1,223 @@
...
+void TimeSource::setClock(const builtin_interfaces::msg::Time::SharedPtr msg,
+  std::shared_ptr<rclcpp::Clock> clock)
+{
+  // TODO(tfoote) Move enabledROS time logic here.
+
+  rclcpp::Time msg_time = rclcpp::Time(*msg);
+  rclcpp::Time now = clock->now();
+  auto diff = now - msg_time;
+  rclcpp::TimeJump jump;
+  jump.delta_.nanoseconds = diff.nanoseconds();
```
wjwwood: `delta_` is an `rcl_duration_t` type, but it never has `init` and `fini` called on it...

wjwwood: I think `TimeJump` just needs a constructor that takes nanoseconds and a destructor to match.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+void TimeSource::clock_cb(const builtin_interfaces::msg::Time::SharedPtr msg)
+{
+  // RCUTILS_LOG_INFO("Got clock message");
+  if (!this->ros_time_valid_) {
+    enableROSTime();
+  }
+  // TODO(tfoote) switch this to be based on if there are clock publishers or use_sim_time
+  // TODO(tfoote) also setup disable
```
wjwwood: I know it's not part of the API review, but are these going to be addressed before this pr is merged?

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+      }
+    }
+  }
+  for (auto & deleted_parameter : event->deleted_parameters) {
+    if (deleted_parameter.name == "use_sim_time") {
+      // If the parameter is deleted mark it as unset but dont' change state.
+      parameter_state_ = UNSET;
+    }
+  }
```
wjwwood: This function might be an argument for `event` having only one list of events, with each event having an action (new/change/delete). That way you could have looped once in this function rather than three times.

wjwwood: Maybe just a convenience function to zip these three lists together for you with their action type.

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include "rcl/time.h"
+
+namespace rclcpp
+{
+
+class Duration
+{
+public:
+  RCLCPP_PUBLIC
+  Duration(int32_t seconds, uint32_t nanoseconds, rcl_clock_type_t clock_type = RCL_SYSTEM_TIME);
```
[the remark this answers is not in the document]
tfoote: It's obvious that a time point on the ROS clock can't be differentiated from a time on the SYSTEM clock. However if you think of two timers being triggered simultaneously on the two clocks. Such that you get R1 and R2 and S1 and S2 being the time points. Then R2 - R1 gives you a Dr and S2 - S1 gives you Ds. Then if they were type compatible R1 + Dr - Ds should equal R1. But that's only true if the system clock and ROS clock are progressing at the same rate. As this is not an assumption that we can make I think that the durations should encode the clock type for the duration as well as the time points.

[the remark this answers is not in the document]
wjwwood: > It's obvious that a time point on the ROS clock can't be differentiated from a time on the SYSTEM clock.

Why not? How would I calculate the difference in the wall clock and the ROS time? They reference the same starting point, so I think they could be differenced.

> Then if they were type compatible R1 + Dr - Ds should equal R1.

But what does that equation mean?

The resulting difference between two timepoints is measured in seconds, but has no bearing on how much "real" time passes between observing the two timepoints.

For example, if you measure R1 to be 0 at t=0 and R2 to be 1 at t=1, then Dr would be 1s. If you measure S1 to be 0 at t=0 and S2 to be 1 at t=3, then Ds would be 1s despite the real time passing to be 3s. But in either case it doesn't matter how much time passes on the wall.

You shouldn't change the underlying clock implementation while sampling it for use in a timer, that's true, but I still don't see why a duration needs to have an associated clock type.

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -0,0 +1,110 @@
...
+  bool
+  operator>(const rclcpp::Duration & rhs) const;
+
+  RCLCPP_PUBLIC
+  Duration
+  operator+(const rclcpp::Duration & rhs) const;
+
+  RCLCPP_PUBLIC
+  Duration
+  operator-(const rclcpp::Duration & rhs) const;
```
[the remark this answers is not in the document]
tfoote: Good point added free functions for the operators in the other order.

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -0,0 +1,161 @@
...
+bool
+JumpThreshold::is_exceeded(const TimeJump & jump)
+{
+  if (on_clock_change_ &&
+    (jump.jump_type_ == TimeJump::ClockChange_t::ROS_TIME_ACTIVATED ||
+    jump.jump_type_ == TimeJump::ClockChange_t::ROS_TIME_DEACTIVATED))
+  {
+    return true;
+  }
+  if ((uint64_t)jump.delta_.nanoseconds > min_forward_ ||
```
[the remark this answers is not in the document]
tfoote: My goals it to move this logic into C in the next step. So I'd like to keep it this way for now.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -0,0 +1,223 @@
...
+  std::shared_ptr<rclcpp::Clock> clock)
+{
+  // TODO(tfoote) Move enabledROS time logic here.
+
+  rclcpp::Time msg_time = rclcpp::Time(*msg);
+  rclcpp::Time now = clock->now();
+  auto diff = now - msg_time;
+  rclcpp::TimeJump jump;
+  jump.delta_.nanoseconds = diff.nanoseconds();
+  // TODO(tfoote) fill in time source change on jump
```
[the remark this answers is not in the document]
tfoote: Yes. Fixed

rclcpp/src/rclcpp/time_source.cpp
tfoote: I moved this to state the required lifetime for the validity of the callback functions compared to the handler.

rclcpp/src/rclcpp/time_source.cpp
tfoote: Most todo's have been addressed. The remaining ones are for future implementations.

rclcpp/src/rclcpp/time_source.cpp
tfoote: I don't think the datastructure necessarily needs to change, but definitely a function to both zip and filter for desired names would avoid a lot of this boilerplate.

tfoote: @wjwwood Is this the sort of thing you were thinking? https://github.com/ros2/rclcpp/commit/b57983c4f84d68d9c48d30974bbf4d3bcc31e623

wjwwood: It's more sophisticated than I was thinking (with the parameter name filtering as well), but yeah that's more or less what I was thinking.

tfoote: I've isolated that and will submit it separately. I still need to add unit tests for it.

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -0,0 +1,110 @@
...
+    rcl_clock_type_t clock_type = RCL_SYSTEM_TIME);
+
+  RCLCPP_PUBLIC
+  Duration(const builtin_interfaces::msg::Duration & duration_msg);  // NOLINT
+
+  RCLCPP_PUBLIC
+  explicit Duration(const rcl_duration_t & duration);
+
+  RCLCPP_PUBLIC
+  Duration(const Duration & rhs);
```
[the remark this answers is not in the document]
tfoote: I added the std::chrono::nanoseconds to the constructor overload

tfoote:
@wjwwood can you take one more look at this one? I believe I've resolved all your comments but want to make sure.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -0,0 +1,132 @@
...
+  JumpThreshold notice_threshold;
+};
+
+class Clock : public std::enable_shared_from_this<Clock>
+{
+public:
+  RCLCPP_SMART_PTR_DEFINITIONS(Clock)
+
+  RCLCPP_PUBLIC
+  explicit Clock(rcl_clock_type_t clock_type = RCL_SYSTEM_TIME);
```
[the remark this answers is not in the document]
wjwwood: What about this?

rclcpp/src/rclcpp/time_source.cpp
wjwwood: Was this fixed?

rclcpp/test/test_time.cpp
```diff
@@ -32,27 +33,44 @@ class TestTime : public ::testing::Test
...
 
-  // default
-  Time default_now = rclcpp::Time::now();
-  EXPECT_NE(0, default_now.sec);
-  EXPECT_NE(0u, default_now.nanosec);
+  // // default
+  // Time default_now = rclcpp::Time::now();
+  // EXPECT_NE(0, default_now.sec);
+  // EXPECT_NE(0u, default_now.nanosec);
 }
```
wjwwood: Delete or add a todo please.

rclcpp/test/test_time_source.cpp
```diff
@@ -0,0 +1,279 @@
...
+  rclcpp::TimeSource ts(node);
+  auto ros_clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
+  EXPECT_FALSE(ros_clock->ros_time_is_active());
+  ts.attachClock(ros_clock);
+  EXPECT_FALSE(ros_clock->ros_time_is_active());
+
+  // builtin_interfaces::msg::Time::SharedPtr last_msg;
+  // auto clock_sub = node->create_subscription<builtin_interfaces::msg::Time>(
+  //   "clock", [&](builtin_interfaces::msg::Time::SharedPtr msg) {last_msg = msg;},
+  //   rmw_qos_profile_default);
```
wjwwood: Delete or add todo.

rclcpp/src/rclcpp/time_source.cpp
tfoote: `rcl_duration_t` is now a simple struct with no init or fini. https://github.com/ros2/rcl/pull/164/files#diff-698cd67a00a630ab8809b42cd25a4179R77

rclcpp/include/rclcpp/clock.hpp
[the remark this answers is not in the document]
tfoote: I removed the enabled_shared_from_this instead of the constructor.

tfoote:
@wjwwood I removed the old test code and the other items I believe were resolved.

mikaelarguedas:
@tfoote FYI: this introduced failing tests on windows debug:
http://ci.ros2.org/job/ci_windows/3648/

PR #372, opened 2017-09-12
restore . as parameter separator

mikaelarguedas:
Follow up of https://github.com/ros2/rclcpp/pull/367#issuecomment-328581361

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3216/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/536/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2572/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3283/)

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -198,7 +198,9 @@ NodeParameters::list_parameters(const std::vector<std::string> & prefixes, uint6
   std::lock_guard<std::mutex> lock(mutex_);
   rcl_interfaces::msg::ListParametersResult result;
 
-  const char separator = '/';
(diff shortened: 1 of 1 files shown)


Karsten1987:
given this back and forth, I would suggest to put that prefix in a public header so that all other packages can depend on it (this holds for all tests as well as the yaml-parameter server for the ros2_control)

tfoote:
@Karsten1987 we'll need it in multi languages too. It's a little overkill but we could even put this sort of thing into message constants and leverage our generators to get it to any target language as a constant.

mikaelarguedas:
I'll merge this as is and ticket the fact that this separator should be exposed to downstream packages one way or another. I like the suggestion of putting it in a message constant in `rcl_interfaces` to store it in a single place and make it available to downstream code regardless of the language

PR #375, opened 2017-09-19
Improvements to rclcpp::Time

wjwwood:
I found these while working on the rviz port.

- Provide a default argument for the single argument constructor, making the class trivially constructible (? not sure if that's the correct term, or maybe trivially default constructible)
  - This is needed to allow it to be used with Qt's signals and slots
- Change the `operator=` methods to return `Time &`
  - https://stackoverflow.com/questions/9072169/why-should-the-assignment-operator-return-a-reference-to-the-object
- Added the `operator!=`

### DIFF
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -64,6 +64,10 @@ public:
 
   RCLCPP_PUBLIC
   bool
+  operator!=(const rclcpp::Time & rhs) const;
+
+  RCLCPP_PUBLIC
+  bool
   operator<(const rclcpp::Time & rhs) const;
 
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -36,7 +36,7 @@ public:
   Time(int32_t seconds, uint32_t nanoseconds, rcl_time_source_type_t clock = RCL_SYSTEM_TIME);
 
   RCLCPP_PUBLIC
-  explicit Time(uint64_t nanoseconds, rcl_time_source_type_t clock = RCL_SYSTEM_TIME);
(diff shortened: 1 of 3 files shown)


wjwwood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3259/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/549/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2585/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3305/)

rclcpp/include/rclcpp/time.hpp
```diff
@@ -36,7 +36,7 @@ class Time
   Time(int32_t seconds, uint32_t nanoseconds, rcl_time_source_type_t clock = RCL_SYSTEM_TIME);
 
   RCLCPP_PUBLIC
-  explicit Time(uint64_t nanoseconds, rcl_time_source_type_t clock = RCL_SYSTEM_TIME);
+  explicit Time(uint64_t nanoseconds = 0, rcl_time_source_type_t clock = RCL_SYSTEM_TIME);
```
Karsten1987: if this constructor has a default value, shouldn't the upper constructor `Time(int32_t seconds, uint32_t nanoseconds)` have default values as well?

dirk-thomas: No, if two constructors can be called without arguments it wouldn't be ambiguous which one the user wants to call.

wjwwood:
The CI failures on Windows are related to timing (not the Time object AFAICT), so I'll merge this.

rclcpp/include/rclcpp/time.hpp
wjwwood: @Karsten1987 I don't think any number of the other constructor arguments should be optional. That is just a convenience constructor anyways, which I believe calls the nanosecond version of the constructor. Also, as @dirk-thomas said, we can't have two constructors that take no arguments.

Issue #378, opened 2017-09-26
OSX Static assertion failure from test_intra_process_manager.cpp:18


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - mini2 maxOS Sierra
- Installation type:
  - Source - CI job
- Version or commit hash:
  - 5813ba54db8e975e1f3f256ba34614ed62356ec2
- DDS implementation:
  - fastRTPS, connext
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Run a CI job, this is the command that failed
```
. /Users/osrf/jenkins/workspace/ci_osx/ws/build/rclcpp/cmake__build.sh && /usr/bin/make -j4 -l4
```

#### Expected behavior
rclcpp tests compile

#### Actual behavior
Failure in static assertion

```
In file included from /Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/test/test_intra_process_manager.cpp:18:
In file included from /Users/osrf/jenkins/workspace/ci_osx/ws/install/gtest_vendor/src/gtest_vendor/include/gtest/gtest.h:58:
In file included from /Users/osrf/jenkins/workspace/ci_osx/ws/install/gtest_vendor/src/gtest_vendor/include/gtest/internal/gtest-internal.h:58:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/map:820:5: error: static_assert failed "Allocator::value_type must be same type as value_type"
    static_assert((is_same<typename allocator_type::value_type, value_type>::value),
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/osrf/jenkins/workspace/ci_osx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/intra_process_manager_impl.hpp:276:14: note: in instantiation of template class 'std::__1::map<const char *, std::__1::set<unsigned long long, std::__1::less<unsigned long long>, std::__1::allocator<unsigned long long> >, rclcpp::intra_process_manager::IntraProcessManagerImpl<std::__1::allocator<void> >::strcmp_wrapper, std::__1::allocator<std::__1::pair<const std::__1::basic_string<char>, std::__1::set<unsigned long long, std::__1::less<unsigned long long>, std::__1::allocator<unsigned long long> > > > >' requested here
[...]

PR #379, opened 2017-09-27
Fix static assertion on xcode 9

sloretz:
This makes sure an allocator given to an `std::map` has the correct type.

Testing not yet in progress, resolving random CI issues so I have an OSX machine to test on first.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3269/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/554/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2604/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3313/)

Connects to ros2/rclcpp#378

### DIFF
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -266,14 +266,11 @@ private:
     }
   };
 
-  // Make sure allocator given to map has the correct value_type
-  using IDTopic = std::pair<const char *, AllocSet>;
-
   using IDTopicMap = std::map<
-      typename IDTopic::first_type,
-      typename IDTopic::second_type,
+      const char *,
+      AllocSet,
       strcmp_wrapper,
-      RebindAlloc<IDTopic>>;
+      RebindAlloc<std::pair<const char const *, AllocSet>>;
 
   SubscriptionMap subscriptions_;
 
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -269,7 +269,7 @@ private:
       const char *,
       AllocSet,
       strcmp_wrapper,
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/intra_process_manager_impl.hpp
```diff
@@ -265,11 +265,15 @@ class IntraProcessManagerImpl : public IntraProcessManagerImplBase
...
+  using IDTopic = std::pair<const char *, AllocSet>;
+
   using IDTopicMap = std::map<
-      const char *,
-      AllocSet,
+      typename IDTopic::first_type,
+      typename IDTopic::second_type,
       strcmp_wrapper,
-      RebindAlloc<std::pair<const std::string, AllocSet>>>;
+      RebindAlloc<IDTopic>>;
```
wjwwood: Is the use of the alias (i.e. the `using` statement) required or is it only necessary to change either the map declaration to `std::string` or the `RebindAlloc` declaration to `const char *`?

sloretz: I don't think it should be necessary, and this is evidence against this fix working, but I managed to get a CI job to run with that change dbb462b5bcad8ddaeb01468f2cd0fcf07cf53e65 and it encountered the same issue.

wjwwood: Well, either way what you've put here seems fine, though I don't know why `const std::string` or `const char *` is preferable.

sloretz: Looks like the first type in the `std::pair` needs to be type `const Key`, which is `const char * const` (thanks grey). I ran a ci_osx and it successfully built rclcpp. If everything builds using ci_launcher then I'll squash/merge this.

sloretz:
Only test failures due to ros2/build_cop#55. Merging

PR #388, opened 2017-10-23
*_raw function

Karsten1987:
Connects to ros2/demos#185

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -260,6 +260,17 @@ if(BUILD_TESTING)
       ${PROJECT_NAME}
     )
   endif()
+  ament_add_gtest(test_raw_message_allocator test/test_raw_message_allocator.cpp
+    ENV RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation})
+  if(TARGET test_raw_message_allocator)
+    target_include_directories(test_raw_message_allocator PUBLIC
+      ${test_msgs_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_raw_message_allocator
+      ${PROJECT_NAME}
+      ${test_msgs_LIBRARIES}
+    )
+  endif()
   ament_add_gtest(test_service test/test_service.cpp)
   if(TARGET test_service)
     target_include_directories(test_service PUBLIC
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -260,13 +260,13 @@ if(BUILD_TESTING)
       ${PROJECT_NAME}
     )
   endif()
-  ament_add_gtest(test_raw_message_allocator test/test_raw_message_allocator.cpp
+  ament_add_gtest(test_serialized_message_allocator test/test_serialized_message_allocator.cpp
     ENV RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation})
-  if(TARGET test_raw_message_allocator)
-    target_include_directories(test_raw_message_allocator PUBLIC
+  if(TARGET test_serialized_message_allocator)
+    target_include_directories(test_serialized_message_allocator PUBLIC
       ${test_msgs_INCLUDE_DIRS}
     )
-    target_link_libraries(test_raw_message_allocator
+    target_link_libraries(test_serialized_message_allocator
       ${PROJECT_NAME}
       ${test_msgs_LIBRARIES}
     )
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -280,6 +280,13 @@ if(BUILD_TESTING)
     )
     target_link_libraries(test_subscription ${PROJECT_NAME})
   endif()
+  ament_add_gtest(test_subscription_traits test/test_subscription_traits.cpp)
+  if(TARGET test_subscription_traits)
+    target_include_directories(test_subscription_traits PUBLIC
+      ${rcl_interfaces_INCLUDE_DIRS}
+      ${rcl_INCLUDE_DIRS}
+    )
+  endif()
   ament_add_gtest(test_find_weak_nodes test/test_find_weak_nodes.cpp)
   if(TARGET test_find_weak_nodes)
     target_include_directories(test_find_weak_nodes PUBLIC
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -280,12 +280,15 @@ if(BUILD_TESTING)
     )
     target_link_libraries(test_subscription ${PROJECT_NAME})
   endif()
+  find_package(test_msgs REQUIRED)
   ament_add_gtest(test_subscription_traits test/test_subscription_traits.cpp)
   if(TARGET test_subscription_traits)
     target_include_directories(test_subscription_traits PUBLIC
-      ${rcl_interfaces_INCLUDE_DIRS}
       ${rcl_INCLUDE_DIRS}
     )
+  ament_target_dependencies(test_subscription_traits
+    "test_msgs"
+  )
   endif()
   ament_add_gtest(test_find_weak_nodes test/test_find_weak_nodes.cpp)
   if(TARGET test_find_weak_nodes)
--- a/rclcpp/include/rclcpp/message_memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/message_memory_strategy.hpp
@@ -86,7 +86,7 @@ public:
     raw_msg->buffer_length = 0;
     raw_msg->buffer_capacity = 0;
     raw_msg->buffer = NULL;
-    rmw_initialize_raw_message(raw_msg.get(), 0, &rcutils_allocator_);
+    rmw_raw_message_init(raw_msg.get(), 0, &rcutils_allocator_);
     rmw_raw_message_resize(raw_msg.get(), capacity);
     return raw_msg;
   }
--- a/rclcpp/include/rclcpp/message_memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/message_memory_strategy.hpp
@@ -18,9 +18,13 @@
 #include <memory>
 #include <stdexcept>
 
+#include "rcl/types.h"
+
 #include "rclcpp/allocator/allocator_common.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/visibility_control.hpp"
+
+#include "rmw/raw_message.h"
 
 namespace rclcpp
 {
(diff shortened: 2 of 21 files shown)


rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -114,21 +121,26 @@ Node::create_subscription(
...
+  typename MessageT,
+  typename CallbackT,
+  typename Alloc,
+  typename SubscriptionT>
 std::shared_ptr<SubscriptionT>
 Node::create_subscription(
   const std::string & topic_name,
-  size_t qos_history_depth,
   CallbackT && callback,
+  size_t qos_history_depth,
```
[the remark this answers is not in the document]
Karsten1987: i can't really explain why, but with the modifications done in `subscription_traits`, I get template deduction errors without this change. Somehow it can't really differentiate between the two function overloads. With the change done here, it works though.
I am happy to discuss/debug that further and try to resolve this. I just don't know how :)

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -281,22 +281,42 @@ void
...
 {
-  std::shared_ptr<void> message = subscription->create_message();
   rmw_message_info_t message_info;
-
-  auto ret = rcl_take(subscription->get_subscription_handle().get(),
+  // TODO(karsten1987): Can this be set to false directly after initialization?
+  message_info.from_intra_process = false;
+
+  if (subscription->is_raw()) {
+    // TODO(karsten1987): Raise if opensplice !?
```
Karsten1987: Right now, this function contains a lot of code duplication, given that big `if - else`.

Ideally, we also call `take_raw` to get the still serialized data from the wire and only convert it to a ros message if it's a raw callback. 
The reason I didn't do it this way is that Opensplice doesn't support to get the serialized stream yet and thus would stay incompatible.

rclcpp/CMakeLists.txt
```diff
@@ -234,6 +234,17 @@ if(BUILD_TESTING)
       ${PROJECT_NAME}
     )
   endif()
+  ament_add_gtest(test_raw_message_allocator test/test_raw_message_allocator.cpp
+    ENV RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation})
+  if(TARGET test_raw_message_allocator)
```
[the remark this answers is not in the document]
Karsten1987: fair point. I simply adhered to the rest of the file. We can change that in a follow up PR

rclcpp/include/rclcpp/message_memory_strategy.hpp
```diff
@@ -62,15 +80,52 @@ class MessageMemoryStrategy
     return std::allocate_shared<MessageT, MessageAlloc>(*message_allocator_.get());
   }
 
+  virtual std::shared_ptr<rcl_message_raw_t> borrow_raw_message(unsigned int capacity)
```
wjwwood: `size_t` (obviously depends on changing that down the API chain too, but I think it should be)

rclcpp/include/rclcpp/message_memory_strategy.hpp
```diff
@@ -62,15 +80,52 @@ class MessageMemoryStrategy
...
+  {
+    auto raw_msg = std::shared_ptr<rcl_message_raw_t>(new rcl_message_raw_t,
+        [](rmw_message_raw_t * msg) {
+          rmw_raw_message_fini(msg);
+          delete msg;
+        });
+    *raw_msg = rmw_get_zero_initialized_raw_message();
+    rmw_raw_message_init(raw_msg.get(), capacity, &rcutils_allocator_);
```
wjwwood: Check the return code, and do not setup the destructor for the `shared_ptr` until you know this succeeds. Make sure to throw based on the rcl ret code.

rclcpp/include/rclcpp/message_memory_strategy.hpp
```diff
@@ -62,15 +80,52 @@ class MessageMemoryStrategy
     return std::allocate_shared<MessageT, MessageAlloc>(*message_allocator_.get());
   }
 
+  virtual std::shared_ptr<rcl_message_raw_t> borrow_raw_message(unsigned int capacity)
+  {
+    auto raw_msg = std::shared_ptr<rcl_message_raw_t>(new rcl_message_raw_t,
+        [](rmw_message_raw_t * msg) {
+          rmw_raw_message_fini(msg);
```
wjwwood: Check return code.

rclcpp/include/rclcpp/message_memory_strategy.hpp
```diff
@@ -62,15 +80,52 @@ class MessageMemoryStrategy
...
   /// Release ownership of the message, which will deallocate it if it has no more owners.
   /** \param[in] msg Shared pointer to the message we are returning. */
   virtual void return_message(std::shared_ptr<MessageT> & msg)
   {
     msg.reset();
   }
 
+  virtual void return_raw_message(std::shared_ptr<rcl_message_raw_t> & raw_msg)
+  {
+    rmw_raw_message_fini(raw_msg.get());
```
wjwwood: Isn't this already being done in the custom destructor?

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -279,6 +276,19 @@ class Publisher : public PublisherBase
     return this->publish(*msg);
   }
 
+  void
+  publish(const rcl_message_raw_t * raw_msg)
+  {
+    if (store_intra_process_message_) {
+      // not supported atm
+      throw std::runtime_error("storing raw messages in intra process is not supported yet.");
```
wjwwood: Make sure to make an issue about this after merging, and consider putting a TODO in here.

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -104,31 +112,47 @@ class SubscriptionBase
...
+  virtual void
+  return_raw_message(std::shared_ptr<rcl_message_raw_t> & message) = 0;
+
   virtual void
   handle_intra_process_message(
     rcl_interfaces::msg::IntraProcessMessage & ipm,
     const rmw_message_info_t & message_info) = 0;
 
+  rosidl_message_type_support_t
+  get_message_type_support_handle() const;
```
wjwwood: Should this return a const reference instead? Not sure it matters, but worth considering.

rclcpp/include/rclcpp/subscription_traits.hpp
```diff
@@ -0,0 +1,79 @@
...
+/*
+ * The current version of uncrustify has a misinterpretion here
+ * between `:` used for inheritance vs for initializer list
+ * The result is that whenever a templated struct is used,
+ * the colon has to be without any whitespace next to it whereas
+ * when no template is used, the colon has to be separated by a space.
+ * Cheers!
+ */
+template<typename T>
+struct is_raw_subscription_argument : std::false_type
```
wjwwood: Maybe if you used the `public` keyword before the inheritance uncrustify would do the right thing?

```c++
template<typename T>
struct is_raw_subscription_argument: public std::false_type
{};
```

rclcpp/src/rclcpp/executor.cpp
wjwwood: I think until we make `take_raw` required it has to be this way.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -74,7 +74,8 @@ void TimeSource::attachNode(
 
   auto cb = std::bind(&TimeSource::clock_cb, this, std::placeholders::_1);
 
-  clock_subscription_ = rclcpp::create_subscription<MessageT, decltype(cb), Alloc, SubscriptionT>(
+  clock_subscription_ = rclcpp::create_subscription<
+    MessageT, decltype(cb), Alloc, MessageT, SubscriptionT>(
```
wjwwood: This is bad style in my opinion because the template arguments line up with the function arguments. Instead do:

```c++
  clock_subscription_ = rclcpp::create_subscription<
    MessageT, decltype(cb), Alloc, MessageT, SubscriptionT
  >(
    node_topics_.get(),
    topic_name,
    std::move(cb),
    ...
  );
```

Or:

```c++
  clock_subscription_ =
    rclcpp::create_subscription<MessageT, decltype(cb), Alloc, MessageT, SubscriptionT>(
      node_topics_.get(),
      topic_name,
      std::move(cb),
      ...
    );
```

rclcpp/include/rclcpp/message_memory_strategy.hpp
Karsten1987: I chose `unsigned int` because RTI only serializes this. 
https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/api/connext_dds/api_cpp/structDDS__DynamicData.html#ae31195afe3fe6f567ddce38e7f99980e

I can change to `size_t` and cast it to unsigned for connext, but this might lead to data loss.

wjwwood: `size_t` is unsigned, and usually larger than `unsigned int`. Do you mean the user could ask for something larger and that would be an issue? You could check that it isn't too big at the point where you pass the `size_t` to RTI's API.

rclcpp/include/rclcpp/subscription_traits.hpp
Karsten1987: Unfortunately, it still complains:
```
31: +++ include/rclcpp/subscription_traits.hpp.uncrustify
31: @@ -40 +40 @@
31: -struct is_raw_subscription_argument<rcl_message_raw_t> : public std::true_type
31: +struct is_raw_subscription_argument<rcl_message_raw_t>: public std::true_type
```

Good idea though!

rclcpp/src/rclcpp/time_source.cpp
Karsten1987: Both suggest versions don't work, but I'll go for this one then to separate the template arguments from the function arguments:
```
  clock_subscription_ = rclcpp::create_subscription<
    MessageT, decltype(cb), Alloc, MessageT, SubscriptionT
    >(
    node_topics_.get(),
    topic_name,
    std::move(cb),
    rmw_qos_profile_default,
    group,
    false,
    false,
    msg_mem_strat,
    allocator);

```

wjwwood:
I'm going to take a crack at changing the `unsigned int` to a `size_t`, so that's what's blocking my approval atm.

dirk-thomas:
The nightly packaging jobs on Linux failed in the `tlsf_cpp` package due to a signature change in this PR.

Karsten1987:
I forgot to link the `realtime_tools` PR into this one. That's why the build farm didn't complain.
please see:
https://github.com/ros2/realtime_support/pull/64

Issue #393, opened 2017-11-09
wrapper object use after free


gaoethan:
## Bug report

**Required Info:**

- Operating System: 
  -  N/A
- Installation type: 
  -  N/A
- Version or commit hash:
  -  master
- DDS implementation: 
  -  N/A
- Client library (if applicable): 
  -  rclcpp

#### Steps to reproduce issue

It occurs the issue of wrapper object use after free with `c_str()` in the code snippet like the following in [transition.cpp](https://github.com/ros2/rclcpp/blob/master/rclcpp_lifecycle/src/transition.cpp#L45-L55):
  ```
auto start_state = new rcl_lifecycle_state_t;
  start_state->id = start.id();
  start_state->label = start.label().c_str();

  auto goal_state = new rcl_lifecycle_state_t;
  goal_state->id = goal.id();
  goal_state->label = start.label().c_str();

  transition_handle->start = start_state;
  transition_handle->goal = goal_state;
```
because The internal representation of temporary of type `std::__cxx11::string` escapes into `start_state->label`, but is destroyed when it exits scope, I intend to draft a PR to optimize it, for example, using s'th like strcpy() to keep it firstly and then give that to `start_state->label`, but it involves some other change and maybe it's better check more comments before getting down to modify it. any comments ? thanks !

#### Expected behavior
the access to content should be at valid address
#### Actual behavior
use after free
#### Additional information

----

Karsten1987:
Good catch.
It might be worth checking into returning a `const char *` for the `label()` function if you intend on going with `strcpy()`

I'd propose you file a PR for this ticket so we can run CI on it.

Karsten1987:
closed via #418

PR #395, opened 2017-11-10
copy constructors & assignment operators

### LINKED ISSUE #394: Missing copy constructor and assignment operator
## Bug report

**Required Info:**

- Operating System:
  -  N/A
- Installation type:
  -  source
- Version or commit hash:
  -  master
- DDS implementation:
  -  N/A
  -  rclcpp

#### Steps to reproduce issue

In the following design of [state.cpp](https://github.com/ros2/rclcpp/blob/master/rclcpp_lifecycle/src/state.cpp#L47-L58):
```
State::State(const rcl_lifecycle_state_t * rcl_lifecycle_state_handle)
: owns_rcl_state_handle_(false)
{
  state_handle_ = rcl_lifecycle_state_handle;
}

State::~State()
{
  if (owns_rcl_state_handle_) {
    delete state_handle_;
  }
}

```
Class `rclcpp_lifecycle::State` owns resources that are freed in its destructor but has no user-written copy constructor or assignment operator. The destructor frees member state_handle_ when `owns_rcl_state_handle_` is true. but the class is copy-constructed and copy-assigned the the following code snippet, for example:

[lifecycle_node_interface_impl.hpp #261](https://github.com/ros2/rclcpp/blob/master/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp#L261)
[lifecycle_node_interface_impl.hpp #356](https://github.com/ros2/rclcpp/blob/master/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp#L356)

That's to say this class, which frees resources in its destructor, does not have a user-written copy constructor, but is copy-constructed. Unless the resource is managed separately from the copy operation, this may cause use-after-free errors. In addition, the same case to the design in [transition.cpp](https://github.com/ros2/rclcpp/blob/master/rclcpp_lifecycle/src/transition.cpp#L59-L76)

We know that their `owns_rcl_state_handle_` and `owns_rcl_transition_handle_` defaults to false, so..here to confirm that these were designed like this on purpose? and will it be prone to introduce new issues in the future while using ? thanks !

Karsten1987:
fixes ros2/rclcpp#393
fixes #394

@gaoethan fyi

CI:

WinD: [Build Status](http://ci.ros2.org/job/ci_windows/3553/)
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3504/)
OSX:

### DIFF
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -70,10 +70,6 @@ protected:
   RCLCPP_LIFECYCLE_PUBLIC
   void
   reset();
-
-  RCLCPP_LIFECYCLE_PUBLIC
-  void
-  assign(const State & rhs);
 };
 
 }  // namespace rclcpp_lifecycle
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -16,11 +16,11 @@
 #define RCLCPP_LIFECYCLE__STATE_HPP_
 
 #include <string>
-#include <utility>
 
 #include "rclcpp_lifecycle/visibility_control.h"
 
-// forward declare rcl_state_t
+#include "rcutils/allocator.h"
+
 typedef struct rcl_lifecycle_state_t rcl_lifecycle_state_t;
 
 namespace rclcpp_lifecycle
@@ -30,13 +30,18 @@ class State
 {
 public:
   RCLCPP_LIFECYCLE_PUBLIC
-  State();
+  explicit State(rcutils_allocator_t allocator = rcutils_get_default_allocator());
 
   RCLCPP_LIFECYCLE_PUBLIC
-  State(uint8_t id, const std::string & label);
+  State(
+    uint8_t id,
+    const std::string & label,
+    rcutils_allocator_t allocator = rcutils_get_default_allocator());
 
   RCLCPP_LIFECYCLE_PUBLIC
-  explicit State(const rcl_lifecycle_state_t * rcl_lifecycle_state_handle);
+  explicit State(
+    const rcl_lifecycle_state_t * rcl_lifecycle_state_handle,
+    rcutils_allocator_t allocator = rcutils_get_default_allocator());
 
   RCLCPP_LIFECYCLE_PUBLIC
   State(const State & rhs);
@@ -57,10 +62,13 @@ public:
 
   RCLCPP_LIFECYCLE_PUBLIC
   friend void
-  swap(State & lhs, const State & rhs);
+  copy_from(State & lhs, const State & rhs);
 
 protected:
+  rcutils_allocator_t allocator_;
+
   bool owns_rcl_state_handle_;
+
   const rcl_lifecycle_state_t * state_handle_;
 };
 
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -16,6 +16,7 @@
 #define RCLCPP_LIFECYCLE__STATE_HPP_
 
 #include <string>
+#include <utility>
 
 #include "rclcpp_lifecycle/visibility_control.h"
 
@@ -53,6 +54,10 @@ public:
   RCLCPP_LIFECYCLE_PUBLIC
   std::string
   label() const;
+
+  RCLCPP_LIFECYCLE_PUBLIC
+  friend void
+  swap(State & lhs, const State & rhs);
 
 protected:
   bool owns_rcl_state_handle_;
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -68,7 +68,11 @@ protected:
   const rcl_lifecycle_state_t * state_handle_;
 
   RCLCPP_LIFECYCLE_PUBLIC
-  void
+  friend void
+  free_resources(State & instance);
+
+  RCLCPP_LIFECYCLE_PUBLIC
+  friend void
   copy_from(State & lhs, const State & rhs);
 };
 
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -38,7 +38,13 @@ public:
   explicit State(const rcl_lifecycle_state_t * rcl_lifecycle_state_handle);
 
   RCLCPP_LIFECYCLE_PUBLIC
+  State(const State & rhs);
+
+  RCLCPP_LIFECYCLE_PUBLIC
   virtual ~State();
+
+  RCLCPP_LIFECYCLE_PUBLIC
+  State & operator=(const State & rhs);
 
   RCLCPP_LIFECYCLE_PUBLIC
   uint8_t
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -60,16 +60,16 @@ public:
   std::string
   label() const;
 
-  RCLCPP_LIFECYCLE_PUBLIC
-  friend void
-  copy_from(State & lhs, const State & rhs);
-
 protected:
   rcutils_allocator_t allocator_;
 
   bool owns_rcl_state_handle_;
 
   const rcl_lifecycle_state_t * state_handle_;
+
+  RCLCPP_LIFECYCLE_PUBLIC
+  void
+  copy_from(State & lhs, const State & rhs);
 };
 
 }  // namespace rclcpp_lifecycle
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -65,15 +65,15 @@ protected:
 
   bool owns_rcl_state_handle_;
 
-  const rcl_lifecycle_state_t * state_handle_;
+  rcl_lifecycle_state_t * state_handle_;
 
   RCLCPP_LIFECYCLE_PUBLIC
-  friend void
-  free_resources(State & instance);
+  void
+  reset();
 
   RCLCPP_LIFECYCLE_PUBLIC
-  friend void
-  copy_from(State & lhs, const State & rhs);
+  void
+  assign(const State & rhs);
 };
 
 }  // namespace rclcpp_lifecycle
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -35,7 +35,8 @@ State::State(
   const std::string & label,
   rcutils_allocator_t allocator)
 : allocator_(allocator),
-  owns_rcl_state_handle_(true)
+  owns_rcl_state_handle_(true),
+  state_handle_(nullptr)
 {
   if (label.empty()) {
     throw std::runtime_error("Lifecycle State cannot have an empty label.");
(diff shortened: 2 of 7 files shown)


rclcpp_lifecycle/src/state.cpp
```diff
@@ -50,13 +51,48 @@ State::State(const rcl_lifecycle_state_t * rcl_lifecycle_state_handle)
...
+    auto state_handle = new rcl_lifecycle_state_t;
+    state_handle->id = rhs.state_handle_->id;
+    state_handle->label = rcutils_strndup(
+      rhs.state_handle_->label,
+      strlen(rhs.state_handle_->label),
+      rcutils_get_default_allocator());
+    state_handle_ = state_handle;
+  } else {
+    state_handle_ = rhs.state_handle_;
+  }
```
dirk-thomas: I would suggest to deduplicate the logic.

Same below.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -14,17 +14,38 @@
...
 
 namespace rclcpp_lifecycle
 {
 
+void
+swap(rcl_lifecycle_state_t * lhs, const rcl_lifecycle_state_t * rhs)
+{
+  lhs->id = rhs->id;
+  lhs->label = rcutils_strndup(rhs->label, strlen(rhs->label), rcutils_get_default_allocator());
+}
```
dirk-thomas: The semantic of this function is not [swap](http://en.cppreference.com/w/cpp/algorithm/swap). Therefore it should have a different name.

Same below.

Karsten1987: The name goes along with the copy-and-swap idiom.
https://stackoverflow.com/questions/3279543/what-is-the-copy-and-swap-idiom#3279550

dirk-thomas: My point is that if I call `swap(a, b)` I expect those two instances to be "swapped" (a has the values form b, b has the values from a). The current code doesn't accomplish that: https://github.com/ros2/rclcpp/pull/395/files#diff-7731e343ff998d96d3dd4fcba8c97c69R45

gaoethan:
@Karsten1987 sorry, I'm out for several day and thanks for taking this 2 issues :+1:

rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
```diff
@@ -29,27 +30,45 @@ class State
...
   uint8_t
   id() const;
 
   RCLCPP_LIFECYCLE_PUBLIC
   std::string
   label() const;
 
+  RCLCPP_LIFECYCLE_PUBLIC
+  friend void
+  copy_from(State & lhs, const State & rhs);
```
dirk-thomas: This function should either be static (since it doesn't operate on `this`) or be private, operate on this and omit `lhs` instead.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -14,49 +14,99 @@
...
+    auto state_handle = reinterpret_cast<rcl_lifecycle_state_t *>(
+      rhs.allocator_.allocate(sizeof(rcl_lifecycle_state_t), rhs.allocator_.state));
+    state_handle->id = rhs.state_handle_->id;
+    state_handle->label = rcutils_strndup(
+      rhs.state_handle_->label,
+      strlen(rhs.state_handle_->label),
+      rhs.allocator_);
+    lhs.state_handle_ = state_handle;
+  } else {
+    lhs.state_handle_ = rhs.state_handle_;
```
dirk-thomas: A previously allocated handle on `lhs` is being lost here.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -14,49 +14,99 @@
...
+  rcutils_allocator_t allocator)
+: allocator_(allocator),
+  owns_rcl_state_handle_(false)
 {
   state_handle_ = rcl_lifecycle_state_handle;
 }
 
+State::State(const State & rhs)
+: allocator_(rhs.allocator_),
+  owns_rcl_state_handle_(rhs.owns_rcl_state_handle_)
```
dirk-thomas: `state_handle_` is not being initialized here. Therefore the function called later can't safely determine if the handle was valid and might need to be deallocate before being overwritten.

rclcpp_lifecycle/src/state.cpp
Karsten1987: If we don't own the `state_handle_`, we can't simply deallocate the resource.

dirk-thomas: But if we do we have to deallocate the resource before overwriting it.

rclcpp_lifecycle/src/state.cpp
Karsten1987: i don't understand this. The `state_handle_` is getting initialize in the `copy_from` function. What is being overwritten inside the copy constructor?

dirk-thomas: The `copy_from` method is public and can be called for any instance. It must cleanup previously allocated memory as mentioned in the other comment. And that will require the member variables to be initialized here *before* calling the function.

Karsten1987: yeah, changed to a protected member function now.

rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
```diff
@@ -49,8 +61,19 @@ class State
...
+
   const rcl_lifecycle_state_t * state_handle_;
+
+  RCLCPP_LIFECYCLE_PUBLIC
+  friend void
+  free_resources(State & instance);
+
+  RCLCPP_LIFECYCLE_PUBLIC
+  friend void
+  copy_from(State & lhs, const State & rhs);
```
dirk-thomas: Both should still either not be member functions or operate on `this` and not receive `instance` / `lhs`. `copy_from` sounds a lot like `assign` to me. And `free_resources` like `reset`.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -75,4 +102,48 @@ State::label() const
...
+{
+  // nothing to free here
+  if (!instance.state_handle_) {
+    return;
+  }
+  if (!instance.owns_rcl_state_handle_) {
+    return;
+  }
+
+  if (instance.state_handle_) {
```
dirk-thomas: Redundant condition with line 109.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -75,4 +102,48 @@ State::label() const
...
+    if (instance.state_handle_->label) {
+      instance.allocator_.deallocate(
+        const_cast<char *>(instance.state_handle_->label),
+        instance.allocator_.state);
+    }
+    instance.allocator_.deallocate(
+      const_cast<rcl_lifecycle_state_t *>(instance.state_handle_),
+      instance.allocator_.state);
```
dirk-thomas: This should set `state_handle_` to `nullptr`. Even in the cases of the early return.

Maybe `owns_rcl_state_handle_` should be reset too.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -75,4 +102,48 @@ State::label() const
...
+    lhs.owns_rcl_state_handle_ = rhs.owns_rcl_state_handle_;
+
+    auto state_handle = reinterpret_cast<rcl_lifecycle_state_t *>(
+      lhs.allocator_.allocate(sizeof(rcl_lifecycle_state_t), lhs.allocator_.state));
+    state_handle->id = rhs.state_handle_->id;
+    state_handle->label = rcutils_strndup(
+      rhs.state_handle_->label,
+      strlen(rhs.state_handle_->label),
+      lhs.allocator_);
+    lhs.state_handle_ = state_handle;
```
dirk-thomas: Isn't `lhs` now owning the memory of `state_handle_`? But `lhs.owns_rcl_state_handle_` might be whatever `rhs.owns_rcl_state_handle_` was which could be `false.

dirk-thomas:
This looks much better now. I would suggest to move the implementation of `assign` into the `operator=`. That gets rid of the custom method and the copy constructor can just call the assignment operator. That is much more common pattern.

Karsten1987:
CI:
linux: [Build Status](http://ci.ros2.org/job/ci_linux/3565/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/2898/)
win: [Build Status](http://ci.ros2.org/job/ci_windows/3658/) (unrelated)

rclcpp_lifecycle/src/state.cpp
```diff
@@ -14,47 +14,90 @@
...
+  owns_rcl_state_handle_(true),
+  state_handle_(nullptr)
 {
   if (label.empty()) {
     throw std::runtime_error("Lifecycle State cannot have an empty label.");
   }
 
-  auto state_handle = new rcl_lifecycle_state_t;
+  auto state_handle = reinterpret_cast<rcl_lifecycle_state_t *>(
+    allocator_.allocate(sizeof(rcl_lifecycle_state_t), allocator_.state));
```
dirk-thomas: The return value of `allocate` needs to be checked for `NULL`.

Same below.

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -36,43 +47,89 @@ Transition::Transition(uint8_t id, const std::string & label)
...
-    if (transition_handle_->start) {
-      delete transition_handle_->start;
+  reset();
+}
+
+Transition &
+Transition::operator=(const Transition & rhs)
+{
+  if (this != &rhs) {
+    if (owns_rcl_transition_handle_) {
```
dirk-thomas: Why is this condition being used? The `State` class doesn't have that logic. I would suggest to keep the logic here the same as for the `State` class.

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -99,4 +156,47 @@ Transition::goal_state() const
...
+  if (transition_handle_->goal) {
+    if (transition_handle_->goal->label) {
+      allocator_.deallocate(
+        const_cast<char *>(transition_handle_->goal->label), allocator_.state);
+      transition_handle_->goal->label = nullptr;
+    }
+    allocator_.deallocate(transition_handle_->goal, allocator_.state);
+    transition_handle_->goal = nullptr;
+  }
+  if (transition_handle_) {
```
dirk-thomas: This condition should always be true when this line is being reached. I would suggest to keep the logic the same as for the `State` class which seems to be correct / clean.

rclcpp_lifecycle/test/test_state_wrapper.cpp
```diff
@@ -37,15 +37,15 @@ TEST_F(TestStateWrapper, wrapper) {
   }
 
   {
-    rcl_lifecycle_state_t lc_state = {"my_c_state", 2, NULL, NULL, 0};
+    rcl_lifecycle_state_t lc_state = {const_cast<char *>("my_c_state"), 2, NULL, NULL, 0};
```
dirk-thomas: Why is this `const_cast` necessary?

Same below.

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -99,4 +156,47 @@ Transition::goal_state() const
...
+  // can't free anything which is not owned
+  if (!owns_rcl_transition_handle_) {
+    transition_handle_ = nullptr;
+    return;
+  }
+
+  if (transition_handle_->start) {
+    if (transition_handle_->start->label) {
+      allocator_.deallocate(
+        const_cast<char *>(transition_handle_->start->label), allocator_.state);
```
dirk-thomas: The need for a `const_cast` here indicates that the field `transition_handle_->start` / `transition_handle_->start->label` should probably not be declared as `const`. But changing that might be out of scope of this PR.

rclcpp_lifecycle/test/test_state_wrapper.cpp
Karsten1987: C++11 raises a warning here, that a const char literal is assigned to a non-const.

https://stackoverflow.com/questions/20944784/why-is-conversion-from-string-constant-to-char-valid-in-c-but-invalid-in-c
Having it cast to a non-const seems to be a valid solution. Not sure if it's the most elegant one though.

wjwwood: I think the question is, why is the field a `char *` rather than a `const char *`?

Karsten1987: how do I deallocate the field later on if it's const?
https://github.com/ros2/rclcpp/pull/395#discussion_r154537053

rclcpp_lifecycle/src/state.cpp
Karsten1987: addressed in commit https://github.com/ros2/rclcpp/pull/395/commits/dba6601f33c30834d1c9a05f1d577696723ccfad

rclcpp_lifecycle/test/test_state_wrapper.cpp
wjwwood: How do you know whether or not to deallocate it later (cannot deallocate a literal string...)?

Karsten1987:
closing in favor of https://github.com/ros2/rclcpp/pull/419

PR #396, opened 2017-11-13
Reset the error when rcl_take_response failed

minggangw:
When the return value of rcl_take_response() is not equal to RCL_RET_OK,
we need reset the error to avoid being overwritten.


dhood:
Thank you for the patch @minggangw 

`rcl_take_response` returning `RCL_RET_CLIENT_TAKE_FAILED` is not considered an error state, which is why we don't print the error in this case.

As mentioned in https://github.com/ros2/rclcpp/issues/397#issuecomment-344023614, the error message should actually not have been set by rcl in this situation, and then the clearing of the message in the client libraries will not be necessary. I'll close this PR for that reason.

PR #404, opened 2017-11-21
detach nodes from executors in destruction.

tfoote:
I ran into this exception when writing unit tests for time_source. The nodes should be cleared from being associated with the executor when the executor is destructing. Otherwise the node can never be reused.

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3581/)
Linux aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/777/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/2921/)
WIndow:  [Build Status](http://ci.ros2.org/job/ci_windows/3674/)

### DIFF
--- a/rclcpp/test/test_executor.cpp
+++ b/rclcpp/test/test_executor.cpp
@@ -1,0 +1,62 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <chrono>
+#include <limits>
+#include <memory>
+#include <string>
+
+#include "rcl/error_handling.h"
+#include "rcl/time.h"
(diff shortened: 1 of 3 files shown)


rclcpp/test/test_executor.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <chrono>
+#include <limits>
+#include <memory>
+#include <string>
+
+#include "rcl/error_handling.h"
+#include "rcl/time.h"
+#include "rclcpp/clock.hpp"
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/duration.hpp"
```
wjwwood: nitpick: alphabetical order

rclcpp/test/test_executor.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+
```
wjwwood: nitpick: minimize vertical whitespace

dhood:
I spent some time thinking about this so figured I'd write my notes for future reference (or in case it sparks some conversation):

I was considering if we should call to [remove_node](https://github.com/ros2/rclcpp/blob/070b3125c1a7f76fbde3e31973d89a052a2ea652/rclcpp/src/rclcpp/executor.cpp#L128) for symmetry, but I concluded that it's not worthwhile. `remove_node` sets `associated_with_executor_` false as in this PR but additionally (1) notifies the executor (by default) and (2) removes the node's guard condition from the memory strategy. For (1), I think we don't need to notify the executor, because it won't be blocked in another thread since (if I understand correctly), we can't be spinning at this point. For (2), while we need to [remove/release the sigint guard condition](https://github.com/ros2/rclcpp/pull/226/files#diff-97193a162153796658f8e485ff3f9d5aR81) because that gets allocated for the executor's waitset, we don't need to do so for the node's notify guard condition on executor destruction because it's been allocated by the node.

So just setting the flag is all we need to do; this lgtm

PR #405, opened 2017-11-21
Fix the dereference to NULL

gaoethan:
`rmw_*_validation_result_string(validation_result)` may return NULL,
and it's dereferenced by passing arg to `NameValidationError`, this 
leads unexpected behaviors and NULL is deprecated to use in C++ now.

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

Connects to ros2/rmw#130

### DIFF
--- a/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
+++ b/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
@@ -136,7 +136,9 @@ rclcpp::expand_topic_or_service_name(
 
       const char * validation_result_string = rmw_node_name_validation_result_string(
         validation_result);
-      validation_result_string = validation_result_string ? validation_result_string : "";
+      if (!validation_result_string) {
+        validation_result_string = "undefined node name type";
+      }
       throw rclcpp::exceptions::InvalidNodeNameError(
               node_name.c_str(),
               validation_result_string,
@@ -161,7 +163,9 @@ rclcpp::expand_topic_or_service_name(
 
       const char * validation_result_string = rmw_namespace_validation_result_string(
         validation_result);
-      validation_result_string = validation_result_string ? validation_result_string : "";
+      if (!validation_result_string) {
+        validation_result_string = "undefined namespace type";
+      }
       throw rclcpp::exceptions::InvalidNamespaceError(
               namespace_.c_str(),
               validation_result_string,
@@ -190,7 +194,10 @@ rclcpp::expand_topic_or_service_name(
   if (validation_result != RMW_TOPIC_VALID) {
     const char * validation_result_string = rmw_full_topic_name_validation_result_string(
       validation_result);
-    validation_result_string = validation_result_string ? validation_result_string : "";
+    if (!validation_result_string) {
+      validation_result_string = "undefined topic name type";
+    }
+
     if (is_service) {
       throw rclcpp::exceptions::InvalidServiceNameError(
               result.c_str(),
--- a/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
+++ b/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
@@ -104,8 +104,7 @@ rclcpp::expand_topic_or_service_name(
 
       if (validation_result == RCL_TOPIC_NAME_VALID) {
         throw std::runtime_error("topic name unexpectedly valid");
-      }
-      eles {
+      } else {
         const char * validation_message =
           rcl_topic_name_validation_result_string(validation_result);
         if (is_service) {
--- a/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
+++ b/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
@@ -102,9 +102,7 @@ rclcpp::expand_topic_or_service_name(
         throw_from_rcl_error(ret);
       }
 
-      if (validation_result == RCL_TOPIC_NAME_VALID) {
-        throw std::runtime_error("topic name unexpectedly valid");
-      } else {
+      if (validation_result != RCL_TOPIC_NAME_VALID) {
         const char * validation_message =
           rcl_topic_name_validation_result_string(validation_result);
         if (is_service) {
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
```diff
@@ -133,9 +133,13 @@ rclcpp::expand_topic_or_service_name(
           RCL_RET_ERROR, "failed to validate node name",
           rmw_get_error_state(), rmw_reset_error);
       }
+
+      const char * validation_result_string = rmw_node_name_validation_result_string(
+        validation_result);
+      validation_result_string = validation_result_string ? validation_result_string : "";
```
clalancette: While using the ternary operator isn't exactly wrong, it is definitely odd to be re-assigning a variable back to itself.  I'd prefer a simple `if (nullptr == validation_result_string)` here.  Also, we should probably have a slightly different string than `""`; maybe something like `"Unknown error"`.

The same goes for the rest of the calls below.

gaoethan: actually, I've ever tried to make the error string specific, but when the result is valid, it also return NULL which is different from the default NULL in some sense, so it's not accurate if we treat all the NULL to "unknown error", I prefer to tweak the `rmw_*validate_result_string` to get it more accurate, any comments ? thanks.

gaoethan:
please evaluate to unify the undefined case in rmw_*_validation_result_string or deal with separate code while using based on  #https://github.com/ros2/rmw/pull/130

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
clalancette: Yeah, I think I agree that we should ensure that the `rmw_*_validation_result_string` should probably never return NULL.  I'll make a similar comment over in ros2/rmw#130 , but let's leave this open for now just in case we come to a different conclusion over there.

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
```diff
@@ -105,17 +105,18 @@ rclcpp::expand_topic_or_service_name(
...
-      if (!validation_message) {
-        throw std::runtime_error("unable to get validation error message");
-      }
-      if (is_service) {
-        using rclcpp::exceptions::InvalidServiceNameError;
-        throw InvalidServiceNameError(name.c_str(), validation_message, invalid_index);
-      } else {
-        using rclcpp::exceptions::InvalidTopicNameError;
-        throw InvalidTopicNameError(name.c_str(), validation_message, invalid_index);
+      eles {
```
mikaelarguedas: typo ?
This also need to be on the same line as the above closing curly bracket to pass the linters

mikaelarguedas: Also this else is not necessary as if we enter the if we throw an error so the previous code can be kept

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
```diff
@@ -133,10 +134,14 @@ rclcpp::expand_topic_or_service_name(
...
-              node_name.c_str(),
-              rmw_node_name_validation_result_string(validation_result),
-              invalid_index);
+
+      if (validation_result != RMW_NODE_NAME_VALID) {
+        throw rclcpp::exceptions::InvalidNodeNameError(
+                node_name.c_str(),
+                rmw_node_name_validation_result_string(validation_result),
+                invalid_index);
+      }
```
mikaelarguedas: Not sure if what scenario that can happen but should we do something in the else case ? (case where the rcl node name is valid but the rmw node name is not) (same everywhere below)

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
gaoethan: yes, it's a typo (my bad), actually the else got changed to not on the same line after I ran **ament_uncrusitify --reformat**, let me check.

why I add the else is because it will return `NULL` if `validation_result` is `RCL_TOPIC_NAME_VALID`, it will be problematic for `InvalidTopicNameError` and `InvalidServiceNameError`

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
gaoethan: this `if` is under umbrella of `RCL_RET_NODE_INVALID_NAME` and it's currently handling the cases of invalid rmw node name as you point :). or what you really mean the case that **rcl node name is invalid** but **rwm node name is valid** ?

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
mikaelarguedas: > why I add the else is because it will return NULL if validation_result is RCL_TOPIC_NAME_VALID, it will be problematic for InvalidTopicNameError and InvalidServiceNameError

My understanding is that this is taken care of by the if just before 
```
        if (validation_result == RCL_TOPIC_NAME_VALID) {
          throw std::runtime_error("topic name unexpectedly valid");
        }
```

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
mikaelarguedas: Indeed that's what I meant :smile:

gaoethan: @mikaelarguedas now it's not sure when this will happen, I throw exceptions for the case of valid rmw with invalid rcl checking to allow the user to be aware of that for further investigation.

rclcpp/src/rclcpp/expand_topic_or_service_name.cpp
mikaelarguedas: @gaoethan This comment has not been addressed. 
I think that the following snippet should be kept. If this evaluates to true the code below is unreacheable as we throw. If it doesnt we're sure we're not in the case of  `validation_result == RCL_TOPIC_NAME_VALID`
```c++
        if (validation_result == RCL_TOPIC_NAME_VALID) {
          throw std::runtime_error("topic name unexpectedly valid");
        }
```

gaoethan: Actually, I have kept it, :) generally speaking, it has lower possibility to hit that and I put it in the [else statement](https://github.com/gaoethan/rclcpp/blob/4edc0930ca7dacbf3887d08f7920bbc19ca0b88b/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp#L115-117)

PR #406, opened 2017-11-21
Resolve erase race condition.

### LINKED ISSUE #401: The ros_clock create_jump_callback is not safe to overwrite the shared pointer
The tests pass if the callback_holder is created anew instead of overwriting the old one. There's clearly cleanup not working correctly.

tfoote:
Fixes #401 

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/3584/)
Linux aarch64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/779/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/2926/)
WIndows: [Build Status](http://ci.ros2.org/job/ci_windows/3676/)

### DIFF
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -130,15 +130,21 @@ Clock::get_triggered_callback_handlers(const TimeJump & jump)
 {
   std::vector<JumpHandler::SharedPtr> callbacks;
   std::lock_guard<std::mutex> guard(callback_list_mutex_);
-  for (auto wjcb = active_jump_handlers_.begin(); wjcb != active_jump_handlers_.end(); wjcb++) {
-    if (auto jcb = wjcb->lock()) {
-      if (jcb->notice_threshold.is_exceeded(jump)) {
-        callbacks.push_back(jcb);
-      }
-    } else {
-      active_jump_handlers_.erase(wjcb);
-    }
-  }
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/clock.cpp
```diff
@@ -130,13 +130,16 @@ Clock::get_triggered_callback_handlers(const TimeJump & jump)
 {
   std::vector<JumpHandler::SharedPtr> callbacks;
   std::lock_guard<std::mutex> guard(callback_list_mutex_);
-  for (auto wjcb = active_jump_handlers_.begin(); wjcb != active_jump_handlers_.end(); wjcb++) {
+  for (auto wjcb = active_jump_handlers_.begin(); wjcb != active_jump_handlers_.end(); ) {
```
Karsten1987: couldn't you do the loop with a `std::erase(std::remove_if()` call? 
http://en.cppreference.com/w/cpp/algorithm/remove

Then you don't have to deal with the indices yourself :)

tfoote: I switched it to a lambda with remove_if instead. It's actually a line or two longer, but should be more performant on the erase due to remove_if's reordering. Though I don't expect performance to be an issue here.

tfoote:
Linux CI with refactored remove_if [Build Status](http://ci.ros2.org/job/ci_linux/)

PR #411, opened 2017-11-30
Add Logger class and give one to nodes

dhood:
This adds a `Logger` class, the `node->get_logger()` method, and the `rclcpp::get_logger(name)` free function for non-node loggers.

Logging macros _must_ now have a logger object passed as the first argument (instead of a name). I added https://github.com/ros2/rclcpp/blob/ed590ad41e722e7c4015136d956416b6ce1b44a3/rclcpp/resource/logging.hpp.em#L47 to give something more useful than a wall of macro errors if users try to pass a string: open to other ideas.

The Logger class is based of a rough implementation initially by @wjwwood. The node's logger is implemented as an interface on @Karsten1987's suggestion to support the use case of composing a lightweight Node that does not include or "pay for" the logger interface.

Todo before merge:
- [x] remove logger namespace following https://github.com/ros2/rclcpp/pull/410#issuecomment-348305379
- [x] do whatever needs to be done for lifecycle_nodes to have a logger (haven't investigated yet)

Other todos listed in the code are planned as follow-up PRs.

ci windows [Build Status](http://ci.ros2.org/job/ci_windows/3748/)


rclcpp/include/rclcpp/logger.hpp
```diff
@@ -0,0 +1,132 @@
...
+ * name prefix.
+ * Currently there are no such naming conventions but they may be introduced in
+ * the future.
+ *
+ * \param[in] name the name of the logger
+ * \return a logger with the fully-qualified name including naming conventions, or
+ * \return a dummy logger if logging is disabled.
+ */
+RCLCPP_PUBLIC
+Logger get_logger(const std::string & name);
```
Karsten1987: Did you make up your mind whether you want to put that as a static member function? As discussed this would prevent you from friend/forward declaring it. However, I remember you mentioned it may be problematic for future developments.

rclcpp/include/rclcpp/logger.hpp
```diff
@@ -0,0 +1,132 @@
...
+  Logger(const Logger &) = default;
+
+  /// Get the name of this logger.
+  /**
+   * \return the full name of the logger including any prefixes, or
+   * \return `nullptr` if this logger is invalid (e.g. because logging is
+   *   disabled).
+   */
+  RCLCPP_PUBLIC
+  const char * get_name() const
```
Karsten1987: style nitpick: new line between return type and function name

rclcpp/include/rclcpp/node_interfaces/node_logger.hpp
```diff
@@ -0,0 +1,61 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
```
Karsten1987: year

rclcpp/include/rclcpp/logger.hpp
```diff
@@ -0,0 +1,132 @@
...
+namespace rclcpp
+{
+
+// Forward declaration is used for friend statement.
+namespace node_interfaces
+{
+class NodeLogger;
+}
+
+namespace logger
```
Karsten1987: given the discussion https://github.com/ros2/rclcpp/pull/410#issuecomment-348305379 you may want to remove the `namespace logger`.

rclcpp/include/rclcpp/logger.hpp
dhood: yeah, the usability was what I had in mind: I wanted it to be just `rclcpp::get_logger`. After we spoke I did try making it a static function but in the end I wasn't able to promote _just_ that one method to the rclcpp namespace like I thought I could. So for usability I left it as a free function

rclcpp/include/rclcpp/logger.hpp
dhood: yeah that's what I had listed above as a todo :+1:

Karsten1987: sorry, my fault. Only quicked looked over the code :)

dhood: nw, at least we had the _same_ point ;)

rclcpp_lifecycle/test/test_default_state_machine.cpp
```diff
@@ -0,0 +1,205 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
```
dhood: Note to reviewers: I wanted to add and use a `TestLifecycleNode` test fixture. I moved the default_state_machine tests to another file because having two different test fixtures in one file was causing rclcpp::init to be called twice.

dhood:
This is good to review now.

The usage of rclcpp logging macros has been updated in https://github.com/ros2/demos/pull/200, including usage for lifecycle nodes which also have a logger since a9c0ef74323bdb92495e74f0b17eea6acd237476

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3648/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/828/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2985/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3750/)

dhood:
This is an example error with gcc if you try to pass a string instead of a logger to a macro. Still not great user experience but the top and bottom lines mention the `Logger` class at least.
```
In file included from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:21:0,
                 from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/rclcpp.hpp:142,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/include/logging_demo/logger_usage_component.hpp:21,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:15:
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp: In member function ‘void logging_demo::LoggerUsage::on_timer()’:
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:50:28: error: invalid initialization of reference of type ‘const rclcpp::Logger&’ from expression of type ‘const char*’
   RCLCPP_INFO_ONCE(get_name(), "Timer callback called (this will only log once)")
                            ^
/home/dhood/ros2_ws/install_isolated/rcutils/include/rcutils/logging_macros.h:67:47: note: in definition of macro ‘RCUTILS_LOG_COND_NAMED’
     if (rcutils_logging_logger_is_enabled_for(name, severity)) { \
                                               ^
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:186:3: note: in expansion of macro ‘RCUTILS_LOG_INFO_ONCE_NAMED’
   RCUTILS_LOG_INFO_ONCE_NAMED( \
   ^
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:50:3: note: in expansion of macro ‘RCLCPP_INFO_ONCE’
   RCLCPP_INFO_ONCE(get_name(), "Timer callback called (this will only log once)")
   ^
In file included from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/rclcpp.hpp:142:0,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/include/logging_demo/logger_usage_component.hpp:21,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:15:
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:58:21: note: in passing argument 1 of ‘const char* rclcpp::logging_macro_utilities::_get_logger_name(const rclcpp::Logger&)’
 inline const char * _get_logger_name(const rclcpp::Logger & logger)
                     ^
In file included from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:21:0,
                 from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/rclcpp.hpp:142,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/include/logging_demo/logger_usage_component.hpp:21,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:15:
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:50:28: error: invalid initialization of reference of type ‘const rclcpp::Logger&’ from expression of type ‘const char*’
   RCLCPP_INFO_ONCE(get_name(), "Timer callback called (this will only log once)")
                            ^
/home/dhood/ros2_ws/install_isolated/rcutils/include/rcutils/logging_macros.h:69:58: note: in definition of macro ‘RCUTILS_LOG_COND_NAMED’
       rcutils_log(&__rcutils_logging_location, severity, name, __VA_ARGS__); \
                                                          ^
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:186:3: note: in expansion of macro ‘RCUTILS_LOG_INFO_ONCE_NAMED’
   RCUTILS_LOG_INFO_ONCE_NAMED( \
   ^
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:50:3: note: in expansion of macro ‘RCLCPP_INFO_ONCE’
   RCLCPP_INFO_ONCE(get_name(), "Timer callback called (this will only log once)")
   ^
In file included from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/rclcpp.hpp:142:0,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/include/logging_demo/logger_usage_component.hpp:21,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:15:
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:58:21: note: in passing argument 1 of ‘const char* rclcpp::logging_macro_utilities::_get_logger_name(const rclcpp::Logger&)’
 inline const char * _get_logger_name(const rclcpp::Logger & logger)
CMakeFiles/logger_usage_component.dir/build.make:62: recipe for target 'CMakeFiles/logger_usage_component.dir/src/logger_usage_component.cpp.o' failed

```

rclcpp_lifecycle/test/test_default_state_machine.cpp
dhood: mmm this file was just renamed... I didn't think we updated copyright in that situation (but I'm not 100% sure about these things)

Karsten1987: It was my initial fault. It just have been 2016 in the first place :/

wjwwood:
@dhood you could use a `static_assert` in the macro to make a better message, something like:

```c++
  static_assert(std::is_same<decltype(logger), Logger>::value, "Must use Logger class"); \
```

See: https://stackoverflow.com/a/4029310

dhood:
Gah, that was the first thing I tried but I gave up on it and figured it could only be used with standard types... actually the problem was whitespace in my macros..!! I shouldn't have given up so quickly.

Here's the error output with a519b725c4ad6dc9dc809119123c2e400a64ea60
```
In file included from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/rclcpp.hpp:142:0,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/include/logging_demo/logger_usage_component.hpp:21,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:15:
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp: In member function ‘void logging_demo::LoggerUsage::on_timer()’:
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:168:3: error: static assertion failed: First argument to logging macros must be an rclcpp::Logger
   static_assert(std::is_same<decltype(logger), rclcpp::Logger>::value, "First argument to logging macros must be an rclcpp::Logger"); \
   ^
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:50:3: note: in expansion of macro ‘RCLCPP_INFO_ONCE’
   RCLCPP_INFO_ONCE(get_name(), "Timer callback called (this will only log once)")
   ^
In file included from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:23:0,
                 from /home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/rclcpp.hpp:142,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/include/logging_demo/logger_usage_component.hpp:21,
                 from /home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:15:
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:170:12: error: request for member ‘get_name’ in ‘((logging_demo::LoggerUsage*)this)->logging_demo::LoggerUsage::<anonymous>.rclcpp::node::Node::get_name()’, which is of non-class type ‘const char*’
     logger.get_name(), \
            ^
/home/dhood/ros2_ws/install_isolated/rcutils/include/rcutils/logging_macros.h:67:47: note: in definition of macro ‘RCUTILS_LOG_COND_NAMED’
     if (rcutils_logging_logger_is_enabled_for(name, severity)) { \
                                               ^
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:169:3: note: in expansion of macro ‘RCUTILS_LOG_INFO_ONCE_NAMED’
   RCUTILS_LOG_INFO_ONCE_NAMED( \
   ^
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:50:3: note: in expansion of macro ‘RCLCPP_INFO_ONCE’
   RCLCPP_INFO_ONCE(get_name(), "Timer callback called (this will only log once)")
   ^
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:170:12: error: request for member ‘get_name’ in ‘((logging_demo::LoggerUsage*)this)->logging_demo::LoggerUsage::<anonymous>.rclcpp::node::Node::get_name()’, which is of non-class type ‘const char*’
     logger.get_name(), \
            ^
/home/dhood/ros2_ws/install_isolated/rcutils/include/rcutils/logging_macros.h:69:58: note: in definition of macro ‘RCUTILS_LOG_COND_NAMED’
       rcutils_log(&__rcutils_logging_location, severity, name, __VA_ARGS__); \
                                                          ^
/home/dhood/ros2_ws/install_isolated/rclcpp/include/rclcpp/logging.hpp:169:3: note: in expansion of macro ‘RCUTILS_LOG_INFO_ONCE_NAMED’
   RCUTILS_LOG_INFO_ONCE_NAMED( \
   ^
/home/dhood/ros2_ws/src/ros2/demos/logging_demo/src/logger_usage_component.cpp:50:3: note: in expansion of macro ‘RCLCPP_INFO_ONCE’
   RCLCPP_INFO_ONCE(get_name(), "Timer callback called (this will only log once)")
   ^
CMakeFiles/logger_usage_component.dir/build.make:62: recipe for target 'CMakeFiles/logger_usage_component.dir/src/logger_usage_component.cpp.o' failed
```

rclcpp/include/rclcpp/logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+ * Currently there are no such naming conventions but they may be introduced in
+ * the future.
+ *
+ * \param[in] name the name of the logger
+ * \return a logger with the fully-qualified name including naming conventions, or
+ * \return a dummy logger if logging is disabled.
+ */
+RCLCPP_PUBLIC
+Logger
+get_logger(const std::string & name);
```
dhood: do we think this should return a shared ptr [as log4cxx does](https://github.com/apache/logging-log4cxx/blob/master/src/main/include/log4cxx/logger.h#L680) or leave it up to callers to call `make_shared` or similar?

tfoote: In our discussion it sounded like it might make sense to return a shared_ptr here. In particular this would allow us to implement the actual Logger with a specialized subclass that potentially overrides the get_name or other capabilities and keeps a weak backlink to the node and can output a warning if node goes out of scope and the backlink is broken. And although the current logger is trivially copyable right now we wouldn't need to keep it that way looking forward if we return the pointer instead of an instance.

rclcpp_lifecycle/test/test_default_state_machine.cpp
dhood: I removed the test changes completely, it was bloating the diff unnecessarily

dhood:
@ros2/team This is good to review if anyone has time; at the moment this is targeted for release but if it is off-base and doesn't make it in nothing breaks. The usage updates are in https://github.com/ros2/demos/pull/200

As @tfoote mentioned in https://github.com/ros2/rclcpp/pull/411/files#r154480879 there may be changes to how the logger is stored/accessed by the Node class, but we can do them at a later date.

rclcpp/include/rclcpp/logger.hpp
wjwwood: We talked about this off-line (@dhood and I) and I pointed out that if we want to do what you describe @tfoote we don't _have_ swap the return type to a shared pointer, instead we could have the Logger class store a shared pointer and forward all methods calls to the shared pointer. Basically we could accomplish the same thing without inheritance.

That being said, if @dhood wanted to switch this to a pointer, that's fine by me too.

Also, if we really want to return a shared pointer we can do that in the future with an API break.

rclcpp/include/rclcpp/logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+      return nullptr;
+    }
+    return name_->c_str();
+  }
+
+  /// Return a logger that is a descendant of this logger.
+  /**
+   * The child logger's full name will include any hierarchy conventions that
+   * indicate it is a descendant of this logger.
+   * For example, ```get_logger('abc').get_child('def')``` will return a logger
```
wjwwood: nitpick: which use triple back tick here but single elsewhere?

rclcpp/include/rclcpp/node_interfaces/node_logger.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "rclcpp/node_interfaces/node_logger_interface.hpp"
+#include "rclcpp/visibility_control.hpp"
+
+namespace rclcpp
+{
+namespace node_interfaces
+{
+
+/// Implementation of the NodeLogger part of the Node API.
+class NodeLogger : public NodeLoggerInterface
```
wjwwood: I have a small concern here that `NodeLogger` sounds to much like a thing that logs for a node (which is not really what it is). What about `NodeLogging` and `NodeLoggingInterface` instead? I had a similar dilemma with `NodeTopic` versus `NodeTopics`, but I ended up going with `NodeTopics` because `NodeTopicsInterface` sounded better. For the same logic I think decided to suggest `NodeLogging` here, but I'll leave it to your discretion.

rclcpp/src/rclcpp/node_interfaces/node_logger.cpp
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+#include "rclcpp/node_interfaces/node_logger.hpp"
+
+using rclcpp::node_interfaces::NodeLogger;
+
+NodeLogger::NodeLogger(rclcpp::node_interfaces::NodeBaseInterface * node_base)
+: node_base_(node_base)
+{
+  // TODO(dhood): use the namespace (slashes converted to dots)
```
wjwwood: This would be nice to have in too, since that's part of the point of having the node logger right? So users don't have to do something like `convert_slash_to_dot(node->get_namespace()) + node->get_name()` for every logging call? (but still get the benefit of better namespaces in logging by default)

I understand this is higher risk to merge though...

rclcpp/include/rclcpp/logger.hpp
dhood: it wasn't rendering the full expression without the triple (the others are simpler)

rclcpp/src/rclcpp/node_interfaces/node_logger.cpp
dhood: It's definitely planned, I agree that it is the main point of the node logger. My thinking was to put the "logger name from node name/namespace" in rcl and then wrap it with python as well. Because of the reach of that change and the fact that things at the user level shouldn't change once we make the swap under the hood though, I consider it a separate "step" and so scoped it out of this PR for time reasons. next step though :smile:

dhood: It's definitely planned, I agree that it is the main point of the node logger. My thinking was to put the "logger name from node name/namespace" in rcl and then wrap it with python as well. Because of the reach of that change and the fact that things at the user level shouldn't change once we make the swap under the hood though, I consider it a separate "step" and so scoped it out of this PR for time reasons. It'll be the very next step though :smile:

rclcpp/include/rclcpp/node_interfaces/node_logger.hpp
dhood: Yeah, the `NodeLogger` attribute of the `Node`/`LifecycleNode` class is not a logger but a way to access the logger. will switch to `NodeLogging` as proposed

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3714/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/874/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3049/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3812/)

those are all an unrelated linter failure that's been fixed

rclcpp/src/rclcpp/node_interfaces/node_logger.cpp
dhood: Following up: https://github.com/ros2/rclcpp/pull/433

Issue #409, opened 2017-11-27
Can't create ParameterService from within node constructor


dhood:
Working on https://github.com/ros2/demos/pull/192 I realised that classes that inherit from Node don't have an appropriate constructor they can call (from what I can see) to create a parameter service.

The only ParameterService constructor there is [takes a node shared pointer](https://github.com/ros2/rclcpp/blob/070b3125c1a7f76fbde3e31973d89a052a2ea652/rclcpp/include/rclcpp/parameter_service.hpp#L44), which can't be used from within a Node constructor because its weak pointer won't have been initialised yet. From a usability standpoint this means the parameter service needs to be created after the node is instantiated, not in its constructor, which is not ideal.

@Karsten1987 mentioned that this was at one point an issue for regular Clients/Services as well but was addressed by https://github.com/ros2/rclcpp/commit/f0afcabf052161c7b44b10665e74e34593dc55bf (see https://github.com/ros2/rclcpp/pull/277#discussion_r91813549).

There's a workaround available for parameter clients by using [the AsyncParametersClient constructor that takes the interface pointers](https://github.com/ros2/rclcpp/blob/022b2b1b807d8239b492e81a8fc156e36c60724d/rclcpp/include/rclcpp/parameter_client.hpp#L52), as the interfaces will have been initialised by that point. However, the call will look like:
```
parameters_client_ = std::make_shared<rclcpp::parameter_client::AsyncParametersClient>(
  get_node_base_interface(),
  get_node_topics_interface(),
  get_node_graph_interface(),
  get_node_services_interface());
```
which is not a great user experience either.

---

Would it make sense to:
1. Have a constructor for the ParameterService that takes the relevant interfaces, and the version that takes the node shared pointer will just forward to it (as is done for AsyncParametersClient)? This will allow users to create a parameter service in the constructor, but it will not be pretty.
1. Make convenience functions in the Node class for creating parameter services and parameter clients, so that the work of getting the node interfaces (shown above) is hidden from users?

Neither of these changes are necessary if users instead do something like:
```
auto node = std::make_shared<ParameterNode>();
auto parameters_client_ = std::make_shared<rclcpp::parameter_client::AsyncParametersClient>(node);
auto sub = parameters_client_->on_parameter_event(std::bind(&ParameterNode::on_parameter_event, node, std::placeholders::_1));
```

[...]

PR #413, opened 2017-12-03
Allow creating parameter client from constructor of Node subclass

dhood:
Closes ros2/rclcpp#409

I only did this for parameter clients: parameter services still need to be created outside of the constructor, but this is not something that we expect users to have to do manually in the long term.

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -64,6 +64,12 @@ public:
     const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
 
   RCLCPP_PUBLIC
+  AsyncParametersClient(
+    rclcpp::node::Node * node,
+    const std::string & remote_node_name = "",
+    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
+
+  RCLCPP_PUBLIC
   std::shared_future<std::vector<rclcpp::parameter::ParameterVariant>>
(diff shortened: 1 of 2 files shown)


dhood:
Example usage in https://github.com/ros2/demos/pull/203/files#diff-653dd97d51d2e5cbd57891eb644a8654R26 (note that the parameter server is still created after the node has been instantiated. We could make it possible from the constructor, but I was avoiding adding API we'd likely remove later on.

dhood:
I'm putting this into review because I've successfully tested https://github.com/ros2/demos/pull/203 that makes use of this (see https://github.com/ros2/demos/pull/203#issuecomment-348819190)

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -63,6 +63,12 @@ class AsyncParametersClient
     const std::string & remote_node_name = "",
     const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
 
+  RCLCPP_PUBLIC
+  AsyncParametersClient(
+    rclcpp::node::Node * node,
```
dhood: note that this is not `const` because things like [`node->get_node_base_interface()`](https://github.com/ros2/rclcpp/blob/284dc1791874c8d6141dc83f3817405a95006cae/rclcpp/include/rclcpp/node.hpp#L374) are not

dhood:
yeah, but waffle doesn't propagate the label to connected tickets sometimes

PR #414, opened 2017-12-03
Prevent callback from being captured as a (potentially invalid) refererence

dhood:
Async parameter clients have an optional callback in the `set_parameters` method, but it's not used in the demos.

I tried to use it in https://github.com/ros2/demos/pull/203/files#diff-a5a151889abb3214250ba9db05358e4bR92 but it caused a segfault because the callback was being captured as a reference that became invalid by the time it was called from the callback of the parameter clients's internal client.


dhood:
regression test added in https://github.com/ros2/system_tests/pull/253

dhood:
I'm waiting for this OS X build: [Build Status](http://ci.ros2.org/job/ci_osx/3039/)

PR #415, opened 2017-12-03
Allow client to trigger another service call from its callback

dhood:
In https://github.com/ros2/demos/pull/203/files#diff-a5a151889abb3214250ba9db05358e4bR83 I was trying to call `set_parameters()` on an async parameter client in the callback of another `set_parameters` call, but it would never send the request.

@wjwwood helped me realise it had deadlocked waiting for the mutex.


dhood:
regression test added in https://github.com/ros2/system_tests/pull/254

PR #416, opened 2017-12-04
Remove e.g. `node::` namespaces and namespace escalation

dhood:
This is following a decision we made Nov '16 but hadn't followed up on. Summary in https://github.com/ros2/rclcpp/pull/410#issuecomment-348305379.

The reasoning for doing it now is because doing it while in beta we won't have to deprecate the old format. If that's too aggressive we can look into the suggestion in https://github.com/ros2/rclcpp/pull/410#issuecomment-348361651.

My opinion on the priority is that this should be in before we tag but doesn't necessarily need to be in before testing since it will be clear from compilation whether or not it's been done correctly. I have a bunch of changes to other packages ready.

So far I've removed namespaces that only had classes. They are not always exactly matching the class name e.g. `parameter_client::` had `AsyncParamtersClient` and `SyncParametersClient`, so let me know if I went too extreme and we can easily revert individual commits.
- `publisher::`
- `subscription::`
- `client::`
- `service::`
- `parameter_client::`
- `paramter_service::`
- `rate::`
- `timer::`
- `node::`
- `any_service_callback::`
- `any_subscription_callback::`
- `context::`
- `event::`
- `executors::single_threaded_executor::SingleThreadedExecutor` -> `executors::SingleThreadedExecutor`
- `executors::multi_threaded_executor::MultiThreadedExecutor` -> `executors::MultiThreadedExecutor`

I removed the now-unnecessary namespace escalations from `rclcpp/rclcpp.hpp` but left the includes for those files because FWIU we still want users to not have to include the headers themselves.

---

There are some namespaces that have more than just classes so I wanted to check what to do with them:

- `executor::` has `executor::Executor` the class, but also:
  - `executor::AnyExecutable` a struct
  - `executor::FutureReturnCode` an enum
  - `executor::ExecutorArgs` a struct

- `parameter::` has
  - `parameter::ParameterVariant` class
  - `parameter::ParameterType` an enum
  - `parameter::operator<<`

Perhaps just move `rclcpp::executor::Executor` -> `rclcpp:Executor` and leave the others as is?

### DIFF
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -45,7 +45,7 @@ struct AnyExecutable
   rclcpp::SubscriptionBase::SharedPtr subscription;
   rclcpp::SubscriptionBase::SharedPtr subscription_intra_process;
   rclcpp::timer::TimerBase::SharedPtr timer;
-  rclcpp::service::ServiceBase::SharedPtr service;
+  rclcpp::ServiceBase::SharedPtr service;
   rclcpp::ClientBase::SharedPtr client;
   // These are used to keep the scope on the containing items
   rclcpp::callback_group::CallbackGroup::SharedPtr callback_group;
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -44,7 +44,7 @@ struct AnyExecutable
   // Only one of the following pointers will be set.
   rclcpp::SubscriptionBase::SharedPtr subscription;
   rclcpp::SubscriptionBase::SharedPtr subscription_intra_process;
-  rclcpp::timer::TimerBase::SharedPtr timer;
+  rclcpp::TimerBase::SharedPtr timer;
   rclcpp::ServiceBase::SharedPtr service;
   rclcpp::ClientBase::SharedPtr client;
   // These are used to keep the scope on the containing items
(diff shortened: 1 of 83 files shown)


wjwwood:
Doesn't matter to me either, I don't expect normal users to use those other classes. Long term we should probably just put them each in their own files, at which point we might want to put them in a subfolder like `executor` or something, but for now I don't think it matters.

Even `Executor` is sort of moot because users will always use the single threaded or multi threaded executor, unless they're making their own which inherits from `Executor`.

dhood:
ok I'll leave the `executor` and `parameter` namespaces as is. Users do interact with `parameter::ParameterVariant` but we can address it later if appropriate.

@wjwwood I forgot to ask: what do we want done with `rclcpp::ok()` and friends? https://github.com/ros2/rclcpp/pull/416/files#diff-690567624ee368716009118718985564R157
Remove the `utilities` namespace, or stick to the escalation?

wjwwood:
Remove them I think, they're just in a utilities file, not a utilities folder right?

dhood:
correct (but most other things we've spoken about were classes not free functions). I'll move em up!

dhood:
Good to review now

CI for this change and all connected PRs that change usage to not include removed namespaces:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3739/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/892/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3081/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3835/)

Turtlebot ci including many more repos:
[Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/108/)

@wjwwood (and any others) I can help with adapting rviz to work with this change or hold off merging until appropriate

PR #418, opened 2017-12-05
Fix 393

Karsten1987:
Connects to https://github.com/ros2/rclcpp/issues/393

This is a redo of https://github.com/ros2/rclcpp/pull/395 split up into each individual issue, which is associated to it. This should help the review process.

Regarding https://github.com/ros2/rclcpp/pull/395#discussion_r154537015:
The `label` stays a `const char *` and get deallocated with a const cast. This is then the same practice as https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_cpp/src/rmw_node.cpp#L142

### DIFF
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -41,16 +41,18 @@ State::State(
     throw std::runtime_error("Lifecycle State cannot have an empty label.");
   }
 
-  auto state_handle = reinterpret_cast<rcl_lifecycle_state_t *>(
+  state_handle_ = static_cast<rcl_lifecycle_state_t *>(
     allocator_.allocate(sizeof(rcl_lifecycle_state_t), allocator_.state));
-  if (!state_handle) {
+  if (!state_handle_) {
     throw std::runtime_error("failed to allocate memory for rcl_lifecycle_state_t");
   }
-  state_handle->id = id;
-  state_handle->label =
+  state_handle_->id = id;
+  state_handle_->label =
     rcutils_strndup(label.c_str(), label.size(), allocator_);
-
-  state_handle_ = state_handle;
+  if (!state_handle_->label) {
+    reset();
+    throw std::runtime_error("failed to duplicate label for rcl_lifecycle_state_t");
+  }
 }
 
 State::State(
@@ -60,37 +62,42 @@ State::State(
   owns_rcl_state_handle_(false),
   state_handle_(nullptr)
 {
+  if (!rcl_lifecycle_state_handle) {
+    throw std::runtime_error("rcl_lifecycle_state_handle is null");
+  }
   state_handle_ = const_cast<rcl_lifecycle_state_t *>(rcl_lifecycle_state_handle);
 }
 
 State::~State()
 {
-  if (owns_rcl_state_handle_) {
-    if (state_handle_->label) {
-      allocator_.deallocate(const_cast<char *>(state_handle_->label), allocator_.state);
-      state_handle_->label = nullptr;
-    }
-    allocator_.deallocate(state_handle_, allocator_.state);
-    state_handle_ = nullptr;
-  }
+  reset();
 }
 
 uint8_t
 State::id() const
 {
-  if (!state_handle_) {
-    throw std::runtime_error("internal state_handle is null");
-  }
   return state_handle_->id;
 }
 
 std::string
 State::label() const
 {
-  if (!state_handle_) {
-    throw std::runtime_error("internal state_handle is null");
-  }
   return state_handle_->label;
 }
 
+void
+State::reset()
+{
+  if (!owns_rcl_state_handle_) {
+    state_handle_ = nullptr;
+    return;
+  }
+
+  if (state_handle_->label) {
+    allocator_.deallocate(const_cast<char *>(state_handle_->label), allocator_.state);
+    state_handle_->label = nullptr;
+  }
+  allocator_.deallocate(state_handle_, allocator_.state);
+  state_handle_ = nullptr;
+}
 }  // namespace rclcpp_lifecycle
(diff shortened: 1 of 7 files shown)


rclcpp_lifecycle/src/state.cpp
```diff
@@ -25,43 +25,61 @@
...
+: allocator_(allocator),
+  owns_rcl_state_handle_(true),
+  state_handle_(nullptr)
 {
   if (label.empty()) {
     throw std::runtime_error("Lifecycle State cannot have an empty label.");
   }
 
-  auto state_handle = new rcl_lifecycle_state_t;
+  auto state_handle = reinterpret_cast<rcl_lifecycle_state_t *>(
```
dirk-thomas: A `static_cast` should be used to cast a `void *` to a type specific pointer.

Same below.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -25,43 +25,61 @@
...
+    allocator_.allocate(sizeof(rcl_lifecycle_state_t), allocator_.state));
+  if (!state_handle) {
+    throw std::runtime_error("failed to allocate memory for rcl_lifecycle_state_t");
+  }
   state_handle->id = id;
   state_handle->label =
-    rcutils_strndup(label.c_str(), label.size(), rcutils_get_default_allocator());
+    rcutils_strndup(label.c_str(), label.size(), allocator_);
```
dirk-thomas: The return code should be checked for `null`.

Same below.

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -36,54 +50,109 @@ Transition::Transition(uint8_t id, const std::string & label)
...
   transition_handle->id = id;
-  transition_handle->label = label.c_str();
+  transition_handle->label = rcutils_strndup(
+    label.c_str(), label.size(), allocator_);
 
-  auto start_state = new rcl_lifecycle_state_t;
+  auto start_state = reinterpret_cast<rcl_lifecycle_state_t *>(
+    allocator_.allocate(sizeof(rcl_lifecycle_state_t), allocator_.state));
+  if (!start_state) {
+    throw std::runtime_error("failed to allocate memory for rcl_lifecycle_state_t");
```
dirk-thomas: This would leak `transition_handle`.

Same below.

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -36,54 +50,109 @@ Transition::Transition(uint8_t id, const std::string & label)
...
-  transition_handle_ = rcl_lifecycle_transition_handle;
+  transition_handle_ = const_cast<rcl_lifecycle_transition_t *>(rcl_lifecycle_transition_handle);
 }
 
 Transition::~Transition()
 {
-  if (owns_rcl_transition_handle_) {
-    if (transition_handle_->start) {
-      delete transition_handle_->start;
+  if (!transition_handle_) {
```
dirk-thomas: Why does the logic in this destructor different from the one for the `State` class?. E.g. it shouldn't be possible for `transition_handle_` to be null.

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -36,54 +50,109 @@ Transition::Transition(uint8_t id, const std::string & label)
...
-: owns_rcl_transition_handle_(false)
+Transition::Transition(
+  const rcl_lifecycle_transition_t * rcl_lifecycle_transition_handle,
+  rcutils_allocator_t allocator)
+: allocator_(allocator),
+  owns_rcl_transition_handle_(false),
+  transition_handle_(nullptr)
 {
-  transition_handle_ = rcl_lifecycle_transition_handle;
+  transition_handle_ = const_cast<rcl_lifecycle_transition_t *>(rcl_lifecycle_transition_handle);
```
dirk-thomas: Should it be checked that `transition_handle_` is not `null`?

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -36,54 +50,109 @@ Transition::Transition(uint8_t id, const std::string & label)
...
+  allocator_.deallocate(transition_handle_, allocator_.state);
+  transition_handle_ = nullptr;
 }
 
 uint8_t
 Transition::id() const
 {
+  if (!transition_handle_) {
+    throw std::runtime_error("internal transition_handle is null");
+  }
```
dirk-thomas: If the constructor protects the class from passing `null` explicitly this can never happen.

rclcpp_lifecycle/src/transition.cpp
Karsten1987: I added the `reset()` function in this PR and call it whenever one of the allocation fails.

dirk-thomas: I think the allocated handles are not fully initialized with zeros therefore the `reset()` function will not be able to detect if a specific field needs to be deallocated since it might contain non-null values without containing actual data.

dirk-thomas: > I think the allocated handles are not fully initialized with zeros therefore the reset() function will not be able to detect if a specific field needs to be deallocated since it might contain non-null values without containing actual data.

I am not sure where I noticed this. Looking again at the current state of the patch all members of the `*_state_` variables are initialized. So please ignore this comment.

Karsten1987:
@dirk-thomas up for another round of review.
https://github.com/ros2/rcl/pull/197/commits/26159891008cc06d1408b189e89e833c3523d9f1 introduces a set of init and fini function for the rcl_lifecycle structs.

CI:
[Build Status](http://ci.ros2.org/job/ci_linux/3746/)
[Build Status](http://ci.ros2.org/job/ci_osx/3086/)
[Build Status](http://ci.ros2.org/job/ci_windows/3840/)

PR #419, opened 2017-12-05
Fix 394

Karsten1987:
Connects to https://github.com/ros2/rclcpp/issues/394

This is a redo of https://github.com/ros2/rclcpp/pull/395 split up into each individual issue, which is associated to it. This should help the review process.

This can only be merged after #418 has been merged.

### DIFF
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -153,6 +153,10 @@ State::reset()
     return;
   }
 
+  // TODO(karsten1987): Fini currently deallocate the state_handle_ instance as well
+  // this should be changed to only deallocate members of the pointer so that stack
+  // variables can be correctly used as well.
+  // see https://github.com/ros2/rclcpp/pull/419#discussion_r155157098
   auto ret = rcl_lifecycle_state_fini(state_handle_, &allocator_);
   if (ret != RCL_RET_OK) {
     rclcpp::exceptions::throw_from_rcl_error(ret);
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -48,6 +48,9 @@ State::State(
   if (!state_handle_) {
     throw std::runtime_error("failed to allocate memory for rcl_lifecycle_state_t");
   }
+  // zero initialize
+  state_handle_->id = 0;
+  state_handle_->label = nullptr;
 
   auto ret = rcl_lifecycle_state_init(state_handle_, id, label.c_str(), &allocator_);
   if (ret != RCL_RET_OK) {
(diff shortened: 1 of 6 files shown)


dirk-thomas:
If the classes ensure that the `state_handle_` / `transition_handle_` are never `null` (as described in the comments on #418) this should simplify some of the logic in this patch.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -86,30 +86,33 @@ State::~State()
...
+  // we own the handle, so we have to deep-copy the rhs object
+  state_handle_ = static_cast<rcl_lifecycle_state_t *>(
+    allocator_.allocate(sizeof(rcl_lifecycle_state_t), allocator_.state));
+  if (!state_handle_) {
+    throw std::runtime_error("failed to allocate memory for rcl_lifecycle_state_t");
+  }
+  auto ret = rcl_lifecycle_state_init(
+    state_handle_, rhs.id(), rhs.label().c_str(), &allocator_);
+  if (ret != RCL_RET_OK) {
+    reset();
```
dirk-thomas: Again, `state_handle_` has not been zero initialized, so no guarantee that `reset` can do the right thing.

Same below.

Karsten1987: I don't see how. If the init returns false, it means that the label pointer is null and thus can be freed correctly.
The same with transitions. We call init on it with two nullptr.

Karsten1987:
CI:
[Build Status](http://ci.ros2.org/job/ci_linux/3754/)
[Build Status](http://ci.ros2.org/job/ci_osx/3094/)
[Build Status](http://ci.ros2.org/job/ci_windows/3848/)

rclcpp_lifecycle/src/state.cpp
```diff
@@ -70,11 +72,58 @@ State::State(
...
+  if (!owns_rcl_state_handle_) {
+    state_handle_ = rhs.state_handle_;
+    return *this;
+  }
+
+  // we own the handle, so we have to deep-copy the rhs object
+  state_handle_ = static_cast<rcl_lifecycle_state_t *>(
+    allocator_.allocate(sizeof(rcl_lifecycle_state_t), allocator_.state));
+  if (!state_handle_) {
+    throw std::runtime_error("failed to allocate memory for rcl_lifecycle_state_t");
```
wjwwood: nitpick: probably should be `std::bad_alloc` (http://en.cppreference.com/w/cpp/memory/new/bad_alloc)

(I don't think this needs to be in for this release, just a fyi)

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -96,16 +108,110 @@ Transition::Transition(
...
+
+  // we don't own the handle, so we can return straight ahead
+  if (!owns_rcl_transition_handle_) {
+    transition_handle_ = rhs.transition_handle_;
+    return *this;
+  }
+
+  // we own the handle, so we have to deep-copy the rhs object
+  transition_handle_ = static_cast<rcl_lifecycle_transition_t *>(
+    allocator_.allocate(sizeof(rcl_lifecycle_transition_t), allocator_.state));
```
wjwwood: So I noticed that the allocation here's matching deallocation is in the fini function for this struct (same for above rcl struct in State class). While this isn't wrong, it's definitely different from how the rest of the fini functions work. Rather than iterate over the internet, I'm giving @Karsten1987 the ok to merge (since it's not broken, just weird) and tomorrow I'll help him clean up the implementation.

PR #420, opened 2017-12-05
switch to send clock with type rcl_interfaces/Time

tfoote:
This is a workaround for https://github.com/ros2/rmw_opensplice/issues/216 if we can't resolve it for the pending release this should make things work. 

In local testing I'm seeing some general overall timeouts with opensplice.


dirk-thomas:
I propose to close this in favor of ros2/rmw_opensplice#217.

PR #422, opened 2017-12-05
Fix linter issues that have appeared

dhood:
I think there might be some linter failures the _just_ started appearing on master.. still checking it's not just from my PR; opening to avoid duplicating efforts.

[Build Status](http://ci.ros2.org/job/ci_linux-aarch64/891/)


mikaelarguedas:
for reference: linters pass locally on the current master

dhood:
uncrustify must have just started behaving differently when I removed an encompassing namespace from that file... this will probably end up closed

PR #424, opened 2017-12-06
deallocate state and transition handles after call to fini

Karsten1987:
Connects to https://github.com/ros2/rcl/issues/199

### DIFF
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -154,11 +154,11 @@ State::reset()
   }
 
   auto ret = rcl_lifecycle_state_fini(state_handle_, &allocator_);
+  allocator_.deallocate(state_handle_, allocator_.state);
+  state_handle_ = nullptr;
   if (ret != RCL_RET_OK) {
     rclcpp::exceptions::throw_from_rcl_error(ret);
   }
-  allocator_.deallocate(state_handle_, allocator_.state);
-  state_handle_ = nullptr;
 }
 
 }  // namespace rclcpp_lifecycle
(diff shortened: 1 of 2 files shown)


rclcpp_lifecycle/src/state.cpp
```diff
@@ -153,14 +153,12 @@ State::reset()
...
-  // TODO(karsten1987): Fini currently deallocate the state_handle_ instance as well
-  // this should be changed to only deallocate members of the pointer so that stack
-  // variables can be correctly used as well.
-  // see https://github.com/ros2/rclcpp/pull/419#discussion_r155157098
   auto ret = rcl_lifecycle_state_fini(state_handle_, &allocator_);
   if (ret != RCL_RET_OK) {
     rclcpp::exceptions::throw_from_rcl_error(ret);
   }
+  allocator_.deallocate(state_handle_, allocator_.state);
+  state_handle_ = nullptr;
```
wjwwood: Shouldn't this still be done when there is an error?

rclcpp_lifecycle/src/transition.cpp
```diff
@@ -257,13 +257,11 @@ Transition::reset()
...
-  // TODO(karsten1987): Fini currently deallocate the transition_handle_ instance as well
-  // this should be changed to only deallocate members of the pointer so that stack
-  // variables can be correctly used as well.
-  // see https://github.com/ros2/rclcpp/pull/419#discussion_r155157098
   auto ret = rcl_lifecycle_transition_fini(transition_handle_, &allocator_);
   if (ret != RCL_RET_OK) {
     rclcpp::exceptions::throw_from_rcl_error(ret);
   }
+  allocator_.deallocate(transition_handle_, allocator_.state);
+  transition_handle_ = nullptr;
```
wjwwood: Same.

rclcpp_lifecycle/src/state.cpp
Karsten1987: agreed.

PR #426, opened 2017-12-18
Fix the potential application crash issues

gaoethan:
`get_topic_name` and `rcl_service_get_service_name` may return NULL

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rclcpp/src/rclcpp/publisher.cpp
+++ b/rclcpp/src/rclcpp/publisher.cpp
@@ -172,7 +172,7 @@ PublisherBase::setup_intra_process(
 {
   const char * topic_name = this->get_topic_name();
   if (!topic_name) {
-    throw std::runtime_error("Get a invalid null topic name");
+    throw std::runtime_error("failed to get topic name");
   }
 
   auto intra_process_topic_name = std::string(topic_name) + "/_intra";
--- a/rclcpp/src/rclcpp/publisher.cpp
+++ b/rclcpp/src/rclcpp/publisher.cpp
@@ -171,12 +171,11 @@ PublisherBase::setup_intra_process(
   const rcl_publisher_options_t & intra_process_options)
 {
   const char * topic_name = this->get_topic_name();
-  std::string intra_process_topic_name = "";
-  if (topic_name) {
-    intra_process_topic_name = std::string(topic_name) + "/_intra";
-  } else {
+  if (!topic_name) {
     throw std::runtime_error("Get a invalid null topic name");
   }
+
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/publisher.cpp
```diff
@@ -170,7 +170,13 @@ PublisherBase::setup_intra_process(
   StoreMessageCallbackT callback,
   const rcl_publisher_options_t & intra_process_options)
 {
-  auto intra_process_topic_name = std::string(this->get_topic_name()) + "/_intra";
+  const char * topic_name = this->get_topic_name();
+  if (topic_name) {
+    auto intra_process_topic_name = std::string(topic_name) + "/_intra";
```
dirk-thomas: Have you tried to compile the patch locally (and ran the tests)? It look to me that the variable has a local scope and therefore isn't available below where it is being used.

gaoethan: badly, it becomes a local scope within if{}, and I changed it. Now all have passed the build and tests, thanks !

rclcpp/src/rclcpp/publisher.cpp
```diff
@@ -170,7 +170,14 @@ PublisherBase::setup_intra_process(
...
   const rcl_publisher_options_t & intra_process_options)
 {
-  auto intra_process_topic_name = std::string(this->get_topic_name()) + "/_intra";
+  const char * topic_name = this->get_topic_name();
+  std::string intra_process_topic_name = "";
+  if (topic_name) {
+    intra_process_topic_name = std::string(topic_name) + "/_intra";
+  } else {
+    throw std::runtime_error("Get a invalid null topic name");
+  }
```
dirk-thomas: Please use a similar structure here as above. Check for `!topic_name` and conditionally `throw`, Then the `else` case becomes obsolete.

gaoethan: make sense, changed :smile:

dirk-thomas:
I modified the error message slight. Looks good to me now:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3798/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/934/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3135/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3884/)

clalancette:
Builds are all green, @dirk-thomas thinks it is OK, looks OK to me.  Merging.

PR #427, opened 2018-01-11
Move clear wait set from after rcl_wait to ahead

jwang11:
Current code do clear wait set after rcl_wait, it is not respond latency
friendly. In fact, clear wait set operation is not urgent, making sure it is
done before next rcl_wait should be fine. With this commit, ready task
will get executed more quickly 

Signed-off-by: jwang <jing.j.wang@intel.com>

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -376,7 +376,7 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
       )
     );
   }
-  // clear wait set 
+  // clear wait set
   if (rcl_wait_set_clear_subscriptions(&wait_set_) != RCL_RET_OK) {
     throw std::runtime_error("Couldn't clear subscriptions from wait set");
   }
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -376,6 +376,22 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
       )
     );
   }
+  // clear wait set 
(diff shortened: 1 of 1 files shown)


wjwwood:
I'm combining the CI for this pr and https://github.com/ros2/rcl/pull/207:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3863/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/987/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3192/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3941/)

dirk-thomas:
Before the `wait_set_` was already "cleared" when the function `wait_for_work()` returned. Now that is not the case anymore (since it is being done on the next call of `wait_for_work()`). Does that require that the cleanup logic in the destructor needs to take care of that instead?

wjwwood:
I would say no, since it only prepares the storage to be used in a repeated loop, it doesn't clean up any resources or anything. It's analogous to memset to zero. The docs reflect this as well:

https://github.com/ros2/rcl/blob/c09f6428d57ea715ad26067cc4c3e581343eea9b/rcl/include/rcl/wait.h#L212-L234

wjwwood:
@jwang11 this was failing our linters due to trailing whitespace, but I think I fixed it in 77ab3bc.

New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3882/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1001/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3208/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3971/)

Issue #349, opened 2017-08-04
resetting subscriber leads to crash


Karsten1987:
given the following code: https://gist.github.com/Karsten1987/82837b87fd11ef23353754b9fa31cf7a

the situation:

1.) node spawns a subscriber and a publisher
2.1) main thread spins the executor, having sub and pub talking
2.2) second thread randomly stops the subscription (resetting the shared_ptr of it)
3.) program crashes with the backtrace at the bottom of the gist.

quickly discussed the problem with @wjwwood, when resetting the subscription, the executor still has a raw handle on it which leads to race conditions. The problem can be addressed by turning the handle into a shared pointer. 

concretely speaking, [this](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/subscription.hpp#L81-L83) function should return a shared_ptr instead of a raw pointer.

dhood:
this is the permalink to the function @Karsten1987 is referencing in his post (in case those lines move around): https://github.com/ros2/rclcpp/blob/b90676871d397cbfdcfecdf265869efb0bd97d9f/rclcpp/include/rclcpp/subscription.hpp#L81-L83

deng02:
We have been hitting this issue when using the ROS2-to-ROS1 ros1_bridge with bridge_all_2to1_topics set to false.  When the ROS1 subscriber for a topic disappers the ROS2 subscriber is destroyed and sometimes this happens while the wait set is still holding a raw pointer to a member of that object.  The result is various segfault errors within rmw_wait.

wjwwood:
I believe https://github.com/ros2/rclcpp/pull/431 closes this, please comment here if this is still an issue.

PR #431, opened 2018-01-16
Store the subscriber, client, service and timer

deng02:
objects as shared pointers (#349)

To prevent an object from being deleted while the rcl_wait_set is
using raw pointers to internal members, we store them as shared
pointers.

The subscriptions are stored as a pair because a single
subscription handle can have both an intra and non-intra
process handle being used by the wait set.  In order to
remove objects based on null wait set handles, we need both.

### DIFF
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -32,6 +32,7 @@
 #include "rclcpp/type_support_decl.hpp"
 #include "rclcpp/expand_topic_or_service_name.hpp"
 #include "rclcpp/visibility_control.hpp"
+#include "rclcpp/logging.hpp"
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
 
@@ -122,8 +123,8 @@ public:
       new rcl_service_t, [ = ](rcl_service_t * service)
       {
         if (rcl_service_fini(service, node_handle_.get()) != RCL_RET_OK) {
-          RCUTILS_LOG_ERROR_NAMED(
-            "rclcpp",
+          RCLCPP_ERROR(
+            rclcpp::get_logger(rcl_node_get_name(node_handle.get())).get_child("rclcpp"),
             "Error in destruction of rcl service handle: %s",
             rcl_get_error_string_safe());
           rcl_reset_error();
@@ -202,7 +203,6 @@ public:
     service_handle_ = std::shared_ptr<rcl_service_t>(new rcl_service_t);
     service_handle_->impl = service_handle->impl;
   }
-
 
   Service() = delete;
 
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -21,6 +21,8 @@
 #include <sstream>
 #include <string>
 
+#include "rcutils/logging_macros.h"
+
 #include "rcl/error_handling.h"
 #include "rcl/service.h"
 
@@ -58,11 +60,11 @@ public:
   get_service_name();
 
   RCLCPP_PUBLIC
-  rcl_service_t *
+  std::shared_ptr<rcl_service_t>
   get_service_handle();
 
   RCLCPP_PUBLIC
-  const rcl_service_t *
+  std::shared_ptr<const rcl_service_t>
   get_service_handle() const;
 
   virtual std::shared_ptr<void> create_request() = 0;
@@ -84,7 +86,7 @@ protected:
 
   std::shared_ptr<rcl_node_t> node_handle_;
 
-  rcl_service_t * service_handle_ = nullptr;
+  std::shared_ptr<rcl_service_t> service_handle_;
   std::string service_name_;
   bool owns_rcl_handle_ = true;
 };
@@ -116,11 +118,22 @@ public:
     auto service_type_support_handle = get_service_type_support_handle<ServiceT>();
 
     // rcl does the static memory allocation here
-    service_handle_ = new rcl_service_t;
-    *service_handle_ = rcl_get_zero_initialized_service();
+    service_handle_ = std::shared_ptr<rcl_service_t>(
+      new rcl_service_t, [ = ](rcl_service_t * service)
+      {
+        if (rcl_service_fini(service, node_handle_.get()) != RCL_RET_OK) {
+          RCUTILS_LOG_ERROR_NAMED(
+            "rclcpp",
+            "Error in destruction of rcl service handle: %s",
+            rcl_get_error_string_safe());
+          rcl_reset_error();
+        }
+        delete service;
+      });
+    *service_handle_.get() = rcl_get_zero_initialized_service();
 
     rcl_ret_t ret = rcl_service_init(
-      service_handle_,
+      service_handle_.get(),
       node_handle.get(),
       service_type_support_handle,
       service_name.c_str(),
@@ -143,16 +156,36 @@ public:
 
   Service(
     std::shared_ptr<rcl_node_t> node_handle,
+    std::shared_ptr<rcl_service_t> service_handle,
+    AnyServiceCallback<ServiceT> any_callback)
+  : ServiceBase(node_handle),
+    any_callback_(any_callback)
+  {
+    // check if service handle was initialized
+    if (!rcl_service_is_valid(service_handle.get(), nullptr)) {
+      // *INDENT-OFF* (prevent uncrustify from making unnecessary indents here)
+      throw std::runtime_error(
+        std::string("rcl_service_t in constructor argument must be initialized beforehand."));
+      // *INDENT-ON*
+    }
+
+    const char * service_name = rcl_service_get_service_name(service_handle.get());
+    if (!service_name) {
+      throw std::runtime_error("failed to get service name");
+    }
+    service_handle_ = service_handle;
+    service_name_ = std::string(service_name);
+  }
+
+  Service(
+    std::shared_ptr<rcl_node_t> node_handle,
     rcl_service_t * service_handle,
     AnyServiceCallback<ServiceT> any_callback)
   : ServiceBase(node_handle),
     any_callback_(any_callback)
   {
     // check if service handle was initialized
-    // TODO(karsten1987): Take this verification
-    // directly in rcl_*_t
-    // see: https://github.com/ros2/rcl/issues/81
-    if (!service_handle->impl) {
+    if (!rcl_service_is_valid(service_handle, nullptr)) {
       // *INDENT-OFF* (prevent uncrustify from making unnecessary indents here)
       throw std::runtime_error(
         std::string("rcl_service_t in constructor argument must be initialized beforehand."));
@@ -163,26 +196,18 @@ public:
     if (!service_name) {
       throw std::runtime_error("failed to get service name");
     }
-    service_handle_ = service_handle;
     service_name_ = std::string(service_name);
-    owns_rcl_handle_ = false;
+
+    // In this case, rcl owns the service handle memory
+    service_handle_ = std::shared_ptr<rcl_service_t>(new rcl_service_t);
+    service_handle_->impl = service_handle->impl;
   }
+
 
   Service() = delete;
 
   virtual ~Service()
   {
-    // check if you have ownership of the handle
-    if (owns_rcl_handle_) {
-      if (rcl_service_fini(service_handle_, node_handle_.get()) != RCL_RET_OK) {
-        std::stringstream ss;
-        ss << "Error in destruction of rcl service_handle_ handle: " <<
-          rcl_get_error_string_safe() << '\n';
-        (std::cerr << ss.str()).flush();
-        rcl_reset_error();
-      }
-      delete service_handle_;
-    }
   }
 
   std::shared_ptr<void> create_request()
@@ -211,7 +236,7 @@ public:
     std::shared_ptr<rmw_request_id_t> req_id,
     std::shared_ptr<typename ServiceT::Response> response)
   {
-    rcl_ret_t status = rcl_send_response(get_service_handle(), req_id.get(), response.get());
+    rcl_ret_t status = rcl_send_response(get_service_handle().get(), req_id.get(), response.get());
 
     if (status != RCL_RET_OK) {
       rclcpp::exceptions::throw_from_rcl_error(status, "failed to send response");
--- a/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
@@ -98,22 +98,22 @@ public:
   {
     for (size_t i = 0; i < wait_set->size_of_subscriptions; ++i) {
       if (!wait_set->subscriptions[i]) {
-        subscription_handles_[i] = nullptr;
+        subscription_handles_[i].reset();
       }
     }
     for (size_t i = 0; i < wait_set->size_of_services; ++i) {
       if (!wait_set->services[i]) {
-        service_handles_[i] = nullptr;
+        service_handles_[i].reset();
       }
     }
     for (size_t i = 0; i < wait_set->size_of_clients; ++i) {
       if (!wait_set->clients[i]) {
-        client_handles_[i] = nullptr;
+        client_handles_[i].reset();
       }
     }
     for (size_t i = 0; i < wait_set->size_of_timers; ++i) {
       if (!wait_set->timers[i]) {
-        timer_handles_[i] = nullptr;
+        timer_handles_[i].reset();
       }
     }
 
@@ -188,7 +188,7 @@ public:
   bool add_handles_to_wait_set(rcl_wait_set_t * wait_set)
   {
     for (auto subscription : subscription_handles_) {
-      if (rcl_wait_set_add_subscription(wait_set, subscription) != RCL_RET_OK) {
+      if (rcl_wait_set_add_subscription(wait_set, subscription.get()) != RCL_RET_OK) {
         RCUTILS_LOG_ERROR_NAMED(
           "rclcpp",
           "Couldn't add subscription to wait set: %s", rcl_get_error_string_safe());
@@ -197,7 +197,7 @@ public:
     }
 
     for (auto client : client_handles_) {
-      if (rcl_wait_set_add_client(wait_set, client) != RCL_RET_OK) {
+      if (rcl_wait_set_add_client(wait_set, client.get()) != RCL_RET_OK) {
         RCUTILS_LOG_ERROR_NAMED(
           "rclcpp",
           "Couldn't add client to wait set: %s", rcl_get_error_string_safe());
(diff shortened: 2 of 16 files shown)


rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -379,10 +379,10 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
...
   VectorRebind<const rcl_guard_condition_t *> guard_conditions_;
 
-  VectorRebind<const rcl_subscription_t *> subscription_handles_;
-  VectorRebind<const rcl_service_t *> service_handles_;
-  VectorRebind<const rcl_client_t *> client_handles_;
-  VectorRebind<const rcl_timer_t *> timer_handles_;
+  VectorRebind<std::pair<const rcl_subscription_t *, SubscriptionBase::SharedPtr>> sub_pair_ptrs_;
+  VectorRebind<rclcpp::ServiceBase::SharedPtr> service_ptrs_;
+  VectorRebind<rclcpp::ClientBase::SharedPtr> client_ptrs_;
+  VectorRebind<rclcpp::TimerBase::SharedPtr> timer_ptrs_;
```
wjwwood: I think these should instead store `shared_ptr`'s to the `rcl_*` types. Here's my rationale why, consider this case (current behavior):

- user creates pub and sub which talk
- user spins
- asynchronously user reset's the sub pointer
- when a message is received for the deleted subscription:
  - a segfault occurs in `rcl_wait()` for using "finalized" pointer to a `rcl_subscription_t`

The existing behavior, described in https://github.com/ros2/rclcpp/issues/349, this resulted in segfault's when `rcl_wait` was still using an `rcl_*` object, because that object had had it's `_fini()` function called on it from the destructor of one of our C++ classes.

With the implementation in this pr it goes like this:

- user creates pub and sub which talk
- user spins
- asynchronously user reset's the sub pointer
- when a message is received for the deleted subscription:
  - a callback is called for the subscription the user has already deleted

With the proposal from #349, i.e. using `shared_ptr` to the `rcl_*` types, it would go like this (I think):

- user creates pub and sub which talk
- user spins
- asynchronously user reset's the sub pointer
- when a message is received for the deleted subscription:
  - the message is "dropped" because the `weak_ptr` to the subscription the executor has fails to materialize into a `shared_ptr` because it no longer exists

wjwwood:
It would also be great to have an automated test for this case, so we don't regress on this point. Maybe someone on our team has time to do that if @deng02 doesn't.

deng02:
@wjwwood  Thanks very much for the feedback. I just want to confirm my understanding of the changes you want.

The PR in its current state artificially delays the destruction of the ROS2 subscriber to ensure that the 'finalized' pointer doesn't become invalidated until we're done with it, which somewhat violates the idea that the user is the owner and is managing the lifecycle of that subscriber.

Instead if we replace the raw pointers with `weak_ptr`s, we can safely dereference within `rcl_wait` when the user resets a sub pointer while preservering user-driven ownership.

The one thing I want to clarify is about having the `SubscriptionBase::get_handle()` return a `shared_ptr`.  Isn't `SubscriptionBase` essentially the owner of the `subscription_handle_` and all others are just observers, in which case the member itself would be of type `std::unique_ptr<rcl_subscription_t>` and `get_handle()` would return `std::weak_ptr<rcl_subscription_t>'?  I'm not sure I see why we need `shared_ptr` at all.  The `rcl_wait_set_t` would need to be modified to use the `weak_ptr` so that when `rcl_wait` tries to access, it can lock to determine if the pointer is still valid or not.

deng02:
Also I definitely agree with the need for an automated test around this.  I will try to write one if the team does not have time to create one.

wjwwood:
Hmm, I think we're not on the same page, let me try to clarify what I mean, but first let me respond to your latest feedback inline.

> The PR in its current state artificially delays the destruction of the ROS2 subscriber to ensure that the 'finalized' pointer doesn't become invalidated until we're done with it, which somewhat violates the idea that the user is the owner and is managing the lifecycle of that subscriber.

Right, except we need to differentiate in the C++ class and the C `rcl` handle for the resource, e.g. `rclcpp::Subscription` versus `rcl_subscription_t`. Which due to intra-process communication there's not even a 1-to-1 mapping, e.g. a `rclcpp::Subscription` typically has shared ownership of two different `rcl_subscription_t`.

But you're right to summarize the issue as violating the contract with the user about the lifecycle of the `rclcpp::Subscription`.

> The one thing I want to clarify is about having the `SubscriptionBase::get_handle()` return a `shared_ptr`. Isn't `SubscriptionBase` essentially the owner of the `subscription_handle_` and all others are just observers, in which case the member itself would be of type `std::unique_ptr<rcl_subscription_t>` and `get_handle()` would return `std::weak_ptr<rcl_subscription_t>`?

I would say no, I would word it as: `SubscriptionBase` starts with unique ownership of the `subscription_handle_` but may share the ownership with callers of `get_handle()`. In which case I think it does need to return a `shared_ptr` and not a `weak_ptr` (though the caller could always upgrade the `weak_ptr` to a `shared_ptr`)

> I'm not sure I see why we need `shared_ptr` at all. The `rcl_wait_set_t` would need to be modified to use the `weak_ptr` so that when `rcl_wait` tries to access, it can lock to determine if the pointer is still valid or not.

Well, `rcl_wait_set_t` cannot use `weak_ptr` or `shared_ptr` since it's a C api. But the problem is that it has to lock it to put it into the wait set and pass it to `dds_wait` at the bottom. So while waiting it needs to have a lock on it. But while it has a lock on it you cannot call fini on it from the destructor of the C++ class.

So basically I'm proposing that we decouple the C++ class and the `rcl` type, such that the user can only control when the C++ class is destroyed. So basically the sequence would go like this:

- [thread 1] user creates subscription and adds the node to an executor
- [thread 1] user spins executor
  - each time it enters `Executor::wait_for_work()` (and calls `AllocatorMemoryStrategy::collect_entities()`) the executor will gain shared ownership of the `rcl_*_t`'s it is waiting on (but not the `rclcpp::*` classes as this pr does)
- [thread 2] user deletes subscription
  - after `~Subscription()` returns the `rcl_subscription_t` will still exist (due to shared ownership with the executor) and therefore it will still appear to exist on the ROS graph, however data received for it will not be delivered
- [thread 1] at some point (for any reason) the executor loops and lets go of it's shared ownership of the `rcl_subscription_t`, at which point it is passed to `rcl_subscription_fini()` in a custom destructor for the `shared_ptr<rcl_subscription_t>`

So the documentation for `rclcpp::Subscription<T>` (specifically `rclcpp::~Subscriptions`) would say something like "on deletion, the subscription is scheduled for removal from the ROS graph and will be actually removed at some undetermined point in the future".

This is obviously not ideal, as the user would like to have some more determinism in how the shutdown of the subscription works. However, this is the simplest solution for the errors/segfaults (and what is implicitly happening right now).

----

A separate, but more completely (and more complicated) solution is to make it so that the destructors of our C++ classes force the executor(s) to give up shared ownership of the `rcl_*_t` classes before calling the equivalent `rcl_*_fini()` function on them and returning. For example, here's a similar sequence of events as above:

- [thread 1] user creates subscription and adds the node to an executor
- [thread 1] user spins executor
  - each time it enters `Executor::wait_for_work()` (and calls `AllocatorMemoryStrategy::collect_entities()`) the executor will gain shared ownership of the `rcl_*_t`'s it is waiting (still necessary to do the `rcl_*_t` instead of the `rclcpp::*` class otherwise the user cannot reliably trigger the destructor)
- [thread 1 or 2] user deletes subscription
  - if in thread 1, that would be from within another callback
    - special consider would need to be given for deleting a subscription from within it's own callback
  - within `~Subscription()` it would try to acquire unique ownership of the `rcl_*_t` by interrupting the executor and preventing it from entering `Executor::wait_for_work()` again
  - at this point, the executor is guaranteed to not be using the `rcl_*_t`
  - then `rcl_*_fini()` would be called and then the executor would be released so it could continue
  - when the executor continues and it calls `get_handle()` on this subscription (assuming it still gets the `rclcpp::Subscription` from the node or callback group) it will get `nullptr` back and will not add anything to the `rcl_wait_set_t` for this subscription (even if it continues before this destructor returns)

This would change the documentation of the classes, e.g. `rclcpp::Subscription()`, to be something more like "after destruction the subscription will have been removed from the ROS graph" which is much more deterministic and easier to understand for the user.

However, it is not a trivial thing to do (in my opinion), since you need to avoid deadlocks and starvation (usually using a pair of mutex, of which one is a "barrier" mutex, e.g. I use these in the `rclcpp::GraphListener`: https://github.com/ros2/rclcpp/blob/b81f55e5df0c3a43ba454b741e772ba77de7375b/rclcpp/include/rclcpp/graph_listener.hpp#L162-L163). It's possible, but tricky code. However, it might be the right solution long term. One downside is that the single threaded executors basically don't need _all_ of the sophisticated checks, only the multi-threaded executors need everything, so it would be best if the single threaded executors could avoid this more sophisticated logic somehow (they still need to interrupt the executor, but they don't need the extra mutex to ensure they get the next lock since there's only one thread in the executor).

Hope that clears up what I mean.

deng02:
@wjwwood  Yes that clears things up, thanks.  I'll move forwards with the changes as described above.

guillaumeautran:
@wjwwood I've re-worked the patch to convert the `rcl_*_t` to `std::shared_ptr`. Please take look and let me know if that works. 

Note: I've tested this change quite a bit with topic subscriptions and also used valgrind to track memory leaks. What is the best way for me to test the other components?

wjwwood:
@guillaumeautran awesome, I'll have a look at it as soon as I can, though I can't promise an immediate turn around right now.

guillaumeautran:
Ping @wjwwood :)

wjwwood:
Sorry @guillaumeautran, I'll get to it as soon as I can, I just have a lot of other stuff on my plate at the moment.

rclcpp/include/rclcpp/service.hpp
```diff
@@ -33,6 +35,7 @@
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
 
+
```
wjwwood: nitpick: minimize vertical whitespace and also avoid pure whitespace changes in pr's

rclcpp/include/rclcpp/service.hpp
```diff
@@ -116,11 +119,22 @@ class Service : public ServiceBase
...
 
     // rcl does the static memory allocation here
-    service_handle_ = new rcl_service_t;
-    *service_handle_ = rcl_get_zero_initialized_service();
+    service_handle_ = std::shared_ptr<rcl_service_t>(
+      new rcl_service_t, [=](rcl_service_t *service)
+      {
+        if (rcl_service_fini(service, node_handle_.get()) != RCL_RET_OK) {
+          RCUTILS_LOG_ERROR_NAMED(
+            "RCLCPP",
```
wjwwood: I think the logger names ought to be lowercase.

wjwwood: Also, this might be a good candidate for using a `rclcpp` sub logger of the node's logger so that it appears as `my_node.rclcpp` rather than just `rclcpp`. Imagine this error comes out on a process with several nodes in it. Having the node name as context would help you narrow it down.

rclcpp/include/rclcpp/service.hpp
```diff
@@ -141,6 +155,32 @@ class Service : public ServiceBase
...
+    std::shared_ptr<rcl_node_t> node_handle,
+    std::shared_ptr<rcl_service_t> service_handle,
+    AnyServiceCallback<ServiceT> any_callback)
+  : ServiceBase(node_handle),
+    any_callback_(any_callback)
+  {
+    // check if service handle was initialized
+    // TODO(karsten1987): Take this verification
+    // directly in rcl_*_t
+    // see: https://github.com/ros2/rcl/issues/81
```
wjwwood: Looks like https://github.com/ros2/rcl/issues/81 is closed, can you try replacing this TODO and code snippet with `rcl_service_is_valid()`? http://docs.ros2.org/ardent/api/rcl/service_8h.html#ae3f8159e4c6c43f9f2cc10dd4c5f5f3f

rclcpp/include/rclcpp/service.hpp
```diff
@@ -163,26 +203,18 @@ class Service : public ServiceBase
...
-    service_handle_ = service_handle;
     service_name_ = std::string(service_name);
-    owns_rcl_handle_ = false;
+
+    // In this case, rcl owns the service handle memory
+    service_handle_ = std::shared_ptr<rcl_service_t>(new rcl_service_t);
+    service_handle_->impl = service_handle->impl;
   }
 
+
```
wjwwood: nitpick: minimize vertical whitespace (for reference, this is part of the Google Style guide which our style is based on, see: https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/src/rclcpp/subscription.cpp
```diff
@@ -18,6 +18,8 @@
 #include <memory>
 #include <string>
 
+#include <rcutils/logging_macros.h>
+
```
wjwwood: nitpick: this should be included after the `rclcpp` headers or the `rmw/*` headers included below should be moved up, again following the Google Style guides recommendation but with some flexibility. In my opinion the only thing that's important is that there is separation between "this project's headers" and "other project's headers" of which the `rclcpp/*` includes and `rmw/*`/`rcutils/*` includes fit into respectively.

rclcpp/src/rclcpp/subscription.cpp
```diff
@@ -34,8 +36,28 @@ SubscriptionBase::SubscriptionBase(
   const rcl_subscription_options_t & subscription_options)
 : node_handle_(node_handle)
 {
+
```
wjwwood: nitpick: unnecessary whitespace

rclcpp/src/rclcpp/subscription.cpp
```diff
@@ -34,8 +36,28 @@ SubscriptionBase::SubscriptionBase(
...
+  auto custom_deletor = [=](rcl_subscription_t *rcl_subs)
+  {
+    std::cout << "Deleting subscription handle\n";
```
wjwwood: Looks like vestigial debug output.

rclcpp/src/rclcpp/timer.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "rclcpp/timer.hpp"
+#include "rcutils/logging_macros.h"
```
wjwwood: This should be included after the system headers.

rclcpp/src/rclcpp/timer.cpp
```diff
@@ -21,11 +22,29 @@ using rclcpp::TimerBase;
 
 TimerBase::TimerBase(std::chrono::nanoseconds period)
 {
+
```
wjwwood: nitpick: unnecessary vertical whitespace

rclcpp/src/rclcpp/timer.cpp
```diff
@@ -62,16 +81,16 @@ std::chrono::nanoseconds
 TimerBase::time_until_trigger()
 {
   int64_t time_until_next_call = 0;
-  if (rcl_timer_get_time_until_next_call(&timer_handle_, &time_until_next_call) != RCL_RET_OK) {
+  if (rcl_timer_get_time_until_next_call(timer_handle_.get(), &time_until_next_call) != RCL_RET_OK) {
     throw std::runtime_error(
-            std::string("Timer could not get time until next call: ") +
-            rcl_get_error_string_safe());
+      std::string("Timer could not get time until next call: ") +
+      rcl_get_error_string_safe());
```
wjwwood: Did you run `ament_uncrustify` or the test suite on this? This is a known oddity of how uncrustify makes use format the code. I would actually expect it to fail the tests with this diff.

rclcpp/include/rclcpp/service.hpp
wjwwood: Here and elsewhere in the pr.

rclcpp/include/rclcpp/service.hpp
wjwwood: @Karsten1987 FYI

rclcpp/include/rclcpp/service.hpp
```diff
@@ -116,11 +119,22 @@ class Service : public ServiceBase
     auto service_type_support_handle = get_service_type_support_handle<ServiceT>();
 
     // rcl does the static memory allocation here
-    service_handle_ = new rcl_service_t;
-    *service_handle_ = rcl_get_zero_initialized_service();
+    service_handle_ = std::shared_ptr<rcl_service_t>(
+      new rcl_service_t, [=](rcl_service_t *service)
+      {
+        if (rcl_service_fini(service, node_handle_.get()) != RCL_RET_OK) {
+          RCUTILS_LOG_ERROR_NAMED(
```
wjwwood: You should use the RCLCPP version of these macros now, see:

http://docs.ros2.org/ardent/api/rclcpp/logging_8hpp.html

This also removes the need for (and therefore implicitly address several other style comments I made) the new rcutils includes.

rclcpp/include/rclcpp/service.hpp
guillaumeautran: Ok, I've fixed the casing for RCLCPP => rclcpp. 

Regarding the node sublogger, shouldn't this be a rcutils concerns (as opposed to encoding the node name when writing the log)?

rclcpp/include/rclcpp/service.hpp
guillaumeautran: done.

rclcpp/src/rclcpp/subscription.cpp
guillaumeautran: oops...

rclcpp/include/rclcpp/service.hpp
guillaumeautran: Looking into it now...

rclcpp/include/rclcpp/service.hpp
wjwwood: Well not if you use the RCLCPP_* logging macros, then it’s rclcpp’s issue. Also how would rcutils know anything about node subloggers. It seems pretty obvious to me that the name needs to be injected by the user.

guillaumeautran: I've done it for this message only. But the code looks ugly to me. I wonder if there is a smarter way to simplify that...

wjwwood: If the service had a handle to the C++ node, then it could call `rclcpp::Node::get_logger()`, which would  be slightly cleaner looking, but it does exactly the same thing. I'll try to re-review this asap.

wjwwood:
Looks good on CI too, thanks for your contribution and patience @guillaumeautran and @deng02!

wjwwood:
@guillaumeautran if you have time to contribute a test for this that would be great too. Realistically, I'll not have time to do it, nor will anyone on our team most likely.

PR #433, opened 2018-01-23
Get node's logger name from rcl

dhood:
connects to ros2/rcl#212

I have a question on this PR. Which approach is more appropriate in the context of node interfaces:

1. logging interface directly calling through to rcl using the node handle [(this commit)](https://github.com/ros2/rclcpp/commit/c440b0857c0cf97a5c2349c4ffdb74a75a71e712), or
1. going through node_base [(this commit)](https://github.com/ros2/rclcpp/commit/88af1f2e39157f26257e30bf2826c0c063407355)?

It's not clear to me since on one hand the logger name is used (only) by the logging interface, but on the other hand at the rcl level it is tightly coupled to what seems to be the node "base".

### DIFF
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -57,6 +57,11 @@ public:
 
   RCLCPP_PUBLIC
   virtual
+  const char *
+  get_logger_name() const;
+
+  RCLCPP_PUBLIC
+  virtual
   rclcpp::Context::SharedPtr
   get_context();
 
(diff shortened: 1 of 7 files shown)


dirk-thomas:
I am leaning towards the first option since imo the second one keeps adding more and more API to the node / interfaces (which I am not a big fan of). But I don't mind either way and would let others comment on this which are more familiar with the pros / cons of the two options.

wjwwood:
Option 3?

Provide the `get_logger_name()` function on the `NodeLoggingInterface` and "dog food" it (if possible, essential option 1 if not possible) in the `NodeLogging` implementation, so that if it is used from outside, it can be mocked via the `NodeLoggingInterface` pure virtual base class.

dhood:
waiting for review @ros2/team

PR #445, opened 2018-02-28
update style

dirk-thomas:
Connect to ~~ament/uncrustify#20~~. Connect to ament/uncrustify#21.

Ready for review.

### DIFF
--- a/rclcpp/include/rclcpp/allocator/allocator_deleter.hpp
+++ b/rclcpp/include/rclcpp/allocator/allocator_deleter.hpp
@@ -94,11 +94,11 @@ void set_allocator_for_deleter(AllocatorDeleter<T> * deleter, Alloc * alloc)
 
 template<typename Alloc, typename T>
 using Deleter = typename std::conditional<
-    std::is_same<typename std::allocator_traits<Alloc>::template rebind_alloc<T>,
-    typename std::allocator<void>::template rebind<T>::other>::value,
-    std::default_delete<T>,
-    AllocatorDeleter<Alloc>
-    >::type;
+  std::is_same<typename std::allocator_traits<Alloc>::template rebind_alloc<T>,
+  typename std::allocator<void>::template rebind<T>::other>::value,
+  std::default_delete<T>,
+  AllocatorDeleter<Alloc>
+  >::type;
 }  // namespace allocator
 }  // namespace rclcpp
 
--- a/rclcpp/include/rclcpp/function_traits.hpp
+++ b/rclcpp/include/rclcpp/function_traits.hpp
@@ -49,7 +49,7 @@ template<typename FunctionT>
 struct function_traits
 {
   using arguments = typename tuple_tail<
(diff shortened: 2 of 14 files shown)


rclcpp/include/rclcpp/allocator/allocator_deleter.hpp
```diff
@@ -94,11 +94,11 @@ void set_allocator_for_deleter(AllocatorDeleter<T> * deleter, Alloc * alloc)
...
-    std::is_same<typename std::allocator_traits<Alloc>::template rebind_alloc<T>,
-    typename std::allocator<void>::template rebind<T>::other>::value,
-    std::default_delete<T>,
-    AllocatorDeleter<Alloc>
-    >::type;
+  std::is_same<typename std::allocator_traits<Alloc>::template rebind_alloc<T>,
+  typename std::allocator<void>::template rebind<T>::other>::value,
+  std::default_delete<T>,
+  AllocatorDeleter<Alloc>
+  >::type;
```
wjwwood: This is better. It's unfortunate that the last line `>::type` isn't indented one less (should end on the same column as the line that starts it).

wjwwood: A similar comment applies to all follow multi-line templates.

rclcpp/include/rclcpp/exceptions.hpp
```diff
@@ -116,7 +116,7 @@ throw_from_rcl_error(
   rcl_ret_t ret,
   const std::string & prefix = "",
   const rcl_error_state_t * error_state = nullptr,
-  void (*reset_error)() = rcl_reset_error);
+  void (* reset_error)() = rcl_reset_error);
```
wjwwood: Not sure this is better (more correct), but I'm fine with it too.

rclcpp/include/rclcpp/function_traits.hpp
```diff
@@ -130,20 +130,20 @@ struct function_traits<FunctionT &&>: function_traits<FunctionT>
...
 struct arity_comparator : std::integral_constant<
-    bool, (Arity == function_traits<FunctorT>::arity)>{};
+    bool, (Arity == function_traits<FunctorT>::arity)> {};
 
 template<typename FunctorT, typename ... Args>
 struct check_arguments : std::is_same<
     typename function_traits<FunctorT>::arguments,
     std::tuple<Args ...>
-  >
+>
```
wjwwood: This one has the right indentation for the closing `>`, but does 4 spaces rather than two on the inside... Which is somehow the counterpart of the one I commented on above. If only we could combine the two...

rclcpp/include/rclcpp/allocator/allocator_deleter.hpp
dirk-thomas: When targeting the latest release this "improvement" is sadly not possible. Therefore I revert this change.

PR #454, opened 2018-03-21
Remove ros arguments

mjcarroll:
Addresses rclcpp component of ros2/rcl#219

Connects to ros2/rcl#219

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -55,11 +55,23 @@ RCLCPP_PUBLIC
 void
 init(int argc, char const * const argv[]);
 
+/// Initialize communications via the rmw implementation and set up a global signal handler.
+/**
+ * Additionally removes ROS-specific arguments from the argument vector.
+ * \param[in] argc Number of arguments.
+ * \param[in] argv Argument vector.
+ * \returns Members of the argument vector that are not ROS arguments.
+ */
+RCLCPP_PUBLIC
+std::vector<std::string>
+init_and_remove_ros_arguments(int argc, char const * const argv[]);
+
 /// Remove ROS-specific arguments from argument vector.
 /**
- * Some arguments may not have been intended as ROS arguments. This function
- * populates a the aruments in a vector. Since the first argument is always assumed
- * to be a process name, the vector will always contain the process name.
+ * Some arguments may not have been intended as ROS arguments.
+ * This function populates a the aruments in a vector.
+ * Since the first argument is always assumed to be a process name, the vector
+ * will always contain the process name.
  *
  * \param[in] argc Number of arguments.
  * \param[in] argv Argument vector.
@@ -68,18 +80,6 @@ init(int argc, char const * const argv[]);
 RCLCPP_PUBLIC
 std::vector<std::string>
 remove_ros_arguments(int argc, char const * const argv[]);
-
-/// Initialize communications via the rmw implementaiton and set up a global signal handler.
-/**
- * This does everything that init does, but additionally removes ROS-specific command line
- * arguments
- * \param[in] argc Number of arguments.
- * \param[in] argv Argument vector.
- * \return Members of the argument vector that are not ROS arguments.
- */
-RCLCPP_PUBLIC
-std::vector<std::string>
-init_and_remove_ros_arguments(int argc, char const * const argv[]);
 
 /// Check rclcpp's status.
 /** \return True if SIGINT hasn't fired yet, false otherwise. */
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -18,6 +18,7 @@
 #include <chrono>
 #include <functional>
 #include <limits>
+#include <vector>
 
 #include "rclcpp/visibility_control.hpp"
 
@@ -45,7 +46,6 @@ std::string to_string(T value)
 
 namespace rclcpp
 {
-
 /// Initialize communications via the rmw implementation and set up a global signal handler.
 /**
  * \param[in] argc Number of arguments.
@@ -54,6 +54,32 @@ namespace rclcpp
 RCLCPP_PUBLIC
 void
 init(int argc, char const * const argv[]);
+
+/// Remove ROS-specific arguments from argument vector.
+/**
+ * Some arguments may not have been intended as ROS arguments. This function
+ * populates a the aruments in a vector. Since the first argument is always assumed
+ * to be a process name, the vector will always contain the process name.
+ *
+ * \param[in] argc Number of arguments.
+ * \param[in] argv Argument vector.
+ * \returns Members of the argument vector that are not ROS arguments.
+ */
+RCLCPP_PUBLIC
+std::vector<std::string>
+remove_ros_arguments(int argc, char const * const argv[]);
+
+/// Initialize communications via the rmw implementaiton and set up a global signal handler.
+/**
+ * This does everything that init does, but additionally removes ROS-specific command line
+ * arguments
+ * \param[in] argc Number of arguments.
+ * \param[in] argv Argument vector.
+ * \return Members of the argument vector that are not ROS arguments.
+ */
+RCLCPP_PUBLIC
+std::vector<std::string>
+init_and_remove_ros_arguments(int argc, char const * const argv[]);
 
 /// Check rclcpp's status.
 /** \return True if SIGINT hasn't fired yet, false otherwise. */
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -53,7 +53,7 @@ namespace rclcpp
  */
 RCLCPP_PUBLIC
 void
-init(int argc, char * argv[]);
+init(int argc, char const * const argv[]);
 
 /// Check rclcpp's status.
 /** \return True if SIGINT hasn't fired yet, false otherwise. */
--- a/rclcpp/src/rclcpp/utilities.cpp
+++ b/rclcpp/src/rclcpp/utilities.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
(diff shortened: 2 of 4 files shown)


mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4127/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1208/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3410/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4234/)

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -45,15 +46,40 @@ std::string to_string(T value)
...
 RCLCPP_PUBLIC
 void
-init(int argc, char * argv[]);
+init(int argc, char const * const argv[]);
+
+/// Remove ROS-specific arguments from argument vector.
+/**
+ * Some arguments may not have been intended as ROS arguments. This function
+ * populates a the aruments in a vector. Since the first argument is always assumed
+ * to be a process name, the vector will always contain the process name.
```
wjwwood: Same style comment here, one line per sentence.

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -45,15 +46,40 @@ std::string to_string(T value)
...
+/**
+ * This does everything that init does, but additionally removes ROS-specific command line
+ * arguments
+ * \param[in] argc Number of arguments.
+ * \param[in] argv Argument vector.
+ * \return Members of the argument vector that are not ROS arguments.
+ */
+RCLCPP_PUBLIC
+std::vector<std::string>
+init_and_remove_ros_arguments(int argc, char const * const argv[]);
```
wjwwood: We could just change the signature of `init` to always return the pruned list, and users could just ignore it.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -191,6 +191,54 @@ rclcpp::init(int argc, char * argv[])
...
+std::vector<std::string>
+rclcpp::remove_ros_arguments(int argc, char const * const argv[])
+{
+  rcl_allocator_t alloc = rcl_get_default_allocator();
+  rcl_arguments_t parsed_args;
+
+  if (rcl_parse_arguments(argc, argv, alloc, &parsed_args) != RCL_RET_OK) {
+    std::string msg = "failed to parse arguments: ";
+    msg += rcl_get_error_string_safe();
+    rcl_reset_error();
```
wjwwood: There is a function for this: https://github.com/ros2/rclcpp/blob/947e3f7e677df87db4e4b76bc13c0b8d217fbeb1/rclcpp/include/rclcpp/exceptions.hpp#L100-L119

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -191,6 +191,54 @@ rclcpp::init(int argc, char * argv[])
 #endif
 }
 
+std::vector<std::string>
+rclcpp::remove_ros_arguments(int argc, char const * const argv[])
+{
+  rcl_allocator_t alloc = rcl_get_default_allocator();
+  rcl_arguments_t parsed_args;
+
+  if (rcl_parse_arguments(argc, argv, alloc, &parsed_args) != RCL_RET_OK) {
```
wjwwood: @sloretz was there going to be a function to get the already parsed arguments?

`rcl_init` already does this work and keeps the `rcl_arguments_t` around for later use. So we could use that here. The downside is that this would need `rcl_init` to have been called before, which might actually be a reason to leave this as-is.

Also, we'd need to ensure the `argc` and `argv` are the same ones given to the previously called `rcl_init`.

So with those two caveat's maybe this is the best approach.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: There are other opportunities to use this below too.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -191,6 +191,54 @@ rclcpp::init(int argc, char * argv[])
...
+      &parsed_args,
+      alloc,
+      &nonros_argc,
+      &nonros_argv) != RCL_RET_OK)
+  {
+    std::string msg = "failed to remove ROS arguments: ";
+    msg += rcl_get_error_string_safe();
+    rcl_reset_error();
+    if (NULL != nonros_argv) {
+      alloc.deallocate(nonros_argv, alloc.state);
```
wjwwood: Same question as on the other pr, I would expect this to not be required. I would expect `rcl_remove_ros_arguments` to clean up memory allocations in an error condition, so that it is not required here.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -191,6 +191,54 @@ rclcpp::init(int argc, char * argv[])
...
+    }
+    throw std::runtime_error(msg);
+  }
+
+  std::vector<std::string> return_arguments;
+  return_arguments.resize(nonros_argc);
+
+  for (int ii = 0; ii < nonros_argc; ++ii) {
```
wjwwood: I'd use `size_t` rather than `int`, unless you had a specific reason here.

rclcpp/test/test_utilities.cpp
```diff
@@ -0,0 +1,34 @@
...
+    "__ns:=/fiz/buz", "--foo=bar", "--baz"};
+  int argc = sizeof(argv) / sizeof(const char *);
+  rclcpp::init(argc, argv);
+  auto args = rclcpp::remove_ros_arguments(argc, argv);
+
+  ASSERT_EQ(4u, args.size());
+  ASSERT_EQ(std::string{"process_name"}, args[0]);
+  ASSERT_EQ(std::string{"-d"}, args[1]);
+  ASSERT_EQ(std::string{"--foo=bar"}, args[2]);
+  ASSERT_EQ(std::string{"--baz"}, args[3]);
```
wjwwood: It would be good to see a test for `init_and_remove_ros_arguments` also or instead.

Also, I'd like to see tests for corner cases, like `argc = 0` and/or `argv = nullptr`.

rclcpp/src/rclcpp/utilities.cpp
sloretz: @wjwwood You mean a function that gets the global parsed arguments? There is [`rcl_get_global_arguments`](https://github.com/ros2/rcl/blob/e48a445f46f1da21ff0c4bf3d52bec9120ab7b57/rcl/include/rcl/arguments.h#L181). Both your caveats apply.

rclcpp/src/rclcpp/utilities.cpp
sloretz: @wjwwood if it is `size_t` will windows warn about signed/unsigned comparison?

rclcpp/src/rclcpp/utilities.cpp
wjwwood: Oh yeah, forgot you ended up adding that. Any opinion on how this function should work, i.e. should it reused the result of init or do it again itself to avoid the corner cases?

rclcpp/src/rclcpp/utilities.cpp
wjwwood: That's a fair point, it might.

rclcpp/src/rclcpp/utilities.cpp
sloretz: As a standalone function with `argc` and `argv` passed in +1 to parsing the args again to prevent a mismatch between the arguments and the global state.

A non-public `rclcpp` function that made an `std::vector` of non-ros arguments given `argc`, `argv`, and `rcl_arguments_t` could be used by both `remove_ros_arguments` and  `init_and_remove_ros_arguments` to avoid the double-parsing. However, inefficiency in `init..()` is a one-time cost in most use cases so it may not be worth the added work.

wjwwood: Sounds good to me too. @mjcarroll I'll leave it up to you to either use the common function or not as @sloretz described.

mjcarroll: This issue would be OBE if we collapse init into a single signature that always returns a std::vector.

rclcpp/src/rclcpp/utilities.cpp
mjcarroll: Also MISRA, but that's not as important here.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4128/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1209/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3411/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4235/)

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
```
wjwwood: minimize vertical whitespace: https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -191,6 +193,57 @@ rclcpp::init(int argc, char * argv[])
...
+{
+  rclcpp::init(argc, argv);
+  return rclcpp::remove_ros_arguments(argc, argv);
+}
+
+std::vector<std::string>
+rclcpp::remove_ros_arguments(int argc, char const * const argv[])
+{
+  rcl_allocator_t alloc = rcl_get_default_allocator();
+  rcl_arguments_t parsed_args;
```
sloretz: Need to zero initialize this before calling `rcl_get_zero_initialized_arguments()`. This could be the source of the random segfault.

sloretz: Possible cause of #458

Issue #392, opened 2017-11-08
Single-threaded executor can be starved by timers


dhood:
**Context** (comes from https://github.com/ros2/demos/pull/187): A single-threaded executor has a timer scheduled for every N seconds, and also a service server.

**Bug**: If the timer callback takes >=N seconds to complete, the executor will never process service requests once the timer is triggered for the first time.

This is because of the combination of the following:

1. When the spinning executor calls to `get_next_executable`, [it does _not_ call `rcl_wait` since the timer's ready](https://github.com/ros2/rclcpp/blob/070b3125c1a7f76fbde3e31973d89a052a2ea652/rclcpp/src/rclcpp/executor.cpp#L546..L550), so the service is not marked as ready.
1. Even if you force `get_next_executable` to always call `wait_for_work` so the server can get receive its request, the server will still _not_ get processed by the executor because [the timer will always get chosen first as the `next_ready_executable`](https://github.com/ros2/rclcpp/blob/070b3125c1a7f76fbde3e31973d89a052a2ea652/rclcpp/src/rclcpp/executor.cpp#L517..L521).

Forcing `get_next_executable` to always call `wait_for_work` _and_ giving timers lower priority in `get_next_ready_executable` will fix this situation, but it is inefficient to wait when it's not necessary, and I'm not sure checking timers last is a fix-all (could there be a parallel situation when the server needs to be the lowest priority?).

@dirk-thomas mentioned that a queue of some sort is probably more appropriate, so that `get_next_executable` processes events in the order that they were received.

For now, we have to recommend that users not permit timer callbacks to block for longer than the duration at which they're scheduled (this might be in our documentation somewhere already?).

mikaelarguedas:
related to executor starvation reported https://github.com/ros2/rclcpp/issues/280
Right now we will match the behavior of rclpy and provide documentation as of how to make custom executors.

ubald:
:wave: I ended up here trying to find our why when I had multiple timers with short intervals (and longer callbacks) only the first one worked, the others are never executed. 

Is it possible that both issues are related?

wjwwood:
@ubald I do not believe what you've described is the issue here. This issue is essential that we do not have fair scheduling and so as long as timers are always ready then other things like Services will never get handled. However, more than "the first one" work in this case.

ubald:
Thanks for the quick reply!

[...]

PR #456, opened 2018-03-29
Fix for single threaded executor timer starvation condition

mjcarroll:
This is still a work in progress, but I wanted to capture the test case.

Connected to ros2/rclcpp#392


clalancette:
Closing this out, as the solution likely involves an executor redesign (being pursued elsewhere).

Issue #458, opened 2018-04-02
Access violation in new test_utilities test


dhood:
See for example https://ci.ros2.org/view/nightly/job/nightly_win_rep/1034/testReport/junit/(root)/projectroot/test_utilities/
```
-- run_test.py: invoking following command in 'C:\J\workspace\nightly_win_rep\ws\src\ros2\rclcpp\rclcpp':
 - C:/J/workspace/nightly_win_rep/ws/build/rclcpp/Release/test_utilities.exe --gtest_output=xml:C:/J/workspace/nightly_win_rep/ws/build/rclcpp/test_results/rclcpp/test_utilities.gtest.xml
-- run_test.py: return code 3221225477
```

RC 3221225477 is [access violation](http://www.febooti.com/products/automation-workshop/online-help/events/run-dos-cmd-command/exit-codes/), so there is likely a bug in the memory usage of this PR https://github.com/ros2/rclcpp/pull/454 or its counterpart in `rcl`

For reference, we have seen a similar issue in the past: https://github.com/ros2/build_cop/issues/19 and the fix boiled down to this PR: https://github.com/ros2/rmw_fastrtps/pull/106/files

FYI @mjcarroll; you may have the best idea of where to look for the issue if you get a chance.

mjcarroll:
Address Sanitize seems to indicate some issues, tracking down the source now.

PR #459, opened 2018-04-05
Correctly clean up arguments structure.

mjcarroll:
Clean up the arguments structure, which was previously leaking.

Connects to ros2/rclcpp#458

### DIFF
--- a/rclcpp/src/rclcpp/utilities.cpp
+++ b/rclcpp/src/rclcpp/utilities.cpp
@@ -252,7 +252,7 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
 
   ret = rcl_arguments_fini(&parsed_args);
   if (RCL_RET_OK != ret) {
-    exceptions::throw_from_rcl_error(ret, "failed to cleanup parsed arguments, leaking memory: ");
+    exceptions::throw_from_rcl_error(ret, "failed to cleanup parsed arguments, leaking memory");
   }
 
   return return_arguments;
--- a/rclcpp/src/rclcpp/utilities.cpp
+++ b/rclcpp/src/rclcpp/utilities.cpp
@@ -210,16 +210,7 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
 
   ret = rcl_parse_arguments(argc, argv, alloc, &parsed_args);
   if (RCL_RET_OK != ret) {
-    // Not using throw_from_rcl_error, because we may need to append deallocation failures.
-    exceptions::RCLErrorBase base_exec(ret, rcl_get_error_state());
-    rcl_reset_error();
-    if (RCL_RET_OK != rcl_arguments_fini(&parsed_args)) {
-      base_exec.formatted_message += std::string(
-        ", failed also to cleanup parsed arguments, leaking memory: ") +
-        rcl_get_error_string_safe();
-      rcl_reset_error();
-    }
-    throw base_exec;
+    exceptions::throw_from_rcl_error(ret, "failed to parse arguments");
   }
 
   int nonros_argc = 0;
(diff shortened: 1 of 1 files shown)


rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -210,7 +210,16 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
 
   ret = rcl_parse_arguments(argc, argv, alloc, &parsed_args);
   if (RCL_RET_OK != ret) {
-    exceptions::throw_from_rcl_error(ret, "Failed to parse arguments");
+    // Not using throw_from_rcl_error, because we may need to append deallocation failures.
+    exceptions::RCLErrorBase base_exec(ret, rcl_get_error_state());
+    rcl_reset_error();
+    if (RCL_RET_OK != rcl_arguments_fini(&parsed_args)) {
```
sloretz: It looks like `rcl_parse_arguments` [cleans up after itself](https://github.com/ros2/rcl/blob/d41c923927850274f63aee1bde565d3a8266a7bd/rcl/src/rcl/arguments.c#L303-L311) when something goes wrong. I would expect this call to `rcl_arguments_fini` to always fail [here](https://github.com/ros2/rcl/blob/d41c923927850274f63aee1bde565d3a8266a7bd/rcl/src/rcl/arguments.c#L432).

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -241,6 +250,11 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
     alloc.deallocate(nonros_argv, alloc.state);
   }
 
+  ret = rcl_arguments_fini(&parsed_args);
+  if (RCL_RET_OK != ret) {
+    exceptions::throw_from_rcl_error(ret, "failed to cleanup parsed arguments, leaking memory: ");
```
wjwwood: Is the `: ` needed in this case? I think this adds it automatically: https://github.com/ros2/rclcpp/blob/45dcd0c6e58bf54d771ae8635d30cc91f8a08e2d/rclcpp/src/rclcpp/exceptions.cpp#L60

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -224,10 +224,19 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
...
       alloc.deallocate(nonros_argv, alloc.state);
     }
-    exceptions::throw_from_rcl_error(ret, "Failed to remove ROS arguments: ");
+    if (RCL_RET_OK != rcl_arguments_fini(&parsed_args)) {
+      base_exec.formatted_message += std::string(
+        ", failed also to cleanup parsed arguments, leaking memory: ") +
+        rcl_get_error_string_safe();
+      rcl_reset_error();
+    }
+    throw base_exec;
```
dhood: I think our default behaviour is to throw an `RCLError` not an `RCLErrorBase`. I would change this to
```
throw RCLError(base_exc, "");
```
following 
 https://github.com/ros2/rclcpp/blob/c70f2f145239ca4dfe86c6d806f331602a0e5c68/rclcpp/src/rclcpp/exceptions.cpp#L71..L72

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -224,10 +224,19 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
     &nonros_argv);
 
   if (RCL_RET_OK != ret) {
+    // Not using throw_from_rcl_error, because we may need to append deallocation failures.
+    exceptions::RCLErrorBase base_exec(ret, rcl_get_error_state());
```
dhood: unless it's a convention that I haven't seen before, `exec` doesn't seem like an appropriate variable name (`base_exc` instead?)

PR #461, opened 2018-04-14
Add CLI args to Node constructor

sloretz:
This adds a parameter for command line arguments to `Node`, and a parameter to ignore global CLI arguments. I think roslaunch might need something like this to dynamically compose nodes without their arguments affecting other nodes, though I'm not sure if there is another strategy planned.

Kind of related to ros2/examples#200 and ros2/ros2#450

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -19,6 +19,7 @@
 
 #include "rclcpp/node_interfaces/node_base.hpp"
 
+#include "rcl/arguments.h"
 #include "rclcpp/exceptions.hpp"
 #include "rcutils/logging_macros.h"
 #include "rmw/validate_node_name.h"
@@ -31,7 +32,9 @@ using rclcpp::node_interfaces::NodeBase;
 NodeBase::NodeBase(
   const std::string & node_name,
   const std::string & namespace_,
-  rclcpp::Context::SharedPtr context)
+  rclcpp::Context::SharedPtr context,
+  const std::vector<std::string> & arguments,
+  bool use_global_arguments)
 : context_(context),
   node_handle_(nullptr),
   default_callback_group_(nullptr),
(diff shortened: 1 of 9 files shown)


sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4222/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1266/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3494/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4327/)

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -84,18 +87,33 @@ NodeBase::NodeBase(
   }
 
   // Create the rcl node and store it in a shared_ptr with a custom destructor.
-  rcl_node_t * rcl_node = new rcl_node_t(rcl_get_zero_initialized_node());
+  std::unique_ptr<rcl_node_t> rcl_node(new rcl_node_t(rcl_get_zero_initialized_node()));
```
sloretz: I changed this only to get rid of false positive "double free" error from cppcheck. It didn't like that it saw two `delete rcl_node;` statements: one in the error handling after calling `rcl_parse_arguments()`, and another in the error handling of `rcl_node_init()`.

sloretz:
checking if windows warning is fixed [Build Status](https://ci.ros2.org/job/ci_windows/4333/)

sloretz:
Remainging CI

Karsten1987:
I believe this PR introduced some memory leaks when `node_name` or `namespace_` are incorrect.
It primarily allocates memory which doesn't get freed when one of the following checks fails and raises an exception.

I ran valgrind on `test_node` and the following tests leaks twice the memory of the node arguments
https://github.com/ros2/rclcpp/blob/master/rclcpp/test/test_node.cpp#L37-L53
As the first test runs correctly, no memory is leaked. However, as the second and third node result in an exception, the node argument memory is leaked.

PR #463, opened 2018-04-16
pass AnyExecutable objects as reference to avoid memory allocation

wjwwood:
We've been trying to eliminate steady-state memory allocation, and this was identified as one of the easiest things we could change to avoid an unnecessary memory allocation and customization point.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4232/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1271/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3502/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4341/)

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -206,9 +206,13 @@ Executor::spin_some()
     throw std::runtime_error("spin_some() called while already spinning");
   }
   RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
-  AnyExecutable::SharedPtr any_exec;
-  while ((any_exec = get_next_executable(std::chrono::milliseconds::zero())) && spinning.load()) {
-    execute_any_executable(any_exec);
+  while (spinning.load()) {
+    AnyExecutable any_exec;
+    if (get_next_executable(any_exec, std::chrono::milliseconds::zero())) {
+      execute_any_executable(any_exec);
+    } else {
+      break;
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


rclcpp/src/rclcpp/executor.cpp
```diff
@@ -218,9 +222,11 @@ Executor::spin_once(std::chrono::nanoseconds timeout)
   if (spinning.exchange(true)) {
     throw std::runtime_error("spin_once() called while already spinning");
   }
-  RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
-  auto any_exec = get_next_executable(timeout);
-  if (any_exec) {
+  RCLCPP_SCOPE_EXIT({
+    this->spinning.store(false);
+  });
```
sloretz: Just curious, is this a bug fix or style fix? It looks like the macro already adds `{}` around its arguments.

wjwwood: Style only.

sloretz: For consistency, there's another use of the scope exit macro on line 208 that could have `{}` added to it.

wjwwood: I'll remove this diff and maybe make a separate pr for this change then, since its growing in scope :)

wjwwood:
The style undo is net zero diff, and the Windows warning is a false positive, so I'll merge as-is.

PR #439, opened 2018-02-17
rclcpp logging still uses fprintf all over the place.

guillaumeautran:
Remove all printf log lines and replace with RCLUTILS_LOG_XXX macros.

Issue: #438

PR #466, opened 2018-04-27
update to rcutils logging in commented code as well

mikaelarguedas:
Follow up of #439 

Otherwise we may forget to update to the logging macros if we decide to uncomment that piece of code

### DIFF
--- a/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
+++ b/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
@@ -359,9 +359,9 @@ public:
       } catch (const std::exception &) {
         // TODO(karsten1987): Windows CI doens't let me print the msg here
         // the todo is to forward the exception to the on_error callback
-        // fprintf(stderr, "Caught exception in callback for transition %d\n",
-        //  it->first);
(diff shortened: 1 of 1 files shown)


rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -359,9 +359,9 @@ class LifecycleNode::LifecycleNodeInterfaceImpl
       } catch (const std::exception &) {
         // TODO(karsten1987): Windows CI doens't let me print the msg here
         // the todo is to forward the exception to the on_error callback
-        // fprintf(stderr, "Caught exception in callback for transition %d\n",
-        //  it->first);
-        // fprintf(stderr, "Original error msg: %s\n", e.what());
+        // RCUTILS_LOG_ERROR("Caught exception in callback for transition %d\n",
```
dhood: I'd say this should use _NAMED for consistency with the other logging calls

mikaelarguedas: All logging calls in this file are not using _NAMED so I matched it

Issue #475, opened 2018-05-18
Expected vs Unexpected parameters Node::create_param()


sloretz:
## Feature request
Allow node author to define expected parameters and what happens when an unexpected parameter is set.

#### Feature description

This issue is a first step towards larger features like parameter constraints (allowed types, read-only parameters, allowed ranges, etc...). These will be ticketed separately. 

If a node does not use parameters:
  * Universal parameters like `use_sim_time` should still work
  * Reject attempts to set parameters to give feedback to run-time user

If a node uses parameters and knows all parameters it can use at compile time
  * Initial values might be needed if parameter is required and run-time user doesn't set it
  * Run-time user may set a new value for a known parameter provided it is a legal value
    * Future work: parameter constraints
    * For now: If a type changes then handle the update the same as an unexpected parameter

If a node uses parameters but does not know all parameters it can use at compile time
  * Initial values needed for known parameters that don't get set by a user
  * Node author needs ability to accept/reject parameters that are not known at compile time
    * Example:  `costmap_2d::ObstacleLayer` does not know all of the parameter names until after `observation_sources` is set.

#### Implementation considerations

Some parameters functionality will move to rcl (ros2/rcl#194). This ticket assumes the client library (`rclcpp`, `rclpy`, etc) will be responsible for parameter storage including default values and having knowledge of expected versus unexpected parameters.

##### Lifecycle considerations
If a node author wants custom logic for unexpected parameters then that logic must be set up before a parameter can be set by a run-time user. This includes defining expected parameters. Parameters cannot be set until a node is added to an executor, and that can't happen until a node is constructed.  Therefore, a node author may define parameters and set up unexpected parameter handling in their constructor.

##### Creating a parameter
```c++
/// Create a parameter with an optional initial value
/*
 * \param[in] initial_value the name and initial value of the parameter
 * \raise an exception if an error occurs
 */
void
NodeParameters::create_parameter(const ParameterVariant & initial_value);

void
NodePamameters::create_parameter(const std::string & name)
{
  /* TODO it does not appear possible to create a ParameterVariant with a name but no type */
}

// Convenience call
template <typename T>
void
NodePamameters::create_parameter(const std::string & name, T value)
{
  ParameterVariant param(name, value);
  return create_parameter(param);
}
```

##### Deleting/undeclaring a parameter
```c++
/// Undeclare parameter with a given name
/* Deleted parameters are treated as unexpected the next time a user tries to set them
 * \param[in] name the name of the parameter to declare
[...]

PR #477, opened 2018-05-23
[WIP] Initialize parameters

sloretz:
Work in progress towards initializing parameters.
## Current state:
When the node is constructed yaml files passed as CLI args are parsed, and the parameters within are set on the node. 

## RCL
- [x] need yaml parameter parser (ros2/rcl#235)
- [x] need `rcl` function to get yaml file name from command line arguments (ros2/rcl#253)

## Questions/Assumptions
* Assuming `rclcpp` needs to call methods to parse yaml because `rcl` doing it would cause a circular dependency between it and `rcl_yaml_parse_params`
* ~~Assuming `rcl` will have a method for getting a path to a yaml file from an `rcl_arguments_t`~~ Has method for getting all yaml paths specified
* ~~Does `rcl_yaml_param_parser` dependency need to be exported?~~ Yes
* Should `rcl_yaml_param_parser` namespace + nodename include a leading slash? currently it does not
* What was the idea of `ParameterInitializationMap` for? Why not `std::vector<ParameterVariant>`?

## TODO
Bare minimum:
- [x] `NodeParameters` should start parameter services in its constructor (#478)
  * Need to pass full `Node` instance to `NodeParameters` to give to `ParameterService`
- [ ] `NodeParameters` needs to get yaml file from CLI arguments (#488)
  1. If `rcl_arguments_t` passed to node options is not zero initialized, then get yaml files from it
  2. if `node options `use_global_arguments` is `True` then get yaml files from global arguments.

Desired:
- [x] Split `ParameterVariant` into `Parameter` which has a `ParameterValue` (#481)
- [x] `rclcpp` method to convert `rcl_params_t` to `std::vector<Parameter>` (or `std::map<string, ParameterValue>` or `std::map<string, Parameter>` ? ) (#485)
- [x] Allow passing `std::vector<Parameter>` into `NodeParameters` constructor. (#486)
    * Needed to initialize a node with parameters generated by roslaunch after it merges multiple yaml files together.
- [ ] `node::create_parameter(name, value)`
   * Declares a parameter is known by a node
- [ ] `std::vector<Parameter> node::unknown_initial_parameters()`
   * Returns a list of parameters that are in the yaml file but not known to the node
   * To be called by a user after all parameters have been created
- [ ] `void node::raise_if_unknown_initial_parameters()`
   * Raises exception if there is an unknown initial parameter
   * To be called by a user after all parameters have been created

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -6,6 +6,7 @@ find_package(ament_cmake_ros REQUIRED)
 find_package(builtin_interfaces REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rcl_interfaces REQUIRED)
+find_package(rcl_yaml_param_parser REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rmw_implementation REQUIRED)
 find_package(rosgraph_msgs REQUIRED)
@@ -99,6 +100,7 @@ add_library(${PROJECT_NAME}
 # specific order: dependents before dependencies
 ament_target_dependencies(${PROJECT_NAME}
   "rcl"
+  "rcl_yaml_param_parser"
   "builtin_interfaces"
   "rosgraph_msgs"
   "rosidl_typesupport_cpp"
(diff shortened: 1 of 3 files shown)


sloretz:
I should have put this info into an issue instead of a PR. The content is covered by #488 and #475. Closing.

PR #481, opened 2018-05-29
Split ParameterVariant into Parameter and ParameterValue

sloretz:
In progress while CI runs

This PR looks large, but most of the changed lines in this PR are due to renaming `ParameterVariant` to `Parameter`. I can split it into smaller PRs if need be.

Changes
* Renamed `ParameterVariant` to `Parameter`
* Removed `rclcpp::parameter` namespace (as was done for other namespaces in #416)
* Moved functions related to getting a parameter value into new `ParameterValue` class.
* Raise `ParameterTypeException` instead of `std::runtime_error` when there is a type mismatch
* Rename `Parameter::get_parameter_value()` to `Parameter::get_value_message()`
* Added `const` to `Parameter::to_parameter()`

PR #486, opened 2018-05-31
Pass initial parameter values to node constructor

sloretz:
~~PR complete, waiting for #481 to be merged first, then this needs to be targeted at `master`.~~

This allows the initial parameter values for a node to be passed in via it's constructor. These parameters are then immediately set on the node.

~~edit: invoked CI wrong, will wait to run until #481 is in.~~

blocks ros2/rclcpp#477
requires ros2/rclcpp#481


sloretz:
CI with 28ad77b
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4514/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1434/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3697/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4589/)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -50,6 +51,15 @@ NodeParameters::NodeParameters(
     rmw_qos_profile_parameter_events,
     use_intra_process,
     allocator);
+
+  // TODO(sloretz) store initial values and use them when a parameter is created
```
dhood: can you explain this TODO a bit more for my understanding please? The impression I get is that one day we want the node to store a list of possible initial parameter values, but not actually set those parameters until a different step?

What we are doing here -- which at the moment is essentially passing in a list of initial parameters -- is not the end goal? Why not?

I have seen in the parent ticket that this is "Needed to initialize a node with parameters generated by roslaunch after it merges multiple yaml files together." but it wasn't enough for me to answer my own question

sloretz: > The impression I get is that one day we want the node to store a list of possible initial parameter values, but not actually set those parameters until a different step?

Yeah, the TODO says what should happen here when (#475) is implemented. The reason for having a different step is to allow a node author to define parameters, including things like a `read_only` parameter that can be initialized but not modified. The parameter shouldn't be set until a node-author finishes defining parameters so that it's possible to error when a run-time user has a typo in their configuration.

> What we are doing here -- which at the moment is essentially passing in a list of initial parameters -- is not the end goal? Why not?

Passing in parameters via constructor will still exist, but setting them right away instead of storing them is the back-up plan. If the whole feature doesn't get in by bouncy release then this at least allows run-time users to initialize parameters via the command line.

rclcpp/test/test_node_initial_parameters.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "rclcpp/rclcpp.hpp"
+
+class TestNodeWithInitialValues : public ::testing::Test
+{
+protected:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, NULL);
+  }
+};
```
mikaelarguedas: Nit: I know other tests don't all do it but it seems cleaner to have a tear down calling `rclcpp::shutdown` rather then just exiting the process

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
dhood: OK, 475 seems to be the missing piece in the explanation puzzle, thanks for explaining

rclcpp/include/rclcpp/node.hpp
```diff
@@ -96,6 +96,7 @@ class Node : public std::enable_shared_from_this<Node>
     const std::string & namespace_,
     rclcpp::Context::SharedPtr context,
     const std::vector<std::string> & arguments,
+    const std::vector<Parameter> & initial_values,
```
dhood: In my opinion, as a user, `initial_values` is not descriptive enough; I think we're relying on the type to convey extra information. When I first looked at this PR I thought "initial values for what?" (and then as you saw, I wondered why it wasn't just called "initial parameters", especially since it sets the names as well). Were there other names that you considered?

rclcpp/test/test_node_initial_parameters.cpp
sloretz: Added call to `shutdown()` in 04cb6a9d061fb3a7b71f32763ce82185600bf3cf

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
sloretz: Referenced issue in TODO comment f30c64cb494351387813395c85e64f40ae3ee4ba

rclcpp/include/rclcpp/node.hpp
sloretz: > Were there other names that you considered?
There were not

changed to `initial_paramters` plus added missing doxygen `\param[in] initial_parameters` in 29046af5d29a085c4edb81a698bdac9124978823

sloretz:
CI as of ad83f5a

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4540/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1453/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3727/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4615/)

Issue #490, opened 2018-06-04
Impossible to tell which parameters were gotten if some don't exist


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  -  source
- Version or commit hash:
  - 2cb890e1f235af3ba110dc0e44eb30f50a7f2285
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
ros2 run demo_nodes_cpp listener
```

In another terminal
```
ros2 param set listener bool_param1 true
ros2 param set listener bool_param2 true
ros2 service call /listener/get_parameters rcl_interfaces/GetParameters "{'names': ['bool_param1', 'does_not_exist', 'bool_param2']}"
```

#### Expected behavior

I would expect the returned output from the service call to indicate which parameter did not exist

#### Actual behavior

Response to service call has 2 values, but no indication of which of the 3 requested parameters were returned.

```
rcl_interfaces.srv.GetParameters_Response(values=[rcl_interfaces.msg.ParameterValue(type=1, bool_value=True, integer_value=0, double_value=0.0, string_value='', byte_array_value=[], bool_array_value=[], integer_array_value=[], double_array_value=[], string_array_value=[]), rcl_interfaces.msg.ParameterValue(type=1, bool_value=True, integer_value=0, double_value=0.0, string_value='', byte_array_value=[], bool_array_value=[], integer_array_value=[], double_array_value=[], string_array_value=[])])
```

#### Additional information

The cause seems to be this code that only adds a result if the parameter exists. One solution might be to add an else case adding a parameter value with type `NOT_SET` to the result.

https://github.com/ros2/rclcpp/blob/2a17232ad0c5b54fabea0bc10f4575428f7c7227/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L126-L132

dirk-thomas:
> One solution might be to add an else case adding a parameter value with type `NOT_SET` to the result.

Based on the documentation of the service interface that is exactly what needs to happen: https://github.com/ros2/rcl_interfaces/blob/8e826d4342a70a17358a27bcf384f695ad32cefd/rcl_interfaces/srv/GetParameters.srv#L7

PR #493, opened 2018-06-06
Get parameters that aren't set

sloretz:
Parameters that are queried but don't exist on the node should return a value with type NOT_SET to comply with the service definition in `rcl_interfaces`.

This PR only changes the behavior of the `get_parameters` service. The C++ API doesn't have the same problem (it returns both the name and the value of the parameter), so this PR adds documentation to it instead.

Fixes ros2/rclcpp#490
connects to ros2/rclcpp#490

### DIFF
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
@@ -49,16 +49,35 @@ public:
   set_parameters_atomically(
     const std::vector<rclcpp::Parameter> & parameters) = 0;
 
+  /// Return descriptions of parameters given their names.
+  /*
+   * \param[in] names a list of parameter names to check.
+   * \return the list of parameters that were found.
+   * Any parameter not found is omitted from the returned list.
+   */
   RCLCPP_PUBLIC
   virtual
   std::vector<rclcpp::Parameter>
   get_parameters(const std::vector<std::string> & names) const = 0;
 
+  /// Return the description of one parameter given a name.
+  /*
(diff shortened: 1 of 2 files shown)


sloretz:
Keeping In progress label, while CI runs

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4551/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1462/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3736/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4632/)

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
```diff
@@ -49,16 +49,35 @@ class NodeParametersInterface
...
+   * \param[in] name the name of the parameter to look for.
+   * \return the parameter if it exists on the node.
+   * \throws std::out_of_range if the parameter does not exist on the node.
+   */
   RCLCPP_PUBLIC
   virtual
   rclcpp::Parameter
   get_parameter(const std::string & name) const = 0;
 
+  /// Return the description of one parameter given a name.
```
mikaelarguedas: Copy-n-paste ?

sloretz: Changed "Return" to "Get" in 886cf2a53c727d7e7dc775f4e1939d1eed9ac891.

sloretz:
Assuming 886cf2a53c727d7e7dc775f4e1939d1eed9ac891 won't break CI since it only shortens comments, and merging.

PR #488, opened 2018-06-01
[WIP] Init params via yaml

sloretz:
Blocks ros2/rclcpp#477
Requires ros2/rclcpp#485
Requires ros2/rclcpp#486
Requires rcl branch `yaml_param_arguments`

This initializes the parameters on a node from yaml files. There are multiple sources of initial values passed to a node. A user can specify multiple yaml files on the command line, and a node can have its own specific arguments as well as explicit `initial_values`. This PR merges all the sources such that `initial_values` overwrites local arguments which overwrites global arguments. Yaml files specified later overwrite ones specified earlier.

i.e. this works
```
ros2 run demo_nodes_cpp talker __params:=generic-default-values.yaml __params:=more-specific-values.yaml
```

The only commit is 2cb890e. This PR needs to be rebased when #485 and #486 are merged.

Issue #296, opened 2016-12-22
Support for multiple parameter change callbacks


rohbotics:
This will be long, and might be better suited for the discourse, let me know.

Currently rclcpp only supports registering one `param_change_callback`, and if you try to register a second one, it overwrites the first callback. 

The problem with having multiple callbacks is supporting atomically setting parameters. Currently the callbacks that are registered do two things, verify that the parameters are within any constraints, as well as update the operation of the node to work with the new parameters (if verification passed). With only one callback per node, this all works fine, and the parameters are updated atomically across the node.
(Ex of current callback: https://github.com/rohbotics/turtlebot2_demo/blob/0f33a3c3caadd13fe89be060690d6e4b797cd2ba/turtlebot2_drivers/src/kobuki_node.cpp#L42) 

With multiple callbacks, this approach doesn't work any more, the parameters can only be atomic per registered callbacks. Here are some proposals I have to fix this. 

## A: Namespacing
Figure out how we are going to namespace parameters within the node, and then only support atomic operations on parameters within a single namespace, and allow for one callback per namespace. 

Pros: Easy to interface with at an API level. 
Cons: Pretty restrictive, could be hard to implement cleanly/correctly

## B: Seperation of Concerns
After discussing with @tfoote, this is the approach that we came up with. The user will provide two functions, one that only verifies the parameter updates, which provides another function that will actually execute updating the operation of the node with the new parameters. This could be made easier with dynamic-configure ish code generation. 

Pros: Keeps atomicity support everywhere
Cons: Could be messy from a user perspective (without code gen)

## C: Make Atomicity Optional
Not every node requires support for atomically setting a group of variables, so make support for it optional. For nodes that want support they can use approach B, and those who don't can use basic callbacks that verify and set parameters. 

Pros: Easy to add parameters without worrying about atomicity
Cons: Might make nobody bother with atomic parameter updates

------------------------------------------------
Footnote: there should also be a way to de-register callbacks if we support multiple

wjwwood:
Thanks for starting the discussion @rohbotics, at some point this might belong on the parameter's design doc (once we have somewhat of a consensus and a list of the possible options).

---

I think option B is the most flexible but also the most complicated to use. However, I am actually leaning towards A. I think A becomes necessary anyways to allow group verification of some parameters but not others.

You could also implement A+B, where the process is still split into pre/verification and post parameter change (B) but also support registering callbacks to parts of the namespace (A).

[...]

PR #495, opened 2018-06-08
Read only parameters

sloretz:
This adds `Node::create_parameter()` and a boolean `allow_undeclared_parameters` to the node's constructor. Parameters may be given a type, a default value, and a bool indicating whether or not they're read only. Read only parameters are initialized, but may not be changed after initialization.

The features work except for a test failure in `test_time_source.cpp`, and I have no idea why. The failure symptom is the `on_parameter_event` callback in `TimeSource` does not get called.

The time_source test passes if I step through the problem area in GDB. That is, I set a breakpoint at the start of the failing test then set a break point inside `TimeSource::on_parameter_event`. If I hit 'c' for continue the test fails and the breakpoint is not reached. If I instead hit `n` repeatedly then the breakpoint is reached in the `spin()` call and the test passes.

There are a couple missing features. Since this PR adds the concept of declaring parameters, it should be possible for a node to error if a run-time user tries to set an undeclared parameter. This could be done with a method that throws if any undeclared parameters where provided, but a node-author would be responsible for calling this method after all parameters have been declared.

Other nodes may allow undeclared parameters (like the observation source parameters in obstacle layer in costmap 2d). If they pass `allow_undeclared_parameters = true` to the constructor then the initial parameters provided by a run-time user are immediately set on a node. A future enhancement could let a node say parameters starting with `foo.bar.` may be undeclared.

requires ros2/rclcpp#488


dejanpan:
@sloretz @mikaelarguedas could we do anything to get this picked up again and merged?

@anup-pem

wjwwood:
I think I got this rebased correctly (I'll be building on this now at least), but some of the merges were not obvious due to recent changes and since it was based on #488 but not rebased on it before it was merged to master.

dejanpan:
@wjwwood: should @anup-pem and I be reviewing this PR?

wjwwood:
No, it's not ready for review yet (I will put the in review label and comment when that happens).

wjwwood:
@sloretz could I trouble you to amend and then force push your commits on this branch, adding a sign-off while you do it? If you'd rather not then I can merge this with the DCO check failing.

wjwwood:
Also, this pr and connected pr's are ready for review. I will have follow up pr's after it is merged.

sloretz:
> @sloretz could I trouble you to amend and then force push your commits on this branch, adding a sign-off while you do it?

I added signed-off-by to those commits and pushed to branch [read_only_parameters_shane_dco](https://github.com/ros2/rclcpp/tree/read_only_parameters_shane_dco). Mind double checking that I did it right? If it looks ok then this branch can be reset to that one.

wjwwood:
It lgtm @sloretz, thanks.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -144,25 +146,19 @@ NodeParameters::NodeParameters(
...
-    combined_values.emplace_back(kv.second);
+    initial_parameter_values_[param.get_name()] =
+      rclcpp::ParameterValue(param.get_value_message());
   }
 
-  // TODO(sloretz) store initial values and use them when a parameter is created ros2/rclcpp#475
-  // Set initial parameter values
-  if (!combined_values.empty()) {
-    rcl_interfaces::msg::SetParametersResult result = set_parameters_atomically(combined_values);
+  if (!initial_parameters.empty() && allow_undeclared_) {
```
clalancette: In the case where `!initial_parameters.empty()` and `!allow_undeclared_`, it seems like we are going to silently ignore the values coming from YAML.  Is that intentional?  Should we throw an exception instead?

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -256,6 +271,8 @@ class NodeOptions
 
   rmw_qos_profile_t parameter_event_qos_profile_ {rmw_qos_profile_parameter_events};
 
+  bool allow_undeclared_parameters_ {false};
```
clalancette: Disallowing undeclared parameters by default is going to make a lot of people unhappy when they switch to Dashing, since they have had no chance to change to the "declared parameters" world before this.  It seems like we should allow undeclared parameters by default in Dashing (possibly with a warning), then maybe switch this to `false` in E-turtle.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -258,6 +258,36 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * \throws std::runtime_error if parameter has already been declared.
+   * \throws std::runtime_error if a parameter name is invalid.
+   * \throws rclcpp::exceptions::InvalidParameterValueException if initial value fails to be set.
+   */
+  RCLCPP_PUBLIC
+  void
+  declare_parameter(
+    const std::string & name,
+    const rclcpp::ParameterValue & default_value = rclcpp::ParameterValue(),
+    bool read_only = false);
```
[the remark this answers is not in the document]
wjwwood: It will need to either evolve to include more arguments or a structure that defines the meta data for the parameter. I don't know which is better right now, so that's why I didn't change it, and presumably why @sloretz started it this way.

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -213,6 +213,21 @@ class NodeOptions
...
+
+  /// Set the allow_undeclared_parameters, return this for parameter idiom.
+  /**
+   * If true, allow any parameter name to be set on the node without first
+   * being declared.
+   * Otherwise, setting an undeclared parameter will raise an exception.
+   */
+  RCLCPP_PUBLIC
+  NodeOptions &
+  allow_undeclared_parameters(const bool & allow_undeclared_parameters);
```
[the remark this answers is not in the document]
wjwwood: Nothing, I was just making it the same as the others. On the other hand it doesn't cost anything either. I can change it if you want.

rclcpp/include/rclcpp/parameter.hpp
```diff
@@ -28,6 +28,25 @@
...
+}  // namespace node_interfaces
+
+namespace detail
+{
+
+// This helper function is required because you cannot do specialization on a
+// class method, so instead we specialize this template function and call it
+// from the unspecialized, but dependent, class method.
+template<typename T>
+auto
```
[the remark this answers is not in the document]
wjwwood: I don't know why auto return value is inherently bad, but I don't know of another way to avoid it. @sloretz wrote this part I believe, but I think it's correct.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
wjwwood: It is intentional, I have a follow up pr where the user can choose what to do with "extra" parameters in the YAML which are not declared in the source code. As well as they can run a checker for unused YAML parameters (i.e. raise an error if any extra YAML parameters are unused). The latter cannot be done automatically because we don't know when the user will be done declaring parameters, so they have to invoke the "checker".

This branch has (some of) my WIP:

https://github.com/ros2/rclcpp/commit/36ca81381355247ac03c3b2f0bd787422655a91e#diff-b2109335f72aba8ba2975b698da15ebd

rclcpp/include/rclcpp/node_options.hpp
wjwwood: I'd be willing to do that, though it will make all of this more complicated because this goes from a bool to a choice (allow, disallow, allow but do deprecation warning).

On the flip side, they can simply create their nodes in dashing with a different value for this option, so the impact is relatively small. Or they can do more work, i.e. declare their parameters, which has more long term benefits.

Anyone else have an opinion on this?

wjwwood:
@sloretz I can't figure out what GitHub's issue is with displaying this correctly. It looks different locally with `git-log`. Also, DCO is still failing...

I think I'm going to rebuild it by cherry-picking each commit, one at a time.

sloretz:
> @sloretz I can't figure out what GitHub's issue is with displaying this correctly. It looks different locally with `git-log`. Also, DCO is still failing...

I think I've seen github do this before. IIRC changing the target branch away and back to master fixed the UI. No idea why the DCO is failing though.

sloretz:
> Also, DCO is still failing...


Oops, it's unhappy about the email in the sign off. I'll fix that
```
Commit sha: f9707db, Author: Shane Loretz, Committer: Shane Loretz; Expected "Shane Loretz sloretz@osrfoundation.org", but got "Shane Loretz sloretz@openrobotics.org".
```

sloretz:
@wjwwood Fixed email for DCO check and pushed to branch [read_only_parameters_shane_dco](https://github.com/ros2/rclcpp/tree/read_only_parameters_shane_dco) again

wjwwood:
Ok, I think I sorted it out, I think it was a combination of issues related to the difference in commit order versus commit date.

rclcpp/include/rclcpp/node_options.hpp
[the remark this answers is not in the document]
wjwwood: 46c207f

rclcpp/include/rclcpp/parameter.hpp
```diff
@@ -142,6 +166,49 @@ RCLCPP_PUBLIC
...
+  return parameter->get_parameter_value();
+}
+
+// Specialization allowing Parameter::get() to return a const ref to the parameter itself.
+template<>
+inline
+auto
+get_value_helper<rclcpp::Parameter>(const rclcpp::Parameter * parameter)
+{
+  // Use this labmda to ensure it's a const reference being returned (and not a copy).
```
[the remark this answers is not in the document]
wjwwood: 8e28200d63518dd20e06545614600a15debc998a

wjwwood:
Ok, I think this is ready for re-review from @clalancette or anyone else that wants to chime in on the above comments.

rclcpp/include/rclcpp/parameter.hpp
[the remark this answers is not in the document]
clalancette: I wouldn't say "inherently bad", but it does make it more difficult for the reader later to know which type is returned (the most common place I run into this is when trying to make a `std::vector` of what's returned here, and I need to go grubbing around in the source to find what is actually returned).  So generally I prefer to only use auto when the type is glaringly obvious.

rclcpp/include/rclcpp/node.hpp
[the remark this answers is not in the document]
clalancette: I'm mostly worried about having to change the API later to add those additional features; if we don't do it now, then it will be an API break later and will be harder to do.  Personally, it sounds like there are enough additional features we'd want to add that it would warrant adding a structure here.  Thoughts?

rclcpp/include/rclcpp/node_options.hpp
clalancette: > On the flip side, they can simply create their nodes in dashing with a different value for this option, so the impact is relatively small. Or they can do more work, i.e. declare their parameters, which has more long term benefits.

True.  My point of view is coming from code that I'm going to need to port from Crystal to Dashing, and with this change I know things won't work out of the box.  I think there are 2 questions:

1.  What do we *eventually* want the default state of `allow_undeclared_parameters` to be?
1.  How do we get there?

The first question basically comes down to flexibility vs. reliability.  Right now (and in ROS 1) we're basically on the side of flexibility, but switching this to false gets us closer to reliability.  I'm OK with that, but it is a change from what we've traditionally done.

For the second question we should consider whether we want to start tick-tocking for a change like this, or whether we should just add it to the porting guide.

Putting it that way, I'd be fine with either tick-tocking with a warning in Dashing and switching to false in E-Turtle, or changing this to false and adding some notes to the Dashing porting guide.

That being said, I'd like to hear other opinions on this change.

rclcpp/include/rclcpp/parameter.hpp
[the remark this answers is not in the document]
wjwwood: Maybe the return type could be `T`? I'm not sure, I think the auto here is to avoid redundancy with the function it calls (it's more of a forwarding of template arguments). Like if the called function (in the implementation of this function) returns int64_t when T = int, then you'd want to avoid the logic needed to do that here as well, therefore auto.

I'd prefer not to muck with this anymore, as I had to do some stuff with `decltype(auto)` and things like that to make it work.

It would all be a lot simpler with constexpr-if, though I'd still need auto return type in that case.

rclcpp/include/rclcpp/node.hpp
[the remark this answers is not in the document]
wjwwood: I thought about this at the time, and I couldn't come up with how constraints might work (like must be in this range, or one of these choices, etc.) which put me off from deciding they should all live in a single struct, but I guess no matter what we come up with we can fit it in an options struct or "description"/"descriptor" struct, since that's the matching IDL equivalent (`ParameterDescriptor.msg`).

rclcpp/include/rclcpp/node_options.hpp
dirk-thomas: I would argue that the default of `allow_undeclared_parameters` should be `true` since that matches past behavior (in ROS 2 as well as ROS 1). To consider the opposite a clear argument should be made why that should be the preferred default.

sloretz: Whether or not `allow_undeclared_parameters` is the default, I think any node with a fixed set of parameters should set it to `false` and declare them.

If undeclared parameters are allowed then misspellings of parameters can only be caught if the node author explicitly checks them. If only declared parameters are allowed, the node author still has to check them after declaring their parameters, but they could use a common function that asserts no undeclared parameters remain in the initial parameter list.

If a user interacting with a node is partially trusted, that is they may change parameters but should not be able to bring the system down then undeclared parameters are a potential attack. A malicious user could set many parameters until the node runs out of memory. A node author can't know if someone running their node is going to have a use case with partially trusted users, so if allowing undeclared parameters is the default then someone with this use case must modify existing nodes to declare their parameters.

wjwwood: I basically agree with @sloretz, it's safer against both typo's (misspelled parameter names) and denial of service attacks (intentional or accidental). It also has implications for safety, in that if you only allow declared parameters, you're unlikely to use more memory than you expect or something like that (declared parameters can be pre-allocated at some level).

wjwwood: That's the argument for making it true at some point, separately, I'm fine with doing a tick-tock roll out if people are concerned about disruption to Crystal users that are upgrading. I defaulted to just changing it because we hadn't been doing backwards compatible changes in other cases (like the `NodeOptions` pr). I figured documenting it in the release notes would be sufficient.

wjwwood: Summarizing from ROS 2 meeting:

- long term we do want to require parameters to be declared
- we will not do the tick-tock, because it would require an immediately deprecated option (allow undeclared but warn) and because the burden of setting one flag in the `NodeOptions` is small if migrating users do not want to deal with declaring parameters

I will rework the tests and documentation to be clearer about this, and then I will draft a pr to the dashing migration notes that is as thorough as possible.

wjwwood:
I still have some loose ends to tie up, like documentation for list_parameters, and some more tests for `set_parameters_atomically` and the variations of `get_parameter(s)`, which I have locally, but need polish.

I also need to track down any new deprecation warnings in higher packages.

But I think the documentation that's been added, the tests that go with it, and the implementation are ready for review (again).

rclcpp/include/rclcpp/node.hpp
```diff
@@ -258,133 +258,584 @@ class Node : public std::enable_shared_from_this<Node>
...
   rcl_interfaces::msg::ListParametersResult
   list_parameters(const std::vector<std::string> & prefixes, uint64_t depth) const;
 
+  using OnParametersSetCallbackType =
+    rclcpp::node_interfaces::NodeParametersInterface::OnParametersSetCallbackType;
+
+  /// Register a callback to be called anytime a parameter is about to be changed.
+  /**
+   * The callback signature is designed to allow handling of any of the above
+   * `set_parameter*` methods, and so it takes a const reference to a vector
```
jacobperron: or `declare_parameter*` methods.

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
```diff
@@ -42,12 +42,49 @@ class NodeParametersInterface
...
+   * \sa rclcpp::Node::set_parameters
+   */
   RCLCPP_PUBLIC
   virtual
   std::vector<rcl_interfaces::msg::SetParametersResult>
-  set_parameters(
-    const std::vector<rclcpp::Parameter> & parameters) = 0;
+  set_parameters(const std::vector<rclcpp::Parameter> & parameters) = 0;
 
+  /// Declare and initialize a parameter, all at once.
```
jacobperron: The Node documentation for this function does not state that parameters are declared. Maybe this should read "Set one or more parameters, all at once."?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -258,133 +258,584 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * \param[out] values The map used to store the parameter names and values,
+   *   respectively, with one entry per parameter matching prefix.
+   * \returns True if output "values" was changed, false otherwise.
+   * \throws rclcpp::ParameterTypeException if the requested type does not
+   *   match the value of the parameter which is stored.
+   */
+  template<typename ParameterT>
+  bool
+  get_parameters(
+    const std::string & prefix,
```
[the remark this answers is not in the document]
wjwwood: Yeah, this is a bit wonky, because you can do something like `foo.b` to match `foo.bar` and `foo.baz`... The addressing of parameters needs to be cleaned up further and the terminology needs to be clearer, but I didn't have time to follow through on that.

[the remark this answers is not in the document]
wjwwood: Sorry, didn't complete my thought... So that means in this case it's a prefix rather than a namespace technically, though a namespace is also technically a kind of prefix, and this function will remove any leading namespace separator as I described in the docs. But at this point I'm just trying to document existing behavior, I don't think this is the best way to have all this work...

rclcpp/include/rclcpp/node.hpp
```diff
@@ -258,133 +258,584 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * The name and type in the given rcl_interfaces::msg::ParameterDescriptor
+   * are ignored, and should be specified using the name argument to this
+   * function and the default value's type instead.
+   *
+   * This method, if successful, will result in any callback registered with
+   * set_on_parameters_set_callback to be called.
+   * If that callback prevents the initial value for the parameter from being
+   * set then rclcpp::exceptions::InvalidParameterValueException is thrown.
+   *
+   * \param[in] name the name of the parameter
```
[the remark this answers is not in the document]
wjwwood: I went through and tried to do this everywhere.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -258,133 +258,584 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * be made to set the parameters which come after.
+   *
+   * If a parameter fails to be set due to any other reason, like being
+   * rejected by the user's callback (basically any reason other than not
+   * having been declared beforehand), then that is reflected in the
+   * corresponding SetParametersResult in the vector returned by this function.
+   *
+   * This method will result in any callback registered with
+   * set_on_parameters_set_callback to be called, once for each parameter.
+   * If the callback prevents the parameter from being set, then it will be
```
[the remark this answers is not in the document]
wjwwood: I didn't remove the line, as above I feel this case is being used as an example (one of a few things that might happen), whereas this is this paragraph is specifically talking about this behavior.

I did reword it slightly, please have a look.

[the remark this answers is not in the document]
jacobperron: :+1:

rclcpp/test/test_node_initial_parameters.cpp
```diff
@@ -42,28 +42,93 @@ TEST_F(TestNodeWithInitialValues, no_initial_values) {
...
+  node->list_parameters({}, 0);
   EXPECT_TRUE(node->get_parameter("foo").get_value<bool>());
   EXPECT_STREQ("hello world", node->get_parameter("bar").get_value<std::string>().c_str());
   std::vector<double> double_array = node->get_parameter("baz").get_value<std::vector<double>>();
   ASSERT_EQ(2u, double_array.size());
   EXPECT_DOUBLE_EQ(3.14, double_array.at(0));
   EXPECT_DOUBLE_EQ(2.718, double_array.at(1));
 }
+
+TEST_F(TestNodeWithInitialValues, undeclared_initial_values_throw) {
```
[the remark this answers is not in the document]
wjwwood: Sorry, I sort of rewrote all the tests and put them all in `test_node.cpp`. I plan on removing this file, but I was just keeping it around so I could go through it once more to make sure I didn't miss an existing case.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -258,133 +258,584 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * callback is returned when this method is used, or nullptr will be returned
+   * if no callback was previously set.
+   *
+   * \param[in] callback The callback to be called when the value for a
+   *   parameter is about to be set.
+   * \return The previous callback that was registered, if there was one,
+   *   otherwise nullptr.
+   */
+  rclcpp::Node::OnParametersSetCallbackType
+  set_on_parameters_set_callback(rclcpp::Node::OnParametersSetCallbackType callback);
```
wjwwood: I’d like feedback on this function (which sort of already exists, see `register_param_change_callback` below) that I’m adding/refactoring as part of this read-only parameter refactor.

My main concern that there will be times when two pieces of code that do not know about each other both need to influence this behavior using this callback. The first concrete example of this is that I’d like to use this callback for checking the type of the `use_sim_time` parameter (which is implicitly declared by each node) before setting it, but if I did that then a user could easily override that callback when they set their own and ignore the return value, breaking the behavior of `use_sim_time`.

Alternatives are that I can simply push back callbacks onto a vector and call of them in the order they were added, or I could instead setup a feature where each parameter can have a single callback in addition to the proposed function (one callback for all parameters). I think the cases where you need to have more than one callback for a single parameter are low, but not impossible, so I think the former alternative is preferable (though I'd like a way to also remove them later).

The reason I’m bringing this up now, is that if I want to fix this in the future and use the alternative which keeps a vector of these, then I feel that we’re introducing a function which should basically be deprecated as soon as possible, and that doesn’t sit well with me.

I didn't do this yet, because even though this pr has ballooned into basically a refactor of the parameter's API, I really was trying to have the least impact on existing behavior, but allowing for undeclared parameters (or not), basically touches every part of the API.

Note, I’m interested in what we would like, rather than when we should do it (now or after Dashing). I can figure that out afterwards.

sloretz: > The first concrete example of this is that I’d like to use this callback for checking the type of the use_sim_time parameter (which is implicitly declared by each node) before setting it, but if I did that then a user could easily override that callback when they set their own and ignore the return value, breaking the behavior of use_sim_time.

This doesn't seem too bad at the moment while there is only one parameter `use_sim_time`. Downstream code would have to duplicate validating `use_sim_time` though.

> Alternatives are that I can simply push back callbacks onto a vector and call of them in the order they were added

Are all callbacks given the same parameters? If so, a custom callback that errored when an unknown parameter was given would still need to know about built in parameters like `use_sim_time`.

> or I could instead setup a feature where each parameter can have a single callback in addition to the proposed function (one callback for all parameters)

I like this option with the callback added by the proposed function only getting parameters which don't have a dedicated validation function. I think that allows adding a built in parameter in the future without needing to update a callback set here.

tfoote: We discussed supporting multiple validation callbacks as an enhancement going forward.  It's ticketed in in #296

wjwwood: > This doesn't seem too bad at the moment while there is only one parameter use_sim_time. Downstream code would have to duplicate validating use_sim_time though.

I dunno, the first example I write for how to use this callback will already need to be careful to store the return of this function and call it before after the new callback code, so it's pretty bad, imo.

> Are all callbacks given the same parameters? If so, a custom callback that errored when an unknown parameter was given would still need to know about built in parameters like use_sim_time.

Yes, the idea is that you ignore parameters you don't care about.

> I like this option with the callback added by the proposed function only getting parameters which don't have a dedicated validation function. I think that allows adding a built in parameter in the future without needing to update a callback set here.

It wouldn't work like that, because it isn't clear how to handle a case where you get a `set_parameters_atomically({{"p1", ...}, {"p2", ...}})` and `p1` has a dedicated callback, but `p2` doesn't. Do you call `p1`'s callback and the general one, or just `p1's`, or just the general one?

Another reason I favor the former alternative (basically just a vector of callbacks that fire for all parameter set actions) is that it's simpler. You call all the callbacks in order, for each set action unconditionally. The only thing you have to decide is whether to call all and then see if any failed or stop at the first failure (I like the latter better).

wjwwood: > We discussed supporting multiple validation callbacks as an enhancement going forward. It's ticketed in in #296

Thanks, I had forgotten about that issue. I suppose I could move the function there, but I wanted to make some headway on _what we want_, so that when we are planning the next release we have something concrete to evaluate.

Also, this brings in a new concern (which my previous response to @rohbotics didn't consider) which is that a single "set verification" callback doesn't make it very easy to have two parts of the code both influence verification without coordination (like with `use_sime_time` and a single user parameter).

So, I'm really asking about that.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -172,9 +169,157 @@ NodeParameters::NodeParameters(
...
+  // If it failed to be set, then throw an exception.
+  if (!result.successful) {
+    throw rclcpp::exceptions::InvalidParameterValueException(
+      "parameter '" + name + "' could not be set: " + result.reason);
+  }
+
+  // Publish the event.
+  events_publisher_->publish(parameter_event);
+
+  return parameters_.at(name).value;
```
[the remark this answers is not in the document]
wjwwood: If the type changes at all then it is invalid. I'd like to know what others think, because I'm not sure.

On the one hand, we return `const &` to private members often and `std::variant` (the closet alternative to our `ParameterValue`) allows you to get const and non-const references to storage:

https://en.cppreference.com/w/cpp/utility/variant/get

Even though, I guess it must be invalid after you do an `emplace` or `swap` on it...

Perhaps it is sufficient to document what should be obvious, which is that the reference is only valid until you change the parameter. I think it would be good to have this feature as reading the parameters is a common case (done often) and could be much faster if a reference is returned when the parameter value is large.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -184,63 +329,186 @@ NodeParameters::set_parameters(
...
+    // change other things from the ParameterInfo.
+    parameters_[kv_pair.first] = kv_pair.second;
+  }
+
+  // Undeclare parameters that need to be.
+  for (auto parameter_to_undeclare : parameters_to_be_undeclared) {
+    auto it = parameters_.find(parameter_to_undeclare->get_name());
+    // assumption: the parameter to be undeclared should be in the parameter infos map
+    assert(it != parameters_.end());
+    if (it != parameters_.end()) {
```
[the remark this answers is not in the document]
wjwwood: Yes, but if in release mode I'd still rather not access the iterator when it points to end, but in debug you'll have it die earlier. Either way this is defensive, and should no happen in normal operation.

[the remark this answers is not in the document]
jacobperron: :ok_hand:

dirk-thomas:
How about support for this in `rclpy`?

rclcpp/include/rclcpp/node.hpp
clalancette: There are two diverging thoughts I had while reading this thread:

1.  For the specific example you bring up (checking the type and then doing additional verification), it seems to me like checking that the type doesn't change should essentially be the default.  While there are uses for having a parameter that can change types, I think that for the vast majority of cases you just want a parameter to have one type and then error (or really, an ignored update) if someone tries to change that type.  So really, I think the right way to solve this particular example is to introduce type-checking into the low-level callback that happens when a parameter is changed, and then you still only need a single user callback.

1.  With the above being said, I can see that there are uses for having multiple callbacks for a particular parameter that are un-coordinated.  While I think we'll eventually want to end up in a place where we have a vector of callbacks, I'll bring up the point that having multiple callbacks can interact in somewhat unexpected ways for the user and/or the writer of the callbacks.

wjwwood:
Support for this is entirely client side, so support for `rclpy` has to be added separately. Also, it will likely be quite difference since several of these signatures are C++ specific. But declare/undeclare/allow undeclared are concepts that are transferable.

I don't have any plan to do that right now, however, since the original request for this feature only needs to be in c++.

Also, the declare/undeclare features are entirely client library facing, I think that Python and C++ will still communicate over the ROS interface just the same. Though the tests for that could be improved.

rclcpp/include/rclcpp/node.hpp
wjwwood: > 1. For the specific example you bring up ...

What about a case where you want to allow int or float? Or boolean or string? I think the user needs to be able to express logic like this if they desire.

However, we can use the meta data about a parameter to check some of these things automatically, like type or immutability or range/choice.

I would prefer to see this kind of checking be the default (as you suggested), but I think it also needs to be possible to opt out of the automatic checking and then do the checking manually.

So the user can, for example, provide a callback which does some custom logic checking and then also calls the verification function that automatically verifies the new value according to the parameter descriptor.

> I'll bring up the point that having multiple callbacks can interact in somewhat unexpected ways for the user and/or the writer of the callbacks.

Yeah, but is there any way around that? Currently if someone sets a callback it's up to them to store and call the old callback if they want. If they overlook this then it will just break things. I don't think that's better or worse than allowing multiple callbacks and failing at the first one (run in the reverse? order of being added).

Most callbacks will only look at a narrow set of parameters or a single parameter. In which case, interference will be minimal.

In the future we can add syntactic sugar that makes it easy for the user to write a callback for a single parameter, getting rid of the boilerplate of checking the parameter name and stuff like that. But to support that and the more advanced use case (a callback looking for a group of parameters or verifying something about all parameters) at the same time we need a vector of callbacks and not a single one, imo.

clalancette: > What about a case where you want to allow int or float? Or boolean or string? I think the user needs to be able to express logic like this if they desire.

I agree, but I also think this is a niche use-case.  Most parameters just need to be a specific type that is settable at runtime.

> Yeah, but is there any way around that?

The basic way around it is to only use a single installed callback.  To be clear, I'm not advocating that we restrict it like it is today; I think that a vector of callbacks makes sense for advanced use-cases.  But I also think that if we get enough "generic" features into the underlying parameter implementation (type checking, read-only, ranges, etc), then most users will only need to bother with 0 or 1 callbacks.

wjwwood: >  But I also think that if we get enough "generic" features into the underlying parameter implementation (type checking, read-only, ranges, etc), then most users will only need to bother with 0 or 1 callbacks.

That's a fair point, but I am concerned about designing ourselves into a corner where the niche cases are no longer possible to support without breaking API significantly or having lots of redundant API's.

I also would really like to see us implement the (let's call them) "parameter constraints" using the same mechanism available to the user. So it's transparent (just another callback in the list), composable (user can call from their callback conditionally), and possible to disable the default checks (remove from callbacks list).

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -595,14 +597,25 @@ NodeParameters::describe_parameters(const std::vector<std::string> & names) cons
...
-      }))
-    {
-      results.push_back(kv.second.descriptor);
+
+  for (const auto & name : names) {
+    auto it = parameters_.find(name);
+    if (it != parameters_.cend()) {
+      results.push_back(it->second.descriptor);
+    } else if (allow_undeclared_) {
+      // parameter not found, but undeclared allowed, so return emtpy
```
jacobperron: "empty"

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -595,14 +597,25 @@ NodeParameters::describe_parameters(const std::vector<std::string> & names) cons
...
+      rcl_interfaces::msg::ParameterDescriptor default_description;
+      default_description.name = name;
+      results.push_back(default_description);
+    } else {
+      throw rclcpp::exceptions::ParameterNotDeclaredException(name);
     }
   }
+
+  if (results.size() != names.size()) {
+    throw std::runtime_error("results and size unexpectedly different sizes");
```
jacobperron: "result and names" ?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -611,16 +624,23 @@ NodeParameters::get_parameter_types(const std::vector<std::string> & names) cons
...
+    } else if (allow_undeclared_) {
+      // parameter not found, but undeclared allowed, so return not set
       results.push_back(rcl_interfaces::msg::ParameterType::PARAMETER_NOT_SET);
+    } else {
+      throw rclcpp::exceptions::ParameterNotDeclaredException(name);
     }
   }
+
+  if (results.size() != names.size()) {
+    throw std::runtime_error("results and size unexpectedly different sizes");
```
jacobperron: ditto

rclcpp/test/test_node.cpp
```diff
@@ -1360,26 +1362,339 @@ TEST_F(TestNode, get_parameters_undeclared_parameters_allowed) {
...
+    EXPECT_TRUE(results[0].read_only);
+
+    EXPECT_EQ(results[1].name, name1);
+    EXPECT_EQ(results[1].type, rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER);
+    EXPECT_FALSE(results[1].read_only);
+
+    EXPECT_EQ(results[2].name, name2);
+    EXPECT_EQ(results[2].type, rcl_interfaces::msg::ParameterType::PARAMETER_STRING);
+    EXPECT_TRUE(results[2].read_only);
+  }
```
jacobperron: Consider also testing with a declared parameter and undeclared parameter together.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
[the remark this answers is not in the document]
jacobperron: Yeah, I think as long as it's documented it should be fine.

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -131,7 +132,7 @@ class NodeOptions
 
   /// Return a reference to the use_global_arguments flag.
```
hidmic: @wjwwood here and below, getter methods no longer return a reference.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -236,34 +256,56 @@ Node::get_parameter(const std::string & name) const
...
+  return result.front();
+}
+
+std::vector<rcl_interfaces::msg::ParameterDescriptor>
+Node::describe_parameters(const std::vector<std::string> & names) const
 {
   return node_parameters_->describe_parameters(names);
 }
 
 std::vector<uint8_t>
```
hidmic: @wjwwood would it make sense to return an `std::vector<rclcpp::ParameterType>` instead?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -253,12 +523,16 @@ NodeParameters::get_parameters(const std::vector<std::string> & names) const
   std::vector<rclcpp::Parameter> results;
```
hidmic: @wjwwood nit: consider doing `results.reserve(names.size())` to avoid incremental reallocations below.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -317,17 +597,25 @@ NodeParameters::describe_parameters(const std::vector<std::string> & names) cons
 {
   std::lock_guard<std::mutex> lock(mutex_);
   std::vector<rcl_interfaces::msg::ParameterDescriptor> results;
```
hidmic: @wjwwood nit: consider doing `results.reserve(names.size())` to avoid incremental reallocations below.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -336,16 +624,23 @@ NodeParameters::get_parameter_types(const std::vector<std::string> & names) cons
 {
   std::lock_guard<std::mutex> lock(mutex_);
   std::vector<uint8_t> results;
```
hidmic: @wjwwood nit: consider doing `results.reserve(names.size())` to avoid incremental reallocations below.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -172,9 +169,157 @@ NodeParameters::NodeParameters(
 NodeParameters::~NodeParameters()
 {}
 
+RCLCPP_LOCAL
+bool
+__lockless_has_parameter(
```
hidmic: @wjwwood out of curiosity, I see below that sometimes this function is used and sometimes its implementation is copied _almost_ verbatim. Why is that?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
wjwwood: derp, this is what happens when you code late

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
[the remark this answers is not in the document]
wjwwood: Actually, I think I was wrong (I assumed this was the version that returned the actual value and not a `ParameterValue`). I think this reference will be valid until you undeclare it, because even if you set it, then it should be changed with the assignment operator (it's not stored as a pointer). Also, it is robust to changes when you mutate the `parameters_` map, since the reference returned by `at()` is valid even when inserting or deleting:

https://stackoverflow.com/a/4555645/671658

So long as you don't delete the parameter you're reference points to (indirectly in this case).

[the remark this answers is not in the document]
wjwwood: I added a note about this in the docs, however.

rclcpp/src/rclcpp/node.cpp
wjwwood: Maybe, I didn't change it because that is how it was before, and I thought it might have something to do with the enum existing in two places (once in `rclcpp` and once as constants in `rcl_interfaces::msg::ParameterType`. I think I'll leave it for now?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
wjwwood: I see a lot of cases where this is happening, but all are either of a different type for the first parameter or they explicitly store the iterator, compare it to end(), and then use the iterator, which this function does not allow you to do because the iterator is not returned.

wjwwood:
@jacobperron and @hidmic I think I addressed all of your feedback.

wjwwood:
@jacobperron I cleaned up the time_source tests only to realize they've been cleanup on master already 😛. But I also punted on documenting/testing `list_parameters` (it was previously undocumented), and I'll make an issue about it and other remaining problems after this get's in.

I also fixed the linter issues and merged it with master.

Can you check the new commits as well as the connected pr's?

- https://github.com/ros2/rcl_interfaces/pull/35
- https://github.com/ros2/demos/pull/241
- https://github.com/ros2/system_tests/pull/278
- https://github.com/ros2/robot_state_publisher/pull/18

rclcpp/test/test_node.cpp
```diff
@@ -260,3 +262,1609 @@ TEST_F(TestNode, now) {
...
+
+    EXPECT_EQ(results[0], rcl_interfaces::msg::ParameterType::PARAMETER_NOT_SET);
+
+    EXPECT_EQ(results[1], rcl_interfaces::msg::ParameterType::PARAMETER_NOT_SET);
+
+    EXPECT_EQ(results[2], rcl_interfaces::msg::ParameterType::PARAMETER_NOT_SET);
+  }
+}
+
+// // test get_parameter with undeclared allowed
```
jacobperron: I think this commented section can be removed? Looks like they're duplicates of tests above.

wjwwood: Oops didn't mean to commit that, I'll remove them.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6744/) (known issues `test_param_yaml` and `flake8`)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3052/) (known issues `test_param_yaml` and `flake8`)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5542/) (known issues `test_param_yaml` and `flake8`)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6569/) (5 new compiler warnings 🙃, fixing those now)

rclcpp/src/rclcpp/node.cpp
```diff
@@ -27,7 +27,17 @@
 #include "rclcpp/node_interfaces/node_clock.hpp"
 #include "rclcpp/node_interfaces/node_graph.hpp"
 #include "rclcpp/node_interfaces/node_logging.hpp"
+// When compiling this file, Windows produces a deprecation warning for the
+// deprecated function prototype of NodeParameters::register_param_change_callback().
```
jacobperron: oof, I'll remember this next time I consider deprecating a function.

Can we move these directives around the deprecated function in node_parameters.hpp instead of everywhere it is included?

rclcpp/test/test_node.cpp
```diff
@@ -442,11 +442,13 @@ TEST_F(TestNode, declare_parameters_with_no_initial_values) {
...
     // with namespace, defaults, no custom descriptors, no initial
-    auto values = node->declare_parameters<int>("namespace1", {
+    int64_t bigger_than_int = INT64_MAX - 42;
+    auto values = node->declare_parameters<int64_t>("namespace1", {
       {"parameter_a", 42},
       {"parameter_b", 256},
+      {"parameter_b", bigger_than_int},
     });
-    std::vector<int> expected = {42, 256};
+    std::vector<int64_t> expected = {42, 256, bigger_than_int};
```
jacobperron: :thinking: Why was there a warning to begin with?

rclcpp/src/rclcpp/node.cpp
wjwwood: I tried that first, but luckily I didn't trust it and checked, when you do that it prevents the deprecation warning in valid cases too...

rclcpp/test/test_node.cpp
wjwwood: So, parameters are always stored as `int64_t`, and the type in this vector causes the deduced template argument to be the same, and therefore you ended up with some template code that boiled down to `int = int64_t` which triggers the warning.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6764/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3066/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5557/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6589/)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -168,10 +169,16 @@ NodeParameters::NodeParameters(
       rclcpp::ParameterValue(param.get_value_message());
   }
 
-  if (!initial_parameters.empty() && allow_undeclared_) {
-    rcl_interfaces::msg::SetParametersResult result = set_parameters_atomically(initial_parameters);
-    if (!result.successful) {
-      throw std::runtime_error("Failed to set initial parameters");
+  // If asked, initialize any parameters that ended up in the initial parameter values,
+  // but did not get declared explcitily by this point.
+  if (automatically_declare_initial_parameters && !this->get_initial_parameter_values().empty()) {
```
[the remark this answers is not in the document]
wjwwood: Good catch.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6769/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3070/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5562/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6595/)

wjwwood:
Ok, merged 🎉!

I'll follow up by:

- adding documentation to the migration guide for dashing
- making an issue to track the documentation/testing of `list_parameters`
- making an issue to track the need to address callback concerns (only one right now, probably need more than one at a time)
- making an issue to track the need to fix parameter addressing and how to support subnode properly
- making an issue to track rclpy support

wjwwood:
Another TODO:

- update https://index.ros.org/doc/ros2/Tutorials/Node-arguments/#parameters

wjwwood:
Documentation and tutorial updated in https://github.com/ros2/ros2_documentation/pull/187

PR #501, opened 2018-06-19
Further expand test tolerance to address flakiness.

mjcarroll:
I tested this on mini3 with the expanded tolerance and a Debug build, and it corrects it.  The failures were right outside of the tolerance (by a few milliseconds), so it may have been some jitter in the timers causing the tests to fail.

I have also changed the assertions to be more informative if it fails again.

### DIFF
--- a/rclcpp/test/executors/test_multi_threaded_executor.cpp
+++ b/rclcpp/test/executors/test_multi_threaded_executor.cpp
@@ -71,8 +71,11 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
   std::atomic_int timer_count {0};
 
   auto timer_callback = [&timer_count, &executor, &system_clock, &last_mutex, &last]() {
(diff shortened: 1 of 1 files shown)


mjcarroll:
With build type set to Debug.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4755/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1623/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3944/)
* Windows [Build Status]

rclcpp/test/executors/test_multi_threaded_executor.cpp
```diff
@@ -98,3 +101,4 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
   executor.add_node(node);
   executor.spin();
 }
+
```
dirk-thomas: Please remove this extra newline.

PR #504, opened 2018-06-22
Fix a bug when publishing unique_ptr

Jrdevil-Wang:
A do_inter_process_publish is always called when publishing unique_ptr, even for the intra-process situation.
It will cause the performance of intra-process publishing degrade as the inter-process situation.
Since other cases of intra-process publishing (i.e. publishing shared_ptr, publishing message reference, and publishing message pointer) all call this version of publish method, it will basically affect all intra-process publishing performance.

This fix will bring the expected intra-process low-latency performance.

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -188,7 +188,6 @@ public:
   virtual void
   publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
   {
-    this->do_inter_process_publish(msg.get());
     if (store_intra_process_message_) {
       // Take the pointer from the unique_msg, release it and pass as a void *
       // to the ipm. The ipm should then capture it again as a unique_ptr of
@@ -209,6 +208,7 @@ public:
         rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish intra process message");
       }
     } else {
+      this->do_inter_process_publish(msg.get());
       // Always destroy the message, even if we don't consume it, for consistency.
       msg.reset();
     }
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -207,6 +207,7 @@ public:
       if (RCL_RET_OK != status) {
         rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish intra process message");
       }
+      this->do_inter_process_publish(msg_ptr);
     } else {
       this->do_inter_process_publish(msg.get());
       // Always destroy the message, even if we don't consume it, for consistency.


dirk-thomas:
Have you tried to subscribe to the message from a different process with this patch applied?

wjwwood:
This is not an appropriate fix, external subscriptions will not work when intra-process is on, as @dirk-thomas suggested.

Jrdevil-Wang:
Yes, you are right, @wjwwood @dirk-thomas. Sorry, I miss the requirement that intra- and inter- should be able to work together.
On the other hand, I think you get my point too. Even though the current intra- implement DO provide zero-copy feature, the key purpose is missed if the performance of intra- is the same as inter-.
I will try to come up with an appropriated fix.

Jrdevil-Wang:
OK. I add a do_inter_process_publish AFTER the intra- message is published. This will enable intra- and inter-process publishing to work together.
Tests have been done by publishing an 800x600x24bit image. The intra- performance is better than inter- (about 5-10 times faster), but is still a little worse (about 2 times) than the expected zero-copy performance.
Please try and review. @dirk-thomas @wjwwood 

It could still be improved, I think. I guess that the subscriber should subscribe the intra- version DDS handle only, and do not subscribe the normal version DDS handle.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -208,7 +207,9 @@ class Publisher : public PublisherBase
       if (RCL_RET_OK != status) {
         rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish intra process message");
       }
+      this->do_inter_process_publish(msg_ptr);
```
dirk-thomas: The function call `store_intra_process_message_()` above is taking ownership of the `msg_ptr`. So we can't rely on `msg_ptr` being still valid here. That is reason why the "inter" call currently happens *before* the "intra" call.

It would certainly be good to address the latency issue you describe but it will probably be more complicated than this.

dirk-thomas:
> I guess that the subscriber should subscribe the intra- version DDS handle only, and do not subscribe the normal version DDS handle.

The subscriber might additionally receive a message from another process. So it still needs to subscribe to the DDS topic in general.

Jrdevil-Wang:
> The subscriber might additionally receive a message from another process. So it still needs to subscribe to the DDS topic in general.

OK, I get your point.
At this point, I have to admit that I am confused why ros2 abandons the pub-sub link design in ros1. 
Tagging a node, rather than a pub-sub link, as "intra_comms" do not really make sense to me.

Jrdevil-Wang:
> Can you clarify what feature or API in ROS 1 you're referring to? I just want to make sure we're talking about the same thing.

@wjwwood , please correct me if I am wrong.

It seems in ros2, whether using intra-process communication is decided by a parameter on Node construction (use_intra_process_comms in the following code).
https://github.com/ros2/rclcpp/blob/4d67a8671b55b28a6ae63697310a11632a888ed4/rclcpp/include/rclcpp/node.hpp#L77-L81

In ros1, whether a message is published with intra- mode is decided per subscriber-link.
https://github.com/ros/ros_comm/blob/89ca9a7bf288b09d033a1ab6966ef8cfe39e1002/clients/roscpp/src/libros/publication.cpp#L410-L417

wjwwood:
Further more, you can see that the "local" connection is made for all publishers and subscriptions unconditionally in their respective creation functions:

- For publishers in `advertise`:
  - https://github.com/ros/ros_comm/blob/89ca9a7bf288b09d033a1ab6966ef8cfe39e1002/clients/roscpp/src/libros/topic_manager.cpp#L388
- For subscribers in `registerSubscriber`:
  - https://github.com/ros/ros_comm/blob/89ca9a7bf288b09d033a1ab6966ef8cfe39e1002/clients/roscpp/src/libros/topic_manager.cpp#L531

Jrdevil-Wang:
OK, let me give an example to explain.

Say there are two processes: process A has one publisher and one subscriber, and process B has one publisher and subscriber. All these publishers and subscribers are with the same topic.

From my understanding, in ros1, there will be four subscriber links, and two of them are intra-process link, the other two of them are inter-process link. When A.pub publishes a message, this message is sent to A.sub through intra-process link, and to B.sub through inter-process link. It is fine to me.

However, in ros2, all publishers and subscribers are with the same DDS topic. Therefore, when A.pub publishes a message, this message is sent to A.sub AND B.sub through DDS (DDS don't know we have another channel), and to A.sub with intra-process mechanism.

Problem is, I should prevent DDS to send this message, but I can't, right?

wjwwood:
Yes, that's the double delivery problem. Borrowing from your example A.sub will receive the message from A.pub twice, once via intra process and once via inter process.

DDS doesn't know about our "side channel", but we can tell it to not send data to itself when we have a side channel. That's what this option is for in our middleware API:

https://github.com/ros2/rmw/blob/b4e6596161f122d520e5085742dd2e48dcdca82b/rmw/include/rmw/rmw.h#L246

Even in the case that there is double delivery, subscriptions which have intra process enabled will ignore interprocess messages from publishers in the same node, see:

https://github.com/ros2/rclcpp/blob/4d67a8671b55b28a6ae63697310a11632a888ed4/rclcpp/include/rclcpp/subscription.hpp#L218-L224

This is inefficient because we've needlessly sent the message to ourselves over the middleware (DDS in this case), but we don't deliver data to the user twice.

---

Unfortunately, the `ignore_local_publications` option is not consistently supported at the moment in our rmw implementations, so we're not using it right now.

That's the remaining work that needs to be done (in my opinion) in order to make intra process on by default in Nodes.

---

A further enhancement would be to have publishers watch the "matched subscriptions" (we don't expose this information from DDS yet) in order to know if publishing to DDS would actually end up sending data to anyone, and then as an optimization not doing inter process publish if there are no matched subscriptions from other participants.

---

Each of these are sort of related, but different topics than the original question, or proposed enhancement in my opinion and perhaps deserve their own issues.

To the best of my knowledge, the original proposal was to publish the intra process message first to improve latency, but that cannot be done without copying the message first to send to the inter process manager.

The other option is to convert the unique pointer to a shared pointer before giving it to the intra process manager so that the inter process publish which is called later may be assured it will still exist. This causes other issues however, and makes truly zero copy across several pub/sub links impossible, because you can never be sure that the user removed read-write access to the message they received in a subscription callback and then published again.

Jrdevil-Wang:
@dirk-thomas , @wjwwood . I get an idea about protecting the msg_ptr. Please check if it is acceptable before I implement it.

During store_intra_process_message_, a record is created in a MappedRingBuffer. 
Each record is an element struct shown below.

https://github.com/ros2/rclcpp/blob/4d67a8671b55b28a6ae63697310a11632a888ed4/rclcpp/include/rclcpp/mapped_ring_buffer.hpp#L219-L224

I plan to add another field "bool is_locked;". Initially, it is set to true, and will be set to false after finishing do_inter_process_publish. After that, nothing is different.
During the time when is_locked is true, any calls to MappedRingBuffer::pop_at_key will return a copy rather than pop the unique_ptr.
Thus, the msg_ptr here is guaranteed to be valid.

Jrdevil-Wang:
> With your proposed change, if the inter-process copying of the message isn't fast enough, then the pointers on the image would not be the same, as the call to pop_key_at would now be returning a copy, not the original.

I think the image_pipeline_all_in_one demo will still work with my proposed change, because it uses the  SingleThreadedExecutor executor. The executor has to complete of the publishing process, before any subscribers get the chance to execute. But yeah, when using multi-threaded executor, my proposed change will bring uncertainty.

> To do this we would need to implement the "count matched subscriptions" feature I described above.

This is a good solution, but I don't think it is easy to keep this count accurate, considering inter-machine subscriptions.

wjwwood:
> I think the image_pipeline_all_in_one demo will still work with my proposed change, because it uses the SingleThreadedExecutor executor. The executor has to complete of the publishing process, before any subscribers get the chance to execute. But yeah, when using multi-threaded executor, my proposed change will bring uncertainty.

With a single threaded executor it would always copy before it can take it out on the other side. With a multi threaded executor or with two executors of any kind, it would only sometimes make a copy first.

> This is a good solution, but I don't think it is easy to keep this count accurate, considering inter-machine subscriptions.

Well, so long as we use the same information the middleware keeps it should be accurate, after all this is how the middleware knows to create unicast connections to other endpoints. But we still need the feature.

---

So what should we do with this pull request? Personally I'd prefer to keep the copy in place as there are only specific cases where we might be able to avoid it with the current proposal and instead we should strive to remove the copy when we can do it always, i.e. when there are no inter-process subscriptions.

Separately we could consider merging this proposed optimization, but I think we should have an opt out option so that people can select determinism versus lower average latency.

wjwwood:
Sorry I made a mistake in the previous comment, with a single threaded executor it will always avoid the copy, which is pretty good. I'll update the comment, but I still think having the ability to opt-out of this optimization would be best.

wjwwood:
Actually it's more complicated than even what I said in the previous comment, it is actually still possible that a copy is made on delivery with a single threaded executor if the publish call is made outside of the executor thread (asynchronously).

Jrdevil-Wang:
@wjwwood Yeah, this problem is much more complicated than I thought, and there seems no easy way to solve here. Maybe I should just close this PR, and wait for a more thorough solution?

wjwwood:
@Jrdevil-Wang if you find the enhancement a benefit for your use case, I'd be ok iterating on this to make it opt-in or opt-out, but otherwise I would agree that we should close this and look towards one of the other optimizations/improvements I described above.

I do appreciate you looking into it though! It was good to get written down somewhere and explain it.

wjwwood:
I'll let you decide to close it or not, or at the very least I'll wait for your feedback before closing it myself.

Jrdevil-Wang:
@wjwwood  Thanks!
I really think my enhancement would be a benefit for a lot of use cases. Despite of the unsafe pointer operation, my enhancement should work well with SingleThreadExecutor. Of course, there is a more severe problem when using MultiThreadExecutor.
In particular, I think most users start trying the intra-process zero-copy feature with the image_pipeline demo, and among them, a lot of users (including myself) will find the performance is not more efficient than inter-process communication, even for a demo, which is pretty frustrating.
Make it an opt seems a good way to let more people notice the problem here. What should I do if I would like to make it an opt?

wjwwood:
Ok, so first you'd need to implement what you proposed above (https://github.com/ros2/rclcpp/pull/504#issuecomment-400604000), where the "final" call to `pop_at_key` for a message instance will return a copy rather than the original.

Then we'd want to expose an option (probably at creation of the publisher) which opted into or out of the new behavior. Then in the demos (like the image pipeline demo) we'd need to ensure we're using the current behavior (because zero-copy is more important than the latency). At best we'd update the demo (and wiki page) to highlight this trade-off (lower average latency for occasionally needing to copy the message).

One thing to watch out for in the implementation of the suggestion above, is that the message still needs to get removed from the ring buffer, even if the "last" call to `pop_at_key` was called _before_ `do_inter_process_publish` finishes. Normally the last call to `pop_at_key` removes the original unique pointer.

ivanpauno:
@Jrdevil-Wang #664 together with #690 have solved what you were proposing here.

I will summarize a little bit:
- #664 avoided doing the inter process publish when only having intra process subscriptions. It used the count matching api introduced in #628.
- #664 also allowed configuring intra process comm at publisher/subscription level. By default, the node intra process configuration was used. api for doing that will probably change with #713.
- #690 allowed storing shared_ptr messages in the IPM. Using that, now the intra process publish is always done before the inter process publish.

https://github.com/ros2/rclcpp/blob/2b3e4d9c98b17bbed6a57e36a0466008454a1e4d/rclcpp/include/rclcpp/publisher.hpp#L91-L120

- #690 also allowed zero-copy intraprocess comm with multiple-subscriptions, using `shared_ptr<const T>` callback signature and `unique_ptr<T>` publish signature.

I think this PR could be closed now, but I will wait for your feedback before doing it.

Jrdevil-Wang:
Dear ivanpauno,

Good to know that the issue is solved! Please close PR #504.

Best regards,

Jrdevil-Wang

From: ivanpauno
Date: 2019-05-03 22:42
To: ros2/rclcpp
CC: Yu-Ping Wang; Mention
Subject: Re: [ros2/rclcpp] Fix a bug when publishing unique_ptr (#504)
@Jrdevil-Wang #664 together with #690 have solved what you were proposing here.
I will summarize a little bit:
#664 avoided doing the inter process publish when only having intra process subscriptions. It used the count matching api introduced in #628. 
#664 also allowed configuring intra process comm at publisher/subscription level. By default, the node intra process configuration was used. api for doing that will probably change with #713. 
#690 allowed storing shared_ptr messages in the IPM. Using that, now the intra process publish is always done before the inter process publish. 
https://github.com/ros2/rclcpp/blob/2b3e4d9c98b17bbed6a57e36a0466008454a1e4d/rclcpp/include/rclcpp/publisher.hpp#L91-L120
#690 also allowed zero-copy intraprocess comm with multiple-subscriptions, using shared_ptr<const T> callback signature and unique_ptr<T> publish signature. 
I think this PR could be closed now, but I will wait for your feedback before doing it.

PR #505, opened 2018-06-27
keep shared pointer reference to rcl_node handle for subscription fini

wjwwood:
Should fix https://github.com/ros2/examples/issues/209 and fix https://github.com/ros2/examples/issues/211, and it should also make https://github.com/ros2/examples/pull/210 unnecessary, but I would leave https://github.com/ros2/examples/pull/210 because it is more correct in my opinion.

After thinking about this a long time and doing some targeted testing, I'm still not completely convinced myself that this change is correct.  If we choose to merge it, we should be very suspicious of it and review it closely.

So I think we should not merge this and instead I'll keep looking into it. But we'll need to make a decision to risk this workaround or leave the error output.

### DIFF
--- a/rclcpp/src/rclcpp/subscription.cpp
+++ b/rclcpp/src/rclcpp/subscription.cpp
@@ -37,23 +37,14 @@ SubscriptionBase::SubscriptionBase(
   type_support_(type_support_handle),
   is_serialized_(is_serialized)
 {
-  std::weak_ptr<rcl_node_t> weak_node_handle(node_handle_);
-  auto custom_deletor = [weak_node_handle](rcl_subscription_t * rcl_subs)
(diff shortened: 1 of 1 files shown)


wjwwood:
We decided to go ahead with this and fix it in a better way after the release, CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4826/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1682/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4005/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4873/)

wjwwood:
The test failures are a regression due to a new version of PyYAML. That will be fixed separately.

wjwwood:
Follow up issue: https://github.com/ros2/rclcpp/issues/506

PR #507, opened 2018-07-03
Fix rosidl dependencies

mikaelarguedas:
This package does not generate any message so it doesnt need to depend on the generators, instead it should link against the libraries that provide the symbols it uses (in this case rosidl_typesupport_cpp)

Connects to ros2/demos#264


mikaelarguedas:
CI for this at https://github.com/ros2/demos/pull/264#issuecomment-402314880

PR #517, opened 2018-07-17
Support jump handlers with only pre- or post-jump callback

dhood:
We are assuming both pre- and post-jump callback are set for a JumpHandler, which can lead to `bad_function_call` exceptions.

To me it seems reasonable that only one is provided and not the other, so I took the approach of checking before calling them, but if we think they should both be required and users have to put noops as appropriate, we can add a check to the constructor instead.


dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4922/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1749/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4087/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4943/)

Issue #518, opened 2018-07-19
Add way to check if init has been called (i.e. isInitialized)


chapulina:
#### Feature description

It would be convenient to have a way to check if `rclcpp::init` has been called, similar to [ros::isInitialized](http://docs.ros.org/diamondback/api/roscpp/html/namespaceros.html#accd0b48f347e82ae96b261f680af41dd).

#### Implementation considerations

If this has been left out due to some design decision, I suppose users can live without it, and instead try to catch specific errors resulting from not calling init. I searched a bit and couldn't find an explanation, so I imagine it may just be in the backlog?

Thanks!

sloretz:
I was going to say `rclcpp::ok()`, but it looks like it returns `true` if `rclcpp` hasn't been initialized. Maybe it should return `false` instead?

mikaelarguedas:
would [`rclcpp::ok()`](https://github.com/ros2/rclcpp/blob/8f2052d65a8b05e19958db5484f5425c688098b8/rclcpp/include/rclcpp/utilities.hpp#L86-L88) work for your use case? (though there may be some subtle differences as it's using a global that I think would default to returning true). We could make rclcpp::init assign a value to the global and flip the default

There is indeed a related issue in the backlog: https://github.com/ros2/rclcpp/issues/3

chapulina:
Yeah if `g_signal_status` is [initialized](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/utilities.cpp#L43) to something other than 0 and set to 0 later, then maybe `rclcpp::ok` could be enough.

sloretz:
It  seems like the discussion in the waffle meeting today leaned towards having separate APIs for tracking whether a signal has been received versus whether rclcpp is currently initialized.

wjwwood:
Closed by #522.

Issue #154, opened 2015-11-13
rclcpp lifecycle: refine design of init, shutdown


jacquelinekay:
See discussion on ros2/rclcpp#152

Karsten1987:
closing this as the related PR is closed.

wjwwood:
I don't think this was the right action, as the related PR was closed (declined) but the topic is still an issue, see: https://discourse.ros.org/t/how-to-shutdown-and-reinitialize-a-publisher-node-in-ros-2/4090/1f

wjwwood:
We could clean this issue up and make it clearer what needs to be done.

wjwwood:
I think this was addressed in https://github.com/ros2/rclcpp/pull/587.

Issue #3, opened 2014-09-04
Consider renaming rclcpp::ok


wjwwood:
Discussed names were:
- `bool rclcpp::is_shutdown` and `bool rclcpp::is_not_shutdown`
- `bool rclcpp::is_shutting_down`
- `State rclcpp::state` and compare to predefined states
  - e.g. `while (rclcpp::state() == rclcpp::RUNNING)`
  - or `while (rclcpp::state() != rclcpp::SHUTDOWN)`

Karsten1987:
@wjwwood @dirk-thomas Do you think it's worth iterating over this ticket?

dirk-thomas:
We should make a decision and then just implement it that way.

dhood:
To summarise the relationship between `rclcpp::ok()`, signals, and `rclcpp::shutdown()` in the current state of the code:

- `rclcpp::ok()` returning true suggests that a signal has been received (and not cleared).
- The rclcpp signal handler will call the user's signal handler, then call `trigger_interrupt_guard_condition`.
- `trigger_interrupt_guard_condition` will wake up waitsets and set the signal received.
- `rclcpp::shutdown` will call `trigger_interrupt_guard_condition` and will call whatever `on_shutdown` callbacks have been registered.

Today it looks that `rclcpp::ok()` is not directly related to any shutdown processes, it just is usually followed by a shutdown. But not always, as pointed out by @dirk-thomas in [this thread](https://github.com/ros2/rclcpp/pull/353#issuecomment-321657654):

```
init()
...
spin()  // waiting for sigint to return from wait
// do some else
spin()  // a second SIGINT while waiting here should wake up the wait again
...
shutdown()
```

dhood:
To conclude (wanted to get that post out before the meeting), my opinion is that if we want to rename `rclcpp::ok()` it should not be to something that implies shutdown, but interruption.

I'd vote for keeping it as `rclcpp::ok`, or returning a state enum.

I considered `rclcpp::signal_received()` or (not entirely accurate) `rclcpp::interrupted()` but both introduce concepts that users are not necessarily familiar with (and don't need to be for the most part).

dirk-thomas:
I don't think `ok` is a good name since there is not clear what "ok" / "not ok" stands for. `Interrupted` seems to describe best what the boolean value currently expresses.

Karsten1987:
I personally also vote for keeping it as `ok`. I believe that the snippet like
```
while rclcpp::ok() {
[...]

PR #522, opened 2018-07-19
Implement rclcpp::is_initialized()

chapulina:
Addresses https://github.com/ros2/rclcpp/issues/518.

* Named the function `is_initialized` instead of `isInitialized` because that seemed to be the style throughout the utilities
* It turns out that `rcl_ok` was already doing what was desired, so `rclcpp::is_initialized` is just a wrapper around that.
* Added new test file so it could run isolated

### DIFF
--- a/rclcpp/test/test_init.cpp
+++ b/rclcpp/test/test_init.cpp
@@ -17,7 +17,6 @@
 #include "rclcpp/utilities.hpp"
 
 TEST(TestInit, is_initialized) {
-
   EXPECT_FALSE(rclcpp::is_initialized());
 
   rclcpp::init(0, nullptr);
@@ -28,4 +27,3 @@ TEST(TestInit, is_initialized) {
 
   EXPECT_TRUE(rclcpp::is_initialized());
 }
-
--- a/rclcpp/test/test_init.cpp
+++ b/rclcpp/test/test_init.cpp
@@ -1,0 +1,31 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "rclcpp/utilities.hpp"
+
+TEST(TestInit, is_initialized) {
+
+  EXPECT_FALSE(rclcpp::is_initialized());
+
+  rclcpp::init(0, nullptr);
+
+  EXPECT_TRUE(rclcpp::is_initialized());
+
+  rclcpp::shutdown();
+
+  EXPECT_TRUE(rclcpp::is_initialized());
+}
+
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -386,6 +386,14 @@ if(BUILD_TESTING)
     target_link_libraries(test_utilities ${PROJECT_NAME})
   endif()
 
+  ament_add_gtest(test_init test/test_init.cpp
+    APPEND_LIBRARY_DIRS "${append_library_dirs}")
+  if(TARGET test_init)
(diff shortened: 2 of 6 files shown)


mikaelarguedas:
hmm that's where the naming will get confusing..
Change is good as `rcl_ok` actually means "is initialized". It's just confusing that it doesn't have the same semantic as `rclcpp::ok`.

:+1: lgtm with green CI

wjwwood:
@mikaelarguedas that's why originally I had suggested that should be the semantic of `rclcpp::ok()`: https://github.com/ros2/rclcpp/issues/3#issuecomment-369387423

I'm fine with having an `is_initialized()` as @chapulina has implemented here, but I think `ok` should probably also change in `rclcpp` to match how it worked (or at least how it is documented) in `roscpp`.

wjwwood:
Or the name in `rcl` should also change to `rcl_is_initialized()`. That would also be ok by me. At the very least we should ticket the need to change the name of `rcl_ok` if we merge this without doing it directly.

rclcpp/test/test_init.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+  EXPECT_FALSE(rclcpp::is_initialized());
+
+  rclcpp::init(0, nullptr);
+
+  EXPECT_TRUE(rclcpp::is_initialized());
+
+  rclcpp::shutdown();
+
+  EXPECT_TRUE(rclcpp::is_initialized());
```
wjwwood: What happens if you do:

- `init`
- `shutdown`
- `init`
- `EXPECT_TRUE(is_initialized)`

I would expect `is_initialized()` to be false after shutdown, is that not the expectation? I believe `rcl_ok` should return false if you call `rcl_shutdown` (maybe we're not doing that properly yet).

wjwwood: Sorry the previous comment was really two different questions, let me edit it real quick.

wjwwood: Yeah, it looks like we're not calling `rcl_shutdown` ever, which until now hasn't had any consequence (none of our middlewares do anything on `rcl_shutdown`/`rmw_shutdown`):

https://github.com/ros2/rclcpp/blob/8f2052d65a8b05e19958db5484f5425c688098b8/rclcpp/src/rclcpp/utilities.cpp#L270-L288

But now it means that `rcl_ok()`'s return value doesn't change when `rclcpp::shutdown()` is used, despite that really being the intention:

- https://github.com/ros2/rcl/blob/d3a734f2661a05fb1ab0746907a3477a02de8dbd/rcl/include/rcl/rcl.h#L137
- https://github.com/ros2/rcl/blob/d3a734f2661a05fb1ab0746907a3477a02de8dbd/rcl/test/rcl/test_rcl.cpp#L140-L143

chapulina: I added the last expectation just to make it clear that the current behaviour is that once initialized, there's no way back.

---

I don't think `init` -> `shutdown` -> `init` works, I tried shutting down in the end of `TestUtilities.init_with_args` and adding a new `init` test after that but an exception is thrown.

---

I'm not sure what the consequences of calling `rcl_shutdown` could be and whether it would be worth doing it on this PR. Whatcha think?

wjwwood: I don't know, but I don't want to block this. I think it's clear more work needs to be done though (after merging this).

wjwwood:
Btw, @chapulina should this be in review? or are you still working on it?

chapulina:
Not working on it, unless the maintainers think the `shutdown` issue must be fixed before getting this in.

wjwwood:
@mikaelarguedas said it lgth, and I don't want to block it. I'll just make a follow up issue.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4948/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1768/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4107/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4964/)

chapulina:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4953/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1773/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4969/)

wjwwood:
I think this issue covers the improvements I talked about before (being able to do init-shutdown-init, etc.): https://github.com/ros2/rclcpp/issues/154

Issue #482, opened 2018-05-29
Duretion: why no .seconds()?


AndyZe:
## Bug report

**Required Info:**

- Operating System:
 Ubuntu 16.04
- Installation type:
  - binary
- Version or commit hash:
  - rclcpp: 0.4.0-0xenial-20180208-1432
- Client library (if applicable):
  - rclcpp

#### Feature description
Feedback from a new ROS2 user: I think the new clock is unnecessarily confusing. Why does the constructor take seconds and nanoseconds, but there's only a method to access nanoseconds?

When I retrieve nanoseconds like so:
```
rclcpp::Duration delta_t_;
delta_t_.nanoseconds();
```

Am I getting the total time (seconds+nanoseconds)? Or is it the modulo?

My initial thought was that I should do this to convert to seconds, but I have a feeling this is not correct:
`delta_t_.nanoseconds()/1000000000;`

wjwwood:
There's not floating point seconds method right now. You could do `duration.nanoseconds() / 1e9` to get floating seconds. There's no reason to not have a `seconds() -> double` method, I think a pr would be accepted.

I will say that the Time interfaces (along with others) are still subject to change in C++. For one thing, the steady clock and system clocks should not be diffable and should produce a compiler error, but I don't think they do right now.

dirk-thomas:
Might such a method be subject to precision loss? In that case it might be better to not provide it.

tfoote:
A conversion to double will potentially loose information. If we provide it it should have a explicit/unique signature so it's not accidentally used. Can you just use the datatypes instead of an untyped double?

AndyZe:
My use case is multiplying a double to integrate:

`error_integral_ += error_.at(0) * delta_t_.nanoseconds()/1e9;`

wjwwood:
As long as the loss of precision is documented I think it would be ok, but I agree avoiding doing that lossy conversion for the user is not ideal.

@AndyZe to answer your question

> Why does the constructor take seconds and nanoseconds, but there's only a method to access nanoseconds?

more directly: it takes seconds and nanoseconds because that's the message storage format, and it has an option to return nanoseconds (that's "as nanoseconds" not the nanoseconds component of the time) because that's underlying storage format.

Personally, I'd like to see the time and duration messages use a single integer for storage rather than the split pair, but that would be a disruptive change and I haven't had time to push for that to happen.

> Am I getting the total time (seconds+nanoseconds)? Or is it the modulo?

[...]

PR #526, opened 2018-07-27
Add rclcpp::Time::seconds()

sloretz:
~~(edit) ros2/rclcpp#482~~ Edit oops, this doesn't fix 482.  That ticket is about duration types.

This adds `seconds()` which returns the seconds since epoch as a double.

Connects to ros2/geometry2#67

### DIFF
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -103,6 +103,8 @@ public:
   nanoseconds() const;
 
   /// \return the seconds since epoch as a floating point number.
+  /// \warning Depending on sizeof(double) there could be significant precision loss.
+  /// When an exact time is required use nanoseconds() instead.
   RCLCPP_PUBLIC
   double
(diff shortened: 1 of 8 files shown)


rclcpp/include/rclcpp/time.hpp
```diff
@@ -102,6 +102,11 @@ class Time
   rcl_time_point_value_t
   nanoseconds() const;
 
+  /// \return the seconds since epoch as a floating point number.
```
dirk-thomas: Should the docblock contain a note/warning about potential rounding errors? I think that was one reason to omit the function until now.

sloretz: Added warning in 7b736c955cddc636c2bf57e16d8c8353a02ae225

sloretz:
Thanks for the reviews. CI in https://github.com/ros2/geometry2/pull/67#issuecomment-411537227 looks green, so I'll merge this.

PR #530, opened 2018-08-09
Compile flag fixes

jpsamper2009:
## Description
- Courtesy of @serge-nikulin 
- This PR fixes warning generated when building with:

```
-Wall
-Wextra
-pedantic
-Wcast-align
-Wunused
-Wconversion
-Wsign-conversion
-Wdouble-promotion
-fvisibility=hidden
```

- Related to ros2/rcl#276 and ros2/rcutils#114

connects to ros2/rcutils#114

### DIFF
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -55,7 +55,8 @@ Duration::Duration(const Duration & rhs)
 Duration::Duration(
   const builtin_interfaces::msg::Duration & duration_msg)
 {
-  rcl_duration_.nanoseconds = RCL_S_TO_NS(static_cast<uint64_t>(duration_msg.sec));
+  rcl_duration_.nanoseconds =
+    static_cast<rcl_duration_value_t>(RCL_S_TO_NS(static_cast<uint64_t>(duration_msg.sec)));
   rcl_duration_.nanoseconds += duration_msg.nanosec;
 }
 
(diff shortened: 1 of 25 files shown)


rclcpp/src/rclcpp/duration.cpp
```diff
@@ -185,9 +186,11 @@ bounds_check_duration_scale(int64_t dns, double scale, uint64_t max)
 {
   auto abs_dns = static_cast<uint64_t>(std::abs(dns));
   auto abs_scale = std::abs(scale);
-
-  if (abs_scale > 1.0 && abs_dns > static_cast<uint64_t>(max / abs_scale)) {
-    if ((dns > 0 && scale > 0) || (dns < 0 && scale < 0)) {
+  long double max_ld = static_cast<long double>(max);
```
clalancette: Do we actually need the extra precision of a `long double` here?  I'm particularly worried that this will have a performance impact because this will probably end up calling software long double routines.  Further, we already know that (in the one and only caller to this), that max is going to be `std::numeric_limites<rcl_duration_value_t>::max()`, so probably INT64_MAX, which I don't think we need to store in a `long double`.  I haven't fully thought through all of the implications here, so I'm glad to hear arguments about why a `long double` would be needed.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -96,9 +96,10 @@ set_signal_handler(int signal_value, signal_handler_t signal_handler)
     strerror_s(error_string, error_length, errno);
 #endif
     // *INDENT-OFF* (prevent uncrustify from making unnecessary indents here)
-    throw std::runtime_error(
-      std::string("Failed to set SIGINT signal handler: (" + std::to_string(errno) + ")") +
-      error_string);
+    std::string error_message("Failed to set SIGINT signal handler: (");
```
clalancette: Any particular reason for this change?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -240,9 +241,9 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
   }
 
   std::vector<std::string> return_arguments;
-  return_arguments.resize(nonros_argc);
+  return_arguments.resize(static_cast<uint32_t>(nonros_argc));
```
clalancette: I think this should probably be a `static_cast<size_t>`, given https://en.cppreference.com/w/cpp/container/vector/resize, no?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -240,9 +241,9 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
...
 
-  for (int ii = 0; ii < nonros_argc; ++ii) {
+  for (uint32_t ii = 0U; ii < static_cast<uint32_t>(nonros_argc); ++ii) {
```
clalancette: If we are casting `nonros_argc` to a number twice, we should probably just store it above and reuse it rather than continually casting.  This may also change to `size_t` based on the previous comment.

jpsamper2009:
This PR was a for a much older version of the code. I'll close it and readdress with a newer version (if the issues still exist).

clalancette:
> This PR was a for a much older version of the code. I'll close it and readdress with a newer version (if the issues still exist).

I'm actually closing this for now.  @jpsamper2009 feel free to reopen if you disagree.

Issue #524, opened 2018-07-25
create a Time and Duration Max static method


tfoote:
This will avoid requiring users to do things like this in algorithms and tests.

`rclcpp::Duration(std::numeric_limits<int32_t>::max(),999999999)`

PR #538, opened 2018-08-20
Add TIME_MAX and DURATION_MAX functions

sagniknitr:
Related to issue #524 .

-> I am not sure about the tests which I added to validate the functions.Would require some help in them.
-> I have done the linting checks locally in my workstation.

### DIFF
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -217,7 +217,7 @@ Duration::nanoseconds() const
 Duration
 Duration::DURATION_MAX()
 {
-  Duration duration_max(std::numeric_limits<int32_t>::max(),999999999);
+  Duration duration_max(std::numeric_limits<int32_t>::max(), 999999999);
   return duration_max;
 }
 
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -215,10 +215,9 @@ Duration::nanoseconds() const
 }
 
 Duration
-Duration::DURATION_MAX()
+Duration::max()
 {
-  Duration duration_max(std::numeric_limits<int32_t>::max(), 999999999);
-  return duration_max;
+  return Duration(std::numeric_limits<int32_t>::max(), 999999999);
 }
 
 double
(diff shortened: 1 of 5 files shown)


dirk-thomas:
The PR currently has merge conflicts. After resolving them please also run the tests of the package and address any linter warnings.

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -214,6 +214,13 @@ Duration::nanoseconds() const
   return rcl_duration_.nanoseconds;
 }
 
+Duration
+Duration::DURATION_MAX()
+{
+  Duration duration_max(std::numeric_limits<int32_t>::max(), 999999999);
```
sloretz: @tfoote This PR and #524 say a duration is limited by the max value that can be given to the constructor accepting seconds and nanoseconds, but there is [another constructor taking nanoseconds directly](https://github.com/ros2/rclcpp/blob/25a9b4e339867bacb8ba5b92b955420cec86596a/rclcpp/src/rclcpp/duration.cpp#L45), and that seems to accept a number of nanoseconds as large as the [64bit `rcutils` type](https://github.com/ros2/rcutils/blob/4cbd05d9c09886754f41747afa705f15d02c6818/include/rcutils/time.h#L46). Should this instead be `std::numeric_limits<rcl_duration_value_t>::max()`? Alternatively, should the nanoseconds constructor restrict the maximum size to int32_t seconds + 999999999 nanoseconds?

tfoote: I think that the nanoseconds constructor should limit it to this representation. Otherwise we'll overflow the message datastructure.

rclcpp/src/rclcpp/time.cpp
```diff
@@ -251,5 +251,12 @@ operator+(const rclcpp::Duration & lhs, const rclcpp::Time & rhs)
   return Time(lhs.nanoseconds() + rhs.nanoseconds(), rhs.get_clock_type());
 }
 
+Time
+TIME_MAX()
+{
+  Time time_max(std::numeric_limits<int32_t>::max(), 999999999);
```
tfoote: This seems like it would be easier to read as a one line function. No need to name the temporary variable, it could just be returned.

rclcpp/src/rclcpp/duration.cpp
tfoote: This seems like it would be easier to read as a one line function. No need to name the temporary variable, it could just be returned.

dirk-thomas:
Thank you for the patch and for iterating on it.

Issue #532, opened 2018-08-16
Remove asymmetry in subscribing/publishing serialized messages


Martin-Idel-SI:
## Small Feature request

Remove asymmetry in subscribing/publishing to serialized message

#### Current behavior

I can subscribe to a serialized message directly by calling `create_subscription` with a callback of the form
```
[](std::shared_ptr<rcutils_char_array_t> msg) { ... }
```
However, I call publish handing in a raw pointer:
https://github.com/ros2/rclcpp/blob/6b34bcc94c12e2213b6de7af57c3d10109402906/rclcpp/include/rclcpp/publisher.hpp#L280

#### Improved behavior

I can call publish also by handing in a `shared_ptr` of a serialized message.

sloretz:
ping @Karsten1987 , mind having a look at this one?

Karsten1987:
@Martin-Idel-SI Sorry for the late response. I somehow didn't see that issue in my inbox.
I'll have a look at it.

PR #541, opened 2018-08-24
publish shared_ptr of rcl_serialized_message

Karsten1987:
fixes ros2/rclcpp#532

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -289,6 +289,12 @@ public:
     }
   }
 
+  void
+  publish(std::shared_ptr<rcl_serialized_message_t> serialized_msg)
+  {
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -289,6 +289,12 @@ class Publisher : public PublisherBase
     }
   }
 
+  void
+  publish(std::shared_ptr<rcl_serialized_message_t> serialized_msg)
```
dirk-thomas: Should the parameter be const to avoid modification?

Karsten1987: done in https://github.com/ros2/rclcpp/pull/541/commits/7e5b4168b402f50b576fe7d94d2e82cb8adb7479

PR #551, opened 2018-09-06
Implement get_parameter_or_set_default.

clalancette:
This is syntactic sugar to allow the user to get a parameter.
If the parameter is already set on the node, it gets the value
of the parameter.  If it is not set, then it gets the alternative
value and sets it on the node, ensuring that it exists.

I personally think that the name makes some sense, but I'm definitely open to other ideas for the API.  If approved and merged, this will be used in https://github.com/ros2/teleop_twist_joy/pull/8 .


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5139/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1917/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4264/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5118/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -321,6 +321,24 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * is assigned the "alternative_value", and the parameter is set to the
+   * "alternative_value" on the node.
+   *
+   * \param[in] name The name of the parameter to get.
+   * \param[out] value The output where the value of the parameter should be assigned.
+   * \param[in] alternative_value Value to be stored in output and parameter if the parameter was not set.
+   */
+  template<typename ParameterT>
+  void
+  get_parameter_or_set_default(
```
mjcarroll: What's the difference between this and the function above?

clalancette: I probably should have been more clear in the initial description.  `get_parameter_or` gets the parameter if it exists, and stores it in the output value.  If it doesn't exist, it stores the alternative value in the output value.  In all cases, it doesn't change the parameter existence on the node (so `ros2 param list` would not show the parameter in the list).  `get_parameter_or_set_default` gets the parameter if it exists, and stores it in the output value.  If it doesn't exist, it stores the alternative value in the output value.  In all cases, the parameter is guaranteed to exist on the node after the call, either with the previously set value, or with the new default value.  Does that make sense?

mjcarroll: I didn't read the docstring close enough to get the distinction, but this certainly makes sense.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -321,6 +321,24 @@ class Node : public std::enable_shared_from_this<Node>
     ParameterT & value,
     const ParameterT & alternative_value) const;
 
+  /// Get the parameter value; if not set, set the "alternative value" and store it in the node.
+  /**
+   * If the parameter is set, then the "value" argument is assigned the value
+   * in the parameter.  If the parameter is not set, then the "value" argument
```
sloretz: nit 1 sentence per line

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -250,6 +250,22 @@ Node::get_parameter_or(
   return got_parameter;
 }
 
+template<typename ParameterT>
+void
+Node::get_parameter_or_set_default(
```
sloretz: I think this could be shortened to `get_parameter_or_set`, but that's subjective of course.

rclcpp/include/rclcpp/node.hpp
clalancette: Fixed now.

rclcpp/include/rclcpp/node_impl.hpp
clalancette: > I think this could be shortened to `get_parameter_or_set`, but that's subjective of course.

True.  I don't have a strong opinion, though I typically prefer a bit more verbose.  I'm going to leave it as `get_parameter_or_set_default` unless someone voices a strong opinion for something else.

(and thanks for the review!)

clalancette:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5176/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1950/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4302/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5158/)

rclcpp/include/rclcpp/node_impl.hpp
dirk-thomas: Since it isn't setting the default value (since a parameter in the server side doesn't have a default value I would second the suggestion from @sloretz.

clalancette: All right, I renamed to `get_parameter_or_set` in  0bd546d.

PR #557, opened 2018-09-14
Output rcl error message when yaml parsing fails

dhood:
Now it'll give something like:

```
  what():  Failed to parse parameters from file '/tmp/launch_params_1dz53yuw': Sequence should be of same type. Value type 'integer' do not belong at line_num 8, at /home/dhood/ros2_ws/src/ros2/rcl/rcl_yaml_param_parser/src/parser.c:914
```

for linters: [Build Status](https://ci.ros2.org/job/ci_linux/5190/)

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -18,6 +18,7 @@
 
 #include <map>
 #include <memory>
+#include <sstream>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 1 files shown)


rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -120,7 +121,11 @@ NodeParameters::NodeParameters(
       throw std::bad_alloc();
     }
     if (!rcl_parse_yaml_file(yaml_path.c_str(), yaml_params)) {
-      throw std::runtime_error("Failed to parse parameters " + yaml_path);
+      std::ostringstream ss;
+      ss << "Failed to parse parameters from file '" << yaml_path << "': " <<
+        rcl_get_error_string_safe();
+      rcl_reset_error();
```
wjwwood: Maybe use the logger?

wjwwood: Disregard :) Not `cout`'ing...

PR #558, opened 2018-09-15
Add max_duration to spin_some()

sloretz:
This adds a parameter to `spin_some()` for the maximum duration to spend executing work. It prevents `spin_some()` from blocking forever if there are always callbacks ready. The reason for adding it is to allow rviz to execute all callbacks without being limited by the gui or blocking it.

connects to ros2/ros1_bridge#133
Blocks ros2/rviz#354

CI https://github.com/ros2/rviz/pull/354#issuecomment-421519940

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -190,10 +190,14 @@ public:
    * single-threaded model of execution.
    * Adding subscriptions, timers, services, etc. with blocking callbacks will cause this function
    * to block (which may have unintended consequences).
+   *
+   * \param[in] max_duration The maximum amount of time to spend executing work, or 0 for no limit.
+   * Note that spin_some() may take longer than this time as it only returns once max_duration has
+   * been exceeded.
    */
   RCLCPP_PUBLIC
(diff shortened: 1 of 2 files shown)


rclcpp/include/rclcpp/executor.hpp
```diff
@@ -190,10 +190,14 @@ class Executor
...
    * to block (which may have unintended consequences).
+   *
+   * \param[in] max_duration The maximum amount of time to spend executing work, or 0 for no limit.
+   * Note that spin_some() may take longer than this time as it only returns once max_duration has
+   * been exceeded.
    */
   RCLCPP_PUBLIC
   virtual void
-  spin_some();
+  spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0));
```
dirk-thomas: Would `timout` match existing namig better?

sloretz: I think the behavior is different. I would expect `timeout` to mean the function blocks until that amount of time elapses, but this function will return immediately when no work is available regardless of `max_duration`.

Issue #515, opened 2018-07-17
Node unable to avoid using sim time if /clock published


dhood:
Even if a node has its `use_sim_time` parameter set to false, if a message is received by the TimeSource on the `/clock` topic, the node's clock will start using sim time.

This branch has added a test that fails but I would expect to pass: https://github.com/ros2/rclcpp/blob/07567dfbcd56803d0cb5ff8afe23e6b3171d02b6/rclcpp/test/test_time_source.cpp#L339..L341

The presence of a publisher on the clock topic currently causes nodes with unset `use_sim_time` parameter value to start using sim time (see https://github.com/ros2/rclcpp/issues/514 if you want to discuss that), but I don't think it should happen for those that have explicitly set the parameter to false.

If others agree we can label this as a bug and then only enable ROS time [upon receipt of /clock msg](https://github.com/ros2/rclcpp/blob/ae6f8e3e9a6cab18877ce885ad795c1c339b7d8e/rclcpp/src/rclcpp/time_source.cpp#L203) if the parameter has not explicitly been set. That would keep the rest of the existing behaviour the same (again, if you want to discuss the behaviour on unset `use_sim_time` parameter, see https://github.com/ros2/rclcpp/issues/514).

wjwwood:
In my opinion, it should definitely be opt-in (presence of a publisher on `/clock` is not sufficient), though maybe a one time warning is appropriate if `use_sim_time` is unset but there is a publisher (it might be an indication that there is a configuration issue). Explicitly setting `use_sim_time` to False could disable the warning.

sloretz:
+1 to opt-in

Is a `/clock` subscription necessary when `use_sim_time` is `false`? Not having it would prevent a warning like @wjwwood described. It also would add a delay if the subscription had to be created later when `use_sim_time` is dynamically enabled, but I'm not sure if that's a problem.

tfoote:
The behavior in ROS1 is to not even subscribe unless the parameter is set. It's actually required that the parameteter is set before startup as the subscription is only created during the startup phase of a ROS1 node. I implemented this off of memory from the first implementation of ROS1 time but that evolved since that generation. As I mentioned in #514 we might want to take advantage of the parameter callbacks and support switching based on the parameter instead of requiring it at startup. Code can be robust to the change in time by using the TimeJump callbacks.

dhood:
[...]

PR #559, opened 2018-09-20
Don't auto-activate ROS time if clock topic is being published

### LINKED ISSUE #514: Nodes default to using ROS time if /clock is published
The presence of a publisher on the `/clock` topic causes nodes with unset `use_sim_time` parameter value to start using sim time. [See this test](https://github.com/ros2/rclcpp/blob/d298fa4445c49ce2229752f79e92533bab1caaaa/rclcpp/test/test_time_source.cpp#L112..L113).

I haven't found a discussion where we decided that should be the case, so wanted to highlight the current state of ROS time being "opt out" instead of "opt in" and verify it's what we want.

dhood:
closes #514 and #515

Previously, the clock subscription was always created. When messages were received on the clock topic, ROS time was set to active (#514), even if `use_sim_time` was set to false (#515).

Now, ROS time will only be set active if the `use_sim_time` parameter is explicitly set true. The subscription is only created if the parameter is set (can be set after node startup).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5218/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1980/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4338/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5196/)

In progress because CI might uncover things relying on the previous behaviour (the /chatter topic won't appear by default now)

### DIFF
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -189,10 +189,6 @@ void TimeSource::create_clock_sub()
 void TimeSource::destroy_clock_sub()
 {
   std::lock_guard<std::mutex> guard(clock_sub_lock_);
-  if (!clock_subscription_) {
-    // Subscription already destroyed/never created.
-    return;
-  }
   clock_subscription_.reset();
 }
 
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -34,10 +34,10 @@ namespace rclcpp
 {
 
 TimeSource::TimeSource(std::shared_ptr<rclcpp::Node> node)
-: ros_time_active_(false)
+: clock_subscription_(nullptr),
+  ros_time_active_(false)
 {
   this->attachNode(node);
-  clock_subscription_.reset();
 }
 
 TimeSource::TimeSource()
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -65,28 +65,6 @@ void TimeSource::attachNode(
   node_services_ = node_services_interface;
   // TODO(tfoote): Update QOS
 
-  const std::string topic_name = "/clock";
-
-  rclcpp::callback_group::CallbackGroup::SharedPtr group;
-  using rclcpp::message_memory_strategy::MessageMemoryStrategy;
-  auto msg_mem_strat = MessageMemoryStrategy<MessageT, Alloc>::create_default();
-  auto allocator = std::make_shared<Alloc>();
-
-  auto cb = std::bind(&TimeSource::clock_cb, this, std::placeholders::_1);
-
-  clock_subscription_ = rclcpp::create_subscription<
-    MessageT, decltype(cb), Alloc, MessageT, SubscriptionT
-    >(
-    node_topics_.get(),
-    topic_name,
-    std::move(cb),
-    rmw_qos_profile_default,
-    group,
-    false,
-    false,
-    msg_mem_strat,
-    allocator);
-
   parameter_client_ = std::make_shared<rclcpp::AsyncParametersClient>(
     node_base_,
     node_topics_,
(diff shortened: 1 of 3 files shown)


rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -190,6 +194,7 @@ void TimeSource::on_parameter_event(const rcl_interfaces::msg::ParameterEvent::S
     if (it.second->value.bool_value) {
       parameter_state_ = SET_TRUE;
       enable_ros_time();
+      create_clock_sub();
```
wjwwood: Is this in danger of being called more than once and therefore recreating the topic?

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -190,6 +194,7 @@ void TimeSource::on_parameter_event(const rcl_interfaces::msg::ParameterEvent::S
...
     } else {
       parameter_state_ = SET_FALSE;
       disable_ros_time();
```
wjwwood: Should the topic be destroyed here explicitly? (opposite of `create_clock_sub()`)

dhood: I had reasoned that if the sim param had been set at some point then it wouldn't be so bad to leave the subscription around, but being symmetric might be better hey (so it reflects the same state a node that never had sim time set would have).

wjwwood: Yeah, I'm not sure it will ever come up, but conversely I don't think it would be wrong to remove the subscription if the parameter were allowed to be changed during runtime (not sure where we landed on that point).

dhood: yeah, it can be changed at runtime (there are parameter callbacks that will get triggered): https://github.com/ros2/rclcpp/pull/559/commits/9ddb1ae7faf2dbf5405b470cec1f456186284a9d

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5229/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1990/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4348/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5207/)

rclcpp/src/rclcpp/time_source.cpp
dhood: for posterity: 9ddb1ae7faf2dbf5405b470cec1f456186284a9d

dhood:
oops this was sitting in the in-progress column accidentally. Good for review thanks!

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -37,6 +37,7 @@ TimeSource::TimeSource(std::shared_ptr<rclcpp::Node> node)
 : ros_time_active_(false)
 {
   this->attachNode(node);
+  clock_subscription_.reset();
```
wjwwood: Is there a reason this has to go after `this->attachNode(node);`? It would be better (in my opinion) to do this in the member initialization section of the constructor as `clock_subscription_(nullptr)`.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -180,6 +186,16 @@ void TimeSource::create_clock_sub()
...
 }
 
+void TimeSource::destroy_clock_sub()
+{
+  std::lock_guard<std::mutex> guard(clock_sub_lock_);
+  if (!clock_subscription_) {
+    // Subscription already destroyed/never created.
+    return;
+  }
+  clock_subscription_.reset();
```
wjwwood: This call is always safe, so you can do it ever if `!clock_subscription_`.

rclcpp/src/rclcpp/time_source.cpp
dhood: nah, there wasn't a reason: 73d4a7bf0c945a03a1a35d189394e2a9f5299bb0

rclcpp/src/rclcpp/time_source.cpp
dhood: Updated in e34e0d71dc91143b9915aaa12dc247d4f79c59c9

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5231/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1992/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4350/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5209/)

PR #561, opened 2018-09-25
Remove std::binary_function usage

anis-ladram:
This pull request removes an inherit of `std::binary_function` for the struct `strcmp_wrapper`, a C-style string comparator.

The struct `strcmp_wrapper` is used by the IDTopicMap typedef (line 276) which is a std::map, and std::map requires strcmp_wrapper to fulfill the [Compare named requirement](https://en.cppreference.com/w/cpp/named_req/Compare). The inherit of `std::binary_function` only defines 3 types: `first_argument_type`, `second_argument_type` and `result_type` (typedefs of types passed in argument), which is useless here because they are not part of the requirement.

 `std::binary_function` has been deprecated in C++11, removed in C++17. This change may seem useless, but this change enables us to compile C++17 ROS2 packages using rclcpp (or at least some of them), which may be useful.

### DIFF
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -262,7 +262,7 @@ private:
     std::hash<uint64_t>, std::equal_to<uint64_t>,
     RebindAlloc<std::pair<const uint64_t, SubscriptionBase::WeakPtr>>>;
 
-  struct strcmp_wrapper : public std::binary_function<const char *, const char *, bool>
+  struct strcmp_wrapper
   {
     bool
     operator()(const char * lhs, const char * rhs) const


wjwwood:
I wrote this originally, but I don't remember much about it :D, and I was probably following some online recommendation on how to make the strcmp wrapper.

This seems reasonable to me, but we'll have to run CI on it to see if it breaks older compilers/c++ standards or not.

wjwwood:
Here's CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5234/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1995/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4353/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5212/)

anis-ladram:
Running CI on a pull request, this also seems reasonable to me! ;) Thanks for triggering the CI build.

wjwwood:
lgtm to me, I agree the test failure is unrelated.

PR #567, opened 2018-10-10
Adding a factory method to create a Duration from seconds

crdelsey:
There are many places in the ROS codebase where a time duration is
specified as a floating point number of seconds. A factory function
to create a Duration object from these values makes the code a
bit simpler in many cases.

Sorry this change is based off the 0.5.1 tag. I was having trouble building
the tip of master. I will rebase this properly if you want to implement this
functionality this way.

### DIFF
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -28,10 +28,12 @@ class RCLCPP_PUBLIC Duration
 public:
   Duration(int32_t seconds, uint32_t nanoseconds);
 
+  // This constructor matches any numeric value - ints or floats
   explicit Duration(rcl_duration_value_t nanoseconds);
 
   explicit Duration(std::chrono::nanoseconds nanoseconds);
 
+  // This constructor matches any std::chrono value other than nanoseconds
   // intentionally not using explicit to create a conversion constructor
   template<class Rep, class Period>
   // cppcheck-suppress noExplicitConstructor
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -94,6 +94,10 @@ public:
   double
   seconds() const;
 
+  // Create a duration object from a floating point number representing seconds
+  static Duration
+  from_seconds(double seconds);
+
   template<class DurationT>
   DurationT
   to_chrono() const
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -233,4 +233,10 @@ Duration::to_rmw_time() const
   return result;
 }
 
+Duration
+Duration::from_seconds(double seconds)
+{
+  return Duration(static_cast<int64_t>(RCL_S_TO_NS(seconds)));
+}
+
 }  // namespace rclcpp
--- a/rclcpp/test/test_duration.cpp
+++ b/rclcpp/test/test_duration.cpp
@@ -136,3 +136,20 @@ TEST(TestDuration, maximum_duration) {
 
   EXPECT_EQ(max_duration, max);
 }
+
+static const int64_t HALF_SEC_IN_NS = 500 * 1000 * 1000;
+static const int64_t ONE_AND_HALF_SEC_IN_NS = 3 * HALF_SEC_IN_NS;
+
+TEST(TestDuration, from_seconds) {
+  EXPECT_EQ(rclcpp::Duration(0), rclcpp::Duration::from_seconds(0.0));
+  EXPECT_EQ(rclcpp::Duration(0), rclcpp::Duration::from_seconds(0));
+  EXPECT_EQ(rclcpp::Duration(1, HALF_SEC_IN_NS), rclcpp::Duration::from_seconds(1.5));
+  EXPECT_EQ(rclcpp::Duration(-ONE_AND_HALF_SEC_IN_NS), rclcpp::Duration::from_seconds(-1.5));
+}
+
+TEST(TestDuration, std_chrono_constructors) {
+  EXPECT_EQ(rclcpp::Duration(0), rclcpp::Duration(0.0s));
+  EXPECT_EQ(rclcpp::Duration(0), rclcpp::Duration(0s));
+  EXPECT_EQ(rclcpp::Duration(1, HALF_SEC_IN_NS), rclcpp::Duration(1.5s));
+  EXPECT_EQ(rclcpp::Duration(-1, 0), rclcpp::Duration(-1s));
+}


wjwwood:
> I was having trouble building the tip of master.

You likely need the latest of all packages, not just `rclcpp`. I'd recommend using this `ros2.repos` file as a starting place:

https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos

crdelsey:
Thanks. I just had various repos in a state where I couldn't update easily. I fixed that now by pulling a clean version of the workspace. The PR has been rebased to master now.

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -226,4 +226,10 @@ Duration::seconds() const
   return std::chrono::duration<double>(std::chrono::nanoseconds(rcl_duration_.nanoseconds)).count();
 }
 
+Duration
+duration_from_seconds(double seconds)
```
wjwwood: What do you think about making this a static method on the `Duration` class itself? `rclcpp::Duration::from_seconds(...);` is clear to me than what's in your tests at the moment.

mjcarroll: ROS1 also had a Time/Duration constructor of the signature:

`explicit TimeBase(double t)`
`explicit DurationBase(double t)`

Is there any reason that we couldn't include them here?

wjwwood: Well, with the constructors you have no hint (other than docs) what the units of the double is. At least with `from_seconds(double)` you know it's seconds.

I'd be ok with something like `explicit Duration(std::chrono::nanoseconds)` and `explicit Duration(std::chrono::duration<double>)`, or even:

```c++
template<ChronoDurationT>
Duration(ChronoDurationT duration)
: storage(std::chrono::duration_cast<std::chrono::nanoseconds>(duration).count())
{}
```
(would need more than that, but something like that would get you started)

crdelsey: I was preferring the option with the two separate constructors for `nanoseconds` and `duration<double>`. That way I could keep template complexity out of the header, but it fails when I do `Duration(1s)`. Overload resolution is ambiguous in that case, so I guess I'll go with the template version.

Do you prefer to define the type `ChronoDurationT` in the class or outside the class?

I'm going to go ahead and add the static `from_seconds` as well. I think that works better when the time is a variable instead of a constant. eg
```c++
double t = 1.0;
auto d = Duration::from_seconds(t);  // with static from_seconds method
Duration d(std::chrono::duration<double>(t));  // without from_seconds
```

wjwwood: Yeah, `chrono` can be tricky sometimes, we do something like this here:

- https://github.com/ros2/rclcpp/blob/144c24c8fd15d57382fa994c7ecf00b847b1c339/rclcpp/include/rclcpp/executor.hpp#L162-L166

But I'm not sure if that would let you have two constructors, one for `int64_t` and one for `double` or not. We then forward that to the non-templated one that is always nanoseconds (which can also be implemented in the `.cpp`):

- https://github.com/ros2/rclcpp/blob/144c24c8fd15d57382fa994c7ecf00b847b1c339/rclcpp/include/rclcpp/executor.hpp#L282-L286

I don't mind about the name of the template argument, `TimeT` would also be fine I guess. We don't see to have a consistent convention unfortunately.

Thanks for iterating.

crdelsey: That might work if the disambiguation rules prefer matching any integer type over double. However, it still brings template baggage into the header. I hate to do that. It is so beautifully simple right now :-) I'll get something working soon and update the PR for further feedback

clalancette:
CI, to see how this fairs:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5318/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2061/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4424/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5288/)

crdelsey:
I see that the change I made caused an uncrustify failure. I just pushed a change to fix that. I also added a couple of comments to clarify which constructors get matched when called with a numeric-like value.

clalancette:
Sorry for the delay.  Kicking off CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5496/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2165/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5399/)

clalancette:
Apologies, but there have been some API changes since this PR was opened.  Would you mind rebasing this?  I can then kick off CI again on it.  Thanks for your patience.

clalancette:
Thanks.  Another CI try:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5515/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2179/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4551/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5415/)

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -29,17 +29,20 @@ class Duration
...
+  explicit Duration(std::chrono::duration<rep, period> duration)
+  {
+    rcl_duration_.nanoseconds =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(duration).count();
+  }
 
   RCLCPP_PUBLIC
-  Duration(
-    const builtin_interfaces::msg::Duration & duration_msg);
+  Duration(const builtin_interfaces::msg::Duration & duration_msg);
```
clalancette: According to CI, it looks like this constructor needs to be marked explicit.

crdelsey: I didn't want to change that since it was unrelated to this PR. I think that change needs some discussion. IMHO, this constructor is better left implicit. With the implicit constructor, we can say:
```
rclcpp::Duration foo = mymsg.foo;
```
instead of 
```
rclcpp::Duration foo = rclcpp::Duration(mymsg.foo);
```

That seems to me to be a common enough use case that not being explicit would be convenient.

If you agree, I can mark this with // NOLINT

wjwwood: This is a thing, it's called a conversion constructor:

https://en.cppreference.com/w/cpp/language/converting_constructor

And they _explicitly_ do not use the keyword `explicit` (ironically?), and in this case I think the linter is providing a hint that we should make sure we want this and in this case I believe we do. So I would +1 the use of `// NOLINT(runtime/explicit): this is a conversion constructor` here.

clalancette: Fine by me.  I was just confused by the fact that:

1.  There is no logical change to this line, so I'm not sure why the linter is picking it up now (maybe because of the line breaks?), and
1.  Why there is any change to this line in this PR.  Maybe putting it back to the old indentation will suffice?

wjwwood: For sure, that's a valid question. If reverting the whitespace only change also quiets the linter that's fine too (though it doesn't make sense :p).

crdelsey: Apparently the world is a strange place :-) I confirmed that reverting the spacing does indeed make the linter error go away.

I think that's a bit better than the //NOLINT at the end of the line, so I'll revert the space change.

clalancette:
Another CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5540/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2197/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4572/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5433/)

clalancette:
FYI; the Windows build has warnings, some of which are not the fault of this PR, but some of which look like they might be.  In particular, the problems in urdfdom color.h are not related to this PR (they are because of https://github.com/ros/urdfdom_headers/pull/47), but the other ones in test_duration.cpp probably are because of this PR.

crdelsey:
Ok. I'll look into them. Sorry for the churn. Is there a way I can trigger the CI so you don't need to do it manually each time I submit a change?

cottsay:
It doesn't look like there has been any activity on this ticket for a few months. @crdelsey, how would you describe the status of this change? Are you just waiting for further review?

crdelsey:
@cottsay Yes. I believe everything had been addressed as of the last commit. It looks like it is a bit out of date now, though, and needs to be rebased. I'll try to do that in the next day or two.

crdelsey:
@cottsay This PR is rebased and passing my tests locally. I believe it is ready for review again.

cottsay:
@crdelsey, would you mind doing another rebase? We'd like to take another shot at getting this in.

crdelsey:
@cottsay It's rebased again

clalancette:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7456/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3600/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6117/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7293/)

clalancette:
The windows and Linux failures all showed up in the nightlies.  Given that, I'm going to merge; thanks for the contribution!

PR #574, opened 2018-10-27
issue a warning if publishing on a not active publisher

fmrico:
Hi all,

Issue #552 seems to be blocked. @Myzhar did not a PR with his work. This is a PR with his changes.

This PR provides the machanism to issue a WARN when publishing on a not active publisher.

Best

### DIFF
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -91,7 +91,8 @@ def is_supported_feature_combination(feature_combination):
  */
 #define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in get_macro_parameters(feature_combination).keys()]))...) \
   static_assert( \
-    ::std::is_same<std::remove_reference<decltype(logger)>::type, ::rclcpp::Logger>::value, \
+    ::std::is_same<typename std::remove_reference<decltype(logger)>::type, \
+    typename ::rclcpp::Logger>::value, \
     "First argument to logging macros must be an rclcpp::Logger"); \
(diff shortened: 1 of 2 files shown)


Karsten1987:
@fmrico thanks for patch. I believe the same comment for PR #552 applies here:
https://github.com/ros2/rclcpp/pull/562#discussion_r222103054

fmrico:
Hi @Karsten1987 ,

I fixed it following the comments in #562 . Is was not so simple, because I had to change logging macros, to make it compile. Macro has changed to:
```
::std::is_same<typename std::remove_reference<decltype(logger)>::type, \
    typename ::rclcpp::Logger>::value, \
    "First argument to logging macros must be an rclcpp::Logger")
```
adding `typename`, because std::remove_reference<Ty>::type is dependant on a template parameter, as I found [here](https://stackoverflow.com/questions/25342985/nesting-standard-type-traits-with-sfinae-fails-to-compile-with-g-4-8).

I run the tests and everything worked ok.

Best

Karsten1987:
Great to hear you made it work. The patch looks good to me so far.
I've triggered the buildfarrm to see if this patch introduces any regression on other tests:

CI:
[Build Status](https://ci.ros2.org/job/ci_linux/5382/)
[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/2100/)
[Build Status](https://ci.ros2.org/job/ci_osx/4464/)
[Build Status](https://ci.ros2.org/job/ci_windows/5334/)

Myzhar:
Great @fmrico 💪

Issue #529, opened 2018-08-06
Allow `Parameter` API to get/set maps of values


clalancette:
#### Feature description

There are some users of the Parameter API in ROS1 that depend on getting a map of values, e.g. https://github.com/ros-teleop/teleop_twist_joy/blob/indigo-devel/src/teleop_twist_joy.cpp#L78 .  It would be great to have a helper API to do this in ROS2 as well, to reduce the amount of boilerplate that users would have to do to get parameters one at a time.

mikaelarguedas:
Permalink: https://github.com/ros-teleop/teleop_twist_joy/blob/10d79bebbf112f95f32713895f7cd0668faa83fc/src/teleop_twist_joy.cpp#L78

dirk-thomas:
> ... to reduce the amount of boilerplate that users would have to do to get parameters one at a time.

I am not sure what you mean with "one at a time".

Couldn't you create a "syntactic sugar" API which wraps the existing [get_parameters](https://github.com/ros2/rclcpp/blob/4ddb76f466d331e4ca8ff4df7ba25a5dce8496af/rclcpp/include/rclcpp/parameter_client.hpp#L188-L190 method and instead of returning a `std::vector<rclcpp::Parameter>` return a `std::map<std::string, rclcpp::ParameterValue>`?

nuclearsandwich:
> Couldn't you create a "syntactic sugar" API which wraps the existing get_parameters method and instead of returning a std::vector<rclcpp::Parameter> return a std::map<std::string, rclcpp::ParameterValue>?

From talking with @clalancette about this a get_parameters overload which can return a map or fill a provided one would be a good enhancement, although not needed for the work on teleop_twist_joy.
Another method that would be useful (but again isn't necessary for teleop_twist_joy) is one which can get an entire prefix. An example would be something like `std::map<std::string, rclcpp::ParameterValue> get_parameters_by_prefix(std::string prefix);` Given the parameters in our [example on the wiki](https://github.com/ros2/ros2/wiki/Node-arguments#parameters) this method would be used to get a map of all parameters under the prefix `some_lists`

PR #575, opened 2018-10-29
Get parameter map

clalancette:
This PR adds the ability to get a `std::map` set of key-value pairs of parameters.  In this case, parameters that form part of a map are all of the form:

```
param.x
param.y
```

Assuming `param.x` has value 0.1 and `param.y` has value 1.5, then calling this API like:

```
std::map<std::string, double> params;
node->get_parameter("param", params);
```

Would fill in the `params` map with keys of `x` and `y` with values of `0.1` and `1.5`, respectively.  There is also a `set_parameter_if_not_set` that works much in the same way for setting values.  Tests for this new functionality are added in a new test file; I'll move the rest of the "local" parameter tests from system_tests once this is in.  This should solve https://github.com/ros2/rclcpp/issues/529 , and will unblock https://github.com/ros2/teleop_twist_joy/pull/8

connects to ros2/rclcpp#529

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -277,6 +277,12 @@ public:
     const std::string & name,
     const ParameterT & value);
 
+  template<typename MapValueT>
+  void
+  set_parameter_if_not_set(
+    const std::string & name,
+    const std::map<std::string, MapValueT> & values);
+
   RCLCPP_PUBLIC
   std::vector<rclcpp::Parameter>
   get_parameters(const std::vector<std::string> & names) const;
@@ -302,6 +308,23 @@ public:
   template<typename ParameterT>
   bool
   get_parameter(const std::string & name, ParameterT & parameter) const;
+
+  /// Assign the value of the map parameter if set into the values argument.
+  /**
+   * Parameter names that are part of a map are of the form "name.member".
+   * This API gets all parameters that begin with name, storing them into the
+   * map with a key of "member" and their value.  If there are no members in
+   * the named map, then the "values" argument is not changed.
+   *
+   * \param[in] name The name of the map parameter to get.
+   * \param[out] values The map of output values, with one std::string,MapValueT
+   *                    per parameter.
+   * \returns true if values was changed, false otherwise
+   */
+  template<typename MapValueT>
+  bool
+  get_parameter(const std::string & name,
+                std::map<std::string, MapValueT> & values) const;
 
   /// Get the parameter value, or the "alternative value" if not set, and assign it to "value".
   /**
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -323,8 +323,9 @@ public:
    */
   template<typename MapValueT>
   bool
-  get_parameter(const std::string & name,
-                std::map<std::string, MapValueT> & values) const;
+  get_parameter(
+    const std::string & name,
+    std::map<std::string, MapValueT> & values) const;
 
   /// Get the parameter value, or the "alternative value" if not set, and assign it to "value".
   /**
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -223,6 +223,28 @@ Node::set_parameter_if_not_set(
   }
 }
 
+// this is a partially-specialized version of set_parameter_if_not_set above,
+// where our concrete type for ParameterT is std::map, but the to-be-determined
+// type is the value in the map.
+template<typename MapValueT>
+void
+Node::set_parameter_if_not_set(
+  const std::string & name,
+  const std::map<std::string, MapValueT> & values)
+{
+  std::vector<rclcpp::Parameter> params;
+
+  for (const auto & val : values) {
+    std::string param_name = name + "." + val.first;
+    rclcpp::Parameter parameter;
+    if (!this->get_parameter(param_name, parameter)) {
+      params.push_back(rclcpp::Parameter(param_name, val.second));
+    }
+  }
+
+  this->set_parameters(params);
+}
+
 template<typename ParameterT>
 bool
 Node::get_parameter(const std::string & name, ParameterT & value) const
@@ -231,6 +253,25 @@ Node::get_parameter(const std::string & name, ParameterT & value) const
   bool result = get_parameter(name, parameter);
   if (result) {
     value = parameter.get_value<ParameterT>();
+  }
+
+  return result;
+}
+
+// this is a partially-specialized version of get_parameter above,
+// where our concrete type for ParameterT is std::map, but the to-be-determined
+// type is the value in the map.
+template<typename MapValueT>
+bool
+Node::get_parameter(const std::string & name,
+                    std::map<std::string, MapValueT> & values) const
+{
+  std::map<std::string, rclcpp::Parameter> params;
+  bool result = node_parameters_->get_parameters_by_prefix(name, params);
+  if (result) {
+    for (const auto & param : params) {
+      values[param.first] = param.second.get_value<MapValueT>();
+    }
   }
 
   return result;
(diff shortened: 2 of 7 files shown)


clalancette:
CI (up to test_rclcpp, testing only rclcpp and test_rclcpp):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5395/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2111/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4476/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5346/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -277,6 +277,12 @@ class Node : public std::enable_shared_from_this<Node>
     const std::string & name,
     const ParameterT & value);
 
+  template<typename MapValueT>
+  void
+  set_parameter_if_not_set(
```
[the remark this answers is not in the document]
clalancette: I don't feel strongly about the name, but I will point out that in ROS1 all of the methods were named the same, regardless of whether they were setting a map, a vector, or a single value.  I think there is value to this here as well, since there is only "one" name to remember, regardless of the type you want to set.  However, if you insist, I will change the name to be plural.

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -223,6 +223,28 @@ Node::set_parameter_if_not_set(
...
+Node::set_parameter_if_not_set(
+  const std::string & name,
+  const std::map<std::string, MapValueT> & values)
+{
+  std::vector<rclcpp::Parameter> params;
+
+  for (const auto & val : values) {
+    std::string param_name = name + "." + val.first;
+    rclcpp::Parameter parameter;
+    if (!this->get_parameter(param_name, parameter)) {
```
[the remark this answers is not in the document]
clalancette: I can make this work using `get_parameters()`, but I think it is much uglier and less efficient:

```C++
template<typename MapValueT>
void
Node::set_parameter_if_not_set(
  const std::string & name,
  const std::map<std::string, MapValueT> & values)
{
  std::vector<std::string> names;

  for (const auto & val : values) {
    std::string param_name = name + "." + val.first;
    names.push_back(param_name);
  }

  std::vector<rclcpp::Parameter> params = node_parameters_->get_parameters(names);
  std::vector<rclcpp::Parameter> params_to_set;

  for (const auto & val : values) {
    std::string param_name = name + "." + val.first;
    if (std::find_if(params.begin(), params.end(),
                     [&param_name](const rclcpp::Parameter& param) {
                       return param.get_name() == param_name;
                     }) == params.end()) {
      params_to_set.push_back(rclcpp::Parameter(param_name, val.second));
    }
  }

  this->set_parameters(params_to_set);
}
```

This is due to the fact that [get_parameters()](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L236) returns only what is currently in the parameters, and silently ignores the ones that aren't there.  Unless you can think of a way to reduce the number of loops and finds here, I'd prefer to stick to the current implementation.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -303,6 +309,24 @@ class Node : public std::enable_shared_from_this<Node>
   bool
   get_parameter(const std::string & name, ParameterT & parameter) const;
 
+  /// Assign the value of the map parameter if set into the values argument.
+  /**
+   * Parameter names that are part of a map are of the form "name.member".
```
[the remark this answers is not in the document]
clalancette: While that is true (and makes some sense for `set`), it doesn't really make much sense for `get`.  If we just make it so that there is no "grouping" with this API, then there is little reason to add this API; we can do the same thing by calling `get_parameters` with a vector of names.  The explicit purpose of the map-based one is to have a group of variables that we can get into a single map.  This is similar to the map-based [getParam](https://github.com/ros/ros_comm/blob/melodic-devel/clients/roscpp/include/ros/node_handle.h#L1806) overload in ROS1.

clalancette:
@ros2/team There are two open questions on this PR:

1) Whether the APIs that modify multiple values should be plural.  On the one hand, this makes sense from a grammatical sense.  On the other hand, all of the APIs in ROS 1 were named the same, regardless of how many items they touched (I'm thinking of getParam() and friends here).  Opinions?
1)  Should we have a map-based overload at all?  I think it makes it much more convenient to get a group of similar values, but it is simple syntactic sugar and all of the values can just be fetched using a vector of names.  If the answer to this is "no", then I'll just close out this PR and do something different in https://github.com/ros2/teleop_twist_joy/pull/8

clalancette:
(also, I'm going to close and re-open this PR to retrigger the build, since I think the failure was unrelated to this PR)

tfoote:
1. Regarding plural vs singular I'd suggest that we go with plural we're already using the plural in most other situations like `describe parameters` and `get_parameter_types`

2. I think that the std::map based interface is a significant improvement for searching and querying compared to a vector that exposing it will be a significant benefit to the user. It might make sense to implement it as the syntactic sugar, where the API just vectorizes the requests before interacting with the underlying implementation.

jacobperron:
1. The vector based interface is already plural, so I think making the map based interface plural is consistent.
2. +1 for std:map based interface.

clalancette:
Updated to use plurals.  Here's CI up to rclcpp:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6039/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2534/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4979/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5909/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -279,6 +279,12 @@ class Node : public std::enable_shared_from_this<Node>
     const std::string & name,
     const ParameterT & value);
 
+  template<typename MapValueT>
```
wjwwood: We should really have documentation for this method. I know it's a variant of another method that has no documentation, but I feel we should be moving toward better documentation with each new pr.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -305,6 +311,24 @@ class Node : public std::enable_shared_from_this<Node>
   bool
   get_parameter(const std::string & name, ParameterT & parameter) const;
 
+  /// Assign the value of the map parameter if set into the values argument.
+  /**
+   * Parameter names that are part of a map are of the form "name.member".
+   * This API gets all parameters that begin with name, storing them into the
+   * map with a key of "member" and their value.  If there are no members in
```
wjwwood: One sentence per line please.

wjwwood: Also, `with a key of "member"` is confusing to me, is the key literally "member" or the name of the parameter?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -305,6 +311,24 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * the named map, then the "values" argument is not changed.
+   *
+   * \param[in] name The name of the map parameter to get.
```
wjwwood: From my perspective there's no such thing as a "map parameter", i.e. there's no map or dict type for parameters. I think this would be more accurately described as "the prefix of the parameters to get", or "the namespace from which to get parameters".

rclcpp/include/rclcpp/node.hpp
clalancette: Good idea.  Done in f039865cb0cf3b555f7aeefe15f85b4c2af67b12

rclcpp/include/rclcpp/node.hpp
clalancette: I fixed one sentence per line in f039865cb0cf3b555f7aeefe15f85b4c2af67b12

Also in that commit, I changed the wording here to use "name of the parameter" to reduced confusion.

rclcpp/include/rclcpp/node.hpp
clalancette: Yeah, I like "prefix of the parameters to get", so I switched to that terminology.

Issue #531, opened 2018-08-10
add "NodeHandle" like concept


wjwwood:
In ROS 1, there's a construct in `roscpp` called a `NodeHandle`, and it has most of the interface of the Node singleton, but can additionally have "extra" namespace. Having a node handle instance with a namespace does not affect the node namespace/name combo, but all things created using the namespace node handle will have the extra namespace added to it, e.g. publishers, subscriptions, services, parameters, etc...

From https://wiki.ros.org/roscpp/Overview/NodeHandles:

> The ros::NodeHandle class serves two purposes. First, it provides RAII-style startup and shutdown of the internal node inside a roscpp program. Second, it provides an extra layer of namespace resolution that can make writing subcomponents easier.

In `rclcpp`, the `Node` class provides the RAII style lifecycle of the node, so this issue is just about having a new concept in `rclcpp` that satisfies the "provides an extra layer of namespace resolution that can make writing subcomponents easier" part.

However, there is a question about lifetime of the node handle: should it be up to the user to ensure that the node instance out lives all node handles, or should node handles have a weak reference to the node?

Unlike the constructible `ros::NodeHandle`, this new object would be generated by a factor function of the original Node. e.g. `node->get_node_handle("namespace", remapping args, additional parameters, etc.)`.

Again, like the node handles in `roscpp`, it should be possible to build a hierarchy of these objects, something like this:

```c++
auto node_ptr = std::make_shared<rclcpp::Node>("my_node", "my_ns");
auto sub_node = node_ptr->create_sub_node("foo_ns");
auto another_sub_node = sub_node->create_sub_node("bar_ns");
// topic would be '/my_ns/foo_ns/bar_ns/chatter'
auto sub = another_sub_node->create_subscription<...>("chatter", ...);
```

This request is being driven by existing use cases in the porting of gazebo ros packages, and is especially useful if you have components which may be "attached" to nodes in arbitrary configurations. These components would take something that looks like a node and use it to create pub/sub, services, etc... but ideally would not need to manually apply namespaces itself.

----

Summary of design issues and goals:

- Provide API of node, but modify namespace of topics, services, parameters, etc...
- Provide custom but scoped remappings
- Provide additional parameter value initialization/declaration (see: https://github.com/ros2/rclcpp/blob/6b34bcc94c12e2213b6de7af57c3d10109402906/rclcpp/include/rclcpp/node.hpp#L101)
- Nestable
- Name bikesheding, how to call it?
[...]

PR #578, opened 2018-11-02
SubNode: add "NodeHandle" like concept

fmrico:
Hi all,

This PR is related to issue #531 , providing a SubNode class (implemented as a Proxy Pattern). This let to have code like:

``` 
 auto node_ptr = std::make_shared<rclcpp::Node>("my_node", "/my_ns");

  auto sub_node = node_ptr->create_sub_node("/foo_ns");
  auto another_sub_node = sub_node->create_sub_node("/bar_ns");

  // topic would be '/my_ns/foo_ns/bar_ns/chatter'
  auto sub0 = node_ptr->create_subscription<std_msgs::msg::String>("chatter0",topic_callback);
  auto sub1 = sub_node->create_subscription<std_msgs::msg::String>("chatter1",topic_callback);
  auto sub2 = another_sub_node->create_subscription<std_msgs::msg::String>("chatter2",topic_callback);
```
which produces these topics:

```
/my_ns/chatter0
/my_ns/foo_ns/bar_ns/chatter2
/my_ns/foo_ns/chatter1
```

This is a possible starting point to have the desired functionality. 

Best

### DIFF
--- a/rclcpp/src/rclcpp/sub_node.cpp
+++ b/rclcpp/src/rclcpp/sub_node.cpp
@@ -31,122 +31,111 @@
 #include "rclcpp/node_interfaces/node_services.hpp"
 #include "rclcpp/node_interfaces/node_timers.hpp"
 #include "rclcpp/node_interfaces/node_topics.hpp"
+#include "rmw/validate_namespace.h"
+
 #include "rclcpp/sub_node.hpp"
 
-using rclcpp::Node;
+
+using rclcpp::SubNode;
 using rclcpp::exceptions::throw_from_rcl_error;
 
-Node::Node(
-  const std::string & node_name,
-  const std::string & namespace_,
-  bool use_intra_process_comms)
-: Node(
-    node_name,
-    namespace_,
-    rclcpp::contexts::default_context::get_global_default_context(),
-    {},
-    {},
-    true,
-    use_intra_process_comms,
-    true)
-{}
-
-Node::Node(
-  const std::string & node_name,
-  const std::string & namespace_,
-  rclcpp::Context::SharedPtr context,
-  const std::vector<std::string> & arguments,
-  const std::vector<rclcpp::Parameter> & initial_parameters,
-  bool use_global_arguments,
-  bool use_intra_process_comms,
-  bool start_parameter_services)
-: node_base_(new rclcpp::node_interfaces::NodeBase(
-      node_name, namespace_, context, arguments, use_global_arguments)),
-  node_graph_(new rclcpp::node_interfaces::NodeGraph(node_base_.get())),
-  node_logging_(new rclcpp::node_interfaces::NodeLogging(node_base_.get())),
-  node_timers_(new rclcpp::node_interfaces::NodeTimers(node_base_.get())),
-  node_topics_(new rclcpp::node_interfaces::NodeTopics(node_base_.get())),
-  node_services_(new rclcpp::node_interfaces::NodeServices(node_base_.get())),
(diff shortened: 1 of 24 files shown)


rclcpp/src/rclcpp/sub_node.cpp
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  return ret_ns;
+}
+
+RCLCPP_PUBLIC
+const std::string
+SubNode::add_extended_ns_prefix(const std::string & original_name) const
+{
+  return std::string(original_->get_namespace()) + "/" + extended_namespace_ + "/" + original_name;
```
sloretz: This works for relative names `foo` or `foo/bar`, but I'm not sure about the behavior for fully qualified or private names `~/bar`.

ROS 1 threw `ros::InvalidNameException` when trying to create a publisher/subscriber/etc with a private name, asking instead that the `NodeHandle` be created with a private namespace. In ROS 2 the `create_...()` APIs allow private names. I'm not sure how to handle all the cases, but there is a proposal here: https://github.com/ros2/rclcpp/issues/531#issuecomment-437201989

rclcpp/test/test_node.cpp
```diff
@@ -75,6 +75,39 @@ TEST_F(TestNode, get_name_and_namespace) {
...
+    auto node = std::make_shared<rclcpp::Node>("my_node", "/ns");
+    auto subnode = node->create_sub_node("sub_ns");
+    EXPECT_STREQ("my_node", subnode->get_name());
+    EXPECT_STREQ("/ns/sub_ns", subnode->get_namespace());
+  }
+  {
+    auto node = std::make_shared<rclcpp::Node>("my_node", "/ns");
+    auto subnode = node->create_sub_node("/sub_ns");
+    EXPECT_STREQ("my_node", subnode->get_name());
+    EXPECT_STREQ("/ns/sub_ns", subnode->get_namespace());
```
sloretz: Given that the subnode's namespace is fully qualified, I think this result should be `/sub_ns`. This also matches ROS 1 when a fully qualified name is given as a child `NodeHandle` namespace.

```c++
#include "ros/ros.h"
#include "std_msgs/String.h"

int main(int argc, char ** argv)
{
  ros::init(argc, argv, "name");

  ros::NodeHandle n;
  ros::NodeHandle nsub(n, "/sub_ns");
  ros::Publisher pub = nsub4.advertise<std_msgs::String>("nsub/relaive", 1000);

  ros::spin();
  return 0;
}
```

launched with `/path/to/above/executable __ns:=/orig_ns`

Creates the topic `/sub_ns/nsub/relative`, not `/orig_ns/sub_ns/...`

wjwwood: We could also consider making it so that "sub nodes" can only use relative name prefix or a prefix starting with a `~`. That would avoid the issue of ignoring the node namespace.

wjwwood:
Yeah, unfortunately I think the design issues here (like the amount of duplicated code) are going to impede progress. Maybe we could try and setup a voice chat at some point to hash out some of the details.

@fmrico are you interested at all in doing that?

fmrico:
Hi, 

Yes, there is a lot of duplicated code because I did not want that SubNode inherits from Node, but the interface had to be the same. I could reconsider this or find another solution.

@wjwwood I am interested in doing that.

fmrico:
Hi @wjwwood  and @sloretz ,

I have been iterating on this. I provide another alternative in PR #578

Is uses Node Class, so there is not duplciated code.

Best

PR #579, opened 2018-11-02
Skeleton for Action Server and Client

sloretz:
This is the start of C++ code matching the API in ros2/examples#216

### DIFF
--- a/rclcpp_action/CMakeLists.txt
+++ b/rclcpp_action/CMakeLists.txt
@@ -40,6 +40,10 @@ target_compile_definitions(${PROJECT_NAME}
   PRIVATE "RCLCPP_ACTION_BUILDING_LIBRARY")
 
 install(
+  DIRECTORY include/
+  DESTINATION include)
+
+install(
   TARGETS ${PROJECT_NAME}
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
--- a/rclcpp_action/CMakeLists.txt
+++ b/rclcpp_action/CMakeLists.txt
@@ -63,23 +63,21 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_client test/test_client.cpp)
   if(TARGET test_client)
-    target_include_directories(test_client PUBLIC
-      ${test_msgs_INCLUDE_DIRS}
+    ament_target_dependencies(test_client
+      "test_msgs"
     )
     target_link_libraries(test_client
       ${PROJECT_NAME}
-      ${test_msgs_LIBRARIES}
     )
   endif()
 
   ament_add_gtest(test_server test/test_server.cpp)
   if(TARGET test_server)
-    target_include_directories(test_server PUBLIC
-      ${test_msgs_INCLUDE_DIRS}
+    ament_target_dependencies(test_server
+      "test_msgs"
     )
     target_link_libraries(test_server
       ${PROJECT_NAME}
-      ${test_msgs_LIBRARIES}
     )
   endif()
 
--- a/rclcpp_action/CMakeLists.txt
+++ b/rclcpp_action/CMakeLists.txt
@@ -57,6 +57,7 @@ ament_export_libraries(${PROJECT_NAME})
 ament_export_dependencies(ament_cmake)
 ament_export_dependencies(action_msgs)
 ament_export_dependencies(rclcpp)
+ament_export_dependencies(rcl_action)
 ament_export_dependencies(rosidl_generator_c)
 ament_export_dependencies(rosidl_generator_cpp)
 
--- a/rclcpp_action/CMakeLists.txt
+++ b/rclcpp_action/CMakeLists.txt
@@ -5,6 +5,7 @@ project(rclcpp_action)
 find_package(ament_cmake_ros REQUIRED)
 find_package(action_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rcl_action REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
 # Default to C++14
@@ -28,6 +29,7 @@ add_library(${PROJECT_NAME}
 
 ament_target_dependencies(${PROJECT_NAME}
   "action_msgs"
+  "rcl_action"
   "rclcpp"
   "rosidl_generator_c"
   "rosidl_generator_cpp")
--- a/rclcpp_action/CMakeLists.txt
+++ b/rclcpp_action/CMakeLists.txt
@@ -59,7 +59,6 @@ ament_export_dependencies(rosidl_generator_cpp)
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
-  find_package(test_msgs REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
   ament_add_gtest(test_client test/test_client.cpp)
--- a/rclcpp_action/CMakeLists.txt
+++ b/rclcpp_action/CMakeLists.txt
@@ -1,0 +1,85 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(rclcpp_action)
+
+find_package(ament_cmake_ros REQUIRED)
+find_package(action_msgs REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rosidl_generator_cpp REQUIRED)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+include_directories(include)
+
+set(${PROJECT_NAME}_SRCS
+  src/client.cpp
+  src/server.cpp
+  src/server_goal_handle.cpp
+)
+
+add_library(${PROJECT_NAME}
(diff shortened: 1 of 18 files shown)


sloretz:
CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5562/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2208/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4583/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5451/)

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
```diff
@@ -0,0 +1,68 @@
...
+  friend Server<ACTION>;
+
+  RCLCPP_ACTION_PUBLIC
+  ServerGoalHandle(
+    rcl_action_server_t * rcl_server, const typename ACTION::Goal goal,
+    rcl_action_goal_handle_t * rcl_handle);
+
+  // TODO(sloretz) shared pointer to keep rcl_server_ alive while goal handles are alive
+  rcl_action_server_t * rcl_server_;
+  rcl_action_goal_handle_t * rcl_handle_;
```
hidmic: @sloretz I agree we should use smart pointers, for both (I'd initially say shared and unique respectively).

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
```diff
@@ -0,0 +1,64 @@
...
+template<typename ACTION>
+class ClientGoalHandle
+{
+public:
+  RCLCPP_ACTION_PUBLIC
+  virtual ~ClientGoalHandle();
+
+  /// TODO(sloretz) examples have this on client as `async_cancel_goal(goal_handle)`
+  std::future<bool>
+  async_cancel();
```
hidmic: @sloretz I do like the idea of calling cancel on the goal handle, but we have to support cancellation requests that expand multiple goals (e.g. cancel all goals). So I'd propose to keep this `async_cancel()` method here, plus `async_cancel_all_goals()` and `async_cancel_goals_before()` methods in the client, and maybe expose the low level method that just takes goal ID and timestamp for completeness.

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+  // TODO(sloretz) `async_result` to make method names consistent?
+  std::future<typename ACTION::Result>
+  result_future();
```
hidmic: @sloretz hmm, I do prefer `async_result`, but it's not a strong preference.

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+private:
+  // The templated Server creates goal handles
+  friend Client<ACTION>;
```
hidmic: @sloretz I don't know if this is a typical pattern in ROS2, but why not using an abstract base class and have the client implement their own?

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+  RCLCPP_ACTION_PUBLIC
+  ClientGoalHandle(rcl_action_client_t * rcl_client, const rcl_action_goal_info_t rcl_info);
+
+  // TODO(sloretz) shared pointer to keep rcl_client_ alive while goal handles are alive
+  rcl_action_client_t * rcl_client_;
+  rcl_action_goal_info_t rcl_info_;
```
hidmic: @sloretz I agree we should use smart pointers.

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
```diff
@@ -0,0 +1,64 @@
...
+};
+}  // namespace rclcpp_action
+
+#include <rclcpp_action/client_goal_handle_impl.hpp>  // NOLINT(build/include_order)
+#endif  // RCLCPP_ACTION__CLIENT_GOAL_HANDLE_HPP_
+
+
+// TODO(sloretz) goal_handle->result_future()
+// TODO(sloretz) goal_handle->cancel()
```
hidmic: @sloretz I think we can remove these TODOs.

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+  // TODO(sloretz) examples has this attribute as 'goal'
+  /// The original request message describing the goal.
+  const typename ACTION::Goal goal_;
+
+  // TODO(sloretz) `set_cancelled`, `set_succeeded`, `set_aborted`
+
+private:
+  // The templated Server creates goal handles
+  friend Server<ACTION>;
```
hidmic: @sloretz same as above about using `friend`.

rclcpp_action/CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+  endif()
+
+  ament_add_gtest(test_server test/test_server.cpp)
+  if(TARGET test_server)
+    target_include_directories(test_server PUBLIC
+      ${test_msgs_INCLUDE_DIRS}
+    )
+    target_link_libraries(test_server
+      ${PROJECT_NAME}
+      ${test_msgs_LIBRARIES}
```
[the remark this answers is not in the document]
sloretz: `ament_target_dependencies()` is the recommended one. I keep forgetting the `ament_cmake` macros and fallback to the usual CMake functions.

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
sloretz: Sounds good. I just need to remember to update this line in the examples https://github.com/ros2/examples/pull/216/files#diff-9df9aa44ab873a3a211a764112190449R37

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
sloretz: I don't using `friend` and a private constructor is a typical pattern. The base class approach would work too.

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
sloretz: 9e514fb7616ca3bd7821393c11ec0ec882258c28

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
sloretz: 4fc2246f448f7ce442c46e64ac9531e7e8d9536e

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
sloretz: e79f5da8032e02e89d0a1b66ba379745f46a21fc

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
sloretz: @hidmic is 193eb4a8bd8010f7ed5f5429e8c4d1a375fc759d what you had in mind?

hidmic: Hmm, I was actually thinking of something like (in client_goal_handle.hpp):

```c++
class ClientGoalHandle {
  public:
    virtual std::future<bool> async_cancel() = 0;
};
```

and then (in `client.cpp`):
```c++
class SimpleClientGoalHandle : public ClientGoalHandle {
   public:
      std::future<bool> async_cancel() override {
         ...
      }
};
```
External code never instantiates a `*GoalHandle` because it knows nothing of it. We have to watch out for slicing though.

sloretz: I don't think `ClientGoalHandleImpl` can be in `client.cpp` because `std::future<typename ACTION::Result> async_result()` requires knowing the user's type.

sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5581/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2216/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4591/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5461/)

rclcpp_action/include/rclcpp_action/client_goal_handle.hpp
hidmic: Argh, right, there's no common base type for action user defined types, silly me. 

I think it's fine. Another alternative (for the sake of completeness :) ) would be to define the goal handle subclass as a private nested class of the client/server class. Same code, no `friend`ship.

hidmic:
I agree with Jacob.

rclcpp_action/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+
+  ament_add_gtest(test_server test/test_server.cpp)
+  if(TARGET test_server)
+    ament_target_dependencies(test_server
+      "test_msgs"
+    )
+    target_link_libraries(test_server
+      ${PROJECT_NAME}
+    )
+  endif()
```
dirk-thomas: Both of these tests only run with the default RMW impl. It would be good to update this to run the tests for each RMW impl.

PR #582, opened 2018-11-13
Fixing the unit test for the logging interface change in rcl.

nburek:
## Summary
This change is part of a set of changes for new logging features. These changes include the following features:

### rcutils
- Slight refactor to the buffer used in logging to make it more generic and reusable
- A new output handler that allows multiple output handlers to be set (default is still only the console output function). This includes the following three output handlers: 1) The console output handler 2) The external logging lib output handler 3) The rosout output handler
- Change to the output format to accept only the log string instead of the format and vargs. The log will be formatted before being sent to the output so that there will be a consistent format across all the outputs
- Added an interface for an external logging library, an output handler function for sending logs to this library, and cmake changes so that the external logging lib can be specified (default is rc_logging_log4cxx)

### rcl
- Added new command line parameters for disabling various logging output handlers
- Added new command line parameter for passing in a config file to the external logging library
- Called the new configuration function in the rcutils logging to configure the logging during the ROS init process

### rclcpp
- Fixing a broken unit test

### rc_logging_log4cxx
- An external logging implementation that uses log4cxx under the hood
- Sets up a default FileAppender  logger to forward logs to files
- Uses the process Id as the filename by default

## Remaining Feature Work
- The rosout handlers is stubbed out in rcutils right now, but more work will need to be done to actually support sending logs to a rosout topic on a node. This will likely include moving the rosout handler up to rcl so that it can use the node interfaces to publish. 

https://discourse.ros.org/t/ros2-logging/6469
https://github.com/nburek/rc_logging_log4cxx


tfoote:
@nburek fyi if you put an entry like:
`Connects to ros2/rcl#327` in the description it will get grouped in https://waffle.io/ros2/ros2 where we triage and review tickets. Just pick a main issue/ticket and connect the others to it.

nburek:
Rebased on master.

nburek:
This entire PR will probably be discarded after the change. I'll run tests to verify though and then close it out if it's no longer needed.

wjwwood:
> This entire PR will probably be discarded after the change. I'll run tests to verify though and then close it out if it's no longer needed.

That's ok, I just wanted to make sure it was ready if needed.

PR #587, opened 2018-11-19
refactor init to allow for non-global init

wjwwood:
Connects to ros2/rmw#154


rclcpp/src/rclcpp/rcl_context_wrapper.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+struct RclContextWrapper
+{
+  RclContextWrapper() : context_(nullptr) {this->reset();}
+
+  ~RclContextWrapper()
+  {
+    rcl_ret_t ret = rcl_shutdown(context_.get());
+    if (RCL_RET_OK != ret && RCL_RET_ALREADY_SHUTDOWN != ret) {
+      rclcpp::exceptions::throw_from_rcl_error(ret, "failed to shutdown context");
```
jacobperron: Consider not throwing in the destructor; as I understand, it is generally considered not a good idea.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -162,14 +169,36 @@ signal_handler(int signal_value)
...
+  int argc,
+  char const * const argv[],
+  rclcpp::Context::SharedPtr context_out,
+  const rclcpp::InitOptions & init_options)
+{
+  auto rcl_context_wrapper = context_out->get_sub_context<RclContextWrapper>();
+  rcl_ret_t ret = rcl_init(
+    argc, argv, init_options.get_rcl_init_options(), rcl_context_wrapper->get_context().get());
+  if (RCL_RET_OK != ret) {
+    rclcpp::exceptions::throw_from_rcl_error(ret, "failed to initialize rcl: ");
```
jacobperron: nit: other calls to this function don't append `: `
```suggestion
    rclcpp::exceptions::throw_from_rcl_error(ret, "failed to initialize rcl");
```

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -252,63 +297,98 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
...
   return return_arguments;
 }
 
 bool
-rclcpp::ok()
+rclcpp::ok(rclcpp::Context::SharedPtr context)
 {
-  return ::g_signal_status == 0;
+  using rclcpp::contexts::default_context::get_global_default_context;
+  if (nullptr != context) {
```
jacobperron: I think this is a bug?
```suggestion
  if (nullptr == context) {
```

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -252,63 +297,98 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
...
+  return rcl_context_is_valid(rcl_context_wrapper->get_context().get());
 }
 
 bool
-rclcpp::is_initialized()
+rclcpp::is_initialized(rclcpp::Context::SharedPtr context)
 {
-  return rcl_ok();
+  using rclcpp::contexts::default_context::get_global_default_context;
+  if (nullptr != context) {
```
jacobperron: ditto
```suggestion
  if (nullptr == context) {
```

jacobperron: Maybe just replace this implementation with `return ok(context);` ?

Out of curiosity, when do you think is an appropriate time to remove deprecated functions, such as this one?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -252,63 +297,98 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
...
+on_shutdown_callbacks_;
 
 void
-rclcpp::shutdown()
+rclcpp::shutdown(rclcpp::Context::SharedPtr context)
 {
   trigger_interrupt_guard_condition(SIGINT);
 
+  using rclcpp::contexts::default_context::get_global_default_context;
+  if (nullptr != context) {
```
jacobperron:
```suggestion
  if (nullptr == context) {
```

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -252,63 +297,98 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
...
 {
   std::lock_guard<std::mutex> lock(g_sigint_guard_cond_handles_mutex);
   auto kv = g_sigint_guard_cond_handles.find(wait_set);
   if (kv != g_sigint_guard_cond_handles.end()) {
     return &kv->second;
   } else {
-    rcl_guard_condition_t handle =
-      rcl_get_zero_initialized_guard_condition();
+    using rclcpp::contexts::default_context::get_global_default_context;
+    if (nullptr != context) {
```
jacobperron:
```suggestion
    if (nullptr == context) {
```

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -252,63 +297,98 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
...
+  }
+  auto rcl_context_wrapper = context->get_sub_context<RclContextWrapper>();
+  auto ret = rcl_shutdown(rcl_context_wrapper->get_context().get());
+  if (RCL_RET_OK != ret) {
+    rclcpp::exceptions::throw_from_rcl_error(ret);
+  }
+  rcl_context_wrapper->reset();
+
+  // Uninstall the signal handlers.
+  uninstall_signal_handlers();
```
jacobperron: In the case of multiple contexts, what are the consequences of uninstalling the global signal handlers when the first context is shutdown? Should we consider waiting until the last context is shutdown before uninstalling signal handlers?

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -49,27 +51,100 @@ namespace rclcpp
...
+RCLCPP_PUBLIC
+void
+init_local(
+  int argc,
+  char const * const argv[],
+  rclcpp::Context::SharedPtr context_out,
+  const InitOptions & init_options = InitOptions());
+
+/// Convenience version of init_local with a templated Context type.
+template<typename ContextT = rclcpp::Context, typename... Args>
```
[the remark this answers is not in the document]
wjwwood: So that you can have a derived type there (one that inherits from `rclcpp::Context`) that has different constructor arguments.

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -82,27 +157,66 @@ std::vector<std::string>
...
+ok(rclcpp::Context::SharedPtr context = nullptr);
 
-/// Returns true if init() has already been called.
-/** \return True if init() has been called, false otherwise. */
+/// Return true if init() has already been called for the given context.
+/**
+ * If nullptr is given for the context, then the global context is used, i.e.
+ * the context initialized by rclcpp::init().
+ *
+ * Deprecated, as it is no longer different from rcl_ok().
```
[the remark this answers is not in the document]
wjwwood: Originally I deprecated this (with `[[deprecated]]`), but thought better of it after. I just need to remove the line in the docs.

rclcpp/src/rclcpp/rcl_context_wrapper.hpp
```diff
@@ -0,0 +1,68 @@
...
+  void
+  reset()
+  {
+    context_.reset(new rcl_context_t, [](rcl_context_t * context) {
+      if (nullptr == context) {
+        return;
+      }
+      rcl_ret_t ret = rcl_context_fini(context);
+      if (RCL_RET_OK != ret && RCL_RET_INVALID_ARGUMENT != ret) {
+        rclcpp::exceptions::throw_from_rcl_error(ret, "failed to fini context");
```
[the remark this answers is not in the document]
wjwwood: Good point, I'll change these to logged error messages.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: Definitely, I had that fix locally, sorry I didn't push sooner, I was still working out other issues.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: So, there's been a long debate about `rclcpp::ok()`'s name, and I didn't really want to revisit it. `is_initialized()` was added to differentiate between ok (whether sigint was called or not) and whether or not init had been called. Now those are essentially the same thing, so there isn't a good reason for one or the other. Personally I'd drop `is_initialized()` since no one is probably using it very much yet, as opposed to `ok()` which is used in several examples and in rclcpp itself. This despite the fact that rcl calls it "is_initialized" and that we've discussed changing the `ok()` name before (though I'm fine with it).

rclcpp/src/rclcpp/utilities.cpp
wjwwood: Personally, I'm leaning toward no, simply because it's somewhat tricky to do that and because it's hard to describe what uninstalling the signal handler would exactly be doing at that point. In my opinion if you create multiple context and then call shutdown, you need to handle the signals at that point or avoid init/shutdown and use install/uninstall signal handlers yourself.

jacobperron: The first option seems reasonable to me: only uninstall if the default context is used.
Another option is to be explicit about the install/uninstall of the signal handlers by adding an optional argument to the `init()` and `shutdown()` functions, perhaps with defaults set to `true`.

rclcpp/src/rclcpp/utilities.cpp
sloretz: ping @wjwwood

wjwwood: fixed

rclcpp/src/rclcpp/context.cpp
```diff
@@ -14,6 +14,202 @@
...
+  // notify all blocking calls, if asked
+  if (notify_all) {
+    rclcpp::notify_all();
+  }
+  // remove self from the global contexts
+  std::lock_guard<std::mutex> context_lock(g_contexts_mutex);
+  for (auto it = g_contexts.begin(); it != g_contexts.end(); ) {
+    auto shared_context = it->lock();
+    if (shared_context.get() == this) {
+      it = g_contexts.erase(it);
```
sloretz: Is the context expected to be in `g_contexts` more than once? I would have expected a `break` here.

PR #593, opened 2018-11-27
rclcpp_action Server implementation

sloretz:
The example server in PR ros2/examples#216 has been updated to work with this PR https://github.com/ros2/examples/blob/75e3f7d4c1d4f26853fe5c894d3940e7cd7af1b1/rclcpp/minimal_action_server/not_composable.cpp

This PR requires ros2/rclcpp#594 in a convoluted way: checkout that branch and merge this one into it.

requires ros2/rcl#343


sloretz:
Implementation is almost complete, running CI to check other platforms.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5684/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2289/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4698/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5592/)

sloretz:
Checking clang warnings on osx [Build Status](https://ci.ros2.org/job/ci_osx/4705/)

sloretz:
CI again

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5696/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2299/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4707/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5610/)

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,477 @@
...
+
+    std::shared_ptr<rcl_action_goal_handle_t> handle(new rcl_action_goal_handle_t, deleter);
+    // Copy out goal handle since action server storage disappears when it is fini'd
+    *handle = *rcl_handle;
+
+    // publish status since a goal's state has changed (was accepted)
+    publish_status();
+
+    // Change status to executing
+    ret = rcl_action_update_goal_state(handle.get(), GOAL_EVENT_EXECUTE);
```
sloretz: One problem with this approach is an action server that wants to execute goals sequentially while keeping a queue of goals will report all of them as executing in the status message.

One way to fix this would be to add a `GoalResponse` called `ACCEPT_AND_DEFER` that skipped the automatic transition to executing. The user would become responsible for calling a method on the goal handle to begin execution. This would need another callback between the goal handle and `ServerBase` that could trigger publishing another status message.

hidmic: Why do we have to transition to `EXECUTING` right away? Can't we stay in `ACCEPTED`?

sloretz: You're write it can stay in accepted, just need to add the plumbing from the goal handle to allow the user to transition to executing on their own time.

sloretz: Actually, if the user accepts it they won't have a goal handle. The first time it's sent to the user is via the `handle_execute` callback. Need to rethink this one

sloretz: After talking offline I changed `handle_execute` to `handle_accepted` in 339a36e85039c6d95dbd8e553c29f311f86dd0ed

sloretz: Added all the stuff I missed and a test for deferred executing in e7fd566c1c38d9d0ac4710db3cda1632efca4526

rclcpp_action/include/rclcpp_action/create_server.hpp
```diff
@@ -31,13 +34,37 @@ create_server(
   rclcpp::Node * node,
```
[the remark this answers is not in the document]
sloretz: Dropped in in 73e3c26df087d0037a977d2af8ba4f0e82d7e7b1. Turns out the smart pointer is necessary to safely call `remove_waitable()` in the action server deleter.

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
```diff
@@ -15,41 +15,250 @@
...
-  explicit ServerGoalHandle(const typename ACTION::Goal goal)
-  : goal_(goal) {}
+  /// \internal
+  ServerGoalHandle(
+    std::shared_ptr<rcl_action_server_t> rcl_server,
+    std::shared_ptr<rcl_action_goal_handle_t> rcl_handle,
+    std::array<uint8_t, 16> uuid,
+    std::shared_ptr<const typename ACTION::Goal> goal,
+    std::function<void(const std::array<uint8_t, 16>&, std::shared_ptr<void>)> on_terminal_state,
+    std::function<void(const std::array<uint8_t, 16>&)> on_executing
```
wjwwood: Can we have an alias for `std::array<uint8_t, 16>`? like `rclcpp_action::GoalID` or whatever?

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -31,65 +36,428 @@ namespace rclcpp_action
...
+  // -----------------------------------------------------
+  // API for communication between ServerBase and Server<>
+
+  // ServerBase will call this function when a goal request is received.
+  // The subclass should convert to the real type and call a user's callback.
+  /// \internal
+  RCLCPP_ACTION_PUBLIC
+  virtual
+  std::pair<GoalResponse, std::shared_ptr<void>>
+  call_handle_goal_callback(std::array<uint8_t, 16> &, std::shared_ptr<void> request) = 0;
```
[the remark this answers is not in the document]
sloretz: Definitely. I'm hoping to use the `GoalID` typedef from the client once that's merged

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
wjwwood: Sorry, just saw https://github.com/ros2/rclcpp/pull/593#discussion_r238888237

sloretz: Definitely. It's in the client PR. I'm hoping to update this one after the client is merged.

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -31,65 +36,428 @@ namespace rclcpp_action
...
+  /// \internal
+  std::pair<CancelResponse, std::shared_ptr<rcl_action_goal_handle_t>>
+  call_handle_cancel_callback(const std::array<uint8_t, 16> & uuid) override
+  {
+    CancelResponse resp = CancelResponse::REJECT;
+    std::shared_ptr<rcl_action_goal_handle_t> rcl_handle;
+    auto element = goal_handles_.find(uuid);
+    if (element != goal_handles_.end()) {
+      std::shared_ptr<ServerGoalHandle<ACTION>> goal_handle = element->second.lock();
+      if (goal_handle) {
```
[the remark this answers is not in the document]
sloretz: I don't think so. It would happen if the original `goal_handle` was destructed, so there's nothing left that care's about the cancel request.

Though maybe the `goal_handle` should try transition itself to `ABORTED` when it gets destructed without completing.

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -31,65 +36,428 @@ namespace rclcpp_action
...
+
+  /// \internal
+  void
+  call_goal_accepted_callback(
+    std::shared_ptr<rcl_action_server_t> rcl_server,
+    std::shared_ptr<rcl_action_goal_handle_t> rcl_goal_handle,
+    std::array<uint8_t, 16> uuid, std::shared_ptr<void> goal_request_message) override
+  {
+    std::shared_ptr<ServerGoalHandle<ACTION>> goal_handle;
+    // TODO(sloretz) how to make sure this lambda is not called beyond lifetime of this?
```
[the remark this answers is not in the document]
sloretz: That's not a bad idea. This TODO is trying to describe a different case where a server goal handle lives longer than the action server itself.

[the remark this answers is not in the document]
wjwwood: I'd wrap `this` in a weak pointer, try to lock it in the lambda and just return quietly when the `lock()` of the weak pointer fails.

I do this in the graph listener:

https://github.com/ros2/rclcpp/blob/9d7b50e4f7ad48846eccafd22a0d0ea4fe885172/rclcpp/src/rclcpp/graph_listener.cpp#L89-L95

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,474 @@
...
+      uuid[i] = goal_info.goal_id.uuid[i];
+    }
+    auto response_pair = call_handle_cancel_callback(uuid);
+    if (CancelResponse::ACCEPT == response_pair.first) {
+      ret = rcl_action_update_goal_state(response_pair.second.get(), GOAL_EVENT_CANCEL);
+      if (RCL_RET_OK != ret) {
+        rcl_ret_t fail_ret = rcl_action_cancel_response_fini(&cancel_response);
+        (void)fail_ret;  // TODO(sloretz) do something with error during cleanup
+        rclcpp::exceptions::throw_from_rcl_error(ret);
+      } else {
```
jacobperron: Since we're throwing above, we can remove the `else` block (unnecessary indentation),

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,474 @@
...
+        uuid[i] = expired_goals[0].goal_id.uuid[i];
+      }
+      pimpl_->goal_results_.erase(uuid);
+      pimpl_->result_requests_.erase(uuid);
+    }
+  }
+}
+
+void
+ServerBase::publish_status()
```
jacobperron: Just a thought:

IIRC, there was discussion in the design doc about making the status (and feedback) publisher optional. I'm not sure if we had any ideas about how the implementation would look, but given time constraints it's probably good to capture this feature in a ticket to tackle at a later time.
It might make sense to instead have the option to disable subscribers in the action client so then CLI tools can always interact with action servers.

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -31,65 +36,428 @@ namespace rclcpp_action
...
 class Server : public ServerBase
 {
 public:
   RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(Server)
 
-  using GoalCallback = std::function<void (rcl_action_goal_info_t &, typename ACTION::Goal *)>;
-  using CancelCallback = std::function<void (std::shared_ptr<ServerGoalHandle<ACTION>>)>;
+  /// Signature of a callback that accepts or rejects goal requests.
+  using GoalCallback = std::function<GoalResponse(
+        std::array<uint8_t, 16>&, std::shared_ptr<typename ACTION::Goal>)>;
```
wjwwood: Why is this shared pointer not `const` here, but it is for `get_goal()` on the `ServerGoalHandle<ActionT>`? Should it be const here too, or is this a copy of the goal?

sloretz: I think it should be `const`

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,474 @@
...
+  if (RCL_RET_OK != ret) {
+    rclcpp::exceptions::throw_from_rcl_error(ret);
+  }
+}
+
+void
+ServerBase::publish_result(const std::array<uint8_t, 16> & uuid, std::shared_ptr<void> result_msg)
+{
+  // Check that the goal exists
+  rcl_action_goal_info_t goal_info;
```
jacobperron: Since the uuid -> `goal_info` conversion is happening in several places, it would be nice to wrap in a utility function. Perhaps after we have the GoalID wrapper introduced by #594

rclcpp_action/test/test_server.cpp
```diff
@@ -32,15 +33,609 @@ class TestServer : public ::testing::Test
...
 
 TEST_F(TestServer, construction_and_destruction)
 {
-  auto node = std::make_shared<rclcpp::Node>("my_node", "/rclcpp_action/test/server");
+  auto node = std::make_shared<rclcpp::Node>("construct_node", "/rclcpp_action/construct");
 
   using GoalHandle = rclcpp_action::ServerGoalHandle<test_msgs::action::Fibonacci>;
-  auto as = rclcpp_action::create_server<test_msgs::action::Fibonacci>(node.get(), "fibonacci",
-      [](rcl_action_goal_info_t &, test_msgs::action::Fibonacci::Goal *) {},
+  auto as = rclcpp_action::create_server<test_msgs::action::Fibonacci>(node, "fibonacci",
```
jacobperron: We could add tests for invalid constructions (e.g. invalid node or action name).

rclcpp_action/include/rclcpp_action/server.hpp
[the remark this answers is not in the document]
sloretz: Thanks. Implemented in d848580df70c037257cf0da6c942d584544a6cf8

rclcpp_action/include/rclcpp_action/server.hpp
sloretz: const in b1828727242a48d6206b1b97307c5bf49d626a89

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
```diff
@@ -15,41 +15,250 @@
...
 public:
-  virtual ~ServerGoalHandle();
-
   /// Indicate if client has requested this goal be cancelled.
   /// \return true if a cancelation request has been accepted for this goal.
-  virtual bool
-  is_cancel_request() const = 0;
+  RCLCPP_ACTION_PUBLIC
+  bool
+  is_cancel_request() const;
```
wjwwood: Just curious, why did you name it like this, rather than say something like `is_canceled()`?

sloretz: It returns `true` in state `CANCELLING` but `false` in state `CANCELED`. Returning `false` would be a strange case if the API was  called `is_canceled()`. 

Ideas? It's the API an action server author should check and call `set_canceled()` after they see this returns `true`.

Alternatively, it could be called `is_canceled()` and return `true` in both cases.

rclcpp_action/src/server.cpp
sloretz: Fixed, i think. I don't see an else condition anymore.

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
sloretz: changed to `is_canceling()` in 52ac6d0bb493e709ffbb678db62a4d8c2a501d3c

rclcpp_action/include/rclcpp_action/server.hpp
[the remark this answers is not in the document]
sloretz: Uses `GoalID` in b9e7fc8

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
sloretz: Done in 
b9e7fc8

rclcpp_action/src/server.cpp
sloretz: Added conversion function in f717e9b5fcda8d36b3a7fb5d3557682b6bac745e

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -15,81 +15,468 @@
...
+    return std::make_pair(user_response, ros_response);
+  }
+
+  /// \internal
+  CancelResponse
+  call_handle_cancel_callback(const GoalID & uuid) override
+  {
+    std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+    CancelResponse resp = CancelResponse::REJECT;
+    std::shared_ptr<rcl_action_goal_handle_t> rcl_handle;
```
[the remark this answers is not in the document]
sloretz: Removed in 03c9cf0fda35ccf0d74d34dfc546927948f61f58

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
```diff
@@ -15,41 +15,244 @@
...
+
+  /// Get the original request message describing the goal.
+  const std::shared_ptr<const typename ACTION::Goal>
+  get_goal() const
+  {
+    return goal_;
+  }
+
+  /// Get the unique identifier of the goal
+  const GoalID
```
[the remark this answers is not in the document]
sloretz: Constant reference returned in ea33c8810128ce955615d05eb1649c58993816b7

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,554 @@
...
+  const rosidl_action_type_support_t * type_support,
+  const rcl_action_server_options_t & options
+)
+: pimpl_(new ServerBaseImpl)
 {
-  // TODO(sloretz) use rcl_action API when available
-  (void)node_base;
-  (void)name;
-  (void)type_support;
+  rcl_ret_t ret;
```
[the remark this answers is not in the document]
sloretz: Moved below in 8a21dec3d19ec00c91b5b595da67c6980ec1d712

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,554 @@
...
+    &pimpl_->result_request_ready_,
+    &pimpl_->goal_expired_);
+
+  if (RCL_RET_OK != ret) {
+    rclcpp::exceptions::throw_from_rcl_error(ret);
+  }
+
+  return pimpl_->goal_request_ready_ ||
+         pimpl_->cancel_request_ready_ ||
+         pimpl_->result_request_ready_;
```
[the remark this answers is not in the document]
sloretz: Whoops, added in bf4797a7a6af486f9f425afa055bcc7aeaacaffe

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,554 @@
...
+
+  if (RCL_RET_OK != ret) {
+    rclcpp::exceptions::throw_from_rcl_error(ret);
+  }
+
+  pimpl_->goal_request_ready_ = false;
+  GoalID uuid = get_goal_id_from_goal_request(message.get());
+  for (size_t i = 0; i < 16; ++i) {
+    goal_info.goal_id.uuid[i] = uuid[i];
+  }
```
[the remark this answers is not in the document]
sloretz: I'm not familiar with `std::copy`. Is that like a default copy constructor? This is converting from a C-style array to a C++ std::array. It was moved into a function in f717e9b5fcda8d36b3a7fb5d3557682b6bac745e

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,554 @@
...
+      rclcpp::exceptions::throw_from_rcl_error(ret);
+    }
+  } else {
+    // Store the request so it can be responded to later
+    std::lock_guard<std::mutex> lock(pimpl_->results_mutex_);
+    auto iter = pimpl_->result_requests_.find(uuid);
+    if (iter == pimpl_->result_requests_.end()) {
+      pimpl_->result_requests_[uuid] = std::vector<rmw_request_id_t>{request_header};
+    } else {
+      iter->second.push_back(request_header);
```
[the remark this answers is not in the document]
sloretz: Ah yup, C++ `map`s are weird. ae676d84ec27615e17cbd2284b7edee711f5eb18

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,570 @@
...
+    }
+  }
+}
+
+void
+ServerBase::execute_check_expired_goals()
+{
+  // Allocate expecting only one goal to expire at a time
+  rcl_action_goal_info_t expired_goals[1];
```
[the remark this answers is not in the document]
sloretz: No good reason. The API it's passed to expects an array, so it felt like describing it as an array of size 1 matched nicer.

rclcpp_action/src/server.cpp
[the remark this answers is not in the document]
hidmic: Something like `std::copy(cpp_array.begin(), cpp_array.end(), c_array)` and `std::copy(std::begin(c_array), std::end(c_array), cpp_array.begin())` should work. Just an FYI though.

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,570 @@
...
+    msg.status = goal_state;
+    // Convert C goal info to C++ goal info
+    for (size_t i = 0; i < 16; ++i) {
+      msg.goal_info.goal_id.uuid[i] = goal_info.goal_id.uuid[i];
+    }
+    msg.goal_info.stamp.sec = goal_info.stamp.sec;
+    msg.goal_info.stamp.nanosec = goal_info.stamp.nanosec;
+
+    status_msg->status_list.push_back(msg);
+  }
```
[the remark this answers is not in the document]
sloretz: Didn't know about it. Used in 039854d3762e01f1d09e3ae8d6c7d6809e4c5a18

rclcpp_action/test/test_server.cpp
```diff
@@ -32,15 +34,609 @@ class TestServer : public ::testing::Test
...
 };
 
 TEST_F(TestServer, construction_and_destruction)
 {
-  auto node = std::make_shared<rclcpp::Node>("my_node", "/rclcpp_action/test/server");
+  auto node = std::make_shared<rclcpp::Node>("construct_node", "/rclcpp_action/construct");
+
+  using GoalHandle = rclcpp_action::ServerGoalHandle<Fibonacci>;
+  auto as = rclcpp_action::create_server<Fibonacci>(node, "fibonacci",
+      [](const std::array<uint8_t, 16> &, std::shared_ptr<const Fibonacci::Goal>) {
```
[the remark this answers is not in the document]
sloretz: GoalID in 71c7100707cfe93b6804d6446b860ac8609536c6

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,570 @@
...
+  {
+    std::lock_guard<std::mutex> lock(pimpl_->server_mutex_);
+    goal_exists = rcl_action_server_goal_exists(pimpl_->action_server_.get(), &goal_info);
+  }
+  if (!goal_exists) {
+    // Goal does not exists
+    result_response = create_result_response(action_msgs::msg::GoalStatus::STATUS_UNKNOWN);
+  } else {
+    // Goal exists, check if a result is already available
+    std::lock_guard<std::mutex> lock(pimpl_->results_mutex_);
```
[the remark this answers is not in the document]
sloretz: Use a single reentrant mutex for everything in b7a3de1ace4fbf8754c30cf3064b493f4ad4cdaa

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,570 @@
...
+  bool goal_exists;
+  {
+    std::lock_guard<std::mutex> lock(pimpl_->server_mutex_);
+    goal_exists = rcl_action_server_goal_exists(pimpl_->action_server_.get(), &goal_info);
+  }
+  if (!goal_exists) {
+    throw std::runtime_error("Asked to publish result for goal that does not exist");
+  }
+
+  std::lock_guard<std::mutex> lock(pimpl_->results_mutex_);
```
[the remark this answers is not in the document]
sloretz: b7a3de1ace4fbf8754c30cf3064b493f4ad4cdaa

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -15,81 +15,472 @@
...
-  // TODO(sloretz) accept clock instance
+  /// Construct an action server.
+  /**
+   * This constructs an action server, but it will not work until it has been added to a node.
+   * Use `rclcpp_action::create_server()` to both construct and add to a node.
+   *
+   * Three callbacks must be provided:
+   *  - one to accept or reject goals sent to the server,
+   *  - one to accept or reject requests to cancel a goal,
+   *  - one to given a goal handle after a goal has been accepted.
```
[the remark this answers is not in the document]
sloretz: Reworded in de1510b

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -15,81 +15,472 @@
...
+  }
+
+  /// \internal
+  void
+  call_goal_accepted_callback(
+    std::shared_ptr<rcl_action_goal_handle_t> rcl_goal_handle,
+    GoalID uuid, std::shared_ptr<void> goal_request_message) override
+  {
+    std::shared_ptr<ServerGoalHandle<ActionT>> goal_handle;
+    std::weak_ptr<Server<ActionT>> weak_this = this->shared_from_this();
```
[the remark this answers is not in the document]
sloretz: Yeah, this is making sure the server or server goal handle can be safely destructed in either order.

rclcpp_action/src/server.cpp
```diff
@@ -12,30 +12,554 @@
...
+    }
+  }
+
+  if (!response->goals_canceling.empty()) {
+    // at least one goal state changed, publish a new status message
+    publish_status();
+  }
+
+  // TODO(sloretz) make this fini happen in an exception safe way
+  ret = rcl_action_cancel_response_fini(&cancel_response);
```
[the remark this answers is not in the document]
sloretz: used `RCLCPP_SCOPE_EXIT` in 1c8c86d

sloretz:
All comments so far addressed, checking CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5798/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2368/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4794/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5697/)

sloretz:
Windows attempt 2 [Build Status](https://ci.ros2.org/job/ci_windows/5699/)

sloretz:
Remaining CI jobs

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/5801/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/2370/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/4796/)

rclcpp_action/src/server_goal_handle.cpp
```diff
@@ -12,4 +12,140 @@
...
+    rclcpp::exceptions::throw_from_rcl_error(ret);
+  }
+}
+
+bool
+ServerGoalHandleBase::try_canceling() noexcept
+{
+  std::lock_guard<std::mutex> lock(rcl_handle_mutex_);
+  // Check if the goal reached a terminal state already
+  const bool active = rcl_action_goal_handle_is_active(rcl_handle_.get());
```
jacobperron: We can use a short-cut and check if the goal is cancelable:

```c
const bool is_cancelable = rcl_action_goal_handle_is_cancelable(rcl_handle.get());
```

sloretz: What's the short cut? IIUC it looks like the code below would be the same.

jacobperron: I think this is equivalent to the current implementation:

```c
const bool is_cancelable = rcl_action_goal_handle_is_cancelable(rcl_handle_.get());
if (is_cancelable) {
  ret = rcl_action_update_goal_state(rcl_handle_.get(), GOAL_EVENT_CANCEL);
  if (RCL_RET_OK != ret) {
    return false;
  }
}

// Get the state again
ret = rcl_action_goal_handle_get_status(rcl_handle_.get(), &state);
if (RCL_RET_OK != ret) {
  return false;
}

// If it's canceling, cancel it
if (GOAL_STATE_CANCELING == state) {
  ret = rcl_action_update_goal_state(rcl_handle_.get(), GOAL_EVENT_SET_CANCELED);
  return RCL_RET_OK == ret;
}
return false;
```

PR #594, opened 2018-11-28
[rclcpp_action] Action client implementation

hidmic:
This pull requests adds the first `rclcpp` implementation of an action client.

PR #598, opened 2018-12-04
add wait_for_action_server() for action clients

wjwwood:
This is based on a glob of changes needed for me to write the action tests in `test_communication`. It will have to be rebased when https://github.com/ros2/rclcpp/pull/594 and https://github.com/ros2/rclcpp/pull/593 are merged.

c7c1738 is the only commit that is "new" to this pr.

Connects to ros2/system_tests#316

### DIFF
--- a/rclcpp_action/src/client.cpp
+++ b/rclcpp_action/src/client.cpp
@@ -34,11 +34,13 @@ class ClientBaseImpl
 public:
   ClientBaseImpl(
     rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base,
+    rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph,
     rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr node_logging,
     const std::string & action_name,
     const rosidl_action_type_support_t * type_support,
     const rcl_action_client_options_t & client_options)
-  : node_handle(node_base->get_shared_rcl_node_handle()),
+  : node_graph_(node_graph),
+    node_handle(node_base->get_shared_rcl_node_handle()),
     logger(node_logging->get_logger().get_child("rclcpp_acton")),
     random_bytes_generator(std::random_device{} ())
   {
@@ -96,6 +98,8 @@ public:
   bool is_cancel_response_ready{false};
   bool is_result_response_ready{false};
 
+  rclcpp::Context::SharedPtr context_;
+  rclcpp::node_interfaces::NodeGraphInterface::WeakPtr node_graph_;
   std::shared_ptr<rcl_action_client_t> client_handle{nullptr};
   std::shared_ptr<rcl_node_t> node_handle{nullptr};
   rclcpp::Logger logger;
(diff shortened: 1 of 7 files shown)


rclcpp_action/src/client.cpp
```diff
@@ -117,18 +121,93 @@ class ClientBaseImpl
...
+  auto node_ptr = pimpl_->node_graph_.lock();
+  if (!node_ptr) {
+    throw rclcpp::exceptions::InvalidNodeError();
+  }
+  auto event = node_ptr->get_graph_event();
+  // check to see if the server is ready immediately
+  if (this->action_server_is_ready()) {
+    return true;
+  }
+  if (timeout == std::chrono::nanoseconds(0)) {
```
sloretz: What should a negative timeout do? It could be nonblocking by making this check `<=`

rclcpp_action/src/client.cpp
```diff
@@ -117,18 +121,93 @@ class ClientBaseImpl
...
+    // check was non-blocking, return immediately
+    return false;
+  }
+  // update the time even on the first loop to account for time spent in the first call
+  // to this->server_is_ready()
+  std::chrono::nanoseconds time_to_wait = timeout - (std::chrono::steady_clock::now() - start);
+  if (timeout > std::chrono::nanoseconds(0) && time_to_wait < std::chrono::nanoseconds(0)) {
```
sloretz: `timeout > std::chrono::nanoseconds(0)` part might be unnecessary if negative timeout is handled above

rclcpp_action/src/client.cpp
hidmic: Negative timeout seems to be handled as indefinitely blocking below.

rclcpp_action/src/client.cpp
```diff
@@ -117,18 +121,93 @@ class ClientBaseImpl
...
+    node_ptr->wait_for_graph_change(
+      event, std::min(time_to_wait, std::chrono::nanoseconds(RCL_MS_TO_NS(100))));
+    // Because of the aforementioned race condition, we check if the service is ready even if the
+    // graph event wasn't triggered.
+    event->check_and_clear();
+    if (this->action_server_is_ready()) {
+      return true;
+    }
+    // server is not ready, loop if there is time left
+    time_to_wait = timeout - (std::chrono::steady_clock::now() - start);
```
hidmic: @wjwwood the only objection I have is that this line is assuming that `timeout` won't ever become largely negative. You may end up trying to wait for a negative time [here](https://github.com/ros2/rclcpp/pull/598/commits/df248de081e18ebcf4bcb5c5ca97f4a4d7ce04f5#diff-4ec96e599f7fb55d453fc9cbe6f4da4aR198) if someone were to pass a large negative timeout.

rclcpp_action/src/client.cpp
wjwwood: How would you handle negative timeout above, i.e. what would you change timeout to be?

rclcpp_action/src/client.cpp
wjwwood: Why would that be a problem? If timeout is a large negative (or a small one) then the while loop will exit just afte this and `false` will be returned.

(your link doesn't work for me btw)

wjwwood:
W.r.t. to the waiting code, it was copied mostly verbatim from the existing wait for service code:

https://github.com/ros2/rclcpp/blob/fe09d937b73fe84af570e5ae40d9e8e3bc6d3fd1/rclcpp/src/rclcpp/client.cpp#L112-L162

So any comments about it will also apply to that code.

rclcpp_action/src/client.cpp
hidmic: Why would it exit? If `timeout < std::chrono::nanoseconds(0)` holds, then it'll loop. Yes, I saw that code too. Same observation.

(Fixed the link, not sure what happened)

wjwwood: Sounds good to me, but we should also change the service one at the same time, since it actually has a lot of code utilizing it (via tests and stuff).

rclcpp_action/src/client.cpp
sloretz: I misunderstood that negative means block forever. @hidmic's changes LGTM

PR #605, opened 2018-12-08
defer signal handling to a singleton thread

### LINKED ISSUE #604: unsafe code in signal handlers
We discovered this due to a new flakey test: https://github.com/ros2/build_cop/issues/152

Basically we're locking mutex's in the signal handler in many ways (one of which is by triggering guard conditions in `rmw_fastrtps_cpp`'s code), which can lead to deadlocks if those functions are reentered  during signal handling.

After a lot of tracing and digging, @hidmic and I are pretty certain this is the core issue and that it could be an issue in a lot of other places, but the above flakey tests were just unlucky enough to setup an environment in which it happens fairly regularly.

We need to fix this for Crystal, so it's a high priority, but we have a plan to fix it.

Basically we're going to take all of the code which may lock a mutex (and therefore may deadlock) in the signal handler and move it into a global class which will call that code from a dedicated thread. The thread will be started when the signal handler is installed and joined when it is uninstalled or when the global class is destructed. In the thread it will simply wait to be safely notified from the signal handler and when notified it will do the necessary work like triggering guard conditions and running `shutdown()` (and therefore `on_shutdown()` callbacks) for each context.

@hidmic is going to get started on that, and I'm going to start working on refactoring of the notification system for `rclcpp::sleep_for` to make it context specific, removing yet more global state of which the signal handler class no longer needs to keep track.

wjwwood:
Fixes #604

### DIFF
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -17,7 +17,6 @@
 
 #include <condition_variable>
 #include <functional>
-#include <map>
 #include <memory>
 #include <mutex>
 #include <string>
@@ -349,7 +348,7 @@ private:
   /// Mutex to protect sigint_guard_cond_handles_.
   std::mutex interrupt_guard_cond_handles_mutex_;
   /// Guard conditions for interrupting of associated wait sets on interrupt_all_wait_sets().
-  std::map<rcl_wait_set_t *, rcl_guard_condition_t> interrupt_guard_cond_handles_;
+  std::unordered_map<rcl_wait_set_t *, rcl_guard_condition_t> interrupt_guard_cond_handles_;
 };
 
 /// Return a copy of the list of context shared pointers.
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -150,8 +150,8 @@ public:
    * - rcl_shutdown() is called on the internal rcl_context_t instance
    * - the shutdown reason is set
    * - each on_shutdown callback is called, in the order that they were added
-   * - if notify all blocking sleep_for() calls, so they return early due to shutdown
-   * - if notify all blocking executors and wait sets
+   * - if interrupt blocking sleep_for() calls, so they return early due to shutdown
+   * - if interrupt blocking executors and wait sets
    *
    * The underlying rcl context is not finalized by this function.
    *
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -284,6 +284,11 @@ public:
   void
   release_interrupt_guard_condition(rcl_wait_set_t * wait_set);
 
+  /// Nothrow version of release_interrupt_guard_condition(), logs to RCLCPP_ERROR instead.
+  RCLCPP_PUBLIC
+  void
+  release_interrupt_guard_condition(rcl_wait_set_t * wait_set, const std::nothrow_t &) noexcept;
+
   /// Interrupt any blocking executors, or wait sets associated with this context.
   RCLCPP_PUBLIC
   virtual
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -16,6 +16,7 @@
 #define RCLCPP__CONTEXT_HPP_
 
 #include <functional>
+#include <map>
 #include <memory>
 #include <mutex>
 #include <string>
@@ -26,6 +27,8 @@
 #include <vector>
 
 #include "rcl/context.h"
+#include "rcl/guard_condition.h"
+#include "rcl/wait.h"
 #include "rclcpp/init_options.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/visibility_control.hpp"
@@ -147,21 +150,21 @@ public:
    * - rcl_shutdown() is called on the internal rcl_context_t instance
    * - the shutdown reason is set
    * - each on_shutdown callback is called, in the order that they were added
-   * - if notify_all is true, rclcpp::notify_all is called to unblock some ROS functions
+   * - if notify all blocking sleep_for() calls, so they return early due to shutdown
+   * - if notify all blocking executors and wait sets
    *
    * The underlying rcl context is not finalized by this function.
    *
    * This function is thread-safe.
    *
    * \param[in] reason the description of why shutdown happened
-   * \param[in] notify_all if true, then rclcpp::notify_all will be called
    * \return true if shutdown was successful, false if context was already shutdown
    * \throw various exceptions derived from RCLErrorBase, if rcl_shutdown fails
    */
   RCLCPP_PUBLIC
   virtual
   bool
-  shutdown(const std::string & reason, bool notify_all = true);
+  shutdown(const std::string & reason);
 
   using OnShutdownCallback = std::function<void ()>;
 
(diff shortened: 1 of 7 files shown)


nuclearsandwich:

nuclearsandwich:
The `Cpr` CI failures look like they might actually useful. The errors relate to contents of the PR and dependency packages are all available.

nuclearsandwich:
Optimistically started CI builds

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5843/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2407/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4836/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5739/)

mjcarroll:
Looks like MacOS may have failed due to network issues, then failed again on `test_get_node_names`: https://ci.ros2.org/job/ci_osx/4837/testReport/

nuclearsandwich:
`curl`'d the build log locally and `grep`'d through it and it does appear that every test which ran passed so I'm suggesting we treat that mac build as passing.

dirk-thomas:
> it does appear that every test which ran passed

The summary line explicitly says: 24522 tests, 0 errors, 0 failures, 50 skipped :+1:

rclcpp/include/rclcpp/context.hpp
```diff
@@ -261,6 +340,16 @@ class Context : public std::enable_shared_from_this<Context>
...
+
+  /// Condition variable for timed sleep (see sleep_for).
+  std::condition_variable interrupt_condition_variable_;
+  /// Mutex for protecting the global condition variable.
+  std::mutex interrupt_mutex_;
+
+  /// Mutex to protect sigint_guard_cond_handles_.
+  std::mutex interrupt_guard_cond_handles_mutex_;
+  /// Guard conditions for interrupting of associated wait sets on interrupt_all_wait_sets().
+  std::map<rcl_wait_set_t *, rcl_guard_condition_t> interrupt_guard_cond_handles_;
```
Karsten1987: just my 2cts, but a `unordered_map` has generally a faster lookup and does not need to reshuffle the tree when `insert` or `erase` is called on it.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -42,116 +43,233 @@
...
 #else
-  if (old_signal_handler) {
-    old_signal_handler(signal_value);
-  }
+    if (old_signal_handler) {
+      old_signal_handler(signal_value);
+    }
 #endif
+    signal_received.exchange(true);
+    events_condition_variable.notify_one();
```
sloretz: I think @hidmic mentioned this before, [calling  `notify_one()` from a signal handler is undefined behavior](https://stackoverflow.com/questions/31521182/c11-use-condition-variable-in-signal-handler/31521467). Is there anything else that can be used to wake the thread?

rclcpp/include/rclcpp/context.hpp
```diff
@@ -147,21 +151,21 @@ class Context : public std::enable_shared_from_this<Context>
    * - rcl_shutdown() is called on the internal rcl_context_t instance
    * - the shutdown reason is set
    * - each on_shutdown callback is called, in the order that they were added
-   * - if notify_all is true, rclcpp::notify_all is called to unblock some ROS functions
+   * - if notify all blocking sleep_for() calls, so they return early due to shutdown
```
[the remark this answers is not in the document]
wjwwood: Sorry, I guess I'm using `notify` and `interrupt` interchangeably here.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -42,116 +43,233 @@
...
+      throw;
+    }
+    RCLCPP_DEBUG(rclcpp::get_logger("rclcpp"), "signal handler installed");
+    return true;
+  }
 
+  bool
+  uninstall()
+  {
+    bool installed = installed_.exchange(false);
```
sloretz: Is the goal to make `install()` and `uninstall()` thread-safe with each other?

If so, starting from signal handlers uninstalled, it looks like thread1 calling `install()` could set `installed_` to `true`, context switch to thread2 which calls `uninstall()` that sets `installed_` to `false` and completes, then switch back to thread1 which installs the signal handler. The end result is the signal handler is installed but `is_installed()` would return `false`.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -42,116 +43,233 @@
...
+  }
 
-  for (auto context_ptr : rclcpp::get_contexts()) {
-    if (context_ptr->get_init_options().shutdown_on_sigint) {
-      // do not notify all, instead do that once after all are shutdown
-      context_ptr->shutdown("signal handler", false /* notify_all */);
+  void
+  deferred_signal_handler()
+  {
+    while (true) {
```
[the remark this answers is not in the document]
wjwwood: I don't think this would work because the location of the `if (!is_installed()) {` is important, in that it comes after the loop over the context's. So I'd pass on changing this logic unless @hidmic thinks it's a good idea to do so.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -42,116 +43,233 @@
...
+  {
+    return installed_.load();
+  }
 
+private:
+  SignalHandler() = default;
+
+  ~SignalHandler()
+  {
+    uninstall();
```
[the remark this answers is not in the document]
wjwwood: Indeed, I'll fix that up by catching and logging instead.

rclcpp/src/rclcpp/utilities.cpp
Karsten1987: that sounds correct to me. I guess these two functions have to be locked by a mutex then.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: I asked him about this, but I don't remember his answer. Do you know of an alternative? Only thing I can think of is a busy wait in the signal handler loop and an atomic, but that doesn't sound very nice.

sloretz: Searching only turns up things that are platform specific.

`sem_post()` for POSIX platforms http://pubs.opengroup.org/onlinepubs/009695399/functions/sem_post.html

I'm not sure about windows. I can't tell from the documentation if windows semaphores are signal safe, https://docs.microsoft.com/en-us/windows/desktop/sync/using-semaphore-objects

wjwwood:
I addressed some of the feedback, still need to address the locking of the install/uninstall functions.

Also the safety of notifying from a signal.

rclcpp/src/rclcpp/utilities.cpp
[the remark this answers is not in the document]
Karsten1987: that's fine for me. I was just wondering if we could somehow change the `while(true)`, but I guess in this case it's all right.

rclcpp/src/rclcpp/utilities.cpp
hidmic: @sloretz Yeah, that's the thing with signal handlers :sweat_smile: 

AFAIK there's no completely safe cross-platform synchronization mechanism that can be used from a signal handler (if only Windows was POSIX compliant...). Busy waits, with or without a timeout, are a portable option. Not a very neat one though. We can also do some more research on how to do this right for each platform.

I'd expect the current implementation to work in almost all cases though. Unless a signal comes right when the thread is about to do something with the condition variable. and that only takes place during a negligible fraction of the process lifetime.

sloretz: >Which I gather isn't safe enough? Is that your conclusion as well @sloretz?

It uses `pthread_cond_signal`, which [this page says](https://linux.die.net/man/3/pthread_cond_signal) 
```
It is not safe to use the pthread_cond_signal() function in a signal handler that is invoked 
asynchronously. Even if it were safe, there would still be a race between the test of the Boolean 
pthread_cond_wait() that could not be efficiently eliminated. 
```

wjwwood: Ok, so I was wrong, this is not safe for pthread, see:

> It is not safe to use the pthread_cond_signal() function in a signal handler that is invoked asynchronously. Even if it were safe, there would still be a race between the test of the Boolean pthread_cond_wait() that could not be efficiently eliminated.
>
> Mutexes and condition variables are thus not suitable for releasing a waiting thread by signaling from code running in a signal handler.
> -- https://linux.die.net/man/3/pthread_cond_signal

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5855/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2415/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4847/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5749/)

sloretz:
I think this is a step in the right direction, though it still doesn't solve ros2/build_cop#152. I can reproduce the same timed-out test with this PR using

```
colcon test --retest-until-fail 100 --event-handlers console_direct+ --packages-select test_communication --ctest-args -R test_action_client_server__Fibonacci__rclcpp__rmw_fastrtps_cpp\$
```

hidmic:
It used to happen roughly once every 10 times or so for me, so I agree here.

wjwwood:
So is this good to merge then @sloretz? @hidmic have you had a chance to try and reproduce it too with this branch? @sloretz and I were not able to do so, despite the fact that it appeared this way at first (it was failing, not hanging, for a different reason).

PR #607, opened 2018-12-11
use signal safe synchronization with platform specific semaphores

wjwwood:
This is a follow up of this comment: https://github.com/ros2/rclcpp/pull/605#discussion_r240439968

This is API and ABI neutral so we can merge it at anytime, but my preference would be after https://github.com/ros2/rclcpp/pull/605 but in for Crystal.


rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -277,15 +311,121 @@ typedef void (* signal_handler_t)(int);
...
+{
+  if (!g_wait_for_signal_is_setup.exchange(false)) {
+    return;
+  }
+#if defined(_WIN32)
+  CloseHandle(g_signal_handler_sem);
+#elif defined(__APPLE__)
+  dispatch_release(g_signal_handler_sem);
+#else  // posix
+  if (EINVAL == sem_destroy(&g_signal_handler_sem)) {
```
hidmic: @wjwwood I believe `sem_destroy` returns `-1` and sets `errno` to EINVAL instead.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -277,15 +311,121 @@ typedef void (* signal_handler_t)(int);
...
+#else  // posix
+  int s;
+  do {
+    s = sem_wait(&g_signal_handler_sem);
+  } while (-1 == s && EINTR == errno);
+#endif
+}
+
+void
+rclcpp::notify_signal_handler() noexcept
```
hidmic: @wjwwood same consideration about notifying the caller of an error condition here.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -277,15 +311,121 @@ typedef void (* signal_handler_t)(int);
...
+  dispatch_release(g_signal_handler_sem);
+#else  // posix
+  if (EINVAL == sem_destroy(&g_signal_handler_sem)) {
+    RCLCPP_ERROR(g_logger, "invalid semaphore in teardown_wait_for_signal()");
+  }
+#endif
+}
+
+void
+rclcpp::wait_for_signal() noexcept
```
hidmic: @wjwwood It seems to me this function could fail and the caller wouldn't know about it. If we want to keep it `noexcept`, consider returning a boolean.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -44,9 +44,37 @@
...
+// This is not thread-safe.
+RCLCPP_LOCAL
+void
+wait_for_signal() noexcept;
+
+// This notifies the deferred_signal_handler() thread to wake up in a signal safe way.
+// This is thread-safe.
+RCLCPP_LOCAL
+void
+notify_signal_handler() noexcept;
```
hidmic: @wjwwood consider making above functions private `static` methods of the `SignalHandler` class.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -44,9 +44,37 @@
 #define HAS_SIGACTION
 #endif
 
+rclcpp::Logger g_logger = rclcpp::get_logger("rclcpp");
```
hidmic: @wjwwood consider making this a `static` member of the `SignalHandler` or exposing it as a singleton like in [line 81](https://github.com/ros2/rclcpp/pull/607/files#diff-fe7ed4c8f373fc8cba1b5a0c60312137R81).

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -44,9 +44,37 @@
...
+
+// Undo all setup done in setup_wait_for_signal().
+// Must not call wait_for_signal() or notify_signal_handler() after calling this.
+// This is not thread-safe.
+RCLCPP_LOCAL
+void
+teardown_wait_for_signal() noexcept;
+
+// This waits for a notification from notify_signal_handler() in a signal safe way.
+// This is not thread-safe.
```
hidmic: @wjwwood do you mean it's not safe to be called from a signal handler?

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -277,15 +311,121 @@ typedef void (* signal_handler_t)(int);
...
+#endif
+
+static std::atomic_bool g_wait_for_signal_is_setup = ATOMIC_VAR_INIT(false);
+#if defined(_WIN32)
+HANDLE g_signal_handler_sem;
+#elif defined(__APPLE__)
+dispatch_semaphore_t g_signal_handler_sem;
+#else  // posix
+sem_t g_signal_handler_sem;
+#endif
```
hidmic: @wjwwood I think these variables could be pushed into the `SignalHandler` singleton.

rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -227,32 +258,37 @@ class SignalHandler final
...
         }
       }
       if (!is_installed()) {
+        RCLCPP_DEBUG(g_logger, "deferred_signal_handler(): signal handling uninstalled");
         break;
       }
-      std::unique_lock<std::mutex> events_lock(events_mutex);
-      events_condition_variable.wait(events_lock);
+      RCLCPP_DEBUG(g_logger, "deferred_signal_handler(): waiting for SIGINT or uninstall");
+      wait_for_signal();
```
hidmic: @wjwwood IIUC `wait_for_signal()` could fail. Should we log it / act upon it (e.g. forcing a shutdown)?

rclcpp/src/rclcpp/utilities.cpp
wjwwood: That's true.

sloretz:
It doesn't block this PR, just noting ros2/build_cop#152 is still present with this PR tested on linux with

```
colcon test --retest-until-fail 100 --event-handlers console_direct+ --packages-select test_communication --ctest-args -R test_action_client_server__Fibonacci__rclcpp__rmw_fastrtps_cpp\$
```

It hangs after a few runs.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: In this case, this function cannot fail and there's really no point in returning something here, since it would just mean I do `if (failed) // same thing I'm doing in the function now, i.e. RCLCPP_ERROR`.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: I'll just throw in this case, since it's only called from the singleton's thread.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: I did this in my latest round of changes, but retrospect I feel it's not necessary to stuff everything into the class. I'll leave them as-is though (in the class).

rclcpp/src/rclcpp/utilities.cpp
wjwwood: Again, lateral change, not really an improvement in my opinion, in fact kind of worse because you have initialize them outside of the class anyways (violates D.R.Y. more than necessary).

wjwwood: I did actually move them in my latest changes, however, the above is just in retrospect.

rclcpp/src/rclcpp/utilities.cpp
wjwwood: I'll throw from here, which will cause the program to terminate, which I think is the best thing to do, as this function really _shouldn't_ fail, it just _can_.

wjwwood:
> It hangs after a few runs.

I'm looking into it.

rclcpp/src/rclcpp/utilities.cpp
hidmic: That's exactly the argument I was going to make on whether this function should be silent about errors or not. I'm not sure if just throwing will do though, AFAIK it'll kill this thread but not the rest. Maybe logging a fatal level error and going into shutdown or raising an abort signal?

wjwwood: Throwing will kill it, since the exception is unhandled it will call `std::terminate()`.

wjwwood:
Ok, I tested this a bunch after @sloretz's comment and neither he nor I could get anything to hang on ctrl-c (sigint).

I also tried to address all the feedback, but I also had started refactoring the `SignalHandler` class into its own files and to clean it up to avoid so much `#ifdef`'ing in weird places. Sorry for the large diff in that case.

This is ready for re-review! Thanks :)

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -0,0 +1,356 @@
...
+SignalHandler::install()
+{
+  std::lock_guard<std::mutex> lock(install_mutex_);
+  bool already_installed = installed_.exchange(true);
+  if (already_installed) {
+    return false;
+  }
+  try {
+    setup_wait_for_signal();
+    signal_received_.exchange(false);
```
jacobperron:
```suggestion
    signal_received_.store(false);
```
Since we're not using the return value of `exchange`.

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -0,0 +1,356 @@
...
+    signal_handler_argument.sa_flags = SA_SIGINFO;
+#else
+    signal_handler_argument = signal_handler;
+#endif
+
+    old_signal_handler_ = SignalHandler::set_signal_handler(SIGINT, signal_handler_argument);
+
+    signal_handler_thread_ = std::thread(&SignalHandler::deferred_signal_handler, this);
+  } catch (...) {
+    installed_.exchange(false);
```
jacobperron:
```suggestion
    installed_.store(false);
```

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -0,0 +1,356 @@
...
+    throw std::runtime_error(msg);
+  }
+
+  return old_signal_handler;
+}
+
+void
+SignalHandler::signal_handler_common()
+{
+  SignalHandler::signal_received_.exchange(true);
```
jacobperron:
```suggestion
  SignalHandler::signal_received_.store(true);
```

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -0,0 +1,356 @@
...
+  RCLCPP_DEBUG(
+    get_logger(),
```
jacobperron: Since `get_logger()` is static, do we want this to be `SignalHandler::get_logger()` (along with all other calls to `get_logger()`)?

jacobperron: I guess the same could be said regarding all other class variable/function usages. I just noticed because L189 and L193 have the class name prefix (maybe they can be removed for consistency instead).

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -0,0 +1,356 @@
...
+    throw std::runtime_error(std::string("sem_init() failed: ") + strerror(errno));
+  }
+#endif
+  wait_for_signal_is_setup_.store(true);
+}
+
+void
+SignalHandler::teardown_wait_for_signal() noexcept
+{
+  if (!wait_for_signal_is_setup_.exchange(false)) {
```
jacobperron: Should we have a lock on setup/teardown wait for signal?

wjwwood: They are explicitly documented as not thread-safe, but no I don't think it's necessary, though I could be wrong about that.

If you have reason to believe they should be synchronized let me know.

jacobperron: I didn't think to hard about it, but thought I'd raise the question just in case. In retrospect, since they are only used within `install()` and `uninstall()`, which are already synchronized, and are private I guess it's okay.

rclcpp/src/rclcpp/signal_handler.cpp
wjwwood: These are redundant, but sometimes helpful to indicate where you're pull information from (e.g. like using `this->method()` rather than just `method()`), so usually I try to just leave that up to the developer's whim, but in this case I'll remove some of the ones you pointed out for consistency.

wjwwood:
@jacobperron can you quickly re-review since I addressed more feedback, fixed cpplint, and rebased?

rclcpp/src/rclcpp/signal_handler.hpp
```diff
@@ -0,0 +1,189 @@
...
+  static
+  SignalHandler &
+  get_global_signal_handler();
+
+  /// Return a global singleton logger to avoid needing to create it everywhere.
+  static
+  rclcpp::Logger &
+  get_logger();
+
+  /// Install the signal handler for SIGINT and start the dedicated singal handling thread.
```
jacobperron:
```suggestion
  /// Install the signal handler for SIGINT and start the dedicated signal handling thread.
```

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5880/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2430/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4863/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5771/)

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -0,0 +1,357 @@
...
+    signal_handler_argument.sa_flags = SA_SIGINFO;
+#else
+    signal_handler_argument = signal_handler;
+#endif
+
+    old_signal_handler_ = SignalHandler::set_signal_handler(SIGINT, signal_handler_argument);
+
+    signal_handler_thread_ = std::thread(&SignalHandler::deferred_signal_handler, this);
+  } catch (...) {
+    installed_.store(false);
```
hidmic: @wjwwood I'm having second thoughts about this. I initially wrote it, but `installed_` isn't the only state that we're mutating here. We could end up with half installed things (e.g. `setup_wait_for signal()` succeeds but `set_signal_handler` does not and throws). Maybe we could make both class and instance methods idempotent? What do you think?

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -0,0 +1,357 @@
...
+  try {
+    // TODO(wjwwood): what happens if someone overrides our signal handler then calls uninstall?
+    //   I think we need to assert that we're the current signal handler, and mitigate if not.
+    set_signal_handler(SIGINT, old_signal_handler_);
+    RCLCPP_DEBUG(get_logger(), "SignalHandler::uninstall(): notifying deferred signal handler");
+    notify_signal_handler();
+    signal_handler_thread_.join();
+    teardown_wait_for_signal();
+  } catch (...) {
+    installed_.exchange(true);
```
hidmic: @wjwwood same as above.

rclcpp/src/rclcpp/signal_handler.cpp
wjwwood: You can do that if you like, but let’s do it in a follow up pr so this doesn’t continue to block.

wjwwood:
Waiting on Windows ci still.

rclcpp/src/rclcpp/signal_handler.cpp
hidmic: Sounds good !

wjwwood:
I had a small bug in the Windows `#ifdef`, I just fixed it in dd6a771 and retriggered Windows CI only (since it's within an `#if defined(_WIN32)` block):

[Build Status](https://ci.ros2.org/job/ci_windows/5785/)

I'm still looking for another approval if anyone has time.

wjwwood:
Oops, I had to fix one more thing (actually it should have worked since `errno` is global, but still) in beaed30, but CI hasn't started yet, so it should still be valid to cover both changes.

nuclearsandwich:
> Build Status

The failing tests here are also failing in the most recent nightlies. Compared to them, this PR introduces no new failing tests. With the green status of other tests and the investigation into the new failures ongoing I would propose that we consider this PR merge-able.

wjwwood:
Sounds good to me, merging.

Issue #599, opened 2018-12-04
SyncParametersClient creation for a class inheriting from rclcpp::Node


alsora:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - from binary
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
```
#include "rclcpp/rclcpp.hpp"

class MyNode : public rclcpp::Node {

public:

    MyNode() : Node("my_node")
    {
        _async_param_client = std::make_shared<rclcpp::AsyncParametersClient>(this, "other_node");
        _sync_param_client = std::make_shared<rclcpp::SyncParametersClient>(this, "other_node");
    }

private:

    rclcpp::SyncParametersClient::SharedPtr _sync_param_client;
    rclcpp::AsyncParametersClient::SharedPtr _async_param_client;
};


```

#### Expected behavior

I create a new class which inherits from `rclcpp::Node` and I want to create synchronous and asynchronous parameter clients.

#### Actual behavior

The syncrhonous parameter client creation fails due to 

> /usr/include/c++/7/ext/new_allocator.h:136:4: error: no matching function for call to ‘rclcpp::SyncParametersClient::SyncParametersClient(MyNode*, const char [11])’
>   { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }

On the other hand the asynchronous parameter client works.

It still do not work if I pass as argument `this->get_node_base_interface()`

clalancette:
Right, it looks like that on bouncy, we just don't have an overloaded `SyncParametersClient` constructor that takes in an `rclcpp::Node *`, while we do have one for `AsyncParametersClient`.  I'll mark this as an enhancement, though it is a fairly low priority for us at the moment.  If you have time to add that constructor, that would move it along faster.  Thanks.

alsora:
Ok perfect, I will make a PR very soon.

wjwwood:
You could use `this->shared_from_this()`, but that won't work in the constructor.

So a raw pointer version would be good.

PR #612, opened 2018-12-12
Added new constructors for SyncParameterClient

alsora:
Following issue https://github.com/ros2/rclcpp/issues/599

Allows to create SyncParameterClient in a class which inherits from rclcpp::Node

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -199,6 +199,16 @@ public:
     const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
 
   RCLCPP_PUBLIC
+  SyncParametersClient(
+    rclcpp::executor::Executor::SharedPtr executor,
+    const rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_interface,
+    const rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr node_topics_interface,
+    const rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph_interface,
+    const rclcpp::node_interfaces::NodeServicesInterface::SharedPtr node_services_interface,
+    const std::string & remote_node_name = "",
+    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
+
+  RCLCPP_PUBLIC
   std::vector<rclcpp::Parameter>
   get_parameters(const std::vector<std::string> & parameter_names);
 
@@ -281,7 +291,7 @@ public:
 
 private:
   rclcpp::executor::Executor::SharedPtr executor_;
-  rclcpp::Node * node_;
+  const rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_interface_;
   AsyncParametersClient::SharedPtr async_parameters_client_;
 };
 
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -281,7 +281,7 @@ public:
 
 private:
   rclcpp::executor::Executor::SharedPtr executor_;
-  rclcpp::Node* node_;
+  rclcpp::Node * node_;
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -356,6 +356,28 @@ SyncParametersClient::SyncParametersClient(
...
+    remote_node_name,
+    qos_profile)
+{}
+
+SyncParametersClient::SyncParametersClient(
+  rclcpp::executor::Executor::SharedPtr executor,
+  rclcpp::Node * node,
+  const std::string & remote_node_name,
+  const rmw_qos_profile_t & qos_profile)
+: executor_(executor), node_(node)
```
wjwwood: This is dangerous, because you're creating a shared pointer to Node (in `node_`) from a pointer that is already owned by another shared pointer (`node` is the pointer and all nodes are created as shared pointers to support `std::enable_shared_from_this`).

The SyncParametersClient needs to be changed to hold a raw pointer (optionally or always) rather than always having a shared pointer to node which cannot be safely fulfilled with a raw pointer as input.

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -185,6 +185,19 @@ class SyncParametersClient
     const std::string & remote_node_name = "",
     const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
 
+  RCLCPP_PUBLIC
+  explicit SyncParametersClient(
+    rclcpp::Node * node,
```
wjwwood: Instead a signature which took the component parts of the node that this client needs would be better, e.g. maybe the `NodeParameterInterface` and the `NodeBaseInterface` (and perhaps others?).

wjwwood: You could keep a signature like this one which simply extracted those components from a `rclcpp::Node` and delegated them to another constructor. That way this could be more easily used with other forms of Node's like `rclcpp_lifecycle::LifecycleNode`.

rclcpp/src/rclcpp/parameter_client.cpp
alsora: I changed the SyncParametersClient member node_ to a raw pointer.
SyncParametersClient has still constructors which accept a raw or shared pointer to a node.

rclcpp/include/rclcpp/parameter_client.hpp
alsora: From what I have understood, SyncParameterClient only uses `NodeBaseInterface` for calling `spin_until_future_complete`.
All the other interfaces are used by AsyncParameterClient and are "extracted" from the node pointer in the AsyncParameterClient constructor. 
I do not get what's the advantage of having a SyncParameterClient constructor signature which takes interfaces, since all the logic using them is contained in the AsyncParameterClient

wjwwood:
> In the last commit I took into account the possibility for objects which does not inherit from rclcpp::Node to create a SyncParametersClient.

Thanks! I'll give it a look.

> Do you have any idea why changing the ParameterClient can affect the rclcpp_action client ?
The test error is a timeout during async_cancel_some_goals

It may or may not be related, we do occasionally have flaky tests, @jacobperron and @sloretz do either of you know if that is a known flaky test?

jacobperron:
> It may or may not be related, we do occasionally have flaky tests, @jacobperron and @sloretz do either of you know if that is a known flaky test?

I'm not aware of any flaky tests in rclcpp_action. 
With either master branch or this PR, the following command passes for me locally:

```
colcon test --executor sequential --retest-until-fail 10 --packages-select rclcpp_action
```

alsora:
@jacobperron yes, you are right, now the checks are marked as passed.

I repeated all the rclcpp tests more than one time and apparently sometimes errors in the rclcpp_action package arise.

wjwwood:
Restarted Windows CI due to failure: [Build Status](https://ci.ros2.org/job/ci_windows/5825/)

wjwwood:
The failing tests are due to another unrelated issue, so I'll merge this. Thanks @alsora!

PR #618, opened 2019-01-11
Bind is no longer in std::__1

kuzai:
The namespace of the __bind function in llvm is now in std and not std::__1


wjwwood:
Thanks!

PR #619, opened 2019-01-17
Get parameter map (#575) (crystal-backport)

clalancette:
This is the backport of 99dd0313ab0f0396dd01a37b827ab6963075ee41 onto the crystal branch, since I'd like to get it into the 2nd crystal sync https://github.com/ros2/ros2/issues/647 .  There are no differences between this change and the one that was merged onto master, and this change only adds an API, so there is no API break.

* Add in the ability to get parameters in a map.

Any parameters that have a "." in them will be considered to
be part of a "map" (though they can also be get and set
individually).  This PR adds two new template specializations
to the public node API so that it can take a map, and store
the list of values (so setting the parameter with a name of
"foo" and a key of "x" will end up with a parameter of "foo.x").
It also adds an API to get all of the keys corresponding to
a prefix, and returing that as a map (so a get of "foo" will
get all parameters that begin with "foo.").  Note that all
parameters within the map must have the same type, otherwise
an rclcpp::ParameterTypeException will be thrown.

* Fix style problems pointed out by uncrustify/cpplint.

* Move tests for set_parameter_if_not_set/get_parameter map to rclcpp.

* Rename get_parameter -> get_parameters.

* Add in documentation from review.

### DIFF
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
@@ -91,13 +91,6 @@ public:
 
   RCLCPP_PUBLIC
   virtual
-  bool
-  get_parameters_by_prefix(
-    const std::string & prefix,
-    std::map<std::string, rclcpp::Parameter> & parameters) const;
-
-  RCLCPP_PUBLIC
-  virtual
   std::vector<rcl_interfaces::msg::ParameterDescriptor>
   describe_parameters(const std::vector<std::string> & names) const;
 
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
@@ -91,6 +91,13 @@ public:
 
   RCLCPP_PUBLIC
   virtual
+  bool
+  get_parameters_by_prefix(
+    const std::string & prefix,
+    std::map<std::string, rclcpp::Parameter> & parameters) const;
+
+  RCLCPP_PUBLIC
+  virtual
   std::vector<rcl_interfaces::msg::ParameterDescriptor>
   describe_parameters(const std::vector<std::string> & names) const;
 
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
@@ -15,6 +15,7 @@
 #ifndef RCLCPP__NODE_INTERFACES__NODE_PARAMETERS_INTERFACE_HPP_
 #define RCLCPP__NODE_INTERFACES__NODE_PARAMETERS_INTERFACE_HPP_
 
+#include <map>
 #include <string>
 #include <vector>
 
@@ -88,6 +89,20 @@ public:
   get_parameter(
     const std::string & name,
     rclcpp::Parameter & parameter) const = 0;
+
+  /// Get all parameters that have the specified prefix into the parameters map.
+  /*
+   * \param[in] prefix the name of the prefix to look for.
+   * \param[out] parameters a map of parameters that matched the prefix.
+   * \return true if any parameters with the prefix exists on the node, or
+   * \return false otherwise.
+   */
+  RCLCPP_PUBLIC
+  virtual
+  bool
+  get_parameters_by_prefix(
+    const std::string & prefix,
+    std::map<std::string, rclcpp::Parameter> & parameters) const = 0;
 
   RCLCPP_PUBLIC
   virtual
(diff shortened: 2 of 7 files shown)


dirk-thomas:
> This is the backport of 99dd031 onto the crystal branch, since I'd like to get it into the 2nd crystal sync 

Is there a specific use case / problem why this additional API is needed in Crystal?

clalancette:
> Is there a specific use case / problem why this additional API is needed in Crystal?

Yes, it's so that we can release https://github.com/ros2/teleop_twist_joy/pull/8 into Crystal (once it has been approved/merged).

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
```diff
@@ -89,6 +89,13 @@ class NodeParameters : public NodeParametersInterface
     const std::string & name,
     rclcpp::Parameter & parameter) const;
 
+  RCLCPP_PUBLIC
+  virtual
```
sloretz: I think adding a virtual method breaks ABI

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
```diff
@@ -89,6 +90,20 @@ class NodeParametersInterface
...
 
+  /// Get all parameters that have the specified prefix into the parameters map.
+  /*
+   * \param[in] prefix the name of the prefix to look for.
+   * \param[out] parameters a map of parameters that matched the prefix.
+   * \return true if any parameters with the prefix exists on the node, or
+   * \return false otherwise.
+   */
+  RCLCPP_PUBLIC
+  virtual
```
sloretz: Same comment about adding a virtual method

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
clalancette: Never mind, I found another way.  Update to this PR coming.

dirk-thomas:
Since this patch has changed from a backport to a modified patch with different logic can you please describe how the backported change is different than what is available on master and if you are planning to update the code on master to match this patch or if it should stay as it is.

clalancette:
> Since this patch has changed from a backport to a modified patch with different logic can you please describe how the backported change is different than what is available on master and if you are planning to update the code on master to match this patch or if it should stay as it is.

Right.  For whatever reason when I did the original patch, I totally missed that [list_parameters](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L345) existed, which more-or-less provides the same functionality as [get_parameters_by_prefix](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L287).  The main benefit to using `get_parameters_by_prefix` is that it is atomic; it takes the mutex, iterates over all of the keys with the prefix, gets their values, and stuffs them in the map to be returned.  When using `list_parameters` as in this patch, the mutex is taken to get the list of parameters with the prefix, but then it is dropped.  We then have to call `get_parameter` in a loop to get the value for each parameter.  This leads to a possible TOCTTOU bug, where we could fetch the list of parameters, one is deleted, and then we try to `get_parameter` on the now-deleted parameter.

My view on it is that we should keep `get_parameters_by_prefix` for master since it is atomic (and slightly faster), but we can live with the (probably rare) race in crystal to make sure we don't break ABI.  Thoughts?

dirk-thomas:
I am not convinced if it is worth introducing a new API into Crystal in a patch release which will not be present in master / future distros.

clalancette:
> I am not convinced if it is worth introducing a new API into Crystal in a patch release which will not be present in master / future distros.

But that's not what would happen here.

On master, #575 added the following new APIs:
* Node class:
  * `set_parameters_if_not_set`
  * `get_parameters`
* NodeParameters class:
  * `get_parameters_by_prefix`

In Crystal, this PR would add the following new APIs:
* Node class:
  * `set_parameters_if_not_set`
  * `get_parameters`

So the APIs are compatible, its just that code that compiles against master would be able to call `get_parameters` slightly faster and with atomicity.  Either way, code that calls `set_parameters_if_not_set` or `get_parameters` would work across Crystal and master.

dirk-thomas:
> So the APIs are compatible, its just that code that compiles against master would be able to call `get_parameters` slightly faster and with atomicity.

Do you mean `get_parameters_by_prefix` here?

If `get_parameters_by_prefix` is so much better should `get_parameters` be removed so that users don't accidentally call a function which isn't really safe to use?

clalancette:
> 
> If `get_parameters_by_prefix` is so much better should `get_parameters` be removed so that users don't accidentally call a function which isn't really safe to use?

`get_parameters_by_prefix` is really an implementation detail of how to get parameters from the lower-level `NodeParameters` class.  Users should really be interacting with the top-level `Node` APIs, which are `get_parameters` and `set_parameters_if_not_set`.  On crystal, the Node API for `get_parameters` will be imlpemented by calling `list_parameters`, followed by a call to `get_parameter` for each of the returned values (this PR).  This has the benefit of using existing APIs (preserving ABI), but has the downside that fetching the set of values with a prefix isn't atomic.  On master, `get_parameters` is implemented by calling `get_parameters_by_prefix`, which has the downside of needing to change ABI, but has the benefit of fetching the set of values atomically.

In my mind, the question is really whether we want to add `get_parameters_by_prefix` to master at all.  I think we do, since it provides atomicity of fetching the set of values, but I would definitely listen to arguments against that.

PR #621, opened 2019-01-28
Avoid race that triggers timer too often

durko:
We are experiencing sporadic failures of `TestMultiThreadedExecutor.timer_over_take` in our CI, where the measured time difference between two timer executions is close to zero.

Looking at the code there is a subtle opportunity for a race condition in `rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp`:
```c
      ...
      if (!get_next_executable(any_exec)) {                                                                                                                                                    
        continue;                                                                                                                                                                              
      }  
                                                                                                                                                                                      
      /* 
        IMAGINE OR INSERT A SHORT FEW MILLISECOND SLEEP HERE 
      */

      if (any_exec.timer) {                                                                                                                                                                    
        // Guard against multiple threads getting the same timer.                                                                                                                              
        std::lock_guard<std::mutex> lock(scheduled_timers_mutex_);                                                                                                                             
        if (scheduled_timers_.count(any_exec.timer) != 0) {                                                                                                                                    
          continue;                                                                                                                                                                            
        }                                                                                                                                                                                      
        scheduled_timers_.insert(any_exec.timer);                                                                                                                                              
      }
      ...                                                                                                                                                                                      
```

The race condition is triggered if:
- a thread gets an already scheduled timer
- acquiring the second lock_guard is slightly delayed

In this case a previously fetched and scheduled timer handle could be removed from the list of scheduled timers by the thread that first got the timer before the delayed check. This results in a second execution of the timer callback.

For me replacing the above comment with 
```
std::this_thread::sleep_for(std::chrono::milliseconds(5));
```
causes the test to always fail.

To mitigate, `get_next_executable` and `if (scheduled_timers_.count ... )` have to be protected by one `lock_guard`. This pull request takes this requirement into account and removes the now redundant second `lock_guard`.

### DIFF
--- a/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
@@ -78,7 +78,6 @@ private:
   size_t number_of_threads_;
   bool yield_before_execute_;
 
-  std::mutex scheduled_timers_mutex_;
   std::set<TimerBase::SharedPtr> scheduled_timers_;
 };
 
--- a/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
@@ -82,7 +82,6 @@ MultiThreadedExecutor::run(size_t)
       }
       if (any_exec.timer) {
         // Guard against multiple threads getting the same timer.
-        std::lock_guard<std::mutex> lock(scheduled_timers_mutex_);
         if (scheduled_timers_.count(any_exec.timer) != 0) {
           continue;
         }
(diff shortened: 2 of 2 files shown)


wjwwood:
Looks like this is gap left over from https://github.com/ros2/rclcpp/pull/383

wjwwood:
Here's CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6460/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2830/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5305/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6279/)

I'll merge if CI is green. Sorry it took so long, I needed a junk of time to really test it and think about it clearly.

Thanks for the patch!

wjwwood:
Looks like this needs to be rebased now. @durko if you have time, that would be great, otherwise I'll try to rebase it for you as soon as I have time.

wjwwood:
I rebased it, here's new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6491/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2846/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5321/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6295/)

Issue #492, opened 2018-06-05
Do argument parsing outside of node constructor


sloretz:
## Feature request
Things that come from command line arguments should be separately passed into the node's constructor rather than passing in arguments and asking the node to do the parsing.

#### Feature description
Currently command line arguments may be passed to a node via an argument to the node's constructor. This separates the responsibility of parsing arguments from actually using them. An entity like ros2 launch launch composable nodes would invoke parsing of any command line arguments, and then pass the results into the node.

#### Implementation considerations
The current arguments passed to the node are https://github.com/ros2/rclcpp/blob/8f793fdb4aa6d21ea2442a31547546a3c90692a5/rclcpp/include/rclcpp/node.hpp#L94-L101

The information that should be passed in if not through the arguments is:
* Topic and service remap rules
* node name and namespace remap rules `__ns:=...`, `__node:=...`
* Initial parameter values `__params:=...`
* Log level (ros2/ros2#498)

Since new features may be added as command line arguments it may be an advantage to pass `rclcpp::NodeArguments()` so future features don't have to change all the places where arguments are passed to a node's constructor

requires ros2/rcl#254

wjwwood:
@mjcarroll is this actually done?

mjcarroll:
Ah, sorry.  With a quick glance I thought this was covered with `NodeOptions`, but the parsing is still happening inside the node, will re-open.

PR #622, opened 2019-01-29
Replace node constructor arguments with NodeOptions

mjcarroll:
Creates a NodeOptions object to pass all of the necessary options to the node constructor.

Connects to #492

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -81,7 +81,7 @@ public:
   explicit Node(
     const std::string & node_name,
     const std::string & namespace_ = "",
-    bool use_intra_process_comms = false);
+    const NodeOptions & options = NodeOptions());
 
   /// Create a node based on the node name and a rclcpp::Context.
   /**
@@ -102,11 +102,7 @@ public:
     const std::string & node_name,
     const std::string & namespace_,
     rclcpp::Context::SharedPtr context,
-    const std::vector<std::string> & arguments,
-    const std::vector<Parameter> & initial_parameters,
-    bool use_global_arguments = true,
-    bool use_intra_process_comms = false,
-    bool start_parameter_services = true);
+    const NodeOptions & options);
 
   RCLCPP_PUBLIC
   virtual ~Node();
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -79,22 +79,19 @@ public:
   RCLCPP_PUBLIC
   explicit Node(
     const std::string & node_name,
-    const std::string & namespace_ = "",
-    const NodeOptions & options = NodeOptions());
+    const NodeOptions & options = NodeOptions::Builder().build());
 
-  /// Create a node based on the node name and a rclcpp::Context.
+  /// Create a new node with the specified name.
   /**
    * \param[in] node_name Name of the node.
    * \param[in] namespace_ Namespace of the node.
-   * \param[in] context The context for the node (usually represents the state of a process).
    * \param[in] options Additional options to control creation of the node.
    */
   RCLCPP_PUBLIC
-  Node(
+  explicit Node(
     const std::string & node_name,
     const std::string & namespace_,
-    rclcpp::Context::SharedPtr context,
-    const NodeOptions & options);
+    const NodeOptions & options = NodeOptions::Builder().build());
 
   RCLCPP_PUBLIC
   virtual ~Node();
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -74,8 +74,7 @@ public:
   /**
    * \param[in] node_name Name of the node.
    * \param[in] namespace_ Namespace of the node.
-   * \param[in] use_intra_process_comms True to use the optimized intra-process communication
-   * pipeline to pass messages between nodes in the same process using shared memory.
+   * \param[in] options Additional options to control creation of the node.
    */
   RCLCPP_PUBLIC
   explicit Node(
@@ -88,14 +87,7 @@ public:
    * \param[in] node_name Name of the node.
    * \param[in] namespace_ Namespace of the node.
    * \param[in] context The context for the node (usually represents the state of a process).
-   * \param[in] arguments Command line arguments that should apply only to this node.
-   * \param[in] initial_parameters a list of initial values for parameters on the node.
-   * This can be used to provide remapping rules that only affect one instance.
-   * \param[in] use_global_arguments False to prevent node using arguments passed to the process.
-   * \param[in] use_intra_process_comms True to use the optimized intra-process communication
-   * pipeline to pass messages between nodes in the same process using shared memory.
-   * \param[in] start_parameter_services True to setup ROS interfaces for accessing parameters
-   * in the node.
+   * \param[in] options Additional options to control creation of the node.
    */
   RCLCPP_PUBLIC
   Node(
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -78,7 +78,7 @@ public:
   RCLCPP_PUBLIC
   explicit Node(
     const std::string & node_name,
-    const NodeOptions & options = NodeOptions::Builder().build());
+    const NodeOptions & options = NodeOptions());
 
   /// Create a new node with the specified name.
   /**
@@ -90,7 +90,7 @@ public:
   explicit Node(
     const std::string & node_name,
     const std::string & namespace_,
-    const NodeOptions & options = NodeOptions::Builder().build());
+    const NodeOptions & options = NodeOptions());
 
   RCLCPP_PUBLIC
   virtual ~Node();
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -73,7 +73,6 @@ public:
   /// Create a new node with the specified name.
   /**
    * \param[in] node_name Name of the node.
-   * \param[in] namespace_ Namespace of the node.
    * \param[in] options Additional options to control creation of the node.
    */
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -41,6 +41,7 @@
 #include "rclcpp/logger.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/message_memory_strategy.hpp"
+#include "rclcpp/node_options.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/node_interfaces/node_clock_interface.hpp"
 #include "rclcpp/node_interfaces/node_graph_interface.hpp"
--- a/rclcpp/include/rclcpp/node_options.hpp
+++ b/rclcpp/include/rclcpp/node_options.hpp
@@ -28,7 +28,7 @@
 namespace rclcpp
 {
 
-/// Encapsulation of options for initializing node.
+/// Encapsulation of options for node initialization.
 class NodeOptions
 {
 public:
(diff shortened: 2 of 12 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -73,39 +74,27 @@ class Node : public std::enable_shared_from_this<Node>
...
   Node(
     const std::string & node_name,
     const std::string & namespace_,
     rclcpp::Context::SharedPtr context,
-    const std::vector<std::string> & arguments,
-    const std::vector<Parameter> & initial_parameters,
-    bool use_global_arguments = true,
-    bool use_intra_process_comms = false,
-    bool start_parameter_services = true);
+    const NodeOptions & options);
```
wjwwood: Why not include context in the options?

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -0,0 +1,127 @@
...
+  bool & start_parameter_services();
+
+protected:
+  /// Clean up internal rcl_node_options_t structure.
+  void
+  finalize_node_options();
+
+  /// Retrieve the ROS_DOMAIN_ID environment variable and populate options.
+  void
+  set_domain_id_from_env();
```
wjwwood: If you're going to use the new line between return type and function name (I personally like that style because it vertically aligns the start of all the function names), then use it everywhere in the class, but if not change these two to not wrap. Either way, I'd just ask you to change them to be consistent within the class.

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -0,0 +1,127 @@
...
+  const bool & use_global_arguments() const;
+  bool & use_global_arguments();
+
+  /// Return a reference to the use_intra_process_comms flag
+  /**
+   * True to use the optimized intra-process communication pipeline to pass messages bewteen nodes
+   * in the same process using shared memory.
+   */
+  const bool & use_intra_process_comms() const;
+  bool & use_intra_process_comms();
```
wjwwood: Do you like this style? Imagine this:

```c++
no.use_intra_process_comms() = true;
no.start_parameter_services() = false;
```

Versus:

```c++
no.use_intra_process_comms(true);
no.start_parameter_services(false);
```

Or even the parameter idiom:

```c++
no.use_intra_process_comms(true).start_parameter_services(false);
```

I was personally hoping for an interface that supports the parameter idiom and maybe even a generic Options class that helps setup other options classes (because we're going to need it for publishers and subscribing, etc).

Another related style is the builder pattern:

```c++
NodeOptionsBuilder nob;

auto no = nob.use_intra_process_comms(true).start_parameter_services(false).build();
```

Which is better because it makes the `NodeOptions` class simpler, but introduces another class, which I don't care for.

mjcarroll:
CI (up to rclcpp)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6098/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2573/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5026/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5966/)

mjcarroll:
Windows failures are likely due to missing symbol visibility.

rclcpp/test/test_node_global_args.cpp
```diff
@@ -33,32 +34,29 @@ class TestNodeWithGlobalArgs : public ::testing::Test
...
-  auto context = rclcpp::contexts::default_context::get_global_default_context();
-  const std::vector<std::string> arguments = {"__node:=local_arguments_test"};
-  const std::vector<rclcpp::Parameter> initial_values = {};
-  const bool use_global_arguments = true;
-  const bool use_intra_process = false;
-  auto node = rclcpp::Node::make_shared(
-    "orig_name", "", context, arguments, initial_values, use_global_arguments, use_intra_process);
+  auto options = rclcpp::NodeOptions::Builder()
+    .arguments({"__node:=local_arguments_test"})
+    .build();
```
wjwwood: I'd like to cast a vote (again) in favor of dropping the builder pattern for the parameter idiom pattern.

I really don't see the benefit of the builder pattern in this case. The only advantage it gives (that I can see) is that the resulting options structure is immutable, but I don't see why that's important or advantages.

On the other hand, the parameter idiom has the advantage of almost always being less code (both on the user's side and on the implementation side), and makes it easier to start with a common options structure, and then customize from there. For example:

```
auto node = Node::make_shared("my_node", get_my_projects_default_node_options().use_intra_process(true));
```

You could enable this by having a version of the constructor for the builder that takes an existing NodeOptions, but it would still require more code.

Maybe you guys have an idea of why an immutable NodeOptions would be nice to have that I can't think of right now, but without that I think the parameter idiom version would be better.

mjcarroll: I don't think that there is a substantial reason for the immuatable, and I think in retrospect, the parameter idiom looks better.  I'll rework this.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2591/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5045/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5986/)

mjcarroll:
Added missing lifecycle fixes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6119/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2592/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5046/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5987/)

rclcpp/src/rclcpp/node_options.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+NodeOptions &
+NodeOptions::allocator(rcl_allocator_t allocator)
+{
+  this->allocator_ = allocator;
+  return *this;
+}
+
+size_t
+NodeOptions::get_domain_id_from_env() const
```
hidmic: @mjcarroll maybe out of scope, but shouldn't we be using [`rcutils_get_env()`](https://github.com/ros2/rcutils/blob/master/include/rcutils/get_env.h#L62) instead?

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -0,0 +1,197 @@
...
+
+private:
+  // This is mutable to allow for a const accessor which lazily creates the node options instance.
+  /// Underlying rcl_node_options structure.
+  mutable std::unique_ptr<rcl_node_options_t, void (*)(rcl_node_options_t *)> node_options_;
+
+  // IMPORTANT: if any of these default values are changed, please update the
+  // documentation in this class.
+
+  rclcpp::Context::SharedPtr context_ {
```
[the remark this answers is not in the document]
wjwwood: `uncrustify` was happy when I ran it.

rclcpp/src/rclcpp/node_options.cpp
```diff
@@ -0,0 +1,243 @@
...
+const std::vector<std::string> &
+NodeOptions::arguments() const
+{
+  return this->arguments_;
+}
+
+NodeOptions &
+NodeOptions::arguments(const std::vector<std::string> & arguments)
+{
+  this->node_options_.reset();  // reset node options to make it be recreated on next access.
```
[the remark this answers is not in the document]
wjwwood: This was actually my change, I think there are more that should, let me recheck this.

rclcpp/src/rclcpp/node_options.cpp
wjwwood: This does more than read the environment variable (I think), but it was probably a todo.

Maybe related: https://github.com/ros2/rcl/issues/119

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -0,0 +1,197 @@
...
+  context(rclcpp::Context::SharedPtr context);
+
+  /// Return a reference to the list of arguments for the node.
+  RCLCPP_PUBLIC
+  const std::vector<std::string> &
+  arguments() const;
+
+  /// Set the arguments, return this for parameter idiom.
+  /**
+   * This will cause the internal rcl_node_options_t struct to be invalidated.
```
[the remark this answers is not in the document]
wjwwood: This goes along with resetting the unique ptr, I'll double check this.

wjwwood:
I had to force-push to address the DCO issue.

rclcpp/src/rclcpp/node_options.cpp
mjcarroll: It's just pulling the environment variable and making sure that it's integral. It could probably be replaced.  This was primarily to move it out of the `NodeBase` constructor.

hidmic: @wjwwood I see, and that issue comes with long standing discussions. Disregard my comment but I think we should have a TODO here too.

wjwwood:
OK, I added a todo and @mjcarroll and I tested it locally, so here's full CI, hopefully it's green:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6124/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2597/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5051/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5992/)

wjwwood:
@mjcarroll do you think it would be possible to make this change tick/tock? I guess we could have left the existing constructors and deprecated them?

A separate question is whether or not it is worth doing that? @nuclearsandwich do you have an opinion on whether or not we should try to do that retroactively?

----

Either way we need an entry in the change summary page for dashing:

https://index.ros.org/doc/ros2/Releases/Release-Dashing-Diademata/#id4

nuclearsandwich:
> @nuclearsandwich do you have an opinion on whether or not we should try to do that retroactively?

I'm slightly confused about the tick and tock. Would the tick deprecating the old constructors be Dashing and then they'd be removed from E-turtle? From offline conversation yesterday I think this new API is required for a lot of things we want to introduce. Does maintaining the old constructors interfere with that? APIs that land in Dashing will be with us until May 2021 whether they be deprecated or not. I think that as long as we update the change guide now that this has merged we will be giving people plenty of time to respond to it and make the necessary updates.

wjwwood:
> I'm slightly confused about the tick and tock. Would the tick deprecating the old constructors be Dashing and then they'd be removed from E-turtle?

Exactly.

> From offline conversation yesterday I think this new API is required for a lot of things we want to introduce. Does maintaining the old constructors interfere with that? APIs that land in Dashing will be with us until May 2021 whether they be deprecated or not.

It wouldn't affect those, as we'd only add new options to the `NodeOptions` struct and not the constructor. So the existing constructor could be used for some existing options, but would be deprecated and new options would only be available via the new `NodeOptions` constructor.

> I think that as long as we update the change guide now that this has merged we will be giving people plenty of time to respond to it and make the necessary updates.

That was my original thinking, but I just wanted to leave the door open if that wasn't ok with everyone.

nuclearsandwich:
> That was my original thinking, but I just wanted to leave the door open if that wasn't ok with everyone.

It's okay with me at least :+1: please do link the ros2_documentation PR when it's up.

dirk-thomas:
Merging this PR while the corresponding rcl PR ros2/rcl#379 hasn't been merged breaks master: https://ci.ros2.org/job/ci_linux/6132/

wjwwood:
I merged the outstanding pr, that's my fault. I opened all connected prs in waffle, but either that one didn't show in waffle or I misclicked. Should be fixed now.

mjcarroll:
It was actually my fault, I connected the `rcl` issue to this pull request and not the other issue.  Sorry about that.

wjwwood:
No worries, I shouldn't 🙈 trust waffle :)

PR #629, opened 2019-02-12
Replace 'auto' to 'const auto &' in for loops

youtalk:
I found some call by value in the for loop codes but they are preferred to replace to call by reference.
This PR simply solves it.


wjwwood:
@youtalk did you intend to close this pull request? I think it makes sense and would start CI so it could be merged if you opened it again.

Issue #634, opened 2019-02-18
Characterization test for rclcpp comm


ivanpauno:
## Feature request

Related with https://github.com/ros2/ros2/issues/649. The idea is to measure the performance of rclcpp communication, so in the future we can compare different intraprocess mechanisms, intraprocess vs interprocess, etc.

#### Feature description
Check here: https://gist.github.com/ivanpauno/daee6069a034dabb6374c0d987162cf6

#### Implementation considerations

Planned tasks:
1. [x] Use cpp timers to measure publish-to-subscription latency and publish/subscription rate. Log results (https://github.com/ros2/rclcpp/pull/636).
2. [ ] Use gperftools to measure cpu/memory usage, publish/subscribe operation time.
3. [ ] Parameterize the test, for varying message length, publishing/subscribers, publishing rate.

ivanpauno:
This are the results of the first task.
https://docs.google.com/document/d/1YRpWajYvXF_RA7nIH3Qe8RFUyVJwJoxVTSpQFJKN18U/edit?usp=sharing

To repro the results:
- Move to branch ivanpauno/rclcpp-performance (https://github.com/ros2/rclcpp/pull/636)
- In a terminal: ros2 run rclcpp_performance performance_test -d __OUTPUT_DIR__
- For reproducing the results skipping the interprocess publish, the patch in the document above should be applied.

ivanpauno:
This was used to test https://github.com/ros2/ros2/issues/649, but it's outdated. Closing!

PR #636, opened 2019-02-21
Added rclcpp_performance package

ivanpauno:
This targets the first task of https://github.com/ros2/rclcpp/issues/634

This package adds three executables:
1. Script for running some predefined tests, and then do the post-processing and plotting of all of them.
   Example usage: ros2 run  rclcpp_performance performance_test -d output_dir [--skip]
   - output_dir: Directory where the logging files will be created.
   - -s|--skip: Don't run the tests again, only post-process and plot results.

2. communication_performance: Which logs the timestamps.
   Example usage: ros2 run communication_performance [-d output_dir] [-p publish_period_ms] [-m M][-n N][-unique] [-intra]
   - output_dir: directory where the log files will be created (it should be created before running). Default: ./logs
   - -p publish_period_ms: Publish period in ms. Default: 100ms
   - -m M: Number of messages that will be published by each publisher before exiting. Default: 1000
   - -n N: Number of publisher/subscriber pairs created. Defualt: 100.
   - -unique: If option exists, the publish method taking an unique_ptr reference is used. If not, shared_ptr publish method is used.
   - -intra: Set intra process comm.

3. posprocess_logging: Plot results.
Example usage: ros2 run posprocess_logging -d output_dir
Where output_dir contains the logged file.
It plots three things:
   - The measured average publish period of each publisher.
   - The measured average publish period of each subscriber.
   - The measured latencys of each publish-subscriber pair.
In all the cases, y axes is in ms and x axes is the result of one of the parts (one publisher, one subscriber, or one publisher-subscriber pair).

My results:
https://docs.google.com/document/d/1YRpWajYvXF_RA7nIH3Qe8RFUyVJwJoxVTSpQFJKN18U/edit?usp=sharing
Old results (https://github.com/ros2/rclcpp/pull/636/commits/841e206942eed33d7396eeacdf042ecd53d3143e):
https://docs.google.com/document/d/11RBqAjcjQ4FBYkwfVtudOaHJSP_wz_TW9AGneo1_ePE/edit?usp=sharing

TODO:
- It needs a general cleaning, and add some error handling.

### DIFF
--- a/rclcpp_performance/CMakeLists.txt
+++ b/rclcpp_performance/CMakeLists.txt
@@ -39,7 +39,7 @@ ament_export_dependencies(rosidl_default_runtime)
 ament_package()
 
 install(
-  PROGRAMS scripts/posprocess_logging
+  PROGRAMS scripts/posprocess_logging scripts/performance_test
   DESTINATION lib/${PROJECT_NAME}
 )
 
--- a/rclcpp_performance/CMakeLists.txt
+++ b/rclcpp_performance/CMakeLists.txt
@@ -1,0 +1,49 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(rclcpp_performance)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+find_package(ament_cmake REQUIRED)
+# find_package(ament_cmake_python REQUIRED)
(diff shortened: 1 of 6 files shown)


dirk-thomas:
Would this new package be better places into a different / separate repository? I am not sure that tying it to the release of `rcl*` is necessary. Also how does this related to the already ongoing effort of performance benchmarking?

ivanpauno:
> Would this new package be better places into a different / separate repository? I am not sure that tying it to the release of `rcl*` is necessary. Also how does this related to the already ongoing effort of performance benchmarking?

It can be moved to other place later. I've added now an script with the test I made. This is a simple measuring of publish/subscription rate and of latency. It would be good to continue the analysis with a performance tool like gperftools.

ivanpauno:
Closing as this is not up to date, and there have been better proposals for measuring performance:
- https://github.com/ApexAI/performance_test
- https://github.com/irobot-ros/ros2-performance/.

PR #638, opened 2019-02-26
Set symbol visibility to hidden for rclcpp

bhatsach:
Enabling symbol visibility feature in gcc and clang compilers.
This will hep find symbol export related issues in linux and
potentially reduce compile times.

Discourse topic link:
https://discourse.ros.org/t/set-symbol-visibility-to-hidden-for-rmw-and-rcl-packages/7981

Signed-off-by: Sachin Suresh Bhat <bhatsach@amazon.com>

Connects to ros2/rcl#391

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -109,6 +109,7 @@ ament_target_dependencies(${PROJECT_NAME}
   "rosidl_typesupport_cpp"
   "rosidl_generator_cpp")
 
+configure_rcl(${PROJECT_NAME})
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(${PROJECT_NAME}
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -23,6 +23,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 include_directories(include)
+include(cmake/configure_rclcpp.cmake)
 
 set(${PROJECT_NAME}_SRCS
   src/rclcpp/any_executable.cpp
@@ -109,7 +110,7 @@ ament_target_dependencies(${PROJECT_NAME}
   "rosidl_typesupport_cpp"
   "rosidl_generator_cpp")
 
-configure_rcl(${PROJECT_NAME})
+configure_rclcpp(${PROJECT_NAME})
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
 target_compile_definitions(${PROJECT_NAME}
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -54,27 +54,21 @@ class RCLCPP_PUBLIC ClientBase
 public:
   RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(ClientBase)
 
-  RCLCPP_PUBLIC
   ClientBase(
     rclcpp::node_interfaces::NodeBaseInterface * node_base,
     rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph);
 
-  RCLCPP_PUBLIC
   virtual ~ClientBase();
 
-  RCLCPP_PUBLIC
   const char *
   get_service_name() const;
 
-  RCLCPP_PUBLIC
   std::shared_ptr<rcl_client_t>
   get_client_handle();
 
-  RCLCPP_PUBLIC
   std::shared_ptr<const rcl_client_t>
   get_client_handle() const;
 
-  RCLCPP_PUBLIC
   bool
   service_is_ready() const;
 
@@ -96,15 +90,12 @@ public:
 protected:
   RCLCPP_DISABLE_COPY(ClientBase)
 
-  RCLCPP_PUBLIC
   bool
   wait_for_service_nanoseconds(std::chrono::nanoseconds timeout);
 
-  RCLCPP_PUBLIC
   rcl_node_t *
   get_rcl_node_handle();
 
-  RCLCPP_PUBLIC
   const rcl_node_t *
   get_rcl_node_handle() const;
 
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -49,7 +49,7 @@ namespace node_interfaces
 class NodeBaseInterface;
 }  // namespace node_interfaces
 
-class ClientBase
+class RCLCPP_PUBLIC ClientBase
 {
 public:
   RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(ClientBase)
--- a/rclcpp/cmake/configure_rclcpp.cmake
+++ b/rclcpp/cmake/configure_rclcpp.cmake
@@ -1,0 +1,66 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Configures ros client library with custom settings.
+# The custom settings are all related to library symbol visibility, see:
+# https://gcc.gnu.org/wiki/Visibility
+# http://www.ibm.com/developerworks/aix/library/au-aix-symbol-visibility/
+#
+# Below code is heavily referenced from a similar functionality in rmw:
+# https://github.com/ros2/rmw/blob/master/rmw/cmake/configure_rmw_library.cmake
+#
+# :param library_target: the library target
+# :type library_target: string
+# :param LANGUAGE: Optional flag for the language of the library.
+#   Allowed values are "C" and "CXX". The default is "CXX".
+# :type LANGUAGE: string
+#
+# @public
+#
+macro(configure_rclcpp library_target)
+  cmake_parse_arguments(_ARG "" "LANGUAGE" "" ${ARGN})
+  if(_ARG_UNPARSED_ARGUMENTS)
+    message(FATAL_ERROR "configure_rclcpp() called with unused arguments: ${_ARG_UNPARSED_ARGUMENTS}")
+  endif()
+
+  if(NOT _ARG_LANGUAGE)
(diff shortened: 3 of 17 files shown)


thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6272/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2697/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5154/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6113/)

rclcpp/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ ament_target_dependencies(${PROJECT_NAME}
   "rosidl_typesupport_cpp"
   "rosidl_generator_cpp")
 
+configure_rcl(${PROJECT_NAME})
```
thomas-moulard: Needs to be rclcpp?

bhatsach: I am trying to reuse the **configure_rcl** macro in ros2/rcl. This is in a way similar to how the [configure_rmw_library](https://github.com/ros2/rmw/blob/master/rmw/cmake/configure_rmw_library.cmake) macro is being reused in [rmw_fastrtps](https://github.com/ros2/rmw_fastrtps/blob/328771c18382bd053d4a6eb34992914df40fe3c4/rmw_fastrtps_cpp/CMakeLists.txt#L99) and other packages. The single macro takes care of both C and C++ code

thomas-moulard: This is not possible because they are not part of the same shared library. See doc here: https://gcc.gnu.org/wiki/Visibility

traversaro: A possible alternative may be to use the CMake standard macro [`generate_export_header`](https://cmake.org/cmake/help/v3.5/module/GenerateExportHeader.html) and the [`DEFINE_SYMBOL`](https://cmake.org/cmake/help/v3.5/prop_tgt/DEFINE_SYMBOL.html) target property.

See https://github.com/ros/console_bridge/pull/43 for a use in a ROS-related repo, and https://bitbucket.org/osrf/gazebo/issues/2262/symbol-visibility-issues-in-windows for a related issue in Gazebo.

bhatsach: Quick note: This is the way in which we are sharing the "configure_rcl" macro from rcl to rclcpp: https://github.com/ros2/rmw/blob/master/rmw/CMakeLists.txt#L62-L65
I am not sure if sharing just the macro is allowing librcl_cpp.so to access librcl.so private symbols.

traversaro: The critical part is in https://github.com/ros2/rcl/pull/391/files#diff-19c13a4e66284bae327c94111b226946R71, in which you are defining the `RCL_BUILDING_DLL` macro also for compilation units that are not part of the `rcl`. The effect will be that when including https://github.com/ros2/rcl/blob/aa4ac7d06cc5e59d693408260868edb004154b69/rcl/include/rcl/visibility_control.h as part of a compilation unit **outside of rcl**, `RCL_PUBLIC` (in Windows) will have the value `__declspec(dllexport)` instead of the correct `__declspec(dllimport)`. This is exactly the problem described in https://bitbucket.org/osrf/gazebo/issues/2262/symbol-visibility-issues-in-windows .

wjwwood: > A possible alternative may be to use the CMake standard macro [`generate_export_header`](https://cmake.org/cmake/help/v3.5/module/GenerateExportHeader.html) and the [`DEFINE_SYMBOL`](https://cmake.org/cmake/help/v3.5/prop_tgt/DEFINE_SYMBOL.html) target property.

I personally don't like the generated code approach (I prefer a single, checked-in boil-plate file) and a custom define.

> I am trying to reuse the **configure_rcl** macro in ros2/rcl. This is in a way similar to how the [configure_rmw_library](https://github.com/ros2/rmw/blob/master/rmw/cmake/configure_rmw_library.cmake) macro is being reused in [rmw_fastrtps](https://github.com/ros2/rmw_fastrtps/blob/328771c18382bd053d4a6eb34992914df40fe3c4/rmw_fastrtps_cpp/CMakeLists.txt#L99) and other packages. The single macro takes care of both C and C++ code

As the others said, you can do this, but only for the visibility compiler options, not the visibility compiler definition.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -49,7 +49,7 @@ namespace node_interfaces
 class NodeBaseInterface;
 }  // namespace node_interfaces
 
-class ClientBase
+class RCLCPP_PUBLIC ClientBase
```
wjwwood: Can you explain why? We explicitly avoided this, choosing to make each method public manually. We've had issues making these public on Windows due to the std data structures which are private members of the class.

rclcpp/CMakeLists.txt
bhatsach: Thanks for the inputs @thomas-moulard , @traversaro and @wjwwood. I have removed the target definition specific code in [configure_rcl](https://github.com/ros2/rcl/pull/391/files#diff-19c13a4e66284bae327c94111b226946R31) in https://github.com/ros2/rcl/pull/391 . This is because each rcl or rclcpp has its own target definition specified in the CMakeList. For eg:
https://github.com/ros2/rclcpp/blob/master/rclcpp/CMakeLists.txt#L112-L115
https://github.com/ros2/rclcpp/blob/master/rclcpp_action/CMakeLists.txt#L38-L41

Is this a reasonable solution? Do you think we should have a separate configure_rclcpp for rclcpp packages?

rclcpp/include/rclcpp/client.hpp
bhatsach: Here's a sample error I ran into while running rclcpp package with visibility set to hidden. Is there any other way to handle this instead of making the class visible?

```
collect2: error: ld returned 1 exit status
make[2]: *** [test_service] Error 1
make[1]: *** [CMakeFiles/test_service.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
CMakeFiles/test_publisher.dir/test/test_publisher.cpp.o: In function `std::shared_ptr<rclcpp::Publisher<rcl_interfaces::msg::IntraProcessMessage_<std::allocator<void> >, std::allocator<void> > > std::dynamic_pointer_cast<rclcpp::Publisher<rcl_interfaces::msg::IntraProcessMessage_<std::allocator<void> >, std::allocator<void> >, rclcpp::PublisherBase>(std::shared_ptr<rclcpp::PublisherBase> const&)':
test_publisher.cpp:(.text._ZSt20dynamic_pointer_castIN6rclcpp9PublisherIN14rcl_interfaces3msg20IntraProcessMessage_ISaIvEEES5_EENS0_13PublisherBaseEESt10shared_ptrIT_ERKS9_IT0_E[_ZSt20dynamic_pointer_castIN6rclcpp9PublisherIN14rcl_interfaces3msg20IntraProcessMessage_ISaIvEEES5_EENS0_13PublisherBaseEESt10shared_ptrIT_ERKS9_IT0_E]+0x30): undefined reference to `typeinfo for rclcpp::PublisherBase'
CMakeFiles/test_publisher.dir/test/test_publisher.cpp.o:(.data.rel.ro._ZTIN6rclcpp9PublisherIN14rcl_interfaces3msg20IntraProcessMessage_ISaIvEEES4_EE[_ZTIN6rclcpp9PublisherIN14rcl_interfaces3msg20IntraProcessMessage_ISaIvEEES4_EE]+0x10): undefined reference to `typeinfo for rclcpp::PublisherBase'
collect2: error: ld returned 1 exit status
make[2]: *** [test_publisher] Error 1
make[1]: *** [CMakeFiles/test_publisher.dir/all] Error 2
CMakeFiles/test_externally_defined_services.dir/test/test_externally_defined_services.cpp.o:(.data.rel.ro._ZTIN6rclcpp7ServiceINS_3srv4MockEEE[_ZTIN6rclcpp7ServiceINS_3srv4MockEEE]+0x10): undefined reference to `typeinfo for rclcpp::ServiceBase'
collect2: error: ld returned 1 exit status
make[2]: *** [test_externally_defined_services] Error 1
make[1]: *** [CMakeFiles/test_externally_defined_services.dir/all] Error 2
CMakeFiles/test_client.dir/test/test_client.cpp.o:(.data.rel.ro._ZTIN6rclcpp6ClientIN14rcl_interfaces3srv14ListParametersEEE[_ZTIN6rclcpp6ClientIN14rcl_interfaces3srv14ListParametersEEE]+0x10): undefined reference to `typeinfo for rclcpp::ClientBase'
```

rclcpp/CMakeLists.txt
traversaro: > I personally don't like the generated code approach (I prefer a single, checked-in boil-plate file) and a custom define.

This is probably a bit OT in this PR, but for the sake of clarity: the generated header file (both CMake's and the example linked by @thomas-moulard ) is useful for supporting the case in which a library is compiled as a static library under Windows. Without that, in any downstream compilation unit `RCLCPP_PUBLIC` is always defined as `__declspec(dllimport)` and this annotation will generate a Visual Studio error LNK2019 if the RCLCPP library is a actually compiled as static. A possible alternative **without** having a generated file is to have an additional logic that defines `RCLCPP_PUBLIC` as an empty macro if `RCLCPP_STATIC` is defined, but then it is necessary some kind of build system logic to make sure that `RCLCPP_STATIC` is always defined (even downstream) if RCLCPP is build as static, for example a conditional `target_compile_definitions(rclcpp PUBLIC RCLCPP_STATIC)` called only if the library is built is static, or similar.

See https://github.com/ros/console_bridge/pull/40#issuecomment-269460721 for a similar comment.

rclcpp/include/rclcpp/client.hpp
wjwwood: I think adding it to the class may be required, but you need to test it on Windows too, because I think it might be in conflict.

If you do leave the public in the class declaration, then the `RCLCPP_PUBLIC` for each method in the class is redundant and should be removed.

bhatsach: I removed all the inner RCLCPP_PUBLIC keywords from the modified classes. Will request @thomas-moulard to run this on windows CI for a quick check

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6281/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2702/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5159/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6118/)

wjwwood:
This is what I was referring to (dll warnings from the linker):

https://ci.ros2.org/job/ci_windows/6118/warnings43Result/new/

bhatsach:
> This is what I was referring to (dll warnings from the linker):
> 
> https://ci.ros2.org/job/ci_windows/6118/warnings43Result/new/

:( let me look into this tomorrow. @wjwwood, can you help close out the rcl related changes if they are good to go? https://github.com/ros2/rcl/pull/391

bhatsach:
Discarding this PR for now as we need further investigation before getting back onto this task:
* Investigate the above windows warnings and identify ways to resolve the same
* Investigate the above os x test failures and identify ways to resolve the same
* Identify a solution which works well on all the platforms. One of the main objectives here is to identify dllexport issues earlier.

PR #639, opened 2019-02-27
Fix test_time_source test

pbaughman:
We had a lot of trouble with this test recently.  The way to repro the issue was to do something like:

```
terminal 1>colcon build --packages-up-to rclcpp
terminal 1>while ./build/rclcpp/test_time_source ; do :; done
```
If you let it run long enough, eventually you'd see it fail.  To make it fail faster, do
```
terminal 2> stress -c 12  # Where 12 is the number of cores on my computer
terminal 3> stress -m 12
```
There were a few different failures that would sometimes happen:
Failure 1 (more common):
```
27: rclcpp/rclcpp/test/test_time_source.cpp:334: Failure   
27: Value of: ros_clock->ros_time_is_active()                                                                         
27:   Actual: false                                                                                                   
27: Expected: true      
```

Failure 2 (less common):
```
27: [ RUN      ] TestTimeSource.callbacks                                                                             
27: rclcpp/rclcpp/test/test_time_source.cpp:249: Failure   
27: Value of: ros_clock->ros_time_is_active()                                                                         
27:   Actual: false                                                                                                   
27: Expected: true                                                                                                    
27: rclcpp/rclcpp/test/test_time_source.cpp:255: Failure   
27: Expected: (0L) != (t_out.nanoseconds()), actual: 0 vs 0                                                           
27: rclcpp/rclcpp/test/test_time_source.cpp:256: Failure   
27: Expected: (t_low.nanoseconds()) < (t_out.nanoseconds()), actual: 1000000000 vs 0                                  
```

## Cause of problem(s)

The call to spin_once [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/test_time_source.cpp#L76) is meant to run the subscriber callbacks of the ros_clock's subscription to the "clock" topic, but that's far from guaranteed.  If you printed out the simulated time at the end of the test, you'd expect it to be 4s + 1000ns, but the majority of the time it was 1s + 1000ns.  Sometimes it was 0s + 1000ns, and sometimes it was 0s + 0ns.

Also, the call to spin_some [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/test_time_source.cpp#L96) is largely ineffectual.  Using GDB, I was able to tell that almost 100% of the time, this spin_some returned without processing any executable actions from the node's queue.

## Fixes
@wjwwood has some ideas about infrastructure changes to make this test easier to write.  Short term, I'm working around the issue by having the test spin until it sees the action it's waiting for instead of spinning once and hoping the action it's waiting for is first in the queue of things to be processed

### DIFF
--- a/rclcpp/test/test_time_source.cpp
+++ b/rclcpp/test/test_time_source.cpp
@@ -55,16 +55,85 @@ protected:
   rclcpp::Node::SharedPtr node;
 };
 
-void trigger_clock_changes(
-  rclcpp::Node::SharedPtr node)
+void spin_until_time(
+    rclcpp::Clock::SharedPtr clock,
+    rclcpp::Node::SharedPtr node,
+    std::chrono::nanoseconds end_time,
+    bool expect_time_update)
 {
-  auto clock_pub = node->create_publisher<rosgraph_msgs::msg::Clock>("clock",
-      rmw_qos_profile_default);
+  // Call spin_once on the node until either:
+  // 1) We see the ros_clock's simulated time change to the expected end_time
+  // -or-
+  // 2) 1 second has elapsed in the real world
+  // If 'expect_time_update' is True, and we timed out waiting for simulated time to
+  // update, we'll have the test fail
 
   rclcpp::executors::SingleThreadedExecutor executor;
   executor.add_node(node);
 
-  rclcpp::WallRate loop_rate(50);
+  auto start = std::chrono::system_clock::now();
+  while (std::chrono::system_clock::now() < (start + 1s))
+  {
+    if (!rclcpp::ok()) {
+      break;  // Break for ctrl-c
+    }
+
+    executor.spin_once(10ms);
+
+    if (clock->now().nanoseconds() == end_time.count()) {
+        return;
+    }
+  }
+
+  if (expect_time_update) {
+    // If we were expecting ROS clock->now to be updated and we didn't take the early return from
+    // the loop up above, that's a failure
+    ASSERT_TRUE(false) << "Timed out waiting for ROS time to update";
+  }
+}
+
+void spin_until_ros_time_updated(
+    rclcpp::Clock::SharedPtr clock,
+    rclcpp::Node::SharedPtr node,
+    rclcpp::ParameterValue value)
+{
+  // Similar to above:  Call spin_once until we see the clock's ros_time_is_active method
+  // match the ParameterValue
+  // Unlike spin_until_time, there aren't any test cases where we don't expect the value to
+  // update.  In the event that the ParameterValue is not set, we'll pump messages for a full second
+  // but we don't cause the test to fail
+
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(node);
+
+  auto start = std::chrono::system_clock::now();
+  while (std::chrono::system_clock::now() < (start + 1s))
+  {
+    if (!rclcpp::ok()) {
+      break;  // Break for ctrl-c
+    }
+
+    executor.spin_once(10ms);
+
+    // In the case where we didn't intend to change the parameter, we'll still pump
+    if (value.get_type() == rclcpp::ParameterType::PARAMETER_NOT_SET) {
+      continue;
+    }
+
+    if (clock->ros_time_is_active() == value.get<bool>()) {
+      return;
+    }
+  }
+}
+
+void trigger_clock_changes(
+  rclcpp::Node::SharedPtr node,
+  std::shared_ptr<rclcpp::Clock> clock,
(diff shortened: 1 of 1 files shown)


rclcpp/test/test_time_source.cpp
```diff
@@ -55,16 +55,85 @@ class TestTimeSource : public ::testing::Test
...
+  // Similar to above:  Call spin_once until we see the clock's ros_time_is_active method
+  // match the ParameterValue
+  // Unlike spin_until_time, there aren't any test cases where we don't expect the value to
+  // update.  In the event that the ParameterValue is not set, we'll pump messages for a full second
+  // but we don't cause the test to fail
 
   rclcpp::executors::SingleThreadedExecutor executor;
   executor.add_node(node);
 
-  rclcpp::WallRate loop_rate(50);
```
pbaughman: This loop_rate didn't do much to help the situation.  The problem isn't that we need more wall time for the test to work.  We need more calls to 'spin_once'

rclcpp/test/test_time_source.cpp
```diff
@@ -73,12 +142,20 @@ void trigger_clock_changes(
     msg->clock.sec = i;
     msg->clock.nanosec = 1000;
     clock_pub->publish(msg);
-    executor.spin_once(1000000ns);
```
pbaughman: Another way to make the test work better was to call executor.spin_once ten times in a row here.  Then you'd start seeing the expected time at the end of a test run

rclcpp/test/test_time_source.cpp
```diff
@@ -371,7 +452,6 @@ TEST_F(TestTimeSource, callback_handler_erasure) {
   EXPECT_GT(t_high.nanoseconds(), t_out.nanoseconds());
 }
```
pbaughman: If you printed out the value of t_out right here, you'd see that it was usually 1s + 1000ns.  That's good enough to be between t_low and t_high, but it's not what you'd expect if everything was working correctly.  It should be 4s + 1000ns at the end of the test

rclcpp/test/test_time_source.cpp
```diff
@@ -55,16 +55,85 @@ class TestTimeSource : public ::testing::Test
...
+
+void trigger_clock_changes(
+  rclcpp::Node::SharedPtr node,
+  std::shared_ptr<rclcpp::Clock> clock,
+  bool expect_time_update=true)
+{
+  auto clock_pub = node->create_publisher<rosgraph_msgs::msg::Clock>("clock",
+      rmw_qos_profile_default);
+
   for (int i = 0; i < 5; ++i) {
```
pbaughman: I'm not sure why this test is doing multiple publishes - most of the intermediate clock values (0, 1, 2, 3) are thrown away by the tests.  Maybe it's another attempt at working around executor.spin_once not handling the action you think it's handling

rclcpp/test/test_time_source.cpp
```diff
@@ -55,16 +55,85 @@ class TestTimeSource : public ::testing::Test
...
+  auto start = std::chrono::system_clock::now();
+  while (std::chrono::system_clock::now() < (start + 1s))
+  {
+    if (!rclcpp::ok()) {
+      break;  // Break for ctrl-c
+    }
+
+    executor.spin_once(10ms);
+
+    if (clock->now().nanoseconds() == end_time.count()) {
```
pbaughman: It's important to do an exact match here.  If you do `clock->now().nanoseconds() >= end_time.count()` then we'll trivially return whenever we're in the part of the test that runs before simulated time is active because clock->now().nanoseconds is a very large number before simulated time is turned on

hidmic:
Running CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6325/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2739/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5207/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6175/)

hidmic:
@pbaughman looks like there are lint issues. Could you verify that your PR passes all tests locally?

pbaughman:
@hidmic Sorry about that - I keep forgetting that our tools are slightly behind your tools.  Re-ran tests in the nightly docker image and fixed things up

hidmic:
@pbaughman no worries, I've relaunched CI, let's see what it has to say.

hidmic:
Looks like the test instability comes a cppcheck issue that was solved by https://github.com/ros2/rclcpp/pull/646.

hidmic:
I'll merge, thanks for the contribution!

PR #655, opened 2019-03-08
Add Deadline, Liveliness, and Lifespan QoS Event Callbacks and Waitables

mm318:
Implementation for Deadline, Liveliness, and Lifespan QoS. The design is at ros2/design#212

This change adds ability to register application callbacks for QoS events (interface with the user application) and the ability to wait on QoS events in `rcl_wait_set_t` (interface with `rcl`).


rclcpp/src/rclcpp/client.cpp
```diff
@@ -88,6 +102,86 @@ ClientBase::get_client_handle() const
...
+    } else if (status != RCL_RET_CLIENT_TAKE_FAILED) {
+      RCUTILS_LOG_ERROR_NAMED(
+        "rclcpp",
+        "take response failed for client of service '%s': %s",
+        get_service_name(), rcl_get_error_string().str);
+      rcl_reset_error();
+    }
+  }
+
+  if (event_ready_) {
```
mm318: We are working separating the execute() function so that each executable in the executor will only take one thing and make only one callback.

(Currently, the way this is setup means it is taking one message, making the callback, taking one event, and making that event callback.)

rclcpp/include/rclcpp/create_publisher.hpp
```diff
@@ -25,12 +25,14 @@
...
 
-template<typename MessageT, typename AllocatorT, typename PublisherT>
+template<typename MessageT, typename EventCallbackT, typename AllocatorT, typename PublisherT>
 std::shared_ptr<PublisherT>
 create_publisher(
   rclcpp::node_interfaces::NodeTopicsInterface * node_topics,
   const std::string & topic_name,
   const rmw_qos_profile_t & qos_profile,
+  EventCallbackT && event_callback,
+  rclcpp::callback_group::CallbackGroup::SharedPtr group,
```
wjwwood: This might be a good time to implement something like the `rclcpp::NodeOptions` class, but for publishers, subscriptions, etc... It will be especially helpful when/if we allow for one callback per type of event.

rclcpp/include/rclcpp/event.hpp
```diff
@@ -17,20 +17,31 @@
...
+{
+  DEADLINE_MISSED,
+  LIVELINESS_CHANGED,
+  LIFESPAN_EXPIRED
+};
+
+using ResourceStatusEventCallbackType = std::function<void(ResourceStatusEvent)>;
+
+
+class GraphEvent
```
wjwwood: Why rename this? It is actually a generic event class, it doesn't have anything particular about it that is unique to graph events.

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -79,7 +79,7 @@ class MultiThreadedExecutor : public executor::Executor
   bool yield_before_execute_;
 
   std::mutex scheduled_timers_mutex_;
-  std::set<TimerBase::SharedPtr> scheduled_timers_;
+  std::set<Waitable::SharedPtr> scheduled_timers_;
```
wjwwood: Why change this? They may be `Waitable`, but `TimerBase` is more accurate, e.g. you don't want a subscription to get added here accidentally, I think.

wjwwood:
With respect to changing the various classes to the `Waitable` pattern. I think that's a good improvement, though I'd have to more thuroughly review the impact on removing them from the memory strategy (that was supposed to be the means by which someone could control memory allocations related to the wait set), but in general it's the direction we want to go. That being said, I'm not 100% it's required for this work, but perhaps I'm wrong. In general it goes better if we keep pr's focused on only what is required, making other improvements in separate pr's.

rclcpp/include/rclcpp/event.hpp
mm318: Ah sorry, it looked like it was only ever used for node graph events (and it was otherwise only mentioned in the comments of `exceptions.hpp`). I guess sometime in the future, it will be used for something else. We'll revert this name change.

wjwwood: Perhaps it's purpose/use should be combined with the purpose/use of `rcl_event_t`, I don't know, but in the case a name collision still happens, we could still consider renaming this. I just didn't see another `rclcpp::Event` yet.

Also if we rename it, we should move it to its own file, would have been `graph_event.hpp`. We try (but often fail) to live up to the "one class per file" ideal, where it makes sense of course.

mm318:
> I'm not 100% it's required for this work

We believe it was cleanest design for `PublisherBase` and `SubscriptionBase` to hold their own respective `rcl_event_t` for QoS events, and that means they become very well-suited for the `Waitable` interface and its `get_number_of_ready_subscriptions()` and `get_number_of_ready_events()` member functions. And using the `add_to_wait_set()` member function can leave the `AllocatorMemoryStrategy` implementation clean.

We can probably decouple the use of the `Waitable` and the removal of the vectors from `AllocatorMemoryStrategy` though. I am curious to know how these separate vectors help control memory allocations versus having these vectors collapse into a single one of `Waitable`s.

wjwwood:
> I am curious to know how these separate vectors help control memory allocations versus having these vectors collapse into a single one of Waitables.

They may not, but I would need to look deeper to know for sure. It's possible that you could use something like `std::array` instead of `std::vector` to store them in an alternative implementation of the memory strategy, but again I'd need to verse myself in it again to nail that down. It might be that we can accomplish the same thing with the Waitable pattern as well. I will say the memory strategy makes the code very hard to follow and I've wanted to try and clean that up for a while to keep the benefit but make it cleaner.

gbiggs:
I'm all in favour of cleaning up the memory strategy stuff, but please don't remove or ignore it without an alternative. We need to be able to control memory allocation explicitly.

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
mm318: This is similar to collapsing the vectors in `AllocatorMemoryStrategy` specific to `SubscriptionBase`, `ClientBase`, `ServiceBase`, etc. to one vector of `Waitable`s, `AnyExecutable` effectively can only contain `Waitable`s, no `TimerBase`s. `AnyExecutable::has_timer()` is also checked before being added to this set.

Also, shouldn't this set be able to include a `Waitable` that has a timer, even if the `Waitable` is not a `TimerBase`?

mm318:
Hi @wjwwood, can you take a look again? I've introduced the `SubscriptionOptions` and `PublisherOptions` classes.

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
nburek: I think the argument here is that if the name is `scheduled_timers_` the expectation is that it should only contain `TimerBase` objects, thus keeping the previous type to enforce that expectation. If the context of how this set used is changing to include all types of Waitables then the name should change to reflect the new context, if not then it should be fine keeping it the more concrete type.

nburek:
> I'm not 100% it's required for this work

+1 for moving the Waitable refactor into its own PR instead of coupling it with the QoS changes. Though I also agree that we still want that refactor for the QoS work.

rclcpp/include/rclcpp/create_publisher.hpp
mm318: We will have a separate PR for introducing `rclcpp::SubscriptionOptions` and `rclcpp::PublisherOptions`.

rclcpp/include/rclcpp/event.hpp
mm318: We will back out this change from this PR.

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
mm318: This change will be omitted from this PR.

mm318:
Closing this pull request as we try to get these QoS features in piecemeal.

PR #656, opened 2019-03-14
fix wait for service memory leak bug

reed-lau:
* when call wait for service in an while loop, the event will be make forever and never release
* fix it by: creating it when we need

### DIFF
--- a/rclcpp/src/rclcpp/client.cpp
+++ b/rclcpp/src/rclcpp/client.cpp
@@ -118,7 +118,6 @@ ClientBase::wait_for_service_nanoseconds(std::chrono::nanoseconds timeout)
   if (!node_ptr) {
     throw InvalidNodeError();
   }
-  auto event = node_ptr->get_graph_event();
   // check to see if the server is ready immediately
   if (this->service_is_ready()) {
     return true;
@@ -127,6 +126,7 @@ ClientBase::wait_for_service_nanoseconds(std::chrono::nanoseconds timeout)
     // check was non-blocking, return immediately
     return false;
   }
+  auto event = node_ptr->get_graph_event();
   // update the time even on the first loop to account for time spent in the first call
   // to this->server_is_ready()
   std::chrono::nanoseconds time_to_wait =


dirk-thomas:
Please provide more information about the exact steps you have tested this patch with.

From a first look the lifetime of the created event seems to be managed by smart pointers and some time after the `event` variable goes out of scope it will be deleted from the the internal list of `graph_events_`.

reed-lau:
Reproduce step:  
1. change the offical add_two_int_client.cpp code as following.
2. compile it using `colcon build`
3. `ros2 run demo_nodes_cpp add_two_ints_client` in one terminal
4. `ros2 run demo_nodes_cpp add_two_ints_server` in another terminal
then the add_two_ints_client node's memory leakage happens. 

Explanation:
the offical implementation using shared_ptr to manager `graph_events_`, but even when `wait_for_service` is ok, one new event will be make and be pushed into a vector and then return. shared_ptr's lifecycle in vector will last until someone pop it, but the `clear` operation is after the `return true`.  
https://github.com/ros2/rclcpp/blob/1869b84a0ca60f758f63f6374b641a5bd93c3fca/rclcpp/src/rclcpp/client.cpp#L115
```c
diff --git a/demo_nodes_cpp/src/services/add_two_ints_client.cpp b/demo_nodes_cpp/src/services/add_two_ints_client.cpp
index f92670b..2510075 100644
--- a/demo_nodes_cpp/src/services/add_two_ints_client.cpp
+++ b/demo_nodes_cpp/src/services/add_two_ints_client.cpp
@@ -76,12 +76,15 @@ int main(int argc, char ** argv)
   request->a = 2;
   request->b = 3;
 
-  while (!client->wait_for_service(1s)) {
-    if (!rclcpp::ok()) {
-      RCLCPP_ERROR(node->get_logger(), "Interrupted while waiting for the service. Exiting.")
-      return 0;
-    }
-    RCLCPP_INFO(node->get_logger(), "service not available, waiting again...")
+  while(1) {
+      while (!client->wait_for_service(1s)) {
+          if (!rclcpp::ok()) {
+              RCLCPP_ERROR(node->get_logger(), "Interrupted while waiting for the service. Exiting.")
+                  return 0;
+          }
+          RCLCPP_INFO(node->get_logger(), "service not available, waiting again...")
+      }
+      printf("service \n");
   }
```

dirk-thomas:
Thanks for the explanation. I understand your use case now.

[Build Status](https://ci.ros2.org/job/ci_linux/6423/)

reed-lau:
will this patch backport to bouncy-branch? as we're still using bouncy, compiling for source.

PR #657, opened 2019-03-14
fix wait for service memory leak

reed-lau:
when call wait_for_service in a loop, the get graph event will leak memory.
we fix it by, when service is ok, there's no need to create a new one.


dirk-thomas:
Closing in favor of #656 which applies the same patch to the default branch. If necessary it can be considered for backporting after it has been accepted on the default branch.

@nuclearsandwich FYI

nuclearsandwich:
> @nuclearsandwich FYI

https://github.com/ros2/ros2/issues/575 is the tracking issue for the next potential Bouncy patch release.

PR #664, opened 2019-03-21
Check QoS policy when configuring intraprocess, skip interprocess publish when possible

ivanpauno:
This is part of the intraprocess comm improvement (https://github.com/ros2/ros2/issues/649, https://docs.google.com/document/d/1dM8O34GM-SbFps1vd1rocPMAg_7O9_-OG4NF4axOz7E/edit#)

As the current and proposed IPM doesn't complain with a non-volatile 'Durability' QoS policy, intraprocess comm shouldn't be configured in those cases. 
- Advantage: Solves compliance issues with QoS policies.
  e.g.: If a publisher and subscriber in the same process had been configured with QoS policy  
  'durability=transient_local" and "history=keep_last", a late subscriber would receive the last message 
  published, as this was not supported by the intra process manager.
- Disadvantage: Intraprocess comm is not supported in those cases.

Previously, interprocess publish was always done. Now, it is skipped when only having intraprocess subscriptions.
- Advantage: Less publish-to-subscribe latency, cpu usage, memory usage when the topic is only used for intraprocess comm.
- Disadvantage: Probably, the time it takes to recognize a interprocess subscription done just before the publishing is longer than before. 
   e.g.:    Process B subscribes to topic /topic.
              Process A publishes to topic /topic just after process B subscriber.
              Since maybe the count updating time it's not the same as how the publish method checks its subscribers, a message is not received when before it was.  
           (This hasn't been checked, it's only a possible drawback of this change)

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -228,7 +228,9 @@ public:
   virtual void
   publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
   {
-    this->do_inter_process_publish(msg.get());
+    if (!use_intra_process_ || get_subscription_count() > get_intra_process_subscription_count()) {
+      this->do_inter_process_publish(msg.get());
+    }
     if (store_intra_process_message_) {
       // Take the pointer from the unique_msg, release it and pass as a void *
       // to the ipm. The ipm should then capture it again as a unique_ptr of
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -181,7 +181,7 @@ protected:
 
   using IntraProcessManagerWeakPtr =
     std::weak_ptr<rclcpp::intra_process_manager::IntraProcessManager>;
-  bool use_intra_process_;
+  bool intra_process_is_enabled_;
   IntraProcessManagerWeakPtr weak_ipm_;
   uint64_t intra_process_publisher_id_;
   StoreMessageCallbackT store_intra_process_message_;
@@ -228,7 +228,9 @@ public:
   virtual void
   publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
   {
-    if (!use_intra_process_ || get_subscription_count() > get_intra_process_subscription_count()) {
+    bool inter_process_subscriptions_exist =
+      get_subscription_count() > get_intra_process_subscription_count();
+    if (!intra_process_is_enabled_ || inter_process_subscriptions_exist) {
       this->do_inter_process_publish(msg.get());
     }
     if (store_intra_process_message_) {
--- a/rclcpp/src/rclcpp/publisher.cpp
+++ b/rclcpp/src/rclcpp/publisher.cpp
@@ -254,7 +254,7 @@ PublisherBase::setup_intra_process(
   if (intra_process_options.qos.durability != RMW_QOS_POLICY_DURABILITY_VOLATILE) {
     RCLCPP_WARN(
       rclcpp::get_logger("rclcpp"),
-      "Enabled intraprocess communication with incompatible QoS policy.");
+      "Skipping intraprocess communication with incompatible QoS policy.");
     return;
   }
   const char * topic_name = this->get_topic_name();
--- a/rclcpp/src/rclcpp/publisher.cpp
+++ b/rclcpp/src/rclcpp/publisher.cpp
@@ -244,18 +244,12 @@ PublisherBase::setup_intra_process(
   IntraProcessManagerSharedPtr ipm,
   const rcl_publisher_options_t & intra_process_options)
 {
-  // Skip intraprocess configuration if the "durability" qos policy is not "volatile".
-  // TODO(ivanpauno):
-  //    RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT is for FastRTPS and Connext also
-  //    equal to "volatile". It would be good to add a
-  //    is_durability_policy_volatile(rcl_publisher_options_t options)
-  //    method to rcl, and corresponging methods in rmw and the rmw vendor
-  //    implementations, which will also checks if "system_default" is "volatile" or not.
-  if (intra_process_options.qos.durability != RMW_QOS_POLICY_DURABILITY_VOLATILE) {
-    RCLCPP_WARN(
-      rclcpp::get_logger("rclcpp"),
-      "Skipping intraprocess communication with incompatible QoS policy.");
-    return;
+  // Intraprocess configuration is only available with "volatile"
+  // durability qos policy.
+  if (this->get_actual_qos().durability != RMW_QOS_POLICY_DURABILITY_VOLATILE) {
+    throw std::runtime_error(
+            "Intraprocess communication is not allowed"
+            " with any qos policy different than volatile.");
   }
   const char * topic_name = this->get_topic_name();
   if (!topic_name) {
--- a/rclcpp/src/rclcpp/publisher.cpp
+++ b/rclcpp/src/rclcpp/publisher.cpp
@@ -43,7 +43,7 @@ PublisherBase::PublisherBase(
   const rosidl_message_type_support_t & type_support,
   const rcl_publisher_options_t & publisher_options)
 : rcl_node_handle_(node_base->get_shared_rcl_node_handle()),
-  use_intra_process_(false), intra_process_publisher_id_(0),
+  intra_process_is_enabled_(false), intra_process_publisher_id_(0),
   store_intra_process_message_(nullptr)
 {
   rcl_ret_t ret = rcl_publisher_init(
@@ -99,7 +99,7 @@ PublisherBase::~PublisherBase()
 
   auto ipm = weak_ipm_.lock();
 
-  if (!use_intra_process_) {
+  if (!intra_process_is_enabled_) {
     return;
   }
   if (!ipm) {
@@ -183,13 +183,12 @@ size_t
 PublisherBase::get_intra_process_subscription_count() const
 {
   auto ipm = weak_ipm_.lock();
-  if (!use_intra_process_) {
+  if (!intra_process_is_enabled_) {
     return 0;
   }
   if (!ipm) {
     // TODO(ivanpauno): should this just return silently? Or maybe return with a warning?
     //                  Same as wjwwood comment in publisher_factory create_shared_publish_callback.
-    //                  If we don't raise an error here, use_intra_process_ flag is unnecessary.
     throw std::runtime_error(
             "intra process subscriber count called after "
             "destruction of intra process manager");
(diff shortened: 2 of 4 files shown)


rclcpp/src/rclcpp/publisher.cpp
```diff
@@ -231,6 +231,19 @@ PublisherBase::setup_intra_process(
   IntraProcessManagerSharedPtr ipm,
   const rcl_publisher_options_t & intra_process_options)
 {
+  // Skip intraprocess configuration if the "durability" qos policy is not "volatile".
+  // TODO(ivanpauno):
```
ivanpauno: It would be good, in general, to have methods to check what is the real value of RMW_QOS_*_SYSTEM_DEFAULT. Or maybe, just delete all these defines and only support explicit configurations. Having unknown default values doesn't seem like a good idea.

ivanpauno: Correct snippet of the TODO:
https://github.com/ros2/rclcpp/blob/c31b273c1fc8802dcdd293266b1b5212ec9583ef/rclcpp/src/rclcpp/publisher.cpp#L235-L246

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -216,7 +216,9 @@ class Publisher : public PublisherBase
   virtual void
   publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
   {
-    this->do_inter_process_publish(msg.get());
+    if (!use_intra_process_ || get_subscription_count() > get_intra_process_subscription_count()) {
```
wjwwood: If the QoS is not volatile, then it should always do inter process.

rclcpp/src/rclcpp/publisher.cpp
wjwwood: Personally I agree with you that we should avoid implicit settings, which is what the `SYSTEM_DEFAULT` gives you, but there's a use case that some people want supported, which allows you to configure QoS settings from external XML files. That is what this allows for, so I don't know if we could realistically get rid of those.

We definitely should have a way to check the effective QoS settings after the publisher is created.

Also, I think this (re)raises the issue that `use_intra_process_comms` needs to be an option on publishers/subscriptions, and not on the node itself. Otherwise you'd be unable to use intra process _and_ have a non-volatile durability on one of your publishers.

Once we have that, I'd say it should be an error (not a warning) to have both intra process enabled and transient local durability on the same publisher/subscription.

rclcpp/include/rclcpp/publisher.hpp
ivanpauno: If QoS is non volatile, use_intra_process_ flag=false, as setup intraprocess doesn't set it to true. It just return without setting-up the intraprocess comm.

wjwwood: You're right, I was thinking that was the flag the user set.

----

This logic could use a comment, I'd even support making an intermediate bool to make it clearer why you're comparing the count's. For example:

```suggestion
    bool intra_process_is_enabled = use_intra_process_;  // might consider renaming this variable entirely
    bool inter_process_subscriptions_exist = get_subscription_count() > get_intra_process_subscription_count();
    if (!intra_process_is_enabled || inter_process_subscriptions_exist) {
```

rclcpp/src/rclcpp/publisher.cpp
ivanpauno: I agree about having use_intra_process_comms option at publish/subscription level. Switching the warning to an error after that also sounds good.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -169,7 +169,7 @@ class PublisherBase
 
   using IntraProcessManagerWeakPtr =
     std::weak_ptr<rclcpp::intra_process_manager::IntraProcessManager>;
-  bool use_intra_process_;
+  bool intra_process_is_enabled_;
```
dirk-thomas: Regarding the naming of the variable I see why you don't want to use `use_` anymore since the decision about if it is actually being used is delayed. I think the order of the parts of the variable name was better before. I would suggest `allow_intra_process_` instead.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -216,7 +216,11 @@ class Publisher : public PublisherBase
   virtual void
   publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
   {
-    this->do_inter_process_publish(msg.get());
+    bool inter_process_subscriptions_exist =
```
dirk-thomas: Similar to the above I would suggest naming this one `has_inter_process_subscriptions`.

rclcpp/src/rclcpp/publisher.cpp
```diff
@@ -231,6 +230,19 @@ PublisherBase::setup_intra_process(
...
+  //    RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT is for FastRTPS and Connext also
+  //    equal to "volatile". It would be good to add a
+  //    is_durability_policy_volatile(rcl_publisher_options_t options)
+  //    method to rcl, and corresponging methods in rmw and the rmw vendor
+  //    implementations, which will also checks if "system_default" is "volatile" or not.
+  if (intra_process_options.qos.durability != RMW_QOS_POLICY_DURABILITY_VOLATILE) {
+    RCLCPP_WARN(
+      rclcpp::get_logger("rclcpp"),
+      "Enabled intraprocess communication with incompatible QoS policy.");
```
dirk-thomas: I think the warning should be more specific about the consequence:

> Skipping intraprocess communication with incompatible QoS policy.

rclcpp/include/rclcpp/publisher.hpp
wjwwood: I don't agree about this name change suggestion, but the other one is fine.

```c++
if (!allow_intra_process_ || has_inter_process_subscriptions) {
```

Isn't as clear to read in my opinion, as it reads to me one of "not allow intra process" or "allow not intra process", neither of which make sense to me, you have to look at it a while to get to "intra process is not allowed" and even then allow implies to me that it may or may not be enabled, which isn't the case. At this point it is either enable or it is not.

Whereas:

```c++
if (!intra_process_is_enabled_ || has_inter_process_subscriptions) {
```

Is very clear immediately, as it reads one of "not intra process is enabled" or "intra process is not enabled", which are better for me than the alternative you suggested.

So 👎 from me to change this variable name.

rclcpp/include/rclcpp/publisher.hpp
wjwwood: I don't find this name any better or worse.

rclcpp/src/rclcpp/publisher.cpp
wjwwood: @ivanpauno are you not planning to add the publisher/subscription option to enable intra-process and therefore make this an error? If you are planning to do so, then I think this warning does not need to be reworded as it will be removed.

ivanpauno: I will do it in another PR. For the moment, I will left the warning reworded.

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6505/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2855/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5330/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6304/)

rclcpp/src/rclcpp/publisher.cpp
wjwwood: I don't really agree, I think warning isn't really sufficient, as I said in my other comment this basically prevents you from using intra-process and transient-local durability at the same time (even if on different entities). So after merging this pr we'll be potentially breaking existing use cases for a very narrow optimization. I would prefer to see the improvement done completely in this pr than have two pr's.

rclcpp/src/rclcpp/publisher.cpp
wjwwood: I still find this point unresolved, until we have the ability to know what the durability is when `SYSTEM_DEFAULT` is in use and until we have the ability to control which publishers and subscriptions have intra-process enabled.

---

To be clear about the first point, I think we need a way to inspect a publisher or subscription after creation to see which durability it is using. Something like `rcl_ret_t rcl_publisher_get_actual_qos(const rcl_publisher_t *, rmw_qos_profile *)`, which may need an `rmw` equivalent.

I know this increases the scope of this issue considerably, but without it the warning is only going to work part of the time and would produce very subtle bugs for the user.

The alternative is an even bigger scope increase, which is to actually support transient-local durability correctly when intra-process is enabled.

---

For the second point, what I imagine is that the creation publishers and subscriptions have a three-way option which is "explicitly enable intra-process", "explicitly disable intra-process", or "default". Where "default" would be based on the intra-process setting at the node level.

I think this is important to do now (again I know it increases the scope of this issue), because otherwise you have to use a warning rather than an error to let the user know that they have conflicting configurations. Without it, you also have scenarios which are impossible to achieve, i.e. using intra-process and transient-local durability at the same time but on different publishers and/or subscriptions.

ivanpauno:
> To be clear about the first point, I think we need a way to inspect a publisher or subscription after creation to see which durability it is using. Something like rcl_ret_t rcl_publisher_get_actual_qos(const rcl_publisher_t *, rmw_qos_profile *), which may need an rmw equivalent.

Ok, I will add those methods.

> The alternative is an even bigger scope increase, which is to actually support transient-local durability correctly when intra-process is enabled.

Supporting non-volatile durability is a really big change, specially in the case HISTORY=KEEP_ALL. I think we can do a big improvement in the intraprocess comm without tackling that.

> For the second point, what I imagine is that the creation publishers and subscriptions have a three-way option which is "explicitly enable intra-process", "explicitly disable intra-process", or "default". Where "default" would be based on the intra-process setting at the node level.

I was thinking in adding a bool option to them, but a three-way option sounds quite better.

wjwwood:
> Ok, I will add those methods.

Thank you, I know it's a bit of a task, so if you need help let me know.

> Supporting non-volatile durability is a really big change, specially in the case HISTORY=KEEP_ALL. I think we can do a big improvement in the intraprocess comm without tackling that.

I agree, that's why I want to instead properly plug this hole in the API until we can do the improvement.

Which also brings up the point, if there are any other QoS the intra-process are incompatible with at the moment, we should use these new "actual QoS" accessors to disable it in those cases as well.

> I was thinking in adding a bool option to them, but a three-way option sounds quite better.

Again, a little more work, but I think it will make it easier for people to transition to from the current state.

ivanpauno:
I didn't merge this intentionally, I don't know when the error happened.
I'm going to revert the changes.

wjwwood:
Wait did you revert it too? Please be careful to only merge things that have been reviewed and pass CI...

wjwwood:
Wait was it merged into something other than master? That makes more sense 😀

ivanpauno:
> Wait was it merged into something other than master? That makes more sense

Yes, thanks for the correction! I merged this into ivanpauno/publisher_get_actual_qos branch unintentionally. I've already reverted the changes in that branch, and opened a follow-up PR [here](https://github.com/ros2/rclcpp/pull/674).

PR #665, opened 2019-03-22
Introduce rclcpp_components to implement composition

mjcarroll:
This introduces `rclcpp_components` and drops the composition CMake macros from `rclcpp`.


rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,197 @@
...
+ComponentManager::ComponentManager(std::weak_ptr<rclcpp::executor::Executor> executor)
+: Node("ComponentManager"),
+  executor_(executor)
+{
+  loadNode_srv_ = create_service<LoadNode>("_container/load_node",
+      std::bind(&ComponentManager::OnLoadNode, this, _1, _2, _3));
+  unloadNode_srv_ = create_service<UnloadNode>("_container/unload_node",
+      std::bind(&ComponentManager::OnUnloadNode, this, _1, _2, _3));
+  listNodes_srv_ = create_service<ListNodes>("_container/list_nodes",
+      std::bind(&ComponentManager::OnListNodes, this, _1, _2, _3));
```
hidmic: @mjcarroll IIUC topics should be private ones i.e. with a `~/` prefix.

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,197 @@
...
+          parameters.push_back(rclcpp::Parameter::from_parameter_msg(p));
+        }
+
+        auto options = rclcpp::NodeOptions()
+          .initial_parameters(parameters)
+          .arguments(request->remap_rules);
+
+        auto node_id = unique_id++;
+
+        node_wrappers_[node_id] = node_factory->create_node_instance(
```
hidmic: Something funky is going w.r.t. node naming, I get:

```sh
[component_container-1] [WARN] [rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed
 it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```

when loading the talker node via launch.

hidmic: Yeap, I can confirm that even though I request for a given node name, the fully qualified node name that the component gets once loaded is the same as that of the container.

mjcarroll: Try again with the new modifications, I should have been handling the name and namespace with remapping.

hidmic: Works perfectly now!

mjcarroll:
Full CI to see if anything else is dependent on current component impl that I didn't catch and to check cross-platform issues.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6531/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2873/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5350/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6326/)

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,201 @@
...
+      "rclcpp_components", request->package_name, content, &base_path))
+  {
+    RCLCPP_ERROR(get_logger(), "Could not find requested resource in ament index");
+    response->error_message = "Could not find requested resource in ament index";
+    response->success = false;
+    return;
+  }
+
+  std::vector<std::string> lines = split(content, '\n', true);
+  for (auto line : lines) {
```
hidmic: @mjcarroll can we make it `const auto &` instead?

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,201 @@
...
+        return;
+      } catch (...) {
+        RCLCPP_ERROR(get_logger(), "Failed to load library");
+        response->error_message = "Failed to load library";
+        response->success = false;
+        return;
+      }
+    }
+    auto classes = loader->getAvailableClasses<rclcpp_components::NodeFactory>();
+    for (auto clazz : classes) {
```
hidmic: @mjcarroll can we make it `const auto &` instead?

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,201 @@
...
+      RCLCPP_INFO(get_logger(), "Found class %s", class_name.c_str());
+      if (clazz == class_name || clazz == fq_class_name) {
+        RCLCPP_INFO(get_logger(), "Instantiate class %s", class_name.c_str());
+        loaders_[library_path] = loader;
+
+        auto node_factory = loader->createInstance<rclcpp_components::NodeFactory>(clazz);
+
+        std::vector<rclcpp::Parameter> parameters;
+        for (auto & p : request->parameters) {
```
hidmic: @mjcarroll can we make it `const auto &` instead?

rclcpp_components/src/component_manager.hpp
```diff
@@ -0,0 +1,78 @@
...
+  void OnListNodes(
+    const std::shared_ptr<rmw_request_id_t> request_header,
+    const std::shared_ptr<ListNodes::Request> request,
+    std::shared_ptr<ListNodes::Response> response);
+
+private:
+  std::weak_ptr<rclcpp::executor::Executor> executor_;
+
+  std::map<uint64_t, rclcpp_components::NodeInstanceWrapper> node_wrappers_;
+  std::map<std::string, class_loader::ClassLoader *> loaders_;
```
hidmic: @mjcarroll who's managing the lifetime of `ClassLoader` instances? Should we make them smart pointers instead?

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,201 @@
...
+          parameters.push_back(rclcpp::Parameter::from_parameter_msg(p));
+        }
+
+        std::vector<std::string> remap_rules {request->remap_rules};
+
+        if (request->node_name.length()) {
```
hidmic: @mjcarroll nit, can ignore: why not use `! request->node_name.empty()` instead?

rclcpp_components/src/component_manager.hpp
mjcarroll: The ComponentManager is managing lifetime, I kept them as raw pointers because that's what the composition demo was doing, but smart pointers probably makes it safer.

mjcarroll:
Fixing CMake and addressing some comments:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6546/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2885/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5363/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6340/)

rclcpp_components/include/rclcpp_components/node_factory.hpp
```diff
@@ -0,0 +1,35 @@
...
+class NodeFactory
+{
+public:
+  NodeFactory() = default;
+
+  virtual ~NodeFactory() = default;
+
+  virtual
+  NodeInstanceWrapper
+  create_node_instance(rclcpp::NodeOptions options) = 0;
```
Karsten1987:
```suggestion
  create_node_instance(const rclcpp::NodeOptions & options) = 0;
```

wjwwood:
They're also available in `plunginlib`:

https://github.com/ros/pluginlib/blob/3bbc242f1cbdeb9a51375fb8941deb386dbf9f2f/pluginlib/include/pluginlib/impl/split.hpp#L42-L51

Really, a "std::filesystem" helper would live best in `rcpputils` now that it exists.

mjcarroll:
Yeah, that's where I copied them from, I'll add the filesystem helper to rcpputils, that makes sense.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6579/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2917/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5395/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6388/)

rclcpp_components/cmake/rclcpp_components_register_nodes.cmake
```diff
@@ -13,38 +13,38 @@
 # limitations under the License.
 
 #
-# Register a node plugin with the ament resource index.
+# Register an rclcpp_component with the ament resource index.
```
wjwwood: `rclcpp_component` -> `rclcpp component`

not using suggestion feature because of DCO :(

rclcpp_components/cmake/rclcpp_components_register_nodes.cmake
```diff
@@ -13,38 +13,38 @@
...
 #
-# The passed library can contain multiple plugins extending the node interface.
+# The passed library can contain multiple nodes
```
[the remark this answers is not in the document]
wjwwood: Was this resolved? It is still missing a period.

rclcpp_components/include/rclcpp_components/node_factory_template.hpp
```diff
@@ -0,0 +1,45 @@
...
+  virtual ~NodeFactoryTemplate() = default;
+
+  virtual
+  NodeInstanceWrapper
+  create_node_instance(const rclcpp::NodeOptions & options)
+  {
+    auto node = std::make_shared<NodeT>(options);
+
+    return NodeInstanceWrapper(node,
+             std::bind(&NodeT::get_node_base_interface, node));
```
wjwwood: Indention is weird here, wrapping both arguments would be preferable to indenting 9? times. Try to only ever indent no more than two spaces with each successive scope.

rclcpp_components/package.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>rclcpp_components</name>
+  <version>0.6.2</version>
+  <description>Package containing container tools for dynamically loadable components</description>
```
wjwwood: Drop `container`? or reword to be something like `containing tools and containers for ...`?

wjwwood: Also punctuation? I guess it's a sentence fragment.

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,203 @@
...
+  {
+    RCLCPP_ERROR(get_logger(), "Could not find requested resource in ament index");
+    response->error_message = "Could not find requested resource in ament index";
+    response->success = false;
+    return;
+  }
+
+  std::vector<std::string> lines = rcpputils::split(content, '\n', true);
+  for (const auto & line : lines) {
+    std::vector<std::string> parts = rcpputils::split(line, ';');
```
wjwwood: It would be nice, I think, to see this kind of logic put into a function or static method, and then unit tested. Most of the other stuff in this pr is hard to test due to the nature of needing end-to-end shared libraries and ament index and such, but unit testing what we can would be nice to have.

rclcpp_components/include/rclcpp_components/node_factory.hpp
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+#ifndef RCLCPP_COMPONENTS__NODE_FACTORY_HPP__
+#define RCLCPP_COMPONENTS__NODE_FACTORY_HPP__
+
+#include "rclcpp_components/node_instance_wrapper.hpp"
+
+namespace rclcpp_components
+{
+class NodeFactory
```
wjwwood: All of these public class could do with some documentation. It doesn't need to overlap with user documentation on how it's used, or the architecture document I started, but just a blurb about what it is and what it is used for would help code spelunkers.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6582/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2920/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5398/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6391/)

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,203 @@
...
+          remap_rules.push_back("__ns:=" + request->node_namespace);
+        }
+
+        auto options = rclcpp::NodeOptions()
+          .initial_parameters(parameters)
+          .arguments(remap_rules);
+
+        auto node_id = unique_id++;
+
+        node_wrappers_[node_id] = node_factory->create_node_instance(options);
```
sloretz: Might be worth wrapping in `try` / `catch` in case the constructor raises an exception.

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+        auto node = node_wrappers_[node_id].get_node_base_interface();
+        if (auto exec = executor_.lock()) {
+          exec->add_node(node, true);
+        }
+        response->full_node_name = node->get_fully_qualified_name();
```
sloretz: The [design pr](https://github.com/ros2/design/pull/206/files#diff-df7db447d3b09c67c5487a923f56a445R527) required this to cancel loading the node if there is a name conflict, but now I'm thinking that sentence should be removed in favor of referencing a design doc that satisfies ros2/design#187. I don't think anything here needs to change.

rclcpp_components/src/component_manager.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+        if (!request->node_namespace.empty()) {
+          remap_rules.push_back("__ns:=" + request->node_namespace);
+        }
+
+        auto options = rclcpp::NodeOptions()
+          .initial_parameters(parameters)
+          .arguments(remap_rules);
+
+        auto node_id = unique_id++;
```
sloretz: If nodes are rapidly loaded/unloaded for a long time then overflow here could be a concern. One option is to increment until an unused ID is found.

wjwwood: I think it's probably fine, but you could also just die if you detect rollover:

https://github.com/ros2/rclcpp/blob/1f2904f980aaaa7ccbfb40d722795b2d9a55a9aa/rclcpp/src/rclcpp/intra_process_manager.cpp#L70-L83

mjcarroll:
@wjwwood and @sloretz Integrated y'alls feedback and also added some tests around the ComponentManager internals and API.  Please take another look.

mjcarroll:
CI all of ros2.repos to make sure I didn't miss any composition dependencies.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6588/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2926/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5407/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6409/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6589/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2927/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5408/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6411/)

mjcarroll:
Fix lint on `logging_demo`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6603/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2937/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5420/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6428/)

Issue #643, opened 2019-02-28
get_node_names should include namespace


wjwwood:
`rcl` returns the node names and their namespaces, but in rclcpp you cannot get the node namespaces, see:

https://github.com/ros2/rclcpp/blob/3919ab189795fbf47b5427add06f45d8a43f81ce/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp#L133-L143

There should probably be a new function that returns the nodes and their namespaces separately, and then keep the existing one for convenience but change it to automatically concatenate the name and namespace, because that's likely what you want in most cases.

oswinso:
I'd like to work on this. Do you mean like changing `NodeGraph::get_node_names()` to return something like `"/namespace/nodename"`, and then having a separate function that returns a `std::pair` of nodes and their namespaces? But isn't that what `NodeGraph::get_node_names_and_namespaces()` is doing?

mjcarroll:
For consistency with: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/node_base.hpp#L49

You could probably do:

```
std::vector<std::pair<std::string, std::string>> get_node_names_and_namespaces();
std::vector<std::string> get_node_names();
std::vector<std::string> get_node_namespaces();
std::vector<std::string> get_node_fully_qualified_names()
```

Is that what you were thinking @wjwwood ?

wjwwood:
Yes, except I don't think `get_node_namespaces()` is needed.

wjwwood:
Well, and I was advocating for just changing `get_node_names()` to return the combined namespace/name, so then `get_node_fully_qualified_names()` is redundant. But that would break "ABI", in the behavior sense, so it would have to go into Dashing only.

jhdcs:
I've also got a possible solution, assuming I've got my environment set up correctly. Would you like me to submit a pull request as well? Also, should we write tests for this functionality?

wjwwood:
> Would you like me to submit a pull request as well?

Ideally you guys could collaborate and avoid duplicated effort. I think the best way to do that is to open a pull request as early as possible, so yes I think that's a good idea.

> Also, should we write tests for this functionality?

Always :)

jhdcs:
Having a few issues creating a pull request. Should I clone the repository to my GitHub account in addition to my local computer or something?

wjwwood:
Yes, GitHub has pretty good documentation on contributing via a pull request:

[...]

PR #669, opened 2019-03-26
[WIP] Add functions to return formatted Node Name-Namespace strings

jhdcs:
Implements functions to add formatted node name and namespace strings. Still needs unit tests. Design tips and suggestions for useful tests is greatly appreciated!

Signed-off-by: Jacob Hassold <jhassold@dcscorp.com>

Connects to ros2/rclcpp#643


rclcpp/include/rclcpp/node.hpp
```diff
@@ -392,6 +392,14 @@ class Node : public std::enable_shared_from_this<Node>
   std::vector<std::string>
   get_node_names() const;
 
+  RCLCPP_PUBLIC
+  std::vector<std::string>
+  get_node_namespaces() const;
+
+  RCLCPP_PUBLIC
+  std::vector<std::string>
+  get_node_names_namespaces_formatted(const char & separator) const;
```
wjwwood: I think this method isn't really necessary, since the separator is know (`/`). It might change in the future, but that's nothing that user should need to care about.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -392,6 +392,14 @@ class Node : public std::enable_shared_from_this<Node>
   std::vector<std::string>
   get_node_names() const;
 
+  RCLCPP_PUBLIC
+  std::vector<std::string>
+  get_node_namespaces() const;
```
wjwwood: How do you know which namespace goes with which node? I think a `get_node_names_and_namespaces() -> std::vector<std::pair<std::string, std::string>>` would be a preferable addition.

rclcpp/include/rclcpp/node.hpp
oswinso: I've removed this method, and changed `get_node_names` to return the fully qualified name.

rclcpp/include/rclcpp/node.hpp
jhdcs: The get_node_namespaces function has been removed. get_node_names_and_namespaces() already has a signature of std::vector<std::pair<std::string, std::string>>, so not sure much needs to be changed here.

jhdcs:
@wjwwood I believe we've addressed the issues that you raised. Any further things that jump out at you?

Also, CLion is warning me of a bunch of Clang violations as I'm going through the code. Should these be cleaned up? Should we make an issue for them? Or should I ignore them?

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -18,6 +18,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <algorithm>
```
wjwwood: nitpick: alphabetically order the includes

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -173,10 +173,12 @@ NodeGraph::get_node_names_and_namespaces() const
     throw std::runtime_error(error_msg);
   }
 
-  std::vector<std::pair<std::string, std::string>> node_names(node_names_c.size);
+  // TODO(jhdcs): If ranges are ever added to ROS2, replace following with range-based filter-transform
```
wjwwood: That would be cool, but this same desire applies to other places in the code, so maybe we just drop the TODO for now. If you really want to track it we can make an issue, but realistically we'd audit the code when we're able and replace all instances where it would be better.

rclcpp/test/test_node.cpp
```diff
@@ -73,6 +73,17 @@ TEST_F(TestNode, get_name_and_namespace) {
...
+  {
+    auto node1 = std::make_shared<rclcpp::Node>("my_node1", "my/ns");
+    auto node2 = std::make_shared<rclcpp::Node>("my_node2", "my/ns");
+    auto node3 = std::make_shared<rclcpp::Node>("my_node3", "/ns2");
+    auto node4 = std::make_shared<rclcpp::Node>("my_node4", "my/ns3");
+    auto names_and_namespaces = node1->get_node_names();
+    EXPECT_EQ("/my/ns/my_node1", names_and_namespaces[0]);
+    EXPECT_EQ("/my/ns/my_node2", names_and_namespaces[1]);
+    EXPECT_EQ("/ns2/my_node3", names_and_namespaces[2]);
+    EXPECT_EQ("/my/ns3/my_node4", names_and_namespaces[3]);
```
wjwwood: The API doesn't guarantee the order of the nodes in this listing, so you're sort of depending on implementation behavior here. I'd suggest just checking that each of these exist somewhere in the output. This would be robust to reordering and also to other nodes existing at the same time (imagine a new test running node or something).

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
jhdcs: Your nitpick is my command.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
jhdcs: I'm just really fond of trying out functional design in areas that don't really have them before. I'll remove this TODO though.

There is a library that adds ranges on GitHub (it's actually the basis for a proposal to add ranges to the standard library), but I don't want to start importing random codebases willy-nilly without consent. _Especially_ since this isn't my project.

jhdcs:
> Make sure you run the tests that our buildfarm would run to help us catch issues early, as CI turn around is pretty long, see: https://index.ros.org/doc/ros2/Tutorials/Colcon-Tutorial/#run-tests

I've been using "colcon test --packages-select rclcpp" from the root of my development workspace. Is this not correct?

rclcpp/test/test_node.cpp
jhdcs: Excellent thinking. Also, didn't realize that about the API. Oops.

I implemented this functionality via an unordered set that's generated from the vector.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
wjwwood: Yeah, unfortunately I think pragmatism wins here (I've been known import future features as third-party packages too, e.g. `std::variant` and `std::filesystem`). Thanks for removing it.

wjwwood:
> I do get several errors about style formatting left around the entire rclcpp package, which was part of the reason I was wondering if you wanted me to go and give the code a good scrubbing.

That's odd, on master it should produce no errors or failures, so any failures should be due to your diff or maybe some misconfiguration on your system (wrong or missing test dependencies?).

We'll see what CI says:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6580/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2918/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5396/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6389/)

jhdcs:
I should clarify - the IDE I'm using (CLion) is giving Clang style warnings. They're more like guidelines that something looks odd, rather than an actual error.

Apologies for the confusion, I misspoke.

jhdcs:
Probably a silly question (I'm new to this), but is there anything I should be doing? The CI is saying that the change was on "Force Reinstallation of Numpy", am I reading it incorrectly?

dirk-thomas:
> The CI is saying that the change was on "Force Reinstallation of Numpy", am I reading it incorrectly?

That is the latest commit in the `ros2/ci` repo which provides the logic for the CI jobs.

jhdcs:
So that change pertains to the CI system, rather than my code. Got it.

So the instability is being caused by something I wrote? I guess I'd better start diving again...

wjwwood:
@jhdcs let me know when you want me to rerun CI, the lastest round has test failures that look related to these changes (based on a cursory look only).

jhdcs:
Yeah. I'm thinking that something might have been expecting the old format. Possibly. I'll let you know once I figure it out on my end.

jhdcs:
Okay, I _think_ I know what's going on.

NodeGraph::get_node_names() is used throughout the API, and the tests seem to expect an unqualified name (i.e: name, but no namespace). I can either update all the tests to accept the new version of get_node_names, or revert get_node_names to its original functionality, and create a new function called NodeGraph::get_qualified_node_names() that implements our desired functionality.

I'm personally partial to the second option, both because it's easier, and because I'm not sure how much of the existing codebase is going to break if we suddenly swap the expected output of an already-in-use function.

@wjwwood, what do you think?

jhdcs:
@wjwwood Would you be able to re-run CI? I think I might have gotten some things fixed, though I did find something interesting: Apparently if you use std::transform in the original get_node_names() function, you get a timeout error in some of the tests (specifically, gtest_multithreaded_rmw_fastrtps/opensplice, and at least on my system). As such, I've reverted it to the for-based implementation until I can figure out what on earth is going on...

wjwwood:
@jhdcs please revert you most recent commit, the idea is to change the behavior of `get_node_names()`, because it was classified as a bug that it did not include the namespace from the beginning, and update any tests we have that need to be. The node name needs to always be associated with the namespace, otherwise you may have duplicate node names and no way to disambiguate them.

Please (re)read my suggested solution in the original issue: https://github.com/ros2/rclcpp/issues/643#issue-415849240

It is to update the behavior of the existing function (and fix anything that breaks due to that) and optionally add a function that returns a list of namespace and names as pairs.

wjwwood:
@jhdcs looks like something got messed up when you force pushed, and now you have changes from master on this branch...

jhdcs:
GAH!

It's just one of those weeks, isn't it?

jhdcs:
Hopefully that fixes it. Sorry for all the trouble, and I really appreciate your guys' patience...

wjwwood:
Actually now the pull request is closed and I cannot reopen it... This usually happens if you delete the upstream branch or fork, but I'm not sure what's happening right now...

jhdcs:
That's basically what happened. Part of the fork was deleted accidentally due to my uncanny ability to mega-derp. No data was lost, but GitHub is really confused.

I was going to open a new pull request that references both this one and the main issue later on, once I feel like I've worked a few more of the kinks out. Unless you want me to open it now, of course.

wjwwood:
> I was going to open a new pull request that references both this one and the main issue later on, once I feel like I've worked a few more of the kinks out. Unless you want me to open it now, of course. 

Oh that's up to you. Feel free to open a new pr when ever you're ready. I just wanted to make sure you knew I could not reopen this one.

PR #670, opened 2019-03-26
Update graph API to support new names and types structure

jacobperron:
Connects to ros2/ros2#677


jacobperron:
Decided to return a delimited string rather than returning a nested list to keep the API simpler.

PR #671, opened 2019-03-28
Replace const char * with a std::array<char, TOPIC_NAME_LENGTH> as the key of IPM IDTopicMap

### LINKED ISSUE #668: Subscription and intra-process subscription counts are not updated in sync
## Bug report

Intra-process subscription count of Publisher is not updated in sync with the subscription count when intra-process communication is used.
 When the `get_subscription_count()` method returns the updated count,  `get_intra_process_subscription_count()` may not return the updated count yet. This behaviour causes the unit tests in [test_publisher_subscription_count_api](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/test_publisher_subscription_count_api.cpp) to fail [occasionally](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1323/testReport/(root)/projectroot/test_publisher_subscription_count_api/).

**Required Info:**

- Operating System:
  - Ubuntu 16.04 
- Installation type:
  - Source 
- Version or commit hash:
  - 8743bcb0a1ef06d6eea5ebbd247e019f9dbb1e10 
- DDS implementation:
  - Fast-RTPS 
  - rclcpp 

#### Steps to reproduce issue
Repeat the unit test [test_publisher_subscription_count_api](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/test_publisher_subscription_count_api.cpp) until failure. Though this may be hard to reproduce locally, here's an instance of such a failure:  https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1323/testReport/(root)/projectroot/test_publisher_subscription_count_api/

Above test logs shows that the equality conditions on `get_subscription_count()`  are satisfied yet the conditions on  `get_intra_process_subscription_count()` are not met.

#### Expected behavior
`get_subscription_count()` and `get_intra_process_subscription_count()` methods should be in sync when intra-process communication is used.

#### Actual behavior
The methods return different results.

ivanpauno:
Closes #668

## Changes

Replaces the `const char *` key in IDTopicMap with a std::array<char, TOPIC_NAME_LENGTH> key. A copy of the string is forced.

## Problem description

After some debugging, I realized what this random error was.
Sometimes, the topic name  wasn't found in the subscription_ids_by_topic_ map here:
https://github.com/ros2/rclcpp/blob/cb20529e5e218cde0106c59b12067a9ca9f43022/rclcpp/include/rclcpp/intra_process_manager_impl.hpp#L266-L270
And a 0 subscription count was returned.

The problem is that a `const char *` was used as a key, an the liveliness of that raw pointer was the liveliness of the subscription:
https://github.com/ros2/rclcpp/blob/cb20529e5e218cde0106c59b12067a9ca9f43022/rclcpp/include/rclcpp/intra_process_manager_impl.hpp#L98-L104
When another subscription was registered in the same topic, and the first subscription was already out of scope, it sometimes resulted in the correct behavior and sometimes failed silently. The OS usually not raise a segfault when accessing a recently freed block of memory, but data could be already corrupted at that moment.

## Bug reproduction

The error is not easy to reproduce. Here is the script I've used:
```bash
#!/bin/bash

set -e
mkdir -p logs_test_intra
for run in {1..50}; do
    echo "Starting test #" $run
    out=$(colcon test --packages-select rclcpp --ctest-arg -R test_publisher_subscription_count_api)
    echo "$out"
    substring="stderr"
    error_list=()
    if [ "${out/$substring}" != "$out" ]; then
        error_list+=($run)
    fi
    cp -L log/latest "logs_test_intra/test_${run}" -R
    echo
    echo "Finished test #" $run
    echo
    echo
done

echo ${error_list[*]}
```
Before I found one error per about 20 runs. I haven't found it again in 100 runs after this change.

## Closes 668

This is not syncing intra-process-subscription count and subscription count as requested in #668.
But it solves the bug, and syncing them is probably not a good idea as has been already discussed in the issue.

### DIFF
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -15,8 +15,6 @@
 #ifndef RCLCPP__INTRA_PROCESS_MANAGER_IMPL_HPP_
 #define RCLCPP__INTRA_PROCESS_MANAGER_IMPL_HPP_
 
-#include <rmw/validate_full_topic_name.h>
-
 #include <algorithm>
 #include <atomic>
 #include <cstring>
@@ -29,6 +27,8 @@
 #include <string>
 #include <unordered_map>
 #include <utility>
+
+#include "rmw/validate_full_topic_name.h"
 
 #include "rclcpp/macros.hpp"
 #include "rclcpp/mapped_ring_buffer.hpp"
@@ -89,20 +89,6 @@ private:
   RCLCPP_DISABLE_COPY(IntraProcessManagerImplBase)
 };
 
-class FixedSizeString
-{
-public:
-  explicit FixedSizeString(const char * str);
-  const char * c_str() const;
-
-private:
-  char str_[RMW_TOPIC_MAX_NAME_LENGTH + 1];
-};
-
-RCLCPP_PUBLIC
-bool
-operator<(const FixedSizeString & lhs, const FixedSizeString & rhs);
-
 template<typename Allocator = std::allocator<void>>
 class IntraProcessManagerImpl : public IntraProcessManagerImplBase
 {
@@ -114,7 +100,7 @@ public:
   add_subscription(uint64_t id, SubscriptionBase::SharedPtr subscription)
   {
     subscriptions_[id] = subscription;
-    subscription_ids_by_topic_[FixedSizeString(subscription->get_topic_name())].insert(id);
+    subscription_ids_by_topic_[fixed_size_string(subscription->get_topic_name())].insert(id);
   }
 
   void
@@ -190,7 +176,7 @@ public:
 
     // Figure out what subscriptions should receive the message.
     auto & destined_subscriptions =
-      subscription_ids_by_topic_[FixedSizeString(publisher->get_topic_name())];
+      subscription_ids_by_topic_[fixed_size_string(publisher->get_topic_name())];
     // Store the list for later comparison.
     if (info.target_subscriptions_by_message_sequence.count(message_seq) == 0) {
       info.target_subscriptions_by_message_sequence.emplace(
@@ -278,7 +264,8 @@ public:
     if (!publisher) {
       throw std::runtime_error("publisher has unexpectedly gone out of scope");
     }
-    auto sub_map_it = subscription_ids_by_topic_.find(FixedSizeString(publisher->get_topic_name()));
+    auto sub_map_it =
+      subscription_ids_by_topic_.find(fixed_size_string(publisher->get_topic_name()));
     if (sub_map_it == subscription_ids_by_topic_.end()) {
       // No intraprocess subscribers
       return 0;
@@ -288,6 +275,16 @@ public:
 
 private:
   RCLCPP_DISABLE_COPY(IntraProcessManagerImpl)
+
+  using FixedSizeString = std::array<char, RMW_TOPIC_MAX_NAME_LENGTH + 1>;
+
+  FixedSizeString
+  fixed_size_string(const char * str) const
+  {
+    FixedSizeString ret = {0};
+    std::strncpy(ret.begin(), str, ret.size());
+    return ret;
+  }
 
   template<typename T>
   using RebindAlloc = typename std::allocator_traits<Allocator>::template rebind_alloc<T>;
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -289,7 +289,7 @@ private:
     std::string,
     AllocSet,
     std::less<std::string>,
-    RebindAlloc<std::pair<const char * const, AllocSet>>>;
+    RebindAlloc<std::pair<const std::string, AllocSet>>>;
 
   SubscriptionMap subscriptions_;
 
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -15,6 +15,8 @@
 #ifndef RCLCPP__INTRA_PROCESS_MANAGER_IMPL_HPP_
 #define RCLCPP__INTRA_PROCESS_MANAGER_IMPL_HPP_
 
+#include <rmw/validate_full_topic_name.h>
+
 #include <algorithm>
 #include <atomic>
 #include <cstring>
@@ -87,6 +89,17 @@ private:
   RCLCPP_DISABLE_COPY(IntraProcessManagerImplBase)
 };
 
+class FixedSizeString
+{
+public:
+  explicit FixedSizeString(const char * str);
+  const char * c_str() const;
+
+private:
+  char str_[RMW_TOPIC_MAX_NAME_LENGTH + 1];
+};
+bool operator<(const FixedSizeString & lhs, const FixedSizeString & rhs);
+
 template<typename Allocator = std::allocator<void>>
 class IntraProcessManagerImpl : public IntraProcessManagerImplBase
 {
@@ -98,9 +111,7 @@ public:
   add_subscription(uint64_t id, SubscriptionBase::SharedPtr subscription)
   {
     subscriptions_[id] = subscription;
-    // subscription->get_topic_name() -> const char * can be used as the key,
-    // since subscriptions_ shares the ownership of subscription
-    subscription_ids_by_topic_[subscription->get_topic_name()].insert(id);
+    subscription_ids_by_topic_[FixedSizeString(subscription->get_topic_name())].insert(id);
   }
 
   void
@@ -175,7 +186,8 @@ public:
     }
 
     // Figure out what subscriptions should receive the message.
-    auto & destined_subscriptions = subscription_ids_by_topic_[publisher->get_topic_name()];
+    auto & destined_subscriptions =
+      subscription_ids_by_topic_[FixedSizeString(publisher->get_topic_name())];
     // Store the list for later comparison.
     if (info.target_subscriptions_by_message_sequence.count(message_seq) == 0) {
       info.target_subscriptions_by_message_sequence.emplace(
@@ -263,7 +275,7 @@ public:
     if (!publisher) {
       throw std::runtime_error("publisher has unexpectedly gone out of scope");
     }
-    auto sub_map_it = subscription_ids_by_topic_.find(publisher->get_topic_name());
+    auto sub_map_it = subscription_ids_by_topic_.find(FixedSizeString(publisher->get_topic_name()));
     if (sub_map_it == subscription_ids_by_topic_.end()) {
       // No intraprocess subscribers
       return 0;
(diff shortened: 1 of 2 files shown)


ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6552/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2891/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5369/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6350/)

ivanpauno:
One the test failures in Linux is related with this PR: https://ci.ros2.org/job/ci_linux/6552/testReport/junit/(root)/projectroot/test_tlsf__rmw_connext_cpp/

It's an allocator test, which check that all "new" or "malloc" calls are using the custom allocator passed, or are being call from the underlying DDS in use.
I tried with the following patch locally:
```patch
+  using AllocString = std::basic_string<char, std::char_traits<char>, RebindAlloc<char>>;
   using IDTopicMap = std::map<
-    std::string,
+    AllocString,
     AllocSet,
-    std::less<std::string>,
-    RebindAlloc<std::pair<const std::string, AllocSet>>>;
+    std::less<AllocString>,
+    RebindAlloc<std::pair<const AllocString, AllocSet>>>;
```
But, it didn't solve the problem.
I don't know why it's only happening in connext, and why is not happening in Linux-aarch64 (the test is not being run in MacOs and Windows).

wjwwood:
I don't know either. Please keep looking into and if you're still stuck after a while I'll try to dig into too.

ivanpauno:
> I don't know either. Please keep looking into and if you're still stuck after a while I'll try to dig into too.

I have checked the stacktrace in test_tlsf and it's calling `new(unsigned long)` from the std::string constructor.
It's not failing in fastrtps and opensplice because test_tlsf is buggy. This line of the stacktrace is marked as `match`:
```
2:   /home/ivanpauno/ros2_ws/build/tlsf_cpp/test_tlsf__rmw_fastrtps_cpp : operator new(unsigned long)+0x3b
```
Only because it has fastrtps in the file name, and it shouldn't. 'fastrtps' and 'openslice' are valid tokens, but 'connext' is not in the token list. I will open an issue explaining the problem.

[Here](https://gist.github.com/ivanpauno/16115ac99e367cf44eff9e97d82da2b7) is the diff I used in rclcpp and tlsf_cpp when testing, and also my test output. The starcktrace with the unexpected new call is above the "BAD HERE". I haven't found a way to avoid the `new(unsigned long)` call.
I also tried writing a basic "string like class" which allocated the memory manually with the allocator without success.

ivanpauno:
As far as I see, IntraProcessManagerImpl is instantiated without passing an allocator to it. Am I right?
https://github.com/ros2/rclcpp/blob/4f2f8def9819a3e37216254a21e3471655529bc0/rclcpp/src/rclcpp/intra_process_manager_impl.cpp#L19-L23
https://github.com/ros2/rclcpp/blob/4f2f8def9819a3e37216254a21e3471655529bc0/rclcpp/include/rclcpp/intra_process_manager.hpp#L130-L132

If I'm right, I don't know why the other allocations done inside IntraProcessManagerImpl haven't been causing the same issue.

wjwwood:
It is created without an allocator, at least by default.

> If I'm right, I don't know why the other allocations done inside IntraProcessManagerImpl haven't been causing the same issue.

Perhaps because they only occur during creation and destruction of pubs or subs? The idea is to eliminate allocations in the common case of publish and take.

ivanpauno:
> Perhaps because they only occur during creation and destruction of pubs or subs? The idea is to eliminate allocations in the common case of publish and take.

It doesn't seem to be the case:
https://github.com/ros2/rclcpp/blob/4f2f8def9819a3e37216254a21e3471655529bc0/rclcpp/include/rclcpp/intra_process_manager_impl.hpp#L163-L195
Store message method is allocating memory in two other cases:
- `info.target_subscriptions_by_message_sequence.emplace` L181-182
- ` std::copy(...)` L186-194.

Both of the allocations are using the standard allocator, because the IntraProcessManagerImpl class template is specialized only once and using it. But those two calls don't causes a failure in test_tlsf. Maybe test_tlsf isn't working well?
With the introduced changes I'm also allocating an string, using a rebind of the same allocator (see [patch](https://github.com/ros2/rclcpp/pull/671#issuecomment-478084021)). So, I expect the same behavior in test_tlsf, but it isn't happening.
I've already found a bug in test_tlsf, explained [here](https://github.com/ros2/realtime_support/issues/73), because it should also be failing in fastrtps and opensplice.

wjwwood:
> Store message method is allocating memory in two other cases:

I could be wrong, but if the destination is larger enough already then neither of those functions will necessarily allocate memory. The `emplace` may essentially do a placement new on the existing memeory, and the `std::copy` may copy the values from source to destination with allocation.

> But those two calls don't causes a failure in test_tlsf. Maybe test_tlsf isn't working well?

It's certainly possible. I haven't looked at that test in a long time, and didn't write it originally.

> I've already found a bug in test_tlsf, explained here, because it should also be failing in fastrtps and opensplice.

Yes, but that doesn't explain why the other allocations (that you think already exist) do not fail with Connext, right?

ivanpauno:
> Yes, but that doesn't explain why the other allocations (that you think already exist) do not fail with Connext, right?

Yes, it's not explaining that. The test is actually doing only one publish, probably it would fail at some point if we loop more. The allocation is only checked inside the publishing loop, so probably that memory was reserved before. I will try to dig into this a little bit more on Wednesday.

IMHO, the bug this PR is correcting is more serious than if the allocator test fails or not, so I would merge it and solve the test_tlfs failure in a follow-up PR if you also agree.

wjwwood:
Well, it would be better to fix the bug without regressing the test. There is a limit on topic name length, we could instead use a fixed sized string for the key to avoid the need for a `std::string`.

ivanpauno:
> Well, it would be better to fix the bug without regressing the test. There is a limit on topic name length, we could instead use a fixed sized string for the key to avoid the need for a `std::string`.

That sounds reasonable. I will work on this on Wednesday.

ivanpauno:
It's passing test_tlsf with the last changes locally.
Here full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6586/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2924/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5403/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6406/)

rclcpp/include/rclcpp/intra_process_manager_impl.hpp
```diff
@@ -175,7 +189,8 @@ class IntraProcessManagerImpl : public IntraProcessManagerImplBase
     }
 
     // Figure out what subscriptions should receive the message.
-    auto & destined_subscriptions = subscription_ids_by_topic_[publisher->get_topic_name()];
+    auto & destined_subscriptions =
+      subscription_ids_by_topic_[FixedSizeString(publisher->get_topic_name())];
```
wjwwood: I never noticed this before, and it's out of scope for this pr, but this is a really expensive operation to do in `store_intra_process_message`. I'll have to think about a better way to do this that doesn't require a string compare based search in the common case.

rclcpp/include/rclcpp/intra_process_manager_impl.hpp
```diff
@@ -15,6 +15,8 @@
 #ifndef RCLCPP__INTRA_PROCESS_MANAGER_IMPL_HPP_
 #define RCLCPP__INTRA_PROCESS_MANAGER_IMPL_HPP_
 
+#include <rmw/validate_full_topic_name.h>
```
wjwwood: This include should go after the system includes, but before the project (`rclcpp`) includes.

wjwwood: Basically line 32, in its own block.

rclcpp/include/rclcpp/intra_process_manager_impl.hpp
hidmic: Sorry to disrupt, but what about a `std::unordered_map` i.e. a hash map?

wjwwood: I was thinking more along the lines of storing this information so that it can be accessed by id (integer), as it only changes when something is added or removed. An `unordered_map` would be an improvement on the existing approach, however.

ivanpauno: I agree that  comparing strings is too expensive. I will only solve the bug in this PR.

wjwwood: > I agree that comparing strings is too expensive. I will only solve the bug in this PR.

👍

ivanpauno:
It has passed all the test locally.
Full CI again:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6592/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2930/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5​41​1/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6​41​5/)

ivanpauno:
MacOS and Windows jobs failed randomly. Again:

* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5415/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6423/)

ivanpauno:
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5417/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6425/)

ivanpauno:
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5​42​3/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6431/)

ivanpauno:
It's finally passing CI, merging!

dirk-thomas:
@ivanpauno This change introduced compiler warning on Windows debug: https://ci.ros2.org/job/ci_windows/6441/warnings43Result/ Please address them asap

ivanpauno:
> @ivanpauno This change introduced compiler warning on Windows debug: https://ci.ros2.org/job/ci_windows/6441/warnings43Result/ Please address them asap

Sorry, I didn't see that "MSBuild Warnings" wasn't empty.
About the problem itself, std::strncpy is deprecated in MS and strncpy_s hasn't been implemented in gcc. I have checked some discussions about this problem, and a "correct solution" doesn't exist. I have three options:
- Use preprocessor to call strncpy_s (string.h) in windows and std::strcpy (cstring) in linux/mac.
- Replace it with:
  `memcpy(ret.data() , str, std::strlen(str)+1 )`
  The size of ret is enough to always copy the entire "str", so no bound checking is needed.
- Code the loop.

I slightly prefer the first option. Do you prefer any of them?

clalancette:
We've typically used `memcpy` to get around this warning, so I'd suggest sticking with that.

PR #667, opened 2019-03-25
Add function to get publisher actual qos settings

ivanpauno:
Implements https://github.com/ros2/rmw/pull/169 at rclcpp level.
Related rcl implementation: https://github.com/ros2/rcl/pull/406

Connects to ros2/rmw#169

PR #674, opened 2019-03-29
Check QoS policy when configuring intraprocess, skip interprocess publish when possible

ivanpauno:
Follow up of https://github.com/ros2/rclcpp/pull/664/commits
Which was unintentionally merged into https://github.com/ros2/rclcpp/pull/667/

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -65,14 +65,14 @@ namespace rclcpp
 {
 
 /// Used as argument in create_publisher and create_subscriber.
-enum class IntraprocessSetting
+enum class IntraProcessSetting
 {
   /// Explicitly enable intraprocess comm at publisher/subscription level.
-  ENABLE,
+  Enable,
   /// Explicitly disable intraprocess comm at publisher/subscription level.
-  DISABLE,
+  Disable,
   /// Take intraprocess configuration from the node.
-  DEFAULT_FROM_NODE
+  NodeDefault
 };
 
 /// Node is the single point of entry for creating publishers and subscribers.
@@ -164,7 +164,7 @@ public:
   create_publisher(
     const std::string & topic_name, size_t qos_history_depth,
     std::shared_ptr<Alloc> allocator = nullptr,
-    IntraprocessSetting use_intra_process_comm = IntraprocessSetting::DEFAULT_FROM_NODE);
+    IntraProcessSetting use_intra_process_comm = IntraProcessSetting::NodeDefault);
 
   /// Create and return a Publisher.
   /**
@@ -181,7 +181,7 @@ public:
     const std::string & topic_name,
     const rmw_qos_profile_t & qos_profile = rmw_qos_profile_default,
     std::shared_ptr<Alloc> allocator = nullptr,
-    IntraprocessSetting use_intra_process_comm = IntraprocessSetting::DEFAULT_FROM_NODE);
+    IntraProcessSetting use_intra_process_comm = IntraProcessSetting::NodeDefault);
 
   /// Create and return a Subscription.
   /**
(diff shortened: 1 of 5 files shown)


wjwwood:
I think by "over" you mean "into", "over" sounds like "as opposed to" or "instead of"?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -64,6 +64,17 @@
 namespace rclcpp
 {
 
+/// Used as argument in create_publisher and create_subscriber.
+enum class IntraprocessSetting
+{
+  /// Explicitly enable intraprocess comm at publisher/subscription level.
+  ENABLE,
```
wjwwood: Let's not use ALL_CAPS, as that is often used by macros, and could collide here. (imagine someone did `#define ENABLE ...`, which is not unthinkable.

Instead, let's do `Enable`, `Disable`, and `Default` (or `DefaultFromNode` but that doesn't look very nice to me, another option is maybe `NodeDefault`).

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -81,17 +83,30 @@ template<typename MessageT, typename Alloc, typename PublisherT>
...
 
+  bool use_intra_process;
+  switch (use_intra_process_comm) {
+    case IntraprocessSetting::ENABLE:
+      use_intra_process = true;
+      break;
+    case IntraprocessSetting::DISABLE:
+      use_intra_process = false;
+      break;
+    default:  // IntraprocessSetting::DEFAULT_FROM_NODE
```
wjwwood: This should be its own case instead, and an error should be thrown if yet something else is used.

rclcpp/src/rclcpp/publisher.cpp
```diff
@@ -231,6 +230,19 @@ PublisherBase::setup_intra_process(
   IntraProcessManagerSharedPtr ipm,
   const rcl_publisher_options_t & intra_process_options)
 {
+  // Skip intraprocess configuration if the "durability" qos policy is not "volatile".
+  // TODO(ivanpauno):
+  //    RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT is for FastRTPS and Connext also
+  //    equal to "volatile". It would be good to add a
+  //    is_durability_policy_volatile(rcl_publisher_options_t options)
+  //    method to rcl, and corresponging methods in rmw and the rmw vendor
+  //    implementations, which will also checks if "system_default" is "volatile" or not.
```
wjwwood: But isn't this exactly what your other pr's are aiming to do?

ivanpauno: I will update when I merge them

rclcpp/include/rclcpp/node.hpp
ivanpauno: Up to now, most of the "enum class" in rclcpp follow the ALL_CAPS style, except of `CallbackGroupType`. I agree we should only use it in macros, so I will change it.

wjwwood: Yeah, there might be other cases that should change, but let's arc towards what we think is right.

rclcpp/src/rclcpp/publisher.cpp
ivanpauno: Sorry that I wasn't clear. But I'm not thinking to merge this before the other PRs. I'm now going to update this.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6585/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2923/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5402/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6405/)

ivanpauno:
[Here] is the characterization of rclcpp comm with this improvement. It's working as expected.
I've also tried the intraprocess demos, and I've checked that interprocess subscriptions receive messages as expected if the publisher is configured with `use_intra_process(true)` option.
@wjwwood Do you think this is ready to be merged?

wjwwood:
I left some comments on the word document, if you could address them. But that document doesn't really apply to this pull request (it is needed, but doesn't tell us if this pr is ready or not).

ivanpauno:
> I left some comments on the word document, if you could address them. But that document doesn't really apply to this pull request (it is needed, but doesn't tell us if this pr is ready or not).

Yes sure. I ran it to check if this improves performance or not.
I'm merging now. I will end addressing the document comments tomorrow.

Issue #640, opened 2019-02-27
std::unique_ptr<> as input parameter instead of std::unique_ptr<>& in publish function


thongpv3:
Why don't use `std::unique_ptr<>` as input parameter instead of `std::unique_ptr<>&` in this function:

https://github.com/ros2/rclcpp/blob/8743bcb0a1ef06d6eea5ebbd247e019f9dbb1e10/rclcpp/include/rclcpp/publisher.hpp#L216-L253

I think it is strange because the the parameter is input parameter, not an input and output parameter. So we can use `std::unique_ptr<>` instead. Here is the guideline: https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

sloretz:
I don't know the reason. Maybe it was added to avoid needing to type `std::move()`?

```c++
#include <memory>

void unique_func_ref(std::unique_ptr<int> & msg) {}
void unique_func_value(std::unique_ptr<int> msg) {}

int main(int argc, char ** argv)
{
   std::unique_ptr<int> foo(new int);
   std::unique_ptr<int> bar(new int);

   unique_func_ref(foo);
   // does not compile
   // unique_func_value(bar);
   unique_func_value(std::move(bar));

   return 0;
}
```

thongpv3:
The `std::unique_ptr` intend to be used like that. Anytime users want to pass the ownership, they should explicitly pass std::move, then they would know that this pointer is not valid to use.

thongpv3:
And I think this block of code also have a problem:
https://github.com/ros2/rclcpp/blob/8743bcb0a1ef06d6eea5ebbd247e019f9dbb1e10/rclcpp/include/rclcpp/publisher.hpp#L255-L306

We can just provide `virtual void publish(const MessageT & msg)` because we don't managed `std::shared_ptr`'s object lifetime. User can use this like `publish(*shared_ptr_to_object)`. There also increase performance because we don't need to copy a `std::shared_ptr`. And no need to copy paste things :)

sloretz:
@thongpv3 would you be willing to make a PR with the `&` removed?

wjwwood:
I think this has been resolved, see: https://github.com/ros2/rclcpp/pull/676#issuecomment-489381722

PR #676, opened 2019-03-29
remove unnecessary publish api on pulisher

thongpv3:
Solve https://github.com/ros2/rclcpp/issues/640
Signed-off-by: Pham Van Thong <thongpv87@gmail.com>

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -23,6 +23,7 @@
 #include <memory>
 #include <sstream>
 #include <string>
+#include <utility>
 
 #include "rcl/error_handling.h"
 #include "rcl/publisher.h"
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -186,7 +186,7 @@ public:
    * \param[in] msg A shared pointer to the message to send.
    */
   virtual void
-  publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
+  publish(std::unique_ptr<MessageT, MessageDeleter> msg)
   {
     this->do_inter_process_publish(msg.get());
     if (store_intra_process_message_) {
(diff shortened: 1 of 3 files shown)


rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -274,16 +233,7 @@ class Publisher : public PublisherBase
...
-  }
-
-  virtual void
-  publish(const MessageT * msg)
-  {
-    if (!msg) {
-      throw std::runtime_error("msg argument is nullptr");
-    }
-    return this->publish(*msg);
+    return this->publish(std::move(unique_msg));
```
sloretz: Looks like CI wants another include for `std::move`

http://build.ros2.org/job/Cpr__rclcpp__ubuntu_bionic_amd64/195/testReport/junit/rclcpp/cpplint/build_include_what_you_use__4____tmp_ws_src_rclcpp_rclcpp_include_rclcpp_publisher_hpp_236_/

```
Add #include <utility> for move
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -233,7 +233,7 @@ NodeParameters::set_parameters_atomically(
 
   std::swap(tmp_map, parameters_);
   parameter_event->stamp = node_clock_->get_clock()->now();
-  events_publisher_->publish(parameter_event);
+  events_publisher_->publish(*parameter_event);
```
wjwwood: It would be better if this kind of change were not needed.

thongpv3: Do you mean that it would be better if the interface accept both pointer and reference

wjwwood: Yeah, specifically it should continue to take a `shared_ptr`, either by reference or value, so that this code doesn't need to change.

thongpv3: I think the API should be hard to misuse. Convenience to use is good to have but it shouldn't confuse the user.
- When you receive a `std::shared_ptr<>`, it mean that you will extend it lifetime (may be you will need to hold the reference of the object in another thread). Do you need it in this case? Remember `std::shared_ptr` is not free.
- `T const *` is good enough but `T const &` is better in this case, because the pointer will accept a `nullptr`, and in the function `publish` will throw an exception when the argument is `nullptr` in this case.

wjwwood: What's the problem (i.e. how will it be misused?) with `const std::shared_ptr<T> &`? It neither implies extending the lifetime (though it could if needed) nor does it incur the cost of copying a shared pointer.

In the future we may want to make a copy of the shared pointer for use in the intra-process manager, so it's convenient to have in the future and right now, many people choose to create shared pointers to (using `std::make_shared`) messages for various reasons, e.g. like having a pool of reusable messages.

thongpv3: `const std::shared_ptr<T> &` don't have lifetime problem, only `publish(std::shared_ptr<const MessageT> msg)` has problem. But if any time you have these pointer, you can make a call to `T const&`, so you will not need these interface. I think `*` is not a big typing problem.

and `const std::shared_ptr<T> &` can make user confuse that this function can modify their variable.

wjwwood: What about `const std::shared_ptr<const T> & ptr`?

The goal here is to not be too disruptive to existing code and to allow people who are using shared pointers to messages to publish easily. I take the point that `*` is not hard to type, but it is a continuing cost, whereas us providing one more signature is also not a burden and a one time cost (for the most part).

So if we can come up with a signature that accurately reflects our intent and is not misleading then we should do that I think.

wjwwood:
This pr cannot be done in crystal. It will have to be targeted at Dashing, since it is removing API calls it is not API compatible with crystal.

wjwwood:
@thongpv3 thanks for proposing this change, after all we ended up agreeing with you (I think) and in Dashing the plan is to only have three non-deprecated publish signatures, and those would be:

- `publish(std::unique_ptr<MessageT>)`
- `publish(const MessageT &)`
- `publish(const rcl_serialized_message_t &)`

See: https://github.com/ros2/rclcpp/pull/709

PR #677, opened 2019-03-30
Rename action state transitions

jacobperron:
Now using active verbs as described in the design doc:

http://design.ros2.org/articles/actions.html#goal-states

Connects to ros2/rcl#399.


jacobperron:
Updated based on design discussion (https://github.com/ros2/design/pull/226). @hidmic, please take another look.

Issue #647, opened 2019-03-04
change throw_from_rcl_error to throw from_rcl_error


sloretz:
## Feature request

Currently rclcpp has a function to throw an exception based on an rcl error. 

https://github.com/ros2/rclcpp/blob/3919ab189795fbf47b5427add06f45d8a43f81ce/rclcpp/src/rclcpp/clock.cpp#L134

This is a request to change that function to return an exception without throwing.

```c++
throw rclcpp::exceptions::from_rcl_error(ret, "Failed to add time jump callback");
```
I think this could help static analysis tools like `cppcheck` recognize when code won't be executed due to an exception. This would remove the need for ros2/rclcpp#646

#### Implementation considerations
`throw_from_rcl_error` could be refactored to use `from_rcl_error` rather than changing all the code at once.

wjwwood:
Seems reasonable, at the very least to factor the creation of the exception and the throw into separate functions.

jhdcs:
Mind if I give this a go? I was thinking of renaming "throw_from_rcl_error" to either "from_rcl_error" or "gen_rcl_exception", then refactoring everything that called the original function to "throw <from_rcl_error>". Unless, of course, you want to do something besides throwing at any of the locations.

Also, out of curiosity, is there a reason we aren't using enums for return codes? And should I put that in a separate question?

wjwwood:
> Mind if I give this a go?

Sure, please have a go at it.

> I was thinking of renaming "throw_from_rcl_error" to either "from_rcl_error" or "gen_rcl_exception", then refactoring everything that called the original function to "throw <from_rcl_error>". Unless, of course, you want to do something besides throwing at any of the locations.

I think `from_rcl_error` is better than `gen_rcl_exception` myself, but `exception_from_rcl_error` would be another alternative that I think. Though in context, `throw rclcpp::exceptions::from_rcl_error` is pretty clear already. On the other hand, if the user does `using rclcpp::exceptions::from_rcl_error` and later does `throw from_rcl_error(...)` it's a bit less clear and `throw exception_from_rcl_error(...)` would be clearer in that case.

I don't mind which you do, but the already proposed ones seem like the best name I've heard so far.

Also, consider the line from the description:

> `throw_from_rcl_error` could be refactored to use `from_rcl_error` rather than changing all the code at once.

Avoiding touching all the existing code would be a nice thing to have, while still enabling users to generate the exception without throwing it in cases where that's needed.

----

[...]

PR #646, opened 2019-03-04
Attempt to fix cppcheck warning

sloretz:
connects to ros2/rmw_connext#348

ros2/build_cop#172

PR #678, opened 2019-04-01
[WIP] Exception Generator function for implementing "from_rcl_error"

jhdcs:
The only way I found to send along exceptions was to make the function wrap the generated exception in a pointer. Which means that you'd have to use std::rethrow_exception instead of just a simple throw to use it.

Still, is this about what you were looking for? And do you have any specific areas that need this functionality?

Connects to issue #647 

Signed-off-by: Jacob Hassold <jhassold@dcscorp.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+.DS_Store
+
+*.swp
+
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 .DS_Store
+
+*.swp
+
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,0 @@
-.DS_Store
-
-*.swp
-tags
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .DS_Store
 
 *.swp
-
 tags
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,1 @@
 .DS_Store
-
-*.swp
-
-tags
--- a/rclcpp/src/rclcpp/exceptions.cpp
+++ b/rclcpp/src/rclcpp/exceptions.cpp
@@ -46,31 +46,11 @@ throw_from_rcl_error(
   const rcl_error_state_t * error_state,
   void (* reset_error)())
 {
-  if (RCL_RET_OK == ret) {
-    throw std::invalid_argument("ret is RCL_RET_OK");
-  }
-  if (!error_state) {
-    error_state = rcl_get_error_state();
-  }
-  if (!error_state) {
-    throw std::runtime_error("rcl error state is not set");
-  }
-  std::string formated_prefix = prefix;
-  if (!prefix.empty()) {
-    formated_prefix += ": ";
-  }
-  RCLErrorBase base_exc(ret, error_state);
-  if (reset_error) {
-    reset_error();
-  }
-  switch (ret) {
-    case RCL_RET_BAD_ALLOC:
-      throw RCLBadAlloc(base_exc);
-    case RCL_RET_INVALID_ARGUMENT:
-      throw RCLInvalidArgument(base_exc, formated_prefix);
-    default:
-      throw RCLError(base_exc, formated_prefix);
-  }
+  // We expect this to either throw a standard error,
+  // or to generate an error pointer (which is caught
+  // in err, and immediately thrown)
+  auto err = from_rcl_error(rt, prefix, error_state, reset_error);
+  std::rethrow_exception(err);
 }
 
 std::exception_ptr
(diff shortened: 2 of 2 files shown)


wjwwood:
> Still, is this about what you were looking for? And do you have any specific areas that need this functionality?

The description of the original issue describes the use case:

https://github.com/ros2/rclcpp/issues/647

> I think this could help static analysis tools like cppcheck recognize when code won't be executed due to an exception. This would remove the need for #646

@sloretz since you raised the feature request original can you comment?

rclcpp/src/rclcpp/exceptions.cpp
```diff
@@ -73,6 +73,40 @@ throw_from_rcl_error(
   }
 }
 
+  std::exception_ptr
```
sloretz: Oops, when I wrote #647 I didn't realize returning a base exception type only triggers the copy constructor of the base type, and the derived exception type would be lost. Returning a pointer makes sense to me now.

rclcpp/src/rclcpp/exceptions.cpp
```diff
@@ -73,6 +73,40 @@ throw_from_rcl_error(
...
+  from_rcl_error(
+    rcl_ret_t ret,
+    const std::string & prefix,
+    const rcl_error_state_t * error_state,
+    void (* reset_error)())
+  {
+    if (RCL_RET_OK == ret) {
+      return std::make_exception_ptr(std::invalid_argument("ret is RCL_RET_OK"));
```
sloretz: This exception is different since it is a problem with the arguments passed to `from_rcl_error` instead of an exception created to match `ret`. Recommend `throw std::invalid_argument("...")` instead.

rclcpp/src/rclcpp/exceptions.cpp
```diff
@@ -73,6 +73,40 @@ throw_from_rcl_error(
...
+    }
+    if (!error_state) {
+      error_state = rcl_get_error_state();
+    }
+    if (!error_state) {
+      return std::make_exception_ptr(std::runtime_error("rcl error state is not set"));
```
sloretz: Same comment about `throw std::runtime_error(...)` instead of returning the exception here since the exception didn't come from the value of `ret`.

jhdcs:
All right, I've reverted the two cases you mentioned.

jhdcs:
So where do we stand on this? Do we still want to pull this into the main repo? Or just stick it on the back-burner?

sloretz:
@jhdcs this looks good to me. Is it possible to make `throw_from_rcl_error` use this function internally using `std::retrhow_exception` without changing the API? If so, would you mind doing so?

jhdcs:
@sloretz Apologies for the late reply. It looks perfectly possible to call "from_rcl_error" from "throw_from_rcl_error". I'm trying to discover if there's any unforseen consequences before I push it to my fork...

jhdcs:
I swear I will get the hang of this someday...

In any case, @sloretz, was that what you were looking for? Apart from the eldritch spawn that is my attempting to sign commits properly?

sloretz:
> In any case, @sloretz, was that what you were looking for?

Yeah, that looks great!

.gitignore
```diff
@@ -1 +1,4 @@
 .DS_Store
+
+*.swp
```
sloretz: One last request, mind removing the `.gitignore` change? Instead I recommend using [`git config --global  core.excludesfile ...`](https://stackoverflow.com/questions/7335420/global-git-ignore). For example, I have `.*.sw?` for vim files in `~/.gitignore` on my machine.

jhdcs: Sure. Should I just revert that commit?

jhdcs:
For some reason, the Dpr_rclpp_ubuntu_bionic_amd64 check seems to be failing. I'm not sure why, though I'm kinda new to this.
It's saying something about "GOAL_EVENT_ABORT" not being declared in-scope inside of rclcpp_action/src/server_goal_action.cpp. However, I don't remember having modified those files.

Did I cross-contaminate something with my commits?

sloretz:
> Did I cross-contaminate something with my commits?

Probably not. That job builds `rclcpp` against `crystal` release, so it is probably failing due to an API change on master.

jhdcs:
> > Did I cross-contaminate something with my commits?
> 
> Probably not. That job builds `rclcpp` against `crystal` release, so it is probably failing due to an API change on master.

So I need to find where in the API it's failing? Or is it out of our hands?

sloretz:
CI (Building all, testing rclcpp, test_rclcpp)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6703/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3019/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5508/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6534/)

sloretz:
> So I need to find where in the API it's failing? Or is it out of our hands?

No action needed. That build can be ignored. I started CI in the comment above.

PR #684, opened 2019-04-05
Replaced strncpy with memcpy

ivanpauno:
Solves [windows ci warning](https://ci.ros2.org/job/ci_windows/6441/warnings43Result/), as requested [here](https://github.com/ros2/rclcpp/pull/671#issuecomment-480141337).

### DIFF
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -283,7 +283,7 @@ private:
   fixed_size_string(const char * str) const
   {
     FixedSizeString ret;
-    std::strncpy(ret.data(), str, ret.size());
+    std::memcpy(ret.data(), str, ret.size());
     return ret;
   }
 
--- a/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager_impl.hpp
@@ -25,9 +25,12 @@
 #include <memory>
 #include <mutex>
 #include <set>
+#include <stdexcept>
 #include <string>
 #include <unordered_map>
 #include <utility>
+
+#include "rcutils/snprintf.h"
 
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/intra_process_manager_impl.hpp
```diff
@@ -283,7 +283,7 @@ class IntraProcessManagerImpl : public IntraProcessManagerImplBase
   fixed_size_string(const char * str) const
   {
     FixedSizeString ret;
-    std::strncpy(ret.data(), str, ret.size());
```
dirk-thomas: The previous code already looks wrong. I always copies `ret.size()` bytes - even if `str` is shorter much implies copying undefined data.

This should check the `strlen(str)`, ensure that it is <= `RMW_TOPIC_MAX_NAME_LENGTH` and only copy its actual data.

ivanpauno: Actually, std::strncpy pads with zeros after the '\0'. I'm using now rcutils_snprintf, which stops copying in the null terminator and solves the same portability issue.
I'm adding the strcmp_wrapper again, as the lexicographical comparison of std::array doesn't stop in the null terminator. I didn't realize before, and as std::strncpy pads with zeros, it was passing.

dirk-thomas: Why use the complexity of `snprintf` if all needed is a 1-to-1 copy?

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6625/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2951/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5436/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6446/)

Issue #683, opened 2019-04-04
[Crystal] rclcpp_action test_client timeout


nuclearsandwich:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04, macOS 10.12, Windows 10
- Installation type:
  - Source
- Version or commit hash:
  - https://github.com/ros2/ros2/blob/crystal-patch-4/ros2.repos
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

The following test timeout occurs in CI for the latest release candidate of ROS 2 Crystal on all platforms. I've also reproduced the timeout locally in an Ubuntu 18.04 container and run the test successfully in a Crystal 20190314 workspace in the same container.

```
test 1
    Start 1: test_client

1: Test command: /home/jenkins-agent/workspace/ci_linux/venv/bin/python3 "-u" "/home/jenkins-agent/workspace/ci_linux/ws/install/ament_cmake_test/share/ament_cmake_test/cmake/run_test.py" "/home/jenkins-agent/workspace/ci_linux/ws/build/rclcpp_action/test_results/rclcpp_action/test_client.gtest.xml" "--package-name" "rclcpp_action" "--output-file" "/home/jenkins-agent/workspace/ci_linux/ws/build/rclcpp_action/ament_cmake_gtest/test_client.txt" "--command" "/home/jenkins-agent/workspace/ci_linux/ws/build/rclcpp_action/test_client" "--gtest_output=xml:/home/jenkins-agent/workspace/ci_linux/ws/build/rclcpp_action/test_results/rclcpp_action/test_client.gtest.xml"
1: Test timeout computed to be: 60
1: -- run_test.py: invoking following command in '/home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rclcpp/rclcpp_action':
1:  - /home/jenkins-agent/workspace/ci_linux/ws/build/rclcpp_action/test_client --gtest_output=xml:/home/jenkins-agent/workspace/ci_linux/ws/build/rclcpp_action/test_results/rclcpp_action/test_client.gtest.xml
1: Running main() from gtest_main.cc
1: [==========] Running 7 tests from 1 test case.
1: [----------] Global test environment set-up.
1: [----------] 7 tests from TestClient
1: [ RUN      ] TestClient.construction_and_destruction
1: [       OK ] TestClient.construction_and_destruction (30 ms)
[...]

PR #685, opened 2019-04-05
Wait for server to be ready in action client tests

jacobperron:
This should address #683 once backported to `crytsal`

### DIFF
--- a/rclcpp_action/test/test_client.cpp
+++ b/rclcpp_action/test/test_client.cpp
@@ -273,6 +273,8 @@ TEST_F(TestClient, async_send_goal_but_ignore_feedback_and_result)
   auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
   ASSERT_TRUE(action_client->wait_for_action_server(WAIT_FOR_SERVER_TIMEOUT));
 
+  ASSERT_TRUE(action_client->wait_for_action_server(std::chrono::seconds(10)));
+
   ActionGoal bad_goal;
   bad_goal.order = -5;
   auto future_goal_handle = action_client->async_send_goal(bad_goal, nullptr, true);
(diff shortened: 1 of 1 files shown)


jacobperron:
:crossed_fingers: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6626/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2952/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5437/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6447/)

/cc @nuclearsandwich

rclcpp_action/test/test_client.cpp
```diff
@@ -273,6 +273,8 @@ TEST_F(TestClient, async_send_goal_but_ignore_feedback_and_result)
   auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
   ASSERT_TRUE(action_client->wait_for_action_server(WAIT_FOR_SERVER_TIMEOUT));
 
+  ASSERT_TRUE(action_client->wait_for_action_server(std::chrono::seconds(10)));
```
sloretz: Does adding 10 seconds to `WAIT_FOR_SERVER_TIMEOUT` on the line above this one also fix the issue?

jacobperron: XD I didn't even see that.

jacobperron: I'll try.

jacobperron: So my grep skills failed me.

It looks like it is already waiting for 10 seconds... odd that an additional wait causes the tests to pass :slightly_frowning_face:

jacobperron: The tests were not actually waiting anywhere close to 10 seconds either.

jacobperron: oooh, my bad. I was working off of the crystal branch, which is missing the wait calls.

jacobperron:
See #686 instead

PR #688, opened 2019-04-08
Created `on_parameter_event` static function

alsora:
Currently, the `TimeSource` class creates an object of type `AsyncParametersClient` with the only purpose of subscribing to the parameter events using the `on_parameter_event` method.

However, the `AsyncParametersClient` is not necessary for doing this, as the `on_parameter_event` creates a regular subscription.

In this PR I added a static implementation for `on_parameter_event` that allows to not create the `AsyncParameterClient` object.
There may be different alternative implementation for decoupling the `on_parameter_event` from the ParameterClient... Let me know if you think that something else would be better than the static methods I used.

If we agree on this change and on how to implement it, I will look into the code if there are other places where to propagate it.

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -121,40 +121,39 @@ public:
       rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
   ))
   {
-    return rclcpp::create_subscription<rcl_interfaces::msg::ParameterEvent>(
+    return this->on_parameter_event(
       this->node_topics_interface_,
+      callback,
+      qos,
+      options);
+  }
+
+  /**
+   * The NodeT type only needs to have a method called get_node_topics_interface()
+   * which returns a shared_ptr to a NodeTopicsInterface, or be a
+   * NodeTopicsInterface pointer itself.
+   */
+  template<
+    typename CallbackT,
+    typename AllocatorT = std::allocator<void>
+    typename NodeT>
+  static typename rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr
+  on_parameter_event(
+    NodeT && node,
+    CallbackT && callback,
+    const rclcpp::QoS & qos = (
+      rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_default))
+    ),
+    const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
+      rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
+  ))
+  {
+    return rclcpp::create_subscription<rcl_interfaces::msg::ParameterEvent>(
+      node,
       "parameter_events",
       qos,
       std::forward<CallbackT>(callback),
       options);
-  }
-
-  template<
-    typename CallbackT,
-    typename Alloc = std::allocator<void>,
-    typename SubscriptionT =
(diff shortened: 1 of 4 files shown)


mjcarroll:
@alsora Any way you could add some tests for this change?

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -134,6 +134,34 @@ class AsyncParametersClient
...
+  on_parameter_event(
+    CallbackT && callback,
+    rclcpp::node_interfaces::NodeTopicsInterface * node_topics)
+  {
+    using rclcpp::message_memory_strategy::MessageMemoryStrategy;
+    auto msg_mem_strat =
+      MessageMemoryStrategy<rcl_interfaces::msg::ParameterEvent, Alloc>::create_default();
+
+    using rcl_interfaces::msg::ParameterEvent;
+    return rclcpp::create_subscription<
```
mjcarroll: This is going to need to be updated to use `SubscriptionOptions`

alsora: Updated to the new subscritpions API.
Sorry for the several force-push, but I had some issues with DCO

mjcarroll:
@alsora I think that this looks good, but I would prefer to see some test coverage as well exercising the new functionality.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7102/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3324/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5818/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6922/)

mjcarroll:
@alsora I think that his may need another look, it's breaking `TestTimeSource` on Linux, as well as some linting issues.

dirk-thomas:
> it's breaking TestTimeSource on Linux

@mjcarroll Aren't these already failing on master, see ros2/build_cop#196?

mjcarroll:
You are correct, just the linting issues, then.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7109/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3329/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5824/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6930/)

alsora:
I added few basic tests for parameter clients and for creating parameter events subscription

dirk-thomas:
New round of CI builds:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7115/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3333/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5828/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6934/)

> I noticed that test_parameter_server is failing.. Is it "normal" or is it caused by my changes?

I would lean towards no, these failures already happen on master. But maybe @mjcarroll can double check?

mjcarroll:
We discussed this issue a bit offline today.  This can be considered an incremental improvement over the current state because it removes the need for the `TimeSource` to contain an `AsyncParametersClient` in favor of a regular subscription.  
This is not, however, sufficient to address (what I believe to be) the root cause of the issue, which is summarized in rclcpp#609.

mjcarroll:
After another look, the current test failures in CI are unrelated to this change.

PR #690, opened 2019-04-10
New IntraProcessManager capable of storing shared_ptr<const T>

ivanpauno:
This is an enhancement of the IntraProcessManager, that allows storing both unique_ptr<T> and shared_ptr<const T>. Design draft [here](https://docs.google.com/document/d/1dM8O34GM-SbFps1vd1rocPMAg_7O9_-OG4NF4axOz7E/edit#heading=h.3q18ubcdx3fd).
It shows a better performance, specially in the case of many subscriptions sharing ownership of the same const message. Document with performance results [here](https://docs.google.com/document/d/1XldfcBP0C8Zb8e4wMcks4zcfvzzE8CRhDPKwn-ZGGNs/edit).

##### List of changes
- MappedRingBuffer can store both kind of smart pointers. push_and_replace, get, and pop methods for both signature.
- MappedRingBuffer updated, covering the new implementation.
- Added store_intra_process_message and take_intra_process_message methods with both signatures to the IntraProcessManager.
- Modified Publisher publish methods to use this new IntraProcessManager.
- Modified Subscription handle_intra_process_message method.
- Modified AnySubscriptionCallback dispatch_intra_process method.
- Stopped bounding intra-process related methods from SubscriptionFactory and PublisherFactory in Subscription and Publisher class.
- Split publisher.h and subscriber.h in two files, one with the base class and the other with the derived.
  The idea is to avoid a circular inclusion with intra_process_manager.h

##### To do
- The intra_process_manager test has been corrected only to make it pass, because of the api changes. But it's actually not covering all the new IPM methods (which are well tested at mapped_ring_buffer level).

### DIFF
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -182,9 +182,9 @@ public:
   {
     (void)message_info;
     if (const_shared_ptr_callback_) {
-      const_shared_ptr_callback_(message);
+      const_shared_ptr_callback_(std::move(message));
     } else if (const_shared_ptr_with_info_callback_) {
-      const_shared_ptr_with_info_callback_(message, message_info);
+      const_shared_ptr_with_info_callback_(std::move(message), message_info);
     } else {
       if (unique_ptr_callback_ || unique_ptr_with_info_callback_ ||
         shared_ptr_callback_ || shared_ptr_with_info_callback_)
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -155,7 +155,6 @@ public:
   void dispatch(
     std::shared_ptr<MessageT> message, const rmw_message_info_t & message_info)
   {
-    (void)message_info;
     if (shared_ptr_callback_) {
       shared_ptr_callback_(message);
     } else if (shared_ptr_with_info_callback_) {
@@ -180,7 +179,6 @@ public:
   void dispatch_intra_process(
     ConstMessageSharedPtr & message, const rmw_message_info_t & message_info)
   {
-    (void)message_info;
     if (const_shared_ptr_callback_) {
       const_shared_ptr_callback_(std::move(message));
     } else if (const_shared_ptr_with_info_callback_) {
@@ -200,7 +198,6 @@ public:
   void dispatch_intra_process(
     MessageUniquePtr & message, const rmw_message_info_t & message_info)
   {
-    (void)message_info;
     if (shared_ptr_callback_) {
       typename std::shared_ptr<MessageT> shared_message = std::move(message);
       shared_ptr_callback_(shared_message);
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -221,10 +221,7 @@ public:
 
   bool use_take_shared_method()
   {
-    if (const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_) {
-      return true;
-    }
-    return false;
+    return const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_;
   }
 
 private:
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -36,6 +36,7 @@ class AnySubscriptionCallback
   using MessageAllocTraits = allocator::AllocRebind<MessageT, Alloc>;
   using MessageAlloc = typename MessageAllocTraits::allocator_type;
   using MessageDeleter = allocator::Deleter<MessageAlloc, MessageT>;
+  using ConstMessageSharedPtr = std::shared_ptr<const MessageT>;
   using MessageUniquePtr = std::unique_ptr<MessageT, MessageDeleter>;
 
   using SharedPtrCallback = std::function<void (const std::shared_ptr<MessageT>)>;
@@ -177,6 +178,26 @@ public:
   }
 
   void dispatch_intra_process(
+    ConstMessageSharedPtr & message, const rmw_message_info_t & message_info)
+  {
+    (void)message_info;
+    if (const_shared_ptr_callback_) {
+      const_shared_ptr_callback_(message);
+    } else if (const_shared_ptr_with_info_callback_) {
+      const_shared_ptr_with_info_callback_(message, message_info);
+    } else {
+      if (unique_ptr_callback_ || unique_ptr_with_info_callback_ ||
+        shared_ptr_callback_ || shared_ptr_with_info_callback_)
+      {
+        throw std::runtime_error("unexpected dispatch_intra_process const shared "
+                "message call with no const shared_ptr callback");
+      } else {
+        throw std::runtime_error("unexpected message without any callback set");
+      }
+    }
+  }
+
+  void dispatch_intra_process(
     MessageUniquePtr & message, const rmw_message_info_t & message_info)
   {
     (void)message_info;
@@ -186,19 +207,24 @@ public:
     } else if (shared_ptr_with_info_callback_) {
       typename std::shared_ptr<MessageT> shared_message = std::move(message);
       shared_ptr_with_info_callback_(shared_message, message_info);
-    } else if (const_shared_ptr_callback_) {
-      typename std::shared_ptr<MessageT const> const_shared_message = std::move(message);
-      const_shared_ptr_callback_(const_shared_message);
-    } else if (const_shared_ptr_with_info_callback_) {
-      typename std::shared_ptr<MessageT const> const_shared_message = std::move(message);
-      const_shared_ptr_with_info_callback_(const_shared_message, message_info);
     } else if (unique_ptr_callback_) {
       unique_ptr_callback_(std::move(message));
     } else if (unique_ptr_with_info_callback_) {
       unique_ptr_with_info_callback_(std::move(message), message_info);
+    } else if (const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_) {
+      throw std::runtime_error("unexpected dispatch_intra_process unique message call"
+              " with const shared_ptr callback");
     } else {
       throw std::runtime_error("unexpected message without any callback set");
     }
+  }
+
+  bool use_take_shared_method()
+  {
+    if (const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_) {
+      return true;
+    }
+    return false;
   }
 
 private:
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -177,12 +177,12 @@ public:
   }
 
   void dispatch_intra_process(
-    ConstMessageSharedPtr & message, const rmw_message_info_t & message_info)
+    ConstMessageSharedPtr message, const rmw_message_info_t & message_info)
   {
     if (const_shared_ptr_callback_) {
-      const_shared_ptr_callback_(std::move(message));
+      const_shared_ptr_callback_(message);
     } else if (const_shared_ptr_with_info_callback_) {
-      const_shared_ptr_with_info_callback_(std::move(message), message_info);
+      const_shared_ptr_with_info_callback_(message, message_info);
     } else {
       if (unique_ptr_callback_ || unique_ptr_with_info_callback_ ||
         shared_ptr_callback_ || shared_ptr_with_info_callback_)
@@ -196,7 +196,7 @@ public:
   }
 
   void dispatch_intra_process(
-    MessageUniquePtr & message, const rmw_message_info_t & message_info)
+    MessageUniquePtr message, const rmw_message_info_t & message_info)
   {
     if (shared_ptr_callback_) {
       typename std::shared_ptr<MessageT> shared_message = std::move(message);
--- a/rclcpp/include/rclcpp/intra_process_manager.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager.hpp
@@ -33,7 +33,7 @@
 #include "rclcpp/mapped_ring_buffer.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/publisher_base.hpp"
-#include "rclcpp/subscription.hpp"
+#include "rclcpp/subscription_base.hpp"
 #include "rclcpp/visibility_control.hpp"
 
 namespace rclcpp
@@ -361,8 +361,7 @@ public:
   }
 
   template<
-    typename MessageT, typename Alloc = std::allocator<void>,
-    typename Deleter = std::default_delete<MessageT>>
+    typename MessageT, typename Alloc = std::allocator<void>>
   void
   take_intra_process_message(
     uint64_t intra_process_publisher_id,
--- a/rclcpp/include/rclcpp/intra_process_manager.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager.hpp
@@ -247,7 +247,7 @@ public:
   uint64_t
   store_intra_process_message(
     uint64_t intra_process_publisher_id,
-    std::unique_ptr<MessageT, Deleter> & message)
+    std::shared_ptr<MessageT> & message)
   {
     using MRBMessageAlloc = typename std::allocator_traits<Alloc>::template rebind_alloc<MessageT>;
     using TypedMRB = typename mapped_ring_buffer::MappedRingBuffer<MessageT, MRBMessageAlloc>;
@@ -334,10 +334,81 @@ public:
     // Return a copy or the unique_ptr (ownership) depending on how many subscriptions are left.
     if (target_subs_size) {
       // There are more subscriptions to serve, return a copy.
-      typed_buffer->get_copy_at_key(message_sequence_number, message);
+      typed_buffer->get(message_sequence_number, message);
     } else {
       // This is the last one to be returned, transfer ownership.
-      typed_buffer->pop_at_key(message_sequence_number, message);
+      typed_buffer->pop(message_sequence_number, message);
+    }
+  }
+
+  /// Take an intra process message.
+  /**
+   * The intra_process_publisher_id and message_sequence_number parameters
+   * uniquely identify a message instance, which should be taken.
+   *
+   * The requesting_subscriptions_intra_process_id parameter is used to make
+   * sure the requesting subscription was intended to receive this message
+   * instance.
+   * This check is made because it could happen that the requester
+   * comes up after the publish event, so it still receives the notification of
+   * a new intra process message, but it wasn't registered with the manager at
+   * the time of publishing, causing it to take when it wasn't intended.
+   * This should be avioded unless latching-like behavior is involved.
+   *
+   * The message parameter is used to store the taken message.
+   * On the last expected call to this method, the ownership is transfered out
+   * of internal storage and into the message parameter.
+   * On all previous calls a copy of the internally stored message is made and
+   * the ownership of the copy is transfered to the message parameter.
+   * TODO(wjwwood): update this documentation when latching is supported.
+   *
+   * The message parameter can be set to nullptr if:
+   *
+   * - The publisher id is not found.
+   * - The message sequence is not found for the given publisher id.
+   * - The requesting subscription's id is not in the list of intended takers.
+   * - The requesting subscription's id has been used before with this message.
+   *
+   * This method may allocate memory to copy the stored message.
+   *
+   * \param intra_process_publisher_id the id of the message's publisher.
+   * \param message_sequence_number the sequence number of the message.
+   * \param requesting_subscriptions_intra_process_id the subscription's id.
+   * \param message the message typed shared_ptr used to return the message.
+   */
+  template<
+    typename MessageT, typename Alloc = std::allocator<void>,
+    typename Deleter = std::default_delete<MessageT>>
+  void
+  take_intra_process_message(
+    uint64_t intra_process_publisher_id,
+    uint64_t message_sequence_number,
+    uint64_t requesting_subscriptions_intra_process_id,
+    std::shared_ptr<MessageT> & message)
+  {
+    using MRBMessageAlloc = typename std::allocator_traits<Alloc>::template rebind_alloc<MessageT>;
+    using TypedMRB = mapped_ring_buffer::MappedRingBuffer<MessageT, MRBMessageAlloc>;
+    message = nullptr;
+
+    size_t target_subs_size = 0;
+    std::lock_guard<std::mutex> lock(take_mutex_);
+    mapped_ring_buffer::MappedRingBufferBase::SharedPtr buffer = impl_->take_intra_process_message(
+      intra_process_publisher_id,
+      message_sequence_number,
+      requesting_subscriptions_intra_process_id,
+      target_subs_size
+    );
+    typename TypedMRB::SharedPtr typed_buffer = std::static_pointer_cast<TypedMRB>(buffer);
+    if (!typed_buffer) {
+      return;
+    }
+    // Return a copy or the unique_ptr (ownership) depending on how many subscriptions are left.
+    if (target_subs_size) {
+      // There are more subscriptions to serve, return a copy.
+      typed_buffer->get(message_sequence_number, message);
+    } else {
+      // This is the last one to be returned, transfer ownership.
+      typed_buffer->pop(message_sequence_number, message);
     }
   }
 
--- a/rclcpp/include/rclcpp/intra_process_manager.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager.hpp
@@ -237,7 +237,7 @@ public:
   uint64_t
   store_intra_process_message(
     uint64_t intra_process_publisher_id,
-    const std::shared_ptr<const MessageT> & message)
+    std::shared_ptr<const MessageT> message)
   {
     using MRBMessageAlloc = typename std::allocator_traits<Alloc>::template rebind_alloc<MessageT>;
     using TypedMRB = typename mapped_ring_buffer::MappedRingBuffer<MessageT, MRBMessageAlloc>;
@@ -266,7 +266,7 @@ public:
   uint64_t
   store_intra_process_message(
     uint64_t intra_process_publisher_id,
-    std::unique_ptr<MessageT, Deleter> & message)
+    std::unique_ptr<MessageT, Deleter> message)
   {
     using MRBMessageAlloc = typename std::allocator_traits<Alloc>::template rebind_alloc<MessageT>;
     using TypedMRB = typename mapped_ring_buffer::MappedRingBuffer<MessageT, MRBMessageAlloc>;
@@ -279,7 +279,7 @@ public:
     }
 
     // Insert the message into the ring buffer using the message_seq to identify it.
-    bool did_replace = typed_buffer->push_and_replace(message_seq, message);
+    bool did_replace = typed_buffer->push_and_replace(message_seq, std::move(message));
     // TODO(wjwwood): do something when a message was displaced. log debug?
     (void)did_replace;  // Avoid unused variable warning.
 
--- a/rclcpp/include/rclcpp/intra_process_manager.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager.hpp
@@ -32,7 +32,7 @@
 #include "rclcpp/intra_process_manager_impl.hpp"
 #include "rclcpp/mapped_ring_buffer.hpp"
 #include "rclcpp/macros.hpp"
-#include "rclcpp/publisher.hpp"
+#include "rclcpp/publisher_base.hpp"
 #include "rclcpp/subscription.hpp"
 #include "rclcpp/visibility_control.hpp"
 
@@ -185,21 +185,11 @@ public:
    * \param buffer_size if 0 (default) a size is calculated based on the QoS.
    * \return an unsigned 64-bit integer which is the publisher's unique id.
    */
-  template<typename MessageT, typename Alloc>
+  RCLCPP_PUBLIC
   uint64_t
   add_publisher(
-    typename Publisher<MessageT, Alloc>::SharedPtr publisher,
-    size_t buffer_size = 0)
-  {
-    auto id = IntraProcessManager::get_next_unique_id();
-    size_t size = buffer_size > 0 ? buffer_size : publisher->get_queue_size();
-    auto mrb = mapped_ring_buffer::MappedRingBuffer<
-      MessageT,
-      typename Publisher<MessageT, Alloc>::MessageAlloc
-      >::make_shared(size, publisher->get_allocator());
-    impl_->add_publisher(id, publisher, mrb, size);
-    return id;
-  }
+    rclcpp::PublisherBase::SharedPtr publisher,
+    size_t buffer_size = 0);
 
   /// Unregister a publisher using the publisher's unique id.
   /**
@@ -247,7 +237,7 @@ public:
   uint64_t
   store_intra_process_message(
     uint64_t intra_process_publisher_id,
-    std::shared_ptr<const MessageT> & message)
+    const std::shared_ptr<const MessageT> & message)
   {
     using MRBMessageAlloc = typename std::allocator_traits<Alloc>::template rebind_alloc<MessageT>;
     using TypedMRB = typename mapped_ring_buffer::MappedRingBuffer<MessageT, MRBMessageAlloc>;
--- a/rclcpp/include/rclcpp/intra_process_manager.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager.hpp
@@ -25,6 +25,7 @@
 #include <memory>
 #include <mutex>
 #include <unordered_map>
+#include <utility>
 #include <set>
 
 #include "rclcpp/allocator/allocator_deleter.hpp"
@@ -242,8 +243,7 @@ public:
    * \return the message sequence number.
    */
   template<
-    typename MessageT, typename Alloc = std::allocator<void>,
-    typename Deleter = std::default_delete<MessageT>>
+    typename MessageT, typename Alloc = std::allocator<void>>
   uint64_t
   store_intra_process_message(
     uint64_t intra_process_publisher_id,
@@ -265,6 +265,35 @@ public:
     (void)did_replace;  // Avoid unused variable warning.
 
     impl_->store_intra_process_message(intra_process_publisher_id, message_seq);
+
+    // Return the message sequence which is sent to the subscription.
+    return message_seq;
+  }
+
+  template<
+    typename MessageT, typename Alloc = std::allocator<void>,
+    typename Deleter = std::default_delete<MessageT>>
+  uint64_t
+  store_intra_process_message(
+    uint64_t intra_process_publisher_id,
+    std::unique_ptr<MessageT, Deleter> & message)
+  {
+    using MRBMessageAlloc = typename std::allocator_traits<Alloc>::template rebind_alloc<MessageT>;
+    using TypedMRB = typename mapped_ring_buffer::MappedRingBuffer<MessageT, MRBMessageAlloc>;
+    uint64_t message_seq = 0;
+    mapped_ring_buffer::MappedRingBufferBase::SharedPtr buffer = impl_->get_publisher_info_for_id(
+      intra_process_publisher_id, message_seq);
+    typename TypedMRB::SharedPtr typed_buffer = std::static_pointer_cast<TypedMRB>(buffer);
+    if (!typed_buffer) {
+      throw std::runtime_error("Typecast failed due to incorrect message type");
+    }
+
+    // Insert the message into the ring buffer using the message_seq to identify it.
+    bool did_replace = typed_buffer->push_and_replace(message_seq, message);
+    // TODO(wjwwood): do something when a message was displaced. log debug?
+    (void)did_replace;  // Avoid unused variable warning.
+
+    impl_->store_intra_process_message(intra_process_publisher_id, std::move(message_seq));
 
     // Return the message sequence which is sent to the subscription.
     return message_seq;
@@ -341,41 +370,6 @@ public:
     }
   }
 
-  /// Take an intra process message.
-  /**
-   * The intra_process_publisher_id and message_sequence_number parameters
-   * uniquely identify a message instance, which should be taken.
-   *
-   * The requesting_subscriptions_intra_process_id parameter is used to make
-   * sure the requesting subscription was intended to receive this message
-   * instance.
-   * This check is made because it could happen that the requester
-   * comes up after the publish event, so it still receives the notification of
-   * a new intra process message, but it wasn't registered with the manager at
-   * the time of publishing, causing it to take when it wasn't intended.
-   * This should be avioded unless latching-like behavior is involved.
-   *
-   * The message parameter is used to store the taken message.
-   * On the last expected call to this method, the ownership is transfered out
-   * of internal storage and into the message parameter.
-   * On all previous calls a copy of the internally stored message is made and
-   * the ownership of the copy is transfered to the message parameter.
-   * TODO(wjwwood): update this documentation when latching is supported.
-   *
-   * The message parameter can be set to nullptr if:
-   *
-   * - The publisher id is not found.
-   * - The message sequence is not found for the given publisher id.
-   * - The requesting subscription's id is not in the list of intended takers.
-   * - The requesting subscription's id has been used before with this message.
-   *
-   * This method may allocate memory to copy the stored message.
-   *
-   * \param intra_process_publisher_id the id of the message's publisher.
-   * \param message_sequence_number the sequence number of the message.
-   * \param requesting_subscriptions_intra_process_id the subscription's id.
-   * \param message the message typed shared_ptr used to return the message.
-   */
   template<
     typename MessageT, typename Alloc = std::allocator<void>,
     typename Deleter = std::default_delete<MessageT>>
--- a/rclcpp/include/rclcpp/intra_process_manager.hpp
+++ b/rclcpp/include/rclcpp/intra_process_manager.hpp
@@ -247,7 +247,7 @@ public:
   uint64_t
   store_intra_process_message(
     uint64_t intra_process_publisher_id,
-    std::shared_ptr<MessageT> & message)
+    std::shared_ptr<const MessageT> & message)
   {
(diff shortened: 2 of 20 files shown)


rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -176,6 +177,26 @@ class AnySubscriptionCallback
...
+  void dispatch_intra_process(
+    ConstMessageSharedPtr & message, const rmw_message_info_t & message_info)
+  {
+    (void)message_info;
+    if (const_shared_ptr_callback_) {
+      const_shared_ptr_callback_(std::move(message));
+    } else if (const_shared_ptr_with_info_callback_) {
+      const_shared_ptr_with_info_callback_(std::move(message), message_info);
+    } else {
+      if (unique_ptr_callback_ || unique_ptr_with_info_callback_ ||
```
mjcarroll: In the case of storing a `ConstMessageSharedPtr` and dispatching to `unique_ptr`, would it make sense to do the copy to a unique ptr here?

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -186,21 +207,26 @@ class AnySubscriptionCallback
...
+      throw std::runtime_error("unexpected dispatch_intra_process unique message call"
+              " with const shared_ptr callback");
     } else {
       throw std::runtime_error("unexpected message without any callback set");
     }
   }
 
+  bool use_take_shared_method()
+  {
+    if (const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_) {
```
mjcarroll: Could be condensed to just `return (const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_);`

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -101,53 +102,91 @@ class MappedRingBuffer : public MappedRingBufferBase
    * \param value if the key is found, the value is stored in this parameter
    */
   void
-  get_copy_at_key(uint64_t key, ElemUniquePtr & value)
+  get(uint64_t key, ElemUniquePtr & value)
```
mjcarroll: Since you changed the method name here, is this actually changing the behavior?

We need to update the documentation here to describe the new behavior if so.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -101,53 +102,91 @@ class MappedRingBuffer : public MappedRingBufferBase
...
+    value.reset();
+    if (it != elements_.end() && it->in_use) {
+      if (!it->shared_value) {
+        it->shared_value = std::move(it->unique_value);
+      }
+      value = it->shared_value;
+    }
+  }
+
+  /// Return a copy of the stored element and reset the stored shared_ptr.
```
mjcarroll: This appears to be returning ownership, rather than a copy.  Please update the documentation to reflect this.

rclcpp/include/rclcpp/any_subscription_callback.hpp
ivanpauno: The signature of `dispatch_intra_process` method depends on the signature of the callback being used, and not on the message being stored.
https://github.com/ros2/rclcpp/blob/85c50717b6ca427f483c60b3129d1f0596b9869c/rclcpp/include/rclcpp/subscription.hpp#L148-L188
e.g.: The signature of the callback is `unique_ptr` and the message stored is a `ConstMessageSharedPtr`.
- `handle_intra_process_message` is call from an executor.
- L159 `any_callback_.use_take_shared_method()` is `false`.
- L175 `take_intra_process_message` subscriber method is called, with `unique_ptr` callback.
https://github.com/ros2/rclcpp/blob/85c50717b6ca427f483c60b3129d1f0596b9869c/rclcpp/include/rclcpp/subscription.hpp#L201-L215
- L213, IPM `take_intra_process_message` is called.
https://github.com/ros2/rclcpp/blob/85c50717b6ca427f483c60b3129d1f0596b9869c/rclcpp/include/rclcpp/intra_process_manager.hpp#L327-L361
- This calls `MappedRingBuffer` `get` or `pop` method with `unique_ptr` signature.
https://github.com/ros2/rclcpp/blob/85c50717b6ca427f483c60b3129d1f0596b9869c/rclcpp/include/rclcpp/mapped_ring_buffer.hpp#L104-L130
- Both of them copy the `ConstMessageSharedPtr` in a `unique_ptr`. e.g. for the `get` method, mapped_ring_buffer.hpp L113 `it->unique_value` is nullptr, so the else clause is executed, which copies the shared_pointer to a unique_ptr.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
ivanpauno: This is the renaming I did:
get_copy_at_key -> get
pop_at_key -> pop
get_ownership_at_key -> Deleted. It was used nowhere.

get, depending on the signature used and the stored message, could return a copy or share the ownership.
pop gives the ownership of the message to the caller. Except when storing a shared_ptr and trying to pop it into unique_ptr, as it's impossible to downgrade a shared_ptr. In this case, the shared_ptr is copied into the unique_ptr, and the MappedRingBuffer releases its ownership of the shared_ptr.

I will update the documentation, because I see it's wrong in some places.

mjcarroll:
CI through rclcpp:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6717/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3031/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5520/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6547/)

mjcarroll:
@ivanpauno Looks to be some warnings on Linux and a failure on macOS, can you get those resolved?

ivanpauno:
CI through rclcpp:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6720/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3034/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5523/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6550/)

ivanpauno:
> @ivanpauno Looks to be some warnings on Linux and a failure on macOS, can you get those resolved?

I still have an obscure warning on Windows. I will try to figure it out on Monday.

ivanpauno:
CI through rclcpp:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6592/)

wjwwood:
I'll try to have a look today.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -101,57 +102,98 @@ class MappedRingBuffer : public MappedRingBufferBase
...
-      auto ptr = ElemAllocTraits::allocate(*allocator_.get(), 1);
-      ElemAllocTraits::construct(*allocator_.get(), ptr, *it->value);
-      value = ElemUniquePtr(ptr);
+      T * raw_ptr;
+      ElemDeleter * deleter;
+      if (it->unique_value) {
+        raw_ptr = it->unique_value.get();
+        deleter = &it->unique_value.get_deleter();
+      } else {
+        raw_ptr = const_cast<T *>(it->shared_value.get());
```
yyu: maybe I haven't understood it but it feels the same `raw_ptr` will be deleted twice because it is owned by two different smart pointers

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -230,97 +90,77 @@ class Publisher : public PublisherBase
...
-    MessageUniquePtr unique_msg(ptr, message_deleter_);
-    return this->publish(unique_msg);
+    MessageAllocTraits::construct(*message_allocator_.get(), ptr, *msg);
+    MessageUniquePtr unique_msg = MessageUniquePtr(ptr, message_deleter_);
+    this->publish(unique_msg);
   }
 
   virtual void
-  publish(std::shared_ptr<const MessageT> msg)
+  publish(const std::shared_ptr<MessageT> && msg)
```
yyu: what is the purpose of a `const` rvalue ref?

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -180,29 +225,52 @@ class MappedRingBuffer : public MappedRingBufferBase
...
    */
   bool
-  push_and_replace(uint64_t key, ElemUniquePtr & value)
+  push_and_replace(uint64_t key, const ConstElemSharedPtr & value)
   {
     std::lock_guard<std::mutex> lock(data_mutex_);
     bool did_replace = elements_[head_].in_use;
     elements_[head_].key = key;
-    elements_[head_].value.swap(value);
+    elements_[head_].unique_value.reset();
```
wjwwood: I realize this was preexisting, but do you guys think it's worth saving a reference to `elements_[head_]` rather than accessing it with the `operator[]` each time? I assume the compiler is amazing and avoids any overhead here, but maybe it would be better than accessing it like this over and over when a reference should work just fine too.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -222,6 +73,15 @@ class Publisher : public PublisherBase
   virtual ~Publisher()
   {}
 
+  mapped_ring_buffer::MappedRingBufferBase::SharedPtr
+  make_mapped_ring_buffer(size_t size) const override
+  {
+    return mapped_ring_buffer::MappedRingBuffer<
+      MessageT,
+      typename Publisher<MessageT, Alloc>::MessageAlloc
+    >::make_shared(size, this->get_allocator());
```
wjwwood: Why was it necessary to move this here? Isn't this the reason that you now have a circular dependency that needed to be broken by moving lots of code around in the headers?

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -230,97 +90,77 @@ class Publisher : public PublisherBase
...
-      if (RCL_RET_OK != status) {
-        rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish intra process message");
-      }
-    } else {
-      // Always destroy the message, even if we don't consume it, for consistency.
       msg.reset();
+      return;
+    }
+    std::unique_ptr<MessageT, MessageDeleter> msg_copy;
+    // If interprocess subscriptors exists, then promote to shared and publish.
```
wjwwood: `subscriptions`, or if you must `subscribers`. I don't think `subscriptors` is a word.

wjwwood: Also the more documentation the better in cases like this. I'd devote a paragraph of comments here to describe that this is an optimization for a common, but narrow case and that it is important for performance. This will help people looking to understand the code for debugging or education and it makes it less likely that this code gets removed accidentally long down the road.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -230,97 +90,77 @@ class Publisher : public PublisherBase
...
+    if (!intra_process_is_enabled_) {
       // In this case we're not using intra process.
       return this->do_inter_process_publish(msg.get());
     }
     // Otherwise we have to allocate memory in a unique_ptr and pass it along.
-    // TODO(wjwwood):
-    //   The intra process manager should probably also be able to store
-    //   shared_ptr's and do the "smart" thing based on other intra process
-    //   subscriptions. For now call the other publish().
+    // NOTE(ivanpauno): As the message is not const, a copy should be made.
```
wjwwood: Please don't use `NOTE`, either use `TODO` if it is a todo, or just leave a comment. `NOTE` is highlighted as a problem along with `TODO` by several linters.

rclcpp/include/rclcpp/publisher.hpp
wjwwood: I don't understand either. `const` reference would be ok by me (though others have a dogmatic belief that you should never take a reference to a shared pointer), and I guess `&&` would be ok, but the `const &&` is odd. I suppose, like `const std::shared_ptr<T>` is implies this function will not modify it, but I'm honestly not sure if that makes sense or not. I'd have to think about it more. @ivanpauno do you have a specific reason for it?

wjwwood: We may want to drop this signature all together or change it to `const std::shared_ptr<const MessageT> &`?

See: https://github.com/ros2/rclcpp/issues/640

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -230,97 +90,77 @@ class Publisher : public PublisherBase
...
   publish(const MessageT & msg)
   {
     // Avoid allocating when not using intra process.
-    if (!store_intra_process_message_) {
+    if (!intra_process_is_enabled_) {
       // In this case we're not using intra process.
       return this->do_inter_process_publish(&msg);
     }
     // Otherwise we have to allocate memory in a unique_ptr and pass it along.
+    // NOTE(ivanpauno): As the message is not const, a copy should be made.
```
wjwwood: nitpick: again no `NOTE` please

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -230,97 +90,77 @@ class Publisher : public PublisherBase
...
+    // unique_ptr<MessageT> or shared_ptr<const MessageT> siganture could
+    // be used here. I decided to use the first one.
     auto ptr = MessageAllocTraits::allocate(*message_allocator_.get(), 1);
     MessageAllocTraits::construct(*message_allocator_.get(), ptr, msg);
-    MessageUniquePtr unique_msg(ptr, message_deleter_);
-    return this->publish(unique_msg);
+    MessageUniquePtr unique_msg = MessageUniquePtr(ptr, message_deleter_);
```
wjwwood: What's the difference here from the original `MessageUniquePtr unique_msg(ptr, message_deleter_);`?

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -376,6 +216,66 @@ class Publisher : public PublisherBase
...
+  }
+
+  uint64_t
+  store_intra_process_message(
+    uint64_t publisher_id,
+    const std::shared_ptr<const MessageT> & msg)
+  {
+    auto ipm = weak_ipm_.lock();
+    if (!ipm) {
+      // TODO(ivanpauno): should this just return silently? Or maybe return with a warning?
```
wjwwood: This is certainly an error case, in my opinion. The IPM is owned by the context and the context should out live the node, which should outlive the publisher.

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -266,89 +149,107 @@ class Subscription : public SubscriptionBase
...
+      ConstMessageSharedPtr msg;
+      take_intra_process_message(
+        ipm.publisher_id,
+        ipm.message_sequence,
+        intra_process_subscription_id_,
+        msg);
+      if (!msg) {
+        // This either occurred because the publisher no longer exists or the
+        // message requested is no longer being stored.
+        // TODO(wjwwood): should we notify someone of this? log error, log warning?
```
wjwwood: Let's add a log warning here (now that we have a logging system, we didn't at the time) and see if anyone complains. I think this shouldn't happen normally.

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -266,89 +149,107 @@ class Subscription : public SubscriptionBase
...
+      MessageUniquePtr msg;
+      take_intra_process_message(
+        ipm.publisher_id,
+        ipm.message_sequence,
+        intra_process_subscription_id_,
+        msg);
+      if (!msg) {
+        // This either occurred because the publisher no longer exists or the
+        // message requested is no longer being stored.
+        // TODO(wjwwood): should we notify someone of this? log error, log warning?
```
wjwwood: Here too.

rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace node_interfaces
+{
+class NodeTopicsInterface;
+}  // namespace node_interfaces
+
+namespace intra_process_manager
+{
+/**
+ * NOTE(ivanpauno): IntraProcessManager is forward declared here, avoiding a circular inclusion between intra_process_manager.hpp and publisher.hpp.
```
wjwwood: This line must be too long right? Also please no `NOTE` and didn't you move this into its own header to avoid the circular dependency? Is this even needed anymore?

rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -0,0 +1,165 @@
...
+    IntraProcessManagerWeakPtr weak_ipm,
+    const rcl_subscription_options_t & intra_process_options);
+
+protected:
+  std::shared_ptr<rcl_subscription_t> intra_process_subscription_handle_;
+  std::shared_ptr<rcl_subscription_t> subscription_handle_;
+  std::shared_ptr<rcl_node_t> node_handle_;
+
+  bool use_intra_process_;
+  IntraProcessManagerWeakPtr weak_ipm_;
```
wjwwood: So did the header changes even do anything if the base needs to reference the IPM?

rclcpp/src/rclcpp/publisher.cpp
```diff
@@ -235,10 +234,16 @@ PublisherBase::operator==(const rmw_gid_t * gid) const
   return result;
 }
 
+rclcpp::mapped_ring_buffer::MappedRingBufferBase::SharedPtr
+PublisherBase::make_mapped_ring_buffer(size_t size) const
```
wjwwood: If you move the `PublisherBase` into `publisher_base.hpp`, then please also make a `publisher_base.cpp` to match.

rclcpp/include/rclcpp/intra_process_manager.hpp
```diff
@@ -264,7 +283,7 @@ class IntraProcessManager
     // TODO(wjwwood): do something when a message was displaced. log debug?
     (void)did_replace;  // Avoid unused variable warning.
 
-    impl_->store_intra_process_message(intra_process_publisher_id, message_seq);
+    impl_->store_intra_process_message(intra_process_publisher_id, std::move(message_seq));
```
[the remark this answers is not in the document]
ivanpauno: That is an error, and doing a std::move here doesn't have any sense. I will correct it.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
ivanpauno: Yes, the `raw_ptr` is only used to construct a new unique pointer, although this is probably a little confusing. I will try to do it cleaner.

rclcpp/include/rclcpp/publisher.hpp
ivanpauno: `publish(const std::shared_ptr<MessageT> && msg)` the extra const is unnecessary,  and it should be deleted as I now realize that it's a bad idea.

I also left two other `publish` methods taking a `shared_ptr`:
`publish(const std::shared_ptr<const MessageT> & msg)` (1)
`publish(const std::shared_ptr<MessageT> & msg)` (2)

(1) Is the one that allows sharing the ownership of the published message with the subscribers.
(2) Constructs a unique_ptr copy and stores it.
According to #640, (2) could be deleted, and (1) could be modified to `publish(std::shared_ptr<const MessageT> msg)`

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -176,6 +177,26 @@ class AnySubscriptionCallback
     }
   }
 
+  void dispatch_intra_process(
+    ConstMessageSharedPtr & message, const rmw_message_info_t & message_info)
+  {
+    (void)message_info;
```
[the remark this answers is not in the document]
ivanpauno: Bad copy-pasting of the old `dispatch_intra_process` which takes a `MessageUniquePtr`.
I will delete that line in both methods.

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -176,6 +177,26 @@ class AnySubscriptionCallback
...
+    if (const_shared_ptr_callback_) {
+      const_shared_ptr_callback_(std::move(message));
```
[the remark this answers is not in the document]
ivanpauno: The shared_ptr passed here is taken from the IPM, which gives you a copy of the shared_ptr stored in the buffer or the ownership of the stored message in case all the other subscription has been handled. So it's fine to move it, as it's not used again.

It's true that this is a little confusing, I will try to do it in a cleaner way.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -101,57 +102,98 @@ class MappedRingBuffer : public MappedRingBufferBase
...
+   */
+  void
+  get(uint64_t key, ConstElemSharedPtr & value)
+  {
+    std::lock_guard<std::mutex> lock(data_mutex_);
+    auto it = get_iterator_of_key(key);
+    value.reset();
+    if (it != elements_.end() && it->in_use) {
+      if (!it->shared_value) {
+        it->shared_value = std::move(it->unique_value);
```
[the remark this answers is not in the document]
ivanpauno: IMHO, the extra logic needed to check what is going to be delivered is quite complex and the amount of copies needed is the same. I don't know if applying that change will end in better performance.

I will add a comment explaining the upgrading logic.

rclcpp/include/rclcpp/publisher.hpp
ivanpauno: I probably deleted and re-wrote it differently. I will come back to the original way, as it's cleaner.

ivanpauno:
@wjwwood I will explain here why I split `publisher.hpp` and `subscriber.hpp`. Sorry for the long story.

Before we had only one `store_intra_process_message_` method, which was bounded.

https://github.com/ros2/rclcpp/blob/6b108414779af49e22ffa13d1b9cb1fa2763d923/rclcpp/include/rclcpp/publisher.hpp#L189
https://github.com/ros2/rclcpp/blob/6b108414779af49e22ffa13d1b9cb1fa2763d923/rclcpp/include/rclcpp/publisher.hpp#L165

The callback took a `void *`, so it was a member of `PublisherBase`, and it was bounded from 'nodes_interfaces::NodeTopics`. The method definition was inside `PublisherFactory`

https://github.com/ros2/rclcpp/blob/6b108414779af49e22ffa13d1b9cb1fa2763d923/rclcpp/src/rclcpp/node_interfaces/node_topics.cpp#L45-L59

With the idea of having two `store_intra_process_message` methods, one taking a `unique_ptr` and one taking a `shared_ptr`, I moved those methods to the publisher and I stopped bounding them.
Those methods call the equivalent `store_intra_process_message` methods in the IPM, so I needed to include it.
As the IPM was including `publisher.hpp` but it didn't need more than `PublisherBase`, I split it in two files and included `publisher_base.hpp` for avoiding a circular inclusion. I still needed to forward declare the IPM in `publisher_base`, as I let `setup_intra_process` method and a `weak_ptr` to the IPM there.
The only place where the IPM needed `Publisher` and not `PublisherBase` was here:

https://github.com/ros2/rclcpp/blob/6b108414779af49e22ffa13d1b9cb1fa2763d923/rclcpp/include/rclcpp/intra_process_manager.hpp#L187-L201

So I moved the MappedRingBuffer creation to the publisher.

In `subscription.hpp`, when I replaced `get_intra_process_message_callback_ ` bounded  method with two `take_intra_process_message` methods I had the same problem. So I applied a similar logic.

This could have been solved in another way: The `get_intra_process_message_callback_` in `Subscription` already took a typed unique_ptr, the only difference with `Publisher` was that it was bounded from `setup_intra_process` method inside the factory.

https://github.com/ros2/rclcpp/blob/6b108414779af49e22ffa13d1b9cb1fa2763d923/rclcpp/include/rclcpp/subscription_factory.hpp#L118-L172

So I could bound the 'store' and 'take' methods to `Publisher` and `Subscription` with both `shared_ptr` and `unique_ptr` signature in a similar way, and avoid moving so many things.

I started splitting the `publisher.hpp` file, and when I realized of this last option I felt that it was more "natural" to don't have bounded methods.

Please if someone prefer the last option I commented or have a third one, let me know. This changes were only a "side effect", and it's not the main intention of this PR.

rclcpp/include/rclcpp/subscription_base.hpp
ivanpauno: Yes, I don't know why the linter didn't complained.

rclcpp/include/rclcpp/intra_process_manager.hpp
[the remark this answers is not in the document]
yyu: @wjwwood I think it's undefined behavior because passing an rvalue means the caller grants the callee permission to do whatever the callee wants

the following is a dumb example to demonstrate this. the idea is to have two `fn`s, one using lvalue the other rvalue. the rvalue version is justifiable when reusing some memory has some performance benefits -- maybe not in this case but you got the idea

```C++
int fn(const int& a, const int& b) { return a + b; }
int fn(const int& a, int&& b) { b += a; return b; }

void lvalue_call()
{
    int a{3}, b{5};
    int c = fn(a, b);
    std::cout << "[lvalue call] a = " << a << ", b = " << b << ", c = " << c << std::endl;
}

void rvalue_call()
{
    int a{3}, b{5};
    int c = fn(a, std::move(b));
    std::cout << "[rvalue call] a = " << a << ", b = " << b << ", c = " << c << std::endl;
}

```

```
$ g++ -std=c++14 foo.cc && ./a.out 
[lvalue call] a = 3, b = 5, c = 8
[rvalue call] a = 3, b = 8, c = 8
```

the point is, if one casts a value into rvalue, he should not expect anything predictable after that, even when the var is as simple as an `int`

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -180,29 +225,52 @@ class MappedRingBuffer : public MappedRingBufferBase
...
-    elements_[head_].value.swap(value);
+    elements_[head_].unique_value.reset();
+    elements_[head_].shared_value.reset();
+    elements_[head_].shared_value = value;
     elements_[head_].in_use = true;
     head_ = (head_ + 1) % elements_.size();
     return did_replace;
   }
 
   bool
```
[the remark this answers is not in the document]
allenh1: Also the `.swap` call should probably stay. I would advocate for

```cpp
elements_[head_].unique_value = nullptr;	    elements_[head_].shared_value.swap(value);
```

I think that makes it a little cleaner (in terms of readability).

rclcpp/include/rclcpp/publisher.hpp
ivanpauno: To be clearer, you could still share ownership of the message between multiple subscriptions using the `publish` method with `unique_ptr` signature, as it's internally promoted.
Creating a `std::shared_ptr<const MessageT>` is maybe a little painful, as the constructor of the messages doesn't allow initializing its members. I don't know if someone would use that `publish` signature.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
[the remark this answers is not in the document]
ivanpauno: If I swap values, the caller will lose the ownership of the message, and that's not the idea. The idea here is that the caller shares the ownership of the message with the value stored in the mapped ring buffer.
In the case of the `push_and_replace` method with `unique_ptr` signature, I could have used `swap`, but I don't think that's clearer.

rclcpp/include/rclcpp/publisher.hpp
wjwwood: > Creating a std::shared_ptr<const MessageT> is maybe a little painful, as the constructor of the messages doesn't allow initializing its members. I don't know if someone would use that publish signature.

You don't have to create the message const, it just tells the user that we won't mutate the message inside of publish.

Consider this:

```c++
#include <cstdio>
#include <memory>
#include <string>

struct Message {
  int a;
  double b;
};

void publish1(const std::shared_ptr<const Message> msg) {
  printf(
    "publishing with a=%s and b=%s\n",
    std::to_string(msg->a).c_str(),
    std::to_string(msg->b).c_str());
}

void publish2(const std::shared_ptr<Message> msg) {
  printf(
    "publishing with a=%s and b=%s\n",
    std::to_string(msg->a).c_str(),
    std::to_string(msg->b).c_str());
}

int main() {
  {
    auto msg = std::make_shared<Message>();
    msg->a = 42;
    msg->b = 3.14;

    publish1(msg);
    publish2(msg);
  }

  {
    auto const_msg = std::make_shared<const Message>();
    // const_msg->a = 42;
    // ^ compier error:
    //     cannot assign to return value because function 'operator->' returns a const value
    // std::shared_ptr<Message> msg = const_msg;
    //                              ^ compiler error:
    //     no viable conversion from 'shared_ptr<const Message>' to 'shared_ptr<Message>'
    publish1(const_msg);  // Ok.
    // publish2(const_msg);  // Not ok, same as most recent error above.
  }

  return 0;
}
```

I'd say we should only have `publish1` and not `publish2`.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -108,24 +109,25 @@ class MappedRingBuffer : public MappedRingBufferBase
...
-        ElemAllocTraits::construct(*allocator_.get(), ptr, *raw_ptr);
+        ElemAllocTraits::construct(*allocator_.get(), ptr, *it->shared_value);
         if (deleter) {
           value = ElemUniquePtr(ptr, *deleter);
         } else {
           value = ElemUniquePtr(ptr);
         }
+        return;
       }
+      throw std::runtime_error("Unexpected empty MappedRingBuffer element.");
```
wjwwood: It's a bit weird that here you throw at the end of the function, but in a similar case below you throw from the else case, use if-else for the shared value case, and do not return from those. I prefer the style in the case below to the one here, but either way it would be better to be consistent.

rclcpp/include/rclcpp/any_subscription_callback.hpp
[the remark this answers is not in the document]
wjwwood: I just wonder if the move is necessary here, if it boils down to essentially:

```c++
std::shared_ptr<T> message = func();
const_shared_ptr_callback_(message);
```

Then the compiler will automatically use move semantics because it knows `message` is not used after being passed to the callback anyways.

At least that's what I think will happen, we could try the `-Wredundant-move` warning to see if the compiler will tell us, see: https://vmpstr.blogspot.com/2015/12/redundant-stdmove.html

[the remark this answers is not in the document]
ivanpauno: I installed clang and built with `--Wredundant-move` and `-Wpessimizing-move`, no new warning appeared.
Still, I will change the signature to `ConstMessageSharedPtr`, stop using `std::move` explicitly, and let the compiler do it efficiently. That's maybe the standard way. To have a consistent style, I will do the same in the old method with `unique_ptr` signature:

https://github.com/ros2/rclcpp/blob/d5dc237099262565622d13a8203ffb0f8367e337/rclcpp/include/rclcpp/any_subscription_callback.hpp#L200-L220

I will apply similar changes in other methods with similar signatures that were modified or added in this PR.

ivanpauno:
According to [this discussion](https://github.com/ros2/rclcpp/pull/690#discussion_r277933981), the comment [here](https://github.com/ros2/rclcpp/pull/690#discussion_r277934665), #640, and the [CppCoreGuideline](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#S-resource) (R.32 to R.36) I changed all the methods (that are subject of this PR) with `unique_ptr` or `shared_ptr` signature to agree with them.
I know that I will probably break code in other repositories (demos, test_rclcpp, tlsf_cpp, ...), specially because now `std::move` is explicitly needed when publishing a `unique_ptr` message.

@wjwwood @mjcarroll Let me know if we want to include the change of:
`publish(std::unique_ptr<MessageT, MessageDeleter> & msg)` to
`publish(std::unique_ptr<MessageT, MessageDeleter> msg)`
in this PR or not. In case that we want to include it, I will start opening the remaining PRs on Monday.

I think that the other open discussion is if we want to split the `publisher.hpp` and `subscription.hpp` files or not. Please, check my previous [comment](https://github.com/ros2/rclcpp/pull/690#issuecomment-486260345) when possible.

ivanpauno:
Complete CI, to check what other packages I'm breaking:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6802/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3098/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5593/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6627/)

rclcpp/include/rclcpp/publisher.hpp
ivanpauno: I think `publish1` is dangerous. Because the caller could modify the message after doing the `publish` call.

I will prefer deleting all the `shared_ptr` `publish` calls, and only left the `unique_ptr` call.
The internal upgrading logic will still allow sharing the message between many subscriptions.

When having a `shared_ptr` message, the standard way of publishing it would be:

```cpp
publish(*shared_msg)
```
Which was also suggested by #640.
Obviously, we could still let a `shared_ptr` publish call, which copies the message to a `unique_ptr` and does the `unique_ptr` publish.

wjwwood: The point of the `publish1` signature is not to keep the same shared pointer, it is to copy it. Obviously you cannot keep a copy of the shared pointer the user gives you, because they could modify it afterwards. So the `publish1` signature is not dangerous so long as you make a copy of the data and not just the shared pointer.

It's only a convenience function beyond `publish(const T &)`, so you don't have to do `publish(*shared_msg)`.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
```diff
@@ -180,29 +235,44 @@ class MappedRingBuffer : public MappedRingBufferBase
...
     std::lock_guard<std::mutex> lock(data_mutex_);
     bool did_replace = elements_[head_].in_use;
-    elements_[head_].key = key;
-    elements_[head_].value.swap(value);
-    elements_[head_].in_use = true;
+    Element & element = elements_[head_];
+    element.key = key;
+    element.unique_value.reset();
+    element.shared_value.reset();
+    element.shared_value = value;
```
wjwwood: Tracing this from `rclcpp::Publisher<T>::publish(std::shared_ptr<T> msg)` down to here, it appears we are going to now share the value reference by `element.shared_value` with the user's shared pointer. This is really dangerous in my opinion. If I implied we should do this, then I was mistaken.

The idea of letting the IPM store share pointers is still valid, but only for the case where converting a unique pointer to a shared pointer has some benefit due to needing to deliver it to multiple subscribers which take `std::shared_ptr<const MessageT>` to their callback.

So in short, I think we need to duplicate the message referenced by `value` here and create our own shared pointer to point to the new copy.

ivanpauno: I have already deleted the `rclcpp::Publisher<T>::publish(std::shared_ptr<T> msg)` method.
I only left the one taking `std::shared_ptr<const T>`

rclcpp/include/rclcpp/intra_process_manager.hpp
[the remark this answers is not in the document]
wjwwood: Ok, thanks @yyu, looks like we're removing that either way, but I think you're right.

rclcpp/include/rclcpp/mapped_ring_buffer.hpp
wjwwood: I don't think `std::shared_ptr<const T>` is sufficiently safe, see:

https://gist.github.com/wjwwood/bf52d3911e7942077071a9089e3148e1

ivanpauno: Yes, that was exactly the doubt I have before. For the moment, I'm going to copy it and do a publish by value. We should later mark it as deprecated.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -222,105 +73,72 @@ class Publisher : public PublisherBase
...
+    // If an interprocess subscription exist, then the unique_ptr is promoted
+    // to a shared_ptr and published.
+    // This allows doing the intraprocess publish first and then doing the
+    // interprocess publish, resulting in lower publish-to-subscribe latency.
+    // It's not possible to do that with an unique_ptr,
+    // as do_intra_process_publish takes the ownership of the message.
+    if (get_subscription_count() > get_intra_process_subscription_count()) {
+      MessageSharedPtr shared_msg =
+        MessageSharedPtr(std::move(msg));
+      return this->publish(shared_msg);
```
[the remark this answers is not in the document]
ivanpauno: If you have both intraprocess and interprocess subscriptions, the idea here is to upgrade the `unique_ptr<T>` to a `shared_ptr<const T>`, then store the shared pointer in the `MRB` and then do the inter process publish. As I'm going to modify the `shared_ptr` publish method now, I will do that explicitly here.

Another option is possible: Do first interprocess publish and then store the `unique_ptr` in the `MRB`. That order is a little worst for publish-to-subscribe latency, although in some cases could end in avoiding one extra copy, e.g.: one `unique_ptr` subscription, one copy vs zero copy.

This is really a detail, I'm not completely sure of which decision is better. I will run a performance test comparing the last example I mentioned and then publish the results here.
It could be easily modified later without any API change, so modifying it again in a follow-up PR is not a big deal.

ivanpauno:
I will deprecate the shared_ptr publish method in another PR, opening all the needed PRs in other repos.
I will also do the `unique_ptr` publish method signature change in another PR, opening PRs in all the needed repos.

Complete CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6809/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3104/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5599/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6633/)

rclcpp/include/rclcpp/publisher.hpp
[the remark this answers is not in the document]
wjwwood: But the whole point is that the publish being called here is unsafe. It may be used safely here, but the user can use it unsafely. If you want to upgrade the unique ptr to a shared ptr, that's fine, but you need to take an internal path to stored the newly created shared ptr in the MRB and then continue the logic.

For a `publish(shared_ptr<T>)` like signature to be safe, it needs to copy the referenced data, not just copy the shared ptr (taking shared ownership), otherwise other people who hold the shared pointer may mutate it after publish returns but before intra process finishes. So if you have to copy it, then this function could really be implemented as `return this->publish(*shared_msg);`.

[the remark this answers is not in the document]
ivanpauno: Yes, see the last commit. I already took an internal path.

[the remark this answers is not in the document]
wjwwood: Yup, sorry still catching up :)

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -230,97 +90,50 @@ class Publisher : public PublisherBase
...
-    // Otherwise we have to allocate memory in a unique_ptr and pass it along.
-    // TODO(wjwwood):
-    //   The intra process manager should probably also be able to store
-    //   shared_ptr's and do the "smart" thing based on other intra process
-    //   subscriptions. For now call the other publish().
-    auto ptr = MessageAllocTraits::allocate(*message_allocator_.get(), 1);
-    MessageAllocTraits::construct(*message_allocator_.get(), ptr, *msg.get());
-    MessageUniquePtr unique_msg(ptr, message_deleter_);
-    return this->publish(unique_msg);
+    this->do_intra_process_publish(message_seq);
```
wjwwood: Does this function not need to do inter-process communication when intra process is enabled?

ivanpauno: Ufff, thanks!
Note: We should add a test with one intraprocess subscription and one interprocess subscription. And check that both receive the message.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -104,8 +104,10 @@ class Publisher : public PublisherBase
     // as do_intra_process_publish takes the ownership of the message.
     uint64_t message_seq;
     if (get_subscription_count() > get_intra_process_subscription_count()) {
+      MessageSharedPtr shared_msg(std::move(msg));
       message_seq =
-        store_intra_process_message(intra_process_publisher_id_, MessageSharedPtr(std::move(msg)));
+        store_intra_process_message(intra_process_publisher_id_, shared_msg);
+      this->do_inter_process_publish(msg.get());
```
wjwwood: Should this not got after the `do_intra_process_publish()`? Otherwise you have to wait on it to finish first...

ivanpauno: Sorry, and I also did `msg.get()` instead of `shared_msg.get()`.

wjwwood:
w.r.t. the splitting of the publisher and subscription headers, that's ok, but please also do the cpp files.

ivanpauno:
One CI more:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6823/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3119/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5611/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6646/)

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6824/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3120/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5612/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6648/)

ivanpauno:
Linux CI failed randomly, again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6828/)

rclcpp/test/test_intra_process_manager.cpp
```diff
@@ -50,6 +53,14 @@ class PublisherBase
     return false;
   }
 
+  virtual
```
dirk-thomas: The class need to have a virtual destructor then: see #723

PR #692, opened 2019-04-14
Fixup utilities

dorezyuk:
Cleans up some minor issues in utility.hpp/cpp

### DIFF
--- a/rclcpp/src/rclcpp/utilities.cpp
+++ b/rclcpp/src/rclcpp/utilities.cpp
@@ -14,6 +14,9 @@
 
 #include "rclcpp/utilities.hpp"
 
+#include <string>
+#include <vector>
+
 #include "./signal_handler.hpp"
 #include "rclcpp/contexts/default_context.hpp"
 #include "rclcpp/exceptions.hpp"
--- a/rclcpp/src/rclcpp/utilities.cpp
+++ b/rclcpp/src/rclcpp/utilities.cpp
@@ -21,45 +21,48 @@
 #include "rcl/error_handling.h"
 #include "rcl/rcl.h"
 
-void
-rclcpp::init(int argc, char const * const argv[], const rclcpp::InitOptions & init_options)
+namespace rclcpp
 {
-  using rclcpp::contexts::default_context::get_global_default_context;
+
+void
+init(int argc, char const * const argv[], const InitOptions & init_options)
+{
+  using contexts::default_context::get_global_default_context;
   get_global_default_context()->init(argc, argv, init_options);
   // Install the signal handlers.
-  rclcpp::install_signal_handlers();
+  install_signal_handlers();
 }
 
 bool
-rclcpp::install_signal_handlers()
+install_signal_handlers()
 {
-  return rclcpp::SignalHandler::get_global_signal_handler().install();
+  return SignalHandler::get_global_signal_handler().install();
 }
 
 bool
-rclcpp::signal_handlers_installed()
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/utilities.cpp
```diff
@@ -84,9 +84,9 @@ rclcpp::remove_ros_arguments(int argc, char const * const argv[])
     &nonros_argc,
     &nonros_argv);
 
-  if (RCL_RET_OK != ret) {
+  if (RCL_RET_OK != ret || nonros_argc < 0) {
```
sloretz: I think the original `RCL_RET_OK != ret` check is enough. It looks like the only case where it could be negative in `rcl_remove_ros_arguments` already returns `RCL_RET_INVALID_ARGUMENT`

https://github.com/ros2/rcl/blob/d94c958a23308299fccb9398153016fc64d81bd3/rcl/src/rcl/arguments.c#L449-L454

sloretz:
CI (build packages above rclcpp, test rclcpp and test_rclcpp)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6687/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3006/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5495/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6520/)

sloretz:
@dorezyuk Looks CI is yellow because of complaints from `cpplint`. Mind fixing them?

https://ci.ros2.org/job/ci_linux/6690/testReport/

rclcpp/src/rclcpp/utilities.cpp
dorezyuk: its true, that when you open up the implementation you can trace back, that the argc will never be smaller then 0 and the RCL_RET_OK. However the implementation may change without the client library getting notice of it. And if it ever returns a negative value, dire things will happen: a -1 will be std::numeric_limits<size_t>::max(), which will lead to the allocation of a very large data chunk. What do you think?

dorezyuk:
I've fixed the white space issue. However I don't understand the missing includes: <vector> and <string> are included in the hpp, where they belong, since they are both part of the function signature. Putting them **only** in the cpp is not an option and putting them in both (hpp/cpp) is really odd. Is there a "trick" to help the linter with that?

wjwwood:
> Putting them only in the cpp is not an option and putting them in both (hpp/cpp) is really odd. Is there a "trick" to help the linter with that?

Putting them in both is what you should do. The principle that the linter is following is "include what you use", which means locally and not relying on a header to do it for you. For a similar reason to why you wanted the redundant check, if in the future the header removed an include the C++ file may no longer work. It's better and more transparent to include it in the cpp file too.

Either way there's no way to reasonably change the linters behavior right now, so I advise you just make it happy :)

sloretz:
LGTM with green CI (build packages above rclcpp, test rclcpp and test_rclcpp)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6739/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3047/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5536/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6563/)

sloretz:
Thanks for the PR @dorezyuk !

PR #697, opened 2019-04-16
Add method to read timer cancellation

dabonnie:
Add method to read timer cancellation
Add relevant unit tests

Signed-off-by: Devin Bonnie <dbbonnie@amazon.com>

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -398,10 +398,10 @@ if(BUILD_TESTING)
   endif()
 
   ament_add_gtest(test_timer test/test_timer.cpp
-          APPEND_LIBRARY_DIRS "${append_library_dirs}")
+    APPEND_LIBRARY_DIRS "${append_library_dirs}")
   if(TARGET test_timer)
     ament_target_dependencies(test_timer
-            "rcl")
+      "rcl")
     target_link_libraries(test_timer ${PROJECT_NAME})
   endif()
 
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -397,6 +397,14 @@ if(BUILD_TESTING)
     target_link_libraries(test_time ${PROJECT_NAME})
   endif()
 
+  ament_add_gtest(test_timer test/test_timer.cpp
+          APPEND_LIBRARY_DIRS "${append_library_dirs}")
+  if(TARGET test_timer)
+    ament_target_dependencies(test_timer
+            "rcl")
+    target_link_libraries(test_timer ${PROJECT_NAME})
+  endif()
+
   ament_add_gtest(test_time_source test/test_time_source.cpp
     APPEND_LIBRARY_DIRS "${append_library_dirs}")
   if(TARGET test_time_source)
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -60,12 +60,13 @@ public:
 
   /// Return the timer cancellation state.
   /**
-   * @return true if the timer has been cancelled, false otherwise
-   * @throw std::runtime_error if the state could not be read
+   * \return true if the timer has been cancelled, false otherwise
+   * \throws std::runtime_error if the rcl_get_error_state returns 0
+   * \throws RCLErrorBase some child class exception based on ret
    */
   RCLCPP_PUBLIC
   bool
-  is_cancelled();
+  is_canceled();
 
   RCLCPP_PUBLIC
   void
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -59,6 +59,10 @@ public:
   cancel();
 
   /// Return the timer cancellation state.
+  /**
+   * @return true if the timer has been cancelled, false otherwise
+   * @throw std::runtime_error if the state could not be read
+   */
   RCLCPP_PUBLIC
   bool
   is_cancelled();
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -58,6 +58,11 @@ public:
   void
   cancel();
 
+  /// Return the timer cancellation state.
+  RCLCPP_PUBLIC
+  bool
+  is_cancelled();
+
   RCLCPP_PUBLIC
   void
   reset();
--- a/rclcpp/src/rclcpp/timer.cpp
+++ b/rclcpp/src/rclcpp/timer.cpp
@@ -19,6 +19,8 @@
 #include <memory>
 
 #include "rclcpp/contexts/default_context.hpp"
+#include "rclcpp/exceptions.hpp"
+
 #include "rcutils/logging_macros.h"
 
 using rclcpp::TimerBase;
@@ -76,12 +78,12 @@ TimerBase::cancel()
 }
 
 bool
-TimerBase::is_cancelled()
+TimerBase::is_canceled()
 {
   bool is_canceled = false;
-  if (rcl_timer_is_canceled(timer_handle_.get(), &is_canceled) != RCL_RET_OK) {
-    throw std::runtime_error(std::string("Couldn't get timer cancelled state: ") +
-            rcl_get_error_string().str);
+  rcl_ret_t ret = rcl_timer_is_canceled(timer_handle_.get(), &is_canceled);
+  if (ret != RCL_RET_OK) {
+    rclcpp::exceptions::throw_from_rcl_error(ret, "Couldn't get timer cancelled state");
   }
   return is_canceled;
 }
--- a/rclcpp/src/rclcpp/timer.cpp
+++ b/rclcpp/src/rclcpp/timer.cpp
@@ -75,6 +75,17 @@ TimerBase::cancel()
   }
 }
 
+bool
+TimerBase::is_cancelled()
+{
+  bool is_canceled = false;
+  if (rcl_timer_is_canceled(timer_handle_.get(), &is_canceled) != RCL_RET_OK) {
+    throw std::runtime_error(std::string("Couldn't get timer cancelled state: ") +
+            rcl_get_error_string().str);
+  }
+  return is_canceled;
+}
+
 void
 TimerBase::reset()
 {
--- a/rclcpp/test/test_timer.cpp
+++ b/rclcpp/test/test_timer.cpp
@@ -14,6 +14,7 @@
 
 #include <gtest/gtest.h>
 
+#include <atomic>
 #include <chrono>
 #include <exception>
 #include <memory>
@@ -29,7 +30,7 @@ using namespace std::chrono_literals;
 class TestTimer : public ::testing::Test
 {
 protected:
-  void SetUp()
+  void SetUp() override
   {
     rclcpp::init(0, nullptr);
     executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
@@ -56,7 +57,7 @@ protected:
     // don't start spinning, let the test dictate when
   }
 
-  void TearDown()
+  void TearDown() override
   {
     timer.reset();
     test_node.reset();
--- a/rclcpp/test/test_timer.cpp
+++ b/rclcpp/test/test_timer.cpp
@@ -20,48 +20,132 @@
 
 #include "rcl/timer.h"
 
+#include "rclcpp/executors/single_threaded_executor.hpp"
 #include "rclcpp/rclcpp.hpp"
 
 using namespace std::chrono_literals;
 
+/// Timer testing bring up and teardown
 class TestTimer : public ::testing::Test
 {
 protected:
   void SetUp()
   {
     rclcpp::init(0, nullptr);
+    executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+
+    has_timer_run.store(false);
+    cancel_timer.store(false);
+
     test_node = std::make_shared<rclcpp::Node>("test_timer_node");
+
+    timer = test_node->create_wall_timer(100ms,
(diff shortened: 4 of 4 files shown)


rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+  try {
+    auto timer = test_node->create_wall_timer(1s,
+        []() -> void {
+          printf("this is a test\n");
+        }
+    );
+
+    // start and cancel
+    ASSERT_FALSE(timer->is_cancelled());
```
thomas-moulard: ASSERT => EXPECT

https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html

rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,67 @@
...
+    test_node.reset();
+    rclcpp::shutdown();
+  }
+
+  rclcpp::Node::SharedPtr test_node;
+};
+
+TEST_F(TestTimer, test_is_cancelled)
+{
+  try {
```
thomas-moulard: You can remove the try...catch, gtest will report it if an exception escapes and print it anyway.

rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,67 @@
...
+    auto timer = test_node->create_wall_timer(1s,
+        []() -> void {
+          printf("this is a test\n");
+        }
+    );
+
+    // start and cancel
```
thomas-moulard: Where is this starting? This is not clear to me.

rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,67 @@
...
+    rclcpp::shutdown();
+  }
+
+  rclcpp::Node::SharedPtr test_node;
+};
+
+TEST_F(TestTimer, test_is_cancelled)
+{
+  try {
+    auto timer = test_node->create_wall_timer(1s,
```
thomas-moulard: This seems fragile, what if the main thread executes too slowly and the timer callback is executed before the assertions run?

rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,67 @@
...
+        []() -> void {
+          printf("this is a test\n");
```
thomas-moulard: Consider doing the following instead:

```c++
#include <atomic>

TEST(...) {
  std::atomic_bool has_timer_callback_run(false);
  auto timer = test_node->create_wall_timer(1s, [&has_timer_callback_run]() -> void {
    has_timer_callback_run.store(true);
  });

  EXPECT_FALSE(has_timer_callback_run.load());
  // etc.
}
```

You could also use an int with an int atomic increment if you expect the callback to run more than once (and you want to verify that it happens)

rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,67 @@
...
+  void TearDown()
+  {
+    test_node.reset();
+    rclcpp::shutdown();
+  }
+
+  rclcpp::Node::SharedPtr test_node;
+};
+
+TEST_F(TestTimer, test_is_cancelled)
```
thomas-moulard: Having a single test case is a code smell:
https://testing.googleblog.com/2018/06/testing-on-toilet-keep-tests-focused.html

If you end up having:

```c++
TEST(...) {
   // I do something
   {
      // some asserts
   }
   // I do something else
   {
      // more asserts
   }
}
```

You basically wrote two tests cases in one and those two should be separated.

I.e. your reset and cancel should be a different test case IMHO

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -58,6 +58,11 @@ class TimerBase
   void
   cancel();
 
+  /// Return the timer cancellation state.
```
thomas-moulard: Would be nice to specify in this doc this may throw a std::runtime_error if the timer state cannot be obtained.

rclcpp/test/test_timer.cpp
dabonnie: > Where is this starting? This is not clear to me.

That's a good question: I need to add the case where the timer has been run / started (spin the node).

rclcpp/test/test_timer.cpp
dabonnie: OK, this seems relevant for the case where the timer should have run. My intent for the current unit test is not testing the timer callback, but rather if the timer cancellation state has been exposed.

rclcpp/test/test_timer.cpp
dabonnie: > This seems fragile, what if the main thread executes too slowly and the timer callback is executed before the assertions run?

In this case I don't see the fragility given that the callback has no bearing on the tests executed. We are only checking timer cancellation state, not the state of the callback (nor does the callback set the state of anything in the test).

For the case where the timer should run (previous comment)

rclcpp/test/test_timer.cpp
dabonnie: I don't agree, as the current test is valid*. Adding a more permutations appears to be appropriate. 

*Checking that the timer state was reset after a cancellation, checking timer state after cancellation, and checking after a reset are atomic, valid permutations of tests.

rclcpp/include/rclcpp/timer.hpp
dabonnie: Ok, can do that. I'll look for some examples in the codebase.

rclcpp/test/test_timer.cpp
dabonnie: Will do.

rclcpp/test/test_timer.cpp
dabonnie: Will do.

rclcpp/test/test_timer.cpp
thomas-moulard: Oh ok, nevermind, I thought the timer was starting automatically.

rclcpp/test/test_timer.cpp
thomas-moulard: :+1:

rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,151 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <chrono>
```
thomas-moulard:
```
#include <atomic>
```

Currently you're depending indirectly on someone else including that for you which is brittle.

rclcpp/test/test_timer.cpp
thomas-moulard: Ok, that makes sense.

rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "rclcpp/executors/single_threaded_executor.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+using namespace std::chrono_literals;
+
+/// Timer testing bring up and teardown
+class TestTimer : public ::testing::Test
+{
+protected:
+  void SetUp()
```
thomas-moulard: nit: I'd add `final` there to explicitly state that this is an inherited method. Ditto for `TearDown`.
I think it's also something recommended in the Google C++ style guide.

rclcpp/test/test_timer.cpp
thomas-moulard: BTW you may want to use https://include-what-you-use.org/ to automate that stuff.

`apt-get install iwyu`
`colcon build --cmake-args -DIWYU_LLVM_ROOT_PATH=$(dirname $(which iwyu)) ...`

I can show you tomorrow if you want!

dabonnie: Oops, good catch. Will add.

rclcpp/test/test_timer.cpp
dabonnie: Ok, I'll use override (from https://google.github.io/styleguide/cppguide.html):

>Explicitly annotate overrides of virtual functions or virtual destructors with exactly one of an override or (less frequently) final specifier. Do not use virtual when declaring an override. Rationale: A function or destructor marked override or final that is not an override of a base class virtual function will not compile, and this helps catch common errors. The specifiers serve as documentation; if no specifier is present, the reader has to check all ancestors of the class in question to determine if the function or destructor is virtual or not.

dabonnie:
@thomas-moulard - please run the following CI job:

* Gist file: https://gist.githubusercontent.com/dabonnie/27cf6a977ebf4bb5ef1d76d0a966794e/raw/b2cbcffcb955304d5247eabc102cf582cb1fa14c/ros2.repos
* Additional BUILD flags: --packages-up-to rclcpp
* Additional TEST flags: --packages-up-to rclcpp
* CI Job: *ci_launcher*

dabonnie:
@wjwwood please review when convenient.

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6721/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3035/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5524/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6551/)

rclcpp/CMakeLists.txt
```diff
@@ -397,6 +397,14 @@ if(BUILD_TESTING)
     target_link_libraries(test_time ${PROJECT_NAME})
   endif()
 
+  ament_add_gtest(test_timer test/test_timer.cpp
+          APPEND_LIBRARY_DIRS "${append_library_dirs}")
```
wjwwood: nitpick: the indentation looks wrong here, looks like the rest of the file uses two spaces

rclcpp/CMakeLists.txt
```diff
@@ -397,6 +397,14 @@ if(BUILD_TESTING)
...
+  if(TARGET test_timer)
+    ament_target_dependencies(test_timer
+            "rcl")
```
wjwwood: nitpick: the indentation looks wrong here, looks like the rest of the file uses two spaces

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -58,6 +58,15 @@ class TimerBase
...
+  /**
+   * @return true if the timer has been cancelled, false otherwise
```
wjwwood: nitpick: we use `\` rather than `@` in our doxygen comments

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -58,6 +58,15 @@ class TimerBase
...
+   * @throw std::runtime_error if the state could not be read
+   */
+  RCLCPP_PUBLIC
+  bool
+  is_cancelled();
```
wjwwood: Can you please use `is_canceled`? We don't use the British variants anywhere else in our code/documentation (e.g. we use behavior not behaviour), see:

https://www.dictionary.com/browse/canceled

rclcpp/src/rclcpp/timer.cpp
```diff
@@ -75,6 +75,17 @@ TimerBase::cancel()
   }
 }
 
+bool
+TimerBase::is_cancelled()
+{
+  bool is_canceled = false;
+  if (rcl_timer_is_canceled(timer_handle_.get(), &is_canceled) != RCL_RET_OK) {
+    throw std::runtime_error(std::string("Couldn't get timer cancelled state: ") +
+            rcl_get_error_string().str);
```
wjwwood: Please use `rclcpp::exceptions::throw_from_rcl_error()` instead:

https://github.com/ros2/rclcpp/blob/713dd0c18411623ea82af1255f583a00c3c5970b/rclcpp/include/rclcpp/exceptions.hpp#L100-L119

wjwwood: (and update the documentation to reflect that it no longer exclusively throws `runtime_error`)

rclcpp/CMakeLists.txt
dabonnie: Addressed in the latest commit.

rclcpp/CMakeLists.txt
dabonnie: Addressed in the latest commit.

rclcpp/include/rclcpp/timer.hpp
dabonnie: Addressed in the latest commit.

rclcpp/include/rclcpp/timer.hpp
dabonnie: My mistake. Addressed in the latest commit.

rclcpp/src/rclcpp/timer.cpp
dabonnie: That's handy! Addressed in the latest commit.

dabonnie:
> relatively small changes requested, otherwise lgtm!

Hopefully good to go now with the most recent changes!

wjwwood:
Please run new CI on it when you're ready and after that it should be good to merge. Thanks for iterating!

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6733/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3044/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5533/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6560/)

wjwwood:
The test failures are the already known flake8 ones.

PR #698, opened 2019-04-17
[WIP / Re-Opened] Add functions to return formatted Node Name-Namespace strings

jhdcs:
This is a re-opening of pull request #669 after I accidentally deleted it. This pull request is intended to fix the issue outlined in issue #643.

In order for this pull request to work, modifications have to be made in other repositories (mainly system_tests). I will link that pull request once it is made.
**EDIT** Pull request is at https://github.com/ros2/system_tests/pull/345

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .DS_Store
 
 rclcpp/\.idea/
-
 tags
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 .DS_Store
+
+rclcpp/\.idea/
+
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,1 @@
 .DS_Store
-
-rclcpp/\.idea/
-tags
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -138,8 +138,11 @@ NodeGraph::get_node_names() const
   std::vector<std::string> nodes;
   auto names_and_namespaces = get_node_names_and_namespaces();
 
-  std::transform(names_and_namespaces.begin(), names_and_namespaces.end(), std::back_inserter(nodes),
-    [](std::pair<std::string, std::string> nns) {return nns.second + "/" + nns.first;});
+  std::transform(names_and_namespaces.begin(),
+    names_and_namespaces.end(),
+    std::back_inserter(nodes),
+    [](std::pair<std::string, std::string> nns) {return nns.second + "/" + nns.first;}
+  );
   return nodes;
 }
 
@@ -174,7 +177,7 @@ NodeGraph::get_node_names_and_namespaces() const
     throw std::runtime_error(error_msg);
   }
 
-  // TODO(jhdcs): If ranges are ever added to ROS2, replace following with range-based filter-transform
+
   std::vector<std::pair<std::string, std::string>> node_names;
   node_names.reserve(node_names_c.size);
   for (size_t i = 0; i < node_names_c.size; ++i) {
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -20,7 +20,6 @@
 #include <utility>
 #include <vector>
 
-
 #include "rcl/graph.h"
 #include "rclcpp/exceptions.hpp"
 #include "rclcpp/event.hpp"
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -137,8 +137,7 @@ NodeGraph::get_node_names() const
   std::vector<std::string> nodes;
   auto names_and_namespaces = get_node_names_and_namespaces();
 
-  std::transform(names_and_namespaces.begin(), names_and_namespaces.end(), std::back_inserter(
-      nodes),
+  std::transform(names_and_namespaces.begin(), names_and_namespaces.end(), std::back_inserter(nodes),
     [](std::pair<std::string, std::string> nns) {return nns.second + "/" + nns.first;});
   return nodes;
 }
@@ -174,10 +173,12 @@ NodeGraph::get_node_names_and_namespaces() const
     throw std::runtime_error(error_msg);
   }
 
-  std::vector<std::pair<std::string, std::string>> node_names(node_names_c.size);
+  // TODO(jhdcs): If ranges are ever added to ROS2, replace following with range-based filter-transform
+  std::vector<std::pair<std::string, std::string>> node_names;
+  node_names.reserve(node_names_c.size);
   for (size_t i = 0; i < node_names_c.size; ++i) {
     if (node_names_c.data[i] && node_namespaces_c.data[i]) {
-      node_names.push_back(std::make_pair(node_names_c.data[i], node_namespaces_c.data[i]));
+      node_names.emplace_back(node_names_c.data[i], node_namespaces_c.data[i]);
     }
   }
 
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -140,7 +140,20 @@ NodeGraph::get_node_names() const
   std::transform(names_and_namespaces.begin(),
     names_and_namespaces.end(),
     std::back_inserter(nodes),
-    [](std::pair<std::string, std::string> nns) {return nns.second + "/" + nns.first;}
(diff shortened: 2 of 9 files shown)


jhdcs:
I would like to mention that this change was a little more extensive than I was first expecting. The system tests in test_rclcpp in node_check_names, and node_name_list seemed to rely on the (now obsolete) original functionality of get_node_names. Namely, that it just returned an unqualified node name with no namespace information attached.

In the previous pull request (#669), the old functionality was deemed a bug, so the updating of the test_rclcpp files was warranted. However, I'm wondering if there's any way to warn users of the rclcpp package that this functionality has changed. Just in case they were using the old bugged version of the code, and relying on the bug.

jhdcs:
@wjwwood Would you be able to run a CI check on this? And who should I contact for testing its counterpart at [https://github.com/ros2/rclcpp/pull/345](url)?

rclcpp_components/test/test_component_manager_api.cpp
```diff
@@ -117,10 +117,10 @@ TEST_F(TestComponentManager, load_components)
       return std::find(node_names.begin(), node_names.end(), name) != node_names.end();
     };
 
-  EXPECT_TRUE(find_in_nodes("test_component_foo"));
-  EXPECT_TRUE(find_in_nodes("test_component_bar"));
-  EXPECT_TRUE(find_in_nodes("test_component_baz"));
-  EXPECT_TRUE(find_in_nodes("test_component_bing"));
+  EXPECT_TRUE(find_in_nodes("//test_component_foo"));
```
sloretz: Where does the double slash come from? A node with name `test_component_foo` and namespace `/` should have a fully qualified name `/test_component_foo`.

From http://design.ros2.org/articles/topic_and_service_names.html a topic name
```md
* must not contain any number of repeated forward slashes (`/`)
```

jhdcs: When the test is being implemented, the code creates a node with a namespace of "/", rather than assigning an empty namespace. This was code written before I got involved, so I'm guessing it was the author's way to get a fully-qualified name of "/test_component_foo".

However, since the new function automatically adds in the slash, this will give us a double slash.

I could fix this by either modifying the tests, or by introducing a check to see if either namespace ends in a slash, or name begins with one, and omitting the addition of a slash in that instance. Any preference for either?

sloretz:
> However, I'm wondering if there's any way to warn users of the rclcpp package that this functionality has changed. Just in case they were using the old bugged version of the code, and relying on the bug.

That would be a good thing to add to the [Breaking changes since X release](https://index.ros.org/doc/ros2/Releases/Release-Crystal-Clemmys/#changes-since-the-bouncy-release) on the release page for Dashing. A comment on ros2/ros2#607 with the text of the change should be enough to get it onto that page once it's created.

rclcpp_components/test/test_component_manager_api.cpp
sloretz: > or by introducing a check to see if either namespace ends in a slash, or name begins with one, and omitting the addition of a slash in that instance. 

I'd prefer this one. This is what `rcl_node_t` does when it computes it's fully qualified name

https://github.com/ros2/rcl/blob/273797e387f083929a5d1abeb6b3d509ffe62f99/rcl/src/rcl/node.c#L244-L248

jhdcs: Okay, got it modified. It's not my prettiest work, but it'll probably do.

I noticed a possible logic bug in the example you gave me: I think if someone enters a namespace that ends with "//", the qualified name computer will generate a qualified name with "//". But this is sort of a garbage-in-garbage-out thing, and the namespace will hopefully have been checked earlier.

jhdcs:
Right, I hope I managed to fix the issue you found with the code... Anything else jumping out at you? Should I break some of the logic in the lambda out into it's own function, or is it not really re-used a lot?

.gitignore
```diff
@@ -1 +1,4 @@
 .DS_Store
+
+rclcpp/\.idea/
+tags
```
sloretz: Would you mind undoing the changes to `.gitignore`? I recommend adding `.idea` to your global `core.excludesfile`.

Just curious, what creates the file or folder `tags`?

jhdcs: "tags" is generated by vim-gutentags. It's a tag file.

And I'll undo the changes.

jhdcs:
All right, have I missed anything? Can I get a CI on this?

sloretz:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6758/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3063/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5554/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6584/)

jhdcs:
So the windows build is getting a timeout error on test_params_yaml.py. Unfortunately it's not going into a lot of detail about which command is timing out (or I'm blind - probably the latter).

Or is this one of those flaky tests again?

sloretz:
> So the windows build is getting a timeout error on test_params_yaml.py. Unfortunately it's not going into a lot of detail about which command is timing out

 This is a known issue, ros2/build_cop#166

jhdcs:
So does that mean this is probably safe to merge? Or is there anything I can do to help debug why the test is flaky in this instance?

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -136,9 +137,24 @@ NodeGraph::get_node_names() const
...
 
-  for (const auto & it : names_and_namespaces) {
-    nodes.push_back(it.first);
-  }
+  std::transform(names_and_namespaces.begin(),
+    names_and_namespaces.end(),
+    std::back_inserter(nodes),
+    [](std::pair<std::string, std::string> nns) {
+      std::string return_string;
+      if ((nns.second.back() == '/') || (nns.first.front() == '/')) {
```
wjwwood: If both of these are true wouldn't you get `/ns//name`?

rclcpp_action/test/test_client.cpp
```diff
@@ -273,12 +273,14 @@ TEST_F(TestClient, async_send_goal_but_ignore_feedback_and_result)
   auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
   ASSERT_TRUE(action_client->wait_for_action_server(WAIT_FOR_SERVER_TIMEOUT));
 
+  printf("Testing Bad Goal...\n");
```
wjwwood: The `printf`'s in this file seem unnecessary, or at least are unrelated changes.

jhdcs: Yeah, I was doing some debugging on that and forgot to get rid of them.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
jhdcs: You are correct. I will modify the return_string to search and replace any "//" substrings. That should probably help.

I added these checks in because the unit tests were entering "/" as a namespace, and I wanted to make sure the output was at least qualified. Ideally, the name and namespace would both be enforced to be correct on node creation, but I figured that's beyond the scope of this change.

wjwwood: They should be enforced, and I think the name may never have a leading slash (I'd have to check), but the name space may have a slash in the `back()` position, specifically for the case where the namespace is `/`.

jhdcs: That makes my life even easier, assuming the name never has a leading slash. Also makes the code more performant (eww, string operations, eww)

jhdcs:
Okay, rebased both this and ros2/system_tests#345, and also updated the two comments you highlighted. I'm making the assumption that names cannot begin with "/", and that people haven't stuck two slashes at the end of a namespace.

sloretz:
CI (this PR plus ros2/system_tests#345,  testing packages above rclcpp)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6877/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3161/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5648/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6699/)

sloretz:
CI looks good. Thanks for the PRs @jhdcs!

dirk-thomas:
Please follow up with a PR to https://github.com/ros2/ros2_documentation/blob/master/source/Releases/Release-Dashing-Diademata.rst to document the change of the behavior.

jhdcs:
Sorry about that. I was pointed to a different issue.

I'll set up a PR ASAP for you.

dirk-thomas:
For reference: ros2/ros2_documentation#182.

PR #701, opened 2019-04-18
Add optional callbacks to action client for goal response and result

### LINKED ISSUE #635: Add more callbacks to rclcpp_action::Client
#### Feature description
Using the action client in `rclcpp_action` requires polling using timers on the future objects returned by `Client::async_send_goal` and `Client::async_get_result`. It would be far more convenient to add callbacks to these functions.

(```spin_until_future_complete``` won't work because my node is already spinning.)

Possible function signatures could look like this:
```
using GoalResultCallback =
  std::function<void(std::shared_future<typename GoalHandle::SharedPtr>)>;
using ResultCallback =
  std::function<void(std::shared_future<Result>)>;

std::shared_future<typename GoalHandle::SharedPtr>
async_send_goal(
  const Goal & goal, GoalResultCallback = nullptr,
  FeedbackCallback callback = nullptr, ResultCallback = nullptr,
  bool ignore_result = false);

std::shared_future<Result>
async_get_result(typename GoalHandle::SharedPtr goal_handle, ResultCallback = nullptr)
```

While these are the callbacks I currently need, it would be consistent to add appropriate callbacks to all functions returning future objects.

jacobperron:
Resolves #635.
This also makes it easier to incorporate action clients in composable nodes since we don't have to rely on waiting on futures.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -267,6 +267,8 @@ public:
   using ResultCallback = typename GoalHandle::ResultCallback;
   using CancelRequest = typename ActionT::Impl::CancelGoalService::Request;
   using CancelResponse = typename ActionT::Impl::CancelGoalService::Response;
+  using CancelOneCallback = std::function<void (typename GoalHandle::SharedPtr, bool)>;
+  using CancelMultipleCallback = std::function<void (typename CancelResponse::SharedPtr)>;
 
   /// Options for sending a goal.
   /**
@@ -419,10 +421,15 @@ public:
    * \throws exceptions::UnknownGoalHandleError If the goal is unknown or already reached a
    *   terminal state.
    * \param[in] goal_handle The goal handle requesting to be canceled.
+   * \param[in] cancel_callback Optional callback that is called when the response is received.
+   *   The callback function takes two parameters: a shared pointer to the goal handle and a bool
+   *   indicating if the action server accepted the cancel request or not.
    * \return A future whose result indicates whether or not the cancel request was accepted.
    */
   std::shared_future<bool>
-  async_cancel_goal(typename GoalHandle::SharedPtr goal_handle)
+  async_cancel_goal(
+    typename GoalHandle::SharedPtr goal_handle,
+    CancelOneCallback cancel_callback = nullptr)
   {
     std::lock_guard<std::mutex> lock(goal_handles_mutex_);
     if (goal_handles_.count(goal_handle->get_goal_id()) == 0) {
@@ -436,7 +443,7 @@ public:
     cancel_request->goal_info.goal_id.uuid = goal_handle->get_goal_id();
     this->send_cancel_request(
       std::static_pointer_cast<void>(cancel_request),
-      [goal_handle, promise](std::shared_ptr<void> response) mutable
+      [goal_handle, cancel_callback, promise](std::shared_ptr<void> response) mutable
       {
         auto cancel_response = std::static_pointer_cast<CancelResponse>(response);
         bool goal_canceled = false;
@@ -446,12 +453,17 @@ public:
           goal_canceled = (canceled_goal_info.goal_id.uuid == goal_handle->get_goal_id());
         }
         promise->set_value(goal_canceled);
+        if (cancel_callback) {
+          cancel_callback(goal_handle, goal_canceled);
+        }
       });
     return future;
   }
 
   /// Asynchronously request for all goals to be canceled.
   /**
+   * \param[in] cancel_callback Optional callback that is called when the response is received.
+   *   The callback takes one parameter: a shared pointer to the CancelResponse message.
    * \return A future to a CancelResponse message that is set when the request has been
    * acknowledged by an action server.
    * See
@@ -459,19 +471,21 @@ public:
    * action_msgs/CancelGoal.srv</a>.
    */
   std::shared_future<typename CancelResponse::SharedPtr>
-  async_cancel_all_goals()
+  async_cancel_all_goals(CancelMultipleCallback cancel_callback = nullptr)
   {
     auto cancel_request = std::make_shared<CancelRequest>();
     // std::fill(cancel_request->goal_info.goal_id.uuid, 0u);
     std::fill(
       cancel_request->goal_info.goal_id.uuid.begin(),
       cancel_request->goal_info.goal_id.uuid.end(), 0u);
-    return async_cancel(cancel_request);
+    return async_cancel(cancel_request, cancel_callback);
   }
 
   /// Asynchronously request all goals at or before a specified time be canceled.
   /**
    * \param[in] stamp The timestamp for the cancel goal request.
+   * \param[in] cancel_callback Optional callback that is called when the response is received.
+   *   The callback takes one parameter: a shared pointer to the CancelResponse message.
    * \return A future to a CancelResponse message that is set when the request has been
    * acknowledged by an action server.
    * See
@@ -479,7 +493,9 @@ public:
    * action_msgs/CancelGoal.srv</a>.
    */
   std::shared_future<typename CancelResponse::SharedPtr>
-  async_cancel_goals_before(const rclcpp::Time & stamp)
+  async_cancel_goals_before(
+    const rclcpp::Time & stamp,
+    CancelMultipleCallback cancel_callback = nullptr)
   {
     auto cancel_request = std::make_shared<CancelRequest>();
     // std::fill(cancel_request->goal_info.goal_id.uuid, 0u);
@@ -487,7 +503,7 @@ public:
       cancel_request->goal_info.goal_id.uuid.begin(),
       cancel_request->goal_info.goal_id.uuid.end(), 0u);
     cancel_request->goal_info.stamp = stamp;
-    return async_cancel(cancel_request);
+    return async_cancel(cancel_request, cancel_callback);
   }
 
   virtual
@@ -618,17 +634,22 @@ private:
 
   /// \internal
   std::shared_future<typename CancelResponse::SharedPtr>
-  async_cancel(typename CancelRequest::SharedPtr cancel_request)
+  async_cancel(
+    typename CancelRequest::SharedPtr cancel_request,
+    CancelMultipleCallback cancel_callback = nullptr)
   {
     // Put promise in the heap to move it around.
     auto promise = std::make_shared<std::promise<typename CancelResponse::SharedPtr>>();
     std::shared_future<typename CancelResponse::SharedPtr> future(promise->get_future());
     this->send_cancel_request(
       std::static_pointer_cast<void>(cancel_request),
-      [promise](std::shared_ptr<void> response) mutable
+      [cancel_callback, promise](std::shared_ptr<void> response) mutable
       {
         auto cancel_response = std::static_pointer_cast<CancelResponse>(response);
         promise->set_value(cancel_response);
+        if (cancel_callback) {
+          cancel_callback(cancel_response);
+        }
       });
     return future;
   }
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -261,9 +261,43 @@ public:
   using Feedback = typename ActionT::Feedback;
   using GoalHandle = ClientGoalHandle<ActionT>;
   using WrappedResult = typename GoalHandle::WrappedResult;
+  using GoalResponseCallback =
+    std::function<void (std::shared_future<typename GoalHandle::SharedPtr>)>;
   using FeedbackCallback = typename ClientGoalHandle<ActionT>::FeedbackCallback;
+  using ResultCallback = typename ClientGoalHandle<ActionT>::ResultCallback;
   using CancelRequest = typename ActionT::Impl::CancelGoalService::Request;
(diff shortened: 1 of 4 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6726/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3040/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5529/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6556/)

jacobperron:
Test failures appear unrelated. This is ready for review.

rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
```diff
@@ -71,6 +73,9 @@ ClientGoalHandle<ActionT>::set_result(const WrappedResult & wrapped_result)
   std::lock_guard<std::mutex> guard(handle_mutex_);
   status_ = static_cast<int8_t>(wrapped_result.code);
   result_promise_.set_value(wrapped_result);
+  if (nullptr != result_callback_) {
```
hidmic: @jacobperron nit: why not
```suggestion
  if (result_callback_) {
```
like it's done for the `goal_response_callback` above?

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -318,31 +325,47 @@ class Client : public ClientBase
...
-        std::shared_ptr<GoalHandle> goal_handle(new GoalHandle(goal_info, callback));
+        std::shared_ptr<GoalHandle> goal_handle(
+          new GoalHandle(goal_info, feedback_callback, result_callback));
         if (!ignore_result) {
           try {
             this->make_result_aware(goal_handle);
           } catch (...) {
             promise->set_exception(std::current_exception());
+            if (goal_response_callback) {
+              goal_response_callback(nullptr);
```
hidmic: @jacobperron (meta, can ignore) what the callback gets as its sole argument isn't quite the same as what a wait on the future/promise would get. I wonder if we should be passing the resolved future/promise instead.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -307,7 +310,11 @@ class Client : public ClientBase
    */
   std::shared_future<typename GoalHandle::SharedPtr>
   async_send_goal(
-    const Goal & goal, FeedbackCallback callback = nullptr, bool ignore_result = false)
+    const Goal & goal,
+    GoalResponseCallback goal_response_callback = nullptr,
+    FeedbackCallback feedback_callback = nullptr,
+    ResultCallback result_callback = nullptr,
+    bool ignore_result = false)
```
hidmic: @jacobperron (meta, can ignore): I know this is a purely subjective observation, biased by own prejudices, but since C++ lacks keyword arguments, I find having to pass multiple `nullptr` values to get to the one I actually care about somewhat annoying. Can we have some overloads for the most common use cases e.g. goal and feedback cb, goal, feedback cb and result cb, etc.?

wjwwood: We've been going the way of "options structs" which alleviates some of this without expanding the number of signatures, also the signatures get confusing/ambiguous at some point.

But in general I agree that this becomes too fragile and inconvenient at some point.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -318,31 +325,47 @@ class Client : public ClientBase
...
     this->send_goal_request(
       std::static_pointer_cast<void>(goal_request),
-      [this, goal_request, callback, ignore_result, promise](
+      [this,
+      goal_request,
+      goal_response_callback,
+      feedback_callback,
+      result_callback,
+      ignore_result,
+      promise](
```
wjwwood: nitpick: this is a strange indention...

I would have done it like this:

```
[
  this,
  ...,
](...) mutable
{
  // ...
}
```

rclcpp_action/include/rclcpp_action/client.hpp
wjwwood: I have to admit I don't understand what the semantic of passing nullptr here is. What scenario does this occur in?

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: I could be persuaded to update this signature to accept a `SendGoalOptions` struct.

I was also thinking we might consider removing the `ignore_result` parameter and instead rely on checking if `result_callback` is set.

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: I believe this is the scenario where the "send result" request fails unexpectedly. In retrospect, I'm not sure what the correct behavior is in this case. The goal has already been accepted by the server, but we failed to request the result. Maybe we should attempt to cancel the goal? Maybe we actually return a goal handle, but throw an error if they attempt to get the result?

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: eeed6f5

rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
jacobperron: eeed6f5

rclcpp_action/include/rclcpp_action/client.hpp
wjwwood: I'm not sure what the right behavior is either, but it should probably be part of the documented behavior of these callbacks.

rclcpp_action/include/rclcpp_action/client.hpp
wjwwood: Which ever is fine by me.

jacobperron: I've added an options struct in f1c46f5. PTAL.

wjwwood: Looks reasonable to me.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -331,8 +331,7 @@ class Client : public ClientBase
       feedback_callback,
       result_callback,
       ignore_result,
-      promise](
-        std::shared_ptr<void> response) mutable
+      promise](std::shared_ptr<void> response) mutable
```
wjwwood: This wasn't what I was talking about on the style, the weird thing is that `[this` and `goal_request` use the same indention.

wjwwood: Sorry, ignore this, I didn't see that you updated the signature again in the next commit.

jacobperron: Right, FWIW, I recall trying to indent these lines as you suggested but one of the linters was complaining. Maybe because I didn't drop the closing `]` to the next line (I can't remember).

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: I've updated the signature of the response callback to take a future (20757b4). Now, in this case calling `future::get()` will throw the exception.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -283,8 +283,11 @@ class Client : public ClientBase
 
     /// Function called when the goal is accepted or rejected.
     /**
-     * If the goal is accepted, then a shared pointer to a valid goal handle is passed as the
-     * single argument. If the goal is rejected, then pointer has the value `nullptr`.
+     * Takes a single argument that is a future to a goal handle shared pointer.
+     * If the goal is accepted, then the pointer points to a valid goal handle.
+     * If the goal is rejected, then pointer has the value `nullptr`.
+     * If an error occurs while waiting for the goal response an exception will be thrown
+     * when calling `future::get()`.
```
wjwwood: It would be nice to list the at least some of the exceptions we know it can throw here.

jacobperron:
@wjwwood If there is an exception it is set here: 

https://github.com/ros2/rclcpp/blob/20757b4571312609e3e04f9232e70650c4742c40/rclcpp_action/include/rclcpp_action/client.hpp#L374

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: Done b69a1fd

jacobperron:
@hidmic @wjwwood In addition to addressing your feedback (thanks!), I've added optional callback parameters to `async_get_result` and the cancel methods. I plan to squash these into three separate commits before merging. Ready for another round of review.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6776/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3077/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5572/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6604/)

CI also includes connected PRs:

- https://github.com/ros2/examples/pull/230
- https://github.com/ros2/system_tests/pull/346

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -318,31 +352,38 @@ class Client : public ClientBase
...
         }
         GoalInfo goal_info;
         goal_info.goal_id.uuid = goal_request->goal_id.uuid;
         goal_info.stamp = goal_response->stamp;
         // Do not use std::make_shared as friendship cannot be forwarded.
-        std::shared_ptr<GoalHandle> goal_handle(new GoalHandle(goal_info, callback));
-        if (!ignore_result) {
+        std::shared_ptr<GoalHandle> goal_handle(
+          new GoalHandle(goal_info, options.feedback_callback, options.result_callback));
+        if (options.result_callback) {
```
hidmic: @jacobperron just for the record, this now means that if you just want the result future, you need to ask for it explicitly later on, always.

jacobperron: Yeah, that's true. Though, this is a subtle feature. I bet that most use cases would provide a callback anyways and the case for sending a request early only to get a reference to the result future later is rare. The user can still get the future later with the client method, `async_get_result()`.

This brings me to another point. I think it's odd to have two public facing methods for the getting the result, `Client.async_get_result()` and `ClientGoalHandle.async_result()`. Originally, the client method was essentially bypassing the "ignore result" flag. I propose we make the goal handle method private and rely on the client method to return the future. Regarding the "ignore result" flag, in the case where the goal handle is not result aware, we have 2.5 options:

1. Keep the current behavior and make the goal handle result aware. I.e. Does not fail if the user previously set the "ignore result" flag.
    1. Or we can not even provide the "ignore result" option and make result requests on demand (this PR). 
2. Throw an exception, because the user chose to ignore the result explicitly.

Let me know what you think. 

To not drag this PR on any longer, I suggest we make any changes related to this discussion in a follow-up PR.

rclcpp_action/test/test_client.cpp
```diff
@@ -344,14 +381,83 @@ TEST_F(TestClient, async_send_goal_with_feedback_and_result)
...
+        result_callback_received = true;
+      }
+    };
+  auto future_goal_handle = action_client->async_send_goal(goal, send_goal_ops);
+  dual_spin_until_future_complete(future_goal_handle);
+  auto goal_handle = future_goal_handle.get();
+  EXPECT_EQ(rclcpp_action::GoalStatus::STATUS_ACCEPTED, goal_handle->get_status());
+  EXPECT_FALSE(goal_handle->is_feedback_aware());
+  EXPECT_TRUE(goal_handle->is_result_aware());
+  auto future_result = action_client->async_get_result(goal_handle);
```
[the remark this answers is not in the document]
jacobperron: See my comment below proposing we make `goal_handle->async_result()` private.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -318,31 +352,38 @@ class Client : public ClientBase
...
           try {
             this->make_result_aware(goal_handle);
           } catch (...) {
             promise->set_exception(std::current_exception());
+            options.goal_response_callback(future);
```
ipa-mdl: this does not check `goal_response_callback` first

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -318,31 +352,38 @@ class Client : public ClientBase
...
             return;
           }
         }
         std::lock_guard<std::mutex> guard(goal_handles_mutex_);
         goal_handles_[goal_handle->get_goal_id()] = goal_handle;
         promise->set_value(goal_handle);
+        if (options.goal_response_callback) {
+          options.goal_response_callback(future);
```
ipa-mdl: This might get called after `result_callback` (not 100% sure)

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: Good catch. I've swapped the two if-blocks as suggested in https://github.com/ros2/rclcpp/pull/738.

Issue #702, opened 2019-04-18
multithreaded executor concurency problem


guillaumeautran:
When processing a `MutuallyExclusive` `callback_group`, the multi-threaded executor ends up getting into a race condition with the `can_be_taken_from_` atomic boolean. The problem is that the [Executor::execute_any_executable](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executor.cpp#L288) resets the `MutuallyExclusive` `callback_group` `can_be_taken_from_` atomic boolean and the [AnyExecutable destructor](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/any_executable.cpp#L35) also does reset the calllback group flag as well. 
These two action cause the `can_be_taken_from_` exclusion flag to get out of sync with reality causing a concurrency issue.

The resulting observable behaviour could be a re-ordering of messages within the same topic subscription (ie: older messages being delivered after newer messages).  

## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 16.04
- Installation type:
  - From source
- Version or commit hash:
  -0.6.4
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Subscriber node (where the issue will be seen):
- Create a ROS2 rclcpp node with the multi-threaded executor (default arguments are fine)
- Register to a topic caring some basic ordered data (a topic of Header type with a properly field timestamp would be fine)
- In the topic callback, verify that the message timestamps are in the proper order (newer after older)
Publisher Node:
- Create a ROS2 node with the topic the subscriber node subscribed to (as stated earlier)
- Publish Header messages with properly filled timestamps (as fast as possible)

#### Expected behavior
Every Header messages are received in order

#### Actual behavior
Header messages are randomly re-ordered (older message arriving last)

sloretz:
Fixed by #703

PR #703, opened 2019-04-18
Fix a concurrency problem in the multithreaded executor

guillaumeautran:
Both, the `Executor::execute_any_executable` and the destructor for the `AnyExecutable` object used by the multithreaded executor, reset the `can_be_taken_from_` flag on a MutuallyExclusive group. This cause the variable to get out of sync and threads to process executables out of sequence.

This fix clears the callback group variable of the `AnyExecutable` instance effectively preventing its destructor from modifying the variable at the wrong time.

Issue: #702

### DIFF
--- a/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
@@ -101,5 +101,8 @@ MultiThreadedExecutor::run(size_t)
         scheduled_timers_.erase(it);
       }
     }
+    // Clear the callback_group to prevent the AnyExecutable destructor from
+    // resetting the callback group `can_be_taken_from`
+    any_exec.callback_group.reset();
   }
 }


guillaumeautran:
The jenkins job failed on 
```14:10:01 /tmp/ws/src/rclcpp/rclcpp_action/src/server_goal_handle.cpp:134:59: note: suggested alternative: ‘GOAL_EVENT_CANCEL’
14:10:01      ret = rcl_action_update_goal_state(rcl_handle_.get(), GOAL_EVENT_CANCELED);
14:10:01                                                            ^~~~~~~~~~~~~~~~~~~
14:10:01                                                            GOAL_EVENT_CANCEL
14:10:01 CMakeFiles/rclcpp_action.dir/build.make:110: recipe for target 'CMakeFiles/rclcpp_action.dir/src/server_goal_handle.cpp.o' failed
```
Which I believe is unrelated to this change. Or is it?

sloretz:
@guillaumeautran the `Dpr__rclcpp__ubuntu_bionic_amd64` failure can be ignored. It is testing against crystal, but there have been breaking changes on master for the upcomming dashing release. No action needed.

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6759/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3064/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5555/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6585/)

sloretz:
CI looks ok, thanks for the PR @guillaumeautran !

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -101,5 +101,8 @@ MultiThreadedExecutor::run(size_t)
         scheduled_timers_.erase(it);
       }
     }
+    // Clear the callback_group to prevent the AnyExecutable destructor from
+    // resetting the callback group `can_be_taken_from`
+    any_exec.callback_group.reset();
```
wjwwood: I'm concerned this isn't the right fix, @guillaumeautran why did you not change the destructor of `AnyExecutable` instead?

Just changing the callback group to `nullptr` is fragile. There's nothing in the destructor of `AnyExecutable` which implies the callback group might be reset to affect behavior.

Instead, I think, at the very least, `AnyExecutable` should have had a boolean like `should_reset_can_be_taken_from_automatically` which defaulted to true, and the multi threaded executor could change to false. Then it's clearer from both the multi threaded executor's code and any executable's code what's happening.

guillaumeautran: I was not sure if there was a specific reason why the destructor was setup the way it was and did not want to run the risk of leaving the group "locked". I'll make the requested changes then and push another PR.

wjwwood: It's possible that since this exists:

https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executor.cpp#L288

We should just remove the code in the destructor of `AnyExecutable` that changes the `can_be_taken_from` of the callback group.

guillaumeautran: @wjwwood Do you think I can add an argument to the constructor of the `AnyExecutable` class?

guillaumeautran: I would be hesitant from removing the if statement. There is a comment that makes a lot of sense:
https://github.com/ros2/rclcpp/blob/97ed34a042c1d4e1a88ea06986112cbdd05e08a9/rclcpp/src/rclcpp/any_executable.cpp#L34

PR #704, opened 2019-04-22
Reimplement the fix a concurrency problem in the multithreaded executor

guillaumeautran:
Reimplement the fix to correct a concurrency problem in the multithreaded executor.
This time, the AnyExecutable class has a boolean flag to reset / not reset the callback group `can_be_taken_from_` variable on destruction.
The multithreaded executor initializes the executor with that flag set to false.

Issue: #702

### DIFF
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -35,7 +35,7 @@ namespace executor
 struct AnyExecutable
 {
   RCLCPP_PUBLIC
-  AnyExecutable();
+  AnyExecutable(bool auto_reset_group = true);
 
   RCLCPP_PUBLIC
   virtual ~AnyExecutable();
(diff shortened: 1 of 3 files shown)


guillaumeautran:
FYI: @wjwwood

wjwwood:
Thanks @guillaumeautran for opening this follow up.

After reading through all of this, and seeing the comment in the destructor of `AnyExecutable` and looking at what I proposed as an actually pr (this one), I think what you did in the other pull request was sufficient.

So, unless you'd prefer to continue with this pull request, I'm ok to close it.

----

In a different direction, it would be great to have a test that could reproduce this case. It seems like something that could easily regress with future work on the executor.

sloretz:
@guillaumeautran Thanks for fixing the issue. Based on @wjwwood 's comment it sounds like the fix you made in #703 is the way to go, so I'll close this PR for now.

PR #707, opened 2019-04-29
Add missing template functionality to lifecycle_node.

mjeronimo:
Recent changes to the node_parameters interface was accompanied by additions to the node.hpp header and implementation files. However, these additions were not also made to the corresponding lifecycle node files. This PR includes the changes required for the lifecycle node.

Signed-off-by: michael.jeronimo@intel.com

I suggest that template methods that supplement the basic node interfaces should either be factored out into a separate header that both node and lifecycle_node include or included in the basic node interface header, if possible. Also, any regular methods added to the node class that are convenience wrappers for calls to the interfaces should also be added to the lifecycle_node class. That way we can avoid node and lifecycle_node getting out of sync, making porting code from node to lifecycle node easier.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -940,7 +940,6 @@ public:
   void
   register_param_change_callback(CallbackT && callback);
 
-
   /// Get the fully-qualified names of all available nodes.
   /**
    * The fully-qualified name includes the local namespace and name of the node.
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
@@ -333,6 +333,190 @@ public:
     const rmw_qos_profile_t & qos_profile = rmw_qos_profile_services_default,
     rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr);
 
+  /// Declare and initialize a parameter, return the effective value.
+  /**
+   * This method is used to declare that a parameter exists on this node.
+   * If, at run-time, the user has provided an initial value then it will be
+   * set in this method, otherwise the given default_value will be set.
+   * In either case, the resulting value is returned, whether or not it is
+   * based on the default value or the user provided initial value.
+   *
+   * If no parameter_descriptor is given, then the default values from the
+   * message definition will be used, e.g. read_only will be false.
+   *
+   * The name and type in the given rcl_interfaces::msg::ParameterDescriptor
+   * are ignored, and should be specified using the name argument to this
+   * function and the default value's type instead.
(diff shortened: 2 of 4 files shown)


wjwwood:
I actually intentionally didn't add these originally, just to keep the declare parameter pr moving. We planned to do this in a follow up pr, along with changes in `rclpy` to reflect the new approach.

There's an on going discussion about how to make changes like this go more smoothly, but unfortunately that will require a major API change.

wjwwood:
I'll try to pick this pr up and make any required changes soon, in the meantime, everything can be done with the `NodeParameterInterface`, if not as conveniently.

Karsten1987:
Just to add on what was said before. The call to the parameter interface could look like this:
```
get_node_parameters_interface()->declare_parameter("my_parameter", rclcpp::ParameterValue("my_default_value"), rcl_interfaces::msg::ParameterDescriptor());
```

I do believe it makes sense to have a free floating function in the style of `rclcpp::create_subscription()`. Maybe something like `rclcpp::declare_parameter` which takes the parameter interface as a first argument.

wjwwood:
I think all of the functions on `rclcpp::Node` should have free function equivalents like this, but would need to refactor the code a lot, that's what I mean.

My proposal here, already changes `rclcpp::create_publisher` and `rclcpp::create_subscription` to take a "node-like" so that you could use lifecycle node or normal node as the first argument and it would "just work":

https://github.com/ros2/rclcpp/pull/713/files#diff-d8ede312b2071056a6ea5591ae22b4e0R68

mjeronimo:
@wjwwood @Karsten1987 Will the work on the free function equivalents for the lifecycle node be integrated in time for the Dashing release?

wjwwood:
I would like to, but I cannot guarantee it. Since it will only be adding API we have a little more than a week to get it in. I will try to merge this as an incremental improvement if there's time. If there's a lot more time, I will try to refactor it into free functions to reduce code duplication.

wjwwood:
Running CI on this:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7100/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3322/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5816/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6920/)

wjwwood:
Another round of CI after I fixed up a bunch of stuff:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7105/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3325/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5819/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6923/)

PR #709, opened 2019-04-29
Deprecate shared ptr publish

ivanpauno:
See https://github.com/ros2/rclcpp/pull/690#discussion_r279109760 https://github.com/ros2/rclcpp/pull/690#issuecomment-487664648

Depends on #690

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -165,25 +165,25 @@ public:
   }
 
   void
-  publish(const rcl_serialized_message_t * serialized_msg)
+  publish(const rcl_serialized_message_t & serialized_msg)
   {
-    if (intra_process_is_enabled_) {
-      // TODO(Karsten1987): support serialized message passed by intraprocess
-      throw std::runtime_error("storing serialized messages in intra process is not supported yet");
-    }
-    auto status = rcl_publish_serialized_message(&publisher_handle_, serialized_msg, nullptr);
-    if (RCL_RET_OK != status) {
-      rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish serialized message");
-    }
+    return this->do_serialized_publish(&serialized_msg);
   }
 
   [[deprecated(
-    "publishing an unique_ptr is prefered when using intra process communication."
-    " If using a shared_ptr, use publish(*msg).")]]
+    "Use publish(*serialized_msg). Check against nullptr before calling if necessary.")]]
+  void
+  publish(const rcl_serialized_message_t * serialized_msg)
+  {
+    return this->do_serialized_publish(serialized_msg);
+  }
+
+  [[deprecated(
+    "Use publish(*serialized_msg). Check against nullptr before calling if necessary.")]]
   void
   publish(std::shared_ptr<const rcl_serialized_message_t> serialized_msg)
   {
-    return this->publish(serialized_msg.get());
+    return this->do_serialized_publish(serialized_msg.get());
   }
 
   std::shared_ptr<MessageAlloc> get_allocator() const
@@ -208,6 +208,19 @@ protected:
     }
     if (RCL_RET_OK != status) {
       rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish message");
+    }
+  }
+
+  void
+  do_serialized_publish(const rcl_serialized_message_t * serialized_msg)
+  {
+    if (intra_process_is_enabled_) {
+      // TODO(Karsten1987): support serialized message passed by intraprocess
+      throw std::runtime_error("storing serialized messages in intra process is not supported yet");
+    }
+    auto status = rcl_publish_serialized_message(&publisher_handle_, serialized_msg, nullptr);
+    if (RCL_RET_OK != status) {
+      rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish serialized message");
     }
   }
 
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -97,11 +97,10 @@ public:
    * \param[in] msg A shared pointer to the message to send.
    */
   virtual void
-  publish(std::unique_ptr<MessageT, MessageDeleter> & msg)
+  publish(std::unique_ptr<MessageT, MessageDeleter> msg)
   {
     if (!intra_process_is_enabled_) {
       this->do_inter_process_publish(msg.get());
-      msg.reset();
       return;
     }
     // If an interprocess subscription exist, then the unique_ptr is promoted
@@ -150,7 +149,7 @@ public:
     auto ptr = MessageAllocTraits::allocate(*message_allocator_.get(), 1);
     MessageAllocTraits::construct(*message_allocator_.get(), ptr, msg);
     MessageUniquePtr unique_msg(ptr, message_deleter_);
-    this->publish(unique_msg);
+    this->publish(std::move(unique_msg));
   }
 
   virtual void
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -128,6 +128,7 @@ public:
     }
   }
 
+  [[deprecated("use publish(*msg)")]]
   virtual void
   publish(const std::shared_ptr<const MessageT> & msg)
   {
@@ -173,6 +174,7 @@ public:
     }
   }
 
+  [[deprecated("use publish(*msg)")]]
   void
   publish(std::shared_ptr<const rcl_serialized_message_t> serialized_msg)
   {
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -128,7 +128,8 @@ public:
     }
   }
 
-  [[deprecated("use publish(*msg)")]]
+  [[deprecated("publishing an unique_ptr is prefered when using intra process communication."
+               " If using a shared_ptr, use publish(*msg).")]]
   virtual void
   publish(const std::shared_ptr<const MessageT> & msg)
   {
@@ -174,7 +175,8 @@ public:
     }
   }
 
-  [[deprecated("use publish(*msg)")]]
+  [[deprecated("publishing an unique_ptr is prefered when using intra process communication."
+               " If using a shared_ptr, use publish(*msg).")]]
   void
   publish(std::shared_ptr<const rcl_serialized_message_t> serialized_msg)
   {
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -127,8 +127,9 @@ public:
     }
   }
 
-  [[deprecated("publishing an unique_ptr is prefered when using intra process communication."
-               " If using a shared_ptr, use publish(*msg).")]]
+  [[deprecated(
+    "publishing an unique_ptr is prefered when using intra process communication."
+    " If using a shared_ptr, use publish(*msg).")]]
   virtual void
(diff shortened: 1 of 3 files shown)


rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -113,6 +113,7 @@ class Publisher : public PublisherBase
     this->do_intra_process_publish(message_seq);
   }
 
+  [[deprecated("use publish(*msg)")]]
```
dirk-thomas: An alternative suggestion would be to publish a unique ptr. Especially in the intra process case that would be better. Maybe mention that case?

rclcpp/test/test_time_source.cpp
```diff
@@ -139,7 +139,7 @@ void trigger_clock_changes(
     auto msg = std::make_shared<rosgraph_msgs::msg::Clock>();
     msg->clock.sec = i;
     msg->clock.nanosec = 1000;
-    clock_pub->publish(msg);
+    clock_pub->publish(*msg);
```
wjwwood: This is a perfect example of a case where you should use `unique_ptr` instead.

wjwwood: Actually, you could use a stack reference instead.

wjwwood: Depends on what we think is more overhead, if we know the stack reference will result in a heap allocation (to make space in which to store the copy) then we can just do `unique_ptr`.

ivanpauno: I used an stack variable here as it is a test with intra process not configured.

ivanpauno:
I'm deprecating similar methods in lifecycle publisher now

ivanpauno:
I also changed the `unique_ptr` publish call here. As I will use that call in the depending PRs.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -119,6 +118,8 @@ class Publisher : public PublisherBase
     }
   }
 
+  [[deprecated("publishing an unique_ptr is prefered when using intra process communication."
+               " If using a shared_ptr, use publish(*msg).")]]
```
wjwwood: Do not wrap like this, instead indent only two spaces, if you want the strings aligned then line break after the `(`.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -164,6 +165,8 @@ class Publisher : public PublisherBase
...
```
wjwwood: Here too.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -96,6 +96,8 @@ class LifecyclePublisher : public LifecyclePublisherInterface,
    * was enabled or disabled and forwards the message
    * to the actual rclcpp Publisher base class
    */
+  [[deprecated("publishing an unique_ptr is prefered when using intra process communication."
+               " If using a shared_ptr, use publish(*msg).")]]
```
wjwwood: And here.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -96,8 +96,11 @@ class LifecyclePublisher : public LifecyclePublisherInterface,
    * was enabled or disabled and forwards the message
    * to the actual rclcpp Publisher base class
    */
+  [[deprecated(
+    "publishing an unique_ptr is prefered when using intra process communication."
+    " If using a shared_ptr, use publish(*msg).")]]
   virtual void
-  publish(const std::shared_ptr<const MessageT> & msg)
+  publish(std::shared_ptr<const MessageT> msg)
```
ivanpauno: This is quite strange:
If I overload the publish method here, don't add a `deprecated` attribute and call a non-deprecated publish method in the base-class -> I never see a warning.
If I overload the publish method here, add a `deprecated` attribute and call a non-deprecated publish method in the base-class -> I see a warning in template specialization.
If I overload the publish method here, don't add a `deprecated` attribute and call the deprecated publish method in the base-class -> I see a warning when the package is being built, but not when the dependencies are built.
(Ubuntu 18.04, gcc)

As a workaround, I'm using `const std::shared_ptr<const MessageT> & msg` in the base class and `std::shared_ptr<const MessageT> msg` here.

If I deprecate the publish methods taking a ptr, I will have a similar problem. @wjwwood suggestions?

Maybe I could directly delete them from the lifecycle publisher, which has less packages depending on it.

ivanpauno:
CI, just to keep the things going:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6845/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3137/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5628/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6666/)

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
ivanpauno: And the trick doesn't work, because it generates a different [warning in `macOS`](https://ci.ros2.org/job/ci_osx/5629/warnings11Result/package.-1256088273/).
I don't see a way of deprecating a method in the base class and the overloading method in the derived class without generating a warning, I will directly delete the `deprecated` methods in `lifecycle_publisher`, it shouldn't generate much problems.

ivanpauno:
Another CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6900/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3176/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5665/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6721/)

ivanpauno:
Forgot rebasing, another full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6909/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3183/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5​67​2/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6730/)

ivanpauno:
Windows job had a random failure while running tests .... again:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6​74​4/)

ivanpauno:
Corrected badly macOS warning last time, again:

* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5​68​1/)

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -164,6 +168,9 @@ class Publisher : public PublisherBase
     }
   }
 
+  [[deprecated(
+    "publishing an unique_ptr is prefered when using intra process communication."
+    " If using a shared_ptr, use publish(*msg).")]]
```
wjwwood: This suggestion won't work because the only other signature for an `rcl_serialized_message_t` is `publish(const rcl_serialized_message_t *)`.

If we had a `publish(const rcl_serialized_message_t &)` then this message would make more sense. I think that would be a good thing to have, and in which case we could deprecate `publish(const rcl_serialized_message_t *)` too, that way were only ever dealing with `unique_ptr` and references, which is ideal in my opinion.

ivanpauno:
Hopefully, last CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6937/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3202/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5697/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6767/)

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -195,6 +202,19 @@ class Publisher : public PublisherBase
     }
   }
 
+  void
+  do_serialized_publish(const rcl_serialized_message_t * serialized_msg)
```
wjwwood: This could have been `const rcl_serializated_message_t &`, but we can do that in a follow up change. Given that it's an internally used protected, I think it could be updated after API freeze.

ivanpauno:
CI (up to `lifecycle` demo, where we had the warning problem)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6958/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3210/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5708/)

(windows ci was ran using branch `ivanpauno/avoid-deprecate-warning-in-template-specialization`. I cleaned a `#if __clang__` specific macro which does nothing in windows)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6768/)

ivanpauno:
CI (up to `lifecycle` demo)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6965/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3216/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5​71​3/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6​78​7/) (unrelated test failure, with build warnings)

wjwwood:
I'm re-running CI for this:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6978/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3232/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5724/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6802/)

wjwwood:
Sorry for all the traffic on some of the related pr's, I actually didn't notice they existed because they were not connected  in waffle, but instead used blocked and depends on, which made it not show up properly. Anyways I've consolidated my changes (redundant) and moved all pr's that were forks to branches so CI is easier to run:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6979/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3234/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5725/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6803/)

wjwwood:
The CI failures on Windows and aarch64 look unrelated. I believe this is ready for merge, but I'll wait since I don't know what the plan for tomorrow is.

ivanpauno:
Thank you @wjwwood. I created some of them in a fork, because I don't have write access on those repos (I will later ask for it).
I believe is ready for merge too.

wjwwood:
CI, just to the example package that needed to change (since I made a small adjustment there):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6985/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3237/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5730/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6806/)

wjwwood:
Ok, new CI after fixing merge conflict in `ros2/system_tests` (using `--packages-{up-to, select} test_communication test_rclcpp test_security`):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6989/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3240/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5733/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6811/)

PR #713, opened 2019-05-02
change API to encourage users to specify history depth always

wjwwood:
This pr does a few things:

- Improves `rclcpp::Duration` so that specifying things like `deadline` and `lifespan` are more natural.
- Adds an rclcpp version of QoS, which requires the history depth to be specified if using "keep last".
- Deprecates `create_*` signatures that allow you to create something without specifying history depth or "keep all".
- Add signatures that take the new `rclcpp::QoS` as the recommended alternative to the deprecated methods.
- Updates NodeOptions to include new QoS and PublisherOptions for the parameter event topic.
- Changes to NodeBase to avoid dependency on NodeOptions to avoid settings that apply to Node but not NodeBase.
- Changes to `rclcpp::create_publisher` and `rclcpp::create_subscription` free functions to make them easier to use.
  - And improve them so that even less code is duplicated between `rclcpp::Node` and `rclcpp_lifecycle::LifecycleNode`.
- Update the `rclcpp_lifecycle` code to match.
- Updates tests and related code to avoid now deprecated signatures.

I'll also summarize the changes to the signatures and what that means for the existing code that uses this.

Forthcoming is a set of pr's to update downstream packages to avoid deprecation warnings.

Unfortunately, this is still not finished yet, as I am now stuck on template deduction problems on Windows. I am somewhat hopeful, that I can fix those problems with only small changes, and so this would be good to review already, but there might be changes for Windows still...

The goal was to get this in for the API freeze, but the unforeseen Windows issues may have derailed that. So I put it in the TODO column just in case there is time.

I've been trying to reproduce it in a minimal example in compiler explorer, with no luck yet. In case you're curious: https://godbolt.org/z/Gu0hMC

### Publishers

For publishers, this is a summary of what will have changed from Crystal to Dashing (including this pr):

```c++
// Completely new signature.
template<
  typename MessageT,
  typename AllocatorT = std::allocator<void>,
  typename PublisherT = ::rclcpp::Publisher<MessageT, AllocatorT>>
std::shared_ptr<PublisherT>
create_publisher(
  const rclcpp::QoS & qos,
  const PublisherOptionsWithAllocator<AllocatorT> & options = (
    PublisherOptionsWithAllocator<AllocatorT>()));
```

```diff
   typename MessageT,
   typename Alloc = std::allocator<void>,
   typename PublisherT = ::rclcpp::Publisher<MessageT, Alloc>>
+[[deprecated("use create_publisher(const std::string &, const rclcpp::QoS &, ...) instead")]]
 std::shared_ptr<PublisherT>
 create_publisher(
   const std::string & topic_name,
   size_t qos_history_depth,
-  std::shared_ptr<Alloc> allocator = nullptr);
+  std::shared_ptr<Alloc> allocator);
 
 template<
   typename MessageT,
   typename Alloc = std::allocator<void>,
   typename PublisherT = ::rclcpp::Publisher<MessageT, Alloc>>
+[[deprecated("use create_publisher(const std::string &, const rclcpp::QoS &, ...) instead")]]
 std::shared_ptr<PublisherT>
 create_publisher(
   const std::string & topic_name,
   const rmw_qos_profile_t & qos_profile = rmw_qos_profile_default,
   std::shared_ptr<AllocatorT> allocator = nullptr);

```

This is a list of all of the API calls possible in Crystal and what happens to each of them:

- `create_publisher("")`:
  - still works, but generates deprecation warning
  - because it invokes `create_publisher("", rmw_qos_profile_t, allocator)` and uses the default arguments for the rmw qos profile and allocator
- `create_publisher("", size_t)`:
  - still works, invokes `create_publisher("", QoS, PublisherOptions)` and `QoS`'s conversion constructor for `size_t` and gets the default argument for the publisher options
- `create_publisher("", size_t, allocator)`:
  - still works, but generates deprecation warning
  - because it invokes `create_publisher("", size_t, allocator)`
- `create_publisher("", rmw_qos_profile_t)`:
  - still works, but generates deprecation warning
  - because it invokes `create_publisher("", rmw_qos_profile_r, allocator)` and uses the default argument for the allocator
- `create_publisher("", rmw_qos_profile_t, allocator)`:
  - still works, but generates deprecation warning
  - because it invokes `create_publisher("", rmw_qos_profile_r, allocator)`

The qos settings are removed from the new `PublisherOptions`, preferring to always explicitly pass the QoS separately. This also allows the `PublisherOptions` to be default-constructable (unlike `rclcpp::QoS`) and avoids an additional signature where depth is provided as well as the parameter options, for which the internal depth value would be ignored, creating a strange behavior.

### Subscriptions

For subscriptions, this is a summary of what will have changed from Crystal to Dashing (including this pr):

```c++
// Completely new signature.
template<
  typename MessageT,
  typename CallbackT,
  typename AllocatorT = std::allocator<void>,
  typename SubscriptionT = rclcpp::Subscription<
    typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, AllocatorT>>
create_subscription(
  const rclcpp::QoS & qos,
  CallbackT && callback,
  const SubscriptionOptionsWithAllocator<AllocatorT> & options = (
    SubscriptionOptionsWithAllocator<AllocatorT>()
  ),
  typename rclcpp::message_memory_strategy::MessageMemoryStrategy<
    typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, AllocatorT
  >::SharedPtr
  msg_mem_strat = nullptr);
```

```diff
 template<
   typename MessageT,
   typename CallbackT,
   typename Alloc = std::allocator<void>,
   typename SubscriptionT = rclcpp::Subscription<
     typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, Alloc>>
+[[deprecated(
+  "use create_subscription(const std::string &, CallbackT, const rclcpp::QoS &, ...) instead"
+)]]
 std::shared_ptr<SubscriptionT>
 create_subscription(
   const std::string & topic_name,
   CallbackT && callback,
   const rmw_qos_profile_t & qos_profile = rmw_qos_profile_default,
   rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr,
   bool ignore_local_publications = false,
   typename rclcpp::message_memory_strategy::MessageMemoryStrategy<
     typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, Alloc>::SharedPtr
   msg_mem_strat = nullptr,
   std::shared_ptr<Alloc> allocator = nullptr);

 template<
   typename MessageT,
   typename CallbackT,
   typename Alloc = std::allocator<void>,
   typename SubscriptionT = rclcpp::Subscription<
     typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, Alloc>>
+[[deprecated(
+  "use create_subscription(const std::string &, CallbackT, const rclcpp::QoS &, ...) instead"
+)]]
 std::shared_ptr<SubscriptionT>
 create_subscription(
   const std::string & topic_name,
   CallbackT && callback,
   size_t qos_history_depth,
-  rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr,
+  rclcpp::callback_group::CallbackGroup::SharedPtr group,
   bool ignore_local_publications = false,
   typename rclcpp::message_memory_strategy::MessageMemoryStrategy<
     typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, Alloc>::SharedPtr 
   msg_mem_strat = nullptr,
   std::shared_ptr<Alloc> allocator = nullptr);

```

This is a list of all of the API calls possible in Crystal and what happens to each of them:

- `create_subscription("", callback)`:
  - still works, but generates deprecation warning
  - because it invokes `create_subscription("", callback, rmw_qos_profile_t, ...)` and uses the default arguments for the qos profile, and so on
- `create_subscription("", callback, rmw_qos_profile_t)`:
  - still works, but generates deprecation warning
  - because it invokes `create_subscription("", callback, rmw_qos_profile_t, ...)` and uses the default arguments for the rest
- `create_subscription("", callback, rmw_qos_profile_t, ...)`:
  - still works, but generates deprecation warning
  - because it invokes `create_subscription("", callback, rmw_qos_profile_t, ...)` and uses the default arguments for the arguments not provided, if any
- `create_subscription("", callback, size_t)`:
  - still works,
  - invokes the new `create_subscription("", callback, rclcpp::QoS, ...)` using the `rclcpp::QoS` conversion constructor for `size_t`
- `create_subscription("", callback, size_t, group, ...)`:
  - still works, but generates deprecation warning
  - because it invokes `create_subscription("", callback, size_t, group, ...)` and uses the default arguments for the arguments not provided, if any

Like the `PublisherOptions`, the qos settings are removed from the new `SubscriptionOptions`, preferring to always explicitly pass the QoS separately.


rclcpp/test/test_publisher.cpp
```diff
@@ -80,16 +80,91 @@ TEST_F(TestPublisher, construction_and_destruction) {
...
+  // Now deprecated functions.
+#if !defined(_WIN32)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#else  // !defined(_WIN32)
+# pragma warning(push)
+# pragma warning(disable: 4996)
+#endif
+  {
+    // auto publisher = node->create_publisher<IntraProcessMessage, std::allocator<void>, rclcpp::Publisher<IntraProcessMessage, std::allocator<void>>>("topic");
```
ivanpauno: Clean comment

rclcpp/test/test_publisher.cpp
```diff
@@ -80,16 +80,91 @@ TEST_F(TestPublisher, construction_and_destruction) {
...
+    auto publisher = node->create_publisher<IntraProcessMessage>("topic");
+    (void)publisher;
+  }
+#if 0
```
ivanpauno: I suppose the code between `#if 0` and `#endif` will be cleaned.

rclcpp/test/test_publisher.cpp
```diff
@@ -98,9 +173,23 @@ TEST_F(TestPublisher, intraprocess_with_invalid_qos) {
...
   using rcl_interfaces::msg::IntraProcessMessage;
   {
+#if !defined(_WIN32)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#else  // !defined(_WIN32)
+# pragma warning(push)
+# pragma warning(disable: 4996)
+#endif
+#if 0
```
ivanpauno: same

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -30,16 +30,21 @@ class Duration
...
+  // cppcheck-suppress noExplicitConstructor
+  Duration(const std::chrono::duration<Rep, Period> & duration)  // NOLINT(runtime/explicit)
+  : Duration(std::chrono::duration_cast<std::chrono::nanoseconds>(duration))
+  {}
 
   RCLCPP_PUBLIC
-  Duration(
-    const builtin_interfaces::msg::Duration & duration_msg);
+  // cppcheck-suppress noExplicitConstructor
+  Duration(const builtin_interfaces::msg::Duration & duration_msg);
```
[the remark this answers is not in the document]
wjwwood: It occurs when processing a single file (as my editor gives it to cppcheck), but for some reason it does not happen when cppcheck is give data via `ament_cppcheck`.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -144,8 +145,27 @@ class Node : public std::enable_shared_from_this<Node>
 
   /// Create and return a Publisher.
   /**
+   * The rclcpp::QoS has several convenient constructors, including a
+   * conversion constructor for size_t, which mimics older API's that
+   * allows just a string and size_t to create a pull request.
+   *
+   * For example, all of these cases will work:
+   *
+   *   pub = node->create_publisher<MsgT>("chatter", 10);  // implicitly KeepLast
```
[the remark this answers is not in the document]
wjwwood: Actually in Crystal `"", size_t` was allowed, maybe not on master, but that would have been a regression of the previous `PublisherOptions` pr.

rclcpp/include/rclcpp/subscription_traits.hpp
```diff
@@ -18,6 +18,7 @@
 #include <memory>
 
 #include "rclcpp/function_traits.hpp"
+#include "rcl/types.h"
```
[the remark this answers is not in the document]
wjwwood: Because without it this header does not compile, as it uses rcl types.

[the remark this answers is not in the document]
wjwwood: Does not compile by itself, it is relying on another header included before it to include this file.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -109,7 +109,11 @@ Node::Node(
   const std::string & namespace_,
   const NodeOptions & options)
 : node_base_(new rclcpp::node_interfaces::NodeBase(
-      node_name, namespace_, options)),
+    node_name,
+    namespace_,
+    options.context(),
+    *(options.get_rcl_node_options()),
```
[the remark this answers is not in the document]
wjwwood: I added them for readability, but I could remove them if you like, i.e. we're de-referencing the return of `get_rcl_node_options()` not `options`.

rclcpp/test/test_duration.cpp
```diff
@@ -60,6 +60,7 @@ TEST(TestDuration, operators) {
   rclcpp::Duration time = rclcpp::Duration(0, 0);
   rclcpp::Duration copy_constructor_duration(time);
   rclcpp::Duration assignment_op_duration = rclcpp::Duration(1, 0);
+  (void)assignment_op_duration;
```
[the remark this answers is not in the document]
wjwwood: Because otherwise it is unused, again this is a quirk of cppcheck when it is fed a single file (probably cannot tell if `rclcpp::Duration` is RAII or not).

rclcpp/test/test_publisher.cpp
wjwwood: Work in progress, removed just now since I fixed it.

rclcpp/test/test_publisher.cpp
wjwwood: Same.

rclcpp/test/test_publisher.cpp
wjwwood: Same too :)

wjwwood:
For those wondering (/cc @hidmic) 0f80c0c fixes the failure to deduce the template argument... and 7c59d90 fixes the default constructor for the default argument of type `PublisherOptionsWithAllocator`... 💢

hidmic:
https://github.com/ros2/rclcpp/commit/0f80c0c5e22f00e6a23c616296a64c625ef14b92?!? a global Alloc type?

https://github.com/ros2/rclcpp/commit/7c59d9045b6a40a24d6bb0d7e0f46dbd8e0b3f5e?!?!? :o

rclcpp/src/rclcpp/qos.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+QoSInitialization
+QoSInitialization::from_rmw(const rmw_qos_profile_t & rmw_qos)
+{
+  switch (rmw_qos.history) {
+    case RMW_QOS_POLICY_HISTORY_KEEP_ALL:
+      return KeepAll();
+    case RMW_QOS_POLICY_HISTORY_KEEP_LAST:
+    case RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT:
+    case RMW_QOS_POLICY_HISTORY_UNKNOWN:
```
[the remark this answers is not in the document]
ivanpauno: Some more thoughts about this:
I think it's ok assuming that `SYSTEM_DEFAULT` is `KEEP_LAST` (that's the case in fastrtps, connext and opensplice), but I don't know if it's the "correct" thing to do. That was not the purpose of `RMW_QOS_*_SYSTEM_DEFAULT` defines.
In my opinion, is better to raise an error both in `UNKNOWN` and `SYSTEM_DEFAULT`.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -128,6 +131,22 @@ class LifecycleNode : public node_interfaces::LifecycleNodeInterface,
...
+   * \return Shared pointer to the created lifecycle publisher.
+   */
+  template<typename MessageT, typename AllocatorT = std::allocator<void>>
+  std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<MessageT, AllocatorT>>
+  create_publisher(
+    const std::string & topic_name,
+    const rclcpp::QoS & qos,
+    const rclcpp::PublisherOptionsWithAllocator<AllocatorT> & options = (
+      rclcpp::PublisherOptionsWithAllocator<AllocatorT>()
+    ));
```
ivanpauno: Maybe. is better to apply https://github.com/ros2/rclcpp/commit/7c59d9045b6a40a24d6bb0d7e0f46dbd8e0b3f5e style change here too. If in this case windows is happy, I don't mind.

rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+// If NodeType has a method called get_node_topics_interface() which returns a const pointer.
+template<
+  typename NodeType,
+  typename std::enable_if<has_get_node_topics_interface<
+    typename std::remove_pointer<NodeType>::type,
+    const rclcpp::node_interfaces::NodeTopicsInterface *
+  >::value, int>::type = 0
+>
+rclcpp::node_interfaces::NodeTopicsInterface *
```
ivanpauno: Shouldn't return type be `const rclcpp::node_interfaces::NodeTopicsInterface *` here?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
wjwwood: Oh, that's an accident. I'll have to add back the extra `()` to keep uncrustify happy, but for the namespace, I would prefer to use them where I can, it's just using them in the node.hpp, for what ever reason, causes the compiler error...

rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp
wjwwood: I had that for a while, but removed it because I actually needed a non-const version, I think I just forgot to update the doc string.

ivanpauno: Yes, but this is looking for a `get_node_topics_interface()` method in `NodeType` returning a const pointer (see L109 below). So, it can't be converted to a non-const pointer.

https://github.com/ros2/rclcpp/blob/0f80c0c5e22f00e6a23c616296a64c625ef14b92/rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp#L104-L116

Maybe this case can be deleted.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
wjwwood: 😭 I have to remove the namespace ___AND___ the extra parenthesis to avoid the error... It's the same error with one or the other or both... wtf Windows.

wjwwood: At least in `node.hpp` I do, here I think the `rclcpp::` is required, since it's from a different namespace.

wjwwood:
I think I fixed all of the Windows issues...

Preliminary CI to see where we are at:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7000/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3249/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5742/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6821/)

rclcpp/include/rclcpp/subscription_traits.hpp
```diff
@@ -69,7 +73,19 @@ template<typename MessageT, typename Deleter>
 struct extract_message_type<std::unique_ptr<MessageT, Deleter>>: extract_message_type<MessageT>
 {};
 
-template<typename CallbackT>
+template<
+  typename CallbackT,
+  // Do not attempt if CallbackT is an integer (mistaken for depth)
+  typename = std::enable_if_t<!std::is_integral<CallbackT>::value>,
+  // Do not attempt if CallbackT is a QoS (mistaken for qos)
+  typename = std::enable_if_t<!std::is_same<
```
ivanpauno: I think that using here `std::is_base_of` will solve the last build problem in macOS.

rclcpp/include/rclcpp/qos.hpp
```diff
@@ -0,0 +1,203 @@
...
+};
+
+class SensorDataQoS : public QoS
+{
+public:
+  explicit
+  SensorDataQoS(
+    const QoSInitialization & qos_initialization = (
+      QoSInitialization::from_rmw(rmw_qos_profile_sensor_data)
+  ));
```
ivanpauno: See https://github.com/ros2/demos/pull/332#discussion_r281618995.
Same for ParameterQoS, etc.

rclcpp/include/rclcpp/subscription_traits.hpp
wjwwood: Yup, sounds right to me, thanks!

rclcpp/include/rclcpp/qos.hpp
wjwwood: We'll leave this as-is, see the discussion on the linked comment thread.

rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp
wjwwood: Ok, this took a while... But I got this part of the code sorted out I think... I reallllllly want C++17 :)

f23e462

wjwwood:
Another round:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7014/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3257/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5752/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6832/)

wjwwood:
Another, now that I have actually pushed :), set of CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7019/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3258/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5753/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6833/)

wjwwood:
Fixed everything I could see from the last CI, including disabling my "build failure" tests because Jenkins thinks they are warnings to report, but the output is expected. New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7020/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3259/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5754/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6834/)

wjwwood:
Missed some more visibility macros, don't know why that didn't show up locally... Another CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7021/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3260/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5755/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6835/)

wjwwood:
The macOS failures were on nightly too:

https://ci.ros2.org/view/nightly/job/nightly_osx_release/1252/testReport/

wjwwood:
Fixed another case of `no appropriate default constructor available` (the error from hell):

Windows [Build Status](https://ci.ros2.org/job/ci_windows/6836/)

wjwwood:
I was trying to be "nice" and put the work around functions from https://github.com/ros2/rclcpp/pull/713/commits/def728c04f505a403c0892a886d9b460ce0ee219 into a `detail` namespace, but that breaks the build in Windows still... (I expect the Windows CI in the previous comments to fail due to this)

So I introduced https://github.com/ros2/rclcpp/pull/713/commits/c16a3eb58924ea97bee847509a25d0a8dbb7edc7 to get it compiling locally again, new CI:

Windows [Build Status](https://ci.ros2.org/job/ci_windows/6837/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -144,8 +145,27 @@ class Node : public std::enable_shared_from_this<Node>
 
   /// Create and return a Publisher.
   /**
+   * The rclcpp::QoS has several convenient constructors, including a
+   * conversion constructor for size_t, which mimics older API's that
+   * allows just a string and size_t to create a pull request.
```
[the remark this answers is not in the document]
wjwwood: 🙃

wjwwood:
TODO after merging:

- address https://github.com/ros2/rclcpp/pull/713#discussion_r280591043
- update migration guide for Dashing

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -206,18 +145,18 @@ Node::create_subscription(
   typename rclcpp::message_memory_strategy::MessageMemoryStrategy<
     typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, Alloc>::SharedPtr
   msg_mem_strat,
-  std::shared_ptr<Alloc> allocator,
-  IntraProcessSetting use_intra_process_comm)
+  std::shared_ptr<Alloc> allocator)
 {
+  rclcpp::QoS qos(rclcpp::QoSInitialization::from_rmw(qos_profile));
+  qos.get_rmw_qos_profile() = qos_profile;
```
jacobperron: Here we're overriding the special logic in the QoS constructor. In case the history is set to `RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT` or `RMW_QOS_POLICY_HISTORY_UNKNOWN`, it won't default to `KeepAll`. I think this should be replaced with:

```c++
  rclcpp::QoS qos(rclcpp::QoSInitialization::from_rmw(qos_profile), qos_profile);
```

This occurs in a few other places following the use of `QoSInitialization::from_rmw`.

wjwwood: The reason this doesn't work is that `from_rmw` doesn't handle `RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT` or `RMW_QOS_POLICY_HISTORY_UNKNOWN` properly, and so by doing `qos.get_rmw_qos_profile() = qos_profile` after construction we ensure that we store all the settings from `qos_profile`, and the initialization I use is just to successfully create the QoS in the first place, it doesn't actually matter. In fact in at least one place I think I initialized with `rclcpp::KeepAll()` just to make it shorter, before I decided to use the `from_rmw` in a few places.

This is super awkward, I know, but it's also mostly only to support deprecated use cases. So I expect cases like this to go away eventually.

chapulina:
Should we add migration notes to https://index.ros.org/doc/ros2/Releases/Release-Dashing-Diademata/#changes-since-the-crystal-release ?

PR #714, opened 2019-05-02
check for nullptr before publishing event

alsora:
Setting `start_parameter_event_publisher` to false causes a segmentation fault because a check for nullptr is missing before publishing.

@raghaprasad


dirk-thomas:
Thanks for the fix.

Same logic as in other places, e.g.: https://github.com/ros2/rclcpp/blob/59d59b0c1821636d798be419336a7ac7fe477f5f/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L519-L523

PR #725, opened 2019-05-18
Volatile is a keyword

emersonknapp:
I realize why the original was spelled wrong, volatile is a c++ keyword, so https://github.com/ros2/rclcpp/pull/724 was a bad change. Modify to not require learning a misspelling, and note why it can't have that name

So sorry about that @dirk-thomas, I was overeager and didn't verify the build properly.

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/rclcpp/include/rclcpp/qos.hpp
+++ b/rclcpp/include/rclcpp/qos.hpp
@@ -105,8 +105,11 @@ public:
   durability(rmw_qos_durability_policy_t durability);
 
   /// Set the durability setting to volatile.
+  /**
+    * Note that this cannot be named `volatile` because it is a C++ keyword.
+    */
   QoS &
-  volatile();
+  durability_volatile();
 
   /// Set the durability setting to transient local.
   QoS &
--- a/rclcpp/src/rclcpp/qos.cpp
+++ b/rclcpp/src/rclcpp/qos.cpp
@@ -120,7 +120,7 @@ QoS::durability(rmw_qos_durability_policy_t durability)
 }
 
 QoS &
-QoS::volatile()
+QoS::durability_volatile()
 {
   return this->durability(RMW_QOS_POLICY_DURABILITY_VOLATILE);
 }


rclcpp/include/rclcpp/qos.hpp
```diff
@@ -105,8 +105,11 @@ class RCLCPP_PUBLIC QoS
   durability(rmw_qos_durability_policy_t durability);
 
   /// Set the durability setting to volatile.
+  /**
+    * Note that this cannot be named `volatile` because it is a C++ keyword.
+    */
   QoS &
-  volatile();
+  durability_volatile();
```
wjwwood: It's not very important, but when reviewing these pr's (thanks for fixing my spelling error btw :D) I thought that `volatile_durability` might make more sense, because all the other QoS start with their name, i.e. if I see someone used `qos.transient_local()`, I'd expect to be able to type in my editor `qos.vola` and see it auto complete. I might be a bit surprised to see `qos.volatile_durability()` rather than `qos.volatile()`, but at least it would follow the pattern of the other QoS. Instead I'd have to guess to type `qos.dur` and see both this option and `qos.durability(...)`, which is why I think it's not such a big deal, but I think swapping the words makes slightly more sense to me.

What do you guys think?

dirk-thomas:
> I thought that volatile_durability might make more sense,

:+1:

rclcpp/include/rclcpp/qos.hpp
dirk-thomas: > I thought that volatile_durability might make more sense,

:+1:

emersonknapp: I don't feel strongly either way, your argument makes sense to me - though my mental model groups them strongly by qos policy, so seeing the stripped-down names felt a little confusing to me personally. e.g. one wonders if `policy.volatile().transient_local();` is some legit combination, whereas it's obvious that an overwriting is happening if the user is forced to specify the policy type, `policy.durability(volatile).durability(transient_local)`

wjwwood: Yeah, these were meant to be as short as possible and not necessarily as explicit as possible. So they could become a bit confusing.

> one wonders if policy.volatile().transient_local(); is some legit combination, whereas it's obvious that an overwriting is happening if the user is forced to specify the policy type, policy.durability(volatile).durability(transient_local)

Yeah, I see the confusion. I had considered making object for each QoS setting, so you could do either `qos.durability(rclcpp::Volatile).durability(rclcpp::TransientLocal)` or something like `qos | rclcpp::Durability::Volatile | rclcpp::Durability::TransientLocal`.
But I didn't want them unnamespaced, and with the namespace they're both pretty verbose.
So the methods were a bit of a compromise.

If we want to include the QoS category in each "set action" for clarity, I actually prefer renaming them all to start with their category, e.g. `transient_local()` -> `durability_transient_local()` or `best_effort()` -> `history_best_effort()`.

I don't see better alternative. We could also consider removing them (they're just convenience anyways), but I'd like them to stay unless we just don't want them.

----

Short of any of those things, I do think that we should rename them from `durability_volatile` to `volatile_durability`, at least until we come up with a better way to name them or use them that's more clear.

If you are ok with that @emersonknapp, could you let me know so I can make a pr for it (or if you have time make the pr and request a review from me)? That way we can get it in before the distro freeze and people starting using the current version.

emersonknapp: Yeah, reversing the name fine with me. I agree that using `enum class`es would be much more verbose. I think the users will be ok, the API documentation does say which policy each method is configuring.

PR #735, opened 2019-05-23
Enforce parameter ranges

### LINKED ISSUE #728: Range support for parameters
## Feature request
Support enforcing parameter ranges according to ParameterDescriptor.

#### Feature description
See ros2/rcl_interfaces#72 for related discussion.

Current parameter API allows setting descriptors, but the ranges are not currently being enforced.

ivanpauno:
Closes #728.

As discussed in #728, I'm not applying type enforcement. Commits https://github.com/ros2/rclcpp/commit/4661bf3cc69e1f2125178c1f1072d120269ad260 and https://github.com/ros2/rclcpp/commit/479940c17acdce29d563d5d684b862078e727800 are reverting type enforcement (I started implementing it).

I have to add more test cases, but it would be good to get an early review.

### DIFF
--- a/rclcpp/include/rclcpp/exceptions.hpp
+++ b/rclcpp/include/rclcpp/exceptions.hpp
@@ -198,13 +198,6 @@ class InvalidParameterValueException : public std::runtime_error
   using std::runtime_error::runtime_error;
 };
 
-/// Thrown if passed parameter description is invalid.
-class InvalidParameterDescriptionException : public std::runtime_error
-{
-  // Inherit constructors from runtime_error.
-  using std::runtime_error::runtime_error;
-};
-
 /// Thrown if parameter is already declared.
 class ParameterAlreadyDeclaredException : public std::runtime_error
 {
--- a/rclcpp/include/rclcpp/exceptions.hpp
+++ b/rclcpp/include/rclcpp/exceptions.hpp
@@ -198,6 +198,13 @@ class InvalidParameterValueException : public std::runtime_error
   using std::runtime_error::runtime_error;
 };
 
+/// Thrown if passed parameter description is invalid.
+class InvalidParameterDescriptionException : public std::runtime_error
+{
+  // Inherit constructors from runtime_error.
+  using std::runtime_error::runtime_error;
+};
+
 /// Thrown if parameter is already declared.
 class ParameterAlreadyDeclaredException : public std::runtime_error
 {
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -191,7 +191,7 @@ __lockless_has_parameter(
 // see https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon
 RCLCPP_LOCAL
 bool
-__are_doubles_equal(double x, double y, int ulp = 100)
+__are_doubles_equal(double x, double y, size_t ulp = 100)
 {
   return std::abs(x - y) <= std::numeric_limits<double>::epsilon() * std::abs(x + y) * ulp;
 }
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -263,7 +263,6 @@ __check_parameter_value_in_range(
   return result;
 }
 
-
 // Return true if parameter values comply with the descriptors in parameter_infos.
 RCLCPP_LOCAL
 rcl_interfaces::msg::SetParametersResult
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -260,8 +260,6 @@ __check_parameters(
   for (const rclcpp::Parameter & parameter : parameters) {
     const rcl_interfaces::msg::ParameterDescriptor & descriptor =
       parameter_infos[parameter.get_name()].descriptor;
-    result = descriptor.type == parameter.get_type() ||
-      descriptor.type == rclcpp::PARAMETER_NOT_SET;
     result = result && __check_parameter_value_in_range(
       descriptor,
       parameter.get_parameter_value());
@@ -295,8 +293,7 @@ __set_parameters_atomically_common(
   if (result.successful) {
     for (size_t i = 0; i < parameters.size(); ++i) {
       const std::string & name = parameters[i].get_name();
-      // TODO(ivanpauno): Why updating the descriptor name in each set?
-      // Maybe, set the correct name when declare parameter is call, and delete it here.
+      parameter_infos[name].descriptor.type = parameters[i].get_type();
       parameter_infos[name].descriptor.name = parameters[i].get_name();
       parameter_infos[name].value = parameters[i].get_parameter_value();
     }
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -252,7 +252,7 @@ __check_parameter_value_in_range(
     if (fp_range.step == 0.0) {
       return result;
     }
-    long rounded_div = std::lround((v - fp_range.from_value) / fp_range.step);
+    double rounded_div = std::round((v - fp_range.from_value) / fp_range.step);
     if (__are_doubles_equal(v, fp_range.from_value + rounded_div * fp_range.step)) {
       return result;
     }
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -197,28 +197,41 @@ __are_doubles_equal(double x, double y, int ulp = ULP)
   return std::abs(x - y) <= std::numeric_limits<double>::epsilon() * std::abs(x + y) * ulp;
 }
 
+#define FORMAT_REASON(reason, name, range_type) \
+  { \
+    std::ostringstream ss; \
+    ss << "Parameter {" << name << "} doesn't comply with " << range_type << " range."; \
+    reason = ss.str(); \
+  }
+
 RCLCPP_LOCAL
-bool
+rcl_interfaces::msg::SetParametersResult
 __check_parameter_value_in_range(
   const rcl_interfaces::msg::ParameterDescriptor & descriptor,
   const rclcpp::ParameterValue & value)
 {
+  rcl_interfaces::msg::SetParametersResult result;
+  result.successful = true;
   if (!descriptor.integer_range.empty() && value.get_type() == rclcpp::PARAMETER_INTEGER) {
     int64_t v = value.get<int64_t>();
     auto integer_range = descriptor.integer_range.at(0);
     if ((v < integer_range.from_value) || (v > integer_range.to_value)) {
-      return false;
+      result.successful = false;
+      FORMAT_REASON(result.reason, descriptor.name, "integer")
+      return result;
     }
-    if (v == integer_range.to_value) {
-      return true;
+    if (v == integer_range.from_value || v == integer_range.to_value) {
+      return result;
     }
     if (integer_range.step == 0) {
-      return true;
+      return result;
     }
     if (((v - integer_range.from_value) % integer_range.step) == 0) {
-      return true;
+      return result;
     }
-    return false;
+    result.successful = false;
+    FORMAT_REASON(result.reason, descriptor.name, "integer")
+    return result;
   }
 
   if (!descriptor.floating_point_range.empty() && value.get_type() == rclcpp::PARAMETER_DOUBLE) {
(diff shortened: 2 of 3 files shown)


ivanpauno:
CI, just to see is this is breaking sth (only fastrtps):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7218/)

rclcpp/include/rclcpp/exceptions.hpp
```diff
@@ -198,6 +198,13 @@ class InvalidParameterValueException : public std::runtime_error
   using std::runtime_error::runtime_error;
 };
 
+/// Thrown if passed parameter description is invalid.
+class InvalidParameterDescriptionException : public std::runtime_error
+{
+  // Inherit constructors from runtime_error.
+  using std::runtime_error::runtime_error;
+};
+
```
jubeira: Is this used somewhere? 

- Any combination of `to` / `from` and `step` are basically valid.
- The name and type of a given descriptor are ignored, so it matter if they don't match the name and type of the associated parameter.
- Having only an integer range, only a floating point range, both at the same time or none shouldn't cause trouble.

Is there something I'm missing?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,88 @@ __lockless_has_parameter(
...
+  const rcl_interfaces::msg::ParameterDescriptor & descriptor,
+  const rclcpp::ParameterValue & value)
+{
+  if (!descriptor.integer_range.empty() && value.get_type() == rclcpp::PARAMETER_INTEGER) {
+    int64_t v = value.get<int64_t>();
+    auto integer_range = descriptor.integer_range.at(0);
+    if ((v < integer_range.from_value) || (v > integer_range.to_value)) {
+      return false;
+    }
+    if (v == integer_range.to_value) {
```
jubeira: Both bounds `to_value` and `from_value` [are valid](https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/msg/IntegerRange.msg#L14-L22).

jubeira:
```suggestion
    if (v == integer_range.from_value || v == integer_range.to_value) {
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,88 @@ __lockless_has_parameter(
...
+    }
+    return false;
+  }
+
+  if (!descriptor.floating_point_range.empty() && value.get_type() == rclcpp::PARAMETER_DOUBLE) {
+    double v = value.get<double>();
+    auto fp_range = descriptor.floating_point_range.at(0);
+    if (__are_doubles_equal(v, fp_range.from_value) ||
+      __are_doubles_equal(v, fp_range.to_value))
+    {
```
jubeira: nitpick: move `{` to the line above.

ivanpauno:
Same in the remaining platforms:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3417/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5913/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7044/)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,88 @@ __lockless_has_parameter(
...
+      return true;
+    }
+    if ((v < fp_range.from_value) || (v > fp_range.to_value)) {
+      return false;
+    }
+    if (fp_range.step == 0.0) {
+      return true;
+    }
+    int truncated_div = (v - fp_range.from_value) / fp_range.step;
```
jubeira: Instead of checking for `truncated_div` and `trincated_div + `, you could `round` the result here instead of truncating.

I'll exaggerate the math a bit to illustrate the point. Imagine `__are_doubles_equal` allows a tolerance of 0.2, that you have a step of 0.5 from 0 to 10, and you set 4.9 as the value. If you truncate, the `truncated_div` will be `floor((4.9 - 0) / 0.5)`, which would be `floor(9.8) = 9`. When you apply `__are_doubles_equal` in the line below, `from + 9 * step` will be 4.5, which is too far away from the value for the check to pass. When you apply `__are_doubles_equal` with `truncated_div + 1`, it will pass.

If instead of truncating by casting to int you round this calculation, 4.9 would round to 10 steps, and the first check in the line below would pass (which I think is a bit simpler than applying `__are_doubles_equal` twice).

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,88 @@ __lockless_has_parameter(
...
+    if (__are_doubles_equal(v, fp_range.from_value + (++truncated_div) * fp_range.step)) {
+      return true;
+    }
+    return false;
+  }
+  return true;
+}
+
+
+// Return true if parameters type and range conform the descriptors in parameter_infos.
```
jubeira:
```suggestion
// Return true if parameter values comply with the descriptors in parameter_infos.
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -201,13 +286,15 @@ __set_parameters_atomically_common(
...
+  result.successful =
+    result.successful && __check_parameters(parameter_infos, parameters);
 
   // If accepted, actually set the values.
   if (result.successful) {
     for (size_t i = 0; i < parameters.size(); ++i) {
       const std::string & name = parameters[i].get_name();
-      parameter_infos[name].descriptor.name = parameters[i].get_name();
       parameter_infos[name].descriptor.type = parameters[i].get_type();
+      parameter_infos[name].descriptor.name = parameters[i].get_name();
```
jubeira: This is really a detail, but given that the name shouldn't change after the descriptor is set the first time, maybe it's not necessary to set the name of the descriptor every time a parameter is set.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -201,13 +286,15 @@ __set_parameters_atomically_common(
   if (on_set_parameters_callback) {
     result = on_set_parameters_callback(parameters);
   }
+  result.successful =
+    result.successful && __check_parameters(parameter_infos, parameters);
```
jubeira: It's fine for `__check_parameters` to return a bool; maybe I'd add a reason here in case it doesn't succeed.

rclcpp/include/rclcpp/exceptions.hpp
ivanpauno: Forgot to delete it. I was going to use it, but I finally didn't.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Sere. It success later in one of these two places:
https://github.com/ros2/rclcpp/blob/ac852902ae2acc84c328de3661202f732ea7180a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L215-L220

But it's maybe clearer to do it in the way you propose.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: It's not recommended when the condition of the if is multi-line, because L228 and L230 will have the same indentation. Actually, uncrustify will complain.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Yeah, I agree. The original code was doing that.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Thanks! I forgot checking what `rcl_interfaces::msg::SetParametersResult` was.

ivanpauno:
Again CI (fastrtps):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7231/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3425/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5921/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7055/)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,102 @@ __lockless_has_parameter(
...
+// see https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon
+#define ULP 100
+RCLCPP_LOCAL
+bool
+__are_doubles_equal(double x, double y, int ulp = ULP)
+{
+  return std::abs(x - y) <= std::numeric_limits<double>::epsilon() * std::abs(x + y) * ulp;
+}
+
+#define FORMAT_REASON(reason, name, range_type) \
```
wjwwood: I would suggest you make this a function instead of a macro.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,102 @@ __lockless_has_parameter(
   return parameters.find(name) != parameters.end();
 }
 
+// see https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon
+#define ULP 100
```
wjwwood: Is this define better than just putting it as the default argument in the declaration? I don't see this used but in one place...

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,102 @@ __lockless_has_parameter(
...
+RCLCPP_LOCAL
+bool
+__are_doubles_equal(double x, double y, int ulp = ULP)
```
wjwwood: I know the example on cppreference uses an integer, but can `ulp` not be unsigned, e.g. `size_t`?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,102 @@ __lockless_has_parameter(
...
+      return result;
+    }
+    result.successful = false;
+    FORMAT_REASON(result.reason, descriptor.name, "floating point")
+    return result;
+  }
+  return result;
+}
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
wjwwood: Maybe this diff should be eliminated? Not sure what putting the assignment of the name below the assignment of the type does, but I think nothing.

rclcpp/test/test_node.cpp
```diff
@@ -695,6 +695,141 @@ TEST_F(TestNode, set_parameter_undeclared_parameters_not_allowed) {
...
+    auto value = node->get_parameter(name);
+    EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_INTEGER);
+    EXPECT_EQ(value.get_value<int64_t>(), 42);
+
+    EXPECT_TRUE(node->set_parameter(rclcpp::Parameter(name, "asd")).successful);
+    EXPECT_TRUE(node->has_parameter(name));
+    value = node->get_parameter(name);
+    EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_STRING);
+    EXPECT_EQ(value.get_value<std::string>(), "asd");
+  }
```
wjwwood: I think there needs to be a test for the cases where: `from` is greater than `to`, `from` and `to` are equal, `step` > `distance(from, to)`, `step` is not a multiple of the `distance(from, to)`. Basically all of the special cases described here:

https://github.com/ros2/rcl_interfaces/blob/7e2e6a7578f588e775f4f1f3b5b8500e3fcbb4f0/rcl_interfaces/msg/IntegerRange.msg#L9-L21

Also a case for when step in the floating point range is negative.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Sorry, didn't realize that I moved it. At some point I deleted it, and then I added it again in the wrong line.

ivanpauno:
Again CI (fastrtps):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7253/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3434/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5931/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7066/)

rclcpp/test/test_node.cpp
ivanpauno: Added in:
https://github.com/ros2/rclcpp/pull/735/commits/964abc7af7904299df6959ad1d04ce1f48f7c459
Corrected found bug in:
https://github.com/ros2/rclcpp/pull/735/commits/4ecc85f356ffe8ca082e8503f05fc9fa03bc3aab

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Addressed in https://github.com/ros2/rclcpp/pull/735/commits/31a0f2ba353f4fe286e3b06d1f684c4145c69899

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Addressed in https://github.com/ros2/rclcpp/pull/735/commits/2755108e124d6ceb7617bd90f32784c2a5d3af66.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/735/commits/b8f48e1e6e629a3c6f8b912d9401bf1d7630fd60

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/735/commits/d91b2c4a17e961a3d7b9acc963f4db9404b2cf60

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/735/commits/b43d35c9647f8311ba0152b105f233de75586588

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -185,6 +188,103 @@ __lockless_has_parameter(
...
+    }
+    if ((v < fp_range.from_value) || (v > fp_range.to_value)) {
+      result.successful = false;
+      format_reason(result.reason, descriptor.name, "floating point");
+      return result;
+    }
+    if (fp_range.step == 0.0) {
+      return result;
+    }
+    int rounded_div = std::round((v - fp_range.from_value) / fp_range.step);
```
dirk-thomas: Compiler warning on Windows: https://ci.ros2.org/job/ci_windows/7066/warnings43Result/new/

ivanpauno:
Again CI (fastrtps):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7257/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3438/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5936/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7071/)

ivanpauno:
@jubeira sorry for the silly change I did. But I don't know why I used `lround` instead of chaning `rounded_div` type to double.

Again CI (fastrtps, up to rclcpp):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7259/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3439/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5937/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7072/)

Issue #730, opened 2019-05-21
Can't set initial parameters when allowing undeclared


chapulina:
## Bug report

**Required Info:**

- Operating System:
  - Ubunto Bionic
- Installation type:
  - dashing sources
- Version or commit hash:
  - master

#### Steps to reproduce issue

Setting to accept undeclared parameters and passing initial parameters:

```.cpp
  std::vector<rclcpp::Parameter> initial_parameters =
      {rclcpp::Parameter("my_string", std::string("str"))};
  rclcpp::NodeOptions node_options;
  node_options.allow_undeclared_parameters(true);
  node_options.initial_parameters(initial_parameters);
  auto node = std::make_shared<Node>("name", "ns", node_options);
  node->get_parameter(param);  // throws
```
Throws this:

```
12: terminate called after throwing an instance of 'rclcpp::ParameterTypeException'            
12:   what():  expected [string] got [not set]  
```

On the other hand, setting the parameters after construction is happy:

```.cpp
  rclcpp::NodeOptions node_options;
  node_options.allow_undeclared_parameters(true);
  auto node = std::make_shared<Node>("name", "ns", node_options);
  node->set_parameters(initial_parameters);
  node->get_parameter(param); // happy
```

#### Expected behavior

No throw on both cases

#### Actual behavior

Throw on first case

dirk-thomas:
What is `initial_parameters` / `param` in your example?

chapulina:
> What is initial_parameters / param in your example?

Added to the description

ivanpauno:
I haven't been able to reproduce the bug you're reporting in master.

Initial parameters aren't supposed to be automatically declared, except you use `node_options.automatically_declare_initial_parameters(true)`: https://github.com/ros2/rclcpp/blob/30df5cdf36b56aa530e7cca6466d9df88cbf20b9/rclcpp/include/rclcpp/node_options.hpp#L260-L272.

If you add a `has_parameter` call, you will see that the parameter hasn't been declared.

```cpp
  std::vector<rclcpp::Parameter> initial_parameters =
      {rclcpp::Parameter("my_string", std::string("str"))};
  rclcpp::NodeOptions node_options;
  node_options.allow_undeclared_parameters(true);
  node_options.initial_parameters(initial_parameters);
  auto node = std::make_shared<Node>("name", "ns", node_options);
  printf("has_parameter: %d\n", node->has_parameter("my_string")); // prints 0
```

If you add the following line before creating the node:
```cpp
[...]

PR #739, opened 2019-05-24
fix behavior of get_parameter when used with allow_undeclared_parameter

wjwwood:
Should fix https://github.com/ros2/rclcpp/issues/730.

I think I missed this case when testing because I tested `declare_parameter` in combination with `allow_undeclared_parameters` AND `rclcpp::NodeOptions::initial_parameters`, but for `get_parameter` I only tested with and without `allow_undeclared_parameters` but not in conjunction with `rclcpp::NodeOptions::initial_parameters`.

@chapulina I don't know how easy it is for you to try this pr out, but hopefully it should make your porting only _require_ switching the `allows_undeclared_parameter` flag to `true`, even if you decide to do more changes and use `declare_parameters`.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -649,8 +649,11 @@ public:
    *
    * If undeclared parameters are allowed, see the node option
    * rclcpp::NodeOptions::allow_undeclared_parameters, then this method will
-   * not throw an exception, and instead return a default initialized
-   * rclcpp::Parameter, which has a type of
+   * not throw an exception.
+   * It will instead attempt to get a value from the "initial parameter values",
+   * i.e. from rclcpp::NodeOptions::initial_parameters or a YAML file on the
+   * command line, and if it is not found there either, it will return a
+   * default initialized rclcpp::Parameter, which has a type of
    * rclcpp::ParameterType::PARAMETER_NOT_SET.
    *
    * \param[in] name The name of the parameter to get.
@@ -670,6 +673,16 @@ public:
    *
    * If the parameter was not declared, then the output argument for this
    * method which is called "parameter" will not be assigned a value.
+   *
+   * However, if undeclared parameters are allowed, see the node option
+   * rclcpp::NodeOptions::allow_undeclared_parameters, this function will
+   * instead attempt to get a value from the "initial parameter values",
+   * i.e. from rclcpp::NodeOptions::initial_parameters or a YAML file on the
+   * command line.
+   * If a value is found from there, true will returned.
+   * If not found there, it will not assign the "parameter" arugment and then
+   * return false.
+   *
    * If the parameter was declared, and therefore has a value, then it is
    * assigned into the "parameter" argument of this method.
    *
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -559,6 +559,8 @@ NodeParameters::get_parameter(const std::string & name) const
   if (get_parameter(name, parameter)) {
     return parameter;
   } else if (this->allow_undeclared_) {
+    // This case is only reached if the parameter name was found neither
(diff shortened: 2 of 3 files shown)


chapulina:
@ivanpauno brings up several good questions about how the rest of the API should work in "undeclared parameter mode". I wrote a couple of tests to illustrate that and also to point out how passing parameters in the constructor still works differently from setting them afterwards.

All these tests pass with this branch, see questions inline:

```.diff
diff --git a/rclcpp/test/test_node.cpp b/rclcpp/test/test_node.cpp
index 456a3f2..e088b5d 100644
--- a/rclcpp/test/test_node.cpp
+++ b/rclcpp/test/test_node.cpp
@@ -697,6 +697,70 @@ TEST_F(TestNode, set_parameter_undeclared_parameters_not_allowed) {
   }
 }
 
+TEST_F(TestNode, initial_parameters_undeclared_parameters_allowed) {
+  // Parameter at construction
+  rclcpp::NodeOptions node_options;
+  node_options.allow_undeclared_parameters(true);
+  node_options.initial_parameters({rclcpp::Parameter("my_string", std::string("str"))});
+  auto node = std::make_shared<rclcpp::Node>("test_set_parameter_node"_unq, node_options);
+
+  // Check parameter
+  auto value = node->get_parameter("my_string").get_value<std::string>();
+  EXPECT_EQ("str", value);
+  EXPECT_FALSE(node->has_parameter("my_string"));
+
+  // undeclare so that it doesn't interfere with the test
+  node->undeclare_parameter("use_sim_time");
+
+  // QUESTION Is it expected that the undeclared parameter comes up with `get_parameter` but not
+  // with `get_parameters` and `list_parameters`? This can be confusing.
+  {
+    std::map<std::string, std::string> values;
+    EXPECT_FALSE(node->get_parameters("", values));
+    EXPECT_EQ(0u, values.size());
+  }
+
+  {
+    auto values = node->list_parameters({""}, 1);
+    EXPECT_EQ(0u, values.names.size());
+    EXPECT_EQ(0u, values.prefixes.size());
+  }
+}
+
+TEST_F(TestNode, set_parameters_undeclared_parameters_allowed_initial) {
+  // Parameter after construction
+  rclcpp::NodeOptions node_options;
+  node_options.allow_undeclared_parameters(true);
+  auto node = std::make_shared<rclcpp::Node>("test_set_parameter_node"_unq, node_options);
+  node->set_parameters({rclcpp::Parameter("my_string", std::string("str"))});
+
+  // Check parameter
+  auto value = node->get_parameter("my_string").get_value<std::string>();
+  EXPECT_EQ("str", value);
+
+  // undeclare so that it doesn't interfere with the test
+  node->undeclare_parameter("use_sim_time");
+
+  // QUESTION Should we expect `has_parameter` and `get_parameters` to have different results from
+  // the test above?
+  EXPECT_TRUE(node->has_parameter("my_string"));
+
+  {
+    std::map<std::string, std::string> values;
+    EXPECT_TRUE(node->get_parameters("", values));
+    ASSERT_EQ(1u, values.size());
+    EXPECT_EQ("my_string", values.begin()->first);
+    EXPECT_EQ("str", values.begin()->second);
+  }
+
+  // QUESTION is it expected that `list_parameters` has a different result from `get_parameters`?
+  {
+    auto values = node->list_parameters({""}, 1);
+    EXPECT_EQ(0u, values.names.size());
+    EXPECT_EQ(0u, values.prefixes.size());
+  }
+}
+
 TEST_F(TestNode, set_parameter_undeclared_parameters_allowed) {
   auto node = std::make_shared<rclcpp::Node>(
     "test_set_parameter_node"_unq,
```

wjwwood:
> With this as-is, get_parameter returns the value loaded with initial_parameters. But calling has_parameter returns false, which is quite strange.

I agree, in retrospect maybe what's on master is what I intended all along. I think instead, my mistake was in the documentation that `allow_undeclared_parameter` was the only option needed to get back to the "crystal behavior". Which makes some sense because the `automatically_declare_initial_parameters` option was a late addition to the pull request.

I'll close this and continue the discussion on the other issue.

PR #741, opened 2019-05-24
ensure removal of guard conditions of expired nodes from memory strategy

### LINKED ISSUE #272: destroyed nodes can cause Executor to crash
User error alert! This is probably because I'm still learning C++11 :face_with_head_bandage: but I suspect I might not be the only person who tries to do this:
```
executor.add_node(std::make_shared<MyNodeSubclass>());
executor.spin();
```
which results in this:
```
  what():  Invalid argument
```
with this backtrace:
```
Thread 1 "minimal_talker" received signal SIGABRT, Aborted.
0x00007ffff6647428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff6647428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff664902a in __GI_abort () at abort.c:89
#2  0x00007ffff6c8084d in __gnu_cxx::__verbose_terminate_handler() ()
#3  0x00007ffff6c7e6b6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff6c7e701 in std::terminate() ()
#5  0x00007ffff6c7e919 in __cxa_throw ()
#6  0x00007ffff6ca77fe in std::__throw_system_error(int) ()
#7  0x0000000000417d03 in std::mutex::lock() ()
#8  0x0000000000418a36 in std::lock_guard<std::mutex>::lock_guard(std::mutex&)
    ()
#9  0x00007ffff79a4ea8 in GuardCondition::attachCondition(std::mutex*, std::condition_variable*) () from /home/mquigley/ros2/install/lib/librmw_fastrtps_cpp.so
#10 0x00007ffff799d7bc in rmw_wait ()
   from /home/mquigley/ros2/install/lib/librmw_fastrtps_cpp.so
#11 0x00007ffff7767026 in rcl_wait ()
   from /home/mquigley/ros2/install/lib/librcl.so
#12 0x00007ffff7439b3e in rclcpp::executor::Executor::wait_for_work(std::chrono::duration<long, std::ratio<1l, 1000000000l> >) ()
   from /home/mquigley/ros2/install/lib/librclcpp.so
#13 0x00007ffff743ad61 in rclcpp::executor::Executor::get_next_executable(std::chrono::duration<long, std::ratio<1l, 1000000000l> >) ()
   from /home/mquigley/ros2/install/lib/librclcpp.so
#14 0x00007ffff7441f45 in rclcpp::executors::single_threaded_executor::SingleThreadedExecutor::spin() () from /home/mquigley/ros2/install/lib/librclcpp.so
#15 0x0000000000417927 in main ()
```
I'm not sure what the right answer is. If it's possible to somehow generate a warning and say something like "hey there! I hope you're having a gr8 day. You should probably hold a `shared_ptr` to that instance on your stack." But that might be hard to generate (actually I have no idea how one could generate that).

dirk-thomas:
Fixes #272.

The first commit changes the API of the memory strategy from taking a vector of nodes to list of nodes. While this change is not strictly necessary it simplifies the logic in the executor notably.

The second commit stores the guard condition point for each node so that the guard condition can be removed from the memory strategy when the node expires.

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -365,6 +365,7 @@ private:
   RCLCPP_DISABLE_COPY(Executor)
 
   std::list<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr> weak_nodes_;
+  std::list<const rcl_guard_condition_t *> guard_conditions_;
 };
 
 }  // namespace executor
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -364,7 +364,7 @@ protected:
 private:
   RCLCPP_DISABLE_COPY(Executor)
 
-  std::vector<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr> weak_nodes_;
+  std::list<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr> weak_nodes_;
 };
 
 }  // namespace executor
--- a/rclcpp/include/rclcpp/memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/memory_strategy.hpp
@@ -15,8 +15,8 @@
 #ifndef RCLCPP__MEMORY_STRATEGY_HPP_
 #define RCLCPP__MEMORY_STRATEGY_HPP_
 
+#include <list>
 #include <memory>
-#include <vector>
 
 #include "rcl/allocator.h"
 #include "rcl/wait.h"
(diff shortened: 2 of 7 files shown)


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7225/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3419/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5915/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7046/)

dirk-thomas:
> it would be nice to have a new test that exposed the old failure

Here we go: e926a067f486fc1023cf45047a88ae73dc2c38e5

dirk-thomas:
Just `rclcpp`:

nuclearsandwich:
> The first commit changes the API of the memory strategy from taking a vector of nodes to list of nodes. While this change is not strictly necessary it simplifies the logic in the executor notably.

How will consumers of this API be affected by this change?

wjwwood:
That API should not be widely used outside of the `Executor` class itself, but it is technically part of the public API.

nuclearsandwich:
> That API should not be widely used outside of the `Executor` class itself, but it is technically part of the public API.

I'd appreciate a couple of additional co-approvers to agree to this API change.

PR #742, opened 2019-05-25
Bionic use of strerror_r

roderick-koehle:
Since API 23 Android Bionic uses the GNU convention for strerror_r.

Following the string.h definition in bionic/libc, line 96
  https://android.googlesource.com/platform/bionic.git/+/refs/heads/master/libc/include/string.h
I suggest to add the API variant to select between the GNU or XSI variant.


dirk-thomas:
Thank you for the patch.

Can you please add a signed-off-by line to your commit message to satisfy the DCO check.

dirk-thomas:
Thanks for updating the patch.

Please in the future add a comment to the ticket otherwise no notification is being sent and your change might go unnoticed.

PR #745, opened 2019-05-28
Rename parameter options

wjwwood:
Connects to ros2/ros2_documentation#242

### DIFF
--- a/rclcpp/include/rclcpp/node_options.hpp
+++ b/rclcpp/include/rclcpp/node_options.hpp
@@ -247,6 +247,10 @@ public:
    * If true, allow any parameter name to be set on the node without first
    * being declared.
    * Otherwise, setting an undeclared parameter will raise an exception.
+   *
+   * This option being true does not affect parameter_overrides, as the first
+   * set action will implicitly declare the parameter and therefore consider
+   * any parameter overrides.
    */
   RCLCPP_PUBLIC
   NodeOptions &
--- a/rclcpp/include/rclcpp/node_options.hpp
+++ b/rclcpp/include/rclcpp/node_options.hpp
@@ -263,7 +263,8 @@ public:
    * implicitly declare any that have not already been declared.
    * Otherwise, parameters passed to the node's parameter_overrides, and/or the
    * global arguments (e.g. parameter overrides from a YAML file), which are
-   * not explicitly declared will not appear on the node at all.
+   * not explicitly declared will not appear on the node at all, even if
+   * `allow_undeclared_parameters` is true.
    * Already declared parameters will not be re-declared, and parameters
    * declared in this way will use the default constructed ParameterDescriptor.
    */
--- a/rclcpp/include/rclcpp/node_options.hpp
+++ b/rclcpp/include/rclcpp/node_options.hpp
@@ -49,7 +49,7 @@ public:
    *     - with history setting and depth from rmw_qos_profile_parameter_events
    *   - parameter_event_publisher_options = rclcpp::PublisherOptionsBase
    *   - allow_undeclared_parameters = false
-   *   - automatically_declare_initial_parameters = false
+   *   - automatically_declare_parameters_from_overrides = false
    *   - allocator = rcl_get_default_allocator()
    *
    * \param[in] allocator allocator to use in construction of NodeOptions.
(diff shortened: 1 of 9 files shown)


rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -107,31 +107,31 @@ class NodeOptions
...
   RCLCPP_PUBLIC
   const std::vector<rclcpp::Parameter> &
-  initial_parameters() const;
+  parameter_overrides() const;
 
-  /// Set the initial parameters, return this for parameter idiom.
+  /// Set the parameters overrides, return this for parameter idiom.
   /**
-   * These initial parameter values are used to change the initial value
+   * These parameter overrides are used to change the initial value
```
chapulina: I think they do more than changing initial values. It sounds like overrides can never be overridden themselves, even by `set_parameter` at runtime.

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -252,24 +252,25 @@ class NodeOptions
...
   /**
-   * If true, automatically iterate through the node's initial parameters and
+   * If true, automatically iterate through the node's parameter overrides and
    * implicitly declare any that have not already been declared.
-   * Otherwise, parameters passed to the node's initial_parameters, and/or the
-   * global initial parameter values, which are not explicitly declared will
-   * not appear on the node at all.
+   * Otherwise, parameters passed to the node's parameter_overrides, and/or the
+   * global arguments (e.g. parameter overrides from a YAML file), which are
+   * not explicitly declared will not appear on the node at all.
```
chapulina:
```suggestion
   * not explicitly declared will not appear on the node at all, even if
   * `allow_undeclared_parameters` is true.
```

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -252,24 +252,25 @@ class NodeOptions
   NodeOptions &
   allow_undeclared_parameters(bool allow_undeclared_parameters);
```
chapulina: Maybe add to documentation that "this option doesn't apply to parameter_overrides", and even elaborate further that "Parameter overrides can never be treated as undeclared parameters".

rclcpp/include/rclcpp/node_options.hpp
wjwwood: > I think they do more than changing initial values.

Like?

Unless you set `automatically_declare_parameters_from_overrides` to be `true`, they are not used in any other way, AFAIK.

> It sounds like overrides can never be overridden themselves, even by set_parameter at runtime.

I do not think that's true, they only override the first time they are decalred. Afterwards (after you do `declare_parameter`) if you `set_parameter` on it it will take the value given to set not the override.

What could I change to make it not sound like that?

What name would be better? `default_parameter_values_override`?

chapulina: > Afterwards (after you do declare_parameter) if you set_parameter on it it will take the value given to set not the override.

I don't think that's true:

```.diff
diff --git a/rclcpp/test/test_node.cpp b/rclcpp/test/test_node.cpp
index bb3ac35..e5a69a1 100644                    
--- a/rclcpp/test/test_node.cpp                  
+++ b/rclcpp/test/test_node.cpp                                               
@@ -380,6 +380,8 @@ TEST_F(TestNode, declare_parameter_with_initial_values) {
     rclcpp::ParameterValue value = node->declare_parameter("parameter_no_default");
     EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_INTEGER);
     EXPECT_EQ(value.get<int>(), 42);     
+    node->set_parameter(rclcpp::Parameter("parameter_no_default", 44));
+    EXPECT_EQ(value.get<int>(), 42); // I'd expect this to be 44
   }
   {
     // int default, with initial
```

wjwwood: Well that's not the way it's supposed to work 🤦‍♂

wjwwood: Whew, it's not broken, you're just reusing `value` which is a copy, not a reference. Either of these work:

```diff
diff --git a/rclcpp/test/test_node.cpp b/rclcpp/test/test_node.cpp
index 7fd0c1a..b3f3b93 100644
--- a/rclcpp/test/test_node.cpp
+++ b/rclcpp/test/test_node.cpp
@@ -377,9 +377,11 @@ TEST_F(TestNode, declare_parameter_with_initial_values) {
   auto node = std::make_shared<rclcpp::Node>("test_declare_parameter_node"_unq, no);
   {
     // no default, with initial
-    rclcpp::ParameterValue value = node->declare_parameter("parameter_no_default");
+    const rclcpp::ParameterValue & value = node->declare_parameter("parameter_no_default");
     EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_INTEGER);
     EXPECT_EQ(value.get<int>(), 42);
+    node->set_parameter({"parameter_no_default", 44});
+    EXPECT_EQ(value.get<int>(), 44);
   }
   {
     // int default, with initial

```

Or this:

```diff
diff --git a/rclcpp/test/test_node.cpp b/rclcpp/test/test_node.cpp
index 7fd0c1a..e27ac87 100644
--- a/rclcpp/test/test_node.cpp
+++ b/rclcpp/test/test_node.cpp
@@ -380,6 +380,8 @@ TEST_F(TestNode, declare_parameter_with_initial_values) {
     rclcpp::ParameterValue value = node->declare_parameter("parameter_no_default");
     EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_INTEGER);
     EXPECT_EQ(value.get<int>(), 42);
+    node->set_parameter({"parameter_no_default", 44});
+    EXPECT_EQ(node->get_parameter("parameter_no_default").get_value<int>(), 44);
   }
   {
     // int default, with initial

```

chapulina: Ohhh sorry for the noise :fearful: 

So I guess my only suggestion now would be to document somewhere that if `automatically_declare_parameters_from_overrides` + `allow_undeclared_parameters`, that makes (only the first?) `set_parameter` behave like `declare_parameter`, i.e. the value is overridden by overrides.

chapulina: ... with the understanding that people will get confused, but at least we have documentation to point them to.

wjwwood: In the process I found some more renames I missed 54c3a0d and added a test case for this 
b3d792b.

wjwwood: > Ohhh sorry for the noise 😨

That's ok haha, I'm just glad it's not broken this late in the game again.

> So I guess my only suggestion now would be to document somewhere that if allow_undeclared_parameters, that makes (only the first?) set_parameter behave like declare_parameter, i.e. the value is overridden by overrides.

Yeah, the wording for `set_parameter` is something like: if `allow_undeclared_parameters` is false it will throw, but if it is true it will implicitly declare for you on the first set, and therefore make use of the overrides when it declares.

rclcpp/include/rclcpp/node_options.hpp
wjwwood: I added a note, see: 870998d69536c339b38d035d55f0648ae841b101

PR #749, opened 2019-05-30
Avoid 'Intra process message no longer being stored when trying to ha…

ivanpauno:
Fixes https://github.com/ros2/demos/issues/362.

The comment in the code is self-explanatory.


ivanpauno:
Merging without running CI, as I only deleted `RCLCPP_WARN` calls and changed comments.

PR #753, opened 2019-06-06
checking origin of intra-process msg before taking them

alsora:
This PR fixes the following scenario.

There are two processes. In the first process there are a `Publisher` and a `Subscription` to `/MyTopic`. Both have intra-process communication enabled. In the second process there is a `Subscription` to `/MyTopic` and it has intra-process communication enabled.

The `Publisher` sends an intra-process meta-message through the RMW on the `/MyTopic/_intra` topic. Also, the `Subscription` in the second process will be listening to that topic. This results in that the meta-message will be delivered to both `Subscription`s. 

The one in the second topic will still try to use the meta-message to retrieve something from the `IntraProcessManager`.

More details https://github.com/irobot-ros/ros2-performance/pull/1

### DIFF
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -170,6 +170,13 @@ public:
       // However, this can only really happen if this node has it disabled, but the other doesn't.
       return;
     }
+
+    if (!matches_any_intra_process_publishers(&message_info.publisher_gid)) {
+      // This intra-process message has not been created by a publisher from this context.
+      // we should ignore this copy of the message.
+      return;
+    }
+
     if (any_callback_.use_take_shared_method()) {
       ConstMessageSharedPtr msg;
       take_intra_process_message(


ivanpauno:
full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7302/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3477/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5977/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7122/)

dirk-thomas:
While the patch itself looks fine to me I am wondering if something should be done additionally to avoid that other processes even receive the message in the first place.

Currently these messages are published over the topic `<topic_name>/_intra`. Maybe each process should use a process-specific topic name, e.g.: `<topic_name>/_intra_<pid>`.

alsora:
@dirk-thomas that was my first idea, however, there still may be collisions when running the nodes in different machines.

This solution is very simple and has been pointed out by @wjwwood

dirk-thomas:
> there still may be collisions when running the nodes in different machines.

Absolutely, that is why I suggested to do the topic suffix *additionally*.

Otherwise you still have these different processes match their `_intra` topics and exchange message for no reason which is unnecessary.

wjwwood:
I'm fine with adding the pid or something else (random number or whatever), so long as we try to keep it short (as anything extra cuts into the possible topic name length left for the user which has a hard total limit of 256).

ivanpauno:
I will merge this one. The topic name change can be done in a follow-up PR.

> @dirk-thomas that was my first idea, however, there still may be collisions when running the nodes in different machines.

And I think collisions are also possible when more than one context is used in the same process.

PR #754, opened 2019-06-07
Implement get_actual_qos() for subscriptions

mm318:
Currently, publishers have the get_actual_qos() feature/function. It would make sense for subscriptions to have them, too.


ivanpauno:
CI: https://github.com/ros2/rmw/pull/177#issuecomment-501261185

PR #757, opened 2019-06-10
Add rclcpp::create_timer()

sloretz:
This is a convenience PR that I'm hoping to get in the first Dashing patch release. It adds `rclcpp::create_timer()` that contains the boilerplate for creating a timer with any clock, [not just wall time](https://github.com/ros2/rclcpp/blob/4532d9cf784a85c899161a9b82e7d65814269f92/rclcpp/include/rclcpp/node.hpp#L335).

### DIFF
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -19,19 +19,21 @@
 #include <string>
 #include <utility>
 
+#include "rclcpp/duration.hpp"
+# include "rclcpp/node_interfaces/get_node_base_interface.hpp"
+# include "rclcpp/node_interfaces/get_node_timers_interface.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/node_interfaces/node_timers_interface.hpp"
-#include "rclcpp/duration.hpp"
-
 
 namespace rclcpp
 {
-/// Create a timer with a given type.
+/// Create a timer with a given clock
+/// \internal
 template<typename CallbackT>
 typename rclcpp::TimerBase::SharedPtr
 create_timer(
-  std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
-  std::shared_ptr<node_interfaces::NodeTimersInterface> node_timers,
+  node_interfaces::NodeBaseInterface * node_base,
+  node_interfaces::NodeTimersInterface * node_timers,
   rclcpp::Clock::SharedPtr clock,
   rclcpp::Duration period,
   CallbackT && callback,
@@ -45,6 +47,27 @@ create_timer(
 
   node_timers->add_timer(timer, group);
   return timer;
+}
+
+/// Create a timer with a given clock
+template<typename NodeT, typename CallbackT>
+typename rclcpp::TimerBase::SharedPtr
+create_timer(
+  NodeT node,
+  rclcpp::Clock::SharedPtr clock,
+  rclcpp::Duration period,
+  CallbackT && callback,
+  rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr)
+{
+  return create_timer(
+    rclcpp::node_interfaces::get_node_base_interface(node),
+    rclcpp::node_interfaces::get_node_timers_interface(node),
+    clock,
+    period,
+    // *INDENT-OFF*
+    std::forward<CallbackT &&>(callback),
+    // *INDENT-ON*
+    group);
 }
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -42,7 +42,7 @@ create_timer(
   auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
     clock,
     period.to_chrono<std::chrono::nanoseconds>(),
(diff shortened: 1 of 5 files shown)


sloretz:
>  the justification for it isn't clear to me

The only API I found for creating a timer is [`rclcpp::Node::create_wall_timer()`](https://github.com/ros2/rclcpp/blob/4532d9cf784a85c899161a9b82e7d65814269f92/rclcpp/include/rclcpp/node.hpp#L335) which only creates wall clock timers and is unavailable to lifecycle nodes.  Without this PR someone could still create the timer class and call add_timer() themselves, but the risk of adding a convenience method is low since it doesn't change any existing APIs.

> This doesn't look like one of our end-user APIs

I think it would look like [rclcpp::create_publisher()](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/create_publisher.hpp#L73) or [rclcpp::create_subscription()](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/create_subscription.hpp) if there was an overload that took a `Node` object. Would you like me to add that?

> (and doesn't seem to be exposed via the rclcpp facade)

What does this mean?

clalancette:
> 
> I think it would look like [rclcpp::create_publisher()](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/create_publisher.hpp#L73) or [rclcpp::create_subscription()](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/create_subscription.hpp) if there was an overload that took a `Node` object. Would you like me to add that?

Yeah, I think that would be better.  Well, what I think would make it even more like `create_publisher` is to have it be a method on the `rclcpp::Node` class (rather than take a node as a parameter).  Does that fit what you are trying to do?

> > (and doesn't seem to be exposed via the rclcpp facade)
> 
> What does this mean?

More-or-less that a "simple" version is exposed via the `Node` class, and the complex details are hidden away in an implementation somewhere (https://en.wikipedia.org/wiki/Facade_pattern).

sloretz:
> what I think would make it even more like create_publisher is to have it be a method on the rclcpp::Node class (rather than take a node as a parameter). Does that fit what you are trying to do?

I thought `rclcpp` was moving away from methods on `rclcpp::Node` to avoid duplication on lifecycle nodes. @wjwwood is that accurate?

wjwwood:
I have two asks for this pr:

- The `rclcpp::create_timer` should take a templated `NodeT` from which the node base and node timer interfaces are automatically extracted.
  - That way both `rclcpp::create_timer(node)` and `rclcpp::create_timer(lifecycle_node)` work without explicitly calling `get_node_*_interface()` manually.
  - We would probably need to copy the logic for node base/node timers interfaces that we have for getting node topics from an arbitrary class: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp
- Keep `rclcpp::create_timer` even if we add a `Node::create_timer` and `LifecycleNode::create_timer`, and keep as much logic in the free function as possible.

> I thought rclcpp was moving away from methods on rclcpp::Node to avoid duplication on lifecycle nodes. @wjwwood is that accurate?

I would like to do that, but until we haven't made a decision to do that as a group. If/when we do then we'd need to deprecate the methods on the `Node` and `LifecycleNode` and recommend the free functions that operate on a "Node-like" instead.

sloretz:
> The rclcpp::create_timer should take a templated NodeT from which the node base and node timer interfaces are automatically extracted. 

Done in ff19fd8

> We would probably need to copy the logic for node base/node timers interfaces that we have for getting node topics from an arbitrary class: 

Added `get_node_base_interface.hpp` and `get_node_timers_interface.hpp`. They are pure copies of `get_node_topics_interface.hpp` with the following substitutions:

* `TOPICS` -> `BASE` | `TIMERS`
* `Topics` -> `Base` | `Timers`
* `topics` -> `base` | `timers`

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -0,0 +1,75 @@
...
+  CallbackT && callback,
+  rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr)
+{
+  return create_timer(
+    rclcpp::node_interfaces::get_node_base_interface(node),
+    rclcpp::node_interfaces::get_node_timers_interface(node),
+    clock,
+    period,
+    // *INDENT-OFF*
+    std::forward<CallbackT &&>(callback),
```
wjwwood: You should be able to just do `std::forward<CallbackT>`, see: https://en.cppreference.com/w/cpp/utility/forward

sloretz: `std::forward<CallbackT>` in 02e266f2dd377734d97512d579892ffed44fce61

sloretz:
> It would be nice to have a test which uses a "different" class from rclcpp::Node, e.g. that uses this class:

Added this test in c511120

sloretz:
CI (testing only rclcpp)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7338/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3503/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6015/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7159/)

Issue #761, opened 2019-06-12
fail to have const member function as callback


Karsten1987:
When trying to set the callback to a `const` class member function, the function traits are complaining and no suitable overload is found.

To reproduce:

Use the example here: https://github.com/ros2/examples/blob/master/rclcpp/minimal_subscriber/member_function.cpp
and declare the function `topic_callback` to `const`.
This produces the following traceback:

```
In file included from /Users/karsten/workspace/osrf/ros2_full/src/ros2/examples/rclcpp/minimal_subscriber/member_function.cpp:17:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/rclcpp.hpp:144:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/executors.hpp:21:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/executors/multi_threaded_executor.hpp:24:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/executor.hpp:32:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/memory_strategies.hpp:18:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/memory_strategy.hpp:24:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/any_executable.hpp:20:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/callback_group.hpp:23:
In file included from /Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/client.hpp:31:
/Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/function_traits.hpp:52:40: error: reference to overloaded function could not be resolved; did you mean to call it?
    typename function_traits<decltype( &FunctionT::operator())>::arguments>::type;
                                       ^~~~~~~~~~~~~~~~~~~~~~
/Users/karsten/workspace/osrf/ros2_full/install/include/rclcpp/subscription_traits.hpp:91:39: note: in instantiation of template class 'rclcpp::function_traits::function_traits<std::__1::__bind<void (MinimalSubscriber::*)(std::__1::shared_ptr<std_msgs::msg::String_<std::__1::allocator<void> > >) const, MinimalSubscriber *, const std::__1::placeholders::__ph<1> &> >' requested here
[...]

PR #763, opened 2019-06-14
Added support for const member functions

esteve:
Addresses #761

### DIFF
--- a/rclcpp/include/rclcpp/function_traits.hpp
+++ b/rclcpp/include/rclcpp/function_traits.hpp
@@ -96,6 +96,24 @@ struct function_traits<
   : function_traits<ReturnTypeT(Args ...)>
 {};
 
+// std::bind for object const methods
+template<typename ClassT, typename ReturnTypeT, typename ... Args, typename ... FArgs>
+#if defined _LIBCPP_VERSION  // libc++ (Clang)
+struct function_traits<std::__bind<ReturnTypeT (ClassT::*)(Args ...), FArgs ...>>
+#elif defined _GLIBCXX_RELEASE  // glibc++ (GNU C++ >= 7.1)
+struct function_traits<std::_Bind<ReturnTypeT(ClassT::*(FArgs ...))(Args ...) const>>
+#elif defined __GLIBCXX__  // glibc++ (GNU C++)
+struct function_traits<std::_Bind<std::_Mem_fn<ReturnTypeT (ClassT::*)(Args ...) const>(FArgs ...)>>
+#elif defined _MSC_VER  // MS Visual Studio
+struct function_traits<
+  std::_Binder<std::_Unforced, ReturnTypeT(__cdecl ClassT::*)(Args ...), FArgs ...>
+>
+#else
+#error "Unsupported C++ compiler / standard library"
+#endif
+  : function_traits<ReturnTypeT(Args ...)>
+{};
+
 // std::bind for free functions
 template<typename ReturnTypeT, typename ... Args, typename ... FArgs>
 #if defined _LIBCPP_VERSION  // libc++ (Clang)
--- a/rclcpp/include/rclcpp/function_traits.hpp
+++ b/rclcpp/include/rclcpp/function_traits.hpp
@@ -99,7 +99,7 @@ struct function_traits<
 // std::bind for object const methods
 template<typename ClassT, typename ReturnTypeT, typename ... Args, typename ... FArgs>
 #if defined _LIBCPP_VERSION  // libc++ (Clang)
-struct function_traits<std::__bind<ReturnTypeT (ClassT::*)(Args ...), FArgs ...>>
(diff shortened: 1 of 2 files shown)


esteve:
@Karsten1987 I added a test for a const member, which I've been able to test on Linux with GCC >= 7.1 and lower, but I don't have a macOS or a Windows machine right now, could you trigger a CI build to see the output of the error?

Moreover, the tests pass, but when I added a const qualifier to the minimal_subscriber example, I ran into issues with subscription_traits, which I think are caused by function_traits, but I can't figure out why it fails.

rclcpp/include/rclcpp/function_traits.hpp
```diff
@@ -96,6 +96,24 @@ struct function_traits<
   : function_traits<ReturnTypeT(Args ...)>
 {};
 
+// std::bind for object const methods
+template<typename ClassT, typename ReturnTypeT, typename ... Args, typename ... FArgs>
+#if defined _LIBCPP_VERSION  // libc++ (Clang)
+struct function_traits<std::__bind<ReturnTypeT (ClassT::*)(Args ...), FArgs ...>>
+#elif defined _GLIBCXX_RELEASE  // glibc++ (GNU C++ >= 7.1)
+struct function_traits<std::_Bind<ReturnTypeT(ClassT::*(FArgs ...))(Args ...) const>>
```
allenh1: @esteve Nothing questioning any of this, but could you explain to me what exactly this is? I can tell it's the result of a `std::bind` (which makes this compiler dependent), but I'm not following the two variadic templates here at all. If you have time to explain, I'd love to learn why this patch is the fix.

esteve: Sure, no problem. Let's say we have this struct:

```c++
struct A {
    int member_const(float x, double y) const {
        (void)x;
        (void)y;
        return 1;
    }

    int member_non_const(float x, double y) {
        (void)x;
        (void)y;
        return 2;
    }
};
```

They are the same, but one of them is declared as `const`. So because of the slightly different signature, the template specializations we have for `std::bind` don't match. Its real signature for glibc++ and GCC >= 7.1 is the following:

```c++
std::_Bind<int (A::*(A*, std::_Placeholder<1>, std::_Placeholder<2>))(float, double) const>
```

For which we don't have any match in `function_traits`, the fix consists in duplicating the signatures we have for object members and just add a `const` where needed. However, I don't have a macOS or Windows machine at hand, so I haven't updated the signatures for them. To fix it, we just need to run the `function_traits` testsuite, and update the specializations to match the signatures.

allenh1: @esteve Ah, that makes a lot of sense. Thank you very much!

allenh1: @esteve I ran it on my windows machine for you: https://gist.github.com/allenh1/0b879d6a28c831615c68f307934a5e0f

Not building at the moment, it seems. But this should have all you need in it! Let me know if you need anything else.

esteve: @allenh1 awesome, thanks!

esteve: @allenh1 I justed updated the specialization with the info you pasted, could you try it again? Thanks.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7370/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3528/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6045/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7197/)

rclcpp/include/rclcpp/function_traits.hpp
allenh1: @esteve I _think_ it's fixed. I have a failure, but my build is not up to date yet (and it is unrelated to function_traits). I'll build my workspace and let you know if it fails.

allenh1: @esteve Definitely working on my machine

esteve:
Unfortunately the macOS build does not show the template signature error, so I can't figure out what the match should be. Anyway, I've pushed a new revision, which I believe should fix it for macOS, but it's purely a guess.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7371/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3529/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6046/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7198/)

Karsten1987:
@esteve thanks for coming up with another piece to the magic. Makes it work on my OSX (mojave), tested with having the callback in the rclcpp example marked as `const`.

Issue #519, opened 2018-07-19
executors should be able to operate on callback groups rather than nodes


wjwwood:
Right now, you can only add whole nodes to an executor, but it's come up in a few scenarios where you'd rather have the ability to add selective callback groups to an executor rather than the whole node. This would allow you to do things like have a separate executor for one subscription and the main executor can handle everything else in the node. This is also more similar to how callback queues worked in ROS 1.

Some things to consider:

- currently the executor assumes it doesn't share nodes with other executors
- the guard condition that interrupts an executor when new things are added (e.g. new timer) is owned by the node, but now it would need to be owned by the callback group (potentially increasing the total number of guard conditions)
- callback groups should still only be associated with a single executor, and that logic is currently in nodes (so it will need to be moved)

ralph-lange:
I would be happy to make a PR on the callback-group-level Executor presented at ROSCon 2018. Recently, I updated the implementation to Crystal, cf. https://github.com/boschresearch/ros2_rclcpp/commits/cbg-executor-0.6.1 
However, this is a prototypical implementation only, i.e. this PR can be only the basis for a discussion of the necessary refactoring. And, in the last weeks, I came to the conclusion that as a first step it is even more important to improve the current implementation regarding predictability rather than to refine the Executor API. The behavior of the Executor deviates substantially from the ROS 1 queueing mechanism, which is basically FIFO (with some exceptions in case of buffer overflows). In the ROS 2 Executor class, the get_next_ready_executable function (https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executor.cpp#L534) prioritizes timers over subscriptions over service invocations and so on. Inside the collect_entities function of the memory strategy (https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp#L153) the order seems to depend on the insertion order of the nodes rather than the arrival order on the middleware layer. A colleague of mine recently developed a formal scheduling model for the current mechanism. I'll include him in the discussion in the next days.
In my opinion, once we have a clear understanding of the current mechanism (which goes down to the middleware), we should first develop a design for the intended model(s) before we discuss about an improved Executor API and granularity.

peterpena:
@ralph-lange I would like to work on your callback-group-level Executor in `cbg-executor-foxy` branch so that it can merge with upstream if that is fine with you.

ralph-lange:
That's of course fine for me - and I'm happy to support you.

[...]

PR #764, opened 2019-06-19
Add rclcpp::spin() version with executor reference

bpwilcox:
To deal with thread cleanup issues upon destruction in https://github.com/ros2/geometry2/pull/114, for example, we needed to stop spinning the node in the thread before it could be joined. Having realized this change using `spin_some` led to high cpu usage, the pattern in https://github.com/ros-planning/navigation2/pull/866 and https://github.com/ros2/geometry2/pull/119 was proposed. 

To consolidate this pattern in `rclcpp`, this PR adds an interface to enable the user to pass in a reference to an Executor to spin the node. This change would allow for the user to manage the executor in order to do things such as call `executor.cancel()` to end the spin before joining a thread.

Signed-off-by: bpwilcox <bpwilcox@eng.ucsd.edu>

### DIFF
--- a/rclcpp/include/rclcpp/executors.hpp
+++ b/rclcpp/include/rclcpp/executors.hpp
@@ -47,6 +47,23 @@ RCLCPP_PUBLIC
 void
 spin(rclcpp::Node::SharedPtr node_ptr);
 
+/// Pass in an executor and spin the specified node.
+/**
+ * \param[in] node_ptr Shared pointer to the node to spin. 
+ * \param[in] executor The executor which will spin the node. 
+ */
+RCLCPP_PUBLIC
+void
+spin(
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
+  rclcpp::executor::Executor & executor);
+
+RCLCPP_PUBLIC
+void
+spin(
+  rclcpp::Node::SharedPtr node_ptr,
+  rclcpp::executor::Executor & executor);
+
 namespace executors
 {
 
--- a/rclcpp/src/rclcpp/executors.cpp
+++ b/rclcpp/src/rclcpp/executors.cpp
@@ -41,3 +41,21 @@ rclcpp::spin(rclcpp::Node::SharedPtr node_ptr)
 {
   rclcpp::spin(node_ptr->get_node_base_interface());
 }
+
+void
+rclcpp::spin(
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
+  rclcpp::executor::Executor & executor)
+{
+  executor.add_node(node_ptr);
+  executor.spin();
+  executor.remove_node(node_ptr);
+}
+
+void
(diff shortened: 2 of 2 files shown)


rotu:
This seems problematic.
1. It seems like this should maybe be an instance method on the executor, not rclcpp. A node can only belong to one executor at a time or else it raises an error.
2. It's misleading. If specifying an executor, that executor is going to spin all nodes attached to it, not just the one you pass.

bpwilcox:
> This seems problematic.
> 
>     1. It seems like this should maybe be an instance method on the executor, not rclcpp. A node can only belong to one executor at a time or else it raises an error.
> 
This would be the case with spinning a node on an executor in any case, even if using the default `rclcpp::spin`, so this interface wouldn't change that expected behavior. 

>     2. It's misleading. If specifying an executor, that executor is going to spin all nodes attached to it, not just the one you pass.

Yes, you're right. If you were intentionally adding multiple nodes to an executor, it could be fine, but it may be undefined behavior to add and spin a node when the executor is already spinning another.

Karsten1987:
I think I am generally in favor of this patch, but I'd have to think about some corner cases in the behavior first.

As @bpwilcox mentioned, you can achieve the same behavior with the current executor or `rclcpp::spin()`. I am not so much concerned whether this should be an instance of the executor. 
@rotu , your second point is a bit unclear to me. The same consequences occur in the current solution in https://github.com/ros2/geometry2/pull/119 isn't it? As the patch is, the executor only contains one node, because it's scoped, but in the end it's up to the user to make sure that no other nodes are spinning (or have to spin) in order to safely cancel the execution, right?

rotu:
@Karsten1987, Yes. If you create an executor and are completely aware of its state, it's fine to add nodes, spin it, and remove nodes. What this new function does is "adds a node to an executor, spins the executor, and removes the node from the executor". It's no longer appropriate to think of the function as "spin the passed node and only the passed node", which is the current meaning of `rclcpp::spin`.

rotu:
I think I'd be *much* more comfortable with this if it threw a warning or an error when the executor already has nodes associated with it.

Karsten1987:
Given that argumentation, I think you are right that this addition as-is brings more confusion than real benefit. So thanks for that :)

I think ideally, the executor shouldn't really deal with nodes in the first place, but rather only with [callback groups coming from nodes](https://github.com/ros2/rclcpp/issues/519). But that's only slightly related to this PR.

So I would recommend to close this PR unmerged and stick to the situation implemented in the referenced geometry2 PR where it's clear that the executor has only one node attached.

rotu:
What is it about constructing and executor that’s so doggone expensive anyway?

bpwilcox:
I agree with the concerns/feedback on passing an executor with unknown state to spin, so I am in favor to close this PR. However, I would be interested to see an interface to encapsulate something like the solution in the geometry2 PR as I recognize it as a common pattern for spinning nodes in a thread.

Issue #750, opened 2019-06-03
Add an option for ignoring parameter overrides


ivanpauno:
#### Feature description
Follow up of #730 (see https://github.com/ros2/rclcpp/issues/730#issuecomment-496676163). Currently, the user don't have a way of ignoring a parameter override.

My suggestion is to change:

```cpp
  RCLCPP_PUBLIC
  const rclcpp::ParameterValue &
  declare_parameter(
    const std::string & name,
    const rclcpp::ParameterValue & default_value = rclcpp::ParameterValue(),
    const rcl_interfaces::msg::ParameterDescriptor & parameter_descriptor =
    rcl_interfaces::msg::ParameterDescriptor());
```
to
```cpp
  RCLCPP_PUBLIC
  const rclcpp::ParameterValue &
  declare_parameter(
    const std::string & name,
    const rclcpp::ParameterValue & default_value = rclcpp::ParameterValue(),
    const rcl_interfaces::msg::ParameterDescriptor & parameter_descriptor =
    rcl_interfaces::msg::ParameterDescriptor()
    bool ignore_override = false,
  );
```
and maybe, also add:
```cpp
  RCLCPP_PUBLIC
  const rclcpp::ParameterValue &
  declare_parameter(
    const std::string & name,
    const rclcpp::ParameterValue & default_value = rclcpp::ParameterValue(),
    bool ignore_override = false,
  );
```

When undeclared parameters are allowed and `automatically_declare_parameter_overrides` is `false`, the first `set_parameter` call does not actually set the parameter but it takes the override value (see the second example [here](https://github.com/ros2/rclcpp/issues/730#issuecomment-495801806)).

I don't see much usage of that feature, and I think it's a little confusing. I suggest always ignoring the parameter override in that case. That last change would be an api break, but of something with not much use cases. I don't see much value on adding a `ignore_override` argument to `set_parameter` method.

Note: Maybe, we could also add a node option for ignoring all the overrides. But I don't see much usage of that.

#### Implementation considerations
Should be fairly easy to implement. An equivalent would be needed in rclpy.

ivanpauno:
@chapulina @wjwwood any opinions?

chapulina:
Having a way to ignore parameter overrides sounds helpful to me.

As for the implementation, have you considered passing `ignore_overrides` as part of [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/msg/ParameterDescriptor.msg)?

[...]

PR #767, opened 2019-06-21
Add ignore override argument to declare parameter

ivanpauno:
Closes https://github.com/ros2/rclcpp/issues/750.

I also thought about adding:
```cpp
  template<typename ParameterT>
  std::vector<ParameterT>
  declare_parameters(
    const std::map<
      std::string,
      std::tuple<ParameterT, rcl_interfaces::msg::ParameterDescriptor, bool>
    > & parameters);
```

Where the last element of the tuple is a bool indicating if the override of that parameter is ignored or not.

But this:
https://github.com/ros2/rclcpp/blob/f4a979df0bb7162fcde8b6b50fead9a2c020dcf7/rclcpp/test/test_node.cpp#L510-L513
ended up with an ambiguous overload error:
```
call of overloaded ‘declare_parameters(const char [1], <brace-enclosed initializer list>)’ is ambiguous
```

I didn't find a workaround, so I simple deleted that option.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -369,6 +369,8 @@ public:
    * are ignored, and should be specified using the name argument to this
    * function and the default value's type instead.
    *
+   * If `ignore_override` is `true`, the parameter override will be ignored.
+   *
    * This method, if successful, will result in any callback registered with
    * set_on_parameters_set_callback to be called.
    * If that callback prevents the initial value for the parameter from being
@@ -382,6 +384,8 @@ public:
    *   did not override it.
    * \param[in] parameter_descriptor An optional, custom description for
    *   the parameter.
+   * \param[in] ignore_override When `true`, the parameter override is ignored.
+   *    Default to `false`.
    * \return A const reference to the value of the parameter.
    * \throws rclcpp::exceptions::ParameterAlreadyDeclaredException if parameter
    *   has already been declared.
@@ -444,8 +448,7 @@ public:
    *
    * The map contains default values for parameters.
    * There is another overload which takes the std::pair with the default value
-   * and descriptor, and another which takes an std::tuple with the same two
-   * parameters and a bool indicating if overriding the parameter default is allowed.
+   * and descriptor.
    *
    * If `ignore_overrides` is `true`, all the overrides of the parameters declared
    * by the function call will be ignored.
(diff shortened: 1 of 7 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -396,7 +396,8 @@ class Node : public std::enable_shared_from_this<Node>
     const std::string & name,
     const rclcpp::ParameterValue & default_value = rclcpp::ParameterValue(),
     const rcl_interfaces::msg::ParameterDescriptor & parameter_descriptor =
-    rcl_interfaces::msg::ParameterDescriptor());
+    rcl_interfaces::msg::ParameterDescriptor(),
+    bool ignore_override = false);
```
wjwwood: Can you document it above please?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -440,11 +442,13 @@ class Node : public std::enable_shared_from_this<Node>
...
    *
-   * The map may either contain default values for parameters, or a std::pair
-   * where the first element is a default value and the second is a
-   * parameter descriptor.
-   * This function only takes the default value, but there is another overload
-   * which takes the std::pair with the default value and descriptor.
+   * The map contains default values for parameters.
+   * There is another overload which takes the std::pair with the default value
+   * and descriptor, and another which takes an std::tuple with the same two
+   * parameters and a bool indicating if overriding the parameter default is allowed.
```
wjwwood: This seems out of date now?

rclcpp/include/rclcpp/node.hpp
wjwwood: There are a few cases where these are not properly documented.

rclcpp/test/test_node.cpp
```diff
@@ -403,12 +404,23 @@ TEST_F(TestNode, declare_parameter_with_initial_values) {
...
+  {
+    // int default, with override and ignoring it
+    rclcpp::ParameterValue default_value(43);
+    rclcpp::ParameterValue value = node->declare_parameter(
+      "parameter_and_default_ignore_override",
+      default_value,
+      rcl_interfaces::msg::ParameterDescriptor(),
+      true);
+    EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_INTEGER);
+    EXPECT_EQ(value.get<int>(), 43);  // and not 43 which is the default value
```
wjwwood: "// and not 42 which is the override value"?

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7415/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3568/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6084/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7240/)

Issue #722, opened 2019-05-15
address how parameter callbacks work


wjwwood:
There was a long discussion here: https://github.com/ros2/rclcpp/pull/495#discussion_r275565749

Basically right now you can only have one parameter callback at a time that can control whether or not parameters may be set.

Note this is about `set_on_parameters_set_callback`, but some parts of this discuss will apply to the "on change" callback which happens after setting occurs and is just for observation (it cannot reject changes).

This callback is considered insufficient because it makes having one callback per parameter or per group of parameters very difficult to implement. At first it seems trivial as the user could just have a single callback which calls multiple callbacks, but if the system also wants to have callbacks for things like implicitly declared parameters (e.g. `use_sim_time`) or callbacks to check changes to parameters based on some state (maybe lifecycle node's) then it's easy for the user to "clobber" system callbacks or even for system callbacks to interfere with one another.

One way to solve this is to have the NodeParameters interface handle this itself, allowing you to set more than one callback at a time. That would remove the burden from the users to coordinate and not interfere with each other.

However, if we do that we need a way to unregister callbacks that were registered (so some sort of receipt is needed), and we need a policy for what order to call them in and what to do if callbacks which overlap (check the same parameter) conflict. Do we stop at the first that fails or continue and aggregate the failure reasons?

----

This is my personal opinion on what we should do:

I think it should be a list of callbacks stored in the node, which is called in the order that they are added (perhaps with an option to prepend to the callback list for special cases), and we stop calling callbacks when the first one returns `SetParameterResults.successful = false`.

I'm not sure if it's worth calling all the callbacks so that we can take all the ones that failed and join their reasons, e.g. it might result in a reason like "failed to set parameters because: the node is not in the unconfigured state, and the parameter is not in the range x to y, and the parameter is read only" versus just one of them (the first one).

Issue #769, opened 2019-06-24
Exception when declaring parameters after callback is set


Karsten1987:
I am struggling with how to declare parameters for my `diagnostic_updater`. Specifically, I have the following setup: I am spawning the `URG` laser driver, which _is_ a node and _has_ a diagnostics updater instance as a class member variable. That diagnostics updater takes a node (or better node interfaces) in its constructor. The diagnostics updater has a parameter `period` which is meant to be updated dynamically. The URG node itself has multiple laser specific parameters declared as well.

Within code, this kind of looks like this:

```c++
Urg::Urg()
: rclcpp::Node("urg_laser_node")
{
  // Get parameters so we can change these later.
  this->declare_parameter<std::string>("ip_address", ip_address_);
  this->declare_parameter<int>("ip_port", ip_port_);
  this->declare_parameter<std::string>("laser_frame_id", laser_frame_id_);
  ...

  this->set_on_parameters_set_callback(std::bind(&UrgNode::param_change_callback, this, std::placeholders::_1));

  // the diagnostics_updater declares another parameter and results in an exception.
  diagnostic_updater_ = std::make_unique<diagnostic_updater::Updater>(
    this->get_node_base_interface(),
    this->get_node_topics_interface(),
    this->get_node_logging_interface(),
    this->get_node_parameters_interface());
  diagnostic_updater_->add("Hardware Status", this, &UrgNode::populateDiagnosticsStatus);
 
  // Having the parameter callback set at the very end, prevents the exception
  // this->set_on_parameters_set_callback(std::bind(&UrgNode::param_change_callback, this, std::placeholders::_1));
}
```

I am noticing two problems with this:

1.) When having `set_on_parameters_set_callback` called before initializing the diagnostics updater instance, I am unable to declare any parameter within the diagnostics instance as the assigned callback for the URG node is not aware of the diagnostics parameter and I get an exception:
```
libc++abi.dylib: terminating with uncaught exception of type rclcpp::exceptions::InvalidParameterValueException: parameter 'urg_node.diagnostics_updater_period' could not be set: The parameter urg_node.diagnostics_updater_period is not part of the reconfigurable parameters.
```

So I had to make sure that the parameter callback is set at last and no new parameters were declared after that.

[...]

Issue #609, opened 2018-12-11
Provide for local callbacks on parameter events in parameter node API


tfoote:
This is a follow up to discussions in #595 and #608 and somewhat related to #296 but for the post change callbacks not validation stage which would be a 3rd callback type proposed.

It would be valuable to not rely on parameter events being published but use the internal node interface API to get the events in a callback for different interfaces. Along with that change if there was a more fully featured API for registering callbacks for parameter changes locally, the parameter event publishing could also be split out into a separate component that could obviate the need for a parameter to turn that functionality on and off in the constructor of the parameter, and avoid the dependencies on the clock and topics interface for the core parameter API.

0-admin:
Wrere ti puch? kjk?(

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/security-ramifications-of-global-parameter-events/15924/8

clalancette:
This was actually completed for Iron, so closing this out.

jrutgeer:
@clalancette 
It seems this issue is not fully complete, see [this TODO item in the code](https://github.com/ros2/rclcpp/blob/411dbe8212dd753ee60871f8605d0a8beaae5e37/rclcpp/src/rclcpp/time_source.cpp#L280).

I came across this due to someone asking about disabling `/parameter_events` subscribers [on Robotics Stack Exchange](https://robotics.stackexchange.com/questions/105382/how-to-stop-the-subscription-to-parameter-events-that-is-automatically-create).

So I think this should be reopened.

clalancette:
> It seems this issue is not fully complete, see [this TODO item in the code](https://github.com/ros2/rclcpp/blob/411dbe8212dd753ee60871f8605d0a8beaae5e37/rclcpp/src/rclcpp/time_source.cpp#L280).

This issue is indeed complete; we now provide a local callback that can be used.  If you'd like to track changing the code in `time_source.cpp` to use that, please feel free to open a new issue.

fujitatomoya:
@jrutgeer to be honest, i do not remember... probably i was working on the experimental patch? bad memory, sorry 😓 

created tracking issue for this, https://github.com/ros2/rclcpp/issues/2370

PR #772, opened 2019-06-25
Allow registering multiple on_parameters_set_callback

ivanpauno:
(Edit)

Fixes https://github.com/ros2/rclcpp/issues/722.

The API looks like this:

```cpp
  add_on_set_parameters_callback(OnParametersSetCallbackType callback);
```
```cpp
  void
  remove_on_set_parameters_callback(const OnSetParametersCallbackHandle * const handler);
```

----
(Original comment)

Based in discussion in https://github.com/ros2/rclcpp/issues/769 (TL;DR see [this comment](https://github.com/ros2/rclcpp/issues/769#issuecomment-505107740)), it would be nice to allow registering multiple callbacks with `set_on_parameters_set_callback`.

My idea was the following:

- Calling `set_on_parameters_set_callback` adds a new callback to a vector. It returns a lambda which emulates the behavior of calling all the previous registered callbacks (before, it was returning the previous callback and overriding it). If no callback was registered before, it returns `nullptr`.
- Calling `set_on_parameters_set_callback` with `nullptr` as argument clear all the registered callbacks.

Maybe it would be better to return the vector of all the callbacks instead of a lambda, but that would add a new signature (probably deprecating the old method).

Maybe, we want to implement this in a different way, or not implement it at all.
This is just a proposal of a possible implementation.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -942,11 +942,12 @@ public:
    * returned unique pointer is valid.
    * The returned unique pointer can be promoted to a shared version.
    *
-   * Releasing the returned unique pointer unregisters the callback.
+   * Resetting or letting the smart pointer go out of scope unregisters the callback.
    * `remove_on_set_parameters_callback` can also be used.
    *
-   * All the registered callbacks are called when a parameter is set.
-   * The order of the callbacks is not important.
+   * The registered callbacks are called when a parameter is set.
+   * When a callback returns a not successful result, the remaining callbacks aren't called.
+   * The order of the callback execution could be different from the registration order.
    *
    * \param callback The callback to register.
    * \returns A unique pointer. The callback is valid as long as the unique pointer is alive.
@@ -972,10 +973,8 @@ public:
    *
    * Calling `remove_on_set_parameters_callback` more than once with the same handler,
    * or calling it after the shared pointer has been released is an error.
-   * Releasing the smart pointer after calling `remove_on_set_parameters_callback` is not a problem.
-   *
-   * All the registered callbacks are called when a parameter is set.
-   * The order of the callbacks is not important.
+   * Resetting or letting the smart pointer go out of scope after calling
+   * `remove_on_set_parameters_callback` is not a problem.
    *
    * \param handler The callback handler to remove.
    * \throws std::runtime_error if the handler was not created with `add_on_set_parameters_callback`,
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -939,40 +939,40 @@ public:
    * rclcpp::exceptions::ParameterModifiedInCallbackException will be thrown.
    *
    * The callback functions must remain valid as long as the
-   * returned unique pointer is valid.
-   * The returned unique pointer can be promoted to a shared version.
+   * returned smart pointer is valid.
+   * The returned smart pointer can be promoted to a shared version.
    *
    * Resetting or letting the smart pointer go out of scope unregisters the callback.
    * `remove_on_set_parameters_callback` can also be used.
    *
    * The registered callbacks are called when a parameter is set.
    * When a callback returns a not successful result, the remaining callbacks aren't called.
-   * The order of the callback execution could be different from the registration order.
+   * The order of the callback is the reverse from the registration order.
    *
    * \param callback The callback to register.
-   * \returns A unique pointer. The callback is valid as long as the unique pointer is alive.
+   * \returns A shared pointer. The callback is valid as long as the smart pointer is alive.
    * \throws std::bad_alloc if the allocation of the OnSetParametersCallbackHandle fails.
    */
   RCLCPP_PUBLIC
-  OnSetParametersCallbackHandle::UniquePtr
+  OnSetParametersCallbackHandle::SharedPtr
   add_on_set_parameters_callback(OnParametersSetCallbackType callback);
 
   /// Remove a callback registered with `add_on_set_parameters_callback`.
   /**
-   * Releases a handler, returned by `add_on_set_parameters_callback`.
+   * Delete a handler returned by `add_on_set_parameters_callback`.
    *
    * e.g.:
    *
    *    `remove_on_set_parameters_callback(scoped_callback.get())`
    *
-   * As an alternative, the unique pointer can be released:
+   * As an alternative, the smart pointer can be reset:
    *
    *    `scoped_callback.reset()`
    *
    * Supposing that `scoped_callback` was the only owner.
    *
    * Calling `remove_on_set_parameters_callback` more than once with the same handler,
-   * or calling it after the shared pointer has been released is an error.
+   * or calling it after the shared pointer has been reset is an error.
    * Resetting or letting the smart pointer go out of scope after calling
    * `remove_on_set_parameters_callback` is not a problem.
    *
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -981,6 +981,7 @@ public:
    * \throws std::runtime_error if the handler was not created with `add_on_set_parameters_callback`,
    *   or if it has been removed before.
    */
+  RCLCPP_PUBLIC
   void
   remove_on_set_parameters_callback(const OnSetParametersCallbackHandle * const handler);
 
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -889,10 +889,12 @@ public:
   rcl_interfaces::msg::ListParametersResult
   list_parameters(const std::vector<std::string> & prefixes, uint64_t depth) const;
 
+  using OnSetParametersCallbackHandle =
+    rclcpp::node_interfaces::OnSetParametersCallbackHandle;
   using OnParametersSetCallbackType =
     rclcpp::node_interfaces::NodeParametersInterface::OnParametersSetCallbackType;
 
-  /// Register a callback to be called anytime a parameter is about to be changed.
+  /// Add a callback for when parameters are being set.
   /**
    * The callback signature is designed to allow handling of any of the above
    * `set_parameter*` or `declare_parameter*` methods, and so it takes a const
@@ -936,9 +938,60 @@ public:
    * of the latter things,
    * rclcpp::exceptions::ParameterModifiedInCallbackException will be thrown.
    *
-   * There may only be one callback set at a time, so the previously set
-   * callback is returned when this method is used, or nullptr will be returned
-   * if no callback was previously set.
+   * The callback functions must remain valid as long as the
+   * returned unique pointer is valid.
+   * The returned unique pointer can be promoted to a shared version.
+   *
+   * Releasing the returned unique pointer unregisters the callback.
+   * `remove_on_set_parameters_callback` can also be used.
+   *
+   * All the registered callbacks are called when a parameter is set.
+   * The order of the callbacks is not important.
+   *
+   * \param callback The callback to register.
+   * \returns A unique pointer. The callback is valid as long as the unique pointer is alive.
+   * \throws std::bad_alloc if the allocation of the OnSetParametersCallbackHandle fails.
+   */
+  RCLCPP_PUBLIC
+  OnSetParametersCallbackHandle::UniquePtr
+  add_on_set_parameters_callback(OnParametersSetCallbackType callback);
+
+  /// Remove a callback registered with `add_on_set_parameters_callback`.
+  /**
+   * Releases a handler, returned by `add_on_set_parameters_callback`.
+   *
+   * e.g.:
+   *
+   *    `remove_on_set_parameters_callback(scoped_callback.get())`
+   *
+   * As an alternative, the unique pointer can be released:
+   *
+   *    `scoped_callback.reset()`
+   *
+   * Supposing that `scoped_callback` was the only owner.
+   *
+   * Calling `remove_on_set_parameters_callback` more than once with the same handler,
+   * or calling it after the shared pointer has been released is an error.
+   * Releasing the smart pointer after calling `remove_on_set_parameters_callback` is not a problem.
+   *
+   * All the registered callbacks are called when a parameter is set.
+   * The order of the callbacks is not important.
+   *
+   * \param handler The callback handler to remove.
+   * \throws std::runtime_error if the handler was not created with `add_on_set_parameters_callback`,
+   *   or if it has been removed before.
+   */
+  void
+  remove_on_set_parameters_callback(const OnSetParametersCallbackHandle * const handler);
+
+  /// Register a callback to be called anytime a parameter is about to be changed.
+  /**
+   * With this method, only one callback can be set at a time. The callback that was previously
+   * set by this method is returned or `nullptr` if no callback was previously set.
+   *
+   * The callbacks added with `add_on_set_parameters_callback` are stored in a different place.
+   * `remove_on_set_parameters_callback` can't be used with the callbacks registered with this
+   * method. For removing it, use `set_on_parameters_set_callback(nullptr)`.
    *
    * \param[in] callback The callback to be called when the value for a
    *   parameter is about to be set.
@@ -946,7 +999,7 @@ public:
    *   otherwise nullptr.
    */
   RCLCPP_PUBLIC
-  rclcpp::Node::OnParametersSetCallbackType
+  OnParametersSetCallbackType
   set_on_parameters_set_callback(rclcpp::Node::OnParametersSetCallbackType callback);
 
   /// Register the callback for parameter changes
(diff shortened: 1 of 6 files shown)


rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -526,7 +540,7 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
       rcl_interfaces::msg::ParameterDescriptor(),  // Implicit declare uses default descriptor.
       staged_parameter_changes,
       parameter_overrides_,
-      nullptr,  // callback is explicitly null, so that it is called only once, when setting below.
+      {},  // callback is explicitly null, so that it is called only once, when setting below.
```
wjwwood: nitpick: comment should probably be updated (no longer null).

clalancette:
> I think I would prefer a more complete solution which allowed callbacks to be registered and unregistered with random access (rather than having to clear them all with `nullptr` being the only way to remove one).
> 
> I don't know if this solution is worth doing in the meantime, since what I described above will definitely require API changes. Maybe others can comment.

I'd prefer to do the "correct" approach (for some value of correct), and then worrying about API changes later if we want to do a backport.

I agree with you that we should be able to add and remove callbacks at will without clearing out the whole list; it will make the whole feature much more robust.

ivanpauno:
> I think I would prefer a more complete solution which allowed callbacks to be registered and unregistered with random access (rather than having to clear them all with `nullptr` being the only way to remove one).

How will the callbacks be identified? just by an integer index?
Also, I don't have an use case of random access of callbacks in mind. Could you add an example?

> I'd prefer to do the "correct" approach (for some value of correct), and then worrying about API changes later if we want to do a backport.

The proposal as-is now, I think it's not subject of backporting as is certainly breaking api. Maybe a better idea is to maintain the old behavior exactly, and add a new api for adding callbacks, accessing and removing them randomly.
We should call the callback registered in the old way, and then all the callbacks registered in the new way. That will not break the api.

Could be something like:

```
/* Adds a callback, it returns an integer indicating the identifier */
std::size_t
add_on_parameters_set_callback(OnParametersSetCallbackType callback) override;
```
```
/* Delete the callback with the provided identifier. Returns true if the identifier was valid.*/
bool
delete_on_parameters_set_callback(size_t identifier) override;
```

```
/* Get a copy of the callback registered with the provided identifier. Returns true if the identifier was valid.*/
bool
get_on_parameters_set_callback(size_t identifier, OnParametersSetCallbackType & callback) override;
```

```
/* Delete all the stored callbacks.*/
void
clear_on_parameters_set_callback() override;
```

The behavior of the existing `set_on_parameters_set_callback` and `register_param_change_callback` don't change. The callback registered by these methods is not stored in the same place as the others, and can't be accessed with the new methods. Probably, the new `clear_on_parameters_set_callback` should also delete the "old" callback.

wjwwood:
> How will the callbacks be identified? just by an integer index?

I wouldn't use an index, as the index may change if you remove something from the list/vector. In other places we're using a "handle" object. It can be anything, maybe a pointer or some other kind of locally-unique identifier that is used as a key in a map. I think the "jump callbacks" use this pattern:

https://github.com/ros2/rclcpp/blob/0ccac1e3bd0f4314e11c86202c0c3ade57aed8ed/rclcpp/include/rclcpp/clock.hpp#L118

I even think the thing returned is a smart pointer and scoped, so another way to "unregister" the callback is to let the handle go out of scope.

> Also, I don't have an use case of random access of callbacks in mind. Could you add an example?

Normally, I would shy away from adding a feature just because someone might want it, but in this case I think that's appropriate, since I cannot imagine off the top of my head when this would be used, I know that if that case came up, then it might prevent them from using the feature all together. Also, since the code adding callbacks to a single node might come from various sources (e.g. the node itself adds the `use_sim_time` parameter callback and then user's add their own), removing all of the callbacks (as proposed in this pr) might have adverse side effects for one part or the other. Better to let them undo anything they've done without interfering with others unintentionally.

For the most part, I'd say we should do it just so the API has symmetry. I think we should avoid API's that let users create or add things without being able to remove them unless there's a specific reason for it beyond "it was easier not to have the remove interface".

> The proposal as-is now, I think it's not subject of backporting as is certainly breaking api. Maybe a better idea is to maintain the old behavior exactly, and add a new api for adding callbacks, accessing and removing them randomly.

That was my thoughts as well, add the new API that allows you to add multiple, have `set_parameter(s)` call both the single callback if set and all of the multi callback ones (populated from the new API) and then deprecate the API that sets only a single callback at a time. That way next version both will be valid ways to add a callback, but neither will interfere with the other.

> The behavior of the existing set_on_parameters_set_callback and register_param_change_callback don't change. The callback registered by these methods is not stored in the same place as the others, and can't be accessed with the new methods. Probably, the new clear_on_parameters_set_callback should also delete the "old" callback.

I didn't read this until I wrote my previous paragraph, but yes exactly this :)

Though I was imagining something more like the clock's jump callback's API:

```c++
struct OnSetParametersCallbackHandle {
  OnSetParametersCallback callback;
};

// stored in a std::vector<std::weak_ptr<OnSetParametersCallbackHandle>>

std::shared_ptr<OnSetParametersCallbackHandle>
add_on_set_parameters_callback(OnSetParametersCallback callback);

// same as handle.reset(); throws if not found
void
remove_on_set_parameters_callback(OnSetParametersCallbackHandle handle);
```

We can optionally still have `clear_on_parameters_set_callback()`, though I think it's dangerous because people will assume they are the only ones adding callbacks, and so clear should only remove "their" callbacks, but it would for instance also remove the one provided for `use_sim_time`. It would be easier to have scoped handles and the user could just let all of them go out of scope.

I would not have a `get_on_parameters_set_callback(handle/id)`, but maybe a `get_on_parameters_set_callbacks() -> const vector<weak_ptr<OnSetParametersCallbackHandle>> &` would be ok, though this is a case where it's sort of a feature looking for a problem in my opinion.

ivanpauno:
I needed to rebase with master, to solve merge conflicts.

CI (up to test_rclcpp rclcpp, only fastrtps):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7574/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3707/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6200/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7400/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -936,17 +938,69 @@ class Node : public std::enable_shared_from_this<Node>
...
+   *
+   * As an alternative, the unique pointer can be released:
+   *
+   *    `scoped_callback.reset()`
+   *
+   * Supposing that `scoped_callback` was the only owner.
+   *
+   * Calling `remove_on_set_parameters_callback` more than once with the same handler,
+   * or calling it after the shared pointer has been released is an error.
+   * Releasing the smart pointer after calling `remove_on_set_parameters_callback` is not a problem.
```
wjwwood: Releasing != reset, see:

https://en.cppreference.com/w/cpp/memory/unique_ptr/release vs https://en.cppreference.com/w/cpp/memory/unique_ptr/reset, release will not destruct it. We should use consistent terminology here and above.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -936,17 +938,69 @@ class Node : public std::enable_shared_from_this<Node>
...
-   * if no callback was previously set.
+   * The callback functions must remain valid as long as the
+   * returned unique pointer is valid.
+   * The returned unique pointer can be promoted to a shared version.
+   *
+   * Releasing the returned unique pointer unregisters the callback.
+   * `remove_on_set_parameters_callback` can also be used.
+   *
+   * All the registered callbacks are called when a parameter is set.
+   * The order of the callbacks is not important.
```
wjwwood: Is it not? What is the order called? What happens if one of them fails (returns a result with success set to false), do the others that have not been called still get called? If so, how do you combine the results if they disagree or more than one fails?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -936,17 +938,69 @@ class Node : public std::enable_shared_from_this<Node>
...
+   *    `scoped_callback.reset()`
+   *
+   * Supposing that `scoped_callback` was the only owner.
+   *
+   * Calling `remove_on_set_parameters_callback` more than once with the same handler,
+   * or calling it after the shared pointer has been released is an error.
+   * Releasing the smart pointer after calling `remove_on_set_parameters_callback` is not a problem.
+   *
+   * All the registered callbacks are called when a parameter is set.
+   * The order of the callbacks is not important.
```
wjwwood: This seems redundant since you mention it in the add callback.

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
```diff
@@ -75,7 +76,9 @@ class ParameterMutationRecursionGuard
 };
 
 /// Implementation of the NodeParameters part of the Node API.
-class NodeParameters : public NodeParametersInterface
+class NodeParameters
+  : public NodeParametersInterface,
+  public std::enable_shared_from_this<NodeParameters>
```
wjwwood: This is not an ideal change, is it possible you don't actually need shared ownership of this? Would it be better to return a shared_ptr of the callback, and keep a weak ptr in this class?

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
```diff
@@ -180,7 +193,9 @@ class NodeParameters : public NodeParametersInterface
   // declare_parameter, etc).  In those cases, this will be set to false.
   bool parameter_modification_enabled_{true};
 
-  OnParametersSetCallbackType on_parameters_set_callback_ = nullptr;
+  OnParametersSetCallbackType on_parameters_set_callback_;
```
wjwwood: Why did you remove the initialization here?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -588,6 +615,8 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
     // they are actually set on the official parameter storage
     parameters_,
     // this will get called once, with all the parameters to be set
+    on_parameters_set_callback_set_,
```
wjwwood: The comment for this one probably needs to be updated.

rclcpp/include/rclcpp/node.hpp
wjwwood: I'd say, instead, "Resetting or letting the smart pointer go out of scope...".

rclcpp/include/rclcpp/node.hpp
```diff
@@ -936,17 +938,69 @@ class Node : public std::enable_shared_from_this<Node>
...
-   * callback is returned when this method is used, or nullptr will be returned
-   * if no callback was previously set.
+   * The callback functions must remain valid as long as the
+   * returned unique pointer is valid.
+   * The returned unique pointer can be promoted to a shared version.
+   *
+   * Releasing the returned unique pointer unregisters the callback.
+   * `remove_on_set_parameters_callback` can also be used.
+   *
+   * All the registered callbacks are called when a parameter is set.
```
wjwwood: Looking at the implementation, I don't think this is true, they're called in some order, and if one returns success as false, then you return immediately without running the rest.

rclcpp/include/rclcpp/node.hpp
ivanpauno: Yes thanks, I sometimes mix those words.

rclcpp/include/rclcpp/node.hpp
ivanpauno: I've updated the comment, I hope it's clearer now.

When rewriting it, I realized about the following:
Users are using callbacks for updating configurations when they are returning a successful result (e.g.: https://github.com/clalancette/velodyne/blob/34af494b6a8caad6e97378026e353ca9cfdaa97f/velodyne_pointcloud/src/conversions/convert.cc#L186-L219). In case other callback was registered, it may be executed later and reject the result, and the prior callback would have updated the configuration.

I wrote this thinking that each callback will check a disjoint group of parameters, but I don't know if that assumption is reasonable or not. If it is reasonable, I would just document how it works. If not, maybe having pre and post callbacks is a good idea.

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
ivanpauno: I tried to avoid the possibility of UB:
https://github.com/ros2/rclcpp/blob/c7d01d7bf3a6f886577689877205ffb08ff27b42/rclcpp/src/rclcpp/clock.cpp#L128-L140

If not, I can't remove the pointer from the container in the deleter.
I have two possible work-arounds:
- Do what `clock` does, and suppose people will write the program in a way that the callback goes out of scope before the interface (so, catching `this` and not `shared_from_this` is ok). This usually happens, when inheriting from `rclcpp::Node`.
- Let the callback go out of scope, and do not delete it from the container in its deleter (use the default deleter). Store weak ptrs (and not raw ptrs), in the container. The callbacks are removed from the container if they are found to be `nullptr` when trying to call them (or, they aren't removed at all, just skipped).

ivanpauno: I think we are always creating a shared pointer of `NodeParameters`.
So inheriting from `std::enable_shared_from_this` shouldn't be a problem (I should run full ci for confirming it).

rclcpp/include/rclcpp/node.hpp
ivanpauno: Based on #769, I think it's a reasonable assumption.
But, there's maybe an use case where is not.

wjwwood: > I wrote this thinking that each callback will check a disjoint group of parameters, but I don't know if that assumption is reasonable or not.

I don't think that it is. The assumption of the callback must be that the values may still not be set due to another callback rejecting them. That's the only safe thing to do.

There is another set of callbacks "on change" which get called after parameters are added, updated, or removed. Currently it uses a topic and a subscription to generate these callbacks, but should be changed to use some other local mechanism (like directly calling a callback from within set parameter methods): https://github.com/ros2/rclcpp/issues/609

By the way, this pull request should probably somehow mention this issue https://github.com/ros2/rclcpp/issues/722 which it is addressing.

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
wjwwood: I think both of the alternatives you suggested are better than having enable shared from this (personally). The first one is more dangerous for inexperienced users that make a mistake, but the second one is pretty safe, it just incurs the overhead of a shared pointer (versus a unique pointer) for each callback handle returned. Personally I'd go with the latter options. That mimics how callback groups and pub/sub/service's are handled within the node.

> I think we are always creating a shared pointer of NodeParameters.
So inheriting from std::enable_shared_from_this shouldn't be a problem (I should run full ci for confirming it).

That's true, but I feel that we're limiting our options (I'd like to eventually get the node interfaces as a non-shared pointer part of the node. Even if that might not be possible, this change takes us further from that goal, and I think it's not required.

Also it means that when the `NodeParameters` is destructed, all outstanding cleanup is done, and you don't have to worry about incomplete cleanup due to lingering callback handle objects.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -936,17 +938,68 @@ class Node : public std::enable_shared_from_this<Node>
...
+   *   or if it has been removed before.
+   */
+  RCLCPP_PUBLIC
+  void
+  remove_on_set_parameters_callback(const OnSetParametersCallbackHandle * const handler);
+
+  /// Register a callback to be called anytime a parameter is about to be changed.
+  /**
+   * With this method, only one callback can be set at a time. The callback that was previously
+   * set by this method is returned or `nullptr` if no callback was previously set.
```
wjwwood: I think we should deprecate this method as well. So that we end up with (eventually) just one way to add these callbacks. I can't think of a good reason to have this method and the other two long term.

ivanpauno: Yeah, I agree.
I was thinking about doing it in a separate PR (or just in a separate commit).
We could backport everything else without adding a deprecation warning.

ivanpauno:
@wjwwood I've addressed your comments.
I think that deprecating the all method is subject of a follow-up PR. So, this one can be backported independently. Also, I will have to open a bunch of PRs for deprecating it.
I think https://github.com/ros2/rclcpp/issues/609 is subject of another PR too.

wjwwood:
Thanks for working on it and iterating!

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7642/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3754/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6249/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/7487/)

ivanpauno:
All CI failures are also appearing in the nightlies. Merging!

PR #775, opened 2019-07-02
Made options optional. Updated API documentation.

esteve:
This PR updates the constructor to match that of `Node` (i.e. `options` should be optional). I also removed part of the API documentation that was no longer relevant.

### DIFF
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
@@ -115,14 +115,13 @@ public:
   /**
    * \param[in] node_name Name of the node.
    * \param[in] namespace_ Namespace of the node.
-   * \param[in] context The context for the node (usually represents the state of a process).
    * \param[in] options Additional options to control creation of the node.
    */
   RCLCPP_LIFECYCLE_PUBLIC
   LifecycleNode(
     const std::string & node_name,
     const std::string & namespace_,
-    const rclcpp::NodeOptions & options);
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
 
   RCLCPP_LIFECYCLE_PUBLIC
   virtual ~LifecycleNode();


ivanpauno:
CI (up to test_rclcpp rclcpp, fastrtps only)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7470/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3609/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6124/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7301/)

dirk-thomas:
The current change is not ABI compatible and can therefore not be backported to Dashing as-is.

@ivanpauno was this selected for a specific use case to be backported? The simple workaround seems to be to just pass the default value explicitly.

@esteve @ivanpauno If either of you still thinks this should be backported can you please create a custom PR targeting the `dashing` branch which adds a new constructor without the `options` argument which calls the existing constructor with the explicit default value.

sloretz:
> The current change is not ABI compatible and can therefore not be backported to Dashing as-is.

How does this break ABI? It looks like the constructor has the same number of arguments with the same types as before.

ivanpauno:
I think that usually defaults are implemented as temporary objects created at the point of call, and it calls the same symbol when using the default than when passing the parameter explicitly.
Compilers may create two symbols in this case, one for the method taking the default and one for the non default.

In the first case this would be ABI compatible, in the second not.
I think that most compilers do the first thing, but I can't confirm that.

For me, it's okay not doing a backport.

dirk-thomas:
I couldn't come up with a reference. With `g++` I can confirm that the symbol stays the same so for that compiler the change is ABI compatible.

> In the first case this would be ABI compatible, in the second not.

The second case would add a new (non-virtual) constructor with one argument less. That change should be ABI compatible too. I will take back my objection about the ABI compatibility then.

The other question remains though: why should this be backported?

ivanpauno:
> The other question remains though: why should this be backported?

There's not good reason, so I think not doing the backport is fine.
I'm deleting it from the project board.

esteve:
> @dirk-thomas The other question remains though: why should this be backported?

I don't really see why not, it'd make the lifecycle API consistent with the node API, it's not breaking the ABI and also fixes the documentation.

dirk-thomas:
Since there is another backport for this repo I don't mind to also pick this one, see #801.

PR #781, opened 2019-07-09
Switch the NodeParameters lock to recursive.

clalancette:
This is so that the on_set_parameter_callback can successfully
call other parameter methods (like get_parameter) without
deadlocking.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -376,6 +376,11 @@ NodeParameters::declare_parameter(
 {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
 
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot declare a parameter from within set callback");
+  }
+
   // TODO(sloretz) parameter name validation
   if (name.empty()) {
     throw rclcpp::exceptions::InvalidParametersException("parameter name must not be empty");
@@ -388,6 +393,8 @@ NodeParameters::declare_parameter(
   }
 
   rcl_interfaces::msg::ParameterEvent parameter_event;
+  parameter_modification_enabled_ = false;
+  RCLCPP_SCOPE_EXIT({parameter_modification_enabled_ = true;});
   auto result = __declare_parameter_common(
     name,
     default_value,
@@ -397,6 +404,7 @@ NodeParameters::declare_parameter(
     on_parameters_set_callback_,
     &parameter_event,
     ignore_override);
+  parameter_modification_enabled_ = true;
 
   // If it failed to be set, then throw an exception.
   if (!result.successful) {
@@ -416,6 +424,11 @@ void
 NodeParameters::undeclare_parameter(const std::string & name)
 {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
+
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot declare a parameter from within set callback");
+  }
 
   auto parameter_info = parameters_.find(name);
   if (parameter_info == parameters_.end()) {
@@ -470,6 +483,11 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
 {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
 
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot modify a parameter from within set callback");
+  }
+
   rcl_interfaces::msg::SetParametersResult result;
 
   // Check if any of the parameters are read-only, or if any parameters are not
@@ -518,6 +536,11 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
   rcl_interfaces::msg::ParameterEvent parameter_event_msg;
   parameter_event_msg.node = combined_name_;
   for (auto parameter_to_be_declared : parameters_to_be_declared) {
+    // Since the callback is explicitly nullptr in __declare_parameter_common,
+    // this isn't strictly needed.  We do it anyway because it is cheap and in
+    // case the implementation changes in the future.
+    parameter_modification_enabled_ = false;
+
     // This should not throw, because we validated the name and checked that
     // the parameter was not already declared.
     result = __declare_parameter_common(
@@ -529,6 +552,8 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
       nullptr,  // callback is explicitly null, so that it is called only once, when setting below.
       &parameter_event_msg,
       true);
+    parameter_modification_enabled_ = true;
+
     if (!result.successful) {
       // Declare failed, return knowing that nothing was changed because the
       // staged changes were not applied.
@@ -575,7 +600,9 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
     }
   }
 
-  // Set the all of the parameters including the ones declared implicitly above.
+  // Set all of the parameters including the ones declared implicitly above.
+  parameter_modification_enabled_ = false;
+  RCLCPP_SCOPE_EXIT({parameter_modification_enabled_ = true;});
   result = __set_parameters_atomically_common(
     // either the original parameters given by the user, or ones updated with initial values
     *parameters_to_be_set,
@@ -583,6 +610,7 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
     parameters_,
     // this will get called once, with all the parameters to be set
     on_parameters_set_callback_);
+  parameter_modification_enabled_ = true;
 
   // If not successful, then stop here.
   if (!result.successful) {
@@ -817,6 +845,13 @@ NodeParameters::list_parameters(const std::vector<std::string> & prefixes, uint6
 NodeParameters::OnParametersSetCallbackType
 NodeParameters::set_on_parameters_set_callback(OnParametersSetCallbackType callback)
 {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot modify a parameter from within set callback");
+  }
+
   auto existing_callback = on_parameters_set_callback_;
   on_parameters_set_callback_ = callback;
   return existing_callback;
@@ -832,6 +867,13 @@ NodeParameters::set_on_parameters_set_callback(OnParametersSetCallbackType callb
 void
 NodeParameters::register_param_change_callback(ParametersCallbackFunction callback)
 {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot modify a parameter from within set callback");
+  }
+
   if (on_parameters_set_callback_) {
     RCLCPP_WARN(
       node_logging_->get_logger(),
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -376,13 +376,7 @@ NodeParameters::declare_parameter(
 {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
 
-  if (!parameter_modification_enabled_) {
-    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
-            "cannot declare a parameter from within set callback");
-  }
-
-  parameter_modification_enabled_ = false;
-  RCLCPP_SCOPE_EXIT({parameter_modification_enabled_ = true;});
+  ParameterMutationRecursionGuard guard(parameter_modification_enabled_);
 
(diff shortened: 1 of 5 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7509/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3647/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6151/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7334/)

ivanpauno:
I would like to see a test with an example usage. IMHO, this can easily end in problematic code.
e.g.:
Someone calls `set_parameters_atomically` setting two params. For checking is one is accepted or not, the other parameter is get ... but it hasn't been updated yet.

I'm afraid other problematic cases are possible. Probably, most of the problems can be worked-around carefully writing the callback. But it sounds tricky.

clalancette:
> Someone calls `set_parameters_atomically` setting two params. For checking is one is accepted or not, the other parameter is get ... but it hasn't been updated yet.

Actually, this is exactly the case this is trying to solve :).

https://github.com/clalancette/velodyne/blob/param-change-callback/velodyne_pointcloud/src/conversions/convert.cc#L141 is the code that I'm currently trying to make work.  The problem is that the callback gets the *changed* parameters, but in order to successfully do updates I also need the "other" parameters that have not been updated.  So I call `get_parameters` to get all of the parameters initially, but that immediately deadlocks.

Besides a recursive mutex, another way to do this would be to change the semantics of the vector that is passed to the `on_parameters_set_callback`.  Instead of passing just the changed parameters, it would pass a vector of *all* of the parameters, with the new ones updated appropriately.  That would give the callback the "final" state of the parameters to check and act upon.  I don't like this solution quite as much (particularly because it changes semantics without changing the signature), but I would be OK going that way as well.

Additional thoughts?

ivanpauno:
> > Someone calls `set_parameters_atomically` setting two params. For checking is one is accepted or not, the other parameter is get ... but it hasn't been updated yet.
> 
> Actually, this is exactly the case this is trying to solve :).
> 
> https://github.com/clalancette/velodyne/blob/param-change-callback/velodyne_pointcloud/src/conversions/convert.cc#L141 is the code that I'm currently trying to make work. The problem is that the callback gets the _changed_ parameters, but in order to successfully do updates I also need the "other" parameters that have not been updated. So I call `get_parameters` to get all of the parameters initially, but that immediately deadlocks.
> 
> Besides a recursive mutex, another way to do this would be to change the semantics of the vector that is passed to the `on_parameters_set_callback`. Instead of passing just the changed parameters, it would pass a vector of _all_ of the parameters, with the new ones updated appropriately. That would give the callback the "final" state of the parameters to check and act upon. I don't like this solution quite as much (particularly because it changes semantics without changing the signature), but I would be OK going that way as well.
> 
> Additional thoughts?

@clalancette thanks for the example.
I would like to see a new callback signature alternative, that includes all the parameters and its descriptors. We could have different signatures alternatives, doing something similar to `any_subscription_callback`.
For the moment, I think this is the easiest change is just allowing the recursive mutex. The only drawback I see is that the user have to be really careful when writing the callback.

wjwwood:
> Besides a recursive mutex, another way to do this would be to...

I don't think we want to create a duplicate set of parameters with the new values set, just for performance reasons, or at least that should be an optional signature/behavior. It would be better (imo) to let the user introspect the current state using the existing API and sort out what to do based on that.

clalancette:
> The changes lgtm, but I'd like us to consider whether or not we allow setting and/or undeclaring parameters within this callback before merging. I think that setting parameters from within the on set parameter callback will be problematic unless you're extremely careful.

Yeah, agreed.

> 
> The simplest thing to do would be to have a flag in `NodeParameters` which was set true when mutating state and false when leaving, then anything that would like to mutate state would also check to make sure the lock had not been acquired recursively, throwing otherwise (to avoid a dead lock or unexpected behavior). Functions that would be affected (imo) are: `declare_parameter`, `undeclare_parameter`, `set_parameters_atomically`, and `set_on_parameters_set_callback`.

It's unfortunate that there doesn't seem to be a way to ask `std::recursive_mutex` how many times it has been locked; that would be good enough.  But yeah, a flag would work here; I'll go ahead and do that.

> 
> And on that point, `set_on_parameters_set_callback` should probably be locked as well, otherwise you could swap the callback concurrently with calling the callback due to setting a parameter. Maybe it needs a separate lock or maybe it can reuse the existing lock, I'm not sure.

Yeah, agreed.  I think this both needs the lock, and needs to check the recursion as above.  I'll add that as well.

clalancette:
> For the moment, I think that the easiest change is just allowing the recursive mutex. The only drawback I see is that the user have to be really careful when writing the callback.

Yeah, it is not totally ideal.  However, it is a strict improvement over what we have now, so even if we do decide to add an additional callback signature, I think it is still worthwhile to change this to a recursive lock.

clalancette:
OK, so I've now updated this PR in ce8c611 to have a way to prevent modifications while in the callback.  I've also added tests in that commit.

The implementation isn't great, mostly because I had to locate the `parameter_modification_enabled_` flag pretty far away from where we really need to set it.  In turn, that's because `__set_parameters_atomically` and `__declare_parameters` aren't class methods, but are just local functions.  A slightly nicer (but more invasive) alternative is to make those class methods, and then relocate the `parameter_modification_enabled_` flag close to where the callback is called.  I'm willing to do that, but I'm not sure if that is the direction we want to go.  Opinions?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -470,6 +483,11 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
 {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
 
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot modify a parameter from within set callback");
+  }
+
```
ivanpauno: I think we can add here:
```
parameter_modification_enabled_ = false;
RCLCPP_SCOPE_EXIT({parameter_modification_enabled_ = true;});
```
and avoid similar code in each call to `__*_common` within the method. Is that correct?

We can maybe apply that style in the other methods, keeping that code after the 
```cpp
if (!parameter_modification_enabled_) {
``` 
in `declare_parameter` method too.
I think it's a bit clearer.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -832,6 +867,13 @@ NodeParameters::set_on_parameters_set_callback(OnParametersSetCallbackType callb
 void
 NodeParameters::register_param_change_callback(ParametersCallbackFunction callback)
 {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot modify a parameter from within set callback");
```
ivanpauno: Consider: `cannot register a new set callback within a set callback`

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -817,6 +845,13 @@ NodeParameters::list_parameters(const std::vector<std::string> & prefixes, uint6
 NodeParameters::OnParametersSetCallbackType
 NodeParameters::set_on_parameters_set_callback(OnParametersSetCallbackType callback)
 {
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot modify a parameter from within set callback");
```
ivanpauno: Consider: `cannot register a new set callback within a set callback`

ivanpauno:
> The implementation isn't great, mostly because I had to locate the `parameter_modification_enabled_` flag pretty far away from where we really need to set it. In turn, that's because `__set_parameters_atomically` and `__declare_parameters` aren't class methods, but are just local functions. A slightly nicer (but more invasive) alternative is to make those class methods, and then relocate the `parameter_modification_enabled_` flag close to where the callback is called. I'm willing to do that, but I'm not sure if that is the direction we want to go. Opinions?

I think that changing those free functions for private methods would be good. It's certainly out of the scope of the PR, but I agree with applying the changes here if you want.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -397,6 +404,7 @@ NodeParameters::declare_parameter(
     on_parameters_set_callback_,
     &parameter_event,
     ignore_override);
+  parameter_modification_enabled_ = true;
```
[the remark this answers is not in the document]
clalancette: I originally did this because I wanted to set `parameter_modification_enabled_` to true as soon as possible (and not wait for the end of the function).  But looking at it again, I think a better way is to just introduce a local scope and let the `RCLCPP_SCOPE_EXIT` take care of it.  I'll make that change.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -417,6 +425,11 @@ NodeParameters::undeclare_parameter(const std::string & name)
 {
   std::lock_guard<std::recursive_mutex> lock(mutex_);
 
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot declare a parameter from within set callback");
```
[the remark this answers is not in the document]
clalancette: Oops!  Fixing.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
clalancette: Good call, changing.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
clalancette: Fixing.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
clalancette: Oh, I like it.  This keeps all of the code that deals with `parameter_modification_enabled_` in the same place.  It has the slight downside in that it is holding the "lock" for a longer period of time, but that doesn't seem to matter in my testing so far.  I'm going to do a bit more testing here, and assuming that is the case, there is no need to touch `__set_parameters_atomically` or `__declare_parameters`.  Thanks for the idea!

clalancette:
> I think that changing those free functions for private methods would be good. It's certainly out of the scope of the PR, but I agree with applying the changes here if you want.

With the changes from 566f519 (coming from [this](https://github.com/ros2/rclcpp/pull/781#discussion_r302083407) comment), this is no longer relevant, so I won't change anything for this.  I'm going to run CI again; I think this is ready for another review.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7521/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3656/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6159/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7349/)

ivanpauno:
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/7349/)

Please, solve the windows warnings before merging.

clalancette:
> Please, solve the windows warnings before merging.

Yeah, I just noticed.  Thanks for the heads up, will fix.

clalancette:
CI (hopefully last one):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7525/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3662/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6163/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7354/)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -374,7 +374,15 @@ NodeParameters::declare_parameter(
...
-  std::lock_guard<std::mutex> lock(mutex_);
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+
+  if (!parameter_modification_enabled_) {
+    throw rclcpp::exceptions::ParameterModifiedInCallbackException(
+            "cannot declare a parameter from within set callback");
+  }
+
+  parameter_modification_enabled_ = false;
+  RCLCPP_SCOPE_EXIT({parameter_modification_enabled_ = true;});
```
wjwwood: You may want to consider moving this into a class, e.g. `ParameterMutationRecursionGuard` or something, that checks if the variable is set (throw if it is) then sets this variable to false in the constructor, and then sets it back to false in the destructor. Just to de-duplicate this code.

wjwwood:
Specifically, the "requested changes" is just the docs. The de-duplication suggestion could be ignored or done in a follow up pr.

clalancette:
> Specifically, the "requested changes" is just the docs. The de-duplication suggestion could be ignored or done in a follow up pr.

I ended up doing both, since they were both good ideas.  See 2031330.  New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7532/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3669/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6169/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7361/)

PR #786, opened 2019-07-17
Fix broken transient local durability policy

ivanpauno:
We weren't allowing using intraprocess comm with non-volatile durability policy.
But we were actually breaking it, as `inter_process_publish_needed` calculation was wrong.

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -111,7 +111,8 @@ public:
     // as do_intra_process_publish takes the ownership of the message.
     uint64_t message_seq;
     bool inter_process_publish_needed =
-      get_subscription_count() > get_intra_process_subscription_count();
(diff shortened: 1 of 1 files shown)


ivanpauno:
Full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7589/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3720/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6213/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7416/)

alsora:
`intra_process_is_enabled_` can't be `false` when we reach that line.
Its value is already checked at the beginning of the `publish` method

ivanpauno:
> `intra_process_is_enabled_` can't be `false` when we reach that line.
> Its value is already checked at the beginning of the `publish` method

Thanks! Sorry for the false alarm :man_facepalming:.

PR #789, opened 2019-07-19
Basic trace instrumentation

iluetkeb:
This is the first part of the instrumentation for the work described at https://discourse.ros.org/t/alpha-release-of-ros-2-tracing-tools/9833

### DIFF
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -26,6 +26,8 @@
 #include "rclcpp/allocator/allocator_common.hpp"
 #include "rclcpp/function_traits.hpp"
 #include "rclcpp/visibility_control.hpp"
+#include "tracetools/tracetools.h"
+#include "tracetools/utils.hpp"
 
 namespace rclcpp
 {
@@ -155,6 +157,7 @@ public:
   void dispatch(
     std::shared_ptr<MessageT> message, const rmw_message_info_t & message_info)
   {
+    TRACEPOINT(callback_start, (const void *)this, false);
     if (shared_ptr_callback_) {
       shared_ptr_callback_(message);
     } else if (shared_ptr_with_info_callback_) {
@@ -174,11 +177,13 @@ public:
     } else {
       throw std::runtime_error("unexpected message without any callback set");
     }
+    TRACEPOINT(callback_end, (const void *)this);
   }
 
   void dispatch_intra_process(
     ConstMessageSharedPtr message, const rmw_message_info_t & message_info)
   {
+    TRACEPOINT(callback_start, (const void *)this, true);
     if (const_shared_ptr_callback_) {
       const_shared_ptr_callback_(message);
     } else if (const_shared_ptr_with_info_callback_) {
@@ -195,11 +200,13 @@ public:
         throw std::runtime_error("unexpected message without any callback set");
       }
     }
+    TRACEPOINT(callback_end, (const void *)this);
   }
 
   void dispatch_intra_process(
     MessageUniquePtr message, const rmw_message_info_t & message_info)
   {
+    TRACEPOINT(callback_start, (const void *)this, true);
     if (shared_ptr_callback_) {
       typename std::shared_ptr<MessageT> shared_message = std::move(message);
       shared_ptr_callback_(shared_message);
@@ -217,11 +224,37 @@ public:
     } else {
       throw std::runtime_error("unexpected message without any callback set");
     }
+    TRACEPOINT(callback_end, (const void *)this);
   }
 
   bool use_take_shared_method()
   {
     return const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_;
+  }
+
+  void register_callback_for_tracing()
+  {
+    if (shared_ptr_callback_) {
+      TRACEPOINT(
+        rclcpp_callback_register,
+        (const void *)this,
+        get_symbol(shared_ptr_callback_));
+    } else if (shared_ptr_with_info_callback_) {
+      TRACEPOINT(
+        rclcpp_callback_register,
+        (const void *)this,
+        get_symbol(shared_ptr_with_info_callback_));
+    } else if (unique_ptr_callback_) {
+      TRACEPOINT(
+        rclcpp_callback_register,
+        (const void *)this,
+        get_symbol(unique_ptr_callback_));
+    } else if (unique_ptr_with_info_callback_) {
+      TRACEPOINT(
+        rclcpp_callback_register,
+        (const void *)this,
+        get_symbol(unique_ptr_with_info_callback_));
+    }
   }
 
 private:
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -46,6 +46,7 @@
 #include "rclcpp/type_support_decl.hpp"
 #include "rclcpp/visibility_control.hpp"
 #include "rclcpp/waitable.hpp"
+#include "tracetools/tracetools.h"
 
 namespace rclcpp
 {
@@ -117,6 +118,14 @@ public:
         options.event_callbacks.liveliness_callback,
         RCL_SUBSCRIPTION_LIVELINESS_CHANGED);
     }
+    TRACEPOINT(
+      rclcpp_subscription_callback_added,
+      (const void *)get_subscription_handle().get(),
+      (const void *)&any_callback_);
+    // The callback object gets copied, so if registration is done too early/before this point
+    // (e.g. in `AnySubscriptionCallback::set()`), its address won't match any address used later
+    // in subsequent tracepoints.
+    any_callback_.register_callback_for_tracing();
   }
 
   /// Called after construction to continue setup that requires shared_from_this().
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -30,6 +30,8 @@
 #include "rclcpp/rate.hpp"
 #include "rclcpp/utilities.hpp"
 #include "rclcpp/visibility_control.hpp"
+#include "tracetools/tracetools.h"
+#include "tracetools/utils.hpp"
 
 #include "rcl/error_handling.h"
 #include "rcl/timer.h"
@@ -133,6 +135,14 @@ public:
   )
   : TimerBase(clock, period, context), callback_(std::forward<FunctorT>(callback))
   {
+    TRACEPOINT(
+      rclcpp_timer_callback_added,
+      (const void *)get_timer_handle().get(),
+      (const void *)&callback_);
+    TRACEPOINT(
+      rclcpp_callback_register,
+      (const void *)&callback_,
+      get_symbol(callback_));
   }
 
   /// Default destructor.
@@ -152,7 +162,9 @@ public:
     if (ret != RCL_RET_OK) {
       throw std::runtime_error("Failed to notify timer that callback occurred");
     }
+    TRACEPOINT(callback_start, (const void *)&callback_, false);
     execute_callback_delegate<>();
+    TRACEPOINT(callback_end, (const void *)&callback_);
   }
 
   // void specialization
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -14,6 +14,7 @@ find_package(rosgraph_msgs REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 find_package(rosidl_typesupport_c REQUIRED)
 find_package(rosidl_typesupport_cpp REQUIRED)
+find_package(tracetools REQUIRED)
 
 # Default to C++14
 if(NOT CMAKE_CXX_STANDARD)
@@ -114,7 +115,9 @@ ament_target_dependencies(${PROJECT_NAME}
   "builtin_interfaces"
   "rosgraph_msgs"
   "rosidl_typesupport_cpp"
-  "rosidl_generator_cpp")
+  "rosidl_generator_cpp"
+  "tracetools"
+)
 
 # Causes the visibility macros to use dllexport rather than dllimport,
 # which is appropriate when building the dll but not consuming it.
@@ -141,6 +144,7 @@ ament_export_dependencies(rosidl_typesupport_cpp)
 ament_export_dependencies(rosidl_typesupport_c)
 ament_export_dependencies(rosidl_generator_cpp)
 ament_export_dependencies(rcl_yaml_param_parser)
+ament_export_dependencies(tracetools)
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -23,6 +23,7 @@
 #include "rclcpp/function_traits.hpp"
 #include "rclcpp/visibility_control.hpp"
 #include "rmw/types.h"
+#include "tracetools/tracetools.h"
 
 namespace rclcpp
 {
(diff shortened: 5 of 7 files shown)


christophebedard:
Also see https://github.com/ros2/rcl/pull/473

nuclearsandwich:
Apologies for the churn. Keyboard and mouse hand were not cooperating

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -107,6 +108,14 @@ class Subscription : public SubscriptionBase
...
         RCL_SUBSCRIPTION_LIVELINESS_CHANGED);
     }
+    TRACEPOINT(
+      rclcpp_subscription_callback_added,
+      (const void *)get_subscription_handle().get(),
+      (const void *)&any_callback_);
+    // TODO(christophebedard) find another way
+    // this object seems to get copied, so if registration is done too early
+    // (e.g. in `set()`), `this` won't match any address used later
+    any_callback_.register_callback_function();
```
hidmic: @iluetkeb I don't see other reasonable way either, to cope with `std::function` copies I mean. In which case I'm reluctant to keep that TODO, as opposed to clearly documenting that tracepoints should not be registered till `AnyCallback` has hit "stable" storage.

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -214,13 +221,39 @@ class AnySubscriptionCallback
...
     }
+    TRACEPOINT(callback_end, (const void *)this);
   }
 
   bool use_take_shared_method()
   {
     return const_shared_ptr_callback_ || const_shared_ptr_with_info_callback_;
   }
 
+  void register_callback_function()
```
hidmic: @iluetkeb nit: in the context of an `AnyCallback` method, wouldn't it make more sense for this method to be called `register_tracepoints()` or something along those lines?

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -190,6 +201,39 @@ class GenericTimer : public TimerBase
...
+
+private:
+  template<
+    typename CallbackT = FunctorT,
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<CallbackT, VoidCallbackType>::value
+    >::type * = nullptr
+  >
+  void
+  register_callback_function()
```
hidmic: @iluetkeb same nit about naming, though in this case maybe `register_callback_tracepoint()` makes more sense?

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -133,6 +135,13 @@ class GenericTimer : public TimerBase
   )
   : TimerBase(clock, period, context), callback_(std::forward<FunctorT>(callback))
   {
+    TRACEPOINT(
+      rclcpp_timer_callback_added,
+      (const void *)get_timer_handle().get(),
+      (const void *)&callback_);
+    // TODO(christophebedard) maybe there's a simpler way
+    // to treat callback_ as an std::function here explicitly
```
hidmic: @iluetkeb not following this TODO, mind to explain?

rclcpp/include/rclcpp/any_subscription_callback.hpp
christophebedard: registering tracepoints is an actual thing with LTTng/tracers, but at this point I doubt it would cause confusion

rclcpp/include/rclcpp/timer.hpp
christophebedard: I'm not @iluetkeb :laughing:, but IIRC it was about finding a way to avoid needing any `register_callback_function()` methods at all (with `std::enable_if`). It seemed to be necessary, because moving

```
TRACEPOINT(
  rclcpp_callback_register,
  (const void *)&callback_,
  get_symbol(function));
```

to the constructor, instead of calling `this->register_callback_function()` and then casting it to its actual type, does not work. I didn't know how to avoid this.

hidmic:
@iluetkeb friendly ping.

rclcpp/include/rclcpp/timer.hpp
hidmic: Do you happen to know why it does not work?

iluetkeb:
> @iluetkeb one extra question. I've seen these tracepoints declared back in the `tracetools` package and it makes me wonder about the scalability of these tools. 

What do you mean by "scalability"?

> Is that a limitation of the current implementation or is it actually impossible to push those declarations closer to the relevant packages?

With "relevant packages", you mean rcl and rclcpp? Sure, we could put them in there. However, that would also meant that we would have to put the supporting code (which currently uses LTTng) in those packages as well. It's not big, but we figured by having one more indirection, it would overall be easier. Also easier to change and support other tracing packages in the future.

Moreover, right now, tracing is a noop by default. You have to recompile tracetools, but only tracetools, to change that. If we put it into rcl/rclcpp, we would have to recompile those, which is much more of an effort.

rclcpp/include/rclcpp/subscription.hpp
iluetkeb: @christophebedard any objections to removing that?

rclcpp/include/rclcpp/timer.hpp
iluetkeb: Well, as I understand it, a "tracepoint" is not a location in the user code, but a tracing function with a specific signature. This seems slightly counter-intuitive in naming, so @christophebedard correct me if I'm wrong, but it appears to be used like that by the trace libraries.

This means, the tracepoint is already there and what's being registered is the user callback function within the ROS node.

So, I would say the naming is appropriate.

iluetkeb:
> To be forced to commit a each new tracepoint to the `tracetools` package doesn't lend itself to reuse by interested individuals wanting to trace their packages.

Ah! You're thinking of people adding new tracepoints, right?

In the tracing framework we're currently using, LTTng, anybody can define tracepoints wherever and however they like, and LTTng will take care of aggregating the data. So, people can add custom tracepoints to their apps directly using LTTNg without having to touch tracetools.

That said, since ROS 2 is cross-platform, and LTTng is Linux only, the tracetools would be a possible place to implement a cross-platform tracing layer. For example, I haven't used it myself, but I know that Event Tracing for Windows is, in theory, supposed to also be able to emit user-space events. We haven't thought about this much, but if we want to go that way, we might improve tracetools a bit to make that sort of thing easier.

In the past, i.e. for ROS 1, we had a generic "message received" tracepoint. It wasn't very capable, but at least you had something similar to logging integrated with the rest of the tracing. Maybe we should bring that back as a first step.

rclcpp/include/rclcpp/subscription.hpp
christophebedard: sure, no problem!

rclcpp/include/rclcpp/timer.hpp
christophebedard: > It's not immediately obvious that you're registering the symbol of a callback for tracing.

I agree. To be honest, it was a name I used at first to try doing it as a workaround and it just stuck.

Maybe something like `register_callback_symbol()`? Maybe it's still not clear that it's referring to the callback _function_. I'm open to anything obviously (for all instances of this registration pattern).

rclcpp/include/rclcpp/timer.hpp
christophebedard: Not unfortunately. I'd defer to people more knowledgeable about C++ templating. It seems like it just does not know what type it actually is in the constructor, and that I need to use `std::enable_if<>` to figure that out.

christophebedard: I could more into it, though

iluetkeb:
@hidmic @dirk-thomas @nuclearsandwich is this being considered for E?

hidmic:
@iluetkeb I was expecting you guys to resolve pending review comments. 

About getting this into Eloquent, both API and feature freeze are happening on Oct 14th. You can check the distro roadmap [here](https://index.ros.org/doc/ros2/Releases/Release-Eloquent-Elusor/#timeline-before-the-release).

iluetkeb:
> @iluetkeb I was expecting you guys to resolve pending review comments.

Ah, I thought none were open, but I see that in one case, we wanted to rename a function. Is that it?

Sorry, but it would really help us, if you could state after Christophe's comments whether that is an OK solution for you or not.

> About getting this into Eloquent, both API and feature freeze are happening on Oct 14th. You can check the distro roadmap [here](https://index.ros.org/doc/ros2/Releases/Release-Eloquent-Elusor/#timeline-before-the-release).

Thanks for the link. I see Oct 18th for feature freeze there. We could probably make the 14th as well, but could you say which is correct, please?

iluetkeb:
there also seems to be a problem with the dependency statement

rclcpp/include/rclcpp/subscription.hpp
hidmic: Yes, please. Remove that TODO **AND** document the caveats of the tracing system w.r.t. memory copies.

rclcpp/include/rclcpp/timer.hpp
hidmic: Maybe `register_callback_for_tracing()`?  Just to associate this functionality with its final purpose.

rclcpp/include/rclcpp/any_subscription_callback.hpp
hidmic: I think we could apply the same rationale as below on this one i.e. make the name be related to tracing in some way.

rclcpp/include/rclcpp/timer.hpp
hidmic: Did we make any progress on this? Going over this again, and without any additional context. it seems you are forced to cast the provided `FunctorT` callback to a concrete callback type to retrieve its symbol. If you guys used SFINAE on `get_symbol` I think you'd be able to put the macro back in the constructor.

hidmic:
> if you could state after Christophe's comments whether that is an OK solution for you or not.

Done.

> I see Oct 18th for feature freeze there. We could probably make the 14th as well, but could you say which is correct, please?

Doh, my bad. **Oct 18th is the correct date.**

rclcpp/include/rclcpp/timer.hpp
christophebedard: To be honest, while I think I understand SFINAE, I've played around a bit and I'm not sure how I'm supposed to use apply it to `get_symbol()`.

I'm not sure if forward-declaring `rclcpp::TimerBase` in `tracetools` (which is what I'm assuming I would have to do?) really makes this worth being able to move the macro to the constructor.

hidmic: > I'm not sure if forward-declaring `rclcpp::TimerBase` in `tracetools`

In principle, you shouldn't need to do that. And actually, since you know the exact types there's no need for template metaprogramming, just locally overload `get_symbol` appropriately and [ADL](https://en.cppreference.com/w/cpp/language/adl) will do the rest. 

I'm curious about https://github.com/ros2/rclcpp/pull/789#discussion_r320478335 and why/how it did not work though.

iluetkeb: Yeah, I've come to the same conclusion, and I'll try that first.

iluetkeb:
@christophebedard @hidmic I don't understand where the error resolving the "tracetools" rosdep key comes from. We have "tracetools" released both as a package and included in ros2/ros2#748.

dirk-thomas:
> I don't understand where the error resolving the "tracetools" rosdep key comes from. We have "tracetools" released both as a package and included in ros2/ros2#748.

The failing job is an Eloquent PR job. It builds the code for ROS Eloquent. As far as I can see you have **not** released `tracetools` into Eloquent yet.

christophebedard:
Makes sense. I'll create a new version/release once we've addressed the comments on all PRs.

rclcpp/include/rclcpp/timer.hpp
christophebedard: @hidmic @iluetkeb solved! The reason why it wouldn't compile with the call to `get_symbol` directly in the constructor was because the tests use lambdas with captures. Apparently they're different and can't really be `std::function`s. Not sure why the `std::enable_if`s work though, since they only check for `std::function`s.

The solution is just to add a `get_symbol` overload which takes a generic 
type and calls `typeid()`, which is exactly what we want for lambdas.

I think this makes more sense, since `tracetools` isn't actually aware of this intricacy in `rclcpp` and simply provides a generic fallback.

iluetkeb:
Sorry for the churn, too many remotes in one repo makes for the odd mistake.

rclcpp/include/rclcpp/timer.hpp
iluetkeb: The PR has been updated and I consider this resolved, then. Plz reopen otherwise.

christophebedard:
The PR for the release into Eloquent is up! https://github.com/ros/rosdistro/pull/22679

wjwwood:
I think this is also unblocked now that https://github.com/ros2/ros2/pull/748 is merged.

hidmic:
CI for `rcl` and `rclcpp`:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8403/)
- Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4319/)
- MacOS [Build Status](https://ci.ros2.org/job/ci_osx/6843/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8305/)

christophebedard:
It's failing on Windows because `rclcpp` cannot resolve `_demangle_symbol()` and `_get_symbol_funcptr()`, which are defined in `tracetools` [here](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/blob/ad46e04ebe98a37f3faedb0bc1a458c37d3aa971/tracetools/include/tracetools/utils.hpp) and implemented [here](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/blob/ad46e04ebe98a37f3faedb0bc1a458c37d3aa971/tracetools/src/utils.cpp).

I suspect that it might be because they're not using [`TRACETOOLS_PUBLIC`](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/blob/ad46e04ebe98a37f3faedb0bc1a458c37d3aa971/tracetools/include/tracetools/visibility_control.hpp).

I made the fix, but can't test it on Windows. @hidmic or @wjwwood, if you think [this fix](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/commit/6fe7786e123b61c6c9934a73be9cca6914276778) looks good, could you run the CI again using `6fe7786e123b61c6c9934a73be9cca6914276778` for the `ros2_tracing` repo?

This is another situation where being able to completely preprocess out all tracing-related/instrumentation calls would be nice, but now is not really the time :laughing:

iluetkeb:
> This is another situation where being able to completely preprocess out all tracing-related/instrumentation calls would be nice, but now is not really the time 

I implemented that, but you have to specify the right option. Maybe we should set that option on Windows by default, until someone implements tracing on Windows.

hidmic:
Re-running Windows CI against on `6fe7786e123b61c6c9934a73be9cca6914276778`:

- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8340/)

iluetkeb:
@hidmic hold it, I've made more fixes

iluetkeb:
Try febd57834ea9c9cb06324a5a18c3495c99fd61c7 instead.

hidmic:
@iluetkeb if CI passes, you guys still have to update the ros2.repos on ros2/ros2 after merging (and tagging?) those fixes before we merge this one.

christophebedard:
Windows CI passed!

I've created `0.2.9`.

* https://github.com/ros2/ros2/pull/805
* https://github.com/ros/rosdistro/pull/22703

hidmic:
One last CI run for `rcl` and `rclcpp`:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8455/)
- Linux aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4350/)
- MacOS [Build Status](https://ci.ros2.org/job/ci_osx/6871/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8351/)

hidmic:
Alright, going in!

PR #793, opened 2019-07-24
Remove non-package from ament_target_dependencies()

sloretz:
blocks ament/ament_cmake#180

### DIFF
--- a/rclcpp_components/CMakeLists.txt
+++ b/rclcpp_components/CMakeLists.txt
@@ -68,8 +68,7 @@ if(BUILD_TESTING)
   add_library(test_component SHARED test/components/test_component.cpp)
   ament_target_dependencies(test_component
     "class_loader"
-    "rclcpp"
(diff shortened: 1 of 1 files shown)


rclcpp_components/CMakeLists.txt
```diff
@@ -68,8 +68,7 @@ if(BUILD_TESTING)
   add_library(test_component SHARED test/components/test_component.cpp)
   ament_target_dependencies(test_component
     "class_loader"
-    "rclcpp"
-    "rclcpp_components")
```
jacobperron: Because test_component.cpp [uses `rclcpp_components`](https://github.com/ros2/rclcpp/blob/d31ea14985b643a3a4a5364e01d269d5f4d04e9a/rclcpp_components/test/components/test_component.cpp#L61), should we add `target_link_libraries(test_component ${PROJECT_NAME})`?

Issue #783, opened 2019-07-12
Promise already satisfied


ipa-hsd:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04.2 LTS
- Installation type:
  - binaries
- Version or commit hash:
  - 0.7.6
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
      auto send_goal_ops = typename rclcpp_action::Client<ROS2_T>::SendGoalOptions;
      send_goal_ops.result_callback =
        [this](auto res2) mutable
      {

      };
      .
      .
      .
      // send goal and wait for result
      auto gh2_future = client_->async_send_goal(goal2, send_goal_ops);
      auto result = client_->async_get_result(gh2_future.get()).get();
```
This seems to be similar to the test case 
https://github.com/ros2/rclcpp/blob/bfee90ab7a210657e1dc02c6b7754eb917687513/rclcpp_action/test/test_client.cpp#L395-L413

#### Expected behavior
I get the result without any errors

#### Actual behavior
```
terminate called after throwing an instance of 'std::future_error'
  what():  std::future_error: Promise already satisfied
```
#### Additional information
```
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff5698801 in __GI_abort () at abort.c:79
#2  0x00007ffff5ced957 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff5cf3ab6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff5cf3af1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff5cf3d24 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff5cef929 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
[...]

PR #806, opened 2019-07-30
[rclcpp_action] Promise after goal_handle's result_awareness is set

vinnamkim:
Solve #783

As @ipa-mdl said, `make_result_aware` might get called twice. This happens because of setting `promise` value before `is_result_aware_` isn't set properly.

Signed-off-by: vinnamkim <vinnam.kim@gmail.com>

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -372,10 +372,6 @@ public:
           std::lock_guard<std::mutex> guard(goal_handles_mutex_);
           goal_handles_[goal_handle->get_goal_id()] = goal_handle;
         }
-        promise->set_value(goal_handle);
-        if (options.goal_response_callback) {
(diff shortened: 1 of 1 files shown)


ipa-mdl:
If I am not mistaken, this fix might lead to `goal_response_callback` getting called after `result_callback` in some cases.

vinnamkim:
> If I am not mistaken, this fix might lead to `goal_response_callback` getting called after `result_callback` in some cases.

I guess in these cases you told the previous code also makes the same results, since they are asynchronous callbacks

jacobperron:
> If I am not mistaken, this fix might lead to goal_response_callback getting called after result_callback in some cases.

Yeah, good point. Giving this another look, it might be better to guard against the race more explicitly rather than changing the order in which callbacks are called. See #808 for a proposed alternative.

vinnamkim:
The alternative seems better. Close this issue.

PR #809, opened 2019-07-31
Overload on_parameter_events function

alsora:
Currently the on_parameter_event function can't be called from a class that derives from `node`.

The following fails
```
class MyNode : public rclcpp::Node
{
   MyNode() : Node("my_node")
   {
     auto AsyncParameterClient::OnParameterEvent(this, callback);
   }
}
```

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -128,18 +128,12 @@ public:
       options);
   }
 
-  /**
-   * The NodeT type only needs to have a method called get_node_topics_interface()
-   * which returns a shared_ptr to a NodeTopicsInterface, or be a
-   * NodeTopicsInterface pointer itself.
-   */
   template<
     typename CallbackT,
-    typename NodeT,
     typename AllocatorT = std::allocator<void>>
   static typename rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr
   on_parameter_event(
-    NodeT && node,
+    rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr node_topics_interface,
     CallbackT && callback,
     const rclcpp::QoS & qos = (
       rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_parameter_events))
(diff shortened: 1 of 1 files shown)


ivanpauno:
There's an example working here:
https://github.com/ros2/demos/blob/0db63dde477712a02a925424a7587a5f0bb8160a/demo_nodes_cpp/src/parameters/parameter_events_async.cpp#L30-L61

Can you provide a complete example of the problem? And the message of the failure too.

alsora:
The difference between the demo and my code is that I'm not creating a parameter client, but rather I only want to subscribe to the "events" topic.

AsyncParameterClient provides a static function for doing that.
This PR is adding the same overloads used for the AsyncParameterClient also to the static function.

(I will add the error later)

ivanpauno:
> The difference between the demo and my code is that I'm not creating a parameter client, but rather I only want to subscribe to the "events" topic.

Sorry, I see the difference now.

I will take a look, as that template was supposed to accept any "Node like" object (see https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp).
That also includes a `LifecycleNode` (which doesn't inherit from `Node`), so the patch is not ok.

alsora:
Ok, I didn't know that LifecycleNode was not inerhiting from Node... Maybe a better solution would be to expand these templates that accept "Node like" objects to also accept classes derived from Node.

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -149,13 +143,55 @@ class AsyncParametersClient
   ))
   {
     return rclcpp::create_subscription<rcl_interfaces::msg::ParameterEvent>(
-      node,
+      node_topics_interface,
```
ivanpauno: Consider not deleting `NodeT && node` as a template parameter and using here:
```cpp
get_node_topics_interface(node),
```

Also, delete the two other templates you added.
[get_node_topics_interface](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp) should do the magic.

ivanpauno:
> Ok, I didn't know that LifecycleNode was not inerhiting from Node... Maybe a better solution would be to expand these templates that accept "Node like" objects to also accept classes derived from Node.

It was failing when taking any "node like" pointer (it was working if you passed `*this` instead of `this`).
https://github.com/ros2/rclcpp/pull/821 will fix the issue.

I'm closing this one.

PR #811, opened 2019-07-31
Add callback group to AsyncParametersClient

bpwilcox:
Currently, the `AsyncParametersClient` doesn't support a `rclcpp::callback_group::CallbackGroup` other than the default (set as `nullptr`). This limits from some special use cases, such as using the `rclcpp::callback_group::CallbackGroup::Reentrant` with a multithreaded executor to enable parameter client requests/response handling while the node is spinning.

This PR adds the callback group as a member variable and constructor argument (with default `nullptr`).

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -53,19 +53,22 @@ public:
     const rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph_interface,
     const rclcpp::node_interfaces::NodeServicesInterface::SharedPtr node_services_interface,
     const std::string & remote_node_name = "",
-    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
+    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters,
+    rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr);
 
   RCLCPP_PUBLIC
   AsyncParametersClient(
     const rclcpp::Node::SharedPtr node,
     const std::string & remote_node_name = "",
-    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
+    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters,
+    rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr);
 
   RCLCPP_PUBLIC
   AsyncParametersClient(
     rclcpp::Node * node,
     const std::string & remote_node_name = "",
-    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters);
+    const rmw_qos_profile_t & qos_profile = rmw_qos_profile_parameters,
+    rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr);
 
   RCLCPP_PUBLIC
   std::shared_future<std::vector<rclcpp::Parameter>>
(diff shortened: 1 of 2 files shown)


rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -187,6 +190,7 @@ class AsyncParametersClient
   rclcpp::Client<rcl_interfaces::srv::DescribeParameters>::SharedPtr
     describe_parameters_client_;
   std::string remote_node_name_;
+  rclcpp::callback_group::CallbackGroup::SharedPtr callback_group_;
```
ivanpauno: Why adding this? There's no need of storing the shared pointer.

ivanpauno:
@wjwwood Was there any reason not to add this or was it simply missing?

rclcpp/include/rclcpp/parameter_client.hpp
bpwilcox: I believe storing the shared pointer is necessary because when added to the node base it is only added as a WeakPtr in the callback groups vector, see https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/node_base.hpp#L146

This had caused problems for us in adding the proper callback group until we noticed this.

ivanpauno: > I believe storing the shared pointer is necessary because when added to the node base it is only added as a WeakPtr in the callback groups vector, see https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/node_base.hpp#L146
> 
> This had caused problems for us in adding the proper callback group until we noticed this.

But in all the other entities that can be added to a callback group (like subscriptions), we aren't storing a shared pointer to it in them.
The lifetime of the callback group is controlled by the user (i.e.: the user should store a shared pointer to the callback group to maintain it valid), and it's not extended to the life of the entities that are added to the group.

bpwilcox: You have a point, if the user manages the shared pointer. I suppose it seemed a bit unclear when you `create_callback_group`. Perhaps the signature could be `create_callback_group(rclcpp::callback_group::CallbackGroup::SharedPtr)` as opposed to just passing the type, to make it clear that the user owns the object?

ivanpauno: mmm, I understand the confusion. Maybe the signature should be:
```cpp
create_callback_group(rclcpp::callback_group::CallbackGroup::SharedPtr &)
```
We're using the pattern of returning a `shared_ptr` and just storing a `weak_ptr` in many places. I would left that discussion for another issue/PR, and just add callback group support to `AsyncParametersClient` here.

Also, we're taking a `rclcpp::callback_group::CallbackGroup::SharedPtr group` in the constructors of `Subscription`, etc, when we could just take `rclcpp::callback_group::CallbackGroup & group`. It's clearer in the last case that `Subscription` is not taking part ownership of the callback group.
I would also don't include that change in this PR.

bpwilcox: I agree. I'll remove the callback group member variable here and perhaps submit another issue or PR for the signature changes.

nuclearsandwich:
Added wjwwood to increase the chances [the question above](https://github.com/ros2/rclcpp/pull/811#issuecomment-519948621) gets :eyes: when he returns. Feel free to clear the assignment after that.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7956/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3988/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6481/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7808/)

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7810/)

PR #816, opened 2019-08-05
Adapt to '--ros-args ... [--]'-based ROS args extraction

hidmic:
Precisely what the title says. Connected to https://github.com/ros2/rcl/pull/477.


hidmic:
So, I'm having second thoughts about [this](https://github.com/ros2/rclcpp/pull/816/files#diff-ad74ae7209fb5c36ae3a9bc53e695b47R98). I did it to spare the user having to feed `--ros-args` into `NodeOptions::arguments()` but it makes it somewhat asymmetric compared to arguments provided on `rclcpp::init()`.

wjwwood:
I don't feel strongly about this, if both work, i.e. user does or does not include `--ros-args`, then I think it's fine. Does `--ros-args --ros-args ....` cause any issues downstream?

hidmic:
> Does --ros-args --ros-args .... cause any issues downstream?

It depends, the second `--ros-args` instance will be picked up as an unknown ROS argument. But now that I think of it, it probably makes sense to simply ignore it.

hidmic:
A side effect of the current implementation though is that if the user feeds a `--` token, all arguments that follow won't be parsed as ROS specific arguments. I can't decide whether this is a bug or a feature though. Should we give the users the chance to store user-defined arguments in `NodeOptions()` instances?

wjwwood:
> A side effect of the current implementation is that if the user feeds a -- token, all arguments that follow won't be parsed as ROS specific arguments. I can't decide whether this is a bug or a feature though. Should we give the users the chance to store user-defined arguments in NodeOptions() instances?

I think we should leave it as-is, and in the documentation mentioned that if they user wants to have non-ROS arguments in the `NodeOptions.arguments` that they should then manually add `--ros-args` and `--`, putting their non-ROS args outside of that.

Which brings up another edge case, what if the user does this:

```
node_options.arguments = {
  "--some-non-ros-opt1",
  "--ros-args",
  "foo:=bar",
  "--",
  "--some-non-ros-opt2",
};
```

Will your current logic wrap the whole thing into `--ros-args`? If so, You're going to get unexpected results in that case.

hidmic:
> Will your current logic wrap the whole thing into --ros-args? If so, You're going to get unexpected results in that case.

Precisely. A leading `--ros-args` flag is always implicitly assumed, so `--some-non-ros-opt1` would be taken as a ROS argument. To get it to work as intended you'd need to do:

```
node_options.arguments = {
  "--",
  "--some-non-ros-opt1",
  "--ros-args",
  "foo:=bar",
  "--",
  "--some-non-ros-opt2",
};
```

We have to search for a `--ros-args` flag and only force one at the beginning if not found to get the "intuitive" behaviour.

wjwwood:
> We have to search for a --ros-args flag and only force one at the beginning if not found to get the "intuitive" behaviour.

That seems like a better way to do it for me, is there any reason not to do it this way?

hidmic:
Ok, fair enough. Added, plus a test.

rclcpp/src/rclcpp/node_options.cpp
```diff
@@ -93,11 +93,20 @@ NodeOptions::get_rcl_node_options() const
     int c_argc = 0;
     std::unique_ptr<const char *[]> c_argv;
     if (!this->arguments_.empty()) {
-      c_argc = static_cast<int>(this->arguments_.size()) + 1;
+      auto it = std::find(
+        this->arguments_.cbegin(), this->arguments_.cend(), RCL_ROS_ARGS_FLAG);
+
+      c_argc = static_cast<int>(this->arguments_.size());
+      if (it == this->arguments_.cend()) c_argc += 1;
```
wjwwood: nitpick: Please always use `{}` for if statements.

hidmic: See 8772bd0d73d2e9bf958e7f545bf3358ba226f61f.

wjwwood: Thanks!

PR #818, opened 2019-08-07
Adding target_compile_definitions to register_node macro

skucheria:


rclcpp_components/cmake/rclcpp_components_register_node.cmake
```diff
@@ -31,6 +31,9 @@ macro(rclcpp_components_register_node target)
   set(node ${ARGS_EXECUTABLE})
   _rclcpp_components_register_package_hook()
   set(_path "lib")
+  string(TOUPPER ${PROJECT_NAME} project_uppercase)
+  target_compile_definitions(${target}
+    PRIVATE "${project_uppercase}_BUILDING_DLL")
```
sloretz: I'm not sure this is the right place to put this. There's no guarantee that the `target` library has visibility macros, let alone that they use this name for the preprocessor definition to switch them. For example, a project with multiple libraries would need a different name for each one. If they didn't and one library depended on another, then symbols from the headers of the library it depends on would have the wrong visibility settings when the downstream library is built.

I would expect whoever created the library to setup this definition independent of the `rclcpp_components_register_node` macro.

PR #822, opened 2019-08-12
Add callbacks using references instead of smart pointers

rotu:
Smart pointers are inappropriate for these callbacks, as they imply the recipient is responsible for the lifespan of the object (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r30-take-smart-pointers-as-parameters-only-to-explicitly-express-lifetime-semantics). In most cases, the request / response objects are ephemeral and the caller should treat them as such.
Also, the constness of the pointers is prone to mistakes. For instance, a service callback can modify the request object since it is a `const shared_ptr<Request>`, not the `const shared_ptr<const Request>` that was probably intended.

### DIFF
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -42,9 +42,14 @@ private:
       const std::shared_ptr<typename ServiceT::Request>,
       std::shared_ptr<typename ServiceT::Response>
     )>;
+  using RefCallback = std::function<void (const ServiceT::Request &, ServiceT::Response &)>;
+  using RefWithRequestHeaderCallback = std::function<void (const rmw_request_id_t &,
+      const ServiceT::Request &, ServiceT::Response &)>;
 
   SharedPtrCallback shared_ptr_callback_;
   SharedPtrWithRequestHeaderCallback shared_ptr_with_request_header_callback_;
+  RefCallback ref_callback_;
+  RefWithRequestHeaderCallback ref_with_request_header_callback_;
 
 public:
   AnyServiceCallback()
@@ -81,6 +86,30 @@ public:
     shared_ptr_with_request_header_callback_ = callback;
   }
 
+  template<
+    typename CallbackT,
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<CallbackT, RefCallback>::value
+    >::type * = nullptr
+  >
+  template<typename CallbackT>
+  void set(CallbackT callback)
+  {
+    ref_callback_ = callback;
+  }
+
+  template<
+    typename CallbackT,
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<CallbackT, RefWithRequestHeaderCallback>::value
+    >::type * = nullptr
+  >
+  template<typename CallbackT>
+  void set(CallbackT callback)
+  {
+    ref_with_request_header_callback_ = callback;
+  }
+
   void dispatch(
     std::shared_ptr<rmw_request_id_t> request_header,
     std::shared_ptr<typename ServiceT::Request> request,
@@ -91,6 +120,10 @@ public:
       shared_ptr_callback_(request, response);
     } else if (shared_ptr_with_request_header_callback_ != nullptr) {
       shared_ptr_with_request_header_callback_(request_header, request, response);
+    } else if (ref_callback_ != nullptr) {
+      ref_callback_(*request, *response);
+    } else if (ref_with_request_header_callback_ != nullptr) {
+      ref_with_request_header_callback_(*request_header, *request, *response);
     } else {
       throw std::runtime_error("unexpected request without any callback set");
     }
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -48,6 +48,9 @@ class AnySubscriptionCallback
   using UniquePtrCallback = std::function<void (MessageUniquePtr)>;
   using UniquePtrWithInfoCallback =
     std::function<void (MessageUniquePtr, const rmw_message_info_t &)>;
+  using ConstRefCallback = std::function<void (const MessageT &)>;
+  using ConstRefWithInfoCallback =
+    std::function<void (const MessageT &, const rmw_message_info_t &)>;
 
   SharedPtrCallback shared_ptr_callback_;
   SharedPtrWithInfoCallback shared_ptr_with_info_callback_;
@@ -55,6 +58,8 @@ class AnySubscriptionCallback
   ConstSharedPtrWithInfoCallback const_shared_ptr_with_info_callback_;
   UniquePtrCallback unique_ptr_callback_;
   UniquePtrWithInfoCallback unique_ptr_with_info_callback_;
+  ConstRefCallback const_ref_callback_;
+  ConstRefWithInfoCallback const_ref_with_info_callback_;
 
 public:
   explicit AnySubscriptionCallback(std::shared_ptr<Alloc> allocator)
@@ -115,20 +120,6 @@ public:
     typename std::enable_if<
       rclcpp::function_traits::same_arguments<
         CallbackT,
-        ConstSharedPtrWithInfoCallback
-      >::value
-    >::type * = nullptr
-  >
-  void set(CallbackT callback)
-  {
-    const_shared_ptr_with_info_callback_ = callback;
-  }
-
-  template<
-    typename CallbackT,
-    typename std::enable_if<
-      rclcpp::function_traits::same_arguments<
-        CallbackT,
         UniquePtrCallback
       >::value
     >::type * = nullptr
@@ -152,6 +143,34 @@ public:
     unique_ptr_with_info_callback_ = callback;
   }
 
+  template<
+    typename CallbackT,
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<
+        CallbackT,
+        ConstRefCallback
+      >::value
+    >::type * = nullptr
+  >
+  void set(CallbackT callback)
+  {
+    const_ref_callback_ = callback;
+  }
+
+  template<
+    typename CallbackT,
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<
+        CallbackT,
+        const_ref_with_info_callback_
+      >::value
+    >::type * = nullptr
+  >
+  void set(CallbackT callback)
+  {
+    const_ref_with_info_callback_ = callback;
+  }
+
   void dispatch(
     std::shared_ptr<MessageT> message, const rmw_message_info_t & message_info)
   {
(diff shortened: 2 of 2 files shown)


rotu:
Thoughts?

Probably needs some test cases.

Also, not sure why guard code like this:
```
  template<
    typename CallbackT,
    typename std::enable_if<
      rclcpp::function_traits::same_arguments<
        CallbackT,
        SharedPtrCallback
      >::value
    >::type * = nullptr
  >
  void set(CallbackT callback)
  {
    shared_ptr_callback_ = callback;
  }
```

When it seems it could be more simply written as:
```
void set(decltype<shared_ptr_callback_> callback) { shared_ptr_callback_ = callback; }
```

ivanpauno:
In `rclcpp/include/rclcpp/any_subscription_callback.hpp`, `UniquePtrCallback` and `ConstSharedPtrCallback` are useful for intra process communication.
If you use the publish method taking a `unique_ptr` and subscribe with an `UniquePtrCallback`, the message is internally moved from the publish call to the subscription callback without doing copies.
If you use the same publish method and subscribe with many `ConstSharedPtrCallback`, the message is moved from the publish call, internally promoted to a shared ptr, and all the callbacks have part ownership of the same shared object (without a copy being made).
Lifetime of these pointers may be extended by the user.

For inter process comm, `unique_ptr` callback is still useful, the user may extend the lifetime of the object in this way (actually, if they use this callback signature, they should extend the lifetime).

`SharedPtrCallback` is not useful at all, and it's against the cpp core guidelines. I agree that it should be deprecated. If someone wants to extend the lifetime of the object and have part ownership, the unique ptr can be promoted to a shared ptr.

For the same reason, `SharedPtrCallback` of `AnyServiceCallback` should also be deprecated (there's not intra process comm for services).

----

I agree that we could have callbacks taking `const &`, but maybe, it's better to have callbacks taking a rvalue references (`&&`), which I think is the right approach. It's not useful when using intra process communication, but it's the best approach when using inter process communication and you don't want to extend the lifetime of the taken message.

rotu:
Thanks @ivanpauno for giving me a little more understanding. what happens if you have intra-process communication with one publisher and multiple subscribers? It doesn’t seem like `unique_ptr` can gracefully handle that.

ivanpauno:
> Thanks @ivanpauno for giving me a little more understanding. what happens if you have intra-process communication with one publisher and multiple subscribers? It doesn’t seem like `unique_ptr` can gracefully handle that.

In that case, all subscriptions receive a copy, except one that receives the original message.

ivanpauno:
> Smart pointers are inappropriate for these callbacks, as they imply the recipient is responsible for the lifespan of the object

Seeing what are we doing for handling messages:
https://github.com/ros2/rclcpp/blob/25f196989c02c092d6073c8383f0fbdfca6a4dcd/rclcpp/src/rclcpp/executor.cpp#L333-L346

I don't think there's much value of adding a callback taking a `const &` or `&&`.
The callback is responsible of manage the lifetime of the message, so the signatures taking an `unique_ptr` by value are ok.

I agree that `shared_ptr` callbacks should be deprecated.
Also, `borrow_message` and `return_message` of `MessageMemoryStrategy` is handling a `shared_ptr` which I think it's not correct. Only having a `create_message` that creates an `unique_ptr` would be ok. We are copying the message for handling `unique_ptr` callbacks, because we're first unnecessarily creating a `shared_ptr`:

https://github.com/ros2/rclcpp/blob/25f196989c02c092d6073c8383f0fbdfca6a4dcd/rclcpp/include/rclcpp/any_subscription_callback.hpp#L166-L174

ivanpauno:
@rotu are you planning with going ahead with this?

rotu:
> I don't think there's much value of adding a callback taking a const & or &&.
> The callback is responsible of manage the lifetime of the message, so the signatures taking an unique_ptr by value are ok.

I think the `const &` *is of value*. That's how I usually write callbacks to indicate that my code doesn't care about the object's lifecycle, which is *most* of my callbacks. It also leads to shorter lambda declarations, gets rid of an unnecessary layer of indirection, and has no possibility of the object being null.

If you need a long-lived reference to a large object, a shared_ptr-to-const is the only way to go. I don't like them as much as references but they do have the right semantics.

I don't understand the `unique_ptr` callback. It seems like you may as well just pass by value and enjoy a simpler function signature.

@ivanpauno, it seems you have strong opinions and maybe more expertise here, so I'm happy if you'd like to take this over.

ivanpauno:
> gets rid of an unnecessary layer of indirection

There's not performance difference between passing a pointer or passing a reference.
Moreover, there's not performance difference between `unique_ptr` and raw pointers (`unique_ptrs` are truly zero cost).
The only difference, is that in the case of passing a `const &`, we can create the object in the stack and pass it to the callback. When using `unique_ptr`, an allocation is needed (dynamic memory can be avoided by using a custom allocator).

> and has no possibility of the object being null.

Though that's true, the object passed to the callback is never null by design.

> I don't understand the `unique_ptr` callback. It seems like you may as well just pass by value and enjoy a simpler function signature.

Passing an `unique_ptr` is not the same than passing by value (it's "equivalent" to passing a reference).
 
> @ivanpauno, it seems you have strong opinions and maybe more expertise here, so I'm happy if you'd like to take this over.

I prefer not adding a new kind of callback if there is not specific use case where it behaves better, to avoid extra complexity to the user (deciding what callback to use). I can obviously be wrong.
I'm not planning to address the changes I [detailed before](https://github.com/ros2/rclcpp/pull/822#issuecomment-521673963) in a near future. I created an [issue](https://github.com/ros2/rclcpp/issues/826) for tracking the enhancement.

----

The signature `void callback (std::unique_ptr<Message>)` is clear for me. The unique ptr is passed by value, so the only owner is the callback. Extending the lifetime of the message or not is something optional, by default, the message will be destructed at the end of the callback.

rotu:
> There's not performance difference between passing a pointer or passing a reference.

Yes, but there's *cognitive* overhead. A shared pointer has swap, reset, etc. methods, may be null, has an extra layer of constness to think about, and you have to dereference it before doing anything with its contents. A reference means there's no wrapper object methods to even think about - just the properties of the object itself. Is that enough to warrant an additional method? Your call.

> Passing an unique_ptr is not the same than passing by value (it's "equivalent" to passing a reference).

It's equivalent to passing by a reference in some ways. The advantage of a `unique_ptr` (and the only justifiable reason to have it be a callback type here) is you know you own the underlying object, so you can mutate it willy-nilly. The code invoking the callback *needs to make a copy* of the message object for each such callback to avoid accidentally creating shared mutable state. So it may as well be pass-by-value (and if you really want to optimize, you can invoke the last callback with an `std::move(M)` to get rid of one copy).

And again, passing by value is cognitively simpler. It may not be faster code but it's one less layer of object to worry about and makes function signatures more readable.

Between the two of these, I think the second is more readable code.
```
    create_service<srv::AddTwoInts>(
      service_name_, [this](
        const std::shared_ptr<srv::AddTwoInts::Request> request,
        std::shared_ptr<srv::AddTwoInts::Response> response) {
        // ...
    });

    create_service<srv::AddTwoInts>(
      service_name_, [this](
        const srv::AddTwoInts::Request & request,
        srv::AddTwoInts::Response & response) {
        //...
      });
```

ivanpauno:
> Yes, but there's cognitive overhead. A shared pointer has swap, reset, etc. methods, may be null, has an extra layer of constness to think about, and you have to dereference it before doing anything with its contents. A reference means there's no wrapper object methods to even think about - just the properties of the object itself. Is that enough to warrant an additional method? Your call.

I was comparing the method taking an `unique_ptr`, `shared_ptr` has a performance overhead compared with raw pointers. I'm not convinced about the cognitive overhead, but if other people agree I think it's ok adding the signature taking a reference.

> It's equivalent to passing by a reference in some ways. The advantage of a unique_ptr (and the only justifiable reason to have it be a callback type here) is you know you own the underlying object, so you can mutate it willy-nilly.

We're doing a copy of the message that is not needed at all (just because we first unnecessary create a shared pointer). You can create the object, take the message, and move the unique_ptr to the callback without doing any copy (unique pointers can't be copied, so you have to use std::move when passing it to a function that takes the unique pointer by value).

----

When using inter process communication, all the callbacks can own its own message without the need of a copy, because they are receiving it separately.
When using intra process communication, if the callback wants to be the unique owner, a extra copy is needed.

tfoote:
> I'm not convinced about the cognitive overhead, but if other people agree I think it's ok adding the signature taking a reference.

I think that providing more prototypes will actually be more confusing overall. The new std smart pointers are relatively new for all of us. There's a level that we do have to learn to use them properly. But they do provide significant benefits with support for preallocated content, which is useful for pipelining large datatypes as well as realtime. And if we consistently use them they will become second nature, and in addition they will come to be used much more generally as well. If there are not strong use cases that cannot be implemented without adding another set of APIs I would strongly suggest that we not provide more options that require more ongoing maintenance and increase the documentation requirements, and provide opportunities for people to get confused about which type they're using when trying to put together potentially elements from different examples.

And for readabilty we provide automatically generated simpler aliases of the datatypes for reference `::SharedPtr` and `::ConstSharedPtr` for all generated messages, so your example would change to `srv::AddTwoInts::Request::SharedPtr` vs `srv::AddTwoInts::Request::ConstSharedPtr` for the const version.

rotu:
> I think that providing more prototypes will actually be more confusing overall

It's pretty hard to use already, since `create_subscription` and `create_service` are hard to read and reason about, and this needs to be simpler to understand and/or use. Looking at the code, there is no indication of what `CallbackT` can be, at at least a `function<void(MessageT)>` or `function<void(const MessageT &>` seem like the best guesses given this signature:

```
  /**
   * \param[in] topic_name The topic to subscribe on.
   * \param[in] callback The user-defined callback function to receive a message
   * \param[in] qos_history_depth The depth of the subscription's incoming message queue.
   * \param[in] options Additional options for the creation of the Subscription.
   * \param[in] msg_mem_strat The message memory strategy to use for allocating messages.
   * \return Shared pointer to the created subscription.
   */
  template<
    typename MessageT,
    typename CallbackT,
    typename AllocatorT = std::allocator<void>,
    typename SubscriptionT = rclcpp::Subscription<
      typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, AllocatorT>>
  create_subscription(
    const rclcpp::QoS & qos,
    CallbackT && callback,
    const SubscriptionOptionsWithAllocator<AllocatorT> & options =
    SubscriptionOptionsWithAllocator<AllocatorT>(),
    typename rclcpp::message_memory_strategy::MessageMemoryStrategy<
      typename rclcpp::subscription_traits::has_message_type<CallbackT>::type, AllocatorT
    >::SharedPtr
    msg_mem_strat = nullptr);
```

rotu:
I’m gonna think about how to make this better without making it more complicated. I’m not sure I have a good answer

rotu:
Still thinking about this and researching. It seems:

1. The object itself is shared between threads, so it should be const.
2. The shared_ptr may or may not be retained by the callee, so it should be passed by reference, as per https://stackoverflow.com/questions/3310737/should-we-pass-a-shared-ptr-by-reference-or-by-value

So the most correct API would be `shared_ptr<const T> const &` if the callback might or might not need the object to persist beyond the call.

rotu:
I’m working on a better implementation. I think I can make this code much better by using a template class instead of SFINAE guards. WIP:

https://gist.github.com/rotu/18b3fce4bc93cbf55d0738cde57c10b6

rotu:
@ivanpaono What is the expectation when there are multiple subscribers in a single process but not using intra-process communication? Should the recipients share the same message instance?

I assumed

PR #829, opened 2019-08-22
Add ParameterEventsSubscriber class

bpwilcox:
This PR adds a new class `ParameterEventsSubscriber` which provides function interfaces to handle parameter events.

The aim of this class is to address shortcomings and one of the unimplemented features (see https://design.ros2.org/articles/ros_parameters.html#unimplemented) for parameter events. Namely, this class:
- can register multiple custom callbacks by parameter name and node (either your own or remote)
- ~~register a parameter to automatically update a reference variable~~
- add a custom event callback by namespace (same behavior as `on_parameter_event()` in `AsyncParametersClient`) 
- can track different namespaces for parameter events (I believe the subscription created in the parameter client doesn't necessarily subscribe to parameter events on the namespace of the remote node passed in but of its own node)

In my opinion, this class would separate the parameter event subscriptions created in the parameter client (I'd remove the `on_parameter_event` in the future)


rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,157 @@
...
+  rclcpp::Parameter & parameter,
+  const std::string parameter_name,
+  const std::string node_name)
+{
+  if (event->node == resolve_path(node_name)) {
+    rclcpp::ParameterEventsFilter filter(event, {parameter_name},
+      {rclcpp::ParameterEventsFilter::EventType::NEW,
+        rclcpp::ParameterEventsFilter::EventType::CHANGED});
+    if (!filter.get_events().empty()) {
+      auto param_msg = filter.get_events()[0].second;
```
ivanpauno: I would prefer:
```cpp
const auto & events = filter.get_events();
auto param_msg = events[events.size()-1].second;
```

It sounds strange, but someone could call `set_parameters_atomically` repeating the same parameter more than once. In that case, you would like to get the last update, and not the first.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+void ParameterEventsSubscriber::register_parameter_callback(
+  const std::string & parameter_name,
+  std::function<void(const rclcpp::Parameter &)> callback,
+  const std::string & node_name)
+{
+  auto full_node_name = resolve_path(node_name);
+  add_namespace_event_subscriber(split_path(full_node_name).first);
+  parameter_node_map_[parameter_name] = full_node_name;
+  parameter_callbacks_[parameter_name] = callback;
```
ivanpauno: What happens if you want to register two different parameter callbacks for different nodes with same parameter name?
Wouldn't you be overriding the same callback?
If that's the case, I would just have one map: node name TO parameter name TO callback

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,157 @@
...
+      parameter_callbacks_[it->first](p);
+    }
+  }
+
+  if (user_callback_) {
+    user_callback_(event);
+  }
+}
+
+std::string ParameterEventsSubscriber::resolve_path(const std::string & path)
```
ivanpauno: I feel like some of the functionality in `resolve_path`, 'split_path`, `join_path` should be already provided in `rcl`.
We will probably need to do something similar in other places (if we're not already doing it).

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,159 @@
...
+  // Node Interfaces used for logging and creating subscribers
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_;
+  rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr node_topics_;
+  rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr node_logging_;
+
+  rclcpp::QoS qos_;
+
+  // Map containers for registered parameters
+  std::map<std::string, std::string> parameter_node_map_;
+  std::map<std::string, std::function<void(const rclcpp::Parameter &)>> parameter_callbacks_;
```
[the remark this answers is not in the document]
bpwilcox: Agreed

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
bpwilcox: That would be odd usage, but I agree we should take the last update

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
bpwilcox: Yes, you are correct. So we could do something like:
`parameter_callbacks_[{parameter_name, full_node_name}] = callback;`

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
bpwilcox: I see there is a split string functionality in rcutils: https://github.com/ros2/rcutils/blob/master/include/rcutils/split.h, and I also see there is `rclcpp::expand_topic_or_service_name`. The closest to `resolve_path` I see is `rclcpp::Node::create_effective_namespace` but that is not a public function.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
ivanpauno: Sounds good too

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
ivanpauno: And we also have https://github.com/ros2/rcpputils/blob/a5a973c2f3cf403af55e312d0054dc889995ba10/include/rcpputils/join.hpp.

If you can centralize some of this functions somewhere it would be good (some common string functionality in `rcutils` or `rcpputils`, and some specific functions for node names, etc in `rcl`).

But I think that's outside the scope of this PR, so you can left the methods here if you want and I will open an issue for tracking the enhancement.

bpwilcox: Neat, didn't know about the `join`. I could get rid of the `split_path` and `join_path` using those utils.

bpwilcox:
@ivanpauno Added a commit to address most of your feedback, including a test case for having the same parameter name on two separate nodes.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,160 @@
...
+  const std::string parameter_name,
+  const std::string node_name)
+{
+  if (event->node == resolve_path(node_name)) {
+    rclcpp::ParameterEventsFilter filter(event, {parameter_name},
+      {rclcpp::ParameterEventsFilter::EventType::NEW,
+        rclcpp::ParameterEventsFilter::EventType::CHANGED});
+    if (!filter.get_events().empty()) {
+      const auto & events = filter.get_events();
+      auto param_msg = events[events.size() - 1].second;
```
ivanpauno: nit: I forgot in the original review, but `std::vector` has a [back](https://en.cppreference.com/w/cpp/container/vector/back) method.

ivanpauno:
CI up to rclcpp test_rclcpp, only fastrtps:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8102/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4101/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6603/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7980/)

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/parameter_events_filter.hpp"
+
+namespace rclcpp
+{
+
+class ParameterEventsSubscriber
+{
+public:
+  ParameterEventsSubscriber(
```
ivanpauno: Add `RCLCPP_PUBLIC` macro here

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * If no namespace is provided, a subscription will be created for the current namespace
+   * Repeated calls to this function will overwrite the callback
+   * If more than one namespace already has a subscription to its parameter events topic, then the
+   * provided callback will be applied to all of them
+   *
+   * \param[in] callback Function callback to be evaluated on event
+   * \param[in] Name of namespace for which a subscription will be created
+   */
+  void set_event_callback(
```
ivanpauno: same

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+  /// adds a custom callback for a specified parameter
+  /**
+   * If a node_name is not provided, defaults to the current node
+   *
+   * \param[in] parameter_name Name of parameter
+   * \param[in] callback Function callback to be evaluated upon parameter event
+   * \param[in] node_name Name of node which hosts the parameter
+   */
+  void register_parameter_callback(
```
ivanpauno: same

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+  {
+    auto callback =
+      [&value, this](const rclcpp::Parameter & param) {
+        get_parameter_update<ParameterT>(param, value);
+      };
+
+    register_parameter_callback(parameter_name, callback, node_name);
+  }
+
+  bool get_parameter_from_event(
```
ivanpauno: same here

ivanpauno:
Also, check this warnings from the windows build:

```
10:17:38 C:\J\workspace\ci_windows\ws\src\ros2\rclcpp\rclcpp\test\test_parameter_events_subscriber.cpp(214): warning C4244: '=': conversion from 'int64_t' to 'int', possible loss of data [C:\J\workspace\ci_windows\ws\build\rclcpp\test_parameter_events_subscriber.vcxproj]
10:17:38 C:\J\workspace\ci_windows\ws\src\ros2\rclcpp\rclcpp\include\rclcpp/parameter_events_subscriber.hpp(116): warning C4244: '=': conversion from 'int64_t' to 'ParameterT', possible loss of data [C:\J\workspace\ci_windows\ws\build\rclcpp\test_parameter_events_subscriber.vcxproj]
10:17:38           with
10:17:38           [
10:17:38               ParameterT=int
10:17:38           ]
10:17:38   C:\J\workspace\ci_windows\ws\src\ros2\rclcpp\rclcpp\include\rclcpp/parameter_events_subscriber.hpp(92): note: see reference to function template instantiation 'void rclcpp::ParameterEventsSubscriber::get_parameter_update<ParameterT>(const rclcpp::Parameter &,ParameterT &)' being compiled
10:17:38           with
10:17:38           [
10:17:38               ParameterT=int
10:17:38           ]
10:17:38   C:\J\workspace\ci_windows\ws\src\ros2\rclcpp\rclcpp\test\test_parameter_events_subscriber.cpp(150): note: see reference to function template instantiation 'void rclcpp::ParameterEventsSubscriber::register_parameter_update<int>(const std::string &,ParameterT &,const std::string &)' being compiled
10:17:38           with
10:17:38           [
10:17:38               ParameterT=int
10:17:38           ]
```

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,160 @@
...
+  const rcl_interfaces::msg::ParameterEvent::SharedPtr event)
+{
+  std::string node_name = event->node;
+  RCLCPP_DEBUG(node_logging_->get_logger(), "Parameter event received for node: %s",
+    node_name.c_str());
+
+  std::unordered_map<
+    std::pair<std::string, std::string>,
+    std::function<void(const rclcpp::Parameter &)>,
+    stringPairHash>::iterator it;
```
wjwwood: `auto`?

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  for (it = parameter_callbacks_.begin(); it != parameter_callbacks_.end(); ++it) {
+    rclcpp::Parameter p;
+    if (get_parameter_from_event(event, p, it->first.first, it->first.second)) {
+      it->second(p);
+    }
+  }
+
+  if (user_callback_) {
```
wjwwood: What if this is set to `nullptr` with `set_event_callback()` after this line, but before the next?

rclcpp/test/test_parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,231 @@
+// Copyright (c) 2019 Intel Corporation
```
wjwwood: Can you follow the copyright statement style that we use? i.e. without the `(c)`:

https://github.com/ros2/rclcpp/blob/64c0f86f14dfe2337816079c4d5991268aeb0635/rclcpp/include/rclcpp/any_executable.hpp#L1

ivanpauno:
> First, I think the feature of updating the variable automatically is inherently unsafe. I could perhaps be convinced otherwise, but not without a new interface in place to protect the user.

I agree with @wjwwood  that is unsafe.
I originally thought that it could be useful in some single threaded scenarios, but after reading the last comments I agree that it doesn't add extra value compared with using `get_parameter` directly.

----

I also agree with the other thread-safety concerns that I originally missed.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
bpwilcox: I agree that `register` may be more ambiguous. Since in the current state we can only have one callback per param/node and per event, then it makes sense to use `set` for now

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+    const rcl_interfaces::msg::ParameterEvent::SharedPtr event,
+    rclcpp::Parameter & parameter,
+    const std::string parameter_name,
+    const std::string node_name = "");
```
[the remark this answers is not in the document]
bpwilcox: I think making it static is appropriate

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+   * \param[in] parameter_name Name of parameter
+   * \param[in] value Reference to variable receiving update
+   * \param[in] node_name Name of node which hosts the parameter
+   */
+  template<typename ParameterT>
+  void register_parameter_update(
+    const std::string & parameter_name, ParameterT & value, const std::string & node_name = "")
+  {
+    auto callback =
+      [&value, this](const rclcpp::Parameter & param) {
```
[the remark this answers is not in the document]
bpwilcox: Yes, a user could call `get_parameter` in a custom callback set via `register_parameter_callback`, but that would only apply for parameters existing on the node itself, not for remote nodes for which it has subscribed parameters. To access those parameters, it would need to create a `AsyncParameterClient` or `SyncParametersClient` to that node and call `get_parameter` to make a service call. I was purposely trying to move away from requiring the clients for this use-case. That could be all left to the user, I suppose, but I thought it convenient to provide the utility to update variables automatically.

Concerning thread-safety, I'd like to look into making these methods thread safe using mutex locks to address the concerns you've mentioned.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+  void get_parameter_update(
+    const rclcpp::Parameter & param, ParameterT & value)
+  {
+    try {
+      value = param.get_value<ParameterT>();
+      RCLCPP_DEBUG(node_logging_->get_logger(), "Updating parameter: %s", param.get_name().c_str());
+    } catch (...) {
+      RCLCPP_WARN(node_logging_->get_logger(),
+        "Parameter '%s' has different type (%s), cannot update registered parameter",
+        param.get_name().c_str(), param.get_type_name().c_str());
```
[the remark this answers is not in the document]
bpwilcox: I'd say if the user chooses to use this method `register_parameter_update` as opposed to `register_parameter_callback` with their own callback and error handling, then they are electing to accept this default error handling for updating value of said parameter.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,160 @@
...
+{}
+
+void ParameterEventsSubscriber::add_namespace_event_subscriber(const std::string & node_namespace)
+{
+  if (std::find(node_namespaces_.begin(), node_namespaces_.end(),
+    node_namespace) == node_namespaces_.end())
+  {
+    node_namespaces_.push_back(node_namespace);
+    auto topic = join_path(node_namespace, "parameter_events");
+    RCLCPP_DEBUG(node_logging_->get_logger(), "Subscribing to topic: %s", topic.c_str());
```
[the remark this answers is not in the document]
bpwilcox: The logger belongs to the node which subscribes to the namespace, so I'm unsure what you mean about using a sub-logger?

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  /// Adds a subscription (if unique) to a namespace parameter events topic
+  void add_namespace_event_subscriber(const std::string & node_namespace);
```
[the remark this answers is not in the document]
bpwilcox: Right now you can't. I suppose I could add a function like `remove_namespace_subscriber` and/or `remove_parameter_callback` (which could call `remove_namespace_subscriber` if there are no more parameters being listened to on that namespace).

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include <unordered_map>
+#include <vector>
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/parameter_events_filter.hpp"
+
+namespace rclcpp
+{
+
+class ParameterEventsSubscriber
```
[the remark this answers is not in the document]
bpwilcox: I agree, I was planning to provide additional documentation at some point. Would an example go in https://github.com/ros2/demos/tree/master/demo_nodes_cpp/src/parameters ?

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+  template<typename NodeT>
+  ParameterEventsSubscriber(
+    NodeT node,
+    const rclcpp::QoS & qos =
+    rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_parameter_events)))
+  : ParameterEventsSubscriber(
+      node->get_node_base_interface(),
+      node->get_node_topics_interface(),
+      node->get_node_logging_interface(),
```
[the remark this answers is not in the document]
bpwilcox: I'm not sure I understand, can you clarify what you mean?

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
ivanpauno: It's true that this adds the benefit of getting a parameter update from an external node without using the service.

The problem is that the reader of the variable that's being updated can be "external" to the class, so I don't see a way of using a mutex lock in a clear way here.
It may be clearer if the user directly writes its callback, using a `mutex` as needed.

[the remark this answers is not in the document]
wjwwood: You can have a callback that gives the user the parameter (extracted from the event rather than using get_parameter), and the user can decide how to get it out of the callback safely. I think this is the easiest thing for us to do.

Or we could have a proxy class where the user needs to access it via a method, like `get()` which can acquire a lock before getting the variable out as a copy (not a reference), protecting it from asynchronous updates.

bpwilcox:
> First, I think the feature of updating the variable automatically is inherently unsafe. I could perhaps be convinced otherwise, but not without a new interface in place to protect the user.
> 
I'd like to look into making this feature thread-safe as well as address thread-safety through the class

> Second, I think there needs to be more documentation about how to use this class. There are some not-so-obvious details about it for users, like the fact that you need to spinning on the Node for it to work.
> 
Yes, I'll be working on documentation. 

> Some what out-of-scope, but I think it applies to this classes future at least, is that when using this on a local node (if you have a reference rather than just a name), then you should be able to get updates without using the topic and without spinning. This is really needed for things like the TimeSource, see this issue:
> 
> #609
> 
I agree this would be preferable behavior, maybe a subsequent PR

> Also perhaps out-of-scope, but it would be a preferable end state, I think that this should support multiple callbacks not just one per node/parameter. For instance more than one party might care when `use_sim_time` changes. Right now they'd step on each other if they both tried to use the same instance of this class. So you could use two instances of this class, but then you'd need additional subscriptions which is wasteful. Therefore it would be better to support more than one callback at a time (at any granularity).
> 
Hmm, so perhaps saving a vector of callbacks per node/parameter, similar to what's being done in https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node.hpp#L962?

> Also, it is not currently possible with this class to get a set of parameters, which is a common use case. For example, "if any of P, I, or D change, give me all of them" kind of thing. A stretch goal for sure, but it would require either changing this classes interface or adding nearly identical functionality along side of it in the future. Seems like it would be better to consider this case from the beginning.

This is interesting. I suppose you could add a custom callback on each parameter and provide your own inspection into the other parameters, but it sounds like there could be a `set_parameter_group_callback` type function that accepts multiple parameter names and has a callback associated with that grouping if any change? You might have to create a client to remote nodes unless you're caching their values somewhere, though.

bpwilcox:
@wjwwood I've addressed much of your feedback in my latest commits. Namely:
- adding mutexes for thread safety
- removed `register_parameter_update` feature
  - given our conversation and further investigation, I decided it is best for now to just remove
- made `get_parameter_from_event` a static method
- changed naming to from `register_parameter_callback` to `add_parameter_callback`
- added feature to add multiple callbacks per {parameter, node}
- added functions to remove parameter callbacks (either by callback handle or entire parameter) and removal of event callback
- internally, subscriptions will be erased if there are no more callbacks associated with the `<namespace>/parameter_events` topic
- updated tests to cover added features

Can you review the changes and let me know if there are any remaining blocking issues for this PR?

mkhansen-intel:
@wjwwood - can you review again please?

wjwwood:
I might be able to get to it this week. It's already on my list, but there are other big pull requests that are older (e.g. https://github.com/ros2/rclcpp/pull/778) which I am working on now.

mkhansen-intel:
Thanks for the update!

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+  event_callback_ = callback;
+}
+
+void
+ParameterEventsSubscriber::remove_event_callback()
+{
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  // Clear current vector of event namespaces and remove subscriptions
+  auto temp_namespaces = event_namespaces_;
```
ivanpauno: nit: Using the move constructor might be better.
Prefer not using `auto`, except the type is obvious or too complicated to write.
```cpp
  std::vector<std::string> temp_namesaces(std::move(event_namespaces_));
```

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,247 @@
...
+      hash_combine(seed, s.first);
+      hash_combine(seed, s.second);
+      return seed;
+    }
+  };
+  // *INDENT-ON*
+
+  // Map container for registered parameters.
+  std::unordered_map<
+    std::pair<std::string, std::string>,
```
ivanpauno: Maybe using a struct instead of a pair. It's more self-documenting.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,308 @@
...
+}
+
+rclcpp::Parameter
+ParameterEventsSubscriber::get_parameter_from_event(
+  const rcl_interfaces::msg::ParameterEvent::SharedPtr event,
+  const std::string parameter_name,
+  const std::string node_name)
+{
+  rclcpp::Parameter p(parameter_name);
+  get_parameter_from_event(event, p, parameter_name, node_name);
```
ivanpauno: Maybe, it's better to  throw when you can't get the parameter from the event ...

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,308 @@
...
+          shared_handle->callback(p);
+        } else {
+          cb = it->second.erase(cb);
+        }
+      }
+    }
+  }
+
+  if (event_callback_) {
+    event_callback_(event);
```
ivanpauno: Shouldn't call this one only when the namespace is in `event_namespaces_;`?

wjwwood:
> @wjwwood Why do we have a parameter event topic per namespace? Was that intentional?
I think it's ok, but also having only one is reasonable.

Can you link to the line where this is done (where the topic name is specified)?

I believe this was discussed somewhere else, but I can't find it. Basically, if the topic is currently `parameter_events`, then I agree it's not right and should be changed. I think either `/parameter_events` or `~/parameter_events` makes sense, but not `parameter_events`. I don't know what the hangup was on changing that, if there was any. Personally, I'd go for `/parameter_events`, because I judge that filtering undesired events is more efficient then having many subscriptions (one for each node) when you want to monitor all changes. Plus we could consider using a keyed topic (keyed on node name most likely) in the future to gain some performance back.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
bpwilcox: I agree that makes more sense. 

Actually, I had a couple more thoughts I'm considering:
1) Apply the event callbacks per node as opposed to namespace
2) Support multiple event callbacks, either multiple on a same node, or multiple different nodes. This is similar to what I've done for the individual parameter callbacks

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
bpwilcox: Hmm, that may be better than requiring the user to check themselves

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
bpwilcox: I don't have any strong position on this, the details are internal in either case.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,308 @@
...
+ParameterEventsSubscriber::set_event_callback(
+  std::function<void(const rcl_interfaces::msg::ParameterEvent::SharedPtr &)> callback,
+  const std::vector<std::string> & node_namespaces)
+{
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  remove_event_callback();
+  std::string full_namespace;
+  for (auto & ns : node_namespaces) {
+    if (ns == "") {
+      full_namespace = node_base_->get_namespace();
```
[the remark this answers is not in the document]
bpwilcox: Would it be better to create a function overload then that doesn't take any namespaces but only applies callback to itself? 

As mentioned in another comment, I am thinking to set the callbacks on the nodes as opposed to the namespace.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,247 @@
...
+  should_unsubscribe_to_namespace(const std::string & node_namespace);
+
+  /// Callback for parameter events subscriptions.
+  void
+  event_callback(const rcl_interfaces::msg::ParameterEvent::SharedPtr event);
+
+  // Utility functions for string and path name operations.
+  std::string resolve_path(const std::string & path);
+  std::pair<std::string, std::string> split_path(const std::string & str);
+  std::string join_path(std::string path, std::string name);
```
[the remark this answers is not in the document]
bpwilcox: I will see if I can use these.

ivanpauno:
> Can you link to the line where this is done (where the topic name is specified)?

https://github.com/ros2/rclcpp/blob/59e7bbbe7cd4295485dc3dab443cde3a4624debd/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L72

> I believe this was discussed somewhere else, but I can't find it.

I remember an offline discussion, but no idea is something was written about it.

> I'd go for /parameter_events, because I judge that filtering undesired events is more efficient then having many subscriptions

I agree. I'm going to open a PR proposing that.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
bpwilcox: Though, if we remove the namespace from the topic name as your PR proposes, then I no longer need to worry about splitting/joining strings.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,247 @@
...
+   * \param[in] event Event msg to be inspected.
+   * \param[out] parameter Reference to rclcpp::Parameter to be assigned.
+   * \param[in] parameter_name Name of parameter.
+   * \param[in] node_name Name of node which hosts the parameter.
+   * \returns true if requested parameter name & node is in event, false otherwise
+   */
+  RCLCPP_PUBLIC
+  static bool
+  get_parameter_from_event(
+    const rcl_interfaces::msg::ParameterEvent::SharedPtr event,
```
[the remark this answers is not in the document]
bpwilcox: The `ParameterEventsFilter` used in this function requires a shared ptr to the event msg,
https://github.com/ros2/rclcpp/blob/59e7bbbe7cd4295485dc3dab443cde3a4624debd/rclcpp/include/rclcpp/parameter_events_filter.hpp#L63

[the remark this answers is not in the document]
ivanpauno: I really don't like the `ParameterEventsFilter` class design.
The problem could be solved with a free function, without the need of taking a `shared_ptr` to an event.
But that discussion is probably out of the scope of this PR.

ivanpauno:
@bpwilcox Can you update the PR after https://github.com/ros2/rclcpp/pull/929?
I think it will simplify the logic a bit here.

bpwilcox:
@ivanpauno I believe I've addressed your comments and updated to use the absolute parameter events topic. I've also added support for multiple event callbacks.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,253 @@
...
+  // *INDENT-OFF* Uncrustify doesn't handle indented public/private labels
+  // Hash function for string pair required in std::unordered_map
+  class StringPairHash
+  {
+  public:
+    template<typename T>
+    inline void hash_combine(std::size_t & seed, const T & v) const
+    {
+      std::hash<T> hasher;
+      seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
ivanpauno: nit: Add a comment, referencing from where this comes.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,253 @@
...
+
+  // Map container for registered parameters.
+  std::unordered_map<
+    std::pair<std::string, std::string>,
+    CallbacksContainerType,
+    StringPairHash
+  > parameter_callbacks_;
+
+  template<typename CallbackHandleT>
+  struct HandleCompare
```
ivanpauno: Is this needed in the header?
If not, move it to the source file.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,206 @@
...
+{
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  auto it = std::find_if(
+    event_callbacks_.begin(),
+    event_callbacks_.end(),
+    HandleCompare<ParameterEventCallbackHandle>(handle));
+  if (it != event_callbacks_.end()) {
+    event_callbacks_.erase(it);
+  } else {
+    throw std::runtime_error("Callback doesn't exist");
```
ivanpauno: nit: Add a test case, checking it actually throws.

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,206 @@
...
+    container.begin(),
+    container.end(),
+    HandleCompare<ParameterCallbackHandle>(handle));
+  if (it != container.end()) {
+    container.erase(it);
+    if (container.empty()) {
+      parameter_callbacks_.erase({handle->parameter_name, handle->node_name});
+    }
+  } else {
+    throw std::runtime_error("Callback doesn't exist");
```
ivanpauno: same

rclcpp/src/rclcpp/parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+rclcpp::Parameter
+ParameterEventsSubscriber::get_parameter_from_event(
+  const rcl_interfaces::msg::ParameterEvent::SharedPtr event,
+  const std::string parameter_name,
+  const std::string node_name)
+{
+  rclcpp::Parameter p;
+  if (!get_parameter_from_event(event, p, parameter_name, node_name)) {
+    throw std::runtime_error(
```
ivanpauno: same

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
wjwwood: I think this comment still applies, you could either have a small example in the doc block here, or you could add an example to either the `demo_nodes_cpp` package or this repository:

https://github.com/ros2/examples/tree/master/rclcpp

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
wjwwood: This still applies, as @ivanpauno said, you need the equivalent of these files:

- https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/get_node_base_interface.hpp
- https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp

for the logging interface. It's essentially copy-paste-replace "base" or "topics" with "logging". And then use those here.

For example it is used here:

https://github.com/ros2/rclcpp/blob/81049c42c0b04d0cc80a3e4c3a0819a735e23cdc/rclcpp/include/rclcpp/create_publisher.hpp#L52-L54

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,183 @@
...
+    const rclcpp::QoS & qos =
+    rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_parameter_events)));
+
+  template<typename NodeT>
+  ParameterEventsSubscriber(
+    NodeT node,
+    const rclcpp::QoS & qos =
+    rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_parameter_events)))
+  : ParameterEventsSubscriber(
+      node->get_node_base_interface(),
```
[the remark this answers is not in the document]
wjwwood: This comment still applies.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
wjwwood: I think this is a valid critique of the existing filter class and would be worth changing now so that we don't have to compromise this new API in order to use the existing filter class API...

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
wjwwood: I think this should be resolved before merge too.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
wjwwood: 👍

rclcpp/test/test_parameter_events_subscriber.cpp
wjwwood: This seems to still be using `(c)`...

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
bpwilcox: I've removed the `ParameterEventsFilter` dependency from the `get_parameter_from_event` function and modified code from the filter to work for this use case.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
bpwilcox: Added a `get_node_logging_interface.hpp`

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
bpwilcox: I think it would be more clear in either `examples` or `demo_nodes_cpp`. I can make a separate PR in one of those repos after this one?

bpwilcox:
@wjwwood @ivanpauno  I've addressed your latest feedback.

fujitatomoya:
> Also, it is not currently possible with this class to get a set of parameters, which is a common use case. For example, "if any of P, I, or D change, give me all of them" kind of thing. A stretch goal for sure, but it would require either changing this classes interface or adding nearly identical functionality along side of it in the future. Seems like it would be better to consider this case from the beginning.

i do agree on this.
off the top of my head, this could be resolved by DDS ContentFilteredTopics, this means that parameter_events reader can register the parameter contents only interested. this is the efficient way to reduce network resource and cpu consumption, if filtering is done on writer side.

wjwwood:
I don't think ContentFilteredTopics has anything to do with my concern you quoted. The class already has to send notifications when one a few different parameter changes, you just cannot have it deliver them as a group to you.

ContentFilteredTopics could be used in the implementation of this class, but it wouldn't affect the API the user sees in the class.

bpwilcox:
This might be a separate issue/PR, but I wanted to field thoughts @wjwwood @ivanpauno on a matching interface on the `node_parameters_interface` to add event and parameter callbacks using this class, similar to the current `on_set_parameters_callback` for validation.

ivanpauno:
> This might be a separate issue/PR, but I wanted to field thoughts @wjwwood @ivanpauno on a matching interface on the `node_parameters_interface` to add event and parameter callbacks using this class, similar to the current `on_set_parameters_callback` for validation.

Yeah, I think that would be great, to avoid creating a subscription for local nodes.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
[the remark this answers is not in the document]
wjwwood: Even if you put something in `examples` or `demo_nodes_cpp` this class needs documentation.

ros-discourse:

mjeronimo:
@ivanpauno @wjwwood I've updated Brian's work for the latest ROS2 code and would appreciate another review. 

Also notifying @bpwilcox @jacobperron 

Scanning the comment thread here, I see the following issues:

* **Automatic variable update unsafe** -- This feature was removed
* **Better documentation required** -- A demo and tutorial created

    The demo is here:

    https://github.com/ros2/demos/blob/mjeronimo/parameter-events-demo-cpp/demo_nodes_cpp/src/parameters/parameter_events_subscriber.cpp

    and a tutorial here:

    https://github.com/ros2/ros2_documentation/blob/mjeronimo/add-monitor-parameter-changes-cpp/source/Tutorials/Monitoring-For-Parameter-Changes-CPP.rst

* **Allow multiple callbacks per node/parameter** -- The code has been updated to allow multiple callbacks on the same parameter

There were also a couple items suggested for future improvements:

* **Provide parameter updates without using parameter topic and spinning** -- Could do this if the benefits are worth the change
* **Monitor a set of parameters** -- We need to clarify the use case, but this seems like something we could add fairly easily

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+{
+  RCLCPP_SMART_PTR_DEFINITIONS(ParameterEventCallbackHandle)
+
+  using ParameterEventCallbackType =
+    std::function<void (const rcl_interfaces::msg::ParameterEvent::SharedPtr &)>;
+
+  ParameterEventCallbackType callback;
+};
+
+class ParameterEventsSubscriber
```
wjwwood, line 61: I wonder about the name of this class, maybe it should just be `ParameterEventMonitor` or `ParameterEventNotifier` or `ParameterEventHandler` or something, since the fact that it subscribes to accomplish this is kind of an implementation detail.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  using ParameterEventCallbackType =
+    ParameterEventCallbackHandle::ParameterEventCallbackType;
+
+  /// Set a custom callback for parameter events.
+  /**
+   * Multiple callbacks are allowed.
+   *
+   * \param[in] callback Function callback to be evaluated on event.
```
wjwwood, line 87: We should document the return type as well (just to have doc coverage).

wjwwood: I'm not going to comment on each case, but in general we should have full documentation for all the classes, methods, and functions. Including their parameters, return values, and possible exceptions at the very least.

rclcpp/test/rclcpp/test_parameter_events_subscriber.cpp
```diff
@@ -0,0 +1,348 @@
...
+protected:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+
+  void SetUp()
+  {
+    rclcpp::NodeOptions options;
+    options.allow_undeclared_parameters(true);
```
wjwwood, line 45: Is this necessary? I looked through the code looking for `get/set_parameter` like functions being called on the node and I didn't see any.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+    event_subscription_ = rclcpp::create_subscription<rcl_interfaces::msg::ParameterEvent>(
+      node_topics, "/parameter_events", qos,
+      std::bind(&ParameterEventsSubscriber::event_callback, this, std::placeholders::_1));
+  }
+
+  using ParameterEventCallbackType =
+    ParameterEventCallbackHandle::ParameterEventCallbackType;
+
+  /// Set a custom callback for parameter events.
```
wjwwood, line 83:
```suggestion
  /// Add a custom callback for parameter events.
```

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * Multiple callbacks are allowed.
```
wjwwood, line 85:
```suggestion
   * This may be called multiple times.
```

The previous wording sounded like you could provide multiple callbacks at the same time (i.e. in one call), at least to me.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+    ParameterEventCallbackType callback);
+
+  /// Remove parameter event callback.
+  /**
+   * Calling this function will set the event callback to nullptr.
+   */
+  RCLCPP_PUBLIC
+  void
+  remove_parameter_event_callback(
+    const ParameterEventCallbackHandle * const handle);
```
wjwwood, line 101: It's kind of odd that `add_parameter_event_callback()` returns a shared pointer but this takes a raw pointer.

I think this should either take a shared pointer (maybe as a const reference), or maybe a const reference to the handle. Is the address of the handle important or are the contents sufficient?

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+   * \param[in] callback Function callback to be evaluated on event.
+   */
+  RCLCPP_PUBLIC
+  ParameterEventCallbackHandle::SharedPtr
+  add_parameter_event_callback(
+    ParameterEventCallbackType callback);
+
+  /// Remove parameter event callback.
+  /**
+   * Calling this function will set the event callback to nullptr.
```
wjwwood, line 96: This sentence isn't clear to me, what even callback? Why is it getting set to nullptr rather than being removed?

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+   * The parameter name and node name are inspected from the callback handle. The callback handle
+   * is erased from the list of callback handles on the {parameter_name, node_name} in the map.
+   * An error is thrown if the handle does not exist and/or was already removed.
+   *
+   * \param[in] handle Pointer to callback handle to be removed.
+   */
+  RCLCPP_PUBLIC
+  void
+  remove_parameter_callback(
+    const ParameterCallbackHandle * const handle);
```
wjwwood, line 131: Same comment as for `remove_parameter_event_callback()`, does this need to be a raw pointer here? or instead does the add function need to return a shared pointer?

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * \param[in] event Event msg to be inspected.
+   * \returns A vector rclcpp::Parameter values from the event
+   */
+  RCLCPP_PUBLIC
+  static std::vector<rclcpp::Parameter>
+  get_parameters_from_event(
+    const rcl_interfaces::msg::ParameterEvent & event);
+
+  using CallbacksContainerType = std::list<ParameterCallbackHandle::WeakPtr>;
```
wjwwood, line 181: Does this need to be public? Seems like it should be below where it is used (only place I can see), in the protected section.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
mjeronimo: Changed to ParameterEventHandler.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
mjeronimo: I went through all of the functions and updated the doc comments, including return types.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
mjeronimo: Updated to "This function may be called multiple times to set multiple parameter event callbacks."

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
mjeronimo: Yes, I agree: it seems like implementation details. I've removed that sentence and have updated the main description: "Remove parameter event callback registered with add_parameter_event_callback", which should be sufficient.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
mjeronimo: It's a convenience function for users of the class and is helpful when doing your own filtering. For example,
```
  // We can also monitor all parameter changes and do our own filtering/searching
  auto cb3 =
    [fqn, remote_param_name, &node](const rcl_interfaces::msg::ParameterEvent::SharedPtr & event) {
      // Use a regular expression to scan for any updates to parameters in "/a_namespace"
      // as well as any parameter changes to our own node
      std::regex re("(/a_namespace/.*)|(/this_node)");
      if (regex_match(event->node, re)) {
        // You can use 'get_parameter*s*_from_event' to enumerate all changes that came
        // in on this event
        auto params = rclcpp::ParameterEventsSubscriber::get_parameters_from_event(*event);
        for (auto & p : params) {
          RCLCPP_INFO(
            node->get_logger(), "cb3: Received an update to parameter \"%s\" of type: %s: \"%s\"",
            p.get_name().c_str(),
            p.get_type_name().c_str(),
            p.value_to_string().c_str());
        }
      }
    };
```

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
mjeronimo: I believe @bpwilcox  was following suit with a similar usage in the rclcpp::Node class:

```
  // node.hpp
  RCLCPP_PUBLIC
  void
  remove_on_set_parameters_callback(const OnSetParametersCallbackHandle * const handler);
```

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
mjeronimo: Done in same style as node.hpp.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
ivanpauno: I think that the original method in rclcpp::Node used that signature because:
- you don't need shared ownership, so taking the shared pointer by value isn't needed (cppcoreguidelines [R34](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-owner)).
- you don't need to "maybe take ownership", so taking a const reference to the shared pointer also isn't needed ([R36](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-const)).

We're using a raw pointer to identify the handle, so we only need a raw pointer to remove it from the map.

rclcpp/include/rclcpp/parameter_event_handler.hpp
```diff
@@ -0,0 +1,241 @@
...
+{
+  RCLCPP_SMART_PTR_DEFINITIONS(ParameterEventCallbackHandle)
+
+  using ParameterEventCallbackType =
+    std::function<void (const rcl_interfaces::msg::ParameterEvent::SharedPtr &)>;
+
+  ParameterEventCallbackType callback;
+};
+
+class ParameterEventHandler
```
wjwwood, line 61: So, I think this comment stilllll applies :)

https://github.com/ros2/rclcpp/pull/829#discussion_r362639274

https://github.com/ros2/rclcpp/pull/829#discussion_r537996420

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
wjwwood: But it's not documented that letting the shared pointer returned by `add_parameter_event_callback` go out of scope will automatically remove it, unlike the inspiration for these signatures, `Node::add_on_set_parameters_callback()`. As far as I can tell from the code, it is not the case that letting this shared pointer go out of scope will result in the callback getting removed. So I don't see any reason for it to be returned as a shared_ptr.

So going back to my original thing, I think the add function should return a raw pointer (as there's no need to have shared ownership) or the remove function should take a shared pointer so it is symmetric with the add function.

Since automatic removal is not happening, it sounds like it should never be a shared pointer in the first place.

And since the handle is trivially copied, maybe it should just return a `OnSetParametersCallbackHandle` from add and take a `const OnSetParametersCallbackHandle &` in remove?

I still just don't see why we're returning a shared ptr from add but taking a raw pointer here.

wjwwood (commented):
There are some lingering questions, but this is really close.

rclcpp/include/rclcpp/parameter_events_subscriber.hpp
ivanpauno: > Since automatic removal is not happening, it sounds like it should never be a shared pointer in the first place.

That sounds good to me.
I think that the inconsistency with the "parameters acceptation callbacks" is a bit undesired, but it sounds ok.

i.e.: I would either always have scoped handles or unscoped ones.

rclcpp/include/rclcpp/parameter_event_handler.hpp
```diff
@@ -0,0 +1,241 @@
...
+   * \param[in] qos The QoS settings to use for any subscriptions.
+   */
+  template<typename NodeT>
+  ParameterEventHandler(
+    NodeT node,
+    const rclcpp::QoS & qos =
+    rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_parameter_events)))
+  {
+    node_base_ = rclcpp::node_interfaces::get_node_base_interface(node);
+    node_logging_ = rclcpp::node_interfaces::get_node_logging_interface(node);
```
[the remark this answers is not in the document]
mjeronimo: Removed.

rclcpp/src/rclcpp/parameter_event_handler.cpp
```diff
@@ -0,0 +1,222 @@
...
+{
+
+ParameterEventCallbackHandle::SharedPtr
+ParameterEventHandler::add_parameter_event_callback(
+  ParameterEventCallbackType callback)
+{
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
+  auto handle = std::make_shared<ParameterEventCallbackHandle>();
+  handle->callback = callback;
+  event_callbacks_.emplace_front(handle);
```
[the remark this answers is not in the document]
mjeronimo: Done.

mjeronimo:
@fujitatomoya  @wjwwood  Any further input?

rclcpp/include/rclcpp/parameter_event_handler.hpp
```diff
@@ -0,0 +1,330 @@
...
+ *
+ * To remove a parameter callback, call remove_parameter_callback, passing the handle returned
+ * from add_parameter_callback:
+ *
+ *   param_handler->remove_parameter_callback(handle2);
+ *
+ * You can also monitor for *all* parameter changes, using add_parameter_event_callback. In this
+ * case, the callback will be invoked whenever any parameter changes in the system. You are likely
+ * interested in a subset of these parameter changes, so in the callback it is convenient to use
+ * a regular expression on the node names or namespaces of interest. For example,
```
wjwwood, line 110:
```suggestion
 * a regular expression on the node names or namespaces of interest. For example:
```

rclcpp/include/rclcpp/parameter_event_handler.hpp
```diff
@@ -0,0 +1,330 @@
...
+};
+
+/// A class used to "handle" (monitor and respond to) changes to parameters.
+/**
+ * The ParameterEventHandler class allows for the monitoring of changes to node parameters,
+ * either a node's own parameters or parameters owned by other nodes in the system. Multiple
+ * parameter callbacks can be set and will be invoked when the specified parameter changes.
+ *
+ * The first step is to instance a ParameterEventHandler, providing a ROS node to use
+ * to create any required subscriptions.
```
wjwwood, line 66:
```suggestion
 * to create any required subscriptions:
```

rclcpp/include/rclcpp/parameter_event_handler.hpp
```diff
@@ -0,0 +1,330 @@
...
+ *             p.value_to_string().c_str());
+ *         }
+ *       }
+ *     };
+ *   auto handle3 = param_handler->add_parameter_event_callback(cb3);
+ *
+ * For both parameter callbacks and parameter event callbacks, when multiple callbacks are added,
+ * the callbacks are invoked last-in, first-called order (LIFO).
+ *
+ * To remove a parameter event callback, use
```
wjwwood, line 148:
```suggestion
 * To remove a parameter event callback, use:
```

rclcpp/src/rclcpp/parameter_event_handler.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+
+  for (auto event_cb = event_callbacks_.begin(); event_cb != event_callbacks_.end(); ++event_cb) {
+    auto shared_event_handle = event_cb->lock();
+    if (nullptr != shared_event_handle) {
+      shared_event_handle->callback(event);
+    } else {
+      event_cb = event_callbacks_.erase(event_cb);
+    }
```
wjwwood, line 199: Did the documentation mention that the handle was scoped?

wjwwood (approved):
Mostly a bunch of nitpicks about the documentation.

I think this is ready to merge regardless. So take or leave the suggestions and lgtm 👍

nuclearsandwich:
Merging this introduced build failures on Windows [Build Status](https://ci.ros2.org/job/ci_windows/13880/). We reverted the change in #1572 and there is a new PR in #1573 to re-apply them.

rclcpp/include/rclcpp/parameter_event_handler.hpp
```diff
@@ -0,0 +1,336 @@
...
+    std::pair<std::string, std::string>,
+    CallbacksContainerType,
+    StringPairHash
+  > parameter_callbacks_;
+
+  rclcpp::Subscription<rcl_interfaces::msg::ParameterEvent>::SharedPtr event_subscription_;
+
+  std::list<ParameterEventCallbackHandle::WeakPtr> event_callbacks_;
+
+  std::recursive_mutex mutex_;
```
wjwwood, line 331: @bpwilcox we're investigating some strange behavior and this class is involved, but during that investigation I noticed this mutex is recursive, but it wasn't immediately obvious to me why that's required. The callback is used in a subscription which is used in a mutually exclusive callback group (the default) and so it will never be called in parallel with itself. Likewise I don't see why any of the other functions (add/remove ones) use the mutex in a way where recursion is needed.

I know this is an ancient issue, but it was just a shot in the dark if you remembered why that was the case.

PR #837, opened 2019-08-29
add mutex in add/remove_node and wait_for_work to protect concurrent use/change of memory_strategy_

dirk-thomas:
Fixes race condition reported in https://github.com/ros-planning/navigation2/issues/1079#issuecomment-526230684

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8037/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4044/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6544/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7892/) (the other test failures are already present on master)

The `cpplint` warning has been addressed, see the retriggered Linux build for `rclcpp`: [Build Status](http://ci.ros2.org/job/ci_linux/8038/)

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -22,6 +22,7 @@
 #include <iostream>
 #include <list>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
The diff ignoring whitespace changes is probably easier to read: https://github.com/ros2/rclcpp/pull/837/files?w=1

mm318:
When we tried back porting this to "dashing" branch of `rclcpp`, we are getting lots of
```
[gzserver-1] terminate called after throwing an instance of 'eprosima::fastcdr::exception::NotEnoughMemoryException'
[gzserver-1]   what():  Not enough memory in the buffer stream
```
Any idea why this might be, @dirk-thomas?

dirk-thomas:
@mm318 Sorry, I don't have an idea how this could be triggered.

ivanpauno:
> When we tried back porting this to "dashing" branch of rclcpp, we are getting lots of

That error doesn't seem to be related with this PR. IIUC, `fastcdr` is only used for serializing messages.

PR #842, opened 2019-09-03
Fail on invalid and unknown ROS specific arguments

hidmic:
Precisely what the title says. This pull request makes `rclcpp` throw if either an invalid or an unknown ROS argument is found.

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -15,6 +15,7 @@
 #ifndef RCLCPP__UTILITIES_HPP_
 #define RCLCPP__UTILITIES_HPP_
 
+
 #include <chrono>
 #include <functional>
 #include <limits>
@@ -26,7 +27,6 @@
 #include "rclcpp/visibility_control.hpp"
 
 #ifdef ANDROID
-#include <sstream>
 
 namespace std
 {
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -15,7 +15,6 @@
 #ifndef RCLCPP__UTILITIES_HPP_
 #define RCLCPP__UTILITIES_HPP_
 
-
 #include <chrono>
 #include <functional>
 #include <limits>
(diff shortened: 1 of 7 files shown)


rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -282,6 +284,27 @@ sub_will_underflow(const T x, const T y)
   return (y > 0) && (x < (std::numeric_limits<T>::min() + y));
 }
 
+/// Join values in a container turned into strings by a given delimiter
```
ivanpauno: I would rather add this in `rcpputils`. We should also move some stuff here to that package.

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -15,6 +15,7 @@
 #ifndef RCLCPP__UTILITIES_HPP_
 #define RCLCPP__UTILITIES_HPP_
 
+
```
ivanpauno: nit: unneeded new line

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -26,7 +27,6 @@
 #include "rclcpp/visibility_control.hpp"
 
 #ifdef ANDROID
-#include <sstream>
```
ivanpauno: Why was this deleted?

rclcpp/include/rclcpp/utilities.hpp
hidmic: Leftover of when `join()` was here. Reverted.

rclcpp/include/rclcpp/utilities.hpp
hidmic: Leftover of when `join()` was here. Reverted.

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -27,7 +27,6 @@
 
 #ifdef ANDROID
 #include <sstream>
-
```
ivanpauno: nit: There's now an extra line deletion in `utilities.hpp` :P

hidmic: Argh! Fixed.

PR #841, opened 2019-09-02
check valid timer handler 1st to reduce the time window for scan.

fujitatomoya:
related to the following discussion for SingleThreadedExecutor CPU consumption,
https://discourse.ros.org/t/singlethreadedexecutor-creates-a-high-cpu-overhead-in-ros-2/10077/6

timer_handler_ should be checked if it is valid or not at the 1st entry. this is done for other objects such as subscriptions etc...but not for timer.

confirmed the following optimization result with typical case using ros2_performance nopub.

[Before]
```
root@2e44b65b8217:~/docker_colcon_ws# top | grep nopub
 7301 root      20   0 3220996  79812  17116 S  18.3  0.5   0:07.10 nopub                                               
 7301 root      20   0 3220996  79812  17116 R  17.3  0.5   0:07.62 nopub                                               
 7301 root      20   0 3220996  79812  17116 S  17.3  0.5   0:08.14 nopub                                               
 7301 root      20   0 3220996  79812  17116 S  18.7  0.5   0:08.70 nopub                                               
 7301 root      20   0 3220996  79812  17116 S  19.0  0.5   0:09.27 nopub                                               
 7301 root      20   0 3220996  79812  17116 R  17.6  0.5   0:09.80 nopub                                               
 7301 root      20   0 3220996  79812  17116 R  15.7  0.5   0:10.27 nopub                                               
 7301 root      20   0 3220996  79812  17116 S  16.9  0.5   0:10.78 nopub                                               
 7301 root      20   0 3220996  79812  17116 S  15.0  0.5   0:11.23 nopub                                               
 7301 root      20   0 3220996  79812  17116 S  16.7  0.5   0:11.73 nopub        
```

[After]
```
root@2e44b65b8217:~/docker_colcon_ws# top | grep nopub
11847 root      20   0 3221000  80312  17272 S  13.3  0.5   0:00.57 nopub                                               
11847 root      20   0 3221000  80312  17272 S  12.0  0.5   0:00.93 nopub                                               
11847 root      20   0 3221000  80312  17272 S  12.3  0.5   0:01.30 nopub                                               
11847 root      20   0 3221000  80312  17272 S  12.3  0.5   0:01.67 nopub                                               
11847 root      20   0 3221000  80312  17272 S  11.3  0.5   0:02.01 nopub                                               
11847 root      20   0 3221000  80312  17272 S  11.3  0.5   0:02.35 nopub                                               
[...]

PR #844, opened 2019-09-04
Make Executor::spin_some() only check for work once

roger-strain:
The goal of this change is to allow `spin_some` to only queue work items once at the start of an execution, without adversely affecting other spin methods.

Previously, work could come either from `get_next_timer`, which would always return a ready timer, even if it were not previously ready; or from the `wait_for_work` method, which was called when no work was currently ready, regardless of whether existing work had been completed or not.

This change modifies `get_next_timer` to instead pull from a list of cached timers which were ready at the previous call to `wait_for_work`. `wait_for_work` itself is still called from `get_next_executable`, but is now guarded by an `allow_wait` flag. By default, this flag is true, (nearly) maintaining current behavior. However, `spin_some` will set this flag true on its first iteration only, setting the flag to false for subsequent calls until `spin_some` itself returns. This causes the list of ready timers to be cached only once, and other waitables to be collected only once. The list is not refreshed during execution of `spin_some`, allowing the method to return when its workload is complete.

Note:

1. If `spin_some` does not complete its workload, a subsequent call will resume where it left off, without refreshing the workload available. While this may mean that a call will not pick up newly available work for that execution, it also means that tasks which are perpetually at the end of the list of executables will not be starved.

2. The "nearly" exception above refers to this same behavior, but in the context of `spin_once`. Previously, if `spin_once` were called with a very short duration timer, that timer may be executed to the exclusion of other work. `spin_once` will now follow the same logic regarding caching of timers and other work, so that all executables are eventually serviced.

I have attempted to create some unit tests to cover the `spin_some` behavior in the ros2/system_tests repo, and successfully ran the timer test, but have not gotten a publish/subscribe test working yet, possibly due to faults in my development environment. If possible, I'd love to push those up and get feedback on how to make those tests work, but I'm not sure the best approach just yet.

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -234,9 +234,11 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
     throw std::runtime_error("spin_some() called while already spinning");
   }
   RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
+  // non-blocking call to pre-load all available work
+  wait_for_work(std::chrono::milliseconds::zero());
   while (spinning.load() && max_duration_not_elapsed()) {
     AnyExecutable any_exec;
-    if (get_next_executable(any_exec, std::chrono::milliseconds::zero())) {
+    if (get_next_ready_executable(any_exec)) {
       execute_any_executable(any_exec);
     } else {
       break;
@@ -531,51 +533,39 @@ Executor::get_group_by_timer(rclcpp::TimerBase::SharedPtr timer)
 bool
 Executor::get_next_ready_executable(AnyExecutable & any_executable)
 {
+  bool success = false;
   // Check the timers to see if there are any that are ready
   memory_strategy_->get_next_timer(any_executable, weak_nodes_);
   if (any_executable.timer) {
-    return true;
+    success = true;
   }
-  // Check the subscriptions to see if there are any that are ready
-  memory_strategy_->get_next_subscription(any_executable, weak_nodes_);
-  if (any_executable.subscription) {
-    return true;
-  }
-  // Check the services to see if there are any that are ready
-  memory_strategy_->get_next_service(any_executable, weak_nodes_);
-  if (any_executable.service) {
-    return true;
-  }
-  // Check the clients to see if there are any that are ready
-  memory_strategy_->get_next_client(any_executable, weak_nodes_);
-  if (any_executable.client) {
-    return true;
-  }
-  // Check the waitables to see if there are any that are ready
-  memory_strategy_->get_next_waitable(any_executable, weak_nodes_);
-  if (any_executable.waitable) {
-    return true;
-  }
-  // If there is no ready executable, return a null ptr
-  return false;
-}
-
-bool
-Executor::get_next_executable(AnyExecutable & any_executable, std::chrono::nanoseconds timeout)
-{
-  bool success = false;
-  // Check to see if there are any subscriptions or timers needing service
-  // TODO(wjwwood): improve run to run efficiency of this function
-  success = get_next_ready_executable(any_executable);
-  // If there are none
   if (!success) {
-    // Wait for subscriptions or timers to work on
-    wait_for_work(timeout);
-    if (!spinning.load()) {
-      return false;
+    // Check the subscriptions to see if there are any that are ready
+    memory_strategy_->get_next_subscription(any_executable, weak_nodes_);
+    if (any_executable.subscription) {
+      success = true;
     }
-    // Try again
-    success = get_next_ready_executable(any_executable);
+  }
+  if (!success) {
+    // Check the services to see if there are any that are ready
+    memory_strategy_->get_next_service(any_executable, weak_nodes_);
+    if (any_executable.service) {
+      success = true;
+    }
+  }
+  if (!success) {
+    // Check the clients to see if there are any that are ready
+    memory_strategy_->get_next_client(any_executable, weak_nodes_);
+    if (any_executable.client) {
+      success = true;
+    }
+  }
+  if (!success) {
+    // Check the waitables to see if there are any that are ready
+    memory_strategy_->get_next_waitable(any_executable, weak_nodes_);
+    if (any_executable.waitable) {
+      success = true;
+    }
   }
   // At this point any_exec should be valid with either a valid subscription
   // or a valid timer, or it should be a null shared_ptr
@@ -594,6 +584,27 @@ Executor::get_next_executable(AnyExecutable & any_executable, std::chrono::nanos
       // any_exec is destructued
       any_executable.callback_group->can_be_taken_from().store(false);
     }
+  }
+  // If there is no ready executable, return a null ptr
+  return success;
+}
+
+bool
+Executor::get_next_executable(AnyExecutable & any_executable, std::chrono::nanoseconds timeout)
+{
+  bool success = false;
+  // Check to see if there are any subscriptions or timers needing service
+  // TODO(wjwwood): improve run to run efficiency of this function
+  success = get_next_ready_executable(any_executable);
+  // If there are none
+  if (!success) {
+    // Wait for subscriptions or timers to work on
+    wait_for_work(timeout);
+    if (!spinning.load()) {
+      return false;
+    }
+    // Try again
+    success = get_next_ready_executable(any_executable);
   }
   return success;
 }


rclcpp/src/rclcpp/executor.cpp
```diff
@@ -234,13 +234,18 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
     throw std::runtime_error("spin_some() called while already spinning");
   }
   RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
+  // during the first run only, the process is allowed to check for new work
+  bool allow_wait = true;
   while (spinning.load() && max_duration_not_elapsed()) {
     AnyExecutable any_exec;
-    if (get_next_executable(any_exec, std::chrono::milliseconds::zero())) {
+    if (get_next_executable(any_exec, std::chrono::milliseconds::zero(), allow_wait)) {
```
ivanpauno: We're passing `std::chrono::milliseconds::zero()`, so this is never waiting at all.

ivanpauno:
@roger-strain friendly ping

roger-strain:
@ivanpauno Apologies, apparently I don't have things set up to get notifications properly. I've got this on my radar now, so I'll take a look as soon as I have a free cycle.

rclcpp/src/rclcpp/executor.cpp
roger-strain: The reason for the added flag is actually beyond the amount of time that's used for waiting. Even with a zero wait duration, the `wait_for_work` method is still called, and it does several things beyond simply waiting for the timeout to occur. Most notably, it calls `add_handles_to_wait_set`, which will requery the various sources for work. This is one of the two ways that "extra" work can be considered during a run of `spin_some` (the other being the timers). By adding a flag that prevents `wait_for_work` from being called repeatedly, we can avoid this rebuilding of the work queues.

Perhaps it would be better if the flag were renamed, perhaps to `allow_new_work`?

roger-strain:
> Also, there's another proposal to change how we get the next ready timer in #841, which I prefer.

I'll take a look at the changes there and see how to integrate a fix for this issue using that approach.

roger-strain:
@ivanpauno Finally had an opportunity to revisit this. The other PR you mentioned actually was a beautiful change for what I was trying to help resolve here, as the timers now line up with the same behavior as the other work sources. I've clarified the flag name and the code is significantly simpler this way. See if this one makes more sense.

roger-strain:
Updated the commit to clean up the formatting check that previously failed.

hidmic:
> While this may mean that a call will not pick up newly available work for that execution, it also means that tasks which are perpetually at the end of the list of executables will not be starved.

I'm a bit concerned about this, as IIUC it suggests we won't meet timer deadlines either or any work submitted during `spin_some`. It almost seems to me we're coping with inadequate queuing / scheduling (or its complete lack thereof) simply by temporarily disabling it.  `Executor::spin_some()` documentation isn't very precise though, so this may well be a valid interpretation, but I _think_ we should be aiming for correct scheduling instead. @wjwwood what do you think?

wjwwood:
I agree with @hidmic that `spin_some` should not try to solve the starvation issue, as that is something that should be solved by a better default scheduler which is more fair, and by potentially having more than one scheduling algorithm as an option.

To that end, I think `spin_some` should not preserve state from call to call, and instead I think `spin_some` should collect all the work that could be done at the start, execute as much of it as it can during the time allotted, and re-evaluating the "work that can be done at the start" again on  subsequent calls.

----

W.r.t. the new flag, I would have prefered a solution which just called deeper functions in the API rather than passing a new flag to the top level API currently used by `spin` and `spin_once`. This might require shuffling some functions around internally, but it would be better in my opinion. 

If we follow the change in this pr, which makes timers pull from a cached list, rather than re-evaluating it at each check, then basically the structure I imagine would be more like this (greatly simplified):

```c++
void
spin_once(DurationT timeout)
{
  LockT<MutexT> lock_for_spinning(mutex_for_spinning_);
  auto any_executable = get_next_executable(timeout);
  if (any_executable) {
    execute_any_executable(any_executable);
  }
}

AnyExecutable
get_next_executable(DurationT timeout)
{
  // simplified for brevity
  // get_next_ready_executable() is where any scheduling happens
  auto any_executable = get_next_ready_executable();  // non-blocking
  if (any_executable) {
    // skipping some house keeping here
    return any_executable;
  }
  // otherwise wait for work
  wait_for_work(timeout);  // time limited blocking
  return get_next_ready_executable();  // non-blocking
}

// note max_spin_time is not a strong guarantee, just like sleep_for
// it's a target that the system does it's best to hit
// also it is a limit for how long to do work, not how long to wait for it
// this function would not wait for work if there's none to be done right now
void
spin_some(DurationT max_spin_time)
{
  LockT<MutexT> lock_for_spinning(mutex_for_spinning_);
  auto start_time = now();
  // do a non-blocking wait for work to make
  // sure everything new is being considered
  wait_for_work(0ms);  // non-blocking
  while (now() -  start_time <= max_spin_time) {
    // scheduling happens in this function (for better or worse)
    auto any_executable = get_next_ready_executable();  // non-blocking
    if (!any_executable) {
      return;
    }
    execute_any_executable(any_executable);
  }
} 
```

I actually think this may be a reasonable approach.

----

The other approach would involve extracting all the work to be done into a vector of "AnyExecutables" and processing them in order afterwards. The downside to this one is that any unprocessed (due to running out of spin time) "AnyExecutables" need to be safely returned or disposed of so they can be considered for future spin calls. I haven't investigated any hitches in this design yet though.

roger-strain:
@wjwwood I've changed the approach to match the (simpler) style you were discussing. Of note, I had to move a block of code dealing with mutually exclusive callback groups from `get_next_executable` to `get_next_ready_executable`, since the former is no longer called from each execution point. 

Not sure what the build failure is; didn't show up locally, and doesn't immediately seem related to the changeset, but maybe I'm misinterpreting it.
```
14:48:59 /tmp/ws/src/rclcpp/rclcpp/resource/logging.hpp.em:80: error: <class 'TypeError'>: can only concatenate tuple (not "str") to tuple
14:48:59 make[2]: *** [include/rclcpp/logging.hpp] Error 1
```

ivanpauno:
This test is consistently failing in all the implementations:

https://github.com/ros2/system_tests/blob/95680e750cff6b4f5e8fb377c0857f34dc20e954/test_rclcpp/test/test_executor.cpp#L58-L75

IMO, it's logical that is now failing.
@roger-strain Can you send a PR updating that test?
I would change the test to actually do some work, and check that is finished before 10ms.

roger-strain:
@ivanpauno I've got an update that I believe fixes that unit test, but I'm also seeing a failure in `spin_some_only_checks_waitable_once_subscription`; is that one expected, or should I dive deeper on it as well?

ivanpauno:
> @ivanpauno I've got an update that I believe fixes that unit test, but I'm also seeing a failure in `spin_some_only_checks_waitable_once_subscription`; is that one expected, or should I dive deeper on it as well?

Please, copy a link to that test (I searched for `spin_some_only_checks_waitable_once_subscription`, and I didn't find anything).

roger-strain:
> Please, copy a link to that test (I searched for `spin_some_only_checks_waitable_once_subscription`, and I didn't find anything).

Okay, that's kind of embarassing. I actually had written some unit tests in my earlier iteration of this, and that was one of them. I didn't clean my workspace before working on this one. I'll remove that and get a PR in on that repo shortly.

roger-strain:
Corresponding unit test PR submitted as ros2/system_tests#394

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8698/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4534/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7105/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8617/)

ivanpauno:
@roger-strain Please, correct the linter failure https://ci.ros2.org/job/ci_linux/8702/testReport/junit/test_rclcpp/uncrustify/test_test_executor_cpp/.

----

> Okay, that's kind of embarassing. I actually had written some unit tests in my earlier iteration of this, and that was one of them. I didn't clean my workspace before working on this one. I'll remove that and get a PR in on that repo shortly.

:rofl: haha, that happens.

roger-strain:
> @roger-strain Please, correct the linter failure https://ci.ros2.org/job/ci_linux/8702/testReport/junit/test_rclcpp/uncrustify/test_test_executor_cpp/.

@ivanpauno Updated commit should be there now.

ivanpauno:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7117/)

ivanpauno:
Going in, thanks @roger-strain !!

PR #848, opened 2019-09-10
Fix typo in deprecated warning.

luca-della-vedova:
Hi all,

Just a small typo that I incurred in while building other packages, "it's" instead of "its".

Cheers,
/Luca


dirk-thomas:
Thanks for the patch.

PR #853, opened 2019-09-16
Trait tests for generated actions

mjcarroll:
Adds test coverage for the changes in ros2/rosidl#412.

Currently, `action_msgs` are required in order to be able to generate action, so the tests are here to prevent any sort of circular dependency in `rosidl_generator_cpp`.

### DIFF
--- a/rclcpp_action/test/test_traits.cpp
+++ b/rclcpp_action/test/test_traits.cpp
@@ -1,0 +1,72 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include <test_msgs/action/fibonacci.hpp>
+
+using rosidl_generator_traits::is_message;
+using rosidl_generator_traits::is_service;
+using rosidl_generator_traits::is_action;
+using rosidl_generator_traits::is_action_goal;
+using rosidl_generator_traits::is_action_result;
+using rosidl_generator_traits::is_action_feedback;
+
+TEST(TestActionTraits, is_action) {
+  using Fibonacci = test_msgs::action::Fibonacci;
+
+  // Top level definition is an action
+  ASSERT_FALSE(is_message<Fibonacci>());
+  ASSERT_FALSE(is_service<Fibonacci>());
+  ASSERT_TRUE(is_action<Fibonacci>());
+  ASSERT_FALSE(is_action_goal<Fibonacci>());
+  ASSERT_FALSE(is_action_result<Fibonacci>());
+  ASSERT_FALSE(is_action_feedback<Fibonacci>());
+
+  // Goal is an action_goal as well as a message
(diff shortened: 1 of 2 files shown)


dirk-thomas:
> `action_msgs` are required in order to be able to generate action

Can you point to what exactly requires `action_msgs` to *generate* actions? Afaik it only requires that package when you want to include the `action__struct.hpp` which is not necessary to use/test the traits.

mjcarroll:
Maybe I'm unclear, but doesn't message (and action) generation happen during the build phase?

jacobperron:
We are referencing action_msgs interfaces when generating code for actions:

https://github.com/ros2/rosidl/blob/07a2ab5c452c462ae067769b752a1811cfb3612e/rosidl_generator_cpp/resource/action__struct.hpp.em#L91-L94

AFAICT, this is the only place, so maybe it's easy to avoid.

mjcarroll:
What happens in the case when you need to generate an action and then immediately use it in the same package?  In that case, it seems that `action_msgs` to need to be a `build_depend`.  Is that something we want to leave up to the user or should it be handled by the generator itself?

dirk-thomas:
> What happens in the case when you need to generate an action and then immediately use it in the same package? In that case, it seems that `action_msgs` to need to be a `build_depend`.

Yes, if a package does want to do that it needs to declare that extra dependency.

jacobperron:
I tried removing `action_msgs` as a dependency from `rosidl_cmake` (ie. the find_package call) and the action_msgs includes from the action struct template. But, `action_msgs`is is transitively bringing in a needed dependency to `unique_identifier_msgs`. When I try to build an action, includes are added for `unique_identfier_msgs`, e.g.:

```
/home/jacob/ws/latest_ws/build/test_msgs/rosidl_generator_cpp/test_msgs/action/fibonacci__struct.hpp:356:10: fatal error: unique_identifier_msgs/msg/uuid__struct.hpp: No such file or directory
 #include "unique_identifier_msgs/msg/uuid__struct.hpp"
```

Maybe we could add a special case for them in the template? I try for a little longer to see if it can be done.

rclcpp_action/test/test_traits.cpp
```diff
@@ -0,0 +1,72 @@
...
+  ASSERT_FALSE(is_action<Fibonacci::Feedback>());
+  ASSERT_FALSE(is_action_goal<Fibonacci::Feedback>());
+  ASSERT_FALSE(is_action_result<Fibonacci::Feedback>());
+  ASSERT_TRUE(is_action_feedback<Fibonacci::Feedback>());
+}
+
+TEST(TestActionTraits, is_action_impl) {
+  using Fibonacci = test_msgs::action::Fibonacci;
+
+  // Test traits on some of the internal implementation of actionlib
```
jacobperron: nit: *actionlib* refers to the ROS 1 implementation. 
```suggestion
  // Test traits on some of the internal implementation of actions
```

rclcpp_action/test/test_traits.cpp
```diff
@@ -0,0 +1,72 @@
...
+  ASSERT_TRUE(is_service<Fibonacci::Impl::SendGoalService>());
```
jacobperron: What about adding some ASSERT_FALSE cases for completeness (like in the previous test)?

rclcpp_action/test/test_traits.cpp
```diff
@@ -0,0 +1,72 @@
...
+using rosidl_generator_traits::is_action;
+using rosidl_generator_traits::is_action_goal;
+using rosidl_generator_traits::is_action_result;
+using rosidl_generator_traits::is_action_feedback;
+
+TEST(TestActionTraits, is_action) {
+  using Fibonacci = test_msgs::action::Fibonacci;
+
+  // Top level definition is an action
+  ASSERT_FALSE(is_message<Fibonacci>());
```
jacobperron: nit: using `EXPECT_*` instead, we'd see multiple failures instead of just the first to fail.

PR #858, opened 2019-09-19
add more context to exception message

dirk-thomas:
Helpful when the exception happens to know what the two clock types actually are.

### DIFF
--- a/rclcpp/src/rclcpp/time.cpp
+++ b/rclcpp/src/rclcpp/time.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <limits>
+#include <string>
 #include <utility>
 
 #include "rclcpp/clock.hpp"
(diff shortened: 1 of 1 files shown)


rclcpp/src/rclcpp/time.cpp
```diff
@@ -194,7 +194,7 @@ Duration
 Time::operator-(const rclcpp::Time & rhs) const
 {
   if (rcl_time_.clock_type != rhs.rcl_time_.clock_type) {
-    throw std::runtime_error("can't subtract times with different time sources");
+    throw std::runtime_error(std::string("can't subtract times with different time sources [") + std::to_string(rcl_time_.clock_type) + " != " + std::to_string(rhs.rcl_time_.clock_type) + "]");
```
jacobperron: Are linters happy with the length of this line?

dirk-thomas: Sorry, good point. Fixed in ef7511c3ff2db0d8e475640f06c003aa574352ed.

PR #862, opened 2019-09-20
Remove timeout on tests

zmichaels11:
* timeout was not required since the request will fail if the node is not found

Signed-off-by: Zachary Michaels <zmichaels11@gmail.com>

### DIFF
--- a/rclcpp_components/test/test_component_manager_api.cpp
+++ b/rclcpp_components/test/test_component_manager_api.cpp
@@ -56,7 +56,7 @@ TEST_F(TestComponentManager, load_components)
     request->plugin_name = "test_rclcpp_components::TestComponentFoo";
 
     auto result = client->async_send_request(request);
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Removing this timeout here doesn't make any sense imo. If the call doesn't finish within 5s it is extremely unlikely that it will finish in the remaining time of the (usually using a CTest timeout of 30 or 60 seconds). Though if it fails to complete within 5s it is safe to fail the test.

The node not being found is only one of the possible scenarios. It can absolutely happen that the node was found and the request sent but no response is ever received. And in that case the explicit timeout makes sure the test fails quickly.

mjcarroll:
@hidmic believes that there may be a race condition specifically effecting FastRTPS, and has a proposed solution here: https://github.com/ros2/rclcpp/pull/876

zmichaels11:
The intention of this PR was to improve stability with this test. Instead it looks like there's another underlying issue. 

There's an open issue on the flakiness of this test here: #863

PR #864, opened 2019-09-21
Zero copy api

Karsten1987:
connects to https://github.com/ros2/ros2/issues/785


rclcpp/include/rclcpp/loaned_message_sequence.hpp
```diff
@@ -0,0 +1,111 @@
...
+   * is being casted to the right message type.
+   *
+   * \param[in] pos Index indicating the position of the desired message.
+   * \return Reference to the index message if available
+   * \throw out of range exception if sequence is empty or index out of bounds.
+   */
+  template<typename MessageT>
+  std::shared_ptr<MessageT> at(size_t pos) const
+  {
+    return std::static_pointer_cast<MessageT>(at(pos));
```
wjwwood: How do you return a `shared_ptr` here? how does it share ownership? won't the destructor of the `shared_ptr` do the wrong thing here (call delete)?

rclcpp/include/rclcpp/loaned_message_sequence.hpp
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * \param[in] sub Pointer to rclcpp::SubscriptionBase where the samples are associated to.
+   */
+  LoanedMessageSequence(const rclcpp::SubscriptionBase * sub);
+
+  /// Move constructor of the LoanedMessageSequenceBase class
+  /**
+   * \param[in] other LoanedMessageSequenceBase to move.
+   */
+  LoanedMessageSequence(LoanedMessageSequence && other);
```
wjwwood: By convention, constructors should come first (before the methods above) in the class declaration.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -132,6 +136,37 @@ class Publisher : public PublisherBase
...
+   *
+   * With a call to \sa `publish` the LoanedMessage instance is being returned to the middleware
+   * or free'd accordingly to the allocator.
+   * If the message is not being published but processed differently, the message has to be
+   * returned by calling \sa `return_loaned_message`.
+   * \sa rclcpp::LoanedMessage for details of the LoanedMessage class.
+   *
+   * \return LoanedMessage containing memory for a ROS message of type MessageT
+   */
+  rclcpp::LoanedMessage<MessageT, AllocatorT>
```
wjwwood: Should the AllocatorT be derived from the MessageT's allocator? Or is there a reason to be able to specify them separately I wonder?

wjwwood: I guess when you do a vector of strings, you can specify the allocator for both separately, so maybe what's here already makes sense.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -132,6 +136,37 @@ class Publisher : public PublisherBase
...
+  {
+    return rclcpp::LoanedMessage<MessageT, AllocatorT>(this, this->get_allocator());
+  }
+
+  /// Return and deallocate the memory for the loaned message.
+  /**
+   * \param loaned_message The LoanedMessage instance to be returned.
+   */
+  void
+  return_loaned_message(rclcpp::LoanedMessage<MessageT, AllocatorT> && loaned_msg) const
```
wjwwood: I'm dubious about this method. Either it should result in a post condition that the loan has been returned (not the case here) or it shouldn't exist, in my opinion.

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -159,12 +160,26 @@ class Subscription : public SubscriptionBase
...
-  void handle_message(std::shared_ptr<void> & message, const rmw_message_info_t & message_info)
+  void *
+  loan_message() override
+  {
+    CallbackMessageT * msg = nullptr;
+    return msg;
+  }
+
+  rclcpp::LoanedMessageSequence
+  loan_message_sequence() override
```
wjwwood: What are these methods used for?

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -175,22 +190,49 @@ class Subscription : public SubscriptionBase
     any_callback_.dispatch(typed_message, message_info);
   }
 
+  void
+  handle_loaned_message(
+    void * loaned_message, const rmw_message_info_t & message_info) override
+  {
+    // make messageT pointer out of it.
+    auto typed_message = reinterpret_cast<CallbackMessageT*>(loaned_message);
```
wjwwood: Is `reinterpret_cast` the right thing here?

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -175,22 +190,49 @@ class Subscription : public SubscriptionBase
...
+  {
+    auto ret = rcl_return_loaned_message(subscription_handle_.get(), msg);
+    if (RCL_RET_OK != ret) {
+      RCLCPP_ERROR(rclcpp::get_logger("rclcpp"), "%s", rcl_get_error_string().str);
+      rcl_reset_error();
+    }
+  }
+
+  void
+  return_loaned_message_sequence(
```
wjwwood: Again, not sure about this method if it literally does nothing.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -331,6 +331,23 @@ Executor::execute_subscription(
       rcl_reset_error();
     }
     subscription->return_serialized_message(serialized_msg);
+  } else if (subscription->can_loan_messages()) {
+    auto loaned_msg = subscription->loan_message();
+    auto ret = rcl_take_loaned_message(
```
wjwwood: Shouldn't think return a loaned message? Why are we getting one from the subscription first?

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,170 @@
...
+template<typename MessageT, typename AllocatorT = std::allocator<void>>
+class LoanedMessage
+{
+  using MessageAllocatorTraits = allocator::AllocRebind<MessageT, AllocatorT>;
+  using MessageAllocator = typename MessageAllocatorTraits::allocator_type;
+
+protected:
+  const rclcpp::PublisherBase * pub_;
+
+  std::unique_ptr<MessageT> message_;
```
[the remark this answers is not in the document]
Karsten1987: we discussed this offline that with having a `std::unique_ptr` the class becomes implicitly non-trivial copyable. Do having a move constructor is really only needed for RVO.

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+  const std::shared_ptr<MessageAllocator> message_allocator_;
+
+  /// Deleted copy constructor to preserve memory integrity
```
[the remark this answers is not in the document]
Karsten1987: with the `unique_ptr` I think I don't need to make the copy constructor private. I'll check that.

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,170 @@
...
+    } else {
+      RCLCPP_WARN(
+        rclcpp::get_logger("rclcpp"),
+        "Currently used middleware can't loan messages. Local allocator will be used.");
+      message_memory = message_allocator_->allocate(1);
+    }
+    if (!message_memory) {
+      throw std::runtime_error("unable to allocate memory for loaned message");
+    }
+    message_.reset(new (message_memory) MessageT());
```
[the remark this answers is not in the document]
Karsten1987: I understand that it might have been called in the second case. I don't think it has been called in the first case though. The first one really only allocates enough space.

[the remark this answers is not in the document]
wjwwood: I'm not sure I agree... What if it is reusing a message that was already there? Like reusing items from a pool? Why would the middleware not construct it after allocating?

rclcpp/include/rclcpp/loaned_message.hpp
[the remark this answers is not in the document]
wjwwood: Ok, that's fine, it just seems a bit odd to be using `reset` and `release` in the constructor/destructor of this class rather than using the deleter of the unique pointer.

You could also just disable the copy constructors and use a raw pointer.

rclcpp/include/rclcpp/subscription.hpp
Karsten1987: I guess they are more for consistency when implementing the executor (in the style of `sub->create_message()`, `sub->handle_message()`, `sub->return_message()`). But I am fine with only having the `return_loaned_message()` functions.

rclcpp/src/rclcpp/executor.cpp
Karsten1987: I opted for an _in-out_ parameter to have a return value indicating if any error occurred.
Would you recommend just returning `nullptr` if error and check accordingly?

wjwwood: It's not an in-out parameter is it? If so why do you initialize it with `subscription->loan_message()`?

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,170 @@
...
+    message_allocator_(allocator)
+  {
+    if (!pub) {
+      throw std::runtime_error("publisher pointer is null");
+    }
+
+    void * message_memory = nullptr;
+    if (pub_->can_loan_messages()) {
+      message_memory =
+        rcl_allocate_loaned_message(pub_->get_publisher_handle(), nullptr, sizeof(MessageT));
```
Karsten1987:
```suggestion
        auto ret = rcl_borrow_loaned_message(pub_->get_publisher_handle(), ts, void ** loaned_msg);
        MsgT * msg = static_cast<MsgT*>(loaned_msg);
```

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,170 @@
...
+      return;
+    }
+
+    // release allocated memory from unique_ptr
+    MessageT * message_memory = message_.release();
+
+    if (pub_->can_loan_messages()) {
+      // return allocated memory to the middleware
+      auto ret =
+        rcl_deallocate_loaned_message(pub_->get_publisher_handle(), message_memory);
```
Karsten1987:
```suggestion
        rcl_return_loaned_message(pub_->get_publisher_handle(), msg);
```

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,170 @@
...
+   * The contract here is that the memory for this message is valid as long as this instance
+   * of the LoanedMessage class is alive.
+   */
+  virtual ~LoanedMessage()
+  {
+    auto error_logger = rclcpp::get_logger("LoanedMessage");
+    if (!pub_) {
+      RCLCPP_ERROR(error_logger, "Can't deallocate message memory. Publisher instance is NULL");
+      return;
+    }
```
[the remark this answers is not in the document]
mjcarroll: @Karsten1987 can you address this?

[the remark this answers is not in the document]
Karsten1987: isn't `pub_` already a reference?

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -132,6 +136,26 @@ class Publisher : public PublisherBase
...
+  /// Loan memory for a ROS message from the middleware
+  /**
+   * If the middleware is capable of loaning memory for a ROS message instance,
+   * the loaned message will be directly allocated in the middleware.
+   * If not, the message allocator of this rclcpp::Publisher instance is being used.
+   *
+   * With a call to \sa `publish` the LoanedMessage instance is being returned to the middleware
+   * or free'd accordingly to the allocator.
+   * If the message is not being published but processed differently, the message has to be
+   * returned by calling \sa `return_loaned_message`.
```
wjwwood: Is this still a thing?

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -191,6 +215,41 @@ class Publisher : public PublisherBase
     return this->do_serialized_publish(&serialized_msg);
   }
 
+  /// Publish an instance of a LoanedMessage
```
wjwwood: Punctuation.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -191,6 +215,41 @@ class Publisher : public PublisherBase
...
+  /**
+   * When publishing a loaned message, the memory for this ROS instance will be deallocated
```
wjwwood: "ROS instance"?

rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -138,6 +138,12 @@ class SubscriptionBase : public std::enable_shared_from_this<SubscriptionBase>
   void
   handle_message(std::shared_ptr<void> & message, const rmw_message_info_t & message_info) = 0;
 
+  // TODO(karsten1987): Does it make sense to pass in a weak_ptr?
```
wjwwood: A `weak_ptr` to what? Can you please create an issue to follow up on this...

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -331,6 +331,32 @@ Executor::execute_subscription(
...
+    if (RCL_RET_OK == ret) {
+      subscription->handle_loaned_message(loaned_msg, message_info);
+    } else if (RCL_RET_SUBSCRIPTION_TAKE_FAILED != ret) {
+      RCUTILS_LOG_ERROR_NAMED(
+        "rclcpp",
+        "take_loaned failed for subscription on topic '%s': %s",
+        subscription->get_topic_name(), rcl_get_error_string().str);
+      rcl_reset_error();
+    }
+    ret = rcl_release_loaned_message(
```
wjwwood: `release`? Why not `return`?

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -132,6 +136,26 @@ class Publisher : public PublisherBase
...
+   * With a call to \sa `publish` the LoanedMessage instance is being returned to the middleware
+   * or free'd accordingly to the allocator.
+   * If the message is not being published but processed differently, the message has to be
+   * returned by calling \sa `return_loaned_message`.
+   * \sa rclcpp::LoanedMessage for details of the LoanedMessage class.
+   *
+   * \return LoanedMessage containing memory for a ROS message of type MessageT
+   */
+  rclcpp::LoanedMessage<MessageT, AllocatorT>
+  loan_message()
```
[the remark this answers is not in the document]
mjcarroll: I think loan is a verb in this case.  Loaned message would be a noun.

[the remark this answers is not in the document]
wjwwood: Then it should be `loan_loaned_message()`, but that would be the wrong direction, the user calling this is not loaning a message to the publisher, it is borrowing one from it.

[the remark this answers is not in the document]
wjwwood: Looking at it another way, this should really be `get_loaned_message()` and `borrow` is just a more specific version of `get`.  Either way, I think `loan_message()` isn't correct, at best it's a somewhat backward (due to loan versus borrow) abbreviation.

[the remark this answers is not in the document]
Karsten1987: I don't feel strong about either, but I feel like changing it to `borrow_loaned_message()` makes more sense here as the `rcl/rmw` functions are called accordingly.

rclcpp/include/rclcpp/subscription_base.hpp
Karsten1987: I was wondering if a `weak_ptr` could help around placing a custom deleter in the shared pointer for the `any_callback_.dispatch()` function. But the solution implemented here does work, so I don't think we need that TODO any longer. 
I'll address that with the other documentation touchups in a follow up PR

rclcpp/src/rclcpp/executor.cpp
Karsten1987: I thought we agreed on having `borrow/publish/return` on the publishing side and `take/release` on the subscription side.
At least on rmw level we need a separation as we can't overload functions and have `rmw_return_loaned_message` take either a `rmw_subscription_t` or `rmw_publisher_t`.

wjwwood: Oh I remember that now, but I thought we went with `rmw_return_loaned_publisher_message()` and `rmw_return_loaned_subscription_message()` (or something similar)?

Karsten1987: I think we were also mentioning once having `discard_loaned_message` on the publisher side and then have `return_loaned_message` for subscriptions. 
I guess we should settle on the names soonish, as this will bring just another set of PRs with it ;-)

Karsten1987: I propose we stick with the suggestion of 
`rmw_return_loaned_message_from_publisher` and `rmw_return_loaned_message_from_subscription`

wjwwood: Sounds good thanks.

sgf201:
> connects to [ros2/ros2#785](https://github.com/ros2/ros2/issues/785)
> connects to [ros2/rmw_cyclonedds#469](https://github.com/ros2/rmw_cyclonedds/issues/469)
image
Is it possible that  using shared_ptr to manage loaned_msg, while calling rcl_return_loaned_message_from_ subscriptions in the destructor to achieve true sharedptr charactors of callback message.

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,206 @@
...
+  using MessageAllocatorTraits = allocator::AllocRebind<MessageT, AllocatorT>;
+  using MessageAllocator = typename MessageAllocatorTraits::allocator_type;
+
+public:
+  /// Constructor of the LoanedMessage class.
+  /**
+   * The constructor of this class allocates memory for a given message type
+   * and associates this with a given publisher.
+   *
+   * Given the publisher instance, a case differentiation is being performaned
```
Zard-C, line 42: performed? 😸

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * which decides whether the underlying middleware is able to allocate the appropriate
+   * memory for this message type or not.
+   * In the case that the middleware can not loan messages, the passed in allocator instance
```
Zard-C, line 45: `passed in` -> `passed-in`

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -0,0 +1,206 @@
...
+   * is being used to allocate the message within the scope of this class.
+   * Otherwise, the allocator is being ignored and the allocation is solely performaned
+   * in the underlying middleware with its appropriate allocation strategy.
+   * The need for this arises as the user code can be written explicitly targeting a middleware
+   * capable of loaning messages.
+   * However, this user code is ought to be usable even when dynamically linked against
```
Zard-C, line 51: ~~is~~ ought to be

Issue #825, opened 2019-08-15
Optimization of the SingleThreadedExecutor and MultiThreadedExecutor to reduce CPU overhead.


MartinCornelis2:
## Feature request / Bug report

Optimization of the SingleThreadedExecutor and MultiThreadedExecutor to reduce CPU overhead.
Possible solutions:
- Overhaul design of current implementation 
  - Work with callback groups instead of nodes
  - Use events to notify executor of changes to system instead of constant checking with weak nodes

OR
- Add a separate optimized Executor that is less dynamic (more user configuration)

#### Feature description
We did research on the large CPU overhead introduced by the Executor in rclcpp and found it to be a blocking issue for our ROS 2 platform. The results of the research can be found here: https://github.com/nobleo/ros2_performance . As a result of the research a discussion was started here:
https://discourse.ros.org/t/singlethreadedexecutor-creates-a-high-cpu-overhead-in-ros-2/10077 . People who joined the discussion also looked into the CPU overhead and came to the same conclusions. In our opinion changes to the Executor or alternatives to the current existing Executor are necessary for normal ROS 2 to run on embedded boards.

We are continuing our research by implementing a custom static Executor as a POC to get an idea of the maximum CPU gain that is possible. Code will be made available when we have something functional and the results of the research will be shared on our github.

Additionally the current implementation of the Executor has some weird effects that are discussed here https://t-blass.de/papers/response-time-analysis-of-ros2.pdf . The paper focuses more on how to measure response time than on the nature of the Executor. However, the paper does highlight the Executors peculiar preference for timers and the weird priority that is based on CallbackType and RegistrationTime. 

The improvements that could be made to scheduling (that are required for real-time) and the optimization with respect to CPU overhead could/should both be encapsulated in a new and improved Executor.

### Related issues
These issues need to be considered when redesigning the executor.
#392 
#519 
#726 

### Implementation considerations
Major changes to the Executor in rclcpp (and possibly to the node class and others) could break existing user code.  Implementing changes will take a lot of effort and may also involve changes in other abstraction layers of ROS2.

iluetkeb:
> Work with callback groups instead of nodes
So far I have not seen data that points to nodes vs. callback groups being a decisive factor. Have I missed something?

> Use events to notify executor of changes to system instead of constant checking with weak nodes

Here, you mean changes such as adding/removing subscribers, right?

MartinCornelis2:
[...]

PR #873, opened 2019-09-25
Added static single threaded executor functionality

ishugoel20:
Signed-off-by: Ishu Goel <ishu.goel@nobleo.nl>

This pull request aims to provide a static alternative to the existing single-threaded executor. In response to a  [question](https://answers.ros.org/question/327477/ros2-uses-6-times-more-cpu-than-fastrtps/) posted on ROS answers, we did research on the high CPU overhead of the existing implementation. The results of the research are [here](https://github.com/nobleo/ros2_performance). Based on the results we started a discussion on ROS discourse [here](https://discourse.ros.org/t/singlethreadedexecutor-creates-a-high-cpu-overhead-in-ros-2/) and we posted an [issue](https://github.com/ros2/rclcpp/issues/825) on the rclcpp GitHub. 

We contribute this static executor as an intermediate solution until another alternative is present or the existing executor is optimized. Our executor can be used by ROS users that have a static system to reduce CPU overhead. 

An explanation of our static single-threaded executor is not included in this pull request but can instead be found [here](https://github.com/nobleo/rclcpp-static-executor). This is an older Dashing version we made for testing. The README provides an overview in the form of a flowchart with a short description of functionality.


ishugoel20:
@alsora @fujitatomoya @ivanpauno @gbiggs @Karsten1987 can you have a look at this?

tfoote:
This current implementation seems like it would be great to operate in environments such as when using the lifecycle nodes: https://design.ros2.org/articles/node_lifecycle.html

Currently the implementation requires that the all the setup is done before starting the spin which is relatively limiting. Would it be possible to modify this to react to events such as new subscriptions at which time it will rebuild the index that it is then iterating over until there's a new event triggering a reevaluation of the list to iterate. 

There could be flags to keep it in permissive mode versus enforcing mode that tools like the lifecycle management and realtime environments might like to setup. But it seems that this would allow us to use this as a replacement for the current executor and get the benefits of faster iteration without requiring the developer to switch to a different executor.

MartinCornelis2:
We did already think about this. When a trigger occurs we can simply collect the handles, resize the wait-set and add the handles again. The list of executables can probably just be appended and may not require rebuilding. The only thing to consider in this case is waitables. Waitables may interfere with the order of things since they are added in a slightly different way(no longer possible to simply use indices to figure out what to execute).  We could look into making a seperate waitable struct, or figure out a way to retain a mapping between the wait-set and the list of executables. I think right now waitables work because they are assessed last and are always added at the end of the lists(I may be wrong here).

fujitatomoya:
@MartinCornelis2 

> The list of executables can probably just be appended and may not require rebuilding. The only thing to consider in this case is waitables. Waitables may interfere with the order of things since they are added in a slightly different way(no longer possible to simply use indices to figure out what to execute). 

can we just recreate entities and executable list via user's request for reconstruction? i believe that it is easier to implement, and besides this user event or request for reconstruction would not be happening every single time just like SingleThreadExecutor does. so i think that it is reasonable.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -0,0 +1,329 @@
...
+}
+
+void
+StaticSingleThreadedExecutor::execute_ready_executables(
+  ExecutableList & exec_list, std::chrono::nanoseconds timeout)
+{
+    refresh_wait_set(timeout);
+    //Execute all the ready subscriptions
+    for (size_t i = 0; i < wait_set_.size_of_subscriptions; ++i) {
+      if (wait_set_.size_of_subscriptions && i < exec_list.number_of_subscriptions) {
```
fujitatomoya: check exec_list.number_of_subscriptions at 1st, so we do not do that every single loop of wait_set_.size_of_subscriptions

ishugoel20:
@fujitatomoya 

> can we just recreate entities and executable list via user's request for reconstruction? i believe that it is easier to implement, and besides this user event or request for reconstruction would not be happening every single time just like SingleThreadExecutor does. so i think that it is reasonable.

We are working on this, would share when this is ready.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
ishugoel20: > check exec_list.number_of_subscriptions at 1st, so we do not do that every single loop of wait_set_.size_of_subscriptions ? also this applies to time, service and clients.

@fujitatomoya thanks for the suggestion, we would consider making this change.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -362,7 +362,6 @@ class Executor
   /// The context associated with this executor.
   std::shared_ptr<rclcpp::Context> context_;
 
-private:
```
allenh1: Is this line being removed to allow access to something in particular? Instead, could there be mutators added to this (the base) instead?

I don't think it's wise to make that list public.

ishugoel20: I have made weak_nodes_ and guard_conditions_ lists protected to be able to access them in StaticSingleThreadedExecutor class.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -0,0 +1,324 @@
...
+  for (size_t i = 0; i < exec_list.number_of_waitables; ++i) {
+    if (exec_list.number_of_waitables && exec_list.waitable[i]->is_ready(&wait_set_)) {
+      exec_list.waitable[i]->execute();
+    }
+  }
+  // Check the guard_conditions to see if anything is added to the executor
+  for (size_t i = 0; i < wait_set_.size_of_guard_conditions; ++i) {
+    if (wait_set_.guard_conditions[i]) {
+      // rebuild the wait_set and ExecutableList struct
+      run_collect_entities();
```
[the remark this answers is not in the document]
MartinCornelis2: At first we implemented the executor to be fully static, but this was quite restrictive. Now we use the node guard_conditions to check if something was added to a node at runtime. If this happens then we rebuild. This means that if nothing is added at runtime (everything is in place before .spin() ), this executor still behaves as a fully static one. So this check and rebuilding step only adds functionality. This means the executor can now be used for a wider variety of source code and will most likely still perform better than the original executor, since only rebuilding when necessary to me seems better than always rebuilding.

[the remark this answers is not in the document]
MartinCornelis2: @fujitatomoya 
> can we just recreate entities and executable list via user's request for reconstruction? i believe that it is easier to implement, and besides this user event or request for reconstruction would not be happening every single time just like SingleThreadExecutor does. so i think that it is reasonable.

Instead of rebuilding on request, we handled it like this (event trigger from node).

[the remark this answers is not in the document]
fujitatomoya: i think that your idea is better than giving the responsibility to user for reconstruction.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -458,6 +458,7 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
       memory_strategy_->number_of_guard_conditions(), memory_strategy_->number_of_ready_timers(),
       memory_strategy_->number_of_ready_clients(), memory_strategy_->number_of_ready_services(),
       memory_strategy_->number_of_ready_events());
+
```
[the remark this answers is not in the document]
ishugoel20: My mistake, I will undo it.

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -20,6 +20,7 @@
 
 #include "rclcpp/executors/multi_threaded_executor.hpp"
 #include "rclcpp/executors/single_threaded_executor.hpp"
+#include "rclcpp/executors/static_single_threaded_executor.hpp"
```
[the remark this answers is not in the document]
ishugoel20: I do not use StaticSingleThreadedExecutor in any of the below spin functions, so it is better not to include it in my view.

rclcpp/include/rclcpp/memory_strategy.hpp
```diff
@@ -22,6 +22,7 @@
 #include "rcl/wait.h"
 
 #include "rclcpp/any_executable.hpp"
+#include "rclcpp/executable_list.hpp"
```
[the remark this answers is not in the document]
ishugoel20: It is not required anymore, I will remove it. Thanks for pointing out.

rclcpp/include/rclcpp/executable_list.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef RCLCPP__EXECUTABLE_LIST_HPP_
```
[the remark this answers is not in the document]
ishugoel20: I was not sure if I had to add that. Thanks for the information. I will add the license note.

rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
```diff
@@ -0,0 +1,196 @@
+#ifndef RCLCPP__EXECUTORS__STATIC_SINGLE_THREADED_EXECUTOR_HPP_
```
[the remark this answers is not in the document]
ishugoel20: Will do here as well

rclcpp/src/rclcpp/executable_list.cpp
```diff
@@ -0,0 +1,14 @@
+#include "rclcpp/executable_list.hpp"
```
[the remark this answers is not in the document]
ishugoel20: Will do

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -0,0 +1,324 @@
+#include "rclcpp/executors/static_single_threaded_executor.hpp"
```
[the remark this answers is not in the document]
ishugoel20: Will do

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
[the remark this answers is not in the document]
fujitatomoya: i believe that this reconstruction only needs to be done here once, so find out if anything is notified, and then reconstruct just once and exit from this loop. this is performance concern if some of entities are updated.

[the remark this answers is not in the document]
MartinCornelis2: Yes this is true, I believe we did realize this, but apparently haven't updated it here. We can add a "break;" after get_executable_list, that should get the desired result right?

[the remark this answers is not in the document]
ivanpauno: A guard condition is for "waking up" the executor, and I think that shouldn't be coupled to re-collecting entities (I'm not sure if you always want to re-collect entities when a guard condition is triggered).
IMO, it's better to provide a way of doing it under user request.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -0,0 +1,335 @@
...
+  // Get all the clients
+  get_client_list(executable_list);
+
+  // Get all the waitables
+  get_waitable_list(executable_list);
+}
+
+void
+StaticSingleThreadedExecutor::execute_ready_executables(
+  ExecutableList & exec_list, std::chrono::nanoseconds timeout)
```
ivanpauno: Why do we need both the `exec_list` and the `memory_strategy_`?
I think there is some duplication there, as they are mostly collecting the same stuff.

Why not using directly `get_next_ready_executable` instead of re-implementing this?

My idea is that this should look quite similar to the current `SingleThreadedExecutor`, but avoiding calling `clear_handles` and `collect_entities` each time. I don't understand the need of another collection.

MartinCornelis2:
Hello Everyone,

Ishu and I are very happy to see that so many people have responded to this PR. 
What we see right now is that there is not really a consensus on how the executor will be refactored for F-turtle. So we plan to wait on implementing more changes until a decision has been made. 

The intention of this version was to show the inefficiencies of the original executor. With a few changes to the code, we managed to gain quite some performance. However, the executor is now sitting on top of old data-structures and functions that do not perfectly fit the implementation. A proper refactored executor would require a bigger rewrite and re-organization of rclcpp.

This is why we rather wait until we are close to the Foxy release and then rebase and help with the refactored executor.

dirk-thomas:
> This is why we rather wait until we are close to the Foxy release and then rebase and help with the refactored executor.

@MartinCornelis2 I would suggest to try to get your proposed changes reviewed and merged rather earlier than later in the release cycle. If you need certain changes or decisions to happen before that please push on those ones as well.

If you wait until close to the Foxy release it becomes likely that a significant PR like this won't be accepted in time and misses the freeze deadline and therefore can only be considered for the next ROS distro.

wjwwood:
@dirk-thomas please read up on the discussion: https://discourse.ros.org/t/singlethreadedexecutor-creates-a-high-cpu-overhead-in-ros-2/10077/21

One option is to get this reviewed and merged first, but it's not clear to me that is the best way.

If @MartinCornelis2 wants to do it that way, then I'm happy to try and get it merged though. Also if, after discussing the roadmap for Foxy, we will not have time for the executor improvements (very bad idea, imo) then that might be another reason to resurrect this pr early, since the aforementioned improvements won't come anyways.

alsora:
Any updates on this?
Is it going to be in the Foxy release?

wjwwood:
> Any updates on this?
> Is it going to be in the Foxy release?

I don't have any updates right now, but that is still the plan.

alsora:
@wjwwood  @ivanpauno  @MartinCornelis2

Thank you for the update.

We are currently running some benchmarks and it looks like this executor performs better than the standard one in all situations.
Moreover, we may have some ideas for improving the CPU usage even further.

What's the best way for proceeding in this case?
should we do PRs towards `nobleo:nobleo/feature/static_executor` or should we fork it and open a new PR from the new branch?

Few more questions:
 - what's the deadline for having this feature merged in order to see it in Foxy? 8th April ?
 - is also the general refactors of the executors still planned for next release?

Thank you

wjwwood:
> What's the best way for proceeding in this case?

I'd suggest a pr to this pr, or a new pr to master which builds on this pr (you can rebase the second one, when/if this one is merged.

I thought you guys had one that was in a completely separate package too right? You can always continue development there without waiting on me to approve each step.

I thought the idea was to wait for the rclcpp executor refactor then re-evaluate the differences with this design. In the meantime you could continue to offer your alternative design as a separate package.

> what's the deadline for having this feature merged in order to see it in Foxy? 8th April ?

Yes. https://index.ros.org/doc/ros2/Releases/Release-Foxy-Fitzroy/#timeline-before-the-release

> is also the general refactors of the executors still planned for next release?

I'll just copy from above...

> > Is it going to be in the Foxy release?

> I don't have any updates right now, but that is still the plan.

MartinCornelis2:
Hi Everyone,

@ishugoel20 and I are of course still very interested in contributing to this, however we are quite busy with client projects. Further development of this will either come out of our free time, or when we have some downtime between projects.

> I thought you guys had one that was in a completely separate package too right? You can always continue development there without waiting on me to approve each step.

Yes this package is here https://github.com/nobleo/static_executor and the plan is to add a Foxy branch when Foxy is released (straight port from the Dashing/Eloquent version). We are not planning to develop the static executor a lot more in terms of additional features or functionality. 

We definitely see the possibility for additional features (mostly with respect to scheduling), but this would require more information from rmw that is currently not available at the rclcpp level (Something the guys from Bosch looked into @iluetkeb ).

> I thought the idea was to wait for the rclcpp executor refactor then re-evaluate the differences with this design.

Yes, this is still very much our plan. After the refactor we will have another look at the executor and contribute where possible. We still hope some of our idea or code ends up in the core of ROS2, because that would simply be very cool :) .

> In the meantime you could continue to offer your alternative design as a separate package.

As stated above, we'll make sure there is a Foxy branch of our separate package sometime after the Foxy release.

@iluetkeb @Karsten1987 @wjwwood is there any update on the work/possibility of exposing more information from DDS to the rclcpp layer for scheduling/executing? Things like timestamps for instance were a discussion point back in the day. I'm not really up to date on other progress made in ROS2 that is indirectly related to this work.

wjwwood:
> @iluetkeb @Karsten1987 @wjwwood is there any update on the work/possibility of exposing more information from DDS to the rclcpp layer for scheduling/executing?

I'm not aware of any work being done here at the moment. If everyone is waiting on me to do it, then I think it's at serious risk of not being part of foxy. I wasn't intending to work on it, though it would obviously be valuable.

One issue that I know of is performance. Getting this information out of DDS is not efficient. You essentially have to do a `read`, getting (in most implementations a copy of) all of the data along with the sample info's which contain the timestamps we need. Also, to answer "how many items are in the history cache" you need to take or read them, even though this should be an atomic `load` of an integer (`size_t`) in most implementation cases that I could imagine... Unfortunately this is not part of the DDS API. However, we may have luck getting this information out of Fast-RTPS and cyclone just because of their open source nature (we could add the needed accessor). These problems need to be figured out a bit before we add stuff to the rmw API, imo.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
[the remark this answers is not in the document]
alsora: One of the problems that we found is that this condition causes `run_collect_entities` to be called every time a guard condition is triggered, even if no nodes/pubs/subs/etc are added

iluetkeb:
> > @iluetkeb @Karsten1987 @wjwwood is there any update on the work/possibility of exposing more information from DDS to the rclcpp layer for scheduling/executing?
> 
> I'm not aware of any work being done here at the moment. If everyone is waiting on me to do it, then I think it's at serious risk of not being part of foxy. I wasn't intending to work on it, though it would obviously be valuable.

We had indeed been hoping for @wjwwood  to take the lead here, but thanks for making it clear that we need to pick this up ourselves.

In my view, the question of how we get the timestamp from the middleware can be left to a later stage. Yes, currently getting the source timestamp from DDS is expensive and cumbersome but firstly, as you said, this is likely possible to address and secondly, we could also work with arrival timestamps in the interim, which can be acquired fairly easily in the rmw-layer itself. For middleware implementations that do not have source timestamps, the arrival timestamp would likely have to be substituted anyway.

However, it is not clear how we pass on the timestamp. Currently, I see three choices:
 1) Put a (variable size) array into the wait_set that contains timestamps
 2) Add a function by which we can query timestamps on a per-subscription basis.
 3) Leave ordering to the middleware.

Personally, I'm leaning towards option 2, because it would avoid having to mess with memory-management in the wait_set, and also because it seems we could implement it more efficiently.

If anybody has a better idea, please tell us :-)

wjwwood:
> We had indeed been hoping for @wjwwood to take the lead here, but thanks for making it clear that we need to pick this up ourselves.

Sorry that wasn't clear, I do plan on doing improvements to the executor design, but I didn't expect to get to exposing additional information through the rmw layer that might help with scheduling.

> In my view, the question of how we get the timestamp from the middleware can be left to a later stage. Yes, currently getting the source timestamp from DDS is expensive and cumbersome but firstly, as you said, this is likely possible to address and secondly, we could also work with arrival timestamps in the interim

The arrival time stamp is just as cumbersome to get at the moment. Only the middleware has the arrival and the source time stamps for each sample. Usually our rmw implementations only get notified that there is data, not how much nor any other information. For each "DDS sample" there is a sample info which contains the arrival and source timestamps. Accessing the sample info requires a read or take.

If the intention is to side step the performance issues by using the arrival timestamp I don't think that will help you.

Maybe you could use a "last time data became available on this subscription" like timestamp that would be managed above the DDS layer. But that's not a fully baked idea, there might be issues with it. Notably, we use a "read condition" rather than a "data received condition", so that if we leave data untaken in a subscription and then try to sleep again it wakes up because there's still data that could be read. This has been a problem in one case where the DDS client being used didn't support the read condition...

wjwwood:
> Leave ordering to the middleware.

With DDS the middleware always has control over ordering, AFAIK. You can influence how DDS presents (orders) the data in the history cache, but if you look at the signature of take, you don't get to randomly access data, e.g. RTI's docs:

https://community.rti.com/static/documentation/connext-dds/6.0.1/doc/api/connext_dds/api_cpp/structFooDataReader.html#a61a138ab1a8aa7adf766dc4cbdab32dd

If you want to access (read) the Nth element you need to read with `max_samples >= N`. If you want to remove (take) the Nth element, AFAICT, you need to take the preceding samples as well.

This limitation probably has to do with the asynchronous nature of accessing the history cache. Consider this:

- check timestamps of three samples in the history
- decide to take the middle one (index 1)
- (asynchronously) one of the messages is removed from the history cache, or one is added, due to lifespan of a sample expiring or incoming network traffic
- try to take index 1, get something other than you intended

Personally, I've always disliked that queuing was part of the DDS API, because it tightly couples reception of data with ordering it and keeping it alive. The consequnce is that you're limited to ordering that is supported by DDS (presentation order QoS https://community.rti.com/static/documentation/connext-dds/6.0.1/doc/api/connext_dds/api_cpp/structDDS__PresentationQosPolicy.html) and that DDS needs to deserialize your data to check things like key values (which makes it harder for us to get data from DDS without deserializing it first, for things like recording with rosbag). By contrast, in ROS 1 we always took data as it was available and if we need to more complex queuing or presentation order it was done in the user space using `message_filters`. The nice thing about that is that anyone can add new kinds of message filters, e.g. the tf message filter. The upside to letting the middleware manage queuing, is that it can be more tightly integrated with the wire protocol to relay information like deadline missed or instances being disposed, etc...

---

Coming back to your original point/question, the more correct thing to say, I think, is that we have control over the order in which we access each DDS entity, but not how we order data within each, i.e. we can decide to take from one DataReader or another, but not which sample within each DataReader.

That all being said, we can probably do more that what is possible with the plain DDS API when using Fast-RTPS or cycloneDDS, since they're open source, or with Connext with one of its many extensions to the DDS API.

wjwwood:
> However, it is not clear how we pass on the timestamp.

My thought was that it would be added to the `rmw_message_info_t` struct:

https://github.com/ros2/rmw/blob/28f91855510c3c832b500fae40fdcc4169d26a0d/rmw/include/rmw/types.h#L343-L348

You can see we already have a commented out field there. This would be returned when you took the data. However, we should also have a "read" like function that gives you the message this message info struct without taking it out of the history. That is what I imagined the executor would use to make scheduling decisions and also what would be expensive since it needed to call `DataReader::read(...)` in order to get the info.

iluetkeb:
I agree that if you want to implement more complex message selection, such as taking the n'th message while keeping the others, this would be something better done on the user-side, e.g. using message_filters or similar things.

That said, right now we're mainly concerned with two issues:
 1) more than one message is waiting, and we would normally process all of them as soon as they are available, but we're not seeing the second one until after the next call to rmw_wait due to the fact that the wait_set only has binary availability information. therefore, we need information on the number of messages available. last I checked, this should be trivial to acquire by maintaining a count of notifications.
 2) once we get into processing more than one message per subscription during one execution cycle, we want to make sure we're fair. one way of doing that would be to process messages in something close to their sending order. this doesn't need to rely middleware-provided timestamps, we could probably maintain our own within the rmw-layer.

Regarding item 2, I had a look at rmw_fastrtps, and what happens there is that they register a listener to each DataReceiver to get a notification about data availability. Right now, the listener only notifies, but it could certainly be extended to maintain some data internally. This would not be entirely trivial to implement, but certainly fast enough, and doesn't need modifications to Fast-RTPS. If the vendor thinks this is useful more generally, they could later incorporate it into their own code, but we don't have to rely on that.

Personally, I don't think that for our normal, tightly coupled systems with low communication latencies there is much value in using the middleware-provided send timestamp over pure arrival time, but of course, YMMV.

Regarding how we get this information: rmw_message_info_t is passed when `take`ing the data. In execution, we normally defer taking until just prior to invocation of the callback. So, we are looking for a way to get this information *before* taking. Also, we're usually looking for such information per receiver, not per message.

btw, I do think there is value in handling many queueing and QoS things close to the receiving side, if only just to maintain some control over how much memory we need for buffering (across both user-space and the kernel). I also think it is useful to delegate some of the more complex communications issues to people who do this for a living. That said, I have been a bit surprised that our robotics requirements seem to have been a challenge for some of the vendors involved, but thankfully it appears they are learning fast.

MartinCornelis2:
Hi Everyone, 

I merged master with this PR (resolved conflicts) and made some small changes to get the executor to compile again. I haven't had time to thoroughly test the executor functionality, but some quick ros2 topic inspections appear to indicate it is in working condition again.

I am OK with reviewing further additions to this PR, but like stated before don't have time to commit fully to further development on this. 

@alsora had some suggestions to improve the rebuilding of the wait-set and executable list further. Right now we consider too many guard-conditions instead of just the "add stuff to a node"-related ones so there is still some unnecessary rebuilding going on.

Changes and small fixes like the one alsora mentioned can still quite easily be added to the executor before the Foxy release (alsora offered to contribute, I offered to review). Additionally, the static executor in this PR is implemented in such a way that it is optional next to the original executor, so (aside from a slightly bigger install) there is no harm in including it. From @iluetkeb and @wjwwood s input it looks like tackling "proper" scheduling is (currently) out of scope.

**In short:**
In my opinion there is some value in polishing this PR a little bit and including it in Foxy.
A better executor that covers more issues should still be developed for future releases, or alternatively the pool of executors to choose from should be expanded to cover the more common use-cases. In future releases the "old" executor should probably be dropped in favor of executors that perform better or fulfill a specific task better.

Let me know what you guys think. If I am needed for anything please @ me or send me an e-mail. 

Kind regards!

iluetkeb:
@MartinCornelis2 I support including your executor.

Regarding "proper" scheduling: IMHO this is an absolute *MUST* for Foxy. The current situation is very problematic in that it mostly works but breaks down in important edge cases. Everybody I talked to that uses ROS 2 for serious applications is using workarounds.

MartinCornelis2:
Of course scheduling is very important, but do you think it is realistic given the timescale to still get something in before Foxy? Is it possible to implement one of these workarounds you mention before April? 

Additionally, the improved scheduling will cost "some" performance. I think the discussion on performance vs determinism (is determinism the correct term here?) is difficult and (for me) still full of question marks. 

Of course if we only lose a small amount of performance and suddenly have (fully) predictable scheduling then by all means shoot in a PR!

@iluetkeb I think your experience can maybe give some clarity on this subject? Do you know of any fixes that are, like this executor, a good short-term improvement that can be implemented somewhat quickly? Or will this just lead to a half-solution that benefits nobody and will still require a huge rewrite?

iluetkeb:
> Of course scheduling is very important, but do you think it is realistic given the timescale to still get something in before Foxy?

Certainly. It's not that hard. So far, I had assumed OpenRobotics is already working on it, and since this goes across many layers, they are in the best position to do so. 

I personally have some IP-related restrictions that would make it harder for me to do this, but having learned of the situation, I will certainly discuss internally how we could proceed here.

mauropasse:
Just opened a PR with some _StaticSingleThreadedExecutor_ improvements: https://github.com/nobleo/rclcpp/pull/5
Below are the benchmark performances [1] we got so far (measured in RPi 1 - Model A+ v1.1, single core @700Mhz)

Executor | CPU [%] | RAM [Mb] | Late [%] | Too late [%] | Lost [%]
-- | -- | -- | -- | -- | --
SingleThreadedExecutor | 72 | 14.12 | 0.64 | 0 | 0.03
StaticSingleThreadedExecutor | 53.71 | 13.34 | 21.04 | 9.96 | 27.15
StaticSingleThreadedExecutor (improved in the PR) | **40.32** | 13.88 | 0.05 | 0 | 0

The gain in CPU performance is about ~13% compared to the _StaticSingleThreadedExecutor_, and 32% compared to the _SingleThreadedExecutor_.
[1] https://github.com/irobot-ros/ros2-performance/tree/master/performances/benchmark

alsora:
I recently started looking into spurious awakes with the ROS 2 executor.
Apparently, both the standard and the static executors are affected, but for different reasons.
See https://github.com/ros2/rclcpp/issues/1021 and https://github.com/nobleo/static_executor/issues/2 

I want to move the discussion here to get some feedback from more people on the static executor issue.

Results from recent experiments:
 - kernel tracing tools show a lot of context switching
 - I ran the static executor with 1 node with 1 timer and nothing else. I commented out some lines in the executor to not invoke any callback (like [here](https://github.com/nobleo/rclcpp/blob/nobleo/feature/static_executor/rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp#L230) and the other "execute"). This means that this system is only waking up and not doing anything else.
Since there are no messages published/received the spurious awakes do not happen. 
Using a frequency of 100Hz I get a CPU usage of ~5%. Doubling the frequency I get ~8.7%.
This roughly should be the overhead of the double awakes in a very small system.
Note that it will be much bigger in a real system.

I think that this is a relevant issue that must be addressed.
P.S. For comparison, I tried to disable IPC and use the "built-in IPC" of cycloneDDS, but the results were worst, even if double awakes were not present

alsora:
@ivanpauno @wjwwood 
FYI @mauropasse recently opened a PR for the static executor to improve the mechanism for refreshing the executable list
https://github.com/nobleo/rclcpp/pull/6

ros-discourse:

https://discourse.ros.org/t/singlethreadedexecutor-creates-a-high-cpu-overhead-in-ros-2/10077/23

rclcpp/include/rclcpp/executable_list.hpp
```diff
@@ -0,0 +1,68 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
```
wjwwood: I need you guys to claim a copyright at the top of each file. See some examples:

- https://github.com/ros2/rclcpp/blob/d399fef9c6f90462c15b57d9871a5e3c7ceda548/rclcpp/src/rclcpp/waitable.cpp#L1
- https://github.com/ros2/rcl/blob/d45503839ef8db33395130e0fb27222f43602251/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h#L1
- https://github.com/ros2/rosbag2/blob/1370c7f8f12b53598bb4c3ef52cc1b936bd1df85/rosbag2_storage/src/rosbag2_storage/storage_factory.cpp#L1-L2

You can assign it to your company or you can use the `Open Source Robotics Foundation, Inc.`, which is the non-profit that owns a lot of the IP already.

wjwwood: @ishugoel20 specifically.

MartinCornelis2:
@wjwwood, Hi William I just saw your post on Discourse and your additions here. Besides the copyright lines, what is expected from our side?
I myself am not very familiar with the process of getting PRs merged into the core-stack and I'm not entirely certain what to do.

My two main questions are:
1. Are we supposed to do something with https://github.com/ros2/rclcpp/pull/1034 ?
2. Is it still ok if I merge this PR (when I'm done reviewing it) from @alsora @mauropasse https://github.com/nobleo/rclcpp/pull/6 , or do you prefer no more changes at this point?

I don't want to accidentally make changes that impede your work. Or have you wait on an action or response from us (which we might not be aware we are supposed to be taking/giving).

Clear and concise instructions (even if blunt) are much appreciated :)

wjwwood:
> what is expected from our side?
I myself am not very familiar with the process of getting PRs merged into the core-stack and I'm not entirely certain what to do.

Nothing at this point. If I have questions, being available to answer them quickly will help me get it in quicker (and be much appreciated). Sometimes we ask or expect contributors to address test failures and such, but given the accelerated timeline for getting changes in, I'll be editing the pr directly if I can.

> Are we supposed to do something with #1034 ?

You don't have to do anything. You could, however, fix this pull request using that pull request's branch as a source. This pull request currently contains commits from master which typically indicates some sort of issue from past rebases done incorrectly, but I'm not sure in this specific case. I could have fix this one directly, but since you had issues and pull requests on your fork I didn't want to disrupt those without asking.

> Is it still ok if I merge this PR (when I'm done reviewing it) from @alsora @mauropasse nobleo#6 , or do you prefer no more changes at this point?

I'll have to have a look at it, but in general I'm fine to consider additional changes now or later (as a follow up pull request to rclcpp).

> I don't want to accidentally make changes that impede your work.

I made the other pull request out of a similar concern, I wouldn't worry about it, I can work around anything you do.

> Or have you wait on an action or response from us

No, I've just decided I want to get this in and then change it based on my changes to the executor I have prepared rather than the other way around.

rclcpp/include/rclcpp/executable_list.hpp
```diff
@@ -38,7 +38,31 @@ struct ExecutableList
   ExecutableList();
 
   RCLCPP_PUBLIC
-  virtual ~ExecutableList();
+  ~ExecutableList();
```
wjwwood: Why remove this from being `virtual`? It should either be `virtual`, non-virtual but class is marked `final`, or non-virtual but destructor is protected.

I'll change this unless someone speaks up...

rclcpp/src/rclcpp/executable_list.cpp
```diff
@@ -24,3 +24,57 @@ ExecutableList::ExecutableList()
...
+  this->number_of_clients = 0;
+
+  this->waitable.clear();
+  this->number_of_waitables = 0;
+}
+
+void
+ExecutableList::add_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
+{
+  this->subscription.push_back(std::move(subscription));
```
wjwwood: Why move here?

rclcpp/include/rclcpp/executable_list.hpp
wjwwood: @ishugoel20 I absolutely must have these copyright statements to move forward with these pull requests. Please have a look as soon as you can. It should be just a few lines changed once you decide who you're attributing the copyright to.

rclcpp/include/rclcpp/executable_list.hpp
mauropasse: I removed `virtual` because the class is not intended to be derived. The class should be marked `final` as you propose.

rclcpp/src/rclcpp/executable_list.cpp
mauropasse: As `subscription` is passed by value this `std::move` avoids the slow increasing/decreasing of the atomic reference count. We "steal" the `subscription` atomic reference count and we nullify its `shared_ptr`.

wjwwood: But move semantics will cause this to happen anyways. My point was that the move is redundant, in my opinion. I suppose it is fine to leave it, but I do think it is redundant.

rclcpp/include/rclcpp/executable_list.hpp
wjwwood: Ok I'll do that on my pull request.

rclcpp/src/rclcpp/executable_list.cpp
wjwwood: See 6b15351

rclcpp/include/rclcpp/executable_list.hpp
wjwwood: See 6b15351

rclcpp/src/rclcpp/executable_list.cpp
alsora: https://stackoverflow.com/a/29644158/7108533

wjwwood: @alsora I don't know what that answer is supposed to indicate. I know moving it is an optimization, but my point is that the compiler knows `subscription` is a local variable and not used after line 50, so it will move it automatically AFAIK. Due to copy elision https://en.cppreference.com/w/cpp/language/copy_elision, but maybe that's only guaranteed in C++17...

My understanding is that the situation is like this:

```c++
void foo(std::string pong);

void bar(std::string ping) {
  foo(ping);  // this automatically moves s into the input parameter of foo()
  // and further more is preferable to this:
  // foo(std::move(ping));
}

void main(void) {
  std::string something = "something";
  bar(something);  // copy occurs making 'ping', but 'ping' is moved into 'pong'
}
```

But maybe I'm wrong.

alsora: As far as I know, in C++ 17 you are guaranteed that copy elision will happen in that situation, but in C++14 it may or may not happen.

rclcpp/include/rclcpp/executable_list.hpp
ishugoel20: @wjwwood sorry for the late response, I have been busy on other projects. I have added the copyrights to the files added by us. Please let me know if there is something else that I have to do specifically for this PR else @MartinCornelis2 can do the needful.

rclcpp/src/rclcpp/executable_list.cpp
mauropasse: https://www.onlinegdb.com/SysgHtoLL did a quick test, using `C++17` and debugging the pointer.
Without `std::move`, the atomic counter is incremented/decremented.

wjwwood:
Closing since I merged #1034 as an alternative.

Thanks everyone for their contributions and patience with us. Please feel free to open new issues to continue discussions that have been taking place here.

Also, I've been working on other structural changes to the executor design in rclcpp, so these new classes may be update soon and/or subsumed into other classes.

ivanpauno:
There were some clang build warnings that seem to be related with this PR:
- https://ci.ros2.org/job/ci_osx/7981/clang/new/
- https://ci.ros2.org/view/nightly/job/nightly_osx_release/1598/clang-tidy/new/

ros-discourse:

https://discourse.ros.org/t/ros-2-real-time-working-group-online-meeting-6-nov-27-2019-7am-pdt-utc-8/11673/1

PR #878, opened 2019-10-01
use parameters interface not subscribe to events via topic ticketed #609

fujitatomoya:
This is a fix for the following TODO,

https://github.com/ros2/rclcpp/blob/4afd1cd5ad2d0f8ac44e9bdd08860d52c9d73009/rclcpp/src/rclcpp/time_source.cpp#L101-L102

confirmed that use_sim_time can be controlled via parameter interface.

```
[Process-A]
root@6cc2c62d0dd0:~/ros2_colcon_ws# ros2 run demo_nodes_cpp parameter_blackboard 
[INFO] [parameter_blackboard]: Parameter blackboard node named '/parameter_blackboard' ready, and serving '1' parameters already!

[Process-B]
root@6cc2c62d0dd0:~/ros2_colcon_ws# ros2 param get /parameter_blackboard use_sim_time
Boolean value is: False
root@6cc2c62d0dd0:~/ros2_colcon_ws# ros2 param set /parameter_blackboard use_sim_time true
Set parameter successful
root@6cc2c62d0dd0:~/ros2_colcon_ws# ros2 param get /parameter_blackboard use_sim_time
Boolean value is: True
```


rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -105,17 +105,36 @@ void TimeSource::attachNode(
       rclcpp::to_string(use_sim_time_param.get_type()).c_str());
   }
 
-  // TODO(tfoote) use parameters interface not subscribe to events via topic ticketed #609
-  parameter_subscription_ = rclcpp::AsyncParametersClient::on_parameter_event(
```
ivanpauno: This shouldn't be deleted.
`set_on_parameters_set_callback` callbacks are for rejecting/accepting parameter, but not for updating other things based on the result (multiple callbacks can be registered, and you don't know if the parameter will finally be accepted or not).
For that we're using `on_parameter_event`, though it may be good to have a "local" alternative to it (the current approach is creating a subscription).

PR #879, opened 2019-10-01
Enable throttling logs

### LINKED ISSUE #797: rclcpp "throttle" logging
### Feature request:  
`rclcpp` `logging` client library support for `throttle` loggers.

I found a [TODO in the source code](https://github.com/ros2/rclcpp/blob/master/rclcpp/resource/logging.hpp.em#L53-L58), and figured it deserved to be called out in an issue for visibility:
```
# TODO(dhood): Implement the throttle macro using time sources available in rclcpp
excluded_features = ['named', 'throttle']
def is_supported_feature_combination(feature_combination):
    is_excluded = any([ef in feature_combination for ef in excluded_features])
    return not is_excluded
}@
```
This is also sort-of captured in a bullet point tucked away in a nearly-completed ticket in the ros2 repo: https://github.com/ros2/ros2/issues/425

A description of the desired functionality can be found in the [ROS1 wiki](http://wiki.ros.org/roscpp/Overview/Logging):
#### Throttle Logger
```
Throttle

    ROS_DEBUG_THROTTLE(period, ...)

    ROS_DEBUG_STREAM_THROTTLE(period, args)

    ROS_DEBUG_THROTTLE_NAMED(period, name, ...)

    ROS_DEBUG_STREAM_THROTTLE_NAMED(period, name, args) 

Throttled output will print a message at most once per "period".

Delayed throttle

    ROS_DEBUG_DELAYED_THROTTLE(period, ...)

    ROS_DEBUG_STREAM_DELAYED_THROTTLE(period, args)

    ROS_DEBUG_DELAYED_THROTTLE_NAMED(period, name, ...)

    ROS_DEBUG_STREAM_DELAYED_THROTTLE_NAMED(period, name, args) 

Delayed throttled output will print a message at most once per "period" and no message will be printed before one "period" has elapsed.

```

BMarchi:
Fixes #797. As the title says, enables throttling logs.

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8497/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4382/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8350/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/6913/)

### DIFF
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -123,16 +123,15 @@ public:
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
 
+  /// Get current time from the time source specified by Clock::get_clock_type() const noexcept.
   /**
-   * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
-   *
-   * \param now. A pointer to time point value, which will be
+   * \param now A pointer to time point value, which will be
    * set to the current time according to the underlying clock.
    * \return RCUTILS_RET_OK if time was successfully retrieved
    */
   RCLCPP_PUBLIC
   rcutils_ret_t
-  get_now_as_time_point_for_logging(rcutils_time_point_value_t * now);
+  get_now_as_rcutils_time_point(rcutils_time_point_value_t * now);
 
 private:
   // Invoke time jump callback
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -123,18 +123,6 @@ public:
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
 
-  /**
-   * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
-   *
-   * \return RCUTILS_RET_OK if time was successfully retrieved
-   * \param now. A time point value pointer that will contain the
-   * current time from the underlying clock
-   * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
-   */
-  RCLCPP_PUBLIC
-  rcutils_ret_t
-  get_now_as_timepoint(rcutils_time_point_value_t * now);
-
 private:
   // Invoke time jump callback
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -123,16 +123,6 @@ public:
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
 
-  /// Get current time from the time source specified by Clock::get_clock_type() const noexcept.
-  /**
-   * \param now A pointer to time point value, which will be
-   * set to the current time according to the underlying clock.
-   * \return RCUTILS_RET_OK if time was successfully retrieved
-   */
-  RCLCPP_PUBLIC
-  rcutils_ret_t
-  get_now_as_rcutils_time_point(rcutils_time_point_value_t * now);
-
 private:
   // Invoke time jump callback
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -123,6 +123,18 @@ public:
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
 
+  /**
+   * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
+   *
+   * \return RCUTILS_RET_OK if time was successfully retrieved
+   * \param now. A time point value pointer that will contain the
+   * current time from the underlying clock
+   * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
+   */
+  RCLCPP_PUBLIC
+  rcutils_ret_t
+  get_now_as_timepoint(rcutils_time_point_value_t * now);
+
 private:
   // Invoke time jump callback
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -131,7 +131,7 @@ public:
    */
   RCLCPP_PUBLIC
   rcutils_ret_t
-  get_now_as_ns(rcutils_time_point_value_t * now);
+  get_now_as_timepoint(rcutils_time_point_value_t * now);
 
 private:
   // Invoke time jump callback
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -124,9 +124,11 @@ public:
     const rcl_jump_threshold_t & threshold);
 
   /**
-   * Get current time from the time source specified by clock_type as nanoseconds.
+   * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
    *
    * \return RCUTILS_RET_OK if time was successfully retrieved
+   * \param now. A time point value pointer that will contain the
+   * current time from the underlying clock
    * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
    */
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -22,6 +22,8 @@
 #include "rclcpp/visibility_control.hpp"
 
 #include "rcl/time.h"
+#include "rcutils/time.h"
+#include "rcutils/types/rcutils_ret.h"
 
 namespace rclcpp
 {
@@ -120,6 +122,16 @@ public:
     JumpHandler::pre_callback_t pre_callback,
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
+
+  /**
+   * Get current time from the time source specified by clock_type as nanoseconds.
+   *
+   * \return RCUTILS_RET_OK if time was successfully retrieved
+   * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
+   */
+  RCLCPP_PUBLIC
+  rcutils_ret_t
+  get_now_as_ns(rcutils_time_point_value_t * now);
 
 private:
   // Invoke time jump callback
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -126,14 +126,13 @@ public:
   /**
    * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
    *
+   * \param now. A pointer to time point value, which will be
+   * set to the current time according to the underlying clock.
    * \return RCUTILS_RET_OK if time was successfully retrieved
-   * \param now. A time point value pointer that will contain the
-   * current time from the underlying clock
-   * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
    */
   RCLCPP_PUBLIC
   rcutils_ret_t
-  get_now_as_timepoint(rcutils_time_point_value_t * now);
+  get_now_as_time_point_for_logging(rcutils_time_point_value_t * now);
 
 private:
   // Invoke time jump callback
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -44,20 +44,66 @@
 #define RCLCPP_LOG_MIN_SEVERITY RCLCPP_LOG_MIN_SEVERITY_DEBUG
 #endif
 
+/**
+ * \def RCLCPP_LOG_CONDITION_THROTTLE_BEFORE
+ * A macro initializing and checking the `throttle` condition.
+ */
+#define RCLCPP_LOG_CONDITION_THROTTLE_BEFORE(clock, duration) { \
+    static rcl_time_point_value_t __rclcpp_logging_duration = RCUTILS_MS_TO_NS((rcl_time_point_value_t)duration); \
+    static rcl_time_point_value_t __rclcpp_logging_last_logged = 0; \
+    rcl_time_point_value_t __rclcpp_logging_now = clock.now().nanoseconds(); \
+    bool __rclcpp_logging_condition = __rclcpp_logging_now >= __rclcpp_logging_last_logged + __rclcpp_logging_duration; \
+    if (RCUTILS_LIKELY(__rclcpp_logging_condition)) { \
+      __rclcpp_logging_last_logged = __rclcpp_logging_now;
+
 @{
 from collections import OrderedDict
+from rcutils.logging import default_args
 from rcutils.logging import feature_combinations
+from rcutils.logging import Feature
+from rcutils.logging import get_macro_arguments
 from rcutils.logging import get_macro_parameters
 from rcutils.logging import get_suffix_from_features
+from rcutils.logging import name_args
+from rcutils.logging import name_doc_lines
+from rcutils.logging import name_params
 from rcutils.logging import severities
+from rcutils.logging import skipfirst_args
+from rcutils.logging import skipfirst_doc_lines
 from rcutils.logging import throttle_args
+from rcutils.logging import throttle_doc_lines
 from rcutils.logging import throttle_params
 
-throttle_args['condition_before'] = 'RCUTILS_LOG_CONDITION_THROTTLE_BEFORE(clock, duration)'
+default_args['name'] = 'logger.get_name()'
+throttle_args['condition_before'] = 'RCLCPP_LOG_CONDITION_THROTTLE_BEFORE(clock, duration)'
 del throttle_params['get_time_point_value']
 throttle_params['clock'] = 'rclcpp::Clock that will be used to get the time point.'
 throttle_params.move_to_end('clock', last=False)
 
+feature_combinations[('skip_first', 'throttle')] = Feature(
+        params=throttle_params,
+        args={
+            'condition_before': ' '.join([
+                throttle_args['condition_before'],
+                skipfirst_args['condition_before']]),
+            'condition_after': ' '.join([
+                throttle_args['condition_after'], skipfirst_args['condition_after']]),
+        },
+        doc_lines=skipfirst_doc_lines + throttle_doc_lines)
+
+feature_combinations[('skip_first', 'throttle', 'named')] = Feature(
+        params=OrderedDict((*throttle_params.items(), *name_params.items())),
+        args={
+            **{
+                'condition_before': ' '.join([
+                    throttle_args['condition_before'],
+                    skipfirst_args['condition_before']]),
+                'condition_after': ' '.join([
+                    throttle_args['condition_after'],
+                    skipfirst_args['condition_after']]),
+            }, **name_args
+        },
+        doc_lines=skipfirst_doc_lines + throttle_doc_lines + name_doc_lines)
 
 excluded_features = ['named']
 def is_supported_feature_combination(feature_combination):
@@ -101,6 +147,7 @@ def is_supported_feature_combination(feature_combination):
  * It also accepts a single argument of type std::string.
  */
 @{params = get_macro_parameters(feature_combination).keys()}@
+@[ if not 'throttle' in feature_combination]
 #define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in params]))...) \
   do { \
     static_assert( \
@@ -108,7 +155,6 @@ def is_supported_feature_combination(feature_combination):
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
     RCUTILS_LOG_@(severity)@(suffix)_NAMED( \
-@{params = ['clock.get_now_as_timepoint' if p == 'clock' and 'throttle' in feature_combination else p for p in params]}@
 @[ if params]@
 @(''.join(['      ' + p + ', \\\n' for p in params]))@
 @[ end if]@
@@ -116,7 +162,20 @@ def is_supported_feature_combination(feature_combination):
       rclcpp::get_c_string(RCLCPP_FIRST_ARG(__VA_ARGS__, "")), \
         RCLCPP_ALL_BUT_FIRST_ARGS(__VA_ARGS__,"")); \
   } while (0)
-
+@[ else]
+#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in params]))...) \
+  do { \
+    static_assert( \
+      ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
+      typename ::rclcpp::Logger>::value, \
+      "First argument to logging macros must be an rclcpp::Logger"); \
+    RCUTILS_LOG_COND_NAMED( \
+      RCUTILS_LOG_SEVERITY_@(severity), \
+      @(''.join([str(a) + ', ' for a in get_macro_arguments(feature_combination)])) \
+      rclcpp::get_c_string(RCLCPP_FIRST_ARG(__VA_ARGS__, "")), \
+        RCLCPP_ALL_BUT_FIRST_ARGS(__VA_ARGS__,"")); \
+  } while (0)
+@[ end if]@
 @[ end for]@
 #endif
 ///@@}
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -44,66 +44,20 @@
 #define RCLCPP_LOG_MIN_SEVERITY RCLCPP_LOG_MIN_SEVERITY_DEBUG
 #endif
 
-/**
- * \def RCLCPP_LOG_CONDITION_THROTTLE_BEFORE
- * A macro initializing and checking the `throttle` condition.
- */
-#define RCLCPP_LOG_CONDITION_THROTTLE_BEFORE(clock, duration) { \
-    static rcl_time_point_value_t __rclcpp_logging_duration = RCUTILS_MS_TO_NS((rcl_time_point_value_t)duration); \
-    static rcl_time_point_value_t __rclcpp_logging_last_logged = 0; \
-    rcl_time_point_value_t __rclcpp_logging_now = clock.now().nanoseconds(); \
-    bool __rclcpp_logging_condition = __rclcpp_logging_now >= __rclcpp_logging_last_logged + __rclcpp_logging_duration; \
-    if (RCUTILS_LIKELY(__rclcpp_logging_condition)) { \
-      __rclcpp_logging_last_logged = __rclcpp_logging_now;
-
 @{
 from collections import OrderedDict
-from rcutils.logging import default_args
 from rcutils.logging import feature_combinations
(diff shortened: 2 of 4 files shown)


rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,16 @@ class Clock
...
 
+  /**
+   * Get current time from the time source specified by clock_type as nanoseconds.
+   *
+   * \return RCUTILS_RET_OK if time was successfully retrieved
+   * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
+   */
+  RCLCPP_PUBLIC
+  rcutils_ret_t
+  get_now_as_ns(rcutils_time_point_value_t * now);
```
hidmic: @BMarchi nit: why not `get_now_as_timepoint` or even just `timepoint`? The fact that it's just an integer counting nanoseconds is a bit of an implementation detail.

hidmic: Also, should `Clock::now()` use this function instead of duplicating code?

rclcpp/test/test_logging.cpp
```diff
@@ -163,6 +164,38 @@ TEST_F(TestLoggingMacros, test_logging_skipfirst) {
   }
 }
 
+TEST_F(TestLoggingMacros, test_throttle) {
+  using namespace std::chrono_literals;
+  for (uint64_t i = 0; i < 3; ++i) {
+    RCLCPP_DEBUG_THROTTLE(g_logger, rcutils_steady_time_now, 10000, "Throttling");
+  }
+  EXPECT_EQ(1u, g_log_calls);
+  RCLCPP_DEBUG_SKIPFIRST_THROTTLE(g_logger, rcutils_steady_time_now, 1, "Skip first throttling");
```
hidmic: @BMarchi hmm, having to pass an `rcutils` function to an `rclcpp` logging macro is somewhat odd. Consider taking `rclcpp::Clock` instances instead and filling the gap between layers.

rclcpp/include/rclcpp/clock.hpp
dirk-thomas: A name including `now` and `time_point` would be preferred by me too.

rclcpp/resource/logging.hpp.em
```diff
@@ -87,19 +86,25 @@ def is_supported_feature_combination(feature_combination):
 @[ end for]@
  * \param logger The `rclcpp::Logger` to use
 @[ for param_name, doc_line in feature_combinations[feature_combination].params.items()]@
+@[ if not param_name == 'time_source']@
  * \param @(param_name) @(doc_line)
+@[ else]@
+ * \param clock rclcpp::Clock that will be used to get the time point.
+@[ end if]@
```
hidmic: @BMarchi qq, wouldn't updating the feature combination dict be more robust? This will pick up any `param_name` called `time_source`.

rclcpp/resource/logging.hpp.em
```diff
@@ -87,19 +86,25 @@ def is_supported_feature_combination(feature_combination):
...
 @[ end for]@
  * \param ... The format string, followed by the variable arguments for the format string.
  * It also accepts a single argument of type std::string.
  */
-#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in get_macro_parameters(feature_combination).keys()]))...) \
+@{params = get_macro_parameters(feature_combination).keys()}@
+@{params = [p if not p == 'time_source' else 'clock' for p in params]}@
```
hidmic: @BMarchi same as above.

rclcpp/resource/logging.hpp.em
```diff
@@ -87,19 +86,25 @@ def is_supported_feature_combination(feature_combination):
...
+#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in params]))...) \
   do { \
     static_assert( \
       ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
     RCUTILS_LOG_@(severity)@(suffix)_NAMED( \
-@{params = get_macro_parameters(feature_combination).keys()}@
+@{params = [p if not p == 'clock' else 'clock.get_now_as_timepoint' for p in params]}@
```
hidmic: @BMarchi could we make this one a bit more specific, so that it only targets `_THROTTLE_` macros?

rclcpp/test/test_logging.cpp
```diff
@@ -163,6 +164,39 @@ TEST_F(TestLoggingMacros, test_logging_skipfirst) {
...
+  for (uint64_t i = 0; i < 2; ++i) {
+    RCLCPP_DEBUG_THROTTLE(g_logger, ros_clock, 10, "Throttling");
+    if (i == 0) {
+      EXPECT_EQ(5u, g_log_calls);
+      rcl_time_point_value_t clock_ns = ros_clock.now().nanoseconds() + RCUTILS_MS_TO_NS(10);
+      EXPECT_EQ(RCL_RET_OK, rcl_set_ros_time_override(clock, clock_ns));
+    } else {
+      EXPECT_EQ(6u, g_log_calls);
+    }
+  }
```
hidmic: @BMarchi do we need a `for` loop here? I'd rather duplicate line 189.

BMarchi: The throttle is based on file and line number, so if I duplicate the code, it will log twice

hidmic: Of course! Forget about my comment.

rclcpp/resource/logging.hpp.em
BMarchi: Bear in mind that if I modify the dictionaries, I have to recreate all the other methods inside this configure file or modify the global variable, are you okay with that?

rclcpp/resource/logging.hpp.em
```diff
@@ -45,61 +45,65 @@
 #endif
 
 @{
-from rcutils.logging import feature_combinations
-from rcutils.logging import get_macro_parameters
-from rcutils.logging import get_suffix_from_features
-from rcutils.logging import severities
```
dirk-thomas: Why remove all these imports and update their usage?

rclcpp/resource/logging.hpp.em
```diff
@@ -45,61 +45,65 @@
...
+@{params = rcutils.logging.get_macro_parameters(feature_combination).keys()}@
+#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in params]))...) \
   do { \
     static_assert( \
       ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
     RCUTILS_LOG_@(severity)@(suffix)_NAMED( \
-@{params = get_macro_parameters(feature_combination).keys()}@
+@{params = ['clock.get_now_as_timepoint' if p == 'clock' and '_THROTTLE' in suffix else p for p in params]}@
```
hidmic: @BMarchi nit: can we check the feature combination instead of the resulting suffix?

rclcpp/resource/logging.hpp.em
hidmic: I'd say it's a leftover from a previous commit. +1 to rollback these changes.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,16 @@ class Clock
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
 
+  /**
+   * Get current time from the time source specified by clock_type as nanoseconds.
```
hidmic: @BMarchi nit: `clock_type` is out of context here, can we link to `get_clock_type()` method instead?

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,16 @@ class Clock
...
+   *
+   * \return RCUTILS_RET_OK if time was successfully retrieved
```
hidmic: @BMarchi nit: `now` is not documented.

rclcpp/test/test_logging.cpp
```diff
@@ -163,6 +164,39 @@ TEST_F(TestLoggingMacros, test_logging_skipfirst) {
...
+  rclcpp::Clock steady_clock(RCL_STEADY_TIME);
+  for (uint64_t i = 0; i < 3; ++i) {
+    RCLCPP_DEBUG_THROTTLE(g_logger, steady_clock, 10000, "Throttling");
+  }
+  EXPECT_EQ(1u, g_log_calls);
+  RCLCPP_DEBUG_SKIPFIRST_THROTTLE(g_logger, steady_clock, 1, "Skip first throttling");
+  EXPECT_EQ(1u, g_log_calls);
+  for (uint64_t i = 0; i < 3; ++i) {
+    RCLCPP_DEBUG_THROTTLE(g_logger, steady_clock, 10, "Throttling");
+    std::this_thread::sleep_for(10ms);
```
hidmic: @BMarchi shouldn't we call the macro more than 3 times in that same time lapse to verify throttling?

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,18 @@ class Clock
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
 
+  /**
+   * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
+   *
+   * \return RCUTILS_RET_OK if time was successfully retrieved
+   * \param now. A time point value pointer that will contain the
```
wjwwood:
```suggestion
   * \param now A time point value pointer that will contain the
```

wjwwood: Also, nitpick: `\param` should come before `\return` which should be before `\throws`

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,18 @@ class Clock
...
+   * current time from the underlying clock
+   * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
+   */
+  RCLCPP_PUBLIC
+  rcutils_ret_t
+  get_now_as_timepoint(rcutils_time_point_value_t * now);
```
wjwwood: `time_point` is two words and therefore should have an `_` between the two words, e.g. https://en.cppreference.com/w/cpp/chrono/time_point.

rclcpp/include/rclcpp/clock.hpp
wjwwood: Words is a bit off too, maybe this?

> \param now A pointer to time point value, which will be set to the current time according to the underlying clock.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,18 @@ class Clock
...
+  /**
+   * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
+   *
+   * \return RCUTILS_RET_OK if time was successfully retrieved
+   * \param now. A time point value pointer that will contain the
+   * current time from the underlying clock
+   * \throws anything rclcpp::exceptions::throw_from_rcl_error can throw.
+   */
+  RCLCPP_PUBLIC
+  rcutils_ret_t
```
wjwwood: This should not return an `rcutils_ret_t`, this is a C++ function, if it is not `RCUTILS_RET_OK`, then you should throw an exception. In fact the documentation says it will throw an exception.

In which case, I think this function should take no arguments and should return an `rcutils_time_point_value_t` or a C++ equivalent, if there is an appropriate one.

rclcpp/resource/logging.hpp.em
wjwwood: I assume this is resolved? (please resolve comments if they are)

rclcpp/resource/logging.hpp.em
```diff
@@ -45,13 +45,21 @@
...
+from rcutils.logging import throttle_params
 
-# TODO(dhood): Implement the throttle macro using time sources available in rclcpp
-excluded_features = ['named', 'throttle']
+throttle_args['condition_before'] = 'RCUTILS_LOG_CONDITION_THROTTLE_BEFORE(clock, duration)'
+del throttle_params['get_time_point_value']
+throttle_params['clock'] = 'rclcpp::Clock that will be used to get the time point.'
+throttle_params.move_to_end('clock', last=False)
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

wjwwood: One blank line is fine, two is unnecessary.

rclcpp/include/rclcpp/clock.hpp
dirk-thomas: Imo the cost of reimplementing the macros outweighs making this function be more C++-like and throw. I would suggest to instead remove the throw completely from this signature and just return an `rcutils_ret_t`.

If we want to make it more clear that this function is only supposed to be used for the logging macro we could mangle the name to something like: `get_now_as_time_point_for_logging`.

BMarchi: I read this comment after implementing the new macro. What are your thoughts about it now?

dirk-thomas:
Duplicating the logic in the duplicated macro (in 1b78e76c58aab0aac91cc3a6a6d6539a8ce8d7e7) goes against the goal that we want to reuse the logic from `rcutils` (see my other comment: https://github.com/ros2/rclcpp/pull/879#discussion_r333060831). With this custom macro in `rclcpp` we need to also duplicate all testing related to that functionality since it might behave differently than the one implemented in `rcutils`.

hidmic:
@dirk-thomas @wjwwood we have conflicting approaches here: either we introduce a new `rclcpp`-only macro to keep the API C++-like or we reuse `rcutils`' macros and bear with a C-like API. We have to make a choice here. I don't have a strong stand, but the latter is easiest to maintain.

dirk-thomas:
After an offline conversation we decided to use the `rcutils` macro and make the function used from `rclcpp` not throw and return an error code.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,17 @@ class Clock
     JumpHandler::post_callback_t post_callback,
     const rcl_jump_threshold_t & threshold);
 
+  /**
+   * Get current time from the time source specified by Clock::get_clock_type() const noexcept.
```
wjwwood: nitpick: the brief can be on it's own line with `///`, e.g.:

```
/// Get current time from the time source specified by Clock::get_clock_type().
/**
 * \param now ...
 * ...
 */
```

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,17 @@ class Clock
...
+   *
+   * \param now. A pointer to time point value, which will be
```
wjwwood:
```suggestion
   * \param now A pointer to time point value, which will be
```

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -121,6 +123,17 @@ class Clock
...
+   * set to the current time according to the underlying clock.
+   * \return RCUTILS_RET_OK if time was successfully retrieved
+   */
+  RCLCPP_PUBLIC
+  rcutils_ret_t
+  get_now_as_time_point_for_logging(rcutils_time_point_value_t * now);
```
wjwwood: Why is `_for_logging` in the name? If it's to differentiate from returning a C++ time point type, I'd use `get_now_as_rcutils_time_point()` instead. Putting the current user of the function in the name doesn't make much sense to me.

wjwwood: On this point, why not use the existing `now()` method on this class and then call this:

https://github.com/ros2/rclcpp/blob/b8f723708757b0d0b7f29de47544798bafc24a6e/rclcpp/include/rclcpp/time.hpp#L101-L103

It returns an `rcl_time_point_value_t`, but that's the same thing as the rcutils one:

https://github.com/ros2/rcl/blob/e6739884151c302838cfbe908fc9963ef4836f8f/rcl/include/rcl/time.h#L44

That way you don't even need a new method on this class. Right?

wjwwood: (answering my own question, perhaps) Maybe because it could throw? But then you could try-catch in the RCLCPP macro.

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -52,7 +52,7 @@ Clock::now()
 {
   Time now(0, 0, rcl_clock_.type);
 
-  auto ret = rcl_clock_get_now(&rcl_clock_, &now.rcl_time_.nanoseconds);
+  auto ret = get_now_as_time_point_for_logging(&now.rcl_time_.nanoseconds);
```
wjwwood: Why did this change?

rclcpp/resource/logging.hpp.em
```diff
@@ -45,13 +45,20 @@
...
 from rcutils.logging import severities
+from rcutils.logging import throttle_args
+from rcutils.logging import throttle_params
 
-# TODO(dhood): Implement the throttle macro using time sources available in rclcpp
-excluded_features = ['named', 'throttle']
+throttle_args['condition_before'] = 'RCUTILS_LOG_CONDITION_THROTTLE_BEFORE(clock, duration)'
+del throttle_params['get_time_point_value']
+throttle_params['clock'] = 'rclcpp::Clock that will be used to get the time point.'
+throttle_params.move_to_end('clock', last=False)
```
dirk-thomas: It is unfortunate that this requires changing the global state in `rcutils`. It would be good to refactor this in the future that the same can be done without altering it (e.g. using function to generate the state on demand).

rclcpp/include/rclcpp/clock.hpp
BMarchi: > Why is _for_logging in the name?

The name came from Dirk's comment
https://github.com/ros2/rclcpp/pull/879#discussion_r333060831.

If we keep the idea of reusing rcutils macro, I can't use the `now` method along with `nanoseconds` since the [rcutils macro](https://github.com/ros2/rcutils/blob/0815d74809bae0158b8e92ca502017eec5779fa4/resource/logging_macros.h.em#L181) takes a pointer as parameter. The compiler won't like that.

rclcpp/src/rclcpp/clock.cpp
BMarchi: Dirk's suggestion. https://github.com/ros2/rclcpp/pull/879#discussion_r333060831

rclcpp/include/rclcpp/clock.hpp
dirk-thomas: > ... we could mangle the name to something like: `get_now_as_time_point_for_logging`.

I am fine with anything - that a merely a suggestion. Naming it `get_now_as_rcutils_time_point` as William mentioned above is fine with me too.

rclcpp/src/rclcpp/clock.cpp
wjwwood: There's not much code being duplicated... but ok I suppose.

rclcpp/include/rclcpp/clock.hpp
wjwwood: > If we keep the idea of reusing rcutils macro, I can't use the now method along with nanoseconds since the rcutils macro takes a pointer as parameter. The compiler won't like that.

Wait what? Can you not:

```c++
auto rcutils_time_point = clock->now().nanoseconds();
MACRO(&rcutils_time_point);
```

Maybe I don't understand what you mean?

BMarchi: > Wait what? Can you not:

No. My PR for [rcutils](https://github.com/ros2/rcutils/pull/183) makes use of a function rather than a plain value. These two PRS are connected.

wjwwood: Well then this:

```c++
std::function<rcutils_ret_t (rcutils_time_point_value_t *)> get_time_point = [&clock](rcutils_time_point_value_t * time_point) -> rcutils_ret_t {
  *time_point = clock->now().nanoseconds();
}
```

I just don't think we want/need another way to get an `rcl_time_point_value_t`/`rcutils_time_point_value_t` from the Clock/Time objects.

rclcpp/resource/logging.hpp.em
```diff
@@ -92,14 +99,21 @@ def is_supported_feature_combination(feature_combination):
...
+@{params = get_macro_parameters(feature_combination).keys()}@
+#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in params]))...) \
   do { \
     static_assert( \
       ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
+@[ if 'throttle' in feature_combination]@ \
+      std::function<rcutils_ret_t (rcutils_time_point_value_t *)> get_time_point = [&clock](rcutils_time_point_value_t * time_point) -> rcutils_ret_t { \
+        *time_point = clock.now().nanoseconds(); \
```
dirk-thomas: `now()` can throw which needs to be catched here.

rclcpp/resource/logging.hpp.em
```diff
@@ -92,14 +99,21 @@ def is_supported_feature_combination(feature_combination):
...
-#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in get_macro_parameters(feature_combination).keys()]))...) \
+@{params = get_macro_parameters(feature_combination).keys()}@
+#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in params]))...) \
   do { \
     static_assert( \
       ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
+@[ if 'throttle' in feature_combination]@ \
+      std::function<rcutils_ret_t (rcutils_time_point_value_t *)> get_time_point = [&clock](rcutils_time_point_value_t * time_point) -> rcutils_ret_t { \
```
wjwwood: You can use `auto` here, I just used the full type before for demonstration, the type is easily deduced from the lambda declaration.

rclcpp/resource/logging.hpp.em
wjwwood: Agreed.

rclcpp/test/test_logging.cpp
hidmic: I think it still applies. Sleeping for 5ms and duplicating the amount of iterations form 3 to 6 would prove the point.

rclcpp/resource/logging.hpp.em
```diff
@@ -92,14 +99,25 @@ def is_supported_feature_combination(feature_combination):
...
       ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
+@[ if 'throttle' in feature_combination]@ \
+    auto get_time_point = [&clock](rcutils_time_point_value_t * time_point) -> rcutils_ret_t { \
+    try { \
+        *time_point = clock.now().nanoseconds(); \
+    } catch (...) { \
+        return RCUTILS_RET_ERROR; \
+    } \
```
hidmic: @BMarchi nit: watch out for indentation here.

BMarchi:
I didn't run CI because I wanted to run it with all the proper changes. I forgot to sing off one of the commits but I was thinking on rebasing later on so it should go away

rclcpp/resource/logging.hpp.em
```diff
@@ -92,14 +99,25 @@ def is_supported_feature_combination(feature_combination):
...
     static_assert( \
       ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
+@[ if 'throttle' in feature_combination]@ \
+    auto get_time_point = [&clock](rcutils_time_point_value_t * time_point) -> rcutils_ret_t { \
+    try { \
+      *time_point = clock.now().nanoseconds(); \
+    } catch (...) { \
+      return RCUTILS_RET_ERROR; \
```
wjwwood: We should try to print something here, otherwise it silently fails. There is a "safe" write to stderr macro you could use, e.g.:

https://github.com/ros2/rcutils/blob/f8c58a19386a9cf31044a716d7e0946ad75766e8/src/split.c#L115-L117

rclcpp/resource/logging.hpp.em
```diff
@@ -92,14 +99,25 @@ def is_supported_feature_combination(feature_combination):
...
+@{params = get_macro_parameters(feature_combination).keys()}@
+#define RCLCPP_@(severity)@(suffix)(logger, @(''.join([p + ', ' for p in params]))...) \
   do { \
     static_assert( \
       ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
+@[ if 'throttle' in feature_combination]@ \
+    auto get_time_point = [&clock](rcutils_time_point_value_t * time_point) -> rcutils_ret_t { \
+    try { \
```
wjwwood: nitpick: the indentation is wrong here, should be indent once more here and below as it's part of the lambda.

BMarchi:
I rerun CI, windows is having some delay. Linux and aarch are looking good.

rclcpp/resource/logging.hpp.em
```diff
@@ -108,12 +108,13 @@ def is_supported_feature_combination(feature_combination):
...
-    try { \
-      *time_point = clock.now().nanoseconds(); \
-    } catch (...) { \
-      return RCUTILS_RET_ERROR; \
-    } \
-      return RCUTILS_RET_OK; \
+      try { \
+        *time_point = clock.now().nanoseconds(); \
+      } catch (...) { \
+        RCUTILS_SAFE_FWRITE_TO_STDERR("could not get current time stamp\n"); \
```
wjwwood: Please put some context here, imagine you saw this on the command line from your program, you'd have no idea where to investigate.

Either prefix it with the function that is having the issue or include the file name and line number, e.g.:

https://github.com/ros2/rcutils/blob/3daf12617c2bfe7d6eb49c68e2710d7727ef2e6d/src/error_handling.c#L158-L160

hidmic: @wjwwood unlike the function you linked, these are macros and both `__FILE__` and `__LINE__` will be resolved in the sources in which these appear. I'm fine with that but it's not the same as referring to this file. Which one did you meant?

wjwwood: I think you should refer to this file, in the code I linked I did something like `[<pkg>:<file_without_dir.c>]: ....` which is fine. Just some indication as to where this print is coming from. I included the line number because I could but if `__LINE__` will not work correctly in this case, maybe include the macro name instead.

BMarchi: Can you check if it's okay @wjwwood?

Issue #859, opened 2019-09-19
get_actual_qos should return an rclcpp::QoS rather than an rmw_qos_profile_t


wjwwood:
Introduced to the Publisher class in https://github.com/ros2/rclcpp/pull/667, this function really ought to return the rclcpp version of the QoS. Users can always use these methods on the `rclcpp::QoS` class to get the rmw profile still:

https://github.com/ros2/rclcpp/blob/b6d18ccc813d9f35b080795de764b1730702981a/rclcpp/include/rclcpp/qos.hpp#L71-L77

wjwwood:
I'm not sure if we should tick-tock this or just change it directly. I don't think it needs to be backported either way.

PR #883, opened 2019-10-08
make get_actual_qos return a rclcpp::QoS

wjwwood:
Fixes https://github.com/ros2/rclcpp/issues/859

----

Connected pull requests:

- https://github.com/ros2/demos/pull/395
- https://github.com/ros2/ros2_documentation/pull/348

### DIFF
--- a/rclcpp/src/rclcpp/publisher_base.cpp
+++ b/rclcpp/src/rclcpp/publisher_base.cpp
@@ -215,16 +215,7 @@ PublisherBase::get_actual_qos() const
     throw std::runtime_error(msg);
   }
 
-  class RMWQoSIntializedQoS : public rclcpp::QoS
-  {
-  public:
-    explicit
-    RMWQoSIntializedQoS(const rmw_qos_profile_t & rmw_qos):
-    QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos), rmw_qos)
-    {}
-  };
-
-  return RMWQoSIntializedQoS(*qos);
+  return rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(*qos), *qos);
 }
 
 bool
--- a/rclcpp/src/rclcpp/publisher_base.cpp
+++ b/rclcpp/src/rclcpp/publisher_base.cpp
@@ -205,7 +205,7 @@ PublisherBase::get_intra_process_subscription_count() const
   return ipm->get_subscription_count(intra_process_publisher_id_);
 }
 
-rmw_qos_profile_t
+rclcpp::QoS
 PublisherBase::get_actual_qos() const
 {
   const rmw_qos_profile_t * qos = rcl_publisher_get_actual_qos(&publisher_handle_);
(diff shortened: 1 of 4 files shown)


rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -214,7 +214,17 @@ PublisherBase::get_actual_qos() const
...
+  class RMWQoSIntializedQoS : public rclcpp::QoS
+  {
+  public:
+    explicit
+    RMWQoSIntializedQoS(const rmw_qos_profile_t & rmw_qos):
+    QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos), rmw_qos)
+    {}
+  };
+
+  return RMWQoSIntializedQoS(*qos);
```
sloretz: This subclass looks intentional; why is it necessary? Would this have worked too?

```cpp
return rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(*qos), *qos)
```

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -264,7 +274,8 @@ PublisherBase::setup_intra_process(
   const rcl_publisher_options_t & intra_process_options)
 {
   // Intraprocess configuration is not allowed with "durability" qos policy non "volatile".
-  if (this->get_actual_qos().durability != RMW_QOS_POLICY_DURABILITY_VOLATILE) {
+  auto actual_durability = this->get_actual_qos().get_rmw_qos_profile().durability;
```
sloretz: Separate enhancement idea; it might be nice if `QoS` had a getter `rmw_qos_durability_policy_t durability() const`

rclcpp/src/rclcpp/publisher_base.cpp
wjwwood: Oh, for some reason I thought that constructor was protected. I don't need the subclass. I'll update it. Thanks :)

rclcpp/src/rclcpp/publisher_base.cpp
wjwwood: Yeah, I considered adding those now, but I wanted the pr to be small (accessors could be backported, this couldn't). It will add a lot of bloat to the QoS class but maybe it's worth it. However, I imagine cases like this are rare, maybe the only other use case is printing the QoS, but I think adding them is a good idea in general.

rclcpp/src/rclcpp/publisher_base.cpp
wjwwood: 863ead4

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8302/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4253/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6769/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8194/)

wjwwood:
The three test failures (`ament_package` and two in `test_launch_ros`) are unrelated.

The compiler warnings on Windows is related to an issue in rviz, and is unrelated.

Issue #759, opened 2019-06-11
Implement SimpleActionServer and SimpleAction Client


orduno:
#### Feature description
ROS2 is missing an `rclcpp_action::SimpleActionServer` that is equivalent to ROS1  `actionlib::SimpleActionServer`, same for the action client. 

ROS2 does come with some basic examples for implementing action clients and servers, but these don't fully implement the ROS1 `SimpleActionServer` policy.

orduno:
We've created a [simple action server ](nav2_util/include/nav2_util/simple_action_server.hpp) for nav2.

jacobperron:
@orduno The link appears to be broken.

I'm happy to iterate with you if you would like to work on getting this into `rclcpp_action`.

orduno:
@jacobperron Ups should work now.

Yes, let's work together to get something into `rclcpp_action`.

Here's the [PR](https://github.com/ros-planning/navigation2/pull/849) with some additional details.

mkhansen-intel:
I think this is needed. @jacobperron - have you looked at what we have in nav2? We could submit this here instead. 

As a side note, I think this would be useful for projects like MoveIt2, I know MoveIt uses SimpleActionClient & Server. @davetcoleman, care to comment on that?

jacobperron:
I'm not convinced that we need a separate entity like "SimpleActionServer" to achieve what we want.
IIUC, what we want is a single goal policy ([described here](http://wiki.ros.org/actionlib#SimpleActionServer_Goal_Policies)). For better semantics, I would use the name *single-goal action server*.

Depending on ones use-case, achieving this policy is fairly straight-forward (without much boilerplate).
For example, I've recently added what is effectively a single-goal action server to turtlesim. When the action server is created we make it so all goal requests and cancel requests are accepted:

https://github.com/ros/ros_tutorials/blob/cd34759c848955c0d4e3a2e500481d86f1affc62/turtlesim/src/turtle.cpp#L68-L81

And in the accepted callback, we add two lines (plus a log statement) to abort any current goal:

https://github.com/ros/ros_tutorials/blob/cd34759c848955c0d4e3a2e500481d86f1affc62/turtlesim/src/turtle.cpp#L138-L142

This is similar to what is done in nav2's SimpleActionServer, except in the turtlesim scenario we don't need to worry about locks since it uses a single threaded executor.

I can think of some improvements we could make to the existing action server API related to a single-goal policy:

[...]

PR #884, opened 2019-10-08
[rclcpp_action] Add public cancel goal method to server goal handle

jacobperron:
This gives the option to cancel an active goal from the action server implementation.
In the context of action servers with a single-goal policy, this let's the server trigger
a cancel event upon receiving a new goal request.

See related discussion in #759

### DIFF
--- a/rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
+++ b/rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
@@ -60,6 +60,12 @@ public:
   bool
   is_executing() const;
 
+  /// Start canceling the goal.
+  /// \throws rclcpp::exceptions::RCLException if the transition to the canceling state is invalid.
+  RCLCPP_ACTION_PUBLIC
+  void
+  cancel_goal();
+
   RCLCPP_ACTION_PUBLIC
   virtual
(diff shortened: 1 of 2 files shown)


jacobperron:
I'm second guessing this change. The only difference between the CANCELED and ABORTED states is semantics: CANCELED implies the goal was preempted by an action client and ABORTED implies the goal was preempted by the action server. If we open up the possibility for an action server to trigger a cancel event, then it kind of removes the meaning of these states.

wjwwood:
> I'm second guessing this change. The only difference between the CANCELED and ABORTED states is semantics: CANCELED implies the goal was preempted by an action client and ABORTED implies the goal was preempted by the action server. If we open up the possibility for an action server to trigger a cancel event, then it kind of removes the meaning of these states.

I wasn't aware of the distinction. Seems like the server should just abort it? Fine by me.

jacobperron:
> Seems like the server should just abort it? Fine by me.

Yes, though we then leave it up to the user to implement an intermediate state *aborting* if they want to ensure a previous goal is finished before starting the next one. I believe this is what's being done in navigation2's simple action server, but I'm not sure if it's necessary or if it is how common.

Issue #863, opened 2019-09-20
rclcpp_compnents: test_api consistently times out


mjcarroll:
#  Bug report

**Required Info:**

- Operating System:
  - Linux
- Installation type:
  - Source
- Version or commit hash:
  - Dashing
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

* Execute `test_component_manager_api`

#### Expected behavior

Test does not time out

#### Actual behavior

Test times out

#### Additional information

This was originally reported as a build_cop issue: https://github.com/ros2/build_cop/issues/184

It's recently become more flaky, causing issues with PR builders and CI jobs (https://github.com/ros2/rclcpp/pull/857) with an attempted fix here https://github.com/ros2/rclcpp/pull/862

hidmic:
So, it's reported for Fast-RTPS and I cannot reproduce it locally using any other `rmw` implementation i.e. RTI Connext and ADLink Opensplice work just fine. I suspect it has to do with the time it takes for Fast-RTPS to bring up / tear down node state, but I have to dig in further.

hidmic:
Alright, some debugging suggests it's a race. Running just two tests, the first one passes and the second gets as far as sending a service request. The service server never gets that request, spinning times out and the test ends up waiting for an `std::future` that never comes (no pun intended).

I went into Fast-RTPS, looking for global state that could explain an interaction between consecutively run tests but I couldn't find any. Maybe eProsima folks can shed some light.

hidmic:
@richiware maybe?

richiware:
We are backporting fixed issues on FastRTPS 1.9 to [1.8](https://github.com/eProsima/Fast-RTPS/issues/709). ROS2 Dashing is using FastRTPS 1.8.x releases. Can you test with this [branch](https://github.com/eProsima/Fast-RTPS/tree/1.8.x)? It will be next 1.8.2 release. In case it still fails, I will try to reproduce the problem and take a look.

hidmic:
Thanks for reaching out @richiware! But it seems to me you're suggesting that this is specific to Fast-RTPS 1.8.x and, by extension, to Dashing. This is also happening on Eloquent. I just noticed the issue was not clear in this regard -- it's updated now.

hidmic:
[...]

PR #885, opened 2019-10-09
call rclcpp::shutdown() after tests

dirk-thomas:
Might address #863?

### DIFF
--- a/rclcpp_components/test/test_component_manager.cpp
+++ b/rclcpp_components/test/test_component_manager.cpp
@@ -27,6 +27,10 @@ protected:
   {
     rclcpp::init(0, nullptr);
   }
+  static void TearDownTestCase()
+  {
+    rclcpp::shutdown();
+  }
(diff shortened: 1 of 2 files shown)


dirk-thomas:
> pending green CI

CI will pass green but that wouldn't answer the question if this change help for the issue. So someone who was able to reproduce it in the past should test it.

wjwwood:
Calling shutdown at the end shouldn't be required... Which problem do you think this will fix? (I skimmed the other issue, but I don't know why you think only calling init is an issue?)

dirk-thomas:
Without a shutdown the node won't communicate to other entities that it is going down.

wjwwood:
Let's consolidate over here: https://github.com/ros2/rclcpp/issues/863#issuecomment-540217007

dirk-thomas:
If others think this isn't necessary I will close the PR.

PR #887, opened 2019-10-11
0.7.11

ivanpauno:
Preparing release.

Will rebase-merge and push tag after merging this.


nuclearsandwich:
> Will rebase-merge and push tag after merging this.

You can also fast-forward it onto the dashing branch locally and GitHub will recognize the commit and count the PR as merged. I think either way works equally well so do whichever you're comfortable with.

PR #888, opened 2019-10-11
[rclcpp_action] Do not throw exception in action client if take fails

### LINKED ISSUE #886: Action executor doesn't allow spurious client wakeup
ClientBase::execute doesn't gracefully handle a client that suffered a spurious wakeup: in that case the ``rmw_take_response`` returns ``RMW_RET_OK`` with ``taken = false``, which gets converted to ``RCL_RET_CLIENT_TAKE_FAILED``.

The regular client executor in
https://github.com/ros2/rclcpp/blob/b8f723708757b0d0b7f29de47544798bafc24a6e/rclcpp/src/rclcpp/executor.cpp#L404-L420 simply does nothing when this error is returned, but the action client executor throws an exception: https://github.com/ros2/rclcpp/blob/b8f723708757b0d0b7f29de47544798bafc24a6e/rclcpp_action/src/client.cpp#L406-L416

Given that the regular client executor handles them gracefully, I suspect the action one should do the same.

(This is related to https://github.com/ros2/rmw_cyclonedds/issues/47, which is ultimately caused by that exception. I'll do a change there to avoid the spurious wakeups, but thought it worthwhile to point it out here in case it really is not behaving.)

jacobperron:
Fixes #886 

As documented in rcl_action, a return code of RCL_RET_ACTION_CLIENT_TAKE_FAILED does not mean that an error occurred.

### DIFF
--- a/rclcpp_action/src/client.cpp
+++ b/rclcpp_action/src/client.cpp
@@ -388,20 +388,20 @@ ClientBase::execute()
     rcl_ret_t ret = rcl_action_take_feedback(
       pimpl_->client_handle.get(), feedback_message.get());
     pimpl_->is_feedback_ready = false;
-    if (RCL_RET_OK != ret) {
-      rclcpp::exceptions::throw_from_rcl_error(ret, "error taking feedback");
-    } else {
+    if (RCL_RET_OK == ret) {
       this->handle_feedback_message(feedback_message);
+    } else if (RCL_RET_ACTION_CLIENT_TAKE_FAILED != ret) {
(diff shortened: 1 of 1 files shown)


jacobperron:
Testing dependencies of `rclcpp_action`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8340/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4276/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6797/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8235/)

Issue #846, opened 2019-09-04
Creating `rclcpp_action` clients and servers in a `Node` constructor


jdlangs:
## Feature request

Similar to the [issue encountered with SyncParametersClient](https://github.com/ros2/rclcpp/issues/599), `rclcpp_action::create_client` and `create_server` expect an `rclcpp::Node::SharedPtr` when called. This is a problem for any class that inherits from `rclcpp::Node` as it cannot use `shared_from_this()` in its constructor to obtain a `shared_ptr` to pass on.

This can be resolved if `rclcpp_action` provides overloads that take a raw pointer. Since the current implementations directly access the node interfaces instead of keeping the `shared_ptr` around, it looks like another brief overload is all that's needed.

jacobperron:
This is possible using the signature accepting node interfaces. See [this demo code](https://github.com/ros2/demos/blob/97651912ebad439a6bc62c34671cad6a5bfee785/action_tutorials/action_tutorials_cpp/src/fibonacci_action_client.cpp#L39-L44) for an example:

```c++
this->client_ptr_ = rclcpp_action::create_client<Fibonacci>(
  this->get_node_base_interface(),
  this->get_node_graph_interface(),
  this->get_node_logging_interface(),
  this->get_node_waitables_interface(),
  "fibonacci");
```

jdlangs:
Hi @jacobperron, I'm unsure how to interpret you closing this issue since it was a request for a specific enhancement. Are you saying that changing the API to enable the use of raw pointers is not desired / will not be accepted? I'll note that the rclcpp parameter clients also have constructors that take the node interfaces but provide raw pointer overloads as well.

Karsten1987:
I think a templated solution could work here as well.

```c++
template<typename MessageT, typename NodePointerT>
auto create_client(NodePointerT node, std::string name)
{
  // static assert that node indeed is a pointer type
  static_assert(rcpputils::is_pointer<declval(node)>::value);

  return rclcpp::create_client<MessageT>(
    node->get_node_base_interface(),
    node->get_node_graph_interface(),
    node->get_node_logging_interface(),
    node->get_node_waitables_interface(),
    name));
}
```


We applied similar things to [diagnostics](https://github.com/ros/diagnostics/blob/dashing/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp#L365-L382) to take a templated pointer to a node. That template can either be deducted as a raw pointer or shared pointer.

[...]

PR #892, opened 2019-10-14
template node type for rclcpp action server and clients

SteveMacenski:
~Addresses #846~ and has the `rclcpp::Node` as the default template argument so it shouldn't break users of it.

I'd add a second test to work with lifecycle nodes, but I didn't want to add lifecycle as a dependency to actions.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/create_client.hpp
+++ b/rclcpp_action/include/rclcpp_action/create_client.hpp
@@ -97,10 +97,10 @@ create_client(
  * \param[in] group The action client will be added to this callback group.
  *   If `nullptr`, then the action client is added to the default callback group.
  */
-template<typename ActionT>
+template<typename ActionT, typename NodeT = rclcpp::Node>
 typename Client<ActionT>::SharedPtr
 create_client(
-  rclcpp::Node::SharedPtr node,
+  typename NodeT::SharedPtr node,
   const std::string & name,
   rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr)
 {
--- a/rclcpp_action/include/rclcpp_action/create_client.hpp
+++ b/rclcpp_action/include/rclcpp_action/create_client.hpp
@@ -97,10 +97,10 @@ create_client(
  * \param[in] group The action client will be added to this callback group.
  *   If `nullptr`, then the action client is added to the default callback group.
  */
-template<typename ActionT, typename NodeT = rclcpp::Node>
+template<typename ActionT, typename NodeT>
 typename Client<ActionT>::SharedPtr
 create_client(
-  typename NodeT::SharedPtr node,
+  NodeT node,
   const std::string & name,
   rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr)
 {
--- a/rclcpp_action/test/test_client.cpp
+++ b/rclcpp_action/test/test_client.cpp
@@ -272,7 +272,7 @@ TEST_F(TestClient, construction_and_destruction)
 
 TEST_F(TestClient, async_send_goal_no_callbacks)
 {
-  auto action_client = rclcpp_action::create_client<ActionType, rclcpp::Node>(client_node,
+  auto action_client = rclcpp_action::create_client<ActionType>(client_node,
       action_name);
   ASSERT_TRUE(action_client->wait_for_action_server(WAIT_FOR_SERVER_TIMEOUT));
 
--- a/rclcpp_action/test/test_client.cpp
+++ b/rclcpp_action/test/test_client.cpp
@@ -272,8 +272,7 @@ TEST_F(TestClient, construction_and_destruction)
 
 TEST_F(TestClient, async_send_goal_no_callbacks)
 {
-  auto action_client = rclcpp_action::create_client<ActionType>(client_node,
-      action_name);
+  auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
   ASSERT_TRUE(action_client->wait_for_action_server(WAIT_FOR_SERVER_TIMEOUT));
 
   ActionGoal bad_goal;
--- a/rclcpp_action/test/test_client.cpp
+++ b/rclcpp_action/test/test_client.cpp
@@ -272,7 +272,8 @@ TEST_F(TestClient, construction_and_destruction)
 
 TEST_F(TestClient, async_send_goal_no_callbacks)
 {
-  auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
+  auto action_client = rclcpp_action::create_client<ActionType, rclcpp::Node>(client_node,
+      action_name);
   ASSERT_TRUE(action_client->wait_for_action_server(WAIT_FOR_SERVER_TIMEOUT));
 
   ActionGoal bad_goal;
--- a/rclcpp_action/include/rclcpp_action/create_server.hpp
+++ b/rclcpp_action/include/rclcpp_action/create_server.hpp
@@ -128,10 +128,10 @@ create_server(
  * \param group[in] The action server will be added to this callback group.
  *   If `nullptr`, then the action server is added to the default callback group.
  */
-template<typename ActionT, typename NodeT = rclcpp::Node>
+template<typename ActionT, typename NodeT>
 typename Server<ActionT>::SharedPtr
 create_server(
-  typename NodeT::SharedPtr node,
+  NodeT node,
   const std::string & name,
   typename Server<ActionT>::GoalCallback handle_goal,
   typename Server<ActionT>::CancelCallback handle_cancel,
--- a/rclcpp_action/include/rclcpp_action/create_server.hpp
+++ b/rclcpp_action/include/rclcpp_action/create_server.hpp
@@ -128,10 +128,10 @@ create_server(
  * \param group[in] The action server will be added to this callback group.
  *   If `nullptr`, then the action server is added to the default callback group.
  */
-template<typename ActionT>
+template<typename ActionT, typename NodeT = rclcpp::Node>
 typename Server<ActionT>::SharedPtr
 create_server(
-  rclcpp::Node::SharedPtr node,
+  typename NodeT::SharedPtr node,
   const std::string & name,
   typename Server<ActionT>::GoalCallback handle_goal,
   typename Server<ActionT>::CancelCallback handle_cancel,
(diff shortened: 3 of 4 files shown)


Karsten1987:
Why not templating it to any pointer type? There is a type trait within rcpputils which could be used for a static assert. 
But I believe that forcing a shared pointer would still not work to address the shared_from_this() problem within the constructor when used within a class inheriting from rclcpp::Node. Whereas a raw pointer to ‘this’ would. 
I might be wrong though.

SteveMacenski:
@Karsten1987 looking across the repo, all instances I can find of node and message type templating are all base types that are later called with `T::SharedPtr`. I don't see why actions should differ. If this one was templated as a pointer, we'd have to template them all that way for consistency. 

A few examples:
https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/create_service.hpp#L33
https://github.com/ros2/rclcpp/blob/af9ae4a61c24ba207f3db6a32c41e0e52cb29cb9/rclcpp_components/include/rclcpp_components/node_factory_template.hpp#L45
https://github.com/ros2/rclcpp/blob/dfb144d3cbadd3f5147d95ca6fcff7ad871804b5/rclcpp/include/rclcpp/executor.hpp#L169

Similarly, I'm not seeing anywhere in here doing a `static_assert` checking if ptr for creating comms, `create_*`,  it seems inconsistent to do that here and no where else. If you try to give it a non-shared ptr it'll fail to compile so I'm not really afraid of asserting here. 

Similar statements for the `shared_from_this()` - that seems out of scope and the design of a node can be simply done to avoid creating in the constructor, the same way you'd have to for any other overload using the shared ptr. 

I'm trying to resolve the missing feature of `rclcpp_action` not being able to take in various node types to make navigation2 code cleaner. Looking back on the ticket, it looks like they wanted something else - removing the addressing in the PR description

jacobperron:
+1 for a more generic template type. E.g. similar to `rclcpp` publishers and subscriptions:

https://github.com/ros2/rclcpp/blob/8e69b7d505eea36410900eeb9f708f2374215feb/rclcpp/include/rclcpp/create_publisher.hpp#L37-L50

Which passes the node to `get_node_topics_interfaces()` (which has templates for pointer and reference types):

https://github.com/ros2/rclcpp/blob/8e69b7d505eea36410900eeb9f708f2374215feb/rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp#L120-L124

https://github.com/ros2/rclcpp/blob/8e69b7d505eea36410900eeb9f708f2374215feb/rclcpp/include/rclcpp/node_interfaces/get_node_topics_interface.hpp#L133-L138

Assuming the proper template functions already exist for the various interfaces, we should be able to make the change:

```diff
-  rclcpp::Node::SharedPtr node,
+  NodeT node
```

And you can default `NodeT` to `rclcpp::Node::SharedPtr`.

I haven't tried it, but maybe we have to use free functions for getting the interfaces instead of the node methods, e.g. 

```c++
get_node_base_interface(node);
```
instead of
```c++
node->get_node_base_interface();
```

SteveMacenski:
@jacobperron I've read that a few times and I'm not entirely certain if you're suggesting a change or expressing a line of thought. Is there something specific you'd like me to do? I'm hearing maybe change from `node->get_xyz()` to `get_xyz(node)` and templating in the ptr type.

I intentionally didn't template the `SharedPtr` because it seemed more readable from user code to do `rclcpp_action::create_client<rclcpp_lifecycle::LifecycleNode>` than `rclcpp_action::create_client<rclcpp_lifecycle::LifecycleNode::SharedPtr>` but I suppose I don't have a strong opinion beyond just convention I generally follow.

jacobperron:
> I'm not entirely certain if you're suggesting a change or expressing a line of thought. Is there something specific you'd like me to do? 

Sorry, I guess it was a bit of both. I'm no wizard with templates, but I would expect the following definition to work not only for shared pointers, but a variety of pointer and reference types for node-like objects (ie. objects that implement the required node interfaces):

```c++
template<typename ActionT, typename NodeT>
typename Client<ActionT>::SharedPtr
create_client(
  NodeT node,
  const std::string & name,
  rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr)
{
  return create_client<ActionT>(
    rclcpp::node_interfaces::get_node_base_interface(node),
    rclcpp::node_interfaces::get_node_graph_interface(node),
    rclcpp::node_interfaces::get_node_logging_interface(node),
    rclcpp::node_interfaces::get_node_waitables_interface(node),
    name,
    group);
}
```

Then, I think template argument deduction will make it so we don't need to worry about specifying the template type. E.g.

```c++
// 'node' is a node-like thing that could be a pointer or passed in by reference
auto client = rclcpp_action::create_client<MyAction>(node, "action_name");
```

I'd like to see the `create_client` and `create_server` functions updated to look like what I've got above. I haven't tested it, so I'm not sure if it works out-of-the-box with lifecycle nodes.

jacobperron:
I just realized that to do what I proposed requires the existence of the following functions:

* `get_node_graph_interface`
* `get_node_logging_interface`
* `get_node_waitables_interface`

I guess we could add those missing free functions.

rclcpp_action/include/rclcpp_action/create_client.hpp
```diff
@@ -97,10 +97,10 @@ create_client(
  * \param[in] group The action client will be added to this callback group.
  *   If `nullptr`, then the action client is added to the default callback group.
  */
-template<typename ActionT>
+template<typename ActionT, typename NodeT = rclcpp::Node>
```
jacobperron: I don't think the default value is necessary. Template argument deduction should suffice.
```suggestion
template<typename ActionT, typename NodeT>
```

Same for `create_server`.

SteveMacenski:
> I guess we could add those missing free functions.

I leave it up to you, if you feel that's necessary to merge this I can take a crack at it. I wanted to get this in before the Friday freeze so I scoped it pretty low. I can file a ticket to implement the other free methods and revisit this topic as well at a later time. I am also not a templating wizard, but I'm willing to try to become one ;-)

rclcpp_action/include/rclcpp_action/create_client.hpp
```diff
@@ -97,10 +97,10 @@ create_client(
...
 typename Client<ActionT>::SharedPtr
 create_client(
-  rclcpp::Node::SharedPtr node,
+  typename NodeT::SharedPtr node,
```
jacobperron: I think we should drop the shared pointer suffix. Then this function should also work with raw pointers.
```suggestion
  NodeT node,
```

IMO, an even better approach would be to implement what I proposed in https://github.com/ros2/rclcpp/pull/892#issuecomment-542465517, which requires implementing some missing free functions. This can be done in this PR or as a follow-up.

Same for `create_server`.

rclcpp_action/test/test_client.cpp
```diff
@@ -272,7 +272,8 @@ TEST_F(TestClient, construction_and_destruction)
 
 TEST_F(TestClient, async_send_goal_no_callbacks)
 {
-  auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
+  auto action_client = rclcpp_action::create_client<ActionType, rclcpp::Node>(client_node,
```
jacobperron: I may be wrong, but I don't think we need to specify the template argument:
```suggestion
  auto action_client = rclcpp_action::create_client<ActionType>(client_node,
```

SteveMacenski: This is to explicitly provide a working test of the template, you're 100% right  its not necessary

rclcpp_action/include/rclcpp_action/create_client.hpp
SteveMacenski: I'm not necessarily convinced that's true, but I can verify tomorrow

rclcpp_action/test/test_client.cpp
jacobperron: If it's test the construction with a provided template argument, then I suggest adding an extra test to the `construction_and_destruction` test above (L270). It's seems like a more relevant place to test this.

rclcpp_action/include/rclcpp_action/create_client.hpp
SteveMacenski: Agreed, my goals are different than that of the ticket but I'm happy to give it a go to implement the free functions. I'm prefer if it were in another PR so that this isn't touch too many things & miss Friday's cutoff.

Looking over it, I think removing the `SharedPtr`would make sense if your prior comment on implicit deduction works. I would just be irritated as the user of this code if I had to specify the type of ptr each time I created a server & pretty much all nodes are going to be shared. If the deduction without specification doesn't work, I might push back to leave this as shared.

jacobperron: Dropping `SharedPtr` seems to be working for me :slightly_smiling_face:

rclcpp_action/test/test_client.cpp
SteveMacenski: Fair enough. I actually did do that but theres a bug in gtest that doesn't allow multiple template arguments in an assert. Its a known problem google filed and then closed a ticket with `won't fix`

jacobperron: It's okay if  you don't wrap it in an assert macro. Simply calling the method should be enough (if it throws I think we'll notice).

rclcpp_action/include/rclcpp_action/create_client.hpp
SteveMacenski: nice, ok, will play around tomorrow

rclcpp_action/test/test_client.cpp
SteveMacenski: 10-4

SteveMacenski:
Alright cool sounds like I have the concrete action items :+1:

SteveMacenski:
@jacobperron changes made. Everything seems to be fine

rclcpp_action/test/test_client.cpp
```diff
@@ -272,7 +272,8 @@ TEST_F(TestClient, construction_and_destruction)
 
 TEST_F(TestClient, async_send_goal_no_callbacks)
 {
-  auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
+  auto action_client = rclcpp_action::create_client<ActionType>(client_node,
```
jacobperron: Please undo this change.

SteveMacenski: done

SteveMacenski:
@jacobperron :+1: 

Real stickler. I like it.

jacobperron:
> This should be squashed now with these many small commits

We typically do a squash merge :+1:

PR #899, opened 2019-10-23
De-flake tests for rmw_connext

mjcarroll:
The `test_node` tests take a bit longer to execute on Connext, so this

- Removes some duplicated test logic
- Increases the timeout for the test (it took approx 200 seconds to execute locally)


mjcarroll:
Linux CI: [Build Status](https://ci.ros2.org/job/ci_linux/8528/)

PR #902, opened 2019-10-24
documentation and reference fix 

monidzik:
3 small fixes, they are unrelated to each other, but I didn't want to create separate PRs for such small changes, please let me know if it's fine;
Improvements are (as in commits):
- Removing reference from nanoseconds parameter in sleep_for function
- Removing unnecessary parameter in function description in `lifecycle_node.hpp`

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -212,7 +212,7 @@ on_shutdown(std::function<void()> callback, rclcpp::Context::SharedPtr context =
 RCLCPP_PUBLIC
 bool
 sleep_for(
-  const std::chrono::nanoseconds & nanoseconds,
+  const std::chrono::nanoseconds nanoseconds,
   rclcpp::Context::SharedPtr context = nullptr);
 
 /// Safely check if addition will overflow.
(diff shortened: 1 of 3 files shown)


rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -211,7 +211,7 @@ on_shutdown(std::function<void()> callback, rclcpp::Context::SharedPtr context =
 RCLCPP_PUBLIC
 bool
 sleep_for(
-  const std::chrono::nanoseconds & nanoseconds,
+  const std::chrono::nanoseconds nanoseconds,
```
Karsten1987: I am bit confused here. You mention in the description that you make it a reference, but this change looks like you remove the reference? Can you elaborate your motivation behind this change?

> Making nanoseconds parameter a reference in sleep_for function

monidzik: @Karsten1987 sorry for super late response, yes, I made a mistake in a description - I corrected it now.
The motivation here is that the reference prevents the implicit conversion to nanoseconds - if I call the function with for example 1s, it would not work.

monidzik:
@mjcarroll I guess it got a little bit forgotten, can you review it?

rclcpp/include/rclcpp/utilities.hpp
mjcarroll: @monidzik See my comment below about this change.

monidzik:
@mjcarroll You are right, apparently my assumptions were wrong here, thanks for double checking. I'll close this MR, since the documentation fix was already applied.

PR #903, opened 2019-10-24
Extend timeout of test_node test for opensplice

BMarchi:
As the title says, this just extends the timeout for the test_node that it was failing for opensplice.


BMarchi:
Closing since this change was merged already

Issue #893, opened 2019-10-14
`Context` destructor not being called


ivanpauno:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 18.04
- Installation type:
  -  from source
- Version or commit hash:
  -  current master
- DDS implementation:
  -  Fast-RTPS (should be DDS implementation agnostic)
- Client library (if applicable):
  -  rclcpp

#### Steps to reproduce issue

- Add a log here https://github.com/ros2/rclcpp/blob/8e69b7d505eea36410900eeb9f708f2374215feb/rclcpp/src/rclcpp/context.cpp#L41
- Execute any demo, e.g. `ros2 run demo_nodes_cpp talker`
- `ctrl-c` it

#### Expected behavior
The log should be printed after `ctrl-c`.

#### Actual behavior
It's never printed.
https://github.com/ros2/rclcpp/blob/8e69b7d505eea36410900eeb9f708f2374215feb/rclcpp/src/rclcpp/contexts/default_context.cpp#L25-L26

For some reason, the destructor of the static variable is not being call.

#### Additional information

I realized of this while working in https://github.com/ros2/rmw_fastrtps/pull/312.
I'm removing the participant from the domain when the context is destructed.
If that doesn't happen, other Participants will not immediately detect that it's not more available (Participant disposal is much slower than removal).

May be related with https://stackoverflow.com/questions/38510621/destructor-of-a-global-static-variable-in-a-shared-library-is-not-called-on-dlcl, as the example is written as a component https://github.com/ros2/demos/blob/master/demo_nodes_cpp/src/topics/talker.cpp.

ivanpauno:
@wjwwood any idea on how to solve this?

wjwwood:
> @wjwwood any idea on how to solve this?

Not off hand. How are you "printing", because at some point `std::cout` is destroyed and even the stdin/stdout handles are closed, but that's usually after static globals are destructed.

It would be less error prone to set a break point and see if you reach it instead.

> May be related with https://stackoverflow.com/questions/38510621/destructor-of-a-global-static-variable-in-a-shared-library-is-not-called-on-dlcl, as the example is written as a component https://github.com/ros2/demos/blob/master/demo_nodes_cpp/src/topics/talker.cpp.

Does it happen with a non-composable node?

ivanpauno:
> How are you "printing", because at some point std::cout is destroyed and even the stdin/stdout handles are closed, but that's usually after static globals are destructed.
>
[...]

PR #906, opened 2019-10-25
Use weak_ptr to store context in GraphListener

Barry-Xu-2018:


ivanpauno:
Fixes #893.

@wjwwood can you take a look to this one and my comment here https://github.com/ros2/rclcpp/issues/893#issuecomment-546346171.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -162,6 +162,9 @@ Context::shutdown(const std::string & reason)
       ++it;
     }
   }
+  // clear sub contexts
+  std::lock_guard<std::recursive_mutex> lock(sub_contexts_mutex_);
+  sub_contexts_.clear();
```
wjwwood: I don't think this is a safe solution to this issue because things may re-add the sub contexts after shutdown. See my comment:

https://github.com/ros2/rclcpp/issues/893#issuecomment-555170318

Barry-Xu-2018: @wjwwood 

Thanks for your comments.   
Based on your comments https://github.com/ros2/rclcpp/issues/893#issuecomment-555170318, I will use weak pointer to fix this issue

Barry-Xu-2018:
@ivanpauno @wjwwood @fujitatomoya 

According to suggestion https://github.com/ros2/rclcpp/issues/893#issuecomment-555170318, I updated codes to use weak pointer way to fix this issue.  
Please review second commit '5bd65eb22ed71e74488481cbfdb89ec01feea9e3'

fujitatomoya:
@Barry-Xu-2018 

i do agree with @ivanpauno , context owns subcontext, and subcontext refers to context using weak_ptr.

Barry-Xu-2018:
@ivanpauno  @fujitatomoya 

> IIUC, @wjwwood comment ([#893 (comment)](https://github.com/ros2/rclcpp/issues/893#issuecomment-555170318)) was suggesting that the graph listener should keep a weak pointer to the context (i.e. avoid storing this
> 
> https://github.com/ros2/rclcpp/blob/a512d58a4ffe9fcd4bb8d91a73a4f1fbc402047b/rclcpp/include/rclcpp/graph_listener.hpp#L168
> 
> ).
> I think that the context should own the subcontext, and not store a weak pointer to them.

Thanks for your comments. 

If updating GraphListener with
```
rclcpp::Context::WeakPtr parent_context_;
 ```

There might be a problem we need to notice 

``` void GraphListener::__shutdown(bool should_throw) {
   ...
    if (shutdown_guard_condition_) {
      auto parent_context_ptr = parent_context_.lock();
      if (parent_context_ptr) {                              <==== parent_context_ptr will be nullptrl, and then not to call release
        if (should_throw) {
          parent_context_ptr->release_interrupt_guard_condition(&wait_set_);
        } else {
          parent_context_ptr->release_interrupt_guard_condition(&wait_set_, std::nothrow);
        }
      } else {
        RCLCPP_ERROR(
            rclcpp::get_logger("rclcpp"),
            "always called here");
      }
      shutdown_guard_condition_ = nullptr;
    }
```

Because Context is destroyed before GraphListener that is stored in Context::sub_contexts_. ( clear Context::sub_contexts_ before destructor ? It seems like my first solution. )

```
(gdb) bt
#0  rclcpp::graph_listener::GraphListener::~GraphListener (this=0x55555594f540, __in_chrg=<optimized out>) at /root/workspace/src/ros2/rclcpp/rclcpp/src/rclcpp/graph_listener.cpp:68
#1  0x00007ffff75d1dce in std::shared_ptr<rclcpp::graph_listener::GraphListener> rclcpp::Context::get_sub_context<rclcpp::graph_listener::GraphListener, std::shared_ptr<rclcpp::Context> >(std::shared_ptr<rclcpp::Context>&&)::{lambda(rclcpp::graph_listener::GraphListener*)#1}::operator()(rclcpp::graph_listener::GraphListener*) const (__closure=0x55555594f790, 
    sub_context_ptr=0x55555594f540) at /root/workspace/src/ros2/rclcpp/rclcpp/include/rclcpp/context.hpp:313
#2  0x00007ffff75d8b30 in std::_Sp_counted_deleter<rclcpp::graph_listener::GraphListener*, std::shared_ptr<rclcpp::graph_listener::GraphListener> rclcpp::Context::get_sub_context<rclcpp::graph_listener::GraphListener, std::shared_ptr<rclcpp::Context> >(std::shared_ptr<rclcpp::Context>&&)::{lambda(rclcpp::graph_listener::GraphListener*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() (this=0x55555594f780) at /usr/include/c++/7/bits/shared_ptr_base.h:470
#3  0x0000555555567256 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x55555594f780) at /usr/include/c++/7/bits/shared_ptr_base.h:154
#4  0x00005555555668eb in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x5555557f0d78, __in_chrg=<optimized out>)
    at /usr/include/c++/7/bits/shared_ptr_base.h:684
#5  0x0000555555566628 in std::__shared_ptr<void, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x5555557f0d70, __in_chrg=<optimized out>)
    at /usr/include/c++/7/bits/shared_ptr_base.h:1123
#6  0x0000555555566644 in std::shared_ptr<void>::~shared_ptr (this=0x5555557f0d70, __in_chrg=<optimized out>) at /usr/include/c++/7/bits/shared_ptr.h:93
#7  0x00007ffff7585362 in std::pair<std::type_index const, std::shared_ptr<void> >::~pair (this=0x5555557f0d68, __in_chrg=<optimized out>) at /usr/include/c++/7/bits/stl_pair.h:208
#8  0x00007ffff7585382 in __gnu_cxx::new_allocator<std::pair<std::type_index const, std::shared_ptr<void> > >::destroy<std::pair<std::type_index const, std::shared_ptr<void> > > (
    this=0x7fffffffccff, __p=0x5555557f0d68) at /usr/include/c++/7/ext/new_allocator.h:140
#9  0x00007ffff7584665 in std::allocator_traits<std::allocator<std::pair<std::type_index const, std::shared_ptr<void> > > >::destroy<std::pair<std::type_index const, std::shared_ptr<void> > > (__a=..., __p=0x5555557f0d68) at /usr/include/c++/7/bits/alloc_traits.h:487
#10 0x00007ffff7583162 in std::__detail::_Hashtable_alloc<std::allocator<std::__detail::_Hash_node<std::pair<std::type_index const, std::shared_ptr<void> >, false> > >::_M_deallocate_node (this=0x5555557ec328, __n=0x5555557f0d60) at /usr/include/c++/7/bits/hashtable_policy.h:2084
#11 0x00007ffff75818d0 in std::__detail::_Hashtable_alloc<std::allocator<std::__detail::_Hash_node<std::pair<std::type_index const, std::shared_ptr<void> >, false> > >::_M_deallocate_nodes (this=0x5555557ec328, __n=0x0) at /usr/include/c++/7/bits/hashtable_policy.h:2097
#12 0x00007ffff75803ea in std::_Hashtable<std::type_index, std::pair<std::type_index const, std::shared_ptr<void> >, std::allocator<std::pair<std::type_index const, std::shared_ptr<void> > >, std::__detail::_Select1st, std::equal_to<std::type_index>, std::hash<std::type_index>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::clear (this=0x5555557ec328) at /usr/include/c++/7/bits/hashtable.h:2032
#13 0x00007ffff757f620 in std::_Hashtable<std::type_index, std::pair<std::type_index const, std::shared_ptr<void> >, std::allocator<std::pair<std::type_index const, std::shared_ptr<void> > >, std::__detail::_Select1st, std::equal_to<std::type_index>, std::hash<std::type_index>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::~_Hashtable (this=0x5555557ec328, __in_chrg=<optimized out>) at /usr/include/c++/7/bits/hashtable.h:1358
#14 0x00007ffff757f364 in std::unordered_map<std::type_index, std::shared_ptr<void>, std::hash<std::type_index>, std::equal_to<std::type_index>, std::allocator<std::pair<std::type_index const, std::shared_ptr<void> > > >::~unordered_map (this=0x5555557ec328, __in_chrg=<optimized out>) at /usr/include/c++/7/bits/unordered_map.h:101
#15 0x00007ffff757ca8c in rclcpp::Context::~Context (this=0x5555557ec2a0, __in_chrg=<optimized out>) at /root/workspace/src/ros2/rclcpp/rclcpp/src/rclcpp/context.cpp:37
#16 0x00007ffff75872b4 in rclcpp::contexts::default_context::DefaultContext::~DefaultContext (this=0x5555557ec2a0, __in_chrg=<optimized out>)
    at /root/workspace/src/ros2/rclcpp/rclcpp/include/rclcpp/contexts/default_context.hpp:28
#17 0x00007ffff75874cf in __gnu_cxx::new_allocator<rclcpp::contexts::default_context::DefaultContext>::destroy<rclcpp::contexts::default_context::DefaultContext> (this=0x5555557ec2a0, 
    __p=0x5555557ec2a0) at /usr/include/c++/7/ext/new_allocator.h:140
#18 0x00007ffff758749b in std::allocator_traits<std::allocator<rclcpp::contexts::default_context::DefaultContext> >::destroy<rclcpp::contexts::default_context::DefaultContext> (
    __a=..., __p=0x5555557ec2a0) at /usr/include/c++/7/bits/alloc_traits.h:487
#19 0x00007ffff7587385 in std::_Sp_counted_ptr_inplace<rclcpp::contexts::default_context::DefaultContext, std::allocator<rclcpp::contexts::default_context::DefaultContext>, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=0x5555557ec290) at /usr/include/c++/7/bits/shared_ptr_base.h:535
#20 0x0000555555567256 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x5555557ec290) at /usr/include/c++/7/bits/shared_ptr_base.h:154
#21 0x00005555555668eb in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (
    this=0x7ffff79889b8 <rclcpp::contexts::default_context::get_global_default_context()::default_context+8>, __in_chrg=<optimized out>)
    at /usr/include/c++/7/bits/shared_ptr_base.h:684
#22 0x0000555555566336 in std::__shared_ptr<rclcpp::contexts::default_context::DefaultContext, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (
    this=0x7ffff79889b0 <rclcpp::contexts::default_context::get_global_default_context()::default_context>, __in_chrg=<optimized out>) at /usr/include/c++/7/bits/shared_ptr_base.h:1123
#23 0x0000555555566352 in std::shared_ptr<rclcpp::contexts::default_context::DefaultContext>::~shared_ptr (
    this=0x7ffff79889b0 <rclcpp::contexts::default_context::get_global_default_context()::default_context>, __in_chrg=<optimized out>) at /usr/include/c++/7/bits/shared_ptr.h:93
#24 0x00007ffff663d041 in __run_exit_handlers (status=0, listp=0x7ffff69e5718 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:108
#25 0x00007ffff663d13a in __GI_exit (status=<optimized out>) at exit.c:139 ---Type <return> to continue, or q <return> to quit---
#26 0x00007ffff661bb9e in __libc_start_main (main=0x55555556538d <main(int, char**)>, argc=1, argv=0x7fffffffd878, init=<optimized out>, fini=<optimized out>, 
    rtld_fini=<optimized out>, stack_end=0x7fffffffd868) at ../csu/libc-start.c:344
#27 0x00005555555650fa in _start ()
```

ivanpauno:
@Barry-Xu-2018 In the case the context was destructed before the graph listener, there's not need to call `release_interrupt_guard_condition`.

The code should look like:
```cpp
    parent_context = parent_context_.lock();
    if (parent_context && shutdown_guard_condition_) {
      if (should_throw) {
        parent_context_->release_interrupt_guard_condition(&wait_set_);
      } else {
        parent_context_->release_interrupt_guard_condition(&wait_set_, std::nothrow);
      }
      shutdown_guard_condition_ = nullptr;
    }
```

wjwwood:
@ivanpauno well... there may still be a resource ownership issue here. The context owns the guard condition and the wait set in the graph listener uses the guard condition. So by removing strong ownership of the context from the graph listener, the graph listener no longer has ownership of the guard condition it is using. This could lead to the context cleaning up the guard condition during its  destruction and then later the graph listener trying to continue to use the guard condition in correctly...

This is the kind of "gotcha" that I was talking about in the other issue when I said:

> it will need to use a weak pointer rather than a shared pointer to the context **and deal with any tear-down issue that follow.**

ivanpauno:
> @ivanpauno well... there may still be a resource ownership issue here. The context owns the guard condition and the wait set in the graph listener uses the guard condition

The [shutdown_guard_condition_](https://github.com/ros2/rclcpp/blob/a512d58a4ffe9fcd4bb8d91a73a4f1fbc402047b/rclcpp/src/rclcpp/graph_listener.cpp#L58) should still be valid, as the context only clears it if you call [release_interrupt_guard_condition](https://github.com/ros2/rclcpp/blob/20143856714ff766a1352205e58fa19202123afd/rclcpp/src/rclcpp/context.cpp#L236-L250).
It's true that in case you can't call `release_interrupt_guard_condition`, `rcl_guard_condition_fini` should be manually called.

I might not be following correctly your comment, but I don't see the issue.

wjwwood:
> It's true that in case you can't call release_interrupt_guard_condition, rcl_guard_condition_fini should be manually called.

Ok that should be sufficient. I was assuming the context would fini them even if they had not been released.

Barry-Xu-2018:
@ivanpauno 


1. about the below code you mentioned.
  
    ```c++
      parent_context = parent_context_.lock();
      if (parent_context && shutdown_guard_condition_) {
        if (should_throw) {
          parent_context_->release_interrupt_guard_condition(&wait_set_);
        } else {
          parent_context_->release_interrupt_guard_condition(&wait_set_, std::nothrow);
        }
        shutdown_guard_condition_ = nullptr;
      }
    ```

    please note that

    ```
    parent_context = parent_context_.lock();
    ```
    It always get nullptr. So not "in case".


2. If using rcl_guard_condition_fini() instead of Context::release_interrupt_guard_condition() to release shutdown_guard_condition_, it could work.  
    But this make a little confused for user here.

    ```c++
    GraphListener::GraphListener(std::shared_ptr<rclcpp::Context> parent_context)
    {
        shutdown_guard_condition_ = parent_context->get_interrupt_guard_condition(&wait_set_);
    }


    void
    GraphListener::__shutdown(bool should_throw)
    {
      // not to use parent_context_ here, because it always gets nullptr by lock() here
      if (shutdown_guard_condition_) {
        ret = rcl_guard_condition_fini(shutdown_guard_condition_);
        ^^^^^^^^^^^^^^^^^^^^^    to finalize shutdown_guard_condition_ here,  but not to clear the pair value inside Context which was done by interrupt_guard_cond_handles_.erase(kv) of Context::release_interrupt_guard_condition.  (I think semantics is weird here, right?)
        if (RCL_RET_OK != ret) {
          throw_from_rcl_error(ret, "failed to finalize shutdown guard condition");
        }
        shutdown_guard_condition_ = nullptr;
      }

    }

    ```

3. 
    About parameter "should_throw" of  GraphListener::__shutdown, the GraphListener::__shutdown(bool should_throw) will be updated as follows,

    ```c++
    void
    GraphListener::__shutdown(bool should_throw)
    {
      // not to use parent_context_ here, because it always get nullptr by lock() here
      if (shutdown_guard_condition_) {
        ret = rcl_guard_condition_fini(shutdown_guard_condition_);
        if (RCL_RET_OK != ret) {
          if (should_throw) {
              throw_from_rcl_error(ret, "failed to finalize shutdown guard condition");
          } else {
              RCLCPP_ERROR(
                  rclcpp::get_logger("rclcpp"),
                  "failed to finalize shutdown guard condition");
          }
        }
        shutdown_guard_condition_ = nullptr;
      }

    }
    ```

    Is my understanding right?

ivanpauno:
> It always get nullptr. So not "in case".

It always get a `nullptr` if you are using the `GraphListener` subcontext, but if you manually create one no. The `GraphListener` code shouldn't depend on how it's created, so it has to handle both cases (context destructed before and after it).

About the second point, it may be confusing but it's ok. Also, it is less confusing if you handle both cases (which is needed).
About the third point, that's fine.

ivanpauno:
@Barry-Xu-2018 Have you been able to move forward with this?

Barry-Xu-2018:
@ivanpauno 

> @Barry-Xu-2018 Have you been able to move forward with this?

Sorry. Since you do self-assigned, I think you will handle this modification.  
Now I continue to modify codes.

Barry-Xu-2018:
@ivanpauno 

> It always get a nullptr if you are using the GraphListener subcontext, but if you manually create one no. The GraphListener code shouldn't depend on how it's created, so it has to handle both cases (context destructed before and after it).

Sorry, I don't understand your meaning. Could you describe it more clear?  

I supposed you mean manually create GraphListener.  
As I know, there is only one GraphListener in an application even if there are many nodes, and I think Context is the best place to store GraphListener. If so, I can't find a reason why to handle both cases.

ivanpauno:
> Sorry. Since you do self-assigned, I think you will handle this modification.

Ups, it was not the idea. I'm sorry to generate confusion.

> As I know, there is only one GraphListener in an application even if there are many nodes, and I think Context is the best place to store GraphListener. If so, I can't find a reason why to handle both cases.

Yes, we only create it once, and I agree that the context is the best place to store it.
But IMO, it's a bad practice to just handle one case based on the current "external" usage.
It tightly couples the class to the current usage, and may generate strange errors (and headaches) in later refactors.
Handling both cases is safer.

rclcpp/src/rclcpp/graph_listener.cpp
```diff
@@ -80,7 +85,7 @@ GraphListener::start_if_not_started()
       0,  // number_of_clients
       0,  // number_of_services
       0,  // number_of_events
-      this->parent_context_->get_rcl_context().get(),
+      parent_context_ptr->get_rcl_context().get(),
```
ivanpauno: The `rcl` context was stored in `interrupt_guard_condition_context_`.
You can directly use `interrupt_guard_condition_context_.get()`, and completely avoid the `parent_context_` lock checking above.

Barry-Xu-2018: Thanks. I will fix it

Barry-Xu-2018:
@ivanpauno 

> > As I know, there is only one GraphListener in an application even if there are many nodes, and I think Context is the best place to store GraphListener. If so, I can't find a reason why to handle both cases.
> 
> Yes, we only create it once, and I agree that the context is the best place to store it.
> But IMO, it's a bad practice to just handle one case based on the current "external" usage.
> It tightly couples the class to the current usage, and may generate strange errors (and headaches) in later refactors.
> Handling both cases is safer.

Thanks for further explanation. Got it.  
I continue to modify codes.

ivanpauno:
All right, merging! Thanks for the fix @Barry-Xu-2018 !

Issue #904, opened 2019-10-24
Build Farm - Tolerance flakiness in TestMultiThreadedExecutor


piraka9011:
## Bug report

**Required Info:**

- Operating System: OS X
- Installation type: Source (Build Farm)
- Version or commit hash: bf89dc0
- DDS implementation: FastRTPS
- Client library (if applicable): rclcpp

#### Steps to reproduce issue
Run the build farm. TestMultiThreadExecutor fails due to tolerance.

#### Expected behavior
TestMultiThreadExecutor passes.

#### Actual behavior
TestMultiThreadExecutor fails.

A recent build in the [Nightly OS X - Release](https://ci.ros2.org/view/nightly/job/nightly_osx_release/1428/testReport/) failed because the [tolerance](https://github.com/ros2/rclcpp/blob/8cc2a9ba8304b61f240f040bc31414502aaa412c/rclcpp/test/executors/test_multi_threaded_executor.cpp#L78) in the TestMultiThreadExecutor was too low.
Should the tolerance be increased the test is running on an OS X system or is there another way to minimize flakiness?

jacobperron:
Based on the comment [here](https://github.com/ros2/rclcpp/blob/8cc2a9ba8304b61f240f040bc31414502aaa412c/rclcpp/test/executors/test_multi_threaded_executor.cpp#L74-L76), we may consider relaxing the test to check that more than X seconds has passed and not have an upper bound:

```diff
-  if (diff < PERIOD - TOLERANCE || diff > PERIOD + TOLERANCE) {
+  if (diff < PERIOD - TOLERANCE) {

PR #836, opened 2019-08-29
Fix hang with timers in MultiThreadedExecutor (#835)

malsbat:
Signed-off-by: Todd Malsbary <todd.malsbary@intel.com>

PR #907, opened 2019-10-25
Relax multithreaded test constraint

piraka9011:
Address issue #904.
Review by @jacobperron requested.

Signed-off-by: Anas Abou Allaban <aabouallaban@pm.me>

### DIFF
--- a/rclcpp/test/executors/test_multi_threaded_executor.cpp
+++ b/rclcpp/test/executors/test_multi_threaded_executor.cpp
@@ -89,10 +89,9 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
         double diff = std::abs((now - last).nanoseconds()) / 1.0e9;
         last = now;
 
-        if (diff < PERIOD - TOLERANCE || diff > PERIOD + TOLERANCE) {
+        if (diff < PERIOD - TOLERANCE) {
           executor.cancel();
           ASSERT_GT(diff, PERIOD - TOLERANCE);
-          ASSERT_LT(diff, PERIOD + TOLERANCE);
         }
       }
     };


rclcpp/test/executors/test_multi_threaded_executor.cpp
```diff
@@ -89,7 +89,7 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
         double diff = std::abs((now - last).nanoseconds()) / 1.0e9;
         last = now;
 
-        if (diff < PERIOD - TOLERANCE || diff > PERIOD + TOLERANCE) {
+        if (diff < PERIOD - TOLERANCE) {
           executor.cancel();
           ASSERT_GT(diff, PERIOD - TOLERANCE);
           ASSERT_LT(diff, PERIOD + TOLERANCE);
```
jacobperron: This second ASSERT is no longer necessary and can be removed.

piraka9011: Shouldn't it be the first one since the if statement will only execute if `diff` < `PERIOD - TOLERANCE`?
So only `ASSERT_LT(diff, PERIOD - TOLERANCE);`

jacobperron: We want the assertion to fail, which is what the first assertion will do if `diff` < `PERIOD - TOLERANCE`. Unless I'm misunderstanding the test.

piraka9011: timer __over__ take :) Should've inferred that.

jacobperron: I'm not sure if you meant to remove the assert, but it's still there.

piraka9011: Strange, pushed again. Ready for review now.

jacobperron:
@piraka9011 Thanks for PR!

rclcpp/test/executors/test_multi_threaded_executor.cpp
```diff
@@ -89,10 +89,9 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
         double diff = std::abs((now - last).nanoseconds()) / 1.0e9;
         last = now;
 
-        if (diff < PERIOD - TOLERANCE || diff > PERIOD + TOLERANCE) {
+        if (diff < PERIOD - TOLERANCE) {
           executor.cancel();
           ASSERT_GT(diff, PERIOD - TOLERANCE);
```
ivanpauno, line 94: this is first checking `diff < UPPER_BOUND`, and then asserting `diff > UPPER_BOUND`? (with `UPPER_BOUND = PERIOD - TOLERANCE`)
how the second condition can happen after the first check?

this is failing sometimes in CI: https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1858/testReport/junit/rclcpp/TestMultiThreadedExecutor/timer_over_take/.
I'm not sure why, maybe a floating point rounding error?

jacobperron: If I understand my previous comment (https://github.com/ros2/rclcpp/pull/907#discussion_r339241075), I think the intention is that if `diff < UPPER_BOUND`, then we want the test fail. I admit it is confusing, and we could probably clarify with a comment.

ivanpauno: oh, so in that case we want the test to unconditionally fail, right?
I think that either adding a comment or using `FAIL()` would be more clear (now that we're checking only one of the conditions).

jacobperron: Makes sense to me.

Regarding the test failure, I'm still not sure if the bug relies in the test code or not. Having timing dependent tests are tricky.

ivanpauno: > Regarding the test failure, I'm still not sure if the bug relies in the test code or not. Having timing dependent tests are tricky.

I have taken a look, and it seems to be a flaky test (though the timer is triggered MUCH BEFORE than expected).
I've read the original PR https://github.com/ros2/rclcpp/pull/383 that motivated these changes, the only change I see in the multithreaded executor since then is https://github.com/ros2/rclcpp/pull/836, which doesn't seem to be related.

The test seems to be super flaky. I have checked the last few repeated jobs, and it has only failed once.
IMO, it sounds like we can ignore it.

@wjwwood do you think that any of the reset changes can be related to this? or is it just a flaky test?

wjwwood: > @wjwwood do you think that any of the reset changes can be related to this? or is it just a flaky test?

No, not that I can think of. Nothing about the executor really changed. I didn't get to that part yet.

chapulina: >The test seems to be super flaky. I have checked the last few repeated jobs, and it has only failed once.
IMO, it sounds like we can ignore it.

Maybe that's the case for `master`, but on `eloquent` it fails more than it succeeds:

http://build.ros2.org/job/Eci__nightly-connext_ubuntu_bionic_amd64/307/testReport/rclcpp/TestMultiThreadedExecutor/timer_over_take/history/

I'd like to either remove this line or add an `xfail` label to it. What would you prefer?

ivanpauno: I think we should improve the test, instead of that, to check if there's actually a regression or not.
Incrementing both period and tolerance by an order of magnitude should make the test reliable (if there's not a real regression).

I will create a PR today/tomorrow-ish.

ivanpauno: I proposed a PR increasing period/tolerance by an order of magnitude, and it failed too https://github.com/ros2/rclcpp/pull/1105#issuecomment-624757046.

There seems to be a real bug, so it should be triagged/fixed.

ivanpauno: It might be connected with this other failure too https://github.com/ros2/rcl/pull/640.

chapulina: > I proposed a PR

Thanks!

> it should be triagged/fixed.

Until it's fixed, the test should be either commented out or marked as `xfail` so it stops showing on CI results. What do you think?

ivanpauno: I don't know if `xfail` is intended to be used only on flaky tests, or also in tests that are failing due to something being actually broken.

If that mark is also intended for the second case, then it's ok to do it.

chapulina: > If that mark is also intended for the second case, then it's ok to do it.

I believe so: https://github.com/ros2/ros2/issues/900

Issue #913, opened 2019-11-06
Intra-process subscriber does not call get_actual_qos() while publisher does, causing gtest_intra_process to fail


malsbat:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - 2ed445647400bacbff6420cfb8ccc7bee86ee469
- DDS implementation:
  - DPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Run gtest_intra_process__rmw_dps_cpp. 

#### Expected behavior
Test passes.

#### Actual behavior
Test fails.

#### Additional information
The test fails due to can_communicate returning false from the qos reliability check.  The publisher qos is the get_actual_qos from the the RMW layer while the subscriber qos is the one supplied by the application (not get_actual_qos).

ivanpauno:
@malsbat Can you open a PR proposing a fix?

Thanks!

ivanpauno:
I saw you have already opened #914

wjwwood:
Can this be closed?

ivanpauno:
> Can this be closed?

Yes, forgo about this. Closing!

PR #914, opened 2019-11-06
Intra-process subscriber should use RMW actual qos. (ros2#913)

malsbat:
Signed-off-by: Todd Malsbary <todd.malsbary@intel.com>

### DIFF
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -124,15 +124,16 @@ public:
       using rclcpp::detail::resolve_intra_process_buffer_type;
 
       // Check if the QoS is compatible with intra-process.
-      if (qos.get_rmw_qos_profile().history == RMW_QOS_POLICY_HISTORY_KEEP_ALL) {
+      rmw_qos_profile_t qos_profile = get_actual_qos().get_rmw_qos_profile();
+      if (qos_profile.history == RMW_QOS_POLICY_HISTORY_KEEP_ALL) {
         throw std::invalid_argument(
                 "intraprocess communication is not allowed with keep all history qos policy");
       }
-      if (qos.get_rmw_qos_profile().depth == 0) {
+      if (qos_profile.depth == 0) {
         throw std::invalid_argument(
                 "intraprocess communication is not allowed with 0 depth qos policy");
       }
-      if (qos.get_rmw_qos_profile().durability != RMW_QOS_POLICY_DURABILITY_VOLATILE) {
+      if (qos_profile.durability != RMW_QOS_POLICY_DURABILITY_VOLATILE) {
         throw std::invalid_argument(
                 "intraprocess communication allowed only with volatile durability");
       }
@@ -149,7 +150,7 @@ public:
         options.get_allocator(),
         context,
         this->get_topic_name(),    // important to get like this, as it has the fully-qualified name
-        qos.get_rmw_qos_profile(),
+        qos_profile,
         resolve_intra_process_buffer_type(options.intra_process_buffer_type, callback)
         );
 
--- a/rclcpp/test/test_subscription.cpp
+++ b/rclcpp/test/test_subscription.cpp
@@ -287,10 +287,6 @@ static std::vector<TestParameters> invalid_qos_profiles()
     TestParameters(
       rclcpp::QoS(rclcpp::KeepAll()),
       "keep_all_qos"));
-  parameters.push_back(
-    TestParameters(
-      rclcpp::QoS(rclcpp::KeepLast(0)),
-      "keep_last_zero_depth_qos"));
 
   return parameters;
 }


ivanpauno:
In #913 you refer to a test failing, can you copy a link to the test?
I don't know what test are you talking about.

malsbat:
https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_intra_process.cpp is the test that fails.

ivanpauno:
That test isn't failing in any of the nightly ci runs https://ci.ros2.org/view/nightly/.
Can you point out what is different in intel dps rmw implementation? (I understand it's failing with that rmw implementation). It would be good if you can provide the test output.

The changes still look find to me, but I want to make sure I'm not missing something.

malsbat:
The relevant difference here is that the dps rmw does not support all the qos values that e.g. the fastrtps rmw does.  See for example: https://github.com/ros2/rmw_dps/blob/f508310a7a385133bdbceaea11f70a09675a2b53/rmw_dps_cpp/src/rmw_subscription.cpp#L133.

So, in this test the messages fail to get delivered because the requested publisher qos is rewritten to the actual publisher qos supported by dps, while the requested subscriber qos is not, and this causes the can_communicate check to fail.  Once both sides use the actual qos then can_communicate passes and the message is delivered.

The test output when dps fails is the "spin_node_once() - callback (1) expected - try %d/%d\n" repeated until max_loops is reached: the subscription callback is not called.

Let me know if that explains it well enough.  If not I can rebuild before and after this patch and attach the logs.

ivanpauno:
> Let me know if that explains it well enough. If not I can rebuild before and after this patch and attach the logs.

No need, sounds good to me.

> The relevant difference here is that the dps rmw does not support all the qos values that e.g. the fastrtps rmw does. See for example: https://github.com/ros2/rmw_dps/blob/f508310a7a385133bdbceaea11f70a09675a2b53/rmw_dps_cpp/src/rmw_subscription.cpp#L133.

As a side note, I wouldn't recommend to do that in the rmw dps implementation.
Not supported features should fail and log and error, and not be overridden by other setting.
I understand that's more difficult to support, because some tests that are part of the standard ros 2 packages will fail. But just overriding a setting without the user being notified doesn't sound like a good idea to me.

----

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8659/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4501/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7069/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8579/)

malsbat:
> As a side note, I wouldn't recommend to do that in the rmw dps implementation.
> Not supported features should fail and log and error, and not be overridden by other setting.
> I understand that's more difficult to support, because some tests that are part of the standard ros 2 packages will fail. But just overriding a setting without the user being notified doesn't sound like a good idea to me.
> 
Initially it was implemented as you suggest, but with the introduction of rmw_publisher_get_actual_qos() and rmw_subscriber_get_actual_qos(), I understood that the requested and actual could be different.  Am I misunderstanding the intent of those APIs?

ivanpauno:
> Initially it was implemented as you suggest, but with the introduction of rmw_publisher_get_actual_qos() and rmw_subscriber_get_actual_qos(), I understood that the requested and actual could be different. Am I misunderstanding the intent of those APIs?

It was added to figure out what is the actual qos policy used when `RMW_QOS_POLICY_*_SYSTEM_DEFAULT` was set originally (the system default value is rmw implementation dependent). See https://github.com/ros2/rmw/blob/813b94ddd5650444b312304ad156f3b5d9f05e39/rmw/include/rmw/types.h#L263.

It wasn't thought as a way of setting a different qos policy when one is not available.
If you want to do that, I would recommend to at least log a big warning when that happens.

ivanpauno:
@malsbat can you check the test failures in the above CI jobs? Thanks!

malsbat:
> > Initially it was implemented as you suggest, but with the introduction of rmw_publisher_get_actual_qos() and rmw_subscriber_get_actual_qos(), I understood that the requested and actual could be different. Am I misunderstanding the intent of those APIs?
> 
> It was added to figure out what is the actual qos policy used when `RMW_QOS_POLICY_*_SYSTEM_DEFAULT` was set originally (the system default value is rmw implementation dependent). See https://github.com/ros2/rmw/blob/813b94ddd5650444b312304ad156f3b5d9f05e39/rmw/include/rmw/types.h#L263.
> 
> It wasn't thought as a way of setting a different qos policy when one is not available.
> If you want to do that, I would recommend to at least log a big warning when that happens.

Thanks for the clarification, I'll need to look into this a bit more deeply.

malsbat:
> @malsbat can you check the test failures in the above CI jobs? Thanks!

Yes, I'll take a look.

malsbat:
@ivanpauno, I've looked into the test failure and it appears that the test is incorrect, but I'd like your input.

The test that's failing is asserting that a QoS of KEEP_LAST with depth of 0 is an invalid intra-process subscription QoS.  However, depth 0 is defined as RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT in rmw/types.h so appears to me to be a valid QoS.  In fact, the fastrtps rmw checks for and handles RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT explicitly when filling in the QoS supplied to the lower layers.

The test previously passed because the intra-process check explicitly looks for depth 0 and throws.  Prior to this change it would see depth 0 (from the requested QoS), but with this change it now sees the system default depth (actual QoS, non-0) and does not throw.

ivanpauno:
> @ivanpauno, I've looked into the test failure and it appears that the test is incorrect, but I'd like your input.
> 
> The test that's failing is asserting that a QoS of KEEP_LAST with depth of 0 is an invalid intra-process subscription QoS. However, depth 0 is defined as RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT in rmw/types.h so appears to me to be a valid QoS. In fact, the fastrtps rmw checks for and handles RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT explicitly when filling in the QoS supplied to the lower layers.
> 
> The test previously passed because the intra-process check explicitly looks for depth 0 and throws. Prior to this change it would see depth 0 (from the requested QoS), but with this change it now sees the system default depth (actual QoS, non-0) and does not throw.

Sounds good to me, can you update the test?

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8696/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4532/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7103/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8615/)

ivanpauno:
Forgot to check this again, going in! Thanks for the PR @malsbat !

PR #916, opened 2019-11-12
Avoid copying AnySubscriptionCallback objects for intraprocess comms

christophebedard:

`ros2_tracing` and the analysis tools use pointers in order to link a subscription to its `AnySubscriptionCallback` object(s) and callback instances. See the [`Subscription` constructor](https://github.com/ros2/rclcpp/blob/2ed445647400bacbff6420cfb8ccc7bee86ee469/rclcpp/include/rclcpp/subscription.hpp#L163-L170).

The new intraprocess comms (https://github.com/ros2/rclcpp/pull/778) changed this by having the `AnySubscriptionCallback` object actually used by `SubscriptionIntraProcess`. However, the problem is that it creates a copy for itself, and thus we can't link the callback back to the original subscription. I wrote a longer explanation [over on the `ros2_tracing` repo](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/issues/55#note_243141152).

This makes it clear that the `AnySubscriptionCallback` object belongs to the `Subscription` and not `SubscriptionIntraProcess`, even if it used by the latter. However, this could of course be changed. Let me know if that would be preferable.

I'm hoping this can get in in time for eloquent!

### DIFF
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -58,7 +58,7 @@ public:
     >::UniquePtr;
 
   SubscriptionIntraProcess(
-    AnySubscriptionCallback<CallbackMessageT, Alloc> callback,
+    AnySubscriptionCallback<CallbackMessageT, Alloc> & callback,
     std::shared_ptr<Alloc> allocator,
     rclcpp::Context::SharedPtr context,
     const std::string & topic_name,
@@ -153,7 +153,7 @@ private:
     }
   }
 
-  AnySubscriptionCallback<CallbackMessageT, Alloc> any_callback_;
+  AnySubscriptionCallback<CallbackMessageT, Alloc> & any_callback_;
   BufferUniquePtr buffer_;
 };
 
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -145,12 +145,12 @@ public:
           AllocatorT,
           typename MessageUniquePtr::deleter_type
         >>(
-        callback,
+        any_callback_,
         options.get_allocator(),
         context,
         this->get_topic_name(),    // important to get like this, as it has the fully-qualified name
         qos.get_rmw_qos_profile(),
-        resolve_intra_process_buffer_type(options.intra_process_buffer_type, callback)
+        resolve_intra_process_buffer_type(options.intra_process_buffer_type, any_callback_)
         );
 
       // Add it to the intra process manager.


christophebedard:
@hidmic @nuclearsandwich @wjwwood as today is the deadline for bug fixes for Eloquent (IIRC), could this be considered? It fixes something that should have worked after #789.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -153,7 +153,7 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
     }
   }
 
-  AnySubscriptionCallback<CallbackMessageT, Alloc> any_callback_;
+  AnySubscriptionCallback<CallbackMessageT, Alloc> & any_callback_;
```
hidmic: @christophebedard I'm somewhat concerned about the lifetime of `SubscriptionIntraProcess` w.r.t `Subscription` and thus `any_callback_` storage. I see we give away references to `SubscriptionIntraProcess` via `get_intra_process_waitable()`:
https://github.com/ros2/rclcpp/blob/b92db52bb14ec097946bfe44d2b337359dddfbab/rclcpp/src/rclcpp/subscription_base.cpp#L179-L196
Then, `SubscriptionIntraProcess` could outlive its associated `Subscription`. Whether that's bad on its own I can't tell. @wjwwood thoughts?

christophebedard: @hidmic then, considering what you pointed out, we should probably define/clarify who owns the callback object.

If you'd prefer to leave it as it was before (`SubscriptionIntraProcess` having its own copy of the `AnySubscriptionCallback` object, and not giving it a reference), I can simply copy the tracepoints/registration over to `SubscriptionIntraProcess` so that it takes care of it when intraprocess is enabled.

hidmic: > `SubscriptionIntraProcess` having its own copy of the `AnySubscriptionCallback` object, 

That's probably the safest thing to do. The new intra-process communication architecture is under `experimental` so I'd expect changes to happen and I'd rather not constrain future development defining ownership or banning copies.

christophebedard: Got it! I'll submit a different fix soon.

christophebedard: @hidmic done. It's over at #918

christophebedard: @hidmic done: #918

christophebedard: @hidmic done: #918

wjwwood: I agree that we should leave it for now and change the tracing logic instead to avoid lifetime issues. I'll close this in favor of the other one.

wjwwood:
Closing in favor of #918

PR #918, opened 2019-11-13
Put subscription callback tracepoints in intraprocess subscription too

christophebedard:

Alternative to #916, following discussion at https://github.com/ros2/rclcpp/pull/916#discussion_r345920151

In order to get a valid `AnySubscriptionCallback` address for tracing, we can have `SubscriptionIntraProcess` call the tracepoints/perform the registration instead (if it is enabled).

The only downside is that I have to pass the subscription handle through to `SubscriptionIntraProcess`. I think this is the simplest fix, but we can always revisit later (as intra-process subscription is considered experimental anyway).

### DIFF
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -64,8 +64,7 @@ public:
     rclcpp::Context::SharedPtr context,
     const std::string & topic_name,
     rmw_qos_profile_t qos_profile,
-    rclcpp::IntraProcessBufferType buffer_type,
-    rcl_subscription_t * subscription_handle)
+    rclcpp::IntraProcessBufferType buffer_type)
   : SubscriptionIntraProcessBase(topic_name, qos_profile),
     any_callback_(callback)
   {
@@ -93,7 +92,7 @@ public:
 
     TRACEPOINT(
       rclcpp_subscription_callback_added,
-      (const void *)subscription_handle,
+      (const void *)this,
       (const void *)&any_callback_);
     // The callback object gets copied, so if registration is done too early/before this point
     // (e.g. in `AnySubscriptionCallback::set()`), its address won't match any address used later
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -30,6 +30,7 @@
 #include "rclcpp/experimental/subscription_intra_process_base.hpp"
 #include "rclcpp/type_support_decl.hpp"
 #include "rclcpp/waitable.hpp"
+#include "tracetools/tracetools.h"
 
 namespace rclcpp
 {
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -63,7 +63,8 @@ public:
     rclcpp::Context::SharedPtr context,
     const std::string & topic_name,
     rmw_qos_profile_t qos_profile,
-    rclcpp::IntraProcessBufferType buffer_type)
+    rclcpp::IntraProcessBufferType buffer_type,
+    rcl_subscription_t * subscription_handle)
   : SubscriptionIntraProcessBase(topic_name, qos_profile),
     any_callback_(callback)
   {
@@ -88,6 +89,15 @@ public:
     if (RCL_RET_OK != ret) {
       throw std::runtime_error("SubscriptionIntraProcess init error initializing guard condition");
     }
+
+    TRACEPOINT(
+      rclcpp_subscription_callback_added,
+      (const void *)subscription_handle,
+      (const void *)&any_callback_);
+    // The callback object gets copied, so if registration is done too early/before this point
+    // (e.g. in `AnySubscriptionCallback::set()`), its address won't match any address used later
+    // in subsequent tracepoints.
+    any_callback_.register_callback_for_tracing();
   }
 
   bool
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -150,9 +150,12 @@ public:
         context,
         this->get_topic_name(),    // important to get like this, as it has the fully-qualified name
         qos.get_rmw_qos_profile(),
-        resolve_intra_process_buffer_type(options.intra_process_buffer_type, callback),
-        get_subscription_handle().get()
+        resolve_intra_process_buffer_type(options.intra_process_buffer_type, callback)
         );
+      TRACEPOINT(
+        rclcpp_subscription_init,
+        (const void *)get_subscription_handle().get(),
+        (const void *)subscription_intra_process.get());
 
       // Add it to the intra process manager.
       using rclcpp::experimental::IntraProcessManager;
@@ -162,8 +165,12 @@ public:
     }
 
     TRACEPOINT(
-      rclcpp_subscription_callback_added,
+      rclcpp_subscription_init,
       (const void *)get_subscription_handle().get(),
+      (const void *)this);
+    TRACEPOINT(
+      rclcpp_subscription_callback_added,
+      (const void *)this,
       (const void *)&any_callback_);
     // The callback object gets copied, so if registration is done too early/before this point
     // (e.g. in `AnySubscriptionCallback::set()`), its address won't match any address used later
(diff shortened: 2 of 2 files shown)


rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -88,6 +89,15 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
     if (RCL_RET_OK != ret) {
       throw std::runtime_error("SubscriptionIntraProcess init error initializing guard condition");
     }
+
+    TRACEPOINT(
```
wjwwood: Please "include what you use", the tracetools header is not included in this file explicitly.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -88,6 +89,15 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+      rclcpp_subscription_callback_added,
+      (const void *)subscription_handle,
```
wjwwood: Why do you need the handle?

wjwwood: I don't have a lot of context, but maybe there's a better thing to put here like the intra process subscription's address? I'm not sure what's being traced in this case.

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -150,24 +150,25 @@ class Subscription : public SubscriptionBase
...
     }
-
-    TRACEPOINT(
-      rclcpp_subscription_callback_added,
-      (const void *)get_subscription_handle().get(),
-      (const void *)&any_callback_);
-    // The callback object gets copied, so if registration is done too early/before this point
-    // (e.g. in `AnySubscriptionCallback::set()`), its address won't match any address used later
-    // in subsequent tracepoints.
-    any_callback_.register_callback_for_tracing();
```
hidmic: @christophebedard I'm not entirely sure you want to remove this bit. Bear in mind that `Subscription` and `SubscriptionIntraProcess` instances keep separate copies of `AnyCallback` and dispatch to each separately. Don't you want to trace both calls?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
christophebedard: Oops. Totally missed that. Fixed in c8fe4f1f09db798ba067857926d8ecd0efb48b48

rclcpp/include/rclcpp/subscription.hpp
christophebedard: @hidmic it was my understanding that `Subscription`'s copy of `AnySubscriptionCallback` is _never_ used if intra-process is enabled. Thus, in this case, we only need to worry about/register `SubscriptionIntraProcess`'s `AnySubscriptionCallback`'s copy. Is this correct? Or can both be used?

hidmic: `Subscription`'s copy of `AnySubscriptionCallback` is used for inter-process communication, see [here](https://github.com/ros2/rclcpp/blob/b92db52bb14ec097946bfe44d2b337359dddfbab/rclcpp/include/rclcpp/subscription.hpp#L210-L220).

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
christophebedard: @wjwwood this is to link the callback object to the subscription (and on). The subscription itself is linked to its node when initializing the subscription in `rcl` here: https://github.com/ros2/rcl/blob/9034c406c2d5bc4ec291f8e1ad4303cde0184cbc/rcl/src/rcl/subscription.c#L188

This is how we can "link" tracepoint events together to get a big picture when performing analyses. See a simple representation of the output data [here](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/snippets/1912239#L57). Events are in separate dataframes, but we can link a callback instance back to its node, e.g.

* callback instance at line 143
* linked back to its callback object at line 129 through its callback_object (which is the address of the `AnySubscriptionCallback` object)
* linked back to its subscription at line 83 through its subscription handle (`rcl_subscription_t` pointer)
* linked back to its node at line 67 through its node handle (`rcl_node_t` pointer)

We can't know _all_ of this information just by putting a tracepoint here in `SubscriptionIntraProcess`. Let me know if this makes sense.

rclcpp/include/rclcpp/subscription.hpp
christophebedard: I see. In this case, having them use the same `AnySubscriptionCallback` object would really be nice, but like we discussed this is the better solution for now. Fixed with 416423f675efef362eb4ff3971c07469a9a57afb!

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
wjwwood: But would it not make sense to "attach" the rcl subscription handle to the intra process subscription object, and then the any subscription callback to the intra process subscription object?

It feels wrong to me to change the interface of the intra process subscription object just for tracing...

christophebedard: > But would it not make sense to "attach" the rcl subscription handle to the intra process subscription object, and then the any subscription callback to the intra process subscription object?

That does make sense; I think that would work, yes.

So we'd have something like this for both `Subscription` and `SubscriptionIntraProcess`:

* `rcl_subscription_t` handle <=> subscription object
* subscription object <=> `AnySubscriptionCallback` object

And the first step could be done in the `Subscription` constructor for both subscription classes.

> It feels wrong to me to change the interface of the intra process subscription object just for tracing...

Of course, I agree. I was just trying to find a solution :smiley:

christophebedard: @wjwwood the only thing is that it requires a new tracepoint, i.e. a modification to `tracetools` and thus an update to the `ros2_tracing` package for this bug fix.

wjwwood: Unfortunately, I think if `ros2_tracing` is so tightly coupled to the internal layout of rclcpp then we have no choice but to update them together... Maybe there's a better way to decouple them, but I still don't think that's a reason to mutate the rclcpp API if it can be avoided.

christophebedard: @wjwwood for sure. I've made the changes with 03c0a89bd5c674f0909536647a96cf89b27e6307.

This should still pass CI, even if `tracetools`/`ros2_tracing` has not been updated (because the tracepoint calls simply get preprocessed out here, and the tracing tests are only run on the CI over on GitLab). What is the deadline for updating the version of the `ros2_tracing` package used in ros2/ros2 for Eloquent? I'm working on it, but it might not be ready until this weekend.

wjwwood: The hard deadline is Tuesday I believe, but you should aim to get all this in by Monday just in case there are issues.

christophebedard:
@wjwwood I've opened a PR to update the `ros2_tracing` version for the `ros2.repos` file https://github.com/ros2/ros2/pull/823

CI for this PR will fail without that new version, but CI for the above ros2/ros2 PR will probably pass.

wjwwood:
Thanks @christophebedard, we're synchronizing our latest release candidate with the public facing repositories today, so we won't attempt to merge this until this afternoon or tomorrow morning.

We'll probably try to do it this afternoon if at all possible. @mjcarroll, who is the Eloquent ROS Boss, will let me know when/if I can proceed.

wjwwood:
@ros-pull-request-builder test this please

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8663/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4503/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7072/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8583/)

wjwwood:
CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8664/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4504/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7073/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8584/)

christophebedard:
@wjwwood your `.repos` file for all jobs above uses `ros2_tracing` `0.2.9` but should use `0.2.10`!

wjwwood:
Yup, we merged it (prematurely), but hopefully we'll merge this in a few minutes.

wjwwood:
I see what you mean now... Needed to rebuild the repos file, even after merging...

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8666/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4506/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7075/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8586/)

Issue #919, opened 2019-11-13
Action client unnecessarily creates an event every time wait_for_action_server is called


astere-cpr:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - from source
- Version or commit hash:
  - b92db52bb14ec097946bfe44d2b337359dddfbab
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

Create an action client to connect to an existing running action server and call `wait_for_action_server()' in a loop 

#### Expected behavior

The number reported by calling `count_graph_users()' should not increase

#### Actual behavior

The number increases steadily.

#### Additional information

Each call to `wait_for_action_server` adds an event to the node's graph listener event list. If no graph change events occur that vector will increase indefinitely.

jacobperron:
Fixed in #920

PR #920, opened 2019-11-13
issue-919 Fixed "memory leak" in action clients

astere-cpr:
Whenever a call is made to `rclcpp_action::Client::wait_for_action_server`
a weak pointer to an Event object gets added to the graph_event_ vector
of the NodeGraph interface. This vector will be cleared on a node graph
change event, but if no such event occurs the weak pointer will be stuck
in the vector.  Furthermore, if client code issues repeated calls to
`wait_for_action_server` the vector will keep growing.

The fix moves the Event object creation right after the early return from
`wait_for_action_server` so that the Event object is not created in the
case that the server is known to be present and therefore there is no
need to wait for a node graph change event to occur.

### DIFF
--- a/rclcpp_action/src/client.cpp
+++ b/rclcpp_action/src/client.cpp
@@ -165,11 +165,11 @@ ClientBase::wait_for_action_server_nanoseconds(std::chrono::nanoseconds timeout)
   if (!node_ptr) {
     throw rclcpp::exceptions::InvalidNodeError();
   }
-  auto event = node_ptr->get_graph_event();
   // check to see if the server is ready immediately
   if (this->action_server_is_ready()) {
     return true;
   }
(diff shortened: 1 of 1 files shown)


ivanpauno:
Fixes #919

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8669/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4509/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7078/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8589/)

ivanpauno:
Ok, going in. Thanks @astere-cpr for the PR!

PR #921, opened 2019-11-14
Create node clock calls const

SteveMacenski:
Per this issue: https://github.com/ros-planning/navigation2/issues/844

In navigation-land we'd like to make more const methods and this is the link stopping us.


SteveMacenski:
One sec... I messed this up somehow

PR #931, opened 2019-11-29
Add basic nominal test of a client

monidzik:


rclcpp/test/test_client.cpp
```diff
@@ -123,3 +126,41 @@ TEST_F(TestClientSub, construction_and_destruction) {
...
+  auto server_cb = [request_str](const std::shared_ptr<RequestType> request,
+      std::shared_ptr<ResponseType> response) -> void
+  {
+    EXPECT_EQ(request->names.size(), 1U);
+    EXPECT_EQ(request->names.front(), request_str);
+
+    rcl_interfaces::msg::ParameterValue resp;
+    resp.string_value = request_str;
+    response->values.push_back(resp);
+    return;
```
jacobperron: nitpick: you can remove this return statement.

wjwwood:
> Since this test is already covered elsewhere and adding it creates more code to maintain, I'm inclined to reject this PR, unless someone thinks otherwise (@ros2/team)

I don't feel strongly about it, in the past we had the `test_rclcpp` package because it was not possible to make tests that ran against multiple rmw implementations from within the rclcpp package (it was a technical limitation that has since been removed). So in principal it would make sense for many, if not all of those test to get "repatriated" to `rclcpp` so that the tests could live closer to the code. In the mean time, however, there may be new reasons to keep things in `test_rclcpp`, like dependencies we don't want in rclcpp or just keeping the test time down in `rclcpp` itself...

As for duplication, I agree that we should try to avoid it, but I'm ambivalent about having a basic service client test here.

Another dimension to consider is a service client/server test in a single process/node versus in separate processes/nodes, where the latter seems to be more common in `test_rclcpp`.

ivanpauno:
Closing as there's already a test in https://github.com/ros2/system_tests/blob/99ec3f9d03366f521b9f188159ab1acbd3f87017/test_rclcpp/test/test_services_client.cpp.
Tests in `test_rclcpp` could be moved to `rclcpp`, to avoid this confusions.
Opened tracking issue https://github.com/ros2/system_tests/issues/396.

PR #922, opened 2019-11-14
Create node clock calls const (try 2)

SteveMacenski:
Per this issue: ros-planning/navigation2#844

In navigation-land we'd like to make more const methods and this is the link stopping us.

Sorry about that, something in my branching got messed up briefly

PR #932, opened 2019-12-02
Change rclcpp::Node::now() a const member function

mm318:
Adding const qualifier to the `now()` member function, so that it is easier to use.


dirk-thomas:
This change is already part of #922.

PR #942, opened 2019-12-11
New README

maryaB-osr:
It's a bit sparse; is there anything important missing? Happy to add more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ rclcpp provides the accepted, standard C++ API for interacting with ROS 2.
 
 `#include "rclcpp/rclcpp.hpp"` allows use of the most common elements of the ROS 2 system.
 
-Visit the [rclcpp API documentation](http://docs.ros2.org/dashing/api/rclcpp/) for a complete list of its main components.
+Visit the [rclcpp API documentation](http://docs.ros2.org/eloquent/api/rclcpp/) for a complete list of its main components.
 
 ### Examples
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This repository contains the source code for the ROS Client Library for C++ package, included with a standard install of any ROS 2 distro.
 
-rclcpp provides the accepted, standard C++ API for interacting with ROS 2.
+rclcpp provides the standard C++ API for interacting with ROS 2.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,18 @@
...
+
+This repository contains the source code for the ROS Client Library for C++ package, included with a standard install of any ROS 2 distro.
+
+rclcpp provides the accepted, standard C++ API for interacting with ROS 2.
+
+## Usage
+
+`#include "rclcpp/rclcpp.hpp"` allows use of the most common elements of the ROS 2 system.
+
+Visit the [rclcpp API documentation](http://docs.ros2.org/dashing/api/rclcpp/) for a complete list of its main components.
```
Karsten1987: could that maybe point to eloquent at this point? I don't know if docs.ros2.org has a substitution for `latest` or so to keep it also pointing to the latest version of the API doc.

maryaB-osr: I got to the API docs from ros index, which doesn't seem to have eloquent as a distro option on package search result pages. That seems like a problem...

README.md
```diff
@@ -0,0 +1,18 @@
+# rclcpp
+
+This repository contains the source code for the ROS Client Library for C++ package, included with a standard install of any ROS 2 distro.
+
+rclcpp provides the accepted, standard C++ API for interacting with ROS 2.
```
clalancette: I think this would be better without the word "accepted":

```suggestion
rclcpp provides the standard C++ API for interacting with ROS 2.
```

README.md
```diff
@@ -0,0 +1,18 @@
...
+
+`#include "rclcpp/rclcpp.hpp"` allows use of the most common elements of the ROS 2 system.
+
+Visit the [rclcpp API documentation](http://docs.ros2.org/eloquent/api/rclcpp/) for a complete list of its main components.
+
+### Examples
+
+The ROS 2 tutorials [Writing a simple publisher and subscriber](https://index.ros.org/doc/ros2/Tutorials/Writing-A-Simple-Cpp-Publisher-And-Subscriber/)
+and [Writing a simple service and client](https://index.ros.org/doc/ros2/Tutorials/Writing-A-Simple-Cpp-Service-And-Client/)
+contain examples of rclcpp components in use.
```
clalancette: Those examples only show some of the components, and I think "APIs" is maybe a better term.  So this last piece of the sentence could be:

```suggestion
contain some examples of rclcpp APIs in use.
```

README.md
```diff
@@ -0,0 +1,18 @@
...
+
```
clalancette: Nit: trailing whitespace.

maryaB-osr: I don't actually see what you're talking about here. I just made this with the GitHub UI, so I don't know if it's a consequence of that, but when I look at the file's changes there's no line 18, but then when I edit it here, there is a line 18.

clalancette: Huh, weird, you are right.  It shows in the Github UI, but not in the file when I check it out.  OK, that's fine, it was a nit anyway :).

PR #946, opened 2019-12-17
Avoid retaining ownership of the rcl context

ivanpauno:
Fixes https://github.com/ros2/build_cop/issues/254.

See https://github.com/ros2/build_cop/issues/254#issuecomment-566160083.

### DIFF
--- a/rclcpp/src/rclcpp/graph_listener.cpp
+++ b/rclcpp/src/rclcpp/graph_listener.cpp
@@ -146,8 +146,8 @@ GraphListener::run_loop()
     // Ensure that the context doesn't go out of scope.
     auto parent_context = parent_context_.lock();
     if (!parent_context) {
-      // the parent context may be destroyed before this loop is stopped.
-      // in that case, just return silently.
+      // The parent context may be destroyed before this loop is stopped.
+      // In that case, the loop is broken and the function just returns silently.
       return;
     }
 
--- a/rclcpp/src/rclcpp/graph_listener.cpp
+++ b/rclcpp/src/rclcpp/graph_listener.cpp
@@ -140,7 +140,7 @@ GraphListener::run_loop()
     }
     // This lock is released when the loop continues or exits.
     std::lock_guard<std::mutex> nodes_lock(node_graph_interfaces_mutex_, std::adopt_lock);
-    // Ensure that the context doesn't goes out of scope.
+    // Ensure that the context doesn't go out of scope.
     auto parent_context = parent_context_.lock();
 
     // Resize the wait set if necessary.
--- a/rclcpp/src/rclcpp/graph_listener.cpp
+++ b/rclcpp/src/rclcpp/graph_listener.cpp
@@ -39,17 +39,15 @@ GraphListener::GraphListener(std::shared_ptr<rclcpp::Context> parent_context)
 : parent_context_(parent_context),
   is_started_(false),
   is_shutdown_(false),
-  interrupt_guard_condition_context_(nullptr),
   shutdown_guard_condition_(nullptr)
 {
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/graph_listener.cpp
```diff
@@ -39,17 +39,15 @@ GraphListener::GraphListener(std::shared_ptr<rclcpp::Context> parent_context)
 : parent_context_(parent_context),
   is_started_(false),
   is_shutdown_(false),
-  interrupt_guard_condition_context_(nullptr),
   shutdown_guard_condition_(nullptr)
 {
   // TODO(wjwwood): make a guard condition class in rclcpp so this can be tracked
```
jacobperron: What about this comment?

ivanpauno: Yeah, I saw that, and I think that it would simplify this class implementation.
I didn't do that as part of this PR, as this tries to be a minimal backportable change.

rclcpp/src/rclcpp/graph_listener.cpp
```diff
@@ -80,7 +79,7 @@ GraphListener::start_if_not_started()
       0,  // number_of_clients
       0,  // number_of_services
       0,  // number_of_events
-      interrupt_guard_condition_context_.get(),
+      parent_context->get_rcl_context().get(),
```
wjwwood: Is it possible for `parent_context->get_rcl_context()` to return `nullptr` here? Should we not check that?

ivanpauno: Good question, I think it isn't possible except if the `parent_context` was never inited.
If it was inited once, it should still not be a `nullptr` after a call to `shutdown` (as shutdown doesn't clear the rcl context).

The rcl functions are already checking if the context is not null and valid, and returning an approppiate error in those cases.
I think that's enough. If you want I can add another check here, maybe just in the constructor, or in all places that `get_rcl_context` is called.

----

That comment makes me realize that I didn't check if the `parent context` wasn't null after locking.
Adding that.

wjwwood: > That comment makes me realize that I didn't check if the parent context wasn't null after locking.
Adding that.

Did you add that? I don't see any new commits.

ivanpauno: Sorry, I forgot to push the changes :man_facepalming:.

rclcpp/src/rclcpp/graph_listener.cpp
```diff
@@ -142,6 +145,11 @@ GraphListener::run_loop()
     std::lock_guard<std::mutex> nodes_lock(node_graph_interfaces_mutex_, std::adopt_lock);
     // Ensure that the context doesn't go out of scope.
     auto parent_context = parent_context_.lock();
+    if (!parent_context) {
+      // the parent context may be destroyed before this loop is stopped.
+      // in that case, just return silently.
```
wjwwood: nitpick: if a comment has punctuation then it should be a sentence and should have capitalization

```c++
// this is ok
// This is also ok.
// some_variable_name at the beginning makes a lack of capitalization ok too.

// this is weird.
// Weird because not a sentence, actually sentence fragment.
// not weird because not a sentence, so sentence fragments are ok
```

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8963/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4733/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7335/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8889/)

Issue #943, opened 2019-12-11
Implement functions to get publisher and subcription information from topic name


mm318:
#### Feature description

Add an API for obtaining participant information, like QoS policies, of the publishers and subscriptions on a topic name.

#### Implementation considerations

This is essentially implementing for `rclcpp` what was done for `rclpy` in https://github.com/ros2/rclpy/pull/454, which can be used as a reference.

God512:
@mm318 

Do you start to implement it ?
If not, I am interesting to implement it.

Barry-Xu-2018:
@mm318 

Do you start to implement this function ?
If not, I am interesting to implement it.

suab321321:
@ivanpauno @wjwwood  sir I would to contribute to this.

mm318:
We have not started implementation. From our perspective, it is still open for anyone to implement. Thanks!

wjwwood:
I feel like this should be implemented as part of https://github.com/ros2/rmw/pull/186, ideally by the original authors.

I'm still planning on reviewing that set of pull requests, and that would be my first bit of feedback, that we shouldn't merge those until an rclcpp API is proposed. If you guys (@jaisontj and friends) want to work with @Barry-Xu-2018, @suab321321, or others that's fine, but please iterate with them before asking for merge on the other pull requests.

mm318:
The proposed API would just be the addition of `get_publishers_info_by_topic()` and `get_subscriptions_info_by_topic()` member functions under the `rclcpp::Node` class.

I think the most we can do additionally as part of https://github.com/ros2/rmw/pull/186 is to stub out these functions in `rclcpp`, otherwise the set of PRs is becoming unwieldy.

wjwwood:
> The proposed API would just be the addition of get_publishers_info_by_topic() and get_subscriptions_info_by_topic() member functions under the rclcpp::Node class.

I think it should be at least in the `NodeGraphInterface` (https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp), but ideally it should be symmetric to the rclpy changes, including using C++ data structures where appropriate (rather than using the C-style rmw ones).

ivanpauno:
> I'm still planning on reviewing that set of pull requests, and that would be my first bit of feedback, that we shouldn't merge those until an rclcpp API is proposed.

[...]

PR #959, opened 2020-01-03
Stub functions to get publisher and subcription informations like QoS policies from topic name

mm318:
For addressing https://github.com/ros2/rclcpp/issues/943#issuecomment-570355568

This is similar to https://github.com/ros2/rclpy/pull/454, but for `rclcpp` and will have no implementation right now.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
@@ -266,20 +266,20 @@ NodeGraph::count_subscribers(const std::string & topic_name) const
   return count;
 }
 
-std::vector<rmw_topic_info_t>
+std::vector<rmw_topic_endpoint_info_t>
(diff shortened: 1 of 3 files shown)


rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -266,6 +266,22 @@ NodeGraph::count_subscribers(const std::string & topic_name) const
   return count;
 }
 
+std::vector<rmw_topic_info_t>
```
ivanpauno: `rmw_topic_info_t` is too "C style", specially because of the C style strings.
I think that the way to go is creating an equivalent `cpp` class.
It will force some extra unnecessary copying, but performance is not much important in graph instronspection functions.

mm318:
Closing this pull request in favor of https://github.com/ros2/rclcpp/pull/960.

PR #960, opened 2020-01-07
Implement functions to get publisher and subcription informations like QoS policies from topic name

Barry-Xu-2018:
Signed-off-by: Barry Xu <Barry.Xu@sony.com>

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -1084,7 +1084,7 @@ public:
    * ROS topic name conventions.
    *
    * `topic_name` may be a relative, private, or fully qualified topic name.
-   * A relative or private topic will be expanded using this node's namespace and name,
+   * A relative or private topic will be expanded using this node's namespace and name.
    * The queried `topic_name` is not remapped.
    *
    * \param[in] topic_name the topic_name on which to find the publishers.
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -1078,16 +1078,19 @@ public:
    * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
    * profile.
    *
-   * "topic_name" may be a relative, private or fully qualified topic name.
-   * A relative or private topic will be expanded using this node's namespace and name, if
-   * the "no_mangle" param is set to false.
+   * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
+   * name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps).
+   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
+   * ROS topic name conventions.
    *
-   * The queried topic name is not remapped.
+   * `topic_name` may be a relative, private, or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name,
+   * The queried `topic_name` is not remapped.
    *
    * \param[in] topic_name the topic_name on which to find the publishers.
-   * \param[in] no_mangle if false, the given topic name will be expanded
-   *    to its fully qualified name. Default to `false`.
-   * \return a list of TopicInfo representing all the publishers on this topic.
+   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
+   *            otherwise it should be a valid ROS topic name. Defaults to `false`.
+   * \return a list of TopicEndpointInfo representing all the publishers on this topic.
    * \throws InvalidTopicNameError if the given topic_name is invalid.
    * \throws std::runtime_error if internal error happens.
    */
@@ -1101,16 +1104,19 @@ public:
    * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
    * profile.
    *
-   * "topic_name" may be a relative, private or fully qualified topic name.
-   * A relative or private topic will be expanded using this node's namespace and name, if
-   * the "no_mangle" param is set to false.
+   * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
+   * name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps).
+   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
+   * ROS topic name conventions.
    *
-   * The queried topic name is not remapped.
+   * `topic_name` may be a relative, private, or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name.
+   * The queried `topic_name` is not remapped.
    *
    * \param[in] topic_name the topic_name on which to find the subscriptions.
-   * \param[in] no_mangle if false, the given topic name will be expanded
-   *    to its fully qualified name. Default to `false`.
-   * \return a list of TopicInfo representing all the subscriptions on this topic.
+   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
+   *            otherwise it should be a valid ROS topic name. Defaults to `false`.
+   * \return a list of TopicEndpointInfo representing all the subscriptions on this topic.
    * \throws InvalidTopicNameError if the given topic_name is invalid.
    * \throws std::runtime_error if internal error happens.
    */
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -1072,10 +1072,10 @@ public:
   bool
   assert_liveliness() const;
 
-  /// Return a list of publishers about topic endpoint information to a given topic.
+  /// Return the topic endpoint information about publishers on a given topic.
   /**
    * The returned parameter is a list of topic endpoint information, where each item will contain
-   * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
+   * the node name, node namespace, topic type, endpoint type, topic endpoint's GID, and its QoS
    * profile.
    *
    * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
@@ -1089,7 +1089,7 @@ public:
    *
    * \param[in] topic_name the topic_name on which to find the publishers.
    * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
-   *            otherwise it should be a valid ROS topic name. Defaults to `false`.
+   *   otherwise it should be a valid ROS topic name. Defaults to `false`.
    * \return a list of TopicEndpointInfo representing all the publishers on this topic.
    * \throws InvalidTopicNameError if the given topic_name is invalid.
    * \throws std::runtime_error if internal error happens.
@@ -1098,10 +1098,10 @@ public:
   std::vector<rclcpp::TopicEndpointInfo>
   get_publishers_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
 
-  /// Return a list of subscriptions about topic endpoint information to a given topic.
+  /// Return the topic endpoint information about subscriptions on a given topic.
   /**
    * The returned parameter is a list of topic endpoint information, where each item will contain
-   * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
+   * the node name, node namespace, topic type, endpoint type, topic endpoint's GID, and its QoS
    * profile.
    *
    * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
@@ -1115,7 +1115,7 @@ public:
    *
    * \param[in] topic_name the topic_name on which to find the subscriptions.
    * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
-   *            otherwise it should be a valid ROS topic name. Defaults to `false`.
+   *   otherwise it should be a valid ROS topic name. Defaults to `false`.
    * \return a list of TopicEndpointInfo representing all the subscriptions on this topic.
    * \throws InvalidTopicNameError if the given topic_name is invalid.
    * \throws std::runtime_error if internal error happens.
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -1072,6 +1072,54 @@ public:
   bool
   assert_liveliness() const;
 
+  /// Return a list of publishers about topic information to a given topic.
+  /**
+   * The returned parameter is a list of topic info, where each topic info will
+   * contain the node name, node namespace, topic type, participant's GID and its QoS profile.
+   *
+   * "topic_name" may be a relative, private or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name, if
+   * the "no_mangle" param is set to false.
+   *
+   * The queried topic name is not remapped.
+   *
+   * "no_mangle" defaults to false.
+   *
+   * \param[in] topic_name the topic_name on which to find the publishers.
+   * \param[in] no_mangle if false, the given topic name will be expanded
+   *    to its fully qualified name.
+   * \return a list of TopicInfo representing all the publishers on this topic.
+   * \throws InvalidTopicNameError if the given topic_name is invalid.
+   * \throws std::runtime_error if internal error happens.
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::TopicInfo>
+  get_publishers_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
+
+  /// Return a list of subscriptions about topic information to a given topic.
+  /**
+   * The returned parameter is a list of topic info, where each topic info will
+   * contain the node name, node namespace, topic type, participant's GID and its QoS profile.
+   *
+   * "topic_name" may be a relative, private or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name, if
+   * the "no_mangle" param is set to false.
+   *
+   * The queried topic name is not remapped.
+   *
+   * "no_mangle" defaults to false.
+   *
+   * \param[in] topic_name the topic_name on which to find the subscriptions.
+   * \param[in] no_mangle if false, the given topic name will be expanded
+   *    to its fully qualified name.
+   * \return a list of TopicInfo representing all the subscriptions on this topic.
+   * \throws InvalidTopicNameError if the given topic_name is invalid.
+   * \throws std::runtime_error if internal error happens.
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::TopicInfo>
+  get_subscriptions_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
+
 protected:
   /// Construct a sub-node, which will extend the namespace of all entities created with it.
   /**
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -1072,10 +1072,11 @@ public:
   bool
   assert_liveliness() const;
 
-  /// Return a list of publishers about topic information to a given topic.
+  /// Return a list of publishers about topic endpoint information to a given topic.
   /**
-   * The returned parameter is a list of topic info, where each topic info will
-   * contain the node name, node namespace, topic type, participant's GID and its QoS profile.
+   * The returned parameter is a list of topic endpoint information, where each item will contain
+   * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
+   * profile.
    *
    * "topic_name" may be a relative, private or fully qualified topic name.
    * A relative or private topic will be expanded using this node's namespace and name, if
@@ -1083,23 +1084,22 @@ public:
    *
    * The queried topic name is not remapped.
    *
-   * "no_mangle" defaults to false.
-   *
    * \param[in] topic_name the topic_name on which to find the publishers.
    * \param[in] no_mangle if false, the given topic name will be expanded
-   *    to its fully qualified name.
+   *    to its fully qualified name. Default to `false`.
    * \return a list of TopicInfo representing all the publishers on this topic.
    * \throws InvalidTopicNameError if the given topic_name is invalid.
    * \throws std::runtime_error if internal error happens.
    */
   RCLCPP_PUBLIC
-  std::vector<rclcpp::TopicInfo>
+  std::vector<rclcpp::TopicEndpointInfo>
   get_publishers_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
 
-  /// Return a list of subscriptions about topic information to a given topic.
+  /// Return a list of subscriptions about topic endpoint information to a given topic.
   /**
-   * The returned parameter is a list of topic info, where each topic info will
-   * contain the node name, node namespace, topic type, participant's GID and its QoS profile.
+   * The returned parameter is a list of topic endpoint information, where each item will contain
+   * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
+   * profile.
    *
    * "topic_name" may be a relative, private or fully qualified topic name.
    * A relative or private topic will be expanded using this node's namespace and name, if
@@ -1107,17 +1107,15 @@ public:
    *
    * The queried topic name is not remapped.
    *
-   * "no_mangle" defaults to false.
-   *
    * \param[in] topic_name the topic_name on which to find the subscriptions.
    * \param[in] no_mangle if false, the given topic name will be expanded
-   *    to its fully qualified name.
+   *    to its fully qualified name. Default to `false`.
    * \return a list of TopicInfo representing all the subscriptions on this topic.
    * \throws InvalidTopicNameError if the given topic_name is invalid.
    * \throws std::runtime_error if internal error happens.
    */
   RCLCPP_PUBLIC
-  std::vector<rclcpp::TopicInfo>
+  std::vector<rclcpp::TopicEndpointInfo>
   get_subscriptions_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
 
 protected:
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -867,6 +867,58 @@ public:
   size_t
   count_subscribers(const std::string & topic_name) const;
 
+  /// Return the topic endpoint information about publishers on a given topic.
+  /**
+   * The returned parameter is a list of topic endpoint information, where each item will contain
+   * the node name, node namespace, topic type, endpoint type, topic endpoint's GID, and its QoS
+   * profile.
+   *
+   * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
+   * name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps).
+   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
+   * ROS topic name conventions.
+   *
+   * `topic_name` may be a relative, private, or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name.
+   * The queried `topic_name` is not remapped.
+   *
+   * \param[in] topic_name the topic_name on which to find the publishers.
+   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
+   *   otherwise it should be a valid ROS topic name. Defaults to `false`.
+   * \return a list of TopicEndpointInfo representing all the publishers on this topic.
+   * \throws InvalidTopicNameError if the given topic_name is invalid.
+   * \throws std::runtime_error if internal error happens.
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::TopicEndpointInfo>
+  get_publishers_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
+
+  /// Return the topic endpoint information about subscriptions on a given topic.
+  /**
+   * The returned parameter is a list of topic endpoint information, where each item will contain
+   * the node name, node namespace, topic type, endpoint type, topic endpoint's GID, and its QoS
+   * profile.
+   *
+   * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
+   * name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps).
+   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
+   * ROS topic name conventions.
+   *
+   * `topic_name` may be a relative, private, or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name.
+   * The queried `topic_name` is not remapped.
+   *
+   * \param[in] topic_name the topic_name on which to find the subscriptions.
+   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
+   *   otherwise it should be a valid ROS topic name. Defaults to `false`.
+   * \return a list of TopicEndpointInfo representing all the subscriptions on this topic.
+   * \throws InvalidTopicNameError if the given topic_name is invalid.
+   * \throws std::runtime_error if internal error happens.
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::TopicEndpointInfo>
+  get_subscriptions_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
+
   /// Return a graph event, which will be set anytime a graph change occurs.
   /* The graph Event object is a loan which must be returned.
    * The Event object is scoped and therefore to return the loan just let it go
@@ -1071,58 +1123,6 @@ public:
   RCUTILS_WARN_UNUSED
   bool
   assert_liveliness() const;
-
-  /// Return the topic endpoint information about publishers on a given topic.
-  /**
-   * The returned parameter is a list of topic endpoint information, where each item will contain
-   * the node name, node namespace, topic type, endpoint type, topic endpoint's GID, and its QoS
-   * profile.
-   *
-   * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
-   * name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps).
-   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
-   * ROS topic name conventions.
-   *
-   * `topic_name` may be a relative, private, or fully qualified topic name.
-   * A relative or private topic will be expanded using this node's namespace and name.
-   * The queried `topic_name` is not remapped.
-   *
-   * \param[in] topic_name the topic_name on which to find the publishers.
-   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
-   *   otherwise it should be a valid ROS topic name. Defaults to `false`.
-   * \return a list of TopicEndpointInfo representing all the publishers on this topic.
-   * \throws InvalidTopicNameError if the given topic_name is invalid.
-   * \throws std::runtime_error if internal error happens.
-   */
-  RCLCPP_PUBLIC
-  std::vector<rclcpp::TopicEndpointInfo>
-  get_publishers_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
-
-  /// Return the topic endpoint information about subscriptions on a given topic.
-  /**
-   * The returned parameter is a list of topic endpoint information, where each item will contain
-   * the node name, node namespace, topic type, endpoint type, topic endpoint's GID, and its QoS
-   * profile.
-   *
-   * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
-   * name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps).
-   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
-   * ROS topic name conventions.
-   *
-   * `topic_name` may be a relative, private, or fully qualified topic name.
-   * A relative or private topic will be expanded using this node's namespace and name.
-   * The queried `topic_name` is not remapped.
-   *
-   * \param[in] topic_name the topic_name on which to find the subscriptions.
-   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
-   *   otherwise it should be a valid ROS topic name. Defaults to `false`.
-   * \return a list of TopicEndpointInfo representing all the subscriptions on this topic.
-   * \throws InvalidTopicNameError if the given topic_name is invalid.
-   * \throws std::runtime_error if internal error happens.
-   */
-  RCLCPP_PUBLIC
-  std::vector<rclcpp::TopicEndpointInfo>
-  get_subscriptions_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
 
 protected:
   /// Construct a sub-node, which will extend the namespace of all entities created with it.
--- a/rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
@@ -30,7 +30,7 @@
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/node_interfaces/node_graph_interface.hpp"
 #include "rclcpp/visibility_control.hpp"
-#include "rmw/topic_info_array.h"
+#include "rmw/topic_endpoint_info_array.h"
 
 namespace rclcpp
 {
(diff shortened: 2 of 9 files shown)


Barry-Xu-2018:
This commit is for https://github.com/ros2/rclcpp/issues/943

Barry-Xu-2018:
@ivanpauno 

Thanks for your comments.  
Since https://github.com/ros2/rclcpp/issues/943#issuecomment-572297738, I have to pend updating codes.   
After rcl codes have been fixed, I will continue to update codes.

mm318:
The set of pull requests remaining in this discussion is:
1. https://github.com/ros2/rclcpp/pull/960
2. https://github.com/ros2/rclpy/pull/454
3. https://github.com/ros2/ros2cli/pull/385

mm318:
Hi @Barry-Xu-2018, can you add me as a collaborator to https://github.com/Barry-Xu-2018/rclcpp?

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
```diff
@@ -117,9 +118,35 @@ class NodeGraph : public NodeGraphInterface
...
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::TopicInfo>
+  get_subscriptions_info_by_topic(
+    const std::string & topic_name,
+    bool no_mangle = false) const override;
+
 private:
   RCLCPP_DISABLE_COPY(NodeGraph)
 
+  typedef rcl_ret_t (* rcl_get_info_by_topic_func_t)(
```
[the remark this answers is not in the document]
ivanpauno: Yeah, or a template parameter can be used to specify the type, instead of using `std::function`.
note: `std::function` is not a "zero cost" abstraction of a function pointer, I prefer limiting its usage as a container of "any function".

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
```diff
@@ -117,9 +118,35 @@ class NodeGraph : public NodeGraphInterface
...
+    const rcl_node_t * node,
+    rcutils_allocator_t * allocator,
+    const char * topic_name,
+    bool no_mangle,
+    rmw_topic_info_array_t * info_array);
+
+  std::vector<rclcpp::TopicInfo>
+  get_info_by_topic(
```
ivanpauno: This can be moved as a free function in `node_graph.cpp` (it will need a `NodeBaseInterface` argument, instead of `this`).
`rcl_get_info_by_topic_func_t` can be replaced with a template parameter, as commented above.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,103 @@ NodeGraph::count_graph_users()
...
+  rcl_get_info_by_topic_func_t rcl_get_info_by_topic) const
+{
+  std::vector<rclcpp::TopicInfo> topic_info_list;
+
+  auto rcl_node_handle = node_base_->get_rcl_node_handle();
+  auto fqdn = rclcpp::expand_topic_or_service_name(
+    topic_name,
+    rcl_node_get_name(rcl_node_handle),
+    rcl_node_get_namespace(rcl_node_handle),
+    false);    // false = not a service
```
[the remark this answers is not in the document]
ivanpauno: > What specifically did you mean should be done in the rcl layer? This whole function?

Yes.

> Is rclcpp::expand_topic_or_service_name() getting the fully qualified name of topic_name, and there's an equivalent in rcl?

Is getting a name and converting it to a "fully qualified name" (actually, it's just expanding, remappings are missing).
In `rcl`, there is a function that does the `expand`, and another that does the remapping. That combination is repeated in many place in the codebase. I think it would be good to combine the whole `expand and remap` logic in one function (a `is_service` flag will be needed).

----

Most `rcl` functions, like `create_publisher`, accept a topic name that's then expanded and remapped.
The only exception I know is `rcl_count_publishers`, which requires a fully qualified name.
The `rclcpp` counterpart doesn't requires a fully qualified name, but does the expansion (and, AFAIS the remapping part is missing). I wonder if both the expansion and remapping should be done in `rcl`, which would improve code reusage.

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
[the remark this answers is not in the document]
mm318: Yeah, the type erasure in `std::function` has overhead, but I didn't think it was that much. Using a template parameter would be good, too. Just something more idiomatic in C++ than a function pointer.

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
mm318: This fits nicely with my [comment](https://github.com/ros2/rclcpp/pull/960/files/9866d7e16ed76463b70432d1e30d67057183dc64#r366542431) as well: `type` can be replaced with template parameter as well.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,103 @@ NodeGraph::count_graph_users()
...
+
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  rmw_topic_info_array_t info_array = rmw_get_zero_initialized_topic_info_array();
+  auto ret =
+    rcl_get_info_by_topic(rcl_node_handle, &allocator, fqdn.c_str(), no_mangle, &info_array);
+  if (ret != RCL_RET_OK) {
+    // *INDENT-OFF*
+    auto error_msg =
+      std::string("Failed to get information by topic for: ") + type + std::string(":");
```
mm318: And if `type` were a template parameter, this can be a string literal.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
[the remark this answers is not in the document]
mm318: > I wonder if both the expansion and remapping should be done in `rcl`, which would improve code reusage.

Yeah, my first thought was that it's surprising that there's a `expand_topic_or_service_name()`
in `rclcpp`. I thought it was going to be at most just a thin wrapper around something in `rcl`.

Barry-Xu-2018:
@mm318 

> Hi @Barry-Xu-2018, can you add me as a collaborator to https://github.com/Barry-Xu-2018/rclcpp?

OK. I added you.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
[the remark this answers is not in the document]
Barry-Xu-2018: I agree that "rclcpp::expand_topic_or_service_name()" should not at rclcpp layer.  
rclcpp::expand_topic_or_service_name() not only expand the name, but to check if the name is valid and then throw exception of rclcpp.  
If not calling "rclcpp::expand_topic_or_service_name()" at rclcpp layer, I think rcl should check name at each interfaces.  
and then rclcpp check the returned value from rcl to throw rclcpp itself exception.  
Maybe need a new issue to handle this.

Barry-Xu-2018:
@ivanpauno  @mm318 

I have updated codes based on your comments.  
Please review again.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
mm318: I realized that this can't be done without some extremely extensive variadic template magic. Sorry.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,103 @@ NodeGraph::count_graph_users()
...
+    topic_info_list.push_back(topic_info_item);
+  }
+  return topic_info_list;
+}
+
+std::vector<rclcpp::TopicInfo>
+NodeGraph::get_info_by_topic(
+  const std::string & topic_name,
+  bool no_mangle,
+  const std::string & type,
```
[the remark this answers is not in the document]
mm318: Sorry, I didn't know doing this was non-trivial. However, it is possible. Please see https://github.com/Barry-Xu-2018/rclcpp/pull/2.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
mm318: I realized that this can't be done without some extremely extensive variadic template magic. Sorry.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -28,6 +28,12 @@
 using rclcpp::node_interfaces::NodeGraph;
 using rclcpp::exceptions::throw_from_rcl_error;
 using rclcpp::graph_listener::GraphListener;
+using rcl_get_info_by_topic_func_t = rcl_ret_t (* )(
+  const rcl_node_t *,
+  rcutils_allocator_t *,
+  const char *,
+  bool,
+  rmw_topic_endpoint_info_array_t *);
```
mm318: This shouldn't be needed. Please see https://github.com/Barry-Xu-2018/rclcpp/pull/2.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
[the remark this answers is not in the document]
mm318: So to answer @ivanpauno's question

> I wonder if this should be done in rcl layer.

Yeah, I think so, but it's outside the scope of this PR.

Barry-Xu-2018:
@mm318 @ivanpauno @fujitatomoya 

Sorry. Next I am going to take Chinese New Year holiday. During holiday, maybe I cannot response your comments. I will be back on Feb 1st.

mm318:
> Sorry. Next I am going to take Chinese New Year holiday. During holiday, maybe I cannot response your comments. I will be back on Feb 1st.

I should be able to address comments since I'm a collaborator.

Have a great holidays!

Barry-Xu-2018:
> > Sorry. Next I am going to take Chinese New Year holiday. During holiday, maybe I cannot response your comments. I will be back on Feb 1st.
> 
> I should be able to address comments since I'm a collaborator.
> 
> Have a great holidays!

Thank you :)

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
```diff
@@ -25,10 +26,40 @@
...
 #include "rclcpp/event.hpp"
 #include "rclcpp/macros.hpp"
+#include "rclcpp/qos.hpp"
 #include "rclcpp/visibility_control.hpp"
+#include "rmw/topic_endpoint_info.h"
 
 namespace rclcpp
 {
+
+using EndpointType = rmw_endpoint_type_t;
```
ivanpauno: It would be good to define an `enum class`, instead of just using de underlying `enum`.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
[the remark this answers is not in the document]
ivanpauno: Sounds reasonable, the changes are broader that just this new functions.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,87 @@ NodeGraph::count_graph_users()
...
+    topic_name,
+    rcl_node_get_name(rcl_node_handle),
+    rcl_node_get_namespace(rcl_node_handle),
+    false);    // false = not a service
+
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  rmw_topic_endpoint_info_array_t info_array = rmw_get_zero_initialized_topic_endpoint_info_array();
+  auto ret =
+    rcl_get_info_by_topic(rcl_node_handle, &allocator, fqdn.c_str(), no_mangle, &info_array);
+  if (ret != RCL_RET_OK) {
```
ivanpauno: According the GSG, the order should be `literal != variable`.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,87 @@ NodeGraph::count_graph_users()
...
+    auto error_msg =
+      std::string("Failed to get information by topic for ") + EndpointType + std::string(":");
+    if (ret == RCL_RET_UNSUPPORTED) {
```
ivanpauno: same here

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,87 @@ NodeGraph::count_graph_users()
...
+      error_msg += std::string("function not supported by RMW_IMPLEMENTATION");
+    } else {
+      error_msg += rcl_get_error_string().str;
+    }
+    rcl_reset_error();
+    if (rmw_topic_endpoint_info_array_fini(&info_array, &allocator) != RMW_RET_OK) {
```
ivanpauno: same here

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,87 @@ NodeGraph::count_graph_users()
...
+  return topic_info_list;
+}
+
+static const char kPublisherEndpointTypeName[] = "publishers";
+std::vector<rclcpp::TopicEndpointInfo>
+NodeGraph::get_publishers_info_by_topic(
+  const std::string & topic_name,
+  bool no_mangle) const
+{
+  return get_info_by_topic<kPublisherEndpointTypeName>(
```
ivanpauno: why not just `get_info_by_topic<"publishers">`.

mm318: The compiler will complain about lack of external linkage.

ivanpauno: Didn't know that that wasn't supported, thanks for clarifying!

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -369,3 +369,87 @@ NodeGraph::count_graph_users()
...
+get_info_by_topic(
+  rclcpp::node_interfaces::NodeBaseInterface * node_base,
+  const std::string & topic_name,
+  bool no_mangle,
+  FunctionT rcl_get_info_by_topic)
+{
+  std::vector<rclcpp::TopicEndpointInfo> topic_info_list;
+
+  auto rcl_node_handle = node_base->get_rcl_node_handle();
+  auto fqdn = rclcpp::expand_topic_or_service_name(
```
mm318: I'm curious, how does `expand_topic_or_service_name()` work without a `no_mangle` parameter?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,52 @@ class Node : public std::enable_shared_from_this<Node>
...
+  /**
+   * The returned parameter is a list of topic endpoint information, where each item will contain
+   * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
+   * profile.
+   *
+   * "topic_name" may be a relative, private or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name, if
+   * the "no_mangle" param is set to false.
+   *
+   * The queried topic name is not remapped.
```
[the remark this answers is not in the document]
mm318: So `rclcpp::expand_topic_or_service_name()` and `_rclpy.rclpy_expand_topic_name()` will expand the topic name, as the function name suggests, right? Does the remapping happen somewhere else under the hood?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,52 @@ class Node : public std::enable_shared_from_this<Node>
...
+   *
+   * \param[in] topic_name the topic_name on which to find the publishers.
+   * \param[in] no_mangle if false, the given topic name will be expanded
+   *    to its fully qualified name. Default to `false`.
```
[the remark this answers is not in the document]
mm318: Fixed.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
mm318: Done.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
mm318: Done.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
mm318: Done.

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
mm318: Done.

mm318:
> The set of pull requests remaining in this discussion is:
> ...

The current set of pull requests:
1. https://github.com/ros2/rcl/pull/557 (doc string fix)
2. https://github.com/ros2/rclcpp/pull/960
3. https://github.com/ros2/rclpy/pull/454
4. https://github.com/ros2/ros2cli/pull/385

rclcpp/include/rclcpp/node.hpp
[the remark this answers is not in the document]
ivanpauno: Here, the remapping is missing and it's not happening.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
ivanpauno: So, one thing is mangling, i.e. generate a valid middleware name (e.g. DDS) from the ROS name.
Another thing is expansion, i.e. get full names from relative names ('topic' -> '/namespace/topic/') and full names from private names ('~/topic' -> '/namespace/node_name/topic/').
Remapping should be always done after expansion, but it's missing here (can be added later, as it's a preexisting problem in `get_publisher_count` too).

We're everywhere always doing remapping and expansion, including when we do `no_mangle=true` (or when [this](https://github.com/ros2/rmw/blob/a3774531e9b7289574501275aafbc3f93120360e/rmw/include/rmw/types.h#L334) qos setting is set to `true`). IMO, always doing remapping and expansion is ok, though is arguable. Considering that, `expand_topic_or_service_name` doesn't need to know about `no_mangle`.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1078,16 +1078,19 @@ class Node : public std::enable_shared_from_this<Node>
...
-   * the "no_mangle" param is set to false.
+   * When the `no_mangle` parameter is `true`, the provided `topic_name` should be a valid topic
+   * name for the middleware (useful when combining ROS with native middleware (e.g. DDS) apps).
+   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
+   * ROS topic name conventions.
    *
-   * The queried topic name is not remapped.
+   * `topic_name` may be a relative, private, or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name,
+   * The queried `topic_name` is not remapped.
```
ivanpauno: nit: shouldn't start in caps

mm318: Oops. There was supposed to be a period instead of a comma.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
wjwwood: Has this been resolved?

rclcpp/include/rclcpp/node.hpp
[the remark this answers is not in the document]
wjwwood: What's the status of this conversation? The `rclpy` code also does not remap.

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
wjwwood: Is this still a thing? (conversation was not resolved, so I'm unsure)

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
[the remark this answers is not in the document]
wjwwood: Is this resolved?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,58 @@ class Node : public std::enable_shared_from_this<Node>
   bool
   assert_liveliness() const;
 
+  /// Return a list of publishers about topic endpoint information to a given topic.
```
wjwwood: nitpick: This is a very awkward sentence.

Maybe "Return the topic endpoint information about publishers on a given topic."?

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,58 @@ class Node : public std::enable_shared_from_this<Node>
...
+  /**
+   * The returned parameter is a list of topic endpoint information, where each item will contain
+   * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
```
wjwwood: Same as in `rclpy`, `participant` is not a ROS term, please limit documentation to ROS terminology.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,58 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * When the `no_mangle` parameter is `false`, the provided `topic_name` should follow
+   * ROS topic name conventions.
+   *
+   * `topic_name` may be a relative, private, or fully qualified topic name.
+   * A relative or private topic will be expanded using this node's namespace and name.
+   * The queried `topic_name` is not remapped.
+   *
+   * \param[in] topic_name the topic_name on which to find the publishers.
+   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
+   *            otherwise it should be a valid ROS topic name. Defaults to `false`.
```
wjwwood: nitpick: do not indent like this, just do two spaces

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,58 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * \return a list of TopicEndpointInfo representing all the publishers on this topic.
+   * \throws InvalidTopicNameError if the given topic_name is invalid.
+   * \throws std::runtime_error if internal error happens.
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::TopicEndpointInfo>
+  get_publishers_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
+
+  /// Return a list of subscriptions about topic endpoint information to a given topic.
```
wjwwood: Same as publisher version.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,58 @@ class Node : public std::enable_shared_from_this<Node>
...
+  /**
+   * The returned parameter is a list of topic endpoint information, where each item will contain
+   * the node name, node namespace, topic type, endpoint type, participant's GID and its QoS
```
wjwwood: Same as publisher.

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
```diff
@@ -25,10 +26,45 @@
...
+
+enum class EndpointType
+{
+  Invalid = RMW_ENDPOINT_INVALID,
+  Publisher = RMW_ENDPOINT_PUBLISHER,
+  Subscription = RMW_ENDPOINT_SUBSCRIPTION
+};
+
+/**
+ * Use to get topic endpoint information that containing the node name, node namespace, topic type,
```
wjwwood: nitpick: this is hard to read, and I'm not sure what it's trying to say

"Struct that contains topic endpoint information like the associated node name, node namespace, ..., and its QoS."?

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
```diff
@@ -25,10 +26,45 @@
...
+ * endpoint type, GID of endpoint and its QoS.
```
wjwwood: nitpick: include the Oxford comma

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
```diff
@@ -25,10 +26,45 @@
...
+struct RCLCPP_PUBLIC TopicEndpointInfo
+{
+  std::string node_name;
+  std::string node_namespace;
+  std::string topic_type;
+  rclcpp::EndpointType endpoint_type;
+  std::array<uint8_t, RMW_GID_STORAGE_SIZE> endpoint_gid;
+  rclcpp::QoS qos;
+
+  /// Constructor which convert rmw_topic_endpoint_info_t to TopicEndpointInfo.
```
wjwwood: `converts`?

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
```diff
@@ -150,6 +186,24 @@ class NodeGraphInterface
   virtual
   size_t
   count_graph_users() = 0;
+
+  /// Return a list of publishers about topic endpoint information to a given topic.
```
wjwwood: These briefs are awkward too (look like copies from above).

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
wjwwood: I think we need to get a holistic summary of how we think `no_mangle` should be used with each of these functions, and then make sure the implementation matches. I feel like there have been several such conversations and potentially different solutions in each case. Also I haven't heard a compelling story for why one way or the other.

I would prefer to see clarity on this point before moving forward, even if that means we need to follow up after this pr and fix other functions.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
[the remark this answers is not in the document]
wjwwood: Following my comment in the conversation about the line just above, I think we need to decide if it is needed or not, and if it is needed then we should probably do it in rcl if we can, but I agree we could do it in a follow up. However, I would like to have a decision on what we should do before continuing here.

rclcpp/include/rclcpp/node.hpp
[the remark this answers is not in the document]
mm318: The docstring will continue to say that `topic_name` will not be remapped.

rclcpp/include/rclcpp/node.hpp
mm318: Done.

rclcpp/include/rclcpp/node.hpp
mm318: Switched over to using `topic endpoint`.

rclcpp/include/rclcpp/node.hpp
mm318: Switched over to using `topic endpoint`.

rclcpp/include/rclcpp/node.hpp
mm318: Fixed.

rclcpp/include/rclcpp/node.hpp
mm318: Done.

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
mm318: Done.

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
mm318: Done.

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
mm318: Done.

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
mm318: Done.

mm318:
The updated set of pull requests:
1. https://github.com/ros2/rcl/pull/558
2. https://github.com/ros2/rclcpp/pull/960
3. https://github.com/ros2/rclpy/pull/454
4. https://github.com/ros2/ros2cli/pull/385

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
mm318: I think this should be resolved.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
mm318: > Another thing is expansion, i.e. get full names from relative names ('topic' -> '/namespace/topic/') and full names from private names ('~/topic' -> '/namespace/node_name/topic/').

I meant to confirm if the expansion algorithm will work for a middleware topic names well? For example `rt/topic` -> `rt/namespace/topic`. (And I guess `rt/~/topic` is an invalid topic name regardless of the value of `no_mangle`.)

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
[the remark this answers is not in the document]
mm318: I think there's a couple of things going on here.

`rclcpp::expand_topic_or_service_name()` does topic name expansion and validation, and does not do remapping.

`rclpy.expand_topic_name.expand_topic_name()` does topic name expansion, does not do validation (the caller is supposed to do it manually), and also does not do remapping.

(My understanding of remapping here is that the returned/output topic name is the remapped version of the input topic name.)

The follow-up should be to implement in `rcl` a function like `rcl_get_expanded_and_remapped_topic_name()` (and it also does validation). This function supposedly should be agnostic to the `no_mangle` parameter as well.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
ivanpauno: I will answer to both this discussion and https://github.com/ros2/rclcpp/pull/960#discussion_r364215929.

> I think we need to get a holistic summary of how we think no_mangle should be used with each of these functions, and then make sure the implementation matches. I feel like there have been several such conversations and potentially different solutions in each case. Also I haven't heard a compelling story for why one way or the other.

Considering that the intention of `no_mangle=true` is to pass a valid middleware name, I think that in some cases it might not play well with remapping expansions. I don't see a problem particularly with DDS based `rmw` implementations, but if others need something more than just adding a prefix a suffix (e.g.: replacing the separator '/' with other specific separator), it might create problems.
Moreover, the idea of `no_mangle` is interacting with native applications of the underlying middleware, which probably don't have somthing similar to remapping and expansion, so it might make sense to ignore it.
I don't have arguments in favor of applying remapping and expansion when using `no_mangle=true`.

----

About adding a function that do both remapping and expansion in `rcl`, I think that's the correct path to go. I don't see an use case when you want to apply one and not the other to a topic/service name. Doing one and not the other just seems wrong.

wjwwood: Maybe we could chat in person about this? I don't think we're talking about all the same things. For me (I may have the wrong idea), `no_mangle` may affect the output (as in do not demangle topic names in the returned information) or it may mean that the input should be used verbatim and not mangled (as in I'm passing a middleware topic name that should be used as-is, and not a ROS topic). In the latter case, I think that expanding the input may not be desired, but I could be wrong.

mm318: My understanding is that this is indeed the latter case: if `no_mangle==true`, then the input into this `get_info_by_topic()` function and subsequently the `expand_topic_or_service_name()` function is to be treated as a middleware topic name.

-----------------------

> I don't have arguments in favor of applying remapping and expansion when using `no_mangle=true`.

> In the latter case, I think that expanding the input may not be desired, but I could be wrong.

Then it sounds like `no_mangle==true` does imply that the input `topic_name` parameter will need to be fully qualified? (Though we've been removing statements about that from the docstrings.)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1072,6 +1072,58 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * \param[in] topic_name the topic_name on which to find the publishers.
+   * \param[in] no_mangle if `true`, `topic_name` needs to be a valid middleware topic name,
+   *   otherwise it should be a valid ROS topic name. Defaults to `false`.
+   * \return a list of TopicEndpointInfo representing all the publishers on this topic.
+   * \throws InvalidTopicNameError if the given topic_name is invalid.
+   * \throws std::runtime_error if internal error happens.
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::TopicEndpointInfo>
+  get_publishers_info_by_topic(const std::string & topic_name, bool no_mangle = false) const;
```
wjwwood: These two methods were not also added to the `LifecycleNode` class.

This keeps coming up and so to avoid it we could consider leaving these methods off of `Node` and instead have free functions that take the node as a template argument instead, something like:

```c++
auto publisher_infos = get_publishers_info_by_topic(node, "topic_name");
```

Where `node` could be a `rclcpp::Node` or a `rclcpp_lifecycle::LifecycleNode`.

wjwwood: Either way, this pull request should make sure `Node` and `LifecycleNode` can both be used the same way (either each having these methods, or the free function).

mm318: I think this is actually simpler than that. There's no implementation in `Node::get_publishers_info_by_topic()`, it is just a wrapper around `node_graph_->get_publishers_info_by_topic()`. The same wrapper can be added to `LifecycleNode`.

wjwwood: > I think this is actually simpler than that. There's no implementation in `Node::get_publishers_info_by_topic()`, it is just a wrapper around `node_graph_->get_publishers_info_by_topic()`. The same wrapper can be added to `LifecycleNode`.

No doubt it can be added, my point was simply that if we standardized on a templated free function that takes a node instead, we'd not need to have the "wrapper" methods for each class. Keeping them in sync has not been trivial.

But until we do an API review and make a decision on that point, adding it to the `LifecyleNode` manually is fine.

mm318:
Are we missing anymore reviews on this set of pull requests?

wjwwood:
@ivanpauno and I are going to chat about the use of `expand_topic_or_service_name()` tomorrow, and then I think this will be ready for final review from us. Sorry for the delay.

mm318:
What was the conclusion to the discussion, @wjwwood and @ivanpauno? Is anything gating this set of pull requests?

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
wjwwood: @ivanpauno summarized our discussion here:

https://github.com/ros2/ros2/issues/862#issuecomment-579464354

Basically we want all of the graph query functions (including this one) to do both expansion (`foo` becomes `/node_ns/foo`) and remapping, by default. And then we want them all to have an option to skip both expansion and remapping.

Also, any functions that return topic or type names should be able to avoid _de_-mangling of those names.

Currently this implementation does expansion but not remapping (as far as I can tell), and it has an option `no_mangle` to make the query assume the input topic should be considered as is, without any additional prefixes or other mangling. It does not, however, have a `no_demangle`, which some of the other functions do have.

This `no_mangle` option does not prevent the topic name expansion here, which is a problem in my opinion, what do you think @ivanpauno?

----

As I see it we need to do these things to get this merged:

- add remapping of the input topic name to this function
- avoid expansion and remapping when `no_mangle` is specified

Later we should make a set of pull requests which add a `no_demangle` option to these query functions. This will affect all the layers unfortunately.

wjwwood: We should also make sure the `rclpy` version of this matches.

ivanpauno:
> What was the conclusion to the discussion, @wjwwood and @ivanpauno? Is anything gating this set of pull requests?

Please, check https://github.com/ros2/ros2/issues/862.
IMO, this is ready to be merged, changes about how `no_mangle` is handled, and if `remapping` should be done or note, can be addressed by a follow-up (as there's already some inconsistency in other places).

I will run CI if @wjwwood agrees.

wjwwood:
@ivanpauno please read https://github.com/ros2/rclcpp/pull/960#discussion_r373078924

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
ivanpauno: > Basically we want all of the graph query functions (including this one) to do both expansion (foo becomes /node_ns/foo) and remapping, by default. And then we want them all to have an option to skip both expansion and remapping.

I completly agree with this.
I think it should be done in `rcl`, repeating the code in `rclcpp` and `rclpy` doesn't make much sense.

> Also, any functions that return topic or type names should be able to avoid de-mangling of those names.

Currently, the topic types are being demangled or not based on `no_mangle` argument.
Considering the use cases mentioned in the issue, I think that's fine.
We could have a better name, like `use_ros_conventions`, default to true:
- `true` means both input and output are ros names.
- `false` means both input and output are dds names (input shouldn't be expanded/remapped/mangled, output shouldn't be demangled).

> This no_mangle option does not prevent the topic name expansion here, which is a problem in my opinion, what do you think @ivanpauno?

Yeah, as we talked, I think it's a problem.

----

We can left this as-is for now, and then correct all the places where we have this error together (publisher/subscription creation is applying remapping and expansion when ros_conventions qos is false , publisher/subscription count is not doing remapping, etc). I don't feel strongly about which path to take.

wjwwood: > I think it should be done in rcl, repeating the code in rclcpp and rclpy doesn't make much sense.

In principle yes, but actually I think the sub-nodes feature in rclcpp prevent this for now. It needs to be more carefully considered I think. I would postpone that to another pull request.

----

> Currently, the topic types are being demangled or not based on no_mangle argument.
Considering the use cases mentioned in the issue, I think that's fine.
We could have a better name, like use_ros_conventions, default to true: ...

I think having just one option can be ok, though it does make the odd cases impossible, because there's no way to manually do mangling/demangling.

If we want to go the route of having one option for both behaviors, then we need to change the `no_mangle` name, as it doesn't cover the existing behavior correctly, imo.

----

> We can left this as-is for now, and then correct all the places where we have this error together (publisher/subscription creation is applying remapping and expansion when ros_conventions qos is false , publisher/subscription count is not doing remapping, etc). I don't feel strongly about which path to take.

I don't think the two points I highlighted (add remapping and avoid expansion/remapping when `no_mangle` is true) are very difficult so I'd prefer those to be done now. Everything else can be done in follow up.

mm318:
Okay, so it sounds like the next steps for these PRs now is to:
- Add remapping of the input topic name to `get_info_by_topic()` in `rclcpp` and `rclpy`
- Avoid expansion and remapping of the input topic name in `get_info_by_topic()` in `rclcpp` and `rclpy` when `no_mangle==true`

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -381,26 +382,59 @@ convert_to_topic_info_list(const rcl_topic_endpoint_info_array_t & info_array)
...
+      rcl_node_get_namespace(rcl_node_handle),
+      false);    // false = not a service
+
+    // Get the node options
+    const rcl_node_options_t * node_options = rcl_node_get_options(rcl_node_handle);
+    if (nullptr == node_options) {
+      throw std::runtime_error("Need valid node options in get_info_by_topic()");
+    }
+
+    // global before local so that local overwrites global
```
ivanpauno: nit: what is this comment referring to?

wjwwood: I'm not sure what the comment is meant to say, but the way the remapping works is that there are "local" remappings in the `node_options->arguments` and "global" remappings in the `global_args`. But the global args are only used if `node_options->use_global_arguments` is true.

mm318: I got the content of this comment from `rclcpp`'s `node_parameters.cpp`. I didn't notice that the context is very different. I'll remove this comment.

ivanpauno:
CI run [here](https://github.com/ros2/rclpy/pull/454#issuecomment-584138107).

mm318:
Hi @ivanpauno, any advice on how to fix the Windows CI failure in `rclcpp` at https://ci.ros2.org/job/ci_windows/9214/? It looks like `std::string` members needs to be exported? Has this (or for member variables that are some other type of `std` container) been done elsewhere in `rclcpp` before?

ivanpauno:
@mm318 I've read a bit, and this seems to be a problem when a member variable of an exported class is template defined in another library:

- (discussion) https://stackoverflow.com/questions/2132747/warning-c4251-when-building-a-dll-that-exports-a-class-containing-an-atlcstrin
- (discussion) https://web.archive.org/web/20170811142318/http://www.microsoft-questions.com/microsoft/VC-Language/30952961/a-solution-to-warning-c4251--class-needs-to-have-dllinterface.aspx
- (doc) https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4251?view=vs-2019

Considering that we already have a test that compiles and access those member variables without problems, I would just ignore the warning with a pragma. I don't think that would be an issue.
Another option is switching from a plain `struct` to a `class` with setters and getters.

@wjwwood opinions?

rclcpp/include/rclcpp/node_interfaces/node_graph_interface.hpp
```diff
@@ -15,20 +15,57 @@
...
+  Invalid = RMW_ENDPOINT_INVALID,
+  Publisher = RMW_ENDPOINT_PUBLISHER,
+  Subscription = RMW_ENDPOINT_SUBSCRIPTION
+};
+
+/**
+ * Struct that contains topic endpoint information like the associated node name, node namespace,
+ * topic type, endpoint type, endpoint GID, and its QoS.
+ */
+struct RCLCPP_PUBLIC TopicEndpointInfo
```
wjwwood: You can try removing this `RCLCPP_PUBLIC` and instead place it on each member to see if that fixes the warnings.

mm318: I have just finished implementing a solution that involves adding member functions that return `const` and non-`const` references. Would that be acceptable?

mm318:
In my opinion, this is ready for a CI re-run, @ivanpauno.

ivanpauno:
I think that the boilerplate is fine, but the windows warning seems to be non-sense. Ignoring it with a pragma would be fine too.

ivanpauno:
New CI round: https://github.com/ros2/rclpy/pull/454#issuecomment-585214475

wjwwood:
> but the windows warning seems to be non-sense. Ignoring it with a pragma would be fine too.

I'm not sure that's true, but we're not doing that so it's fine.

ivanpauno:
Merged! Thanks @Barry-Xu-2018 and @mm318 for pushing this!

PR #926, opened 2019-11-15
 add support for STREAM logging macros

dirk-thomas:
Fixes #915. Requires ros2/rcutils#192.

The changes to the `logging.hpp.em` template only result in additional lines and don't change/remove any existing lines.

Since the `STREAM` macros only differ in the vargs / stream_arg handling the new test only covers one of the feature combinations.

The first commit is a pure refactoring without changing the expanded template. The second commit adds `stream` to the feature combinations.

The following details show the added lines for the `DEBUG` log level:

<details>
<summary>New empty STREAM macros for all existing feature combinations in case logging has been disabled.</summary>

> /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
> #define RCLCPP_DEBUG_STREAM(...)
> /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
> #define RCLCPP_DEBUG_STREAM_ONCE(...)
> /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
> #define RCLCPP_DEBUG_STREAM_EXPRESSION(...)
> /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
> #define RCLCPP_DEBUG_STREAM_FUNCTION(...)
> /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
> #define RCLCPP_DEBUG_STREAM_SKIPFIRST(...)
> /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
> #define RCLCPP_DEBUG_STREAM_THROTTLE(...)
> /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
> #define RCLCPP_DEBUG_STREAM_SKIPFIRST_THROTTLE(...)

</details>

<details>
<summary>New STREAM macros for all existing feature combinations.</summary>

> // The RCLCPP_DEBUG_STREAM macro is surrounded by do { .. } while (0)
> // to implement the standard C macro idiom to make the macro safe in all
> // contexts; see http://c-faq.com/cpp/multistmt.html for more information.
> /**
>  * \def RCLCPP_DEBUG_STREAM
>  * Log a message with severity DEBUG.
>  * \param logger The `rclcpp::Logger` to use
>  * \param stream_arg The argument << into a stringstream
>  */
[...]

PR #964, opened 2020-01-16
logging.hpp: Added missing include

hdino:
Including just `logging.hpp` and using the new stream macros (#926) is causing the following error:
```
error: aggregate ‘std::stringstream ss’ has incomplete type and cannot be defined
RCLCPP_INFO_STREAM(logger, "Test");
```
I added the required include line to fix this.

### DIFF
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -17,6 +17,7 @@
 #ifndef RCLCPP__LOGGING_HPP_
 #define RCLCPP__LOGGING_HPP_
 
+#include <sstream>
 #include <type_traits>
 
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9059/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4807/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7406/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8988/)

clalancette:
The failures on macOS and Windows were also there in the nightlies, so were not caused by this patch.  Merging, thanks for the contribution!

PR #967, opened 2020-01-21
Deprecated is_initialized()

jacobperron:
The function was previously documented as being deprecated, but this change adds compiler warnings if it is used.
Ignore compiler warnings where the function is being tested and change to the preferred usage elsewhere.

It doesn't look like where using `is_initialized()` in any other `ros2` packages, but running a build for all of them just in case: [Build Status](https://ci.ros2.org/job/ci_linux/9099/) (edit: rebuild with connected PR)

### DIFF
--- a/rclcpp/include/rclcpp/utilities.hpp
+++ b/rclcpp/include/rclcpp/utilities.hpp
@@ -154,6 +154,7 @@ ok(rclcpp::Context::SharedPtr context = nullptr);
  * \return true if the context is initialized, and false otherwise
  */
 RCLCPP_PUBLIC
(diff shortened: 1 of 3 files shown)


jacobperron:
Ah, missed that usage. Fixed in https://github.com/ros-visualization/rqt/pull/215

clalancette:
It looks like this PR is causing issues on macOS; see https://ci.ros2.org/job/ci_osx/7443/console for one example.

jacobperron:
Clang doesn't seem to like the introduced syntax. See https://github.com/ros2/rclcpp/pull/968 for a workaround for macOS.

PR #971, opened 2020-01-29
Accept taking an rvalue reference future in spin_until_future_complete

ivanpauno:
I don't see a reason why not (`wait`, `wait_for`, `wait_until`, `valid` are all const methods).
In this way, a temporal is admitted.

Another question:
Why not accepting a `std::future` by value too?

that would transform:
```cpp
std::future<RET_TYPE> future = ...;
...
rclcpp::spin_until_future_complete(node, future.share());
```

in:
```cpp
std::future<RET_TYPE> future = ...;
...
rclcpp::spin_until_future_complete(node, std::move(future));
```

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -223,7 +223,7 @@ public:
   template<typename ResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli>
   FutureReturnCode
   spin_until_future_complete(
-    std::shared_future<ResponseT> & future,
+    const std::shared_future<ResponseT> & future,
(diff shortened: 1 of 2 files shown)


wjwwood:
> Why not accepting a std::future by value too?

I don't think there is a reason, as far as I know. Likely just added the signatures needed at the time. Services return shared futures I believe.

ivanpauno:
> I don't think there is a reason, as far as I know. Likely just added the signatures needed at the time. Services return shared futures I believe.

Sounds good, created issue https://github.com/ros2/rclcpp/issues/972.
CI jobs in https://github.com/ros2/demos/pull/427#issuecomment-579925930.

PR #819, opened 2019-08-07
Dnae adas/serialized intra process communication

DensoADAS:
Description:
Extended rclcpp to allow serialized intra process communication. Therefore, we extended mapped ring buffer to allow to handle the message type and the (possible custom) deleter of the serialized data. If the subscription or publisher are not serialized the serialized message is converted on demand.

Possible applications:
 * direct recording of serialized data without serialization (with IPM)
 * non-templated publisher of serialized messages (with IPM)

Features:
* extended mapped_ring_buffer to handle serialized data (including deleter)
* added serialization and deserialization functions in intraprocess manager
* converting data content if needed (if non/serialized subscription vs. publisher)
* moved publish(rcl_serialized_message_t) from Publisher<T> to PublisherBase to make it more generic
* added unit tests: test_intra_process_communication

PR #973, opened 2020-01-30
Dnae adas/serialized ipm

DensoADAS:
Description:
Extended rclcpp to allow serialized intra process communication. Therefore, we added a second connection to the intraprocess manager to the publisher and subscriber for serialized messages. To ease up the memory management, the rcl_serialized_message_t is wrapped in SerializedContainer with a delete to prevent memory loss. To prevent double free the ownership of the serialized message is passed to the publisher. The conversion between serialized and deserialized messages is done in the SubscriptionIntraProcess on demand. The changes also allow general purpose publisher and subscriber as needed e. g. for rosbag2.

Possible applications:
•	direct recording of serialized data without serialization (with IPM)
•	non-templated publisher of serialized messages (with IPM)

Features:
•	added serialization and deserialization functions in intraprocess manager
•	implemented publish of serialized messages (call by value and unique pointer) in Publisher and PublisherLifecycle
•	claiming ownership of serialized messages by publish(rcl_serialized_message_t)
•	create_publisher and create_subscription support now general message type of rcl_serialized_message_t by passing message_type (so non-templated version is possible)
•	added unit tests: test_intra_process_communication


DensoADAS:
updated from https://github.com/ros2/rclcpp/pull/819

DensoADAS:
Has anyone reviewed the PR already?

wjwwood:
No, we have not sorry. We'll get to it as soon as we can.

DensoADAS:
@ivanpauno @wjwwood Any updates? This PR is blocking the rosbag2 PR.

ivanpauno:
> @ivanpauno @wjwwood Just a small after-Easter question: Any updates? This PR is blocking the rosbag2 PR.

Sorry, but I don't think that I will be able to do a review before `Foxy` release.

Karsten1987:
@DensoADAS can you please rebase this branch? The connected PR for rosbag2 please as well.

rclcpp/include/rclcpp/experimental/serialized_container.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include "rmw/serialized_message.h"
+
+namespace rclcpp
+{
+namespace experimental
+{
+
+/// Object oriented version of rcl_serialized_message_t with destructor to avoid memory leaks
+class SerializedContainer : public rcl_serialized_message_t
```
Karsten1987, line 30: I wouldn't call it a container. I associate some sort of array or list with a container in C++. Why not just calling it `rclcpp::SerializedMessage`? That would go ahead with things like a `rclcpp::LoanedMessage`.

Karsten1987, line 30: The inheritance model goes okay here, yet I would prefer to be consistent with what we have in `rclcpp::LoanedMessage` and have the C++ class only wrap the `rcl_serialized_message_t`.

rclcpp/include/rclcpp/experimental/serialized_container.hpp
```diff
@@ -0,0 +1,75 @@
...
+      std::memcpy(buffer, sc.buffer, sc.buffer_length);
+    }
+    buffer_length = sc.buffer_length;
+  }
+
+  explicit SerializedContainer(rcl_serialized_message_t && msg)
+  : rcl_serialized_message_t(msg)
+  {
+    // reset buffer to prevent double free
+    msg = rmw_get_zero_initialized_serialized_message();
```
Karsten1987, line 56: don't you leak memory here if `msg` is not empty?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -64,11 +71,15 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
-    rclcpp::IntraProcessBufferType buffer_type)
+    rclcpp::IntraProcessBufferType buffer_type,
+    std::shared_ptr<SerializationBase> serializer)
   : SubscriptionIntraProcessBase(topic_name, qos_profile),
-    any_callback_(callback)
+    any_callback_(callback), serializer_(serializer)
   {
-    if (!std::is_same<MessageT, CallbackMessageT>::value) {
+    if (!std::is_same<MessageT, CallbackMessageT>::value &&
+      !std::is_same<MessageT, rclcpp::experimental::SerializedContainer>::value &&
```
Karsten1987, line 80: How can that ever be true? That's to say whether `MessageT` is equal to `rcl_serialized_message_t`? That's never the case.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -141,14 +152,40 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
   }
 
   template<typename T>
-  typename std::enable_if<std::is_same<T, rcl_serialized_message_t>::value, void>::type
+  typename std::enable_if<
+    std::is_same<T, rcl_serialized_message_t>::value &&
+    !std::is_same<MessageT, rclcpp::experimental::SerializedContainer>::value,
```
Karsten1987, line 157: again, I don't understand why we need that check.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -141,14 +152,40 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+    void>::type
   execute_impl()
   {
-    throw std::runtime_error("Subscription intra-process can't handle serialized messages");
+    if (serializer_) {
```
Karsten1987, line 161: nitpick here: I would invert the check and throw/return early `if (nullptr == serializer_)`

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -141,14 +152,40 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+      rmw_message_info_t msg_info;
+      msg_info.from_intra_process = true;
+
+      ConstMessageSharedPtr msg = buffer_->consume_shared();
+      auto serialized_msg =
+        serializer_->serialize_message(reinterpret_cast<const void *>(msg.get()));
```
Karsten1987, line 167: do we need that `reinterpret_cast` here?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -141,14 +152,40 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+
+      if (serialized_msg == nullptr) {
```
Karsten1987, line 169:
```suggestion
      if (nullptr == serialized_msg) {
```

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
Karsten1987: EDIT: Help me to get this right here.

Is the idea to separate between the four cases
* publish(MsgT) -> subscribe(MsgT)
* publish(MsgT) -> subscribe(rcl_serialized_message_t)
* publish(rcl_serialized_msg_t) -> subscribe(MsgT)
* publish(rcl_serialized_msg_t) -> subscribe(rcl_serialized_msg_t)

where this precise specialization is `publish(rcl_serialized_msg_t) -> subscribe(MsgT)` ?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -141,14 +152,40 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+    } else {
+      throw std::runtime_error("Subscription intra-process can't handle serialized messages");
+    }
   }
 
   template<class T>
-  typename std::enable_if<!std::is_same<T, rcl_serialized_message_t>::value, void>::type
+  typename std::enable_if<
+    !std::is_same<T, rcl_serialized_message_t>::value &&
+    !std::is_same<MessageT, rclcpp::experimental::SerializedContainer>::value,
```
Karsten1987, line 187: Connecting to the comment above, this specialization is `publish(rcl_serialized_msg_t) -> subscribe(rcl_serailized_msg_t)` ?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -163,8 +200,75 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+    std::is_same<MessageT, rclcpp::experimental::SerializedContainer>::value,
+    void>::type
+  execute_impl()
+  {
+    rmw_message_info_t msg_info;
+    msg_info.from_intra_process = true;
+
+    if (any_callback_.use_take_shared_method()) {
+      ConstMessageSharedPtr msg = buffer_->consume_shared();
+      if (msg == nullptr) {
```
Karsten1987, line 215:
```suggestion
      if (nullptr == msg) {
```

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -163,8 +200,75 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
   }
 
+  template<typename T>
+  typename std::enable_if<
+    std::is_same<T, rcl_serialized_message_t>::value &&
+    std::is_same<MessageT, rclcpp::experimental::SerializedContainer>::value,
+    void>::type
+  execute_impl()
+  {
+    rmw_message_info_t msg_info;
```
Karsten1987, line 210:
```suggestion
    rmw_message_info_t msg_info {};
```

Karsten1987: same comment applies to all four functions.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -163,8 +200,75 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+        serializer_->deserialize_message(*serialized_container,
+          reinterpret_cast<void *>(msg.get()));
+        any_callback_.dispatch_intra_process(std::move(msg), msg_info);
+      }
+    } else {
+      throw std::runtime_error("Subscription intra-process can't handle unserialized messages");
+    }
+  }
+
+  CallbackMessageUniquePtr construct_unique()
```
Karsten1987, line 257: This could be a (inline) free function (anonymous namespace) which takes the allocator as a first argument.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -172,34 +192,34 @@ class Publisher : public PublisherBase
...
-    // A shared_ptr<const MessageT> could also be constructed here.
-    auto ptr = MessageAllocatorTraits::allocate(*message_allocator_.get(), 1);
-    MessageAllocatorTraits::construct(*message_allocator_.get(), ptr, msg);
-    MessageUniquePtr unique_msg(ptr, message_deleter_);
-    this->publish(std::move(unique_msg));
+    this->do_publish_message(msg);
   }
 
-  void
-  publish(const rcl_serialized_message_t & serialized_msg)
```
Karsten1987, line 200: I don't think this function should be deleted. There should still be the option of publishing a serialized message, just like we do in the demos: https://github.com/ros2/demos/blob/master/demo_nodes_cpp/src/topics/talker_serialized_message.cpp#L111

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -158,6 +173,11 @@ class Publisher : public PublisherBase
   virtual void
   publish(std::unique_ptr<MessageT, MessageDeleter> msg)
   {
+    if (std::is_same<MessageT, rcl_serialized_message_t>::value) {
+      this->template publish<MessageDeleter>(std::move(msg));
+      return;
+    }
```
Karsten1987, line 179: I think this call should not be possible if we keep `publish(const rcl_serialized_message_t * msg)` as mentioned below.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -172,34 +192,34 @@ class Publisher : public PublisherBase
...
-  void
-  publish(const rcl_serialized_message_t & serialized_msg)
+  /// Publish a serialized message. Non specialized version to prevent compiling errors.
+  template<typename TDeleter, typename T>
+  void publish(std::unique_ptr<T, TDeleter> serialized_msg)
   {
-    return this->do_serialized_publish(&serialized_msg);
+    (void)serialized_msg;
+    throw std::runtime_error(
+            "publishing unique_ptr with custom deleter only supported for serialized messages");
```
Karsten1987, line 215: why is that? IF called with our existing function this is templated on the `MessageT` as well as the `Deleter`

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -265,15 +330,24 @@ class Publisher : public PublisherBase
...
+      if (RCL_RET_OK != status) {
+        rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish serialized message");
+      }
     }
-    auto status = rcl_publish_serialized_message(&publisher_handle_, serialized_msg, nullptr);
-    if (RCL_RET_OK != status) {
-      rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish serialized message");
+
+    auto msg = std::make_unique<rclcpp::experimental::SerializedContainer>(
+      std::move(*serialized_msg));
```
Karsten1987, line 347: I have a problem with that. If I call publish with a serialized message allocated on the stack and passed to `publish(&msg)` I get a stack corruption when the unique_ptr tries to deallocate that message.

rclcpp/include/rclcpp/publisher_factory.hpp
```diff
@@ -63,17 +63,20 @@ struct PublisherFactory
...
 PublisherFactory
-create_publisher_factory(const rclcpp::PublisherOptionsWithAllocator<AllocatorT> & options)
+create_publisher_factory(
+  const rclcpp::PublisherOptionsWithAllocator<AllocatorT> & options,
+  const rosidl_message_type_support_t & type_support)
 {
   PublisherFactory factory {
     // factory function that creates a MessageT specific PublisherT
-    [options](
+    [options, type_support](
```
Karsten1987, line 72: wouldn't it suffice to get the typesupport here?

rclcpp/include/rclcpp/subscription_factory.hpp
```diff
@@ -88,7 +89,7 @@ create_subscription_factory(
 
   SubscriptionFactory factory {
     // factory function that creates a MessageT specific SubscriptionT
-    [options, msg_mem_strat, any_subscription_callback](
+    [options, msg_mem_strat, any_subscription_callback, type_support](
```
Karsten1987, line 92: same thing. I think that's the place to extract the typesupport.

rclcpp/src/rclcpp/intra_process_manager.cpp
```diff
@@ -56,7 +59,9 @@ IntraProcessManager::add_publisher(rclcpp::PublisherBase::SharedPtr publisher)
 }
 
 uint64_t
-IntraProcessManager::add_subscription(SubscriptionIntraProcessBase::SharedPtr subscription)
+IntraProcessManager::add_subscription(
+  SubscriptionIntraProcessBase::SharedPtr subscription,
+  const bool is_serialized)
```
Karsten1987, line 64: we don't need that parameter here. It's part of the subscription base: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/subscription_base.hpp#L203-L205

rclcpp/src/rclcpp/intra_process_manager.cpp
```diff
@@ -32,7 +32,9 @@ IntraProcessManager::~IntraProcessManager()
 {}
 
 uint64_t
-IntraProcessManager::add_publisher(rclcpp::PublisherBase::SharedPtr publisher)
+IntraProcessManager::add_publisher(
+  rclcpp::PublisherBase::SharedPtr publisher,
+  const bool is_serialized)
```
Karsten1987, line 37: I don't see an easy way to set a flag on the publisher base similar to the subscription base which indicates that it's a serialized publisher. Bummer.

Karsten1987: That being said, what happens if I take the same publisher and publish first a serialized message and then a regular `MsgT`?

rclcpp/src/rclcpp/intra_process_manager.cpp
```diff
@@ -220,6 +226,11 @@ IntraProcessManager::can_communicate(
     return false;
   }
 
+  // a publisher and a subscription with different content type can't communicate
+  if (sub_info.is_serialized != pub_info.is_serialized) {
+    return false;
+  }
```
Karsten1987, line 232: Just when I thought I understood what's happening the intraprocess manager, I read this comment and I am confused again. ;-)
I thought the four different variants are so that a non-serialized subscription could obtain values published from a serialized message?

rclcpp/src/rclcpp/intra_process_manager.cpp
Karsten1987: EDIT: I just browsed through the tests to see what's happening on the user side. What about we don't call a `create_publisher` function, but rather a `create_serialized_publisher` which implements the publisherbase interface (sets the `is_serialized` parameter) and only allows a call to publish with serialized messages.

rclcpp/include/rclcpp/create_publisher.hpp
```diff
@@ -66,6 +67,27 @@ create_publisher(
   return std::dynamic_pointer_cast<PublisherT>(pub);
 }
 
+template<
+  typename MessageT,
+  typename AllocatorT = std::allocator<void>,
+  typename PublisherT = rclcpp::Publisher<MessageT, AllocatorT>,
+  typename NodeT>
+std::shared_ptr<PublisherT>
+create_publisher(
```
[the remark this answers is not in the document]
DensoADAS: this allows an instantion with type "rcl_serialized_message_t" for generic publishers without previously known message type (e.g. for rosbag2)

Example:
`auto publisher = rclcpp::create_publisher<rcl_serialized_message_t>(
      node,
      "/topic",
      *rosidl_typesupport_cpp::get_message_type_support_handle<test_msgs::msg::Strings>(),
      rclcpp::QoS(10));`

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -76,6 +78,40 @@ create_subscription(
...
+  typename CallbackMessageT =
+  typename rclcpp::subscription_traits::has_message_type<CallbackT>::type,
+  typename SubscriptionT = rclcpp::Subscription<CallbackMessageT, AllocatorT>,
+  typename MessageMemoryStrategyT = rclcpp::message_memory_strategy::MessageMemoryStrategy<
+    CallbackMessageT,
+    AllocatorT
+  >,
+  typename NodeT>
+typename std::shared_ptr<SubscriptionT>
+create_subscription(
```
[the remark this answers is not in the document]
DensoADAS: same reason as for publisher

rclcpp/include/rclcpp/experimental/serialization.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include <rmw/rmw.h>
+
+#include <memory>
+#include <string>
+
+#include "rcl/error_handling.h"
+
+namespace rclcpp
+{
+namespace experimental
```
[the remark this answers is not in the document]
DensoADAS: is there a definition for "experimental"? Should we move it?

rclcpp/include/rclcpp/experimental/serialized_container.hpp
DensoADAS: agree

rclcpp/include/rclcpp/experimental/serialized_container.hpp
DensoADAS: the ownership is moved to the current object, yet the pointer in msg is (by default) not reset

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
DensoADAS: for this implementation rcl_serialized_message_t is supported

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
DensoADAS: I have added comments for clarification to the code

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
DensoADAS: to stasify the checks, that conversion to void* is intended

rclcpp/include/rclcpp/publisher_factory.hpp
DensoADAS: same as create_publisher

rclcpp/include/rclcpp/subscription_factory.hpp
DensoADAS: same as create_subscription

rclcpp/include/rclcpp/publisher.hpp
DensoADAS: I enabled on the cost of copying the message (otherwise it's not known when the data is/can be released)

rclcpp/include/rclcpp/experimental/serialization.hpp
```diff
@@ -0,0 +1,112 @@
...
+{
+
+/// Interface to (de)serialize a message
+class SerializationBase
+{
+public:
+  virtual ~SerializationBase() {}
+
+  virtual std::shared_ptr<rcl_serialized_message_t> serialize_message(const void * message) = 0;
```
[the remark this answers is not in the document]
DensoADAS: yeah, I understand your concern. The Serialization class hides the rmw functions. It's difficult to avoid the void* without knowing the actual content (cannot be templated here). For the current PR we skip changing this

DensoADAS:
@Karsten1987 Thank you for your review. I've implemented most of your comments (some stuff like raw pointers need more time) We'd appreciate it if the PR could make it in foxy. Have a nice, healthy and relaxing weekend!

rclcpp/include/rclcpp/experimental/serialization.hpp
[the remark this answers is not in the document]
Karsten1987: I forgot the rationale why we introduced "experimental". But I see the `Serialization` class being easily usable outside of the context of the intraprocess communication. Therefore, I'd propose putting it in the regular `rclcpp` namespace.

rclcpp/include/rclcpp/experimental/serialization.hpp
```diff
@@ -0,0 +1,125 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
Karsten1987, line 1:
```suggestion
// Copyright 2020 Open Source Robotics Foundation, Inc.
```

rclcpp/include/rclcpp/experimental/serialization.hpp
```diff
@@ -0,0 +1,125 @@
...
+    const auto ret = rmw_serialized_message_init(serialized_message, 0, &rcutils_allocator_);
+    if (ret != RCUTILS_RET_OK) {
+      throw std::runtime_error(
+              "Error allocating resources for serialized message: " +
+              std::string(rcutils_get_error_string().str));
+    }
+
+    if (nullptr == message) {
+      throw std::runtime_error("Message is nullpointer while serialization.");
+    }
```
Karsten1987, line 75: still leaking memory here. This check should be first thing in this function, before any allocation is done

rclcpp/include/rclcpp/experimental/serialized_message.hpp
```diff
@@ -0,0 +1,90 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
Karsten1987, line 1:
```suggestion
// Copyright 2020 Open Source Robotics Foundation, Inc.
```

rclcpp/include/rclcpp/experimental/serialized_message.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+/// Object oriented version of rcl_serialized_message_t with destructor to avoid memory leaks
+class SerializedMessage : public rcl_serialized_message_t
+{
+public:
+  SerializedMessage()
+  : rcl_serialized_message_t(rmw_get_zero_initialized_serialized_message())
+  {}
+
+  explicit SerializedMessage(const SerializedMessage & sc)
```
Karsten1987, line 37: can you simply call the other constructor (`SerializedMessage(const rcl_serialized_message_t & sc)`? Alternatively provide a generic `init` or `swap` function to de-duplicate the code within the two constructors.

Karsten1987, line 37:
```suggestion
  explicit SerializedMessage(const SerializedMessage & serialized_message)
```

rclcpp/include/rclcpp/experimental/serialized_message.hpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // do not call memcpy if the pointer is "static"
+    if (buffer != sc.buffer) {
+      std::memcpy(buffer, sc.buffer, sc.buffer_length);
+    }
+    buffer_length = sc.buffer_length;
+  }
+
+  explicit SerializedMessage(const rcl_serialized_message_t & sc)
```
Karsten1987, line 52:
```suggestion
  explicit SerializedMessage(const rcl_serialized_message_t & serialized_message)
```

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -191,34 +192,42 @@ class Publisher : public PublisherBase
       get_subscription_count() > get_intra_process_subscription_count();
 
     if (inter_process_publish_needed) {
-      auto shared_msg = this->do_intra_process_publish_and_return_shared(std::move(msg));
+      auto shared_msg = this->do_intra_process_publish_and_return_shared(
+        std::move(
```
Karsten1987, line 196: don't think we need that line break here.

rclcpp/src/rclcpp/intra_process_manager.cpp
Karsten1987: @DensoADAS can you comment on this?

rclcpp/src/rclcpp/subscription_base.cpp
```diff
@@ -218,23 +219,29 @@ SubscriptionBase::can_loan_messages() const
...
-            "SubscriptionBase::get_intra_process_waitable() called "
+            "SubscriptionBase::get_intra_process_waitables() called "
             "after destruction of intra process manager");
   }
 
-  // Use the id to retrieve the subscription intra-process from the intra-process manager.
-  return ipm->get_subscription_intra_process(intra_process_subscription_id_);
+  std::vector<rclcpp::Waitable::SharedPtr> waitables(intra_process_subscription_ids_.size());
+
+  for (size_t i = 0; i < intra_process_subscription_ids_.size(); ++i) {
```
Karsten1987, line 239:
```suggestion
  for (auto i = 0u; i < intra_process_subscription_ids_.size(); ++i) {
```

rclcpp/test/test_intra_process_communication.cpp
```diff
@@ -0,0 +1,367 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
Karsten1987, line 1:
```suggestion
// Copyright 2020 Open Source Robotics Foundation, Inc.
```

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -61,8 +61,9 @@ class LifecyclePublisher : public LifecyclePublisherInterface,
     rclcpp::node_interfaces::NodeBaseInterface * node_base,
     const std::string & topic,
     const rclcpp::QoS & qos,
-    const rclcpp::PublisherOptionsWithAllocator<Alloc> & options)
-  : rclcpp::Publisher<MessageT, Alloc>(node_base, topic, qos, options),
+    const rclcpp::PublisherOptionsWithAllocator<Alloc> & options,
+    const rosidl_message_type_support_t & type_support)
```
Karsten1987, line 65: Can we have an overloaded constructor for this to not break with existing code?
The first constructor could get the type support as done above with the macro from `MessageT`

rclcpp/include/rclcpp/experimental/serialization.hpp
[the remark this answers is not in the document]
DensoADAS: The message is not converted in this sense. The resulting shared pointer is a new object. (And I want to prevent objects without cleanup functions as much as possible, so I prioritized preventing memory loss)

rclcpp/include/rclcpp/experimental/serialized_message.hpp
```diff
@@ -0,0 +1,90 @@
...
+      std::memcpy(buffer, sc.buffer, sc.buffer_length);
+    }
+    buffer_length = sc.buffer_length;
+  }
+
+  explicit SerializedMessage(rcl_serialized_message_t && msg)
+  : rcl_serialized_message_t(msg)
+  {
+    // reset buffer to prevent double free
+    msg = rmw_get_zero_initialized_serialized_message();
```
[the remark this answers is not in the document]
DensoADAS: yes, deletion is now handled by SerializedMessage

rclcpp/include/rclcpp/experimental/serialized_message.hpp
DensoADAS: just removed it, as implicit conversion is possible

DensoADAS:
@Karsten1987 another update

Karsten1987:
@DensoADAS branch has (still) conflicts

rclcpp/include/rclcpp/experimental/serialization.hpp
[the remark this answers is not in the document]
Karsten1987: I thought I had mentioned this somewhere else already, but what about using the `SerializedMessage` for this?
```
virtual void serialize_message(const void * ros_message, SerializedMessage & serialized_message);

virtual void deserialize_message(const SerializedMessage & serialized_message, void * ros_message);
```

rclcpp/src/rclcpp/intra_process_manager.cpp
DensoADAS: I think the point here is that [here](https://github.com/DensoADAS/rclcpp/blob/241164204d5fe9fb87a885c25cc3d48c49aef06c/rclcpp/include/rclcpp/publisher.hpp#L135) always the serialized and not serialized publishers are added to the interprocess manager.

kfabian: I think the point here is that here always the serialized and not serialized publishers are added to the interprocess manager.

Karsten1987: Does this make sense though? Either the publisher is a `serialized` one or not, am I right?
That is, I think we can just only add one id plus the flag whether it's serialized or not.

DensoADAS:
@Karsten1987 We do not have an osx machine available. From your error log I assume there is a problem with overloading the "create_subscription" function and template deduction. Therefore I renamed the "generic" one (create_generic_subscription/create_generic_publisher). Could you test this? Thank you.

Karsten1987:
@DensoADAS not sure if you've seen this, but I've opened a PR towards your original branch which addresses this: https://github.com/DensoADAS/rclcpp/pull/1
There is also a second follow-up one, where I tried to summarize the other issues I had with this: https://github.com/DensoADAS/rclcpp/pull/2

Karsten1987:
So I've just played around with a bit more and I actually think we should re-architecture this PR, maybe break it up into a few smaller ones even.

1.) Introduce a C++ version for the serialization. That includes the `rclcpp::SerializedMessage` and the `rclcpp::Serialization`. Additionally, the callback signature should be changed/enhanced to take the C++ class. The need for wrapping this into a shared pointer should vanish with this as well as the destructor of the C++ instance can make sure that the buffer is cleanup correctly. 
That work was started in here: 
https://github.com/DensoADAS/rclcpp/pull/2/commits/d6bf9833b3bbe97cd57691c45f2edca65fa86735 & https://github.com/DensoADAS/rclcpp/pull/2/commits/dd6564518d498e55813962d3ea5607ac1746fdd1

2.) Create a template specialization for the publisher, which is instantiated only for `rclcpp::SerializedMessage`. This should be less impactful on the `rclcpp::Publisher`, the intra process manager and waitset API, as all the differentiation in the current PR (i.e. `std::is_same<T, rcl_serialized_message_t>`) can go away. The main motivation for this is that I don't see this scale up if new functionality gets added to the publisher as again the functions have to be cross-checked with a serialized message.
That also removes the need to constantly add two `id`s to the intra-process manager. This specialization could then further take the typesupport as a argument in the constructor whereas the other publisher does not need to be changed at all.
Work started here: https://gist.github.com/Karsten1987/bae218680ab400292dabd4537e516f64
That specialization has further benefits that all ambiguous member functions can go. In this specialization, there is no need to support a publish to a ROS message or if so, only provide a `const void * ros_msg` function. Similarly, a `borrow_loaned_message` can be implemented differently for this specialization - or not at all.
The differentiation template to be used can be done solely in the `rclcpp::create_publisher` function.
Also, in this case, I'd recommend to deprecate the `publish(const rcl_serialized_message_t & msg)` function overload on the regular Publisher.

3.) The same thing could technically be done for the subscription class as well. If the user wants to retrieve the data in a serialized way, `rclcpp::create_subscription` returns a specialized version of it. That doesn't really matter as for now given that we anyway hide this already away, but I think it would ease the change needed for the intra process communication. 

---

The changes in 2.) might have drawbacks as to two different publisher instances have to be used when trying to publish ROS messages vs serialized messages. However, I generally also don't really see a big use-case for this in the first place. I could see how one wants to use one or the other publisher - c.f. rosbags when not being able to depend on the message types during compile time, but never really the mixed version.

DensoADAS:
@Karsten1987 I replaced this PR by 3 new ones

PR #977, opened 2020-01-31
code style only: wrap after open parenthesis if not in one line

dirk-thomas:
Style update to match the ROS 2 development guide and pass with the updated linter configuration from ament/ament_lint#210.

### DIFF
--- a/rclcpp/test/test_node.cpp
+++ b/rclcpp/test/test_node.cpp
@@ -687,7 +687,8 @@ TEST_F(TestNode, declare_parameter_with_cli_overrides) {
     test_resources_path / "test_parameters.yaml").string();
   // test cases with overrides
   rclcpp::NodeOptions no;
-  no.arguments({
+  no.arguments(
+  {
     "--ros-args",
     "-p", "parameter_bool:=true",
     "-p", "parameter_int:=42",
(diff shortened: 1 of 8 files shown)


rclcpp/test/test_node.cpp
```diff
@@ -687,7 +687,8 @@ TEST_F(TestNode, declare_parameter_with_cli_overrides) {
     test_resources_path / "test_parameters.yaml").string();
   // test cases with overrides
   rclcpp::NodeOptions no;
-  no.arguments({
+  no.arguments(
+  {
```
wjwwood: This one doesn't make sense to me, but whatever.

rclcpp/test/test_node_options.cpp
```diff
@@ -35,14 +35,16 @@ TEST(TestNodeOptions, ros_args_only) {
   ASSERT_EQ(0, rcl_arguments_get_count_unparsed_ros(&rcl_options->arguments));
 
   char * node_name = nullptr;
-  EXPECT_EQ(RCL_RET_OK, rcl_remap_node_name(
+  EXPECT_EQ(
+    RCL_RET_OK, rcl_remap_node_name(
```
wjwwood: This one is odd to me too. I understand why it did this one, but it would be better if there was a "don't wrap all if only the last (or only) argument is multiline", but I guess it probably doesn't have that.

PR #998, opened 2020-02-20
Ensure logging is initialized just once

ivanpauno:
Alternative to ros2/rcl#560, based on https://github.com/ros2/rcl/pull/560#discussion_r373282259.

I didn't find a solution that:
- Doesn't break ABI.
- Doesn't have problems if somebody creates a global context in another library (both order of inizialization and order of destruction).

I chose to break ABI, and avoid the second problem (though that unfortunatelly will block this from being backported as-is).
I'm avoiding to finish the logger when all context were shut down, and I'm calling `rcl_logging_fini` when all the contexts were finished (not shutdown).

P.S.: The current `g_contexts_mutex` might cause both initialization/destruction order problems if somebody creates a global context in another library, but that's subject of another PR.

Connects to ros2/rcl#579

### DIFF
--- a/rclcpp/include/rclcpp/init_options.hpp
+++ b/rclcpp/include/rclcpp/init_options.hpp
@@ -42,7 +42,7 @@ public:
   RCLCPP_PUBLIC
   InitOptions(const InitOptions & other);
 
-  /// Return `true` if logging should be initialized.
+  /// Return `true` if logging should be initialized when `rclcpp::Context::init` is called.
   RCLCPP_PUBLIC
   bool
   auto_initialize_logging() const;
@@ -72,7 +72,7 @@ protected:
 
 private:
   std::unique_ptr<rcl_init_options_t> init_options_;
-  bool initialize_logging_ = true;
+  bool initialize_logging_{true};
 };
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/init_options.hpp
+++ b/rclcpp/include/rclcpp/init_options.hpp
@@ -42,6 +42,16 @@ public:
   RCLCPP_PUBLIC
   InitOptions(const InitOptions & other);
 
+  /// Return `true` if logging should be initialized.
+  RCLCPP_PUBLIC
+  bool
+  initialize_logging() const;
+
+  /// Set flag indicating if logging should be initialized or not.
+  RCLCPP_PUBLIC
+  InitOptions &
+  initialize_logging(bool initialize_logging);
+
   /// Assignment operator.
   RCLCPP_PUBLIC
   InitOptions &
@@ -62,6 +72,7 @@ protected:
 
 private:
   std::unique_ptr<rcl_init_options_t> init_options_;
+  bool initialize_logging_ = true;
 };
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/init_options.hpp
+++ b/rclcpp/include/rclcpp/init_options.hpp
@@ -45,12 +45,12 @@ public:
   /// Return `true` if logging should be initialized.
   RCLCPP_PUBLIC
   bool
-  initialize_logging() const;
+  auto_initialize_logging() const;
 
   /// Set flag indicating if logging should be initialized or not.
   RCLCPP_PUBLIC
   InitOptions &
-  initialize_logging(bool initialize_logging);
+  auto_initialize_logging(bool initialize_logging);
 
   /// Assignment operator.
   RCLCPP_PUBLIC
--- a/rclcpp/src/rclcpp/context.cpp
+++ b/rclcpp/src/rclcpp/context.cpp
@@ -22,6 +22,7 @@
 #include <utility>
 
 #include "rcl/init.h"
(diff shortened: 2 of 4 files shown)


ivanpauno:
Connected to https://github.com/ros2/rcl/pull/579.

wjwwood:
> Doesn't have problems if somebody creates a global context in another library (both order of inizialization and order of destruction).

I don't think this can be addressed until logging is no longer required to be global.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -34,8 +35,31 @@ static std::vector<std::weak_ptr<rclcpp::Context>> g_contexts;
 
 using rclcpp::Context;
 
+static
+std::shared_ptr<std::mutex>
+get_global_logging_configure_mutex()
+{
+  static auto mutex = std::make_shared<std::mutex>();
```
wjwwood: Because of this, I think it should be possible to create a Context without initializing logging. At least then a user could have their own globals and avoid the global shutdown issues if they manage to init/shutdown logging themselves.

As-is, logging is always configured unconditionally.

wjwwood: Probably an option in `rclcpp::InitOptions` and don't get the global mutex in the constructor, instead acquire it conditionally in the `init()` method.

fujitatomoya: @wjwwood @ivanpauno 

maybe it is better for user application to get the flexibility via rclcpp::InitOption, i do agree on this very much.

rclcpp/include/rclcpp/init_options.hpp
```diff
@@ -42,6 +42,16 @@ class InitOptions
   RCLCPP_PUBLIC
   InitOptions(const InitOptions & other);
 
+  /// Return `true` if logging should be initialized.
+  RCLCPP_PUBLIC
+  bool
+  initialize_logging() const;
```
wjwwood: nitpick: this sounds oddly imperative, like it should actually initialize the logging when called. I know it's an accessor to this member variable, but I had to double take when reading it the first time. Maybe `should_initialize_logging` or `auto_initialize_logging` would be more obvious as a configuration rather than an action.

But also feel free to leave as-is if you want.

ivanpauno: done

ivanpauno:
CI jobs in https://github.com/ros2/rcl/pull/579#issuecomment-595304625.

rclcpp/include/rclcpp/init_options.hpp
```diff
@@ -62,6 +72,7 @@ class InitOptions
 
 private:
   std::unique_ptr<rcl_init_options_t> init_options_;
+  bool initialize_logging_ = true;
```
hidmic, line 75: @ivanpauno nit^N: 
```suggestion
  bool initialize_logging_{true};
```

rclcpp/include/rclcpp/init_options.hpp
```diff
@@ -42,6 +42,16 @@ class InitOptions
   RCLCPP_PUBLIC
   InitOptions(const InitOptions & other);
 
+  /// Return `true` if logging should be initialized.
```
hidmic, line 45: @ivanpauno initialized when? I presume when the context this options apply to is. Mind to be explicit?

rclcpp/src/rclcpp/context.cpp
```diff
@@ -34,8 +35,27 @@ static std::vector<std::weak_ptr<rclcpp::Context>> g_contexts;
...
+  return mutex;
+}
+
+static
+size_t &
+get_logging_reference_count()
+{
+  static size_t ref_count = 0;
+  return ref_count;
+}
```
hidmic, line 52: @ivanpauno meta: I won't block this PR on it, but I _think_ this reference counting should be down in `rcl`. We haven't explored multi-language ROS2 executables much yet, but this could be a problem if we (or anyone) wanted to go down that path.

ivanpauno: IMHO, no.
We probably want to drop the message count in the future and have a more elegant solution, and so I would avoid adding a reference count API in `rcl`.

rclcpp/include/rclcpp/init_options.hpp
ivanpauno: https://github.com/ros2/rclcpp/pull/998/commits/0c95f168d584923b8ccb00d6080561df82b95db5

rclcpp/include/rclcpp/init_options.hpp
ivanpauno: See https://github.com/ros2/rclcpp/pull/998/commits/0c95f168d584923b8ccb00d6080561df82b95db5

rclcpp/src/rclcpp/context.cpp
hidmic: I see. Then if this aims to be just a temporary solution, a TODO and/or an follow-up issue would be nice.

PR #999, opened 2020-02-20
Ensure all Clock accesses are thread-safe

hidmic:
Precisely what the title says. Before this patch, both `rclcpp::Clock` and `rclcpp::TimerBase` (though indirectly through `rcl`) would access non thread-safe `rcl_clock_*` API without locking, leading to internal state corruption on concurrent access in multi-threaded applications.

This change is ABI breaking w.r.t. to previous distro releases (e.g. Eloquent) but sets the scene for future non ABI breaking changes using a PIMPL.

### DIFF
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -23,6 +23,30 @@
 namespace rclcpp
 {
 
+class Clock::Impl
+{
+public:
+  explicit Impl(rcl_clock_type_t clock_type)
+  : allocator_{rcl_get_default_allocator()}
+  {
+    rcl_ret_t ret = rcl_clock_init(clock_type, &rcl_clock_, &allocator_);
+    if (ret != RCL_RET_OK) {
+      exceptions::throw_from_rcl_error(ret, "could not get current time stamp");
+    }
+  }
+
+  ~Impl()
+  {
+    rcl_ret_t ret = rcl_clock_fini(&rcl_clock_);
+    if (ret != RCL_RET_OK) {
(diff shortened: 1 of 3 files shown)


dirk-thomas:
It would be good to keep these two changes (PIMPL and adding a mutex) separate since together in a single commit it just makes the review harder.

hidmic:
@dirk-thomas fair enough, see https://github.com/ros2/rclcpp/pull/999/commits/a91ba82799c09fa8592a19c58cc807f0e10821d5 and https://github.com/ros2/rclcpp/pull/999/commits/f4ff7f1a55508551d4ec4793007774471d2e5792.

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -15,6 +15,8 @@
 #include "rclcpp/clock.hpp"
 
 #include <memory>
+#include <mutex>
```
dirk-thomas: Nitpick: since the `clock.hpp` includes this header already we usually don't repeat the include in the `cpp` file.

hidmic: Last commit amended to keep history clean.

hidmic:
CI up to `rclcpp` and `test_rclcpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9404/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5081/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7696/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9322/)

hidmic:
Alright, going in!

PR #986, opened 2020-02-09
Add missing service callback registration tracepoint

christophebedard:
This was an oversight on my part when writing the instrumentation. See https://gitlab.com/micro-ROS/ros_tracing/tracetools_analysis/issues/25#note_284822856.

I've updated the `ros2_tracing` tests to cover this: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/merge_requests/138

I'd like for this to be considered for a backport to eloquent (https://github.com/ros2/ros2/issues/859)!

PR #1001, opened 2020-02-25
Add ifdefs to remove tracing-related calls if tracing is disabled

christophebedard:
While it's not the default value, `TRACETOOLS_DISABLED` can be defined through `ros2_tracing`/`tracetools` to remove all `TRACEPOINT` calls. However, we never applied this logic to some other tracing-related function calls.

~~Based on #986 because it was just simpler this way.~~ Rebased

### DIFF
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -100,6 +100,7 @@ public:
     TRACEPOINT(callback_end, (const void *)this);
   }
 
+#ifndef TRACETOOLS_DISABLED
   void register_callback_for_tracing()
   {
     if (shared_ptr_callback_) {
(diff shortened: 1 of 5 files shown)


rclcpp/include/rclcpp/any_service_callback.hpp
```diff
@@ -100,6 +100,7 @@ class AnyServiceCallback
     TRACEPOINT(callback_end, (const void *)this);
   }
 
+#ifndef TRACETOOLS_DISABLED
```
dirk-thomas: Is it a good idea to change the API based on the define? Code calling this API must then also ifdef the call. Also building with/without the define changes API/ABI.

How about putting the ifdef only around the content of the function?

christophebedard: Done in 25dda366a2ed498a892d7fc13b2c055c5bc5433f (while keeping the `#ifndef`s around the function calls)

dirk-thomas:
Thanks for the patch.

Issue #1012, opened 2020-03-04
Timer does not working when reset a cancelled timer


DongheeYe:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - from source 
- Version or commit hash:
  - Dashing and latest
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
I tested it using changed code from https://github.com/ros2/examples/blob/master/rclcpp/minimal_timer/lambda.cpp
```cpp
#include <chrono>
#include <memory>
#include <glib.h>


#include "rclcpp/rclcpp.hpp"

using namespace std::chrono_literals;

class MinimalTimer : public rclcpp::Node
{
public:
  MinimalTimer()
  : Node("minimal_timer")
  {
    auto timer_callback = [this]() -> void {
      RCLCPP_INFO(this->get_logger(), "Timer Expired. Timer cancel");
      timer_->cancel();
    };
    timer_ = create_wall_timer(500ms, timer_callback);
  }

// for test, set public visibility
// private:
   rclcpp::TimerBase::SharedPtr timer_;
   rclcpp::TimerBase::SharedPtr timer2_;
};

GMainLoop *mainloop;
std::shared_ptr<rclcpp::executors::SingleThreadedExecutor> exec;
std::shared_ptr<MinimalTimer> node;

gboolean TimerExpired(gpointer data) {
   if (node->timer_->is_canceled()){
     RCLCPP_INFO(node->get_logger(), "Timer is cancelled and reset");
     node->timer_->reset();
   }else {
     RCLCPP_INFO(node->get_logger(), "Timer is not cancelled");
   }
   return true;
}

int main(int argc, char * argv[])
{
  rclcpp::init(argc, argv);

  exec = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
  node = std::make_shared<MinimalTimer>();
  exec->add_node(node);

  mainloop = g_main_loop_new(nullptr, false);

  g_timeout_add(1000, TimerExpired, nullptr);

  g_thread_new(nullptr,
        [](gpointer data) -> gpointer {
          exec->spin();
          return nullptr;
        },
        static_cast<void *>(node.get()));

  g_main_loop_run(mainloop);
  rclcpp::shutdown();
  return 0;
}

```

And change rclcpp/minimal_timer/CMakeLists.txt
```diff
 find_package(ament_cmake REQUIRED)
[...]

PR #1013, opened 2020-03-04
[timer] Fix malfunction when reset cancelled timer

DongheeYe:
Signed-off-by: Donghee Ye <donghee.ye@samsung.com>

Related to https://github.com/ros2/rclcpp/issues/1012

To reset wait_set after reset a cancelled timer, there is needed node handle or context handle. When constructing a timer, there are context handle parameter. To trigger guard condition to reset wait_set, save context handle in constructor and trigger when reset timer.

### DIFF
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -104,6 +104,7 @@ public:
 protected:
   Clock::SharedPtr clock_;
   std::shared_ptr<rcl_timer_t> timer_handle_;
+  rclcpp::Context::SharedPtr context_;
 };
 
 
--- a/rclcpp/src/rclcpp/timer.cpp
+++ b/rclcpp/src/rclcpp/timer.cpp
@@ -72,6 +72,7 @@ TimerBase::TimerBase(
       rcl_reset_error();
     }
   }
+  context_ = context;
 }
 
 TimerBase::~TimerBase()
@@ -102,6 +103,7 @@ TimerBase::reset()
   if (rcl_timer_reset(timer_handle_.get()) != RCL_RET_OK) {
     throw std::runtime_error(std::string("Couldn't reset timer: ") + rcl_get_error_string().str);
   }
+  context_->interrupt_all_wait_sets();
 }
 
 bool


ivanpauno:
@DongheeYe Thanks for opening the PR.
There seems to be a bunch of issues here:
- Methods of the `Timer` class should be thread safe, and they don't seem to be.
- Timers should trigger a guard condition when being reset/cancelled, but that should be done in `rcl` and not here. Probably, [here](https://github.com/ros2/rcl/blob/40ba55bc1743b480b3e1c76f6c3c009e13b07412/rcl/src/rcl/timer.c#L394) and [here](https://github.com/ros2/rcl/blob/40ba55bc1743b480b3e1c76f6c3c009e13b07412/rcl/src/rcl/timer.c#L375). [This](https://github.com/ros2/rcl/blob/40ba55bc1743b480b3e1c76f6c3c009e13b07412/rcl/src/rcl/timer.c#L37) same guard condition should be triggered. I'm not sure about if should be triggered when it's cancelled/reset or both.

DongheeYe:
@ivanpauno Thank to comments. I'd considered to use the rcl timer guard condition, either. ([here](https://github.com/ros2/rcl/blob/40ba55bc1743b480b3e1c76f6c3c009e13b07412/rcl/src/rcl/timer.c#L37)). However, I thought, the purpose of current rcl timer guard condition is to wake up of waitset when waiting timer is expired. In this case, cancelled timer guard condition is not added to the node waitset [rcl_wait](https://github.com/ros2/rcl/blob/40ba55bc1743b480b3e1c76f6c3c009e13b07412/rcl/src/rcl/wait.c#L551). So, additional guard condition is needed to wake up node waitset to add timer waitset again like. I will try to make a rcl patch that add additional guard condition to rcl timer and trigger when rcl_timer_reset.

ivanpauno:
> In this case, cancelled timer guard condition is not added to the node waitset rcl_wait

I think you can change the logic to make sure is always added, instead of adding a new guard condition.

wjwwood:
Interrupting all wait set associated with the context is overkill. It should only interrupt the wait set/executor of which the timer is associated. Pub/sub accomplish this by being lazily removed when the executor notices that the publisher or subscription have gone out of scope (it stores them as weak pointers), but they don't have a "reset" like function.

I think the best thing to do would be to get the node's node_base_interface and trigger the guard condition there, like `NodeTimers::add_timer()` does:

https://github.com/ros2/rclcpp/blob/96ebf59a6045a535730d98fff25e522807c7aa75/rclcpp/src/rclcpp/node_interfaces/node_timers.cpp#L42-L46

You could add a method to the `NodeTimersInterface` that does this and call it from `NodeTimers::add_timer()` as well as `TimerBase::reset()`.

ivanpauno:
Closed in favor of https://github.com/ros2/rcl/pull/589 (I added the reviewers here by error, the intention was to add them to the rcl PR).

ivanpauno:
> Methods of the Timer class should be thread safe, and they don't seem to be.

This comment is still valid. `rcl` functions aren't thread safe (they aren't intended to be), and `rclcpp` should provide thread safety on top of that. It would be good to add a mutex (a recursive mutex sounds suitable) to protect access to `timer_handle_` member of `Timer` class.

Can you open a PR proposing that?

PR #1020, opened 2020-03-09
expose get_period and set_period from rcl

jabbenseth:
Expose the `std::chrono::nanoseconds get_period()` and `set_period(std::chrono::nanoseconds)` functions of the rcl timers to rclcpp.

reset behaves similar to the reset parameter of ros::Timer::setPeriod

### DIFF
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -76,7 +76,7 @@ public:
 
   RCLCPP_PUBLIC
   void
-  set_period(std::chrono::nanoseconds new_period, bool reset=true);
+  set_period(std::chrono::nanoseconds new_period, bool reset = true);
 
   RCLCPP_PUBLIC
   std::chrono::nanoseconds
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -75,6 +75,14 @@ public:
   reset();
 
   RCLCPP_PUBLIC
+  void
+  set_period(std::chrono::nanoseconds new_period, bool reset=true);
+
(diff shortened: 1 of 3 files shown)


claireyywang:
what's the benefit of this change?

ipa-jba:
The perceived benefit from that change came up while porting something like the "twist_mux" from ROS to ROS 2 where one Timer was used as Watchdog, where different input Topics have different durations of being the valid input before falling back to a lower priority input.

Where the ROS Code was
```
    if (prio >= last_prio_) {
      last_prio_ = prio;
      output_.publish(t);
      timer_.setPeriod(decay, true);
    }
```
and used one Timer all the time, the current ROS 2 Code (without this change) creates new timers in every callback 
```
    if (prio >= last_prio_)
    {
      last_prio_ = prio;
      output_->publish(t);
      timer_ = nh_->create_wall_timer(decay, [this]() -> void { last_prio_ = 0; });
```
The intention was to save some allocation/reallocation efforts by reusing an existing timer

ivanpauno (changes requested):
@ipa-jba Thanks for opening the PR, this addition makes sense to me.

Can you target the PR to `master` instead of eloquent?
Could you add a test case that involves a more real scenario, e.g. modifying period while spinning or modifying the period when some condition happens in the timer callback, etc.

fujitatomoya (commented):
overall lgtm with https://github.com/ros2/rclcpp/pull/1020#pullrequestreview-579603960

fujitatomoya:
@jabbenseth are you planning to work on this?

clalancette:
No response to this in quite a long time, so I'm going to go ahead and close it out.  Feel free to reopen if you still want to work on it.

PR #1024, opened 2020-03-12
Let integer values be assigned to double parameters

### LINKED ISSUE #979: Implicitly cast integer values for double parameters
#### Feature description

If we try to pass an integer value to a double parameter from the command line or from a parameters YAML file we get a `rclcpp::ParameterTypeException`. For example, passing a parameter from the command line:

```
ros2 run foo_package foo_node --ros-args -p foo_arg:=1
```

results in the following error:

```
terminate called after throwing an instance of 'rclcpp::ParameterTypeException'
  what():  expected [double] got [integer]
``` 

and we can fix it by explicitly making our value a floating point number:

```
ros2 run foo_package foo_node --ros-args -p foo_arg:=1.0
```

But, it seems reasonable to me that if a user forgets to explicitly provide a floating point value that we should implicitly cast an integer to a float (as is done in many programming languages).

jacobperron:
Fixes #979.

When constructing integer ParameterValue objects, also set the double member.
When requesting a double value for ParameterValue objects of type integer, do not throw an exception.

### DIFF
--- a/rclcpp/include/rclcpp/parameter_value.hpp
+++ b/rclcpp/include/rclcpp/parameter_value.hpp
@@ -144,7 +144,7 @@ public:
 
   template<ParameterType type>
   constexpr
-  typename std::enable_if<type == ParameterType::PARAMETER_BOOL, const bool &>::type
+  typename std::enable_if<type == ParameterType::PARAMETER_BOOL, bool>::type
   get() const
   {
     if (value_.type != rcl_interfaces::msg::ParameterType::PARAMETER_BOOL) {
@@ -155,7 +155,7 @@ public:
 
   template<ParameterType type>
   constexpr
-  typename std::enable_if<type == ParameterType::PARAMETER_INTEGER, const int64_t &>::type
+  typename std::enable_if<type == ParameterType::PARAMETER_INTEGER, int64_t>::type
   get() const
   {
     if (value_.type != rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER) {
@@ -166,7 +166,7 @@ public:
 
   template<ParameterType type>
   constexpr
-  typename std::enable_if<type == ParameterType::PARAMETER_DOUBLE, const double &>::type
+  typename std::enable_if<type == ParameterType::PARAMETER_DOUBLE, double>::type
   get() const
   {
     if (value_.type != rcl_interfaces::msg::ParameterType::PARAMETER_DOUBLE) {
@@ -250,7 +250,7 @@ public:
 
   template<typename type>
   constexpr
-  typename std::enable_if<std::is_same<type, bool>::value, const bool &>::type
+  typename std::enable_if<std::is_same<type, bool>::value, bool>::type
   get() const
   {
     return get<ParameterType::PARAMETER_BOOL>();
@@ -259,7 +259,7 @@ public:
   template<typename type>
   constexpr
   typename std::enable_if<
-    std::is_integral<type>::value && !std::is_same<type, bool>::value, const int64_t &>::type
+    std::is_integral<type>::value && !std::is_same<type, bool>::value, int64_t>::type
   get() const
   {
     return get<ParameterType::PARAMETER_INTEGER>();
@@ -267,7 +267,7 @@ public:
 
   template<typename type>
   constexpr
-  typename std::enable_if<std::is_floating_point<type>::value, const double &>::type
+  typename std::enable_if<std::is_floating_point<type>::value, double>::type
   get() const
   {
     return get<ParameterType::PARAMETER_DOUBLE>();
--- a/rclcpp/include/rclcpp/parameter_value.hpp
+++ b/rclcpp/include/rclcpp/parameter_value.hpp
@@ -169,10 +169,7 @@ public:
   typename std::enable_if<type == ParameterType::PARAMETER_DOUBLE, const double &>::type
   get() const
   {
-    // Allow integers to be cast to doubles
-    if (value_.type != rcl_interfaces::msg::ParameterType::PARAMETER_DOUBLE &&
-      value_.type != rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER)
-    {
+    if (value_.type != rcl_interfaces::msg::ParameterType::PARAMETER_DOUBLE) {
       throw ParameterTypeException(ParameterType::PARAMETER_DOUBLE, get_type());
     }
     return value_.double_value;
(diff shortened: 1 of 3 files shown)


rclcpp/src/rclcpp/parameter_value.cpp
```diff
@@ -189,12 +193,16 @@ ParameterValue::ParameterValue(const std::vector<bool> & bool_array_value)
 ParameterValue::ParameterValue(const std::vector<int> & int_array_value)
 {
   value_.integer_array_value.assign(int_array_value.cbegin(), int_array_value.cend());
+  // Allow integer arrays to be cast to double arrays
+  value_.double_array_value.assign(int_array_value.cbegin(), int_array_value.cend());
```
wjwwood: I'm not a fan of this. For primitive types it is not as bad (though I still don't like the idea that the value has two values stored), but for the array I think there's a chance that could be lots of overhead.

clalancette:
> I would not advocate for changing the `ParameterValue::get<vector<int/double>/string>` to return a copy rather than a const reference. Because we'd like people to be able to access their parameter values without memory allocations. However, we could consider adding a `get_copy` for these, or even change `get` to return a copy but add a `get_ref` for these types, and people who care about the memory allocations could explicitly use `get_ref`.

I'm generally fine with a `get`/`get_copy` split (or however we want to split them), though I am concerned that their semantics will be different.  That is, I don't want it to be the case that `get_copy` can take an integer for a double parameter while `get` cannot.  If we can make the semantics exactly the same for both, then I'd be all for that split.

wjwwood:
>  If we can make the semantics exactly the same for both, then I'd be all for that split.

The only reason to have both is so the automatic conversion can happen... Otherwise you could just have a const and non-const method for each called `get()`, i.e. overloading.

The idea was, @jacobperron correct me if I'm wrong, to have `get()` never convert (because it has to store both versions because it cannot convert in place because that would return a copy to a temporary) and have `get_copy()` where it can convert (because it is returning a value not a reference).

clalancette:
> The only reason to have both is so the automatic conversion can happen... Otherwise you could just have a const and non-const method for each called `get()`, i.e. overloading.

> The idea was, @jacobperron correct me if I'm wrong, to have `get()` never convert (because it has to store both versions because it cannot convert in place because that would return a copy to a temporary) and have `get_copy()` where it can convert (because it is returning a value not a reference).

Personally, I think that is a worse place to be in from an API point of view.

It means you'll end up with some downstream API users using `get()`, some using `get_copy()`, and inconsistent conversion semantics.

As an alternative, I wonder if it would be better to just improve the error messages that people get when we can't do conversions.  Right now it's pretty unclear both why the conversion can't happen, and which parameter is causing the problem.  Adding error messages explaining both of those things would go a long way towards improving the user experience.

wjwwood:
Personally, I see the frustration from the user perspective, but I also think it's not our place to do this conversion for them, simply because `std::variant` doesn't do this for you:

```c++
std::variant<int, double> foo{1};
std::get<double>(f);  // exception here
```

Whereas in the proposed example of ours:

```c++
ParameterValue p(1);
p.get<double>();  // exception right now, no exception with this pr
```

clalancette:
> simply because `std::variant` doesn't do this for you:

Just to clarify: would you say that you agree with my proposal from the previous comment?  That is, we should not take this PR and instead improve the error messages that happen when automatic conversion fails?

wjwwood:
Sort of I guess.

I would like to solve the issue for users since it is annoying, but on the other hand its very tricky to do correctly and is part of the reason why std::variant doesn't do it for you...

@jacobperron's original pr addresses your concern @clalancette, but it requires that you store all integers as doubles as well. And if we ever updated to use std::variant instead then we wouldn't be able to continue doing that.

jacobperron:
I agree that having separate APIs for getting copies and references is not a great solution. What if we just do the conversion for the primitive (i.e int -> double), but not for the array variant? And then we can provide a good error message for the array case.

I think we can accomplish this by returning primitives by value (always), and returning arrays by reference. I think the overhead for returning primitives by value is negligible.

jacobperron:
@wjwwood @clalancette I've reverted the original commit (b65f95e) and added two new changes:

1. Return primitive parameter values by value (9d17592)
2. Cast integer parameters to double if requested (b67522e)

If users try to get a double array from an integer array they will still get a `ParameterType` exception. I'll look into improving the error message for users in a separate PR.

rclcpp/include/rclcpp/parameter_value.hpp
```diff
@@ -166,13 +166,17 @@ class ParameterValue
 
   template<ParameterType type>
   constexpr
-  typename std::enable_if<type == ParameterType::PARAMETER_DOUBLE, const double &>::type
+  typename std::enable_if<type == ParameterType::PARAMETER_DOUBLE, double>::type
   get() const
   {
-    if (value_.type != rcl_interfaces::msg::ParameterType::PARAMETER_DOUBLE) {
-      throw ParameterTypeException(ParameterType::PARAMETER_DOUBLE, get_type());
+    // Allow integer values to be cast to double
```
wjwwood: We should really document this somewhere, maybe in the high level `get/declare_parameter()` APIs, or maybe just right here.

clalancette:
> This works for me, but I'm interested to hear from @clalancette again.

I hate to be the naysayer, but I still think that having it different in different contexts is a worse API.  That is, the fact that it is different between arrays and primitives just feels really ugly to me.  It's also kicking the can down the road; we're going to get exactly the same complaint about arrays before too long.

I have a longer post to write here to try and come up with yet another solution.  I'll get to that later today, but I wanted to get my thoughts in now since this has been sitting for a couple of days.

clalancette:
Sorry this totally fell off of my radar.  Here are some further thoughts.

The ultimate goal here is to make things easier for users, so that they don't have to write `0.0` and instead can just write `0`.  The first question to answer, which I don't think we've examined enough, is whether we actually want to do this.  I'm personally a fan of being very explicit, so I'd vote for *not* doing this, and instead improve the error messages so that it is explicit about what is failing.  I think we've done this somewhat elsewhere; if need be, we can press more on that front.  Maybe it is worthwhile doing a discourse post or similar to get other ideas here?

(I may have some ideas on different ways we could accomplish it if we decide to go forward, but I think we really need to answer that question first)

jacobperron:
I'm okay with forcing users to be explicit. This change was driven by interactions with users that thought that this implicit cast should happen for them, and I think we'll get similar feedback on Discourse, but we should make a post to confirm.

If we do decide to go forward with this change, it sounds like it will need some significant refactoring so it's not going to make it for Foxy. We can table it until after Foxy.

jacobperron:
Here's an improvement on the error message: https://github.com/ros2/rclcpp/pull/1027

jacobperron:
Closing this since I don't have any intention on completing it.

PR #1031, opened 2020-03-23
[rclcpp] API review March 2020

wjwwood:
Record of API review process for `rclcpp`.

For external observers, we're doing this review in stages. I will circulate for public feedback after the core team has had a few rounds with it to avoid unnecessarily repeating suggestions (better to say "please add to this" rather than "just send me suggestions").

### DIFF
--- a/rclcpp/doc/api_review_march_2020.md
+++ b/rclcpp/doc/api_review_march_2020.md
@@ -91,9 +91,18 @@
 - https://github.com/ros2/rclcpp/issues/506
 - https://github.com/ros2/rclcpp/issues/726
 
+**Suggested Action**: Consolidate to a single issue, and defer.
+
 **Notes from 2020-03-23**:
 
-- 
+- (chris) Putting ownership mechanics on user is hard.
+- (dirk) add documentation clearly outlining ownership
+- (shane) warn on unused to catch issues with immediately deleted items
+- (tfoote) debugging output for destruction so it easy to see when reviewing logs
+- (chris) possible to create API that checks for destruction
+  - (william) might lead to complex synchronization issues
+- (tfoote) could add helper classes to make scoped things non-scoped
+  - (shane) concerned that there is no longer "one good way" to do it
 
 ### Allow QoS to be configured externally, like we allow remapping of topic names
 
@@ -208,11 +217,15 @@
 
 - https://github.com/ros2/rclcpp/issues/492
 
-**Suggested Action**: ?
+**Suggested Action**: Defer until after foxy.
 
 **Notes from 2020-03-23**:
 
-- 
+- (dirk) may be related to ROS 1 heritage of argc/argv being passed to node directly
+- (shane) impacts rcl API as well, two parts "global options" as well node specific options
+- (dirk) what is the recommendation to users that want to add arguments programmatically
+  - user should be able to get non-ros argc/argv somehow (seems like you can now)
+- (jacob) the argument in NodeOptions are used for application specific argument via component loading as well
 
 ## Timer
 
@@ -257,11 +270,11 @@
 
 - https://github.com/ros2/rclcpp/issues/491
 
-**Suggested Action**: ?
+**Suggested Action**: Defer.
 
 **Notes from 2020-03-23**:
 
-- 
+- (dirk) likely new API, so possible to backport
 
 ## Service Client
 
@@ -287,11 +300,19 @@
 - https://github.com/ros2/rclcpp/issues/975
 - https://github.com/ros2/demos/blob/948b4f4869298f39cfe99d3ae517ad60a72a8909/demo_nodes_cpp/src/services/add_two_ints_client.cpp#L24-L39
 
-**Suggested Action**: ?
+**Suggested Action**: Update issue and defer. Also defer decision on reconciling rclpy's send_request.
 
 **Notes from 2020-03-23**:
 
-- 
+- (karsten/shane) async spinner helps in rclpy version, rclcpp could emulate
+- (chris) sees three options:
+  - only async (current case in rclcpp)
+  - have sync version, add lots of docs that spinning needs to happen elsewhere (current case for rclpy)
+  - reentrant spinning
+- (william) you either need async/await from language or ".then" syntax (we have this in async_send_request())
+- (chris) more error checking for recursive spinning
+- (chris) weird that rclcpp and rclpy have different API
+- (shane) thinks it is ok to have different API, but rclpy is not ideal
 
 ## Parameters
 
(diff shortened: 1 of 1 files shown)


rclcpp/doc/api_review_march_2020.md
```diff
@@ -0,0 +1,433 @@
...
+- https://github.com/ros2/rclcpp/issues/360
+- https://github.com/ros2/rclcpp/blob/96ebf59a6045a535730d98fff25e522807c7aa75/rclcpp/src/rclcpp/parameter_client.cpp#L453-L468
+
+**Suggested Action**: Update issue, decide if it can be taken for Foxy or not.
+
+**Notes from 2020-03-23**:
+
+- (tfoote) Seems like adding a timeout is a good idea.
+
+### Name of AsyncParametersClient inconsistent
```
wjwwood: @ros2/team please comment on this API review point asynchronously

rclcpp/doc/api_review_march_2020.md
```diff
@@ -0,0 +1,433 @@
...
+
+- https://github.com/ros2/rclcpp/blob/7c1721a0b390be8242a6b824489d0bc861f6a0ad/rclcpp/include/rclcpp/parameter_client.hpp#L44
+
+**Suggested Action**: Reconcile class and file name, switch to singular name?
+
+**Notes from on-line, post 2020-03-23 meeting**:
+
+- 
+
+### `SyncParametersClient::get_parameters` doesn't allow you to detect error cases
```
wjwwood: @ros2/team please comment on this API review point asynchronously

rclcpp/doc/api_review_march_2020.md
```diff
@@ -0,0 +1,433 @@
...
+
+**Suggested Action**: ?
+
+**Notes from on-line, post 2020-03-23 meeting**:
+
+- 
+
+## Clock
+
+### Clock Jump callbacks on System or Steady time?
```
wjwwood: @ros2/team please comment on this API review point asynchronously

rclcpp/doc/api_review_march_2020.md
```diff
@@ -0,0 +1,433 @@
...
+
+- (tfoote) Seems like adding a timeout is a good idea.
+
+### Name of AsyncParametersClient inconsistent
+
+> AsyncParameter**s**Client uses plural, when filename is singular (and ParameterService is singular):
+
+- https://github.com/ros2/rclcpp/blob/7c1721a0b390be8242a6b824489d0bc861f6a0ad/rclcpp/include/rclcpp/parameter_client.hpp#L44
+
+**Suggested Action**: Reconcile class and file name, switch to singular name?
```
tfoote: +1 for homogenizing to singular

rclcpp/doc/api_review_march_2020.md
```diff
@@ -0,0 +1,433 @@
...
+
+## Clock
+
+### Clock Jump callbacks on System or Steady time?
+
+> Currently time jump callbacks are registered via Clock::create_jump_handler(). Jump handlers are only invoked by TimeSource::set_clock(). This is only called if the clock type is RCL_ROS_TIME and ROS time is active.
+
+- https://github.com/ros2/rclcpp/issues/528
+
+**Suggested Action**: ?
```
tfoote: There should be no jumps in steady time. If there's a big change in system time, it doesn't necessarily mean that time jumped, just that you might have been sleeping for a long time. Most ntp systems adjust the slew rate these days instead of jumping but still that's an external process and I don't know of any APIs to introspect the state of the clock.

I'm not sure that we have a way to detect jumps in time for system or steady time. To that end I think that we should be clear that we only provide callbacks when simulation time starts or stops, or simulation time jumps.

We should also strongly recommend that operators not actively adjust their system clocks while running ROS nodes.

rclcpp/doc/api_review_march_2020.md
```diff
@@ -0,0 +1,433 @@
...
+- 
+
+### `SyncParametersClient::get_parameters` doesn't allow you to detect error cases
+
+> E.g. https://github.com/ros2/rclcpp/blob/249b7d80d8f677edcda05052f598de84f4c2181c/rclcpp/src/rclcpp/parameter_client.cpp#L246-L257 return an empty vector if something goes wrong which is also a valid response.
+
+- https://github.com/ros2/rclcpp/issues/200
+- https://github.com/ros2/rclcpp/blob/96ebf59a6045a535730d98fff25e522807c7aa75/rclcpp/src/rclcpp/parameter_client.cpp#L412-L426
+
+**Suggested Action**: ?
```
tfoote: An empty list is not a valid response unless you passed in an empty list. The return should have the same length as the request in the same order. Any parameters that are not set should return a `ParameterVariant` with type [`PARAMETER_NOT_SET`. ](https://github.com/ros2/rclcpp/blob/249b7d80d8f677edcda05052f598de84f4c2181c/rclcpp/include/rclcpp/parameter.hpp#L36) to indicate that it was polled and determined to not be set.

Suggested action improve documentation of the API to clarify a short or incomplete

jacobperron: I think throwing an exception is also a valid action, making it clear that an error occurred.

rclcpp/doc/api_review_march_2020.md
jacobperron: I agree with Tully, if we don't have a way to detect system time jumps then I think we should just document that this only works with ROS time. In addition to documentation, we could log an info or warning message if the user registers jump callback with steady or system time, but it may be unnecessarily noisy.

hidmic:
One thing to add to the API review. Availability of node interfaces' getter functions is somewhat limited as of today i.e. many interface types are lacking a corresponding getter function. No getter function yields ownership either, which may result in client code not using them if access to the underlying shared pointer is necessary.

I've opened https://github.com/ros2/rclcpp/pull/1035 to address these issues through duplication.

rclcpp/doc/api_review_march_2020.md
```diff
@@ -0,0 +1,433 @@
...
+> The testing goal would still be met, either by keeping the "node_interface" classes as-is or by mocking the classes that node would multiple inherit from, however it's harder to indicate that a function needs a class that multiple inherits from several classes but not others.
+> Also having interdependency between the classes which are inherited from is a bit complicated in this design pattern.
+>
+> For (b), we would be changing how we recommend all code be written (not a trivial thing to do at all), because example code like `auto pub = node->create_publsiher(...);` would be come `auto pub = create_publisher(node, ...);`.
+> This has some distinct advantages, however, in that it allows us to write these functions, like `create_publisher(node, ...)`, so that the node argument can be any class that meets the criteria of the function.
+> That not only means that when we add a feature it automatically works with `Node` and `LifecycleNode` without adding anything to them, it also means that use defined `Node`-like classes will also work, even if they do not inherit from or provide the complete interface for `rclcpp::Node`.
```
Rayman, line 30:
```suggestion
> That not only means that when we add a feature it automatically works with `Node` and `LifecycleNode` without adding anything to them, it also means that user defined `Node`-like classes will also work, even if they do not inherit from or provide the complete interface for `rclcpp::Node`.
```

wjwwood:
> I've opened #1035 to address these issues through duplication (almost literally).

Since this was closed, superseded, then merged already, I'll omit this from the API review for now.

PR #1033, opened 2020-03-24
Add start-parameter-event-subcriber option

mauropasse:
Signed-off-by: Mauro <mpasserino@irobot.com>

@alsora

### DIFF
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -137,7 +137,8 @@ Node::Node(
       node_services_,
       node_logging_,
       node_clock_,
-      node_parameters_
+      node_parameters_,
+      options.start_parameter_event_subscriber()
     )),
   node_waitables_(new rclcpp::node_interfaces::NodeWaitables(node_base_.get())),
   node_options_(options),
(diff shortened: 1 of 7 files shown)


mauropasse:
Hello! Any updates on this?

ivanpauno:
@wjwwood @hidmic for more feedback.

@mauropasse I think this will have to wait post-Foxy release.

hidmic:
> If we would have a "on parameter changed" callback, that could be used instead of a subscription (and it will be much cheaper).

My first reaction is to agree with @ivanpauno. That'd boost performance without introducing subtle issues with overall functionality. 

I can understand that on resource constrained platforms one may want to drop unnecessary features (unnecessary for a given application), but I'm not convinced that's something you'd do in runtime.

fujitatomoya:
@mauropasse 

> Benchmarks on RPi1 (single core running at 700Mhz) shows that CPU is reduced about 2% when this parameter event subscriber is not created.

just out of curiosity, only not creating subscription for parameter events lead this? I think cpu consumption for parameter event subscription depends on parameter activity.

mauropasse:
@fujitatomoya I didn't do too much CPU profiling to get the exact reason, but my guess is that when IPC is used the subscription is implemented as a waitable, and the current StaticSingleThreadExecutor checks it's predicate at every iteration (IsReady?) 
Originally I did this to have a clean debug about the general system, setting ros params off, the subscription was still on and being checked.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -137,7 +137,8 @@ Node::Node(
       node_services_,
       node_logging_,
       node_clock_,
-      node_parameters_
+      node_parameters_,
+      options.start_parameter_event_subscriber()
```
fujitatomoya, line 141: So this option is actually only for time_source? I was expecting that this works for all user operations as long as related to node.

ivanpauno:
> If use_sim_time is modified locally, it will be ignored.

Maybe https://github.com/ros2/rclcpp/pull/1947 solves this issue, in which case I think adding this option will be fine.

PR #1049, opened 2020-04-03
Add a QoS profile equality check that ignores all fields that don't affect compatibility

emersonknapp:
Some of the values in our QoS profiles only affect local behavior, or choice of topic names. These don't affect whether a publisher and subscription are compatible. Add a utility check on `rclcpp::QoS` that compares only the compatibility-affecting QoS policies for equality 

Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>

### DIFF
--- a/rclcpp/src/rclcpp/qos.cpp
+++ b/rclcpp/src/rclcpp/qos.cpp
@@ -184,6 +184,21 @@ QoS::avoid_ros_namespace_conventions(bool avoid_ros_namespace_conventions)
   return *this;
 }
 
+bool QoS::compatibility_policies_equal(const QoS & other) const
+{
+  // Note that these policies do not affect QoS compatibility
+  // * History + Depth (relevant only for local behavior)
+  // * Lifespan (only relevant locally on Publishers)
+  // * avoid ros namespace conventions (this affects choice of topic names, not quality of service)
+  const auto & pl = get_rmw_qos_profile();
+  const auto & pr = other.get_rmw_qos_profile();
+  return pl.reliability == pr.reliability &&
+         pl.durability == pr.durability &&
+         pl.deadline == pr.deadline &&
+         pl.liveliness == pr.liveliness &&
+         pl.liveliness_lease_duration == pr.liveliness_lease_duration;
+}
+
 namespace
 {
 /// Check if two rmw_time_t have the same values.
--- a/rclcpp/include/rclcpp/qos.hpp
+++ b/rclcpp/include/rclcpp/qos.hpp
@@ -147,6 +147,13 @@ public:
   QoS &
   avoid_ros_namespace_conventions(bool avoid_ros_namespace_conventions);
 
+  /// Determine if all policies that affect QoS compatibility are equal.
+  /*
+    Note that this method does not check if the QoS policies are compatible.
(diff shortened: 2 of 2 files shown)


emersonknapp:
@Karsten1987 ping for review (I can't request reviewers on rclcpp)

emersonknapp:
@wjwwood what do you think about this? should i move the logic into `rcl` or `rmw` instead? I'd like to define this centrally to have a consistent definition. I need it in rosbag2 in order to subscribe intelligently, but I only want to define that logic there as a last resort because it would be at risk of becoming out of sync

rclcpp/src/rclcpp/qos.cpp
```diff
@@ -193,6 +193,21 @@ bool operator==(const rmw_time_t & left, const rmw_time_t & right)
...
+bool QoS::compatibility_policies_equal(const QoS & other) const
+{
+  // Note that these policies do not affect QoS compatibility
+  // * History + Depth (relevant only for local behavior)
+  // * Lifespan (only relevant locally on Publishers)
+  // * avoid ros namespace conventions (this affects choice of topic names, not quality of service)
+  const auto & pl = get_rmw_qos_profile();
+  const auto & pr = other.get_rmw_qos_profile();
+  return pl.reliability == pr.reliability &&
+         pl.durability == pr.durability &&
```
wjwwood, line 205: It's not that simple is it?

For instance, see:

https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/#qos-compatibilities

For instance with durability, if a publisher is `volatile` and a subscribing node asks for `transient local`, they will not communicate, but if the publisher is `transient local` and the subscribing node asks for `volatile`, then they will talk, just as `volatile` instead.

There's a similar asymmetry with reliability.

It seems like this check is insufficient atm, and further more you need to know in what context the QoS will be used, i.e. for either a publisher or subscription.

emersonknapp: In the comment about this function I address exactly this concern.

This is _not_ a compatibility check. This is a check that all fields that _affect_ compatibility are _exactly equal_

emersonknapp: In a sense, it is checking a very limited (but bidirectional) subset of compatibility.

emersonknapp: In rosbag2, we are currently giving up if there are publishers offering different qos policies, we don't have time to make it smarter than that for Foxy. Most ROS systems will have a limited number of publishers on any given topic, and it's extremely likely that all of them will be offering the same QoS policies (at least within the fields that affect compatibility).

wjwwood: Well, I guess I'm just dubious of the value of having such a check in the rclcpp API, rather than in rosbag2, where those assumptions hold...

Karsten1987: We've been talking about this and essentially thought it might be well placed here for maintenance purposes. Meaning, if at some point the QoS have different attributes to compare, we'd only have to modify it in one central place. 
But I am also okay with having this check in rosbag2 if it's really only that tool which would profit from this function.

PR #1050, opened 2020-04-07
Topic Statistics: Add SubscriptionTopicStatistics class

dabonnie:
This PR is part of [aws-roadmap/#225](https://github.com/ros-tooling/aws-roadmap/issues/225) to implement Topic Statistics in ROS2. This PR adds the relevant class and dependencies in order to calculate and publish Topic Statistics for a node's subscription. 

Currently a work in progress while I'm adding unit tests.

Signed-off-by: Devin Bonnie <dbbonnie@amazon.com>

### DIFF
--- a/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
@@ -41,7 +41,7 @@ namespace
 /**
  * For now, use hard coded time instead of a node's clock (to support sim time and playback)
  * due to node clock lifecycle issues.
- * @return the current nanoseconds (count) since epoch
+ * \return the current nanoseconds (count) since epoch
  */
 int64_t GetCurrentNanosecondsSinceEpoch()
 {
@@ -66,7 +66,7 @@ using libstatistics_collector::moving_average_statistics::StatisticData;
  * Class used to collect, measure, and publish topic statistics data. Current statistics
  * supported for subscribers are received message age and received message period.
  *
- * @tparam CallbackMessageT the subscribed message type
+ * \tparam CallbackMessageT the subscribed message type
  */
 template<typename CallbackMessageT>
 class SubscriberTopicStatistics
@@ -84,10 +84,10 @@ public:
   * This object wraps utilities, defined in libstatistics_collector, to collect,
   * measure, and publish topic statistics data.
    *
-  * @param node the node creating the subscription, used to create the publisher and timer to
+  * \param node the node creating the subscription, used to create the publisher and timer to
   * publish topic statistics.
-  * @param publishing_topic the topic to publish statistics to
-  * @param publishing_period the period at which topic statistics messages are published
+  * \param publishing_topic the topic to publish statistics to
+  * \param publishing_period the period at which topic statistics messages are published
   */
   SubscriberTopicStatistics(
     rclcpp::Node & node,
@@ -118,7 +118,7 @@ public:
 
   /// Set the timer used to publish statistics messages.
   /**
-   * @param measurement_timer the timer to fire the publisher, created by the node
+   * \param measurement_timer the timer to fire the publisher, created by the node
    */
   void SetPublisherTimer(const rclcpp::TimerBase::SharedPtr & publisher_timer)
   {
@@ -127,8 +127,8 @@ public:
 
   /// Handle a message received by the subscription to collect statistics.
   /**
-   * @param received_message the message received by the subscription
-   * @param now_nanoseconds current time in nanoseconds
+   * \param received_message the message received by the subscription
+   * \param now_nanoseconds current time in nanoseconds
    */
   virtual void OnMessageReceived(
     const CallbackMessageT & received_message,
@@ -143,7 +143,7 @@ public:
 
   /// Return a vector of all the currently collected data
   /**
-   * @return a vector of all the collected data
+   * \return a vector of all the collected data
    */
   std::vector<StatisticData> GetCurrentCollectorData() const
   {
--- a/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
@@ -32,6 +32,20 @@
 #include "libstatistics_collector/topic_statistics_collector/received_message_period.hpp"
 #include "libstatistics_collector/moving_average_statistics/types.hpp"
 
+namespace
+{
+/// Return the current nanoseconds (count) since epoch
+/**
+ * For now, use hard coded time instead of a node's clock (to support sim time and playback)
+ * due to node clock lifecycle issues.
+ * @return the current nanoseconds (count) since epoch
+ */
+int64_t GetCurrentNanosecondsSinceEpoch()
+{
+  const auto now = std::chrono::system_clock::now();
+  return std::chrono::duration_cast<std::chrono::nanoseconds>(now.time_since_epoch()).count();
+}
+}  // namespace
 
 namespace rclcpp
 {
@@ -51,10 +65,10 @@ using libstatistics_collector::moving_average_statistics::StatisticData;
 template<typename CallbackMessageT>
 class SubscriberTopicStatistics
 {
-  using topic_stats_collector =
+  using TopicStatsCollector =
     libstatistics_collector::topic_statistics_collector::TopicStatisticsCollector<
     CallbackMessageT>;
-  using received_message_period =
+  using ReceivedMessagePeriod =
     libstatistics_collector::topic_statistics_collector::ReceivedMessagePeriodCollector<
     CallbackMessageT>;
 
@@ -76,9 +90,9 @@ public:
   {
     // TODO(dbbonnie): ros-tooling/aws-roadmap/issues/226, received message age
 
-    auto rmp = std::make_unique<received_message_period>();
-    rmp->Start();
-    subscriber_statistics_collectors_.emplace_back(std::move(rmp));
+    auto received_message_period = std::make_unique<ReceivedMessagePeriod>();
+    received_message_period->Start();
+    subscriber_statistics_collectors_.emplace_back(std::move(received_message_period));
 
     window_start_ = rclcpp::Time(GetCurrentNanosecondsSinceEpoch());
   }
@@ -86,24 +100,6 @@ public:
   virtual ~SubscriberTopicStatistics()
   {
     TearDown();
-  }
-
-  /// Stop all collectors, clear measurements, stop publishing timer, and reset publisher.
-  virtual void TearDown()
-  {
-    for (auto & collector : subscriber_statistics_collectors_) {
-      collector->Stop();
-    }
-
-    subscriber_statistics_collectors_.clear();
-
-    if (publisher_timer_) {
-      publisher_timer_->cancel();
-      publisher_timer_.reset();
-    }
-    if (publisher_) {
-      publisher_.reset();
-    }
   }
 
   /// Set the timer used to publish statistics messages.
@@ -126,7 +122,7 @@ public:
   {
     (void) received_message;
 
-    for (auto & collector : subscriber_statistics_collectors_) {
+    for (const auto & collector : subscriber_statistics_collectors_) {
       collector->OnMessageReceived(received_message, now_nanoseconds);
     }
   }
@@ -145,6 +141,23 @@ public:
   }
 
 private:
+  /// Stop all collectors, clear measurements, stop publishing timer, and reset publisher.
+  virtual void TearDown()
+  {
+    for (auto & collector : subscriber_statistics_collectors_) {
+      collector->Stop();
+    }
+
+    subscriber_statistics_collectors_.clear();
+
+    if (publisher_timer_) {
+      publisher_timer_->cancel();
+      publisher_timer_.reset();
+    }
+
+    publisher_.reset();
+  }
+
   /// Publish a populated MetricsStatisticsMessage
   virtual void PublishMessage()
   {
@@ -165,20 +178,8 @@ private:
     window_start_ = window_end;
   }
 
-  /// Return the current nanoseconds (count) since epoch
-  /**
-   * Based on design discussions, using harccoded time instead of a node's clock
-   * due to lifecycle issues.
-   * @return the current nanoseconds (count) since epoch
-   */
-  int64_t GetCurrentNanosecondsSinceEpoch()
-  {
-    const auto now = std::chrono::system_clock::now();
-    return std::chrono::duration_cast<std::chrono::nanoseconds>(now.time_since_epoch()).count();
-  }
-
   /// Collection of statistics collectors
-  std::vector<std::unique_ptr<topic_stats_collector>> subscriber_statistics_collectors_{};
+  std::vector<std::unique_ptr<TopicStatsCollector>> subscriber_statistics_collectors_{};
   /// Node name used to generate topic statistics messages to be published
   const std::string node_name_;
   /// Publisher, created by the node, used to publish topic statistics messages
--- a/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
@@ -20,20 +20,19 @@
 #include <utility>
 #include <vector>
 
-
-#include "rclcpp/create_publisher.hpp"
-#include "rcl/time.h"
-#include "rclcpp/node.hpp"
-#include "rclcpp/time.hpp"
-#include "rclcpp/publisher.hpp"
-#include "rclcpp/timer.hpp"
+#include "libstatistics_collector/collector/generate_statistics_message.hpp"
+#include "libstatistics_collector/moving_average_statistics/types.hpp"
+#include "libstatistics_collector/topic_statistics_collector/constants.hpp"
+#include "libstatistics_collector/topic_statistics_collector/received_message_period.hpp"
 
 #include "metrics_statistics_msgs/msg/metrics_message.hpp"
 
-#include "libstatistics_collector/collector/generate_statistics_message.hpp"
-#include "libstatistics_collector/topic_statistics_collector/constants.hpp"
-#include "libstatistics_collector/topic_statistics_collector/received_message_period.hpp"
-#include "libstatistics_collector/moving_average_statistics/types.hpp"
+#include "rcl/time.h"
+#include "rclcpp/create_publisher.hpp"
+#include "rclcpp/node.hpp"
+#include "rclcpp/publisher.hpp"
+#include "rclcpp/time.hpp"
+#include "rclcpp/timer.hpp"
 
 namespace
 {
--- a/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
@@ -116,15 +116,6 @@ public:
     TearDown();
   }
 
-  /// Set the timer used to publish statistics messages.
-  /**
-   * \param measurement_timer the timer to fire the publisher, created by the node
-   */
-  void SetPublisherTimer(const rclcpp::TimerBase::SharedPtr & publisher_timer)
-  {
-    publisher_timer_ = std::move(publisher_timer);
-  }
-
   /// Handle a message received by the subscription to collect statistics.
   /**
    * \param received_message the message received by the subscription
--- a/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
@@ -20,7 +20,10 @@
 #include <utility>
 #include <vector>
 
+
+#include "rclcpp/create_publisher.hpp"
 #include "rcl/time.h"
+#include "rclcpp/node.hpp"
 #include "rclcpp/time.hpp"
 #include "rclcpp/publisher.hpp"
 #include "rclcpp/timer.hpp"
@@ -34,7 +37,7 @@
 
 namespace
 {
-/// Return the current nanoseconds (count) since epoch
+/// Return the current nanoseconds (count) since epoch.
 /**
  * For now, use hard coded time instead of a node's clock (to support sim time and playback)
  * due to node clock lifecycle issues.
@@ -51,6 +54,9 @@ namespace rclcpp
 {
 namespace topic_statistics
 {
+
+constexpr const char kDefaultPublishTopicName[]{"topic_statistics"};
+constexpr const std::chrono::milliseconds kDefaultPublishingPeriod{std::chrono::minutes(1)};
 
 using libstatistics_collector::collector::GenerateStatisticMessage;
 using metrics_statistics_msgs::msg::MetricsMessage;
@@ -75,26 +81,34 @@ class SubscriberTopicStatistics
 public:
   /// Construct a SubcriberTopicStatistics object.
   /**
-   * This object wraps utilities, defined in libstatistics_collector, to collect,
-   * measure, and publish topic statistics data.
+  * This object wraps utilities, defined in libstatistics_collector, to collect,
+  * measure, and publish topic statistics data.
    *
-   * @param node_name the name of the node, which created this instance, in order to denote
-   * topic source
-   * @param publisher instance constructed by the node in order to publish statistics data
-   */
+  * @param node the node creating the subscription, used to create the publisher and timer to
+  * publish topic statistics.
+  * @param publishing_topic the topic to publish statistics to
+  * @param publishing_period the period at which topic statistics messages are published
+  */
   SubscriberTopicStatistics(
-    const std::string & node_name,
-    const rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr & publisher)
-  : node_name_(node_name),
-    publisher_(std::move(publisher))
+    rclcpp::Node & node,
+    const std::string & publishing_topic = kDefaultPublishTopicName,
+    const std::chrono::milliseconds & publishing_period = kDefaultPublishingPeriod)
   {
-    // TODO(dbbonnie): ros-tooling/aws-roadmap/issues/226, received message age
+    publisher_ =
+      node.create_publisher<metrics_statistics_msgs::msg::MetricsMessage>(
+      publishing_topic,
+      10);
 
-    auto received_message_period = std::make_unique<ReceivedMessagePeriod>();
-    received_message_period->Start();
-    subscriber_statistics_collectors_.emplace_back(std::move(received_message_period));
+    auto callback = [this]()
+      {
+        this->PublishMessage();
+      };
 
-    window_start_ = rclcpp::Time(GetCurrentNanosecondsSinceEpoch());
+    publisher_timer_ = node.create_wall_timer(publishing_period, callback);
+
+    node_name_ = node.get_name();
+
+    BringUp();
   }
 
   virtual ~SubscriberTopicStatistics()
@@ -141,8 +155,18 @@ public:
   }
 
 private:
+  /// Construct and start all collectors and set window_start_.
+  void BringUp()
+  {
+    auto received_message_period = std::make_unique<ReceivedMessagePeriod>();
+    received_message_period->Start();
+    subscriber_statistics_collectors_.emplace_back(std::move(received_message_period));
+
+    window_start_ = rclcpp::Time(GetCurrentNanosecondsSinceEpoch());
+  }
+
   /// Stop all collectors, clear measurements, stop publishing timer, and reset publisher.
-  virtual void TearDown()
+  void TearDown()
   {
     for (auto & collector : subscriber_statistics_collectors_) {
       collector->Stop();
@@ -181,11 +205,11 @@ private:
   /// Collection of statistics collectors
   std::vector<std::unique_ptr<TopicStatsCollector>> subscriber_statistics_collectors_{};
   /// Node name used to generate topic statistics messages to be published
-  const std::string node_name_;
+  std::string node_name_;
   /// Publisher, created by the node, used to publish topic statistics messages
-  rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr publisher_{nullptr};
+  rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr publisher_;
   /// Timer which fires the publisher
-  rclcpp::TimerBase::SharedPtr publisher_timer_{nullptr};
+  rclcpp::TimerBase::SharedPtr publisher_timer_;
   /// The start of the collection window, used in the published topic statistics message
   rclcpp::Time window_start_;
 };
--- a/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef RCLCPP__SUBSCRIBER_TOPIC_STATISTICS_HPP_
-#define RCLCPP__SUBSCRIBER_TOPIC_STATISTICS_HPP_
+#ifndef RCLCPP__TOPIC_STATISTICS__SUBSCRIBER_TOPIC_STATISTICS_HPP_
+#define RCLCPP__TOPIC_STATISTICS__SUBSCRIBER_TOPIC_STATISTICS_HPP_
 
-#include <vector>
+#include <memory>
 #include <string>
+#include <utility>
+#include <vector>
 
 #include "rcl/time.h"
 #include "rclcpp/time.hpp"
@@ -25,18 +27,20 @@
 
 #include "metrics_statistics_msgs/msg/metrics_message.hpp"
 
+#include "libstatistics_collector/collector/generate_statistics_message.hpp"
 #include "libstatistics_collector/topic_statistics_collector/constants.hpp"
-#include "libstatistics_collector/topic_statistics_collector/received_message_age.hpp"
 #include "libstatistics_collector/topic_statistics_collector/received_message_period.hpp"
+#include "libstatistics_collector/moving_average_statistics/types.hpp"
 
 
-namespace rclcpp {
-namespace topic_statistics {
+namespace rclcpp
+{
+namespace topic_statistics
+{
 
 using libstatistics_collector::collector::GenerateStatisticMessage;
 using metrics_statistics_msgs::msg::MetricsMessage;
-using metrics_statistics_msgs::msg::StatisticDataPoint;
-using metrics_statistics_msgs::msg::StatisticDataType;
+using libstatistics_collector::moving_average_statistics::StatisticData;
 
 /**
  * Class used to collect, measure, and publish topic statistics data. Current statistics
@@ -45,14 +49,14 @@ using metrics_statistics_msgs::msg::StatisticDataType;
  * @tparam CallbackMessageT the subscribed message type
  */
 template<typename CallbackMessageT>
-class SubscriberTopicStatistics {
-
-using topic_stats_collector =
-  topic_statistics_collector::TopicStatisticsCollector<CallbackMessageT>;
-using received_message_age =
-  topic_statistics_collector::ReceivedMessageAgeCollector<CallbackMessageT>;
-using received_message_period =
-  topic_statistics_collector::ReceivedMessagePeriodCollector<CallbackMessageT>;
+class SubscriberTopicStatistics
+{
+  using topic_stats_collector =
+    libstatistics_collector::topic_statistics_collector::TopicStatisticsCollector<
+    CallbackMessageT>;
+  using received_message_period =
+    libstatistics_collector::topic_statistics_collector::ReceivedMessagePeriodCollector<
+    CallbackMessageT>;
 
 public:
   /// Construct a SubcriberTopicStatistics object.
@@ -64,16 +68,15 @@ public:
    * topic source
    * @param publisher instance constructed by the node in order to publish statistics data
    */
-  SubscriberTopicStatistics(const std::string & node_name,
+  SubscriberTopicStatistics(
+    const std::string & node_name,
     const rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr & publisher)
-    : node_name_(node_name),
+  : node_name_(node_name),
     publisher_(std::move(publisher))
   {
-    const auto rma = std::make_unique<received_message_age>();
-    rma->Start();
-    subscriber_statistics_collectors_.emplace_back(std::move(rma));
+    // TODO(dbbonnie): ros-tooling/aws-roadmap/issues/226, received message age
 
-    const auto rmp = std::make_unique<received_message_period>();
+    auto rmp = std::make_unique<received_message_period>();
     rmp->Start();
     subscriber_statistics_collectors_.emplace_back(std::move(rmp));
 
@@ -88,7 +91,7 @@ public:
   /// Stop all collectors, clear measurements, stop publishing timer, and reset publisher.
   virtual void TearDown()
   {
-    for (auto & collector: subscriber_statistics_collectors_) {
+    for (auto & collector : subscriber_statistics_collectors_) {
       collector->Stop();
     }
 
@@ -123,9 +126,22 @@ public:
   {
     (void) received_message;
 
-    for (auto & collector: subscriber_statistics_collectors_) {
+    for (auto & collector : subscriber_statistics_collectors_) {
       collector->OnMessageReceived(received_message, now_nanoseconds);
     }
+  }
+
+  /// Return a vector of all the currently collected data
+  /**
+   * @return a vector of all the collected data
+   */
+  std::vector<StatisticData> GetCurrentCollectorData() const
+  {
+    std::vector<StatisticData> data;
+    for (const auto & collector : subscriber_statistics_collectors_) {
+      data.push_back(collector->GetStatisticsResults());
+    }
+    return data;
   }
 
 private:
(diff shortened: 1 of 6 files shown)


dabonnie:
FYI @wjwwood @thomas-moulard

dabonnie:
@ros2/aws-oncall - please run this CI job
Gist: https://gist.githubusercontent.com/dabonnie/b651968aa96a5d53f3fbbbae73710255/raw/3faf954eb646facec4071b3147e7e1f62e970950/ros2.repos
BUILD args: --packages-up-to rclcpp
TEST args: --packages-select rclcpp
Job: ci_launcher

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * Class used to collect, measure, and publish topic statistics data. Current statistics
+ * supported for subscribers are received message age and received message period.
+ *
+ * @tparam CallbackMessageT the subscribed message type
+ */
+template<typename CallbackMessageT>
+class SubscriberTopicStatistics
+{
+  using topic_stats_collector =
```
thomas-moulard, line 54: `TopicStatsCollector`

https://google.github.io/styleguide/cppguide.html#Type_Names

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+    libstatistics_collector::topic_statistics_collector::TopicStatisticsCollector<
+    CallbackMessageT>;
+  using received_message_period =
```
thomas-moulard, line 57: ditto

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+   * This object wraps utilities, defined in libstatistics_collector, to collect,
+   * measure, and publish topic statistics data.
+   *
+   * @param node_name the name of the node, which created this instance, in order to denote
+   * topic source
+   * @param publisher instance constructed by the node in order to publish statistics data
+   */
+  SubscriberTopicStatistics(
+    const std::string & node_name,
+    const rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr & publisher)
```
thomas-moulard, line 73: Why do we need to share ownership of the Publisher here? Do we have cases where this would survive the Publisher lifetime? If not, a pointer is enough.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+  : node_name_(node_name),
+    publisher_(std::move(publisher))
+  {
+    // TODO(dbbonnie): ros-tooling/aws-roadmap/issues/226, received message age
+
+    auto rmp = std::make_unique<received_message_period>();
```
thomas-moulard, line 79:
```suggestion
    auto received_message_period = std::make_unique<received_message_period>();
```

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+    window_start_ = rclcpp::Time(GetCurrentNanosecondsSinceEpoch());
+  }
+
+  virtual ~SubscriberTopicStatistics()
+  {
+    TearDown();
+  }
+
+  /// Stop all collectors, clear measurements, stop publishing timer, and reset publisher.
+  virtual void TearDown()
```
thomas-moulard, line 92: Does this need to be public?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+    auto rmp = std::make_unique<received_message_period>();
+    rmp->Start();
+    subscriber_statistics_collectors_.emplace_back(std::move(rmp));
+
+    window_start_ = rclcpp::Time(GetCurrentNanosecondsSinceEpoch());
+  }
+
+  virtual ~SubscriberTopicStatistics()
+  {
+    TearDown();
```
thomas-moulard, line 88: Should the user call explicitly or not? If yes, we don't need this call in the dtor (or we should log a warning at least). If no, `TearDown` should be private.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+  /// Stop all collectors, clear measurements, stop publishing timer, and reset publisher.
+  virtual void TearDown()
+  {
+    for (auto & collector : subscriber_statistics_collectors_) {
+      collector->Stop();
+    }
+
+    subscriber_statistics_collectors_.clear();
+
+    if (publisher_timer_) {
```
thomas-moulard, line 100: Are there valid cases where it can be null?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+      publisher_timer_->cancel();
+      publisher_timer_.reset();
+    }
+    if (publisher_) {
```
thomas-moulard, line 104: Are there valid cases where it can be null?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+      publisher_.reset();
+    }
+  }
+
+  /// Set the timer used to publish statistics messages.
+  /**
+   * @param measurement_timer the timer to fire the publisher, created by the node
+   */
+  void SetPublisherTimer(const rclcpp::TimerBase::SharedPtr & publisher_timer)
```
thomas-moulard, line 113: Could we pass that in the constructor directly? Don't we always need this timer?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+    publisher_timer_ = std::move(publisher_timer);
+  }
+
+  /// Handle a message received by the subscription to collect statistics.
+  /**
+   * @param received_message the message received by the subscription
+   * @param now_nanoseconds current time in nanoseconds
+   */
+  virtual void OnMessageReceived(
+    const CallbackMessageT & received_message,
```
thomas-moulard, line 124:
```suggestion
    const CallbackMessageT &,
```
You can then remove `(void) received_message`

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+    return data;
+  }
+
+private:
+  /// Publish a populated MetricsStatisticsMessage
+  virtual void PublishMessage()
+  {
+    rclcpp::Time window_end{GetCurrentNanosecondsSinceEpoch()};
+
+    for (auto & collector : subscriber_statistics_collectors_) {
```
thomas-moulard, line 153:
```suggestion
    for (const auto & collector : subscriber_statistics_collectors_) {
```

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+    window_start_ = window_end;
+  }
+
+  /// Return the current nanoseconds (count) since epoch
+  /**
+   * Based on design discussions, using harccoded time instead of a node's clock
+   * due to lifecycle issues.
+   * @return the current nanoseconds (count) since epoch
+   */
+  int64_t GetCurrentNanosecondsSinceEpoch()
```
thomas-moulard, line 174:
```suggestion
  int64_t GetCurrentNanosecondsSinceEpoch() const
```

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+        collected_stats);
+      publisher_->publish(message);
+    }
+    window_start_ = window_end;
+  }
+
+  /// Return the current nanoseconds (count) since epoch
+  /**
+   * Based on design discussions, using harccoded time instead of a node's clock
+   * due to lifecycle issues.
```
thomas-moulard, line 171: Which design discussions?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /// Collection of statistics collectors
+  std::vector<std::unique_ptr<topic_stats_collector>> subscriber_statistics_collectors_{};
+  /// Node name used to generate topic statistics messages to be published
+  const std::string node_name_;
+  /// Publisher, created by the node, used to publish topic statistics messages
+  rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr publisher_{nullptr};
+  /// Timer which fires the publisher
+  rclcpp::TimerBase::SharedPtr publisher_timer_{nullptr};
```
thomas-moulard, line 187: Why SharedPtr?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,194 @@
...
+    const auto now = std::chrono::system_clock::now();
+    return std::chrono::duration_cast<std::chrono::nanoseconds>(now.time_since_epoch()).count();
+  }
+
+  /// Collection of statistics collectors
+  std::vector<std::unique_ptr<topic_stats_collector>> subscriber_statistics_collectors_{};
+  /// Node name used to generate topic statistics messages to be published
+  const std::string node_name_;
+  /// Publisher, created by the node, used to publish topic statistics messages
+  rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr publisher_{nullptr};
```
thomas-moulard, line 185: Why SharedPtr?

rclcpp/test/topic_statistics/test_subscriber_topic_statistics.cpp
```diff
@@ -0,0 +1,115 @@
...
+      };
+    subscription_ = create_subscription<Empty,
+        std::function<void(Empty::UniquePtr)>>(
+      topic,
+      0 /*history_depth*/,
+      callback);
+  }
+
+private:
+  void ReceiveMessage(const Empty & msg) const
```
thomas-moulard, line 63:
```suggestion
  void ReceiveMessage(const Empty &) const
```
and remove `(void) msg`

rclcpp/test/topic_statistics/test_subscriber_topic_statistics.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Test fixture to bring up and teardown rclcpp
+ */
+class TestSubscriberTopicStatisticsFixture : public ::testing::Test
+{
+protected:
+  void SetUp()
+  {
+    rclcpp::init(0, nullptr);
```
thomas-moulard, line 79:
```suggestion
    rclcpp::init(0 /* argc */, nullptr /* argv */);
```

IIRC?

rclcpp/test/topic_statistics/test_subscriber_topic_statistics.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+  auto empty_subscriber = std::make_shared<EmptySubscriber>(
+    kTestNodeName,
+    kTestSubStatsTopic);
+
+  // manually create publisher tied to the node
+  auto topic_stats_publisher =
+    empty_subscriber->create_publisher<MetricsMessage>(
+    kTestTopicStatisticsTopic,
+    10);
```
thomas-moulard, line 98: why 10? For a test, KeepAll() may be acceptable, and (potentially?) reduce flakyness.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: The `SubscriberTopicStatistics` owns the Publisher and its lifetime, where the `shared_ptr<Publisher>` is created via the [Node API](http://docs.ros2.org/eloquent/api/rclcpp/classrclcpp_1_1Node.html#ad1dfc9d04d67ab93353e04a7df72bc9a). It was chosen to do this via the Node API as it became quite complex to do otherwise. For an example of the creation during subscription please see [this PoC](https://github.com/aws-ros-dev/rclcpp/blob/dabonnie/topic_statistics/rclcpp/include/rclcpp/node_impl.hpp#L112).

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Will make private.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: No (agree with your above comment).

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: This is a good point. I have local changes with a factory method that ensures this doesn't happen, so this check can be removed there.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Good point, as it's non-null on construction and only null on destruction.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Not quite, I was having issues with this via the [PoC](https://github.com/aws-ros-dev/rclcpp/blob/dabonnie/topic_statistics/rclcpp/include/rclcpp/node_impl.hpp#L123). Here you can see because the parent `Node` creates the timer, we need to define the lambda with the `SubscriberTopicStatistics` instance. There may be a better way to accomplish what we want here though.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Sorry, this was an email thread with @wjwwood .

dabonnie:
> Mainly LGTM. A few things to look at:
> 
> 1/ Lifecycle - do we need this class to have ownership through SharedPtr for both the publisher, and timer? Could we just use a pointer to avoid having to deal with ownership issues?

This is a great question and not at all obvious unless you can see the larger picture. IMHO the [PoC](https://github.com/aws-ros-dev/rclcpp/blob/dabonnie/topic_statistics/rclcpp/include/rclcpp/node_impl.hpp#L108) provides a good rationale of the design choices.

The intent is to create a `SubscriberTopicStatistics` instance when a Node Subscription is created, where the Subscription itself maintains single ownership of the `SubscriberTopicStatistics` instance. As part of the topic statistics functionality, the `SubscriberTopicStatistics` needs to do the following:

1. Aggregate data
2. Publish aggregated data on a topic at a periodic interval

Item 1 is simply hooking this class into `Subscription::handle_message(std::shared_ptr<void> & message, const rmw_message_info_t & message_info)` and passing the input message into the `SubscriberTopicStatistics` instance.

Item 2  demonstrates that the ownership of the publisher (where topic statistics data is sent) and the timer belong to the `SubscriberTopicStatistics` instance. Currently they are shared pointers simply because they are created via the Node API (see comments on your review remarks) as that was the simplest approach. My early attempts to try and create these resources, without the Node, became complicated and duplicative.

> 1b/ Lifecycle - let's choose our side! Either things get teardown in the destructor, and we don't expose teardown, or the opposite. Pros of dtor: exception safety! Cons: if TearDown throws, [we may crash](http://www.cs.technion.ac.il/users/yechiel/c++-faq/dtors-shouldnt-throw.html)

Since the lifecycle of the `SubscriberTopicStatistics` class will be tied to the Node's Subscription, we should teardown in the constructor since there appear to be no other places to teardown. We can be explicit and declare `noexcept` in the `libstatistics_collector` specific teardown calls, as a quick glance reveals fairly safe calls.

> 2/ Visibility - this class has a lot of methods as `public`, could you double check to see if we could make some of those methods private? `OnMessageReceived`? `TearDown`?

Agree and commented on your review remarks.

> 3/ Simplicity - could we remove `SetPublishTimer`?
> 

I don't believe so (but maybe possible?). See my comment here: https://github.com/ros2/rclcpp/pull/1050/files#r405218682

> 4/ Consistency - `TearDown` is not consistent with the constructor. It checks if a publisher is set, but actually it should never be the case. Let's remove those if-s for cases which can't happen, and clearly decide (and document/comment) on what's optional and what's not.
> 

Agreed

> 5/ Simplicity - `GetCurrentNanosecondsSinceEpoch` Could we avoid putting this in a header? This does not depend on the class structure. The relationship between this and the `publish_timer_` is not clear in the PR.
> 

Can we leave as an anonymous namespace function? I did this to attempt to avoid a [long method](https://refactoring.guru/smells/long-method) and encapsulate the time source. This way, if and when Node Time (sim, playback, etc) is a desirable feature, the time source can be easily updated.

> Generally, the code structure is OK, but I'd like all those implicit design decisions to be explained. It's not necessarily a lot of work, but a few lines of comments here and there could go a long way. I have other nits which are pretty minor in the diff.

I'll address those, thanks!

prajakta-gokhale:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9996/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5597/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8132/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9879/)

rclcpp/test/topic_statistics/test_subscriber_topic_statistics.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+#include "test_msgs/msg/empty.hpp"
+
+
+namespace
+{
+constexpr const char kTestNodeName[]{"test_sub_stats_node"};
+constexpr const char kTestSubStatsTopic[]{"/test_sub_stats_topic"};
+constexpr const char kTestTopicStatisticsTopic[]{"/test_topic_statistics_topic"};
+constexpr const uint16_t kNoSamples = 0;
```
zmichaels11, line 36:
```suggestion
constexpr const uint16_t kNoSamples{0};
```
Any reason why this is `uint16_t`?

dabonnie: As discussed this will be the same type as `data.sample_count`

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10097/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5677/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8210/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9996/)

Targets: Focal

piraka9011:
Round 2: New Gist
https://gist.githubusercontent.com/dabonnie/25d7282e24a689d91ae1c0e7ca4764b5/raw/a24a0c844384371db99ef9eb69b14f1ce1b388a1/ros2.repos

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10103/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5681/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8214/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10001/)

ros-discourse:

https://discourse.ros.org/t/ros2-tooling-wg-next-meeting/12545/17

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+namespace
+{
+/// Return the current nanoseconds (count) since epoch
+/**
+ * For now, use hard coded time instead of a node's clock (to support sim time and playback)
+ * due to node clock lifecycle issues.
+ * @return the current nanoseconds (count) since epoch
+ */
+int64_t GetCurrentNanosecondsSinceEpoch()
```
wjwwood, line 43: This style doesn't follow the rest of rclcpp, it should be `get_current_nanoseconds_since_epoch()`. There are lots of instances of this in the new code. I believe it should be changed to be consistent with rclcpp.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,195 @@
...
+    TearDown();
+  }
+
+  /// Set the timer used to publish statistics messages.
+  /**
+   * @param measurement_timer the timer to fire the publisher, created by the node
+   */
+  void SetPublisherTimer(const rclcpp::TimerBase::SharedPtr & publisher_timer)
+  {
+    publisher_timer_ = std::move(publisher_timer);
```
wjwwood, line 111: This is not good right? moving a const reference?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
wjwwood: This doesn't look like it was ever done.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
wjwwood: https://github.com/ros2/rclcpp/blob/0f0b83368a9650b76152f017c21d51f4366245fd/rclcpp/include/rclcpp/publisher_base.hpp#L55 It uses shared from this, and I assume this class owns it? If not I guess it could be a `weak_ptr` instead.

dabonnie: The `SubscriberTopicsStatistics` class owns the publisher. Please see the new constructor, that should hopefully clarify this, [where the publisher is created](https://github.com/ros2/rclcpp/pull/1050/files#diff-756335e5efe4ced64021eaf3abf4df29R97).

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: This is now resolved given the new constructor.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: This is now resolved given the new constructor.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -32,6 +32,20 @@
 #include "libstatistics_collector/topic_statistics_collector/received_message_period.hpp"
 #include "libstatistics_collector/moving_average_statistics/types.hpp"
 
+namespace
+{
+/// Return the current nanoseconds (count) since epoch
+/**
+ * For now, use hard coded time instead of a node's clock (to support sim time and playback)
+ * due to node clock lifecycle issues.
+ * @return the current nanoseconds (count) since epoch
```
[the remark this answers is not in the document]
dabonnie: Will do,

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,195 @@
...
+   *
+   * @param node_name the name of the node, which created this instance, in order to denote
+   * topic source
+   * @param publisher instance constructed by the node in order to publish statistics data
+   */
+  SubscriberTopicStatistics(
+    const std::string & node_name,
+    const rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr & publisher)
+  : node_name_(node_name),
+    publisher_(std::move(publisher))
```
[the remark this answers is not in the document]
dabonnie: Good question. This should now (hopefully) be resolved given the new constructor, where the publisher is created by an input node.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Will do. The style you're seeing is from the [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Function_Names); I'll make the necessary changes to be consistent with the `rclcpp` style.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: This is now outdated given that the timer is now created in the new constructor.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
wjwwood: When contributing to a core ROS 2 package, refer to the ROS 2 style guide:

https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#function-and-method-naming

wjwwood:
Sorry, looks like there's a merge conflict on master now, it was probably my pr that just got merged.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+#ifndef RCLCPP__TOPIC_STATISTICS__SUBSCRIBER_TOPIC_STATISTICS_HPP_
+#define RCLCPP__TOPIC_STATISTICS__SUBSCRIBER_TOPIC_STATISTICS_HPP_
+
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+
```
wjwwood, line 23: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,208 @@
...
+#include "rclcpp/time.hpp"
+#include "rclcpp/publisher.hpp"
+#include "rclcpp/timer.hpp"
+
+#include "metrics_statistics_msgs/msg/metrics_message.hpp"
+
+#include "libstatistics_collector/collector/generate_statistics_message.hpp"
+#include "libstatistics_collector/topic_statistics_collector/constants.hpp"
+#include "libstatistics_collector/topic_statistics_collector/received_message_period.hpp"
+#include "libstatistics_collector/moving_average_statistics/types.hpp"
```
wjwwood, line 36: nitpick: according to GSG the `rclcpp` headers should be included last, after these

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /// Handle a message received by the subscription to collect statistics.
+  /**
+   * \param received_message the message received by the subscription
+   * \param now_nanoseconds current time in nanoseconds
+   */
+  virtual void handle_message(
+    const CallbackMessageT & received_message,
+    const rcl_time_point_value_t now_nanoseconds) const
```
wjwwood, line 124: Since this is a public function, it might be nice to avoid the C types, and prefer something like `std::chrono::nanoseconds`.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
wjwwood: Could this not be made `const`?

rclcpp/test/topic_statistics/test_subscriber_topic_statistics.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "rclcpp/create_publisher.hpp"
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/qos.hpp"
+#include "rclcpp/node.hpp"
+
+#include "rclcpp/topic_statistics/subscriber_topic_statistics.hpp"
+
+#include "test_msgs/msg/empty.hpp"
+
+
```
wjwwood, line 33: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Would this be an issue if we were to use ros time in the future?

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10192/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5745/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8285/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10079/)

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
wjwwood: Well `rcl_time_point_value_t` is implicitly a steady clock, whether or not it comes from ROS time or the system is an implementation detail. What is that time stamp used for? It seems like you would most often be interested in statistics in terms of actual wall time, but I don't know. Maybe you'd want both. Either way you could use `rclcpp::Time` if you want to potentially support ROS Time. My point here is that you should try to use a C++ construct in a C++ interface if possible.

dabonnie: Sounds good, I've made the changes to use `rclcpp::Time`.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+  using ReceivedMessagePeriod =
+    libstatistics_collector::topic_statistics_collector::ReceivedMessagePeriodCollector<
+    CallbackMessageT>;
+
+public:
+  /// Construct a SubcriberTopicStatistics object.
+  /**
+  * This object wraps utilities, defined in libstatistics_collector, to collect,
+  * measure, and publish topic statistics data.
+   *
```
wjwwood, line 83:
```suggestion
  *
```

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,207 @@
...
+  /// Handle a message received by the subscription to collect statistics.
+  /**
+   * \param received_message the message received by the subscription
+   * \param now_nanoseconds current time in nanoseconds
+   */
+  virtual void handle_message(
+    const CallbackMessageT & received_message,
+    const rclcpp::Time now_nanoseconds) const
+  {
+    (void) received_message;
```
wjwwood, line 125: Is this redundant now?

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Done

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: You are correct: removed.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,205 @@
...
+{
+/// Return the current nanoseconds (count) since epoch.
+/**
+ * \return the current nanoseconds (count) since epoch
+ */
+int64_t get_current_nanoseconds_since_epoch()
+{
+  const auto now = std::chrono::system_clock::now();
+  return std::chrono::duration_cast<std::chrono::nanoseconds>(now.time_since_epoch()).count();
+}
```
wjwwood, line 47: Are you not worried about breaking the one-definition-rule here by including the implementation in the header?

dabonnie: That's a really good point. I will move this to be a private class function to avoid creating an impl file while also removing the anon namespace.

dabonnie:
Just an FYI: trying to figure out the Windows build failures with `libstatistics_collector` (seems to be a `CMakeLists.txt` issue) and, in parallel, separating the dependencies into their own repos (https://github.com/ros-tooling/aws-roadmap/issues/248, https://github.com/ros-tooling/aws-roadmap/issues/249) and updating the `ros2.repos` file (https://github.com/ros-tooling/aws-roadmap/issues/250).

dabonnie:
Updated the message dependency to use the `statistics_msgs` defined in `rcl_interfaces' (https://github.com/ros2/rcl_interfaces/pull/98). However, FYI @wjwwood  this PR is blocked by https://github.com/ros2/ros2/pull/897 (can verify CI after that is merged).

wjwwood:
https://github.com/ros2/ros2/pull/897 is merged now, I think this also needs CI then?

dabonnie:
> [ros2/ros2#897](https://github.com/ros2/ros2/pull/897) is merged now, I think this also needs CI then?

@wjwwood sadly not quite: I need you to re-review the changes from my latest commit, where the full contest can be seen in this draft PR ( https://github.com/ros2/rclcpp/pull/1072 - note that this is working code and publishes topic statistics for the `demo_nodes_cpp` talker and listener).

> Why these changes? 

Any attempt to create a publisher or timer at the level of `create_subscription` or `subscription_factory` resulted in numerous errors from simply including the following headers:

```
#include "rclcpp/timer.hpp"
#include "rclcpp/create_publisher.hpp"
```

and in the case of our convenience constructor it included 

```
#include "rclcpp/node.hpp"
```

I'm currently unable to find a better approach that what I've proposed in https://github.com/ros2/rclcpp/pull/1072. 

> Why wasn't this caught sooner?

It was my mistake not ensuring the proposed review changes actually worked offline in an integration branch. However, thankfully I was able to catch these issues today and provide the full picture for review.

dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10298/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5832/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8378/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10195/)

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+#include "statistics_msgs/msg/metrics_message.hpp"
+
+namespace rclcpp
+{
+namespace topic_statistics
+{
+
+constexpr const char kDefaultPublishTopicName[]{"topic_statistics"};
+constexpr const std::chrono::milliseconds kDefaultPublishingPeriod{std::chrono::minutes(1)};
```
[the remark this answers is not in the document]
dabonnie: Will update

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,198 @@
...
+    for (const auto & collector : subscriber_statistics_collectors_) {
+      collector->OnMessageReceived(received_message, now_nanoseconds.nanoseconds());
+    }
+  }
+
+  /// Return a vector of all the currently collected data
+  /**
+   * \return a vector of all the collected data
+   */
+  std::vector<StatisticData> get_current_collector_data() const
```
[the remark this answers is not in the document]
dabonnie: Sure

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,198 @@
...
+    libstatistics_collector::topic_statistics_collector::ReceivedMessagePeriodCollector<
+    CallbackMessageT>;
+
+public:
+  /// Construct a SubscriberTopicStatistics object.
+  /**
+   * This object wraps utilities, defined in libstatistics_collector, to collect,
+   * measure, and publish topic statistics data.
+   *
+   * @param node_name the name of the node, which created this instance, in order to denote
```
wjwwood, line 69: This is still using the other doxygen style.

```suggestion
   * \param node_name the name of the node, which created this instance, in order to denote
```

Please check the rest of the file.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,198 @@
...
+   * topic source
+   * @param publisher instance constructed by the node in order to publish statistics data.
+   * This class owns the publisher/
```
wjwwood, line 72:
```suggestion
   * This class owns the publisher.
```

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,198 @@
...
+   */
+  SubscriberTopicStatistics(
```
wjwwood, line 74:
```suggestion
  SubscriptionTopicStatistics(
```

And any fallout from that rename.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,198 @@
...
+    const std::string & node_name,
+    rclcpp::Publisher<statistics_msgs::msg::MetricsMessage>::SharedPtr publisher)
+  : node_name_(node_name),
+    publisher_(std::move(publisher))
+  {
+    // TODO(dbbonnie): ros-tooling/aws-roadmap/issues/226, received message age
+    // TODO(dbbonnie) throw if publisher is null
```
wjwwood, line 81: I really don't think this is worth leaving as a TODO.

```suggestion
    if (nullptr == publisher_) {
      throw std::invalid_argument("publisher pointer is nullptr");
    }
```

And add a line to the doc block that it throws.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
```diff
@@ -0,0 +1,198 @@
...
+      data.push_back(collector->GetStatisticsResults());
+    }
+    return data;
+  }
+
+  /// Set the timer used to publish statistics messages.
+  /**
+   * @param measurement_timer the timer to fire the publisher, created by the node
+   */
+  void set_publisher_timer(const rclcpp::TimerBase::SharedPtr & publisher_timer)
```
wjwwood, line 121: Based on https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-sharedptrparam-owner, I'd say this should not take a `const &` and instead should take a copy and move it into the internal shared pointer.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Sorry, those were from the reverted changes. Will double check and fix.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Applied check and added doc.

rclcpp/include/rclcpp/topic_statistics/subscriber_topic_statistics.hpp
dabonnie: Renamed classes and relevant files.

dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10319/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5848/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8404/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10226/)

wjwwood:
The test failure on macOS is also on nightly: https://ci.ros2.org/view/nightly/job/nightly_osx_release/1622/testReport/(root)/projectroot/test_publisher_subscription_count_api/

We also discussed it elsewhere.

wjwwood:
This needs a rebase after merging the other pr.

dabonnie:
> This needs a rebase after merging the other pr.

Rebased and re-running CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10320/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5849/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8405/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10227/)

dabonnie:
> The test failure on macOS is also on nightly: https://ci.ros2.org/view/nightly/job/nightly_osx_release/1622/testReport/(root)/projectroot/test_publisher_subscription_count_api/
> 
> We also discussed it elsewhere.

@wjwwood same test failure, otherwise appears gtg?

Issue #1048, opened 2020-04-01
Add default Incompatible QoS callbacks


mm318:
#### Feature description

With https://github.com/ros2/rclcpp/pull/924 merged in, a user is able to provide a callback for when an participant on the topic is discovered but unable to be matched due to incompatible QoS policies. This feature request is for leveraging this feature even when a user does not provide a custom callback function, by setting a default callback function that prints a warning (initially suggested at https://github.com/ros2/ros2/issues/822#issuecomment-598024481) on the discovery of an incompatible participant on the topic.

#### Implementation considerations

What should the behavior be if the underlying middleware does not support this kind of QoS event callback?

mm318:
Relevant pull request opened at https://github.com/ros2/rclcpp/pull/1051.

PR #1051, opened 2020-04-07
Create a default warning for qos incompatibility

mm318:
Closes https://github.com/ros2/rclcpp/issues/1048

This pull request adds a default callback for incompatible offered/requested QoS events. In the case that the underlying middleware doesn't support such type of events, the default callback will not be registered and a warning will be emitted.

### DIFF
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -88,15 +88,17 @@ public:
       this->add_event_handler(
         options_.event_callbacks.incompatible_qos_callback,
         RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
-    } else {
+    } else if (options_.use_default_callbacks) {
       // Register default callback when not specified
       try {
         this->add_event_handler(
-          QOSOfferedIncompatibleQoSCallbackType(default_incompatible_qos_callback),
+          [this](QOSOfferedIncompatibleQoSInfo & info) {
+            this->default_incompatible_qos_callback(info);
+          },
           RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
       } catch (UnsupportedEventTypeException & /*exc*/) {
         RCLCPP_WARN_ONCE(
-          rclcpp::get_logger("rclcpp::Publisher"),
+          rclcpp::get_logger(rcl_node_get_logger_name(rcl_node_handle_.get())),
           "This rmw implementation does not support ON_OFFERED_INCOMPATIBLE_QOS "
           "events, you will not be notified when Publishers offer an incompatible "
           "QoS profile to Subscriptions on the same topic.");
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -88,6 +88,19 @@ public:
       this->add_event_handler(
         options_.event_callbacks.incompatible_qos_callback,
         RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
+    } else {
+      // Register default callback when not specified
+      try {
+        this->add_event_handler(
+          QOSOfferedIncompatibleQoSCallbackType(default_incompatible_qos_callback),
+          RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
+      } catch (UnsupportedEventTypeException & /*exc*/) {
+        RCLCPP_WARN_ONCE(
+          rclcpp::get_logger("rclcpp"),
+          "This rmw implementation does not support ON_OFFERED_INCOMPATIBLE_QOS "
+          "events, you will not be notified when Publishers offer an incompatible "
+          "QoS profile to Subscriptions on the same topic.");
+      }
     }
     // Setup continues in the post construction method, post_init_setup().
   }
--- a/rclcpp/include/rclcpp/publisher.hpp
+++ b/rclcpp/include/rclcpp/publisher.hpp
@@ -96,7 +96,7 @@ public:
           RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
       } catch (UnsupportedEventTypeException & /*exc*/) {
         RCLCPP_WARN_ONCE(
-          rclcpp::get_logger("rclcpp"),
+          rclcpp::get_logger("rclcpp::Publisher"),
           "This rmw implementation does not support ON_OFFERED_INCOMPATIBLE_QOS "
           "events, you will not be notified when Publishers offer an incompatible "
           "QoS profile to Subscriptions on the same topic.");
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -122,15 +122,17 @@ public:
       this->add_event_handler(
         options.event_callbacks.incompatible_qos_callback,
         RCL_SUBSCRIPTION_REQUESTED_INCOMPATIBLE_QOS);
-    } else {
+    } else if (options_.use_default_callbacks) {
       // Register default callback when not specified
       try {
         this->add_event_handler(
-          QOSRequestedIncompatibleQoSCallbackType(default_incompatible_qos_callback),
+          [this](QOSRequestedIncompatibleQoSInfo & info) {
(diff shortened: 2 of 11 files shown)


mm318:
Hi @ivanpauno and @wjwwood, can you take a look at this pull request? It's for addressing the comment

> I think unless the user defines their own callback it would be nice to have a default callback which logs a warning when this event occurs. It would be ok, in this default case, to ignore such an error as "unimplemented" I think. Though I'd like to see a notice or warning that this is the case, something like (once) "[WARN] the rmw implementation being used does not support ON_REQUESTED_INCOMPATIBLE_QOS events, you will not be notified when publishers and subscribers on the same topic do not have matching QoS".

at https://github.com/ros2/ros2/issues/822#issuecomment-598024481.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -88,6 +88,19 @@ class Publisher : public PublisherBase
...
+      try {
+        this->add_event_handler(
+          QOSOfferedIncompatibleQoSCallbackType(default_incompatible_qos_callback),
+          RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
+      } catch (UnsupportedEventTypeException & /*exc*/) {
+        RCLCPP_WARN_ONCE(
+          rclcpp::get_logger("rclcpp::Publisher"),
+          "This rmw implementation does not support ON_OFFERED_INCOMPATIBLE_QOS "
+          "events, you will not be notified when Publishers offer an incompatible "
+          "QoS profile to Subscriptions on the same topic.");
```
ivanpauno, line 102: I don't like to log a warning knowing that this will always be shown with `fastrtps` (until they support this events).

wjwwood:
> There may be valid reasons why people create publishers/subscriptions in the same topic that do not match.

They can suppress the warning by defining their own callback for this case. We could also have an option to disable this when creating the pub/sub.

> If there's enough debugging tools to check if a publisher/subscription pair matched, I don't think this is needed.

I guess, but this would be nice to have still imo. As a hint to people who don't understand why their system is not working.

rclcpp/include/rclcpp/publisher.hpp
wjwwood: Why?

We can add a way to suppress this warning, either programmatically or via an environment variable.

@mm318 Does fastrtps not support this already?

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -88,6 +88,19 @@ class Publisher : public PublisherBase
...
         RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
+    } else {
+      // Register default callback when not specified
+      try {
+        this->add_event_handler(
+          QOSOfferedIncompatibleQoSCallbackType(default_incompatible_qos_callback),
+          RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
+      } catch (UnsupportedEventTypeException & /*exc*/) {
+        RCLCPP_WARN_ONCE(
+          rclcpp::get_logger("rclcpp::Publisher"),
```
wjwwood, line 99: This should use the node's logger no?

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -122,6 +122,19 @@ class Subscription : public SubscriptionBase
...
         RCL_SUBSCRIPTION_REQUESTED_INCOMPATIBLE_QOS);
+    } else {
+      // Register default callback when not specified
+      try {
+        this->add_event_handler(
+          QOSRequestedIncompatibleQoSCallbackType(default_incompatible_qos_callback),
+          RCL_SUBSCRIPTION_REQUESTED_INCOMPATIBLE_QOS);
+      } catch (UnsupportedEventTypeException & /*exc*/) {
+        RCLCPP_WARN_ONCE(
+          rclcpp::get_logger("rclcpp::Subscription"),
```
wjwwood, line 133: Same.

wjwwood: And more cases below.

rclcpp/include/rclcpp/publisher.hpp
mm318: Correct, @wjwwood, Fast-RTPS currently still does not support this type of callback (tracked at https://github.com/eProsima/Fast-RTPS/issues/852).

rclcpp/include/rclcpp/publisher.hpp
mm318: I am unsure how to get the node's logger from a `Publisher`/`Subscription` class. The `Publisher` and `Subscription` only has access to the `NodeBaseInterface`. How do you get access to the `NodeLoggingInterface`?

rclcpp/include/rclcpp/publisher.hpp
mm318: > We can add a way to suppress this warning, either programmatically or via an environment variable.

Is there a standard way in ROS2 to suppress warnings?

rclcpp/include/rclcpp/publisher.hpp
ivanpauno: I think that adding the logging interface to the publisher can be done in a follow-up.
For example, we're doing this currently in the code: https://github.com/ros2/rclcpp/blob/57fddca648a2c0cc422717bc2360bef2d0d90b77/rclcpp/src/rclcpp/publisher_base.cpp#L103.

wjwwood: > I think that adding the logging interface to the publisher can be done in a follow-up.

I actually disagree, it will require an API change for the publisher class if we now want to pass in the logging interface, which I think we should. Better to do that now...

----

If not, you can copy what the `NodeLogger` does in the meantime:

https://github.com/ros2/rclcpp/blob/c1a7a655374a8ed68ca3f3de54adc4875c5e885d/rclcpp/src/rclcpp/node_interfaces/node_logging.cpp#L38

rclcpp/include/rclcpp/publisher.hpp
wjwwood: Not really no.

rclcpp/include/rclcpp/publisher.hpp
mm318: Done.

rclcpp/include/rclcpp/publisher.hpp
mm318: This will essentially be controlled by `options_.use_default_callbacks`.

rclcpp/include/rclcpp/subscription.hpp
mm318: Done.

mm318:
> * Adding a `options_.use_default_callbacks`, default to `true`.

I have implemented avoiding default callbacks this way.

I have addressed the suggestions made here, and applied them to `rclpy` (in https://github.com/ros2/rclpy/pull/536) as well.

mm318:
@ros2/aws-oncall - please run this CI job
Gist: `https://gist.githubusercontent.com/mm318/972b53569da7d10427bf13d8992290c2/raw/0ab88f1d350b57d36fad04bb0f6358aa8b87cb14/ros2_incompatible_qos_callbacks.repos`
BUILD args: `--packages-up-to rclcpp rclpy`
TEST args: `--packages-select rclcpp rclpy`
Job: `ci_launcher`

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10133/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5700/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8234/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10028/)

mm318:
> Edit: ament_export_warnings only, no test failures.

Thanks for the clarification!

piraka9011:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10029/)

piraka9011:
@mm318 mac OS has test failures tho.

mm318:
I believe the Windows and macOS test failures are unrelated to the changes in this pull request.

PR #1052, opened 2020-04-08
Fix indents after paren and newline

rotu:
as per https://github.com/ament/ament_lint/pull/174


rclcpp_lifecycle/test/test_state_machine_info.cpp
```diff
@@ -63,13 +63,13 @@ TEST_F(TestStateMachineInfo, available_transitions) {
...
       (transition.start_state().id() >= 10 &&
-      (transition.start_state().id() <= 15)));
+        (transition.start_state().id() <= 15)));
     EXPECT_FALSE(transition.start_state().label().empty());
 
     EXPECT_TRUE(
       transition.goal_state().id() <= 4 ||
       (transition.goal_state().id() >= 10 &&
-      (transition.goal_state().id() <= 15)));
+        (transition.goal_state().id() <= 15)));
```
hidmic, line 72: @rotu forcing an indent here is somewhat awkward.

rclcpp_components/include/rclcpp_components/register_node_macro.hpp
```diff
@@ -32,7 +32,7 @@
  */
 #define RCLCPP_COMPONENTS_REGISTER_NODE(NodeClass) \
   CLASS_LOADER_REGISTER_CLASS( \
-    rclcpp_components::NodeFactoryTemplate<NodeClass>, \
-    rclcpp_components::NodeFactory)
+  rclcpp_components::NodeFactoryTemplate<NodeClass>, \
+  rclcpp_components::NodeFactory)
```
hidmic, line 36: @rotu losing indentation here is unfortunate.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -246,8 +246,8 @@ void TimeSource::on_parameter_event(const rcl_interfaces::msg::ParameterEvent::S
   }
   // Filter for only 'use_sim_time' being added or changed.
   rclcpp::ParameterEventsFilter filter(event, {"use_sim_time"},
-    {rclcpp::ParameterEventsFilter::EventType::NEW,
-      rclcpp::ParameterEventsFilter::EventType::CHANGED});
+  {rclcpp::ParameterEventsFilter::EventType::NEW,
+    rclcpp::ParameterEventsFilter::EventType::CHANGED});
```
hidmic, line 250: @rotu uh, would leaving the opening curly brace on line 248 improve things a bit?

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -401,12 +401,12 @@ SyncParametersClient::SyncParametersClient(
...
-    node_graph_interface,
-    node_services_interface,
-    remote_node_name,
-    qos_profile);
+      node_base_interface,
+      node_topics_interface,
+      node_graph_interface,
+      node_services_interface,
+      remote_node_name,
+      qos_profile);
```
hidmic, line 409: @rotu this is a clear improvement IMHO

rclcpp_lifecycle/test/test_state_machine_info.cpp
rotu: It seems reasonable to me. That line is a continuation of the parenthetical started on the previous line

rclcpp_components/include/rclcpp_components/register_node_macro.hpp
rotu: It is unfortunate. https://github.com/ament/ament_lint/pull/174#issuecomment-582688531

I don't understand why, and I think it's an uncrustify bug.

rclcpp/src/rclcpp/time_source.cpp
rotu: Maybe, but there's a bigger issue here - for some reason, Uncrustify is not recognizing this as a mult-line function call. There should be a line break after `filter`:

https://github.com/ros2/rclcpp/blob/ee6ab95cfc8ecab353d4cf1efdb81de237424f11/rclcpp/src/rclcpp/time_source.cpp#L248-L250

rotu: submitted https://github.com/ament/ament_lint/issues/231

rclcpp/src/rclcpp/parameter_client.cpp
rotu: Yeah. It's weird that this wasn't the case before. `uncrustify` must be pretending there's an implicit parenthesis on the previous line and aligning to it.

ivanpauno:
Closing, until we have an agreement in https://github.com/ament/ament_lint/pull/174 that this is the desired style (IIUC, this doesn't pass with the current linters configuration, sorry if I'm wrong).

rotu:
@ivanpauno this is the desired style as documented in the developer guide.

This PR has some “examples of the proposed changes” as requested in https://github.com/ament/ament_lint/pull/174#issuecomment-611158357

It seems fixing ament_lint is an uphill battle, and there needs to be a better way than a billion PRs more or less simultaneously. Maybe ament_uncrustify should only check changed code to prevent this sort of “churny” PR. Or maybe we should version the checking rules so that uncrustify and consuming packages don’t need to move in lock-step.

@dirk-thomas, thoughts?

ivanpauno:
> This PR has some “examples of the proposed changes” as requested in ament/ament_lint#174 (comment)

I think that having an example is great, and the `ament_lint` issue still references this PR.

> It seems fixing ament_lint is an uphill battle, and there needs to be a better way than a billion PRs more or less simultaneously. Maybe ament_uncrustify should only check changed code to prevent this sort of “churny” PR. Or maybe we should version the checking rules so that uncrustify and consuming packages don’t need to move in lock-step.

The disadvantage of that is that we will have dissimilar styles in different packages.
Though doing the change in a lot of packages at the same time isn't nice, I think it's the correct thing to do if we want to update the style.

dirk-thomas:
> Maybe ament_uncrustify should only check changed code to prevent this sort of “churny” PR.

The concept of "changed code" doesn't exist when you e.g. run the tests for a workspace since there is not reference point to determine a change against.

> Or maybe we should version the checking rules so that uncrustify and consuming packages don’t need to move in lock-step.

Configuration files could be versioned and downstream packages could choose a specific version. That is already technically possible since you can pass a specific config file to the linter. But as a matter of fact all existing packages use the default config and therefore would still need to be updated at least one.

Also as soon as the uncrustify version is changes there will likely be breaking changes. So you would need to ship different uncrustify versions in parallel.

If you are interested in either of these ideas please consider to contribute towards this (with pull requests and if the changes are more complex consider to start with a design issue to discuss the proposal).

rotu:
Created https://github.com/ament/ament_lint/issues/233 to trace the issue with improving linters.

PR #1054, opened 2020-04-09
Add -DNOMINMAX compiler definition for Windows

jacobperron:
This fixes an issue where max() and min() are treated as macros defined in windows.h

Alternative to https://github.com/ros2/rclcpp/pull/1053

Windows: [Build Status](https://ci.ros2.org/job/ci_windows/9921/)

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -24,6 +24,11 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+# Avoid confusing max() and min() for Windows macros
+if(WIN32)
+  add_definitions(-DNOMINMAX)
+endif()
+
 include_directories(include)
 
 set(${PROJECT_NAME}_SRCS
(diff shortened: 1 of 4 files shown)


wjwwood:
I think this is also ok to do, but there are cases where people do want the min/max macros, e.g.:

https://developercommunity.visualstudio.com/content/problem/114149/gdiplush-does-not-compile-when-nominmax-is-defined.html

Also, unless we export that definition it will not work when someone trys to use our headers (without adding the compiler definition themselves).

jacobperron:
> Also, unless we export that definition it will not work when someone trys to use our headers (without adding the compiler definition themselves).

I did not want to export the definition. I made this change assuming that we don't want to use the min/max macros in the rclcpp packages. Users can still use them if they want, and will have to workaround our definitions if they happen to include the macros before our headers.

dirk-thomas:
Since this is not necessary anymore and wouldn't prevent downstream packages to fail I don't think this should be merged.

jacobperron:
I see it as defensive. We're currently relying on external packages (e.g. rcutils) to not include Windows.h or have them define `NOMINMAX` for us. Knowing that we use max() and min() in rclcpp packages, this change avoids accidental breaks due to upstream packages introducing the Windows macros.

dirk-thomas:
If any dependency introduces such an include this patch I would expect the patch to only work around the problem when building this package but downstream code would still fail to build, right?

jacobperron:
> If any dependency introduces such an include this patch I would expect the patch to only work around the problem when building this package but downstream code would still fail to build, right?

That's right, I was only trying to ensure that this package builds. In retrospect, maybe not a great patch considering it might be better for us to avoid shipping a "broken" rclcpp accidentally.

PR #1056, opened 2020-04-09
[eloquent] Ensure rcl_clock accesses are thread-safe (ABI-safe version).

clalancette:
This is an alternative PR to https://github.com/ros2/rclcpp/pull/999
The code in this one is less nice, but it is ABI-safe.  Thus,
it can be used as a short-term workaround for users hurt by
the problem now, and/or used as the solution for Eloquent.

I'm still testing this, so consider this a request for comment on the approach.  Once I run a full CI on it, I'll be more confident in it.

### DIFF
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -143,10 +143,8 @@ Clock::create_jump_callback(
     throw std::bad_alloc{};
   }
 
-  // Lookup the per-object mutex
-  std::mutex & clock_mutex = get_clock_mutex(this);
   {
-    std::lock_guard<std::mutex> clock_guard(clock_mutex);
+    std::lock_guard<std::mutex> clock_guard(get_clock_mutex(this));
     // Try to add the jump callback to the clock
     rcl_ret_t ret = rcl_clock_add_jump_callback(
       &rcl_clock_, threshold, Clock::on_time_jump,
@@ -160,8 +158,7 @@ Clock::create_jump_callback(
   // create shared_ptr that removes the callback automatically when all copies are destructed
   // TODO(dorezyuk) UB, if the clock leaves scope before the JumpHandler
   return JumpHandler::SharedPtr(handler.release(), [this](JumpHandler * handler) noexcept {
-    std::mutex & clock_mutex = get_clock_mutex(this);
-    std::lock_guard<std::mutex> clock_guard(clock_mutex);
+    std::lock_guard<std::mutex> clock_guard(get_clock_mutex(this));
 
     rcl_ret_t ret = rcl_clock_remove_jump_callback(&rcl_clock_, Clock::on_time_jump,
         handler);
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -28,6 +28,12 @@ namespace rclcpp
 std::shared_timed_mutex g_clock_map_mutex;
 std::unordered_map<rclcpp::Clock *, std::mutex> g_clock_mutex_map;
 
+std::mutex & get_clock_mutex(rclcpp::Clock * clock)
+{
+  std::shared_lock<std::shared_timed_mutex> lk(g_clock_map_mutex);
+  return g_clock_mutex_map.at(clock);
+}
+
 JumpHandler::JumpHandler(
   pre_callback_t pre_callback,
   post_callback_t post_callback,
@@ -138,9 +144,7 @@ Clock::create_jump_callback(
   }
 
   // Lookup the per-object mutex
-  g_clock_map_mutex.lock_shared();
-  std::mutex & clock_mutex = g_clock_mutex_map.at(this);
-  g_clock_map_mutex.unlock_shared();
+  std::mutex & clock_mutex = get_clock_mutex(this);
   {
     std::lock_guard<std::mutex> clock_guard(clock_mutex);
     // Try to add the jump callback to the clock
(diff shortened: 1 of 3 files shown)


clalancette:
CI on Linux (all packages, all tests, Eloquent/bionic): [Build Status](http://ci.ros2.org/job/ci_linux/10019/)

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -118,18 +137,30 @@ Clock::create_jump_callback(
...
 
-  // Try to add the jump callback to the clock
-  rcl_ret_t ret = rcl_clock_add_jump_callback(
-    &rcl_clock_, threshold, Clock::on_time_jump,
-    handler.get());
-  if (RCL_RET_OK != ret) {
-    exceptions::throw_from_rcl_error(ret, "Failed to add time jump callback");
+  // Lookup the per-object mutex
+  g_clock_map_mutex.lock_shared();
+  std::mutex & clock_mutex = g_clock_mutex_map.at(this);
```
wjwwood, line 142: Scoped would be nice here, since if this threw then the mutex would remain locked. You could use a function or a local lambda to enclose the scope but return the reference.

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -118,18 +137,30 @@ Clock::create_jump_callback(
...
+    }
   }
 
   // *INDENT-OFF*
   // create shared_ptr that removes the callback automatically when all copies are destructed
   // TODO(dorezyuk) UB, if the clock leaves scope before the JumpHandler
   return JumpHandler::SharedPtr(handler.release(), [this](JumpHandler * handler) noexcept {
+    g_clock_map_mutex.lock_shared();
+    std::mutex & clock_mutex = g_clock_mutex_map.at(this);
+    g_clock_map_mutex.unlock_shared();
```
wjwwood, line 161: In fact, as often as you do this, it might be nice to have a function for it, like `get_clock_mutex(Clock *) -> std::mutex &`.

clalancette: Good point.  I'll go ahead and make a function for it, which should solve both of your concerns.

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -138,9 +144,7 @@ Clock::create_jump_callback(
   }
 
   // Lookup the per-object mutex
-  g_clock_map_mutex.lock_shared();
-  std::mutex & clock_mutex = g_clock_mutex_map.at(this);
-  g_clock_map_mutex.unlock_shared();
+  std::mutex & clock_mutex = get_clock_mutex(this);
   {
     std::lock_guard<std::mutex> clock_guard(clock_mutex);
```
wjwwood, line 149: Why not this:

```suggestion
    std::lock_guard<std::mutex> clock_guard(g_clock_mutex_map.at(this));
```

And remove the line above?

clalancette: Good call.  Done in 685bd50

clalancette:
CI on Linux (all packages, all tests, Eloquent/bionic): [Build Status](http://ci.ros2.org/job/ci_linux/10124/)

clalancette:

clalancette:
@mjcarroll correctly pointed out that https://ci.ros2.org/job/ci_linux/10124/testReport/junit/(root)/test_rclcpp/gtest_spin__rmw_connext_cpp_gtest_missing_result/ looks suspiciously like a regression from this patch.  I wasn't able to reproduce locally, but I think it needs more looking into.

clalancette:
After doing a bit of debugging here, it turns out that there are some accesses to the `TimerBase` object that don't go through the `Clock` constructor, so they never get a new entry in the global mutex map assigned.  I'm not sure how that can happen (the TimerBase constructor takes a `rclcpp::Clock::SharedPtr`), but it seems to be the cause of the crash we are seeing in CI.  Needs more investigation.

clalancette:
Thanks to some help from @hidmic, I think we've solved the crash.  Here is another CI on Linux (all packages, all tests, Eloquent/Bionic), to be sure: [Build Status](http://ci.ros2.org/job/ci_linux/11172/)

clalancette:
There are 194 failures in the Linux build.  If I filter out all of the flake8 and pep257 ones, that leaves me with the following list:

```
    projectroot.test_publisher_subscriber__rclpy__rmw_connext_cpp__rmw_cyclonedds_cpp
    projectroot.test_publisher_subscriber__rclpy__rclcpp__rmw_connext_cpp__rmw_cyclonedds_cpp
    projectroot.test_publisher_subscriber__rclcpp__rclpy__rmw_connext_cpp__rmw_cyclonedds_cpp
    projectroot.test_publisher_subscriber__rclcpp__rmw_connext_cpp__rmw_cyclonedds_cpp
    projectroot.test_events__rmw_cyclonedds_cpp
    rcl.TestEventFixture__rmw_cyclonedds_cpp.test_pubsub_deadline_missed
    rcl.TestEventFixture__rmw_cyclonedds_cpp.test_pubsub_liveliness_kill_pub
    test_communication.TestPublisherSubscriberAfterShutdown.test_processes_finished_gracefully[WStrings]
    test_communication.TestPublisherSubscriberAfterShutdown.test_processes_finished_gracefully[WStrings]
    test_communication.TestPublisherSubscriberAfterShutdown.test_processes_finished_gracefully[WStrings]
    test_communication.TestPublisherSubscriberAfterShutdown.test_processes_finished_gracefully[WStrings]
```

I believe that the WString interoperability with Connext is a known issue, so we can ignore those.  That leaves us with:

```
    rcl.TestEventFixture__rmw_cyclonedds_cpp.test_pubsub_deadline_missed
    rcl.TestEventFixture__rmw_cyclonedds_cpp.test_pubsub_liveliness_kill_pub
```

But if memory serves correctly, the Cyclone version in Eloquent didn't support liveliness or deadline.  So overall, I think we're pretty good to go here.  I'm going to run one last CI on the other platforms, then go ahead and merge this.

clalancette:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6458/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9129/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11093/)

clalancette:
Windows "failed" because it had too many test failures.  But I honestly don't know what I should expect for failures there.  Here's a Windows CI run *without* this patch:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11120/)

clalancette:
While we could potentially slip this into Eloquent, I think it is too risky given that Eloquent is going EOL in 4-6 weeks.  I'm just going to close this out; the fix is in Foxy and later.

PR #1059, opened 2020-04-10
Make sure to include what you use.

clalancette:

For reasons I don't understand, I could not build on Windows without these includes, even though CI seems OK with it.  Regardless, this seems to be the right thing to do, and with this in place I can now build on Windows again.

### DIFF
--- a/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
+++ b/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
@@ -15,6 +15,7 @@
 #ifndef RCLCPP__EXECUTORS__STATIC_EXECUTOR_ENTITIES_COLLECTOR_HPP_
 #define RCLCPP__EXECUTORS__STATIC_EXECUTOR_ENTITIES_COLLECTOR_HPP_
 
+#include <chrono>
 #include <list>
 #include <memory>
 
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10088/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5674/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8207/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9991/)

clalancette:
> Additions make sense to me, LGTM! Are all test failures unrelated?

I will double-check that before merging.  Thanks for the review.

clalancette:
Looking through the test failures:

1.  All platforms have a CMake warning from `ament_export_interfaces.cmake`, coming from tracetools.  That is unrelated to this PR.
1.  macOS has two test failures in `count_api`, but those are also in the nightlies.
1.  Windows has a host of failures.  Most of those are *not* in the latest nightly that finished (https://ci.ros2.org/view/nightly/job/nightly_win_rel/1518).  I have a somewhat hard time seeing how this PR would cause all of those failures.  I'm going to kick Windows one more time to see if that was flakey for some reason.

clalancette:
All right, looks indeed like Windows was a flake; it succeeded this time with no test failures, though it had the same CMake warning as the other platforms.  I'm going to merge, thanks for the reviews.

PR #1063, opened 2020-04-14
Revert #1047

rotu:
Fix CI breakages in at least test_rclcpp and rosbag2_transport. It seems upstream CI was never run.
As per https://github.com/osrf/osrf_testing_tools_cpp/pull/36#issuecomment-591088977

This reverts commit aaf8b3828cb4da7e5dd3f017e653f132e5b0ba8c.

Signed-off-by: Dan Rose <dan@digilabs.io>


wjwwood:
Normally we revert when we're not sure that someone is working on a fix or if the fix is not trivial. Since I'm on the case and the fix is trivial, we'll just merge the fix to avoid churn.

Sorry for the disruption.

rotu:
@wjwwood could you please make sure CI gets run this time?

wjwwood:
@rotu I will. I already apologized for the mistake, and I am sorry. There's no need to be rude.

wjwwood:
Also, I sort of just assumed snark out of your most recent comment, if you didn't intend that, I'm sorry for assuming so. I know getting an issue closed can seem negative as well, but I appreciate your initiative on opening this and in a different circumstance would have taken it.

rotu:
@wjwwood Thank you! I saw your comment on this before the one on the PR and did not mean to badger you or imply that you’re being careless, so if I did either, I’m sorry. Development is hard and lots of things can go wrong. It’s a miracle that they go *right* so often!

PR #1067, opened 2020-04-16
Remove warning message for failing to register default incompatible QoS callback

mm318:
This is to synchronize `rclcpp` with the change being made to `rclpy` in https://github.com/ros2/rclpy/pull/544.


mm318:
> In case that yes, I would delete the warning in `rclcpp` too: https://github.com/aws-ros-dev/rclcpp/blob/f6d0fc388fbf9baf3262aa158b32569228437de8/rclcpp/include/rclcpp/publisher.hpp#L100.

@ivanpauno, this pull request is for deleting the warning from `rclcpp` (https://github.com/ros2/rclpy/pull/544#issuecomment-614625361).

PR #1069, opened 2020-04-16
Generate node interfaces' getters and traits.

hidmic:
Based on https://github.com/ros2/rclcpp/compare/node_interfaces_traits. Alternative to #1035. Forces explicit shared ownership of interfaces.


Karsten1987:
I hope you don't mind me pushing on you branch. I just pushed a few things I found in order to make this compile. It was mainly the `inline` keyword to avoid duplicate symbols.

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -60,8 +60,8 @@ create_timer(
   rclcpp::callback_group::CallbackGroup::SharedPtr group = nullptr)
 {
   return create_timer(
-    rclcpp::node_interfaces::get_node_base_interface(node),
-    rclcpp::node_interfaces::get_node_timers_interface(node),
+    rclcpp::node_interfaces::get_node_base_interface(node).get(),
+    rclcpp::node_interfaces::get_node_timers_interface(node).get(),
```
Karsten1987, line 64: wouldn't it make sense to change the `create_timer` API to take a shared pointer? Seems kind of odd that other `create_x` functions take shared pointers, but timers don't, e.g. https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/create_client.hpp#L33-L35

rclcpp/resource/get_interface.hpp.em
```diff
@@ -0,0 +1,108 @@
...
+namespace node_interfaces
+{
+namespace detail
+{
+
+// If NodeType has a method called get_@(interface_name)() which returns a shared pointer.
+template<
+  typename NodeType,
+  typename std::enable_if<has_@(interface_name)<
+    typename std::remove_pointer<NodeType>::type
```
Karsten1987, line 55: just like `std::is_pointer`, `std::remove_pointer` doesn't take smart pointers into account. I am working on a thing `rcpputils::remove_pointer` which does that. That would make the second template specialization in `has_interface` obsolete.

Karsten1987:
Connected PR here: https://github.com/ros2/rcpputils/pull/56

rclcpp/include/rclcpp/create_timer.hpp
hidmic: That's true, but I didn't want this PR to cover more ground than it needs to. We can correct that in a follow-up PR.

hidmic:
> I just pushed a few things I found in order to make this compile. It was mainly the inline keyword to avoid duplicate symbols.

Uh, I didn't run into any locally (on a Ubuntu 18.04 box, using gcc 7.4.0). Thanks !

Karsten1987:
CI (build everything, test everything above rclcpp):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10278/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5814/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8359/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10174/)

Karsten1987:
CI (build everything, test everything above rclcpp):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10286/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5821/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8366/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10181/)

hidmic:
Test failures seem unrelated, but there's an odd `cppcheck` failure in MacOS for `test_rclcpp`. @Karsten1987 have you seen it before?

Karsten1987:
from what I can tell it timed out without testing any code. I can't reproduce it locally - cppcheck works correctly on my OSX.

wjwwood:
I'm not really a fan of generated code in this case. I don't think it adds much value over just committing the various files that are being generated here. Obviously DRY is good, but I don't think it's worthwhile trade-off if you have to go around the language to depend on the build system and external template engines. If there were a way to do this C++ (there is it's called meta classes in C++20) then I'd be all for it, but since there's not, I prefer committing the code, repetitive as it is, over generating it, because I think it's easier to read and understand.

But I do think we need the various getters, so if you guys want to push ahead with this direction that's fine. I know I haven't had time to review the alternative pr either, so if you guys have time and desire to push on it, then I think its fair you get to make the design decisions too.

hidmic:
I don't have a strong opinion either way, but we do have to make a call. @ros2/team WDYT?

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -32,8 +32,8 @@ namespace rclcpp
 template<typename CallbackT>
 typename rclcpp::TimerBase::SharedPtr
 create_timer(
-  node_interfaces::NodeBaseInterface * node_base,
-  node_interfaces::NodeTimersInterface * node_timers,
+  std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
+  std::shared_ptr<node_interfaces::NodeTimersInterface> node_timers,
```
ivanpauno, line 36: I think this is reasonable, but we will also break a lot of downstream packages if the old function returning a raw pointer is deleted.
IMO, we should deprecate instead of delete it.

hidmic: I agree, and we could so for `rclcpp::create_timer`, but not for the generated API (at least not seamlessly, template argument deduction won't be carried out for return types). #1035, on the other hand, doesn't drop API. Ultimately, there's a discussion about ownership that has to settled on.

ivanpauno: > Ultimately, there's a discussion about ownership that has to settled on.

I think that it's up to the class requiring the interfaces to decide if they need shared ownership or just a non-owning pointer.

IMO, we should only provide getters returning shared pointers, and deprecate the old ones returning raw pointers (which can be replaced with `get_shared_*_interface().get()`).
But deleting that signature now will require a lot of changes in downstream packages.

> I agree, and we could so for rclcpp::create_timer, but not for the generated API (at least not seamlessly, template argument deduction won't be carried out for return types)

I don't understand why it can be done in the generated API, I will check again tomorrow.

jacobperron: It might not look pretty, but we should be able to write some empy conditions to check if we're generating getters for one of the previously existing interfaces and insert the deprecated API.

ivanpauno: > I agree, and we could so for rclcpp::create_timer, but not for the generated API (at least not seamlessly, template argument deduction won't be carried out for return types).

Yes, template argument deduction won't work, but the same applies for the PR creating the files manually.
Both functions need to have a different name.

hidmic: Precisely, that's what #1035 does. We can make this PR do the same, though it'll take two deprecation cycles for the `std::shared_ptr` flavor API to get the name we intended it to have here.

ivanpauno: > We can make this PR do the same, though it'll take two deprecation cycles for the std::shared_ptr flavor API to get the name we intended it to have here.

That's a great point.
If you want to continue with the API breaking version, I can help with reviewing the PRs needed in the downstream packages.
It would also be good to add a comment in the release notes.

If you want to continue with two functions with different names, and optionally deprecate the ones returning a raw pointer, I also approve it.

hidmic: @jacobperron we need an executive decision here.

jacobperron: I see. I'd rather not go through two deprecation cycles to get the API we want. I think it will be okay to have this API break and document it in the release notes.

hidmic:
Alright, since everything above `rclcpp` is passing CI, I think this is good to go. @Karsten1987 has already approved, but it would be good if you guys can give your approval too @ivanpauno @jacobperron

ivanpauno:
> It would also be good to add a comment in the release notes.

It would be great to do this.

hidmic:
Alright, https://github.com/ros2/ros2_documentation/pull/621 is up. Merging.

PR #1070, opened 2020-04-17
Export component manager

DensoADAS:
Exporting the component manger library and the dependency to composition interfaces is missing.


Karsten1987:
thanks for the contribution

PR #1072, opened 2020-04-18
Integrate topic statistics

dabonnie:
~~This is currently a draft PR and is blocked by https://github.com/ros2/rclcpp/pull/1050 (when #1050 is merged this can come out of draft), where the PR uses components introduced there.~~

~~Unit tests are a work in progress, but I've manually verified that Topic Statistics can be published via the `demo_nodes_cpp` talker + listener demo on Linux (Ubuntu 18.04).~~

This is ready for review.


dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10299/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5833/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8385/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10201/)

I have two local test failures on Linux that I will be investigating, specifically

```
18/47 Test #18: test_publisher_subscription_count_api .......***Failed   32.27 sec
25/47 Test #25: test_subscription_publisher_count_api .......***Failed   16.25 sec
```

dabonnie:
@wjwwood this is ready for review. Passing locally on Linux but need to run CI. 

Hopefully this provides some context to https://github.com/ros2/rclcpp/pull/1074#pullrequestreview-396912517. My initial plan was to create the `SubscriptionTopicStatistics` instance at the `create_subscription` or factory level, but including the Node and publisher headers there led to compile time errors (more context here: https://github.com/ros2/rclcpp/pull/1050#issuecomment-615545995). Currently I don't see a cleaner way to do this and would appreciate your input, thanks!

dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10329/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5857/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8413/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10238/)

dabonnie:
Will re-run CI (for everything on Linux per https://github.com/ros2/rclcpp/pull/1074#issuecomment-617268892). @wjwwood could you please review today?

dabonnie:
[Build Status](https://ci.ros2.org/job/ci_linux/10354/)

rclcpp/test/test_publisher_subscription_count_api.cpp
```diff
@@ -51,6 +51,10 @@ class TestPublisherSubscriptionCount : public ::testing::TestWithParam<TestParam
   {
     rclcpp::init(0, nullptr);
   }
+  static void TearDownTestCase()
+  {
+    rclcpp::shutdown();
+  }
```
wjwwood, line 57: Please, lets resolve this in the other pull request:

https://github.com/ros2/rclcpp/pull/1073

And remove these from this unrelated pull request.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -72,23 +122,25 @@ class EmptySubscriber : public rclcpp::Node
...
     subscription_ = create_subscription<Empty,
         std::function<void(Empty::UniquePtr)>>(
       topic,
       rclcpp::QoS(rclcpp::KeepAll()),
-      callback);
+      callback,
+      options);
   }
 
-  virtual ~EmptySubscriber() = default;
```
wjwwood, line 85: If you remove this, then it is good practice to mark the class as `final`, because it is no longer safe to inherit from it generally, though it may still be in specific cases.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -63,6 +77,42 @@ class TestSubscriptionTopicStatistics : public SubscriptionTopicStatistics<Callb
   }
 };
 
+/**
+ * Empty publisher node: used to publish empty messages
+ */
+class EmptyPublisher : public rclcpp::Node
```
wjwwood, line 83: This should also either be `final` or have a virtual destructor I think.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -63,6 +77,42 @@ class TestSubscriptionTopicStatistics : public SubscriptionTopicStatistics<Callb
...
+private:
+  void publish_message()
+  {
+    ++number_published_;
+    auto msg = Empty{};
+    publisher_->publish(msg);
+  }
+
+  rclcpp::Publisher<Empty>::SharedPtr publisher_;
+  std::atomic<int> number_published_{0};
```
wjwwood, line 112: Does it need to be signed?

```suggestion
  std::atomic<size_t> number_published_{0};
```

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -92,13 +95,40 @@ Node::create_subscription(
...
+    std::shared_ptr<Publisher<statistics_msgs::msg::MetricsMessage>> publisher =
+      this->create_publisher<statistics_msgs::msg::MetricsMessage>(
+      options.topic_stats_options.publish_topic,
+      qos);
+
+    auto sub_topic_stats = std::make_shared<
+      rclcpp::topic_statistics::SubscriptionTopicStatistics<CallbackMessageT>
+      >(this->get_name(), publisher);
+
+    auto timer = this->create_wall_timer(
```
wjwwood, line 108: Is it really the best idea to have a timer and publisher for each subscription with this setting turned on?

The most common use case in ROS 1 was to turn on statistics for all the topics in the graph so you could introspect multiple things simultaneously and see which topics were using the most resources or lagging behind (you didn't know ahead of time which you wanted to look at), in which case that is a lot of extra timers and publishers... Really these should share a resource for this, no?

Each subscription could collect statistical info asynchronously (as callbacks are called) and a single timer could trigger to collect them all and publish them all on a single publisher. I think that would be a much better approach, would it not?

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -92,13 +95,40 @@ Node::create_subscription(
...
   typename MessageMemoryStrategyT::SharedPtr msg_mem_strat)
 {
-  return rclcpp::create_subscription<MessageT>(
-    *this,
-    extend_name_with_sub_namespace(topic_name, this->get_sub_namespace()),
-    qos,
-    std::forward<CallbackT>(callback),
-    options,
-    msg_mem_strat);
+  if (options.topic_stats_options.state == rclcpp::TopicStatisticsState::Enable) {
```
wjwwood, line 98: I don't understand why this needs to be in the `rclcpp::Node` class. I know you said there were circular references or something preventing you from putting it into the `Subscription` class (I ran out of time investigating this, but I believe it should be possible with some rearranging or forward declaring), but this branching logic should at least be in `rclcpp::create_subscription` instead. Putting it here means that `rclcpp_lifecycle::LifecycleNode` does not get this feature, and it generally couples this feature to `rclcpp::Node` only, which is not ideal.

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -260,6 +268,13 @@ class Subscription : public SubscriptionBase
     }
     auto typed_message = std::static_pointer_cast<CallbackMessageT>(message);
     any_callback_.dispatch(typed_message, message_info);
+
+    if (subscription_topic_statistics_) {
+      const auto nanos = std::chrono::time_point_cast<std::chrono::nanoseconds>(
+        std::chrono::steady_clock::now());
+      const auto time = rclcpp::Time(nanos.time_since_epoch().count());
```
wjwwood, line 275: This is a bug, imo. You convert a steady time to a `rclcpp::Time` which default to ROS time, when steady time's epoch and ROS time's epoch are different. Much of the time steady time is counting since the start of the machine, whereas ROS time (and system time) count from the unix epoch on Jan 1, 1970.

wjwwood: If you use the `rclcpp::Time` in the signature below you will either need to set the clock time to steady time:

https://github.com/ros2/rclcpp/blob/649d72f835bfca334e7eab39a8c16752094d6317/rclcpp/include/rclcpp/time.hpp#L46
https://github.com/ros2/rcl/blob/b463168528e99024b32c8c2181b96d3cfeb42084/rcl/include/rcl/time.h#L64

Or use ROS/system time and get the current time from the node's clock.

rclcpp/test/test_publisher_subscription_count_api.cpp
dabonnie: Will do, sorry for the local changes.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
dabonnie: This is a mistake, will revert.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
dabonnie: Will add a virtual default dtor.

rclcpp/include/rclcpp/node_impl.hpp
dabonnie: My fear here was the complexity of the approach and use case. 

Per the global timer: I didn't think this made since given the variety of potential topic publishing periods. For example, the default 1 second publishing period makes sense for something collected at a high frequency, but may not for much lower frequencies. 

We could have a global publisher and this was considered, but for the immediate approach I chose a per subscription publisher for ease of lifetime management. 

I think we could do the following: have a global publisher for topic statistics, where the publisher is created and held by the node, but have separate timers. How does that sound?

wjwwood: Sure ok, that's definitely better.

I see now what you mean about the timers, since each subscription can set its statistics publish rate separately. I wonder though if that make sense now (connecting the publish rate with the subscription). Why not have one publisher and timer per node and have it publish all the statistics at once? What use case(s) supports the design decision to publish them separately and asynchronously? Do you know off hand how it was done in ROS 1? Was there one publisher that gathered all the stats and published them in bulk periodically or a separate publishing event for each contributor of stats (each subscription)?

An in between would be to have the Subscriptions register themselves with a "statistics publisher" class, which periodically queries all registered subscriptions and publishes their stats together, then you could have as many or few of those statistic publishers as you want, and have a default one in the node which is used if nothing else is specified. This is kind of how callback groups work, you can create one and pass it when making subscriptions, but if you don't specify one the default one in the node is used. This could also be extended more easily to incorporate more sources of stats, things other than subscriptions, in the future. They would just register themselves with a statistics publisher, collect stats asynchronously, and wait to be queried by the statistics publisher.

I think by far the dominant use case will be to specify nothing when creating the subscription, and then want to turn on stats externally without changing any code. So I expect most users would be fine with posting stats in bulk at a moderate frequency (~1hz), even if the subscriptions receive data at a slower rate. I think it's nice to be able to further configure this (which subscriptions publish stats, how and how often they publish them), but I expect that will be an uncommon use case.

rclcpp/include/rclcpp/node_impl.hpp
dabonnie: I tried implementing in the `create_subscription` class as well and was getting compiler errors (similar issue when trying to create in the `Subscription` class). I'm again trying to see if it's possible in a branch, but simply moving the working code from `node_impl` to `create_subscription`

```
typename std::shared_ptr<SubscriptionT>
create_subscription(
  NodeT && node,
  const rclcpp::QoS & qos,
  CallbackT && callback,
  const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
    rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
  ),
  typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (
    MessageMemoryStrategyT::create_default()
  )
)
{
  using rclcpp::node_interfaces::get_node_topics_interface;
  auto node_topics = get_node_topics_interface(std::forward<NodeT>(node));

  std::shared_ptr<rclcpp::topic_statistics::SubscriptionTopicStatistics<CallbackMessageT>>
    subscription_topic_stats = nullptr;

  if (rclcpp::detail::resolve_enable_topic_statistics(options, *node.node_base_)) {
    auto  publisher =
      node.create_publisher<statistics_msgs::msg::MetricsMessage>(options.topic_stats_options.publish_topic, 10);

    auto subscription_topic_stats = std::make_shared<
      rclcpp::topic_statistics::SubscriptionTopicStatistics<CallbackMessageT>
    >(node.get_name(), publisher);

    auto timer = node.create_wall_timer(
      options.topic_stats_options.publish_period, [subscription_topic_stats]() {
        subscription_topic_stats->publish_message();
      });

    subscription_topic_stats->set_publisher_timer(timer);
  }
(rest of the method omitted)
```
 results in an error I can't get past yet

```
[ 12%] Building CXX object CMakeFiles/rclcpp.dir/src/rclcpp/executor.cpp.o
In file included from /home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/node_impl.hpp:40:0,
                 from /home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp:1167,
                 from /home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/src/rclcpp/executor.cpp:25:
/home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/create_subscription.hpp: In function ‘std::shared_ptr<SubscriptionT> rclcpp::create_subscription(NodeT&&, const string&, const rclcpp::QoS&, CallbackT&&, const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>&, typename MessageMemoryStrategyT::SharedPtr)’:
/home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/create_subscription.hpp:80:65: error: expected primary-expression before ‘>’ token
       node.create_publisher<statistics_msgs::msg::MetricsMessage>(options.topic_stats_options.publish_topic, 10);
                                                                 ^
CMakeFiles/rclcpp.dir/build.make:403: recipe for target 'CMakeFiles/rclcpp.dir/src/rclcpp/executor.cpp.o' failed
make[2]: *** [CMakeFiles/rclcpp.dir/src/rclcpp/executor.cpp.o] Error 1
CMakeFiles/Makefile2:1182: recipe for target 'CMakeFiles/rclcpp.dir/all' failed
make[1]: *** [CMakeFiles/rclcpp.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
```

rclcpp/include/rclcpp/subscription.hpp
dabonnie: I'll set the clock type to steady time. Based on past discussions it was unclear how to obtain the node clock time at the subscriber level.

rclcpp/include/rclcpp/node_impl.hpp
dabonnie: >I see now what you mean about the timers, since each subscription can set its statistics publish rate separately. I wonder though if that make sense now (connecting the publish rate with the subscription). Why not have one publisher and timer per node and have it publish all the statistics at once? What use case(s) supports the design decision to publish them separately and asynchronously? Do you know off hand how it was done in ROS 1? Was there one publisher that gathered all the stats and published them in bulk periodically or a separate publishing event for each contributor of stats (each subscription)?

ROS1: each [StatisticsLogger](https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/clients/roscpp/include/ros/statistics.h#L89) has a [publisher](https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/clients/roscpp/src/libros/statistics.cpp#L234), but publishes the data by checking the time using its [callback](https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/clients/roscpp/src/libros/statistics.cpp#L237) instead of a separate timer.  

A single timer does make sense, but now we need to rethink ownership. The node creates the publisher and the timer on behalf of the topic statistics instance, but now the node needs to own both the publisher and timer. This places more in node_impl (unless you have thoughts otherwise) which based on your previous comments seems undesirable. 

>An in between would be to have the Subscriptions register themselves with a "statistics publisher" class, which periodically queries all registered subscriptions and publishes their stats together, then you could have as many or few of those statistic publishers as you want, and have a default one in the node which is used if nothing else is specified. This is kind of how callback groups work, you can create one and pass it when making subscriptions, but if you don't specify one the default one in the node is used. This could also be extended more easily to incorporate more sources of stats, things other than subscriptions, in the future. They would just register themselves with a statistics publisher, collect stats asynchronously, and wait to be queried by the statistics publisher.

That makes sense and would be useful for publishers, as you noted. However, could we approach this as a separate improvement?

rclcpp/include/rclcpp/node_impl.hpp
dabonnie: Basically it comes down to the fact that we need a lot of things from the node level, create a publisher (seems we can use `rclcpp:create_publisher`?), the node name (ok, make a topic statistics parameter in the options), and also need to create a timer tied to the node. I don't see how that exists at the `create_subscription` level.

rclcpp/include/rclcpp/node_impl.hpp
wjwwood: > This places more in node_impl (unless you have thoughts otherwise) which based on your previous comments seems undesirable.

I could be owned by `NodeBase` or `NodeTopics`, right? No need for it to be in `Node`.

> However, could we approach this as a separate improvement?

Yes, but it would definitely break API in my opinion, so I assume it will have to be G-turtle and cannot be added as a non-breaking change.

wjwwood: > ROS1: each StatisticsLogger has a publisher,

One thing to note is that in ROS 1 those publishers were actually a singleton...

https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/clients/roscpp/src/libros/topic_manager.cpp#L324-L356

rclcpp/include/rclcpp/node_impl.hpp
wjwwood: You can definitely do it, but you will likely need to pass more interfaces into create_subscription, which makes total sense because the subscription is now pulling in all of these dependencies (creating a subscription now requires the interface to make timers because it now creates a timer...).

For the timer you can pass additionally the NodeTimerInterface and emulate this:

https://github.com/ros2/rclcpp/blob/649d72f835bfca334e7eab39a8c16752094d6317/rclcpp/include/rclcpp/node_impl.hpp#L104-L117

Ideally we would create a `rclcpp::create_*timer` function(s) to emulate how it works with publishers and subscriptions, and then use that. But in the meantime we'd have to copy what Node does above.

And for the publisher you can use `rclcpp::create_publisher`, as you suggested.

The node name can be accessed via the NodeBaseInterface, which can be accessed via the NodeTopicsInterface:

https://github.com/ros2/rclcpp/blob/649d72f835bfca334e7eab39a8c16752094d6317/rclcpp/include/rclcpp/node_interfaces/node_topics_interface.hpp#L79-L82

The other option would be to invert the dependencies, as I suggested, and instead register the Subscription with a statistics publisher which is created separately. This is close to what you have now, except instead of a separate stand alone class, you just have a publisher and a timer which are separately created for each subscription, so there's no chance for reuse between subscriptions.

wjwwood: So, to summarize, you can either add the necessary interfaces to the `rclcpp::create_subscription()` function or leave it mostly as-is and update lifecycle node to also do what `rclcpp::Node::create_subscription()` is doing.

rclcpp/include/rclcpp/subscription.hpp
wjwwood: This is another case of interface leaking. Now we've added a dependency on ROS Time from subscriptions, which we need to handle properly. Usually by either adding the clock interface to subscriptions (maybe problematic due to cycles) or by inverting the dependencies and injecting the clock dependency into the subscription some how. This could be done by passing a "get current time" std::function into the subscription which could be a lambda that calls now on the node clock, just as an example. That way the subscription can get the current ros time without requiring that interface.

wjwwood: We can continue to use steady time, if that make sense. But if you think system/ros time would make more sense, then there are solutions we just have to make sure we maintain the interfaces correctly.

rclcpp/include/rclcpp/node_impl.hpp
dabonnie: @wjwwood thanks for your help: in the latest commit I was able to move everything over to `create_subscription` and out of `node_impl`.

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -58,16 +64,49 @@ create_subscription(
...
+  rclcpp::node_interfaces::NodeTimersInterface::SharedPtr node_timer_interface = nullptr
 )
 {
   using rclcpp::node_interfaces::get_node_topics_interface;
   auto node_topics = get_node_topics_interface(std::forward<NodeT>(node));
 
+  std::shared_ptr<rclcpp::topic_statistics::SubscriptionTopicStatistics<CallbackMessageT>>
+  subscription_topic_stats = nullptr;
+
+  if (options.topic_stats_options.state == rclcpp::TopicStatisticsState::Enable) {
```
prajakta-gokhale, line 77: You'd need to use `rclcpp::detail::resolve_enable_topic_statistics()` here.

dabonnie:
CI (run according to [this](https://github.com/ros2/rclcpp/pull/1074#issuecomment-617268892) suggestion):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10368/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5884/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8444/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10265/)

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -43,7 +43,7 @@ class NodeBase : public NodeBaseInterface
     rclcpp::Context::SharedPtr context,
     const rcl_node_options_t & rcl_node_options,
     bool use_intra_process_default,
-    bool enable_topic_statistics_default);
+    bool enable_topic_statistics_default = false);
```
prajakta-gokhale, line 46: Why make this optional again?

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -58,16 +64,49 @@ create_subscription(
   ),
   typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (
     MessageMemoryStrategyT::create_default()
-  )
+  ),
+  rclcpp::node_interfaces::NodeTimersInterface::SharedPtr node_timer_interface = nullptr
```
prajakta-gokhale, line 68: Should there be validation for this value before it is used on line 99?

rclcpp/include/rclcpp/subscription.hpp
dabonnie: Ok. I think long term time that was obtained from the node's clock makes sense, but we had earlier design discussions (email) where we decided that this was a future improvement given the potential issues with dependency injection. 

As much as I'd like to try and integrate this now I believe we're running out of time (ugh, poor pun) unless you see an easy path forward (I'm currently digging now).

wjwwood: No we can do it better in the future, but we should try to follow up. Things like this tend to get lost after a big push like this.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
dabonnie: Think this is a rebase (merge conflicts) mistake. Will revert.

rclcpp/include/rclcpp/create_subscription.hpp
dabonnie: Yes, that's a good point. If the user wants to use topic statistics and the interface is null then we should throw. I'll add the throw and document.

wjwwood: This should be extracted from the `NodeT`. See line 72 for an example of another interface extracting it.

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -58,16 +64,49 @@ create_subscription(
...
+      >(node_topics->get_node_base_interface()->get_name(), publisher);
+
+    auto sub_call_back = [subscription_topic_stats]() {
+        subscription_topic_stats->publish_message();
+      };
+
+    auto timer = create_timer(
+      std::chrono::duration_cast<std::chrono::nanoseconds>(
+        options.topic_stats_options.
+        publish_period),
```
wjwwood, line 95:
```suggestion
        options.topic_stats_options.publish_period),
```

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -76,6 +115,34 @@ create_subscription(
...
+  rclcpp::node_interfaces::NodeBaseInterface * node_base_,
+  const rclcpp::node_interfaces::NodeTimersInterface::SharedPtr & node_timers_)
+{
+  auto timer = rclcpp::WallTimer<CallbackT>::make_shared(
+    nanos,
+    std::move(callback),
+    node_base_->get_context());
+  node_timers_->add_timer(timer, group);
+  return timer;
+}
```
wjwwood, line 144: This needs to be in its own header...

wjwwood: And should replace the code in `node_impl.hpp`.

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -43,6 +43,7 @@
 #include "rclcpp/qos.hpp"
 #include "rclcpp/type_support_decl.hpp"
 #include "rclcpp/visibility_control.hpp"
+#include "rclcpp/timer.hpp"
```
wjwwood, line 46: Alpha

rclcpp/include/rclcpp/create_subscription.hpp
wjwwood: You can also just copy the content of this function into the body of create_subscription, if you don't want to deal with all of that, but if we add this function publicly we should do it properly.

dabonnie: I'm happy to make a new `create_timer` header if with the same signature.

rclcpp/include/rclcpp/create_subscription.hpp
dabonnie: I had tried that with compile errors that NodeT is a `rclcpp::node_interfaces::NodeTopicsInterface`, I'll double check.

wjwwood: Being called from where? This one should be passing a `rclcpp::Node &`:

https://github.com/ros2/rclcpp/blob/0f0a4a8e397720ace85299eb30592e710b1b339b/rclcpp/include/rclcpp/node_impl.hpp#L96

dabonnie: I thought the same given what was being passed, but maybe I'm interpreting the below incorrectly.

   Doing this
   ```
   create_subscription(
  NodeT && node,
  const rclcpp::QoS & qos,
  CallbackT && callback,
  const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
    rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
  ),
  typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (
    MessageMemoryStrategyT::create_default()
  )
)
{
  using rclcpp::node_interfaces::get_node_timers_interface;
  auto node_timer_interface = get_node_timers_interface(std::forward<NodeT>(node));

  using rclcpp::node_interfaces::get_node_topics_interface;
  auto node_topics = get_node_topics_interface(std::forward<NodeT>(node));
  ```

  resulted in this

  ```
In file included from /home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/create_subscription.hpp:27:0,
                 from /home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/node_impl.hpp:40,
                 from /home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp:1167,
                 from /home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/src/rclcpp/time_source.cpp:27:
/home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/get_node_timers_interface.hpp: In instantiation of ‘rclcpp::node_interfaces::NodeTimersInterface* rclcpp::node_interfaces::detail::get_node_timers_interface_from_pointer(NodeType) [with NodeType = std::shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface>*; typename std::enable_if<std::is_same<NodeType, std::shared_ptr<typename std::remove_pointer<_Tp>::type::element_type>*>::value, int>::type <anonymous> = 0]’:
/home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/get_node_timers_interface.hpp:143:56:   required from ‘rclcpp::node_interfaces::NodeTimersInterface* rclcpp::node_interfaces::get_node_timers_interface(NodeType&&) [with NodeType = std::shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface>&; typename std::enable_if<(! std::is_pointer<typename std::remove_reference< <template-parameter-1-1> >::type>::value), int>::type <anonymous> = 0]’
/home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/create_subscription.hpp:75:56:   required from ‘std::shared_ptr<SubscriptionT> rclcpp::create_subscription(NodeT&&, const string&, const rclcpp::QoS&, CallbackT&&, const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>&, typename MessageMemoryStrategyT::SharedPtr) [with MessageT = rosgraph_msgs::msg::Clock_<std::allocator<void> >; CallbackT = std::_Bind<void (rclcpp::TimeSource::*(rclcpp::TimeSource*, std::_Placeholder<1>))(std::shared_ptr<rosgraph_msgs::msg::Clock_<std::allocator<void> > >)>; AllocatorT = std::allocator<void>; CallbackMessageT = rosgraph_msgs::msg::Clock_<std::allocator<void> >; SubscriptionT = rclcpp::Subscription<rosgraph_msgs::msg::Clock_<std::allocator<void> >, std::allocator<void> >; MessageMemoryStrategyT = rclcpp::message_memory_strategy::MessageMemoryStrategy<rosgraph_msgs::msg::Clock_<std::allocator<void> >, std::allocator<void> >; NodeT = std::shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface>&; std::__cxx11::string = std::__cxx11::basic_string<char>; typename MessageMemoryStrategyT::SharedPtr = std::shared_ptr<rclcpp::message_memory_strategy::MessageMemoryStrategy<rosgraph_msgs::msg::Clock_<std::allocator<void> >, std::allocator<void> > >]’
/home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/src/rclcpp/time_source.cpp:232:3:   required from here
/home/ubuntu/ros2_worksapce/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/get_node_timers_interface.hpp:115:48: error: no matching function for call to ‘get_node_timers_interface_from_pointer(std::__shared_ptr<rclcpp::node_interfaces::NodeTopicsInterface, (__gnu_cxx::_Lock_policy)2>::element_type*)’
   return get_node_timers_interface_from_pointer(node_shared_pointer->get());
   ```

rclcpp/include/rclcpp/create_subscription.hpp
dabonnie: Actually, there's already a `create_timer` header, so will use that and replace the code in `node_impl.hpp`.

rclcpp/include/rclcpp/create_subscription.hpp
wjwwood: It's a specific instance, `time_source.cpp`. Likely it is trying to create the subscription on the cheap (without holding more interfaces).

So, one way to fix this would be to make `NodeTopicsInterface` depend on `NodeTimersInterface`, like it does with `NodeBaseInterface` right now.

We seem to be retracing some of my steps (which I'm not remembering until we happen on them again, lol).

I think we need to change the `NodeTopicsInterface` to have a `get_node_timers_interface()` equivalent to this:

https://github.com/ros2/rclcpp/blob/0f0a4a8e397720ace85299eb30592e710b1b339b/rclcpp/include/rclcpp/node_interfaces/node_topics_interface.hpp#L79-L82

And pass it when creating the `NodeTopics` instance:

https://github.com/ros2/rclcpp/blob/0f0a4a8e397720ace85299eb30592e710b1b339b/rclcpp/include/rclcpp/node_interfaces/node_topics.hpp#L42

Then, anything that has a `NodeTopicsInterface` also has the `NodeTimersInterface` needed to do this work. I think this is the least invasive way to do it, and it's what I did when I needed the node base interface in the topic specific part of rclcpp.

Sorry for the epic journey, but I'm fairly confident in this last approach. I think there's only a few places `NodeTopics` is instantiated:

- https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node.cpp#L111
- https://github.com/ros2/rclcpp/blob/0f0a4a8e397720ace85299eb30592e710b1b339b/rclcpp_lifecycle/src/lifecycle_node.cpp#L70

And in both cases you have `NodeTimers` available right next to it.

ros-discourse:

https://discourse.ros.org/t/ros2-tooling-wg-next-meeting/12545/20

rclcpp/include/rclcpp/create_subscription.hpp
dabonnie: Thanks for the walkthrough, I was able to implement as you outlined.

rclcpp/include/rclcpp/create_subscription.hpp
dabonnie: Moved the definition to `create_timer.hpp`, need to replace the code in `node_impl.hpp`.

dabonnie: The new timer code is now used in `node_impl.hpp`.

dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10375/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5890/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8451/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10271/)

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -64,10 +74,46 @@ create_subscription(
   using rclcpp::node_interfaces::get_node_topics_interface;
   auto node_topics = get_node_topics_interface(std::forward<NodeT>(node));
 
+  auto node_timer_interface = node_topics->get_node_timers_interface();
```
wjwwood, line 77: nitpick: the scope of this variable can be reduced to only be within in the if statement's true block.

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -68,6 +70,46 @@ create_timer(
...
+ * \param group
+ * \param node_base
+ * \param node_timers
+ * \return
+ * \throws std::invalid argument if either node_base or node_timers
+ * are null
+ */
+template<typename DurationRepT, typename DurationT, typename CallbackT>
+typename rclcpp::WallTimer<CallbackT>::SharedPtr
+create_node_timer(
```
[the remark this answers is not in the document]
dabonnie: SGTM

rclcpp/include/rclcpp/create_subscription.hpp
dabonnie: Good point, will apply.

dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5899/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8468/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10282/)

wjwwood:
Since this is changing public API we need to run a wider CI than just up to `rclcpp`.

dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10385/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5900/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8469/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10283/)

dabonnie:
@wjwwood I think this is good to go?

wjwwood:
Yeah, let me annotate the CI with why it's ok to merge...

> Building everything and testing via `--packages-above rclcpp`:
>
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/10385/)
  * cmake warning is known assimp issue
  * compiler warnings are known cyclone issue (https://github.com/ros2/rmw_cyclonedds/pull/162)
  * tests failures appear on nightly
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5900/)
  * cmake warning is known assimp issue
  * compiler warnings are known cyclone issue (https://github.com/ros2/rmw_cyclonedds/pull/162)
> * macOS [Build Status](http://ci.ros2.org/job/ci_osx/8469/)
  * build is broken due to fast-rtps issue upstream
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/10283/)
  * compiler warnings are TF false positives
  * I did my best to categorize the issues on Windows, but there are 152 failing tests
    * most seem to be related to a cyclone test communication issue
    * others all seem to be flaky (timing related) and/or appear on nightly
    * I didn't see anything novel or that I think could be related to this

Karsten1987:
I get a `unused variable` warning on this:
```
[Processing: rclcpp]
--- stderr: rclcpp
/Users/karsten/workspace/ros2/ros2_master/src/ros2/rclcpp/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp:49:43: error: unused variable 'kTestStatsPublishPeriod' [-Werror,-Wunused-const-variable]
constexpr const std::chrono::milliseconds kTestStatsPublishPeriod{5000};
```

wjwwood:
Crap, I'm just not going to ever merge anything ever again, haha.

@dabonnie I know it's getting late, but can you have a look at fixing this warning asap?

dabonnie:
> Crap, I'm just not going to ever merge anything ever again, haha.
> 
> @dabonnie I know it's getting late, but can you have a look at fixing this warning asap?

Let me take a look.

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -64,10 +74,46 @@ create_subscription(
...
+
+  if (rclcpp::detail::resolve_enable_topic_statistics(
+      options,
+      *node_topics->get_node_base_interface()))
+  {
+    std::shared_ptr<Publisher<statistics_msgs::msg::MetricsMessage>> publisher =
+      create_publisher<statistics_msgs::msg::MetricsMessage>(
+      node,
+      options.topic_stats_options.publish_topic,
+      qos);
```
ivanpauno, line 88: qq: why does the topics statistics publisher use the same qos as the subscription being created?

I haven't checked the topics statistics demo, but it seems that if different entities are publishing statistics using different QoS, a variety of problems can happen (e.g.: pub/sub not matching).

dabonnie: That's a good question. It's possible we could provide publishing QoS options as part of the topic statistics options. What do you think is the best path forward?

ivanpauno: >  It's possible we could provide publishing QoS options as part of the topic statistics options. What do you think is the best path forward?

I think that adding an extra option for that (like `options.topic_stats_options.publish_topic`) is a good idea.

PR #1076, opened 2020-04-21
Dnae adas/intraprocess manager support for serialized message

DensoADAS:
added support for serialized messages in intraprocess manager, allowing on demand serialization/deserialization


DensoADAS:
depends on #1075

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -356,18 +358,37 @@ class IntraProcessManager
...
+      // same message type, so it can just be passed through
+      if (subscription_base->is_serialized() == is_serialized_publisher) {
+        auto subscription = std::static_pointer_cast<
+          rclcpp::experimental::SubscriptionIntraProcess<MessageT>
+          >(subscription_base);
+
+        subscription->provide_intra_process_message(message);
+      } else if (is_serialized_publisher) {
+        // publisher provides a serialized message, while subscriber expects a ROS message
+        provide_serialized_intra_process_message(subscription_base, *message);
```
ivanpauno, line 379: The name of the function is confusing to me.
IMO, it sounds like it's passing a serialized message along, but it's deserializing it and the providing the deserialized message to the intraprocess subscription.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -384,33 +405,71 @@ class IntraProcessManager
...
 
+  template<typename T>
+  static std::enable_if_t<!rclcpp::is_serialized_message_class<T>::value>
+  provide_serialized_intra_process_message(
+    rclcpp::experimental::SubscriptionIntraProcessBase::SharedPtr & subscription,
+    const T & serialized_message)
+  {
+    (void)subscription;
+    (void)serialized_message;
+    // prevent call if actual message type does not match
```
ivanpauno, line 462: could a `static_assert` be added here?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -134,6 +148,20 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
     return buffer_->use_take_shared_method();
   }
 
+  void
+  provide_serialized_intra_process_message(const SerializedMessage & serialized_message)
```
ivanpauno, line 152: maybe, `provide_ros_massage_from_serialized_message` would be clearer, though a bit verbose.
The opposite version: `provide_serialized_message_from_ros_message` could be added to avoid some duplication in intraprocess manager code.

ivanpauno: This method should only be available if the message type is a serialized message.
Probably, you can add a SFINAE check above to avoid it to be called in non-serialized message specializations.

ivanpauno: Though that would not satisfy the base class interface.
Better, if the message type is not serialized, the specialization should throw an error.

ivanpauno:
I talked offline with @Karsten1987 and I will explain a bit more about how I think this [optimization](https://github.com/ros2/rclcpp/pull/1076#pullrequestreview-397566868) should be done.

Instead of having only two vectors [here](https://github.com/ros2/rclcpp/blob/1a343416645a93fdc88bc5ad204e3cb27f80c9b4/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp#L329-L330), we could have four:
- One for subscriptions that take ownership (`std::unique_ptr`).
- One for subscriptions that don't take ownership. (`std::shared_ptr`).
- One for serialized subscriptions that take ownership. (`std::unique_ptr` to serialized message).
- One for serialized subscriptions that don't take ownership. (`std::shared_ptr` to serialized message).

Then the logic [here](https://github.com/ros2/rclcpp/blob/1a343416645a93fdc88bc5ad204e3cb27f80c9b4/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp#L264-L290) can be replaced with something like this:
- If the publisher is serialized and we have at least one subscription taking a non-serialized message, a serialized message `unique_ptr` is created.
- Then, we use the same logic we're using right now:
  - If there are not subscriptions taking a `unique_ptr`, the original non-serialized `unique_ptr` is promoted to a `shared_ptr` and added to all the buffers (https://github.com/ros2/rclcpp/blob/1a343416645a93fdc88bc5ad204e3cb27f80c9b4/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp#L266-L269).
  - If there is at least one subscription requiring an `unique_ptr`, the message of the original `unique_ptr` is [copied](https://github.com/ros2/rclcpp/blob/1a343416645a93fdc88bc5ad204e3cb27f80c9b4/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp#L275) in a `shared_ptr`. The `unique_ptr` is used for the [buffers requiring ownership of the message](https://github.com/ros2/rclcpp/blob/1a343416645a93fdc88bc5ad204e3cb27f80c9b4/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp#L282-L287), the `shared_ptr` [for the others](https://github.com/ros2/rclcpp/blob/1a343416645a93fdc88bc5ad204e3cb27f80c9b4/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp#L277-L281).

The same logic is then used for serialized subscriptions:
- If the publisher is non-serialized and we have at least one subscription taking a serialized message, a serialized message `unique_ptr` is created.
- Then, the same story about promoting the `unique_ptr` to a `shared_ptr` or not depending if a subscription taking a `unique_ptr` to a serialized message is required or not.

@DensoADAS I'm sorry for my super late review.
I think the optimization can be done in a follow-up, as it doesn't break public API.
It will break the intraprocess manager ABI, but that class isn't publicly available.
The only place where we use [it is being opaqued in a pointer](https://github.com/ros2/rclcpp/blob/911291f8d32da8a9a70f69adc7bcf1e41477e017/rclcpp/include/rclcpp/publisher.hpp#L123), so I think it'll be safe to backport the optimization.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
Karsten1987: I also don't think that these two static methods have to be introduced here. I think the `static_assert` could be part of the function above.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
Karsten1987: question: Why is that function called `_serialized_`? Could that not be part of the overload with `provide_intra_process_message(const SerializedMessage & serialized_message)`?

DensoADAS:
> t doesn't break public API.
> It will break the intraprocess manager ABI, but that class i

I have already thought about this. But in lack of time and this is more complicated I skipped this optimization for now. I will provide this later.

I also would add support for reusing deserialized messages (from subscriber side).

DensoADAS:
- updated to changes in serialized_messages
- rebased

Karsten1987:
thanks @DensoADAS  for iterating with us over this. It looks like there is still a conflict in this PR.

DensoADAS:
@Karsten1987 I merged it and updated it (also the other PR)

Karsten1987:
You might want to rebase once again as it looks like you've got some duplicated commits on your branch.

I've just checked and your PR compiles locally on my OSX machine :)
I'd be happy if you could address the review comments on the remaining two PRs in order to advance with these as soon as possible. thanks.

rclcpp/include/rclcpp/serialization.hpp
```diff
@@ -97,6 +97,34 @@ class Serialization : public SerializationBase
   }
 };
 
+/// Specialized serialization for rcl_serialized_message_t (because type support is not defined)
```
Karsten1987, line 100: I still don't quite understand why we need this specialization. Does the `static_assert` really introduce so much trouble?

DensoADAS: alternative: you can add enable_if around the functions where it's used

ivanpauno:
> I have already thought about this. But in lack of time and this is more complicated I skipped this optimization for now. I will provide this later.

Sounds good, thanks!

> I also would add support for reusing deserialized messages (from subscriber side).

Can you explain further? I don't get what you mean.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
ivanpauno: I agree with @Karsten1987 on this.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -134,6 +148,20 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
     return buffer_->use_take_shared_method();
   }
 
+  void
+  provide_intra_process_message(const SerializedMessage & serialized_message)
```
ivanpauno, line 152: IMO, there's a minimal asymmetry in the implementation, that makes this extremely confusing:

- If both subscription and publisher type matches, we use this other [overload](https://github.com/ros2/rclcpp/blob/0577f390242de78eeec1ae75fda9be7193995e47/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp#L139), and that's also true if both publisher/subscription are using a serialized message (the only difference is that the other signature uses an `unique_ptr`).
- If the publisher is serialized and the subscription isn't, this method is used, which provides a non-serialized message from a serialized one.
- If the publisher isn't serialized and the subscription is, the [original overload]([overload](https://github.com/ros2/rclcpp/blob/0577f390242de78eeec1ae75fda9be7193995e47/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp#L139)) is used, and serialization is handled in `add_owned_msg_to_buffers`/`add_shared_msg_to_buffers`

We should either have methods clearly named here (`provide_serialized_intra_process_message_from_ros_message` and `provide_ros_intra_process_message_from_serialized_message`), or always handle serialization/deserialization in `add_owned_msg_to_buffers`/`add_shared_msg_to_buffers`.

rclcpp/include/rclcpp/serialization.hpp
ivanpauno: I'm also not sure why this is needed.

DensoADAS:
> 
> 
> > I have already thought about this. But in lack of time and this is more complicated I skipped this optimization for now. I will provide this later.
> 
> Sounds good, thanks!
> 
> > I also would add support for reusing deserialized messages (from subscriber side).
> 
> Can you explain further? I don't get what you mean.

On publisher side the message get's serialized. On susbscriber side deserialized. Both operations only need to be done once for a message.

ivanpauno:
> On publisher side the message get's serialized. On susbscriber side deserialized. Both operations only need to be done once for a message.

Oh yeah, the comment I made above handles that correctly (serialization/deserialization is only done once if needed).

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
Karsten1987: In my opinion, it makes sense to handle the serialization in the `add_*msg_to_buffers` or some central place.

From what I understand, the `provide_intra_process_message` functions should simply be overloaded with `SerializedMessages` and shouldn't do any further logic than message passing - pretty much exactly how the other functions work. The serialization/deserialization should be handled outside of this, i.e. in the `add` functions.

Karsten1987:
> Isn't this PR required #1081 for this change to make sense?

I would try holding off from #1081 until these changes are settled. The changes introduced in the other PR are designed to be a follow up - and given time, I would assume these changes to be landed past Foxy.

ivanpauno:
AFAIS [here](https://github.com/ros2/rclcpp/blob/0577f390242de78eeec1ae75fda9be7193995e47/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp#L116), the intraprocess communication will happen if the subscription is taking a `SerializedMessage` object, and I don't see how that will happen without #1081 merged.

Karsten1987:
The typetrait will return correctly as well with `rcl_serialized_message_t`.
But I agree, before calling the callback, I believe that `get_rcl_serialized_message()` has to be called to get the rcl type.

ivanpauno:
> The typetrait will return correctly as well with ``.
> But I agree, before calling the callback, I believe that `get_rcl_serialized_message()` has to be called to get the rcl type.

That's true, but I think it is worth it to do that change now.
If not, the code for handling all the cases (including mismatchs between `rmw_serialized_message_t` and `SerializedMessage`) is going to be really tricky (more tricky than now).

It would also be great to add some testing in https://github.com/ros2/rclcpp/blob/0577f390242de78eeec1ae75fda9be7193995e47/rclcpp/test/test_intra_process_manager.cpp, to have some confidence that this actually works well :wink: .

Karsten1987:
> That's true, but I think it is worth it to do that change now.

Sounds good. I am on it ;-)

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
Karsten1987: I just played around a bit more with this and honestly, I don't know what the nicest solution for this is. 

I am generally in favor of handling the serialization in one place, ideally in the intra-process manager. In order to do that, we need a couple of functions in the subscription base:

This handle essentially returns the typesupport pointer which is required in order to run `rmw_serialize`.
```
const rosidl_type_support_t *
get_type_support_handle() const;
```
The implemenation of the intraprocess subscription can then do the SFINAE thing and either return the rosidl handle for `MessageT` if it's not a serialized message or return a `nullptr` otherwise.

The second message you'd need on the subscription base is to create/instantiate a default `MessageT` and return this as a `shared_ptr<void>` or `unique_ptr<void>` depending on the number of subscription.
```
std::shared_ptr<void> generate_message() const
{
  return std::make_shared<MessageT>();  // same for unique ptr
}
```
That returned pointer can then be filled by the serialization and passed back via `provide_intra_process_message`.

The `provide_intra_process_message` itself has to be overloaded with `shared_ptr<void>` and `unique_ptr<void>` whose implementation then casts these pointers back to the `MessageT` type and proceed as usual.
```
void provide_intra_process_message(const std::shared_ptr<void> message) {
  provide_intra_process_message(std::static_pointer_cast<const MessageT>(message));
}
```

That's quite a bit of flip-flop but doable. The alternatively, we could just invert the logic and do all the serialization inside the intraprocess subscription.

---

as for the SFINAE part, what I came up with which works nicely without too many empty implemented specializations is the following:

```
  const rosidl_message_type_support_t *
  get_type_support_handle() const override
  {
    return get_type_support_handle_impl<serialization_traits::is_serialized_message_class<MessageT>::value>();
  }

  template<bool is_serialized>
  const rosidl_message_type_support_t *
  get_type_support_handle_impl() const
  {
    return rosidl_typesupport_cpp::get_message_type_support_handle<MessageT>();
  }

  template<>
  const rosidl_message_type_support_t *
  get_type_support_handle_impl<true>() const
  {
    return nullptr;
  }
```

The serialization can then be conveniently be instantiated with:
```
rclcpp::SerializationBase serialization(get_type_support_handle());
```

DensoADAS:
@Karsten1987 I made changes regarding to your request, mainly:
* added method for getting type support (instead of specialization)
* added the optimization @ivanpauno  suggested (but I used a shared pointer to avoid the need to know the allocator on the publisher side)

Karsten1987:
@DensoADAS friendly ping

DensoADAS:
@Karsten1987 yeah, we are continuing on the PRs

DensoADAS:
@Karsten1987 I rebased the branch. Is anything else open? Perhaps I lost the overview a little by my spam filter.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -324,8 +301,39 @@ class IntraProcessManager
...
-    std::vector<uint64_t> take_ownership_subscriptions;
+    enum
+    {
+      IndexSharedTyped = 0, IndexSharedSerialized = 1,
+      IndexOwnershipTyped = 2, IndexOwnershipSerialized = 3,
+      IndexNum = 4
+    };
+
+    /// get the index for "take_subscriptions" depending on shared/serialized
+    constexpr static auto
```
Karsten1987, line 312:
```suggestion
    constexpr static size_t
```

I had to change it to `size_t` in order to make it work on OSX. Otherwise I got:
```
intra_process_manager.hpp:325:67: error: function 'get_index' with deduced return type cannot be used before it is defined
    constexpr static auto ownership_same = SplittedSubscriptions::get_index(
                                                                  ^
/Users/karsten/workspace/ros2/ros2_master/src/ros2/rclcpp/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp:313:5: note: 'get_index' declared here
    get_index(const bool is_shared, const bool is_serialized)
```

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -344,12 +352,162 @@ class IntraProcessManager
...
+    typename Alloc = std::allocator<void>,
+    typename Deleter = std::default_delete<MessageT>>
+  void
+  do_intra_process_publish_same_type(
+    std::unique_ptr<MessageT, Deleter> message,
+    std::shared_ptr<typename allocator::AllocRebind<MessageT, Alloc>::allocator_type> allocator,
+    const std::vector<uint64_t> & take_ownership_subscriptions,
+    const std::vector<uint64_t> & take_shared_subscriptions)
+  {
+    // subsriber and publisher have same message types
```
Karsten1987, line 374:
```suggestion
    // subscriber and publisher have same message types
```

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -344,12 +352,162 @@ class IntraProcessManager
...
+
+      this->template add_shared_msg_to_buffers<MessageT>(msg, take_shared_subscriptions);
+    } else if (!take_ownership_subscriptions.empty() && // NOLINT
+      take_shared_subscriptions.size() <= 1)
+    {
+      // There is at maximum 1 buffer that does not require ownership.
+      // So this case is equivalent to all the buffers requiring ownership
+
+      // Merge the two vectors of ids into a unique one
+      std::vector<uint64_t> concatenated_vector(take_shared_subscriptions);
```
Karsten1987, line 390: consider calling `reserve` here as the total amount of elements in the merged vector is known.would it also make sense to invert this here? Copying the `take_ownership_subscriptions` assuming that it's the bigger vector of the two.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -198,40 +203,22 @@ class IntraProcessManager
...
     {
-      // Construct a new shared pointer from the message
-      // for the buffers that do not require ownership
-      auto shared_msg = std::allocate_shared<MessageT, MessageAllocatorT>(*allocator, *message);
-
-      this->template add_shared_msg_to_buffers<MessageT>(
-        shared_msg, sub_ids.take_shared_subscriptions);
-      this->template add_owned_msg_to_buffers<MessageT, Alloc, Deleter>(
-        std::move(message), sub_ids.take_ownership_subscriptions, allocator);
+      do_intra_process_publish_other_type<MessageT>(
```
Karsten1987, line 210: should this return here? if not, shouldn't we check below as well that we have subscriptions which require "same type" to prevent doing always that work?

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -344,12 +352,162 @@ class IntraProcessManager
...
+  }
+
+  template<typename MessageT>
+  void
+  do_intra_process_publish_other_type(
+    const MessageT * unsupported_message,
+    const std::vector<uint64_t> & take_ownership_subscriptions,
+    const std::vector<uint64_t> & take_shared_subscriptions)
+  {
+    // subsriber and publisher have different message types
```
Karsten1987, line 421:
```suggestion
    // subscriber and publisher have different message types
```

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -344,12 +352,162 @@ class IntraProcessManager
...
+      take_shared_subscriptions.size() > 1)
+    {
+      // Construct a new shared pointer from the message
+      // for the buffers that do not require ownership
+      auto shared_msg = subscription_base->create_shared_message(message.get());
+
+      this->template add_shared_msg_to_buffers<void>(
+        shared_msg, take_shared_subscriptions);
+      add_owned_msg_to_buffers(std::move(message), take_ownership_subscriptions);
+    }
```
Karsten1987, line 464: Is there a way to deduplicate this logic?

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -411,6 +565,84 @@ class IntraProcessManager
...
+  /**
+   * Method to serialize a ros message to rclcpp::SerializedMessage.
+   * The publisher has a template type while the subscribers is serialized.
+   *
+   * \param message the ros message from publisher.
+   * \param subscription a subscriber used for creating the serialized message.
+   * \return a shared pointer (containing rclcpp::SerializedMessage) with deleter.
+   */
+  template<typename MessageT>
+  std::shared_ptr<void> convert_message(
```
Karsten1987, line 604: consider renaming this function to `serialize_message`.

ivanpauno:
@DensoADAS There seem to be conflicts again :confused:

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -115,10 +129,9 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
   }
 
   void
-  provide_intra_process_message(ConstMessageSharedPtr message)
+  provide_intra_process_message(std::shared_ptr<const void> message)
```
ivanpauno, line 132: why do you need `void` here?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -134,6 +147,24 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
     return buffer_->use_take_shared_method();
   }
 
+  std::shared_ptr<void>
+  create_shared_message(const void * message_to_copy)
```
ivanpauno, line 151: I would really avoid `void` usage if possible

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -324,8 +301,39 @@ class IntraProcessManager
...
+             (is_shared ? IndexSharedTyped : IndexOwnershipTyped);
+    }
+
+    std::vector<uint64_t> take_subscriptions[IndexNum];
+  };
+
+  template<bool is_serialized>
+  struct SplittedSubscriptionsIndicies
+  {
+    constexpr static auto ownership_same = SplittedSubscriptions::get_index(
```
ivanpauno, line 325: `owned_same` or `unique_same` sound better to me

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
ivanpauno: More precisely, this should return an `Index`.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -324,8 +301,39 @@ class IntraProcessManager
...
+      IndexOwnershipTyped = 2, IndexOwnershipSerialized = 3,
+      IndexNum = 4
+    };
+
+    /// get the index for "take_subscriptions" depending on shared/serialized
+    constexpr static auto
+    get_index(const bool is_shared, const bool is_serialized)
+    {
+      return (is_serialized ? IndexSharedTyped : IndexSharedSerialized) +
+             (is_shared ? IndexSharedTyped : IndexOwnershipTyped);
```
ivanpauno, line 316: please, avoid the sum.
It might look like a clever way to calculate the index, but it's extremely confusing.

ivanpauno: It also looks like the first line was wrong:
```suggestion
      return (is_serialized ? IndexSharedSerialized : IndexSharedTyped) +
             (is_shared ? IndexSharedTyped : IndexOwnershipTyped);
```

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -259,33 +247,22 @@ class IntraProcessManager
...
-          sub_ids.take_ownership_subscriptions,
-          allocator);
-      }
-
-      return shared_msg;
+    // check if (de)serialization is needed
+    if (sub_ids.take_subscriptions[Indicies::ownership_other].size() +
+      sub_ids.take_subscriptions[Indicies::shared_other].size() > 0)
+    {
+      do_intra_process_publish_other_type<MessageT>(
```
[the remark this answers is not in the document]
ivanpauno: I don't think so

rclcpp/src/rclcpp/intra_process_manager.cpp
```diff
@@ -141,9 +140,10 @@ IntraProcessManager::get_subscription_count(uint64_t intra_process_publisher_id)
     return 0;
   }
 
-  auto count =
-    publisher_it->second.take_shared_subscriptions.size() +
-    publisher_it->second.take_ownership_subscriptions.size();
+  auto count = 0u;
+  for (auto i = 0u; i < SplittedSubscriptions::IndexNum; ++i) {
```
ivanpauno, line 144: you can use [std::accumulate](https://en.cppreference.com/w/cpp/algorithm/accumulate) if you want, though there's no big benefit on using it.

rclcpp/test/rclcpp/test_intra_process_manager.cpp
```diff
@@ -23,6 +23,8 @@
 #define RCLCPP_BUILDING_LIBRARY 1
 #include "rclcpp/allocator/allocator_common.hpp"
 #include "rclcpp/macros.hpp"
+#include "rclcpp/serialization.hpp"
```
ivanpauno, line 26: It would be great to add tests verifying that the intraprocess manager is working well with intraprocess messages.

PR #1077, opened 2020-04-21
Dnae adas/generic subscriber and publisher

DensoADAS:
using SerializedMessage and support in intraprocess manager to create a generic subscriber / publisher without previously known message type


DensoADAS:
@Karsten1987 I did not create a spezialzation of Publisher for SerializedMessages as I disklike copying so much redudant code. I redesigned it so the changes to the Publisher are reduced. I hope it does not break to much with the current api.

DensoADAS:
depends on #1077

ivanpauno:
A few questions before reviewing, I'm probably missing some context from https://github.com/ros2/rclcpp/pull/973 discussion:

- what's the goal of the "generic" publisher/subscription?
- how they differ from `Publisher`/`Subscription`?

rclcpp/include/rclcpp/create_publisher.hpp
```diff
@@ -29,6 +29,46 @@
...
+template<
+  typename MessageT,
+  typename AllocatorT = std::allocator<void>,
+  typename PublisherT = rclcpp::Publisher<MessageT, AllocatorT>,
+  typename NodeT>
+std::shared_ptr<PublisherT>
+create_generic_publisher(
+  NodeT & node,
+  const std::string & topic_name,
+  const rosidl_message_type_support_t & type_support,
```
ivanpauno, line 46: Why both `MessageT` template parameter and the `type_support` argument are needed?

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -29,6 +29,53 @@
...
+  typename CallbackMessageT =
+  typename rclcpp::subscription_traits::has_message_type<CallbackT>::type,
+  typename SubscriptionT = rclcpp::Subscription<CallbackMessageT, AllocatorT>,
+  typename MessageMemoryStrategyT = rclcpp::message_memory_strategy::MessageMemoryStrategy<
+    CallbackMessageT,
+    AllocatorT
+  >,
+  typename NodeT>
+typename std::shared_ptr<SubscriptionT>
+create_generic_subscription(
```
ivanpauno, line 51: same about this

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -259,24 +274,10 @@ class Publisher : public PublisherBase
...
-        if (nullptr != context && !rcl_context_is_valid(context)) {
-          // publisher is invalid due to context being shutdown
-          return;
-        }
-      }
-    }
-    if (RCL_RET_OK != status) {
-      rclcpp::exceptions::throw_from_rcl_error(status, "failed to publish message");
-    }
+    generic_publisher::do_inter_process_publish(publisher_handle_, msg);
```
ivanpauno, line 280: I don't understand why the two new free functions are needed, and why a "normal" `Publisher` is calling the functions in `generic_publisher` namespace.
@DensoADAS can you expain?

rclcpp/include/rclcpp/publisher_factory.hpp
```diff
@@ -60,6 +60,35 @@ struct PublisherFactory
   const PublisherFactoryFunction create_typed_publisher;
 };
 
+/// Return a PublisherFactory with functions setup for creating a PublisherT<MessageT, AllocatorT>.
+template<typename MessageT, typename AllocatorT, typename PublisherT>
```
ivanpauno, line 64: Is `MessageT` being used?

rclcpp/include/rclcpp/subscription_factory.hpp
```diff
@@ -117,6 +117,62 @@ create_subscription_factory(
   // return the factory now that it is populated
   return factory;
 }
+/// Return a SubscriptionFactory setup to create a SubscriptionT<MessageT, AllocatorT>.
+template<
+  typename MessageT,
```
ivanpauno, line 122: Same about `MessageT`

rclcpp/include/rclcpp/create_publisher.hpp
ivanpauno: Naming should also be more clear.
Maybe, just an override of `create_publisher`?

rclcpp/include/rclcpp/create_publisher.hpp
```diff
@@ -29,6 +29,46 @@
...
+ * The NodeT type only needs to have a method called get_node_topics_interface()
+ * which returns a shared_ptr to a NodeTopicsInterface.
+ */
+template<
+  typename MessageT,
+  typename AllocatorT = std::allocator<void>,
+  typename PublisherT = rclcpp::Publisher<MessageT, AllocatorT>,
+  typename NodeT>
+std::shared_ptr<PublisherT>
+create_generic_publisher(
```
Karsten1987, line 43: given that the return type is just a regular publisher, I would advise against renaming the function to `generic` and have it part of the overloads as originally designed.

rclcpp/include/rclcpp/create_subscription.hpp
Karsten1987, line 51: same comment as above for the `create_publisher` function.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -32,6 +32,7 @@
 #include "rclcpp/allocator/allocator_deleter.hpp"
 #include "rclcpp/detail/resolve_use_intra_process.hpp"
 #include "rclcpp/experimental/intra_process_manager.hpp"
+#include "rclcpp/generic_publisher.hpp"
```
Karsten1987, line 35: I am not sure if the two template functions are worth putting them into a separate file. Also, given that there are only functions, I don't think they should live in a `generic_publisher` namespace.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -259,24 +274,10 @@ class Publisher : public PublisherBase
   }
 
 protected:
-  void
+  virtual void
```
Karsten1987, line 277: is that change necessary?

DensoADAS: no, sorry slipped in

DensoADAS:
> 
> 
> A few questions before reviewing, I'm probably missing some context from #973 discussion:
> 
>     * what's the goal of the "generic" publisher/subscription?
> 
>     * how they differ from `Publisher`/`Subscription`?
> 
> 
> Thanks!

The "generic" publisher/subscriber allows communication without knowing the message type to compile time. Therefore a serialized message can bei published or received and the endpoint has to handle the content itself. This is used for example in rosbag2.

This implementation supports now intraprocess communication when using the C++ version of serialized messages which make the memory handling more safe.

rclcpp/include/rclcpp/publisher.hpp
DensoADAS: for the SerializedMessage the original "do_inter_process_publish" will fail as it's no ROS message

ivanpauno: can we just are an override for a serialized message in the `Publisher` class?

ivanpauno:
> The "generic" publisher/subscriber allows communication without knowing the message type to compile time. Therefore a serialized message can bei published or received and the endpoint has to handle the content itself. This is used for example in rosbag2.

Ah ok. So in those cases, you use `MessageT=SerializedMessageBase`, and you don't need to know the message type at compile time, right?

rclcpp/include/rclcpp/create_publisher.hpp
DensoADAS: I think MessageT is used for the allocator

Karsten1987:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10380/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5895/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8463/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10277/)

DensoADAS:
Switching back to create_generic_... would solve this. Do you have another suggestion?

DensoADAS:
@Karsten1987 I lost a little bit the track of the conversation. What's the current status?

Karsten1987:
In my opinion, https://github.com/ros2/rclcpp/pull/1076#discussion_r413487908 this is where we stand. I think the intraprocess manager should be sorted out before pushing forward with this change.

Karsten1987:
Thanks @DensoADAS for iterating over this PR.

Unfortunately, we've passed API freeze for Foxy and after talking with @jacobperron (the Foxy lead), we've decided to table this for G-Turtle.

nyxaria:
Is a generic subscriber available in Foxy? I don't think I can use ROS2 until it is in. :/

jacobperron:
> Is a generic subscriber available in Foxy? 

Not in rclcpp. Unless something like this PR can be done in an ABI compatible way, I don't think we can expect it to land in Foxy.

[As mentioned on ROS Answers](https://answers.ros.org/question/359681/generic-subscriber-in-ros2/), you can do this in Python and/or copy the C++ implementation from rosbag2 for your application.

PR #1080, opened 2020-04-21
Add received message age metric to topic statistics

prajakta-gokhale:
Waiting on #1072 for unit testing. 

Update: This includes part of changes from #1072 for unit tests. Once #1072 gets merged this will be rebased for CI.

Signed-off-by: Prajakta Gokhale <prajaktg@amazon.com>

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -619,10 +619,6 @@ if(BUILD_TESTING)
     target_link_libraries(test_wait_set ${PROJECT_NAME})
   endif()
 
-  set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
-  if(WIN32)
-    set(append_library_dirs "${append_library_dirs}/$<CONFIG>")
-  endif()
   ament_add_gtest(test_subscription_topic_statistics test/topic_statistics/test_subscription_topic_statistics.cpp
     APPEND_LIBRARY_DIRS "${append_library_dirs}"
   )
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -229,6 +229,7 @@ if(BUILD_TESTING)
     test/msg/MessageWithHeader.msg
     DEPENDENCIES builtin_interfaces
     LIBRARY_NAME ${PROJECT_NAME}
+    SKIP_INSTALL
   )
 
   ament_add_gtest(test_client test/test_client.cpp)
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -217,6 +217,7 @@ if(BUILD_TESTING)
 
   find_package(rmw_implementation_cmake REQUIRED)
 
+  find_package(sensor_msgs REQUIRED)
   find_package(test_msgs REQUIRED)
 
   include(cmake/rclcpp_add_build_failure_test.cmake)
@@ -614,11 +615,13 @@ if(BUILD_TESTING)
   if(TARGET test_subscription_topic_statistics)
     ament_target_dependencies(test_subscription_topic_statistics
       "libstatistics_collector"
+      "rcl"
       "rcl_interfaces"
       "rcutils"
       "rmw"
       "rosidl_runtime_cpp"
       "rosidl_typesupport_cpp"
+      "sensor_msgs"
       "statistics_msgs"
       "test_msgs")
     target_link_libraries(test_subscription_topic_statistics ${PROJECT_NAME})
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -12,6 +12,7 @@ find_package(rcpputils REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rosgraph_msgs REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
 find_package(rosidl_runtime_cpp REQUIRED)
 find_package(rosidl_typesupport_c REQUIRED)
 find_package(rosidl_typesupport_cpp REQUIRED)
@@ -222,6 +223,13 @@ if(BUILD_TESTING)
   include(cmake/rclcpp_add_build_failure_test.cmake)
 
   add_definitions(-DTEST_RESOURCES_DIRECTORY="${CMAKE_CURRENT_BINARY_DIR}/test/resources")
+
+  rosidl_generate_interfaces(${PROJECT_NAME}_test_msgs
+    test/msg/Header.msg
+    test/msg/MessageWithHeader.msg
+    DEPENDENCIES builtin_interfaces
+    LIBRARY_NAME ${PROJECT_NAME}
+  )
 
   ament_add_gtest(test_client test/test_client.cpp)
   if(TARGET test_client)
@@ -613,6 +621,7 @@ if(BUILD_TESTING)
   ament_add_gtest(test_subscription_topic_statistics test/topic_statistics/test_subscription_topic_statistics.cpp)
   if(TARGET test_subscription_topic_statistics)
     ament_target_dependencies(test_subscription_topic_statistics
+      "builtin_interfaces"
       "libstatistics_collector"
       "rcl_interfaces"
       "rcutils"
@@ -621,8 +630,10 @@ if(BUILD_TESTING)
       "rosidl_typesupport_cpp"
       "statistics_msgs"
       "test_msgs")
+    rosidl_target_interfaces(test_subscription_topic_statistics ${PROJECT_NAME}_test_msgs "rosidl_typesupport_cpp")
     target_link_libraries(test_subscription_topic_statistics ${PROJECT_NAME})
   endif()
+
 
   ament_add_gtest(test_subscription_options test/test_subscription_options.cpp)
   if(TARGET test_subscription_options)
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -619,8 +619,12 @@ if(BUILD_TESTING)
     target_link_libraries(test_wait_set ${PROJECT_NAME})
   endif()
 
+  set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
+  if(WIN32)
+    set(append_library_dirs "${append_library_dirs}/$<CONFIG>")
+  endif()
   ament_add_gtest(test_subscription_topic_statistics test/topic_statistics/test_subscription_topic_statistics.cpp
-    APPEND_LIBRARY_DIRS "${CMAKE_CURRENT_BINARY_DIR}"
+    APPEND_LIBRARY_DIRS "${append_library_dirs}"
   )
   if(TARGET test_subscription_topic_statistics)
     ament_target_dependencies(test_subscription_topic_statistics
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -634,7 +634,6 @@ if(BUILD_TESTING)
     target_link_libraries(test_subscription_topic_statistics ${PROJECT_NAME})
   endif()
 
-
   ament_add_gtest(test_subscription_options test/test_subscription_options.cpp)
   if(TARGET test_subscription_options)
     ament_target_dependencies(test_subscription_options "rcl")
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -217,7 +217,6 @@ if(BUILD_TESTING)
 
   find_package(rmw_implementation_cmake REQUIRED)
 
-  find_package(sensor_msgs REQUIRED)
   find_package(test_msgs REQUIRED)
 
   include(cmake/rclcpp_add_build_failure_test.cmake)
@@ -615,13 +614,11 @@ if(BUILD_TESTING)
   if(TARGET test_subscription_topic_statistics)
     ament_target_dependencies(test_subscription_topic_statistics
       "libstatistics_collector"
-      "rcl"
       "rcl_interfaces"
       "rcutils"
       "rmw"
       "rosidl_runtime_cpp"
       "rosidl_typesupport_cpp"
-      "sensor_msgs"
       "statistics_msgs"
       "test_msgs")
     target_link_libraries(test_subscription_topic_statistics ${PROJECT_NAME})
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -619,7 +619,9 @@ if(BUILD_TESTING)
     target_link_libraries(test_wait_set ${PROJECT_NAME})
   endif()
 
-  ament_add_gtest(test_subscription_topic_statistics test/topic_statistics/test_subscription_topic_statistics.cpp)
+  ament_add_gtest(test_subscription_topic_statistics test/topic_statistics/test_subscription_topic_statistics.cpp
+    APPEND_LIBRARY_DIRS "${CMAKE_CURRENT_BINARY_DIR}"
+  )
   if(TARGET test_subscription_topic_statistics)
     ament_target_dependencies(test_subscription_topic_statistics
       "builtin_interfaces"
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -271,8 +271,8 @@ public:
 
     if (subscription_topic_statistics_) {
       const auto nanos = std::chrono::time_point_cast<std::chrono::nanoseconds>(
-        std::chrono::steady_clock::now());
-      const auto time = rclcpp::Time(nanos.time_since_epoch().count(), RCL_STEADY_TIME);
+        std::chrono::system_clock::now());
+      const auto time = rclcpp::Time(nanos.time_since_epoch().count());
       subscription_topic_statistics_->handle_message(*typed_message, time);
     }
   }
--- a/rclcpp/package.xml
+++ b/rclcpp/package.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="3">
+<package format="2">
   <name>rclcpp</name>
   <version>0.8.3</version>
   <description>The ROS client library in C++.</description>
@@ -39,8 +39,6 @@
   <test_depend>rmw_implementation_cmake</test_depend>
   <test_depend>rosidl_default_generators</test_depend>
   <test_depend>test_msgs</test_depend>
-
-  <member_of_group>rosidl_interface_packages</member_of_group>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/rclcpp/package.xml
+++ b/rclcpp/package.xml
@@ -37,7 +37,6 @@
   <test_depend>ament_lint_common</test_depend>
   <test_depend>rmw</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
-  <test_depend>sensor_msgs</test_depend>
   <test_depend>test_msgs</test_depend>
 
   <export>
--- a/rclcpp/package.xml
+++ b/rclcpp/package.xml
@@ -37,6 +37,7 @@
   <test_depend>ament_lint_common</test_depend>
   <test_depend>rmw</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
+  <test_depend>sensor_msgs</test_depend>
   <test_depend>test_msgs</test_depend>
 
   <export>
--- a/rclcpp/package.xml
+++ b/rclcpp/package.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<package format="3">
   <name>rclcpp</name>
   <version>0.8.3</version>
   <description>The ROS client library in C++.</description>
@@ -37,7 +37,10 @@
   <test_depend>ament_lint_common</test_depend>
   <test_depend>rmw</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
+  <test_depend>rosidl_default_generators</test_depend>
   <test_depend>test_msgs</test_depend>
+
+  <member_of_group>rosidl_interface_packages</member_of_group>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
+++ b/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
@@ -25,6 +25,7 @@
 #include "libstatistics_collector/moving_average_statistics/types.hpp"
 
 #include "rclcpp/create_publisher.hpp"
+#include "rclcpp/msg/message_with_header.hpp"
 #include "rclcpp/node.hpp"
 #include "rclcpp/qos.hpp"
 #include "rclcpp/rclcpp.hpp"
@@ -36,7 +37,6 @@
 #include "statistics_msgs/msg/statistic_data_point.hpp"
 #include "statistics_msgs/msg/statistic_data_type.hpp"
 
-#include "test_msgs/msg/message_with_header.hpp"
 #include "test_msgs/msg/empty.hpp"
 
 #include "test_topic_stats_utils.hpp"
@@ -51,7 +51,7 @@ constexpr const uint64_t kNoSamples{0};
 constexpr const std::chrono::seconds kTestDuration{10};
 }  // namespace
 
-using test_msgs::msg::MessageWithHeader;
+using rclcpp::msg::MessageWithHeader;
 using test_msgs::msg::Empty;
 using rclcpp::topic_statistics::SubscriptionTopicStatistics;
 using statistics_msgs::msg::MetricsMessage;
--- a/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
+++ b/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
@@ -102,13 +102,11 @@ public:
 private:
   void publish_message()
   {
-    ++number_published_;
     auto msg = Empty{};
     publisher_->publish(msg);
   }
 
   rclcpp::Publisher<Empty>::SharedPtr publisher_;
-  std::atomic<size_t> number_published_{0};
   rclcpp::TimerBase::SharedPtr publish_timer_;
 };
 
@@ -135,7 +133,6 @@ public:
 private:
   void publish_message()
   {
-    ++number_published_;
     auto msg = MessageWithHeader{};
     auto now = this->now();
     auto nanos = now.nanoseconds() - 1000 * 1000;
@@ -145,7 +142,6 @@ private:
   }
 
   rclcpp::Publisher<MessageWithHeader>::SharedPtr publisher_;
-  std::atomic<size_t> number_published_{0};
   rclcpp::TimerBase::SharedPtr publish_timer_;
 };
 
@@ -302,8 +298,6 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_stats_for_message_no
   // Message age statistics will not be calculated because Empty messages
   // don't have a `header` with timestamp.
 
-  // TODO(prajakta-gokhale): Change Empty message type to something with a `header`,
-  // and have below assertions work for all collectors.
   for (const auto & msg : received_messages) {
     if (msg.metrics_source == "message_period") {
       for (const auto & stats_point : msg.statistics) {
--- a/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
+++ b/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
@@ -25,7 +25,6 @@
 #include "libstatistics_collector/moving_average_statistics/types.hpp"
 
 #include "rclcpp/create_publisher.hpp"
-#include "rclcpp/clock.hpp"
 #include "rclcpp/node.hpp"
 #include "rclcpp/qos.hpp"
 #include "rclcpp/rclcpp.hpp"
@@ -138,11 +137,13 @@ private:
   {
     ++number_published_;
     auto msg = Imu{};
-    msg.header.stamp = clock_.now();
+    auto now = this->now();
+    auto nanos = now.nanoseconds() - 1000*1000;
+    msg.header.stamp.sec = static_cast<std::int32_t>(RCL_NS_TO_S(nanos));
+    msg.header.stamp.nanosec = static_cast<std::uint32_t>(nanos % (1000 * 1000 * 1000));
     publisher_->publish(msg);
   }
 
-  rclcpp::Clock clock_;
   rclcpp::Publisher<Imu>::SharedPtr publisher_;
   std::atomic<size_t> number_published_{0};
   rclcpp::TimerBase::SharedPtr publish_timer_;
@@ -386,16 +387,16 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_stats_for_message_wi
           EXPECT_LT(0, stats_point.data) << "unexpected sample count";
           break;
         case StatisticDataType::STATISTICS_DATA_TYPE_AVERAGE:
-          EXPECT_NE(std::nan(""), stats_point.data) << "unexpected avg";
+          EXPECT_LT(0, stats_point.data) << "unexpected avg";
           break;
         case StatisticDataType::STATISTICS_DATA_TYPE_MINIMUM:
-          EXPECT_NE(std::nan(""), stats_point.data) << "unexpected min";
+          EXPECT_LT(0, stats_point.data) << "unexpected min";
           break;
         case StatisticDataType::STATISTICS_DATA_TYPE_MAXIMUM:
-          EXPECT_NE(std::nan(""), stats_point.data) << "unexpected max";
+          EXPECT_LT(0, stats_point.data) << "unexpected max";
           break;
         case StatisticDataType::STATISTICS_DATA_TYPE_STDDEV:
-          EXPECT_NE(std::nan(""), stats_point.data) << "unexpected stddev";
+          EXPECT_LE(0, stats_point.data) << "unexpected stddev";
           break;
         default:
           FAIL() << "received unknown statistics type: " << std::dec <<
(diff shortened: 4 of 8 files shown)


prajakta-gokhale:
FYI @wjwwood this is the last PR for topic statistics functionality.

prajakta-gokhale:
@wjwwood @dabonnie please take a look.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -138,3 +185,66 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_manual_construction)
...
+    kTestDuration);
+
+  // compare message counts, sample count should be the same as published and received count
+  EXPECT_EQ(2, statistics_listener->GetNumberOfMessagesReceived());
+
+  // check the received message and the data types
+  const auto received_messages = statistics_listener->GetReceivedMessages();
+
+  EXPECT_EQ(2u, received_messages.size());
+  EXPECT_EQ("message_age", received_messages[0].metrics_source);
```
dabonnie, line 221: How is the order of the messages guaranteed?

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -138,3 +185,66 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_manual_construction)
...
+  ex.add_node(statistics_listener);
+  ex.add_node(empty_subscriber);
+
+  // spin and get future
+  ex.spin_until_future_complete(
+    statistics_listener->GetFuture(),
+    kTestDuration);
+
+  // compare message counts, sample count should be the same as published and received count
+  EXPECT_EQ(2, statistics_listener->GetNumberOfMessagesReceived());
```
dabonnie, line 215: Can you please comment / document how we'll test a message that has a timestamp in the header?

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
prajakta-gokhale: Changed this to make it order-agnostic.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
prajakta-gokhale: Added a `TODO` below.

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -138,3 +186,74 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_manual_construction)
...
+    received_metrics.insert(msg.metrics_source);
+  }
+  EXPECT_TRUE(received_metrics.find("message_age") != received_metrics.end());
+  EXPECT_TRUE(received_metrics.find("message_period") != received_metrics.end());
+
+  // Check the collected statistics for message period.
+  // Message age statistics will not be calculated because Empty messages
+  // don't have a `header` with timestamp.
+
+  // TODO(prajakta-gokhale): Change EMpty message type to something with a `header`,
```
dabonnie, line 234: Can you add a new test to use a message with a header and leave this test? I think it's good that we test both message types, i.e., those that have a header timestamp and those that don't.

wjwwood:
This needs rebase after https://github.com/ros2/rclcpp/pull/1072

prajakta-gokhale:
> This needs rebase after #1072

Rebase done. Please take a look at the diff now, it should be easier to review.

prajakta-gokhale:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10411/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5918/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8491/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10300/)

prajakta-gokhale:
@wjwwood I am in the process of adding a test dependency on `sensor_msgs` in order to use `sensor_msgs::msg::Imu` message in this unit test, so we can assert on correct message age calculation also. This is necessary because the message age metric only gets calculated when the message has a `header` value in it.

What are your thoughts on that?

wjwwood:
I would instead recommend you extend `test_msgs` with a new message to accomplish this.

dabonnie:
> I would instead recommend you extend `test_msgs` with a new message to accomplish this.

We need to test using the `std_msgs` `Header`: so we'll be adding a dependency on that package. Is that ok as a full dependency for `test_msgs` or is it easier to add a test dependency here?

jacobperron:
I would prefer a separate test message. We have to be careful adding a dependency to `std_msgs` in `test_msgs` though; this would create a circular dependency at the repository level. `test_msgs` (rcl_interfaces) -> `std_msgs` (common_interfaces) -> `builtin_interfaces` (rcl_interfaces).

jacobperron:
Do does the test message have to contain the type `std_msgs/msg/Header`, or can it just contain `builtin_interfaces/msg/Time`?

dabonnie:
> Does the test message have to contain the type `std_msgs/msg/Header`, or can it just contain `builtin_interfaces/msg/Time`?

The header is required. See how we extract a message's timestamp [here](https://github.com/ros-tooling/libstatistics_collector/blob/master/include/libstatistics_collector/topic_statistics_collector/received_message_age.hpp#L71).

FYI @jacobperron @wjwwood : https://github.com/ros2/rclcpp/pull/1080#issuecomment-618143080. I believe @prajakta-gokhale will provide a patch here that can use the message proposed (have verified locally).

wjwwood:
True, if it has to be a `std_msgs/msg/Header` then we may need to depend on some other message package, though that's not ideal.

prajakta-gokhale:
> this would create a circular dependency at the repository level. `test_msgs` (rcl_interfaces) -> `std_msgs` (common_interfaces) -> `builtin_interfaces` (rcl_interfaces).

This isn't a circular dependency between packages, if I'm not wrong? Will it be a problem if it's circular dependncy between repositories?

dabonnie:
> I would prefer a separate test message. We have to be careful adding a dependency to `std_msgs` in `test_msgs` though; this would create a circular dependency at the repository level. `test_msgs` (rcl_interfaces) -> `std_msgs` (common_interfaces) -> `builtin_interfaces` (rcl_interfaces).

Given the circular dependency I'm not seeing a path forward, especially since we were under the assumption that a user who desires to timestamp their message would use the `std_msgs/msg/Header`. Any thoughts here?

wjwwood:
> Will it be a problem if it's circular dependncy between repositories?

Honestly, it's technically feasible, but it's not a nice situation to have, and some folks are very particular on that point that we don't have repository level dependency cycles. So yes it is a problem.

jacobperron:
> This isn't a circular dependency between packages, if I'm not wrong? Will it be a problem if it's circular dependncy between repositories?

In particular, it gets annoying when it comes to releasing the packages, since we typically make releases per repository.

jacobperron:
@dabonnie @prajakta-gokhale If I understand the code correctly, I think we don't necessarily need to use `std_msgs/msg/Header`. We just need a message that contains a field named `header` that itself contains a field of type `builtin_interfaces/msg/Time`.

prajakta-gokhale:
Agreed, it's not at all preferable to have a circular dependency of any kind. 

> We just need a message that contains a field named header that itself contains a field of type builtin_interfaces/msg/Time

Yes, it could work with that.

jacobperron:
I think maybe the best path forward is to generate a test message in `rclcpp`, where we can add a test dependency on `std_msgs`.

prajakta-gokhale:
> I think maybe the best path forward is to generate a test message in `rclcpp`, where we can add a test dependency on `std_msgs`.

I went down that path a bit this morning. I got a build error when I added `<member_of_group>rosidl_interface_packages</member_of_group>` in rclcpp/package.xml.

```bash
Error(s):
- The manifest of package "rclcpp" (with format version 2) must not contain the following tags: member_of_group
```

jacobperron:
> I went down that path a bit this morning. I got a build error when I added <member_of_group>rosidl_interface_packages</member_of_group> in rclcpp/package.xml

I think that error could be resolved with changing the package.xml format to `3`, **but** I also don't think we need to use the `<member_of_group>` tag at all since we don't expect other packages outside `rclcpp` to depend on the message.

prajakta-gokhale:
Here's what can work as well:

- Add a new message in `test_msgs` called `Header` that has a `builtin_interfaces/Time` field
- Add another new message in `test_msgs` called TestHeader that has a field `Header`
- Use `test_msgs::msg::TestHeader` in this PR

prajakta-gokhale:
Honestly, if it is possible to easily add a new message in rclcpp/test I'd prefer that. That's how we're testing the message age collection feature in [libstatistics_collector](https://github.com/ros-tooling/libstatistics_collector/blob/master/test/msg/DummyMessage.msg).

jacobperron:
> Honestly, if it is possible to easily add a new message in rclcpp/test I'd prefer that. That's how we're testing the message age collection feature in libstatistics_collector.

I prefer this approach :+1:

prajakta-gokhale:
Okay I'll have a revision with that soon. Thanks for your help @jacobperron and @wjwwood !

prajakta-gokhale:
We implemented a `DummyMessage` in `rclcpp/test` similar to [libstatistics_collector](https://github.com/ros-tooling/libstatistics_collector), but it isn't usable in tests and throws this error:

```
Failed to find library 'rclcpp__rosidl_typesupport_fastrtps_cpp'
unknown file: Failure
C++ exception with description "could not create publisher: type support not from this implementation, at /../../ros2_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/publisher.cpp:75, at /../../ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:173" thrown in the test body.
```

Is implementing the messages in test_msgs a viable alternative to implementing them in rclcpp? I think the problem may be because of some other setting in rclcpp/CMakeLists.txt.

prajakta-gokhale:
I moved the new messages to `test_msgs` (w/o introducing any new dependencies) and that seems to be working locally. I'll open a PR in `test_msgs` and add the corresponding changes here.

prajakta-gokhale:
Opened https://github.com/ros2/rcl_interfaces/pull/101 for new messages required in tests here.

prajakta-gokhale:
CI for `rclcpp` and `test_msgs` together:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10424/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5930/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8505/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10317/)

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -271,8 +271,8 @@ class Subscription : public SubscriptionBase
 
     if (subscription_topic_statistics_) {
       const auto nanos = std::chrono::time_point_cast<std::chrono::nanoseconds>(
-        std::chrono::steady_clock::now());
-      const auto time = rclcpp::Time(nanos.time_since_epoch().count(), RCL_STEADY_TIME);
+        std::chrono::system_clock::now());
```
jacobperron, line 274: IIUC, this change is because we're assuming that the timestamp in message headers is using system time, right?

What happens if a node is using simulation time (ie. ROS_TIME)?

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -217,31 +280,122 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_single_empty_stats_m
...
+
+  std::set<std::string> received_metrics;
+  for (const auto & msg : received_messages) {
+    received_metrics.insert(msg.metrics_source);
+  }
+  EXPECT_TRUE(received_metrics.find("message_age") != received_metrics.end());
+  EXPECT_TRUE(received_metrics.find("message_period") != received_metrics.end());
+
+  // Check the collected statistics for message period.
+
```
jacobperron, line 375: nit: remove extra vertical space

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -217,31 +280,122 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_single_empty_stats_m
...
+  }
+  EXPECT_TRUE(received_metrics.find("message_age") != received_metrics.end());
+  EXPECT_TRUE(received_metrics.find("message_period") != received_metrics.end());
+
+  // Check the collected statistics for message period.
+  // Message age statistics will not be calculated because Empty messages
+  // don't have a `header` with timestamp.
+
+  for (const auto & msg : received_messages) {
+    if (msg.metrics_source == "message_period") {
```
jacobperron, line 302: nit: we can avoid extra indentation by skipping the iteration early:

```c++
if (msg.metrics_source != "message_period") {
  continue;
}
for (const auto & stats_point : msg.statistics) {
...
```

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -217,31 +280,122 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_single_empty_stats_m
...
+  for (const auto & msg : received_messages) {
+    received_metrics.insert(msg.metrics_source);
+  }
+  EXPECT_TRUE(received_metrics.find("message_age") != received_metrics.end());
+  EXPECT_TRUE(received_metrics.find("message_period") != received_metrics.end());
+
+  // Check the collected statistics for message period.
+  // Message age statistics will not be calculated because Empty messages
+  // don't have a `header` with timestamp.
+
```
jacobperron, line 300: nit : remove extra vertical space

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -97,21 +99,49 @@ class EmptyPublisher : public rclcpp::Node
...
+  }
+
+  rclcpp::Publisher<Empty>::SharedPtr publisher_;
+  rclcpp::TimerBase::SharedPtr publish_timer_;
+};
+
+/**
+ * MessageWithHeader publisher node: used to publish MessageWithHeader with `header` value set
+ */
+class DummyPublisher : public rclcpp::Node
```
jacobperron, line 116: I would renaming the new publisher/subscription nodes to something more informative, e.g. containing the name of the message: `MessageWithHeaderPublisher` and `MessageWithHeaderSubscriber`

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -97,21 +99,49 @@ class EmptyPublisher : public rclcpp::Node
...
 private:
   void publish_message()
   {
-    ++number_published_;
-    auto msg = Empty{};
+    auto msg = MessageWithHeader{};
+    auto now = this->now();
+    auto nanos = now.nanoseconds() - 1000 * 1000;
+    msg.header.stamp.sec = static_cast<std::int32_t>(RCL_NS_TO_S(nanos));
+    msg.header.stamp.nanosec = static_cast<std::uint32_t>(nanos % (1000 * 1000 * 1000));
```
jacobperron, line 140: I think we can avoid the manual conversion and use the assignment operator directly:

```c++
msg.header.stamp = this->now();
```

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -147,6 +177,38 @@ class EmptySubscriber : public rclcpp::Node
...
+public:
+  DummySubscriber(const std::string & name, const std::string & topic)
+  : Node(name)
+  {
+    // manually enable topic statistics via options
+    auto options = rclcpp::SubscriptionOptions();
+    options.topic_stats_options.state = rclcpp::TopicStatisticsState::Enable;
+
+    auto callback = [this](MessageWithHeader::UniquePtr msg) {
+        this->receive_message(*msg);
```
jacobperron, line 194: I guess we can remove the method `receive_message` and just pass an empty lambda.

rclcpp/include/rclcpp/subscription.hpp
dabonnie: That is correct. There is also no way to specify a message's time source when we were assuming one would use the std_msgs/Header (I brought this up with @wjwwood and @tfoote during the message review).

jacobperron: Okay, so it sounds like a known limitation of topic statistics for the moment; not compatible with simulation.

jacobperron:
> I implemented a DummyMessage in rclcpp/test similar to libstatistics_collector, but it isn't usable in tests and throws this error:
>
>        Failed to find library 'rclcpp__rosidl_typesupport_fastrtps_cpp'
>        unknown file: Failure
>        C++ exception with description "could not create publisher: type support not from this imp

@prajakta-gokhale @dabonnie It looks like the issue we had here was due to a limitation of `rosidl_generate_interfaces`. In order for it to infer the correct names of the generated rmw libraries the target name should be the same as the project name. I.e.

```cmake
rosidl_generate_interfaces(${PROJECT_NAME}  # <-- must be ${PROJECT_NAME}
```

So we don't collide with the target name for the library, I've opened https://github.com/ros2/rclcpp/pull/1093, which renames the library. After #1093 is merged you can rebase and move the message generation code back into rclcpp. I think adding a test dependency directly on `std_msgs` should be okay if you prefer to use `std_msgs/msg/Header` directly in the tests.

Thanks to @dirk-thomas for finding a solution :tada:

prajakta-gokhale:
@jacobperron thanks for diving into it and finding a solution! @dabonnie and I will address your comments and move the dummy messages back here. At this point I think we're okay without the added dependency on `std_msgs`, by using a fake `Header` message. 

I wonder how `rosidl_generate_interfaces` works in [libstatistics_collector](https://github.com/ros-tooling/libstatistics_collector/blob/master/CMakeLists.txt#L94-L99) where we're naming the msg library different from actual generated library name.

jacobperron:
> I wonder how rosidl_generate_interfaces works in libstatistics_collector where we're naming the msg library different from actual generated library name.

I'm not sure, maybe it is a difference in usage. In libstatistics_collector, the test does not instantiate a publisher :man_shrugging:

dabonnie:
> > I wonder how rosidl_generate_interfaces works in libstatistics_collector where we're naming the msg library different from actual generated library name.
> 
> I'm not sure, maybe it is a difference in usage. In libstatistics_collector, the test does not instantiate a publisher 🤷‍♂️

We do something similar in the [ros-tooling/system_metrics_collector](https://github.com/ros-tooling/system_metrics_collector):

 - [DummyMessage](https://github.com/ros-tooling/system_metrics_collector/tree/master/system_metrics_collector/test/msg)
 - [TestPublisher](https://github.com/ros-tooling/system_metrics_collector/blob/master/system_metrics_collector/test/topic_statistics_collector/test_subscriber_topic_statistics.cpp)
 - [Cmake](https://github.com/ros-tooling/system_metrics_collector/blob/master/system_metrics_collector/CMakeLists.txt#L43)

rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -97,21 +99,47 @@ class EmptyPublisher : public rclcpp::Node
...
+  virtual ~MessageWithHeaderPublisher() = default;
+
 private:
   void publish_message()
   {
-    ++number_published_;
-    auto msg = Empty{};
+    auto msg = MessageWithHeader{};
+    // Subtract 1 sec from current time so the received message age is always > 0
+    msg.header.stamp = this->now() - rclcpp::Duration{1, 0};
```
dabonnie, line 138: Much cleaner!

dabonnie:
@jacobperron I believe this is ready for review while we wait for #1093 to be merged.

jacobperron:
> We do something similar in the ros-tooling/system_metrics_collector:

Ah, I think I see now! There is already a way to resolve this with rosidl_generate_interfaces, with the [LIBRARY_NAME](https://github.com/ros2/rosidl/blob/b311072c4b141be17ce67c11a5fcecf86bf39abb/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L39-L40) argument. :upside_down_face: 

Instead, I think you can just update this PR to do something like:
```
  rosidl_generate_interfaces(${PROJECT_NAME}_interfaces
    test/msg/MessageWithHeader.msg
    DEPENDENCIES buildtin_interfaces
    LIBRARY_NAME ${PROJECT_NAME}  # <-- what we were missing
    SKIP_INSTALL
  )
```

I'm going to close #1093.

prajakta-gokhale:
> Ah, I think I see now! There is already a way to resolve this with rosidl_generate_interfaces, with the [LIBRARY_NAME](https://github.com/ros2/rosidl/blob/b311072c4b141be17ce67c11a5fcecf86bf39abb/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L39-L40) argument.

I think I tried that as well, and it had a different error (which I can't remember now). I can give it a shot again. If that doesn't work we can merge #1093 and rebase this to use the same library name.

jacobperron:
> I think I tried that as well, and it had a different error (which I can't remember now). I can give it a shot again. If that doesn't work we can merge #1093 and rebase this to use the same library name

Sounds good to me, Thanks for iterating!

jacobperron:
@prajakta-gokhale FWIW, I've pushed a branch to confirm that I can generate and use a custom message in a test with a publisher: https://github.com/ros2/rclcpp/blob/37fefec0a5a4c9e1a27de10626d2076c9f6282d1/rclcpp/CMakeLists.txt#L228-L244

rclcpp/include/rclcpp/subscription.hpp
wjwwood: The type in `std_msgs::msg::Header` can be ROS time... Both ROS Time and System time share an epoch, the unix epoch. It's only when you want to use a different Epoch or behavior for the clock that the stamp in header is insufficient, e.g. like steady time or TAI time.

So in this situation I don't  understand why you cannot use ROS Time...

jacobperron: I guess we could ROS time, and assume the incoming messages are using the same epoch as this node.

At the end of the day we can't know if the time in the header will have the same epoch as the time from this node, right? Since one of them may being use simulation time and the other not.

prajakta-gokhale:
@jacobperron the latest revision adds the message types we need here.

It differs a little from your version because I got a compile error if I didn't add `<member_of_group>`
```bash
CMake Error at /.../ros2_ws/install/rosidl_cmake/share/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake:229 (message):
  Packages installing interfaces must include
  '<member_of_group>rosidl_interface_packages</member_of_group>' in their
  package.xml
Call Stack (most recent call first):
  CMakeLists.txt:632 (rosidl_generate_interfaces)
```

And to do that I had to change package.xml version to from 2 to 3.
```bash
Error(s):
- The manifest of package "rclcpp" (with format version 2) must not contain the following tags: member_of_group
```

rclcpp/CMakeLists.txt
```diff
@@ -221,6 +222,13 @@ if(BUILD_TESTING)
 
   add_definitions(-DTEST_RESOURCES_DIRECTORY="${CMAKE_CURRENT_BINARY_DIR}/test/resources")
 
+  rosidl_generate_interfaces(${PROJECT_NAME}_test_msgs
+    test/msg/Header.msg
+    test/msg/MessageWithHeader.msg
+    DEPENDENCIES builtin_interfaces
+    LIBRARY_NAME ${PROJECT_NAME}
```
jacobperron, line 229: Since we're only generating the messages for the tests in this package, we don't need to install them. We can add `SKIP_INSTALL`.
```suggestion
    LIBRARY_NAME ${PROJECT_NAME}
    SKIP_INSTALL
```

This should also eliminate the need to add the `<member_of_group>` tag.

rclcpp/CMakeLists.txt
```diff
@@ -618,9 +627,11 @@ if(BUILD_TESTING)
       "rosidl_typesupport_cpp"
       "statistics_msgs"
       "test_msgs")
+    rosidl_target_interfaces(test_subscription_topic_statistics ${PROJECT_NAME}_test_msgs "rosidl_typesupport_cpp")
     target_link_libraries(test_subscription_topic_statistics ${PROJECT_NAME})
   endif()
 
+
```
jacobperron, line 634: nit: the extra blank line isn't necessary.

rclcpp/package.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<package format="3">
```
jacobperron, line 3: If we skip the installation of generated messages, we can revert this change and remove the `<member_of_group>` tag below.

rclcpp/include/rclcpp/subscription.hpp
wjwwood: Simulated time uses the same epoch as system time, at least that's my understanding. If you are simulating time and then set the time to 1 (as in 1 nanosecond) then it's 1 nanosecond after jan 1 1970.

At least that's how I interrpret it:

- this is the rcl time point:
  - https://github.com/ros2/rcl/blob/643f6ec5300a24b8f9076019232f98b8d6aa76b6/rcl/include/rcl/time.h#L43-L44
- which the rclcpp::Time is based on:
  - https://github.com/ros2/rclcpp/blob/df3c2ffa8a9d0b1b6f5cf707986accbc091575f9/rclcpp/include/rclcpp/time.hpp#L129

Perhaps it should be clarified, but I think ROS time and system time should be usable interchangeably.

rclcpp/CMakeLists.txt
prajakta-gokhale: When I add `SKIP_INSTALL` I get the same error as before:

```bash
Failed to find library 'rclcpp__rosidl_typesupport_fastrtps_cpp'
unknown file: Failure
C++ exception with description "could not create publisher: type support not from this implementation, at /Users/prajaktg/ros2_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/publisher.cpp:75, at /Users/prajaktg/ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:173" thrown in the test body.
```

prajakta-gokhale: @dabonnie can you verify that with the latest changes on this branch? Trying to see if it's just my workspace doing that.

jacobperron: I can reproduce the same issue with your branch.. I'm not sure why skipping installation causes the issue :thinking:

jacobperron: In any case, I don't think it's a blocker. We could figure it out later and patch it after.

jacobperron:
Testing `rclcpp` and `test_rclcpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10466/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5953/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10351/)

rclcpp/CMakeLists.txt
dabonnie: @prajakta-gokhale yes, I was able to build with all of @jacobperron's suggestions. Will push shortly.

dabonnie: @jacobperron here's my diff, are we missing something?

```
diff --git a/rclcpp/CMakeLists.txt b/rclcpp/CMakeLists.txt
index d57289c..68b8f5a 100644
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -227,6 +227,7 @@ if(BUILD_TESTING)
     test/msg/MessageWithHeader.msg
     DEPENDENCIES builtin_interfaces
     LIBRARY_NAME ${PROJECT_NAME}
+    SKIP_INSTALL
   )
 
   ament_add_gtest(test_client test/test_client.cpp)
diff --git a/rclcpp/package.xml b/rclcpp/package.xml
index 9c4b6e2..4e7f15b 100644
--- a/rclcpp/package.xml
+++ b/rclcpp/package.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="3">
+<package format="2">
   <name>rclcpp</name>
   <version>0.8.3</version>
   <description>The ROS client library in C++.</description>
@@ -40,8 +40,6 @@
   <test_depend>rosidl_default_generators</test_depend>
   <test_depend>test_msgs</test_depend>
 
-  <member_of_group>rosidl_interface_packages</member_of_group>
-
   <export>
     <build_type>ament_cmake</build_type>
   </export>
```

jacobperron: Ah, since we're not installing the package, we need to point the test executable to the right place.

Modifying the CMakeList.txt like so resolves it for me:

```diff
- ament_add_gtest(test_subscription_topic_statistics test/topic_statistics/test_subscription_topic_statistics.cpp)
+ ament_add_gtest(test_subscription_topic_statistics test/topic_statistics/test_subscription_topic_statistics.cpp
+   APPEND_LIBRARY_DIRS "${CMAKE_CURRENT_BINARY_DIR}"
+ )

**Comment by jacobperron** [comment]
We have some known flaky build failures (https://github.com/ros2/build_farmer/issues/273), so I'll just keep retrying. I'll wait for my latest comments to be addressed though.

File: rclcpp/CMakeLists.txt (hunk 12 above)
Comment (dabonnie) [reply rclcpp/CMakeLists.txt:229]: @jacobperron still results in the same failure with a clean build :-(

**Comment by jacobperron** [comment]
Testing rclcpp with Fast-RTPS only to see if the most recent changes are okay:

Linux: [Build Status](https://ci.ros2.org/job/ci_linux/10470/)

**Comment by jacobperron** [comment]
Testing `rclcpp` and `test_rclcpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10473/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5958/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8541/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10354/)

**Comment by jacobperron** [comment]
I think all except one of the failures is unrelated.
On Windows, we see the familiar test failure:

> C++ exception with description "could not create publisher: type support not from this implementation, at C:\ci\ws\src\ros2\rmw_fastrtps\rmw_fastrtps_cpp\src\publisher.cpp:75, at C:\ci\ws\src\ros2\rcl\rcl\src\rcl\publisher.c:173" 

https://ci.ros2.org/job/ci_windows/10354/testReport/rclcpp/TestSubscriptionTopicStatisticsFixture/test_receive_stats_for_message_with_header/

I guess what we pass to `APPEND_LIBRARY_DIRS` should be changed if we are Windows ([for example](https://github.com/ros2/system_tests/blob/f9dd335393b32ee9985037cd37cd80cb084bf79a/test_communication/CMakeLists.txt#L464-L467)).

**Comment by dabonnie** [comment]
> I think all except one of the failures is unrelated.
> On Windows, we see the familiar test failure:
> 
> > C++ exception with description "could not create publisher: type support not from this implementation, at C:\ci\ws\src\ros2\rmw_fastrtps\rmw_fastrtps_cpp\src\publisher.cpp:75, at C:\ci\ws\src\ros2\rcl\rcl\src\rcl\publisher.c:173"
> 
> https://ci.ros2.org/job/ci_windows/10354/testReport/rclcpp/TestSubscriptionTopicStatisticsFixture/test_receive_stats_for_message_with_header/
> 
> I guess what we pass to `APPEND_LIBRARY_DIRS` should be changed if we are Windows ([for example](https://github.com/ros2/system_tests/blob/f9dd335393b32ee9985037cd37cd80cb084bf79a/test_communication/CMakeLists.txt#L464-L467)).

Ok I'll give that a shot.

**Comment by prajakta-gokhale** [comment]
> I guess what we pass to `APPEND_LIBRARY_DIRS` should be changed if we are Windows ([for example](https://github.com/ros2/system_tests/blob/f9dd335393b32ee9985037cd37cd80cb084bf79a/test_communication/CMakeLists.txt#L464-L467)).

Yes, that'll also make it consistent with [other tests](https://github.com/ros2/rclcpp/blob/master/rclcpp/CMakeLists.txt#L510-L513) in this package.

File: rclcpp/CMakeLists.txt [hunk 15]
```diff
@@ -619,8 +619,12 @@ if(BUILD_TESTING)
     target_link_libraries(test_wait_set ${PROJECT_NAME})
   endif()
 
+  set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
+  if(WIN32)
+    set(append_library_dirs "${append_library_dirs}/$<CONFIG>")
+  endif()
```
prajakta-gokhale, line 625: This is already set [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/CMakeLists.txt#L510-L513). The test only needs to use it.

jacobperron: :+1:  I missed that the first time around.

dabonnie: My mistake as well, fixing.

jacobperron:
Windows up to rclcpp: [Build Status](https://ci.ros2.org/job/ci_windows/10357/)

hidmic:
@prajakta-gokhale FYI these tests fail when being run against RTI Connext

```
[ RUN      ] TestSubscriptionTopicStatisticsFixture.test_receive_stats_for_message_no_header
Expected: (0) < (stats_point.data), actual: 0 vs nan
unexpected avg
Expected: (0) < (stats_point.data), actual: 0 vs nan
unexpected max
Expected: (0) < (stats_point.data), actual: 0 vs nan
unexpected min
Expected: (0) < (stats_point.data), actual: 0 vs 0
unexpected sample count
Expected: (0) < (stats_point.data), actual: 0 vs nan
```

dabonnie:
> @prajakta-gokhale FYI these tests fail when being run against RTI Connext:
> 
> ```
> [ RUN      ] TestSubscriptionTopicStatisticsFixture.test_receive_stats_for_message_no_header
> Expected: (0) < (stats_point.data), actual: 0 vs nan
> unexpected avg
> Expected: (0) < (stats_point.data), actual: 0 vs nan
> unexpected max
> Expected: (0) < (stats_point.data), actual: 0 vs nan
> unexpected min
> Expected: (0) < (stats_point.data), actual: 0 vs 0
> unexpected sample count
> Expected: (0) < (stats_point.data), actual: 0 vs nan
> ```

@hidmic do you happen to have any more details, e.g., is the output from a job on [Jenkins](https://ci.ros2.org/)? or observed elsewhere?

dirk-thomas:
> these tests fail when being run against RTI Connext:

I created a separate ticket for the test failure since otherwise this will get lost (as it has been for the past two months): see #1273.

PR #1083, opened 2020-04-22
deprecate redundant namespaces

wjwwood:
I collected most of the API breaking changes I came across while working on the executor changes. This pull request primarily deprecates old namespaces that we've been wanting to get rid of for a while.

It also moves some classes into their own files and renames one class (with deprecation).

Connected pull requests:

- https://github.com/ros2/demos/pull/442
- https://github.com/ros2/examples/pull/267
- https://github.com/ros2/realtime_support/pull/92
- https://github.com/ros2/system_tests/pull/426


rclcpp/include/rclcpp/exceptions.hpp
```diff
@@ -15,265 +15,6 @@
...
-/// Thrown if parameter is modified while in a set callback.
-class ParameterModifiedInCallbackException : public std::runtime_error
-{
-  // Inherit constructors from runtime_error.
-  using std::runtime_error::runtime_error;
-};
-
-}  // namespace exceptions
-}  // namespace rclcpp
+#include "rclcpp/exceptions/exceptions.hpp"
```
sloretz, line 18: why keep the `exceptions` namespace?

rclcpp/include/rclcpp/future_return_code.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+/// Stream operator for FutureReturnCode.
+RCLCPP_PUBLIC
+std::ostream &
+operator<<(std::ostream & os, const FutureReturnCode & future_return_code);
+
+/// String conversion function for FutureReturnCode.
+RCLCPP_PUBLIC
+std::string
+to_string(const FutureReturnCode & future_return_code);
```
sloretz, line 43: does there need to be a deprecated signature for `rclcpp::executor::to_string` as well?

rclcpp/include/rclcpp/future_return_code.hpp
```diff
@@ -0,0 +1,53 @@
...
+ *          This does not indicate that the operation succeeded; "get" may still throw an exception.
+ * INTERRUPTED: The future is not complete, spinning was interrupted by Ctrl-C or another error.
+ * TIMEOUT: Spinning timed out.
+ */
+enum class FutureReturnCode {SUCCESS, INTERRUPTED, TIMEOUT};
+
+/// Stream operator for FutureReturnCode.
+RCLCPP_PUBLIC
+std::ostream &
+operator<<(std::ostream & os, const FutureReturnCode & future_return_code);
```
sloretz, line 38: Does there need to be a deprecated signature for `rclcpp::executor::operator<<` as well?

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
```diff
@@ -17,6 +17,8 @@
 
 #include <memory>
 
+#include "rosidl_runtime_cpp/traits.hpp"
```
sloretz, line 20: Seems unrelated; is this addition intentional?

rclcpp/include/rclcpp/executor_options.hpp
```diff
@@ -0,0 +1,49 @@
...
+  rclcpp::Context::SharedPtr context;
+  size_t max_conditions;
+};
+
+namespace executor
+{
+
+using ExecutorArgs [[deprecated("use rclcpp::ExecutorOptions instead")]] = ExecutorOptions;
+
+}  // namespace executor
```
sloretz, line 46: Need a deprecation for `rclcpp::executor::create_default_executor_arguments()`?

rclcpp_components/src/component_manager.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "ament_index_cpp/get_resource.hpp"
 #include "class_loader/class_loader.hpp"
+#include "rclcpp/rclcpp.hpp"
```
sloretz, line 25: Recommend using more specific includes. What is used that wasn't brought in by `component_manager.hpp`?

rclcpp/include/rclcpp/executor_options.hpp
```diff
@@ -0,0 +1,49 @@
+// Copyright 2014 Open Source Robotics Foundation, Inc.
```
jacobperron, line 1:
```suggestion
// Copyright 2014-2020 Open Source Robotics Foundation, Inc.
```

rclcpp/include/rclcpp/experimental/executable_list.hpp
```diff
@@ -27,7 +27,7 @@
 
 namespace rclcpp
 {
-namespace executor
+namespace experimental
```
jacobperron, line 30: Why `experimental` versus `detail`?

rclcpp/include/rclcpp/exceptions.hpp
wjwwood: Changing it everywhere was very invasive, and I think in the future we could have each exception in its own file, so this would become something like:

```c++
#include "rclcpp/exceptions/some_exception.hpp"
#include "rclcpp/exceptions/some_other_exception.hpp"
#include "rclcpp/exceptions/throw_from_rcl_error.hpp"
```

Basically I think the exceptions namespace has some value where as the others were pretty redundant, so I just "fixed" it by moving them into a folder so the namespace made sense.

rclcpp/include/rclcpp/future_return_code.hpp
wjwwood: I didn't think so because I thought, like the stream operator, it would be used in a way that https://en.cppreference.com/w/cpp/language/adl would make it "just work", but I can setup an inline forwarding function. It's annoying because we cannot deprecate a using declaration, and we could use the namespace deprecation feature, but it's C++17 only...

rclcpp/include/rclcpp/future_return_code.hpp
wjwwood: I think https://en.cppreference.com/w/cpp/language/adl handles this one.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
wjwwood: Ah, yeah I just noticed that this header would not compile by itself without this include, so basically just a missing include. It's not strictly related, but a worthwhile change, imo.

rclcpp/include/rclcpp/executor_options.hpp
wjwwood: I forgot to mention this in the issue description, thanks for pointing it out.

I ended up just removing it (not moving it) because there's no reason not to just use the `ExecutorOptions` default constructor instead. It also caused the header to not compile without warnings because it was a static function that had no uses in the header.

I audited our code at least, and didn't find it being used in many places, so I would like to just remove it, though I can add a deprecation for it if you like.

sloretz: I agree it's pointless and probably unused. I would recommend adding a deprecation just for the sake of saying we followed this process consistently.

rclcpp/include/rclcpp/experimental/executable_list.hpp
wjwwood: Because I think it's completely redundant with new features related to the new wait set. I'd like to remove it completely, but haven't had time to do that yet.

I chose not to put it in detail, because I'm not sure about it (expect it might change) or not sure it will stick around. Whereas something in detail, for me, we don't expect it to change or be removed.

rclcpp_components/src/component_manager.cpp
wjwwood: I can drop this one, I can't remember what it was. I don't think something wasn't compiling, it was just more of a "include what you use" kind of thing.

rclcpp/include/rclcpp/executor_options.hpp
wjwwood: Ok, will do.

rclcpp/include/rclcpp/executor_options.hpp
wjwwood: 0548061

rclcpp/include/rclcpp/executor_options.hpp
wjwwood: 0548061

rclcpp/include/rclcpp/future_return_code.hpp
wjwwood: 0548061

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10415/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5921/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8493/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10303/)

wjwwood:
CI after fixing another rebase:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10417/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5923/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8495/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10306/)

wjwwood:
CI with just Fast-RTPS, but otherwise full:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10449/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5942/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8527/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10333/)

PR #1085, opened 2020-04-22
Update comment about return value in Executor::get_next_ready_executable

christophebedard:
Goes back to when `get_next_ready_executable()` was changed to return a `bool` instead of a `SharedPtr` here: https://github.com/ros2/rclcpp/commit/1610fc3973541d7064d2592ba5555ee99b45e128#diff-97193a162153796658f8e485ff3f9d5aR563


hidmic:
Since this only shortens a comment line, I'll proceed without CI.

PR #1094, opened 2020-04-24
Enforce a precedence for wildcard matching in parameter overrides.

### LINKED ISSUE #953: Odd parameter overrides' ordering when using wildcards
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Binaries
- Version or commit hash:
  - Eloquent (mostly 0.8.1)
  - rclcpp

#### Steps to reproduce issue

1. Write a parameter YAML file, `explicit_params.yaml`,  setting `use_sim_time` explicitly referring to a node:

    ```yaml
    talker:
        ros__parameters:
            use_sim_time: false    
    ```

2. Write a parameter YAML file, `wildcard_params.yaml`, setting `use_sim_time` implicitly using wildcards:

    ```yaml
    /**:
        ros__parameters:
            use_sim_time: true    
    ``` 

3. Run a `talker` node:

   - Passing the parameter YAML file using wildcards first:
     
     ```sh
     ros2 run demo_nodes_cpp talker --ros-args --params-file path/to/wildcard_params.yaml --params-file path/to/explicit_params.yaml
     ```
   - Passing the parameter YAML file being explicit first:

     ```sh
     ros2 run demo_nodes_cpp talker --ros-args --params-file path/to/explicit_params.yaml --params-file path/to/wildcard_params.yaml
     ```

#### Expected behavior
I'm not sure if we've ever fully specified behavior when using wildcards. 

- If wildcards have lower precedence than fully qualified names, command line order should not matter i.e. `use_sim_time` always `false`.

- If command line order is respected, `use_sim_time` should be `false` in the first case, `true` in the second case.

#### Actual behavior

First given override wins.

#### Additional information

This is inconsistent with how it works for `rclpy` nodes, where command line order is respected. Looking at both [`rclpy`](https://github.com/ros2/rclpy/pull/370/files#diff-80dda03110b5be606b823d6b0558b5c2R4543-R4554) and [`rclcpp`](https://github.com/ros2/rclcpp/pull/762/files#diff-43c1c52da4c73f0a25534075d66ece63R148-R156) implementations, it looks behavior emerged by pure chance. We should probably be matching wildcards where we can enforce a consistent precedence, down in [`rcl`](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/arguments.c).

hidmic:
Fixes #953.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -114,7 +114,7 @@ NodeParameters::NodeParameters(
       for (const auto & node_name : node_matching_names) {
         if (initial_map.count(node_name) > 0) {
           // Combine parameter yaml files, overwriting values in older ones
-          for (auto & param : initial_map.at(node_name)) {
+          for (const rclcpp::Parameter & param : initial_map.at(node_name)) {
             parameter_overrides_[param.get_name()] =
               rclcpp::ParameterValue(param.get_value_message());
           }
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -107,11 +107,15 @@ NodeParameters::NodeParameters(
           rcl_yaml_node_struct_fini(params);
         });
       rclcpp::ParameterMap initial_map = rclcpp::parameter_map_from(params);
-      for (auto iter = initial_map.begin(); initial_map.end() != iter; iter++) {
-        // TODO(cottsay) implement further wildcard matching
-        if (iter->first == "/**" || iter->first == combined_name_) {
+
+      // Enforce wildcard matching precedence
+      // TODO(cottsay) implement further wildcard matching
+      const char * node_matching_names[] = {"/**", combined_name_.c_str()};
+      const size_t node_matching_names_count = sizeof(node_matching_names) / sizeof(char *);
+      for (size_t index = 0U; index < node_matching_names_count; ++index) {
+        if (initial_map.count(node_matching_names[index]) > 0) {
           // Combine parameter yaml files, overwriting values in older ones
-          for (auto & param : iter->second) {
+          for (auto & param : initial_map.at(node_matching_names[index])) {
             parameter_overrides_[param.get_name()] =
               rclcpp::ParameterValue(param.get_value_message());
           }
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -110,12 +110,11 @@ NodeParameters::NodeParameters(
 
       // Enforce wildcard matching precedence
       // TODO(cottsay) implement further wildcard matching
-      const char * node_matching_names[] = {"/**", combined_name_.c_str()};
-      const size_t node_matching_names_count = sizeof(node_matching_names) / sizeof(char *);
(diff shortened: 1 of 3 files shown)


hidmic:
CI up to `rclpy` and `rclcpp`:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/10496/)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -107,11 +107,15 @@ NodeParameters::NodeParameters(
...
+
+      // Enforce wildcard matching precedence
+      // TODO(cottsay) implement further wildcard matching
+      const char * node_matching_names[] = {"/**", combined_name_.c_str()};
+      const size_t node_matching_names_count = sizeof(node_matching_names) / sizeof(char *);
+      for (size_t index = 0U; index < node_matching_names_count; ++index) {
+        if (initial_map.count(node_matching_names[index]) > 0) {
           // Combine parameter yaml files, overwriting values in older ones
-          for (auto & param : iter->second) {
+          for (auto & param : initial_map.at(node_matching_names[index])) {
```
clalancette, line 118: Definitely +1 to the idea, but can we use more modern C++ stuff here?  Something like:

```
  std::vector<std::string> node_matching_names{"/**", combined_name};
  for (const std::string & name : node_matching_names) {
    if (initial_map.count(name.c_str()) > 0) {
      for (auto & param : initial_map.at(name.c_str())) {
      }
    }
  }
```

(alternatively, if you don't want to do `name.c_str()` all of the time, you could make that a `std::vector` of `const char *` instead)

hidmic: I made the `rclpy` patch and the C vibe stuck with me :sweat_smile: 

Changed in https://github.com/ros2/rclcpp/pull/1094/commits/e9ba4f2715cac5867f053bda857057e4e2c81e13.

clalancette: > I made the `rclpy` patch first and the C vibe stuck with me sweat_smile

Ha, yeah, totally understood.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -107,11 +107,14 @@ NodeParameters::NodeParameters(
...
-        if (iter->first == "/**" || iter->first == combined_name_) {
+
+      // Enforce wildcard matching precedence
+      // TODO(cottsay) implement further wildcard matching
+      const std::vector<std::string> node_matching_names{"/**", combined_name_};
+      for (const auto & node_name : node_matching_names) {
+        if (initial_map.count(node_name) > 0) {
           // Combine parameter yaml files, overwriting values in older ones
-          for (auto & param : iter->second) {
+          for (auto & param : initial_map.at(node_name)) {
```
clalancette, line 117: One more nit: `const auto & param` here?

hidmic:
@jacobperron let me know how/if you want to proceed with this and https://github.com/ros2/rclpy/pull/547.

jacobperron:
@hidmic Seeing as it's a behavior change, I think it's okay to land these changes post-Foxy API freeze.

ivanpauno:
> @hidmic Seeing as it's a behavior change, I think it's okay to land these changes post-Foxy API freeze.

Is it really a behavior change? or is it a bug fix?
It sounds more like a bug fix to me, though it's not super clear.

hidmic:
> Is it really a behavior change? or is it a bug fix?

It's arguable. It may be neither, as it's not that wildcards were not working -- evaluation order was simply unspecified.

PR #1101, opened 2020-05-01
Use RCL_RET_SERVICE_TAKE_FAILED and not RCL_RET_CLIENT_TAKE_FAILED when checking a request take

ivanpauno:
Sounds like a typo/bug.

`rcl_take_request` for reference:
https://github.com/ros2/rcl/blob/47501d17c526ab6e1e133f9899b60a6049630cdc/rcl/src/rcl/service.c#L306.

Should solve https://ci.ros2.org/view/nightly/job/nightly_linux_release/1529/testReport/junit/(root)/projectroot/test_services_cpp__rmw_cyclonedds_cpp/ and similar failures in other nightlies.

[taken=false](https://github.com/ros2/rmw_cyclonedds/blob/40ba599462277dd5601a71b5ddb034e92ad6b565/rmw_cyclonedds_cpp/src/rmw_node.cpp#L3063-L3064) seems to be something more likely to happen after https://github.com/ros2/rmw_cyclonedds/pull/176. @eboasson It might worth investigating the why, but this fixes the issue.

### DIFF
--- a/rclcpp/src/rclcpp/service.cpp
+++ b/rclcpp/src/rclcpp/service.cpp
@@ -41,7 +41,7 @@ ServiceBase::take_type_erased_request(void * request_out, rmw_request_id_t & req
     this->get_service_handle().get(),
     &request_id_out,
     request_out);
-  if (RCL_RET_CLIENT_TAKE_FAILED == ret) {
+  if (RCL_RET_SERVICE_TAKE_FAILED == ret) {
     return false;
   } else if (RCL_RET_OK != ret) {
     rclcpp::exceptions::throw_from_rcl_error(ret);


eboasson:
> taken=false seems to be something more likely to happen after ros2/rmw_cyclonedds#176. @eboasson It might worth investigating the why, but this fixes the issue.

Interesting ... does the test process by any chance keep running while nodes come & go by any chance? If that's the case, then before ros2/rmw_cyclonedds#176 the participant would remain and discovery take place only once during the run of test process, where now it would be created and deleted multiple times. The former is definitely nicer behaviour, but hanging on termination is a show stopper.

ivanpauno:
> Interesting ... does the test process by any chance keep running while nodes come & go by any chance? If that's the case, then before ros2/rmw_cyclonedds#176 the participant would remain and discovery take place only once during the run of test process, where now it would be created and deleted multiple times

Yes, that's correct (the test is doing that).

> The former is definitely nicer behaviour, but hanging on termination is a show stopper.

Yes, ros2/rmw_cyclonedds#176 is still ok.
In normal use cases, the user doesn't create and delete nodes.
Just out of curiosity, why can the take of a request be affected by that? (it's still ok if that happens, as `take=false` is not considered an error).

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10602/)

eboasson:
> Just out of curiosity, why can the take of a request be affected by that?

It could be one of those “invalid samples” — a message that shows up in the result of take with the `valid_data` flag false to indicate that it is only a notification of a state change on the instance. If the service readers remains while the client writer disappears, the service will get a notification of that because the last known writer for the instance disappeared — the `instance_state` in the sample info will then go from `ALIVE` to `NOT_ALIVE_NO_WRITERS`.

As an aside: ROS2 may not “do keys” but for DDS that means there is still one instance for which it tracks known active writers. If you have multiple writers for the same topic in ROS2, in Cyclone they will all end up being a writer for the same topic, and so a single writer disappearing may or may not result in the instance state changing.

As an aside to the aside: as I understand it, at least Connext implements treats topics without key fields entirely differently. That does make some sense from a backwards compatibility thing for them if you go back to the time the DDS specification was written. I don’t know what Fast-RTPS does, but given that the RTPS specification suggests they should be treated differently, it might well not be generating these invalid samples. So it may well be that Cyclone has `taken` set to  false relatively often compared to the others.

ivanpauno:
Thanks for the details @eboasson !
Considering that, it sounds like this is the correct fix, and that getting "invalid samples" is expected.

wjwwood:
> It could be one of those “invalid samples” — a message that shows up in the result of take with the valid_data flag false to indicate that it is only a notification of a state change on the instance.

That's definitely the case, in my opinion.

PR #1102, opened 2020-05-04
Added dockblock to ComponentManager class

ahcorde:
Added dockblock to ComponentManager class

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/rclcpp_components/include/rclcpp_components/component_manager.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager.hpp
@@ -77,11 +77,10 @@ public:
     std::string node_name = "ComponentManager",
     const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
 
-  /// Default destructor
   RCLCPP_COMPONENTS_PUBLIC
   virtual ~ComponentManager();
 
-  /// Get the component resource
+  /// Return a list of valid loadable components in a given package.
   /*
    * \param package_name name of the package
    * \param resource_index name of the executable
@@ -94,16 +93,17 @@ public:
     const std::string & package_name,
     const std::string & resource_index = "rclcpp_components") const;
 
-  /// Create the component factory
+  /// Instantiate a component from a dynamic library.
   /*
-   * \param resource list of component resources
+   * \param resource a component resource (class name + library path)
+   * \return a NodeFactory interface
    */
   RCLCPP_COMPONENTS_PUBLIC
   virtual std::shared_ptr<rclcpp_components::NodeFactory>
   create_component_factory(const ComponentResource & resource);
 
 protected:
-  // Service to load a new node in the component
+  /// Service callback to load a new node in the component
   /*
    * This function allows to add parameters, remap rules, a specific node, name a namespace
    * and/or additional arguments.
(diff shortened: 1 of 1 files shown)


rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -60,41 +62,89 @@ class ComponentManager : public rclcpp::Node
...
+   * \param node_name the name of the node that the data originates from.
+   * \param node_options additional options to control creation of the node.
+   */
   RCLCPP_COMPONENTS_PUBLIC
   ComponentManager(
     std::weak_ptr<rclcpp::Executor> executor,
     std::string node_name = "ComponentManager",
     const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
 
+  /// Default destructor
```
brawner, line 80: Not needed.

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -60,41 +62,89 @@ class ComponentManager : public rclcpp::Node
...
   RCLCPP_COMPONENTS_PUBLIC
   virtual ~ComponentManager();
 
-  /// Return a list of valid loadable components in a given package.
+  /// Get the component resource
```
brawner, line 84: The other comment had more information, why the change?

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -60,41 +62,89 @@ class ComponentManager : public rclcpp::Node
...
+   * \throws ComponentManagerException if the resource was not found or a invalid resource entry
+   * \return a list of component resources
+   */
   RCLCPP_COMPONENTS_PUBLIC
   virtual std::vector<ComponentResource>
   get_component_resources(
     const std::string & package_name,
     const std::string & resource_index = "rclcpp_components") const;
 
+  /// Create the component factory
```
brawner, line 97: Is there more detail you can add to the brief?

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -60,41 +62,89 @@ class ComponentManager : public rclcpp::Node
...
+  /*
+   * \param resource list of component resources
+   */
   RCLCPP_COMPONENTS_PUBLIC
   virtual std::shared_ptr<rclcpp_components::NodeFactory>
   create_component_factory(const ComponentResource & resource);
 
 protected:
+  // Service to load a new node in the component
```
brawner, line 106: This is the service callback, right? Say that instead. Also, when might a deriving class want to override these functions?

These questions apply also to the next two functions.

ahcorde: If you derive from this class and change these methods we are going to change the behaviour of this abstraction. do you think we should explicitly indicate it?

brawner: I was just surprised to see protected virtual callbacks, and I wasn't entirely sure when that would be useful to override, so I thought some documentation might help. But seeing as all the public functions are virtual as well, I guess this class was just designed with that sort of flexibility in mind.

ahcorde:
running CI up-to `rclcpp_components` to check linters

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10717/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6130/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8732/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10617/)

PR #1113, opened 2020-05-11
Allow spin_until_future_complete to accept std::future

### LINKED ISSUE #972: Allow passing a `std::future` to `spin_until_future_complete`
#### Feature description

Currently, `spin_until_future_completes` only accepts a `std::shared_future`.
It would make sense to accept both.
See https://github.com/ros2/rclcpp/pull/971.

Just templatize more the current functions/methods.

naiveHobo:
Signed-off-by: Sarthak Mittal <sarthakmittal2608@gmail.com>

closes #972

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -188,59 +188,10 @@ public:
    *   code.
    * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
    */
-  template<typename ResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+  template<typename FutureResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli>
   FutureReturnCode
   spin_until_future_complete(
-    const std::shared_future<ResponseT> & future,
-    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
-  {
-    // TODO(wjwwood): does not work recursively; can't call spin_node_until_future_complete
-    // inside a callback executed by an executor.
-
-    // Check the future before entering the while loop.
-    // If the future is already complete, don't try to spin.
-    std::future_status status = future.wait_for(std::chrono::seconds(0));
-    if (status == std::future_status::ready) {
-      return FutureReturnCode::SUCCESS;
-    }
-
-    auto end_time = std::chrono::steady_clock::now();
-    std::chrono::nanoseconds timeout_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
-      timeout);
-    if (timeout_ns > std::chrono::nanoseconds::zero()) {
-      end_time += timeout_ns;
-    }
-    std::chrono::nanoseconds timeout_left = timeout_ns;
-
-    while (rclcpp::ok(this->context_)) {
-      // Do one item of work.
-      spin_once(timeout_left);
-      // Check if the future is set, return SUCCESS if it is.
-      status = future.wait_for(std::chrono::seconds(0));
-      if (status == std::future_status::ready) {
-        return FutureReturnCode::SUCCESS;
-      }
-      // If the original timeout is < 0, then this is blocking, never TIMEOUT.
-      if (timeout_ns < std::chrono::nanoseconds::zero()) {
-        continue;
-      }
-      // Otherwise check if we still have time to wait, return TIMEOUT if not.
-      auto now = std::chrono::steady_clock::now();
-      if (now >= end_time) {
-        return FutureReturnCode::TIMEOUT;
-      }
-      // Subtract the elapsed time from the original timeout.
-      timeout_left = std::chrono::duration_cast<std::chrono::nanoseconds>(end_time - now);
-    }
-
-    // The future did not complete before ok() returned false, return INTERRUPTED.
-    return FutureReturnCode::INTERRUPTED;
-  }
-
-  template<typename ResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli>
-  FutureReturnCode
-  spin_until_future_complete(
-    const std::future<ResponseT> & future,
+    const FutureResponseT & future,
     std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
   {
     // TODO(wjwwood): does not work recursively; can't call spin_node_until_future_complete
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -237,6 +237,55 @@ public:
     return FutureReturnCode::INTERRUPTED;
   }
 
+  template<typename ResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+  FutureReturnCode
+  spin_until_future_complete(
+    const std::future<ResponseT> & future,
+    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
+  {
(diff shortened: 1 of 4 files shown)


rclcpp/include/rclcpp/executor.hpp
```diff
@@ -231,6 +231,55 @@ class Executor
     return FutureReturnCode::INTERRUPTED;
   }
 
+  template<typename ResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+  FutureReturnCode
+  spin_until_future_complete(
+    const std::future<ResponseT> & future,
+    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
```
ivanpauno, line 238: What I see, is that this is exactly a repetition of the template method above, but not just taking the "future response" as a template parameter, instead of taking ResponseT. i.e. update the current template definition to:

```cpp
  template<typename FutureResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli>
  FutureReturnCode
  spin_until_future_complete(
    const FutureResponseT && future,
    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
```

ivanpauno: This comment applies for all the other methods too

naiveHobo:
@ivanpauno could you elaborate a bit more on how the tests should look like? I can't see any specific tests for `spin_until_future_complete` so I'm not completely sure what to do (sorry first PR on rclcpp).

ivanpauno:
> @ivanpauno could you elaborate a bit more on how the tests should look like? I can't see any specific tests for `spin_until_future_complete` so I'm not completely sure what to do (sorry first PR on rclcpp).

It's true that there is not a unit test for `spin_until_future_complete` currently, so it would be great to add one!
I imagine something like:

```cpp
SingleThreadedExecutor ex;
std::future<void> future = std::async([]() {
  return;
});
FutureReturnCode ret = ex.spin_until_future_complete(future, 1s);
EXPECT_EX(FutureReturnCode::SUCCESS, ret);
```

The version of the test for the `std::shared_future` signature, is the same but with a `std::shared_future sh_future = future.get_shared();` line in the middle.

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -104,7 +104,7 @@ template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::mi
 rclcpp::FutureReturnCode
 spin_until_future_complete(
   rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
-  const std::shared_future<FutureT> & future,
+  const FutureT & future,
```
ivanpauno, line 107: nit: I know this predates this PR, but I would unify all templates parameter names to either `FutureT` or `FutureResponseT`.
Though I suggested the later, the former sounds nicer :smile:.

rclcpp/test/test_executor.cpp
```diff
@@ -67,3 +67,47 @@ TEST_F(TestExecutors, addTemporaryNode) {
...
+  // test success
+  future = std::async([]() {
+    return;
+  });
+  ret = executor.spin_until_future_complete(future, 1s);
+  EXPECT_EQ(rclcpp::FutureReturnCode::SUCCESS, ret);
+
+  // test timeout
+  future = std::async([]() {
+    std::this_thread::sleep_for(20ms);
```
ivanpauno: Use a bigger number here e.g.: `1s`. That will avoid test flakiness (Windows time slice is ~100ms)

rclcpp/test/test_executor.cpp
```diff
@@ -67,3 +67,47 @@ TEST_F(TestExecutors, addTemporaryNode) {
...
+  // test success
+  future = std::async([]() {
+    return;
+  });
+  ret = executor.spin_until_future_complete(future.share(), 1s);
+  EXPECT_EQ(rclcpp::FutureReturnCode::SUCCESS, ret);
+
+  // test timeout
+  future = std::async([]() {
+    std::this_thread::sleep_for(20ms);
```
ivanpauno: same here

ivanpauno:
@naiveHobo I will merge this past Foxy rosdistro freeze (June 3th).

You can ping me if I forget :grin:

naiveHobo:
@ivanpauno thank you for the help!

naiveHobo:
@ivanpauno @fujitatomoya should be good to merge now, right?

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11046/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6370/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8997/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10969/)

dirk-thomas:
When changing public API of a package it is not sufficient to run CI builds limited to `--packages-up-to` that package. Downstream packages using that API need to be built and tested too in order to ensure the change doesn't break existing code.

In this case downstream packages fail to build, e.g. `interactive_markers`: [Build Status](https://ci.ros2.org/job/ci_linux/11064/)

Since this broke all builds I am going to revert it before the nightlies: #1159.

Please follow up with a redo of this patch. Either with updates to affected packages or - even better - with a modified patch which maintains API compatibility (if possible). I reopened #972 to track this.

ivanpauno:
@naiveHobo Can you open another PR?
This one was reverted in https://github.com/ros2/rclcpp/pull/1159 because it caused problems in an upstream package https://ci.ros2.org/job/ci_linux/11064/console#console-section-13, so we will have to sort that out before merging again.

ivanpauno:
The error was in this line:

https://github.com/ros-visualization/interactive_markers/blob/1ad7c9b029b3f582f1fc5ff6fd43fa685c535614/test/interactive_markers/test_interactive_marker_server.cpp#L370

Where the response type was explicitly passed.
We can delete that explicit parameter in `interactive_markers` and re-merge this as-is (and template argument deduction will do the right thing).

After seeing that error, my question is if we prefer an API compatible change instead of this:
```cpp
template<typename ResponseT, typename TimeRepT = int64_t, typename TimeT = std::milli, template<X> class FutureT = std::shared_future>
rclcpp::FutureReturnCode
spin_node_until_future_complete(
  rclcpp::Executor & executor,
  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
  const FutureT<ResponseT> & future,
```

Note: I don't know if that is exactly valid CPP, but something like that.

I prefer the API breaking version, as I find it more clear.
@wjwwood @hidmic opinions?

ivanpauno:
> When changing public API of a package it is not sufficient to run CI builds limited to `--packages-up-to` that package. Downstream packages using that API need to be built and tested too in order to ensure the change doesn't break existing code.

Yeah sorry, I forgot this was breaking API.

naiveHobo:
I think making the change in `interactive_markers` makes more sense. This PR doesn't break any other package downstream. Plus, I feel this makes for a better API. Thoughts?

naiveHobo:
@ivanpauno does this need another PR once https://github.com/ros-visualization/interactive_markers/pull/72 is in?

ivanpauno:
> @ivanpauno does this need another PR once [ros-visualization/interactive_markers#72](https://github.com/ros-visualization/interactive_markers/pull/72) is in?

Yes, please open another PR.

ivanpauno:
> > @ivanpauno does this need another PR once [ros-visualization/interactive_markers#72](https://github.com/ros-visualization/interactive_markers/pull/72) is in?
> 
> Yes, please open another PR.

Doesn't matter, I will open a revert PR.

ivanpauno:
@naiveHobo FYI I've opened https://github.com/ros2/rclcpp/pull/1160.
Thanks a lot for your contribution!!!

PR #1114, opened 2020-05-12
use rosidl_default_generators dependency in test

Karsten1987:
`rosidl_default_generators` are only used in tests and should be included only when building tests.


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10715/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6128/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8730/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10615/)

PR #1119, opened 2020-05-19
Extend the lifetime of rcl_publisher_t by holding onto the shared pointer.

clalancette:
This PR aims to fix a crash that can be seen while using the code in https://github.com/clalancette/rclcpp_event_crash.   That code creates a node, then creates a thread to create publishers in the thread.  The default executor is then run in the main thread.  If you build that code against latest master, put a lot of CPU stress on the machine, and run that code using Connext, it will crash almost every time.

After some debugging, some advice from @wjwwood and @ivanpauno below, we determined that the problem is that the publisher handle (`rcl_publisher_t`, via `PublisherBase` class) could be freed while the executor was still working on QOSEvent handles.  The fix here is to hold onto a shared pointer to the underlying `rcl_{publisher,subscription}_t` structure in QOSEvent, which keeps those structures around until the executor is done with them.

The downside to this change is that it changes the API of `rclcpp`, making `get_publisher_handle` return a shared_ptr rather than a raw pointer.  This also requires https://github.com/ros2/demos/pull/448 and https://github.com/ros2/rosbag2/pull/420 to compile successfully.

### DIFF
--- a/rclcpp/include/rclcpp/publisher_base.hpp
+++ b/rclcpp/include/rclcpp/publisher_base.hpp
@@ -200,10 +200,11 @@ protected:
     const EventCallbackT & callback,
     const rcl_publisher_event_type_t event_type)
   {
-    auto handler = std::make_shared<QOSEventHandler<EventCallbackT>>(
+    auto handler = std::make_shared<QOSEventHandler<EventCallbackT,
+        std::shared_ptr<rcl_publisher_t>>>(
       callback,
       rcl_publisher_event_init,
-      publisher_handle_.get(),
+      publisher_handle_,
       event_type);
     event_handlers_.emplace_back(handler);
   }
--- a/rclcpp/include/rclcpp/publisher_base.hpp
+++ b/rclcpp/include/rclcpp/publisher_base.hpp
@@ -99,13 +99,13 @@ public:
   /// Get the rcl publisher handle.
   /** \return The rcl publisher handle. */
   RCLCPP_PUBLIC
-  rcl_publisher_t *
+  std::shared_ptr<rcl_publisher_t>
   get_publisher_handle();
 
   /// Get the rcl publisher handle.
   /** \return The rcl publisher handle. */
   RCLCPP_PUBLIC
-  const rcl_publisher_t *
+  std::shared_ptr<const rcl_publisher_t>
   get_publisher_handle() const;
 
   /// Get all the QoS event handlers associated with this publisher.
@@ -203,7 +203,7 @@ protected:
     auto handler = std::make_shared<QOSEventHandler<EventCallbackT>>(
       callback,
       rcl_publisher_event_init,
-      &publisher_handle_,
+      publisher_handle_.get(),
       event_type);
     event_handlers_.emplace_back(handler);
   }
@@ -213,7 +213,7 @@ protected:
 
   std::shared_ptr<rcl_node_t> rcl_node_handle_;
 
-  rcl_publisher_t publisher_handle_ = rcl_get_zero_initialized_publisher();
+  std::shared_ptr<rcl_publisher_t> publisher_handle_;
 
   std::vector<std::shared_ptr<rclcpp::QOSEventHandlerBase>> event_handlers_;
 
--- a/rclcpp/src/rclcpp/publisher_base.cpp
+++ b/rclcpp/src/rclcpp/publisher_base.cpp
@@ -47,8 +47,24 @@ PublisherBase::PublisherBase(
 : rcl_node_handle_(node_base->get_shared_rcl_node_handle()),
   intra_process_is_enabled_(false), intra_process_publisher_id_(0)
 {
+  auto custom_deleter = [node_handle = this->rcl_node_handle_](rcl_publisher_t * rcl_pub)
+    {
+      if (rcl_publisher_fini(rcl_pub, node_handle.get()) != RCL_RET_OK) {
+        RCLCPP_ERROR(
+          rclcpp::get_node_logger(node_handle.get()).get_child("rclcpp"),
+          "Error in destruction of rcl publisher handle: %s",
+          rcl_get_error_string().str);
+        rcl_reset_error();
+      }
+      delete rcl_pub;
+    };
+
+  publisher_handle_ = std::shared_ptr<rcl_publisher_t>(
+    new rcl_publisher_t, custom_deleter);
+  *publisher_handle_.get() = rcl_get_zero_initialized_publisher();
+
   rcl_ret_t ret = rcl_publisher_init(
-    &publisher_handle_,
+    publisher_handle_.get(),
     rcl_node_handle_.get(),
     &type_support,
     topic.c_str(),
@@ -67,7 +83,7 @@ PublisherBase::PublisherBase(
     rclcpp::exceptions::throw_from_rcl_error(ret, "could not create publisher");
   }
   // Life time of this object is tied to the publisher handle.
-  rmw_publisher_t * publisher_rmw_handle = rcl_publisher_get_rmw_handle(&publisher_handle_);
+  rmw_publisher_t * publisher_rmw_handle = rcl_publisher_get_rmw_handle(publisher_handle_.get());
   if (!publisher_rmw_handle) {
     auto msg = std::string("failed to get rmw handle: ") + rcl_get_error_string().str;
     rcl_reset_error();
@@ -85,14 +101,6 @@ PublisherBase::~PublisherBase()
   // must fini the events before fini-ing the publisher
   event_handlers_.clear();
 
-  if (rcl_publisher_fini(&publisher_handle_, rcl_node_handle_.get()) != RCL_RET_OK) {
-    RCUTILS_LOG_ERROR_NAMED(
-      "rclcpp",
-      "Error in destruction of rcl publisher handle: %s",
-      rcl_get_error_string().str);
-    rcl_reset_error();
-  }
-
   auto ipm = weak_ipm_.lock();
 
   if (!intra_process_is_enabled_) {
@@ -102,7 +110,7 @@ PublisherBase::~PublisherBase()
     // TODO(ivanpauno): should this raise an error?
     RCLCPP_WARN(
       rclcpp::get_logger("rclcpp"),
-      "Intra process manager died before than a publisher.");
+      "Intra process manager died before a publisher.");
     return;
   }
   ipm->remove_publisher(intra_process_publisher_id_);
@@ -111,13 +119,14 @@ PublisherBase::~PublisherBase()
 const char *
 PublisherBase::get_topic_name() const
 {
-  return rcl_publisher_get_topic_name(&publisher_handle_);
+  return rcl_publisher_get_topic_name(publisher_handle_.get());
 }
 
 size_t
 PublisherBase::get_queue_size() const
 {
-  const rcl_publisher_options_t * publisher_options = rcl_publisher_get_options(&publisher_handle_);
+  const rcl_publisher_options_t * publisher_options = rcl_publisher_get_options(
+    publisher_handle_.get());
   if (!publisher_options) {
     auto msg = std::string("failed to get publisher options: ") + rcl_get_error_string().str;
     rcl_reset_error();
@@ -132,16 +141,16 @@ PublisherBase::get_gid() const
   return rmw_gid_;
 }
 
-rcl_publisher_t *
+std::shared_ptr<rcl_publisher_t>
 PublisherBase::get_publisher_handle()
 {
-  return &publisher_handle_;
+  return publisher_handle_;
 }
 
-const rcl_publisher_t *
+std::shared_ptr<const rcl_publisher_t>
 PublisherBase::get_publisher_handle() const
 {
-  return &publisher_handle_;
+  return publisher_handle_;
 }
 
 const std::vector<std::shared_ptr<rclcpp::QOSEventHandlerBase>> &
@@ -156,13 +165,13 @@ PublisherBase::get_subscription_count() const
   size_t inter_process_subscription_count = 0;
 
   rcl_ret_t status = rcl_publisher_get_subscription_count(
-    &publisher_handle_,
+    publisher_handle_.get(),
     &inter_process_subscription_count);
 
   if (RCL_RET_PUBLISHER_INVALID == status) {
     rcl_reset_error();  /* next call will reset error message if not context */
-    if (rcl_publisher_is_valid_except_context(&publisher_handle_)) {
-      rcl_context_t * context = rcl_publisher_get_context(&publisher_handle_);
+    if (rcl_publisher_is_valid_except_context(publisher_handle_.get())) {
+      rcl_context_t * context = rcl_publisher_get_context(publisher_handle_.get());
       if (nullptr != context && !rcl_context_is_valid(context)) {
         /* publisher is invalid due to context being shutdown */
         return 0;
@@ -195,7 +204,7 @@ PublisherBase::get_intra_process_subscription_count() const
 rclcpp::QoS
 PublisherBase::get_actual_qos() const
 {
-  const rmw_qos_profile_t * qos = rcl_publisher_get_actual_qos(&publisher_handle_);
+  const rmw_qos_profile_t * qos = rcl_publisher_get_actual_qos(publisher_handle_.get());
   if (!qos) {
     auto msg = std::string("failed to get qos settings: ") + rcl_get_error_string().str;
     rcl_reset_error();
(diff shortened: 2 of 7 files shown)


rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -187,7 +187,7 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
         group->find_waitable_ptrs_if(
           [this](const rclcpp::Waitable::SharedPtr & waitable) {
             waitable_handles_.push_back(waitable);
-            return false;
+            return true;
```
wjwwood, line 190: I don't think this is extending the lifetime of the waitable, it is just stopping the iteration through the waitables after visiting the first one:

https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/include/rclcpp/callback_group.hpp#L155

So I think you're just avoiding the bug, not fixing it with this.

These functions are being used for their side effects, essentially always returning false means the "find" function will be called for each entry, it's essentially being used as a "for each" or "map" like function. It's kind of ugly to be honest... Not sure if it's my fault or not, haha. Doesn't matter though.

clalancette: > So I think you're just avoiding the bug, not fixing it with this.

Yeah, you are completely right about this.  I misunderstood the logic.

ivanpauno: [Waitable](https://github.com/ros2/rclcpp/blob/01a6befdde7f78b6f7788e6180cc5ed38390e955/rclcpp/include/rclcpp/waitable.hpp#L26) class doesn't define a virtual destructor, and then:

https://github.com/ros2/rclcpp/blob/01a6befdde7f78b6f7788e6180cc5ed38390e955/rclcpp/include/rclcpp/qos_event.hpp#L79-L83

Sounds like that might be the problem (I really don't know if that's the case, just wondering)

clalancette: > Sounds like that might be the problem (I really don't know if that's the case, just wondering)

Huh, good point.  It does look like this is *a* problem, but putting a virtual destructor on the `Waitable` class just moved the problem around.  More debugging required.

ivanpauno: From the discussion in yesterday's meeting, I have an idea I want to share that might help to fix the problem:

Events can be created both from a [publisher](https://github.com/ros2/rmw/blob/d030ad19eaeb6bee3eda7c0197360442e2db0a7a/rmw/include/rmw/event.h#L79) and a [subscription](https://github.com/ros2/rmw/blob/d030ad19eaeb6bee3eda7c0197360442e2db0a7a/rmw/include/rmw/event.h#L97).
The `rmw` implementation then holds a pointer to underlying implementation the Publisher/Subscription (see [1](https://github.com/ros2/rmw_connext/blob/c90e7752a2f482aef4afe53dc0ce7f45c632152e/rmw_connext_cpp/src/rmw_event.cpp#L33), [2](https://github.com/ros2/rmw_connext/blob/c90e7752a2f482aef4afe53dc0ce7f45c632152e/rmw_connext_cpp/src/rmw_event.cpp#L47), [3](https://github.com/ros2/rmw_connext/blob/c90e7752a2f482aef4afe53dc0ce7f45c632152e/rmw_connext_shared_cpp/src/event.cpp#L46)).

Then, at `rclcpp` level, the created `EventHandler` is not holding the `Publisher`/`Subscription`:

https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/include/rclcpp/qos_event.hpp#L101-L102

when that actually should be hold.

It's not possible to hold the `rclcpp` objects in the event handlers, as that would create a cyclic dependency ([4](https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/include/rclcpp/subscription_base.hpp#L286), [5](https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/include/rclcpp/publisher_base.hpp#L218)), but holding the `rcl` objects should solve the problem.

To avoid creating derived classes of [QOSEventHandler](https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/include/rclcpp/qos_event.hpp#L106) for publisher and subscriptions, we can store a `shared_ptr<void>` to the parent handler.

Disclaimer: I haven't tested the issue and then I'm not sure if this is actually the root cause of the problem, but there is definitely something "unsound" here.

clalancette: Thanks for the thoughts @ivanpauno .  You've got the problem exactly right; we aren't holding a reference to the Publisher, and so some of the underlying rcl structures can get freed and then improperly reused in the executor.

I like your idea about holding onto a shared pointer for the underlying rcl structures as a fix.  I'll start trying to do that and see what happens there.

clalancette: All right, this fix works, and I feel a *lot* better about it in general.  @ivanpauno thanks again for the idea.  @wjwwood another review here would be appreciated.

clalancette:
CI smoke test Linux: [Build Status](http://ci.ros2.org/job/ci_linux/10915/)

rclcpp/include/rclcpp/publisher_base.hpp
```diff
@@ -99,13 +99,13 @@ class PublisherBase : public std::enable_shared_from_this<PublisherBase>
...
   RCLCPP_PUBLIC
-  rcl_publisher_t *
+  std::shared_ptr<rcl_publisher_t>
   get_publisher_handle();
 
   /// Get the rcl publisher handle.
   /** \return The rcl publisher handle. */
   RCLCPP_PUBLIC
-  const rcl_publisher_t *
+  const std::shared_ptr<rcl_publisher_t>
```
ivanpauno, line 108: if I'm not getting confused, it should be `std::shared_ptr<const rcl_publisher_t>`

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -47,8 +47,24 @@ PublisherBase::PublisherBase(
 : rcl_node_handle_(node_base->get_shared_rcl_node_handle()),
   intra_process_is_enabled_(false), intra_process_publisher_id_(0)
 {
+  auto custom_deletor = [node_handle = this->rcl_node_handle_](rcl_publisher_t * rcl_pub)
```
ivanpauno, line 50: nit: deletor -> deleter ?

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -132,16 +140,16 @@ PublisherBase::get_gid() const
...
-rcl_publisher_t *
+std::shared_ptr<rcl_publisher_t>
 PublisherBase::get_publisher_handle()
 {
-  return &publisher_handle_;
+  return publisher_handle_;
 }
 
-const rcl_publisher_t *
+const std::shared_ptr<rcl_publisher_t>
```
ivanpauno, line 149: same about `std::shared_ptr<const rcl_publisher_t>`

ivanpauno:
The API/ABI break seems completely necessary to fix the issue.

The missing `virtual` keyword commented [here](https://github.com/ros2/rclcpp/pull/1119#discussion_r428216447) should also be addressed (in either this PR or a separated one).

rclcpp/src/rclcpp/publisher_base.cpp
clalancette: > nit: deletor -> deleter ?

Heh.  I usually take the `-or` suffix as "the one who deletes".  But in this case, there doesn't seem to be a real word corresponding to this.  This is copied from https://github.com/ros2/rclcpp/blob/eff11d61bbb60220003779eb4a5a1b205a76da29/rclcpp/src/rclcpp/subscription_base.cpp#L47 , but I agree with you on `deleter`; I'll change it here and we can change it in the subscription_base some other time.

ivanpauno: I'm not a native English speaker, so you can ignore my comment :smile:.

I've seen `deleter` many times and I don't remember having seen `deletor`, so I thought it was a typo.

rclcpp/include/rclcpp/publisher_base.hpp
clalancette: This is a copy of the code in https://github.com/ros2/rclcpp/blob/eff11d61bbb60220003779eb4a5a1b205a76da29/rclcpp/include/rclcpp/subscription_base.hpp#L93, but modified for publishers.  As I understand, what this code is doing as-is is making sure that the shared_ptr is const (that is, you can't reassign the returned value).  What you are suggesting is we make the structure const (that is, you can't change the structure).  That's definitely a stronger guarantee, but I don't think that that is the intended use-case.  For instance, one use is https://github.com/ros2/rclcpp/pull/1119/files#diff-8c68c968941e2e05ba8974455ea3f1b5R140 , where we are calling `rcl_return_loaned_message_from_publisher()`.  In turn, the API for that is [here](https://github.com/ros2/rcl/blob/0de31d04830f0b33a3adf4a87e179e0d0d76555b/rcl/include/rcl/publisher.h#L256), where it expects a `const rcl_publisher_t *`.

So while I like the stronger const guarantee in general, I think that is a bigger issue that we'd have to evaluate separately.

clalancette:
> The missing `virtual` keyword commented [here](https://github.com/ros2/rclcpp/pull/1119#discussion_r428216447) should also be addressed (in either this PR or a separated one).

Yeah, I was wondering if I should include it here or not.  I think I will add it; while it is not directly related, the PR isn't all that huge and it is tangentially related, so I'll just put it here so we don't forget about it.

rclcpp/include/rclcpp/publisher_base.hpp
ivanpauno: > This is a copy of the code in https://github.com/ros2/rclcpp/blob/eff11d61bbb60220003779eb4a5a1b205a76da29/rclcpp/include/rclcpp/subscription_base.hpp#L93, but modified for publishers. 

mmm, I didn't check what `Subscription` does (neither why)

> As I understand, what this code is doing as-is is making sure that the shared_ptr is const (that is, you can't reassign the returned value).

Yes

> What you are suggesting is we make the structure const (that is, you can't change the structure).

Yes

> That's definitely a stronger guarantee, but I don't think that that is the intended use-case.  For instance, one use is https://github.com/ros2/rclcpp/pull/1119/files#diff-8c68c968941e2e05ba8974455ea3f1b5R140 , where we are calling `rcl_return_loaned_message_from_publisher()`.  In turn, the API for that is [here](https://github.com/ros2/rcl/blob/0de31d04830f0b33a3adf4a87e179e0d0d76555b/rcl/include/rcl/publisher.h#L256), where it expects a `const rcl_publisher_t *`.
> So while I like the stronger const guarantee in general, I think that is a bigger issue that we'd have to evaluate separately.

* `const rcl_publisher_t *`: The `rcl_publisher_t` structure can't be modified.
* `rcl_publisher_t * const`: The pointer can't be reassigned.

Before, we were returning `const rcl_publisher_t *`.
I don't see why we should change the "constness" when upgrading to a `shared_ptr`.
This is relaxing the previous guarantee.

clalancette: We can't drop the signature; some of the `rcl` functions need a const pointer in order to compile.  But I can change it to what @ivanpauno recommends.

clalancette:
CI on all platforms:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10922/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6271/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8893/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10823/)

rclcpp/include/rclcpp/publisher_base.hpp
wjwwood: What? If you have a non-const pointer, then you can get a const pointer, haha.

clalancette: Hm, you make a good point, but I definitely couldn't compile without this in place.  Let me take a quick look again.

ivanpauno: > We can't drop the signature; some of the rcl functions need a const pointer in order to compile. But I can change it to what @ivanpauno recommends.

I think there's a bit of confusion.
If the method overload with const qualifier is deleted (with const qualifier I mean the one affecting `this` ptr), then in the places where we pass a `const &` or `const *` to a `rclcpp::Publisher` we cannot call `get_publisher_handle`.

It seems that we're doing that somewhere, and thus that overload can't be removed.

clalancette: Yeah, without this signature, I get the following compile error:

```
In file included from /home/ubuntu/ros2_ws/src/ros2/rclcpp/rclcpp/include/rclcpp/publisher.hpp:35:0,
                 from /home/ubuntu/ros2_ws/src/ros2/rclcpp/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp:31,
                 from /home/ubuntu/ros2_ws/src/ros2/rclcpp/rclcpp/include/rclcpp/subscription.hpp:50,
                 from /home/ubuntu/ros2_ws/src/ros2/rclcpp/rclcpp/include/rclcpp/any_executable.hpp:25,
                 from /home/ubuntu/ros2_ws/src/ros2/rclcpp/rclcpp/src/rclcpp/any_executable.cpp:15:
/home/ubuntu/ros2_ws/src/ros2/rclcpp/rclcpp/include/rclcpp/loaned_message.hpp: In constructor ‘rclcpp::LoanedMessage<MessageT, AllocatorT>::LoanedMessage(const rclcpp::PublisherBase&, std::allocator<_Tp1>)’:
/home/ubuntu/ros2_ws/src/ros2/rclcpp/rclcpp/include/rclcpp/loaned_message.hpp:68:35: error: no matching function for call to ‘rclcpp::PublisherBase::get_publisher_handle() const’
         pub_.get_publisher_handle().get(),
```

That comes from this line of code: https://github.com/ros2/rclcpp/blob/1a48a60a75895ae86c1f93a57bc1955b11fb56c9/rclcpp/include/rclcpp/loaned_message.hpp#L68 .  The signature of `rcl_borrow_loaned_message` comes from https://github.com/ros2/rcl/blob/master/rcl/include/rcl/publisher.h#L226 .  I guess it is the lack of `const` on the method itself that is the missing bit, not the const pointer.

Not super important.  Just wanted to followup :).

wjwwood: Ah, sorry, yeah we need a const qualified method for this class, it's not that the pointer which is returned is const or not.

wjwwood: Well also that signature may be wrong (for LoanedMessage), if it wants to take a const reference to a PublisherBase, but then extract a non-const pointer for the rcl publisher from it. Instead maybe it should take a normal reference or a pointer to a PublisherBase...

clalancette: I've opened https://github.com/ros2/rclcpp/issues/1143 to track this for the future.

ivanpauno: > Well also that signature may be wrong (for LoanedMessage), if it wants to take a const reference to a PublisherBase, but then extract a non-const pointer for the rcl publisher from it.

`LoanedMessage` is taking a [const rclcpp::PublisherBase & pub](https://github.com/ros2/rclcpp/blob/1a48a60a75895ae86c1f93a57bc1955b11fb56c9/rclcpp/include/rclcpp/loaned_message.hpp#L60), and then extracting a : [**const** rcl_publisher_t * publisher](https://github.com/ros2/rcl/blob/c212a0dba4c0fbb833429e38992a0f765b84910e/rcl/include/rcl/publisher.h#L227).

I don't see where's the issue.
We've already changed the signature to return a `std::shared_ptr<const rcl_publisher_t>` in https://github.com/ros2/rclcpp/pull/1119/commits/b9bf5dc9c29f1972c4bed0a873df98908ee54fd4, so everything looks correct to me.

wjwwood: Yeah, ok that's fine then.

wjwwood: My (unlikely) concern was that this signature:

https://github.com/ros2/rcl/blob/c212a0dba4c0fbb833429e38992a0f765b84910e/rcl/include/rcl/publisher.h#L227

Is incorrectly accepting a const publisher and later mutating its state anyways, but maybe not, maybe it is just used as a key for lookup in some other structure.

I think this is fine, we don't need to do anything more.

PR #1122, opened 2020-05-20
[rclcpp_action] Action client holds weak pointers to goal handles

### LINKED ISSUE #861: rclcpp_action: In certain circumstances, goal can be gone before client gets result
## Bug report

While investigating a failure as part of https://github.com/ros2/geometry2/pull/172, I found out that action servers and clients in the same process may run into a race.  If the server immediately sets a goal to "success" or "failed" during the "handle_accept" callback (like [here](https://github.com/ros2/geometry2/blob/fc774a02a4e77df80074dbce4e554cfaa8653b4e/tf2_ros/test/test_buffer_client.cpp#L67)), the entire action may be gone by the time the client goes to get the result (via [async_get_result](https://github.com/ros2/rclcpp/blob/b6d18ccc813d9f35b080795de764b1730702981a/rclcpp_action/include/rclcpp_action/client.hpp#L401)).  I was only able to reproduce this particular behavior while using Windows Debug.  It's not entirely clear to me if this is a bug, since I'm not sure if it is valid for a server to set the goal to "success" or "fail" during the "handle_accept" callback.  Nonetheless, @jacobperron thought it would be a good idea to open a bug to track this.

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - Source (Debug build - `colcon build --merge-install --cmake-args -DCMAKE_BUILD_TYPE=Debug`)
- Version or commit hash:
  - b6d18ccc813d9f35b080795de764b1730702981a
- DDS implementation:
  - Fast-RTPS
  - rclcpp

#### Steps to reproduce issue

Comment out the workaround line [here](https://github.com/ros2/geometry2/blob/fc774a02a4e77df80074dbce4e554cfaa8653b4e/tf2_ros/test/test_buffer_client.cpp#L71).

Build on Windows in Debug mode:
```
colcon build --merge-install --event-handlers console_cohesion+ console_package_list+ --cmake-args -DBUILD_TESTING=ON --no-warn-unused-cli -DCMAKE_BUILD_TYPE=Debug -DINSTALL_EXAMPLES=OFF -DSECURITY=ON --packages-up-to tf2_ros tf2_py
```

Run on Windows in Debug mode (you may have to do this several times to make the issue happen):
```
colcon test --merge-install --event-handlers console_direct+ --executor sequential --retest-until-pass 10 --packages-select tf2_ros
```

#### Expected behavior

Test passes every time.

#### Actual behavior

Test fails sometimes with "Goal handle is not known to this client"

jacobperron:
Fixes #861

It is against the design of ROS actions to rely on the status topic for the core implementation,
instead it should just be used for introspection.

Rather than relying on the status topic to remove references to goal handles, the action client
instead holds weak pointers to the goal handles. This way, as long as a user holds a reference to
the goal handle they can use it to interact with the action client.

~~Draft PR for now as I try to determine if this actually fixes the issue (hopefully with a reliable reproduction of the bug).~~

For a reproduction, see https://github.com/ros2/rclcpp/issues/861#issuecomment-631762189

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -331,7 +331,9 @@ public:
    * If the goal is accepted by an action server, the returned future is set to a `ClientGoalHandle`.
    * If the goal is rejected by an action server, then the future is set to a `nullptr`.
    *
-   * The goal handle is used to monitor the status of the goal and get the final result.
+   * The returned goal handle is used to monitor the status of the goal and get the final result.
+   * It is valid as long as you hold a reference to the shared pointer or until the
+   * rclcpp_action::Client is destroyed at which point the goal status will become UNKNOWN.
    *
    * \param[in] goal The goal request.
    * \param[in] options Options for sending the goal request. Contains references to callbacks for
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -342,20 +342,6 @@ public:
   std::shared_future<typename GoalHandle::SharedPtr>
   async_send_goal(const Goal & goal, const SendGoalOptions & options = SendGoalOptions())
   {
-    // To prevent the list from growing out of control, forget about any goals
-    // with no more user references
-    {
-      std::lock_guard<std::mutex> guard(goal_handles_mutex_);
-      auto goal_handle_it = goal_handles_.begin();
-      while (goal_handle_it != goal_handles_.end()) {
-        if (!goal_handle_it->second.lock()) {
-          goal_handle_it = goal_handles_.erase(goal_handle_it);
-        } else {
-          ++goal_handle_it;
-        }
-      }
-    }
-
     // Put promise in the heap to move it around.
     auto promise = std::make_shared<std::promise<typename GoalHandle::SharedPtr>>();
     std::shared_future<typename GoalHandle::SharedPtr> future(promise->get_future());
@@ -400,6 +386,21 @@ public:
           }
         }
       });
+
+    // To prevent the list from growing out of control, forget about any goals
+    // with no more user references
+    {
+      std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+      auto goal_handle_it = goal_handles_.begin();
+      while (goal_handle_it != goal_handles_.end()) {
+        if (!goal_handle_it->second.lock()) {
+          goal_handle_it = goal_handles_.erase(goal_handle_it);
+        } else {
+          ++goal_handle_it;
+        }
+      }
+    }
+
     return future;
   }
 
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -387,6 +387,8 @@ public:
         }
       });
 
+    // TODO(jacobperron): Encapsulate into it's own function and
+    //                    consider exposing an option to disable this cleanup
     // To prevent the list from growing out of control, forget about any goals
     // with no more user references
     {
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -342,6 +342,20 @@ public:
   std::shared_future<typename GoalHandle::SharedPtr>
   async_send_goal(const Goal & goal, const SendGoalOptions & options = SendGoalOptions())
   {
+    // To prevent the list from growing out of control, forget about any goals
+    // with no more user references
+    {
+      std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+      auto goal_handle_it = goal_handles_.begin();
+      while (goal_handle_it != goal_handles_.end()) {
+        if (!goal_handle_it->second.lock()) {
+          goal_handle_it = goal_handles_.erase(goal_handle_it);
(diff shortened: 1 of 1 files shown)


rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -342,6 +342,20 @@ class Client : public ClientBase
   std::shared_future<typename GoalHandle::SharedPtr>
   async_send_goal(const Goal & goal, const SendGoalOptions & options = SendGoalOptions())
   {
+    // To prevent the list from growing out of control, forget about any goals
```
jacobperron, line 345: I'm not a fan of cleaning up goal handles in the send goal function, but I wasn't sure of a better place to put this logic (or an alternative to ensure we don't grow the map of goal handles indefinitely).

SteveMacenski, line 345: I would be happier with this if:
- It were a separate, well named function that could be called by the object using the client to clear itself if it wanted to
- This cleaning up process, if happening in the send goal command, happened at the end of it so it didn't add any additional latency to sending the command, even if the client thread had a little latency for returning. That lock makes me nervous. 
- add a default to `true` parameter in async_send_goal whether or not it wants this method to do the housekeeping for you. Maybe someone wants to do it themselves.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -546,7 +563,12 @@ class Client : public ClientBase
         "Received feedback for unknown goal. Ignoring...");
       return;
     }
-    typename GoalHandle::SharedPtr goal_handle = goal_handles_[goal_id];
+    typename GoalHandle::SharedPtr goal_handle = goal_handles_[goal_id].lock();
+    // Forget about the goal if there are no more user references
+    if (!goal_handle) {
+      goal_handles_.erase(goal_id);
```
SteveMacenski, line 569: This should print something - error / info / debug. Something to indicate that you wrongfully got feedback

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -575,16 +597,13 @@ class Client : public ClientBase
...
-      goal_handle->set_status(status.status);
-      const int8_t goal_status = goal_handle->get_status();
-      if (
-        goal_status == GoalStatus::STATUS_SUCCEEDED ||
-        goal_status == GoalStatus::STATUS_CANCELED ||
-        goal_status == GoalStatus::STATUS_ABORTED)
-      {
+      typename GoalHandle::SharedPtr goal_handle = goal_handles_[goal_id].lock();
+      // Forget about the goal if there are no more user references
+      if (!goal_handle) {
```
SteveMacenski, line 602: I like this so much better!

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: I don't think this is a "wrong" case. It's possible for a user to "fire-and-forget" a goal and I think we would be in this case. I.e. The user send a goal, but doesn't bother retaining the goal handle (maybe they don't care about a result or feedback). Logging a debug message whenever we remove a goal handle isn't a bad idea though.

SteveMacenski: That would also work.

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: > It were a separate, well named function that could be called by the object using the client to clear itself if it wanted to

Good idea. But in order to try and get this fix into Foxy, I think I'll just add a TODO to expose a function. This way we avoid adding/modifying the existing API.

> add a default to `true` parameter in async_send_goal whether or not it wants this method to do the housekeeping for you.

Ditto. I'd like to postpone changes to API until post-Foxy.

> This cleaning up process, if happening in the send goal command, happened at the end of it so it didn't add any additional latency to sending the command, even if the client thread had a little latency for returning. That lock makes me nervous.

I can do this.

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: Logging debug in d06e8e0

jacobperron:
Testing packages above rclcpp_action:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10850/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6220/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8838/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10744/)

jacobperron:
I'll still have to update the examples that aren't already maintaining a reference to the goal handle. Otherwise, I expect them to break as pointed out in https://github.com/ros2/rclcpp/issues/861#issuecomment-596426349.

jacobperron:
So, it turns out this change didn't break our examples, because of a subtle way the lifetime of the goal handle is extended. If the user registers a result callback when sending the goal, then the shared_ptr is kept alive by the action client by passing it to this lambda:

https://github.com/ros2/rclcpp/blob/d06e8e08545d93c905a5d9591c4f1fc26a77268e/rclcpp_action/include/rclcpp_action/client.hpp#L633-L636

I guess this is fine, though I think it's worth adding some documentation regarding the lifetime of the goal handle.

PR #1123, opened 2020-05-21
Deprecate set_on_parameters_set_callback

### LINKED ISSUE #791: Deprecate `set_on_parameters_set_callback`
#### Feature description

In https://github.com/ros2/rclcpp/pull/772, new methods that allow registering multiple callbacks were added.
The old `set_on_parameters_set_callback` should be deprecated.

claireyywang:
This PR deprecates `set_on_parameters_set_callback` and replaces it with the new `add_on_set_parameters_callback` and `remove_on_set_parameters_callback` functions. Closes #791 . 

P.S. I added `add_on_set_parameters_callback` and `remove_on_set_parameters_callback` to `lifecycle_node` cuz they weren't there, and I assumed we'd want to deprecate `set_on_parameters_set_callback` in there as well. #1134 needs to merged first.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -277,7 +277,7 @@ public:
    * If `ignore_override` is `true`, the parameter override will be ignored.
    *
    * This method, if successful, will result in any callback registered with
-   * set_on_parameters_set_callback to be called.
+   * add_on_set_parameters_callback to be called.
    * If that callback prevents the initial value for the parameter from being
    * set then rclcpp::exceptions::InvalidParameterValueException is thrown.
    *
@@ -359,7 +359,7 @@ public:
    * by the function call will be ignored.
    *
    * This method, if successful, will result in any callback registered with
-   * set_on_parameters_set_callback to be called, once for each parameter.
+   * add_on_set_parameters_callback to be called, once for each parameter.
    * If that callback prevents the initial value for any parameter from being
    * set then rclcpp::exceptions::InvalidParameterValueException is thrown.
    *
@@ -401,7 +401,7 @@ public:
   /// Undeclare a previously declared parameter.
   /**
    * This method will not cause a callback registered with
-   * set_on_parameters_set_callback to be called.
+   * add_on_set_parameters_callback to be called.
    *
    * \param[in] name The name of the parameter to be undeclared.
    * \throws rclcpp::exceptions::ParameterNotDeclaredException if the parameter
@@ -435,7 +435,7 @@ public:
    * Parameter overrides are ignored by set_parameter.
    *
    * This method will result in any callback registered with
-   * set_on_parameters_set_callback to be called.
+   * add_on_set_parameters_callback to be called.
    * If the callback prevents the parameter from being set, then it will be
    * reflected in the SetParametersResult that is returned, but no exception
    * will be thrown.
@@ -476,7 +476,7 @@ public:
    * corresponding SetParametersResult in the vector returned by this function.
    *
    * This method will result in any callback registered with
-   * set_on_parameters_set_callback to be called, once for each parameter.
+   * add_on_set_parameters_callback to be called, once for each parameter.
    * If the callback prevents the parameter from being set, then, as mentioned
    * before, it will be reflected in the corresponding SetParametersResult
    * that is returned, but no exception will be thrown.
@@ -507,7 +507,7 @@ public:
    * If the exception is thrown then none of the parameters will have been set.
    *
    * This method will result in any callback registered with
-   * set_on_parameters_set_callback to be called, just one time.
+   * add_on_set_parameters_callback to be called, just one time.
    * If the callback prevents the parameters from being set, then it will be
    * reflected in the SetParametersResult which is returned, but no exception
    * will be thrown.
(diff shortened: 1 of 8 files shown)


claireyywang:
Some questions to resolve before this PR is good to review:

- Currently the deprecation warning is not suppressed yet in `test_node.cpp`. All the directives I found seems to be specific to different compilers. Is there a generic one I can use for all platforms?

- Some documentation for functions like `declare_parameter`, `set_parameter` in `node.hpp` still says 

  > This method will result in any callback registered with set_on_parameters_set_callback to be called."

  Should I update them to use `add_on_set_parameters_callback` instead?

@jacobperron :smiley:

jacobperron:
I'm not aware of a generic way to suppress the warnings; it will vary by compiler. I think you'll have to do something like this before using the deprecated API:

https://github.com/ros2/rclcpp/blob/64bdef61c807bed3750ab21dac2278164f5b44c6/rclcpp/test/test_init.cpp#L20-L26

and then restore the state afterwards:

https://github.com/ros2/rclcpp/blob/64bdef61c807bed3750ab21dac2278164f5b44c6/rclcpp/test/test_init.cpp#L62-L66

Regarding the docs, I think they could be updated as you suggest :+1:

jacobperron:
We should also add a note to the Foxy release notes (there exists a similar note for `rclpy` already).

rclcpp/test/test_node.cpp
```diff
@@ -1377,6 +1376,7 @@ TEST_F(TestNode, set_parameters_undeclared_parameters_not_allowed) {
     EXPECT_EQ(node->get_parameter(name1).get_value<int>(), 2);
     EXPECT_EQ(node->get_parameter(name2).get_value<bool>(), true);  // old value
     EXPECT_EQ(node->get_parameter(name3).get_value<std::string>(), "red");
+    RCLCPP_SCOPE_EXIT({node->remove_on_set_parameters_callback(handler.get());});    // always reset
```
jacobperron, line 1379: I would move this directly after the callback is added above. This way an unexpected failure after calling `add_on_set_parameters_callback` won't prevent this remove function from being called. Also, having them close together makes it easier to associate them with each other.

Same comment applies to other places where the scoped exit macro is not next to the `add_on_set_parameters_callback` call.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -447,13 +447,33 @@ class LifecycleNode : public node_interfaces::LifecycleNodeInterface,
...
   using OnParametersSetCallbackType =
     rclcpp::node_interfaces::NodeParametersInterface::OnParametersSetCallbackType;
 
+  /// Add a callback for when parameters are being set.
+  /**
+   * \sa rclcpp::Node::add_on_set_parameters_callback
+   */
+  RCLCPP_LIFECYCLE_PUBLIC
+  rclcpp_lifecycle::LifecycleNode::OnSetParametersCallbackHandle::SharedPtr
+  add_on_set_parameters_callback(OnParametersSetCallbackType callback);
```
jacobperron, line 461: Ah, it's unfortunate we missed adding this for lifecycle nodes when the feature was added. It would be better to add these functions to lifecycle in a separate PR (prior to merging this PR), but I'm not going to be sticky on that point.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -851,6 +852,7 @@ class Node : public std::enable_shared_from_this<Node>
    * \return The previous callback that was registered, if there was one,
    *   otherwise nullptr.
    */
+  [[deprecate("use add_on_set_parameters_callback(OnParametersSetCallbackType callback) instead")]]
```
jacobperron, line 855: The correct attribute name is [deprecated](https://en.cppreference.com/w/cpp/language/attributes/deprecated).

```suggestion
  [[deprecated("use add_on_set_parameters_callback(OnParametersSetCallbackType callback) instead")]]
```

I missed it, but the compiler didn't.

claireyywang:
> LGTM

Im merging the changes from #1134 to run CI

claireyywang:
CI on all ros2 packages
linux [Build Status](https://ci.ros2.org/job/ci_linux/10898/)
linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6252/)
osx [Build Status](https://ci.ros2.org/job/ci_osx/8874/)
windows [Build Status](https://ci.ros2.org/job/ci_windows/10789/)

claireyywang:
merging since failed tests are consistent with nightly

ivanpauno:
Between all the `rclcpp_lifecycle` failures preexistent to this PR (being fixed by https://github.com/ros2/rcl/pull/662), a new one was introduced here:

* https://ci.ros2.org/job/ci_linux/10898/testReport/junit/rclcpp_lifecycle/TestDefaultStateMachine/check_parameters/ (from the job run in https://github.com/ros2/rclcpp/pull/1123#issuecomment-634327749)
* In nightlies: https://ci.ros2.org/view/nightly/job/nightly_linux_release/1555/testReport/junit/rclcpp_lifecycle/TestDefaultStateMachine/check_parameters/.

See https://github.com/ros2/rclcpp/pull/1136 for a fix.

Issue #1042, opened 2020-03-31
rclcpp::Node constructor and destructor crash with multithreading


rotu:
Use of `rclcpp::Node` is unsafe in an asynchronous setting and can cause segfaults and undefined behavior if such objects are constructed and/or destructed simultaneously.

See https://github.com/ros2/rosbag2/issues/329 where this directly caused a reproducible crash. While the particular crash was averted in https://github.com/ros2/rosbag2/pull/338, the constructor/destructor still use the unsafe functions `rcl_node_init`/`rcl_node_fini`.

Either these functions should be made threadsafe or they should be documented as unsafe and audited/instrumented to detect such unsafe usage.

rotu:
@thomas-moulard re: https://github.com/ros-tooling/aws-oncall/issues/107 (can't comment on my own ticket in aws-oncall repo)

fujitatomoya:
@rotu 

does anybody work on the fix?

rotu:
As far as I can tell, no. @dirk-thomas is listed as the package maintainer so maybe he knows.

ivanpauno:
@rotu how it was concluded that the thread safety issue is in `Node` constructor/destructor?
From the information available in https://github.com/ros2/rosbag2/issues/329, it's not obvious to me.
Can you clarify how that conclusion was achieved?

I will try to track down the issue, but it would be great to have clearer information (and preferably, a simple example with reproduction steps).

dirk-thomas:
> @dirk-thomas is listed as the package maintainer

While I am listed in the manifest this repo is maintainer by the whole @ros2/team.

ivanpauno:
I think I understand the issue:

- `rcl_node_init` calls [rcl_logging_rosout_init_publisher_for_node](https://github.com/ros2/rcl/blob/85ef95514d6ee47bb8285009add84e956595c018/rcl/src/rcl/node.c#L303)
- `rcl_logging_rosout_init_publisher_for_node` [modifies](https://github.com/ros2/rcl/blob/94b5a1d7d0899aa84a1026e21488bee95e67bbd8/rcl/src/rcl/logging_rosout.c#L188) some [global state](https://github.com/ros2/rcl/blob/94b5a1d7d0899aa84a1026e21488bee95e67bbd8/rcl/src/rcl/logging_rosout.c#L83).

We really should get rid of those global variables in `rcl` ...

[...]

PR #1124, opened 2020-05-21
protect rcl_node_init()/fini() with rcl logging mutex.

fujitatomoya:
fix candidate for https://github.com/ros2/rclcpp/issues/1042


rclcpp/include/rclcpp/context.hpp
```diff
@@ -320,6 +320,11 @@ class Context : public std::enable_shared_from_this<Context>
     return sub_context;
   }
 
+  /// Return the global logging configure mutex.
+  RCLCPP_PUBLIC
+  std::shared_ptr<std::mutex>
+  get_rcl_logging_configure_mutex();
```
rotu, line 326: The shared_ptr seems unnecessary

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -65,10 +65,13 @@ NodeBase::NodeBase(
   // Create the rcl node and store it in a shared_ptr with a custom destructor.
   std::unique_ptr<rcl_node_t> rcl_node(new rcl_node_t(rcl_get_zero_initialized_node()));
 
-  ret = rcl_node_init(
-    rcl_node.get(),
-    node_name.c_str(), namespace_.c_str(),
-    context_->get_rcl_context().get(), &rcl_node_options);
+  {
+    std::lock_guard<std::mutex> guard(*(context_->get_rcl_logging_configure_mutex()));
```
rotu, line 69: There are two subtle bugs here due to using a shared_ptr:

1. What if it's null?
2. What if the object is deleted after the shared_ptr is deterrence?

rclcpp/include/rclcpp/context.hpp
rotu: This method should be marked const since it shouldn't mutate the Context

ivanpauno: I choose to hold a `shared_ptr` to the [mutex](https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/include/rclcpp/context.hpp#L342) in https://github.com/ros2/rclcpp/pull/998 because of the following reason:

- We have the [default context]((https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/include/rclcpp/contexts/default_context.hpp#L48)), with static lifetime.
- The [global logging mutex](https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/src/rclcpp/context.cpp#L42) also have static lifetime.
- According to the standard, objects with static lifetime created in function scope in a single translation unit are created on first use, and destructed in the opposite order they well created. There's no ordering ensured for objects in different translation units.

Then imagine the following situation:
- An user has a program that uses two contexts: the default one and another one with static lifetime.
- One of both contexts get initialized first (doesn't matter which one), [get_global_logging_configure_mutex](https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/src/rclcpp/context.cpp#L40) is called and then the mutex is initialized (on first use).
- The other context is created afterwards.
- As destruction ordering is not ensured between different translation units, a possible destruction order is:
  - context1
  - mutex
  - context2

In case `shutdown` for both contexts wasn't explicitly called the program, the `Context` destructor will call it and that [takes the logging mutex](https://github.com/ros2/rclcpp/blob/2aaeee72a6efa3475837245869eb4128f41a57fd/rclcpp/src/rclcpp/context.cpp#L231), which can segfault if the mutex was previously destroyed.

That's the long story of why using the shared pointer ...

Having say that, I should've added a comment in the code. The rationale is not obvious at all.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
ivanpauno: > What if it's null?

We should add a check and handle that case correctly.

> What if the object is deleted after the shared_ptr is deterrence?

If the `Node` holds the shared pointer, that's avoided.

ivanpauno:
We decided to finally add mutual exclusion protection in rclcpp/rclpy instead of rcl, https://github.com/ros2/rcl/pull/660 is a start on that path.

We're near to release date and we are running out of time, for that reason we will continue the fix proposal within the ros2 team to streamline the implementation and iterate on reviews faster.

I will probably borrow some of your code @fujitatomoya, thanks for starting a fix!

fujitatomoya:
@ivanpauno @rotu 

thanks for catching up!!!

PR #1125, opened 2020-05-21
Fix thread safety issues related to logging

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1042.
Replaces https://github.com/ros2/rclcpp/pull/1124.
Kind of a continuation of the work done in https://github.com/ros2/rcl/issues/551.

Depends on https://github.com/ros2/rcl/pull/660 (merged).
Depends on https://github.com/ros2/rcutils/pull/253 (merged).

As far as I can tell this completely fixes the thread safety issue, but it needs a thorough review.

### DIFF
--- a/rclcpp/src/rclcpp/context.cpp
+++ b/rclcpp/src/rclcpp/context.cpp
@@ -67,6 +67,7 @@ rclcpp_logging_output_handler(
       location, severity, name, timestamp, format, args);
   } catch (std::exception & ex) {
     RCUTILS_SAFE_FWRITE_TO_STDERR(ex.what());
+    RCUTILS_SAFE_FWRITE_TO_STDERR("\n");
   } catch (...) {
     RCUTILS_SAFE_FWRITE_TO_STDERR("failed to take global rclcpp logging mutex\n");
   }
--- a/rclcpp/src/rclcpp/context.cpp
+++ b/rclcpp/src/rclcpp/context.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015 Open Source Robotics Foundation, Inc.
+// Copyright 2015-2020 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -23,10 +23,17 @@
 
 #include "rcl/init.h"
 #include "rcl/logging.h"
+
 #include "rclcpp/detail/utilities.hpp"
 #include "rclcpp/exceptions.hpp"
 #include "rclcpp/logging.hpp"
+
+#include "rcutils/error_handling.h"
+#include "rcutils/macros.h"
+
 #include "rmw/impl/cpp/demangle.hpp"
+
+#include "./logging_mutex.hpp"
 
 /// Mutex to protect initialized contexts.
 static std::mutex g_contexts_mutex;
@@ -35,13 +42,27 @@ static std::vector<std::weak_ptr<rclcpp::Context>> g_contexts;
 
 using rclcpp::Context;
 
-static
-std::shared_ptr<std::mutex>
-get_global_logging_configure_mutex()
+extern "C"
 {
-  static auto mutex = std::make_shared<std::mutex>();
-  return mutex;
+void
+rclcpp_logging_output_handler(
+  const rcutils_log_location_t * location,
+  int severity, const char * name, rcutils_time_point_value_t timestamp,
+  const char * format, va_list * args)
+{
+  std::shared_ptr<std::recursive_mutex> logging_mutex = nullptr;
+  try {
+    logging_mutex = get_global_logging_mutex();
+  } catch (std::exception & ex) {
+    RCUTILS_SAFE_FWRITE_TO_STDERR(ex.what());
+    return;
+  }
+
+  std::lock_guard<std::recursive_mutex> guard(*logging_mutex);
+  return rcl_logging_multiple_output_handler(
+    location, severity, name, timestamp, format, args);
 }
+}  // extern "C"
 
 static
 size_t &
@@ -54,7 +75,7 @@ get_logging_reference_count()
 Context::Context()
 : rcl_context_(nullptr),
   shutdown_reason_(""),
-  logging_configure_mutex_(nullptr)
+  logging_mutex_(nullptr)
 {}
 
 Context::~Context()
@@ -116,16 +137,14 @@ Context::init(
   }
 
   if (init_options.auto_initialize_logging()) {
-    logging_configure_mutex_ = get_global_logging_configure_mutex();
-    if (!logging_configure_mutex_) {
-      throw std::runtime_error("global logging configure mutex is 'nullptr'");
-    }
-    std::lock_guard<std::mutex> guard(*logging_configure_mutex_);
+    logging_mutex_ = get_global_logging_mutex();
+    std::lock_guard<std::recursive_mutex> guard(*logging_mutex_);
     size_t & count = get_logging_reference_count();
     if (0u == count) {
-      ret = rcl_logging_configure(
+      ret = rcl_logging_configure_with_output_handler(
         &rcl_context_->global_arguments,
-        rcl_init_options_get_allocator(init_options_.get_rcl_init_options()));
+        rcl_init_options_get_allocator(init_options_.get_rcl_init_options()),
+        rclcpp_logging_output_handler);
       if (RCL_RET_OK != ret) {
         rcl_context_.reset();
         rclcpp::exceptions::throw_from_rcl_error(ret, "failed to configure logging");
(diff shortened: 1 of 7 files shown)


rclcpp/src/rclcpp/context.cpp
```diff
@@ -23,25 +23,46 @@
...
 
-static
-std::shared_ptr<std::mutex>
-get_global_logging_configure_mutex()
+extern "C"
 {
-  static auto mutex = std::make_shared<std::mutex>();
-  return mutex;
+void
+rclcpp_logging_output_handler(
```
fujitatomoya, line 48: this is intended to be called from somewhere else? if that so, interface should be exposed. if not, make it static?

ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1125/commits/bfc98d0e0be5fc89b3bc15be01885e4434f4fc48.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -23,25 +23,47 @@
...
-get_global_logging_configure_mutex()
+void
+rclcpp_logging_output_handler(
+  const rcutils_log_location_t * location,
+  int severity, const char * name, rcutils_time_point_value_t timestamp,
+  const char * format, va_list * args)
 {
-  static auto mutex = std::make_shared<std::mutex>();
-  return mutex;
+  std::shared_ptr<std::recursive_mutex> logging_mutex = nullptr;
```
hidmic, line 54: @ivanpauno nit^N: `nullptr` assignment is redundant with `std::shared_ptr` default constructor. Feel free to leave as-is if you think it helps readability.

ivanpauno: Solved in https://github.com/ros2/rclcpp/pull/1125/commits/8630accf5c8ca7ec688684b51a86bfaeec8700b8.

ivanpauno:
When https://github.com/ros/rosdistro/pull/25019 and https://github.com/ros/rosdistro/pull/25018 get merged I will re-run the checker.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10867/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6232/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/8853/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10758/)

ivanpauno:

ivanpauno:

rclcpp/src/rclcpp/context.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 Open Source Robotics Foundation, Inc.
+// Copyright 2015-2020 Open Source Robotics Foundation, Inc.
```
wjwwood, line 1: Why? I mean it is allowed, but it seems done so inconsistently through our packages, even within this pull request. Is it useful?

rclcpp/src/rclcpp/context.cpp
```diff
@@ -23,25 +23,47 @@
...
 {
-  static auto mutex = std::make_shared<std::mutex>();
-  return mutex;
+  std::shared_ptr<std::recursive_mutex> logging_mutex;
+  try {
+    logging_mutex = get_global_logging_mutex();
+  } catch (std::exception & ex) {
+    RCUTILS_SAFE_FWRITE_TO_STDERR(ex.what());
+    return;
+  }
```
wjwwood, line 60: I think this needs a `catch (...)` statement too.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -23,25 +23,47 @@
...
+void
+rclcpp_logging_output_handler(
+  const rcutils_log_location_t * location,
+  int severity, const char * name, rcutils_time_point_value_t timestamp,
+  const char * format, va_list * args)
 {
-  static auto mutex = std::make_shared<std::mutex>();
-  return mutex;
+  std::shared_ptr<std::recursive_mutex> logging_mutex;
+  try {
```
wjwwood, line 55: Better put the whole function into the try-catch. Throwing from this function is not allowed.

ivanpauno:
Though the rosdistro PRs mentioned [previously](https://github.com/ros2/rclcpp/pull/1125#issuecomment-632747777) were merged, the checker is still using the previous version of the rcl/rcutils debians, and thus it is failing.
I'm not sure of how/when debians are synchronized.

If CI is green, that should be enough testing.

rclcpp/src/rclcpp/logging_mutex.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * This mutex is locked in the following situations:
+ *   - In initialization/destruction of contexts.
+ *   - In initialization/destruction of nodes.
+ *   - In the rcl logging output handler installed by rclcpp,
+ *     i.e.: in all calls to the logger macros, including RCUTILS_* ones.
+ */
+// Implementation detail:
+// A shared pointer to the mutex is used, so that objects that need to use
+// it at destruction time can hold it alive.
+// In that way, a destruction ordering problem between static objects is avoided.
```
wjwwood, line 34: That's a nice thought, but actually it will not avoid it, but instead will throw. For example, if the static in this function is destroyed and then you attempt to shutdown logging, an exception is thrown instead.

wjwwood: Instead we may need to have logic that allows shutdown to happen without a mutex lock in the event that this function returns nullptr? And also causes log functions to fail without trying to log if this returns nullptr? That should allow this to be destroyed first, and then at least let logging shutdown gracefully-ish.

ivanpauno: > That's a nice thought, but actually it will not avoid it, but instead will throw. For example, if the static in this function is destroyed and then you attempt to shutdown logging, an exception is thrown instead.

None of the destructors (NodeBase/Context) is calling `get_global_logging_mutex`. This function is only being called at construction time, ensuring that at that point the static shared pointer will be alive.

In all places where the mutex is needed at destruction time, a copy of the shared pointer that was made a construction time is being used. That copy of the shared pointer is hold by the object that needs to lock the mutex in its destructor, ensuring it will still be alive (if the shared pointer with static lifetime in `get_global_logging_mutex` was already destructed, that won't be a problem as another owner is keeping the mutex alive).

I cannot see the problem you're pointing out. If you think this is a problem, please be more specific in how the use-after-destruction can happen.

rclcpp/src/rclcpp/context.cpp
ivanpauno: I agree we don't consistently do this, but I remember having read that we should.
If you think it's a problem, I can revert the change.

rclcpp/src/rclcpp/context.cpp
ivanpauno: Fixed in https://github.com/ros2/rclcpp/pull/1125/commits/b29cdd698b0b8778d036b73a55f49b976b37fca4.

rclcpp/src/rclcpp/context.cpp
ivanpauno: See https://github.com/ros2/rclcpp/pull/1125/commits/b29cdd698b0b8778d036b73a55f49b976b37fca4.

rclcpp/src/rclcpp/logging_mutex.hpp
wjwwood: Sorry, I got bamboozled by misreading the github view, haha:

image

I thought it was being used in the context destructor.

You're right, I think this ought to work then.

rclcpp/src/rclcpp/context.cpp
wjwwood: I don't think it is necessary, but I also don't care if you leave it. It's just you don't add it to other files you touched in this pull request, so I don't know when and when not to add it.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -23,38 +23,60 @@
...
+  const char * format, va_list * args)
+{
+  try {
+    std::shared_ptr<std::recursive_mutex> logging_mutex;
+    logging_mutex = get_global_logging_mutex();
+    std::lock_guard<std::recursive_mutex> guard(*logging_mutex);
+    return rcl_logging_multiple_output_handler(
+      location, severity, name, timestamp, format, args);
+  } catch (std::exception & ex) {
+    RCUTILS_SAFE_FWRITE_TO_STDERR(ex.what());
```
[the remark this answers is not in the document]
wjwwood: The new line on the new message reminded me of this.

ivanpauno:
Github is not working pretty well, but let's see if I'm lucky:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10871/)

rclcpp/src/rclcpp/logging_mutex.cpp
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include <mutex>
+
+#include "rcutils/macros.h"
+
+std::shared_ptr<std::recursive_mutex>
+get_global_logging_mutex()
```
wjwwood, line 21: I just noticed that this does not include the header file that defines this symbol, should it not?

PR #1140, opened 2020-05-28
`SubscriptionBase::get_subscription_handle() const` should return a shared pointer to const value

ivanpauno:
In https://github.com/ros2/rclcpp/pull/431, `const rcl_subscription_t *` was incorrectly modified to `const std::shared_ptr<rcl_subscription_t>`.

This fixes the issue.

### DIFF
--- a/rclcpp/include/rclcpp/subscription_base.hpp
+++ b/rclcpp/include/rclcpp/subscription_base.hpp
@@ -91,7 +91,7 @@ public:
   get_subscription_handle();
 
   RCLCPP_PUBLIC
-  const std::shared_ptr<rcl_subscription_t>
+  std::shared_ptr<const rcl_subscription_t>
   get_subscription_handle() const;
 
   /// Get all the QoS event handlers associated with this subscription.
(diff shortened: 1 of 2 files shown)


ivanpauno:
Jobs config:
* Build: everything
* Test: rclcpp test_rclcpp

Jobs:
* Linux: [Build Status](http://ci.ros2.org/job/ci_linux/10921/)

ivanpauno:
> It's reasonable enough to me, but I'll suggest we only merge this for Foxy if we also merge #1119 .

I think we should merge #1119 for Foxy, it is (almost) impossible to backport a fix of that.

Talking about that, I really think we should start using PIMPL extensively in `rclcpp`.
IMO, we have run many times in the "I cannot backport a ABI compatible fix" issue.
I will propose that for G-Turtle.

ivanpauno:
Jobs config:
* Build: rclcpp
* Test: rclcpp

Jobs:
* AArch64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6273/)
* osX: [Build Status](http://ci.ros2.org/job/ci_osx/8895/)
* Windows: [Build Status](http://ci.ros2.org/job/ci_windows/10825/)

PR #1141, opened 2020-05-28
Pass shared pointer by value instead than by const & when possible

ivanpauno:
According to the cpp core guidelines (which imo is the best guideline about smart pointer signatures), if the function always retains shared ownership of a `std::shared_ptr`, then it must be passed by value.
See [R.34](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r34-take-a-shared_ptrwidget-parameter-to-express-that-a-function-is-part-owner) and [R.36](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r36-take-a-const-shared_ptrwidget-parameter-to-express-that-it-might-retain-a-reference-count-to-the-object-).

This breaks ABI and not API, so I don't think there are much chances of breaking downstream packages (this functionality is also pretty new and there is not much external usage, it was never part of a ros distribution).

### DIFF
--- a/rclcpp/include/rclcpp/wait_set_policies/static_storage.hpp
+++ b/rclcpp/include/rclcpp/wait_set_policies/static_storage.hpp
@@ -17,6 +17,7 @@
 
 #include <array>
 #include <memory>
+#include <utility>
 
 #include "rclcpp/client.hpp"
 #include "rclcpp/guard_condition.hpp"
--- a/rclcpp/include/rclcpp/wait_set_policies/static_storage.hpp
+++ b/rclcpp/include/rclcpp/wait_set_policies/static_storage.hpp
@@ -60,9 +60,9 @@ public:
 
     /// Conversion constructor, which is intentionally not marked explicit.
     SubscriptionEntry(
-      const std::shared_ptr<rclcpp::SubscriptionBase> & subscription_in = nullptr,
-      const rclcpp::SubscriptionWaitSetMask & mask_in = {})
-    : subscription(subscription_in),
+      std::shared_ptr<rclcpp::SubscriptionBase> subscription_in = nullptr,
+      rclcpp::SubscriptionWaitSetMask mask_in = {})
+    : subscription(std::move(subscription_in)),
       mask(mask_in)
     {}
   };
@@ -100,10 +100,10 @@ public:
   {
     /// Conversion constructor, which is intentionally not marked explicit.
     WaitableEntry(
-      const std::shared_ptr<rclcpp::Waitable> & waitable_in = nullptr,
-      const std::shared_ptr<void> & associated_entity_in = nullptr) noexcept
-    : waitable(waitable_in),
-      associated_entity(associated_entity_in)
+      std::shared_ptr<rclcpp::Waitable> waitable_in = nullptr,
+      std::shared_ptr<void> associated_entity_in = nullptr) noexcept
+    : waitable(std::move(waitable_in)),
+      associated_entity(std::move(associated_entity_in))
     {}
 
     std::shared_ptr<rclcpp::Waitable> waitable;
--- a/rclcpp/include/rclcpp/wait_set_policies/dynamic_storage.hpp
+++ b/rclcpp/include/rclcpp/wait_set_policies/dynamic_storage.hpp
@@ -52,9 +52,9 @@ public:
 
     /// Conversion constructor, which is intentionally not marked explicit.
     SubscriptionEntry(
-      const std::shared_ptr<rclcpp::SubscriptionBase> & subscription_in = nullptr,
+      std::shared_ptr<rclcpp::SubscriptionBase> subscription_in = nullptr,
       const rclcpp::SubscriptionWaitSetMask & mask_in = {})
-    : subscription(subscription_in),
+    : subscription(std::move(subscription_in)),
       mask(mask_in)
     {}
 
@@ -117,10 +117,10 @@ public:
 
     /// Conversion constructor, which is intentionally not marked explicit.
     WaitableEntry(
-      const std::shared_ptr<rclcpp::Waitable> & waitable_in = nullptr,
-      const std::shared_ptr<void> & associated_entity_in = nullptr) noexcept
-    : waitable(waitable_in),
-      associated_entity(associated_entity_in)
+      std::shared_ptr<rclcpp::Waitable> waitable_in = nullptr,
+      std::shared_ptr<void> associated_entity_in = nullptr) noexcept
+    : waitable(std::move(waitable_in)),
+      associated_entity(std::move(associated_entity_in))
     {}
 
     void


hidmic:
> But it equally might cause us to have crashes, since std::move on the shared_ptr causes the caller pointer to go to nullptr.

How? In this patch `std::move` makes the argument go to `nullptr`, but it won't affect the caller.

clalancette:
> How? In this patch `std::move` makes the argument go to `nullptr`, but it won't affect the caller.

I'm new to this idiom, so bear with me.  But running the example program below:

```
#include <cstdio>
#include <memory>

std::shared_ptr<int> g_int;

void store_shared(std::shared_ptr<int> newint)
{
  g_int = newint;
}

int main()
{
  auto i = std::make_shared<int>(25);
  fprintf(stderr, "Count: %ld, pointer %p\n", i.use_count(), (void *)i.get());
  store_shared(std::move(i));
  fprintf(stderr, "Count: %ld, pointer %p\n", i.use_count(), (void *)i.get());
  return 0;
}
```

I get output:

```
Count: 1, pointer 0xfe2ec0
Count: 0, pointer (nil)
```

That seems to suggest to me that the caller's pointer gets set to nullptr, no?

hidmic:
> That seems to suggest to me that the caller's pointer gets set to nullptr, no?

Because in that example `std::move` is in the caller scope. It wouldn't be the case if it was within `store_shared()`.

clalancette:
> Because in that example `std::move` is in the caller scope. It wouldn't be the case if it was within `store_shared()`.

OK, I see.  But if I put it into the `store_shared` scope, then it doesn't get set to nil, but the reference count also gets increased (just like the const ref code that is already there).

So that suggests to me that this change is actually a no-op unless both the caller and callee does `std::move`.  Or am I wrong about that?

ivanpauno:
@clalancette A copy is being made at call time, because the `shared_ptr` is being passed by value.
Then, it's being moved to the place we want to store it, which is safe as the argument is not used again.

That's the recommended approach for a method that always takes ownership of the passed shared pointer, and passing a `const &` is explicitly not recommended (there's actually no problem with it, it's just less idiomatic).

hidmic:
> So that suggests to me that this change is actually a no-op

It is a no-op, to pass by reference and copy-construct vs. to pass by value and move-construct. The thing that @ivanpauno, I think, is trying to point out is that passing by value is clearer about ownership, and consistent with isocpp guidelines.

ivanpauno:
> > So that suggests to me that this change is actually a no-op
> 
> It is a no-op, to pass by reference and copy-construct vs. to pass by value and move-construct. The thing that @ivanpauno, I think, is trying to point out is that passing by value is clearer about ownership, and consistent with isocpp guidelines.

Exactly that.

PS: passing the shared pointer by reference is slightly less performant, but not a big deal. The point here is to choose the signature that makes the ownership semantics of the function more obvious.

ivanpauno:
> I'm good with it, but yeah, we're past freeze. @jacobperron ?

It's an API and feature freeze. This doesn't change API neither adds a new feature.

clalancette:
All right, thanks for all of the explanations.

Given all of that, I'm still going to recommend that we wait until after Foxy to land this; we have to start slowing down what goes in, and this is relatively minor in the scope of things.  I'll leave it to @jacobperron for the final call.

jacobperron:
Note, there's a cpplint error.

wjwwood:
Uh, tldr, I don't agree and 👎 for this pr.

I'll try to address a few things I saw in this thread that I don't fully agree with.

---

> According to the cpp core guidelines (which imo is the best guideline about smart pointer signatures), if the function always retains shared ownership of a std::shared_ptr, then it must be passed by value.
See R.34 and R.36.

That's true, but depending on the storage policy used, they are either stored as a shared pointer or a weak pointer and periodically upgraded to a shared pointer, see the dynamic storage policy which stores it as a weak pointer:

https://github.com/ros2/rclcpp/blob/1a48a60a75895ae86c1f93a57bc1955b11fb56c9/rclcpp/include/rclcpp/wait_set_policies/dynamic_storage.hpp#L70

I did consult the cpp core guidelines when writing this, but since it only sometimes retains ownership I used a const reference. By the way, the constructor for weak pointer takes a const reference:

https://en.cppreference.com/w/cpp/memory/weak_ptr/weak_ptr constructor (2)

> PS: passing the shared pointer by reference is slightly less performant, but not a big deal.

I think you have this backwards, passing a shared pointer by reference is more efficient than copy constructing it because it avoids a reference count increase which may need to be locked or use an atomic.

> In turn this might help us avoid some memory leaks.

Using `std::move` is about avoiding reference count churn, not avoiding memory leaks. We still have to account for all of the shared pointers and avoid circular references, for example.

> But it equally might cause us to have crashes, since std::move on the shared_ptr causes the caller pointer to go to nullptr.

You should never use `std::move` unless the input argument is `shared_ptr<T> &`, which the cpp core guidelines say you should do if the function wants to indicate to the user that it will "reseat" the pointer, which could either be reassign it, reset it, or move it. Thankfully, you cannot std move a shared pointer that is a const reference, and if it isn't a reference then you're moving a temporary which is safe, and so the only unsafe way to do it is a reference to a shared pointer, but that is explicitly called out in the core guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r35-take-a-shared_ptrwidget-parameter-to-express-that-a-function-might-reseat-the-shared-pointer

rclcpp/include/rclcpp/wait_set_policies/static_storage.hpp
```diff
@@ -60,9 +61,9 @@ class StaticStorage : public rclcpp::wait_set_policies::detail::StoragePolicyCom
 
     /// Conversion constructor, which is intentionally not marked explicit.
     SubscriptionEntry(
-      const std::shared_ptr<rclcpp::SubscriptionBase> & subscription_in = nullptr,
-      const rclcpp::SubscriptionWaitSetMask & mask_in = {})
-    : subscription(subscription_in),
+      std::shared_ptr<rclcpp::SubscriptionBase> subscription_in = nullptr,
+      rclcpp::SubscriptionWaitSetMask mask_in = {})
+    : subscription(std::move(subscription_in)),
```
wjwwood, line 66: These changes in the static storage part of the class can be taken maybe. I think I was trying to keep the same signature between the two kinds of SubscriptionEntry, but since they are different and everything is templated, I think we can potentially take this part of the change.

ivanpauno:
> That's true, but depending on the storage policy used, they are either stored as a shared pointer or a weak pointer and periodically upgraded to a shared pointer, see the dynamic storage policy which stores it as a weak pointer:

I saw the dynamic storage policies. But everything is templated, and there is no common base class between `WeakSubscriptionEntry` and `SubscriptionEntry`, so the signatures don't need to be the same. Actually, they should be different, because they have different ownership semantics.

If we would have a common base class for both with a virtual method, I would agree that taking the `const &` is the correct choice.

> > PS: passing the shared pointer by reference is slightly less performant, but not a big deal.
>
> I think you have this backwards, passing a shared pointer by reference is more efficient than copy constructing it because it avoids a reference count increase which may need to be locked or use an atomic.

It wasn't backwards.
If your method sometimes retains shared ownership and sometimes not, passing by value would be inefficient as in some cases the reference count would be increased and then decreased unnecessarily.

If your method is always retaining shared ownership, doing the copy at call time or later is the same (no performance difference). A `shared_ptr` has a small memory footprint, so passing it by value might be more performant than passing it by reference (it avoids an extra de-referencing) --- but that really doesn't matter ... ---.

> You should never use std::move unless the input argument is shared_ptr<T> &, which the cpp core guidelines say you should do if the function wants to indicate to the user that it will "reseat" the pointer, which could either be reassign it, reset it, or move it. Thankfully, you cannot std move a shared pointer that is a const reference, and if it isn't a reference then you're moving a temporary which is safe, and so the only unsafe way to do it is a reference to a shared pointer, but that is explicitly called out in the core guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r35-take-a-shared_ptrwidget-parameter-to-express-that-a-function-might-reseat-the-shared-pointer

This is not resetting the pointer that the user passed ....
This is just moving the argument that was passed to avoid two copies (one at call time, and one copy construct later), the shared pointer of the caller is not affected (as it was passed by value).
AFAIU,  "copy elision" does not optimize out this case (the only case of named copy elision is NRVO https://en.cppreference.com/w/cpp/language/copy_elision).

wjwwood:
> Actually, they should be different, because they have different ownership semantics.

So this pr should be updated no? Currently the signatures are still the same, just both flipped to by value.

> It wasn't backwards. If your method sometimes retains shared ownership and sometimes not, passing by value would be inefficient as in some cases the reference count would be increased and then decreased unnecessarily.

What? You said:

> passing the shared pointer by reference is slightly less performant

I'm so confused 😅, either passing by reference is less performant or passing by value is, right? I agree that passing by value is wasteful if you don't need it, but your original comment said the reference is slower...

>  A shared_ptr has a small memory footprint, so passing it by value might be more performant than passing it by reference

I don't think that will ever be the case, as this rule only applies to values of the same size as a reference (pointer), but a shared pointer will always be a pointer + something (like the control block), and often incrementing the reference count will be non-trivial compared to a dereference. Either way, it's not something we have evidence for, so avoiding the reference to a shared pointer because it might be less efficient doesn't make sense to me.

> This is not resetting the pointer that the user passed ....

Sorry, I wasn't suggesting we are, I was explaining to @clalancette why moving is safe in most cases (When you're doing it to an input parameter within a function). We're not using this pattern at all. Also, I wasn't suggesting anything about copy elisions either, not sure where that came into play...

ivanpauno:
> I don't think that will ever be the case, as this rule only applies to values of the same size as a reference (pointer), but a shared pointer will always be a pointer + something (like the control block), and often incrementing the reference count will be non-trivial compared to a dereference. Either way, it's not something we have evidence for, so avoiding the reference to a shared pointer because it might be less efficient doesn't make sense to me.

Yeah, that part of the discussion doesn't matter.

> So this pr should be updated no? Currently the signatures are still the same, just both flipped to by value.

I don't understand what you mean :confused:.
I think that passing by value in `SubscriptionEntry` is ok, and that passing `const &` in `WeakSubscriptionEntry` is ok.

Please, comment in line what are the required changes.

> Sorry, I wasn't suggesting we are, I was explaining to @clalancette why moving is safe in most cases (When you're doing it to an input parameter within a function). We're not using this pattern at all. Also, I wasn't suggesting anything about copy elisions either, not sure where that came into play...

Sorry, I got confused

ivanpauno:
> Sorry, I thought this was changing the weak subscription entry, or rather than there was only one of them, so this is fine I think. I was wrong.
> 
> lgtm.

haha no problem, I think everybody --including me-- contributed a bit to the general confusion :joy:

----

CI config:

- build: up to rclcpp
- test: only rclcpp

Jobs:

* Linux: [Build Status](http://ci.ros2.org/job/ci_linux/10927/)
* AArch64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6276/)
* osX: [Build Status](http://ci.ros2.org/job/ci_osx/8898/)
* Windows: [Build Status](http://ci.ros2.org/job/ci_windows/10830/)

PR #1146, opened 2020-05-29
Make test_rate more reliable on Windows and improve error output when it fails

ivanpauno:
Test rate is sometimes failing on Windows, e.g.: https://ci.ros2.org/job/ci_windows/10825/testReport/junit/(root)/projectroot/test_rate/.

We've been fixing lately several issues we had on time based test on Windows, where the problem usually is that the test is using a small error-margin/period/etc (compared with the OS time slice).

In this case, the error message isn't clear:

```
C:\ci\ws\src\ros2\rclcpp\rclcpp\test\test_rate.cpp(36): error: Value of: period < delta
  Actual: false
Expected: true
```

This will improve the error message when the test fails.

### DIFF
--- a/rclcpp/test/test_rate.cpp
+++ b/rclcpp/test/test_rate.cpp
@@ -22,9 +22,9 @@
    Basic tests for the Rate and WallRate classes.
  */
 TEST(TestRate, rate_basics) {
-  auto period = std::chrono::milliseconds(100);
-  auto offset = std::chrono::milliseconds(50);
-  auto epsilon = std::chrono::milliseconds(1);
+  auto period = std::chrono::milliseconds(1000);
+  auto offset = std::chrono::milliseconds(500);
+  auto epsilon = std::chrono::milliseconds(100);
   double overrun_ratio = 1.5;
 
   auto start = std::chrono::system_clock::now();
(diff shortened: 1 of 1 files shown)


rclcpp/test/test_rate.cpp
```diff
@@ -33,24 +33,24 @@ TEST(TestRate, rate_basics) {
   ASSERT_TRUE(r.sleep());
   auto one = std::chrono::system_clock::now();
   auto delta = one - start;
-  ASSERT_TRUE(period < delta);
-  ASSERT_TRUE(period * overrun_ratio > delta);
+  EXPECT_LT(period, delta);
```
jacobperron, line 36: Perhaps in a follow-up PR we can help make these checks more robust by checking `period < (delta + constant)`

ivanpauno: I haven't done that modification yet, because my first assumption is that if you slept for `period`, then `delta` should be always greater than that (considering that the start up time of delta was calculated before rate construction, and the finish time was calculated after the sleep call waked up).

I see that this is only failing when the system timer is being using, and not when using the steady timer. It might be the case that my assumption is only valid for a steady timer.

clalancette: > I see that this is only failing when the system timer is being using, and not when using the steady timer. It might be the case that my assumption is only valid for a steady timer.

For a system timer, you can't reliably assume any time has passed on small timescales; corrections may skew the result.

ivanpauno: See https://github.com/ros2/rclcpp/pull/1146/commits/c1017e14420ced48bb1dce70c2bdc12885a52252.

ivanpauno:
CI config:

* Build: Up to rclcpp
* Test: `--retest-until-fail 20 --packages-select rclcpp --ctest-args -R test_rate`

Jobs:

* Windows: [Build Status](http://ci.ros2.org/job/ci_windows/10855/)

PR #1149, opened 2020-06-01
Add missing virtual destructors

ivanpauno:
Recently, we detected a missing virtual dtor in [QOSEventHandlerBase](https://github.com/ros2/rclcpp/pull/1119#discussion_r428216447).
It sounded like something that the compiler could catch, and after looking a bit I found:

>        -Wnon-virtual-dtor (C++ and Objective-C++ only)
>           Warn when a class has virtual functions and an accessible non-virtual destructor itself or in > an accessible polymorphic base class, in which case it is possible but unsafe to delete an instance
>           of a derived class through a pointer to the class itself or base class.  This warning is automatically enabled if -Weffc++ is specified.

i.e.: if the class has a public non virtual dtor and virtual methods, the compiler will warn.

The new compiler option detected missing virtual dtors in `RateBase` and `TimerBase`, which are fixed in this PR too.

---

I propose also adding `-Woveloaded-virtual`:

>        -Woverloaded-virtual (C++ and Objective-C++ only)
>           Warn when a function declaration hides virtual functions from a base class.

We should probably add both warnings to all cpp packages.

---

I'm planning to rebase and merge, one commit adding the warnings and the other fixing the missing virtual dtors.

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -23,7 +23,7 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
 endif()
 
 set(${PROJECT_NAME}_SRCS
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -31,6 +31,7 @@ class RateBase
 public:
   RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(RateBase)
 
+  virtual ~RateBase() {}
   virtual bool sleep() = 0;
   virtual bool is_steady() const = 0;
   virtual void reset() = 0;
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -62,6 +62,7 @@ public:
 
   /// TimerBase destructor
   RCLCPP_PUBLIC
+  virtual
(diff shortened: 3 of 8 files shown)


rclcpp/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
```
hidmic, line 26: @ivanpauno will `-Woverloaded-virtual` result in a warning if said method is both overriden and overloaded?

ivanpauno: gcc man page explanation is better than mine:

> Warn when a function declaration hides virtual functions from a base class

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10949/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6294/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/8916/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10856/)

ivanpauno:
@jacobperron @hidmic I've added the same flags to `rclcpp_action`, `rclcpp_components` and `rclcpp_lifecycle`; added some missing virtual dtors in `rclcpp_lifecycle` and retriggered CI.

ivanpauno:
CI failed because `LifecyclePublisherInterface` destructor didn't have a definition (IDK what I was thinking ...). It didn't show up until `lifecycle` demo was built, because we don't have a test for `LifecyclePublisherInterface` in `rclcpp_lifecycle`.

Retriggered full CI in https://github.com/ros2/rclcpp/pull/1149#issuecomment-636981657...

ivanpauno:

wjwwood:
lgtm since build.ros2.org passed the linter fix.

PR #1150, opened 2020-06-01
Avoid multiple type topics in tests.

hidmic:
Precisely what the title says, as RMW implementation may but are not bound to support it. This is making these tests fail on nightlies running with RMW implementations other than Fast-RTPS (see http://build.ros2.org/view/Fci/job/Fci__nightly-cyclonedds_ubuntu_focal_amd64/128/testReport/junit/(root)/rclcpp/test_subscription_topic_statistics_gtest_missing_result/)

### DIFF
--- a/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
+++ b/rclcpp/test/topic_statistics/test_subscription_topic_statistics.cpp
@@ -210,21 +210,20 @@ protected:
   void SetUp()
   {
     rclcpp::init(0 /* argc */, nullptr /* argv */);
-    empty_subscriber = std::make_shared<EmptySubscriber>(
-      kTestSubNodeName,
-      kTestSubStatsTopic);
   }
 
   void TearDown()
   {
     rclcpp::shutdown();
-    empty_subscriber.reset();
   }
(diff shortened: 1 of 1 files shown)


hidmic:
CI up to `rclcpp` using CycloneDDS only:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10955/)

hidmic:
> This is fine by me, though I don't fully understand how it fixes the issue. Seems like the type should always be Empty before and after.

The same test fixture is being used for other tests below that instantiate pub/sub instances with a different type against the same topic. This simply ensures those `Empty` subscribers are created when they actually should.

wjwwood:
I see, makes sense. We could have changed the topic names too...

hidmic:
> I see, makes sense. We could have changed the topic names too...

We can do that as well. I guess I didn't see the point in keeping an unused subscriber around.

wjwwood:
That's a valid point too, it's just confusing to use the same topic name with different types in the test, even if you ensure they're not used at the same time.

hidmic:
@wjwwood see https://github.com/ros2/rclcpp/pull/1150/commits/bee47ab081859ead4ca24cd45094369d0813ede7.

PR #1156, opened 2020-06-05
Add spin_all method to Executor

ivanpauno:
This is a midpoint between the old implementation of `spin_some` and the one after https://github.com/ros2/rclcpp/pull/844.

The old `spin_some` implementation waited on entities when getting each executable.
That created starvation and other problems as described in https://github.com/ros2/rclcpp/pull/844.

Now, it only collects entities once and then executes all the ones that are ready.
The problem is that executing "all ready" entities doesn't actually execute all of them.

e.g.: If a subscription have a queue of 100 messages ready, it will only take one.

This proposes to wait on entities and execute all the ready executables repeatedly, until all of them are executed or the specified time has elapsed.

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -185,7 +185,7 @@ public:
    * become ready, this method will execute work repeatedly until `max_duration` has elapsed.
    *
    * \param[in] max_duration The maximum amount of time to spend executing work. Must be positive.
-   * Note that spin_some() may take longer than this time as it only returns once max_duration has
+   * Note that spin_all() may take longer than this time as it only returns once max_duration has
    * been exceeded.
    */
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -160,7 +160,7 @@ public:
   void
   spin_node_some(std::shared_ptr<rclcpp::Node> node);
 
-  /// Complete all available queued work without blocking.
+  /// Complete some available queued work without blocking.
   /**
    * This function can be overridden. The default implementation is suitable for a
    * single-threaded model of execution.
@@ -174,6 +174,26 @@ public:
   RCLCPP_PUBLIC
   virtual void
   spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0));
+
+  /// Complete all available queued work without blocking.
+  /**
+   * This function can be overridden. The default implementation is suitable for a
+   * single-threaded model of execution.
+   * Adding subscriptions, timers, services, etc. with blocking callbacks will cause this function
+   * to block (which may have unintended consequences).
+   *
+   * To ensure all available queued work is executed, this function have to check if there is
+   * available work repeatedly.
+   * Thus, if the time that waitables take to be executed is longer than the period on which new waitables
+   * become ready, this will execute work repeatedly until `max_duration` has elapsed.
+   *
+   * \param[in] max_duration The maximum amount of time to spend executing work. Must be positive.
+   * Note that spin_some() may take longer than this time as it only returns once max_duration has
+   * been exceeded.
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  spin_all(std::chrono::nanoseconds max_duration);
 
   RCLCPP_PUBLIC
   virtual void
@@ -269,6 +289,10 @@ protected:
   spin_node_once_nanoseconds(
     rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node,
     std::chrono::nanoseconds timeout);
+
+  RCLCPP_PUBLIC
+  void
+  spin_some_impl(std::chrono::nanoseconds max_duration, bool exhaustive);
 
   /// Find the next available executable and do the work associated with it.
   /**
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -160,7 +160,7 @@ public:
   void
   spin_node_some(std::shared_ptr<rclcpp::Node> node);
 
-  /// Complete some available queued work without blocking.
+  /// Collect work once and execute all available work, optionally within a duration.
   /**
    * This function can be overridden. The default implementation is suitable for a
    * single-threaded model of execution.
@@ -175,17 +175,14 @@ public:
   virtual void
   spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0));
 
-  /// Complete all available queued work without blocking.
+  /// Collect and execute work repeatedly within a duration or until no more work is available.
   /**
    * This function can be overridden. The default implementation is suitable for a
    * single-threaded model of execution.
    * Adding subscriptions, timers, services, etc. with blocking callbacks will cause this function
    * to block (which may have unintended consequences).
-   *
-   * To ensure all available queued work is executed, this function has to check if there is
-   * available work repeatedly.
-   * Thus, if the time that waitables take to be executed is longer than the period on which new waitables
-   * become ready, this will execute work repeatedly until `max_duration` has elapsed.
+   * If the time that waitables take to be executed is longer than the period on which new waitables
+   * become ready, this method will execute work repeatedly until `max_duration` has elapsed.
    *
    * \param[in] max_duration The maximum amount of time to spend executing work. Must be positive.
    * Note that spin_some() may take longer than this time as it only returns once max_duration has
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -182,7 +182,7 @@ public:
    * Adding subscriptions, timers, services, etc. with blocking callbacks will cause this function
    * to block (which may have unintended consequences).
    *
-   * To ensure all available queued work is executed, this function have to check if there is
+   * To ensure all available queued work is executed, this function has to check if there is
    * available work repeatedly.
    * Thus, if the time that waitables take to be executed is longer than the period on which new waitables
    * become ready, this will execute work repeatedly until `max_duration` has elapsed.
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -28,6 +28,8 @@
 
 #include "rcutils/logging_macros.h"
 
+using namespace std::chrono_literals;
+
 using rclcpp::exceptions::throw_from_rcl_error;
 using rclcpp::AnyExecutable;
 using rclcpp::Executor;
@@ -212,8 +214,21 @@ Executor::spin_node_some(std::shared_ptr<rclcpp::Node> node)
   this->spin_node_some(node->get_node_base_interface());
 }
 
+void Executor::spin_some(std::chrono::nanoseconds max_duration)
+{
+  return this->spin_some_impl(max_duration, false);
+}
+
+void Executor::spin_all(std::chrono::nanoseconds max_duration)
+{
+  if (max_duration <= 0ns) {
+    throw std::invalid_argument("max_duration must be positive");
+  }
+  return this->spin_some_impl(max_duration, true);
+}
+
 void
-Executor::spin_some(std::chrono::nanoseconds max_duration)
+Executor::spin_some_impl(std::chrono::nanoseconds max_duration, bool exhaustive)
 {
   auto start = std::chrono::steady_clock::now();
   auto max_duration_not_elapsed = [max_duration, start]() {
@@ -232,14 +247,20 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
     throw std::runtime_error("spin_some() called while already spinning");
   }
   RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
-  // non-blocking call to pre-load all available work
-  wait_for_work(std::chrono::milliseconds::zero());
+  bool work_available = false;
   while (rclcpp::ok(context_) && spinning.load() && max_duration_not_elapsed()) {
     AnyExecutable any_exec;
+    if (!work_available) {
+      wait_for_work(std::chrono::milliseconds::zero());
+    }
     if (get_next_ready_executable(any_exec)) {
       execute_any_executable(any_exec);
+      work_available = true;
     } else {
-      break;
+      if (!work_available || !exhaustive) {
+        break;
+      }
+      work_available = false;
     }
   }
 }
--- a/rclcpp/test/rclcpp/test_executor.cpp
+++ b/rclcpp/test/rclcpp/test_executor.cpp
@@ -115,3 +115,85 @@ TEST_F(TestExecutors, testSpinUntilFutureCompleteSharedFuture) {
   ret = executor.spin_until_future_complete(future.share(), 100ms);
   EXPECT_EQ(rclcpp::FutureReturnCode::TIMEOUT, ret);
 }
+
+class TestWaitable : public rclcpp::Waitable
+{
+public:
+
+  TestWaitable()
+  {
+    rcl_guard_condition_options_t guard_condition_options =
+      rcl_guard_condition_get_default_options();
+
+    gc_ = rcl_get_zero_initialized_guard_condition();
+    rcl_ret_t ret = rcl_guard_condition_init(
+      &gc_,
+      rclcpp::contexts::get_global_default_context()->get_rcl_context().get(),
+      guard_condition_options);
+    if (RCL_RET_OK != ret) {
+      rclcpp::exceptions::throw_from_rcl_error(ret);
+    }
+  }
+
+  bool
+  add_to_wait_set(rcl_wait_set_t * wait_set) override
+  {
+    rcl_ret_t ret = rcl_wait_set_add_guard_condition(wait_set, &gc_, NULL);
+    if (RCL_RET_OK != ret) {
+      return false;
+    }
+    ret = rcl_trigger_guard_condition(&gc_);
+    return RCL_RET_OK == ret;
+  }
+
+  bool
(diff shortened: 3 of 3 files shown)


ivanpauno:
@roger-strain as you proposed #844, I would like to hear your opinion on this change.
Thanks!

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -232,16 +232,18 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
...
-      break;
+  do {
+    // non-blocking call to pre-load all available work
+    wait_for_work(std::chrono::milliseconds::zero());
+    while (spinning.load() && max_duration_not_elapsed()) {
+      AnyExecutable any_exec;
+      if (get_next_ready_executable(any_exec)) {
+        execute_any_executable(any_exec);
+      } else {
+        return;
```
hidmic, line 243: @ivanpauno how's this any different? I'd think `spin_some()` should return only once a non-blocking `wait_for_work()` call results in no executables to be served.

ivanpauno: :joy:
The original version could introduce a busy loop (non-blocking wait with nothing ready tried repeatedly). https://github.com/ros2/rclcpp/pull/1156/commits/0b2f75f80ef720fcbf1dfdda9c1625096ecba28d was a really bad attempt of fixing that.

See https://github.com/ros2/rclcpp/pull/1156/commits/6866819cae28a6297beebd12940989566610c54d for a real fix.

ivanpauno: :man_facepalming: see https://github.com/ros2/rclcpp/pull/1156/commits/55fbb44d117839b017e89207651815996af7d5bd.
I tested it this time.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -232,16 +232,24 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
...
+      } else {
+        if (!something_was_ready) {
+          return;
+        } else {
+          break;
+        }
+      }
     }
-  }
+  } while (max_duration_not_elapsed());
```
hidmic, line 252: @ivanpauno what if `spinning.load()` is false?

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -232,16 +232,24 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
...
   RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
-  // non-blocking call to pre-load all available work
-  wait_for_work(std::chrono::milliseconds::zero());
-  while (spinning.load() && max_duration_not_elapsed()) {
-    AnyExecutable any_exec;
-    if (get_next_ready_executable(any_exec)) {
-      execute_any_executable(any_exec);
-    } else {
-      break;
+  do {
```
hidmic, line 235: @ivanpauno one suggestion:

```c++
bool work_available = false;
AnyExecutable any_exec;
while (spinning.load() && max_duration_not_elapsed()) {
  if (!work_available) {
    wait_for_work(std::chrono::milliseconds::zero());
  }
  if (get_next_ready_executable(any_exec)) {
      work_available = true;
  } else {
     if (!work_available) {
        break;
     }
     work_available = false;
  }
}
```

ivanpauno: that's wrong, because it'll wait again until executing one executable.
That's what https://github.com/ros2/rclcpp/pull/844 fixed.

roger-strain: I think this loop structure is a bit cleaner, so probably would go this way if possible.

ivanpauno: That was almost perfect.
I added an extra check, see https://github.com/ros2/rclcpp/pull/1156/commits/f4dc1768650d14a392f563161eaf9aafffb05cd8.

ivanpauno:
The default timeout of `spin_some` is 0, which is used as "do work until nothing all has been handled", i.e.: it might work forever.
Is that a nice default? shouldn't the user be forced to pass a timeout?

Maybe instead of this, we should have two methods:
* spin_some: Checks for ready entities once
* spin_for: Check for ready entities multiple times

rclcpp/src/rclcpp/executor.cpp
hidmic: Good catch.

roger-strain:
> The default timeout of `spin_some` is 0, which is used as "do work until nothing all has been handled", i.e.: it might work forever.
> Is that a nice default? shouldn't the user be forced to pass a timeout?
> 
> Maybe instead of this, we should have two methods:
> 
>     * spin_some: Checks for ready entities once
> 
>     * spin_for: Check for ready entities multiple times

As I think about it, "it might work forever" was exactly the problem we were trying to avoid the last time around.

First as a side note, if I see a `spin_for` method, I'm probably going to expect it to take that long before returning, whether or not there's work to do. "You told me to spin for five seconds, so I'll spin for five seconds with nothing to do." There could be a use case for that behavior, honestly, so it might be fair to consider that independently of the question of whether zero is a nice default for `spin_some`.

Back to the main question, I'm not sure of what a developer would expect the default behavior to be. If I want to make sure that I get a timely return from that method, I feel it's my responsibility to say so, by specifying a timeout.

That being said, we could also consider introducing an optional flag parameter for whether we allow requerying the work queues after the first instance. Perhaps the default values would be time limit of zero (keep going until you're done), and do *not* requery after the first batch. If you want the "run until you're completely out of things to do" behavior, you can call `spin_some(0, true)`. If you want "run for a set amount of time, but keep at it for that long", you can call `spin_some(100, true)`. If you can't be bothered to specify anything, you get "run for as long as it takes to service each thing zero or one times" by calling `spin_some()`.

ivanpauno:
> As I think about it, "it might work forever" was exactly the problem we were trying to avoid the last time around.

Yeap, I think that the correct behavior should be something like:

* Collect ready entities once.
* Handle all the queue of ready entities (i.e. all the messages of subscriptions, all request of servers, etc) within the passed duration.

Unfortunately, that requires much more work than this.
We should revisit wait_set/executor design, as largely discussed in https://github.com/ros2/design/issues/259, but that is a broader discussion.

> First as a side note, if I see a spin_for method, I'm probably going to expect it to take that long before returning, whether or not there's work to do. "You told me to spin for five seconds, so I'll spin for five seconds with nothing to do."

That makes sence.

> There could be a use case for that behavior, honestly, so it might be fair to consider that independently of the question of whether zero is a nice default for spin_some.

Makes sense too.

> Back to the main question, I'm not sure of what a developer would expect the default behavior to be. If I want to make sure that I get a timely return from that method, I feel it's my responsibility to say so, by specifying a timeout.

I think it might be confusing for some people, so I would prefer not giving a default if we keep the current 0s sementics.

> That being said, we could also consider introducing an optional flag parameter for whether we allow requerying the work queues after the first instance. Perhaps the default values would be time limit of zero (keep going until you're done), and do not requery after the first batch. If you want the "run until you're completely out of things to do" behavior, you can call spin_some(0, true). If you want "run for a set amount of time, but keep at it for that long so long as there's anything to do", you can call spin_some(100, true). If you can't be bothered to specify anything, you get "run for as long as it takes to service each thing zero or one times" by calling spin_some().

:+1:
@hidmic @wjwwood does that make sense to you?

ivanpauno:
We should really improve `AnyExecutable` as using it is confusing (see https://github.com/ros2/rclcpp/pull/1156/commits/0f38c51830c131eaa3cb3ae8bba8adce395b20d9).

hidmic:
> does that make sense to you?

I'd be fine with the added argument, though I fail to see in what circumstances I'd set it to `false`. If I use `spin_some()`, I expect it to run all available executables. If I had an issue with it eventually spinning continuously I'd set a timeout or use `spin_once()` instead. But again, I'm fine with it if people think it's useful.

ivanpauno:
> I'd be fine with the added argument, though I fail to see in what circumstances I'd set it to false. If I use spin_some(), I expect it to run all available executables. If I had an issue with it eventually spinning continuously I'd set a timeout or use spin_once() instead. But again, I'm fine with the proposed change if people think it's useful.

There's a difference with `spin_once`: it only waits once and it executes only ONE executable.

What I don't like of this `spin_some` modification, is that it can block undefinitely (with the default 0s duration). IMO, that's not ideal.

I went ahead with @roger-strain idea, and implemented it in https://github.com/ros2/rclcpp/pull/1156/commits/05625f03539b67f6fa3cfaf24525857330cbeeb0.
I can revert it if that extra parameter is not desired.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -169,10 +169,16 @@ class Executor
    * \param[in] max_duration The maximum amount of time to spend executing work, or 0 for no limit.
    * Note that spin_some() may take longer than this time as it only returns once max_duration has
    * been exceeded.
+   * \param[in] exhaustive When true, it will try to collect entities again after executing executables
+   * until there is no more ready entities or the specified duration has been reached.
+   * When there are subscriptions/clients with long history queues, this will ensure all messages/requestes
+   * available are handled.
```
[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1156/commits/58ac8c9a0b8a355f344c8d1bf7bd451b1891865e.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -232,14 +232,20 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
     throw std::runtime_error("spin_some() called while already spinning");
   }
   RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
-  // non-blocking call to pre-load all available work
-  wait_for_work(std::chrono::milliseconds::zero());
-  while (spinning.load() && max_duration_not_elapsed()) {
+  bool work_available = false;
+  while (rclcpp::ok(this->context_) && spinning.load() && max_duration_not_elapsed()) {
```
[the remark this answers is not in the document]
ivanpauno: The problem is that with the old `spin_some` implementation, no having that check wasn't much of a problem. It would take an unnoticeable time for `spin_some` to finish and react to `ctrl-c`.

Now that we're repeatedly waiting to check if there's more work, that time can be really noticeable.

I will split that change, so we can cherry-pick it separately.

ivanpauno:
> W.r.t to the naming discussion, for me `spin_some()` doesn't necessarily mean "everything", just as much as you can, either at once or in a given period. Maybe it should be `spin_some(max_period=0)` and `spin_all(max_period)` (forcing users to give a period when you do all).

I really like the suggested naming, I've applied it in https://github.com/ros2/rclcpp/pull/1156/commits/58ac8c9a0b8a355f344c8d1bf7bd451b1891865e.

rclcpp/src/rclcpp/executor.cpp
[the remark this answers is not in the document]
wjwwood: Ok makes sense.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -174,6 +174,26 @@ class Executor
...
   spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0));
 
+  /// Complete all available queued work without blocking.
+  /**
+   * This function can be overridden. The default implementation is suitable for a
+   * single-threaded model of execution.
+   * Adding subscriptions, timers, services, etc. with blocking callbacks will cause this function
+   * to block (which may have unintended consequences).
+   *
+   * To ensure all available queued work is executed, this function have to check if there is
```
wjwwood, line 184:
```suggestion
   * To ensure all available queued work is executed, this function has to check if there is
```

ivanpauno:
Rebased after conflicting changes in https://github.com/ros2/rclcpp/pull/1167.
Added a test case in https://github.com/ros2/rclcpp/pull/1156/commits/2fd08b5ad7143176ff7b5f6dab2042d07ce1e625.

rclcpp/include/rclcpp/executor.hpp
ivanpauno: Thanks! Fixed in https://github.com/ros2/rclcpp/pull/1156/commits/7d43e5b27adbf0cfe87ed6ebd4168d47b4ef4283.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6422/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9072/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11044/)

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -232,14 +247,20 @@ Executor::spin_some(std::chrono::nanoseconds max_duration)
     throw std::runtime_error("spin_some() called while already spinning");
   }
   RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
-  // non-blocking call to pre-load all available work
-  wait_for_work(std::chrono::milliseconds::zero());
+  bool work_available = false;
```
clalancette, line 250: Looking through the logic of this, I think that this variable could be named better.  In particular, I think this variable really controls whether to "check if there is new work available".  Is that correct?  If so, I'd suggest `check_for_work_available`, or something like that, and invert the logic.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -175,6 +175,26 @@ class Executor
...
+  /**
+   * This function can be overridden. The default implementation is suitable for a
+   * single-threaded model of execution.
+   * Adding subscriptions, timers, services, etc. with blocking callbacks will cause this function
+   * to block (which may have unintended consequences).
+   *
+   * To ensure all available queued work is executed, this function has to check if there is
+   * available work repeatedly.
+   * Thus, if the time that waitables take to be executed is longer than the period on which new waitables
+   * become ready, this will execute work repeatedly until `max_duration` has elapsed.
```
clalancette, line 188: If I'm reading the logic correctly, the description of this logic isn't quite correct (nor is `spin_some` documentation).

For `spin_some`, the executor will collect work *once*, and then execute all of the work that was available at that time until the timeout expires.

For `spin_all`, the executor will collect work repeatedly, and then execute all of the work that was available (and that comes in) until the timeout expires.

In both cases, once it has executed one callback, if there is no further work to do, it will return immediately.

Does that sound like the correct logic to you?  If so, I'd recommend updating this documentation describing more of that.

rclcpp/src/rclcpp/executor.cpp
ivanpauno: mmm, `work_available` sounds like a better name to me.
`work_available` is `true` if the last `wait_for_work` resulted in at least one available entity.

We're actually checking for more work (i.e. calling `wait_for_work`) when `work_available` is `false`, so the new name sounds confusing (except if I invert the logic).

rclcpp/include/rclcpp/executor.hpp
ivanpauno: That description is correct, I will try to update the doc blocks to make the difference clearer.

ivanpauno: > the executor will collect work repeatedly, and then execute all of the work that was available

To be more precise, `spin_all` collects and executes work repeatedly within the specified duration.

ivanpauno: See https://github.com/ros2/rclcpp/pull/1156/commits/07c84d36a8e665bc8d4c4d90df869b96d62cc012

ivanpauno:
I've force pushed to solve conflicts, no code changes have been done.

ivanpauno:
Checking if the rebase didn't break anything:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11164/)

rclcpp/src/rclcpp/executor.cpp
wjwwood: I have no strong preference here.

clalancette: Yeah, you'd have to to invert the logic for the new name to make sense.

The main reason I was pushing for a change is that it is somewhat subtle how `spin_some` differs from `spin_all`, and it took a good 10 minutes of staring at the code for me to understand the difference.  But I don't feel very strongly about this one, so if you think it is clearer as-is, that's fine with me.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -175,6 +175,23 @@ class Executor
...
+  /**
+   * This function can be overridden. The default implementation is suitable for a
+   * single-threaded model of execution.
+   * Adding subscriptions, timers, services, etc. with blocking callbacks will cause this function
+   * to block (which may have unintended consequences).
+   * If the time that waitables take to be executed is longer than the period on which new waitables
+   * become ready, this method will execute work repeatedly until `max_duration` has elapsed.
+   *
+   * \param[in] max_duration The maximum amount of time to spend executing work. Must be positive.
+   * Note that spin_some() may take longer than this time as it only returns once max_duration has
```
clalancette, line 188: Nit:

```suggestion
   * Note that spin_all() may take longer than this time as it only returns once max_duration has
```

ivanpauno: Ups, thanks for catching that. Fixed in https://github.com/ros2/rclcpp/pull/1156/commits/668a3b94fb769282a203fcfa85f428eb19dd6a4b.

rclcpp/src/rclcpp/executor.cpp
ivanpauno: Yes, I think that the difference between both methods is subtle and might be a bit confusing.
With the last doc block update, I hope it will be clear enough for users.

About naming of that variable, I think that depending on the person one name or the other may be clearer. As I don't see a clear benefit, I will keep the original name.

PR #1157, opened 2020-06-08
Include original exception in ComponentManagerException

mbuijs:
When a component throws an exception in the constructor (e.g. when declaring a parameter) during loading, there's no other feedback than `Component constructor threw an exception`. This change introduces a similar construction to what is done earlier in the same file for constructing `class_loader::ClassLoader`: try to catch `std::exception` and append it's `what()` to the new `ComponentManagerException`.

### DIFF
--- a/rclcpp_components/src/component_manager.cpp
+++ b/rclcpp_components/src/component_manager.cpp
@@ -196,6 +196,11 @@ ComponentManager::OnLoadNode(
 
       try {
         node_wrappers_[node_id] = factory->create_node_instance(options);
+      } catch (const std::exception &ex) {
+        // In the case that the component constructor throws an exception,
+        // rethrow into the following catch block.
+        throw ComponentManagerException(
+                "Component constructor threw an exception: " + std::string(ex.what()));
       } catch (...) {
         // In the case that the component constructor throws an exception,
         // rethrow into the following catch block.
--- a/rclcpp_components/src/component_manager.cpp
+++ b/rclcpp_components/src/component_manager.cpp
@@ -196,7 +196,7 @@ ComponentManager::OnLoadNode(
 
       try {
         node_wrappers_[node_id] = factory->create_node_instance(options);
-      } catch (const std::exception &ex) {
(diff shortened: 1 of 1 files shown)


rclcpp_components/src/component_manager.cpp
```diff
@@ -196,6 +196,10 @@ ComponentManager::OnLoadNode(
 
       try {
         node_wrappers_[node_id] = factory->create_node_instance(options);
+      } catch (const std::exception &ex) {
+        // In the case that the component constructor throws an exception,
+        // rethrow into the following catch block.
+        throw ComponentManagerException("Component constructor threw an exception: " + std::string(ex.what()));
```
mjcarroll, line 202:
```suggestion
        throw ComponentManagerException(
                "Component constructor threw an exception: " +
                std::string(ex.what()));
```

mbuijs: Sorry, I should've checked that.

mbuijs:
> Taking back my approval, because this doesn't pass the linters.

I've clicked the `Re-request review` button, but I'm not sure if it actually does something, so through this message I would like to request you to review this pull request again. I have fixed the linter error about the line length in an attempt to be consistent with the line breaks in the rest of the file.

rclcpp_components/src/component_manager.cpp
```diff
@@ -196,6 +196,11 @@ ComponentManager::OnLoadNode(
 
       try {
         node_wrappers_[node_id] = factory->create_node_instance(options);
+      } catch (const std::exception &ex) {
```
fujitatomoya, line 199: http://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/31/testReport/junit/rclcpp_components/uncrustify/src_component_manager_cpp/

```suggestion
      } catch (const std::exception & ex) {
```

mbuijs: Fixed, thanks for checking.

Issue #1021, opened 2020-03-10
interrupt_guard_condition_ causes executor "spurious" awakes


alsora:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source

#### Steps to reproduce issue

Run a ROS 2 system made of a single node with only a timer. No publishers/subscriptions or anything else.

```
#include <chrono>
#include <memory>
#include "rclcpp/rclcpp.hpp"

using namespace std::chrono_literals;

class MinimalTimer : public rclcpp::Node
{
public:
  MinimalTimer()
  : Node("minimal_timer")
  {
    timer_ = this->create_wall_timer(
      5s, std::bind(&MinimalTimer::timer_callback, this));
  }

private:
  void timer_callback()
  {
    RCLCPP_WARN(this->get_logger(), "Timer executed!");
  }
  rclcpp::TimerBase::SharedPtr timer_;
};

int main(int argc, char * argv[])
{
  rclcpp::init(argc, argv);
  rclcpp::spin(std::make_shared<MinimalTimer>());
  rclcpp::shutdown();
  return 0;
}
```

#### Expected behavior

The timer has a 5 seconds period, so  the executor should wake up once every 5 seconds.

#### Actual behavior

After every real awakes, the executor wakes up a second time for no apparent reason.

 - Executor waits for 5 seconds
 - Executor wakes up to execute the timer callback
 - Executor goes back to wait
 - Executor wakes up IMMEDIATELY
 - There is no work to do, so executor goes back to wait
 - Executor waits for 5 seconds

This pattern keeps going on forever.
This applies also to more complex systems, with the same frequency.

Some logs that I added:

```
[INFO] [1583844157.135606428] [single_thread_exec_cpp]: LINE:34 Spin: While loop iteration
[INFO] [1583844157.135620489] [executor_cpp]: LINE:598 get_next_ready_executable 1st call: Nothing
[INFO] [1583844162.135118217] [executor_cpp]: LINE:602 wait_for_work completed
[INFO] [1583844162.135281008] [executor_cpp]: LINE:540 found ready timer
[INFO] [1583844162.135315670] [executor_cpp]: LINE:607 get_next_ready_executable 2nd call: Success
[WARN] [1583844162.135348415] [minimal_timer]: Timer executed!
[INFO] [1583844162.135504465] [single_thread_exec_cpp]: LINE:34 Spin: While loop iteration
[...]

PR #1162, opened 2020-06-09
Toggle Guard Condition Feature

peterpena:
This pull request is related to #1021. Triggering a guard condition is only necessary when the executor is multithreaded and has at least one non-empty `MutualExclusive` callback group. In this PR I added a virtual function with a `trigger_guard_condition_` boolean that by default is set to false and is overriden by the `MultiThreadedExecutor` to set it to true when the conditions are met. The executor sets the trigger when `add_node` or `remove_node` is invoked. 

The tests verify the `SingleThreadedExecutor` sets the trigger to false when `add_node` and `remove_node` are called and `MultiThreadedExecutor` sets it to true when there is a non-empty `MutualExclusive` callback group and false when the node is removed and executor no longer has a node to process:
```
-- run_test.py: extra environment variables to append:
 - LD_LIBRARY_PATH+=/home/peter/ros2_foxy/build/rclcpp
-- run_test.py: invoking following command in '/home/peter/ros2_foxy/build/rclcpp':
 - /home/peter/ros2_foxy/build/rclcpp/test_guard_condition_trigger --gtest_output=xml:/home/peter/ros2_foxy/build/rclcpp/test_results/rclcpp/test_guard_condition_trigger.gtest.xml

Running main() from /home/peter/ros2_foxy/install/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
[==========] Running 2 tests from 1 test case.
[----------] 2 tests from TestGuardConditionTrigger
[ RUN      ] TestGuardConditionTrigger.set_trigger_guard_condition_multi_threaded
[       OK ] TestGuardConditionTrigger.set_trigger_guard_condition_multi_threaded (21 ms)
[ RUN      ] TestGuardConditionTrigger.set_trigger_guard_condition_single_threaded
[       OK ] TestGuardConditionTrigger.set_trigger_guard_condition_single_threaded (8 ms)
[----------] 2 tests from TestGuardConditionTrigger (29 ms total)

[==========] 2 tests from 1 test case ran. (39 ms total)
[  PASSED  ] 2 tests.


-- run_test.py: return code 0
-- run_test.py: inject classname prefix into gtest result file '/home/peter/ros2_foxy/build/rclcpp/test_results/rclcpp/test_guard_condition_trigger.gtest.xml'
-- run_test.py: verify result file '/home/peter/ros2_foxy/build/rclcpp/test_results/rclcpp/test_guard_condition_trigger.gtest.xml'
```

### DIFF
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -94,6 +94,20 @@ public:
     return _find_ptrs_if_impl<rclcpp::Waitable, Function>(func, waitable_ptrs_);
   }
 
+   /**
+   * Returns size of callback group which includes
+   * subscriptions, timers, clients, services, and
+   * waitable
+   *
+   * \return size of callback group
+   */
+  size_t
+  size() const
+  {
+    return subscription_ptrs_.size() + timer_ptrs_.size() +
+           client_ptrs_.size() + service_ptrs_.size() + waitable_ptrs_.size();
+  }
+
   RCLCPP_PUBLIC
   std::atomic_bool &
   can_be_taken_from();
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -263,6 +263,15 @@ public:
   void
   set_memory_strategy(memory_strategy::MemoryStrategy::SharedPtr memory_strategy);
 
+  /**
+   * Returns wake_after_execute_ flag
+   *
+   * \return wake_after_execute_ flag
+   */
+  RCLCPP_PUBLIC
+  bool
+  get_wake_after_executor_flag() {return wake_after_execute_.load();}
+
 protected:
   RCLCPP_PUBLIC
   void
@@ -279,6 +288,17 @@ protected:
   RCLCPP_PUBLIC
   void
   execute_any_executable(AnyExecutable & any_exec);
+
+  /**
+   * After executing an executable, this function determines
+   * if it should wake the wait in rcl_wait so that the executor
+   * can process any pending executable
+   *
+   * \return wake_after_execute_ flag
+   */
+  RCLCPP_PUBLIC
+  virtual bool
+  set_wake_after_execute_flag() {return wake_after_execute_.load();}
 
   RCLCPP_PUBLIC
   static void
@@ -324,6 +344,9 @@ protected:
 
   /// Spinning state, used to prevent multi threaded calls to spin and to cancel blocking spins.
   std::atomic_bool spinning;
+
+  /// boolean to control whether guard condition is triggered after executing
+  std::atomic_bool wake_after_execute_;
 
   /// Guard condition for signaling the rmw layer to wake up for special events.
   rcl_guard_condition_t interrupt_guard_condition_ = rcl_get_zero_initialized_guard_condition();
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -36,6 +36,7 @@ using rclcpp::FutureReturnCode;
 
 Executor::Executor(const rclcpp::ExecutorOptions & options)
 : spinning(false),
+  wake_after_execute_(false),
   memory_strategy_(options.memory_strategy)
 {
   rcl_guard_condition_options_t guard_condition_options = rcl_guard_condition_get_default_options();
@@ -304,9 +305,13 @@ Executor::execute_any_executable(AnyExecutable & any_exec)
   }
   // Reset the callback_group, regardless of type
   any_exec.callback_group->can_be_taken_from().store(true);
+
+  // Check whether triggering a guard condition is necessary
+  // (will depend on the type of executor and callback groups)
   // Wake the wait, because it may need to be recalculated or work that
   // was previously blocked is now available.
-  if (rcl_trigger_guard_condition(&interrupt_guard_condition_) != RCL_RET_OK) {
+  if (set_wake_after_execute_flag() &&
+      rcl_trigger_guard_condition(&interrupt_guard_condition_) != RCL_RET_OK) {
     throw std::runtime_error(rcl_get_error_string().str);
   }
 }
--- a/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
@@ -70,6 +70,43 @@ MultiThreadedExecutor::get_number_of_threads()
   return number_of_threads_;
 }
 
+// Iterate through all the callback groups of
+// all nodes in the executor and verify
+// there is a non-empty callback group
+// that is "Mutually Exclusive". If there is,
+// then set trigger guard condition to true
+// so that it avoids a thread getting stuck
+// in rcl_wait.
+bool
+MultiThreadedExecutor::set_wake_after_execute_flag()
+{
+  // Initialize as false in the case
+  // it is true and the executor has
+  // zero nodes, nodes with empty
+  // callback groups, or nodes with
+  // all reentrant callback groups
+  wake_after_execute_.store(false);
+  for (auto & weak_node : weak_nodes_) {
+    auto node = weak_node.lock();
+    if (!node) {
+      continue;
+    }
+    for (auto & weak_group : node->get_callback_groups()) {
+      auto callback_group = weak_group.lock();
+
+      // Skip over callback groups that are empty
+      if(!callback_group || callback_group->size() == 0) {
+        continue;
+      }
+      if (callback_group->type() == rclcpp::callback_group::CallbackGroupType::MutuallyExclusive) {
+        wake_after_execute_.store(true);
+        break;
+      }
+    }
+  }
+  return wake_after_execute_.load();
+}
+
 void
 MultiThreadedExecutor::run(size_t)
 {
--- a/rclcpp/test/rclcpp/test_wake_after_execute_flag.cpp
+++ b/rclcpp/test/rclcpp/test_wake_after_execute_flag.cpp
@@ -1,0 +1,105 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <string>
+#include <memory>
+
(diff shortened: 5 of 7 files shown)


rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -70,6 +70,42 @@ MultiThreadedExecutor::get_number_of_threads()
...
+  for (auto & weak_node : weak_nodes_) {
+    auto node = weak_node.lock();
+    if (!node) {
+      continue;
+    }
+    for (auto & weak_group : node->get_callback_groups()) {
+      auto callback_group = weak_group.lock();
+
+      // Skip over callback groups that are empty
+      if(!callback_group || callback_group->size() == 0) {
```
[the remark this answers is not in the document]
peterpena: After considering the workarounds, I decided to add this function inside `execute_any_executable` before triggering the guard condition.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -304,9 +305,13 @@ Executor::execute_any_executable(AnyExecutable & any_exec)
...
   // Reset the callback_group, regardless of type
   any_exec.callback_group->can_be_taken_from().store(true);
+
+  // Check whether triggering a guard condition is necessary
+  // (will depend on the type of executor and callback groups)
   // Wake the wait, because it may need to be recalculated or work that
   // was previously blocked is now available.
-  if (rcl_trigger_guard_condition(&interrupt_guard_condition_) != RCL_RET_OK) {
+  if (set_wake_after_execute_flag() &&
+      rcl_trigger_guard_condition(&interrupt_guard_condition_) != RCL_RET_OK) {
```
peterpena, line 314: This is where it decides if it should wake the `waitset` after executing.

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -94,6 +94,20 @@ class CallbackGroup
     return _find_ptrs_if_impl<rclcpp::Waitable, Function>(func, waitable_ptrs_);
   }
 
+   /**
+   * Returns size of callback group which includes
+   * subscriptions, timers, clients, services, and
+   * waitable
+   *
+   * \return size of callback group
+   */
```
wjwwood, line 103:
```suggestion
  /// Return the total number of entities in the callback group.
  /**
   * The size of the callback group includes subscriptions, timers,
   * clients, services, and waitables.
   *
   * \return size of the callback group
   */
```

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -94,6 +94,20 @@ class CallbackGroup
...
+  size_t
+  size() const
+  {
+    return subscription_ptrs_.size() + timer_ptrs_.size() +
+           client_ptrs_.size() + service_ptrs_.size() + waitable_ptrs_.size();
```
wjwwood, line 108: Make sure you run the tests, I'm a bit surprised this passed the linters, but maybe.

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -94,6 +94,20 @@ class CallbackGroup
...
   }
 
+   /**
+   * Returns size of callback group which includes
+   * subscriptions, timers, clients, services, and
+   * waitable
+   *
+   * \return size of callback group
+   */
+  size_t
```
wjwwood, line 104:
```suggestion
  RCLCPP_PUBLIC
  size_t
```

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -263,6 +263,15 @@ class Executor
   void
   set_memory_strategy(memory_strategy::MemoryStrategy::SharedPtr memory_strategy);
 
+  /**
+   * Returns wake_after_execute_ flag
+   *
+   * \return wake_after_execute_ flag
+   */
```
wjwwood, line 270: This doc block should have a brief above the block which starts with `///`, but also this doc block should be more descriptive. Like maybe "Return true if this executor should cause the executor to wake up after execution." or something like that.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -280,6 +289,17 @@ class Executor
   void
   execute_any_executable(AnyExecutable & any_exec);
 
+  /**
+   * After executing an executable, this function determines
+   * if it should wake the wait in rcl_wait so that the executor
+   * can process any pending executable
+   *
+   * \return wake_after_execute_ flag
+   */
```
wjwwood, line 298: Please update this one to have a brief using `///` as well.

Also the name of this function seems off. I would expect it to change the value of this variable rather than just return it (get it).

rclcpp/src/rclcpp/executor.cpp
wjwwood, line 314: nitpick: If you have a multi-line if statement, put the `{` on its own line:

```suggestion
  if (set_wake_after_execute_flag() &&
      rcl_trigger_guard_condition(&interrupt_guard_condition_) != RCL_RET_OK)
  {
```

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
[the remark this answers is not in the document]
wjwwood: Yeah, unfortunately this is very expensive. Iterating over all of the items each time will cost us I think. Also, it should only be necessary when a callback group is added, removed, or has an item added/removed from it.

rclcpp/test/rclcpp/test_wake_after_execute_flag.cpp
```diff
@@ -0,0 +1,105 @@
...
+class TestWakeAfterExecuteFlag : public ::testing::Test
+{
+protected:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+};
+
+
```
wjwwood, line 39: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/test/rclcpp/test_wake_after_execute_flag.cpp
```diff
@@ -0,0 +1,105 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <string>
+#include <memory>
```
wjwwood, line 19: nitpick: please use alphabetical ordering

rclcpp/test/rclcpp/test_wake_after_execute_flag.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/exceptions.hpp"
+#include "rclcpp/node.hpp"
+#include "rclcpp/rclcpp.hpp"
```
wjwwood, line 24: This include is redundant.

rclcpp/test/rclcpp/test_wake_after_execute_flag.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "rclcpp/executors.hpp"
+#include "rclcpp/executor.hpp"
```
wjwwood, line 26: nitpick: also alphabetical here

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -129,6 +136,7 @@ class NodeBase : public NodeBaseInterface
   std::vector<rclcpp::CallbackGroup::WeakPtr> callback_groups_;
 
   std::atomic_bool associated_with_executor_;
+  std::function<void()> call_executor_function_;
```
wjwwood, line 139: Perhaps a better name for this would be `notify_executor_callback_` or `notify_executor_of_callback_group_change_function_` or something. We're technically calling the executor here, but that doesn't indicate why we're calling it.

I like `notify_executor_of_callback_group_change_function_` the best of the two above, but something more descriptive like that, even if it is not that exactly is what I'd like to see.

rclcpp/src/rclcpp/callback_group.cpp
```diff
@@ -48,6 +48,9 @@ CallbackGroup::add_subscription(
       subscription_ptrs_.end(),
       [](rclcpp::SubscriptionBase::WeakPtr x) {return x.expired();}),
     subscription_ptrs_.end());
+  if (call_executor_function_ != nullptr) {
+    call_executor_function_();
+  }
```
wjwwood, line 53: You could consider putting this into a private method, like `notify_of_callback_group_change_if_needed()`.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
wjwwood: Maybe even better, since this class really shouldn't be coupled to a executor (creates a conceptual cycle), just `notify_of_callback_group_change_function_` might be even better. It doesn't matter who we're notifying, we're just notifying them.

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -70,6 +70,42 @@ MultiThreadedExecutor::get_number_of_threads()
...
+// in rcl_wait.
+bool
+MultiThreadedExecutor::determine_wake_after_execute()
+{
+  // Initialize as false in the case
+  // it is true and the executor has
+  // zero nodes, nodes with empty
+  // callback groups, or nodes with
+  // all reentrant callback groups
+  wake_after_execute_.store(false);
```
wjwwood, line 88: Is there a risk here of `wake_after_execute_` being accessed after this line but before the logic is completed below?

Would it not be safer to store it in a temporary and assign it once at the end of this function?

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -70,6 +70,42 @@ MultiThreadedExecutor::get_number_of_threads()
...
 
+// Iterate through all the callback groups of
+// all nodes in the executor and verify
+// there is a non-empty callback group
+// that is "Mutually Exclusive". If there is,
+// then set trigger guard condition to true
+// so that it avoids a thread getting stuck
+// in rcl_wait.
+bool
+MultiThreadedExecutor::determine_wake_after_execute()
```
wjwwood, line 81: Does this need to return the value? It seems like `get_wake_after_execute_flag()` is used in most places instead.

If not, then maybe this should be called `void evaluate_need_to_wake_after_execute()`?

rclcpp/test/rclcpp/test_wake_after_execute_flag.cpp
```diff
@@ -0,0 +1,90 @@
...
+        ASSERT_EQ(executor.get_wake_after_execute_flag(), true);
+      }
+      timer_count++;
+      if (timer_count > EXECUTION_COUNT) {
+        executor.cancel();
+      }
+    };
+
+  auto timer_ = node->create_wall_timer(
+    2s, timer_callback, cbg);
```
wjwwood, line 63:
```suggestion
  auto timer_ = node->create_wall_timer(2s, timer_callback, cbg);
```

PR #1164, opened 2020-06-10
Add message lost subscription event

ivanpauno:
See https://github.com/ros2/rmw/issues/226.


ivanpauno:
> The changes look fine, but do you plan to follow-up with changes to actually register/call the new callback type?

Good catch.
Up to now, I've tested that the rclpy side works well.
I have to check an example with rclcpp too.

ivanpauno:
I've successfully tested the `rclcpp` side too.
@jacobperron do you thing anything else is needed to merge [this group](https://github.com/ros2/rmw/issues/226#issuecomment-640890428) of PRs?

jacobperron:
> @jacobperron do you thing anything else is needed to merge this group of PRs?

Nothing comes to mind.

PR #1167, opened 2020-06-11
Check if context is valid when looping in spin_some

ivanpauno:
Taken from https://github.com/ros2/rclcpp/pull/1156#discussion_r438432897, so we can backport this change.

PR #1023, opened 2020-03-11
Fix spin_until_future_complete: check spinning value

DongheeYe:
…h executor is cancaled

Signed-off-by: Donghee Ye <donghee.ye@samsung.com>

rclcpp::executor::spin_until_future_complete is also finished when executor is canceled.

Related to #1022

PR #1171, opened 2020-06-11
[dashing] Backports #1023 #1167

ivanpauno:
This PR will be merged with `--ff-only`, to preserve commit hashes.
Backports #1023 #1167.

#1167 was modified, as https://github.com/ros2/rclcpp/pull/844 wasn't backported to eloquent.
I don't think it's a good idea to backport #844, as it introduces a behavior change that can affect some use cases (see discussion in https://github.com/ros2/rclcpp/pull/1156).

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -35,6 +35,7 @@
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/utilities.hpp"
 #include "rclcpp/visibility_control.hpp"
+#include "rclcpp/scope_exit.hpp"
 
 namespace rclcpp
 {
@@ -243,9 +244,14 @@ public:
     }
     std::chrono::nanoseconds timeout_left = timeout_ns;
 
-    while (rclcpp::ok(this->context_)) {
+    if (spinning.exchange(true)) {
+      throw std::runtime_error("spin_until_future_complete() called while already spinning");
+    }
+    RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
+    while (rclcpp::ok(this->context_) && spinning.load()) {
       // Do one item of work.
-      spin_once(timeout_left);
+      spin_once_impl(timeout_left);
+
       // Check if the future is set, return SUCCESS if it is.
       status = future.wait_for(std::chrono::seconds(0));
       if (status == std::future_status::ready) {
@@ -367,6 +373,10 @@ protected:
 
 private:
   RCLCPP_DISABLE_COPY(Executor)
+
+  RCLCPP_PUBLIC
+  void
+  spin_once_impl(std::chrono::nanoseconds timeout);
 
   std::list<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr> weak_nodes_;
   std::list<const rcl_guard_condition_t *> guard_conditions_;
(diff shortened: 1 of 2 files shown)


ivanpauno:
@nuclearsandwich friednly ping

nuclearsandwich:
> I will wait until tomorrow to merge, so all nightly jobs were ran once with these changes in.

This line doesn't make sense to me. Unmerged changes won't be part of nightlies and ci.ros2.org doesn't run Dashing nightlies.

ivanpauno:
> This line doesn't make sense to me. Unmerged changes won't be part of nightlies and ci.ros2.org doesn't run Dashing nightlies.

Sorry, that is old.

I merged the original PR the same day I opened the backport.
CI jobs in the original PR were more limited than full CI.
I wanted to double check that there were no regressions in full CI.

That has been checked, so the comment doesn't apply anymore.

ivanpauno:
The test failing in the PR checker is a known flaky test.

ivanpauno:
@nuclearsandwich friendly ping

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -368,6 +374,10 @@ class Executor
 private:
   RCLCPP_DISABLE_COPY(Executor)
 
+  RCLCPP_PUBLIC
+  void
+  spin_once_impl(std::chrono::nanoseconds timeout);
```
hidmic, line 379: @ivanpauno I'd think the compiler needs no runtime tricks to use this method and so it _should_ be ABI compatible, but why the public visibility?

ivanpauno: > @ivanpauno I'd think the compiler needs no runtime tricks to use this method and so it _should_ be ABI compatible, but why the public visibility?

In the original PR, this method was protected, and not private.
I forgot changing the RCLCPP_PUBLIC macro to RCLCPP_LOCAL (though the change isn't completely necessary).

ivanpauno: Fixed

ivanpauno:
@hidmic based on the [failing PR check](http://build.ros2.org/job/Dpr__rclcpp__ubuntu_bionic_amd64/267/console), `spin_until_future_complete` will be solved at compile time in another package and it needs to link against `spin_once_impl`.

I will undo the last change and use `RCLCPP_PUBLIC` again.

hidmic:
> @hidmic based on the failing PR check, spin_until_future_complete will be solved at compile time in another package and it needs to link against spin_once_impl.

Oh, right, this is a template class. Hmm, that means that all downstream packages have to be recompiled to make use of this :/

ivanpauno:
>  Oh, right, this is a template class.

The method is a template, not the class.

> Hmm, that means that all downstream packages have to be recompiled to make use of this :/

Users will need to recompile to get the fix, but users won't have to recompile if they are not interested in this fix (i.e. this is ABI compatible).
For fixes in templates (and in inline functions), there's no other way to get a fix than recompiling (in the extreme case of a header only library, you always have to recompile to get the latest fixes).

hidmic:
> Users will need to recompile to get the fix, but users won't have to recompile if they are not interested in this fix (i.e. this is ABI compatible).

Yeah, I know. It striked me as a bit odd, but it won't do any harm. The added `throw` could mean trouble but I'm sure `spin_once()` was already throwing.

Issue #1177, opened 2020-06-16
Double free in create_wall_timer if an invalid period is passed


brawner:
In [`create_wall_timer()` in `create_timer.hpp`](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/create_timer.hpp#L105-L110), if a bad period duration is passed (either < 0 or > nanoseconds::max()) then rcl will throw an error. During the destructor of WallTimer, it will call cancel, which results in a double free. I haven't found exactly where.

This issue is to track the several PRs I think will be necessary to resolve the cascade of errors present.

brawner:
Add PR to check arguments in `create_wall_timer`: https://github.com/ros2/rclcpp/pull/1178

brawner:
For reference, this is the error message:
```
[20.871s] 8: [ RUN      ] TestCreateTimer.call_wall_timer_with_bad_arguments
[20.874s] 8: [ERROR] [1592350214.053122631] [rclcpp]: Couldn't initialize rcl timer handle: timer period must be non-negative, at /home/brawner/workspace/ros2_master/src/ros2/rcl/rcl/src/rcl/timer.c:139
[20.874s] 8: 
[20.874s] 8: terminate called after throwing an instance of 'std::runtime_error'
[20.874s] 8:   what():  Couldn't cancel timer: timer is invalid, at /home/brawner/workspace/ros2_master/src/ros2/rcl/rcl/src/rcl/timer.c:379
[20.938s] 8: -- run_test.py: return code -6
```

fujitatomoya:
@brawner 

> During the destructor of WallTimer, it will call cancel, which results in a double free.

i do not think of `doube free` case either, but exception in destructor.

checking argument is also better, but we could change like `only issue cancel on initialized rcl_timer_t successfully`, i think that would be less exception.

fujitatomoya:
problem confirmed,

```
# strace ./build/ros2_test_prover/ros2_644
...<snip>
write(1, "Callback: Getting logger", 24Callback: Getting logger) = 24
write(1, "\n", 1
)                       = 1
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x38} ---
+++ killed by SIGSEGV (core dumped) +++
```

brawner:
Agreed @fujitatomoya. This was the intended second PR to followup and resolve the issue. It throws an exception in the constructor if `rcl_timer_init` fails.

[...]

PR #1178, opened 2020-06-16
Check period duration in create_wall_timer

brawner:
This is the first of the fixes related to #1177. It adds checks for negative duration values and duration values larger than nanoseconds::max(). As described in the included comment, it's difficult to solve the general problem of comparing one duration value to another for safe casting. I intend to follow up with a PR, (though not immediately) to rcpputils to provide a more general solution.

### DIFF
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -76,7 +76,7 @@ create_timer(
  * \tparam DurationRepT
  * \tparam DurationT
  * \tparam CallbackT
- * \param period period to exectute callback
+ * \param period period to execute callback
  * \param callback callback to execute via the timer period
  * \param group
  * \param node_base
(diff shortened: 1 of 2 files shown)


brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11177/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6455/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9119/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11090/)

rclcpp/test/rclcpp/test_create_timer.cpp
```diff
@@ -18,10 +18,10 @@
 #include <chrono>
 #include <memory>
 
+#include "node_interfaces/node_wrapper.hpp"
```
fujitatomoya, line 21: nit: why this change is needed?

brawner: It puts the includes in alphabetical order, which admittedly is unrelated to this PR, but the whole effort falls under "Increasing test coverage of rclcpp". Once this test is added, there then create_timer.hpp is fully covered.

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -102,6 +102,26 @@ create_wall_timer(
...
+  // a signed integer, which is undefined behavior. Checking whether any std::chrono::duration is
+  // greater than nanoseconds::max() is a difficult general problem. This is a more conservative
+  // version of Howard Hinnant's (the <chrono> guy>) response here:
+  // https://stackoverflow.com/a/44637334/2089061
+  // However, this doesn't solve the issue for all possible duration types of period.
+  // Follow-up issue: https://github.com/ros2/rclcpp/issues/1177
+  constexpr auto ns_max_as_double =
+    std::chrono::duration_cast<
+    std::chrono::duration<double, std::chrono::nanoseconds::period>
+    >(std::chrono::nanoseconds::max() - std::chrono::duration<DurationRepT, DurationT>(1));
```
hidmic, line 119: @brawner why substracting an `std::chrono::duration<DurationRepT, DurationT>` unit?

brawner: The DurationRepT of nanoseconds is a 64-bit signed int, however the precision of a 64-bit double is such that the duration cast to double might actually cause period to round down to be less than nanoseconds::max(). Essentially, 

```
duration_cast<duration<double, nanoseconds::period>>(nanoseconds::max() + 1us) == duration_cast<duration<double, nanoseconds::period>>(nanoseconds::max())
```

But, this would overflow.
```
duration_cast<nanoseconds>(1us + nanoseconds::max())
```

In the stackover flow response I posted, they would cast period to a double check if it would overflow, then cast the double to nanoseconds, which also works. I just didn't want to change the behavior for the by far most common case of period << nanoseconds::max().

hidmic: Mind to update the comment to reflect that rationale?

brawner: Added a comment and split it out into a separate variable for clarity.

brawner:
Added some more comments, and a second check in case overflow occurs in some bizarre circumstance. Unfortunately, overflow to a negative is not guaranteed behavior of signed-integer overflow because it's implementation specific.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11208/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6481/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9167/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11125/)

brawner:
RPR build failed for an unrelated reason. MacOS test failures are unrelated and a known issue.

PR #1187, opened 2020-06-19
Update tracetools' QL to 2 in rclcpp's QD

christophebedard:
`tracetools`' QL was just bumped to 2: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/merge_requests/179


christophebedard:
Does this need a second approval?

PR #901, opened 2019-10-24
Type conversions fixes

monidzik:
Couple of improvements in terms of type conversions; mostly adding `static_casts`.

Issue #525, opened 2018-07-26
When can a Duration be negative?


dhood:
Spinning off from https://github.com/ros2/geometry2/pull/67#discussion_r205326838 so I don't hijack that thread (and so the convo doesn't get lost).

I thought that Durations could not store negative values but from @tfoote's comment it sounds like that understanding is incorrect. The source of my misunderstanding is this line:

https://github.com/ros2/rclcpp/blob/af6e86c522d674a58f935cabceea2108c7f64589/rclcpp/src/rclcpp/duration.cpp#L92

Should that check be removed? Or is it appropriate in that situation?

dhood:
this question can be extended to Time as well, because it also seems to have signed storage underneath, but has similar exceptions e.g.:

https://github.com/ros2/rclcpp/blob/3a503685bf83280ffcde3ecb2d33a7d8de7dade2/rclcpp/src/rclcpp/time.cpp#L52

tfoote:
Time cannot have a negative value. It's nanoseconds since the epoch. That looks like a copy and paste error from the time class. There's no such thing as a duration point.

The time point is a point on the timeline. A duration is the difference between two time points on the timeline. Time can be differentiated to a duration, Duration can be added to time, But time cannot be added together. 1990 + 1991 doesn't make any sense.

There are tests for negative values here: https://github.com/ros2/rclcpp/blob/master/rclcpp/test/test_duration.cpp but clearly not coverage on negative values being assigned.

dhood:
I recall now the storage type changed from unsigned to signed: https://github.com/ros2/rclcpp/pull/429

Here's what I suspect:

- At some point, Time couldn't be negative, so the exceptions were added.
- Time can be negative now, the exceptions should have been removed during that change but were overlooked.
- Duration can be negative, but was copied from Time so has exception(s) by accident that should be removed.

dhood:
Durations can be negative and the exceptions were removed in https://github.com/ros2/rclcpp/pull/527 for rclcpp and https://github.com/ros2/rclpy/pull/220 for rclpy.

[...]

PR #1188, opened 2020-06-19
Fix conversion of negative durations to messages

meyerj:
This pull request is a follow-up on https://github.com/ros2/rclcpp/pull/901, to fix conversions of negative `Duration` (and `Time`) instances to the respective message times in `builtin_interfaces`, where the `nanosec` part is represented as an unsigned `uint32` integer type.

The implicit conversion from `rclcpp::Duration` to `builtin_interfaces::msg::Duration` was actually wrong for negative values, because the default integer division in C/C++ rounds towards zero and the modulo operator returns a negative remainder for negative numerators. With the previous implementation this negative remainder was then cast to `uint32_t` without a range check, which then results in timestamps that are 2³² nanoseconds or 4.294967296 seconds too big if converted back to `rclcpp::Duration`.

For
```cpp
rclcpp::Duration duration(-1);
builtin_interfaces::msg::Duration msg_duration = duration;
std::cout
  << "sec: " << msg_duration.sec << '\n'
  << "nanosec: " << msg_duration.nanosec << '\n'
  << "back to Duration(msg_duration).nanoseconds(): " << Duration(msg_duration).nanoseconds() << '\n';
```
the output was
```
sec: 0
nanosec: 4294967295
back to Duration(msg_duration).nanoseconds(): 4294967295
```
instead of
```
sec: -1
nanosec: 999999999
back to Duration(msg_duration).nanoseconds(): -1
```

The conversion from `rclcpp::Duration` to `rmw_time_t` had a similar problem. In `rmw_time_t` both, the `sec` and `nsec` parts are unsigned 64 bit types:
```cpp
typedef struct RMW_PUBLIC_TYPE rmw_time_t
{
  uint64_t sec;
  uint64_t nsec;
} rmw_time_t;
```
So the type cannot represent negative values. Was it intentional to represent `rmw_time_t` with 128 bits in total, or should it have been `uint32_t` instead? I added a check that throws `std::runtime_error` when trying to call `Duration::to_rmw_time()` for negative values.

For `rclcpp::Time` I am not sure whether it is supposed to have negative values (times before the epoch) or not. There was a discussion about that in https://github.com/ros2/rclcpp/issues/525. Some constructors and assignment operators throw, others don't. For example it is possible to construct a negative `rclcpp::Time` instance from a negative `int64_t` nanosecond value:

https://github.com/ros2/rclcpp/blob/79d1ba1cceab26791496b3ead9873dee5232ea9a/rclcpp/src/rclcpp/time.cpp#L60-L64

Because of that, and because also the internal storage type `rcl_time_point_value_t` and [`builtin_interfaces/Time`](https://github.com/ros2/rcl_interfaces/blob/5a4a3bd4da4cfc35ad377f2966714321b2951c89/builtin_interfaces/msg/Time.msg) can also hold negative values, I applied the same patch for the conversion to the message type as in `rclcpp::Duration`. If negative values should be forbidden for `rclcpp::Time`, which is the conclusion of https://github.com/ros2/rclcpp/issues/525#issuecomment-414425011 if I understand correctly, then in my opinion this should be addressed in a separate pull request and the added checks (and unit test) can be removed again.

All other changes are mainly cosmetic, either to make implicit casts between signed and unsigned integer types explicit, or to match the implementation of constructors and respective assignment operators. In order to at least emit a warning and to avoid similar problems with unintended integer or floating point conversions that may alter the value, I suggest to add `-Wconversion` to the compiler options of `rclcpp`. It is not included in `-Wall` or `-Wextra`. But this warning option would not have helped in case of the buggy `Duration` conversions because the bad casts without checking for over- or underflow were explicit. Also the explicit casts added in https://github.com/ros2/rclcpp/pull/901 might hide problems in case the argument is out of range of the target type.

The copy constructors and assignment operators are equivalent to the default implementations, in the sense that they just copy the one and only member of the `Duration` or `Time` instance. Maybe this used to be different in an earlier version? The assignment of `rcl_time_.nanoseconds = rhs.rcl_time_.nanoseconds` in the copy constructor `Time::Time(const Time &)` was redundant because `rcl_time_` is already copied in the initializer list.

There is another problem with the definition of time-related macros defined in [rcutils/time.h](https://github.com/ros2/rcutils/blob/8c1d190fc5e6425f83c3e8716a5ee861bbf82cf1/include/rcutils/time.h) which are also indirectly used for the `Duration` and `Time` implementations in `rclcpp`: They don't have parentheses around the macro argument:

```cpp
/// Convenience macro to convert seconds to nanoseconds.
#define RCUTILS_S_TO_NS(seconds) (seconds * (1000LL * 1000LL * 1000LL))
/// Convenience macro to convert milliseconds to nanoseconds.
#define RCUTILS_MS_TO_NS(milliseconds) (milliseconds * (1000LL * 1000LL))
/// Convenience macro to convert microseconds to nanoseconds.
#define RCUTILS_US_TO_NS(microseconds) (microseconds * 1000LL)

/// Convenience macro to convert nanoseconds to seconds.
#define RCUTILS_NS_TO_S(nanoseconds) (nanoseconds / (1000LL * 1000LL * 1000LL))
/// Convenience macro to convert nanoseconds to milliseconds.
#define RCUTILS_NS_TO_MS(nanoseconds) (nanoseconds / (1000LL * 1000LL))
/// Convenience macro to convert nanoseconds to microseconds.
#define RCUTILS_NS_TO_US(nanoseconds) (nanoseconds / 1000LL)
```

So `RCUTILS_S_TO_NS(1 + 1)` is not `2000000000LL` as one might expect, but `1000000001LL`, because of operator precedence rules. This case is not triggered in `rclcpp`. I will open a separate pull request in https://github.com/ros2/rcutils for that.

Last but not least, new unit test cases have been added to `test_duration.cpp` and `test_time.cpp` to cover all the conversions and potential failure cases above. Without the patches in `time.cpp` and `duration.cpp` the new tests would fail as follows:
```
root@c610b7e4c97e:/ros2_ws# ./build/rclcpp/test/test_duration 
Running main() from /opt/ros/foxy/src/gtest_vendor/src/gtest_main.cc
[==========] Running 8 tests from 1 test case.
[----------] 8 tests from TestDuration
[ RUN      ] TestDuration.operators
[       OK ] TestDuration.operators (0 ms)
[ RUN      ] TestDuration.chrono_overloads
[       OK ] TestDuration.chrono_overloads (0 ms)
[ RUN      ] TestDuration.overflows
[       OK ] TestDuration.overflows (1 ms)
[ RUN      ] TestDuration.negative_duration
[       OK ] TestDuration.negative_duration (0 ms)
[ RUN      ] TestDuration.maximum_duration
[       OK ] TestDuration.maximum_duration (0 ms)
[ RUN      ] TestDuration.from_seconds
[       OK ] TestDuration.from_seconds (0 ms)
[ RUN      ] TestDuration.std_chrono_constructors
[       OK ] TestDuration.std_chrono_constructors (0 ms)
[ RUN      ] TestDuration.conversions
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:203: Failure
  duration_msg.sec
    Which is: 0
  -1
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:204: Failure
  duration_msg.nanosec
    Which is: 3794967296
  HALF_SEC_IN_NS
    Which is: 500000000
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:205: Failure
  rclcpp::Duration(duration_msg).nanoseconds()
    Which is: 3794967296
  -HALF_SEC_IN_NS
    Which is: -500000000
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:207: Failure
Expected: duration.to_rmw_time() throws an exception of type std::overflow_error.
  Actual: it throws nothing.
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:220: Failure
Expected: duration.to_rmw_time() throws an exception of type std::overflow_error.
  Actual: it throws nothing.
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:229: Failure
  duration_msg.sec
    Which is: -1
  -2
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:230: Failure
  duration_msg.nanosec
    Which is: 3794967296
  HALF_SEC_IN_NS
    Which is: 500000000
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:231: Failure
  rclcpp::Duration(duration_msg).nanoseconds()
    Which is: 2794967296
  -ONE_AND_HALF_SEC_IN_NS
    Which is: -1500000000
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_duration.cpp:233: Failure
Expected: duration.to_rmw_time() throws an exception of type std::overflow_error.
  Actual: it throws nothing.
[  FAILED  ] TestDuration.conversions (1 ms)
[----------] 8 tests from TestDuration (2 ms total)

[==========] 8 tests from 1 test case ran. (2 ms total)
[  PASSED  ] 7 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestDuration.conversions

 1 FAILED TEST

root@c610b7e4c97e:/ros2_ws# ./build/rclcpp/test/test_time
Running main() from /opt/ros/foxy/src/gtest_vendor/src/gtest_main.cc
[==========] Running 8 tests from 1 test case.
[----------] 8 tests from TestTime
[ RUN      ] TestTime.clock_type_access
[       OK ] TestTime.clock_type_access (0 ms)
[ RUN      ] TestTime.clock_jump_callback_destruction_order
[       OK ] TestTime.clock_jump_callback_destruction_order (0 ms)
[ RUN      ] TestTime.time_sources
[       OK ] TestTime.time_sources (0 ms)
[ RUN      ] TestTime.conversions
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_time.cpp:167: Failure
  time_msg.sec
    Which is: 0
  -1
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_time.cpp:168: Failure
  time_msg.nanosec
    Which is: 3794967296
  HALF_SEC_IN_NS
    Which is: 500000000
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_time.cpp:171: Failure
Expected: { rclcpp::Time negative_time(time_msg); } throws an exception.
  Actual: it doesn't.
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_time.cpp:193: Failure
  time_msg.sec
    Which is: -1
  -2
/ros2_ws/src/rclcpp/rclcpp/test/rclcpp/test_time.cpp:194: Failure
  time_msg.nanosec
    Which is: 3794967296
  HALF_SEC_IN_NS
    Which is: 500000000
[  FAILED  ] TestTime.conversions (2 ms)
[ RUN      ] TestTime.operators
[       OK ] TestTime.operators (1 ms)
[ RUN      ] TestTime.overflow_detectors
[       OK ] TestTime.overflow_detectors (0 ms)
[ RUN      ] TestTime.overflows
[       OK ] TestTime.overflows (1 ms)
[ RUN      ] TestTime.seconds
[       OK ] TestTime.seconds (0 ms)
[----------] 8 tests from TestTime (4 ms total)

[==========] 8 tests from 1 test case ran. (4 ms total)
[  PASSED  ] 7 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestTime.conversions

 1 FAILED TEST
 ```

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -23,7 +23,11 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wconversion -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
+  # About -Wno-sign-conversion: With Clang, -Wconversion implies -Wsign-conversion. There are a number of
+  # implicit sign conversions in rclcpp and gtest.cc, see https://ci.ros2.org/job/ci_osx/9265/.
+  # Hence disabling -Wsign-conversion for now until all those have eventually been fixed.
+  # (from https://github.com/ros2/rclcpp/pull/1188#issuecomment-650229140)
+  add_compile_options(-Wall -Wextra -Wconversion -Wno-sign-conversion -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
 endif()
 
 set(${PROJECT_NAME}_SRCS
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -23,7 +23,7 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
+  add_compile_options(-Wall -Wextra -Wconversion -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
 endif()
 
 set(${PROJECT_NAME}_SRCS
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -66,14 +66,14 @@ Duration::Duration(const rcl_duration_t & duration)
 Duration::operator builtin_interfaces::msg::Duration() const
 {
   builtin_interfaces::msg::Duration msg_duration;
-  constexpr rcl_duration_value_t kRemainder = RCL_S_TO_NS(1);
-  const auto result = std::div(rcl_duration_.nanoseconds, kRemainder);
+  constexpr rcl_duration_value_t kDivisor = RCL_S_TO_NS(1);
+  const auto result = std::div(rcl_duration_.nanoseconds, kDivisor);
   if (result.rem >= 0) {
     msg_duration.sec = static_cast<std::int32_t>(result.quot);
     msg_duration.nanosec = static_cast<std::uint32_t>(result.rem);
   } else {
     msg_duration.sec = static_cast<std::int32_t>(result.quot - 1);
-    msg_duration.nanosec = static_cast<std::uint32_t>(kRemainder + result.rem);
+    msg_duration.nanosec = static_cast<std::uint32_t>(kDivisor + result.rem);
   }
   return msg_duration;
 }
@@ -84,8 +84,7 @@ Duration::operator=(const Duration & rhs) = default;
 Duration &
 Duration::operator=(const builtin_interfaces::msg::Duration & duration_msg)
 {
-  rcl_duration_.nanoseconds = RCL_S_TO_NS(static_cast<rcl_duration_value_t>(duration_msg.sec));
-  rcl_duration_.nanoseconds += static_cast<rcl_duration_value_t>(duration_msg.nanosec);
+  *this = Duration(duration_msg);
   return *this;
 }
 
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -47,17 +47,14 @@ Duration::Duration(std::chrono::nanoseconds nanoseconds)
   rcl_duration_.nanoseconds = nanoseconds.count();
 }
 
-Duration::Duration(const Duration & rhs)
-{
-  rcl_duration_.nanoseconds = rhs.rcl_duration_.nanoseconds;
-}
+Duration::Duration(const Duration & rhs) = default;
 
 Duration::Duration(
   const builtin_interfaces::msg::Duration & duration_msg)
 {
   rcl_duration_.nanoseconds =
-    static_cast<rcl_duration_value_t>(RCL_S_TO_NS(duration_msg.sec));
-  rcl_duration_.nanoseconds += duration_msg.nanosec;
+    RCL_S_TO_NS(static_cast<rcl_duration_value_t>(duration_msg.sec));
+  rcl_duration_.nanoseconds += static_cast<rcl_duration_value_t>(duration_msg.nanosec);
 }
 
 Duration::Duration(const rcl_duration_t & duration)
@@ -69,24 +66,26 @@ Duration::Duration(const rcl_duration_t & duration)
 Duration::operator builtin_interfaces::msg::Duration() const
 {
   builtin_interfaces::msg::Duration msg_duration;
-  msg_duration.sec = static_cast<std::int32_t>(RCL_NS_TO_S(rcl_duration_.nanoseconds));
-  msg_duration.nanosec =
-    static_cast<std::uint32_t>(rcl_duration_.nanoseconds % (1000 * 1000 * 1000));
+  constexpr rcl_duration_value_t kRemainder = RCL_S_TO_NS(1);
+  const auto result = std::div(rcl_duration_.nanoseconds, kRemainder);
+  if (result.rem >= 0) {
+    msg_duration.sec = static_cast<std::int32_t>(result.quot);
+    msg_duration.nanosec = static_cast<std::uint32_t>(result.rem);
+  } else {
+    msg_duration.sec = static_cast<std::int32_t>(result.quot - 1);
+    msg_duration.nanosec = static_cast<std::uint32_t>(kRemainder + result.rem);
+  }
   return msg_duration;
 }
 
 Duration &
-Duration::operator=(const Duration & rhs)
-{
-  rcl_duration_.nanoseconds = rhs.rcl_duration_.nanoseconds;
-  return *this;
-}
+Duration::operator=(const Duration & rhs) = default;
 
 Duration &
 Duration::operator=(const builtin_interfaces::msg::Duration & duration_msg)
 {
-  rcl_duration_.nanoseconds = RCL_S_TO_NS(static_cast<int64_t>(duration_msg.sec));
-  rcl_duration_.nanoseconds += duration_msg.nanosec;
+  rcl_duration_.nanoseconds = RCL_S_TO_NS(static_cast<rcl_duration_value_t>(duration_msg.sec));
+  rcl_duration_.nanoseconds += static_cast<rcl_duration_value_t>(duration_msg.nanosec);
   return *this;
 }
 
@@ -230,6 +229,10 @@ Duration::seconds() const
 rmw_time_t
 Duration::to_rmw_time() const
 {
+  if (rcl_duration_.nanoseconds < 0) {
+    throw std::runtime_error("rmw_time_t cannot be negative");
+  }
+
   // reuse conversion logic from msg creation
   builtin_interfaces::msg::Duration msg = *this;
   rmw_time_t result;
--- a/rclcpp/src/rclcpp/node_options.cpp
+++ b/rclcpp/src/rclcpp/node_options.cpp
@@ -338,7 +338,7 @@ NodeOptions::get_domain_id_from_env() const
   _dupenv_s(&ros_domain_id, &ros_domain_id_size, env_var);
 #endif
   if (ros_domain_id) {
-    uint32_t number = strtoul(ros_domain_id, NULL, 0);
+    uint32_t number = static_cast<uint32_t>(strtoul(ros_domain_id, NULL, 0));
     if (number == (std::numeric_limits<uint32_t>::max)()) {
 #ifdef _WIN32
       // free the ros_domain_id before throwing, if getenv was used on Windows
--- a/rclcpp/test/rclcpp/test_parameter.cpp
+++ b/rclcpp/test/rclcpp/test_parameter.cpp
@@ -32,7 +32,7 @@ protected:
   }
 };
 
-TEST(TestParameter, not_set_variant) {
+TEST_F(TestParameter, not_set_variant) {
   // Direct instantiation
   rclcpp::Parameter not_set_variant;
   EXPECT_EQ(rclcpp::PARAMETER_NOT_SET, not_set_variant.get_type());
@@ -58,7 +58,7 @@ TEST(TestParameter, not_set_variant) {
     rclcpp::Parameter::from_parameter_msg(not_set_param).get_type());
 }
 
-TEST(TestParameter, bool_variant) {
+TEST_F(TestParameter, bool_variant) {
   // Direct instantiation
   rclcpp::Parameter bool_variant_true("bool_param", true);
   EXPECT_EQ("bool_param", bool_variant_true.get_name());
@@ -116,7 +116,7 @@ TEST(TestParameter, bool_variant) {
     bool_variant_false.get_value_message().type);
 }
 
-TEST(TestParameter, integer_variant) {
+TEST_F(TestParameter, integer_variant) {
   const int TEST_VALUE {42};
 
   // Direct instantiation
@@ -164,7 +164,7 @@ TEST(TestParameter, integer_variant) {
     from_msg.get_value_message().type);
 }
 
-TEST(TestParameter, long_integer_variant) {
+TEST_F(TestParameter, long_integer_variant) {
   const int64_t TEST_VALUE {std::numeric_limits<int64_t>::max()};
 
   // Direct instantiation
@@ -212,7 +212,7 @@ TEST(TestParameter, long_integer_variant) {
     from_msg.get_value_message().type);
 }
 
-TEST(TestParameter, float_variant) {
+TEST_F(TestParameter, float_variant) {
   const float TEST_VALUE {42.0f};
 
   // Direct instantiation
@@ -260,7 +260,7 @@ TEST(TestParameter, float_variant) {
     from_msg.get_value_message().type);
 }
 
-TEST(TestParameter, double_variant) {
+TEST_F(TestParameter, double_variant) {
   const double TEST_VALUE {-42.1};
 
   // Direct instantiation
@@ -308,7 +308,7 @@ TEST(TestParameter, double_variant) {
     from_msg.get_value_message().type);
 }
 
-TEST(TestParameter, string_variant) {
+TEST_F(TestParameter, string_variant) {
   const std::string TEST_VALUE {"ROS2"};
 
   // Direct instantiation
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

File: rclcpp/src/rclcpp/node_options.cpp [hunk 1]
```diff
@@ -338,7 +338,7 @@ NodeOptions::get_domain_id_from_env() const
   _dupenv_s(&ros_domain_id, &ros_domain_id_size, env_var);
 #endif
   if (ros_domain_id) {
-    uint32_t number = strtoul(ros_domain_id, NULL, 0);
+    uint32_t number = static_cast<uint32_t>(strtoul(ros_domain_id, NULL, 0));
```
meyerj, line 341: This static_cast just makes explicit what the compiler would also have done implicitly.

The check below may also not do what is expected: [`strtoul()`](https://en.cppreference.com/w/cpp/string/byte/strtoul) converts the string to `unsigned long`, and returns `ULONG_MAX` in case of an overflow. On my machine (Linux 64-bit) `unsigned long` is a 64 bit type, which is also why the implicit conversion to `uint32_t` triggered a compiler warning with `-Wconversion`. But that also means that only values exceeding `18446744073709551617` and all multiples of `4294967296` minus 1 trigger the run-time exception. `4294967296` is silently cast to `0`, `4294967297` to `1` etc.

Illegal values like `"foo"` would also silently be interpreted as `0` without an error, because `strtoul` simply stops parsing at the first invalid character. You would have to pass a `char **` as the second argument `str_end` instead of `NULL` and check whether it points to the end of the string (the terminating null character).

The conversions should probably use C++ [stringstreams](https://en.cppreference.com/w/cpp/io/basic_istringstream) or the C++11 equivalent [std::stoul](https://en.cppreference.com/w/cpp/string/basic_string/stoul) with an additional range check, instead of error-prone C functions. Or, if the function must be implemented in C to avoid code duplication with `rcl`, as stated in @wjwwood's comment above, the solution would be to use `unsigned long` instead of `uint32_t` without a cast, check that the `str_end` actually points to the terminating null character after conversion, and to add an explicit range check before the cast to the desired output type.

Apparently the issue is already discussed in https://github.com/ros2/rcl/pull/689.

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -47,17 +47,14 @@ Duration::Duration(std::chrono::nanoseconds nanoseconds)
...
-}
+Duration::Duration(const Duration & rhs) = default;
 
 Duration::Duration(
   const builtin_interfaces::msg::Duration & duration_msg)
 {
   rcl_duration_.nanoseconds =
-    static_cast<rcl_duration_value_t>(RCL_S_TO_NS(duration_msg.sec));
-  rcl_duration_.nanoseconds += duration_msg.nanosec;
+    RCL_S_TO_NS(static_cast<rcl_duration_value_t>(duration_msg.sec));
```
ivanpauno, line 56: why changing the order here?
https://github.com/ros2/rcutils/blob/e7eb74c7a11cf207aa9330d77f44098339528cac/include/rcutils/time.h#L30
https://github.com/ros2/rcutils/blob/e7eb74c7a11cf207aa9330d77f44098339528cac/include/rcutils/time.h#L48

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -69,24 +66,26 @@ Duration::Duration(const rcl_duration_t & duration)
 Duration::operator builtin_interfaces::msg::Duration() const
 {
   builtin_interfaces::msg::Duration msg_duration;
-  msg_duration.sec = static_cast<std::int32_t>(RCL_NS_TO_S(rcl_duration_.nanoseconds));
-  msg_duration.nanosec =
-    static_cast<std::uint32_t>(rcl_duration_.nanoseconds % (1000 * 1000 * 1000));
+  constexpr rcl_duration_value_t kRemainder = RCL_S_TO_NS(1);
```
ivanpauno, line 69: maybe kDivisor?

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -69,24 +66,26 @@ Duration::Duration(const rcl_duration_t & duration)
...
-}
+Duration::operator=(const Duration & rhs) = default;
 
 Duration &
 Duration::operator=(const builtin_interfaces::msg::Duration & duration_msg)
 {
-  rcl_duration_.nanoseconds = RCL_S_TO_NS(static_cast<int64_t>(duration_msg.sec));
-  rcl_duration_.nanoseconds += duration_msg.nanosec;
+  rcl_duration_.nanoseconds = RCL_S_TO_NS(static_cast<rcl_duration_value_t>(duration_msg.sec));
+  rcl_duration_.nanoseconds += static_cast<rcl_duration_value_t>(duration_msg.nanosec);
```
ivanpauno, line 88: to avoid duplication, you can take advantage of the constructor taking a `builtin_interfaces::msg::Duration` and then use copy assignment. In that way, we won't need to update this method again in the future.
I don't mind strongly though

rclcpp/test/rclcpp/test_parameter.cpp
```diff
@@ -32,7 +32,7 @@ class TestParameter : public ::testing::Test
   }
 };
 
-TEST(TestParameter, not_set_variant) {
+TEST_F(TestParameter, not_set_variant) {
```
ivanpauno, line 35: the TEST -> TEST_F changes are a bit unrelated, and it would be great to do them in a separate PR.
I don't mind accepting them here though

rclcpp/test/rclcpp/test_time.cpp
```diff
@@ -86,44 +86,124 @@ TEST(TestTime, time_sources) {
   EXPECT_NE(0u, steady_now.nanosec);
 }
 
-TEST(TestTime, conversions) {
+static const int64_t HALF_SEC_IN_NS = 500 * 1000 * 1000;
+static const int64_t ONE_SEC_IN_NS = 1000 * 1000 * 1000;
```
ivanpauno, line 90: 2 * HALF_SEC_IN_NS?

rclcpp/src/rclcpp/duration.cpp
meyerj: For the result it should not matter in this case, because the result of `RCL_S_TO_NS(duration_msg.sec)` has type `int64_t` implicitly and the `static_cast<rcl_duration_value_t>` is a no-op.

It was mainly for consistency with
https://github.com/ros2/rclcpp/blob/683ac64ee4edf52ae3a4cd1a722c17c4bd81b562/rclcpp/src/rclcpp/duration.cpp#L84-L90

So this change addresses the
> All other changes are mainly cosmetic, either to make implicit casts between signed and unsigned integer types explicit, or to match the implementation of constructors and respective assignment operators.

part of the description.

rclcpp/src/rclcpp/duration.cpp
meyerj: Fine for me. But unless the compiler optimizes that, which is probably what happens, constructing a temporary and invoking the copy assignment involves an additional step.

I will apply the same suggestion to [`Time::operator=(const builtin_interfaces::msg::Time & time_msg)`](https://github.com/ros2/rclcpp/blob/683ac64ee4edf52ae3a4cd1a722c17c4bd81b562/rclcpp/src/rclcpp/time.cpp#L110) for consistency.

Strictly speaking the overloads could be removed completely then, because the respective constructors are not marked `explicit` and the compiler will implicitly construct a temporary `Time` instance and invoke the default copy-assignment operator. But doing so would break the ABI compatibility.

rclcpp/test/rclcpp/test_time.cpp
meyerj: I do not see how that would improve readability. But I can change to
```cpp
static const int64_t HALF_SEC_IN_NS = RCL_MS_TO_NS(500);
static const int64_t ONE_SEC_IN_NS = RCL_S_TO_NS(1);
static const int64_t ONE_AND_HALF_SEC_IN_NS = RCL_MS_TO_NS(1500);
```
?

The code has originally been copied from
https://github.com/ros2/rclcpp/blob/a667583821fb3b1c0830420ce3b0517df1e9f8c4/rclcpp/test/rclcpp/test_duration.cpp#L140-L141
and I added `ONE_SEC_IN_NS` to cover the case where the `nanosec` part is zero during message conversion.

rclcpp/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
+  add_compile_options(-Wall -Wextra -Wconversion -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
```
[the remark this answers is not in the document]
meyerj: That triggers many warnings in gtest headers and sources and one in [src/rclcpp/node_interfaces/node_parameters.cpp:199](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L199):
```
/opt/ros/foxy/src/rclcpp/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp: In function ‘rcl_interfaces::msg::SetParametersResult __check_parameter_value_in_range(const ParameterDescriptor&, const rclcpp::ParameterValue&)’:
/opt/ros/foxy/src/rclcpp/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp:199:13: error: conversion to ‘long unsigned int’ from ‘int64_t’ {aka ‘long int’} may change the sign of the result [-Werror=sign-conversion]
  199 |     if (((v - integer_range.from_value) % integer_range.step) == 0) {
      |          ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

So I suggest to leave that for a separate pull request.

rclcpp/src/rclcpp/duration.cpp
ivanpauno: > Fine for me. But unless the compiler optimizes that, which is probably what happens, constructing a temporary and invoking the copy assignment involves an additional step.

That's a far point. I prefer deduplication over a bit of extra performance, but it might be arguable.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11306/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6550/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9265/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11208/)

ivanpauno:
@meyerj It seems that in `clang` `Wconversion` implies `Wsign-conversion` ([docs](https://clang.llvm.org/docs/DiagnosticsReference.html#wconversion)), and thus this introduced 36 build warning on macOS.

You can either choose to:
- Fix the warnings
- Activate the warning in gcc and not in clang (extra cmake logic).
- Don't add the flag in this PR, and iterate in a follow-up PR.

meyerj:
> @meyerj It seems that in `clang` `Wconversion` implies `Wsign-conversion` ([docs](https://clang.llvm.org/docs/DiagnosticsReference.html#wconversion)), and thus this introduced 36 build warning on macOS.
> 
> I think you can choose to:
> 
> * Fix the warnings
> * Activate the warning in gcc and not in clang (extra cmake logic).
> * Don't add the flag in this PR, and iterate in a follow-up PR.

How about adding `-Wno-sign-conversion`?

ivanpauno:
> How about adding -Wno-sign-conversion?

:+1:
Consider adding a TODO clarifying why that flag was added.

ivanpauno:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9281/)

ivanpauno:
Thanks for the contribution @meyerj !

PR #1189, opened 2020-06-20
Unittests for memory strategy files, except allocator_memory_strategy

brawner:
Adds several tests for memory strategy. I will likely split this up into a couple of PRs. I still need to add coverage for memory_strategy.cpp.

### DIFF
--- a/rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
+++ b/rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
@@ -1,0 +1,64 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+
+#include "gtest/gtest.h"
+
+#include "rclcpp/strategies/message_pool_memory_strategy.hpp"
+#include "test_msgs/msg/empty.hpp"
+
+using rclcpp::strategies::message_pool_memory_strategy::MessagePoolMemoryStrategy;
+
+class TestMessagePoolMemoryStrategy : public ::testing::Test
+{
+public:
+  void SetUp()
+  {
+    message_memory_strategy_ =
+      std::make_shared<MessagePoolMemoryStrategy<test_msgs::msg::Empty, 1>>();
+  }
+
+protected:
+  std::shared_ptr<MessagePoolMemoryStrategy<test_msgs::msg::Empty, 1>> message_memory_strategy_;
+};
+
+TEST_F(TestMessagePoolMemoryStrategy, construct_destruct) {
+}
+
+TEST_F(TestMessagePoolMemoryStrategy, borrow_return) {
+  auto message = message_memory_strategy_->borrow_message();
+  ASSERT_NE(nullptr, message);
+
+  EXPECT_NO_THROW(message_memory_strategy_->return_message(message));
+}
+
+TEST_F(TestMessagePoolMemoryStrategy, borrow_too_many) {
+  auto message = message_memory_strategy_->borrow_message();
+  ASSERT_NE(nullptr, message);
+
+  // Size is 1, borrowing second time should fail
+  EXPECT_THROW(message_memory_strategy_->borrow_message(), std::runtime_error);
+  EXPECT_NO_THROW(message_memory_strategy_->return_message(message));
+}
+
+TEST_F(TestMessagePoolMemoryStrategy, return_unrecognized) {
+  auto message = message_memory_strategy_->borrow_message();
+  ASSERT_NE(nullptr, message);
+
+  auto unrecognized = std::make_shared<test_msgs::msg::Empty>();
+  // Unrecognized does not belong to pool
+  EXPECT_THROW(message_memory_strategy_->return_message(unrecognized), std::runtime_error);
+  EXPECT_NO_THROW(message_memory_strategy_->return_message(message));
+}
--- a/rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
+++ b/rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
@@ -18,7 +18,7 @@
 
 #include "rclcpp/strategies/message_pool_memory_strategy.hpp"
 #include "test_msgs/msg/empty.hpp"
-#include "../../utils/gtest_macros.hpp"
+#include "../../utils/rclcpp_gtest_macros.hpp"
 
 using rclcpp::strategies::message_pool_memory_strategy::MessagePoolMemoryStrategy;
 
@@ -54,7 +54,7 @@ TEST_F(TestMessagePoolMemoryStrategy, borrow_too_many) {
   ASSERT_NE(nullptr, message);
 
   // Size is 1, borrowing second time should fail
-  EXPECT_THROW_EQ(
+  RCLCPP_EXPECT_THROW_EQ(
     message_memory_strategy_->borrow_message(),
     std::runtime_error("Tried to access message that was still in use! Abort."));
   EXPECT_NO_THROW(message_memory_strategy_->return_message(message));
@@ -66,7 +66,7 @@ TEST_F(TestMessagePoolMemoryStrategy, return_unrecognized) {
 
   auto unrecognized = std::make_shared<test_msgs::msg::Empty>();
   // Unrecognized does not belong to pool
-  EXPECT_THROW_EQ(
+  RCLCPP_EXPECT_THROW_EQ(
     message_memory_strategy_->return_message(unrecognized),
     std::runtime_error("Unrecognized message ptr in return_message."));
   EXPECT_NO_THROW(message_memory_strategy_->return_message(message));
--- a/rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
+++ b/rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
@@ -18,6 +18,7 @@
 
 #include "rclcpp/strategies/message_pool_memory_strategy.hpp"
 #include "test_msgs/msg/empty.hpp"
+#include "../../utils/gtest_macros.hpp"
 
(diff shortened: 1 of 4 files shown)


brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11231/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6499/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9193/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11145/)

brawner:
[Build Status](https://ci.ros2.org/job/ci_linux/11247/)

brawner:
After splitting into three PRs:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11301/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6547/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9259/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11205/)

rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
```diff
@@ -0,0 +1,64 @@
...
+    message_memory_strategy_ =
+      std::make_shared<MessagePoolMemoryStrategy<test_msgs::msg::Empty, 1>>();
+  }
+
+protected:
+  std::shared_ptr<MessagePoolMemoryStrategy<test_msgs::msg::Empty, 1>> message_memory_strategy_;
+};
+
+TEST_F(TestMessagePoolMemoryStrategy, construct_destruct) {
+}
```
clalancette, line 38: I don't typically like tests with no checks in them; they basically only test whether this compiles and doesn't throw an exception.  Also, if the tests are ever built with optimization enabled, the whole thing might be optimized away.  Can we at least check that the `message_memory_strategy_` pointer is not null, and maybe check something simple on the class?

(this might be made a bit easier by making `TestMessagePoolMemoryStrategy` have an "is-a" relationship with `MessagePoolMemoryStrategy`, but I'll leave it up to you whether you want to do that).

rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  EXPECT_NO_THROW(message_memory_strategy_->return_message(message));
+}
+
+TEST_F(TestMessagePoolMemoryStrategy, borrow_too_many) {
+  auto message = message_memory_strategy_->borrow_message();
+  ASSERT_NE(nullptr, message);
+
+  // Size is 1, borrowing second time should fail
+  EXPECT_THROW(message_memory_strategy_->borrow_message(), std::runtime_error);
```
clalancette, line 52: I sometimes find that these checks are too generic.  In particular, there are lots of things in the system that can throw `std::runtime_error`, and there are several exception classes that derive from it.  So you are not always sure which exception you are getting.  In the past, I've used macros like this:

```
#define EXPECT_THROW_OF_TYPE(type, statement, msg)                                                           \
  do {                                                                                                       \
    try {                                                                                                    \
      statement;                                                                                             \
    } catch (type const& err) {                                                                              \
      if (std::string(err.what()).find(msg) == std::string::npos) {                                          \
        FAIL() << "Expected error msg containing:" << std::endl                                              \
               << msg << std::endl                                                                           \
               << "Saw error msg:" << std::endl                                                              \
               << err.what() << std::endl;                                                                   \
      }                                                                                                      \
    } catch (std::exception const& err) {                                                                    \
      FAIL() << "Expected " #type << std::endl << "Saw exception type: " << typeid(err).name() << std::endl; \
    }                                                                                                        \
  } while (0)

#define EXPECT_RUNTIME_THROW(st, msg) EXPECT_THROW_OF_TYPE(std::runtime_error, st, msg)
#define EXPECT_ARGUMENT_THROW(st, msg) EXPECT_THROW_OF_TYPE(std::invalid_argument, st, msg)
```

It's unfortunately quite wordy, but the benefit is that it tests that the exception is what you expect *and* that the error string is also what you expect.  What do you think about using something like that?

rclcpp/test/rclcpp/test_message_memory_strategy.cpp
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <memory>
+
+#include "rclcpp/message_memory_strategy.hpp"
+#include "test_msgs/msg/empty.hpp"
+
+TEST(TestMemoryStrategies, construct_destruct) {
```
clalancette, line 22: This is another test where I'd like to see some EXPECTs or ASSERTs.  In this case it seems pretty easy to add something since all of the member variables are public.

rclcpp/test/rclcpp/test_message_memory_strategy.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  auto allocator = std::make_shared<std::allocator<void>>();
+  rclcpp::message_memory_strategy::MessageMemoryStrategy<test_msgs::msg::Empty> memory_strategy2(
+    allocator);
+}
+
+TEST(TestMemoryStrategies, standard_allocation) {
+  rclcpp::message_memory_strategy::MessageMemoryStrategy<test_msgs::msg::Empty> memory_strategy;
+  auto default_message = memory_strategy.create_default();
+  ASSERT_NE(nullptr, default_message);
```
clalancette, line 33: I may be misunderstanding this, but it looks to me like `create_default` is actually a factory method that returns a `MessageMemoryStrategy`.  So I'd think a better test here would be:

```
auto memory_strategy = rclcpp::message_memory_strategy::MessageMemoryStrategy<test_msgs::msg::Empty>::create_default();
```

rclcpp/test/rclcpp/test_memory_strategy.cpp
```diff
@@ -0,0 +1,399 @@
...
+  void SetUp() override
+  {
+    rclcpp::init(0, nullptr);
+
+    // This doesn't test AllocatorMemoryStrategy directly, so we cast to the base class.
+    // AllocatorMemoryStrategy is more commonly used than MessagePoolMemoryStrategy
+    // so we use this derived class for these tests.
+    memory_strategy_ =
+      std::make_shared<
+      rclcpp::memory_strategies::allocator_memory_strategy::AllocatorMemoryStrategy<>>();
```
clalancette, line 55: This is actually a dumb C++ question, but: can we elide the `<>` after `AllocatorMemoryStrategy`?  Won't the compiler just use the default template argument in that case?

rclcpp/test/rclcpp/test_memory_strategy.cpp
```diff
@@ -0,0 +1,399 @@
...
+  {
+    return memory_strategy_;
+  }
+
+private:
+  std::shared_ptr<MemoryStrategy> memory_strategy_;
+};
+
+TEST_F(TestMemoryStrategy, construct_destruct) {
+}
```
clalancette, line 74: Another place where it would be nice to add some EXPECTs or ASSERTs.

rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
brawner: > they basically only test whether this compiles and doesn't throw an exception

And segfaults! I find them pretty useful just when I'm debugging unit tests, but checking the invariance of the class on construction is a good habit I think. Since there are no EXPECT's in the SetUp, it makes sense to add some stuff here.

Added a few nullptr checks.

rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
brawner: Finally got back to this. See PR #1232 for the added macros. I took a slightly different tact, what do you think?

rclcpp/test/rclcpp/test_memory_strategy.cpp
brawner: Done.

rclcpp/test/rclcpp/test_message_memory_strategy.cpp
brawner: Done.

rclcpp/test/rclcpp/test_message_memory_strategy.cpp
brawner: You are correct about that. Fixed the test.

brawner:
I rebased this on #1232 to add the gtest macros to the unit test.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11511/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6692/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9426/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11395/)

rclcpp/test/rclcpp/test_memory_strategy.cpp
clalancette: Ha, OK.  Maybe @wjwwood can enlighten us here.

But this isn't really an issue, so I'll resolve this thread.

rclcpp/test/rclcpp/strategies/test_message_pool_memory_strategy.cpp
clalancette: Overall, looking good.  I left some comments on #1232 , so resolving here.

brawner:
Rebased with updates to #1232

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11523/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6698/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9432/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11421/)

brawner:
Rebased on to master after merging #1232

brawner:
Forgot to --signoff

clalancette:
As far as I can tell, Windows CI should succeed in Release (or None) mode.  At least, the nightly succeeded: https://ci.ros2.org/view/nightly/job/nightly_win_rel/1615/ .

It's in debug mode that Windows CI is currently failing.

rclcpp/test/rclcpp/test_memory_strategy.cpp
wjwwood: If there's no arguments (it is a class) it cannot deduce it without the `<>`. Apparently this might be possible in C++17 and newer enough versions of compilers:

https://en.cppreference.com/w/cpp/language/class_template_argument_deduction

brawner:
Trying Windows build again. The farm is a bit backed up, so it might be a little bit for results.

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11521/)

brawner:
Windows passed, looks like we're good.

Thanks for the reviews and help @clalancette!

PR #1205, opened 2020-06-29
Remove usage of domain id in node options

ivanpauno:
This code isn't doing anything:

- https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/src/rcl/node_options.c#L32
- https://github.com/ros2/rcl/blob/b9cfc243bf018091b5c5bcb5585d19591daf4ae6/rcl/src/rcl/node.c#L257-L264

I'm planning to completely remove domain id from `node options`, and use the one in `init options` instead.
That was agreed during the implementation of https://github.com/ros2/rmw/issues/183, but we never finished that API change.

It isn't much of an issue currently, as there's no API in `rclpy` or `rclcpp` to modify the domain id.

### DIFF
--- a/rclcpp/src/rclcpp/node_options.cpp
+++ b/rclcpp/src/rclcpp/node_options.cpp
@@ -92,7 +92,6 @@ NodeOptions::get_rcl_node_options() const
     *node_options_ = rcl_node_get_default_options();
     node_options_->allocator = this->allocator_;
     node_options_->use_global_arguments = this->use_global_arguments_;
-    node_options_->domain_id = this->get_domain_id_from_env();
     node_options_->enable_rosout = this->enable_rosout_;
 
     int c_argc = 0;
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/node_options.cpp
```diff
@@ -322,36 +321,4 @@ NodeOptions::allocator(rcl_allocator_t allocator)
   return *this;
 }
 
-// TODO(wjwwood): reuse rcutils_get_env() to avoid code duplication.
-//   See also: https://github.com/ros2/rcl/issues/119
-size_t
-NodeOptions::get_domain_id_from_env() const
```
fujitatomoya, line 328: we should also delete the declaration? from https://github.com/ros2/rclcpp/blob/a8cd93623964a089ee132d65e8b5eda84c15740d/rclcpp/include/rclcpp/node_options.hpp#L331-L333

ivanpauno: done

ivanpauno:
> @ivanpauno I think we should deprecate get_domain_id_from_env() instead of removing it altogether -- it is part of the public NodeOptions API after all

is it?
The method is protected, and it doesn't make sense to inherit from `NodeOptions` ([slicing ...](https://github.com/ros2/rclcpp/blob/a8cd93623964a089ee132d65e8b5eda84c15740d/rclcpp/include/rclcpp/node.hpp#L1189)).

I understand that we prefer a tick-tock cycle over deleting the API, but in this case I think it will be unnecessary and confusing (as the domain id is not part of node options).

ivanpauno:
> there would be follow-up PR for rcl to delete domain_id from rcl_node_options_t?

I have to create follow ups in `rcl` `rmw` and `rmw_connext`

hidmic:
> The method is protected, and it doesn't make sense to inherit from NodeOptions (slicing ...).

True, and looking closer at it we set ourselves up so that it is impossible to make any use of an override. Oh well, fair enough.

ivanpauno:
> > The method is protected, and it doesn't make sense to inherit from NodeOptions (slicing ...).
> 
> True, and looking closer at it we set ourselves up so that it is impossible to make any use of an override. Oh well, fair enough.

I will take that comment as an approval :heavy_check_mark: :joy:.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11353/)

Issue #1212, opened 2020-07-01
Ready reentrant Waitable objects can attempt to execute multiple times


audrow:
In the `MultiThreadedExecutor`, the `run` function is bound to each running thread; the `run` function builds and checks the wait set for ready executables and seeks to run them.  The `Executor::wait_for_work()` function builds the wait set and then removes all items that do not return `true` for `is_ready()`, which for the `SubscriptionIntraProcess` class looks to see if there is a positive number of entries in the ring buffer by calling `has_data()`. 

The first executable in the wait set is then checked to see if it should be excluded based on its group. At this stage, an executable that belongs to the `Reentrant` callback group type is always returned to be executed, where as executables in the `MutuallyExclusive` callback group may not be, as only one callback from the group can be executed at once.  From there, the executable is popped from the buffer and executed.  

To execute the callback for the `SubscriptionIntraProcess`, the ring buffer dequeues an element.  To dequeue, the ring buffer first checks that it contains a positive number of elements with `has_data()`; if there are zero elements, an error is thrown.

It is possible for the same ring buffer entry to be 'ready' in the `Executor::wait_for_work()` and no be not ready in `Executor::execute_any_executable()`, thus yielding the following error:

```
[ERROR] [1593631031.302392471] [rclcpp]: Calling dequeue on empty intra-process buffer
terminate called after throwing an instance of 'std::runtime_error'
  what():  Calling dequeue on empty intra-process buffer
Aborted (core dumped)
```
<details> <summary>Backtrace</summary>

```
#0  rclcpp::experimental::buffers::RingBufferImplementation<std::unique_ptr<test_rclcpp::msg::UInt32_<std::allocator<void> >, std::default_delete<test_rclcpp::msg::UInt32_<std::allocator<void> > > > >::dequeue (this=0x5555558d0790)
    at /root/ros2_foxy/install/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp:78
[...]

PR #1213, opened 2020-07-02
Make ring buffer thread-safe

### LINKED ISSUE #1007: Race condition in MultiThreadedExecutor with intra_process communication resulting in dequeuing ring buffer before enqueuing
There is a race condition in the `MultiThreadedExecutor` that happens during the `test_multithreahed.multi_consumer_intra_process` test in `test_rclcpp`. The `intra_process_buffer`'s `ring_buffer_implementation` has dequeue() called without any data. It may be that some process is calling dequeue() right after another has checked has_data().

This bug results in a flakey test with test_rclcpp, but it's fairly reliable to reproduce with sufficient repeat-until-fail iterations. 

https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1780/testReport/(root)/projectroot/gtest_multithreaded__rmw_fastrtps_cpp/
https://ci.ros2.org/job/ci_windows/9412/testReport/

audrow:
Fixes #1007, but there is a logical race condition that results in the same error, see #1212.

### DIFF
--- a/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
+++ b/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
@@ -62,7 +62,7 @@ public:
   /// Add a new element to store in the ring buffer
   /**
    * This member function is thread-safe.
-   * 
+   *
    * \param request the element to be stored in the ring buffer
    */
   void enqueue(BufferT request)
@@ -82,7 +82,7 @@ public:
   /// Remove the oldest element from ring buffer
   /**
    * This member function is thread-safe.
-   * 
+   *
    * \return the element that is being removed from the ring buffer
    */
   BufferT dequeue()
@@ -105,7 +105,7 @@ public:
   /// Get the next index value for the ring buffer
   /**
    * This member function is thread-safe.
-   * 
+   *
    * \param val the current index value
    * \return the next index value
    */
@@ -118,7 +118,7 @@ public:
   /// Get if the ring buffer has at least one element stored
   /**
    * This member function is not thread-safe.
-   * 
+   *
    * \return `true` if there is data and `false` otherwise
    */
   inline bool has_data() const
@@ -127,10 +127,10 @@ public:
     return has_data_();
   }
 
-  /// Get if the size of the buffer is equal to its capacity 
+  /// Get if the size of the buffer is equal to its capacity
   /**
    * This member function is thread-safe.
-   * 
+   *
    * \return `true` if the size of the buffer is equal is capacity
    * and `false` otherwise
    */
@@ -146,7 +146,7 @@ private:
   /// Get the next index value for the ring buffer
   /**
    * This member function is not thread-safe.
-   * 
+   *
    * \param val the current index value
    * \return the next index value
    */
@@ -158,7 +158,7 @@ private:
   /// Get if the ring buffer has at least one element stored
   /**
    * This member function is not thread-safe.
-   * 
+   *
    * \return `true` if there is data and `false` otherwise
    */
   inline bool has_data_() const
@@ -166,10 +166,10 @@ private:
     return size_ != 0;
   }
 
-  /// Get if the size of the buffer is equal to its capacity 
+  /// Get if the size of the buffer is equal to its capacity
   /**
    * This member function is not thread-safe.
-   * 
+   *
    * \return `true` if the size of the buffer is equal is capacity
    * and `false` otherwise
    */
--- a/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
+++ b/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
@@ -37,6 +37,10 @@ namespace experimental
 namespace buffers
 {
 
+/// Store elements in a fixed-size, FIFO buffer
+/**
+ * All public member functions are thread-safe.
+ */
 template<typename BufferT>
 class RingBufferImplementation : public BufferImplementationBase<BufferT>
 {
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
```diff
@@ -112,7 +115,7 @@ class RingBufferImplementation : public BufferImplementationBase<BufferT>
   size_t read_index_;
   size_t size_;
 
-  std::mutex mutex_;
+  mutable std::recursive_mutex recursive_mutex_;
```
fujitatomoya, line 118: this has to be std::recursive_mutex instead mutex? as far as i checked, recursive_mutex takes time more than mutex. And i do not see the necessity so far, could you elaborate the case?

rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
```diff
@@ -88,16 +88,19 @@ class RingBufferImplementation : public BufferImplementationBase<BufferT>
 
   inline size_t next(size_t val)
   {
+    std::lock_guard<std::recursive_mutex> lock(recursive_mutex_);
```
fujitatomoya, line 91: I think that methods for `next`, `has_data` and `is_full` can be implemented via reader lock, using std::shared_mutex instead of std::mutex.

audrow: I think you're right.  I'll give this a try.

rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
audrow: I use the recursive mutex because `enqueue()` and `dequeue()` call `next()`, `has_data()`, and `is_full()`, which have locks in them.

rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
audrow: I tried using a `shared_mutex` with a `unique_lock` (also tried `lock_guard`) in the write functions (i.e., `enqueue()` and `dequeue()`) and `shared_lock`s in the read functions (i.e., `next()`, `has_data()`, and `is_full()`).  This doesn't work because these read operations are called in the write functions, causing a deadlock.

Any ideas on how to use a `shared_mutex` on the read functions, considering that the read functions are called in the write functions?

fujitatomoya: basically I think that mutex should be aligned with data but function.

how about the following?

- `enqueue` w/o `next` / `is_full` and use `unique_lock`
- `dequeue` w/o `has_data` / `next` and use `unique_lock`
- `next` with `shared_lock`
- `has_data` with `shared_lock`
- `is_full` with `shared_lock` 

with above,

- it can keep the interfaces compatible.
- it can use reader/writer lock instead of mutex. (also no need for recursive mutex)

I think that we should avoid the complicated code with mutex, but since this is the place for performance, I'd think using shared_mutex would be better.

rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
```diff
@@ -112,7 +131,7 @@ class RingBufferImplementation : public BufferImplementationBase<BufferT>
   size_t read_index_;
   size_t size_;
 
-  std::mutex mutex_;
+  mutable std::shared_timed_mutex mutex_;
```
fujitatomoya, line 134: are we going to use timeout for lock in the future? could you explain why you pick the `shared_timed_mutex` instead of `shared_mutex`?

audrow: Good question.  `rclcpp` defaults to using C++14, and in C++14 `shared_mutex` was renamed to `shared_timed_mutex` since it had timing in it.  `shared_mutex` is available from C++17 onwards. (See [here](https://stackoverflow.com/a/40207246/3335057) or [here](https://stackoverflow.com/a/42302130/3335057).)

audrow:
@fujitatomoya I went back to using regular `mutex`s because it was brought up to me that the functions we are reading from are likely quicker than the logic associated with checking for the `shared_timed_mutex`'s lock.

rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
```diff
@@ -55,55 +59,125 @@ class RingBufferImplementation : public BufferImplementationBase<BufferT>
...
   inline size_t next(size_t val)
   {
-    return (val + 1) % capacity_;
+    std::lock_guard<std::mutex> lock(mutex_);
+    return next_(val);
   }
 
+  /// Get if the ring buffer has at least one element stored
+  /**
+   * This member function is not thread-safe.
```
jacobperron, line 120: Looks like this function *is* thread-safe.

audrow: Good catch!

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11438/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6652/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9380/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11340/)

fujitatomoya:
@audrow 

> likely quicker than the logic associated with checking for the shared_timed_mutex's lock.

interesting, thanks for the information and notification!

PR #1217, opened 2020-07-09
Remove recreation of entities_collector

brawner:
In the implementation of `spin_until_future_complete` in static_single_threaded_executor, it recreated the entities_collector. However, due to the logic of the class, this removed the nodes from being considered for work. It's my understanding of this code that the entities collector should only be created in the constructor.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

PR #1198, opened 2020-06-25
Unittests for allocator_memory_strategy.hpp part 2

brawner:
This adds the remaining tests for allocator_memory_strategy.hpp and is directly dependent on #1197. The new commit is https://github.com/ros2/rclcpp/commit/5a88b09bfe0ad7f068f5d53b7f6de04f69f58a65

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11300/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6546/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9258/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11204/)

PR #1218, opened 2020-07-09
Adding callback groups in executor

peterpena:


brawner:
This might conflict with #1217 and #1220. Also, I've expanded unit test coverage of the executor implementations to about 90%-95% in #1221 and #1222, which might require checking coverage of this PR to ensure it doesn't regress.

It also looks like you're modifying memory strategy files. How would #1189, #1197 and #1198 change due to this PR? I'm still waiting on @wjwwood to review those.

peterpena:
Thank you for letting me know. I will take a look at those PRs and see how much it will change.

peterpena:
#1217 and #1220 change `static_single_threaded_executor` and `exceptions`. This PR does not change either of these files, so it should be fine.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -211,7 +211,8 @@ Node::get_logger() const
 }
 
 rclcpp::CallbackGroup::SharedPtr
-Node::create_callback_group(rclcpp::CallbackGroupType group_type)
+Node::create_callback_group(
```
ralph-lange, line 214: This line break is not required. It was introduced in the demo by the second argument RealTimeClass which is not used here.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -216,7 +215,8 @@ NodeBase::get_shared_rcl_node_handle() const
 }
 
 rclcpp::CallbackGroup::SharedPtr
-NodeBase::create_callback_group(rclcpp::CallbackGroupType group_type)
+NodeBase::create_callback_group(
```
ralph-lange, line 218: This line break is not required. It was introduced in the demo by the second argument RealTimeClass which is not used here.

peterpena:
> Are we confident that callback groups are the right granularity? I'm convinced of it. And I’m sure that 80% of the nodes will neither be used with multiple Executors nor will they create any custom callback groups. However, one might also ask why the new Executor interface does not allow registering even single callbacks. This should be decided explicitly.

I think exposing the user with different granularity is not a bad option, and it is not too hard to implement. We can add `get_associated_with_executor_atomic()` at the callback level as you did for the callback groups and `collect_entities` retrieves the handles that are not associated with the executor (this is the default state). For the callbacks that are associated with executor, they will be added separately in `collect_entities`. 

> The current implementation is not fully backward compatible. If a callback group is created after a node has been added via add_node, this callback group will not be served. We might consider the concept of a default Executor - similarly to the default callback group. That is, we expect that a node is always added to one Executor instance by add_node. This instance is the default Executor for this node. Thereafter, selected callback groups (or even callbacks, cf. Question 1) may be registered with other Executor instances. As long as a callback group is not registered with a different Executor instance, it will be served by the default Executor. This would ensure full backward compatibility and avoid "dangling" callbacks.

I this is a great idea. I did not consider this case and it is true that it will fail with current implementation. 

We can definitely have more discussions 👍

gbiggs:
> We might consider the concept of a default Executor - similarly to the default callback group. That is, we expect that a node is always added to one Executor instance by add_node. This instance is the default Executor for this node. Thereafter, selected callback groups (or even callbacks, cf. Question 1) may be registered with other Executor instances. As long as a callback group is not registered with a different Executor instance, it will be served by the default Executor. This would ensure full backward compatibility and avoid "dangling" callbacks.

I'm stepping into this PR late and haven't had time to read through it in detail, so I may just be missing something, but wouldn't you be _required_ to ensure there is at least one "default" executor to make sure that built-in topics and servives (like parameters) get serviced?

peterpena:
> I'm stepping into this PR late and haven't had time to read through it in detail, so I may just be missing something, but wouldn't you be required to ensure there is at least one "default" executor to make sure that built-in topics and servives (like parameters) get serviced?

Currently, every callback group has an `allowable` state where when created, the `allowable` state is set to `true`. If a node is added to the executor, the executor checks if there are any `allowable` callback groups that have not been assigned. If there is, then the executor will add the callback group right before the memory strategy collects entities. This makes the PR backwards compatible. If the user solely adds callback groups and does not add a node to an executor, then they will be responsible to add the default callback group to the executor. If the user adds a node to the executor and does not add a callback group. All the callback groups will be collected by the executor, even if the callbacks were added after the adding the node. If the user does not want an executor to add callback groups that have not been assigned to an executor from the node, they can switch the 'allowable' state to false, and the user will be responsible to add the callback group manually to an executor.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2020-07-16/15468/1

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,6 +77,50 @@ class Executor
...
+  add_allowable_unassigned_callback_groups();
+
+  /// Add a callback group to an executor.
+  RCLCPP_PUBLIC
+  virtual void
+  add_callback_group(
+    rclcpp::CallbackGroup::SharedPtr group_ptr,
+    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr, bool notify = true);
+
+
```
wjwwood, line 95: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -102,6 +102,14 @@ class CallbackGroup
   const CallbackGroupType &
   type() const;
 
+  RCLCPP_PUBLIC
```
wjwwood, line 105: Doc block here please.

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -102,6 +102,14 @@ class CallbackGroup
...
+  std::atomic_bool &
+  get_associated_with_executor_atomic();
+
+  RCLCPP_PUBLIC
```
wjwwood, line 109: Doc block here too.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,6 +77,50 @@ class Executor
...
+  get_callback_groups();
+
+  /// Remove a callback group from the executor.
+  RCLCPP_PUBLIC
+  virtual void
+  remove_callback_group(
+    rclcpp::CallbackGroup::SharedPtr group_ptr,
+    bool notify = true);
+
+
```
wjwwood, line 123: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
```diff
@@ -103,6 +103,16 @@ class StaticExecutorEntitiesCollector final
   size_t
   get_number_of_ready_guard_conditions() override;
 
+  RCLCPP_PUBLIC
```
wjwwood, line 106: Doc block here.

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
```diff
@@ -103,6 +103,16 @@ class StaticExecutorEntitiesCollector final
...
+  void
+  add_callback_group(
+    rclcpp::CallbackGroup::SharedPtr group_ptr,
+    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr);
+
+  RCLCPP_PUBLIC
```
wjwwood, line 112: Doc block here.

rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
```diff
@@ -78,6 +78,19 @@ class StaticSingleThreadedExecutor : public rclcpp::Executor
   void
   spin() override;
 
+  RCLCPP_PUBLIC
```
wjwwood, line 81: Doc block here.

rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
```diff
@@ -78,6 +78,19 @@ class StaticSingleThreadedExecutor : public rclcpp::Executor
...
+  void
+  add_callback_group(
+    rclcpp::CallbackGroup::SharedPtr group_ptr,
+    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
+    bool notify) override;
+
+  RCLCPP_PUBLIC
```
wjwwood, line 88: Doc block here.

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -136,12 +144,14 @@ class CallbackGroup
...
   // Mutex to protect the subsequent vectors of pointers.
   mutable std::mutex mutex_;
+  std::atomic_bool associated_with_executor_;
   std::vector<rclcpp::SubscriptionBase::WeakPtr> subscription_ptrs_;
   std::vector<rclcpp::TimerBase::WeakPtr> timer_ptrs_;
   std::vector<rclcpp::ServiceBase::WeakPtr> service_ptrs_;
   std::vector<rclcpp::ClientBase::WeakPtr> client_ptrs_;
   std::vector<rclcpp::Waitable::WeakPtr> waitable_ptrs_;
   std::atomic_bool can_be_taken_from_;
+  std::atomic_bool allow_executor_to_add_;
```
ralph-lange, line 154: I propose to consider changing the type to `const bool` as it is set a construction time only. This might make testing this flag slightly faster.

peterpena: I agree. After considering between different options and testing, const bool is what we need because the `allow_executor_to_add_` should not change in the life of the callback group.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -115,72 +116,139 @@ Executor::~Executor()
...
+  std::vector<rclcpp::CallbackGroup::SharedPtr>> node_to_groups,
+  bool notify)
+  {
+    for (auto const& element : node_to_groups) {
+      add_callback_groups(element.second, element.first, notify);
+    }
+  }
+
+void
+Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr, bool notify)
```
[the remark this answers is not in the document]
peterpena: This is now backward compatible. When `add_node` is called. it is stored in a list of nodes that is iterated before calling `collect_entities` to look for callback groups that are in an `allowable` state and have not been assigned to an executor yet.

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11785/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6866/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9624/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11728/)

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11791/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6873/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9628/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11734/)

peterpena:
The failures are not relevant to this PR, so I am running it again. Hopefully, the errors are transient.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,124 @@ class Executor
...
+  RCLCPP_PUBLIC
+  virtual void
+  remove_callback_group(
+    rclcpp::CallbackGroup::SharedPtr group_ptr,
+    bool notify = true);
+
   /// Add a node to the executor.
   /**
    * An executor can have zero or more nodes which provide work during `spin` functions.
+   * If a node is already associated with an executor, the executor throws an exception.
```
wjwwood, line 192:
```suggestion
   * If a node is already associated with an executor, this method throws an exception.
```

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,124 @@ class Executor
...
+   * The node is added to `weak_nodes_` which will be iterated before `collect_entities`
+   * from the memory strategy to find any callback groups that are not associated to
+   * an executor and are in an `allowable` state. All callback groups found in the node
+   * the moment it was added to the executor that are not associated to an executor and
+   * in an 'allowable' state will be added to the executor through `add_callback_group`.
```
wjwwood, line 197: You should mention here what happens to callback groups added to the node later.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -98,6 +214,9 @@ class Executor
 
   /// Remove a node from the executor.
   /**
+   * All the callback groups associated with this node will be removed from the executor
```
wjwwood, line 217: Including callback groups which are part of the node but were created with the second parameter being false?

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -329,6 +448,15 @@ class Executor
   rclcpp::node_interfaces::NodeBaseInterface::SharedPtr
   get_node_by_group(rclcpp::CallbackGroup::SharedPtr group);
 
+  /**
+   * Checks in `weak_groups_to_nodes` whether `node_ptr` has been added to the
```
wjwwood, line 452: Use a brief with `///`, also avoid discussing internals of the class, and instead just describe what should happen.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -58,15 +81,21 @@ StaticSingleThreadedExecutor::add_node(
...
     throw std::runtime_error("Node has already been added to an executor.");
   }
-
   if (notify) {
     // Interrupt waiting to handle new node
     if (rcl_trigger_guard_condition(&interrupt_guard_condition_) != RCL_RET_OK) {
       throw std::runtime_error(rcl_get_error_string().str);
     }
   }
-
```
wjwwood, line 68: These changes to remove whitespace seem unnecessary, I'd recommend avoiding changes like this.

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -150,48 +150,47 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
     );
   }
 
-  bool collect_entities(const WeakNodeList & weak_nodes) override
+  bool collect_entities(const WeakCallbackGroupsToNodesMap & weak_groups_to_nodes)
```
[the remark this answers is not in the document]
peterpena: Yes that was my mistake.

rclcpp/include/rclcpp/executor.hpp
peterpena: It should only remove callback groups that are associated with the executor. I changed the comment to reflect this.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,124 @@ class Executor
...
+   * is not associated to an executor and is in an 'allowable state'. This takes care of
+   * any callback group that has been added to a node but not explicitly added to the
+   * executor such as the default callback group. It is important to node that in order
+   * for the callback groups to be automatically added to an executor through this function,
+   * the node of the callback groups needs to be added through the `add_node` function
+   * provided by the executor.
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  add_allowable_unassigned_callback_groups();
```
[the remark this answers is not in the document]
peterpena: It should be under protected.

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11892/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6933/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9687/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11825/)

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11909/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6952/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9703/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11842/)

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11923/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6969/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9716/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11862/)

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,124 @@ class Executor
...
+  /**
+   * The callback group is removed from the map `weak_groups_to_nodes` and unassociated
+   * with the executor. If the node that is associated with the callback group removed
+   * no longer has any callback groups associated with the executor, the guard
+   * condition is removed from `weak_nodes_to_guard_conditions`.
+   * \param[in] group_ptr Shared pointer to the callback group to be added.
+   * \param[in] notify True to trigger the interrupt guard condition during this function. If
+   * the executor is blocked at the rmw layer while waiting for work and it is notified that a
+   * callback group was removed, it will wake up.
+   */
```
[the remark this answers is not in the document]
peterpena: I added a note about an exception thrown if a callback group is not associated with the executor.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,124 @@ class Executor
...
+   * \param[in] notify True to trigger the interrupt guard condition during this function. If
+   * the executor is blocked at the rmw layer while waiting for work and it is notified that a new
+   * callback group was added, it will wake up.
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  add_callback_groups(
+    std::map<rclcpp::node_interfaces::NodeBaseInterface::SharedPtr,
+    std::vector<rclcpp::CallbackGroup::SharedPtr>> node_to_groups,
+    bool notify = true);
```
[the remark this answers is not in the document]
peterpena: I removed the function.

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11942/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6986/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9733/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11881/)

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11943/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6987/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9734/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11882/)

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,97 @@ class Executor
   virtual void
   spin() = 0;
 
+  /// Add a callback group to an executor.
+  /**
+   * An executor can have zero or more callback groups which provide work during `spin` functions.
+   * When an executor attempts to add a callback group, the executor checks to see if it is already
+   * associated with another executor.
+   * If it is, than an exception is thrown.
```
wjwwood, line 85: Please be specific about what exception type is thrown and please use `\throw` to document it.

wjwwood, line 85:
```suggestion
   * If it is, then an exception is thrown.
```

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,97 @@ class Executor
...
+   *
+   * \param[in] group_ptrs a vector of shared ptrs that point to callback groups
+   * \param[in] node_ptr a shared pointer that points to a node base interface
+   * \param[in] notify True to trigger the interrupt guard condition during this function. If
+   * the executor is blocked at the rmw layer while waiting for work and it is notified that a new
+   * callback group was added, it will wake up.
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  add_callback_groups(
```
wjwwood, line 118: I think we should drop this signature too, since there's no symmetric `remove_callback_groups` which takes a vector and I just don't think it is needed as it is simply a for_each under the hood.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,97 @@ class Executor
...
+   *
+   * \param[in] group_ptr Shared pointer to the callback group to be added.
+   * \param[in] notify True to trigger the interrupt guard condition during this function. If
+   * the executor is blocked at the rmw layer while waiting for work and it is notified that a
+   * callback group was removed, it will wake up.
+   * \throw std::runtime_error if the callback group is not associated with the executor
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  remove_callback_group(
```
wjwwood, line 154: @peterpena and I talked about this offline and I think we should change this so that you may only ever remove a callback group with this method which was explicitly added with `add_callback_group`. It should not be possible to remove callback groups added with `add_node` and it should not be possible to implicitly remove a node that was added with `add_node` by removing its only callback group. These side-effects and interactions make these functions more complicated than they need to be.

This will also affect `remove_node` and `has_node`.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -76,9 +77,97 @@ class Executor
...
+   * Nodes have associated callback groups, and this method adds any of those callback groups
+   * to this executor which have their automatically_add_to_executor_with_node parameter true.
+   * The node is also associated with the executor so that future callback groups which are
+   * created on the node with the automatically_add_to_executor_with_node parameter set to true
+   * are also automatically associated with this executor.
+   *
+   * Callback groups with the automatically_add_to_executor_with_node parameter set to false must
+   * be manually added to an executor using the add_callback_group method.
+   *
+   * If a node is already associated with an executor, this method throws an exception.
```
wjwwood, line 169: Please also use the type and `\throw` here as well.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -343,6 +447,21 @@ class Executor
...
+   * can be automatically added to this executor.
+   * This takes care of any callback group that has been added to a node but not explicitly added
+   * to the executor.
+   * It is important to note that in order for the callback groups to be automatically added to an
+   * executor through this function, the node of the callback groups needs to have been added
+   * through the `add_node` method.
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  add_allowable_unassigned_callback_groups();
```
wjwwood, line 463: Can you rename this method to avoid the `allowable` terminology that we moved away from elsewhere in this pr?

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
```diff
@@ -216,13 +251,52 @@ class StaticExecutorEntitiesCollector final
   rclcpp::Waitable::SharedPtr
   get_waitable(size_t i) {return exec_list_.waitable[i];}
 
+  /// Returns true if the callback group belongs to the collector
+  /** Returns true if the callback group belongs to the collector
+   * \param[in] group_ptr a callback group shared pointer
+   * \return boolean whether a callback group belongs the collector
+   */
```
wjwwood, line 258:
```suggestion
  /// Return true if the callback group belongs to the collector
  /**
   * \param[in] group_ptr a callback group shared pointer
   * \return boolean whether a callback group belongs the collector
   */
```

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
```diff
@@ -216,13 +251,52 @@ class StaticExecutorEntitiesCollector final
...
+  bool
+  has_callback_group(const rclcpp::CallbackGroup::SharedPtr group_ptr) const;
+
+  /// Returns true if the node belongs to the collector
+  /** Returns true if the node belongs to the collector
+   * \param[in] group_ptr a node base interface shared pointer
+   * \return boolean whether a node belongs the collector
+   */
```
wjwwood, line 266:
```suggestion
  /// Return true if the node belongs to the collector
  /**
   * \param[in] group_ptr a node base interface shared pointer
   * \return boolean whether a node belongs the collector
   */
```

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11980/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7012/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9764/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11919/)

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11991/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7020/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9772/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11928/)

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12018/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7042/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9793/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11951/)

peterpena:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12036/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7061/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9811/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11968/)

wjwwood:
Updated CI to check the state:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12185/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7169/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9912/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12068/)

wjwwood:
CI is good now, I'm gonna merge. Thanks again @peterpena.

Btw, do you have the demos/examples you used in your show and tell presentation? I think I can recreate them, but if you have them handy let me know.

wjwwood:
Follow up issue: https://github.com/ros2/rclcpp/issues/1287

----

Also, we should consider this for back port, but give how invasive it is, I don't think that's easily done. Though I think could be possible to back port it in some way (maybe an entirely separate executor class has to be used). @Karsten1987 FYI

peterpena:
@wjwwood Here is the example code that we created if you still need it: https://github.com/ros2/demos/tree/peterpena/cbg-executor-example. Sorry for taking so long. I just settled in.

Karsten1987:
@v-lopez fyi.

The recent changes to the executor should technically allow us for a lock-free execution of the ros2 control rt thread. The precise example here: https://github.com/ros2/demos/commit/667ad10ea63c944d6b630224fe09ea9b77610534

As mentioned, this PR is most likely not going to be backported to Foxy, but we could think about copying the changes to the executor in a ros2 control namespace and use it internally until Galactic.

ralph-lange:
> @wjwwood Here is the example code that we created if you still need it: https://github.com/ros2/demos/tree/peterpena/cbg-executor-example. Sorry for taking so long. I just settled in.

I am willingly ready to port the ping-pong example from https://github.com/boschresearch/ros2_examples/tree/cbg-executor-foxy/rclcpp/cbg-executor_ping-pong to the final (callback-group-level) Executor API of rclcpp and make a pull request for https://github.com/ros2/demos/. Are you interested in this, @wjwwood, @peterpena?

wjwwood:
@ralph-lange if you'd like to do that then that would be ok with me. I think there are more changes coming to that API (like not needing to pass the node pointer when adding a callback group to the executor), but having some examples in the meantime will definitely be good.

brawner:
I think this PR introduced a memory error as exhibited by `TestExecutors.addTemporaryNode`. In `wait_for_work`, it will try to find the guard_condition associated with a weak_ptr to a node. But if that node is expired, then the guard_condition will already have been freed. (https://github.com/ros2/rclcpp/pull/1218/files#diff-e09e1b6262f7c9c86a3fc1a128c0da842729af46311955be86d1c97bf2558fbeR672-R678).

I opened https://github.com/ros2/rclcpp/issues/1402 for tracking

PR #1226, opened 2020-07-13
Add take data to waitable

audrow:


rclcpp/src/rclcpp/executor.cpp
```diff
@@ -321,6 +321,7 @@ Executor::execute_any_executable(AnyExecutable & any_exec)
     execute_client(any_exec.client);
   }
   if (any_exec.waitable) {
+    std::lock_guard<std::mutex> memlock(memory_strategy_mutex_);
```
sloretz, line 324: I don't think the memory strategy is used within this scope. Does this mutex need to be locked here?

audrow:
Closing in favor of #1235.

PR #1220, opened 2020-07-09
Derive and throw exception in spin_some spin_all for StaticSingleThreadedExecutor

brawner:
As a quick-resolution to #1219 this PR proposes to derive and throw a new exception of type "UnimplementedError" in `spin_some`, `spin_all` and `spin_once` for StaticSingleThreadedExecutor. These derived methods currently don't function correctly and so this will help notify users that it is a known issue. I'm happy to take a different approach on this if people have other suggestions.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

Issue #1219, opened 2020-07-09
StaticSingleThreadedExecutor's spin_some, spin_all and spin_once cannot use the base class implementation


brawner:
I've been expanding the unit tests of the executors  for all three classes, and these methods in `StaticSingleThreadedExecutor` should not be using the base class implementation in `rclcpp::Executor`. Because `StaticSingleThreadedExecutor` derives and implements its own `add_node` and `remove_node`, the collection of nodes in `rclcpp::Executor` is not updated and so these spin variants won't do any work. Either `add_node` and `remove_node` need to call the base class implementations as well, which would currently throw an "This node already has an executor" exception or these variants need to be derived specifically for `StaticSingleThreadedExecutor`.

dirk-thomas:
@brawner Can you comment what you think should happen on this ticket: assign it to you, assign it to someone else, enhancement for backlog, ...?

brawner:
It's not exactly an enhancement, after dealing with the backport in #1229, I'll assign it to someone else.

fujitatomoya:
i think we can close this by https://github.com/ros2/rclcpp/pull/1338. CC: @mauropasse

PR #1229, opened 2020-07-14
[foxy backport] Derive and throw exception in spin_some spin_all for StaticSingleThreadedExecutor (#1220)

brawner:
This backports just changes to the headers, not the unit test.

### DIFF
--- a/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
@@ -190,6 +190,38 @@ public:
     return rclcpp::FutureReturnCode::INTERRUPTED;
   }
 
+  /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
+  RCLCPP_PUBLIC
+  void
+  spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0)) override
+  {
+    (void)max_duration;
+    throw rclcpp::exceptions::UnimplementedError(
+            "spin_some is not implemented for StaticSingleThreadedExecutor, use spin or "
+            "spin_until_future_complete");
+  }
+
+  /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
+  RCLCPP_PUBLIC
+  void
+  spin_all(std::chrono::nanoseconds) override
+  {
+    throw rclcpp::exceptions::UnimplementedError(
+            "spin_all is not implemented for StaticSingleThreadedExecutor, use spin or "
+            "spin_until_future_complete");
+  }
+
+  /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
+  RCLCPP_PUBLIC
+  void
+  spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1)) override
+  {
+    (void)timeout;
+    throw rclcpp::exceptions::UnimplementedError(
+            "spin_once is not implemented for StaticSingleThreadedExecutor, use spin or "
+            "spin_until_future_complete");
+  }
+
 protected:
   /// Check which executables in ExecutableList struct are ready from wait_set and execute them.
   /**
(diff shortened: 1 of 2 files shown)


rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
```diff
@@ -190,6 +190,38 @@ class StaticSingleThreadedExecutor : public rclcpp::Executor
...
+    (void)max_duration;
+    throw rclcpp::exceptions::UnimplementedError(
+            "spin_some is not implemented for StaticSingleThreadedExecutor, use spin or "
+            "spin_until_future_complete");
+  }
+
+  /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
+  RCLCPP_PUBLIC
+  void
+  spin_all(std::chrono::nanoseconds) override
```
jacobperron, line 207: `spin_all()` is not part of the `rclcpp::Executor` class in Foxy.

brawner: That's what I get for playing fast and loose and not adequately testing this PR. Removed spin_all.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11452/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6666/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9393/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11353/)

jacobperron:
@brawner I think you need to also provide the [foxy repos file](https://github.com/ros2/ros2/blob/foxy/ros2.repos) for CI.

jacobperron:
I wasn't sure if this change was ABI compatible so I ran an ABI checker and it looks good :+1:

brawner:
Rerunning tests. 

> I wasn't sure if this change was ABI compatible so I ran an ABI checker and it looks good

Thanks! But I think that may because those implementations are header only. If the implementation eventually gets moved to the source file, do we have anything to worry about?

Also, this file wasn't actually being tested anywhere and not linked to. Would the ABI checker recognize it?

jacobperron:
I don't know the answer to either of your questions :sweat_smile: 

Maybe someone from @ros2/team can weigh in?

brawner:
From what I can tell from: https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B under section "Adding a reimplemented virtual function"

This will break ABI compatibility if anyone calls the derived function explicitly:
```
rclcpp::executors::StaticSingleThreadedExecutor executor;
executor.add_node(node);
executor.spin_some();
```

But wouldn't if they called the base class implementation:
```
rclcpp::Executor executor = rclcpp::executors::StaticSingleThreadedExecutor();
executor.add_node(node);
executor.spin_some();
```

Since the compiler will not include the vtable lookup in the first version.

As these functions do not currently work at all, this might be a potentially necessary breaking ABI change that can helps users understand why it broke.

clalancette:
Right, as far as I can tell, this will break ABI, because it will change the offsets into the derived object (as you correctly state above).

I'm on the fence about whether it is worth it to break ABI for this.  On the "yes, let's break ABI" side, Foxy is only a month old, and so the number of users is probably fairly low (and thus the scope of the breakage is low).  Further, there probably aren't very many users of `StaticSingleThreadedExecutor`.  On the "no, let's not break ABI" side, the bar to break ABI is pretty high, and I'm not sure that on its own, this clears it.  The problem doesn't seem severe enough to me to potentially break users (which, if they don't recompile, will break in incomprehensible ways too).

I'd lean towards "no, let's not break ABI, and just document the problem", but I could be persuaded otherwise.  I'd like to hear more opinions.

chapulina:
I often refer to [this guide](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B#The_Do.27s_and_Don.27ts) from KDE when it comes to ABI. They have some tips to workaround [adding a reimplemented virtual function](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B#Adding_a_reimplemented_virtual_function), which could be done for Foxy, then the correct thing can be done for Galactic.

> I'd lean towards "no, let's not break ABI

+1

jacobperron:
I'm also leaning towards not breaking API and documenting the issue. We should add it to the list of [known issues for Foxy](https://index.ros.org/doc/ros2/Releases/Release-Foxy-Fitzroy/#known-issues).

brawner:
For reference, this is the open rclcpp issue. https://github.com/ros2/rclcpp/issues/1219

I'll make a PR for submitting it to known issues.

brawner:
Closing in favor of #1385

PR #1233, opened 2020-07-20
[Dashing] Avoid reference cycle between Context and GraphListener

ivanpauno:
Backport of https://github.com/ros2/rclcpp/pull/906.

This change breaks ABI of the `GraphListener` class.

That class is never used directly by an user.
In the internal places where that class is being used, it's being stored as a `shared_ptr`.
i.e.: ABI of other objects shouldn't be affected, and thus the user shouldn't be affected.

I'm not quite sure if we should backport a fix like this or not as it's hard to verify if it actually breaks "user" ABI or not.
If the answer is no, I will delete this from Dashing and Eloquent patch release board.

### DIFF
--- a/rclcpp/include/rclcpp/graph_listener.hpp
+++ b/rclcpp/include/rclcpp/graph_listener.hpp
@@ -165,7 +165,7 @@ private:
   void
   __shutdown(bool);
 
-  rclcpp::Context::SharedPtr parent_context_;
+  rclcpp::Context::WeakPtr parent_context_;
 
   std::thread listener_thread_;
   bool is_started_;
--- a/rclcpp/src/rclcpp/graph_listener.cpp
+++ b/rclcpp/src/rclcpp/graph_listener.cpp
@@ -80,7 +80,7 @@ GraphListener::start_if_not_started()
       0,  // number_of_clients
       0,  // number_of_services
       0,  // number_of_events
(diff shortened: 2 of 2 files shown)


wjwwood:
I would decline to do this, even with the fair points about the class not being used externally and therefore not a high risk ABI break.

Is someone asking for this fix or we just think it's good to have?

ivanpauno:
> Is someone asking for this fix or we just think it's good to have?

I think nobody is asking for it.
I've added it to the board in the past, and I wanted to double check what we should do in these cases.

I will close and delete it from the Dashing backports board.
If someone specifically asks for this in the future we can reconsider making the backport.

wjwwood:
Ok, yeah perhaps someone on our team would disagree, but I'd prefer to document the issue rather than break ABI, in this case.

Perhaps here, if appropriate: https://index.ros.org/doc/ros2/Releases/Release-Dashing-Diademata/#known-issues

ivanpauno:
> Ok, yeah perhaps someone on our team would disagree, but I'd prefer to document the issue rather than break ABI, in this case.
> 
> Perhaps here, if appropriate: https://index.ros.org/doc/ros2/Releases/Release-Dashing-Diademata/#known-issues

Sounds like a good idea, I've opened https://github.com/ros2/ros2_documentation/pull/826 to document the issue.

PR #1237, opened 2020-07-22
suppress warnings inline

Karsten1987:
This should let `cppcheck` pass with v2.x.

I deliberately opted against the approach taken in https://github.com/ros2/rclpy/pull/577 as it'd increase testing time quite dramatically - it always timeout after 3mins.


rclcpp/src/rclcpp/context.cpp
```diff
@@ -317,7 +317,9 @@ Context::shutdown(const std::string & reason)
       rcl_ret_t rcl_ret = rcl_logging_fini();
       if (RCL_RET_OK != rcl_ret) {
         RCUTILS_SAFE_FWRITE_TO_STDERR(
+          // cppcheck-suppress unknownMacro
```
fujitatomoya, line 320: LGTM. but quick question, why does it intend to print file and line particularly here? can we just print the error message such as `failed to finalize rcl logging` or something like that?

PR #1238, opened 2020-07-23
Warn about unused result of add_on_set_parameters_callback

jacobperron:
If the user doesn't retain a reference to the returned shared pointer there will be zero references and their callback will be unregistered.

I think it would be good to backport this change to Foxy and Eloquent.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -26,6 +26,8 @@
 #include <utility>
 #include <vector>
 
+#include "rcutils/macros.h"
+
 #include "rcl/error_handling.h"
 #include "rcl/node.h"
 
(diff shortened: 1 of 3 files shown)


jacobperron:
> I'm not quite sure of backporting warnings, though in this case it might be ok.

I suggest it because if a user is not retaining the shared pointer result, then their registered callback will not be called as expected. We'd be helping uncover bugs.

jacobperron:
Building everything to check for warnings and testing rclcpp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11606/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6744/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9488/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11568/)

PR #1241, opened 2020-07-27
Add `take_data` to `Waitable` and `data` to `AnyExecutable`

### LINKED ISSUE #1212: Ready reentrant Waitable objects can attempt to execute multiple times

audrow:
Fixes #1212.  This PR extends the `Waitable` class with a `take_data` method and the `AnyExecutable` struct with a `data` field.  `take_data` places the data to be executed in `AnyExecutable.data`, which is then used in the `execute` method.

### DIFF
--- a/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
+++ b/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
@@ -77,7 +77,11 @@ public:
   void
   execute(std::shared_ptr<void> & data) override;
 
-  /// Prepare the waitable for execution.
+  /// Take the data so that it can be consumed with `execute`.
+  /**
+   * For `StaticExecutorEntitiesCollector`, this always return `nullptr`.
+   * \sa rclcpp::Waitable::take_data()
+   */
   RCLCPP_PUBLIC
   std::shared_ptr<void>
   take_data() override;
--- a/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
+++ b/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
@@ -72,9 +72,15 @@ public:
   void
   fini();
 
+  /// Execute the waitable.
   RCLCPP_PUBLIC
   void
-  execute() override;
+  execute(std::shared_ptr<void> & data) override;
+
+  /// Prepare the waitable for execution.
+  RCLCPP_PUBLIC
+  void
+  take_data(std::shared_ptr<void> & data) override;
 
   /// Function to add_handles_to_wait_set and wait for work and
   /**
--- a/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
+++ b/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
@@ -79,8 +79,8 @@ public:
 
   /// Prepare the waitable for execution.
   RCLCPP_PUBLIC
-  void
-  take_data(std::shared_ptr<void> & data) override;
+  std::shared_ptr<void>
+  take_data() override;
 
   /// Function to add_handles_to_wait_set and wait for work and
   /**
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -20,6 +20,7 @@
 #include <functional>
 #include <map>
 #include <memory>
+#include <stdexcept>
 #include <string>
 #include <utility>
 
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -198,7 +198,7 @@ private:
       data);
 
     if (any_callback_.use_take_shared_method()) {
-      ConstMessageSharedPtr shared_msg = (*shared_ptr).first;
+      ConstMessageSharedPtr shared_msg = shared_ptr->first;
       any_callback_.dispatch_intra_process(shared_msg, msg_info);
     } else {
       MessageUniquePtr unique_msg = std::move((*shared_ptr).second);
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -18,6 +18,7 @@
 #include <rmw/rmw.h>
 
 #include <functional>
+#include <map>
 #include <memory>
 #include <string>
 #include <utility>
@@ -115,13 +116,34 @@ public:
   bool
   is_ready(rcl_wait_set_t * wait_set)
   {
-    (void)wait_set;
+    (void) wait_set;
     return buffer_->has_data();
   }
 
-  void execute()
+  void take_data(std::shared_ptr<void> & data)
   {
-    execute_impl<CallbackMessageT>();
+    if (data) {
+      throw std::runtime_error("'data' is not empty");
+    }
+
+    ConstMessageSharedPtr shared_msg;
+    MessageUniquePtr unique_msg;
+
+    if (any_callback_.use_take_shared_method()) {
+      shared_msg = buffer_->consume_shared();
+    } else {
+      unique_msg = buffer_->consume_unique();
+    }
+    data = std::static_pointer_cast<void>(
+      std::make_shared<std::pair<ConstMessageSharedPtr, MessageUniquePtr>>(
+        std::pair<ConstMessageSharedPtr, MessageUniquePtr>(
+          shared_msg, std::move(unique_msg)))
+    );
+  }
+
+  void execute(std::shared_ptr<void> & data)
+  {
+    execute_impl<CallbackMessageT>(data);
   }
 
   void
@@ -154,26 +176,34 @@ private:
 
   template<typename T>
   typename std::enable_if<std::is_same<T, rcl_serialized_message_t>::value, void>::type
-  execute_impl()
+  execute_impl(std::shared_ptr<void> & data)
   {
     throw std::runtime_error("Subscription intra-process can't handle serialized messages");
   }
 
   template<class T>
   typename std::enable_if<!std::is_same<T, rcl_serialized_message_t>::value, void>::type
-  execute_impl()
+  execute_impl(std::shared_ptr<void> & data)
   {
+    if (!data) {
+      throw std::runtime_error("'data' is empty");
+    }
+
     rmw_message_info_t msg_info;
     msg_info.publisher_gid = {0, {0}};
     msg_info.from_intra_process = true;
 
+    auto shared_ptr = std::static_pointer_cast<std::pair<ConstMessageSharedPtr, MessageUniquePtr>>(
+      data);
+
     if (any_callback_.use_take_shared_method()) {
-      ConstMessageSharedPtr msg = buffer_->consume_shared();
-      any_callback_.dispatch_intra_process(msg, msg_info);
+      ConstMessageSharedPtr shared_msg = (*shared_ptr).first;
+      any_callback_.dispatch_intra_process(shared_msg, msg_info);
     } else {
-      MessageUniquePtr msg = buffer_->consume_unique();
-      any_callback_.dispatch_intra_process(std::move(msg), msg_info);
+      MessageUniquePtr unique_msg = std::move((*shared_ptr).second);
+      any_callback_.dispatch_intra_process(std::move(unique_msg), msg_info);
     }
+    shared_ptr.reset();
   }
 
   AnySubscriptionCallback<CallbackMessageT, Alloc> any_callback_;
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -178,6 +178,7 @@ private:
   typename std::enable_if<std::is_same<T, rcl_serialized_message_t>::value, void>::type
   execute_impl(std::shared_ptr<void> & data)
   {
+    (void)data;
     throw std::runtime_error("Subscription intra-process can't handle serialized messages");
   }
 
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -121,12 +121,9 @@ public:
     return buffer_->has_data();
   }
 
-  void take_data(std::shared_ptr<void> & data)
+  std::shared_ptr<void>
+  take_data()
   {
-    if (data) {
-      throw std::runtime_error("'data' is not empty");
-    }
-
     ConstMessageSharedPtr shared_msg;
     MessageUniquePtr unique_msg;
 
@@ -135,7 +132,7 @@ public:
     } else {
       unique_msg = buffer_->consume_unique();
     }
-    data = std::static_pointer_cast<void>(
+    return std::static_pointer_cast<void>(
       std::make_shared<std::pair<ConstMessageSharedPtr, MessageUniquePtr>>(
         std::pair<ConstMessageSharedPtr, MessageUniquePtr>(
           shared_msg, std::move(unique_msg)))
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -201,7 +201,7 @@ private:
       ConstMessageSharedPtr shared_msg = shared_ptr->first;
       any_callback_.dispatch_intra_process(shared_msg, msg_info);
     } else {
-      MessageUniquePtr unique_msg = std::move((*shared_ptr).second);
+      MessageUniquePtr unique_msg = std::move(shared_ptr->second);
       any_callback_.dispatch_intra_process(std::move(unique_msg), msg_info);
     }
     shared_ptr.reset();
--- a/rclcpp/include/rclcpp/qos_event.hpp
+++ b/rclcpp/include/rclcpp/qos_event.hpp
@@ -17,6 +17,7 @@
 
 #include <functional>
 #include <memory>
+#include <stdexcept>
 #include <string>
 
 #include "rcl/error_handling.h"
@@ -137,7 +138,7 @@ public:
   take_data(std::shared_ptr<void> & data) override
   {
     if (data) {
-      throw std::runtime_error("Should not be data in the pointer");
+      throw std::runtime_error("'data' is not empty");
     }
 
     EventCallbackInfoT callback_info;
--- a/rclcpp/include/rclcpp/qos_event.hpp
+++ b/rclcpp/include/rclcpp/qos_event.hpp
@@ -145,9 +145,7 @@ public:
         "Couldn't take event info: %s", rcl_get_error_string().str);
       return nullptr;
     }
-    return std::static_pointer_cast<void>(
-      std::make_shared<EventCallbackInfoT>(
-        callback_info));
+    return std::static_pointer_cast<void>(std::make_shared<EventCallbackInfoT>(callback_info));
   }
 
   /// Execute any entities of the Waitable that are ready.
--- a/rclcpp/include/rclcpp/qos_event.hpp
+++ b/rclcpp/include/rclcpp/qos_event.hpp
@@ -158,7 +158,7 @@ public:
   execute(std::shared_ptr<void> & data) override
   {
     if (!data) {
-      throw std::runtime_error("Data is empty");
+      throw std::runtime_error("'data' is empty");
     }
     auto callback_ptr = std::static_pointer_cast<EventCallbackInfoT>(data);
     event_callback_(*callback_ptr);
--- a/rclcpp/include/rclcpp/qos_event.hpp
+++ b/rclcpp/include/rclcpp/qos_event.hpp
@@ -132,7 +132,7 @@ public:
     }
   }
 
-  /// Take data so that the callback cannot be called again
+  /// Take data so that the callback cannot be scheduled again
   void
   take_data(std::shared_ptr<void> & data) override
   {
--- a/rclcpp/include/rclcpp/qos_event.hpp
+++ b/rclcpp/include/rclcpp/qos_event.hpp
@@ -134,22 +134,18 @@ public:
   }
 
   /// Take data so that the callback cannot be scheduled again
-  void
-  take_data(std::shared_ptr<void> & data) override
+  std::shared_ptr<void>
+  take_data() override
   {
-    if (data) {
-      throw std::runtime_error("'data' is not empty");
-    }
-
     EventCallbackInfoT callback_info;
     rcl_ret_t ret = rcl_take_event(&event_handle_, &callback_info);
     if (ret != RCL_RET_OK) {
       RCUTILS_LOG_ERROR_NAMED(
         "rclcpp",
         "Couldn't take event info: %s", rcl_get_error_string().str);
-      return;
+      return nullptr;
     }
-    data = std::static_pointer_cast<void>(
+    return std::static_pointer_cast<void>(
       std::make_shared<EventCallbackInfoT>(
         callback_info));
   }
--- a/rclcpp/include/rclcpp/qos_event.hpp
+++ b/rclcpp/include/rclcpp/qos_event.hpp
@@ -16,6 +16,7 @@
 #define RCLCPP__QOS_EVENT_HPP_
 
 #include <functional>
+#include <memory>
 #include <string>
 
 #include "rcl/error_handling.h"
@@ -131,12 +132,15 @@ public:
     }
   }
 
-  /// Execute any entities of the Waitable that are ready.
+  /// Take data so that the callback cannot be called again
   void
-  execute() override
+  take_data(std::shared_ptr<void> & data) override
   {
-    EventCallbackInfoT callback_info;
+    if (data) {
+      throw std::runtime_error("Should not be data in the pointer");
+    }
 
+    EventCallbackInfoT callback_info;
     rcl_ret_t ret = rcl_take_event(&event_handle_, &callback_info);
     if (ret != RCL_RET_OK) {
       RCUTILS_LOG_ERROR_NAMED(
@@ -144,8 +148,21 @@ public:
         "Couldn't take event info: %s", rcl_get_error_string().str);
       return;
     }
+    data = std::static_pointer_cast<void>(
+      std::make_shared<EventCallbackInfoT>(
+        callback_info));
+  }
 
-    event_callback_(callback_info);
+  /// Execute any entities of the Waitable that are ready.
+  void
+  execute(std::shared_ptr<void> & data) override
+  {
+    if (!data) {
+      throw std::runtime_error("Data is empty");
+    }
+    auto callback_ptr = std::static_pointer_cast<EventCallbackInfoT>(data);
+    event_callback_(*callback_ptr);
+    callback_ptr.reset();
   }
 
 private:
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -88,10 +88,10 @@ StaticExecutorEntitiesCollector::fini()
   exec_list_.clear();
 }
 
-void
-StaticExecutorEntitiesCollector::take_data(std::shared_ptr<void> & data)
+std::shared_ptr<void>
+StaticExecutorEntitiesCollector::take_data()
 {
-  (void) data;
+  return nullptr;
 }
 
 void
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -77,7 +77,8 @@ StaticExecutorEntitiesCollector::init(
   // Add executor's guard condition
   memory_strategy_->add_guard_condition(executor_guard_condition);
   // Get memory strategy and executable list. Prepare wait_set_
-  execute();
+  std::shared_ptr<void> shared_ptr;
+  execute(shared_ptr);
 }
 
 void
@@ -88,8 +89,15 @@ StaticExecutorEntitiesCollector::fini()
 }
 
 void
-StaticExecutorEntitiesCollector::execute()
+StaticExecutorEntitiesCollector::take_data(std::shared_ptr<void> & data)
 {
+  (void) data;
+}
+
+void
+StaticExecutorEntitiesCollector::execute(std::shared_ptr<void> & data)
+{
+  (void) data;
   // Fill memory strategy with entities coming from weak_nodes_
   fill_memory_strategy();
   // Fill exec_list_ with entities coming from weak_nodes_ (same as memory strategy)
--- a/rclcpp/include/rclcpp/any_executable.hpp
+++ b/rclcpp/include/rclcpp/any_executable.hpp
@@ -47,6 +47,7 @@ struct AnyExecutable
   // These are used to keep the scope on the containing items
   rclcpp::CallbackGroup::SharedPtr callback_group;
   rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base;
+  std::shared_ptr<void> data;
 };
 
 namespace executor
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
@@ -56,8 +56,9 @@ public:
   virtual bool
   is_ready(rcl_wait_set_t * wait_set) = 0;
 
-  virtual void
-  take_data(std::shared_ptr<void> & data) = 0;
+  virtual
+  std::shared_ptr<void>
+  take_data() = 0;
 
   virtual void
   execute(std::shared_ptr<void> & data) = 0;
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
@@ -57,7 +57,10 @@ public:
   is_ready(rcl_wait_set_t * wait_set) = 0;
 
   virtual void
-  execute() = 0;
+  take_data(std::shared_ptr<void> & data) = 0;
+
+  virtual void
+  execute(std::shared_ptr<void> & data) = 0;
 
   virtual bool
   use_take_shared_method() const = 0;
--- a/rclcpp/include/rclcpp/waitable.hpp
+++ b/rclcpp/include/rclcpp/waitable.hpp
@@ -153,14 +153,13 @@ public:
    * // Update the Waitable
    * waitable.update(wait_set);
    * // Execute any entities of the Waitable that may be ready
-   * std::shared_ptr<void> data;
-   * waitable.take_data(data);
+   * std::shared_ptr<void> data = waitable.take_data();
    * ```
    */
   RCLCPP_PUBLIC
   virtual
-  void
-  take_data(std::shared_ptr<void> & data) = 0;
+  std::shared_ptr<void>
+  take_data() = 0;
 
   /// Execute data that is passed in.
   /**
@@ -181,8 +180,7 @@ public:
    * // Update the Waitable
    * waitable.update(wait_set);
    * // Execute any entities of the Waitable that may be ready
-   * std::shared_ptr<void> data;
-   * waitable.take_data(data);
+   * std::shared_ptr<void> data = waitable.take_data();
    * waitable.execute(data);
    * ```
    */
--- a/rclcpp/include/rclcpp/waitable.hpp
+++ b/rclcpp/include/rclcpp/waitable.hpp
@@ -16,6 +16,7 @@
 #define RCLCPP__WAITABLE_HPP_
 
 #include <atomic>
+#include <memory>
 
 #include "rclcpp/macros.hpp"
 #include "rclcpp/visibility_control.hpp"
@@ -125,8 +126,17 @@ public:
   bool
   is_ready(rcl_wait_set_t * wait_set) = 0;
 
-  /// Execute any entities of the Waitable that are ready.
+  /// Take the data so that it can be consumed with `execute`.
   /**
+   * NOTE: take_data is a partial fix to a larger design issue with the
+   * multithreaded executor. This method is likely to be removed when
+   * a more permanent fix is implemented. A longterm fix is currently
+   * being discussed here: https://github.com/ros2/rclcpp/pull/1276
+   *
+   * This method takes the data from the underlying data structure and
+   * writes it to the void shared pointer `data` that is passed into the
+   * method. The `data` can then be executed with the `execute` method.
+   *
    * Before calling this method, the Waitable should be added to a wait set,
    * waited on, and then updated.
    *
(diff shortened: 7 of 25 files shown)


audrow:
To check if this PR breaks upstream packages:  

[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6789/)

ivanpauno:
I'm not convinced that adding a `take_data` method is the solution to the problem, it seems complicated and it doesn't solve the issue for all cases. e.g. we had a similar issue with timers, and an strange way to solve the problem was introduced (see [1](https://github.com/ros2/rclcpp/blob/0e7fed993dcfa0eec28f094fb8518fc0877fb03d/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L86-L98) and [2](https://github.com/ros2/rclcpp/blob/0e7fed993dcfa0eec28f094fb8518fc0877fb03d/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L105-L110)).
In the case of timers, the `take_data` approach won't do.

The issue is that we're waiting for the same waitable to be ready again, before executing it.
Maybe, a thread shouldn't be waiting for a waitable that another thread has already plan to execute.
If the multithreaded executor worked that way, we wouldn't need to introduce this artificial mechanisms to solve the issue.

e.g.: the `rclpy` executor [works this way](https://github.com/ros2/rclpy/blob/1bdd3cea03c5d24a227ac3de7e96b8b8ea5da8e3/rclpy/rclpy/executors.py#L391), and doesn't have any of this problem.

I also think that the `rclpy` threading model is much better than the one used in `rclcpp`.
In the `rclpy` executor there's only one thread collecting entities, and the others doing work.
In `rclcpp`, all threads collect entities and do work. As two threads can't wait on the same entities at the same time, this means that the threads can be blocked for a long time until being able to wait for work again.

I would rather revisit the `MultiThreadedExecutor` design than merging yet another hack to solve its problems.

audrow:
> I would rather revisit the `MultiThreadedExecutor` design than merging yet another hack to solve its problems.

It would probably be good to open a discussion about redesigning `MultiThreadedExecutor`, but I see this PR as being orthogonal to that conversation. This PR fixes one known flaky test and possibly others in our current implementation. Again, a redesign might be a good idea, but it won't help for some time with our existing flaky tests.

ivanpauno:
> It would probably be good to open a discussion about redesigning MultiThreadedExecutor, but I see this PR as being orthogonal to that conversation.

Yeah, my comment was a bit confusing.

> This PR fixes one known flaky test and possibly others in our current implementation. Again, a redesign might be a good idea, but it won't help for some time with our existing flaky tests.

The problem is that this PR isn't only fixing the issue, it's modifying `Waitable` and `AnyExecutable` API to work around it.
Both API changes seem conceptually wrong to me.

Instead of that, `MultiThreadedExecutor` should internally keep track of all the entities that are already being used by another thread (by used I mean either being ready to be executed or being added to a wait set).
Doing that fix would require extending `MemoryStrategy` API, which is where we are currently keeping track of the entities.

ivanpauno:
> e.g. we had a similar issue with timers, and an strange way to solve the problem was introduced (see [1](https://github.com/ros2/rclcpp/blob/0e7fed993dcfa0eec28f094fb8518fc0877fb03d/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L86-L98) and [2](https://github.com/ros2/rclcpp/blob/0e7fed993dcfa0eec28f094fb8518fc0877fb03d/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L105-L110))

The current hack that avoids a timer to be fired "twice", actually doesn't avoid the problem.
Here a recent example of a timer being fired twice: https://ci.ros2.org/view/nightly/job/nightly_osx_release/1712/testReport/projectroot/test/test_multi_threaded_executor/.

The issue happens in the following order:
- Threads A adds the timer to the wait set.
- The timer is ready to be executed by thread A.
- Thread B adds the timer to the wait set, and it's waiting it to be ready.
- Thread B sees that the timer is ready to be executed ([get_next_executable](https://github.com/ros2/rclcpp/blob/0e7fed993dcfa0eec28f094fb8518fc0877fb03d/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L83)).
- Thread A executes the timer and removes it from the scheduled timer collection.
- Thread B checks if the timer has been scheduled ... but thread A has already removed it from the collection. So thread B executes the timer again.

ivanpauno:
@wjwwood I'm considering adding a known issue to all past releases unrecommending the usage of `MultiThreadedExecutor`.
I don't think we will be able to fix this in a backportable way.

The `MultiThreadedExecutor` isn't working well with anything that it's not a subscription/service/service_client.
That is, it's not working well with timers and waitables (if I'm not forgetting something else).

The reason why it's not an issue with subscriptions/etc, is because taking them twice isn't a problem (see [1](https://github.com/ros2/rclcpp/blob/bbf2c4cbfb006245c784f39ee8f8f9e56431c91e/rclcpp/src/rclcpp/executor.cpp#L345) and [2](https://github.com/ros2/rclcpp/blob/bbf2c4cbfb006245c784f39ee8f8f9e56431c91e/rclcpp/src/rclcpp/executor.cpp#L354)).
But the `MultiThreadedExecutor` may currently try to execute anything more than once, and that's a big issue.

clalancette:
> @wjwwood I'm considering adding a known issue to all past releases unrecommending the usage of `MultiThreadedExecutor`.
> I don't think we will be able to fix this in a backportable way.

The problem is that you must use the `MultiThreadedExecutor` to accomplish certain things.  For instance, say you have a node controlling motors, but you want a high-priority ESTOP to be able to come into that node and override the commands currently queued (this is the case in some of the customers that I've written code for).  As far as I know, the only way to accomplish this is with separate callback groups for the subscriptions and the `MultiThreadedExecutor`.

I guess we could recommend that users that need that functionality write their own executor, but it seems like a step back to me.

Maybe another solution here is to introduce `MultiThreadedExecutor2` that is redesigned and has this stuff fixed?

ivanpauno:
> Maybe another solution here is to introduce MultiThreadedExecutor2 that is redesigned and has this stuff fixed?

Sure, that's possible.
What I'm saying is that I don't see a way of backporting a fix for the current class.

Maybe we can fix the issue with some nasty usage of global variables in a backportable fashion, but it isn't going to be simple.

wjwwood:
I think we should have a meeting about this topic with @audrow, @sloretz, me, @ivanpauno, and any one else interested because @audrow, @sloretz, and I already spoke about this and discussed the issues you brought up. Maybe we overlooked something or maybe this pr missed something we discussed before, but this back and forth on the issue isn't ideal, since we're repeating things we already discussed and they're complex enough that you have to catch up on them each time more than a few days passes.

As a primer for the meeting, I think the problem @ivanpauno is talking about existed before and exists after this pr, but is not the problem this pr is trying to avoid. But I could be wrong about that assessment.

ivanpauno:
> I think we should have a meeting about this topic with @audrow, @sloretz, me, @ivanpauno, and any one else interested because @audrow

Sure

> As a primer for the meeting, I think the problem @ivanpauno is talking about existed before and exists after this pr, but is not the problem this pr is trying to avoid. But I could be wrong about that assessment.

According to the PR description, this is trying to fix #1212:
```
Ready reentrant Waitable objects can attempt to execute multiple times
```
It's trying to do so, by taking advantage that [get_next executable](https://github.com/ros2/rclcpp/blob/bbf2c4cbfb006245c784f39ee8f8f9e56431c91e/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L83) is done in a [mutually exclusive fashion](https://github.com/ros2/rclcpp/blob/bbf2c4cbfb006245c784f39ee8f8f9e56431c91e/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L79) and assuming that "taking the data" from the waitable early will make it to no longer be ready.

I have two comments about the approach:
- The assumption that "taking data" from the Waitable will make it to no longer be ready is wrong. A timer is a great counterexample.
- The concept of "taking data" from a waitable is wrong. Why a waitable should take data? (e.g. a timer)
  This is also obvious from the proposed API that returns/takes a `std::shared_ptr<void>`, because what is being "taken" is unknown. I really consider the proposed `take_data` and `execute` signatures to be completely unacceptable, returning/taking a `std::shared_ptr<void>` in an user facing API is really poor c++.

The problem is not `Waitable` API, but that the `MultiThreadedExecutor` is missing some bookkeeping to avoid ready entities to be fired twice.

sloretz:
> e.g.: the `rclpy` executor [works this way](https://github.com/ros2/rclpy/blob/1bdd3cea03c5d24a227ac3de7e96b8b8ea5da8e3/rclpy/rclpy/executors.py#L391), and doesn't have any of this problem.

@ivanpauno I think you might be talking about a different issue. The issue this solves isn't avoided in `rclpy` using the entity `_executor_event`, it's solved by separating [taking data from something about to be executed](https://github.com/ros2/rclpy/blob/1bdd3cea03c5d24a227ac3de7e96b8b8ea5da8e3/rclpy/rclpy/executors.py#L400), and the [execution of the event with that taken data](https://github.com/ros2/rclpy/blob/1bdd3cea03c5d24a227ac3de7e96b8b8ea5da8e3/rclpy/rclpy/executors.py#L407). This separation is why rclpy's `Waitable` [has a take_data method](https://github.com/ros2/rclpy/blob/1bdd3cea03c5d24a227ac3de7e96b8b8ea5da8e3/rclpy/rclpy/waitable.py#L78-L80).

Say `rclcpp` had a flag on every entity like `_executor_event`. When should the executor reset that flag? After the event is executed? Then all entities are mutually exclusive with themselves, and can't be run in parallel. After the execute method starts executing? Then there's a race between the start of execution and when the Waitable takes it's own data for that event.

> Maybe, a thread shouldn't be waiting for a waitable that another thread has already plan to execute.
If the multithreaded executor worked that way, we wouldn't need to introduce this artificial mechanisms to solve the issue.

 I don't think the Multithreaded Executor's design is relevant to this problem. When does a `Waitable` become ready to check if it can be executed again? The answer is it becomes ready after the data for the last event has been taken from it; that way it can be sure the presence of data again means there's another event to execute.

audrow:
In meeting with @sloretz, @wjwwood, and @ivanpauno, we came up with three ways to proceed:
1. Ignore the case where reentrant waitable object is attempted to be executed twice similar to what is done with a timer now. This has a downside of spurious wake ups, which can be expensive @wjwwood pointed out. 
2. Continue with this approach, but add a disclaimer comment in `AnyExecutable` that the data field fixes a current bug and will likely be changed as we develop a more general solution.
3. Similar to option 2, except split `take_data` into two methods: one to take the data and store it locally and another to get that data in an object that doesn't modify `AnyExecutable`. That object would then be passed into execute.

I will plan to do option 2 and perhaps a future PR can be opened for option 3. I will also create a PR that adds an `.rst` file in `rclcpp/docs` to continue discussing a more general solution.

ivanpauno:
Thanks for the summary @audrow. I will add some extra comments on it:

> Ignore the case where reentrant waitable object is attempted to be executed twice similar to what is done with a timer now. This has a downside of spurious wake ups, which can be expensive @wjwwood pointed out.

I wouldn't say "similar to what we do in timer" now, as that has a race condition and doesn't work (see [here](https://github.com/ros2/rclcpp/pull/1241#issuecomment-670530455)).
I would rather say that the `execute` method of all waitables/timers/etc should handle the case of being spuriously executed.
That fix is just a couple of lines of codes, so I will open a PR for clarity.

> Continue with this approach, but add a disclaimer comment in AnyExecutable that the data field fixes a current bug and will likely be changed as we develop a more general solution.

I would like to mention that this proposed method doesn't solve the problem in all cases, but it particularly solve the problem for waitables where the concept of "taking data" makes sense.

> Similar to option 2, except split take_data into two methods: one to take the data and store it locally and another to get that data in an object that doesn't modify AnyExecutable. That object would then be passed into execute.

Rather than splitting it, my idea was having a method `get_execution_callback` that will return a callable object that owns the taken data (in the case that taking any data is needed).
In that way, the API doesn't have to use `void *`, which I think it isn't nice for interfaces that the user is expected to extend.
This has the same problem than option 2: it sometimes fix the problem, but it sometimes doesn't.

I highly prefer going forward with something like 1. or 3.
1. is surprisingly simple and it doesn't involve changing API (the solution isn't ideal because of the reason @wjwwood pointed out).
3. is fine, but I'm not completely sure that's the right API long term, and still doesn't solve the issue completely (we would need to correct the timers issue separately)

> I will also create a PR that adds an .rst file in rclcpp/docs to continue discussing a more general solution.

I think that's great! The topic isn't simple, so discussing and weighing the different alternatives is a great idea.

ivanpauno:
> That fix is just a couple of lines of codes, so I will open a PR for clarity.

https://github.com/ros2/rclcpp/pull/1275 is a draft PR taking this approach.
It fixes the problem for SubscriptionIntraProcess and for Timer, but the fixes are kind of custom.

The general idea is to avoid assuming in `execute` that there's something ready to be executed, which is maybe not super nice.

audrow:
I've added a comment to `Waitable's` `take_data` method that says that this method is not a permanent fix to the problem we've discussed and linked to #1276.

https://github.com/ros2/rclcpp/blob/b92110af9b1f5d462d2de7ae29ee346db1d43268/rclcpp/include/rclcpp/waitable.hpp#L131-L134

audrow:
@wjwwood, do you think that we should go with #1275 because of its simplicity? I like how it simplifies how `MultiThreadedExecutor::run` handles timers. However, as @ivanpauno mentions [here](https://github.com/ros2/rclcpp/pull/1275#issuecomment-676623661), we'll need other PRs to extend other action clients and servers.

wjwwood:
Sorry to leave this hanging for so long.

I actually do not like the patch in https://github.com/ros2/rclcpp/pull/1275, despite its simplicity.

We discussed that in the past and I think I was not in favor:

> Ignore the case where reentrant waitable object is attempted to be executed twice similar to what is done with a timer now. This has a downside of spurious wake ups, which can be expensive @wjwwood pointed out.

I think it addresses a symptom but is not a solution. If we later have different causes for that symptom we might miss it and be erroneously suppressing that "spurious event".

ivanpauno:
> I think it addresses a symptom but is not a solution. If we later have different causes for that symptom we might miss it and be erroneously suppressing that "spurious event".

Yeah, I agree with that.
To be fair, there are other places where we are ignoring spurious events https://github.com/ros2/rclcpp/blob/58bd8d6c210ef5094ddc23d7177c6c10bf6cf4ed/rclcpp/src/rclcpp/executor.cpp#L541-L555.

ivanpauno:
I don't really like the proposed API, but I guess there's not much people actually extending `Waitable` and this will fix a lot of outstanding issues related to `MultiThreadedExecutor` bugs.

I took the liberty of rebasing as there were a lot of conflicts.

rclcpp/include/rclcpp/qos_event.hpp
```diff
@@ -131,21 +132,37 @@ class QOSEventHandler : public QOSEventHandlerBase
...
+        callback_info));
+  }
 
-    event_callback_(callback_info);
+  /// Execute any entities of the Waitable that are ready.
+  void
+  execute(std::shared_ptr<void> & data) override
+  {
+    if (!data) {
+      throw std::runtime_error("Data is empty");
```
ivanpauno, line 161: this wasn't raising an exception before (see https://github.com/ros2/rclcpp/pull/1241/commits/70a55c505254651e92bacac76f19f8acc71cd6bf), but I guess the change is ok.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -115,13 +116,34 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+    ConstMessageSharedPtr shared_msg;
+    MessageUniquePtr unique_msg;
+
+    if (any_callback_.use_take_shared_method()) {
+      shared_msg = buffer_->consume_shared();
+    } else {
+      unique_msg = buffer_->consume_unique();
+    }
+    data = std::static_pointer_cast<void>(
+      std::make_shared<std::pair<ConstMessageSharedPtr, MessageUniquePtr>>(
```
ivanpauno, line 138: this kinds of usage of tuples is a bit weird, but ok

ivanpauno (approved):
looks ok with an extra approver

rclcpp/include/rclcpp/qos_event.hpp
ivanpauno: and https://github.com/ros2/rclcpp/pull/1241/commits/7ff69fad758a68f146d065dadbba9969cb4c9bb9

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12970/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7920/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10691/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12921/)

ivanpauno:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10697/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12948/)

ivanpauno:
Rebased to fix `Waitable::execute` call in recently introduced benchmark.

fujitatomoya (approved):
looks good to me, besides this fixes https://github.com/ros2/rclcpp/issues/1399 (confirmed)

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -115,13 +116,34 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+    (void) wait_set;
     return buffer_->has_data();
   }
 
-  void execute()
+  void take_data(std::shared_ptr<void> & data)
   {
-    execute_impl<CallbackMessageT>();
+    if (data) {
+      throw std::runtime_error("'data' is not empty");
```
clalancette, line 126: Nit: `#include <stdexcept>` for `std::runtime_error`.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -115,13 +116,34 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
   bool
   is_ready(rcl_wait_set_t * wait_set)
   {
-    (void)wait_set;
+    (void) wait_set;
     return buffer_->has_data();
   }
 
-  void execute()
+  void take_data(std::shared_ptr<void> & data)
```
clalancette, line 123: I'm wondering if this signature should just be:

```
std::shared_ptr<void> take_data()
```

It makes a little more sense to me.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -115,13 +116,34 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+      unique_msg = buffer_->consume_unique();
+    }
+    data = std::static_pointer_cast<void>(
+      std::make_shared<std::pair<ConstMessageSharedPtr, MessageUniquePtr>>(
+        std::pair<ConstMessageSharedPtr, MessageUniquePtr>(
+          shared_msg, std::move(unique_msg)))
+    );
+  }
+
+  void execute(std::shared_ptr<void> & data)
```
clalancette, line 144: Given that we are changing public API, do we think we should maintain a deprecated version of this API without the data?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -154,26 +176,35 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
     msg_info.publisher_gid = {0, {0}};
     msg_info.from_intra_process = true;
 
+    auto shared_ptr = std::static_pointer_cast<std::pair<ConstMessageSharedPtr, MessageUniquePtr>>(
+      data);
+
     if (any_callback_.use_take_shared_method()) {
-      ConstMessageSharedPtr msg = buffer_->consume_shared();
-      any_callback_.dispatch_intra_process(msg, msg_info);
+      ConstMessageSharedPtr shared_msg = (*shared_ptr).first;
```
clalancette, line 201: Isn't this equivalent to:

```suggestion
      ConstMessageSharedPtr shared_msg = shared_ptr->first;
```

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -154,26 +176,35 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
...
+      any_callback_.dispatch_intra_process(shared_msg, msg_info);
     } else {
-      MessageUniquePtr msg = buffer_->consume_unique();
-      any_callback_.dispatch_intra_process(std::move(msg), msg_info);
+      MessageUniquePtr unique_msg = std::move((*shared_ptr).second);
```
clalancette, line 204: Same comment here; I think this can be:

```suggestion
      MessageUniquePtr unique_msg = std::move(shared_ptr->second);
```

rclcpp/include/rclcpp/qos_event.hpp
```diff
@@ -131,21 +132,37 @@ class QOSEventHandler : public QOSEventHandlerBase
...
 
-  /// Execute any entities of the Waitable that are ready.
+  /// Take data so that the callback cannot be called again
   void
-  execute() override
+  take_data(std::shared_ptr<void> & data) override
   {
-    EventCallbackInfoT callback_info;
+    if (data) {
+      throw std::runtime_error("Should not be data in the pointer");
```
clalancette, line 140:
```suggestion
      throw std::runtime_error("'data' is not empty");
```

(also, `#include <stdexcept>` at the top for `std::runtime_error`)

rclcpp/include/rclcpp/qos_event.hpp
clalancette, line 161:
```suggestion
      throw std::runtime_error("'data' is empty");
```

clalancette (commented):
I've left a few comments here for improvement.

One other thing that I think bears looking at is the test code coverage with this commit.  A lot of the statements were being tested already, but it would be worthwhile to look at the coverage and make sure we get everything in here covered.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
ivanpauno: Yeah, that sounds like the correct signature to me.
@audrow do you think that change would be ok?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
ivanpauno: The issue is that we have to still call only one from the executor.
Maybe we can always call `execute(data)` from the executor, and provide the following defaults:

```
class Waitable {
  virtual std::shared_ptr<void> take_data() {return nullptr};
  virtual execute(std::shared_ptr<void> data) {(void)data; this->execute()};
  virtual execute() = 0;
}
```

So you can choose to either override only `execute()`, or `execute(data)` and `take_data()`.
We could even make use of that in the places were `take_data` returns `nullptr`.

@audrow @clalancette does that sound good to you?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1241/commits/5b39f357bbe0096af397b0b782542452153ac517.

rclcpp/include/rclcpp/qos_event.hpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1241/commits/5b39f357bbe0096af397b0b782542452153ac517.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1241/commits/bc0849daa4b326cf7a5f46c575452c775d408782

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
ivanpauno: I realized that I cannot really use:
```cpp
  virtual execute() = 0;
```

but rather need:
```cpp
  virtual execute() {throw std::runtime_error{"either execute() or execute(std::shared_ptr<void> &) need to be overriden"};};
```

Because you cannot have a pure virtual method.

I think I would rather break API and add a comment in the release notes than do the above.
I don't expect much people creating custom waitables out of `rclcpp`.

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
```diff
@@ -87,9 +88,16 @@ StaticExecutorEntitiesCollector::fini()
   exec_list_.clear();
 }
 
+std::shared_ptr<void>
+StaticExecutorEntitiesCollector::take_data()
+{
+  return nullptr;
+}
+
```
ivanpauno, line 96: The `StaticExecutorEntitiesCollector` would have some synchronization issues if we would try to use it in a `MultiThreadedExecutor`, but we're not so this is ok for the moment.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13039/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7987/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10763/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13019/)

ivanpauno:
New CI, after https://github.com/ros2/system_tests/pull/444/commits/53b75b86afd3cf8304bb5d6c38ee21e8a926d6b8:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13049/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7996/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10772/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13030/)

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
clalancette: All right, that's OK with me.  I will suggest a release note for Galactic, then, just in case people are creating custom waitables.

clalancette (approved):
I'm tentatively approving this.

The code looks reasonable to me, but there is a lot of change here and I'm not really that familiar with this part of the code.  That's why I'm a bit hesitant.  It would be awesome if someone else more familiar in this area could take a look, but let's not hold up on that.

CI is green on all platforms, which is great.

The one last CI thing we may want to run is this piece of code: https://github.com/ros2/system_tests/blob/c1ccb7281fe387ed5e92c860e4c66f20eb151601/test_rclcpp/test/test_multithreaded.cpp#L338 .  That is, maybe run CI with `--retest-until-fail 100 --packages-select test_rclcpp`.  That will take a long time, but will hopefully convince us that this particular issue is fixed.

ivanpauno:
> The one last CI thing we may want to run is this piece of code: https://github.com/ros2/system_tests/blob/c1ccb7281fe387ed5e92c860e4c66f20eb151601/test_rclcpp/test/test_multithreaded.cpp#L338 . That is, maybe run CI with --retest-until-fail 100 --packages-select test_rclcpp. That will take a long time, but will hopefully convince us that this particular issue is fixed.

I went with

```
--retest-until-fail 200 --packages-select test_rclcpp --ctest-args -R gtest_multithreaded
```

which will run `test_multithreaded.cpp` 200 times or until it fails.
That should not take as long as running all `test_rclcpp` tests 100 times, and is the only test that matters for this PR.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13064/)

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
```diff
@@ -72,9 +72,15 @@ class StaticExecutorEntitiesCollector final
   void
   fini();
 
+  /// Execute the waitable.
   RCLCPP_PUBLIC
   void
-  execute() override;
+  execute(std::shared_ptr<void> & data) override;
+
+  /// Prepare the waitable for execution.
```
wjwwood, line 80: I think the documentation here could be a bit better. Prepare it how?

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
wjwwood: 👍

rclcpp/include/rclcpp/qos_event.hpp
```diff
@@ -131,21 +133,33 @@ class QOSEventHandler : public QOSEventHandlerBase
...
     if (ret != RCL_RET_OK) {
       RCUTILS_LOG_ERROR_NAMED(
         "rclcpp",
         "Couldn't take event info: %s", rcl_get_error_string().str);
-      return;
+      return nullptr;
     }
+    return std::static_pointer_cast<void>(
+      std::make_shared<EventCallbackInfoT>(
+        callback_info));
```
wjwwood, line 150: nitpick: unnecessary new line here

wjwwood (approved):

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
ivanpauno: Fixed in https://github.com/ros2/rclcpp/pull/1241/commits/4d21b6eb6404b32b687303915c9c5b36e73812e9.

rclcpp/include/rclcpp/qos_event.hpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1241/commits/4d21b6eb6404b32b687303915c9c5b36e73812e9.

ivanpauno:
Going in, thanks @audrow for working on this.

ivanpauno:
Adding comment in the release notes in https://github.com/ros2/ros2_documentation/pull/906.

audrow:
@ivanpauno Thanks for pushing this forward!

msmcconnell:
@clalancette anyway this can be backported to foxy? My team has started to see issues from #1212.

clalancette:
> @clalancette anyway this can be backported to foxy? My team has started to see issues from #1212.

Definitely not as it is; it breaks both API and ABI.  If you can figure out a way to backport it without breaking those, we can consider it.

msmcconnell:
Ok, I figured it was unlikely. Thanks for confirming.

ros-discourse:

https://discourse.ros.org/t/ros-2-client-library-working-group-update/37307/10

PR #1202, opened 2020-06-26
Unit tests for node interfaces

brawner:
Signed-off-by: Stephen Brawner <brawner@gmail.com>

PR #1246, opened 2020-07-28
fix failing test with Connext since it doesn't wait for discovery

dirk-thomas:
The failing test was introduced by #1202.

Failing tests:

* Connext: http://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/29/testReport/rclcpp/TestNodeGraph/get_service_names_and_types_by_node/

### DIFF
--- a/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
+++ b/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
@@ -95,6 +95,9 @@ TEST_F(TestNodeGraph, get_service_names_and_types_by_node)
   EXPECT_THROW(
     node_graph->get_service_names_and_types_by_node("not_a_node", "not_absolute_namespace"),
     std::runtime_error);
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+
   auto service_names_and_types1 = node_graph->get_service_names_and_types_by_node("node1", "/ns");
   auto service_names_and_types2 = node_graph->get_service_names_and_types_by_node("node2", "/ns");
   EXPECT_EQ(service_names_and_types1.size(), service_names_and_types2.size());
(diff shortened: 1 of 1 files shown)


rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -95,6 +95,9 @@ TEST_F(TestNodeGraph, get_service_names_and_types_by_node)
   EXPECT_THROW(
     node_graph->get_service_names_and_types_by_node("not_a_node", "not_absolute_namespace"),
     std::runtime_error);
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```
brawner, line 99: The main reason for this test was just to add coverage for the bad arguments above, but I still wanted to add a nominal case. Is there perhaps a better check I could be doing here instead of just checking that the two are the same and can avoid waiting for a full second to establish the node graph discovery?

dirk-thomas: The test needs to wait until node 1 has discovered node 2. It could do so iteratively to finish as early as possible with a maximum timeout. Please feel free to commit directly to this branch to update the test.

brawner: Will do

brawner:
Pushed an update to the unit test to check that an added service exists for one node and not the other.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11676/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6791/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9542/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11643/)

dirk-thomas:
@brawner Please run CI build which check that the test passes with Connext which was the original problem.

brawner:
Rerunning with only connext static checked (also --retest-until-fail 2)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11698/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6798/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9553/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11651/)

PR #1249, opened 2020-07-31
Make parameter clients use template constructors

### LINKED ISSUE #855: Parameters in LifecycleNodes
Hi all,

I am trying to add parameters to a node following the demo at https://github.com/ros2/demos/blob/master/demo_nodes_cpp/src/parameters/set_and_get_parameters.cpp

Problem is, the node is based on the Lyfecycle interface and I suspect rclcpp::SyncParametersClient does not accept them yet. I have seen that this problem appears in multiple other places as the NodeInterface is not yet fully integrated (see 
 https://github.com/ros-perception/image_common/issues/108 for example)

For reference, the code I use is:

```
this->parameters_client_ = std::make_shared<rclcpp::SyncParametersClient>(this);
```

where ```this``` extends from ```rclcpp_lifecycle::LifecycleNode```. I have also tried with ```std::make_shared<rclcpp::SyncParametersClient>(this->shared_from_this());``` in vain. The error is listed as follow:

```
/opt/ros/eloquent/include/rclcpp/parameter_client.hpp:201:12: note:   no known conversion for argument 1 from ‘std::shared_ptr<rclcpp_lifecycle::LifecycleNode>’ to ‘rclcpp::Node::SharedPtr {aka std::shared_ptr<rclcpp::Node>}’
```

which makes me suggest it is the same problem as https://github.com/ros-perception/image_common/issues/108

Thanks in advance for your support.

**EDIT**

After looking closely at the constructors for ```rclcpp::SyncParametersClient``` it seems that the ```rclcpp::node_interfaces::NodeBaseInterface::SharedPtr``` is partially integrated, i.e. only when passing a ```rclcpp::executor::Executor::SharedPtr``` along with it. This means we should be able to initialize the parameters_client outside of the node class.

Could it be possible to extend the Constructor 

```
rclcpp::SyncParametersClient::SyncParametersClient(rclcpp::Node::SharedPtr, const string&, const rmw_qos_profile_t&)
```
to accept a ```rclcpp::node_interfaces::NodeBaseInterface::SharedPtr``` and not only a ```rclcpp::Node::SharedPtr```?

audrow:
Fixes #855 for both synchronous or asynchronous parameter clients.

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -74,7 +74,6 @@ public:
    * \param[in] group (optional) The async parameter client will be added to this callback group.
    */
   template<typename NodeT>
-  RCLCPP_PUBLIC
   AsyncParametersClient(
     const std::shared_ptr<NodeT> node,
     const std::string & remote_node_name = "",
@@ -98,7 +97,6 @@ public:
    * \param[in] group (optional) The async parameter client will be added to this callback group.
    */
   template<typename NodeT>
-  RCLCPP_PUBLIC
   AsyncParametersClient(
     NodeT * node,
     const std::string & remote_node_name = "",
@@ -258,7 +256,6 @@ public:
   RCLCPP_SMART_PTR_DEFINITIONS(SyncParametersClient)
 
   template<typename NodeT>
-  RCLCPP_PUBLIC
   explicit SyncParametersClient(
     std::shared_ptr<NodeT> node,
     const std::string & remote_node_name = "",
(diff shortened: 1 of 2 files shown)


audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11763/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6839/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9597/)
* Windows [Build Status]

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11770/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6847/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9607/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11712/)

Issue #1255, opened 2020-08-03
[BUG] Namespacing the parameter blackboard node leads to wrong events topic


maxlein:
## Bug report

Our system has all nodes namespaced, e.g. /robot_1/param_server.  
Now we switched from eloquent to foxy, and all parameter events failed to fire.  

After a bit of testing and asking a [question](https://answers.ros.org/question/358170/parameter-events-on-foxy/?comment=358296#post-id-358296), I found out that adding a namespace breaks parameter events. 

Output from demo parameter blackboard started with namespace `test_ns`:  

```
ros2 node info /test_ns/parameter_blackboard
/test_ns/parameter_blackboard
  Subscribers:
    /test_ns/parameter_events: rcl_interfaces/msg/ParameterEvent
  Publishers:
    /parameter_events: rcl_interfaces/msg/ParameterEvent
    /rosout: rcl_interfaces/msg/Log
  Service Servers:
    /test_ns/parameter_blackboard/describe_parameters: rcl_interfaces/srv/DescribeParameters
    /test_ns/parameter_blackboard/get_parameter_types: rcl_interfaces/srv/GetParameterTypes
    /test_ns/parameter_blackboard/get_parameters: rcl_interfaces/srv/GetParameters
    /test_ns/parameter_blackboard/list_parameters: rcl_interfaces/srv/ListParameters
    /test_ns/parameter_blackboard/set_parameters: rcl_interfaces/srv/SetParameters
    /test_ns/parameter_blackboard/set_parameters_atomically: rcl_interfaces/srv/SetParametersAtomically
```  

parameter_events node:  

```
ros2 node info /test_ns/parameter_events
/test_ns/parameter_events
  Subscribers:
    /test_ns/parameter_events: rcl_interfaces/msg/ParameterEvent
  Publishers:
    /parameter_events: rcl_interfaces/msg/ParameterEvent
    /rosout: rcl_interfaces/msg/Log
```

As you see the parameter_events topic subscriber has a namespace, whereas the publisher has not.  
This doesn't look right to me..  

I can get it to work by remapping the node local event topic to a global one: `--ros-args -r parameter_events:=/parameter_events`

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries, foxy
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
[...]

PR #1257, opened 2020-08-03
Use global namespace for parameter events subscription topic

jacobperron:
Similar to https://github.com/ros2/rclcpp/pull/929, but for the subscription.

This fixes an issue listening to parameter events from a remote node when the local node has a different namespace.
Originally reported here: https://answers.ros.org/question/358170/parameter-events-on-foxy/

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -177,7 +177,7 @@ public:
   {
     return rclcpp::create_subscription<rcl_interfaces::msg::ParameterEvent>(
       node,
-      "parameter_events",
+      "/parameter_events",
       qos,
       std::forward<CallbackT>(callback),
       options);


jacobperron:
@ivanpauno Was the subscription topic left as a relative namespace on purpose, or simply overlooked?

ivanpauno:
> @ivanpauno Was the subscription topic left as a relative namespace on purpose, or simply overlooked?

It was overlooked.
Actually, I think a similar change is needed in `rclpy`.

clalancette:
@jacobperron To be clear, does this fix #1255 ?

jacobperron:
@clalancette Yeah, I think this should fix #1255. I didn't realize an issue had been opened.

@maxlein Are you able to confirm that this PR fixes the issue for you?

jacobperron:
Similar change in rclpy: https://github.com/ros2/rclpy/pull/613

jacobperron:

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11752/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6830/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9588/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11692/)

jacobperron:
I'm not sure if we want to backport this change since it is not necessarily broken, but changes behavior.

ivanpauno:
> I'm not sure if we want to backport this change since it is not necessarily broken in Foxy and it changes behavior.

Using an absolute topic for publishing an a relative topic for subscribing is IMO broken.
Having different behavior between rclpy and rclcpp is IMO broken too.

It's true that backporting it is a behavior change and might break some code.

ros-discourse:

PR #1261, opened 2020-08-05
[Foxy] Use global namespace for parameter events subscription topic (#1257)

jacobperron:
Backport #1257 to Foxy.

This changes default behavior and could consequently break application code that is relying on the assumption that the parameter events subscription uses a relative namespace (but the publisher is using a global namespace), therefore I'm hesitant about backporting. On the other hand, without the patch it can lead to buggy behavior when trying to use parameter events for the first time in Foxy.

Dashing and Eloquent don't have the same problem since both the publisher and subscription are using relative namespaces.

@ros2/team I'm looking for opinions on whether this fix should be backported or not.

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -177,7 +177,7 @@ public:
   {
     return rclcpp::create_subscription<rcl_interfaces::msg::ParameterEvent>(
       node,
-      "parameter_events",
+      "/parameter_events",
       qos,
       std::forward<CallbackT>(callback),
       options);


jacobperron:
If we decide to move forward with this backport, we should also backport the same change in rclpy: https://github.com/ros2/rclpy/pull/612

jacobperron:
rclpy backport: https://github.com/ros2/rclpy/pull/614

ros-discourse:

jacobperron:
It sounds like we haven't reached a consensus yet on whether to use global or private topic names for parameter events: 

But, since I think the current state in Foxy is "broken" because of the asymmetry between the parameter client and parameter sever, I'm going to move forward on this PR.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12583/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7550/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10295/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12507/)

PR #1274, opened 2020-08-11
fix topic stats test, wait for more messages, only check the ones with samples

### LINKED ISSUE #1273: topic statistic tests failing with Connext
In the nightly jobs using only Connext these test fail since they have been added in #1080:

* http://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/43/testReport/rclcpp/TestSubscriptionTopicStatisticsFixture/test_receive_stats_for_message_no_header/
* http://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/43/testReport/rclcpp/TestSubscriptionTopicStatisticsFixture/test_receive_stats_for_message_with_header/

@prajakta-gokhale @dabonnie Will either of you be able to work on this this in the near term?

dirk-thomas:
Fixes #1273.

Before: [Build Status](https://ci.ros2.org/job/ci_linux/11887/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/11891/)

@prajakta-gokhale @dabonnie FYI

### DIFF
--- a/rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
+++ b/rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
@@ -284,7 +284,7 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_stats_for_message_no
   auto statistics_listener = std::make_shared<rclcpp::topic_statistics::MetricsMessageSubscriber>(
     "test_receive_single_empty_stats_message_listener",
     "/statistics",
-    2);
+    4);
 
   auto empty_subscriber = std::make_shared<EmptySubscriber>(
     kTestSubNodeName,
@@ -301,12 +301,12 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_stats_for_message_no
     kTestDuration);
 
   // Compare message counts, sample count should be the same as published and received count
-  EXPECT_EQ(2, statistics_listener->GetNumberOfMessagesReceived());
+  EXPECT_EQ(4, statistics_listener->GetNumberOfMessagesReceived());
 
   // Check the received message and the data types
   const auto received_messages = statistics_listener->GetReceivedMessages();
 
-  EXPECT_EQ(2u, received_messages.size());
+  EXPECT_EQ(4u, received_messages.size());
 
   std::set<std::string> received_metrics;
   for (const auto & msg : received_messages) {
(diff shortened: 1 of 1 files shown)


rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -318,10 +318,27 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_stats_for_message_no
   // Check the collected statistics for message period.
   // Message age statistics will not be calculated because Empty messages
   // don't have a `header` with timestamp.
+  bool any_samples = false;
   for (const auto & msg : received_messages) {
     if (msg.metrics_source != "message_period") {
       continue;
     }
+    // skip messages without samples
```
dabonnie, line 326: Nit: could make this as a separate function (I should have done the same for some of the reused code) for the multiple tests.

dabonnie: Note: If you receive 4 messages I would expect that 2 of them are the `message_age` measurement and the remaining 2 are the `message_period` measurement. Message period is calculated on the subscriber side, whereas [message age requires that message timestamp is present](https://index.ros.org/doc/ros2/Concepts/About-Topic-Statistics/#types-of-statistics-calculated). From the [behavior](https://index.ros.org/doc/ros2/Concepts/About-Topic-Statistics/#behavior) notice how the [sample count is zero](https://index.ros.org/doc/ros2/Tutorials/Topics/Topic-Statistics-Tutorial/#observe-published-statistic-data), so that would denote the `message_age` measurement. 

My guess is this test was broken from the start because it didn't attempt to check the `message age` case, but only the `message period`. What we could do here instead is simply add a count for each type of measurement received and assert false otherwise, e.g., your 

```
    if (!has_samples) {
      continue;
    }
```

would just turn into checking a counter at the end instead of the boolean flag.

dirk-thomas: Both suggestions would be good for improving the test but for the scope of this PR which just aims to fix the failure with Connext I would rather not do those.

dabonnie: I will pick up adding the improvement. I was also incorrect regarding the counts above: if you receive 4 messages, I would expect 4 `message age` and 4 `message period` messages for a total of 8 statistics messages published: 1 for each of of the messages recieved.

PR #1275, opened 2020-08-12
Fix error when SubscriptionIntraProcess is being executed more than once by the MultiThreadedExecutor

ivanpauno:
This ensures that executing is a noop if there's nothing to be executed.

Alternative to https://github.com/ros2/rclcpp/pull/1241.

### DIFF
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
@@ -25,6 +25,7 @@
 #include "rcl/error_handling.h"
 
 #include "rclcpp/any_subscription_callback.hpp"
+#include "rclcpp/experimental/buffers/buffer_implementation_base.hpp"
 #include "rclcpp/experimental/buffers/intra_process_buffer.hpp"
 #include "rclcpp/experimental/create_intra_process_buffer.hpp"
 #include "rclcpp/experimental/subscription_intra_process_base.hpp"
@@ -159,10 +160,20 @@ private:
 
     if (any_callback_.use_take_shared_method()) {
       ConstMessageSharedPtr msg = buffer_->consume_shared();
-      any_callback_.dispatch_intra_process(msg, msg_info);
+      try {
+        any_callback_.dispatch_intra_process(msg, msg_info);
+      } catch (buffers::BufferEmptyError & e) {
+        // Ignore this error.
+        // The multithreaded executor might have scheduled this waitable for execution more than
+        // once.
+      }
     } else {
       MessageUniquePtr msg = buffer_->consume_unique();
-      any_callback_.dispatch_intra_process(std::move(msg), msg_info);
+      try {
+        any_callback_.dispatch_intra_process(std::move(msg), msg_info);
+      } catch (buffers::BufferEmptyError & e) {
+        // ignore
+      }
     }
   }
 
--- a/rclcpp/include/rclcpp/experimental/buffers/buffer_implementation_base.hpp
+++ b/rclcpp/include/rclcpp/experimental/buffers/buffer_implementation_base.hpp
@@ -35,6 +35,15 @@ public:
   virtual bool has_data() const = 0;
 };
 
+/// Thrown when trying to dequeue data from an emty buffer
+class BufferEmptyError : public std::runtime_error
+{
+public:
+  BufferEmptyError()
+  : std::runtime_error("trying to dequeue from an empty buffer") {}
+};
+
+
 }  // namespace buffers
 }  // namespace experimental
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
+++ b/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp
@@ -84,14 +84,14 @@ public:
    * This member function is thread-safe.
    *
    * \return the element that is being removed from the ring buffer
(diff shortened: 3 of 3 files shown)


rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -157,6 +157,9 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
     msg_info.publisher_gid = {0, {0}};
     msg_info.from_intra_process = true;
 
+    if (!buffer_->has_data()) {
```
audrow, line 160: I think that you want a `lock_guard` before this statement. I can imagine a segmentation fault as follows:
* Thread 1: `!buffer_->has_data()` evaluates to `false` so the thread skips over the if-statement
* Thread 2: consumes the data in `buffer_`
* Thread 1: Tries to consume the data that has already been consumed.

ivanpauno: Yeap, thanks for noting the issue!

ivanpauno: Solved in https://github.com/ros2/rclcpp/pull/1275/commits/3d0f8df8a86d93d362dc79cd7aaa95dcd9f9031b.

ivanpauno:
> I like how minimal this solution is. As you mention that this is just a fix for SubscriptionIntraProcess, the same kind of fix should be applied to all classes that inherit from Waitable

All the other Waitables that https://github.com/ros2/rclcpp/pull/1241 is modifying still have the same issue after and before the PR (for QOSEventHandler, action clients and services, #1241 isn't fixing the problem).
That's a big reason why I don't think #1241 is correct.

I can apply a fix to action clients and services, doing the same that subscriptions do: ignore when taking data fails.
I would prefer to do that in a separate PR.

For `QOSEventHandler`, the only possible fix is to add the concept of a Waitable that is mutually exclusive with itself (which can be implemented by creating a callback group and adding only that single Waitable to it).
Actually, I think this approach is nice for timers too.

ivanpauno:
> All the other Waitables that #1241 is modifying still have the same issue after and before the PR (for QOSEventHandler, action clients and services, #1241 isn't fixing the problem).

Correction: #1241 does fix the issue for action clients and services.

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -85,8 +85,6 @@ class MultiThreadedExecutor : public rclcpp::Executor
   size_t number_of_threads_;
   bool yield_before_execute_;
   std::chrono::nanoseconds next_exec_timeout_;
-
-  std::set<TimerBase::SharedPtr> scheduled_timers_;
```
sloretz, line 89: @ivanpauno @audrow This looks like it breaks ABI. Is it possible to implement this in a way that can be back ported to Foxy? One of the Foxy CI failures looks like it might be an instance of #1212 

http://build.ros2.org/job/Fci__nightly-fastrtps_ubuntu_focal_amd64/193/testReport/junit/(root)/projectroot/gtest_multithreaded__rmw_fastrtps_cpp/

ivanpauno: I can avoid deleting that line, but the patch will still break ABI because of the addition of:
https://github.com/ros2/rclcpp/blob/3d0f8df8a86d93d362dc79cd7aaa95dcd9f9031b/rclcpp/include/rclcpp/timer.hpp#L260

I don't see how to avoid breaking ABI.

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -198,6 +199,15 @@ class GenericTimer : public TimerBase
   void
   execute_callback() override
   {
+    // TODO(ivanpauno): Delete this workaround.
+    // The issue is that the multithreaded executor might schedule the timer
+    // more than once in different threads.
+    // Making `execute_callback` mutually exclusive and rechecking if the timer is ready
+    // avoids the issue.
+    std::lock_guard<std::mutex> lock(execution_mutex_);
```
sloretz, line 207: Is this related to #1212? It doesn't seem related to the intra process ring buffer being taken from twice. Could this be it's own pull request? Would that then allow backporting the #1212 fix to Foxy?

ivanpauno: > Could this be it's own pull request?

Yes, I actually should open an issue describing the problem.
I originally solved both here because they are related.

> Would that then allow backporting the #1212 fix to Foxy?

:+1:

ivanpauno: done, sorry for the delay

fujitatomoya (approved):
lgtm

audrow (approved):

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12864/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7819/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10575/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12807/)

wjwwood:
I generally do not like this approach, as I said here:

https://github.com/ros2/rclcpp/pull/1241#issuecomment-721421457

But I'll leave it to the other `rclcpp` maintainers to decide, since I don't have time to help architect a better solution.

PR #1276, opened 2020-08-12
Redesigning the MultiThreadedExecutor

audrow:
Adds a document for discussing how the `MultiThreadedExecutor` should be redesigned to avoid calling execute more than one time on one executable object. This discussion was prompted by #1241 and #1275.

### DIFF
--- a/rclcpp/doc/multi_threaded_executor_redesign.rst
+++ b/rclcpp/doc/multi_threaded_executor_redesign.rst
@@ -1,0 +1,60 @@
+======================================
+:code:`MultiThreadedExecutor` Redesign
+======================================
+
+The :code:`MultiThreadedExecutor` may attempt to execute different types of executables multiple times. 
+This occurs because one executable object can be identified as ready in two threads and both threads will attempt to execute it. 
+To fix this, a general solution should do the following:
+ * Work for all the types of executable objects: timers, subscriptions, services, clients, and waitables
+ * Avoid spurious wake ups (which can be costly)
+
+Summary of discussion so far
+----------------------------
+
+Broadly speaking, two approaches were discussed:
+
+1. **Handle no data in the executor**: If more than one thread is trying to execute the same object and that object is consumed after it is executed once, check for the executable before trying to execute it:
+
+    .. code-block:: c++
+
+        void execute() {
+          if(buffer->has_data()) {
+            auto data = buffer->consume_data();
+            run_any_callback(data);
+          } else {
+            // do nothing
+          }
+        }
+
+   :Advantages: 
+     * simplicity
+     * doesn't extend interface in a way that doesn't generalize well for all other types of executable objects (i.e., timers)
+   :Disadvantages: 
+     * spurious wake up calls
+     * needs to be documented well so every class that executes :code:`AnyExecutable` objects is thread-safe and doesn't hide a race condition
+
+
+2. **Take the data**: When an executable is ready, take it and store it so that it can be executed when execute is called:
+
+    .. code-block:: c++
+
+        void take_data(std::shared_ptr<void> & data) {
+          data = buffer->consume_data();
+        }
+
+        void execute(std::shared_ptr<void> & data) {
+          run_any_callback(data_);
+        }
+
+   :Advantages:
+     * no spurious wake ups
+   :Disadvantages:
+     * more complex
+     * extends the interface in a way that doesn't generalize well for all other types of executable objects (i.e., timers)
+     * type erasure of data
+
+Ideas
+-----------------------
+ * Add a preparation method to get the event after :code:`take_data`
+ * Take a lambda that can be executed later
+ * Store conditions in the waitset (like DDS)
--- a/rclcpp/doc/multi_threaded_executor_redesign.rst
+++ b/rclcpp/doc/multi_threaded_executor_redesign.rst
@@ -56,9 +56,26 @@ Broadly speaking, two approaches were discussed:
      * extends the interface in a way that doesn't generalize well for all other types of executable objects (i.e., timers)
      * type erasure of data
 
+.. note::
+
+   To avoid erasing the type of the data, :code:`take_data` could be replaced with a :code:`get_handle` method that would return a lambda function.
+
+   .. code-block:: c++
+
+      // called first
+      std::function<void()> get_handle() {
+        return [data=buffer->consume_data()]() {
+          run_any_callback(data);
+        };
+      }
+
+      // called second
+      void execute(std::function<void()> handle) {
+        handle();
+      }
+
 Ideas
 -----------------------
 
 * Add a preparation method to get the event after :code:`take_data`
-* Take a lambda that can be executed later
 * Store conditions in the waitset (like DDS)
--- a/rclcpp/doc/multi_threaded_executor_redesign.rst
+++ b/rclcpp/doc/multi_threaded_executor_redesign.rst
@@ -5,8 +5,9 @@
 The :code:`MultiThreadedExecutor` may attempt to execute different types of executables multiple times. 
 This occurs because one executable object can be identified as ready in two threads and both threads will attempt to execute it. 
 To fix this, a general solution should do the following:
- * Work for all the types of executable objects: timers, subscriptions, services, clients, and waitables
- * Avoid spurious wake ups (which can be costly)
+
+* Work for all the types of executable objects: timers, subscriptions, services, clients, and waitables
+* Avoid spurious wake ups (which can be costly)
 
 Summary of discussion so far
 ----------------------------
@@ -55,6 +56,7 @@ Broadly speaking, two approaches were discussed:
 
 Ideas
 -----------------------
- * Add a preparation method to get the event after :code:`take_data`
- * Take a lambda that can be executed later
- * Store conditions in the waitset (like DDS)
+
+* Add a preparation method to get the event after :code:`take_data`
+* Take a lambda that can be executed later
+* Store conditions in the waitset (like DDS)
--- a/rclcpp/doc/multi_threaded_executor_redesign.rst
+++ b/rclcpp/doc/multi_threaded_executor_redesign.rst
@@ -77,5 +77,26 @@ Broadly speaking, two approaches were discussed:
 Ideas
 -----------------------
 
-* Add a preparation method to get the event after :code:`take_data`
+* Add a preparation method to get the event after :code:`take_data`:
+
+  .. code-block:: c++
+
+     std::shared_ptr<void> & data_;
+
+     // called first
+     void take_data() {
+       data_ = buffer->consume_data();
+     }
+
+     // called second
+     std::shared_ptr<void> get_prepared() {
+       return data_;
+     }
+
+     // called third, with the data returned in the previous step
+     void execute(std::shared_ptr<void> & data) {
+       run_any_callback(data);
+     }
+
+
 * Store conditions in the waitset (like DDS)
--- a/rclcpp/doc/multi_threaded_executor_redesign.rst
+++ b/rclcpp/doc/multi_threaded_executor_redesign.rst
@@ -35,20 +35,22 @@ Broadly speaking, two approaches were discussed:
      * needs to be documented well so every class that executes :code:`AnyExecutable` objects is thread-safe and doesn't hide a race condition
 
 
-2. **Take the data**: When an executable is ready, take it and store it so that it can be executed when execute is called:
+2. **Take the data**: This approach takes the data when collecting the ready entities, and that process is done in a mutually exclusive fashion. Thus, this approach may avoid spurious wake ups.
 
     .. code-block:: c++
 
+        // called first
         void take_data(std::shared_ptr<void> & data) {
           data = buffer->consume_data();
         }
 
+        // called second
         void execute(std::shared_ptr<void> & data) {
           run_any_callback(data_);
         }
 
    :Advantages:
-     * no spurious wake ups
+     * fixes spurious wake ups for events where :code:`take_data` has a meaning (i.e., not timers)
    :Disadvantages:
      * more complex
      * extends the interface in a way that doesn't generalize well for all other types of executable objects (i.e., timers)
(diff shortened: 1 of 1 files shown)


rclcpp/doc/multi_threaded_executor_redesign.rst
```diff
@@ -0,0 +1,62 @@
...
+            auto data = buffer->consume_data();
+            run_any_callback(data);
+          } else {
+            // do nothing
+          }
+        }
+
+   :Advantages: 
+     * simplicity
+     * doesn't extend interface in a way that doesn't generalize well for all other types of executable objects (i.e., timers)
```
gbiggs, line 32:
```suggestion
     * extends interface in a way that generalizes well for all other types of executable objects (i.e., timers)
```

rclcpp/doc/multi_threaded_executor_redesign.rst
```diff
@@ -0,0 +1,62 @@
...
+        void take_data(std::shared_ptr<void> & data) {
+          data = buffer->consume_data();
+        }
+
+        void execute(std::shared_ptr<void> & data) {
+          run_any_callback(data_);
+        }
+
+   :Advantages:
+     * no spurious wake ups
```
ivanpauno, line 51: As commented above, this **may** avoid spurious wake ups.
But for example, it doesn't fix the problem for timers (though, it does fix the problem in all waitables were "taking data" has a meaning).

I still like the idea of somehow "taking the data" early.
I don't think that "taking data" is conceptually correct, more on that in a different comment.

For the entities where "taking data" doesn't' have sense, the problem can be fixed by making them mutually exclusive with themselves (one thread cannot wait or execute the entity if another thread is executing it).

rclcpp/doc/multi_threaded_executor_redesign.rst
```diff
@@ -0,0 +1,62 @@
...
+
+    .. code-block:: c++
+
+        void take_data(std::shared_ptr<void> & data) {
+          data = buffer->consume_data();
+        }
+
+        void execute(std::shared_ptr<void> & data) {
+          run_any_callback(data_);
+        }
```
ivanpauno, line 48: My main concern is how data is type erased here.
I don't think that's a good idea for a user facing API.

I think it's a better idea to have something like:

```cpp
std::function<void()> get_handle() {
  return [data=buffer->consume_data()]() {
    run_any_callback(data);
  };
}
```

In that way, data is not type erased.

IIRC @wjwwood raised the concern that this approach forces entities to have a way of "handling" the data, and it would be nice in the future to be able to create, for example, a `Subscription` where you can directly take the data.

Though that would be nice, all our current API is based on callbacks.
i.e.: you need to actually pass a callback when creating a `Subscription`/etc.

If we want to add support to a WaitSet+take approach (which I think it would be nice for some users), we will actually need to split entities e.g. `Subscription` and `ExecutableSubscription`.

In that approach, I think that waitables actually don't need to have the concept of "taking data" neither.
You only need to be able to check if the waitable is ready, and then you can execute whatever code is needed.

----

Longer term:

I think we should separate the concept of Waitable/Condition from the concept of Executable.
The current design mixes both, and  they aren't necessarily the same.

I would like to write how I think that can be done, but I don't think that's the goal of this discussion.

rclcpp/doc/multi_threaded_executor_redesign.rst
```diff
@@ -0,0 +1,62 @@
...
+     * no spurious wake ups
+   :Disadvantages:
+     * more complex
+     * extends the interface in a way that doesn't generalize well for all other types of executable objects (i.e., timers)
+     * type erasure of data
+
+Ideas
+-----------------------
+
+* Add a preparation method to get the event after :code:`take_data`
```
ivanpauno, line 60: I would like to see an example of this.
I don't understand if that's what I'm proposing, or if it's something different.

ivanpauno: ok, the next item is what I'm proposing, so I don't understand this one much

rclcpp/doc/multi_threaded_executor_redesign.rst
```diff
@@ -0,0 +1,62 @@
...
+* Take a lambda that can be executed later
+* Store conditions in the waitset (like DDS)
```
ivanpauno, line 62: I think that having the concept of conditions (like DDS) is nice, but I'm not sure what this item means exactly.

rclcpp/doc/multi_threaded_executor_redesign.rst
ivanpauno: If we don't erase the type in the interface, I'm convinced that https://github.com/ros2/rclcpp/pull/1275 is a good idea (each time I read the PR I'm a bit confused, but I think I'm convinced now :joy:).
Without that, I still don't like the proposal much (sorry for being a bit hardheaded).

Maybe there's a different way of avoiding erasing the type, instead of my `get_handle` idea. I would like to hear of other alternatives!

rclcpp/doc/multi_threaded_executor_redesign.rst
audrow: > For the entities where "taking data" doesn't' have sense, the problem can be fixed by making them mutually exclusive with themselves (one thread cannot wait or execute the entity if another thread is executing it).

Should this be another solution listed? Would this work for all entity types? If so, could you briefly describe its implementation?

rclcpp/doc/multi_threaded_executor_redesign.rst
```diff
@@ -0,0 +1,62 @@
...
+
+   :Advantages: 
+     * simplicity
+     * extends interface in a way that generalizes well for all other types of executable objects (i.e., timers)
+   :Disadvantages: 
+     * spurious wake up calls
+     * needs to be documented well so every class that executes :code:`AnyExecutable` objects is thread-safe and doesn't hide a race condition
+
+
+2. **Take the data**: When an executable is ready, take it and store it so that it can be executed when execute is called:
```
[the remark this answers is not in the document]
audrow: I've updated the description to use most of your wording, including that this method may avoid spurious wakeups.

rclcpp/doc/multi_threaded_executor_redesign.rst
audrow: I've added the `get_handle` method in with an example. Let me know what you think.

rclcpp/doc/multi_threaded_executor_redesign.rst
audrow: I've added an example.

For what you're proposing, it is now in an example in the note after item 2.

rclcpp/doc/multi_threaded_executor_redesign.rst
audrow: We spoke about it in our meeting. I'm not sure that I understand it well enough to explain. 

It was something like, we could do what DDS does, which is use conditions that can be peaked at or read. Once a condition is read, it is removed from the queue, so they don't come up again. I believe this would avoid spurious wakeups. @wjwwood, is this correct?

rclcpp/doc/multi_threaded_executor_redesign.rst
ivanpauno: > Should this be another solution listed?

Distinguishing between entities that can be executed concurrently (e.g. you can take data from a subscription from different threads) from the ones that cannot sounds like right to me.

> Would this work for all entity types?

In entities like `timer`, waiting on it to be ready on one thread while scheduling it for execution in another thread is a problem.
For other entities (e.g. subscriptions), that would only worsen performance.

> If so, could you briefly describe its implementation?

We could have a flag or something that differentiates between "concurrent" and "non-concurrent" entities.

We could create a [mutually exclusive callback group](https://github.com/ros2/rclcpp/blob/0e0a6a495c452bdf3526a0b7bf457e4186bb344d/rclcpp/include/rclcpp/callback_group.hpp#L45) for each timer instead of putting them in the [default callback group](https://github.com/ros2/rclcpp/blob/0e0a6a495c452bdf3526a0b7bf457e4186bb344d/rclcpp/include/rclcpp/node_interfaces/node_base.hpp#L92).
Currently when an entity on a ME callback group is being executed, all the entities in the group (including the one scheduled for execution) cannot be waited on in another thread neither be executed.

rclcpp/doc/multi_threaded_executor_redesign.rst
ivanpauno: Currently in a ROS 2 wait set you can add different kind of entities (subscription, service, ...).
In `DDS`, you can only [add conditions](https://community.rti.com/static/documentation/connext-dds/5.3.0/doc/api/connext_dds/api_cpp2/classdds_1_1core_1_1cond_1_1WaitSet.html#acd93c6ad7eb1da7f1f942b57e3388da6) to a WaitSet.
[Conditions API](https://community.rti.com/static/documentation/connext-dds/5.3.0/doc/api/connext_dds/api_cpp2/classdds_1_1core_1_1cond_1_1Condition.html) is quite simple, you can only check if them are ready or dispatch an attached callback.

There's no concept of "reading" or "peaking" a condition AFAIU.
I think that having something similar to DDS conditions will simplify our custom handling of different cases in wait set and executor code a lot (and the abstractions looks nice too :smiley: ).

rclcpp/doc/multi_threaded_executor_redesign.rst
wjwwood: >We could create a mutually exclusive callback group for each timer instead of putting them in the default callback group.

How would that help? It breaks the existing expectation of users which is that callbacks are not called concurrently unless they put them in different callback groups (the same as ROS 1).

ivanpauno: > How would that help? It breaks the existing expectation of users which is that callbacks are not called concurrently unless they put them in different callback groups (the same as ROS 1).

AFAIU:
- Callbacks from different callback groups **can** be executed concurrently.
- Callbacks from the same Reentrant Callback Group (RCG) **can** be executed concurrently.
- Callbacks from the same Mutually Exclusive Callback Group (MECG) **cannot** be executed concurrently.

For entities that aren't reentrant (e.g. timers) we could do the following:
- By default each timer is added to a new MECG that only contains that timer, and that MECG will not be used for other entities later.
- The user can pass a custom MECG. Passing a RCG is not allowed.

I think that doesn't break user assumptions. If it does break them, I'm probably miss-understanding how callback groups work.

This might not be the ideal way of implementing non-reentrant entities, maybe having an ad-hoc flag is a better idea.

rclcpp/doc/multi_threaded_executor_redesign.rst
ivanpauno: thanks

rclcpp/doc/multi_threaded_executor_redesign.rst
ivanpauno: looks good

rclcpp/doc/multi_threaded_executor_redesign.rst
wjwwood: Why are timers not reentrant? You should be able to have a timer in a reentrant callback group and have timer callbacks being called concurrently, if that's what you desire.

If you put each timer in its own MECG, then it can be run concurrently with subscription callbacks (assuming the user did not specify a custom CG for either), which is different from ROS 1, where the callbacks are mutually exclusive by default (even between subscriptions and timers).

Putting ROS 1 aside, if you right now create a timer and a subscription, and then spin in a multi-threaded executor they are mutually exclusive, your proposed change would change that behavior, breaking their expectation.

Either way, I don't see how this fixes the problem, but that's likely due to me loosing context on the conversation...

ivanpauno: > Why are timers not reentrant? You should be able to have a timer in a reentrant callback group and have timer callbacks being called concurrently, if that's what you desire.

We have a race condition when trying to wait on timers in a thread while another thread has scheduled its execution, and sometimes the timer is executed twice in a row ([here a failure in CI](https://ci.ros2.org/view/nightly/job/nightly_osx_release/1712/testReport/projectroot/test/test_multi_threaded_executor/), but it's almost impossible to reproduce the failure, it's extremely flaky).
There's a patch in the executor trying to avoid the race condition, but I think the patch suffers from a race condition too as explained [here](https://github.com/ros2/rclcpp/pull/1241#issuecomment-670530455).
The approach proposed in https://github.com/ros2/rclcpp/pull/1241 doesn't solve the problem and I proposed a custom workaround in https://github.com/ros2/rclcpp/pull/1275.

I think that having the concept of entities that cannot be executed reentrantly could also solve the problem, and I think that's particular suitable for timers.

> If you put each timer in its own MECG, then it can be run concurrently with subscription callbacks (assuming the user did not specify a custom CG for either), which is different from ROS 1, where the callbacks are mutually exclusive by default (even between subscriptions and timers).

Sorry, I thought that the default callback group was reentrant and not ME ([code here](https://github.com/ros2/rclcpp/blob/01d6f52e3281f6df2c191690527f9208006a5c50/rclcpp/src/rclcpp/node_interfaces/node_base.cpp#L150)). The change I proposed wouldn't be acceptable.

fujitatomoya (commented):
to avoid spurious wake-up is really nice and cost effective. this can also fix the https://github.com/ros2/ros2/issues/1035 and https://github.com/ros2/rclcpp/issues/1399.

rclcpp/doc/multi_threaded_executor_redesign.rst
wjwwood: I'm still convinced #1275 is _not_ the right solution, also sorry for being hardheaded, and I don't have an issue with the type erasure, since the root cause is that we have to deal with type erased data at the the `rcl`/`rmw` layer anyways.

If we added the "handle" which contains a type erased pointer, which the callback could re-type internally, I really don't see that issue with what was proposed here, w.r.t. the `std::shared_ptr<void> & data`.

ivanpauno: > If we added the "handle" which contains a type erased pointer, which the callback could re-type internally, I really don't see that issue with what was proposed here, w.r.t. the std::shared_ptr<void> & data.

The difference is that you don't expose to the user a `void *`.

audrow:
I'm going to close this because there hasn't been movement on this in a few years. Feel free to open this if it becomes relevant again.

Issue #1265, opened 2020-08-09
support wildcards for just namespaces, or just nodes


rpaaron:
#### Feature description
support wildcards for just namespaces, or just nodes

```yaml
/**:
    ros__parameters:
        applies_to_all_nodes_in_all_namespaces: true
```

would be nice also to have
```yaml
some_namespace:
    *:
        ros__parameters:
            applies_to_all_nodes_in_particular_namespace: true
```
and
```yaml
/*:
    some_node:
        ros__parameters:
            applies_to_particular_node_in_all_namespaces: true
```

is there a way currently to achieve this?

fujitatomoya:
I think that it would be really nice to describe wildcard format in https://index.ros.org/doc/ros2/Tutorials/Node-arguments/#setting-parameters-from-yaml-files. (After this PR is merged.)

iuhilnehc-ynos:
@rpaaron

Let me confirm the feature as following,

- The following configuration can be accepted by a node on 'some_namespace' namespace, whatever the node name is.

```
some_namespace:
    *:
        ros__parameters:
            applies_to_all_nodes_in_particular_namespace: true
```

- The following configuration can be accepted by a node name 'some_node' on whatever the namespace is.

```
/*:
    some_node:
        ros__parameters:
            applies_to_particular_node_in_all_namespaces: true
```

If the maintainer here accepts these features, I'd like to contribute to it.

iuhilnehc-ynos:
@rpaaron 

Sorry, I didn't see your PR. Delete my previous comment.

clalancette:
@rpaaron Thanks for the issue and the pull request implementing it.  Before we go too far on the implementation, though, we should settle on the design first.  Towards that end, could you please propose a modification to http://design.ros2.org/articles/static_remapping.html for the design?  Thanks.

jacobperron:
Correct me if I'm wrong, but using wildcards as proposed here looks consistent with how [name matching is described in the design doc](http://design.ros2.org/articles/static_remapping.html#match-part-of-a-rule). It still might be useful to add concrete examples when it comes to using them in parameter files, either in the design doc or in the tutorial as @fujitatomoya suggests.

rpaaron:
I think these changes are inline with what's written in the design document already.  If not, could someone please advise what needs to change.

jacobperron:
[...]

PR #1280, opened 2020-08-18
Support namespace and node name wildcards

rpaaron:
Addresses #1265 
I'd like to add tests but adding wildcards to params via arguments returns an error.  Alternative is to add a .yml file, but im not sure about adding a filesystem dependency to the tests

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -110,10 +110,11 @@ NodeParameters::NodeParameters(
 
       // Enforce wildcard matching precedence
       // TODO(cottsay) implement further wildcard matching
+      auto ns = std::string(node_base->get_namespace());
       const std::vector<std::string> node_matching_names{
         "/**",
         "/*/" + std::string(node_base->get_name()),
-        std::string(node_base->get_namespace()) + "/*",
+        (ns == "/" ? "" : ns) + "/*",
         combined_name_};
 
       for (const auto & node_name : node_matching_names) {
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -110,7 +110,12 @@ NodeParameters::NodeParameters(
 
       // Enforce wildcard matching precedence
       // TODO(cottsay) implement further wildcard matching
-      const std::vector<std::string> node_matching_names{"/**", combined_name_};
+      const std::vector<std::string> node_matching_names{
+        "/**",
+        "/*/" + std::string(node_base->get_name()),
+        std::string(node_base->get_namespace()) + "/*",
+        combined_name_};
+
       for (const auto & node_name : node_matching_names) {
         if (initial_map.count(node_name) > 0) {
           // Combine parameter yaml files, overwriting values in older ones
--- a/rclcpp/test/resources/test_node_parameters/wildcards.yaml
+++ b/rclcpp/test/resources/test_node_parameters/wildcards.yaml
@@ -17,6 +17,14 @@ ns:
     ros__parameters:
       explicit_in_ns: "explicit_in_ns" 
 
+"*":
+  ros__parameters:
+    node_wild_no_ns: "node_wild_no_ns" 
+
+node:
+  ros__parameters:
+    explicit_no_ns: "explicit_no_ns" 
+
 ns:
   nodeX:
     ros__parameters:
--- a/rclcpp/test/resources/test_node_parameters/wildcards.yaml
+++ b/rclcpp/test/resources/test_node_parameters/wildcards.yaml
@@ -1,0 +1,28 @@
+/**:
+  ros__parameters:
+    full_wild: "full_wild" 
+
+/*:
+  node:
+    ros__parameters:
+      namespace_wild: "namespace_wild" 
+
+ns:
+  "*":
+    ros__parameters:
+      node_wild_in_ns: "node_wild_in_ns" 
+
+ns:
+  node:
+    ros__parameters:
+      explicit_in_ns: "explicit_in_ns" 
+
+ns:
+  nodeX:
+    ros__parameters:
+      should_not_appear: "incorrect_node_name" 
+
+nsX:
+  node:
+    ros__parameters:
+      should_not_appear: "incorrect_namespace" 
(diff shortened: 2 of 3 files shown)


rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -109,8 +109,12 @@ NodeParameters::NodeParameters(
       rclcpp::ParameterMap initial_map = rclcpp::parameter_map_from(params);
 
       // Enforce wildcard matching precedence
-      // TODO(cottsay) implement further wildcard matching
-      const std::vector<std::string> node_matching_names{"/**", combined_name_};
+      const std::vector<std::string> node_matching_names{
+        "/**",
+        "/*/" + std::string(node_base->get_name()),
+        std::string(node_base->get_namespace()) + "/*",
```
fujitatomoya, line 115: what if namespace is NOT specified, this will be `//*`. i think this is not handled as expected.

rpaaron:
> besides inline comment, it would be nice to do test with possible combination.
> 
> > Alternative is to add a .yml file, but im not sure about adding a filesystem dependency to the tests
> 
> https://github.com/ros2/rclcpp/blob/3270aad95eda0bc5ac7a3a79d4b1e07273154340/rclcpp/test/rclcpp/test_node.cpp#L688
> 
> `declare_parameter_with_cli_overrides` test uses 'test_parameters.yaml, i think that you can refer to.

thats great, thank you kindly

fujitatomoya (approved):
LGTM, but like to hear from others about design.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -109,8 +109,13 @@ NodeParameters::NodeParameters(
       rclcpp::ParameterMap initial_map = rclcpp::parameter_map_from(params);
 
       // Enforce wildcard matching precedence
-      // TODO(cottsay) implement further wildcard matching
```
iuhilnehc-ynos, line 112: If removing this line, maybe we need to add more implementations.
such as  (I am not asking you to implement it, I just thought we might keep this line)

```
asia_*:
    *:
        ros__parameters:
            applies_to_all_nodes_in_particular_namespace_with_asia: true
```

rpaaron: Good point, thank you, reinstated

rclcpp/test/resources/test_node_parameters/wildcards.yaml
```diff
@@ -0,0 +1,36 @@
...
+  ros__parameters:
+    full_wild: "full_wild" 
+
+/*:
+  node:
+    ros__parameters:
+      namespace_wild: "namespace_wild" 
+
+ns:
+  "*":
```
iuhilnehc-ynos, line 11: Even if * without double quotes in YAML is invalid, I still think the node wildcards without (") here should be supported, just like the namespace, such as (/*). 
```
ns:
  /*:           # "*" can be also supported, but I think users like this style.
    ros__parameters:
      node_wild_in_ns: "node_wild_in_ns" 
```

It should be discussed with the maintainer. [rcl_yaml_param_parser](https://github.com/ros2/rcl/blob/911088d8cb6ecba14290c399cc65f86adc3e3bf7/rcl_yaml_param_parser/src/parser.c#L205-L223) should be updated if the above suggestion is accepted.

clalancette:
The same comment as https://github.com/ros2/rclcpp/issues/1265#issuecomment-690653120 applies to this PR.

mabelzhang:
@rpaaron Friendly ping. Have you had a chance to look into the design raised in the previous comment?

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
jacobperron: FWIW, the [current design doc](http://design.ros2.org/articles/static_remapping.html#match-part-of-a-rule) explicitly disallows partial matching of tokens (e.g. `*bar`), so I don't think we should worry about supporting this case (unless the design changes).

rclcpp/test/resources/test_node_parameters/wildcards.yaml
jacobperron: Supporting both `/*` and `"*"` makes sense to me :+1:

rclcpp/test/resources/test_node_parameters/wildcards.yaml
```diff
@@ -0,0 +1,36 @@
+/**:
+  ros__parameters:
+    full_wild: "full_wild" 
+
+/*:
```
jacobperron, line 5: I think we should have another test case for matching namespaces with multiple tokens. Following the design doc:

- `/*` should match exactly one token (e.g. `/foo` or `/bar`, but not `/foo/bar` or `/`)
- `/**` should match zero or more tokens (e.g. `/foo` or `/bar` or `/foo/bar` or `/`)

rpaaron:
> @rpaaron Friendly ping. Have you had a chance to look into the design raised in the previous comment?

thanks @mabelzhang can you advise what I need to change here?

clalancette:
> thanks @mabelzhang can you advise what I need to change here?

Essentially this comment: https://github.com/ros2/rclcpp/issues/1265#issuecomment-690653120 .  That is, some updates to the design document so that we are all on the same page.

rclcpp/test/resources/test_node_parameters/wildcards.yaml
jacobperron: I take back my previous comment. Name tokens separated by a YAML colon (`:`) should be treated as if they are separated with a slash (`/`). Therefore, the example:

```yaml
ns:
  /*:
    ros__parameters:
      ...
```

Should resolve to the name `/ns//*`, which is invalid. The correct way to write it would be:

```yaml
/ns/*:
  ros__parameters:
    ...
```

fujitatomoya:
@rpaaron can you rebase this on master?

fujitatomoya:
@rpaaron sorry for the confusion, i need to take back the previous request. could you help the review on https://github.com/ros2/rclcpp/pull/1839 instead. (please see https://github.com/ros2/rclcpp/pull/1839#pullrequestreview-835692216)

fujitatomoya:
I will go ahead to close this one in favor of https://github.com/ros2/rclcpp/pull/1839.
CC: @rpaaron @iuhilnehc-ynos

PR #1290, opened 2020-09-04
Increase coverage rclcpp_action to 95%

brawner:
This adds unit tests for rclcpp_action and increases the coverage to just bit over 95%. As part of this work, an issue was found with the client implementation, which is potentially addressed in #1292. This branch is currently targeting that PR branch for easy review.

Testing `--packages-select rclcpp_action`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12240/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7218/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9960/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12134/)

Signed-off-by: Stephen Brawner <brawner@gmail.com>

PR #738, opened 2019-05-24
Guard against calling null goal response callback

jacobperron:
Also make sure to invoke the goal response callback before the result callback.

Addresses review comment https://github.com/ros2/rclcpp/pull/701#pullrequestreview-233076772.

PR #1292, opened 2020-09-08
Set promise value after try-catch

brawner:
This PR fixes a potential issue, where `promise->set_value` is called before a potential call to `promise->set_exception`. Only one can be called, otherwise a `std::future_error` is thrown. This was the original ordering in https://github.com/ros2/rclcpp/pull/594, but it was rearranged in https://github.com/ros2/rclcpp/pull/738. @jacobperron @hidmic if this is breaking something unintentionally, please let me know. It should pass current unit tests as is. This patch is required in order to pass tests in #1290.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -370,14 +370,6 @@ public:
         // Do not use std::make_shared as friendship cannot be forwarded.
         std::shared_ptr<GoalHandle> goal_handle(
           new GoalHandle(goal_info, options.feedback_callback, options.result_callback));
-        {
-          std::lock_guard<std::mutex> guard(goal_handles_mutex_);
-          goal_handles_[goal_handle->get_goal_id()] = goal_handle;
-        }
-        promise->set_value(goal_handle);
-        if (options.goal_response_callback) {
-          options.goal_response_callback(future);
-        }
 
         if (options.result_callback) {
           try {
@@ -386,6 +378,16 @@ public:
             promise->set_exception(std::current_exception());
             return;
           }
+        }
+
+        {
+          std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+          goal_handles_[goal_handle->get_goal_id()] = goal_handle;
+        }
+
+        promise->set_value(goal_handle);
+        if (options.goal_response_callback) {
+          options.goal_response_callback(future);
         }
       });
 
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -262,7 +262,7 @@ public:
   using GoalHandle = ClientGoalHandle<ActionT>;
   using WrappedResult = typename GoalHandle::WrappedResult;
   using GoalResponseCallback =
-    std::function<void (std::shared_future<typename GoalHandle::SharedPtr>)>;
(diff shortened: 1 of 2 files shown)


brawner:
Testing current unit tests, `--packages-select rclcpp_action`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12239/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7217/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9959/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12133/)

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -387,6 +379,16 @@ class Client : public ClientBase
...
         }
+
+        {
+          std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+          goal_handles_[goal_handle->get_goal_id()] = goal_handle;
+        }
+
+        promise->set_value(goal_handle);
+        if (options.goal_response_callback) {
+          options.goal_response_callback(future);
```
jacobperron, line 390: It was moved in response to https://github.com/ros2/rclcpp/pull/701#pullrequestreview-233076772; if  `make_result_aware` is called first, we risk having the result callback triggered before the goal response callback (unlikely, but I *think* this is possible if the action server is running in the same process).

brawner:
Testing with unit tests introduced in #1290

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12240/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7218/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9960/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12134/)

jacobperron (commented):
Taking a step back, to me the promise pertains to the *goal handle* (which may eventually contain a result). Therefore, I don't think it makes sense to set an exception on the promise if there's an issue requesting the result, rather that exception should somehow manifest itself later when trying to access the result. Does that make sense?

rclcpp_action/include/rclcpp_action/client.hpp
hidmic: Hmm, reflecting on this, it seems to me that the goal response callback should simply be taking a concrete  goal handle instead of a future one. Additionally, failing to make the goal handle result aware should not result in caller code losing access to the goal handle. Getting the future result should throw, but the exception could carry the goal handle (which is still valid).

jacobperron: @hidmic I agree with everything you've said. I think it's okay to not backport changes (though maybe we could by supporting two different signatures for the goal response callback).

brawner: Adjusted the goal_response_callback signature, but I'm still not decided about what to do exception and goal handle bit.

jacobperron: I think we can remove the try-catch altogether. I'll open a PR in a moment with a proposed alternative.

jacobperron: See https://github.com/ros2/rclcpp/pull/1311.

brawner:
Closing in favor of #1311

PR #1294, opened 2020-09-09
Remove unused private function (rclcpp::Node and rclcpp_lifecycle::Node)

brawner:
This function `group_in_node` is declared private, but is otherwise unused. This is a PR for removing it, but I can also move it to protected or public if it's inclusion might be helpful.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -1182,10 +1182,6 @@ protected:
 private:
   RCLCPP_DISABLE_COPY(Node)
 
-  RCLCPP_PUBLIC
-  bool
-  group_in_node(CallbackGroup::SharedPtr group);
-
   rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_;
   rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph_;
   rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr node_logging_;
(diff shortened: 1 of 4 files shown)


fujitatomoya (commented):
what about `Node::group_in_node`? https://github.com/ros2/rclcpp/blob/d66cd96f254648de5fa68f6c28a34968563294fb/rclcpp/include/rclcpp/node.hpp#L1185-L1187

is there concrete use case for this method but `LifecycleNode::group_in_node`?

clalancette:
> This function `group_in_node` is declared private, but is otherwise unused. This is a PR for removing it, but I can also move it to protected or public if it's inclusion might be helpful.

Overall, I'm definitely for removing code that isn't used.  Particularly if it is a pretty simple method like this one.  So I'd be for just removing it.

@fujitatomoya makes a good point about rclcpp though; we should probably remove it from there as well (since it doesn't look to be used there either).

brawner:
Agreed, it also looks like it can be removed from rclcpp::Node as other functions call node_base->callback_group_in_node directly.

brawner:
Testing `--packages-select rclcpp rclcpp_lifecycle`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12263/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7237/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9981/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12153/)

clalancette (approved):
I love removing dead code.  Go go go!

PR #1297, opened 2020-09-10
Log error instead of throwing exception in Transition and State reset(), mark no except

brawner:
This was discovered during fault injection testing, but if for some reason `rcl_lifecycle_state_fini` fails during `rclcpp::Transition::reset()` or `rclcpp::State::reset()`, these destructors will throw an uncatchable exception because destructors are `noexcept` by default.

There are other potential resolutions to this issue, like marking these destructors `noexcept(false)`, but that has performance side effects by poisoning the chain of destructors that eventually call these.

### DIFF
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -83,13 +83,7 @@ State::State(const State & rhs)
 
 State::~State()
 {
-  try {
-    reset();
-  } catch (...) {
-    RCLCPP_ERROR(
-      rclcpp::get_logger("rclcpp_lifecycle"),
-      "Resetting rclcpp_lifecycle::State threw during destruction, leaking memory");
-  }
+  reset();
 }
 
 State &
@@ -150,7 +144,7 @@ State::label() const
 }
 
 void
-State::reset()
+State::reset() noexcept
 {
   if (!owns_rcl_state_handle_) {
     state_handle_ = nullptr;
@@ -164,7 +158,9 @@ State::reset()
   allocator_.deallocate(state_handle_, allocator_.state);
   state_handle_ = nullptr;
   if (ret != RCL_RET_OK) {
-    rclcpp::exceptions::throw_from_rcl_error(ret);
+    RCLCPP_ERROR(
+      rclcpp::get_logger("rclcpp_lifecycle"),
+      "rcl_lifecycle_transition_fini did not complete successfully, leaking memory");
   }
 }
 
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -21,6 +21,7 @@
 #include "rcl_lifecycle/rcl_lifecycle.h"
 
 #include "rclcpp/exceptions.hpp"
+#include "rclcpp/logging.hpp"
 
 #include "rcutils/allocator.h"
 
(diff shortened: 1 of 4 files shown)


clalancette:
> There are other potential resolutions to this issue, like marking these destructors `noexcept(false)`, but that has performance side effects by poisoning the chain of destructors that eventually call these.

Good catch.  I like this solution better than the marking of the destructors.  I do have one comment, but I'll put it inline.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -82,7 +83,13 @@ State::State(const State & rhs)
 
 State::~State()
 {
-  reset();
+  try {
+    reset();
+  } catch (...) {
```
clalancette, line 88: Catching any exception here is "safer", in that anything that happens to throw will be caught.  But on the other hand, that same logic goes for every destructor in our codebase, and we don't wrap them all in `try { } catch(...) {}`.  What I'm trying to get at here is that I think this may be better as:

```
} catch (const RCLErrorBase &) {
```

(looks deeper).  Oh, but `rclcpp::exception::throw_from_rcl_error` can also throw a `std::invalid_argument` or `std::runtime_error`.  That's unfortunate.  Do you have any thoughts here on how we could avoid the catch-all, or do you think using the catch-all is the right solution?

fujitatomoya: I think it would be better to put `try & catch` for now, but i agree with @clalancette . although i do not have any better idea for now.

brawner: I wasn't terribly convinced of any one solution. The catch all is pretty useful, since reset is a function that can throw several different types of exceptions, and it's simple to write. `reset` gets plenty of usage in the unit tests, so I think if it were to start throwing a different type of exception we'd start seeing it show up there.

Other ideas I had...

`rcl_lifecycle_state_fini` currently only fails if it's passed a bad allocator, which would be a programming error. But if it did fail, it would only leak memory. I could move the RCLCPP_ERROR message to `reset()` and just ignore a bad return value. In that case, it might also be good to make `reset()` `noexcept` as well, so other exceptions thrown from inside that function are also identified and resolved.

I could remove the fault injection hook from `rcl_lifecycle_state_fini`. It wouldn't decrease coverage below 95% in `rcl_lifecycle` or `rclcpp_lifecycle` I believe.

I could catch a std::exception, which `RclError` is derived from, and log the `e.what()`. That would be about just as general as the `...` though.

I could catch the `RCLErrorBase`, and let the std::exceptions from `throw_from_rcl_error` still crash since they also represent programming errors but won't be thrown during fault injection testing.

clalancette: > `rcl_lifecycle_state_fini` currently only fails if it's passed a bad allocator, which would be a programming error. But if it did fail, it would only leak memory. I could move the RCLCPP_ERROR message to `reset()` and just ignore a bad return value. In that case, it might also be good to make `reset()` `noexcept` as well, so other exceptions thrown from inside that function are also identified and resolved.

Oh, I like this option.  It seems the cleanest to me.  I don't think we'll run into any problems marking `reset()` as `noexcept` either; other than the `throw_from_rcl_error`, I don't think there is anything else in the method that can throw.

brawner:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12279/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7253/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9997/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12174/)

clalancette (approved):
Great, thanks for iterating here.  This change seems good to me with green CI.

rclcpp_lifecycle/src/state.cpp
brawner: Switched to noexcept option

brawner:
Thanks for the review @clalancette and @fujitatomoya

PR #1305, opened 2020-09-11
Destroy msg extracted from LoanedMessage.

### LINKED ISSUE #1304: Extracted msg from LoanedMessage not destroy
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - source
- Version or commit hash:
  - d66cd96f254648de5fa68f6c28a34968563294fb
- DDS implementation:
  - rmw_fastrtps_cpp
  - rclcpp

#### Steps to reproduce issue

```
$ colcon build --cmake-args -DCMAKE_BUILD_TYPE=Debug --packages-up-to rclcpp
$ . install/setup.bash
$ valgrind --leak-check=full ./build/rclcpp/test/test_loaned_message
```

#### Expected behavior
No memory leak

#### Actual behavior

<details><summary> memory leak log </summary>
<p>

```shell

==1676834== 
==1676834== HEAP SUMMARY:
==1676834==     in use at exit: 52,173 bytes in 106 blocks
==1676834==   total heap usage: 20,136 allocs, 20,030 frees, 9,441,487 bytes allocated
==1676834== 
==1676834== 48 bytes in 1 blocks are definitely lost in loss record 1 of 31
==1676834==    at 0x483DE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1676834==    by 0x14DD6A: __gnu_cxx::new_allocator<test_msgs::msg::BasicTypes_<std::allocator<void> > >::allocate(unsigned long, void const*) (new_allocator.h:114)
==1676834==    by 0x14CFD5: rclcpp::LoanedMessage<test_msgs::msg::BasicTypes_<std::allocator<void> >, std::allocator<void> >::LoanedMessage(rclcpp::PublisherBase const&, std::allocator<test_msgs::msg::BasicTypes_<std::allocator<void> > >) (loaned_message.hpp:80)
==1676834==    by 0x14B90C: rclcpp::LoanedMessage<test_msgs::msg::BasicTypes_<std::allocator<void> >, std::allocator<void> >::LoanedMessage(rclcpp::PublisherBase const*, std::shared_ptr<std::allocator<test_msgs::msg::BasicTypes_<std::allocator<void> > > >) (loaned_message.hpp:109)
==1676834==    by 0x14BE61: rclcpp::Publisher<test_msgs::msg::BasicTypes_<std::allocator<void> >, std::allocator<void> >::borrow_loaned_message() (publisher.hpp:176)
==1676834==    by 0x148F76: TestLoanedMessage_release_Test::TestBody() (test_loaned_message.cpp:70)
==1676834==    by 0x19785B: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2447)
==1676834==    by 0x1903BA: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2483)
==1676834==    by 0x16BECD: testing::Test::Run() (gtest.cc:2522)
==1676834==    by 0x16C8C6: testing::TestInfo::Run() (gtest.cc:2703)
==1676834==    by 0x16CFB7: testing::TestCase::Run() (gtest.cc:2825)
==1676834==    by 0x1786A0: testing::internal::UnitTestImpl::RunAllTests() (gtest.cc:5216)
==1676834== 
==1676834== LEAK SUMMARY:
==1676834==    definitely lost: 48 bytes in 1 blocks
==1676834==    indirectly lost: 0 bytes in 0 blocks
==1676834==      possibly lost: 0 bytes in 0 blocks
==1676834==    still reachable: 52,125 bytes in 105 blocks
==1676834==         suppressed: 0 bytes in 0 blocks
==1676834== Reachable blocks (those to which a pointer was found) are not shown.
==1676834== To see them, rerun with: --leak-check=full --show-leak-kinds=all


```

</p>
</details>

iuhilnehc-ynos:
Fixes #1304

Signed-off-by: Chen Lihui <Lihui.Chen@sony.com>

### DIFF
--- a/rclcpp/include/rclcpp/loaned_message.hpp
+++ b/rclcpp/include/rclcpp/loaned_message.hpp
@@ -179,6 +179,20 @@ public:
     return *message_;
   }
 
+  /// Release ownership of the ROS message instance.
+  /**
+   * A call to `release()` will unmanage the memory for the ROS message.
+   * That means that the destructor of this class will not free the memory on scope exit.
+   *
+   * \return Raw pointer to the message instance.
+   */
+  MessageT * release()
+  {
+    auto msg = message_;
+    message_ = nullptr;
+    return msg;
+  }
+
 protected:
   const rclcpp::PublisherBase & pub_;
 
--- a/rclcpp/include/rclcpp/loaned_message.hpp
+++ b/rclcpp/include/rclcpp/loaned_message.hpp
@@ -130,26 +130,10 @@ public:
    */
   virtual ~LoanedMessage()
   {
-    auto error_logger = rclcpp::get_logger("LoanedMessage");
-
     if (message_ == nullptr) {
       return;
     }
-
-    if (pub_.can_loan_messages()) {
-      // return allocated memory to the middleware
-      auto ret =
-        rcl_return_loaned_message_from_publisher(pub_.get_publisher_handle().get(), message_);
-      if (ret != RCL_RET_OK) {
-        RCLCPP_ERROR(
-          error_logger, "rcl_deallocate_loaned_message failed: %s", rcl_get_error_string().str);
-        rcl_reset_error();
-      }
-    } else {
-      // call destructor before deallocating
-      message_->~MessageT();
-      message_allocator_.deallocate(message_, 1);
-    }
+    DestroyMessage(pub_, message_, message_allocator_);
     message_ = nullptr;
   }
 
@@ -184,13 +168,18 @@ public:
    * A call to `release()` will unmanage the memory for the ROS message.
    * That means that the destructor of this class will not free the memory on scope exit.
    *
-   * \return Raw pointer to the message instance.
+   * \return std::unique_ptr to the message instance.
    */
-  MessageT * release()
+  std::unique_ptr<MessageT, std::function<void(MessageT *)>>
+  release()
   {
     auto msg = message_;
     message_ = nullptr;
-    return msg;
+    return std::unique_ptr<MessageT, std::function<void(MessageT *)>>(
+      msg,
+      [pub = pub_, allocator = message_allocator_](MessageT * msg_ptr) {
+        LoanedMessage::DestroyMessage(pub, msg_ptr, allocator);
+      });
   }
 
 protected:
@@ -202,6 +191,29 @@ protected:
 
   /// Deleted copy constructor to preserve memory integrity.
   LoanedMessage(const LoanedMessage<MessageT> & other) = delete;
+
+private:
+  static void DestroyMessage(
+    const rclcpp::PublisherBase & pub, MessageT * msg_ptr,
+    std::allocator<MessageT> allocator)
+  {
+    auto error_logger = rclcpp::get_logger("LoanedMessage");
+
+    if (pub.can_loan_messages()) {
+      // return allocated memory to the middleware
+      auto ret =
+        rcl_return_loaned_message_from_publisher(pub.get_publisher_handle().get(), msg_ptr);
+      if (ret != RCL_RET_OK) {
+        RCLCPP_ERROR(
+          error_logger, "rcl_deallocate_loaned_message failed: %s", rcl_get_error_string().str);
+        rcl_reset_error();
+      }
+    } else {
+      // call destructor before deallocating
+      msg_ptr->~MessageT();
+      allocator.deallocate(msg_ptr, 1);
+    }
+  }
 };
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/loaned_message.hpp
+++ b/rclcpp/include/rclcpp/loaned_message.hpp
@@ -133,7 +133,7 @@ public:
     if (message_ == nullptr) {
       return;
     }
-    DestroyMessage(pub_, message_, message_allocator_);
+    destroy_message(pub_, message_, message_allocator_);
     message_ = nullptr;
   }
 
@@ -178,7 +178,7 @@ public:
     return std::unique_ptr<MessageT, std::function<void(MessageT *)>>(
       msg,
       [pub = pub_, allocator = message_allocator_](MessageT * msg_ptr) {
-        LoanedMessage::DestroyMessage(pub, msg_ptr, allocator);
+        LoanedMessage::destroy_message(pub, msg_ptr, allocator);
       });
   }
 
@@ -193,8 +193,9 @@ protected:
   LoanedMessage(const LoanedMessage<MessageT> & other) = delete;
 
 private:
-  static void DestroyMessage(
-    const rclcpp::PublisherBase & pub, MessageT * msg_ptr,
+  static void destroy_message(
+    const rclcpp::PublisherBase & pub,
+    MessageT * msg_ptr,
     std::allocator<MessageT> allocator)
   {
     auto error_logger = rclcpp::get_logger("LoanedMessage");
--- a/rclcpp/include/rclcpp/loaned_message.hpp
+++ b/rclcpp/include/rclcpp/loaned_message.hpp
@@ -130,10 +130,26 @@ public:
    */
   virtual ~LoanedMessage()
   {
+    auto error_logger = rclcpp::get_logger("LoanedMessage");
+
     if (message_ == nullptr) {
       return;
     }
-    destroy_message(pub_, message_, message_allocator_);
+
+    if (pub_.can_loan_messages()) {
+      // return allocated memory to the middleware
+      auto ret =
+        rcl_return_loaned_message_from_publisher(pub_.get_publisher_handle().get(), message_);
+      if (ret != RCL_RET_OK) {
+        RCLCPP_ERROR(
+          error_logger, "rcl_deallocate_loaned_message failed: %s", rcl_get_error_string().str);
+        rcl_reset_error();
+      }
+    } else {
+      // call destructor before deallocating
+      message_->~MessageT();
+      message_allocator_.deallocate(message_, 1);
+    }
     message_ = nullptr;
   }
 
@@ -167,8 +183,10 @@ public:
   /**
    * A call to `release()` will unmanage the memory for the ROS message.
    * That means that the destructor of this class will not free the memory on scope exit.
-   * The memory is freed when the unique pointer goes out instead, and the unique pointer
-   * will keep the publisher alive until it is destroyed.
+   * If the message is loaned from the middleware but not be published, the user needs to call
+   * `rcl_return_loaned_message_from_publisher` manually.
+   * If the memory is from the local allocator, the memory is freed when the unique pointer
+   * goes out instead, and the unique pointer will keep the publisher alive until it is destroyed.
    *
    * \return std::unique_ptr to the message instance.
    */
@@ -177,10 +195,17 @@ public:
   {
     auto msg = message_;
     message_ = nullptr;
+
+    if (pub_.can_loan_messages()) {
+      return std::unique_ptr<MessageT, std::function<void(MessageT *)>>(msg, [](MessageT *) {});
+    }
+
     return std::unique_ptr<MessageT, std::function<void(MessageT *)>>(
       msg,
-      [pub = pub_, allocator = message_allocator_](MessageT * msg_ptr) {
-        LoanedMessage::destroy_message(pub, msg_ptr, allocator);
+      [ = ](MessageT * msg_ptr) {
+        // call destructor before deallocating
+        msg_ptr->~MessageT();
+        message_allocator_.deallocate(msg_ptr, 1);
       });
   }
 
@@ -193,30 +218,6 @@ protected:
 
   /// Deleted copy constructor to preserve memory integrity.
   LoanedMessage(const LoanedMessage<MessageT> & other) = delete;
-
-private:
-  static void destroy_message(
-    const rclcpp::PublisherBase & pub,
-    MessageT * msg_ptr,
-    std::allocator<MessageT> allocator)
-  {
-    auto error_logger = rclcpp::get_logger("LoanedMessage");
-
-    if (pub.can_loan_messages()) {
-      // return allocated memory to the middleware
-      auto ret =
-        rcl_return_loaned_message_from_publisher(pub.get_publisher_handle().get(), msg_ptr);
-      if (ret != RCL_RET_OK) {
-        RCLCPP_ERROR(
-          error_logger, "rcl_deallocate_loaned_message failed: %s", rcl_get_error_string().str);
-        rcl_reset_error();
-      }
-    } else {
-      // call destructor before deallocating
-      msg_ptr->~MessageT();
-      allocator.deallocate(msg_ptr, 1);
-    }
-  }
 };
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/loaned_message.hpp
+++ b/rclcpp/include/rclcpp/loaned_message.hpp
@@ -167,6 +167,8 @@ public:
   /**
    * A call to `release()` will unmanage the memory for the ROS message.
    * That means that the destructor of this class will not free the memory on scope exit.
+   * The memory is freed when the unique pointer goes out instead, and the unique pointer
+   * will keep the publisher alive until it is destroyed.
    *
    * \return std::unique_ptr to the message instance.
    */
--- a/rclcpp/include/rclcpp/loaned_message.hpp
+++ b/rclcpp/include/rclcpp/loaned_message.hpp
@@ -179,20 +179,6 @@ public:
     return *message_;
   }
 
-  /// Release ownership of the ROS message instance.
-  /**
-   * A call to `release()` will unmanage the memory for the ROS message.
-   * That means that the destructor of this class will not free the memory on scope exit.
-   *
-   * \return Raw pointer to the message instance.
-   */
-  MessageT * release()
-  {
-    auto msg = message_;
-    message_ = nullptr;
-    return msg;
-  }
-
 protected:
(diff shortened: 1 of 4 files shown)


fujitatomoya (approved):
LGTM

rclcpp/test/rclcpp/test_loaned_message.cpp
```diff
@@ -79,5 +79,7 @@ TEST_F(TestLoanedMessage, release) {
 
   ASSERT_EQ(42.0f, msg->float64_value);
 
+  msg->~MessageT();
+  pub->get_allocator()->deallocate(msg, 1);
```
clalancette, line 83: This seems like a weird API.  That is, is this the API a user of the borrow API is supposed to use?  If not, what is a user who is borrowing a message, but then calling `release` on it, supposed to call to free the memory?

iuhilnehc-ynos: The release of LoanedMessage is to `Release ownership of the ROS message instance`.
https://github.com/ros2/rclcpp/blob/0313417f0213f1741465978c91ff18ba4f99ab55/rclcpp/include/rclcpp/loaned_message.hpp#L192
After `release`, the destructor is not going to destroy the message.
https://github.com/ros2/rclcpp/blob/0313417f0213f1741465978c91ff18ba4f99ab55/rclcpp/include/rclcpp/loaned_message.hpp#L135-L137

So user need to do the same way as
https://github.com/ros2/rclcpp/blob/0313417f0213f1741465978c91ff18ba4f99ab55/rclcpp/include/rclcpp/loaned_message.hpp#L149-L151

iuhilnehc-ynos: OK, I will not use this weird API to destroy it.
Update it with
```
ASSERT_EQ(RCL_RET_OK, rcl_return_loaned_message_from_publisher(pub.get_publisher_handle().get(), msg));
```

clalancette: That seems like a reasonable API to add.  However, I'll ping @mjcarroll and @wjwwood who were initially involved with this feature to see what the expected sequence is supposed to be.

iuhilnehc-ynos: Grad to hear that you also think it's a reasonable API. I'll keep my eyes on it.

Karsten1987: thanks for retagging me here. I did not really see this within all GH notifications.

I think I've added that previously to mimic the behavior of a unique_ptr. Looking roughly through the code base, I think it's safest to simply deprecate this function or directly completely remove it. I don't really see a good use-case for it and as shown here might lead to more confusion than benefits.

iuhilnehc-ynos: @wjwwood @Karsten1987

Thanks for your response. I'll remove this method and the test case.

iuhilnehc-ynos: @clalancette 
If all of you agree to add a unique_ptr version of release pointed by @wjwwood, I'll add it later.

wjwwood: > I don't really see a good use-case for it and as shown here might lead to more confusion than benefits.

I tend to agree with this atm. If we really want to keep it, then I think we should capture the destruction in the `unique_ptr` so it is done safely. Otherwise, I think we should just get rid of it.

iuhilnehc-ynos: After reconsideration, I think this API is useful for some advanced users.

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -202,6 +191,29 @@ class LoanedMessage
 
   /// Deleted copy constructor to preserve memory integrity.
   LoanedMessage(const LoanedMessage<MessageT> & other) = delete;
+
+private:
+  static void DestroyMessage(
```
wjwwood, line 196: This should be `destroy_message()` to follow the style of the package.

```suggestion
  static void destroy_message(
```

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -202,6 +191,29 @@ class LoanedMessage
...
+    const rclcpp::PublisherBase & pub, MessageT * msg_ptr,
```
wjwwood, line 197:
```suggestion
    const rclcpp::PublisherBase & pub,
    MessageT * msg_ptr,
```

rclcpp/test/rclcpp/test_loaned_message.cpp
```diff
@@ -77,7 +77,7 @@ TEST_F(TestLoanedMessage, release) {
     // of the data after a call to release.
   }
 
-  ASSERT_EQ(42.0f, msg->float64_value);
+  ASSERT_EQ(42.0f, msg.get()->float64_value);
```
wjwwood, line 80: Is this needed? Does `unique_ptr` not implement `operator->`?

https://en.cppreference.com/w/cpp/memory/unique_ptr/operator*

wjwwood (changes requested):
Looks reasonable to me, with a few small things to fixup.

rclcpp/test/rclcpp/test_loaned_message.cpp
iuhilnehc-ynos: Thank you.
> Is this needed?

No. Updated.

fujitatomoya (approved):
lgtm with @wjwwood 's comments.

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -184,13 +168,18 @@ class LoanedMessage
    * A call to `release()` will unmanage the memory for the ROS message.
    * That means that the destructor of this class will not free the memory on scope exit.
```
wjwwood, line 169: We should update this documentation, and mention that the memory is free'd when the unique pointer goes out instead, and that the unique pointer will keep the publisher alive until it is destroyed.

wjwwood (changes requested):
one more change request, sorry for missing it before

wjwwood (approved):

wjwwood:
Here's CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12758/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7717/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10467/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12700/)

wjwwood:
@iuhilnehc-ynos this is causing a failure in `tf2_ros`, but I don't think it is related to your changes. I reproduce them locally, so I think it either needs a rebase or a potentially some other investigation.

wjwwood:
I tried to rebase for you, but it doesn't apply cleanly. I tried to resolve the merge conflicts, but then rclcpp doesn't build. So I'll pass this back to you @iuhilnehc-ynos for now.

fujitatomoya:
@wjwwood 

we will take this over. thanks for the intel.

iuhilnehc-ynos:
> I tried to rebase for you, but it doesn't apply cleanly. I tried to resolve the merge conflicts, but then rclcpp doesn't build. So I'll pass this back to you @iuhilnehc-ynos for now.

Thank you for trying to do that. I'll resolve the merge conflicts.

iuhilnehc-ynos:
let the middleware clean up the loaned memory?
https://github.com/ros2/rclcpp/blob/31c202e325476f6e912d6d23c17d13393156dbf4/rclcpp/include/rclcpp/publisher.hpp#L262-L264

If we can't change this design, I think the best way to fix this memory leak is only to remove this test case.
To provide a `release_loaned_memory` in publisher.hpp is also not good, because this loaned memory can be freed at two functions. What do you think about this?

iuhilnehc-ynos:
Friendly ping, @wjwwood
https://github.com/ros2/rclcpp/pull/1305#issuecomment-708868358
Do you have any suggestions?

wjwwood:
Sorry, I'm not sure what to do there.

@mjcarroll what do you think?

ivanpauno:
> let the middleware clean up the loaned memory?

IIUC, you either call `publish_loaned_msg` or `return_loaned_msg`, not both (`publish_loaned` takes ownership of the message).
To avoid the problem I think you can do the following:
- If the middleware doesn't support loaned messages, `release` returns an unique_ptr with a deleter that deallocates the message.
- If the middleware does support loaned messages, `release` returns an unique_ptr with a deleter that does nothing.

It doesn't sound nice, but if I'm not miss-understanding the issue that should work.

iuhilnehc-ynos:
@ivanpauno

Thank you for your reply.

> If the middleware doesn't support loaned messages, release returns an unique_ptr with a deleter that deallocates the message.

Agreed.

> If the middleware does support loaned messages, release returns an unique_ptr with a deleter that does nothing.

If `an unique_ptr with a deleter` does nothing, I need to do some special thing for the test case that only calls release no publish.
Actually, I think @wjwwood's policy that `LoanedMessage::destroy_message` works well.

I was once afraid that user can call both `LoanedMessage::release` and `Publisher::publish(rclcpp::LoanedMessage<MessageT, AllocatorT> && loaned_msg)` at the same time if `LoanedMessage::release` is as a public method.

But for now, I think that's user's responsibility, I will continue to rebase these commit and use `loaned_msg.release().get()` instead of `loaned_msg.release()` as `do_loaned_message_publish` parameter.

iuhilnehc-ynos:
@ivanpauno

Thank you.
Besides moving `release` from `public`  to `private`, I can't figure out another better solution.
So I'll follow your suggestion, at least to make this PR alive, and let's hear others' opinions about this.

rclcpp/include/rclcpp/loaned_message.hpp
```diff
@@ -183,14 +183,30 @@ class LoanedMessage
...
+      return std::unique_ptr<MessageT, std::function<void(MessageT *)>>(msg, [](MessageT *) {});
+    }
+
+    return std::unique_ptr<MessageT, std::function<void(MessageT *)>>(
+      msg,
+      [ = ](MessageT * msg_ptr) {
+        // call destructor before deallocating
+        msg_ptr->~MessageT();
+        message_allocator_.deallocate(msg_ptr, 1);
+      });
```
ivanpauno, line 209: I would be more explicit about what is being captured, I'm not sure if you meant to use:

```suggestion
      [this](MessageT * msg_ptr) {
        // call destructor before deallocating
        msg_ptr->~MessageT();
        message_allocator_.deallocate(msg_ptr, 1);
      });
```

or

```suggestion
      [allocator=message_allocator_](MessageT * msg_ptr) {
        // call destructor before deallocating
        msg_ptr->~MessageT();
        allocator.deallocate(msg_ptr, 1);
      });
```

which by the way the difference doesn't matter, but I prefer being more explicit.

ivanpauno (approved):

rclcpp/include/rclcpp/loaned_message.hpp
iuhilnehc-ynos: Updated. :clap:

fujitatomoya:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13447/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8358/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11170/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13501/)

fujitatomoya:
[macOS failure rclpy.test_create_while_spinning.xunit.missing_result](https://ci.ros2.org/job/ci_osx/11170/testReport/junit/(root)/rclpy/test_create_while_spinning_xunit_missing_result/) is unrelated to the fix.

CI with `--packages-up-to rclcpp`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13467/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8377/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11190/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13530/)

fujitatomoya:
sorry, i am trying to figure out CI parameters, this one should be okay.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13470/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8378/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11191/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13531/)

fujitatomoya:
@clalancette 

sorry to bother you...when you got time, could you confirm this is okay to merge just in case? I don't see any failure on https://ci.ros2.org/job/ci_windows/13531/testReport/junit/projectroot.test/rclcpp/test_node/.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -261,7 +261,7 @@ class Publisher : public PublisherBase
     if (this->can_loan_messages()) {
       // we release the ownership from the rclpp::LoanedMessage instance
       // and let the middleware clean up the memory.
-      this->do_loaned_message_publish(loaned_msg.release());
+      this->do_loaned_message_publish(loaned_msg.release().get());
```
ivanpauno, line 264: can we change `do_loaned_message_publish` so it takes a `rclcpp::LoanedMessage` by value and also make `rclcpp::LoanedMessage` a **move only** type? (like `std::unique_ptr`).

IMO, it's more clear that ownership is being transferred in the following snippet:

```suggestion
      this->do_loaned_message_publish(std::move(loaned_msg));
```

The `release` method isn't needed anymore with that API change (it can be deleted or kept though).
IMO, breaking API here makes sense, this is almost not being used and the current API is a bit broken.

ivanpauno: by the way, `rclcpp::LoanedMessage` seems to be already a move only type.

iuhilnehc-ynos: @ivanpauno 

> this->do_loaned_message_publish(std::move(loaned_msg));

I can update with it, but `release` still needs to be called inside `do_loaned_message_publish`.

> The `release` method isn't needed anymore

Sorry, I don't catch it. I thought that `release` is to set a flag to control whether to `return_loaned_message` in `rclcpp` or `rmw`, and we don't know if users will call the publish() method after creating `LoanedMessage`.

ivanpauno: I have confused some things because of miss-reading the current code, sorry!

> can we change do_loaned_message_publish so it takes a rclcpp::LoanedMessage by value and also make rclcpp::LoanedMessage a move only type? (like std::unique_ptr).

This is already the case!
`rclcpp::Publisher::publish(rclcpp::LoanedMessage &&)` takes ownership.

> I think the release method must be put in private section of LoanedMessage, and then use friend class Publisher<MessageT, AllocatorT> in LoanedMessage.

So the comment [here](https://github.com/ros2/rclcpp/pull/1305#issuecomment-708868358) wasn't an actual leak of memory, only a potential leak if the user calls `release` and doesn't publish the message?
IMO, that's fine.
The idea of calling `release()` is "give me the raw ptr and I will make sure to free it correctly", so returning a raw ptr is the correct thing to do.
If the user call release, they must either publish it, or check with `can_loan_message()` the correct destructor to call and do it manually.

wjwwood: >  The idea of calling release() is "give me the raw ptr and I will make sure to free it correctly", so returning a raw ptr is the correct thing to do.

👍

This pull request changes this for `rclcpp::LoanedMessage::release` just because in the case that the message was not loaned the logic to free it properly is a bit complicated (explicitly call destructor and use correct allocator, which the user may have not specified explicitly originally). But they still need to handle the unique_ptr with care, either publishing it or checking if it needs to be returned the middleware, so in that sense it's like a raw pointer half the time.

> If the user call release, they must either publish it, or check with can_loan_message() the correct destructor to call and do it manually.

Yes, but in the case that `can_loan_message()` returns false, they should just let it go out of scope.

wjwwood, line 264: Would it make more sense to change the signature of this method to accept a unique_ptr? I suppose it doesn't matter, just a thought.

wjwwood (approved):
lgtm

rclcpp/include/rclcpp/publisher.hpp
iuhilnehc-ynos: OK, I'll update it and some related test source code.

ivanpauno (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13486/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8393/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11207/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13551/)

fujitatomoya:
[windows unstable](https://ci.ros2.org/job/ci_windows/13551/testReport/projectroot.test/rclcpp/test_node/) has all test pass, i will go ahead to merge this.

Issue #1306, opened 2020-09-11
Not finalize guard condition while destructing SubscriptionIntraProcess


Barry-Xu-2018:
## Bug report

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - source
- Version or commit hash:
  - [d66cd96](https://github.com/ros2/rclcpp/commit/d66cd96f254648de5fa68f6c28a34968563294fb)
- DDS implementation:
  - rmw_fastrtps_cpp
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
$ colcon build --packages-up-to rclcpp --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Debug
$ source install/setup.bash
$ valgrind --leak-check=full ./build/rclcpp/test/test_subscription
```

#### Expected behavior

No memory leak

#### Actual behavior

Memory leak occur
```
==190768== 
==190768== HEAP SUMMARY:
==190768==     in use at exit: 46,900 bytes in 100 blocks
==190768==   total heap usage: 10,538 allocs, 10,438 frees, 4,416,678 bytes allocated
==190768== 
==190768== 144 (56 direct, 88 indirect) bytes in 1 blocks are definitely lost in loss record 24 of 34
==190768==    at 0x483F7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==190768==    by 0x50FBF46: __default_allocate (allocator.c:37)
==190768==    by 0x508B690: __rcl_guard_condition_init_from_rmw_impl (guard_condition.c:73)
==190768==    by 0x508B842: rcl_guard_condition_init (guard_condition.c:108)
[...]

PR #1307, opened 2020-09-11
Not finalize guard condition while destructing SubscriptionIntraProcess

Barry-Xu-2018:


Barry-Xu-2018:
Fix #1306

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -102,6 +102,12 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
 #endif
   }
 
+  ~SubscriptionIntraProcess()
+  {
+    rcl_ret_t ret = rcl_guard_condition_fini(&gc_);
+    (void)ret;
```
fujitatomoya, line 108: I think it would be better to throw exception or at least print error via `RCUTILS_LOG_ERROR_NAMED`, instead of ignoring return code.

```suggestion
    if (RCL_RET_OK != ret) {
      rclcpp::exceptions::throw_from_rcl_error(ret, "failed to finalize guard condition");
    }
```

fujitatomoya, line 108: if we keep the current code, we do not even need `ret`?

```suggestion
    (void) rcl_guard_condition_fini(&gc_);
```

fujitatomoya (changes requested):

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
iuhilnehc-ynos, line 108: Maybe we could use this kind of code.

```suggestion
    if (rcl_guard_condition_fini(&gc_) != RCL_RET_OK) {
      RCUTILS_LOG_ERROR_NAMED(
        "rclcpp",
        "failed to destroy guard condition: %s", rcl_get_error_string().str);
    }
```

iuhilnehc-ynos: I prefer to use `RCLCPP_ERROR` to print error log here, which avoids throwing an exception on destructor.

Barry-Xu-2018: If throw exception，there is warning "note: in C++11 destructors default to noexcept".   
So I will add output error log.

Barry-Xu-2018: Writing as that is to avoid build warning 'not deal with return value of rcl_guard_condition_fini()'.  
Now I change code to output error log which also can avoid that build warning.

rclcpp/include/rclcpp/experimental/subscription_intra_process.hpp
```diff
@@ -102,6 +102,16 @@ class SubscriptionIntraProcess : public SubscriptionIntraProcessBase
 #endif
   }
 
+  ~SubscriptionIntraProcess()
+  {
+    if (rcl_guard_condition_fini(&gc_) != RCL_RET_OK) {
+      RCUTILS_LOG_ERROR_NAMED(
+        "rclcpp",
+        "Failed to destory guard condition: %s",
```
clalancette, line 110:
```suggestion
        "Failed to destroy guard condition: %s",
```

clalancette (changes requested):
One small nit to fix, then I'm happy with this.

Barry-Xu-2018:
@clalancette 

I have fixed spelling error.  
CI report error.  
After checking, the error is from rclcpp_lifecycle.
rclcpp_lifecycle.cppcheck.error: preprocessorErrorDirective (/opt/ros/rolling/include/mimick/mock.h:82)  
I find previous CI build also has this error.

ivanpauno:
@Barry-Xu-2018 the PR checker failure doesn't seem related, but it isn't happening in other jobs.
Can you rebase your branch? maybe that fixes the issue.

Barry-Xu-2018:
@ivanpauno 

Rebase was done.  
All check passes.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12445/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7412/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10155/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12360/)

clalancette:
Looks good, thanks for the ping.

PR #1310, opened 2020-09-11
Replace std_msgs with test_msgs in executors test

jacobperron:
Without this change, I am unable to build locally.
std_msgs is not declared as a test dependency or find_package'd anywhere, so
I'm not sure why CI ever passed the build phase. @brawner any ideas?

### DIFF
--- a/rclcpp/test/CMakeLists.txt
+++ b/rclcpp/test/CMakeLists.txt
@@ -498,7 +498,8 @@ ament_add_gtest(
   TIMEOUT 180)
 if(TARGET test_executors)
   ament_target_dependencies(test_executors
-    "rcl")
+    "rcl"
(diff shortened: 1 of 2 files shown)


brawner (approved):
It looks like this was a transitive dependency with libstatistics_collector and was able to be used through that. Thanks for patching this.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12286/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7259/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10003/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12185/)

PR #1311, opened 2020-09-11
Pass goal handle to goal response callback instead of a future

jacobperron:
This resolves an issue where `promise->set_value` is called before a potential call to `promise->set_exception`.
If there is an issue sending a result request, set the exception on the future to the result in the goal handle, instead of the future to the goal handle itself.

Alternative to https://github.com/ros2/rclcpp/pull/1292.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -15,6 +15,7 @@
 #ifndef RCLCPP_ACTION__CLIENT_HPP_
 #define RCLCPP_ACTION__CLIENT_HPP_
 
+#include <rclcpp/exceptions.hpp>
 #include <rclcpp/macros.hpp>
 #include <rclcpp/node_interfaces/node_base_interface.hpp>
 #include <rclcpp/node_interfaces/node_logging_interface.hpp>
@@ -37,6 +38,7 @@
 #include <utility>
 
 #include "rclcpp_action/client_goal_handle.hpp"
+#include "rclcpp_action/exceptions.hpp"
 #include "rclcpp_action/types.hpp"
 #include "rclcpp_action/visibility_control.hpp"
 
@@ -261,8 +263,7 @@ public:
   using Feedback = typename ActionT::Feedback;
   using GoalHandle = ClientGoalHandle<ActionT>;
   using WrappedResult = typename GoalHandle::WrappedResult;
-  using GoalResponseCallback =
-    std::function<void (std::shared_future<typename GoalHandle::SharedPtr>)>;
+  using GoalResponseCallback = std::function<void (typename GoalHandle::SharedPtr)>;
   using FeedbackCallback = typename GoalHandle::FeedbackCallback;
   using ResultCallback = typename GoalHandle::ResultCallback;
   using CancelRequest = typename ActionT::Impl::CancelGoalService::Request;
@@ -284,12 +285,9 @@ public:
 
     /// Function called when the goal is accepted or rejected.
     /**
-     * Takes a single argument that is a future to a goal handle shared pointer.
+     * Takes a single argument that is a goal handle shared pointer.
      * If the goal is accepted, then the pointer points to a valid goal handle.
      * If the goal is rejected, then pointer has the value `nullptr`.
-     * If an error occurs while waiting for the goal response an exception will be thrown
-     * when calling `future::get()`.
-     * Possible exceptions include `rclcpp::RCLError` and `rclcpp::RCLBadAlloc`.
      */
     GoalResponseCallback goal_response_callback;
 
@@ -360,7 +358,7 @@ public:
         if (!goal_response->accepted) {
           promise->set_value(nullptr);
           if (options.goal_response_callback) {
-            options.goal_response_callback(future);
+            options.goal_response_callback(nullptr);
           }
           return;
         }
@@ -376,16 +374,11 @@ public:
         }
         promise->set_value(goal_handle);
         if (options.goal_response_callback) {
-          options.goal_response_callback(future);
+          options.goal_response_callback(goal_handle);
         }
 
         if (options.result_callback) {
-          try {
-            this->make_result_aware(goal_handle);
-          } catch (...) {
-            promise->set_exception(std::current_exception());
-            return;
-          }
+          this->make_result_aware(goal_handle);
         }
       });
 
@@ -632,22 +625,28 @@ private:
     using GoalResultRequest = typename ActionT::Impl::GetResultService::Request;
     auto goal_result_request = std::make_shared<GoalResultRequest>();
     goal_result_request->goal_id.uuid = goal_handle->get_goal_id();
-    this->send_result_request(
-      std::static_pointer_cast<void>(goal_result_request),
-      [goal_handle, this](std::shared_ptr<void> response) mutable
-      {
-        // Wrap the response in a struct with the fields a user cares about
-        WrappedResult wrapped_result;
-        using GoalResultResponse = typename ActionT::Impl::GetResultService::Response;
-        auto result_response = std::static_pointer_cast<GoalResultResponse>(response);
-        wrapped_result.result = std::make_shared<typename ActionT::Result>();
-        *wrapped_result.result = result_response->result;
-        wrapped_result.goal_id = goal_handle->get_goal_id();
-        wrapped_result.code = static_cast<ResultCode>(result_response->status);
-        goal_handle->set_result(wrapped_result);
-        std::lock_guard<std::mutex> lock(goal_handles_mutex_);
-        goal_handles_.erase(goal_handle->get_goal_id());
-      });
+    try {
+      this->send_result_request(
+        std::static_pointer_cast<void>(goal_result_request),
+        [goal_handle, this](std::shared_ptr<void> response) mutable
+        {
+          // Wrap the response in a struct with the fields a user cares about
+          WrappedResult wrapped_result;
+          using GoalResultResponse = typename ActionT::Impl::GetResultService::Response;
+          auto result_response = std::static_pointer_cast<GoalResultResponse>(response);
+          wrapped_result.result = std::make_shared<typename ActionT::Result>();
+          *wrapped_result.result = result_response->result;
+          wrapped_result.goal_id = goal_handle->get_goal_id();
+          wrapped_result.code = static_cast<ResultCode>(result_response->status);
+          goal_handle->set_result(wrapped_result);
+          std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+          goal_handles_.erase(goal_handle->get_goal_id());
+        });
+    } catch (rclcpp::exceptions::RCLError &) {
+      goal_handle->invalidate();
+      std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+      goal_handles_.erase(goal_handle->get_goal_id());
+    }
   }
 
   /// \internal
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -643,9 +643,8 @@ private:
           goal_handles_.erase(goal_handle->get_goal_id());
         });
     } catch (rclcpp::exceptions::RCLError &) {
+      // This will cause an exception when the user tries to access the result
       goal_handle->invalidate();
-      std::lock_guard<std::mutex> lock(goal_handles_mutex_);
-      goal_handles_.erase(goal_handle->get_goal_id());
     }
   }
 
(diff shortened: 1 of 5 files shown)


brawner (approved):
This seems reasonable to me. I'll let @hidmic have the final approval though.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -632,22 +625,28 @@ class Client : public ClientBase
...
+          wrapped_result.result = std::make_shared<typename ActionT::Result>();
+          *wrapped_result.result = result_response->result;
+          wrapped_result.goal_id = goal_handle->get_goal_id();
+          wrapped_result.code = static_cast<ResultCode>(result_response->status);
+          goal_handle->set_result(wrapped_result);
+          std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+          goal_handles_.erase(goal_handle->get_goal_id());
+        });
+    } catch (rclcpp::exceptions::RCLError &) {
+      goal_handle->invalidate();
```
hidmic, line 646: @jacobperron hmm, my only concern here is that the caller won't know that the result request went wrong till much later. It also means a caller can't cancel the goal just sent.

---

What if we just eat up the exception? It's not ideal, but `ClientGoalHandle::is_result_aware()` can still return `false` and `ClientGoalHandle::async_get_result()` (which will eventually be called anyways) can throw.

brawner:
Here are the builds from #1290. I'm still trying to figure out the Windows issue (if it pops up)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12297/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7269/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10013/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12195/)

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: Perhaps it's not as ergonomic, but the caller can still immediately check the future returned by `Client::async_get_result`, and any error that occurred when sending the result request should be visible.

>  It also means a caller can't cancel the goal just sent.

This is a valid concern. I think what you're proposing is to remove the erasure logic below so that the user can still interact with the goal handle (even though it failed to get the result). I think calling `invalidate()` still makes sense so that calling `Client::async_get_result` still produces an exception.

jacobperron: The other option I hadn't considered was to not handle the exception at all. This way we leave it up to the caller how to handle the exception (and they know something went wrong right away).

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -632,22 +625,28 @@ class Client : public ClientBase
...
+          auto result_response = std::static_pointer_cast<GoalResultResponse>(response);
+          wrapped_result.result = std::make_shared<typename ActionT::Result>();
+          *wrapped_result.result = result_response->result;
+          wrapped_result.goal_id = goal_handle->get_goal_id();
+          wrapped_result.code = static_cast<ResultCode>(result_response->status);
+          goal_handle->set_result(wrapped_result);
+          std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+          goal_handles_.erase(goal_handle->get_goal_id());
+        });
+    } catch (rclcpp::exceptions::RCLError &) {
```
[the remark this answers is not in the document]
jacobperron: Correct me if I'm wrong, but I think this is the only type that can be thrown, [here](https://github.com/ros2/rclcpp/blob/0313417f0213f1741465978c91ff18ba4f99ab55/rclcpp_action/src/client.cpp#L334).

rclcpp_action/include/rclcpp_action/client.hpp
hidmic: Right. If `ClientGoalHandle::make_result_aware` fails, the goal handle will not be result aware. Any attempt to get the future [will result in an exception](https://github.com/ros2/rclcpp/blob/0313417f0213f1741465978c91ff18ba4f99ab55/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L63-L65) that the user can handle.

jacobperron: Take a look at 81239f1

As it is currently implemented, the user will get the [exception set in the future](https://github.com/ros2/rclcpp/blob/0313417f0213f1741465978c91ff18ba4f99ab55/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L145) (because we set result awareness to `true` above, before sending a result request).

But, I think it makes sense to throw the error as earlier if possible, so I've make a small change make this happen: d444e09

hidmic: Do we have to do anything besides making the goal handle not result aware? Would it be wrong if the user attempts to request the result again?

jacobperron:
Note to self: examples, demos, and other packages that register a result callback need to be updated.
I'll spend a little time looking at solutions for maintaining, but deprecated, the old signature.

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: I don't think there's anything else to do :thinking: If a user attempts to request the result I'm not actually sure what would happen, but I guess it shouldn't be allowed considering the result future would have already been set to an exception. In retrospect, I think it keeps the logic simpler to let the exception live in the future (1176dad).
Currently, the user is not able to request the result more than once (even if it is successful).

hidmic: Hmm, I see. We're not dealing with exceptions correctly e.g. depending on where `Client::async_get_result()` throws, the handle may be tagged as result aware (banning future result requests) or not. 

In my mind, a user that explicitly makes a goal handle result aware should get an exception if something goes wrong, and the goal handle should remain result unaware. That is, `Client::async_get_result()` should offer strong exception guarantees. It shouldn't be different for a user that implicitly makes a goal handle result aware by passing in a result callback to `Client::async_send_goal()`. If we don't make the latter throw (and there're valid arguments against), by keeping the goal handle result unaware we allow the user to check `ClientGoalHandle::is_result_aware()` or handle the exception that `ClientGoalHandle::async_get_result()` will throw. Does that make sense?

To invalidate the goal handle improves the situation quite a bit, but it delays error generation (until `future::get`, when it could show itself before spinning), it induces an spurious status change (to UNKNOWN until another status message comes along), and it prevents any retries on an otherwise unaware goal handle. 

But I won't block you. CI's green and this is a fairly unusual case :)

hidmic (approved):

rclcpp_action/include/rclcpp_action/client.hpp
jacobperron: Yeah, that makes sense to me. I've made changes so that in the event that a result request fails, `ClientGoalHandle::is_result_aware()` should return `false` and `Client::async_get_result()` raises an exception (9cb1776).

I couldn't find a nice way to do it with the existing state information, so I've introduced a new member to `ClientGoalHandle` for checking if `invalidate()` was called, which doubles as an improved error message.

PTAL

jacobperron:
Here's of list of downstream changes required by this change:

- https://github.com/ros2/demos/pull/463
- https://github.com/ros2/examples/pull/291
- https://github.com/ros2/geometry2/pull/323
- https://github.com/ros/ros_tutorials/pull/100

I couldn't find a nice way to deprecate the old signature. Suggestions welcome that would make the change easier on users.

hidmic (approved):

rclcpp_action/include/rclcpp_action/client.hpp
hidmic: Neat! Two questions. Could a goal handle be invalidated twice? Should we do something about the spurious UNKNOWN status on invalidation?

jacobperron: Good questions. I think we certainly want to guard against multiple calls invalidate (see b145c73). As an aside, I wish it was easier to test this stuff :upside_down_face: It might be worth revisiting the architecture to make validation easier.

I'm not sure what to do about the UNKNOWN status. We could remove setting it when the goal is invalidated. Really I think the status should be reporting the latest state from the action server. Since this is keeping the same behavior (ie. when the client is destroyed), I'll defer deciding to remove the unknown status to another PR.

jacobperron:
Testing all connected PRs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12331/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7300/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10044/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12235/)

brawner:
CI is all green, is this ready to be merged?

jacobperron:
Err, I may have merged too early. The connected PRs need approvals too :fearful:

ivanpauno:
New build failure in nightlies seem to be related with this: https://ci.ros2.org/view/nightly/job/nightly_linux_release/1681/.

jacobperron:
Added a release note for Galactic: https://github.com/ros2/ros2_documentation/pull/871

v-lopez:
I apologise for coming back to this merged changes, but have you found any way of keeping compatibility between new and old code? Or the only solution right now is a branch for foxy and new developments on a separate branch?

jacobperron:
@v-lopez Unfortunately, the original bug was "easily" fixed by changing API. In this case, I think the API break makes sense since the original API of passing a future was not necessary and part of the bug (see discussion in https://github.com/ros2/rclcpp/pull/1292#discussion_r486579163). 

I think we can easily maintain backwards compatibility if we can use C++17, specifically `std::variant`. E.g. change the type of the goal callback member of the `SendGoalOptions` struct to a `std::variant`. I don't know if we've settled on the minimal C++ standard for Galactic. If we can use C++17, I'm happy to add a change to make this backward compatible.

v-lopez:
In my use cases I've managed to workaround it using the returned shared_future from sending the goal.
My idea was more on the lines of adding the new callback to the foxy version somehow, since it is LTS a lot of packages are going to use it for the next years, and the code won't compile on galactic. But I cannot think of a way of doing it easily right now.

jacobperron:
I don't think adding the new callback to Foxy can be done in an ABI (or API) compatible way. Also, my understanding is the error this bug causes happens rarely (if at all), so I think it's very unlikely that we would backport this change.

jacobperron:
@v-lopez Please correct me if I'm wrong. If you are experiencing errors due to the bug this change fixes, perhaps we could patch it in Foxy a different way.

v-lopez:
No, I've not experienced the bug at all.

Was just looking for a way to make the code compatible Foxy and onwards, luckily there's multiple ways of getting the GoalHandle, we can go with that. Thanks for your efforts.

ivanpauno:
> I think we can easily maintain backwards compatibility if we can use C++17, specifically std::variant. E.g. change the type of the goal callback member of the SendGoalOptions struct to a std::variant. I don't know if we've settled on the minimal C++ standard for Galactic. If we can use C++17, I'm happy to add a change to make this backward compatible.

Though using `std::variant` is nice, for this case we can manually implement an API backwards compatible adapter:

```cpp
  struct SendGoalOptions
  {
    SendGoalOptions()
    : goal_response_callback(nullptr),
      feedback_callback(nullptr),
      result_callback(nullptr)
    {
    }

    /// Function called when the goal is accepted or rejected.
    /**
     * Takes a single argument that is a goal handle shared pointer.
     * If the goal is accepted, then the pointer points to a valid goal handle.
     * If the goal is rejected, then pointer has the value `nullptr`.
     */
    GoalResponseCallbackCompatibilitySelector goal_response_callback;

    /// Function called whenever feedback is received for the goal.
    FeedbackCallback feedback_callback;

    /// Function called when the result for the goal is received.
    ResultCallback result_callback;
  };

class GoalResponseCallbackCompatibilitySelector {
public:
  // implicit constructor
  GoalResponseCallbackCompatibilitySelector(GoalResponseCallbackOldSignature old_callback) : old_callback_(old_callback)
{}

  // implicit constructor
  GoalResponseCallbackCompatibilitySelector(GoalResponseCallbackNewSignature new_callback) : new_callback_(new_callback)
{}

private:
  friend rclcpp_action::Client;  // to avoid exposing getters of new_callback and old_callback publicly

  // Consumer of the class will first try to use new_callback_ and if it is `nullptr`, `get_old_callback` will be used.
  GoalResponseCallbackNewSignature new_callback_;
  GoalResponseCallbackOldSignature old_callback_;
};
```

Considering this change might affect several users, I think it's worth going through a ping-pong deprecation cycle.

jacobperron:
Thanks for the suggestion, @ivanpauno  :bow: 
I agree, we avoid a hard break and go through a deprecation cycle. Would you like to open a PR applying your suggestion? Otherwise, I'm happy to do it.

ivanpauno:
@v-lopez see #1495, which enabled backwards compatibility.

PR #1312, opened 2020-09-11
Use best effort, keep last, history depth 1 QoS Profile for '/clock' subscriptions

ivanpauno:
This PRs proposes changing the default QoS profile for `/clock` subscriptions.
The reasons:
- `history_depth = 10` doesn't seem to make sense, if a newer message was received it seems ok to use it directly.
- reliable communication is typically problematic in high hz problems, like clock is.
  - Repair traffic will increase jitter and likely congest the network
  - If nodes are depending on /clock for their event loops, things are likely to a grinding halt.

I'm not quite sure what would the effect of this change in all users, so this need some discussion before being merged.

### DIFF
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -237,7 +237,7 @@ void TimeSource::create_clock_sub()
   clock_subscription_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(
     node_topics_,
     "/clock",
-    rclcpp::QoS(QoSInitialization::from_rmw(rmw_qos_profile_default)),
+    rclcpp::QoS(KeepLast(1)).reliability(RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT),
     std::bind(&TimeSource::clock_cb, this, std::placeholders::_1)
   );
 }
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -237,7 +237,7 @@ void TimeSource::create_clock_sub()
   clock_subscription_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(
     node_topics_,
     "/clock",
-    rclcpp::QoS(KeepLast(1)).reliability(RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT),
+    rclcpp::QoS(KeepLast(1)).best_effort(),
     std::bind(&TimeSource::clock_cb, this, std::placeholders::_1)
   );
 }


rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -237,7 +237,7 @@ void TimeSource::create_clock_sub()
   clock_subscription_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(
     node_topics_,
     "/clock",
-    rclcpp::QoS(QoSInitialization::from_rmw(rmw_qos_profile_default)),
+    rclcpp::QoS(KeepLast(1)).reliability(RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT),
```
wjwwood, line 240:
```suggestion
    rclcpp::QoS(KeepLast(1)).best_effort(),
```

wjwwood (commented):
Seems reasonable to me, but the assumption that makes it ok is that clock is continuously publishing. We might want to make that the case even when time is paused for instance.

fujitatomoya (commented):
agree on this. i believe that is the use case most likely.

wjwwood:
One other thing to mention, we should probably expose this QoS as an option to the node, e.g.:

https://github.com/ros2/rclcpp/blob/0e0a6a495c452bdf3526a0b7bf457e4186bb344d/rclcpp/include/rclcpp/node_options.hpp#L246-L257

clalancette:
> One other thing to mention, we should probably expose this QoS as an option to the node, e.g.:

Oh, I like that.  That way it is easy for a user to override this if this default turns out not to be what they want.

That being said, I do think that besteffort/keeplast(1) is the correct default for most users, so :+1: to the change from me.

rclcpp/src/rclcpp/time_source.cpp
ivanpauno: thanks!

jacobperron (approved):
LGTM

It'd be nice to add a note to the [Galactic release page](https://index.ros.org/doc/ros2/Releases/Release-Galactic-Geochelone/#changes-since-the-foxy-release).

wjwwood (approved):

ivanpauno:
Remaining PRs:
- [rclpy](https://github.com/ros2/rclpy/pull/619)
- [docs](https://github.com/ros2/ros2_documentation/pull/869)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12335/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7304/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10048/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12244/)

ivanpauno:
Follow-up ticket https://github.com/ros2/rclcpp/issues/1318

Issue #1285, opened 2020-08-26
rclcpp_action server could cause deadlock with its wrapper's mutex


daisukes:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Focal
- Installation type:
  - Foxy release src build 
- Version or commit hash:
  - https://github.com/ros2/ros2/commit/4109e26bf96d0a0a9ef755364dabc851556f3089
- DDS implementation:
  - tried with Fast-RTPS, cyclonedds
- Client library (if applicable):
  - Navigation 2

#### Steps to reproduce issue

The original issue is here and I think this is rclcpp issue
https://github.com/ros-planning/navigation2/issues/1961

```
- launch tb3 simulation and make it active
- repeat sending goal action and waiting for the result in a single loop, and wait for the deadlock 
https://gist.github.com/daisukes/46298b083a48e5db7016f3c0efba7e2d
  - I could reproduce the deadlock with this code usually in a few minutes
```

#### Expected behavior

- does not stop, keep working

#### Actual behavior

- the server will not accept goals when the deadlock happens

#### Problem

- user callbacks are called in the `reentlant_mutex_` lock context
  - in the callbacks, the wrapper will lock its mutex

- in another thread, the wrapper first lock its mutex
  - then calls rclcpp_action method, which tries to lock `reentlant_mutex_`

This inconsistent locking order cause deadlocks. 
So I think rclcpp_action needs to call user callbacks outside of `reentlant_mutex_` lock context.

#### Additional information
Here is gdb information
Thread 9 and 14 are trying to hold two mutex in different order
https://gist.github.com/daisukes/712316a97832f5d9ab851ad47c77ad98

Thread 9: server thread
Frame# 6 [nav2_util::SimpleActionServer<nav2_msgs::action::ComputePathToPose, rclcpp::Node>::handle_goal](https://github.com/ros-planning/navigation2/blob/2e0697c64c265f31bd4ef6c5ed43374841242077/nav2_util/include/nav2_util/simple_action_server.hpp#L82)
Frame# 14 [rclcpp_action::ServerBase::execute_goal_request_received()](https://github.com/ros2/rclcpp/blob/4a5eed968ca6fa4979214db8ac945fb163919326/rclcpp_action/src/server.cpp#L213)

Thread 14: working thread [started here](https://github.com/CMU-cabot/navigation2/blob/db7cea3e5e/nav2_util/include/nav2_util/simple_action_server.hpp#L128)
[...]

PR #1313, opened 2020-09-13
Fix action server deadlock (#1285)

daisukes:
To fix the issue https://github.com/ros2/rclcpp/issues/1285

I think the problem is that the server can let the next goal accept during processing `on_terminal_state` callback.
It can be solved by putting them in a single `reentrant_mutex_` context.

This does not make any deadlock with [my test code](https://gist.github.com/daisukes/46298b083a48e5db7016f3c0efba7e2d) so far.

### DIFF
--- a/rclcpp_action/src/server.cpp
+++ b/rclcpp_action/src/server.cpp
@@ -58,10 +58,10 @@ public:
   size_t num_services_ = 0;
   size_t num_guard_conditions_ = 0;
 
-  bool goal_request_ready_ = false;
-  bool cancel_request_ready_ = false;
-  bool result_request_ready_ = false;
-  bool goal_expired_ = false;
+  std::atomic<bool> goal_request_ready_{false};
+  std::atomic<bool> cancel_request_ready_{false};
+  std::atomic<bool> result_request_ready_{false};
+  std::atomic<bool> goal_expired_{false};
 
   // Lock for unordered_maps
   std::recursive_mutex unordered_map_mutex_;
@@ -174,29 +174,41 @@ ServerBase::add_to_wait_set(rcl_wait_set_t * wait_set)
 bool
 ServerBase::is_ready(rcl_wait_set_t * wait_set)
 {
-  std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
-  rcl_ret_t ret = rcl_action_server_wait_set_get_entities_ready(
-    wait_set,
-    pimpl_->action_server_.get(),
-    &pimpl_->goal_request_ready_,
-    &pimpl_->cancel_request_ready_,
-    &pimpl_->result_request_ready_,
-    &pimpl_->goal_expired_);
+  bool goal_request_ready;
+  bool cancel_request_ready;
+  bool result_request_ready;
+  bool goal_expired;
+  rcl_ret_t ret;
+  {
+    std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
+    ret = rcl_action_server_wait_set_get_entities_ready(
+      wait_set,
+      pimpl_->action_server_.get(),
+      &goal_request_ready,
+      &cancel_request_ready,
+      &result_request_ready,
+      &goal_expired);
+  }
+
+  pimpl_->goal_request_ready_ = goal_request_ready;
+  pimpl_->cancel_request_ready_ = cancel_request_ready;
+  pimpl_->result_request_ready_ = result_request_ready;
+  pimpl_->goal_expired_ = goal_expired;
 
   if (RCL_RET_OK != ret) {
     rclcpp::exceptions::throw_from_rcl_error(ret);
   }
 
-  return pimpl_->goal_request_ready_ ||
-         pimpl_->cancel_request_ready_ ||
-         pimpl_->result_request_ready_ ||
-         pimpl_->goal_expired_;
+  return pimpl_->goal_request_ready_.load() ||
+         pimpl_->cancel_request_ready_.load() ||
+         pimpl_->result_request_ready_.load() ||
+         pimpl_->goal_expired_.load();
 }
 
 std::shared_ptr<void>
 ServerBase::take_data()
 {
-  if (pimpl_->goal_request_ready_) {
+  if (pimpl_->goal_request_ready_.load()) {
     rcl_ret_t ret;
     rcl_action_goal_info_t goal_info = rcl_action_get_zero_initialized_goal_info();
     rmw_request_id_t request_header;
@@ -215,7 +227,7 @@ ServerBase::take_data()
         ret,
         goal_info,
         request_header, message));
-  } else if (pimpl_->cancel_request_ready_) {
+  } else if (pimpl_->cancel_request_ready_.load()) {
     rcl_ret_t ret;
     rmw_request_id_t request_header;
 
@@ -232,7 +244,7 @@ ServerBase::take_data()
       std::make_shared
       <std::tuple<rcl_ret_t, std::shared_ptr<action_msgs::srv::CancelGoal::Request>,
       rmw_request_id_t>>(ret, request, request_header));
-  } else if (pimpl_->result_request_ready_) {
+  } else if (pimpl_->result_request_ready_.load()) {
     rcl_ret_t ret;
     // Get the result request message
     rmw_request_id_t request_header;
@@ -244,7 +256,7 @@ ServerBase::take_data()
     return std::static_pointer_cast<void>(
       std::make_shared<std::tuple<rcl_ret_t, std::shared_ptr<void>, rmw_request_id_t>>(
         ret, result_request, request_header));
-  } else if (pimpl_->goal_expired_) {
+  } else if (pimpl_->goal_expired_.load()) {
     return nullptr;
   } else {
     throw std::runtime_error("Taking data from action server but nothing is ready");
@@ -254,17 +266,17 @@ ServerBase::take_data()
 void
 ServerBase::execute(std::shared_ptr<void> & data)
 {
-  if (!data && !pimpl_->goal_expired_) {
+  if (!data && !pimpl_->goal_expired_.load()) {
     throw std::runtime_error("'data' is empty");
   }
 
-  if (pimpl_->goal_request_ready_) {
+  if (pimpl_->goal_request_ready_.load()) {
     execute_goal_request_received(data);
-  } else if (pimpl_->cancel_request_ready_) {
+  } else if (pimpl_->cancel_request_ready_.load()) {
     execute_cancel_request_received(data);
-  } else if (pimpl_->result_request_ready_) {
+  } else if (pimpl_->result_request_ready_.load()) {
     execute_result_request_received(data);
-  } else if (pimpl_->goal_expired_) {
+  } else if (pimpl_->goal_expired_.load()) {
     execute_check_expired_goals();
   } else {
     throw std::runtime_error("Executing action server but nothing is ready");
@@ -289,7 +301,10 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
   rmw_request_id_t request_header = std::get<2>(*shared_ptr);
   std::shared_ptr<void> message = std::get<3>(*shared_ptr);
 
-  pimpl_->goal_request_ready_ = false;
+  bool expected = true;
+  if (!pimpl_->goal_request_ready_.compare_exchange_strong(expected, false)) {
+    return;
+  }
 
   GoalUUID uuid = get_goal_id_from_goal_request(message.get());
   convert(uuid, &goal_info);
--- a/rclcpp_action/src/server.cpp
+++ b/rclcpp_action/src/server.cpp
@@ -47,6 +47,8 @@ public:
 
   // Lock everything except user callbacks
   std::recursive_mutex action_server_reentrant_mutex_;
+  // Lock to gurantee the call order of user callbacks in execute_goal_request_received
+  std::recursive_mutex goal_request_mutex_;
 
   std::shared_ptr<rcl_action_server_t> action_server_;
 
@@ -286,16 +288,20 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
   rmw_request_id_t request_header = std::get<2>(*shared_ptr);
   std::shared_ptr<void> message = std::get<3>(*shared_ptr);
 
-
-  std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
+  std::lock_guard<std::recursive_mutex> goal_lock(pimpl_->goal_request_mutex_);
+  std::unique_lock<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
 
   pimpl_->goal_request_ready_ = false;
 
   GoalUUID uuid = get_goal_id_from_goal_request(message.get());
   convert(uuid, &goal_info);
 
+  lock.unlock();
+
   // Call user's callback, getting the user's response and a ros message to send back
   auto response_pair = call_handle_goal_callback(uuid, message);
+
+  lock.lock();
 
   ret = rcl_action_send_goal_response(
     pimpl_->action_server_.get(),
@@ -346,6 +352,8 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
     // publish status since a goal's state has changed (was accepted or has begun execution)
     publish_status();
 
+    lock.unlock();
+
     // Tell user to start executing action
     call_goal_accepted_callback(handle, uuid, message);
   }
@@ -359,7 +367,7 @@ ServerBase::execute_cancel_request_received(std::shared_ptr<void> & data)
     <std::tuple<rcl_ret_t, std::shared_ptr<action_msgs::srv::CancelGoal::Request>,
       rmw_request_id_t>>(data);
   auto ret = std::get<0>(*shared_ptr);
-  std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
+  std::unique_lock<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
   if (RCL_RET_ACTION_SERVER_TAKE_FAILED == ret) {
     // Ignore take failure because connext fails if it receives a sample without valid data.
     // This happens when a client shuts down and connext receives a sample saying the client is
@@ -396,6 +404,8 @@ ServerBase::execute_cancel_request_received(std::shared_ptr<void> & data)
     }
   });
 
+  lock.unlock();
+
   auto response = std::make_shared<action_msgs::srv::CancelGoal::Response>();
 
   response->return_code = cancel_response.msg.return_code;
@@ -414,6 +424,8 @@ ServerBase::execute_cancel_request_received(std::shared_ptr<void> & data)
       response->goals_canceling.push_back(cpp_info);
     }
   }
+
+  lock.lock();
 
   // If the user rejects all individual requests to cancel goals,
   // then we consider the top-level cancel request as rejected.
--- a/rclcpp_action/src/server.cpp
+++ b/rclcpp_action/src/server.cpp
@@ -344,10 +344,7 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
 
     if (GoalResponse::ACCEPT_AND_EXECUTE == status) {
       // Change status to executing
-      {
-        std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
-        ret = rcl_action_update_goal_state(handle.get(), GOAL_EVENT_EXECUTE);
-      }
+      ret = rcl_action_update_goal_state(handle.get(), GOAL_EVENT_EXECUTE);
       if (RCL_RET_OK != ret) {
         rclcpp::exceptions::throw_from_rcl_error(ret);
       }
--- a/rclcpp_action/src/server.cpp
+++ b/rclcpp_action/src/server.cpp
@@ -45,7 +45,7 @@ public:
   {
   }
 
-  // Lock everything except user callbacks
+  // Lock for action_server_
   std::recursive_mutex action_server_reentrant_mutex_;
 
   std::shared_ptr<rcl_action_server_t> action_server_;
@@ -62,6 +62,9 @@ public:
   bool cancel_request_ready_ = false;
   bool result_request_ready_ = false;
   bool goal_expired_ = false;
+
+  // Lock for unordered_maps
+  std::recursive_mutex unordered_map_mutex_;
 
   // Results to be kept until the goal expires after reaching a terminal state
   std::unordered_map<GoalUUID, std::shared_ptr<void>> goal_results_;
@@ -338,7 +341,7 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
     *handle = *rcl_handle;
 
     {
-      std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
+      std::lock_guard<std::recursive_mutex> lock(pimpl_->unordered_map_mutex_);
       pimpl_->goal_handles_[uuid] = handle;
     }
 
@@ -453,7 +456,6 @@ ServerBase::execute_result_request_received(std::shared_ptr<void> & data)
   auto shared_ptr = std::static_pointer_cast
     <std::tuple<rcl_ret_t, std::shared_ptr<void>, rmw_request_id_t>>(data);
   auto ret = std::get<0>(*shared_ptr);
-  std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
   if (RCL_RET_ACTION_SERVER_TAKE_FAILED == ret) {
     // Ignore take failure because connext fails if it receives a sample without valid data.
     // This happens when a client shuts down and connext receives a sample saying the client is
@@ -473,7 +475,10 @@ ServerBase::execute_result_request_received(std::shared_ptr<void> & data)
   rcl_action_goal_info_t goal_info;
   convert(uuid, &goal_info);
   bool goal_exists;
-  goal_exists = rcl_action_server_goal_exists(pimpl_->action_server_.get(), &goal_info);
+  {
+    std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
+    goal_exists = rcl_action_server_goal_exists(pimpl_->action_server_.get(), &goal_info);
+  }
   if (!goal_exists) {
     // Goal does not exists
     result_response = create_result_response(action_msgs::msg::GoalStatus::STATUS_UNKNOWN);
@@ -487,6 +492,7 @@ ServerBase::execute_result_request_received(std::shared_ptr<void> & data)
 
   if (result_response) {
     // Send the result now
+    std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
     rcl_ret_t ret = rcl_action_send_result_response(
       pimpl_->action_server_.get(), &request_header, result_response.get());
     if (RCL_RET_OK != ret) {
@@ -494,6 +500,7 @@ ServerBase::execute_result_request_received(std::shared_ptr<void> & data)
     }
   } else {
     // Store the request so it can be responded to later
+    std::lock_guard<std::recursive_mutex> lock(pimpl_->unordered_map_mutex_);
     pimpl_->result_requests_[uuid].push_back(request_header);
   }
   data.reset();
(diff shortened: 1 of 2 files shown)


rclcpp_action/src/server.cpp
```diff
@@ -537,6 +537,11 @@ ServerBase::publish_result(const GoalUUID & uuid, std::shared_ptr<void> result_m
       }
     }
   }
+
+  // Publish a status message any time a goal handle changes state
+  publish_status();
+  // notify base so it can recalculate the expired goal timer
+  notify_goal_terminal_state();
```
fujitatomoya, line 544: I think that the problem is lock order but not locking time window. so I believe with this fix, it still makes the deadlock problem. (in case, right after `rclcpp_action::ServerGoalHandle::succeed` is issued, then in another thread `rclcpp_action::ServerBase::execute_goal_request_received` is fired then calling user callback `handle_goal`)

I think `reentrant_mutex_` must be used internally except user callback.

fujitatomoya (changes requested):

rclcpp_action/src/server.cpp
daisukes: Yeah, I agree with this. This could solve only for my test case.

> I think reentrant_mutex_ must be used internally except user callback.
I changed the fix, is this what you thought?

brawner:
Does this target the correct branch? This bug probably exists on the primary branch `master`, yah? We can backport the bug fix to foxy after merging it on master.

Also, are you able to create a unit test out of your test code and add that to rclcpp_action's tests? I'm about to add a bunch of tests with #1290 and I haven't encountered this issue. It would be awesome to add in a unit test that would prevent this bug from reoccurring once this PR is merged.

daisukes:
My project is based on foxy, so I made a PR for foxy, but yes I think this can be for `master`. 

> Also, are you able to create a unit test out of your test code and add that to rclcpp_action's tests?

Okay, I will try it. My test code is based on Navigation2, but I think I can reproduce its mutex behavior.

daisukes:
@brawner I think I could reproduce the error with a unit test. How do you want me to commit the unit test?

1. works on this branch
2. PR to #1290 branch `jacob/actions_refactor` or `brawner/rclcpp_action-test-coverage`

brawner:
If you need the infrastructure in rclcpp_action-test-coverage than you'll have to target that branch. If however, you can just include the test with this PR then that's ideal because it would make backporting to foxy a lot easier.

daisukes:
I put the test here to prove my patch can fix the issue. I changed the order of commit, to make testing before/after the fix easier.

rclcpp_action/src/server.cpp
fujitatomoya: I think that this works, but i'd like to hear from others.

rclcpp_action/src/server.cpp
```diff
@@ -232,9 +232,13 @@ ServerBase::execute_goal_request_received()
   GoalUUID uuid = get_goal_id_from_goal_request(message.get());
   convert(uuid, &goal_info);
 
+  lock.unlock();
+
   // Call user's callback, getting the user's response and a ros message to send back
   auto response_pair = call_handle_goal_callback(uuid, message);
 
+  lock.lock();
+
```
clalancette, line 241: These type of lock/unlock sequences are often problematic.  It often leads to race conditions where two threads come in here simultaneously, then race to the `lock.lock()`.  Only one will win (and the other will have to wait), but what if T1 comes in first, calls the callback, but then T2 is the one that reacquires the lock first?  Is that situation OK?

Can you explain more about the original deadlock situation?  I'm wondering if there are other solutions we can use here.

daisukes: @clalancette 

> if T1 comes in first, calls the callback, but then T2 is the one that reacquires the lock first? Is that situation OK?

There are 3 `call_**_callback` functions that do not call any `rcl` functions that should be protected by the lock (`pimpl_->reentrant_mutex_`)

- `call_handle_goal_callback`: this uses only local variables
- `call_goal_accepted_callback`: 
- `call_handle_cancel_callback`: 
  - These will access to `goal_handles_`, but also protected by `goal_handles_mutex_`

So, I believe these three functions are threadsafe as long as user callbacks are threadsafe which we don't need to worry about. 

> Can you explain more about the original deadlock situation? I'm wondering if there are other solutions we can use here.

I think explaining with my test code is the best way. The user code has its own mutex which is used in the callback functions and also is used in their working thread which calls the succeed function. 

https://github.com/ros2/rclcpp/blob/938f973c99542a52d413e3e3b8d083acec78d6b7/rclcpp_action/test/test_server.cpp#L994-L1003

- user code: lock user mutex and call succeed (eventually lock rclcpp_action)  L997
- rclcpp_action: lock rclcpp_action mutex and call callbacks (lock user mutex) L1002

fujitatomoya: @clalancette 

I think the most complicated case is the following,

in condition, the goals will be accepted always and simply only two threads.

1. (T-1) `goal_request_ready_` ready, then call `execute_goal_request_received` with `reentrant_mutex_` locked
2. (T-1) unlock `reentrant_mutex_`, then it will call `call_handle_goal_callback` (accepted)
3. (T-2) `goal_request_ready_` ready, then call `execute_goal_request_received` with `reentrant_mutex_` locked
4. (T-2) unlock `reentrant_mutex_`, then it will call `call_handle_goal_callback` (accepted)
5. (T-2) lock `reentrant_mutex_`, call `rcl_action_send_goal_response` and the rest.
6. (T-2) unlock `reentrant_mutex_`.
7. (T-1) lock `reentrant_mutex_`, call `rcl_action_send_goal_response` and the rest.
8. (T-1) unlock `reentrant_mutex_`.

I believe this is the new prblem introduced by this fix, because goal order between acceptance and execution is not consistent. (it is really rare, but possible.)

daisukes: If that order should be guranteed, we can put another mutex which is only used at the beginning of `execute_goal_request_received`.

daisukes: @fujitatomoya
I don't think the potential issue is a critical issue because this only happens when the action server is executed by a multithread executor.

fujitatomoya: yes, obviously that is one of the case. I think the `reentrant_mutex_` is to support exclusive access for `execute_goal_request_received` since it was considered as issue. IMO, it would be better not to change the behavior. for example, if user application expects `first comes, first accepted & executed`, which is really common I guess. this cannot be guaranteed by the fix. I'd like to hear from others.

justinIRBT:
I have an application that is frequently changing controller goals and I was running into this deadlock in nearly every single mission.  I applied the diff in this PR to my server.cpp file locally and haven't been able to reproduce the deadlock.

rclcpp_action/src/server.cpp
daisukes: I added another lock to keep the behavior.

daisukes:
@clalancette @fujitatomoya 
Any updates?

rclcpp_action/src/server.cpp
```diff
@@ -45,6 +45,7 @@ class ServerBaseImpl
 
   // Lock everything except user callbacks
   std::recursive_mutex reentrant_mutex_;
+  std::recursive_mutex goal_request_mutex_;
```
fujitatomoya, line 48: i think this can be private member of `ServerBase`.

fujitatomoya (changes requested):
I do not think of any other way to support current behavior instead of this at this moment. but before moving forward, maybe it would be better to discuss and land on the consensus since this provides additional mutex.

rclcpp_action/src/server.cpp
daisukes: Do you mean both mutexes or only `goal_request_mutex_`?

daisukes:
> it would be better to discuss and land on the consensus since this provides additional mutex.

1. without an additional mutex
  - pros: without additional lock cost
  - cons: behavior change (order of `call_handle_goal_callback`, `call_goal_accepted_callback` is not guaranteed with multi-thread executor)

2. with an additional mutex
  - pros: keep behavior
  - cons: additional lock cost

Which do you prefer?
I would say 2. It is a safer choice at a minimal cost.

clalancette:
@daisukes I just ended up merging #1421, which fixed one missing lock here.  I believe this PR also fixed that bug, but in the interest of keeping things small, I went ahead and merged that one first.  Can you please rebase this PR on top of that?  Thanks.

daisukes:
Hi @clalancette, okay I will rebase them on your change. 
But, before working on that, we need to decide how to deal with @fujitatomoya 's concern, which needs to introduce another lock to guarantee the original behavior. What do you think?

https://github.com/ros2/rclcpp/pull/1313#discussion_r490725563

clalancette:
So I want to back way up here and look at the original problem.

I think the most instructive thing to look at is the gist at https://gist.github.com/daisukes/712316a97832f5d9ab851ad47c77ad98 .  In there, I see the deadlock that caused @daisukes the problem in the first place.  Exactly as he says, there is a lock inversion problem:

* thread 9 takes the `rclcpp_action::action_server_reentrant_mutex`, then calls `SimpleActionServer::handle_goal` (callback), which attempts to take the internal `update_mutex`.
* thread 14 is running `SimpleActionServer::handle_accepted()`, which first takes the internal `update_mutex`, then calls into `Server::call_goal_accepted_callback()`, which takes the `rclcpp_action::action_server_reentrant_mutex`.  This causes the deadlock.

However, looking more closely at that code shows me something else that seems weird.  https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/simple_action_server.hpp#L128 seems to show that a new `std::async` should be executing the `work()` function.  If I'm reading https://en.cppreference.com/w/cpp/thread/async properly, that should mean that a separate thread is used to execute `work()`.  However, looking back at the stack trace for Thread 14, I see that `handle_accepted` is in the call stack.  Now, that may just be a detail of how C++ implements `async`.  But it also implies that the async thread is sharing state with the original thread, including the locked lock.

In other words, I think that implementation in `SimpleActionServer` may be incorrect.  @daisukes if you change the `SimpleActionServer` to use a `std::thread` to kick off work, instead of using the `std::async`, does that solve the problem?

Relatedly, can you give detailed instructions on how to reproduce?  I tried your reproducer program at https://gist.github.com/daisukes/46298b083a48e5db7016f3c0efba7e2d , but so far I haven't been able to reproduce it.

daisukes:
@clalancette 
Thank you for dig into the problem!

Here is the test code to check if the deadlock can happen, which I committed within this PR.
https://github.com/ros2/rclcpp/pull/1313/commits/938f973c99542a52d413e3e3b8d083acec78d6b7

Maybe, my explanation was bad, but basically, it happens with two consecutive goal requests.

daisukes:
Hello @clalancette, I have reread your comment

> However, looking back at the stack trace for Thread 14, I see that handle_accepted is in the call stack. Now, that may just be a detail of how C++ implements async. But it also implies that the async thread is sharing state with the original thread, including the locked lock.
The `async` is called with the `std::launch::async` policy, so the `work()` should be executed in a different thread according to the documentation, indeed it is in Thread 14 which is a different thread from the main node's Thread 9. 
Both threads are accessing an instance of rclcpp_action server and sharing lock state is how the lock works, right?


> Relatedly, can you give detailed instructions on how to reproduce? I tried your reproducer program at https://gist.github.com/daisukes/46298b083a48e5db7016f3c0efba7e2d , but so far I haven't been able to reproduce it.

I remember that I could reproduce the lock with this snippet only with my configuration which has a larger map compared to the nav2 default turtle map. So I wrote the test code which I mentioned in the previous comment.

clalancette:
> > 
> 
> The `async` is called with the `std::launch::async` policy, so the `work()` should be executed in a different thread according to the documentation, indeed it is in Thread 14 which is a different thread from the main node's Thread 9.
> Both threads are accessing an instance of rclcpp_action server and sharing lock state is how the lock works, right?

Yes, but my point is a little different.  If this was truly an independent thread, then it would kick off the new thread, then return from the callback back to the `rclcpp_action` code, at which point it would drop the `reentrant_mutex` in the server.  At that point, the deadlock would no longer exist, and the thread could take the lock (indeed, it may already have been waiting for it).

So my question is: why is that not happening?

daisukes:
> Yes, but my point is a little different. If this was truly an independent thread, then it would kick off the new thread, then return from the callback back to the `rclcpp_action` code, at which point it would drop the `reentrant_mutex` in the server. At that point, the deadlock would no longer exist, and the thread could take the lock (indeed, it may already have been waiting for it).

This deadlock problem is very tricky, as I wrote in the issue #1285 
> #### Problem
> - user callbacks are called in the `reentlant_mutex_` lock context
>   - in the callbacks, the wrapper will lock its mutex
> - in another thread, the wrapper first lock its mutex
>   - then calls rclcpp_action method, which tries to lock `reentlant_mutex_`

The deadlock happening in the situation like

- server thread: accept goal 1, invoke async thread
- async thread: do work, lock wrappter mutex, and about to call `succeeed`
- server thread: accept goal 2 (lock rclcpp mutex) and call callback (trying to lock wrapper mutex)
- async thread: call `succeed` (trying to lock rclcpp mutex)

> So my question is: why is that not happening?

I'm not sure, but from my experience, the higher cpu load cause the more deadlocks.

QQting:
This PR solved my issue. Thanks.

QQting:
Hi @daisukes and @justinIRBT ,

I'm developing an application to use the new feature of Navigation2 [Dynamic Object Following](https://navigation.ros.org/tutorials/docs/navigation2_dynamic_point_following.html).
The GIF below is the screen recording of my application:
pedsim-with-nb2

The publish rate of the goal update is 20Hz, and the RateController in my BT file is also 20Hz. Thanks to this PR, my application no longer gets a deadlock.

However, I encounter a new issue that the bt_action_node of **navigate_to_pose** reports `send_goal failed` due to it returns [rclcpp::executor::FutureReturnCode::TIMEOUT](https://github.com/ros-planning/navigation2/blob/3d166b748bd704eca05ef671b0b2ca99968a8c73/nav2_behavior_tree/include/nav2_behavior_tree/bt_action_node.hpp#L233-L238) after I keep running the program for 2 minutes.

I've tried to increase the [server_timeout](https://github.com/ros-planning/navigation2/blob/3d166b748bd704eca05ef671b0b2ca99968a8c73/nav2_bt_navigator/src/bt_navigator.cpp#L128) from default 10ms to 100ms. 10 minutes later, everything seems to be fine, but the controller server begins yelling `Control loop missed its desired rate of 20.0000Hz`. 

Do you guys encounter the same issues like the server_timeout or missed the control loop rate?

SteveMacenski:
@QQting can you email me a copy of that video (stevenmacenski@gmail.com)? I'd love to use that in our documentation, its hilarious.

fujitatomoya:
@QQting 

> However, I encounter a new issue that the bt_action_node of navigate_to_pose reports send_goal failed due to it returns rclcpp::executor::FutureReturnCode::TIMEOUT after I keep running the program for 2 minutes.

I think that it would be nice to keep this issue with `deadlock` problem instead of mixing them up. How about creating new issue then?

QQting:
> @QQting
> 
> > However, I encounter a new issue that the bt_action_node of navigate_to_pose reports send_goal failed due to it returns rclcpp::executor::FutureReturnCode::TIMEOUT after I keep running the program for 2 minutes.
> 
> I think that it would be nice to keep this issue with `deadlock` problem instead of mixing them up. How about creating new issue then?

I'm not sure to create a new issue because my situation is based on this PR which hasn't been merged yet. And I've also tested the latest rclcpp_action of the main-branch, the deadlock still happens in my application.

fujitatomoya:
okay, so maybe after this PR is merged then 👍

iuhilnehc-ynos:
- Lock everything **except user callbacks**

https://github.com/ros2/rclcpp/blob/dd0f97f1790de0384caeeb94a7c1c49e4b9f08ed/rclcpp_action/src/server.cpp#L46-L47


- Why the user callbacks still need to lock by `action_server_reentrant_mutex_`

https://github.com/ros2/rclcpp/blob/dd0f97f1790de0384caeeb94a7c1c49e4b9f08ed/rclcpp_action/src/server.cpp#L236-L237

https://github.com/ros2/rclcpp/blob/dd0f97f1790de0384caeeb94a7c1c49e4b9f08ed/rclcpp_action/src/server.cpp#L353

https://github.com/ros2/rclcpp/blob/dd0f97f1790de0384caeeb94a7c1c49e4b9f08ed/rclcpp_action/src/server.cpp#L289

Maybe we might just unlock the `action_server_reentrant_mutex_` before calling the above user callbacks and lock it again after user callbacks?

fujitatomoya:
> Maybe we might just unlock the action_server_reentrant_mutex_ before calling the above user callbacks and lock it again after user callbacks?

yes, that @daisukes suggests.

iuhilnehc-ynos:
> yes, that @daisukes suggests.

Thank you. I see

daisukes:
I have updated the changes to be able to be merged into the master branch.

daisukes:
I'm not sure why it failed to build on Jenkins.
It had no error on my local environment, but it failed with [rclcpp_lifecyle](http://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/704/console#console-section-8).

SteveMacenski:
Try retriggering it, might be a transient error from a recent commit https://github.com/ros2/rclcpp/commit/a7db1dcca219b69569e6da63eb6331bfec553cff

daisukes:
@SteveMacenski Thank you for the reminder! It is passed now.

SteveMacenski:
@clalancette what's blocking this now?

clalancette:
> @clalancette what's blocking this now?

I'm still not totally convinced that this is the problem.  I'll have to go back to look at it again, but when I last looked at it, https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/simple_action_server.hpp#L128 should not have been holding the lock, and thus the deadlock should not be happening.  I don't think we have an adequate explanation for why that is the case.

Additionally, doing the `lock`/`unlock` thing is a code smell in my opinion.  Usually that leads to other races/deadlocks later on.  

Overall, I don't think we understand the problem enough yet to put in a solution.

SteveMacenski:
@clalancette would it be worth getting on a call? https://github.com/ros2/rclcpp/issues/1285 this ticket clearly explains the issue. Sure it appears in Nav2's simple action server, but that could be anyone's arbitrary wrapper on `rclcpp_action` and many applications need to lock other resources than just the action object itself as info being fed into it or taken out of it need thread control. 

https://gist.github.com/daisukes/46298b083a48e5db7016f3c0efba7e2d there's a snipped provided by @daisukes showing that this fixes a real problem with frequently running goals and user locks. 

Justin also reported this and I don't think it was in a nav2 context. I don't disagree with you that the `lock`/`unlock` isn't the most elegant thing, but it fixes a real issue that is understood by others. I don't think @daisukes would have gone down this route if the simple action server had something within it to fix (since that's where we started).

clalancette:
All right, I've spent some time and taken another look at this problem.  Thanks for bearing with me while I found the time to do this.

First of all, I spent the time to create a small reproducer.  If you clone https://github.com/clalancette/deadlock, you should be able to build it in a workspace (along with `nav2_msgs` and `nav2_common`), then run:

Terminal 1:
```
$ . install/setup.bash
$ ./install/deadlock/bin/dummy_action_server3
```

Terminal 2:
```
$ . install/setup.bash
$ ./install/deadlock/bin/deadlock
```

This will usually deadlock within a few minutes (at least, on my machine).  Having this minimal reproducer is helpful for looking at the stack traces.

And restating what we've already said above, the deadlock is of the AB/BA type.  Inside of the server, in thread 1 we end up calling `SimpleActionServer::handle_accepted()`, which takes the `update_mutex` in that class.  Later on *in that same thread* we end up calling `ServerBase::notify_goal_terminal_state`, which takes the `action_server_reentrant_mutex` in that class.

Thread 1:
`SimpleActionServer::handle_accepted()` (takes the `update_mutex` lock)
`SimpleActionServer::work()` (this seems like it should be on a separate thread, but that's not what the backtrace shows)
`PlannerServer::computePlan()`
`SimpleActionServer::succeeded_current()`
`ServerGoalHandle::succeed()`
`ServerBase::notify_goal_terminal_state()` (takes the `action_server_reentrant_mutex`)

Thread 2:
`NodeThread::spin()`
`ServerBase::execute()`
`ServerBase::execute_goal_request_received()` (takes the `action_server_reentrant_mutex`)
`ServerBase::call_handle_goal_callback()`
`SimpleActionServer::handle_goal()` (takes the `update_mutex` lock)

So that's our deadlock.  I'm actually going to ignore the part I don't quite understand about `std::async`; it turns out that it doesn't matter since this deadlock can happen in several other places as well.

Now that I've looked at the code in https://github.com/ros2/rclcpp/blob/master/rclcpp_action/src/server.cpp , I actually agree with the direction that this PR is heading in.  The main reason I agree with it is that the `action_server_reentrant_mutex` is actually meant to prevent concurrent access to the `action_server_` itself.  Thus, that mutex should only be held while accessing `action_server_`, and not at other times.

That being said, here are the issues I have with how the PR is implemented:

1. I don't think we should add in the extra `goal_request_mutex`, or at least we shouldn't combine it with this PR.  If that is something we really think we need, let's do it in a separate PR to keep this one focused on the current problem.
2. Let's not lock and unlock the `action_server_reentrant_mutex` by hand.  Instead, let's use scoping to lock it; that just fits in better with how we do locking pretty much everywhere else.
3. Related to the last point, let's make the scope of the locks even smaller.  With the one exception of `publish_status` (where we have to hold the lock across the entire method), we should only need to hold that lock when we are accessing the `pimpl->action_server_`.

@daisukes If you can make those changes, and make sure all of the tests pass, I'm happy to then run CI on it and look at merging it.  Thanks again for your patience here.

daisukes:
@clalancette Thank you for your time to see the problem.
I pushed the changes to address your points.

rclcpp_action/src/server.cpp
```diff
@@ -334,11 +337,17 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
     // Copy out goal handle since action server storage disappears when it is fini'd
     *handle = *rcl_handle;
 
-    pimpl_->goal_handles_[uuid] = handle;
+    {
+      std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
+      pimpl_->goal_handles_[uuid] = handle;
+    }
```
clalancette, line 343: I don't think we need this particular lock here.  At least, this lock is meant to protect the `action_server`, and that is not being accessed here.

rclcpp_action/src/server.cpp
```diff
@@ -334,11 +337,17 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
...
 
     if (GoalResponse::ACCEPT_AND_EXECUTE == status) {
       // Change status to executing
-      ret = rcl_action_update_goal_state(handle.get(), GOAL_EVENT_EXECUTE);
+      {
+        std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
+        ret = rcl_action_update_goal_state(handle.get(), GOAL_EVENT_EXECUTE);
+      }
```
clalancette, line 350: Same comment; I don't think we need to lock here since we aren't accessing `pimpl_->action_server`.

clalancette (dismissed):
A couple of changes inline.

Also, I believe we need to take the mutex in `execute_result_request_received` and in `execute_check_expired_goals` to be totally thread-safe.

rclcpp_action/src/server.cpp
daisukes: https://github.com/ros2/rclcpp/blob/6feedbe117cc163a5b7f7d4d66a3446445ea6b5e/rclcpp_action/src/server.cpp#L526

goal_handles_ can be changed simultaneously, so I put a lock guard here too.

rclcpp_action/src/server.cpp
daisukes: You are right, I will remove this lock guard.

daisukes:
> Also, I believe we need to take the mutex in `execute_result_request_received` and in `execute_check_expired_goals` to be totally thread-safe.

They already have a lock that also protects unordered_map. Do you want me to make their scope smaller too?

https://github.com/ros2/rclcpp/blob/6feedbe117cc163a5b7f7d4d66a3446445ea6b5e/rclcpp_action/src/server.cpp#L453-L459
https://github.com/ros2/rclcpp/blob/6feedbe117cc163a5b7f7d4d66a3446445ea6b5e/rclcpp_action/src/server.cpp#L505-L514

rclcpp_action/src/server.cpp
clalancette: I'm sorry to be a bit pedantic here, but the name of the lock is currently `action_server_reentrant_mutex`.  As a reader of the code, then, I expect that that lock will protect the `action_server`, and not necessarily protect anything else.

But I do agree that the `goal_handles` could be changed by two separate threads.  There are 2 ways we can go with this:
1.  Rename the lock to `pimpl_lock` or something like that.  At that point, it makes more sense that the lock protects the entire PIMPL structure.  But we would also need to audit the rest of the code to make sure that the lock is being held anytime anything in the structure is being modified.
2. Introduce another lock specifically to protect the `goal_handles`, and lock that as appropriate.

I'm honestly not sure which is better.

clalancette:
> They already have a lock that also protects unordered_map. Do you want me to make their scope smaller too?

Oops, yeah, that's right.  Let's make the scope smaller there as well.

rclcpp_action/src/server.cpp
daisukes: I missed this message and have made a change, but I added another mutex for unordered maps including goal_handles, because the pimpl structure has three groups of variables, 1) action_server and its attributes (size_t, bool) captured from actsion_server, 2) three unordered maps, and 3) references (clock and logger).
I think we don't need to lock the entire pimpl structure.

One line I'm not sure about is updating goal_request_ready_ without a lock.
All size_t and bool variables are protected because they are updated with action_server.
https://github.com/ros2/rclcpp/blob/5ae0e9a9929138a05adc40d0bfb448fb5816e819/rclcpp_action/src/server.cpp#L292

clalancette: > One line I'm not sure about is updating goal_request_ready_ without a lock.

Good question.  It *should* be atomic since it is just updating a boolean, but that is not guaranteed.  One thing we could do there is make it a `std::atomic<bool>`, which would guarantee it.

clalancette (approved):
I'm going to go ahead and approve this, with or without the change of `goal_request_ready` to `std::atomic<bool>`.  I'm also going to fire up CI on it and see what happens.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13415/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8335/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11140/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13474/)

rclcpp_action/src/server.cpp
daisukes: Thanks for the suggestion! I changed it to `std::atomic<bool>`.

clalancette:
@daisukes Thanks for the update.  One more CI run, then I'm happy to merge:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13425/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8341/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11146/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13482/)

clalancette:
Looking good to me, so I'm going to merge.  Thanks @daisukes for the contribution and for sticking with it.

daisukes:
@clalancette and @SteveMacenski, thank you for your kind support!

I wanted to include this fix in foxy initially and made a backport for my project. Do you want me to make another PR for foxy as well?
https://github.com/daisukes/rclcpp/tree/foxy-deadlock-fix-backport

While working on that, I found that  these lines were deleted by this commit https://github.com/ros2/rclcpp/commit/8c8c268aad3e243ddb7e2717b5c46044db34e055
https://github.com/ros2/rclcpp/blob/579e9d01d6810e5f60a76f896f440b191514aab5/rclcpp_action/src/server.cpp#L308
https://github.com/ros2/rclcpp/blob/579e9d01d6810e5f60a76f896f440b191514aab5/rclcpp_action/src/server.cpp#L392
I'm not sure these can cause problems.

clalancette:
> I'm not sure these can cause problems.

I'm honestly not sure.  With the tests passing, I'm inclined to say things are working.  But if you find during analysis that you think those should be set, please feel free to open a another PR to restore them.

> I wanted to include this fix in foxy initially and made a backport for my project. Do you want me to make another PR for foxy as well?

If you are interested in having the fix in Foxy, please do.  The good news with this PR is that since everything is PIMPL-style, it should be API and ABI compatible.

SteveMacenski:
> @clalancette and @SteveMacenski, thank you for your kind support!

You did the work, I just checked in on occasion. Thanks for sticking with it :partying_face:!

PR #1314, opened 2020-09-16
Call two iterator overload of vector::erase with end iterator in dynamic_storage.hpp

brawner:
In the case where the predicate `p` is false for all elements or these vectors are empty, these calls to std::remove_if will return the `std::vector::end()`, which is not a valid iterator position to `std::vector::erase(std::iterator)` and in turn an exception will be thrown. However, passing two `end()` iterators to `std::vector::erase` is ok.

Testing this PR, which doesn't include the failing unit test. Those will be added separately (`--packages-select rclcpp`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12330/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7299/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10043/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12233/)

### DIFF
--- a/rclcpp/include/rclcpp/wait_set_policies/dynamic_storage.hpp
+++ b/rclcpp/include/rclcpp/wait_set_policies/dynamic_storage.hpp
@@ -382,11 +382,13 @@ public:
         return weak_ptr.expired();
       };
     // remove guard conditions which have been deleted
-    guard_conditions_.erase(std::remove_if(guard_conditions_.begin(), guard_conditions_.end(), p));
-    timers_.erase(std::remove_if(timers_.begin(), timers_.end(), p));
-    clients_.erase(std::remove_if(clients_.begin(), clients_.end(), p));
-    services_.erase(std::remove_if(services_.begin(), services_.end(), p));
-    waitables_.erase(std::remove_if(waitables_.begin(), waitables_.end(), p));
+    guard_conditions_.erase(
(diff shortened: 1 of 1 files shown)


brawner:
Testing with new unit tests introduced in #1316
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12333/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7302/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10046/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12238/)

wjwwood:
Is this exception documented? I don't see it in the reference:

https://en.cppreference.com/w/cpp/container/vector/erase

The only exception it can throw is from the assignment operator of T, according to cppreference.

wjwwood:
Similarly, the documented exceptions for `std::remove_if` also do not include the exception you describe:

https://en.cppreference.com/w/cpp/algorithm/remove

wjwwood (approved):

wjwwood:
It's probably not an exception though, huh? Instead it's probably a segfault or something like that?

brawner:
Oh you're right. It results in a segfault.

brawner:
Failure for windows in the new unit tests is due to an issue unrelated to this PR, but I believe I've fixed it over at #1316.

PR #1321, opened 2020-09-18
Add coverage tests context functions

Blast545:


rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -78,3 +83,130 @@ TEST(TestUtilities, multi_init) {
...
+  }
+
+  {
+    auto context_to_destroy = std::make_shared<rclcpp::contexts::DefaultContext>();
+    auto mock = mocking_utils::inject_on_return(
+      "lib:rclcpp", rcl_trigger_guard_condition, RCL_RET_ERROR);
+    // This will log a message, no throw expected
+    EXPECT_NO_THROW(
+    {
+      // context_to_destroy.~rclcpp::Context();
```
Blast545, line 209: @brawner I had that part of the code working, but it is not working once I add the expect_no_throw, I think it's similar to the code you had in the wait_policies PR

rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -78,3 +83,130 @@ TEST(TestUtilities, multi_init) {
...
+  EXPECT_EQ(0u, context1->get_on_shutdown_callbacks().size());
+  EXPECT_EQ(std::string{""}, context1->shutdown_reason());
+}
+
+TEST(TestUtilities, test_context_basic_access_const_methods) {
+  auto context1 = std::make_shared<const rclcpp::contexts::DefaultContext>();
+
+  EXPECT_NE(nullptr, context1->get_init_options().get_rcl_init_options());
+  EXPECT_EQ(0u, context1->get_on_shutdown_callbacks().size());
+  // EXPECT_EQ(std::string{""}, context1->shutdown_reason()); not available for const
```
brawner, line 99: This seems like a method that should be const, I wonder if there is a better thread-safe way of setting the string in `rclcpp::Context` so it can be.

rclcpp/test/rclcpp/test_utilities.cpp
brawner: If you want to call the constructor, you can call `reset()` on the shared_ptr (as long as its use count is 1).
```
EXPECT_NO_THROW(context_to_destroy.reset());
```

rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -78,3 +83,130 @@ TEST(TestUtilities, multi_init) {
...
+  interrupt_guard_condition = context1->get_interrupt_guard_condition(&wait_set);
+  EXPECT_NE(nullptr, interrupt_guard_condition);
+  EXPECT_NO_THROW(context1->release_interrupt_guard_condition(&wait_set));
+
+  rclcpp::shutdown(context1);
+}
+
+
+TEST(TestUtilities, test_context_init_shutdown_fails) {
+  auto context = std::make_shared<rclcpp::contexts::DefaultContext>();
```
brawner, line 163: I don't think it makes sense to reuse the same context for these tests since you're calling init/shutdown on it. If one check fails it will probably cause the other ones to fail as well. Create a new context in each scope block below for better insulation.

rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -78,3 +83,130 @@ TEST(TestUtilities, multi_init) {
...
+      context1->release_interrupt_guard_condition(nullptr),
+      std::runtime_error);
+  }
+
+  // Test it works after restore mocks
+  interrupt_guard_condition = context1->get_interrupt_guard_condition(&wait_set);
+  EXPECT_NE(nullptr, interrupt_guard_condition);
+  EXPECT_NO_THROW(context1->release_interrupt_guard_condition(&wait_set));
+
+  rclcpp::shutdown(context1);
```
brawner, line 158: Put this in a SCOPE_EXIT after you call init (if it doesn't make sense to put in `Setup/Teardown`)

rclcpp/test/rclcpp/test_utilities.cpp
Blast545: Thanks! eb70a4b

rclcpp/test/rclcpp/test_utilities.cpp
Blast545: I imagine it is not const because the shutdown_reason may change. But, in terms of this specific test, is odd that the method is not available for a const context. Not sure how to proceed here

rclcpp/test/rclcpp/test_utilities.cpp
Blast545: You are right, changed it with cbeae71

rclcpp/test/rclcpp/test_utilities.cpp
Blast545: osrf_tools are not added to the `rclcpp` tests, it seems. Should I create a class for this particular test?

brawner: No need to recreate. Fortunately rclcpp has it's own SCOPE_EXIT in `rclcpp/scope_exit.hpp`

rclcpp/test/rclcpp/test_utilities.cpp
brawner: It looks like it's const because it has to lock the mutex. It might just be worth opening an issue and getting some input from others. I don't know whether it's worth investigating or if it's not an important use case.

Blast545: I opened #1323 to track the conversation. I didn't tag anyone as it might not be a priority at the moment, let me know if you think I should tag someone in particular.

rclcpp/test/rclcpp/test_utilities.cpp
Blast545: Cool, didn't know about it, see ff35723

rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -20,6 +20,7 @@
 #include "rcl/init.h"
 #include "rcl/logging.h"
 
+#include "rclcpp/scope_exit.hpp"
```
brawner, line 23: alphabetical ordering of headers, pls

brawner (approved):

rclcpp/test/rclcpp/test_utilities.cpp
Blast545: Sure f52f2ae

Blast545:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12382/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7351/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10094/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12295/)

Blast545:

Issue #1254, opened 2020-08-03
allocator_common.hpp overallocates 100x memory & invokes UB on dealloc


stevewolter:
rclcpp/include/rclcpp/allocator/allocator_common.hpp has two severe correctness issues in memory allocation:

 1. retyped_allocate takes the "size" parameter (which is size in bytes as defined by rcl_allocator_t) and passes it to it's allocator's allocate function (which takes number of elements as its parameter). So when asked to allocate N bytes, it actually allocates N*sizeof(T) bytes, where T is the type of the allocator. It's called by subscribers and publishers with a typed allocator of type "Message", which is easily ~100 bytes, so it's overallocating memory by a factor of ~100.
2. retyped_deallocate passes 1 to the allocator's deallocate function, instead of the size of the allocated memory. This happens to work for some allocator implementations that ignore the parameter (libstdc++'s malloc_allocator), but will lead to hard-to-diagnose segfaults on other architectures. I've observed a segfault on Linux + tcmalloc, and I guess this is the reason that allocators needed to be commented out to avoid segfaults on Windows. In any case, it's relying on undefined behaviour and a maintenance hazard.

I realize that both of these issues are not easily fixable. Since allocators are not supported yet anyway (#1061), I'd suggest to disable support for custom allocators entirely and depend on the default RCL allocator. If you concur, I'd be happy to help in implementation.

clalancette:
Thanks for bringing up this issue.  After reading the documentation, I agree with both of your points.  I'd be OK with removing this implementation entirely and just using the default rcl allocator for now instead.  But pinging @ros2/team for other opinions.

wjwwood:
Yeah we can just remove it, I think it should be fixable, but I don't think anyone is really using it and we plan to move to the polymorphic allocator (we think) in the future anyways, so we can fix it then.

clalancette:
Great, thanks for the feedback.  @stevewolter if you are interested in working on this, we'll be happy to review.  Thanks.

stevewolter:
Thanks Chris, I'll give it a shot and send a PR.

stevewolter:
Quick update: It will be some time - I'm on vacation for 2 weeks now, but I'll send a PR afterwards.

clalancette:
@stevewolter Thanks for the heads up.

stevewolter:
[...]

Issue #1061, opened 2020-04-10
none of the API compiles for message types using a custom allocator


dirk-thomas:
E.g. trying to compile a call to `create_publisher` with a message type using a custom allocator fails since the corresponding typesupport symbol isn't specialized for the custom allocator type.

Issue #849, opened 2019-09-10
UBSAN: runtime error publisher_options.hpp:67:62: runtime error: reference binding to null pointer of type 'std::__1::shared_ptr<std::__1::allocator<void> >::element_type' (aka 'std::__1::allocator<void>')


oneattosecond:
## Bug report

**Required Info:**

- Operating System:
  - MacOS 10.14.6
- Installation type:
  - compiled from source
- Version or commit hash:
  - dashing repos file
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp, cartographer, navigation, map_server

#### Steps to reproduce issue
1. Enable UBSAN by adding below lines to various CMakeLists.txt incl. cartographer & navigation2 modules

    set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=undefined")
    set (CMAKE_LINKER_FLAGS_DEBUG "${CMAKE_LINKER_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=undefined")

2. Rebuild all TB3 and related NAV packages with 

     colcon build --cmake-args -DBUILD_TESTING=OFF -DCMAKE_BUILD_TYPE=Debug -DFORCE_DEBUG_BUILD=True --symlink-install

3. Run cartographer and navigation and map server logic as per the usual, notice a bunch of runtime errors (UBSAN warnings) that follow same pattern

Showing a sample from cartographer, but same exact pattern happens in occupancy grid, world model, dwb_controller, navfn_planner, etc. It's either worth fixing or just white noise, I'm not sure so I figured it's worth filing

[cartographer_node-1] ~/ros2_osx_dashing/install/rclcpp/include/rclcpp/publisher_options.hpp:67:62: runtime error: reference binding to null pointer of type 'std::__1::shared_ptr<std::__1::allocator<void> >::element_type' (aka 'std::__1::allocator<void>')
[cartographer_node-1] xcode_path/usr/include/c++/v1/memory:4706:63: runtime error: reference binding to null pointer of type 'std::__1::allocator<void>'
[cartographer_node-1] xcode_path/usr/include/c++/v1/type_traits:2299:12: runtime error: reference binding to null pointer of type 'std::__1::allocator<void>'
[...]

PR #1324, opened 2020-09-22
Remove generation of RCL allocators from C++.

stevewolter:
As explained in #1254, there's conceptually no way to implement RCL
allocators in terms of C++ allocators. Since allocator_common can't
be fixed, delete it and inline calls to rcl_get_default_allocator at
the call sites. Since this would introduce surprising behavior for
custom allocator users (using system malloc when a custom allocator
is requested), add compile-time assertions checking for non-default
allocators. Since there's other blocking bugs for allocator use
(#1061), this shouldn't break existing users.

Technically, we could roll back a lot of templating on allocator and
explicitly use std::allocator<void> in all classes. I have refrained
from doing so because you might have plans with this code, but if you
concur, I'll be more than happy to de-template all these classes to
simplify the code.

I've tried to test this by doing a full ROS compilation following the
Dockerfile of the source Docker image. The tslf_cpp package breaks,
but the code is already sublty broken (copy constructor assumes double
ownership of the memory pool) and can't be used in production, so I'm
not sure what to do about it. Should we delete it?


stevewolter:
Sorry for the noise. This is ready for review now.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
```diff
@@ -39,61 +39,59 @@ void * retyped_allocate(size_t size, void * untyped_allocator)
...
 
-// Convert a std::allocator_traits-formatted Allocator into an rcl allocator
-template<
-  typename T,
-  typename Alloc,
-  typename std::enable_if<!std::is_same<Alloc, std::allocator<void>>::value>::type * = nullptr>
+// Deprecated: Generic converter from C++ allocator into RCL allocator.
+// This allocator overallocates memory by 100x and invokes UB on free,
+// see #1254.
+template<typename Alloc>
```
clalancette, line 70: Honestly, if this thing is so broken, I think we should just remove it.  No reason to keep time-bombs around for others to find.

rclcpp/include/rclcpp/publisher_options.hpp
```diff
@@ -75,7 +75,9 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
     using AllocatorTraits = std::allocator_traits<Allocator>;
     using MessageAllocatorT = typename AllocatorTraits::template rebind_alloc<MessageT>;
     auto message_alloc = std::make_shared<MessageAllocatorT>(*this->get_allocator().get());
-    result.allocator = rclcpp::allocator::get_rcl_allocator<MessageT>(*message_alloc);
+    // TODO(stevewolter): This is likely to invoke undefined behavior - message_alloc goes
+    // out of scope at the end of this function, but the allocator doesn't.
+    result.allocator = get_rcl_allocator(*message_alloc);
```
clalancette, line 80: Can you open an issue about this?  It's a good thing to have it in the code, but it is easy to forget about.

clalancette (changes requested):
I've got a few minor things to cleanup; otherwise looks pretty good to me.  Pinging @wjwwood for any additional thoughts.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
stevewolter: +1, I totally want to remove it. tlsf_cpp and demo_nodes still use this overload, though. My plan is to submit this PR, after which I can add overloads for get_rcl_allocator to tlsf_cpp and demo_nodes, and then drop this overload.

rclcpp/include/rclcpp/publisher_options.hpp
stevewolter: Done, #1339.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
wjwwood: Yeah I think we should remove it. There's other allocator work that needs to be done, and in general we will probably end up dropping all the allocator stuff in favor the polymorphic allocators from C++17 when we can.

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: Actually, I does this hit UB? Since we're doing a copy of `MessageAllocatorT`, it should be fine. According to the expected behavior of allocators, they should be trivially copyable:

- https://en.cppreference.com/w/cpp/named_req/Allocator#Requirements
- https://github.com/ros2/rclcpp/issues/849#issuecomment-539758462

So since we're copying it, the allocator instance in `message_alloc` should be fine to go out of scope.

There's no reason for us to be using shared pointers to allocators, which is something that's already been pointed out but we've yet to change. Instead we should just be copying the allocators around.

stevewolter: Note that we're *not* copying the allocator in the generic overload of get_rcl_allocator, we're taking its address and pass it by reference. There's no way to copy it in the context of get_rcl_allocator, because the state pointer in rcl_allocator_t is a non-owning pointer (because there's no way to specify a destructor in rcl_allocator_t).

I think we can get away with it for now (both allocators in ROS are stateless), but the move to polymorphic allocators will need to resolve the issue.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
stevewolter: Seems we agree. Shall we submit this change and I'll send a follow-up PR to be submitted when dependencies are resolved?

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: Ah you're right, the `void * state` field in the allocator structure assumes essentially global state. For a generic allocator there's no way to deal with this, but if your allocator exposes its global state you could store a pointer to it in the state field and reconstruct the allocator each time you wanted to do that.

In fact that's really what we should be doing rather than storing a reference to the allocator, we should just be capturing the allocator type and creating a new instance each time we need to do an operation, using a lambda or something.

But that all being said, we need to just move away from these template based allocators I think.

----

For this particular issue we could keep the `message_alloc` as a member of the options class and at least that would extend the validity of the `rcl_publisher_options_t result` we're returning to match the lifetime of the publisher options object (which is itself stored in the publisher):

https://github.com/ros2/rclcpp/blob/869f3ed873c0cdd0bd179ce7776676492a386e41/rclcpp/include/rclcpp/publisher.hpp#L373

So that should be safe for the typical use case.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
wjwwood: Generally my preference is to do it all at once (remove this API and update the packages using it), but you can do it in any order that works for you. We just need all tests passing and no compiler warnings for each set of pull requests merged.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
```diff
@@ -29,71 +29,21 @@ namespace allocator
...
-  typename T,
-  typename Alloc,
-  typename std::enable_if<!std::is_same<Alloc, std::allocator<void>>::value>::type * = nullptr>
-rcl_allocator_t get_rcl_allocator(Alloc & allocator)
+// Builds the RCL default allocator for the C++ standard allocator.
+// This assumes that the user intent behind both allocators is the
+// same: Using system malloc for allocation.
+//
+// If you're using a custom allocator in ROS, you'll need to provide
+// your own overload for this function.
```
wjwwood, line 39: If you want doxygen to pick this up you'd need something like this:

```suggestion
/// Return the equivalent rcl_allocator_t for the C++ standard allocator.
/**
 * This assumes that the user intent behind both allocators is the
 * same: Using system malloc for allocation.
 *
 * If you're using a custom allocator in ROS, you'll need to provide
 * your own overload for this function.
 */
```

Also, I made a few verbiage suggestions.

wjwwood (commented):
Changes lgtm, with one comment. I think this approach makes sense.

wjwwood (approved):

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12873/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7828/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10589/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12816/)

stevewolter:
Thanks, and sorry for the delay. This will only build together with its companion changes https://github.com/ros2/demos/pull/467 and https://github.com/ros2/realtime_support/pull/103. Do you want to take a look at them as well, and I'll submit all three simultaneously?

wjwwood:
Yeah we have to do them all at once.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
```diff
@@ -29,71 +29,23 @@ namespace allocator
...
+/// Return the equivalent rcl_allocator_t for the C++ standard allocator.
+/**
+ * This assumes that the user intent behind both allocators is the
+ * same: Using system malloc for allocation.
+ *
+ * If you're using a custom allocator in ROS, you'll need to provide
+ * your own overload for this function.
+ */
+template<typename T>
+rcl_allocator_t get_rcl_allocator(std::allocator<T> allocator)
```
wjwwood, line 43: This needs to be in the `rclcpp::allocator` namespace, so that it matches the folder structure `include/rclcpp/allocator/...`.

wjwwood (changes requested):

fujitatomoya:
@stevewolter can you rebase this? this should fix https://github.com/ros2/rclcpp/issues/1925.

stevewolter:
@fujitatomoya We've downprioritized our ROS efforts, so it will be while until I find time, sorry. Please feel free to clone the change & submit it under your name if you need it.

wjwwood:
@fujitatomoya @stevewolter I think I can spend some time to revive this since it came up in one of my projects recently.

tylerjw:
I'd again like to use ASAN in CI for libraries I use that depend on rclcpp. Is there anything I can do to help finish this?

As far as I can tell, this works, and I have tested it locally, but it seems there is some confusion around the namespace this should be in. Could I get some help understanding what changes are blocking this so maybe I could try to update it?

clalancette:
> I'd again like to use ASAN in CI for libraries I use that depend on rclcpp. Is there anything I can do to help finish this?

Yes!  If you had time to do it, it would be wonderful to rebase this onto the latest and get it compiling and working (it probably needs a few fixes before that happens).

As far as the namespace fixes, really we just need to move `get_rcl_allocator` back into the `rclcpp::allocator` namespace; that should be enough.

One other thing to note is that both https://github.com/ros2/realtime_support/pull/103 and https://github.com/ros2/demos/pull/467 may need to be updated along with this to make CI pass.

Thanks for taking a look, it is much appreciated!

rclcpp/include/rclcpp/allocator/allocator_common.hpp
```diff
@@ -29,71 +29,23 @@ namespace allocator
...
-// Convert a std::allocator_traits-formatted Allocator into an rcl allocator
-template<
-  typename T,
-  typename Alloc,
-  typename std::enable_if<!std::is_same<Alloc, std::allocator<void>>::value>::type * = nullptr>
-rcl_allocator_t get_rcl_allocator(Alloc & allocator)
+/// Return the equivalent rcl_allocator_t for the C++ standard allocator.
+/**
+ * This assumes that the user intent behind both allocators is the
+ * same: Using system malloc for allocation.
```
cwecht, line 37: AFAIK this is not exactly true: std::allocator is implemented in terms of new/delete. Although new/delete are often implemented in terms of malloc/free, it is not demanded by the C++ Standard (AFAIK, might be wrong), meaning that implementations are free to implement new/delete independently from malloc/free.

tylerjw:
@clalancette I pushed my attempt at fixing this PR up here: https://github.com/ros2/rclcpp/pull/2046

jmachowinski:
I am going to close this PR in favor for https://github.com/ros2/rclcpp/pull/3058 which will be merged in the near future.

PR #1328, opened 2020-09-23
Avoid timers to be executed twice in the multithreaded executor

ivanpauno:
~Fixes https://github.com/ros2/rclcpp/issues/1374 by adding a mutex in `Timer` class, that makes `is_ready` and `execute` methods mutually exclusive.~

Fixes https://github.com/ros2/rclcpp/issues/1487.

It deletes the old "scheduled timers" mechanism, that has a long history of being error prone.

### DIFF
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -19,6 +19,7 @@
 #include <chrono>
 #include <functional>
 #include <memory>
+#include <mutex>
 #include <sstream>
 #include <thread>
 #include <type_traits>
@@ -198,6 +199,15 @@ public:
   void
   execute_callback() override
   {
+    // TODO(ivanpauno): Delete this workaround.
+    // The issue is that the multithreaded executor might schedule the timer
+    // more than once in different threads.
+    // Making `execute_callback` mutually exclusive and rechecking if the timer is ready
+    // avoids the issue.
+    std::lock_guard<std::mutex> lock(execution_mutex_);
+    if (!is_ready()) {
+      return;
+    }
     rcl_ret_t ret = rcl_timer_call(timer_handle_.get());
     if (ret == RCL_RET_TIMER_CANCELED) {
       return;
@@ -247,6 +257,7 @@ protected:
   RCLCPP_DISABLE_COPY(GenericTimer)
 
   FunctorT callback_;
+  std::mutex execution_mutex_;
 };
 
 template<
--- a/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
@@ -85,8 +85,6 @@ private:
   size_t number_of_threads_;
   bool yield_before_execute_;
   std::chrono::nanoseconds next_exec_timeout_;
-
-  std::set<TimerBase::SharedPtr> scheduled_timers_;
 };
 
 }  // namespace executors
--- a/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
@@ -83,18 +83,6 @@ MultiThreadedExecutor::run(size_t)
       if (!get_next_executable(any_exec, next_exec_timeout_)) {
         continue;
       }
-      if (any_exec.timer) {
-        // Guard against multiple threads getting the same timer.
-        if (scheduled_timers_.count(any_exec.timer) != 0) {
-          // Make sure that any_exec's callback group is reset before
-          // the lock is released.
-          if (any_exec.callback_group) {
-            any_exec.callback_group->can_be_taken_from().store(true);
-          }
-          continue;
-        }
-        scheduled_timers_.insert(any_exec.timer);
-      }
     }
     if (yield_before_execute_) {
       std::this_thread::yield();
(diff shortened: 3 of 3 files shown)


rclcpp/include/rclcpp/timer.hpp
```diff
@@ -198,6 +199,15 @@ class GenericTimer : public TimerBase
   void
   execute_callback() override
   {
+    // TODO(ivanpauno): Delete this workaround.
+    // The issue is that the multithreaded executor might schedule the timer
+    // more than once in different threads.
+    // Making `execute_callback` mutually exclusive and rechecking if the timer is ready
+    // avoids the issue.
+    std::lock_guard<std::mutex> lock(execution_mutex_);
```
fujitatomoya, line 207: I think, with this fix there will be high possibility that multiple threads get the same timer, be exclusive on this lock. This would be a problem that if the user timer callback takes time. in that case, other threads just stops here and cannot to continue to get any executables.

ivanpauno: mmm, two timers being executed in parallel is an uncommon event, so contention will not happen frequently.
But yeah, maybe this is not great.

I'm still confused about https://github.com/ros2/rclcpp/issues/1374, the test checking that a timer is not executed twice is sometimes failing and I don't understand well why.

fujitatomoya: > mmm, two timers being executed in parallel is an uncommon event

agree,

> the test checking that a timer is not executed twice is sometimes failing and I don't understand well why.

time is not executed twice, i think that the jitter because of lock contention.

if there is a lot of threads, it is hard to get lock https://github.com/ros2/rclcpp/blob/5851eebdda5cc2ed9186a034df0ef75ccfe032d3/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L106 because everyone else is just trying to loop to get lock . (this also consumes cpu resource.) https://github.com/ros2/rclcpp/blob/5851eebdda5cc2ed9186a034df0ef75ccfe032d3/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L79 I was thinking that we might solve this problem if we can provide higher priority to get lock https://github.com/ros2/rclcpp/blob/5851eebdda5cc2ed9186a034df0ef75ccfe032d3/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L106 to call https://github.com/ros2/rclcpp/blob/5851eebdda5cc2ed9186a034df0ef75ccfe032d3/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L109

ivanpauno: Ah, I think I understand: we're seeing two timer callbacks being executed close because one was late and lose a period, but not because it was scheduled twice, right?
I have to recheck how the test works, but my description of https://github.com/ros2/rclcpp/issues/1374 is definitely wrong.

fujitatomoya: > we're seeing two timer callbacks being executed close because one was late and lose a period, but not because it was scheduled twice, right?

yes, i do think so.

ivanpauno:
Round of CI testing everything above rclcpp:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13428/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8343/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11148/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13485/)

wjwwood (commented):
But this breaks the Timer so that it cannot be run concurrently in a reentrant callback group with a multi-threaded executor. I think we're just trading one broken thing for another.

ivanpauno:
> But this breaks the Timer so that it cannot be run concurrently in a reentrant callback group with a multi-threaded executor. I think we're just trading one broken thing for another.

I wouldn't consider that a bug, at most I would consider it a non optimal approach to the problem. Though really, I'm not sure if allowing to execute timer callbacks reentrantly is a good idea at all, but maybe it is (?).

On the other hand, #1487 is definitely buggy behavior.

But yes this PR isn't really "the right fix" as you can actually block another worker thread that took the same timer (ideally, if we would consider that reentrant execution of the same executable is not a good idea, a worker thread would never take the same executable than another and thus it wouldn't be blocked), but I don't see an easy solution here without major surgery of the executor code (and I'm not sure what that would be).

fujitatomoya:
> But yes this PR isn't really "the right fix" as you can actually block another worker thread that took the same timer

we cannot expect how long user callback takes. if user callback takes 5 seconds and if other threads are scheduled to execute the same timer (I think this is probable), these threads will be blocked for 5 seconds just waiting on the `GenericTImer::execution_mutex_`. in this case, executor cannot dispatch and process the other executables. (related to https://github.com/ros2/rclcpp/pull/1328/files#r499131414)

wjwwood:
> in this case, executor cannot dispatch and process the other executables

This is exactly the reason I came up with callback groups originally, because otherwise, when using a multi-threaded executor, the user would always have to synchronize their callbacks and that would lead to most threads in the pool just waiting on one of those mutexs, even though other things could be done.

ivanpauno:
Alternatively https://github.com/ros2/rclcpp/pull/1516 also fixes the issue.

ivanpauno:
Closing in favor of https://github.com/ros2/rclcpp/pull/1516

PR #1331, opened 2020-09-24
Add in tests for the static executor

clalancette:
This PR adds in tests for the `StaticSingleThreadedExecutor` and a helper class, `StaticExecutorEntitiesCollector`.  With this PR in place, I see 97% line coverage on static_single_threaded_executor.cpp and 97% line coverage on static_executor_entities_collector.cpp.

Along the way, I found a few bugs and other problems that I've fixed:
1.  There were some typos in comments.
1.  There were a bunch of places where we could add `const` to variables to give the compiler a bit more of a hint.
1.  There were a couple of places where we were calling the copy constructor when passing a map into a method.
1.  There was a bug when calculating whether to add a guard condition after adding a new node.

Note that this is still in draft because I want to run CI on it (particularly for Windows).  I'll pull it out of draft once CI comes back clean.

### DIFF
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -211,7 +211,7 @@ StaticExecutorEntitiesCollector::prepare_wait_set()
 
   if (RCL_RET_OK != ret) {
     throw std::runtime_error(
-            std::string("Couldn't resize the wait set : ") + rcl_get_error_string().str);
+            std::string("Couldn't resize the wait set: ") + rcl_get_error_string().str);
   }
 }
 
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -28,7 +28,7 @@ using rclcpp::executors::StaticExecutorEntitiesCollector;
 
 StaticExecutorEntitiesCollector::~StaticExecutorEntitiesCollector()
 {
-  // Disassocate all callback groups and thus nodes.
+  // Disassociate all callback groups and thus nodes.
   for (auto & pair : weak_groups_associated_with_executor_to_nodes_) {
     auto group = pair.first.lock();
     if (group) {
@@ -218,7 +218,7 @@ StaticExecutorEntitiesCollector::prepare_wait_set()
 void
 StaticExecutorEntitiesCollector::refresh_wait_set(std::chrono::nanoseconds timeout)
 {
-  // clear wait set (memeset to '0' all wait_set_ entities
+  // clear wait set (memset to '0' all wait_set_ entities
   // but keeps the wait_set_ number of entities)
   if (rcl_wait_set_clear(p_wait_set_) != RCL_RET_OK) {
     throw std::runtime_error("Couldn't clear wait set");
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -297,6 +297,8 @@ StaticExecutorEntitiesCollector::add_callback_group(
   if (has_executor.exchange(true)) {
     throw std::runtime_error("Callback group has already been added to an executor.");
   }
+  bool is_new_node = !has_node(node_ptr, weak_groups_associated_with_executor_to_nodes_) &&
+    !has_node(node_ptr, weak_groups_to_nodes_associated_with_executor_);
   rclcpp::CallbackGroup::WeakPtr weak_group_ptr = group_ptr;
   auto insert_info = weak_groups_to_nodes.insert(
     std::make_pair(weak_group_ptr, node_ptr));
@@ -304,8 +306,6 @@ StaticExecutorEntitiesCollector::add_callback_group(
   if (!was_inserted) {
     throw std::runtime_error("Callback group was already added to executor.");
   }
-  bool is_new_node = !has_node(node_ptr, weak_groups_associated_with_executor_to_nodes_) &&
-    !has_node(node_ptr, weak_groups_to_nodes_associated_with_executor_);
   if (is_new_node) {
     rclcpp::node_interfaces::NodeBaseInterface::WeakPtr node_weak_ptr(node_ptr);
     weak_nodes_to_guard_conditions_[node_weak_ptr] = node_ptr->get_notify_guard_condition();
--- a/rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
+++ b/rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
@@ -14,12 +14,17 @@
 
 #include <gtest/gtest.h>
 
+#include <chrono>
 #include <memory>
+#include <stdexcept>
 
 #include "rclcpp/rclcpp.hpp"
 
 #include "test_msgs/msg/empty.hpp"
 #include "test_msgs/srv/empty.hpp"
+
+#include "../../mocking_utils/patch.hpp"
+#include "../../utils/rclcpp_gtest_macros.hpp"
 
 namespace
 {
@@ -105,7 +110,9 @@ TEST_F(TestStaticExecutorEntitiesCollector, add_remove_node) {
   EXPECT_NO_THROW(entities_collector_->add_node(node1->get_node_base_interface()));
 
   // Check adding second time
-  EXPECT_THROW(entities_collector_->add_node(node1->get_node_base_interface()), std::runtime_error);
+  RCLCPP_EXPECT_THROW_EQ(
+    entities_collector_->add_node(node1->get_node_base_interface()),
+    std::runtime_error("Node has already been added to an executor."));
 
   auto node2 = std::make_shared<rclcpp::Node>("node2", "ns");
   EXPECT_FALSE(entities_collector_->remove_node(node2->get_node_base_interface()));
@@ -114,6 +121,10 @@ TEST_F(TestStaticExecutorEntitiesCollector, add_remove_node) {
   EXPECT_TRUE(entities_collector_->remove_node(node1->get_node_base_interface()));
   EXPECT_FALSE(entities_collector_->remove_node(node1->get_node_base_interface()));
   EXPECT_TRUE(entities_collector_->remove_node(node2->get_node_base_interface()));
+
+  auto node3 = std::make_shared<rclcpp::Node>("node3", "ns");
+  node3->get_node_base_interface()->get_associated_with_executor_atomic().exchange(true);
+  EXPECT_FALSE(entities_collector_->remove_node(node3->get_node_base_interface()));
 }
 
 TEST_F(TestStaticExecutorEntitiesCollector, init_bad_arguments) {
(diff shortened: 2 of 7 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12452/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7419/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12368/)

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
```diff
@@ -293,3 +304,319 @@ TEST_F(TestStaticExecutorEntitiesCollector, add_remove_node_with_entities) {
   // Still one for the executor
   EXPECT_EQ(1u, entities_collector_->get_number_of_waitables());
 }
+
+TEST_F(TestStaticExecutorEntitiesCollector, add_callback_group) {
+  auto node = std::make_shared<rclcpp::Node>("node1", "ns");
+  rclcpp::CallbackGroup::SharedPtr cb_group = node->create_callback_group(
```
brawner, line 310: A bit of a nitpick here, but I think using `auto` would let you put these two lines onto one, which would help readability a small bit with all these new tests. Likewise, I would recommend to avoid abbreviations like 'cb_group' if possible (either `group` or `callback_group`). Fix here and below

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
```diff
@@ -293,3 +304,319 @@ TEST_F(TestStaticExecutorEntitiesCollector, add_remove_node_with_entities) {
...
+      entities_collector_->execute(),
+      std::runtime_error("Couldn't resize the wait set : error not set"));
+  }
+
+  EXPECT_TRUE(entities_collector_->remove_node(node->get_node_base_interface()));
+  entities_collector_->init(&wait_set, memory_strategy, &rcl_guard_condition);
+}
+
+TEST_F(TestStaticExecutorEntitiesCollector, refresh_wait_set_not_initialized) {
+  RCLCPP_EXPECT_THROW_EQ(
```
brawner, line 402: There are probably a couple too many line breaks here, maybe put the command on one line and the exception on the other?

```
RCLCPP_EXPECT_THROW_EQ(
      entities_collector_->refresh_wait_set(std::chrono::nanoseconds(1000)), 
      std::runtime_error("rcl_wait() failed: error not set"));
```

Same thing below

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
```diff
@@ -293,3 +304,319 @@ TEST_F(TestStaticExecutorEntitiesCollector, add_remove_node_with_entities) {
...
+}
+
+TEST_F(TestStaticExecutorEntitiesCollector, refresh_wait_set_add_handles_to_wait_set_failed) {
+  auto node = std::make_shared<rclcpp::Node>("node", "ns");
+
+  // Create 1 of each entity type
+  auto subscription =
+    node->create_subscription<test_msgs::msg::Empty>(
+    "topic", rclcpp::QoS(10), [](test_msgs::msg::Empty::SharedPtr) {});
+  auto timer =
```
brawner, line 447: Nit: was this line break intentional?

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
```diff
@@ -293,3 +304,319 @@ TEST_F(TestStaticExecutorEntitiesCollector, add_remove_node_with_entities) {
...
+  rclcpp::GuardCondition guard_condition(shared_context);
+  rcl_guard_condition_t rcl_guard_condition = guard_condition.get_rcl_guard_condition();
+
+  entities_collector_->init(&wait_set, memory_strategy, &rcl_guard_condition);
+
+  {
+    auto mock = mocking_utils::patch_and_return("lib:rclcpp", rcl_wait_set_resize, RCL_RET_ERROR);
+    RCLCPP_EXPECT_THROW_EQ(
+      entities_collector_->execute(),
+      std::runtime_error("Couldn't resize the wait set : error not set"));
```
brawner, line 394: Should it be?

```suggestion
      std::runtime_error("Couldn't resize the wait set: error not set"));
```

rclcpp/test/rclcpp/executors/test_static_single_threaded_executor.cpp
```diff
@@ -47,3 +54,104 @@ TEST_F(TestStaticSingleThreadedExecutor, check_unimplemented) {
...
+TEST_F(TestStaticSingleThreadedExecutor, add_callback_group_trigger_guard_failed) {
+  rclcpp::executors::StaticSingleThreadedExecutor executor;
+  auto node = std::make_shared<rclcpp::Node>("node", "ns");
+  rclcpp::CallbackGroup::SharedPtr cb_group = node->create_callback_group(
+    rclcpp::CallbackGroupType::MutuallyExclusive);
+
+  {
+    auto mock = mocking_utils::patch_and_return(
+      "lib:rclcpp", rcl_trigger_guard_condition,
+      RCL_RET_ERROR);
```
brawner, line 67: I don't think you need to split up the parameters
```
  "lib:rclcpp", rcl_trigger_guard_condition, RCL_RET_ERROR);
```
Here and below

rclcpp/test/rclcpp/executors/test_static_single_threaded_executor.cpp
```diff
@@ -47,3 +54,104 @@ TEST_F(TestStaticSingleThreadedExecutor, check_unimplemented) {
...
+  auto service =
+    node->create_service<test_msgs::srv::Empty>(
+    "service",
+    [](
+      const test_msgs::srv::Empty::Request::SharedPtr,
+      test_msgs::srv::Empty::Response::SharedPtr) {});
+  auto client = node->create_client<test_msgs::srv::Empty>("service");
+
+  std::promise<void> promise;
+  std::future<void> fut = promise.get_future();
```
brawner, line 150:
```suggestion
  std::future<void> future = promise.get_future();
```

brawner (commented):
Most of my comments are just due to some formatting that seems a bit different then I've seen elsewhere in `rclcpp`. Also, I'm curious whether this makes sense as several PRs rather than just one larger one.

clalancette:
> Most of my comments are just due to some formatting that seems a bit different then I've seen elsewhere in `rclcpp`.

I honestly just write code in my own style, and then do what uncrustify tells me to make it pass the linters.  It is often the case that several things would satisfy uncrustify, but I just always take the one it tells me to do.

> Also, I'm curious whether this makes sense as several PRs rather than just one larger one.

Yeah, I debated about that.  I decided to put it into a single PR for a few reasons:
1.  These changes did come up while writing the tests here, so they are mostly related.
1.  My plan was to "Rebase and Merge" this set so that they are individual commits.
1.  Expediency.

If you'd still like me to split it up, how many more PRs would you suggest?  I can see arguments for 1, 2, 3, 4, or 5 :).

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
```diff
@@ -28,23 +28,23 @@ using rclcpp::executors::StaticExecutorEntitiesCollector;
 
 StaticExecutorEntitiesCollector::~StaticExecutorEntitiesCollector()
 {
-  // Disassocate all callback groups and thus nodes.
-  for (auto & pair : weak_groups_associated_with_executor_to_nodes_) {
+  // Disassociate all callback groups and thus nodes.
+  for (auto const & pair : weak_groups_associated_with_executor_to_nodes_) {
```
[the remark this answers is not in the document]
clalancette: Yeah, I agree.  This particular file mostly used `auto const`, which is why I went with that.  But I'm happy to change all of them to more what the rest of the codebase looks like.

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
clalancette: Would that actually, work, though?  If I do that, then the node pointer becomes a temporary, and then wouldn't it be destroyed immediately after this line (since there are no more handles to it)?

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
clalancette: Not with the code that is currently in `static_executor_entities_collector.cpp`.  But since I haven't been shy about making other changes, I suppose I shouldn't be shy here either.  I'll make the change to both the code and the tests.

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
clalancette: Just doing what uncrustify told me to :).

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
clalancette: Same here.

rclcpp/test/rclcpp/executors/test_static_single_threaded_executor.cpp
clalancette: Yeah, that's reasonable.  Will fix.

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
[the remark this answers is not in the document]
brawner: It looks like this file is an anomaly within rclcpp. I'd switch them to const auto

brawner:
> how many more PRs would you suggest?

I waffled myself on how many would make sense. I think separating out the API changes would be good and then you can tag some reviewers that might have more insight into the intended use of these classes than myself. Otherwise, it's just a matter of backportability. 

When adding coverage our general guiding principle has been that bugs should be back portable to whatever ros distro's make sense and tests should be backportable to foxy. However, I was under the impression you were adding coverage for stuff mostly added only in rolling, so separating out into more PRs might not provide much benefit.

clalancette:
All right, I've rebased and left only the changes that don't affect API here.  I'll open a separate PR for the API changes presently.  I'm going to mark this as an in-progress PR now, and run another round of CI here.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12478/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7445/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10191/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12397/)

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
clalancette: We discussed this offline.  @brawner was suggesting that I combine this line with the next one by using `auto`.  My personal preference is to not use `auto`, so I think we'll just leave this here for now.  @brawner Let me know if you disagree.

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
clalancette: I cleaned this up a bit.

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
clalancette: Not sure about this one.  uncrustify definitely told me to make this change, though you are right that it can probably be collapsed to one line.

brawner: Strange, uncrustify is happy with this on my machine, 
```
auto timer = node->create_wall_timer(std::chrono::seconds(60), []() {});
```

rclcpp/test/rclcpp/executors/test_static_executor_entities_collector.cpp
brawner: Disagreements about the awesomeness and power of auto in a unit test aside, I'm fine leaving these as is.

brawner (approved):

clalancette:
Thanks for the reviews!

Issue #1010, opened 2020-03-02
[rclcpp_components] Make component manager public


Karsten1987:
TL;DR: We've discussed about having the ros2_control's controller manager inherit from the `rclcpp_components` component manager as we would like to have similar functionality.

More info here: https://github.com/ros-controls/roadmap/pull/7

In order to inherit though, the component manager header has to be modified to be a public header with `virtual` functions. Is there a reason why this is currently not the case?

fyi @mjcarroll

mjcarroll:
No, there isn't a strong reason other than it was not necessary to support a public API for it at the time.

Since no one is currently using this as an interface, I'm not quite sure what the ABI implications are, but it would seem reasonable to do for Foxy?

Karsten1987:
yeah, Foxy would be great so we can work with Foxy for the ros control release.

mjcarroll:
Are there additions to the API that will also be required in order to support the concept of a "strongly-typed component"? Or have you not quite made it far enough to determine yet?

Karsten1987:
I haven't really thought that through in all detail, but essentially we'd like to restrict the type of nodes the `ControllerManager` can load. 

From what I understand, the `NodeFactory` is already templated in a way that can be used to restrict to load components different from `rclcpp::Node`. 
In terms of API change, I believe it would be great if we could e.g. things like this https://github.com/ros2/rclcpp/blob/master/rclcpp_components/src/component_manager.cpp#L65-L66 turn into a function argument with default value. That way, we don't necessarily have to override that particular function.

mjcarroll:
Ping @Karsten1987 , have you determined if there is anything else you need API-wise?  Otherwise I will proceed with just exposing it as part of the public API and we can go from there.

Karsten1987:
I think we should be okay with the changes mentioned in my comment above. The essential requirements we would have to make it work for ros-control is to be able to subclass from the component manager and modify its logic to restrict loading class instances other than `rclcpp::Node`.

PR #1333, opened 2020-09-25
ComponentManager: switch off parameter services and event publisher

mbuijs:
So far, the component manager has no parameters anyway, so no need for
services to configure them or topics to publish events.

An attempt to limit the number of services / topics.
(https://discourse.ros.org/t/ros2-default-behavior-wifi/13460/35)

Signed-off-by: Martijn Buijs <martijn.buijs@gmail.com>

### DIFF
--- a/rclcpp_components/src/component_container.cpp
+++ b/rclcpp_components/src/component_container.cpp
@@ -22,8 +22,15 @@ int main(int argc, char * argv[])
 {
   /// Component container with a single-threaded executor.
   rclcpp::init(argc, argv);
+
+  rclcpp::NodeOptions options;
+  options.start_parameter_services(false);
+  options.start_parameter_event_publisher(false);
+
   auto exec = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
-  auto node = std::make_shared<rclcpp_components::ComponentManager>(exec);
+  auto node = std::make_shared<rclcpp_components::ComponentManager>(
+    exec, "ComponentManager",
+    options);
   exec->add_node(node);
   exec->spin();
 }
--- a/rclcpp_components/src/component_container.cpp
+++ b/rclcpp_components/src/component_container.cpp
@@ -22,15 +22,8 @@ int main(int argc, char * argv[])
 {
   /// Component container with a single-threaded executor.
   rclcpp::init(argc, argv);
-
-  rclcpp::NodeOptions options;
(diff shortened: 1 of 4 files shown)


rclcpp_components/src/component_container.cpp
```diff
@@ -22,8 +22,15 @@ int main(int argc, char * argv[])
 {
   /// Component container with a single-threaded executor.
   rclcpp::init(argc, argv);
+
+  rclcpp::NodeOptions options;
+  options.start_parameter_services(false);
+  options.start_parameter_event_publisher(false);
```
fujitatomoya, line 28: although this makes sense, I would adjust these default options with constructor `ComponentManager::ComponentManager` since these options seem reasonable to be disabled and reduce redundant code.

mbuijs: That was my original change, however while searching for owners of this code, I found a pull request to make `ComponentManager::ComponentManager` public, so that other packages it can be reused. If any derived class wants to introduce parameters it would be impossible to start using parameter services or the parameter event publisher unless this code is changed again.

References:
- https://github.com/ros2/rclcpp/issues/1010
- https://github.com/ros-controls/roadmap/pull/7

mbuijs: Never mind my reply above, shortly after I typed it, I realized what you meant with 'default options'. Now it's in the default argument of the constructor.

ivanpauno (approved):
Makes sense to me

ivanpauno:
I think that having an argument to enable/disable the parameter services or the parameter events might make sense too.
Similar on how we have an argument to enable/disable [rosout](https://github.com/ros2/rcl/blob/84cac5937b303e0dca3c9c1b8ffc8b73a2bd0197/rcl/include/rcl/arguments.h#L51).

wjwwood (approved):
lgtm, but it would be nice to mention this in the doc block too.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12634/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7595/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10339/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12563/)

PR #1330, opened 2020-09-24
Add coverage tests graph_listener

Blast545:
Checking with CI coverage, I think I have a coverage issue with my local environment at the moment:
* Linux coverage [Build Status](http://ci.ros2.org/job/ci_linux_coverage/207/)

PR #1336, opened 2020-09-25
Add executor unit tests

clalancette:
This PR adds in unit tests for the Executor class.  With this change in place, I'm seeing 91% line coverage of the executor.cpp file.

Besides a handful of single lines that aren't being covered, there is a big chunk of missing coverage in the `execute_subscription` method.  I think this is going to be a lot of work to cover, so I figured I would open what I have so far and we can iterate on that later.

### DIFF
--- a/rclcpp/test/mocking_utils/patch.hpp
+++ b/rclcpp/test/mocking_utils/patch.hpp
@@ -253,7 +253,7 @@ struct PatchTraits<ID, void(ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5)>
     mock_type, void, ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5);
 };
 
-/// Traits specialization for ReturnT(ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5)
+/// Traits specialization for ReturnT(ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5, ArgT6)
 /// free functions.
 /**
  * \tparam ID Numerical identifier of the patch. Ought to be unique.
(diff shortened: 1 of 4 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12487/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7454/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10200/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12406/)

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -0,0 +1,450 @@
...
+#include <string>
+
+#include "rclcpp/executor.hpp"
+#include "rclcpp/memory_strategy.hpp"
+#include "rclcpp/strategies/allocator_memory_strategy.hpp"
+
+#include "../mocking_utils/patch.hpp"
+#include "../utils/rclcpp_gtest_macros.hpp"
+
+class DummyExecutor : public rclcpp::Executor
```
brawner, line 29: You didn't want to add these tests to test_executors.cpp?

brawner (commented):
Looks good to me, just a couple of questions

rclcpp/test/mocking_utils/patch.hpp
```diff
@@ -270,6 +270,62 @@ struct PatchTraits<ID, ReturnT(ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5, ArgT6)>
...
+ * \tparam ID Numerical identifier of the patch. Ought to be unique.
+ * \tparam ReturnT Return value type.
+ * \tparam ArgTx Argument types.
+ */
+template<size_t ID, typename ReturnT,
+  typename ArgT0, typename ArgT1,
+  typename ArgT2, typename ArgT3,
+  typename ArgT4, typename ArgT5,
+  typename ArgT6, typename ArgT7>
+struct PatchTraits<ID, ReturnT(ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5, ArgT6, ArgT7)>
```
Blast545, line 285: FYI: I added those in #1330, one of us will have to rebase it

Blast545 (approved):
LGTM

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -0,0 +1,450 @@
...
+  DummyExecutor dummy;
+  auto node = std::make_shared<rclcpp::Node>("node", "ns");
+  bool timer_fired = false;
+  auto timer =
+    node->create_wall_timer(std::chrono::milliseconds(1), [&timer_fired]() {timer_fired = true;});
+
+  // Wait for the wall timer to have expired.
+  std::this_thread::sleep_for(std::chrono::milliseconds(50));
+  EXPECT_FALSE(timer_fired);
+  dummy.spin_nanoseconds(node->get_node_base_interface());
```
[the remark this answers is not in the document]
clalancette: My understanding is that `spin_once` collects all of the entities that could possibly be executed, and then picks one to execute from a fixed-priority list: https://github.com/ros2/rclcpp/blob/148d295416dba3814162465392e56116195c2304/rclcpp/src/rclcpp/executor.cpp#L488.

Since timers are always the first thing in the priority list, then the only way we could not execute this is if some other entity added a timer to the executor before this one.  But I don't think that can currently happen.

There is a lot of interest in removing that fixed-priority scheme and make it something else (FIFO, or whatever).  If that happens, then there is a small possibility that this test won't work anymore, but I think it is pretty unlikely.

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+TEST_F(TestExecutor, spin_node_once_nanoseconds) {
+  DummyExecutor dummy;
+  auto node = std::make_shared<rclcpp::Node>("node", "ns");
+  bool timer_fired = false;
+  auto timer =
+    node->create_wall_timer(std::chrono::milliseconds(1), [&timer_fired]() {timer_fired = true;});
+
+  // Wait for the wall timer to have expired.
+  std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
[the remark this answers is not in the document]
clalancette: I don't think that 2ms is going to be enough.  10ms is *probably* enough, but I really didn't want to introduce a flaky test here (particularly on Windows, where timers are really jittery).

Just as a datapoint, this entire file full of tests takes 842ms to run on my machine.  If I were to change all of these 50ms sleeps to 10ms, then I'd save ~210ms.  Its a good chunk, but I think the resistance to flakiness is more important.

rclcpp/test/rclcpp/test_executor.cpp
clalancette: Hmpf.  Well, the whole executor/executors naming thing in this repository somewhat confuses me.  I wanted to make this file match the `.cpp` file it was testing.  Also, adding it to `test_executors.cpp` would make that file really long.  But I actually don't feel strongly either way, so if you think they should be combined, I can do that.

rclcpp/test/mocking_utils/patch.hpp
clalancette: Good to know, thanks for the heads up.

rclcpp/test/rclcpp/test_executor.cpp
brawner: The main thing about adding it to test_executors.cpp is that those tests are parameterized against all executors. But these tests you're adding are against a generic executor and testing the base class code. That would be the biggest difference. 

Since you're not using a specific executor here and instead just the base class code, then I think it's fine to leave it as is. I was mainly just curious. I do think you should add a disclaimer comment at the top describing this file is only intended to test the base class and everything else should be considered for specific executor test files or the parameterized test file.

clalancette: Sounds good, I'll go ahead and add the disclaimer.

clalancette: (I did this in the latest, where I also rebased)

rclcpp/test/rclcpp/test_executor.cpp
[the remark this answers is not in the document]
brawner: Very true, and 210ms is peanuts all told. I just thought it was not great practice to depend on sleeps for behavior tests, but I also couldn't think of a way to assert the proper behavior. FWIW on my machine, these tests all pass even removing this sleep.

clalancette:
One more CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12499/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7467/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10211/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12419/)

brawner (approved):

PR #1338, opened 2020-09-28
Add spin_some support to the StaticSingleThreadedExecutor

mauropasse:
Signed-off-by: Mauro <mpasserino@irobot.com>

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -544,7 +544,7 @@ protected:
   RCLCPP_DISABLE_COPY(Executor)
 
   RCLCPP_PUBLIC
-  void
+  virtual void
   spin_once_impl(std::chrono::nanoseconds timeout);
 
   typedef std::map<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr,
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -544,7 +544,7 @@ protected:
   RCLCPP_DISABLE_COPY(Executor)
 
   RCLCPP_PUBLIC
-  virtual void
+  void
   spin_once_impl(std::chrono::nanoseconds timeout);
 
   typedef std::map<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr,
--- a/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
+++ b/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
@@ -69,6 +69,10 @@ public:
 
   /// Finalize StaticExecutorEntitiesCollector to clear resources
   RCLCPP_PUBLIC
+  bool
+  is_init() {return initialized_;}
+
+  RCLCPP_PUBLIC
   void
   fini();
 
@@ -339,6 +343,9 @@ private:
 
   /// Executable list: timers, subscribers, clients, services and waitables
   rclcpp::experimental::ExecutableList exec_list_;
+
+  /// Bool to check if the entities collector has been initialized
+  bool initialized_ = false;
 };
 
 }  // namespace executors
--- a/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
@@ -15,6 +15,7 @@
 #ifndef RCLCPP__EXECUTORS__STATIC_SINGLE_THREADED_EXECUTOR_HPP_
 #define RCLCPP__EXECUTORS__STATIC_SINGLE_THREADED_EXECUTOR_HPP_
 
+#include <chrono>
 #include <cassert>
 #include <cstdlib>
 #include <memory>
@@ -77,6 +78,42 @@ public:
   RCLCPP_PUBLIC
   void
   spin() override;
+
+  /// Static executor implementation of spin some
+  /**
+   * This non-blocking function will execute entities that
+   * were ready when this API was called, until timeout or no
+   * more work available. Entities that got ready while
+   * executing work, won't be taken into account here.
+   *
+   * Example:
+   *   while(condition) {
+   *     spin_some();
+   *     sleep(); // User should have some sync work or
+   *              // sleep to avoid a 100% CPU usage
+   *   }
+   */
+  RCLCPP_PUBLIC
+  void
+  spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0)) override;
+
+  /// Static executor implementation of spin all
+  /**
+   * This non-blocking function will execute entities until
+   * timeout or no more work available. If new entities get ready
+   * while executing work available, they will be executed
+   * as long as the timeout hasn't expired.
+   *
+   * Example:
+   *   while(condition) {
+   *     spin_all();
+   *     sleep(); // User should have some sync work or
+   *              // sleep to avoid a 100% CPU usage
+   *   }
+   */
+  RCLCPP_PUBLIC
+  void
+  spin_all(std::chrono::nanoseconds max_duration) override;
 
   /// Add a callback group to an executor.
   /**
@@ -224,27 +261,6 @@ public:
   /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
   RCLCPP_PUBLIC
   void
-  spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0)) override
-  {
-    (void)max_duration;
-    throw rclcpp::exceptions::UnimplementedError(
-            "spin_some is not implemented for StaticSingleThreadedExecutor, use spin or "
-            "spin_until_future_complete");
-  }
-
-  /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
-  RCLCPP_PUBLIC
-  void
-  spin_all(std::chrono::nanoseconds) override
-  {
-    throw rclcpp::exceptions::UnimplementedError(
-            "spin_all is not implemented for StaticSingleThreadedExecutor, use spin or "
-            "spin_until_future_complete");
-  }
-
-  /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
-  RCLCPP_PUBLIC
-  void
   spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1)) override
   {
     (void)timeout;
@@ -254,14 +270,22 @@ public:
   }
 
 protected:
-  /// Check which executables in ExecutableList struct are ready from wait_set and execute them.
   /**
-   * \param[in] exec_list Structure that can hold subscriptionbases, timerbases, etc
-   * \param[in] timeout Optional timeout parameter.
+   * @brief Executes ready executables from wait set.
+   * @param spin_once if true executes only the first ready executable.
+   * @return true if any executable was ready.
    */
   RCLCPP_PUBLIC
+  bool
+  execute_ready_executables(bool spin_once = false);
+
+  RCLCPP_PUBLIC
   void
-  execute_ready_executables();
+  spin_some_impl(std::chrono::nanoseconds max_duration, bool exhaustive);
+
+  RCLCPP_PUBLIC
+  void
+  spin_once_impl(std::chrono::nanoseconds timeout) override;
 
 private:
   RCLCPP_DISABLE_COPY(StaticSingleThreadedExecutor)
--- a/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
@@ -261,13 +261,7 @@ public:
   /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
   RCLCPP_PUBLIC
   void
-  spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1)) override
-  {
-    (void)timeout;
-    throw rclcpp::exceptions::UnimplementedError(
-            "spin_once is not implemented for StaticSingleThreadedExecutor, use spin or "
-            "spin_until_future_complete");
-  }
+  spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1)) override;
 
 protected:
   /**
@@ -282,10 +276,6 @@ protected:
   RCLCPP_PUBLIC
   void
   spin_some_impl(std::chrono::nanoseconds max_duration, bool exhaustive);
-
-  RCLCPP_PUBLIC
-  void
-  spin_once_impl(std::chrono::nanoseconds timeout) override;
 
 private:
   RCLCPP_DISABLE_COPY(StaticSingleThreadedExecutor)
--- a/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
@@ -192,77 +192,6 @@ public:
   std::vector<rclcpp::CallbackGroup::WeakPtr>
   get_automatically_added_callback_groups_from_nodes() override;
 
-  /// Spin (blocking) until the future is complete, it times out waiting, or rclcpp is interrupted.
-  /**
-   * \param[in] future The future to wait on. If this function returns SUCCESS, the future can be
-   *   accessed without blocking (though it may still throw an exception).
-   * \param[in] timeout Optional timeout parameter, which gets passed to
-   *    Executor::execute_ready_executables.
-   *   `-1` is block forever, `0` is non-blocking.
-   *   If the time spent inside the blocking loop exceeds this timeout, return a TIMEOUT return
-   *   code.
-   * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
-   *
-   *  Example usage:
-   *  rclcpp::executors::StaticSingleThreadedExecutor exec;
-   *  // ... other part of code like creating node
-   *  // define future
-   *  exec.add_node(node);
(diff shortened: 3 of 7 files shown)


mauropasse:
https://github.com/ros2/rclcpp/issues/1219

mauropasse:
The failing test does not fail when run locally:
```
root@IRBT-8424l:~/ros2_cc_ws_x86_master_eq/ros2_sdk/build/rclcpp/test# ./test_publisher_subscription_count_api
Running main() from /root/ros2_cc_ws_x86_master_eq/ros2_sdk/install/src/gtest_vendor/src/gtest_main.cc
[==========] Running 4 tests from 4 test cases.                                 
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_intraprocess_comm, where TypeParam = TwoSubscriptionsIntraprocessComm
[ RUN      ] TestPublisherSubscriptionCount/two_subscriptions_intraprocess_comm.increasing_and_decreasing_counts
[       OK ] TestPublisherSubscriptionCount/two_subscriptions_intraprocess_comm.increasing_and_decreasing_counts (8029 ms)
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_intraprocess_comm (8029 ms total)
                                                                                
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_one_intraprocess_one_not, where TypeParam = TwoSubscriptionsOneIntraprocessOneNot
[ RUN      ] TestPublisherSubscriptionCount/two_subscriptions_one_intraprocess_one_not.increasing_and_decreasing_counts
[       OK ] TestPublisherSubscriptionCount/two_subscriptions_one_intraprocess_one_not.increasing_and_decreasing_counts (8024 ms)
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_one_intraprocess_one_not (8024 ms total)
                                                                                
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_with_intraprocess_comm, where TypeParam = TwoSubscriptionsInTwoContextsWithIntraprocessComm
[ RUN      ] TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_with_intraprocess_comm.increasing_and_decreasing_counts
[WARN] [1601478189.805770540] [rclcpp]: logging was initialized more than once  
[       OK ] TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_with_intraprocess_comm.increasing_and_decreasing_counts (8038 ms)
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_with_intraprocess_comm (8038 ms total)
                                                                                
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_without_intraprocess_comm, where TypeParam = TwoSubscriptionsInTwoContextsWithoutIntraprocessComm
[ RUN      ] TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_without_intraprocess_comm.increasing_and_decreasing_counts
[WARN] [1601478197.845648792] [rclcpp]: logging was initialized more than once  
[       OK ] TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_without_intraprocess_comm.increasing_and_decreasing_counts (8031 ms)
[----------] 1 test from TestPublisherSubscriptionCount/two_subscriptions_in_two_contexts_without_intraprocess_comm (8031 ms total)
                                                                                
[==========] 4 tests from 4 test cases ran. (32129 ms total)                    
[  PASSED  ] 4 tests. 
```

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
```diff
@@ -329,6 +333,9 @@ class StaticExecutorEntitiesCollector final
 
   /// Executable list: timers, subscribers, clients, services and waitables
   rclcpp::experimental::ExecutableList exec_list_;
+
+  /// Bool to check if the entities collector has been initialized
+  bool initialized = false;
```
clalancette, line 338:
```suggestion
  bool initialized_ = false;
```

rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
```diff
@@ -78,6 +78,19 @@ class StaticSingleThreadedExecutor : public rclcpp::Executor
...
+  /**
+   * The default implementation is suitable for a single-threaded model of execution.
+   * Adding subscriptions, timers, services, etc. with blocking callbacks will cause
+   * this function to block (which may have unintended consequences).
+   *
+   * \param[in] max_duration The maximum amount of time to spend executing work, or 0 for no limit.
+   */
+  RCLCPP_PUBLIC
+  void
+  spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0)) override;
```
clalancette, line 92: Nit: `#include <chrono>` at the top of the file for this.

rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp
```diff
@@ -220,16 +233,10 @@ class StaticSingleThreadedExecutor : public rclcpp::Executor
     return rclcpp::FutureReturnCode::INTERRUPTED;
   }
 
-  /// Not yet implemented, see https://github.com/ros2/rclcpp/issues/1219 for tracking
+  /// Initialize entities collector, so we can spin_some without need to rebuilt it
```
clalancette, line 236:
```suggestion
  /// Initialize entities collector, so we can spin_some without having to rebuild it
```

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
```diff
@@ -76,8 +76,12 @@ StaticExecutorEntitiesCollector::init(
 
   // Add executor's guard condition
   memory_strategy_->add_guard_condition(executor_guard_condition);
+
   // Get memory strategy and executable list. Prepare wait_set_
   execute();
+
+  // The entities collector is now initialized
+  initialized = true;
```
clalancette, line 84:
```suggestion
  initialized_ = true;
```

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -50,6 +50,34 @@ StaticSingleThreadedExecutor::spin()
   }
 }
 
+void
+StaticSingleThreadedExecutor::spin_some(std::chrono::nanoseconds max_duration)
```
clalancette, line 54: `#include <chrono>` at the top of the file for this.

rclcpp/test/rclcpp/executors/test_static_single_threaded_executor.cpp
```diff
@@ -50,7 +50,6 @@ TEST_F(TestStaticSingleThreadedExecutor, check_unimplemented) {
   auto node = std::make_shared<rclcpp::Node>("node", "ns");
   executor.add_node(node);
 
-  EXPECT_THROW(executor.spin_some(), rclcpp::exceptions::UnimplementedError);
```
clalancette, line 53: It's good that you fixed this test, but can you please add some tests for the working version of `spin_some`?  We also want to make sure we cover all of the lines, so be sure to add tests for when things aren't initialized, and when `spin_some` is called inside of `spin_some`.

clalancette (changes requested):

rclcpp/test/rclcpp/executors/test_static_single_threaded_executor.cpp
mauropasse: Thanks for the review. I'll add the tests soon

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -50,6 +51,34 @@ StaticSingleThreadedExecutor::spin()
...
+{
+  // Make sure the entities collector has been initialized
+  if (!entities_collector_->is_init()) {
+    throw std::runtime_error("spin_some() called without initializing the entities collector");
```
ivanpauno, line 59: couldn't this implicitly initialize the entities collector?
Other executors don't need an initialization step before calling `spin_some()`.

rclcpp/test/rclcpp/executors/test_static_single_threaded_executor.cpp
ivanpauno: @mauropasse friendly ping

mauropasse: @ivanpauno thanks for the reminder. As soon as I have time I'll continue this work, currently I'm working on the **EventsExecutor**.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -50,6 +51,34 @@ StaticSingleThreadedExecutor::spin()
...
+  }
+
+  if (spinning.exchange(true)) {
+    throw std::runtime_error("spin_some() called while already spinning");
+  }
+  RCLCPP_SCOPE_EXIT(this->spinning.store(false); );
+
+  if (rclcpp::ok(context_) && spinning.load()) {
+    // Refresh wait set, wait for work and execute ready executables
+    entities_collector_->refresh_wait_set(max_duration);
```
fujitatomoya, line 69: if i am not mistaken, i was expecting `max_duration` is time to spend executing work, not to wait on wait_set. and test would be also ideal.

mauropasse: True, I have updated the implementation and added the `StaticSingleThreadedExecutor` to the spin tests.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
mauropasse: Yes, it can. The new implementation implicitly initializes the entities collector

rclcpp/test/rclcpp/executors/test_static_single_threaded_executor.cpp
mauropasse: I added the `StaticSingleThreadedExecutor` to the list of executors tested for the different spin variations

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -544,7 +544,7 @@ class Executor
   RCLCPP_DISABLE_COPY(Executor)
 
   RCLCPP_PUBLIC
-  void
+  virtual void
```
ivanpauno, line 547: why do you need to make this method virtual?

ivanpauno: is it because of `spin_until_future_complete`?

IMO having this method as virtual is a bit confusing, but it might be fine

rclcpp/test/rclcpp/executors/test_executors.cpp
```diff
@@ -525,10 +485,11 @@ TYPED_TEST(TestExecutorsStable, spinSome) {
   // count becomes too large, spin exits, or the 1 second timeout completes.
   auto start = std::chrono::steady_clock::now();
   while (
-    my_waitable->get_count() <= 1 &&
+    my_waitable->get_count() <= 10 &&
```
ivanpauno, line 488: why?

mauropasse: If `spin_some(1s)` works as expected, the executor should execute only a single waitable event, and should ignore new notifications about the waitable being triggered more times.

If in the while loop (which triggers the waitable) we limit the count to 1, we are not testing the possibility of `spin_some` not working properly, that is, executing the waitable more than a single time. In order to test this situation, the waitable should have been triggered more than a single time (which currently could not happen if we limit the count to 1). So the test will be succesful:
 `EXPECT_EQ(1u, my_waitable->get_count())` even if `spin_some` didn't work as expected.

Now lets suppose `spin_some` doesn't work as expected, with the change `get_count() <= 10` , we can trigger the waitable many times (up to 10) and the executor might execute many waitable events. In this case, the test won't be succesful, if `my_waitable->get_count() = 10;`
`EXPECT_EQ(1u, my_waitable->get_count())` 

Furthermore, I think a `sleep(10ms)` should be added before `executor.spin_some(1s);`, to give time the executor to already have events when it starts spinning. Otherwise it might start spinning without having any events, so the spin exits, the test fails, but it was due it didn't have anything to do yet. What do you think?

mauropasse: I have reverted this change also. What I said above is valid for the wait_set executors, which can execute as max one event per entity, per `spin_some` call.
It is still not here, but the `EventsExecutor` can process multiple events for the same entity in the same `spin_some` call, if at the time of calling it there were already many events ready to be executed for the same entity.

ivanpauno: > If in the while loop (which triggers the waitable) we limit the count to 1, we are not testing the possibility of spin_some not working properly, that is, executing the waitable more than a single time. In order to test this situation, the waitable should have been triggered more than a single time (which currently could not happen if we limit the count to 1). So the test will be succesful:

Good point, that seems to make sense to me.

> Furthermore, I think a sleep(10ms) should be added before executor.spin_some(1s);, to give time the executor to already have events when it starts spinning. Otherwise it might start spinning without having any events, so the spin exits, the test fails, but it was due it didn't have anything to do yet. What do you think?

The TestWaitable is always returning ready, so that doesn't seem to be a problem

> It is still not here, but the EventsExecutor can process multiple events for the same entity in the same spin_some call, if at the time of calling it there were already many events ready to be executed for the same entity.

That sounds reasonable to me

rclcpp/include/rclcpp/executor.hpp
ivanpauno: I'm still confused if [this](https://github.com/ros2/rclcpp/blob/7bfda87d32bab7e9caa843b2ededa18e427fd00c/rclcpp/include/rclcpp/executor.hpp#L359) will work as expected.

ivanpauno:
@mauropasse could you check the linters errors https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/1009/testReport/junit/rclcpp/uncrustify/src_rclcpp_executors_static_single_threaded_executor_cpp/.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13997/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8820/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11690/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14076/)

ivanpauno (commented):
I'm not sure if `spin_until_future_complete()` is doing what expected (see https://github.com/ros2/rclcpp/pull/1338#discussion_r597971382), otherwise lgtm

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14032/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8852/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11725/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14123/)

ivanpauno (approved):
At some point we should refactor the executor code, have a abstract base class and share code between implementations that have things in common in a different way.

The current code is getting really confusing...

rclcpp/include/rclcpp/executor.hpp
ivanpauno: I didn't see that `spin_until_future_complete` was being overloaded in the derived class though.
But making `spin_once_impl` virtual doesn't seem to make much harm (though I thing the current code is getting very confusing).

wjwwood:
@ivanpauno is this ok to merge?

ivanpauno:
I'm confused about test failures, though they look unrelated.
@mauropasse can you rebase with master?

wjwwood:
Rerunning CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14232/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9030/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11923/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14336/)

wjwwood:
The compiler warnings on Linux/macOS are due to the tracetools deprecation, which has been fixed.

The cmake warning on Windows is from cyclonedds and is already known.

wjwwood:
@clalancette I think that your feedback was addressed, so I'm gonna merge now. We can follow up if need be.

msy22:
Is this an issue that is going to or has already been fixed in Foxy? Currently I get this error when I call `spin_some` on a StaticSingleThreadedExecutor:

```
[tf_broadcaster_exe-1] terminate called after throwing an instance of 'rclcpp::exceptions::UnimplementedError'
[tf_broadcaster_exe-1]   what():  spin_some is not implemented for StaticSingleThreadedExecutor, use spin or spin_until_future_complete
```

This is dispite the fact that the [StaticSingleThreadedExecutor API](https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1executors_1_1StaticSingleThreadedExecutor.html#details) clearly states that this is a public member function inherited from `rclcpp::Executor`

clalancette:
As-is, this patch wouldn't be eligible for Foxy; it breaks ABI.

If someone wanted to try to make an ABI-compatible version for Foxy, we would certainly review and consider it.

wjwwood:
> This is dispite the fact that the StaticSingleThreadedExecutor API clearly states that this is a public member function inherited from rclcpp::Executor

This was done explicitly because the generic implementation in `rclcpp::Executor::spin_some_impl()` doesn't work for the design of the static executor. So it is explicitly not implemented (until this pull request).

msy22:
Bother, unfortunately I don't know what ABI is, nor do I have the time to make those changes. SingleThreadedExecutor will work, I was just hoping to take advantage of the apparent CPU overhead reductions that I've seen talked about.

Thanks for the explanations!

wjwwood:
https://en.wikipedia.org/wiki/Application_binary_interface in case you're interested.

PR #1342, opened 2020-09-28
Test the remaining node public API

brawner:
This adds tests for the remaining public API in node.hpp/node.cpp.

### DIFF
--- a/rclcpp/test/rclcpp/test_node.cpp
+++ b/rclcpp/test/rclcpp/test_node.cpp
@@ -26,7 +26,14 @@
 #include "rclcpp/rclcpp.hpp"
 
 #include "rcpputils/filesystem_helper.hpp"
+
+#include "rmw/validate_namespace.h"
+
 #include "test_msgs/msg/basic_types.hpp"
+#include "test_msgs/msg/empty.hpp"
+#include "test_msgs/srv/empty.hpp"
+
+#include "../mocking_utils/patch.hpp"
 
 class TestNode : public ::testing::Test
 {
@@ -50,7 +57,19 @@ protected:
 TEST_F(TestNode, construction_and_destruction) {
   {
     auto node = std::make_shared<rclcpp::Node>("my_node", "/ns");
-    (void)node;
+    EXPECT_NE(nullptr, node->get_node_base_interface());
+    EXPECT_NE(nullptr, node->get_node_clock_interface());
+    EXPECT_NE(nullptr, node->get_node_graph_interface());
+    EXPECT_NE(nullptr, node->get_node_logging_interface());
+    EXPECT_NE(nullptr, node->get_node_time_source_interface());
+    EXPECT_NE(nullptr, node->get_node_timers_interface());
+    EXPECT_NE(nullptr, node->get_node_topics_interface());
+    EXPECT_NE(nullptr, node->get_node_services_interface());
+    EXPECT_NE(nullptr, node->get_node_parameters_interface());
+    EXPECT_NE(nullptr, node->get_node_waitables_interface());
+    EXPECT_NE(nullptr, node->get_node_options().get_rcl_node_options());
+    EXPECT_NE(nullptr, node->get_graph_event());
+    EXPECT_NE(nullptr, node->get_clock());
   }
 
   {
@@ -279,8 +298,11 @@ TEST_F(TestNode, get_logger) {
 TEST_F(TestNode, get_clock) {
   auto node = std::make_shared<rclcpp::Node>("my_node", "/ns");
   auto ros_clock = node->get_clock();
-  EXPECT_TRUE(ros_clock != nullptr);
+  EXPECT_NE(nullptr, ros_clock);
   EXPECT_EQ(ros_clock->get_clock_type(), RCL_ROS_TIME);
+
+  const rclcpp::Node & const_node = *node.get();
+  EXPECT_NE(nullptr, const_node.get_clock());
 }
 
 TEST_F(TestNode, now) {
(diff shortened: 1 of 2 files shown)


brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12511/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7479/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10223/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12431/)

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -796,6 +818,19 @@ TEST_F(TestNode, has_parameter) {
   }
 }
 
+TEST_F(TestNode, list_parameters) {
+  auto node = std::make_shared<rclcpp::Node>("test_list_parameter_node"_unq);
+  {
```
clalancette, line 823: Do we need this extra scope?  It seems like everything will go out of scope properly without it, but I may be missing something.

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -2455,3 +2490,89 @@ TEST_F(TestNode, get_publishers_subscriptions_info_by_topic) {
...
+  EXPECT_EQ(2u + num_callback_groups_in_basic_node, node->get_callback_groups().size());
+}
+
+// This is tested more thoroughly in node_interfaces/test_node_graph
+TEST_F(TestNode, get_entity_names) {
+  auto node = std::make_shared<rclcpp::Node>("node", "ns");
+  const auto node_names = node->get_node_names();
+  EXPECT_NE(
+    node_names.end(),
+    std::find(node_names.begin(), node_names.end(), node->get_fully_qualified_name()));
```
clalancette, line 2511: Would it make sense to go for a positive test here?  That is, something like:

```
EXPECT_EQ("/ns/node", std::find(node_names.begin(), node_names.end(), node->get_fully_qualified_name()));
```

(as a side note, reviewing this is exactly why I don't like `auto`.  In order for me to tell what type `node->get_node_names()` returned, I had to go looking at the header file, which isn't part of this PR.  It's not the end of the world, but just makes life more difficult for the reader)

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -2455,3 +2490,89 @@ TEST_F(TestNode, get_publishers_subscriptions_info_by_topic) {
...
+      EXPECT_NO_THROW(node->wait_for_graph_change(event, timeout));
+      thread_completion = std::chrono::steady_clock::now();
+    });
+
+  // Start creating nodes until event triggers in graph_event_wait_thread or until 100 nodes have
+  // been created
+  std::vector<std::shared_ptr<rclcpp::Node>> nodes;
+  while (thread_completion == thread_start && nodes.size() < 100) {
+    nodes.emplace_back(std::make_shared<rclcpp::Node>("node"_unq, "ns"));
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
clalancette, line 2550: I'm concerned that this will be a flaky test, particularly on Windows.  Can we really create 100 nodes in under 10 seconds reliably?  Out of curiosity, why did you choose 100, and not something like 10?

clalancette (commented):
I've got two questions here.  Otherwise, looks good.

Blast545 (approved):
LGTM

rclcpp/test/rclcpp/test_node.cpp
brawner: Fixed. (result of copy-paste from unit test just above this one)

rclcpp/test/rclcpp/test_node.cpp
brawner: I'm not sure it adds anything. Checking that find didn't return `end()` would still be required in some form because dereferencing `end()` would crash the test. Likewise, we're not checking that `std::find` works, just that the node's name is in the collection of node_names.

My counter-argument would be that this unit test is robust to changes in the return type collection (as long as that type is suitable for std::find), which is also valuable.

rclcpp/test/rclcpp/test_node.cpp
brawner: This test isn't about creating 100 nodes, in fact on my machine only one was created before the test succeeded. The race condition is that `node->wait_for_graph_change()` may be called after the first node is created. So this while loop continues to create nodes until `wait_for_graph_change()` detects a change in the node graph. 100 was chosen as a sufficiently large number to assert the failure case. 10 would also work, but with a higher number, this test will be theoretically less flaky.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12533/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7502/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10245/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12452/)

rclcpp/test/rclcpp/test_node.cpp
clalancette: > I'm not sure it adds anything. Checking that find didn't return `end()` would still be required in some form because dereferencing `end()` would crash the test. Likewise, we're not checking that `std::find` works, just that the node's name is in the collection of node_names.

That's a good point.  OK, I'm fine with this test as-is.

rclcpp/test/rclcpp/test_node.cpp
clalancette: Ah, that makes a lot more sense.  Thanks for the explanation.  The comment right above didn't make that totally clear to me; maybe update it to say something like:

```
  // Start creating nodes until at least one event triggers in graph_event_wait_thread or until 100 nodes have
  // been created (at which point this is a failure)
```

That being said, that isn't blocking, so I'll still approve.

clalancette (approved):

rclcpp/test/rclcpp/test_node.cpp
brawner: Updated comment with this language

brawner:
Rebuilding linux for the most recent minor commit
[Build Status](https://ci.ros2.org/job/ci_linux/12542/)

PR #1344, opened 2020-09-29
Complete coverage of Parameter and ParameterValue API

brawner:
This PR adds several more tests to complete the testing of Parameter's and ParameterValue's API. Tests were added for the `<<` operator for both types and a vector of parameters, as well as the `==` and `!=` operators for Parameters. 

### DIFF
--- a/rclcpp/test/rclcpp/test_parameter.cpp
+++ b/rclcpp/test/rclcpp/test_parameter.cpp
@@ -71,6 +71,7 @@ TEST_F(TestParameter, not_set_variant) {
     not_set_variant.get_value<int>(),
     rclcpp::exceptions::InvalidParameterTypeException);
 
+  // Check == and != operators work as expected
   EXPECT_EQ(not_set_variant, not_set_variant);
   rclcpp::Parameter not_set_variant2;
   EXPECT_EQ(not_set_variant, not_set_variant2);
@@ -139,6 +140,7 @@ TEST_F(TestParameter, bool_variant) {
     rcl_interfaces::msg::ParameterType::PARAMETER_BOOL,
     bool_variant_false.get_value_message().type);
 
+  // Check == and != operators work as expected
   EXPECT_EQ(bool_variant_true, bool_variant_true);
   EXPECT_NE(bool_variant_false, bool_variant_true);
   EXPECT_EQ(bool_variant_true, from_msg_true);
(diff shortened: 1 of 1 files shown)


rclcpp/test/rclcpp/test_parameter.cpp
```diff
@@ -44,6 +44,9 @@ TEST_F(TestParameter, not_set_variant) {
   rclcpp::Parameter not_set_variant;
   EXPECT_EQ(rclcpp::PARAMETER_NOT_SET, not_set_variant.get_type());
   EXPECT_EQ("not set", not_set_variant.get_type_name());
+  std::stringstream ss;
+  ss << not_set_variant;
+  EXPECT_EQ("{\"name\": \"\", \"type\": \"not set\", \"value\": \"not set\"}", ss.str());
```
clalancette, line 49: I know you didn't change this, but this is weird thing to be returning when trying to stream `PARAMETER_NOT_SET`.  I'd find it much more intuitive if the stream operator threw an exception.

If you agree, my suggestion is that we leave this test as-is (to test the current behavior), but open a follow-up issue to fix it later.

clalancette (approved):
I've left one comment which doesn't block this.  Looks good to me with green CI.

rclcpp/test/rclcpp/test_parameter.cpp
```diff
@@ -67,6 +70,12 @@ TEST_F(TestParameter, not_set_variant) {
   EXPECT_THROW(
     not_set_variant.get_value<int>(),
     rclcpp::exceptions::InvalidParameterTypeException);
+
+  EXPECT_EQ(not_set_variant, not_set_variant);
```
Blast545, line 74: This comparison to itself looks odd, I suppose you are testing a == operator? Please add a comment or maybe remove this in favor of the test checking `not_set_variant == not_set_variant2`

Blast545: I won't block PR on this though

Blast545 (approved):
LGTM

rclcpp/test/rclcpp/test_parameter.cpp
brawner: I'm not aware of many uses of throwing exceptions in the `<<` operator, but I guess it depends on how much of an error it assumed to be when trying to stream an unset parameter.

Follow-up issue for comment created here. https://github.com/ros2/rclcpp/issues/1348

rclcpp/test/rclcpp/test_parameter.cpp
brawner: Yah, this is checking that `==` does in fact agree that the same object is equal to itself.

brawner: I added a comment though, which is a great idea.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12534/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7503/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10246/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12453/)

PR #1346, opened 2020-09-29
Add in more coverage for expand_topic_or_service_name.

clalancette:
This gets this file up to 97% coverage.

### DIFF
--- a/rclcpp/test/CMakeLists.txt
+++ b/rclcpp/test/CMakeLists.txt
@@ -117,7 +117,7 @@ if(TARGET test_expand_topic_or_service_name)
     "rosidl_runtime_cpp"
     "rosidl_typesupport_cpp"
   )
-  target_link_libraries(test_expand_topic_or_service_name ${PROJECT_NAME})
+  target_link_libraries(test_expand_topic_or_service_name ${PROJECT_NAME} mimick)
 endif()
 ament_add_gtest(test_function_traits rclcpp/test_function_traits.cpp)
 if(TARGET test_function_traits)
--- a/rclcpp/test/rclcpp/test_expand_topic_or_service_name.cpp
+++ b/rclcpp/test/rclcpp/test_expand_topic_or_service_name.cpp
@@ -14,8 +14,19 @@
 
 #include <gtest/gtest.h>
 
+#include <stdexcept>
+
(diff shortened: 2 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12526/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7495/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10238/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12445/)

brawner (approved):
This looks good, but what remains to get the file to 100%?

clalancette:
> This looks good, but what remains to get the file to 100%?

Yeah, that's a good question.  The two missing lines are at https://github.com/ros2/rclcpp/blob/4a2231d7a5bbfacaeef5b914a7f8b97500d89ade/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp#L137 and at https://github.com/ros2/rclcpp/blob/4a2231d7a5bbfacaeef5b914a7f8b97500d89ade/rclcpp/src/rclcpp/expand_topic_or_service_name.cpp#L164 .  They are not totally straightforward to get to because it isn't enough to inject a return value; we actually need to modify one of the "out" parameters.

This morning I couldn't come up with a good way to reach them.  I may have figured out a solution, but I've gotten bogged down in testing utilities.cpp, so I haven't had time to look back.

I'm going to go ahead and merge this as-is, and potentially follow-up with those two lines a little later.  Thanks for the review.

brawner:
When you need to modify an output parameter, you can use `mocking_utils::patch()`, which takes as input a callback to call, instead of just returning a value instead. A good example that looks very similar to these conditions is done in `test_node_base.cpp`,

https://github.com/ros2/rclcpp/blob/554d933f518d06c839194d0ec21b40c436e123de/rclcpp/test/rclcpp/node_interfaces/test_node_base.cpp#L121-L135

clalancette:
> When you need to modify an output parameter, you can use `mocking_utils::patch()`, which takes as input a callback to call, instead of just returning a value instead. A good example that looks very similar to these conditions is done in `test_node_base.cpp`,

Oh, perfect!  Thanks for the example.

PR #1378, opened 2020-10-05
[backport foxy] Log error instead of throwing exception in Transition and State reset() mark no except (#1297)

brawner:
Backports #1297 for foxy

### DIFF
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -87,7 +87,7 @@ public:
 protected:
   RCLCPP_LIFECYCLE_PUBLIC
   void
-  reset() noexcept;
+  reset();
 
   rcutils_allocator_t allocator_;
 
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -87,7 +87,7 @@ public:
 protected:
   RCLCPP_LIFECYCLE_PUBLIC
   void
-  reset();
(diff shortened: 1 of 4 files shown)


brawner:
@jacobperron should I just backport the foxy and not the API change? (marking reset() here noexcept)

jacobperron (approved):

jacobperron:

jacobperron:
> should I just backport the fix and not the API change? (marking reset() here noexcept)

Good question. It isn't clear to me if it is guaranteed to be ABI compatible. Perhaps it is safest to remove the `noexcept`.

rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
```diff
@@ -87,7 +87,7 @@ class State
 protected:
   RCLCPP_LIFECYCLE_PUBLIC
   void
-  reset();
+  reset() noexcept;
```
jacobperron, line 90: It's probably best to avoid backporting this change.

jacobperron (changes requested):

rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
brawner: Done

jacobperron (approved):

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12619/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7581/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10326/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12544/)

PR #1384, opened 2020-10-06
Update maintainers

ivanpauno:


mabelzhang (approved):
Do we care about consistency in the emails or does William prefer osrfoundation.org to openrobotics.org?

ivanpauno:
> Do we care about consistency in the emails or does William prefer osrfoundation.org to openrobotics.org?

My bad, I think `@openrobotics.org` is the correct option.
I will update when @wjwwood confirms (and I can update dthomas mail too).

wjwwood (approved):

wjwwood:
I don't mind which email is used.

Issue #1340, opened 2020-09-28
graph_listener cleaning memory managed by parent context


Blast545:
I found this issue when adding coverage tests for `rclcpp`, PR: #1330.

This is the causing problems, just in MacOS:
``` c++
    auto context_to_destroy = std::make_shared<rclcpp::contexts::DefaultContext>();
    context_to_destroy->init(0, nullptr);
    auto graph_listener_error =
      std::make_shared<rclcpp::graph_listener::GraphListener>(context_to_destroy);
    context_to_destroy.reset();
    RCLCPP_EXPECT_THROW_EQ(
      graph_listener_error->start_if_not_started(),
      std::runtime_error("parent context was destroyed"));
```

#### Expected behavior
graph_listener is destroyed cleanly

#### Actual behavior
graph_listener causes segfault when calling its shutdown method (destruction).

The portion of code causing a problem is: 

``` c++
    if (shutdown_guard_condition_) {
      auto parent_context_ptr = parent_context_.lock();
      if (parent_context_ptr) {
        if (should_throw) {
          parent_context_ptr->release_interrupt_guard_condition(&wait_set_);
        } else {
          parent_context_ptr->release_interrupt_guard_condition(&wait_set_, std::nothrow);
        }
      } else {
        ret = rcl_guard_condition_fini(shutdown_guard_condition_);
        if (RCL_RET_OK != ret) {
          if (should_throw) {
            throw_from_rcl_error(ret, "failed to finalize shutdown guard condition");
          } else {
            RCLCPP_ERROR(
              rclcpp::get_logger("rclcpp"),
              "failed to finalize shutdown guard condition");
          }
        }
      }
      shutdown_guard_condition_ = nullptr;
```

The `shutdown_guard_condition_` memory is allocated at the `context` level [here](https://github.com/ros2/rclcpp/blob/b4b0afc26752ffbd9c6ec07922b51637a966185f/rclcpp/src/rclcpp/context.cpp#L405) so I think it shouldn't be fini at the `graph_listener`, it should be fini at the `context` level, and `graph_listener` would just check the weak pointer referencing the parent context ptr. This would change this portion of code to be:

``` c++
      auto parent_context_ptr = parent_context_.lock();
      if (parent_context_ptr) {
        if (should_throw) {
          parent_context_ptr->release_interrupt_guard_condition(&wait_set_);
        } else {
          parent_context_ptr->release_interrupt_guard_condition(&wait_set_, std::nothrow);
        }
[...]

PR #1404, opened 2020-10-14
Move ownership of shutdown_guard_condition to executors/graph_listener

brawner:
Ownership of the shutdown guard conditions used by the graph_listener and the executor base class was in rclcpp::Context, which would lead to memory errors if the context was cleaned up first. 

This moves ownership of the guard_condition to the executor and graph_listener classes, and removes the guard_condition API from Context.

I'm marking this as a draft PR so I can get some initial review on it since I'm not terribly familiar with rclcpp's API.

Addresses #1391 and #1340

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -25,6 +25,7 @@
 
 #include "rclcpp/exceptions.hpp"
 #include "rclcpp/executor.hpp"
+#include "rclcpp/guard_condition.hpp"
 #include "rclcpp/memory_strategy.hpp"
 #include "rclcpp/node.hpp"
 #include "rclcpp/scope_exit.hpp"
@@ -40,55 +41,9 @@ using rclcpp::Executor;
 using rclcpp::ExecutorOptions;
 using rclcpp::FutureReturnCode;
 
-namespace
-{
-
-std::shared_ptr<rcl_guard_condition_t>
-create_shutdown_guard_condition(std::shared_ptr<rclcpp::Context> context)
-{
-  auto custom_deleter = [](rcl_guard_condition_t * guard_condition) {
-      rcl_ret_t ret = rcl_guard_condition_fini(guard_condition);
-      if (RCL_RET_OK != ret) {
-        RCLCPP_ERROR(
-          rclcpp::get_logger("rclcpp"),
-          "Failed to finalize shutdown_guard_condition, leaking memory!");
-      }
-    };
-
-  auto shutdown_guard_condition =
-    std::shared_ptr<rcl_guard_condition_t>(new rcl_guard_condition_t, custom_deleter);
-  *shutdown_guard_condition = rcl_get_zero_initialized_guard_condition();
-
-  rcl_ret_t ret = rcl_guard_condition_init(
-    shutdown_guard_condition.get(),
-    context->get_rcl_context().get(),
-    rcl_guard_condition_get_default_options());
-  if (RCL_RET_OK != ret) {
-    throw_from_rcl_error(ret, "failed to create shutdown guard condition");
-  }
-
-  context->on_shutdown(
-    [weak_shutdown_guard_condition =
-    std::weak_ptr<rcl_guard_condition_t>(shutdown_guard_condition)]()
-    {
-      auto shared_guard_condition = weak_shutdown_guard_condition.lock();
-      if (nullptr != shared_guard_condition) {
-        rcl_ret_t status = rcl_trigger_guard_condition(shared_guard_condition.get());
-        if (status != RCL_RET_OK) {
-          RCUTILS_LOG_ERROR_NAMED(
-            "rclcpp",
-            "failed to trigger guard condition in Context shutdown callback(): %s",
-            rcl_get_error_string().str);
-        }
-      }
-    });
-  return shutdown_guard_condition;
-}
-
-}  // namespace
-
 Executor::Executor(const rclcpp::ExecutorOptions & options)
 : spinning(false),
+  shutdown_guard_condition_(std::make_shared<rclcpp::GuardCondition>(options.context)),
   memory_strategy_(options.memory_strategy)
 {
   // Store the context for later use.
@@ -101,14 +56,17 @@ Executor::Executor(const rclcpp::ExecutorOptions & options)
     throw_from_rcl_error(ret, "Failed to create interrupt guard condition in Executor constructor");
   }
 
-  // Register a shutdown hook for the executor with context
-  shutdown_guard_condition_ = create_shutdown_guard_condition(context_);
+  context_->on_shutdown(
+    [weak_gc = std::weak_ptr<rclcpp::GuardCondition>{shutdown_guard_condition_}]() {
+      auto strong_gc = weak_gc.lock();
+      if (strong_gc) {
+        strong_gc->trigger();
+      }
+    });
 
   // The number of guard conditions is always at least 2: 1 for the ctrl-c guard cond,
   // and one for the executor's guard cond (interrupt_guard_condition_)
-
-  // Put the global ctrl-c guard condition in
-  memory_strategy_->add_guard_condition(shutdown_guard_condition_.get());
+  memory_strategy_->add_guard_condition(&shutdown_guard_condition_->get_rcl_guard_condition());
 
   // Put the executor's guard condition in
   memory_strategy_->add_guard_condition(&interrupt_guard_condition_);
@@ -179,7 +137,7 @@ Executor::~Executor()
     rcl_reset_error();
   }
   // Remove and release the sigint guard condition
-  memory_strategy_->remove_guard_condition(shutdown_guard_condition_.get());
+  memory_strategy_->remove_guard_condition(&shutdown_guard_condition_->get_rcl_guard_condition());
 }
 
 std::vector<rclcpp::CallbackGroup::WeakPtr>
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -40,29 +40,78 @@ using rclcpp::Executor;
 using rclcpp::ExecutorOptions;
 using rclcpp::FutureReturnCode;
 
+namespace {
+
+std::shared_ptr<rcl_guard_condition_t>
+CreateShutdownGuardCondition(std::shared_ptr<rclcpp::Context> context)
+{
+  auto custom_deleter = [](rcl_guard_condition_t * guard_condition) {
+      rcl_ret_t ret = rcl_guard_condition_fini(guard_condition);
+      if (RCL_RET_OK != ret) {
+        RCLCPP_ERROR(
+          rclcpp::get_logger("rclcpp"),
+          "Failed to finalize shutdown_guard_condition, leaking memory!");
+      }
+    };
+
+  auto shutdown_guard_condition =
+    std::shared_ptr<rcl_guard_condition_t>(new rcl_guard_condition_t, custom_deleter);
+  *shutdown_guard_condition = rcl_get_zero_initialized_guard_condition();
+
+  rcl_ret_t ret = rcl_guard_condition_init(
+    shutdown_guard_condition.get(),
+    context->get_rcl_context().get(),
+    rcl_guard_condition_get_default_options());
+  if (RCL_RET_OK != ret) {
+    throw_from_rcl_error(ret, "failed to create shutdown guard condition");
+  }
+
+  context->on_shutdown(
+    [weak_shutdown_guard_condition =
+    std::weak_ptr<rcl_guard_condition_t>(shutdown_guard_condition)]()
+    {
+      auto shared_guard_condition = weak_shutdown_guard_condition.lock();
+      if (nullptr != shared_guard_condition) {
+        rcl_ret_t status = rcl_trigger_guard_condition(shared_guard_condition.get());
+        if (status != RCL_RET_OK) {
+          RCUTILS_LOG_ERROR_NAMED(
+            "rclcpp",
+            "failed to trigger guard condition in Context shutdown callback(): %s",
+            rcl_get_error_string().str);
+        }
+      }
+    });
+  return shutdown_guard_condition;
+}
+
+}  // namespace
+
 Executor::Executor(const rclcpp::ExecutorOptions & options)
 : spinning(false),
   memory_strategy_(options.memory_strategy)
 {
+  // Store the context for later use.
+  context_ = options.context;
+
   rcl_guard_condition_options_t guard_condition_options = rcl_guard_condition_get_default_options();
   rcl_ret_t ret = rcl_guard_condition_init(
-    &interrupt_guard_condition_, options.context->get_rcl_context().get(), guard_condition_options);
+    &interrupt_guard_condition_, context_->get_rcl_context().get(), guard_condition_options);
   if (RCL_RET_OK != ret) {
     throw_from_rcl_error(ret, "Failed to create interrupt guard condition in Executor constructor");
   }
+
+  // Register a shutdown hook for the executor with context
+  shutdown_guard_condition_ = CreateShutdownGuardCondition(context_);
 
   // The number of guard conditions is always at least 2: 1 for the ctrl-c guard cond,
   // and one for the executor's guard cond (interrupt_guard_condition_)
 
   // Put the global ctrl-c guard condition in
-  memory_strategy_->add_guard_condition(options.context->get_interrupt_guard_condition(&wait_set_));
+  memory_strategy_->add_guard_condition(shutdown_guard_condition_.get());
 
   // Put the executor's guard condition in
   memory_strategy_->add_guard_condition(&interrupt_guard_condition_);
   rcl_allocator_t allocator = memory_strategy_->get_allocator();
-
-  // Store the context for later use.
-  context_ = options.context;
 
   ret = rcl_wait_set_init(
     &wait_set_,
@@ -129,8 +178,7 @@ Executor::~Executor()
     rcl_reset_error();
   }
   // Remove and release the sigint guard condition
-  memory_strategy_->remove_guard_condition(context_->get_interrupt_guard_condition(&wait_set_));
-  context_->release_interrupt_guard_condition(&wait_set_, std::nothrow);
+  memory_strategy_->remove_guard_condition(shutdown_guard_condition_.get());
 }
 
 std::vector<rclcpp::CallbackGroup::WeakPtr>
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -40,10 +40,11 @@ using rclcpp::Executor;
 using rclcpp::ExecutorOptions;
 using rclcpp::FutureReturnCode;
 
-namespace {
+namespace
+{
 
 std::shared_ptr<rcl_guard_condition_t>
-CreateShutdownGuardCondition(std::shared_ptr<rclcpp::Context> context)
+create_shutdown_guard_condition(std::shared_ptr<rclcpp::Context> context)
 {
   auto custom_deleter = [](rcl_guard_condition_t * guard_condition) {
       rcl_ret_t ret = rcl_guard_condition_fini(guard_condition);
@@ -101,7 +102,7 @@ Executor::Executor(const rclcpp::ExecutorOptions & options)
   }
 
   // Register a shutdown hook for the executor with context
-  shutdown_guard_condition_ = CreateShutdownGuardCondition(context_);
+  shutdown_guard_condition_ = create_shutdown_guard_condition(context_);
 
   // The number of guard conditions is always at least 2: 1 for the ctrl-c guard cond,
   // and one for the executor's guard cond (interrupt_guard_condition_)
--- a/rclcpp/src/rclcpp/graph_listener.cpp
+++ b/rclcpp/src/rclcpp/graph_listener.cpp
@@ -33,60 +33,14 @@ using rclcpp::exceptions::throw_from_rcl_error;
 
 namespace rclcpp
 {
-namespace {
-
-std::shared_ptr<rcl_guard_condition_t>
-CreateShutdownGuardCondition(rclcpp::Context * context)
-{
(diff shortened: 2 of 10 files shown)


brawner:
Testing `--packages-above rclcpp`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12728/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7689/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10442/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12675/)

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -39,30 +39,79 @@ using rclcpp::Executor;
 using rclcpp::ExecutorOptions;
 using rclcpp::FutureReturnCode;
 
+namespace {
+
+std::shared_ptr<rcl_guard_condition_t>
+CreateShutdownGuardCondition(std::shared_ptr<rclcpp::Context> context)
```
ivanpauno, line 45: nit^N: this is creating a copy of the shared pointer (inc the ref count), and that's not necessary because the function isn't retaining ownership.
Prefer `rclcpp::Context & context` (`rclcpp::Context * context` is also ok)

ivanpauno (approved):
This looks great to me, I think it makes ownership of the gc more clear.

ivanpauno:
ps: uncrustify isn't passing

wjwwood (commented):
So, I'm wondering if this is a case that should be supported at all.

Since the guard condition is destroyed after the context, I think it leaves the guard condition in an undefined state since it has a pointer to the context which it is associated with:

https://github.com/ros2/rcl/blob/0ca8c6162136594e490ba69fbf288de1b5f7073d/rcl/include/rcl/guard_condition.h#L36

So I think the guard condition needs to be destroyed before the context. This kind of makes sense because the destruction of the context would destroy the participant and then make the DDS guard condition under the hood invalid as well (potentially, I don't know this for sure).

The existing implementation does this by having the rclcpp context own the guard conditions so it can destroy them before the rcl context.

So maybe we should keep that, then synchronize the shutdown of the graph listener (and executors) with the shutdown of the context so that the guard condition is never in use when the context is being destroyed and so the destructor of the context can destroy the guard conditions and then the context.

I know that's a lot harder, but I think it's the right thing to do.

brawner:
@wjwwood That seems reasonable given that rcl_guard_condition_t has a pointer to the context. Another suggestion @ivanpauno had was to change the `rclcpp::Context::get_interrupt_guard_condition()` to return a weak_ptr to the guard_condition, which I think would also work well. Is that the sort of synchronization you were thinking?

ivanpauno:
> So I think the guard condition needs to be destroyed before the context. This kind of makes sense because the destruction of the context would destroy the participant and then make the DDS guard condition under the hood invalid as well (potentially, I don't know this for sure).

That's a good point.

> The existing implementation does this by having the rclcpp context own the guard conditions so it can destroy them before the rcl context.

That's true, but we use a different approach in other places.
e.g.: publishers must be destroyed after nodes, and the current trick is that `rclcpp::Publisher` is keeping shared ownership of the `rcl_node_t` handle ([here](https://github.com/ros2/rclcpp/blob/ed68b4bde771cf7c1affb422f8a8cd87b51bec4c/rclcpp/include/rclcpp/publisher_base.hpp#L215)) (and not the opposite, i.e. the node doesn't own the publisher).

This could also work here, we could have a `rclcpp::GuardCondition` that extends the lifetime of a `rcl_context_t`.
I wonder if this approach is correct, e.g.:

- context A
  - node A
    - pub A
    - pub B

`Pub A` and `pub B` is extending the lifetime of `node A`, that's ok.
But how is the lifetime of context `A` extended?
I mean, if `node A` rclcpp object and `context A` rclcpp object go out of scope, `context A` rcl object will also go out of scope and that's an issue (because `pub A` and `pub B` are extending the lifetime of `node A` rcl handle, which doesn't extend the lifetime of `context A`).
I think we never run into the issue because almost always we're using the "default context" singleton, which will be destroyed after "everything".

> So maybe we should keep that, then synchronize the shutdown of the graph listener (and executors) with the shutdown of the context so that the guard condition is never in use when the context is being destroyed and so the destructor of the context can destroy the guard conditions and then the context.
> 
> I know that's a lot harder, but I think it's the right thing to do.

I'm not pretty sure of how synchronization can be achieved here.

> @wjwwood That seems reasonable given that rcl_guard_condition_t has a pointer to the context. Another suggestion @ivanpauno had was to change the rclcpp::Context::get_interrupt_guard_condition() to return a weak_ptr to the guard_condition, which I think would also work well. Is that the sort of synchronization you were thinking?

I think that's ok, but it's still possible that someone locks the weak pointer and keeps shared ownership of the gc, if the gc handle is not extending the lifetime of the context the same issue is still possible.

wjwwood:
Well you cannot extend the lifetime of the context using the guard condition or the graph listener, because then you'll have a circular dependency. Perhaps the cycle could be broken, but I haven't thought through it enough.

That's why I think it should be undefined behavior to delete the guard condition after the context is destroyed, and due to that we must ensure that the graph listener and others are no longer using the guard condition before destroying the context.

ivanpauno:
> Well you cannot extend the lifetime of the context using the guard condition or the graph listener, because then you'll have a circular dependency. Perhaps the cycle could be broken, but I haven't thought through it enough.

The cycle can be broken by extending the lifetime of the `rcl_context_t` handle ([which is accessible]) instead of extending `rclcpp::Context`.
We do that in other places, but I think that approach has some caveats too (see comment above, though I'm not sure if I'm correct in that example).

[which is accessible]: https://github.com/ros2/rclcpp/blob/31c202e325476f6e912d6d23c17d13393156dbf4/rclcpp/include/rclcpp/context.hpp#L227

clalancette:
I think there is still technical details to sort out.  As such, I've assigned this to @brawner , but let me know if something else should be done here.

ivanpauno:
Some observations based on checking the logic here again:

- In "real code", `GraphListener` is only created as a [subcontext](https://github.com/ros2/rclcpp/blob/e62f3280f55e806d7d63d5bce35c172e3fc2a9bd/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp#L38), so it will always be destroyed in its parent context destructor or outlive the context.
- [this](https://github.com/ros2/rclcpp/blob/a95efa452e0c6dcc4fd6854d36cb800a8e49cbfe/rclcpp/src/rclcpp/graph_listener.cpp#L381) `std::weak_ptr<rclcpp::Context>::lock` will always return `nullptr`, because that's called from the parent context destructor or after it was destructed (see previous point).
- Storage for the interrupt guard condition is managed by [context](https://github.com/ros2/rclcpp/blob/a95efa452e0c6dcc4fd6854d36cb800a8e49cbfe/rclcpp/include/rclcpp/context.hpp#L374).
- Guard condition "finalizer" is called not by context, but by who registered the guard condition (e.g. the GraphListener). See [here](https://github.com/ros2/rclcpp/blob/a95efa452e0c6dcc4fd6854d36cb800a8e49cbfe/rclcpp/src/rclcpp/graph_listener.cpp#L376).

Based on the above points, there are two possibilities:

1.
```
---- context destructor
    ---- GraphListener shared counter is decremented
        ---- No other owner, GraphListener is destroyed.
            ---- GraphListener fails to lock context weak ptr (which will always happen)
                ---- GraphListener calls guard condition finalizer, ok because context is still "alive" (at least, its member destructors haven't been called)
```

2.
```
---- context destructor
    ---- GraphListener shared counter is decremented
        ---- GraphListener not destroyed (other owners).
---- GraphListener shared counter decremented and reaches 0.
     ---- GraphListener is destroyed.
         ---- GraphListener fails to lock context weak ptr (which will always happen)
             ---- GraphListener calls guard condition finalizer. This is not ok, because we're accessing "freed" memory.
```

In "real code", most of the time situation 1. happens, and we don't get a segfault.
That doesn't mean that 2. cannot happen, if anybody is retaining ownership of the `GraphListener` when context is being destroyed, it can absolutely happen.

Tests make the failure easier to reproduce, as it's not creating the GraphListener as a subcontext.

---

I'm not sure what is the right solution, but we have to modify ownership relations somehow, if not it's impossible to "synchronize" `GraphListener` and `Context` finalization correctly now.

IMO, having a `GuardCondition` class that retains ownership only of the `rcl_context` will break the reference cycle and make things easier.
@brawner I can take a look and implement that approach if that sounds ok to you.

brawner:
Please take the reins. I've let this lag long enough.

ivanpauno:
I needed to force push because of rebasing, previous code is still accessible [here](https://github.com/ros2/rclcpp/commit/0f6b928ff8a0bd05c3fcff9eed7abdb22dccd58e).

ivanpauno:
> IMO, having a GuardCondition class that retains ownership only of the rcl_context will break the reference cycle and make things easier.

I couldn't finally do that, because there's already a graph listener class retaining shared ownership of its parent `rclcpp::Context` (see [here](https://github.com/ros2/rclcpp/blob/7d257177e06317e9d30952d9653e59041e87199d/rclcpp/include/rclcpp/guard_condition.hpp#L58-L59)).

I broke the reference cycle by holding `rcl_context_t` directly in `GraphListener`.
I also realized that the `shutdown_guard_condition` was unnecessary, the `interrupt_guard_condition` can be used for the same purpose.
Finally, there was a shutdown callback installed in the global context instead of the parent context, I also fixed that.

ivanpauno:
@brawner I cannot add you as a reviewer (as you opened the PR), but it would be great if you can take a look too.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13043/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7991/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10767/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13024/)

brawner:
Thanks, I'll look it over later today.

brawner:
Building windows debug to see if this resolves the ongoing bugs:

[Build Status](https://ci.ros2.org/job/ci_windows/13026/)

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -40,30 +40,80 @@ using rclcpp::Executor;
 using rclcpp::ExecutorOptions;
 using rclcpp::FutureReturnCode;
 
+namespace
+{
+
+std::shared_ptr<rcl_guard_condition_t>
+create_shutdown_guard_condition(std::shared_ptr<rclcpp::Context> context)
```
brawner, line 47: In order to maintain the same level of coverage, I think the failure cases in this function will need to be covered with new mock tests.

brawner (commented):
Shoot, I didn't realize the job was `--packages-above rclcpp`. I'll have to look at it again in the morning.

If you need some help figuring out what's covered and what's not feel free to ping me. Otherwise, the developer's guide has a section on how to run coverage locally.

rclcpp/src/rclcpp/executor.cpp
ivanpauno: Used `rclcpp::GuardCondition`, which already has 100% test coverage.

ivanpauno: See https://github.com/ros2/rclcpp/pull/1404/commits/564993c3e2920e4205fae2118d545de34bf07f0f

ivanpauno:
> If you need some help figuring out what's covered and what's not feel free to ping me.

After https://github.com/ros2/rclcpp/pull/1404/commits/564993c3e2920e4205fae2118d545de34bf07f0f I think coverage is going to be as good as before. I can run the coverage tests to check if that's the case.

> Otherwise, the developer's guide has a section on how to run coverage locally.

The only thing I found is https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#code-coverage, which doesn't seem to explain how to run it locally.
Is there information somewhere else?

ivanpauno:
Coverage in master:

Starting >>> rclcpp  
Reading tracefile /home/ivanpauno/ros2_ws/build/rclcpp/coverage.info
Summary coverage rate: [rclcpp - 13.1s]
  lines......: 95.3% (17173 of 18020 lines)
  functions..: 79.1% (7536 of 9532 functions)
  branches...: 30.1% (23911 of 79494 branches)
Reading tracefile /home/ivanpauno/ros2_ws/build/rclcpp/coverage.info
Summary coverage rate: [rclcpp - 15.4s]
  lines......: 83.5% (17173 of 20565 lines)
  functions..: 76.3% (7536 of 9880 functions)
  branches...: 30.1% (23911 of 79494 branches)
Finished <<< rclcpp [15.5s]

Coverage in this branch:

  lines......: 95.3% (17093 of 17933 lines)
  functions..: 79.0% (7497 of 9489 functions)
  branches...: 30.1% (23783 of 79016 branches)
Reading tracefile /home/ivanpauno/ros2_ws/build/rclcpp/coverage.info
Summary coverage rate: [rclcpp - 14.7s]
  lines......: 80.9% (17093 of 21125 lines)
  functions..: 74.9% (7497 of 10011 functions)
  branches...: 30.1% (23783 of 79016 branches)
Finished <<< rclcpp [14.8s]

Looks ok to me

brawner:
It's tough for me to look at the summary of rclcpp and notice if a few branches are missing.

I guess I meant the ros2 onboarding guide (not developer's guide). https://index.ros.org/doc/ros2/Contributing/ROS-2-On-boarding-Guide/#how-to-measure-coverage-locally-using-lcov-ubuntu

It will produce an html report with the lines that are missing. I'm only concerned because you removed many tests that checked for failure conditions, but didn't introduce any.

Looking over the code, I don't see major chunks that would be missed. Maybe a line here or there.

rclcpp/test/rclcpp/test_graph_listener.cpp
```diff
@@ -170,26 +172,6 @@ TEST_F(TestGraphListener, error_run_graph_listener_mock_wait_set_add_guard_condi
     std::runtime_error("failed to add interrupt guard condition to wait set: error not set"));
 }
 
-TEST_F(TestGraphListener, error_run_graph_listener_mock_wait_set_add_guard_condition_twice) {
```
ivanpauno, line 173: this one was removed because it only failed when a second wait condition is added to the waitset, but the graphlistener is now only using one

rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -131,59 +131,6 @@ MOCKING_UTILS_BOOL_OPERATOR_RETURNS_FALSE(rcl_guard_condition_options_t, !=)
 MOCKING_UTILS_BOOL_OPERATOR_RETURNS_FALSE(rcl_guard_condition_options_t, >)
 MOCKING_UTILS_BOOL_OPERATOR_RETURNS_FALSE(rcl_guard_condition_options_t, <)
 
-TEST(TestUtilities, test_context_release_interrupt_guard_condition) {
```
ivanpauno, line 134: this was removed because the concept of interrupt guard condition was removed from Context

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -42,28 +43,35 @@ using rclcpp::FutureReturnCode;
...
   if (RCL_RET_OK != ret) {
     throw_from_rcl_error(ret, "Failed to create interrupt guard condition in Executor constructor");
   }
 
+  context_->on_shutdown(
+    [weak_gc = std::weak_ptr<rclcpp::GuardCondition>{shutdown_guard_condition_}]() {
+      auto strong_gc = weak_gc.lock();
+      if (strong_gc) {
+        strong_gc->trigger();
+      }
```
ivanpauno, line 64: I'm not sure if this branch is being covered, I will double check

ivanpauno (commented):
@brawner I've left some comments explaining why the test cases were removed.
I will double check the html report.

rclcpp/test/rclcpp/test_graph_listener.cpp
ivanpauno: By the way, I don't like this kind of test that depends on an implementation detail (a guard condition being added twice to a waitset in the graph listener loop is not part of the API contract).
I'm not sure how to have 95% line coverage while avoiding this kind of tests though.

brawner:
Removing the tests absolutely makes sense, and it also looks like there are fewer failure conditions to check, which is also a great thing. My issue wasn't that tests were removed, only that none were added back in. Which unless the conditions are hard to trigger even with the mocking tools, it's probably worth adding a test for those conditions.

brawner:
Opened https://github.com/ros2/rclcpp/issues/1456 for the new win debug failure.

rclcpp/src/rclcpp/executor.cpp
ivanpauno: it's being covered

rclcpp/src/rclcpp/graph_listener.cpp
```diff
@@ -224,7 +203,7 @@ GraphListener::run_loop()
       if (graph_gc == wait_set_.guard_conditions[graph_gc_indexes[i]]) {
         node_ptr->notify_graph_change();
       }
-      if (shutdown_guard_condition_triggered) {
+      if (is_shutdown_) {
```
ivanpauno, line 206: this branch is being covered

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -42,28 +43,35 @@ using rclcpp::FutureReturnCode;
...
 {
+  // Store the context for later use.
+  context_ = options.context;
+
   rcl_guard_condition_options_t guard_condition_options = rcl_guard_condition_get_default_options();
   rcl_ret_t ret = rcl_guard_condition_init(
-    &interrupt_guard_condition_, options.context->get_rcl_context().get(), guard_condition_options);
+    &interrupt_guard_condition_, context_->get_rcl_context().get(), guard_condition_options);
   if (RCL_RET_OK != ret) {
     throw_from_rcl_error(ret, "Failed to create interrupt guard condition in Executor constructor");
```
ivanpauno, line 56: This line isn't hit anymore.
The reason is that now we're creating two guard conditions, instead of one, and we're hitting the error in the first creation and not here.

ivanpauno (commented):
The overall coverage of rclcpp remains the same.

The coverage in executor.cpp improves, because of deleting many uncovered lines.
There's one added line that's not hit in test.

The coverage in graph_listener.cpp is also improved, and all added lines are hit.

The coverage in context.cpp decreases, but we didn't add any line, and previous lines are still covered.
The only reason that the percentage decreased, is that the lines removed had better coverage that the ones that weren't removed (i.e. unrelated to the PR).

rclcpp/src/rclcpp/executor.cpp
ivanpauno: @brawner Any suggestion to cover this line without adding a mock that fails in the second call to `rcl_guard_condition_init`?
I can apply the mentioned option, but IMO that's a bit "strange".

Maybe fault injection works for this, I will try to give it a try.

P.S.: I don't think that hitting this line matters, as overall coverage has improved.

clalancette: > P.S.: I don't think that hitting this line matters, as overall coverage has improved.

Yeah, that's my opinion too.  If overall coverage is improving, then its OK to lose a line like this.  If it becomes important, we can always revisit it later.

brawner: The fault injection tools were intended for situations like this, but I think it's fine without.

brawner:
I think this looks good to me now. @ivanpauno did you want to give it one more round of CI testing after https://github.com/ros2/rclcpp/pull/1404/commits/564993c3e2920e4205fae2118d545de34bf07f0f?

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8015/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10792/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13049/)

ivanpauno:
This needs a reviewer that hasn't committed changes here, maybe @wjwwood @mabelzhang ?

wjwwood (approved):
lgtm, with a nitpick

rclcpp/src/rclcpp/graph_listener.cpp
```diff
@@ -33,28 +33,27 @@ using rclcpp::exceptions::throw_from_rcl_error;
 
 namespace rclcpp
 {
+
```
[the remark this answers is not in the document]
ivanpauno:
```suggestion
```

brawner:
Thanks again Ivan for taking over this PR!

azeey:
I believe this PR has caused the following regressions on windows [nightlies](https://ci.ros2.org/job/nightly_win_rel/1758/testReport/):

- test_communication.TestActionClientServer.test_client_finishes_in_a_finite_amount_of_time[Fibonacci]
- test_communication.TestActionClientServerAfterShutdown.test_processes_finished_gracefully[Fibonacci]
- test_communication.TestActionClientServer.test_client_finishes_in_a_finite_amount_of_time[NestedMessage]
- test_communication.TestActionClientServerAfterShutdown.test_processes_finished_gracefully[NestedMessage]

I have ran windows CI at the following commits to figure out which PR is the likely cause:
* PR 1459 (27d1b11647adaa7933c00d0c4a399a783fa09b9f) (https://ci.ros2.org/job/ci_windows/13120/)
* PR 1241 (8c8c268aad3e243ddb7e2717b5c46044db34e055) (https://ci.ros2.org/job/ci_windows/13146/)
* PR 1465 (71a58d40f19160b1d1ba8f1f408702acaea94fde) (https://ci.ros2.org/job/ci_windows/13148/)
* PR 1404 (cba6f20988758128377751de85dbc8351aeebefc) (https://ci.ros2.org/job/ci_windows/13153/). The regression only happens here.

brawner:
Thanks @azeey  for bringing this to our attention. I opened https://github.com/ros2/system_tests/issues/452 to track discussion. I'm not sure this one is obvious. It only happens with rmw_cyclonedds_cpp, an rclpp client and an rclpy server (I believe that's how you parse the test name) on Windows.

PR #1408, opened 2020-10-15
Qos configurability

ivanpauno:
Prototype of https://github.com/ros2/design/pull/300.

Depends on https://github.com/ros2/rclcpp/pull/1410.
Diff omitting #1410: https://github.com/ros2/rclcpp/pull/1408/files/ec90cb53091e355dcf194c944fcaf239c5fa6699..9197a8f2bfe9da6e54f3e30779b3b1c0236339f2.

---
Not quite ready yet.
Things to solve before merging:

1. Get the final topic name before creating a publisher/subscription
   - [x] [rcl support](https://github.com/ros2/rcl/pull/835#issue-505063024) for expand and remap.
   - [x] Wrap the rcl function in an `rclcpp` method for `Node` ([PR](https://github.com/ros2/rclcpp/pull/1410)).
1. `create_subscription` and `create_publisher` can be called without a parameters interface

   I'm not quite sure what to do about this.
   If the user wants to only provide a topics interface (without using qos overrides), that should be ok.
   Maybe fail when qos overrides are enabled and a parameters interface isn't provided?

     - [x] Add create_publisher/create_subscription overload taking both a topics interface and a parameters interface (https://github.com/ros2/rclcpp/pull/1408/commits/4d6ad65be53a2beb1f4a7a48d27a07cb5ca97f11).
     - [x] Handle error when qos overrides are enabled and a parameters interface isn't provided: A warning is being logged in this case.
     - [ ] See places that need special handling in `rclcpp` code in `create_publisher.hpp`/`create_subscription.hpp` comments.
       Added notes, this can be resolved in a follow-up.
1. Improve QosOverridingOptions API

   Not a blocker for merging this, but we should have minimal API that won't be deprecated.
   Maybe QosOverridingOptions could use the builder pattern or the parameter idiom to make things easier.
   I definitely also want either builder pattern/parameters idiom in `PublisherOptions`/`SubscriptionOptions`, but that can be done in a follow up.
1. Test coverage

   - [x] Improve test coverage (note: this can still be improved)
1. Move things that can be shared between rclcpp/rclpy to rcl/rmw

   - [x] Conversions between policy values and strings ([PR](https://github.com/ros2/rmw/pull/284)).
   - [x] Extend [rmw_qos_policy_kind_t](https://github.com/ros2/rmw/blob/e02978028f1756cc4c057d06f462ceac8acafcab/rmw/include/rmw/events_statuses/incompatible_qos.h#L28) to cover all possible policies (and move that to a more appropriate header). Define `rclcpp::QosPolicyKind` values based on that. Conversions to/from strings can be implemented there ([PR](https://github.com/ros2/rmw/pull/285)).
   - [x] Use the new rmw functions in this PR (https://github.com/ros2/rclcpp/pull/1408/commits/77eec6d5d703182a624f7b0656e4de16c924b8c4).
1. Support in other client libraries

   - [X] Support in `rclpy` ([PR](https://github.com/ros2/rclpy/pull/635))


ivanpauno:
FYI @jacobperron

ivanpauno:
I was thinking that if `QosOverridingOptions` is an argument independent of `PublisherOptions`/`SubscriptionOptions`, this [runtime check](https://github.com/ros2/rclcpp/blob/9197a8f2bfe9da6e54f3e30779b3b1c0236339f2/rclcpp/include/rclcpp/detail/qos_parameters.hpp#L172-L174) can we avoided (and instead not providing a parameters interface while passing `QosOverridingOptions` can be detected at build time).

@wjwwood @jacobperron opinions?

ivanpauno:
I still have to write a bunch of tests before merging, but this is otherwise ready for review.

jacobperron (commented):
> I was thinking that if `QosOverridingOptions` is an argument separated of `PublisherOptions`/`SubscriptionOptions`, this [runtime check](https://github.com/ros2/rclcpp/blob/9197a8f2bfe9da6e54f3e30779b3b1c0236339f2/rclcpp/include/rclcpp/detail/qos_parameters.hpp#L172-L174) can we avoided

I'm not sure I follow exactly. It what situation will the node not have a parameters interface? Does this mean that someone has created a Node-like object that doesn't implement that interface? I think I prefer bundling the QoS overrides with the pub/sub options to keep the API tidy, but I don't have a strong preference.

ivanpauno:
> I'm not sure I follow exactly. It what situation will the node not have a parameters interface? Does this mean that someone has created a Node-like object that doesn't implement that interface? I think I prefer bundling the QoS overrides with the pub/sub options to keep the API tidy, but I don't have a strong preference.

You can currently do the following:

```cpp
rclcpp::create_publisher(node_topics_interface, "my/topic/name", qos, options);
//                              ^^^^^^^
//                                ||------------------Type `rclcpp::node_interfaces::NodeTopicsInterface`
```
(same for `create_subscription`)
Here [an example](https://github.com/ros2/rclcpp/blob/5d9db5de748e936c102352121f7910c7d64ba2ad/rclcpp/src/rclcpp/time_source.cpp#L236).

In that case, `NodeTopicsInterface` doesn't provide a `NodeParametersInterface`.

The current approach is that a warning is logged if you wanted to generate qos parameters and didn't provide a parameters interface, but I would prefer if the error is detected at compile time.
If `create_publisher` takes `QosOverridingOptions` as a separated argument, that can be achieved.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -233,11 +234,17 @@ void TimeSource::create_clock_sub()
     return;
   }
 
+  using rclcpp::QosPolicyKind;
+  rclcpp::SubscriptionOptions options;
+  options.qos_overriding_options = QosOverridingOptions{
+    QosPolicyKind::Depth, QosPolicyKind::History, QosPolicyKind::LivelinessLeaseDuration,
+    QosPolicyKind::Reliability};
```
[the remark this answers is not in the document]
ivanpauno: The only reason I didn't add durability here is that the publisher in `gazebo_ros_packages` isn't offering transient local durability and that isn't currently reconfigurable.

> If we are going ahead an setting options for the /clock topic here, I think we should add rationale to the design doc.

I think that what QoS should be reconfigurable for each topic can be discussed in a case by case fashion, without adding that to the design doc.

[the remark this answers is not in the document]
jacobperron: It might be nice to document QoS options for common topics so that they can match in multiple client libraries; i.e. I presume we want rclcpp, rclpy, etc to have the same settings.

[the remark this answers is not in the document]
ivanpauno: > I presume we want rclcpp, rclpy, etc to have the same settings.

That's a fair point.

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,298 @@
...
+  if (options.policy_kinds.size()) {
+    RCLCPP_WARN(
+      rclcpp::get_logger("rclcpp"),
+      "qos override options ignored because no parameter interface was provided");
+  }
+  return default_qos;
+}
+
+/// \internal Helper function to get a rmw qos policy value from a string.
+#define RCLCPP_DETAIL_QOS_POLICY_FROM_PARAMETER_STRING( \
```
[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1408/commits/f1ac90fc031a3fd0816438bfeb3f2de0bc592787.

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,298 @@
...
+    std::ostringstream oss{"unknown ", std::ios::ate};
+    oss << kind << " qos policy value: {" << policy_value_stringified << "}";
+    throw std::invalid_argument{oss.str()};
+  }
+  return policy_value_stringified;
+}
+
+inline
+::rclcpp::ParameterValue
+get_default_qos_param_value(rclcpp::QosPolicyKind qpk, const rclcpp::QoS & qos)
```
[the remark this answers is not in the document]
ivanpauno: Yeah, I also don't like abbreviated names, but I was being too lazy :joy:.
Fixed in https://github.com/ros2/rclcpp/pull/1408/commits/b8ca1de029b0bb711cd2de1312d7239ca9fef260.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -67,6 +58,19 @@
 #include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+#include "rclcpp/detail/qos_parameters.hpp"
+
+#include "rclcpp/node_interfaces/node_base_interface.hpp"
```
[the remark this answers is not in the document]
ivanpauno: Agreed, fixed in https://github.com/ros2/rclcpp/pull/1408/commits/3310c759921625336dbe80011b1bd558007dc76a.

rclcpp/test/rclcpp/detail/test_qos_parameters.cpp
```diff
@@ -0,0 +1,79 @@
...
+    ).get_value<std::string>(),
+    "best_effort");
+  EXPECT_EQ(RMW_QOS_POLICY_HISTORY_KEEP_LAST, qos.get_rmw_qos_profile().history);
+  EXPECT_EQ(RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT, qos.get_rmw_qos_profile().reliability);
+  EXPECT_EQ(10u, qos.get_rmw_qos_profile().depth);
+
+  std::map<std::string, rclcpp::Parameter> qos_params;
+  EXPECT_TRUE(
+    node->get_node_parameters_interface()->get_parameters_by_prefix(
+      "qos_overrides./my/fully/qualified/topic_name.publisher", qos_params));
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rclcpp/pull/1408/commits/074f3b475fb8dbf809090c9f13b87258cbf9fd29.

rclcpp/test/rclcpp/detail/test_qos_parameters.cpp
```diff
@@ -0,0 +1,79 @@
...
+  EXPECT_EQ(3u, qos_params.size());
+
+  rclcpp::shutdown();
+}
```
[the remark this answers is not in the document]
ivanpauno: Added tests in https://github.com/ros2/rclcpp/pull/1408/commits/64cd3801b15ba6f06e3649ad3e9f08b619535144.

rclcpp/src/rclcpp/time_source.cpp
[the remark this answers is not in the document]
ivanpauno: I have deleted that change, so that can be discussed in a separated PR: https://github.com/ros2/rclcpp/pull/1408/commits/7bb8462b3c93967413163c5dd7aef1a3b12f0aa9.

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,303 @@
...
+      std::count(options.policy_kinds.begin(), options.policy_kinds.end(), policy))
+    {
+      std::ostringstream param_name{param_prefix, std::ios::ate};
+      param_name << qos_policy_kind_to_cstr(policy);
+      std::ostringstream param_desciption{"qos policy {", std::ios::ate};
+      param_desciption << qos_policy_kind_to_cstr(policy) << param_description_suffix;
+      rcl_interfaces::msg::ParameterDescriptor descriptor{};
+      descriptor.description = param_desciption.str();
+      descriptor.read_only = true;
+      auto value = parameters_interface.declare_parameter(
```
ivanpauno, line 149: I noticed that this can cause a [parameter already declare](https://github.com/ros2/rclcpp/blob/371074523adfeeb69479266a044a7a8a34712efa/rclcpp/include/rclcpp/exceptions/exceptions.hpp#L258) error.

When creating two subscriptions in the same topic, the same qos might be desired for both (i.e. not passing an id to disambiguate) , and in that case the second creation will fail.
Maybe the `QosOverridingOptions` should have a "parameters already declared" option, that uses `get_parameter` instead of `declare_parameter`.

"/clock" is an special case, because we can create many time sources with different subscriptions, and we don't know what's the first one (I think that never happens in a real use case, but it's happening in tests).
[Here](https://github.com/ros2/rclcpp/blob/371074523adfeeb69479266a044a7a8a34712efa/rclcpp/src/rclcpp/time_source.cpp#L85-L92) is how we handle that for `use_sim_time` (by the way, there's a TOCTTOU race condition there).
I'm not sure what's the right way to handle this last case, @jacobperron @wjwwood any ideas?

ivanpauno:
Force pushed to resolve conflicts with master.

jacobperron:
> The current approach is that a warning is logged if you wanted to generate qos parameters and didn't provide a parameters interface, but I would prefer if the error is detected at compile time.
If create_publisher takes QosOverridingOptions as a separated argument, that can be achieved.

Creating a separate arg for QosOverridingOptions find to me. I don't have a strong feeling either way.

rclcpp/include/rclcpp/detail/qos_parameters.hpp
jacobperron: What about wrapping this in a try-catch and if we catch a `ParameterAlreadyDeclaredException` then use `get_parameter`?

jacobperron: Perhaps the race you found can be fixed in a similar way.

ivanpauno: > What about wrapping this in a try-catch and if we catch a ParameterAlreadyDeclaredException then use get_parameter?

That sounds ok to me, but maybe in the general case it should be an error (the `/clock` case sounds pretty exceptional).

> Perhaps the race you found can be fixed in a similar way.

:+1:

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -107,15 +84,18 @@ create_subscription(
               " ms");
     }
 
-    std::shared_ptr<Publisher<statistics_msgs::msg::MetricsMessage>> publisher =
-      create_publisher<statistics_msgs::msg::MetricsMessage>(
-      node,
+    // TODO(ivanpauno): This could have topics statistics enabled, but I'm not sure if it makes
```
jacobperron, line 87: I'm confused. Do you mean the topic statistics publisher could have QoS overrides enabled?

I'm not sure. I think it should be discussed in the design doc ultimately and we can revisit it if/when it is documented there.

rclcpp/include/rclcpp/qos_overriding_options.hpp
```diff
@@ -0,0 +1,105 @@
...
+namespace detail
+{
+// forward declare
+template<typename T>
+class QosParameters;
+}
+
+/// Options that are passed in subscription/publisher constructor to specify QoSConfigurability.
+/**
+ * TODO: Write nice docs here.
```
jacobperron, line 63: Let's resolve this :)

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,303 @@
...
+::rclcpp::ParameterValue
+get_default_qos_param_value(rclcpp::QosPolicyKind policy, const rclcpp::QoS & qos);
+
+/// \internal Modify the given `policy` in `qos` to be `value`.
+inline
+void
+apply_qos_override(
+  rclcpp::QosPolicyKind policy, rclcpp::ParameterValue value, rclcpp::QoS & qos);
+
+/// \internal Declare qos parameters for the given entity.
```
jacobperron, line 91: nit
```suggestion
/// \internal Declare QoS parameters for the given entity.
```

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,303 @@
...
+/**
+ * \tparam NodeT Node pointer or reference type.
+ * \tparam EntityQosParametersTraits A class with two static methods: `entity_type()` and
+ *  `allowed_policies()`. See `PublisherQosParametersTraits` and `SubscriptionQosParametersTraits`.
+ * \param options User provided options that indicate if qos parameter overrides should be
```
jacobperron, line 96:
```suggestion
 * \param options User provided options that indicate if QoS parameter overrides should be
```

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+/// \internal Helper function to get a rmw qos policy value from a string.
+#define RCLCPP_DETAIL_APPLY_QOS_OVERRIDE_FROM_PARAMETER_STRING( \
+    kind_lower, kind_upper, parameter_value, rclcpp_qos) \
+  do { \
+    auto policy_string = (parameter_value).get<std::string>(); \
+    auto policy_value = rmw_qos_ ## kind_lower ## _policy_from_str(policy_string.c_str()); \
+    if (RMW_QOS_POLICY_ ## kind_upper ## _UNKNOWN == policy_value) { \
+      throw std::invalid_argument{"unknown qos policy " #kind_lower " value: " + policy_string}; \
```
jacobperron, line 191:
```suggestion
      throw std::invalid_argument{"unknown QoS policy " #kind_lower " value: " + policy_string}; \
```

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+/// \internal Throw an exception if `policy_value_stringified` is NULL.
+inline
+const char *
+check_if_stringified_policy_is_null(const char * policy_value_stringified, QosPolicyKind kind)
+{
+  if (!policy_value_stringified) {
+    std::ostringstream oss{"unknown ", std::ios::ate};
+    oss << kind << " qos policy value: {" << policy_value_stringified << "}";
```
jacobperron, line 256:
```suggestion
    oss << kind << " QoS policy value: {" << policy_value_stringified << "}";
```

rclcpp/include/rclcpp/detail/qos_parameters.hpp
```diff
@@ -0,0 +1,303 @@
...
+        check_if_stringified_policy_is_null(
+          rmw_qos_liveliness_policy_to_str(rmw_qos.liveliness), kind));
+    case QosPolicyKind::LivelinessLeaseDuration:
+      return ParameterValue(rmw_duration_to_int64_t(rmw_qos.liveliness_lease_duration));
+    case QosPolicyKind::Reliability:
+      return ParameterValue(
+        check_if_stringified_policy_is_null(
+          rmw_qos_reliability_policy_to_str(rmw_qos.reliability), kind));
+    default:
+      throw std::invalid_argument{"unknown qos policy kind"};
```
jacobperron, line 296:
```suggestion
      throw std::invalid_argument{"unknown QoS policy kind"};
```

rclcpp/src/rclcpp/qos_overriding_options.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace rclcpp
+{
+
+const char *
+qos_policy_kind_to_cstr(const QosPolicyKind & qpk)
+{
+  const char * ret = rmw_qos_policy_kind_to_str(static_cast<rmw_qos_policy_kind_t>(qpk));
+  if (!ret) {
+    throw std::invalid_argument{"unknown qos policy kind"};
```
jacobperron, line 34:
```suggestion
    throw std::invalid_argument{"unknown QoS policy kind"};
```

jacobperron (commented):
Left some more minor comments, but LGTM overall.

I think it would be useful to have an example (perhaps in the ros2/demos repository) demonstrating how to use this feature. Maybe just a modified talker demo that has some QoS policies that are configurable and exercises the callback feature; what do you think?

Regarding changing `QosOverridingOptions` to be a separate argument from the publisher/subscription options, I don't have a strong opinion. It would be good to hear from @mabelzhang and/or @wjwwood.

rclcpp/include/rclcpp/detail/qos_parameters.hpp
jacobperron: I guess we still need to resolve this. In the general case, how should users workaround this issue?

IIUC, the situation arise if I do something like this:

```c
...
 rclcpp::QosOverridingOptions qos_overrides{rclcpp::QosOverridingOptions::kDefaultPolicies};
 auto sub_options = rclcpp::SubscriptionOptions();
 sub_options.qos_overriding_options = qos_overrides;
 auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", foo_qos, foo_callback, sub_options);
 auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", bar_qos, bar_callback, sub_options);
...
```
At first glance, I would expect things to work without issue; the user shouldn't have to worry about parameters already being declared since it's an implementation detail. But, then it looks a little confusing since `sub1` and `sub2` have different "default" QoS settings, but the same override abilities. And configuring one QoS policy should modify both. It seems a bit strange, and is probably a exceptional use-case, but seems like something we could easily support.

Correct me if I'm wrong or missing something.

ivanpauno: mmm, the default qos of the second case will be ignored, so I don't think allowing that is a good idea.
We definitely need to support this somehow.

Maybe we should provide user facing `declare_subscription_qos_parameter_overrides` and `get_subscription_qos_parameter_overrides` functions:

```cpp
rclcpp::QosOverridingOptions qos_overrides{rclcpp::QosOverridingOptions::kDefaultPolicies};
rclcpp::QoS qos{KeepLast(10)};
declare_subscription_qos_parameter_overrides(
  qos_overrides, node_parameters, node_topics->resolve_topic_name(topic), qos);  # qos is inout here
auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", qos, foo_callback);
```
 ... in another part of the code later ...
```cpp
rclcpp::QoS qos = get_subscription_qos_from_overrides(
  node_parameters, node_topics->resolve_topic_name(topic));  // qos overriding options aren't needed here
auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", qos, bar_callback, sub_options);
```

The function for declaring the parameters is not strictly necessary, as it can be done in a `create_subscription` call passing qos overriding options. But in some cases, like `/clock`, I think it will be useful.

rclcpp/include/rclcpp/create_subscription.hpp
ivanpauno: Deleted the comment in https://github.com/ros2/rclcpp/pull/1408/commits/f1062cb78d12b04f5893776284a07bd01d0e9c1c.

> I'm confused.

I'm reading the comment and I'm confused too :joy:.
The comment was about being able to reconfigure topic statistics qos, but we can discuss where to enable qos reconfigurability in a follow-up PR.

rclcpp/include/rclcpp/qos_overriding_options.hpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1408/commits/2707b35d86d6de717037c5f68c92ba4da0ef5026.

rclcpp/include/rclcpp/detail/qos_parameters.hpp
mjeronimo: I don't think the user should have to worry about declaring QoS parameters in this context. As Jacob mentioned, this should be a implementation concern. In Navigation 2 we used internally a declare_parameter_if_not_declared function, as this came up in the context of plugins declaring the same parameter: https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/node_utils.hpp#L83-L94.

I would also strive to keep the semantics very clear. In the case of the following example,

```
 rclcpp::QosOverridingOptions qos_overrides{rclcpp::QosOverridingOptions::kDefaultPolicies};

 auto sub_options = rclcpp::SubscriptionOptions();
 sub_options.qos_overriding_options = qos_overrides;

 auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", foo_qos, foo_callback, sub_options);

 auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", bar_qos, bar_callback, sub_options);
```

I would interpret this to mean that the first subscription is using any QoS settings from foo_qos and then overriding those with any settings in sub_options.qos_overriding_options. Similarly, for the second subscription, it will use bar_qos and then override any QoS settings from sub_options.qos_overriding_options. Assuming this is the intent, it is clear what to expect.

In contrast, the second example, which has this:

```
rclcpp::QosOverridingOptions qos_overrides{rclcpp::QosOverridingOptions::kDefaultPolicies};
rclcpp::QoS qos{KeepLast(10)};
declare_subscription_qos_parameter_overrides(
  qos_overrides, node_parameters, node_topics->resolve_topic_name(topic), qos);  # qos is inout here
auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", qos, foo_callback);
```
Then this:

```
rclcpp::QoS qos = get_subscription_qos_from_overrides(
  node_parameters, node_topics->resolve_topic_name(topic));  // qos overriding options aren't needed here
auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", qos, bar_callback, sub_options);
```
(Assuming I've understood correctly) requires the user to declare the QoS parameters, uses separate functions depending on whether it is the first call to get overrides or not, and also seems to have the side effect of setting the QoS state so that get_subscription_qos_from_overrides can apply it. IMO, all of this is more complicated than the first version.

mjeronimo: Regarding the approach in the current code..

```
 rclcpp::QoS qos{rclcpp::KeepLast(10)};
  qos = rclcpp::detail::declare_qos_parameters(
    rclcpp::QosOverridingOptions::kDefaultPolicies,
    node,
    "/my/fully/qualified/topic_name",
    qos,
    rclcpp::detail::PublisherQosParametersTraits{});

  EXPECT_EQ(
    node->get_parameter(
      "qos_overrides./my/fully/qualified/topic_name.publisher.history").get_value<std::string>(),
    "keep_last");

```
Do we want to expose that this QoS override functionality is implemented using parameters? You could abstract this and expose only the ability to set (and get) QoS override values for a topic name and the ability to apply these overrides to a QoS policy.

ivanpauno: > I would interpret this to mean that the first subscription is using any QoS settings from foo_qos and then overriding those with any settings in sub_options.qos_overriding_options. Similarly, for the second subscription, it will use bar_qos and then override any QoS settings from sub_options.qos_overriding_options. Assuming this is the intent, it is clear what to expect.

The issue is the following:

```cpp
 auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", foo_qos, foo_callback, sub_options);
```

supposing that `reliability` is configurable, this will declare a parameter named `qos_overrides./same_topic.subscription.reliability` with a default value of `foo_qos.reliability`.

```cpp
 auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", bar_qos, bar_callback, sub_options);
```

This second call will not declare a parameter (as it was already declared before), and `bar_qos.reliability` will be completely ignored and overridden with the value got from the parameter.

I prefer asymmetries being explicit and not implicit, and the API proposed above makes the asymmetry explicit.

We could also have some extra sugar in the `QosOverridingOptions` struct, like an option to indicate if declare parameters or to get the previously declared ones:

```cpp
 rclcpp::QosOverridingOptions qos_overrides{rclcpp::QosOverridingOptions::kDefaultPolicies};

 auto sub_options = rclcpp::SubscriptionOptions();
 sub_options.qos_overriding_options = qos_overrides;

 auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", foo_qos, foo_callback, sub_options);

sub_options.qos_overriding_options.declare_parameters = false;  // the second construction will get the already declared parameters instead of declaring again.

 auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
    node, "same_topic", bar_qos, bar_callback, sub_options);
```

ivanpauno: > requires the user to declare the QoS parameters, uses separate functions depending on whether it is the first call to get overrides or not, and also seems to have the side effect of setting the QoS state so that get_subscription_qos_from_overrides can apply it

Ah, that's a good point.
An input qos parameters profile would still be needed in `get_subscription_qos_from_overrides` (as having an extra qos state doesn't seem like a good idea).

```cpp
rclcpp::QoS qos{KeepLast{10}};
get_subscription_qos_from_overrides(
  node_parameters, node_topics->resolve_topic_name(topic), qos);
```

ivanpauno: > Do we want to expose that this QoS override functionality is implemented using parameters? I understood that it could be implemented using parameters, but didn't realize this would be part of the interface.

If we're using parameters, the user will always be able to use the "normal" parameters API to get them.
In this case they cannot set the parameter, as we're using the `read-only` option.

> You could abstract this and expose only the ability to set (and get) QoS override values for a topic name and the ability to apply these overrides to a QoS policy.

IIUC, that's the idea of the previously proposed `get_subscription_qos_from_overrides`, `declare_subscription_qos_parameter_overrides`.

jacobperron: IIUC, using the proposed `declare_subscription_qos_parameter_overrides` and `get_subscription_qos_parameter_overrides functions` functions is essentially the same as the user setting the same default QoS for both subscriptions. IMO we should encourage people to set the same defaults for sub/pubs with the same topic name, rather than adding additional functions for them to use. I think this is simpler and easy to document. If possible, it would be good to log a warning if we detect a different default QoS setting in the second subscription being created.

jacobperron: As I understand, even if we introduce the proposed functions we still rely on the user knowing about this pitfall. We should probably document this pitfall in the design doc and somewhere in the code API.

mjeronimo: Ok, I understand that using (read-only) parameters is going to be part of the interface to the user and not a hidden implementation detail. Given that, could the interface be simplified a bit in its operations? The user could perhaps:

- Explicitly declare (once) each parameter that can be overridden using `declare_subscription_qos_parameter_overrides`
- Update an existing QoS using `get_subscription_qos_from_overrides`
- Create the subscription using the update QoS

For example, the previous example:
```
rclcpp::QosOverridingOptions qos_overrides{rclcpp::QosOverridingOptions::kDefaultPolicies};
rclcpp::QoS qos{KeepLast(10)};
declare_subscription_qos_parameter_overrides(
  qos_overrides, node_parameters, node_topics->resolve_topic_name(topic), qos);  # qos is inout here
auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", qos, foo_callback);

rclcpp::QoS qos = get_subscription_qos_from_overrides(
  node_parameters, node_topics->resolve_topic_name(topic));  // qos overriding options aren't needed here
auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", qos, bar_callback, sub_options);
```
Would become:

```
// Declare each override once, which sets the value in the node parameters (without modifying a particular qos)
rclcpp::QosOverridingOptions qos_overrides{rclcpp::QosOverridingOptions::kDefaultPolicies};
declare_subscription_qos_parameter_overrides(
    qos_overrides, node_parameters, node_topics->resolve_topic_name(topic));

// Now, you can override a particular QoS setting, updating any qos fields from the parameters
rclcpp::QoS qos{KeepLast(10)};
rclcpp::QoS updated_qos = get_subscription_qos_from_overrides(
    qos, node_parameters, node_topics->resolve_topic_name(topic));

// Then, create a subscription
auto sub1 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", updated_qos, foo_callback);

// You can override another
rclcpp::QoS qos2{KeepLast(5)};
rclcpp::QoS updated_qos2 = get_subscription_qos_from_overrides(
    qos2, node_parameters, node_topics->resolve_topic_name(topic));

// And create another subscription
auto sub2 = rclcpp::create_subscription<test_msgs::msg::Empty>(
  node, "same_topic", updated_qos2, bar_callback, sub_options);
```

There are basically two operations needed: create a parameter for a QoS override and apply QoS overrides to a QoS structure.

ivanpauno: > As I understand, even if we introduce the proposed functions we still rely on the user knowing about this pitfall

It doesn't sound like a pitfall in `get_subscription_qos_from_overrides` (or with  `qos_overriding_options.declare_parameters = false`), as in that case the user would be explicitly requesting: "take this qos profile and always replace the policies that have a declared parameter with the parameter value".

> Ok, I understand that using (read-only) parameters is going to be part of the interface to the user and not a hidden implementation detail

Yeah, that was the result of some discussions in https://github.com/ros2/design/pull/296.

> Given that, could the interface be simplified a bit in its operations? The user could perhaps:

That sounds ok to me.
It's a bit too verbose in the case you want to only create one subscription/publisher in that topic, so I wanted to provide extra sugar on top of that to avoid the verbosity.

mjeronimo: True, it may be more lines of code, but I think the programming model of: Declare, Override, Create, Override, Create is simpler and easier to understand than DeclareAndOverride, Create, OverrideWithSavedQoS, Create

I'm assuming that this operation is returning a saved QoS from the initial DeclareAndOverride (could be wrong about that)

```
rclcpp::QoS qos = get_subscription_qos_from_overrides(
  node_parameters, node_topics->resolve_topic_name(topic));  // qos overriding options aren't needed here
```

jacobperron: Here's a straw-man example of what I think we should do here:

```c++
if (!parameters_interface.has_parameter()) {
  auto value = parameters_interface.declare_parameter(
    param_name.str(), get_default_qos_param_value(policy, qos), descriptor);
  ::rclcpp::detail::apply_qos_override(policy, value, qos);
} else {
  auto value = parameters_interface.get_parameter(param_name.str());
  // TODO: We would need to add the "is_default()" API, but it should be easy
  if (value.is_default() && value != get_default_qos_param_value(policy, qos)) {
    // Log or output an error here (optionally, we could throw an exception instead)
    std::cerr << "Default setting for QoS policy " << policy << " is being ignored. Previously overridden by another subscription." << std::endl; 
  }
  ::rclcpp::detail::apply_qos_override(policy, value, qos);
}
```

jacobperron: To clarify, I think it is a pitfall in the sense that the user must be aware that they can't write code like my example (https://github.com/ros2/rclcpp/pull/1408#discussion_r514514169), furthermore, they must also be aware of these additional functions to workaround the issue (which are essentially setting the default QoS profile to the same value as a previously declared sub/pub).

ivanpauno: https://github.com/ros2/rclcpp/pull/1408/commits/cab74d8d2de3238d87f9f0853004ae08cb1cee9f takes the approach of https://github.com/ros2/rclcpp/pull/1408#discussion_r517669584.

ivanpauno:
Rebased to fix Rpr checker

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12978/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7930/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10703/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12956/)

jacobperron (approved):

ivanpauno:
[Last commit](https://github.com/ros2/rclcpp/pull/1408/commits/22a541269fb2e03a6c1b080ed0a7659b2a35f559) modifies the return value of the validation callback, so when the profile is rejected the user gets a more insightful message.

jacobperron (approved):

ivanpauno:
Local coverage testing: new files have 100% line coverage, new lines in modified files are completely covered.

Last round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13058/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8005/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10781/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13039/)

rclcpp/test/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ find_package(test_msgs REQUIRED)
 
 include(cmake/rclcpp_add_build_failure_test.cmake)
 
-add_subdirectory(benchmark)
+# add_subdirectory(benchmark)
```
jacobperron, line 7: This change doesn't look like it was intended to be pushed?

rclcpp/test/rclcpp/test_create_subscription.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <chrono>
 #include <memory>
+#include <utility>
```
jacobperron, line 19: I don't see why this is needed, looking at the newly added code.

rclcpp/test/CMakeLists.txt
ivanpauno: yeah, I used that to not have problems when checking coverage.

ivanpauno:
```suggestion
add_subdirectory(benchmark)
```

rclcpp/test/rclcpp/test_create_subscription.cpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1408/commits/da0742316caad9a3de6f33cbe3b17d911b6ccad0.

jacobperron (approved):

wjwwood:
🎉 This has turned out to be a really cool feature, imo.

Issue #1403, opened 2020-10-13
test_subscription_topic_statistics fails with rmw_connext_cpp


brawner:
This is a reoccuring failure over at:
http://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/106/testReport/rclcpp/TestSubscriptionTopicStatisticsFixture/test_receive_stats_for_message_no_header/

It was probably introduced in https://github.com/ros2/rclcpp/pull/1281, which aligns with when this failure started showing up.

iuhilnehc-ynos:
I can reproduce this failure, refer to 
```
$ ros2 topic echo /statistics statistics_msgs/msg/MetricsMessage
measurement_source_name: test_sub_stats_node
metrics_source: message_age
unit: ms
window_start:
  sec: 1603101739
  nanosec: 842547404
window_stop:
  sec: 1603101740
  nanosec: 843078333
statistics:
- data_type: 1
  data: .nan
- data_type: 3
  data: .nan
- data_type: 2
  data: .nan
- data_type: 5
  data: 0.0
- data_type: 4
  data: .nan
---
measurement_source_name: test_sub_stats_node
metrics_source: message_period
unit: ms
window_start:
  sec: 1603101739
  nanosec: 842547404
window_stop:
  sec: 1603101740
  nanosec: 843078333
statistics:
- data_type: 1
  data: .nan
- data_type: 3
  data: .nan
- data_type: 2
  data: .nan
- data_type: 5
  data: 0.0
- data_type: 4
  data: .nan
---
```

I am sure it's introduced in #1281.

#1281 lost logic,
```
    if (!has_samples) {
      continue;
    }
```

ivanpauno:
@dabonnie does @iuhilnehc-ynos comment make sense to you?
Would reintroducing that logic fix the issue?

dabonnie:
I don't think so (context [here](https://github.com/ros2/rclcpp/pull/1274/files#r469641061).

[...]

PR #1411, opened 2020-10-20
Initial benchmark tests for rclcpp::init/shutdown create/destroy node

brawner:
This PR adds some basic benchmarks and reorganizes the test CMakeLists.txt a bit. I added a new CMakeLists.txt for the benchmark directory and separated out one for the test/rclcpp directory.

Edit: Reorganization and refactoring was moved to #1422. This just now adds new benchmarks


rclcpp/test/benchmark/benchmark_init_shutdown.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+BENCHMARK_F(PerformanceTest, rclcpp_shutdown)(benchmark::State & state)
+{
+  for (auto _ : state) {
+    state.PauseTiming();
+    rclcpp::init(0, nullptr);
+    state.ResumeTiming();
+
+    rclcpp::shutdown();
+  }
```
cottsay, line 40: The `PauseTiming` and `ResumeTiming` concepts are evidently pretty expensive and can make the measurements noisy. The approach we've been taking in the other benchmarks is to time the cycle (in this case, startup/shutdown, elsewhere init/fini, etc).

That said, I haven't actually tried to measure how bad Pause/Resume makes things. If you can gather some data on the subject, I could be convinced to start using them, otherwise I think we should opt for the less noisy approach, since the EC2 instances are generating enough noise as it is.

rclcpp/test/benchmark/benchmark_node.cpp
```diff
@@ -0,0 +1,66 @@
...
+  void SetUp(::benchmark::State & state)
+  {
+    rclcpp::init(0, nullptr);
+    performance_test_fixture::PerformanceTest::SetUp(state);
+  }
+
+  void TearDown(::benchmark::State & state)
+  {
+    rclcpp::shutdown();
+    performance_test_fixture::PerformanceTest::TearDown(state);
```
cottsay, line 35: Since `PerformanceTest::TearDown` stops the memory measurements, you should call that first so that any memory operations involved in tearing down `rclcpp` don't get recorded.

```suggestion
    performance_test_fixture::PerformanceTest::TearDown(state);
    rclcpp::shutdown();
```

rclcpp/test/benchmark/benchmark_node.cpp
```diff
@@ -0,0 +1,66 @@
...
+  }
+};
+
+BENCHMARK_F(NodePerformanceTest, create_node)(benchmark::State & state)
+{
+  std::vector<std::unique_ptr<rclcpp::Node>> nodes(10000);
```
cottsay, line 41: This vector allocation will cause memory allocation to occur. You should reset the heap counters so that the statistics are reset before the loop runs. If the `heap_allocations` statistic isn't an integer, it's usually an indication that each iteration of your timing loop is not performing exactly the same operations, which will result in noisy results.

```suggestion
  std::vector<std::unique_ptr<rclcpp::Node>> nodes(10000);
  reset_heap_counters();
```

rclcpp/test/benchmark/benchmark_node.cpp
```diff
@@ -0,0 +1,66 @@
...
+  for (auto _ : state) {
+    auto node = std::make_unique<rclcpp::Node>("node");
+    benchmark::DoNotOptimize(node);
+    benchmark::ClobberMemory();
```
cottsay, line 45: I could never observe any change in behavior when I used these. Could you share the results with/without it?

rclcpp/test/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+  if(RCLCPP_UNPARSED_ARGUMENTS)
+    message(
+      FATAL_ERROR
+      "Unrecognized arguments for 'add_rclcpp_benchmark'"
+      " (${RCLCPP_BENCHMARK_UNPARSED_ARGUMENTS})")
+    return()
+  endif()
+  find_package(${rmw_implementation} REQUIRED)
+  message(STATUS "Adding ${NAME} for '${rmw_implementation}'")
+  set(rmw_implementation_env_var RMW_IMPLEMENTATION=${rmw_implementation})
```
cottsay, line 18: I think `rmw_implementation_env_var` needs to get passed to `add_performance_test`.

cottsay (changes requested):

rclcpp/test/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+  add_performance_test(${full_benchmark_name} ${SOURCE})
+  if(TARGET ${full_benchmark_name})
+    target_link_libraries(${full_benchmark_name} ${PROJECT_NAME} ${LIBRARIES})
+  endif()
+endfunction()
+
+# Add new benchmarks inside this macro
+macro(rclcpp_benchmarks)
+  add_rclcpp_benchmark(benchmark_init_shutdown benchmark_init_shutdown.cpp)
+  add_rclcpp_benchmark(benchmark_node benchmark_node.cpp)
```
ahcorde, line 30: don't you need to link here against ${PROJECT_NAME} ?

ahcorde (changes requested):

rclcpp/test/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# Give cppcheck hints about macro definitions coming from outside this package
+get_target_property(
+  ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS
+  performance_test_fixture::performance_test_fixture
+  INTERFACE_INCLUDE_DIRECTORIES)
+
+function(add_rclcpp_benchmark NAME SOURCE)
+  cmake_parse_arguments(RCLCPP_BENCHMARK "" "LIBRARIES" "")
```
ahcorde, line 8:
```suggestion
  set(multiValueArgs LIBRARIES)
  cmake_parse_arguments(RCLCPP_BENCHMARK "" "" "LIBRARIES" ${ARGN})
```

rclcpp/test/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+    return()
+  endif()
+  find_package(${rmw_implementation} REQUIRED)
+  message(STATUS "Adding ${NAME} for '${rmw_implementation}'")
+  set(rmw_implementation_env_var RMW_IMPLEMENTATION=${rmw_implementation})
+
+  set(full_benchmark_name ${NAME}${target_suffix})
+  add_performance_test(${full_benchmark_name} ${SOURCE})
+  if(TARGET ${full_benchmark_name})
+    target_link_libraries(${full_benchmark_name} ${PROJECT_NAME} ${LIBRARIES})
```
ahcorde, line 23:
```suggestion
    target_link_libraries(${full_benchmark_name} ${PROJECT_NAME} ${RCLCPP_BENCHMARK_LIBRARIES})
```

ahcorde (changes requested):

rclcpp/test/benchmark/benchmark_node.cpp
brawner: The difficult thing about comparing results with this is that `DoNotOptimize()` doesn't remove all optimizations. So sometimes it's important to include and sometimes it makes no difference. Because ultimately the compiler's optimizations will be different depending on the platform and optimization level, I thought it would be better to include it. Even though for this example it doesn't make much difference.

Here are two simple examples which shows what gets optimized and what doesn't.
https://godbolt.org/z/c6aP3d  with DoNotOptimize
https://godbolt.org/z/8ee5sT without

A little more context:
https://github.com/google/benchmark/issues/242#issuecomment-230056319

brawner:
Building with `--cmake-args -DAMENT_RUN_PERFORMANCE_TESTS=ON` and testing `--packages-select rclcpp`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12809/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7767/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10523/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12755/)

brawner:
Fixed some more issues with the CMakeLists and its internal macro.

It looks like the node benchmark may need to be disabled for connext for now. Destroying a node with connext takes upwards of 3 seconds.

cottsay:
> It looks like the node benchmark may need to be disabled for connext for now. Destroying a node with connext takes upwards of 3 seconds.

Isn't that exactly the kind of thing we want to track with this sort of testing?

brawner:
We certainly would have like to have caught the regression back when it first happened. But the current issue is approaching 2 years old...

https://github.com/ros2/rmw_connext/issues/325

brawner:
Moved the find_package calls out of the base CMakeLists following @cottsay's PRs: https://github.com/ros2/performance_test_fixture/pull/8, https://github.com/ament/ament_cmake/pull/293

rclcpp/test/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+find_package(performance_test_fixture REQUIRED)
+
+# Give cppcheck hints about macro definitions coming from outside this package
+get_target_property(
+  ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS
+  performance_test_fixture::performance_test_fixture
+  INTERFACE_INCLUDE_DIRECTORIES)
```
cottsay, line 7: Is this necessary or working? I always assumed it needed to be defined before `ament_cmake_cppcheck` was processed in the CMake.

brawner: Removing it did not seem to impact cppcheck in any way.

brawner:
Adding a runtime early to test_node for rmw_connext_cpp, and an additional create_destroy_node for rmw_connext_cpp. These early returns will print out:

```
NodePerformanceTest/create_destroy_node    0 ns      0 ns          0 heap_allocations=inf
```

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12824/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7781/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10538/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12769/)

rclcpp/test/benchmark/benchmark_init_shutdown.cpp
brawner: @cottsay one downside to using these Pause/Resume calls is that the benchmark library will require running many more iterations to get much higher certainty in the measurements. But if the code inside the Pause/Resume is slow, then the benchmark can take a long time to run. The prime example here is `create_node` in `test_node__rmw_connext_cpp`. There creating the node is fast, but destroying it is very slow (3.18 seconds) so benchmark wants to run that test a many many times, instead of just once, which it does for the `destroy_node` benchmark (and the new `create_destroy_node` one).  Ultimately, this caused the benchmark to timeout.

I think the current default is that it tries to collect about 1 second worth of iterations? If Pause/Resume is then used and the code outside of Pause/Resume is 50% of the total iteration runtime, the benchmark would take 2 seconds to run. If that code is an even smaller percentage it would take many more seconds to run. This will easily lead to longer benchmark runtimes if not used carefully for rclcpp.

Therefore my recommendation is that if the code run outside of the Pause/Resume calls is small (< 5 us) Pause/Resume should be avoided. OR if the benchmark is not critical and if the code outside Pause/Resume is less than 50% of the runtime of the iteration, then it should also be avoided.

cottsay:
> Adding a runtime early return to test_node for rmw_connext_cpp, and an additional create_destroy_node for rmw_connext_cpp.

I would *much* rather see the CMake logic add `SKIP_TEST` when Connext is used instead of generating false results. Then we'll actually see the test reported as "skipped". Thoughts?

brawner:
Though I should have built those in release mode, debug mode showed me the errors of my way. Google test has a runtime skip, but sadly I don't think there is an equivalent for google benchmark.

> /home/jenkins-agent/workspace/ci_linux/ws/build/google_benchmark_vendor/benchmark-1.5.2-prefix/src/benchmark-1.5.2/src/benchmark_runner.cc:120: RunInThread: Check `st.error_occurred() || st.iterations() >= st.max_iterations' failed. Benchmark returned before State::KeepRunning() returned false!

brawner:
Seems like SKIP_TEST in cmake is probably the best way to go.

cottsay:
Wow, the heap allocation counts vary wildly in the node creation test. This indicates that the algorithm isn't performing the same "amount of work" in each iteration, since it's allocating different amounts of memory each time. I don't think we should try to figure out what's going on right now because it's probably a deep rabbit hole, but we should probably follow-up with some answers.

It does mean that the timing probably has unnecessary noise in it.

rclcpp/test/benchmark/benchmark_node.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include "performance_test_fixture/performance_test_fixture.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+using performance_test_fixture::PerformanceTest;
+
+class NodePerformanceTest : public PerformanceTest
+{
+public:
+  void SetUp(::benchmark::State & state)
```
cottsay, line 26: Please choose to either include the leading `::` when referencing the `benchmark` namespace, or remove it here. I don't care which one you choose, but please be consistent.

brawner: Done

brawner:
Disabled connext through cmake since an early return led to assertion failures, and removed that logic from the benchmark itself.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12832/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7787/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10544/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12775/)

brawner:
Due to inconsistencies we had found before (and #1417), I added some warmup loops before the actual benchmark measurements are done.

brawner:
Err, needed to reset_heap_counters after warmup code.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12833/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7788/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10545/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12776/)

ahcorde:
There is a test failing in CI

```
02:55:29 81: Failed to find library 'rclcpp__rosidl_typesupport_fastrtps_cpp'
02:55:29 81: unknown file: Failure
02:55:29 81: C++ exception with description "could not create publisher: type support not from this implementation, at /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/publisher.cpp:90, at /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rcl/rcl/src/rcl/publisher.c:114" thrown in the test body.
02:55:29 81: [  FAILED  ] TestSubscriptionTopicStatisticsFixture.test_receive_stats_for_message_with_header (5 ms)
```

brawner:
test_subscription_topic_statistics is a known flaky failure, and it even fails regularly with connext (https://github.com/ros2/rclcpp/issues/1403). Likewise the test_executor/test_graph_listener failures on windows will hopefully be addressed by #1404 or something similar.

cottsay (approved):

brawner:
Separated the refactoring of the CMakeLists.txt from this PR. See #1422. I'll rerun tests once that's merged and this is rebased back onto master.

brawner:
Test failures are the usual test_subscription_topic_statistics and test_publisher/subscription_counts. test_multi_threaded_executor on macos is a rare flaky failure. All these test failures are unrelated to this PR though.

brawner:
This gave me a heart attack, but I think that's just because it deleted the branch that this PR used to target. Now that it targets master, delete_bot's reference was updated as well? The master branch was *not* actually deleted
Screenshot from 2020-10-22 14-04-21

brawner:
Building: `--cmake-args --no-warn-unused-cli-DINSTALL_EXAMPLES=OFF -DSECURITY=ON -DAMENT_RUN_PERFORMANCE_TESTS=ON --packages-up-to rclcpp`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12846/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7800/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10557/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12790/)

brawner:
Only failing test is TestSubscriptionTopicStatistics which is unrelated. Linux--arch64 and Window's cmake warnings have to do with osrf_testing_tools_cpp not existing on their platforms, which is expected.

PR #1413, opened 2020-10-20
Added executor benchmark tests

ahcorde:
This PR buils in top of this other PR https://github.com/ros2/rclcpp/pull/1411

This one includes benchmark tests for executors

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/rclcpp/test/benchmark/CMakeLists.txt
+++ b/rclcpp/test/benchmark/CMakeLists.txt
@@ -10,6 +10,12 @@ if(TARGET benchmark_client)
   ament_target_dependencies(benchmark_client test_msgs rcl_interfaces)
 endif()
 
+add_performance_test(benchmark_executor benchmark_executor.cpp)
+if(TARGET benchmark_executor)
+  target_link_libraries(benchmark_executor ${PROJECT_NAME})
+  ament_target_dependencies(benchmark_executor test_msgs)
+endif()
+
 add_performance_test(benchmark_init_shutdown benchmark_init_shutdown.cpp)
 if(TARGET benchmark_init_shutdown)
   target_link_libraries(benchmark_init_shutdown ${PROJECT_NAME})
--- a/rclcpp/test/benchmark/benchmark_executor.cpp
+++ b/rclcpp/test/benchmark/benchmark_executor.cpp
@@ -183,7 +183,9 @@ BENCHMARK_F(PerformanceTestExecutorSimple, multi_thread_executor_remove_node)(be
   }
 }
 
-BENCHMARK_F(PerformanceTestExecutorSimple, static_single_thread_executor_add_node)(benchmark::State & st)
+BENCHMARK_F(
+  PerformanceTestExecutorSimple,
+  static_single_thread_executor_add_node)(benchmark::State & st)
 {
   rclcpp::executors::StaticSingleThreadedExecutor executor;
   for (auto _ : st) {
@@ -194,7 +196,9 @@ BENCHMARK_F(PerformanceTestExecutorSimple, static_single_thread_executor_add_nod
   }
 }
 
-BENCHMARK_F(PerformanceTestExecutorSimple, static_single_thread_executor_remove_node)(benchmark::State & st)
+BENCHMARK_F(
+  PerformanceTestExecutorSimple,
+  static_single_thread_executor_remove_node)(benchmark::State & st)
 {
   rclcpp::executors::StaticSingleThreadedExecutor executor;
   for (auto _ : st) {
--- a/rclcpp/test/benchmark/benchmark_executor.cpp
+++ b/rclcpp/test/benchmark/benchmark_executor.cpp
@@ -14,6 +14,7 @@
 
 #include <memory>
 #include <string>
+#include <utility>
 
 #include "performance_test_fixture/performance_test_fixture.hpp"
 
--- a/rclcpp/test/benchmark/benchmark_executor.cpp
+++ b/rclcpp/test/benchmark/benchmark_executor.cpp
@@ -137,6 +137,51 @@ public:
   rclcpp::Node::SharedPtr node;
 };
 
+
+BENCHMARK_F(PerformanceTestExecutorSimple, single_thread_executor_add_node)(benchmark::State & st)
+{
+  rclcpp::executors::SingleThreadedExecutor executor;
+  for (auto _ : st) {
+    executor.add_node(node);
+    st.PauseTiming();
+    executor.remove_node(node);
+    st.ResumeTiming();
+  }
+}
+
+BENCHMARK_F(PerformanceTestExecutorSimple, single_thread_executor_remove_node)(benchmark::State & st)
+{
+  rclcpp::executors::SingleThreadedExecutor executor;
+  for (auto _ : st) {
+    st.PauseTiming();
+    executor.add_node(node);
+    st.ResumeTiming();
+    executor.remove_node(node);
+  }
+}
+
+BENCHMARK_F(PerformanceTestExecutorSimple, multi_thread_executor_add_node)(benchmark::State & st)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  for (auto _ : st) {
+    executor.add_node(node);
+    st.PauseTiming();
+    executor.remove_node(node);
+    st.ResumeTiming();
+  }
+}
+
+BENCHMARK_F(PerformanceTestExecutorSimple, multi_thread_executor_remove_node)(benchmark::State & st)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  for (auto _ : st) {
+    st.PauseTiming();
+    executor.add_node(node);
+    st.ResumeTiming();
+    executor.remove_node(node);
+  }
+}
+
 BENCHMARK_F(
   PerformanceTestExecutorSimple,
   static_single_thread_executor_spin_until_future_complete)(benchmark::State & st)
--- a/rclcpp/test/benchmark/benchmark_executor.cpp
+++ b/rclcpp/test/benchmark/benchmark_executor.cpp
@@ -149,7 +149,8 @@ BENCHMARK_F(PerformanceTestExecutorSimple, single_thread_executor_add_node)(benc
   }
 }
 
-BENCHMARK_F(PerformanceTestExecutorSimple, single_thread_executor_remove_node)(benchmark::State & st)
+BENCHMARK_F(
+  PerformanceTestExecutorSimple, single_thread_executor_remove_node)(benchmark::State & st)
 {
   rclcpp::executors::SingleThreadedExecutor executor;
   for (auto _ : st) {
--- a/rclcpp/test/benchmark/benchmark_executor.cpp
+++ b/rclcpp/test/benchmark/benchmark_executor.cpp
@@ -15,6 +15,7 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "performance_test_fixture/performance_test_fixture.hpp"
 
@@ -25,6 +26,8 @@
 using namespace std::chrono_literals;
 using performance_test_fixture::PerformanceTest;
 
+constexpr unsigned int kNumberOfNodes = 10;
+
 class PerformanceTestExecutor : public PerformanceTest
 {
 public:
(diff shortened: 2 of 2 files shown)


rclcpp/test/benchmark/CMakeLists.txt
```diff
@@ -4,9 +4,12 @@ get_target_property(
   performance_test_fixture::performance_test_fixture
   INTERFACE_INCLUDE_DIRECTORIES)
 
+find_package(test_msgs REQUIRED)
```
brawner, line 7: This is already included in the rclcpp/test/CMakeLists.txt. Is it customary to include it again?

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,309 @@
...
+    publisher.reset();
+    subscription.reset();
+    node.reset();
+    rclcpp::shutdown();
+  }
+
+  rclcpp::Node::SharedPtr node;
+  rclcpp::Publisher<test_msgs::msg::Empty>::SharedPtr publisher;
+  rclcpp::Subscription<test_msgs::msg::Empty>::SharedPtr subscription;
+  int callback_count;
```
brawner, line 55: This needs to be initialized (either on construction or in setup).

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+BENCHMARK_F(PerformanceTestExecutor, single_thread_executor)(benchmark::State & st)
+{
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(node);
+
+  reset_heap_counters();
+
+  for (auto _ : st) {
+    executor.spin_some(std::chrono::nanoseconds(50));
```
brawner, line 66: These executors don't really have any work to do during spin_some, and your timeout is really short. What are you trying to benchmark?

rclcpp/test/benchmark/benchmark_executor.cpp
ahcorde: it's initialized here:

https://github.com/ros2/rclcpp/blob/6873a47ff379c52246cd5c6295994376d090a430/rclcpp/test/benchmark/benchmark_executor.cpp#L33-L39

rclcpp/test/benchmark/benchmark_executor.cpp
ahcorde: there is a publisher and a subscriber sending an empty message. I can increase the timeout

rclcpp/test/benchmark/CMakeLists.txt
ahcorde: solved https://github.com/ros2/rclcpp/pull/1413/commits/2a6c761af3737dedce40293ba96c98e301b13aae

rclcpp/test/benchmark/benchmark_executor.cpp
brawner: this->callback_count is not set to 0 as far as I can tell. Or am I missing something?

rclcpp/test/benchmark/benchmark_executor.cpp
brawner: The publisher doesn't publish though... so the subscriber won't get any messages either.

rclcpp/test/benchmark/benchmark_executor.cpp
ahcorde: I was planning to check this number when the test is finished to raise an error if there is no message published. (like right now)

ahcorde:
Compiling up-to `rclcpp` and testing `rclcpp`

With -DAMENT_RUN_PERFORMANCE_TESTS=ON:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12848/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7802/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10559/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12792/)

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12855/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7810/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10566/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12798/)

rclcpp/test/benchmark/benchmark_executor.cpp
brawner: callback_count still needs to be initialized either during construction or setup of the unit test. Otherwise it's value is undefined and could potentially not be 0.

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,352 @@
...
+  }
+  if (callback_count == 0) {
+    st.SkipWithError("No message was received");
+  }
+}
+
+BENCHMARK_F(PerformanceTestExecutor, multi_thread_executor)(benchmark::State & st)
+{
+  rclcpp::executors::MultiThreadedExecutor executor;
+  executor.add_node(node);
```
cottsay, line 86: There is some extra work that has to be done in `spin` to bring up the new node, from what I can tell. I think you need to add a `spin_some` before the heap counters are reset.

I'm also seeing a single extra allocation that happens the first time the callback is received, so throwing an extra `publish` before resetting the counters would also be a good idea.

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+  reset_heap_counters();
+
+  for (auto _ : st) {
+    st.PauseTiming();
+    publisher->publish(empty_msgs);
```
cottsay, line 92: My tests are showing that the message is always getting through on CycloneDDS, but not Fast-DDS. The dropped messages will certainly contribute to inconsistencies in the data. My guess is that this is caused by sync vs async mode in DDS, so we won't be able to compare the results between the vendors unless we configure them to behave consistently.

This is also causing the test to fail sometimes for me. The first thing Google Benchmark does is invoke the loop exactly once to get a general idea of how complex the operation is, and if that first message gets dropped, it triggers the `SkipWithError` you have below.

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,353 @@
...
+    st.ResumeTiming();
+
+    executor.spin_some(100ms);
+
+    benchmark::ClobberMemory();
```
cottsay, line 97: Does this have any effect without `benchmark::DoNotOptimize()` as well? @brawner?

rclcpp/test/benchmark/CMakeLists.txt
```diff
@@ -64,6 +64,11 @@ macro(rclcpp_benchmarks)
     benchmark_node
     benchmark_node.cpp
     TEST_OPTIONS ${SKIP_TEST})
+  add_rclcpp_benchmark(benchmark_executor benchmark_executor.cpp
+    AMENT_DEPENDS
+      test_msgs
+    TEST_OPTIONS
+      ${SKIP_TEST})
```
cottsay, line 71: Do we actually want to skip this on Connext? I thought the reason behind that was due to performance problems with node shutdown in the timing loop, which I don't see happening in these benchmarks.

cottsay (changes requested):

rclcpp/test/benchmark/benchmark_executor.cpp
brawner: Here's the explanation from benchmark's readme:

> The second tool for preventing optimizations is ClobberMemory(). In essence ClobberMemory() forces the compiler to perform all pending writes to global memory. Memory managed by block scope objects must be "escaped" using DoNotOptimize(...) before it can be clobbered. In the below example ClobberMemory() prevents the call to v.push_back(42) from being optimized away.

It sounds like clobbermemory has no effect if there is not a DoNotOptimize

rclcpp/test/benchmark/benchmark_executor.cpp
ahcorde: We are in the same issue that we already have with `performance_test`. Are we comparing/testing the ''default" behaviour or should we compare all the possible scenarios?

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12907/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7861/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10624/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12856/)

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,408 @@
...
+  reset_heap_counters();
+
+  for (auto _ : st) {
+    st.PauseTiming();
+    publisher->publish(empty_msgs);
+    st.ResumeTiming();
+
+    executor.spin_some(100ms);
+  }
+  if (callback_count == 0) {
```
brawner, line 78: Should callback_count be reset to 0 in line 68 to ensure that the callbacks are being called in the for loop? Fix here and below.

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,408 @@
...
+  }
+
+  test_msgs::msg::Empty empty_msgs;
+  rclcpp::Node::SharedPtr node;
+  rclcpp::Publisher<test_msgs::msg::Empty>::SharedPtr publisher;
+  rclcpp::Subscription<test_msgs::msg::Empty>::SharedPtr subscription;
+  int callback_count;
+};
+
+BENCHMARK_F(PerformanceTestExecutor, single_thread_executor)(benchmark::State & st)
```
brawner, line 62: I would add more descriptive test name to `single_thread_executor`, `multi_thread_executor` and `static_single_thread_executor`. Maybe `single_thread_executor_spin_some`, etc...

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,408 @@
...
+    if (ret != rclcpp::FutureReturnCode::SUCCESS) {
+      st.SkipWithError(rcutils_get_error_string().str);
+      break;
+    }
+  }
+}
+
+BENCHMARK_F(
+  PerformanceTestExecutorSimple,
+  static_single_thread_executor_spin_node_until_future_complete)(benchmark::State & st)
```
brawner, line 210: How is this benchmark different from the current `static_single_thread_executor` benchmark?

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,408 @@
...
+    if (ret != rclcpp::FutureReturnCode::SUCCESS) {
+      st.SkipWithError(rcutils_get_error_string().str);
+      break;
+    }
+  }
+}
+
+BENCHMARK_F(
+  PerformanceTestExecutorSimple,
+  single_thread_executor_spin_node_base_interface_until_future_complete)(benchmark::State & st)
```
brawner, line 233: I don't think you need the `interface` variants of these benchmarks. The `spin_until_future_complete` versions you have above call the exact same code as these.

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,408 @@
...
+    if (ret != rclcpp::FutureReturnCode::SUCCESS) {
+      st.SkipWithError(rcutils_get_error_string().str);
+      break;
+    }
+  }
+}
+
+BENCHMARK_F(
+  PerformanceTestExecutorSimple,
+  node_base_interface_spin_until_future_complete)(benchmark::State & st)
```
brawner, line 346: No need for the `interface` variant, since the above benchmark calls this same code.

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,408 @@
...
+    if (ret != rclcpp::FutureReturnCode::SUCCESS) {
+      st.SkipWithError(rcutils_get_error_string().str);
+      break;
+    }
+  }
+}
+
+BENCHMARK_F(
+  PerformanceTestExecutorSimple,
+  static_executor_entities_collector)(benchmark::State & st)
```
brawner, line 374:
```suggestion
  static_executor_entities_collector_execute)(benchmark::State & st)
```

brawner (commented):
What do you think about benchmarking `add_node` and `remove_node`? That happens a lot when the base-version of `rclcpp::spin_*` functions are called.

brawner:
Also, looking at the benchmarks in the above CI jobs, multithreaded and singlethreaded get the exact same numbers which I assume is due to the fact that they only have one node each. What do you think of running `spin_some` but with something like 10 nodes each?

rclcpp/test/benchmark/benchmark_executor.cpp
ahcorde: One it's using the `node` and the other one the `node_base_interface`

ahcorde: Removed all the `interface` variants

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12944/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7893/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10661/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12893/)

brawner (commented):
It looks good to me, but what did you think about my comment of benchmarking add/remove node?

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+  rclcpp::Node::SharedPtr node;
+};
+
+BENCHMARK_F(
+  PerformanceTestExecutorSimple,
+  static_single_thread_executor_spin_until_future_complete)(benchmark::State & st)
+{
+  rclcpp::executors::StaticSingleThreadedExecutor executor;
+  executor.add_node(node);
```
brawner, line 145: > PerformanceTestExecutorSimple/static_single_thread_executor_spin_until_future_complete            30.3 ns         30.3 ns     23117640 heap_allocations=0

No work is being performed during this benchmark because static_singcle_thread_executor is a special design. You need to add/remove the node each time you call spin. That way the entities collector can 'collect' the work to be done.

rclcpp/test/benchmark/benchmark_executor.cpp
brawner: This benchmark is the same as `static_single_thread_executor_spin_until_future_complete` except there is just one more jump that happens. `rclcpp::executors::spin_node_until_future_complete` just calls add_node, spin_until_future_complete and then remove_node, which is what the above test needs to also be doing.

brawner (changes requested):
Sorry, upon closer inspection of the timing results I realized there were a couple of existing issues here.

brawner:
> PerformanceTestExecutor/single_thread_executor_spin_some                                        449165 ns       449496 ns         1602 heap_allocations=156
PerformanceTestExecutor/multi_thread_executor_spin_some                                         448985 ns       449328 ns         1563 heap_allocations=155.992

Based on these results even though you're using 10 nodes, I take it the aws machine only has one core?

ahcorde:
@brawner, I have a 8 core machine, but the numbers are very similar 

```
9: PerformanceTestExecutor/single_thread_executor_spin_some                                        572711 ns       571053 ns         1211 heap_allocations=155.893
9: PerformanceTestExecutor/multi_thread_executor_spin_some                                         565251 ns       565667 ns         1216 heap_allocations=155.893
9: PerformanceTestExecutorSimple/single_thread_executor_add_node                                     8099 ns         8060 ns        86212 heap_allocations=5.00019
9: PerformanceTestExecutorSimple/single_thread_executor_remove_node                                  9375 ns         9413 ns        73593 heap_allocations=5.00023
9: PerformanceTestExecutorSimple/multi_thread_executor_add_node                                      8093 ns         8023 ns        87132 heap_allocations=5.0002
9: PerformanceTestExecutorSimple/multi_thread_executor_remove_node                                   9597 ns         9632 ns        74300 heap_allocations=5.00023
9: PerformanceTestExecutorSimple/static_single_thread_executor_spin_until_future_complete             579 ns          501 ns      1373339 heap_allocations=1
9: PerformanceTestExecutorSimple/single_thread_executor_spin_node_until_future_complete             15854 ns        15796 ns        43443 heap_allocations=5
9: PerformanceTestExecutorSimple/multi_thread_executor_spin_node_until_future_complete              15928 ns        15869 ns        43874 heap_allocations=5
9: PerformanceTestExecutorSimple/static_single_thread_executor_spin_node_until_future_complete       2731 ns         2721 ns       257257 heap_allocations=1.00001
9: PerformanceTestExecutorSimple/spin_until_future_complete                                         64075 ns        63835 ns        10937 heap_allocations=22
9: PerformanceTestExecutorSimple/static_executor_entities_collector_execute                         15248 ns        15248 ns        45676 heap_allocations=7
```

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -0,0 +1,363 @@
...
+    PerformanceTest::TearDown(st);
+    node.reset();
+    rclcpp::shutdown();
+  }
+
+  rclcpp::Node::SharedPtr node;
+};
+
+
+BENCHMARK_F(PerformanceTestExecutorSimple, single_thread_executor_add_node)(benchmark::State & st)
```
brawner, line 141: No add/remove for static_single_threaded_executor? That class actually overrides the default implementation (whereas single_threaded/multi_threaded use the base class one).

ahcorde: https://github.com/ros2/rclcpp/pull/1413/commits/5fc0cac8814e8b11075d7e2c576e65e3139f8e5c

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12967/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7917/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10688/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12917/)

brawner (approved):
I'll be curious if spin_some ends up giving us some false negatives as the underlying code changes the decideable threshold on what 'some' is. I was playing around with this yesterday and spin_some processed fewer than 10 nodes per call. `spin_all` took something like 30% longer.

ahcorde:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12939/)

PR #1416, opened 2020-10-21
Add EventsExecutor

irobot-ros:
This PR introduces the changes required to implement the `EventsExecutor` design in rclcpp.
See [design](https://github.com/ros2/design/pull/305) and [Discourse post](https://discourse.ros.org/t/ros2-middleware-change-proposal/15863).

The new executor uses an events queue and a timers manager as opposed to waitsets, to efficiently execute entities with work to do.

This new executor greatly reduces CPU usage of a ROS 2 application.
These are the CPU and latency measurements on a RaspberryPi 2 for a 20 nodes system.
See the blog post for more details on the tests that we run.

Screenshot from 2020-10-21 15-39-07

The bulk of the changes for this implementation are in the rclcpp layer, with some minor changes in other repositories (rcl, rmw, rmw_implementation) for forwarding entities, the declaration of some data types in rcutils, and finally some additional changes in the vendor specific rmw implementations..
We currently implemented this only on top of the default ROS middleware fastrtps, while we provided stubs for other middlewares.

The current implementation does not support ROS 2 actions, which will be added in a follow up PR.

Developed by iRobot
Mauro Passerino
Lenny Story
Alberto Soragna

----

Connects to:

- design:
  - https://github.com/ros2/design/pull/305


ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/20

mauropasse:
Hi! In the latest update of this PR we have implemented the ownership model matching the behaviour followed by the **SingleThreadedExecutor**, as requested in the design PR discussion https://github.com/ros2/design/pull/305#discussion_r541155343:

"_The executor should have weak ownership on the entities associated with it, and strong ownership when using them._"

We took special care in trying to maintain the same performances as obtained in the previous model, in which there was no need to lock weak pointers to get ownership of entities.
The CPU usage resulted the same for a 10-nodes system, while for a bigger system (20 nodes) the CPU went from 21% to 26% when using the new ownership model.

Another topic of discussion was the possible unbound growth of the events queue. We already have some ideas for addressing this issue, and we plan to share them on the design proposal. But we'd like to implement the fixes on a follow up PR to avoid taking too long to merge these changes on the codebase, and being able to start using and testing the Events Executor.
Please let us know if you'd agree on this, or if you have another ideas on mind.

Thanks!
-The _iRobot_ team

wjwwood:
Hey guys, thanks for continuing to update this and iterate on our feedback.

We had a meeting about this last week (some of the maintainers of rclcpp and related packages) and we decided to try and get the "listener" related changes merged in first and then continue to work on the executor itself.

That way there's less to keep in sync and we can also enable these features to be used outside of rclcpp and this pull request.

One of the things still blocking the executor in my opinion is how to deal with the event queue. I'm still not convinced that it is good to have an unbounded queue there. And if we don't have one, and instead bound it, how do we deal with overruns?

I've scheduled a bit of time in the next middleware working group meeting (tomorrow) to discuss this and other issues with this pull request.

In the meantime, @ivanpauno already did some reviews of the related pull requests, and I'll do the same. Again, we'll be aiming to merge the listener API changes separately since that's more generally useful and more straightforward than the remaining topics for discussion on the executor.

PR #1421, opened 2020-10-22
Add missing locking to the rclcpp_action::ServerBase.

clalancette:
This patch actually does 4 related things:

1.  Renames the recursive mutex in the ServerBaseImpl class
to action_server_reentrant_mutex_, which makes it a lot
clearer what it is meant to lock.
2.  Adds some additional error checking where checks were missed.
3.  Moves the mutex in execute_check_expired_goals out of the
loop so we aren't constantly locking and unlocking it.
4.  Adds a lock to publish_status so that the action_server
structure is protected.

I found this problem while trying to reproduce https://github.com/ros2/rclcpp/pull/1313 .  It manifested as a segfault that happened sometimes when running a client over and over against a server.  Before this patch, that test would crash within minutes.  With this patch in place, it has been running for several hours without a crash.

@daisukes @fujitatomoya FYI.

### DIFF
--- a/rclcpp_action/src/server.cpp
+++ b/rclcpp_action/src/server.cpp
@@ -44,7 +44,7 @@ public:
   }
 
   // Lock everything except user callbacks
-  std::recursive_mutex reentrant_mutex_;
+  std::recursive_mutex action_server_reentrant_mutex_;
 
   std::shared_ptr<rcl_action_server_t> action_server_;
 
@@ -88,12 +88,13 @@ ServerBase::ServerBase(
       if (nullptr != ptr) {
         rcl_node_t * rcl_node = node_base->get_rcl_node_handle();
         rcl_ret_t ret = rcl_action_server_fini(ptr, rcl_node);
-        (void)ret;
-        RCLCPP_DEBUG(
-          rclcpp::get_logger("rclcpp_action"),
-          "failed to fini rcl_action_server_t in deleter");
+        if (RCL_RET_OK != ret) {
+          RCLCPP_DEBUG(
+            rclcpp::get_logger("rclcpp_action"),
+            "failed to fini rcl_action_server_t in deleter");
+        }
+        delete ptr;
       }
-      delete ptr;
     };
 
   pimpl_->action_server_.reset(new rcl_action_server_t, deleter);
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12842/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7796/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10553/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12785/)

rclcpp_action/src/server.cpp
```diff
@@ -461,6 +464,11 @@ ServerBase::publish_status()
 {
   rcl_ret_t ret;
 
+  // We need to hold the lock across this entire method because
+  // rcl_action_server_get_goal_handles() returns an internal pointer to the
+  // goal data.
+  std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
```
fujitatomoya, line 470: this fix is included in this PR https://github.com/ros2/rclcpp/pull/1313. https://github.com/ros2/rclcpp/blob/daa2640031598c7be32f4efb5671e544cc8b2d3e/rclcpp_action/src/server.cpp#L480

rclcpp_action/src/server.cpp
```diff
@@ -437,9 +439,10 @@ ServerBase::execute_check_expired_goals()
   rcl_action_goal_info_t expired_goals[1];
   size_t num_expired = 1;
 
+  std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
```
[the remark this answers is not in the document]
clalancette: That's fair.  I'm not 100% convinced that we should be doing it this way, but the lock is protecting what it purports to protect.  To keep this PR simpler, I've reverted this bit of the patch.

rclcpp_action/src/server.cpp
clalancette: Yeah, understood.  I was just trying to get the simpler part of that patch in, then take another look at #1313.

jacobperron (approved):

clalancette:
Since I updated the code here, another CI to make sure everything is good to go:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12879/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7834/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10596/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12822/)

fujitatomoya (approved):
lgtm

clalancette:
The failing tests on Windows look to be there already (and they are also at the *rcl* layer, not the rclcpp layer, so can't be caused by this PR).  So I'm going to go ahead and merge, thanks for the reviews.

PR #1432, opened 2020-10-30
rclcpp::Duration constructors might be confusing to users migrating from ROS 1

ivanpauno:
[ros::Duration::Duration(double seconds)](http://docs.ros.org/en/latest/api/rostime/html/classros_1_1Duration.html#ab097cc53395335730c081b6a13d0a12e) provides a constructor from a floating point number, which equivalent in ROS 2 is an static method [rclcpp::Duration::from_seconds](https://github.com/ros2/rclcpp/blob/2309811814b80ad241448aebf51eebd8db17d826/rclcpp/include/rclcpp/duration.hpp#L129-L130).

The only provided constructor from numeric values in rclcpp is [rclcpp::Duration::Duration(rcl_duration_value_t)](https://github.com/ros2/rclcpp/blob/2309811814b80ad241448aebf51eebd8db17d826/rclcpp/include/rclcpp/duration.hpp#L42) (`rcl_duration_value_t -> int64_t`).

The lack of a constructor accepting a `double seconds` together with cpp implicit narrowing conversions might make the user think that `rclcpp::Duration(1.3)` is doing the correct thing, though that's probably not what they want
In gcc or clang, `-Wfloat-conversion` or `-Wconversion` should trigger a warning in this case, but not everybody use that.

[This comment](https://github.com/ros2/rclcpp/blob/2309811814b80ad241448aebf51eebd8db17d826/rclcpp/include/rclcpp/duration.hpp#L41) also provides a bit of extra confusion.

Possible solutions to avoid the issue:
- Add an `rclcpp::Duration::Duration(double seconds)` constructor.
  I'm not a fan of `rclcpp::Duration(1.3)` being 1.3 seconds and `rclcpp::Duration(1)` being 1ns.
- Deprecate `rclcpp::Duration::Duration(rcl_duration_value_t nanoseconds)` in favor of `rclcpp::Duration::from_nanoseconds(rcl_duration_value_t nanoseconds)`.
  This is the approach taken by this drat PR.
- Delete the confusing comment and do nothing.

### DIFF
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -40,7 +40,7 @@ public:
 
   /// Construct duration from the specified nanoseconds.
   [[deprecated(
-    "Use Duration::from_nanoseconds instead or std::chrono literals. For example:"
+    "Use Duration::from_nanoseconds instead or std::chrono_literals. For example:"
     "rclcpp::Duration::from_nanoseconds(int64_variable);"
     "rclcpp::Duration(0ns);")]]
   explicit Duration(rcl_duration_value_t nanoseconds);
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -39,7 +39,10 @@ public:
   Duration(int32_t seconds, uint32_t nanoseconds);
 
   /// Construct duration from the specified nanoseconds.
-  [[deprecated("Use Duration::from_nanoseconds instead")]]
+  [[deprecated(
+    "Use Duration::from_nanoseconds instead or std::chrono literals. For example:"
+    "rclcpp::Duration::from_nanoseconds(int64_variable);"
+    "rclcpp::Duration(0ns);")]]
   explicit Duration(rcl_duration_value_t nanoseconds);
 
   /// Construct duration from the specified std::chrono::nanoseconds.
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -38,10 +38,11 @@ public:
    */
   Duration(int32_t seconds, uint32_t nanoseconds);
 
-  // This constructor matches any numeric value - ints or floats.
+  /// Construct duration from the specified nanoseconds.
+  [[deprecated("Use Duration::from_nanoseconds instead")]]
   explicit Duration(rcl_duration_value_t nanoseconds);
 
-  // This constructor matches std::chrono::nanoseconds.
+  /// Construct duration from the specified std::chrono::nanoseconds.
   explicit Duration(std::chrono::nanoseconds nanoseconds);
 
   // This constructor matches any std::chrono value other than nanoseconds
@@ -129,6 +130,10 @@ public:
   static Duration
   from_seconds(double seconds);
 
+  /// Create a duration object from an integer number representing nanoseconds
+  static Duration
+  from_nanoseconds(rcl_duration_value_t nanoseconds);
+
   /// Convert Duration into a std::chrono::Duration.
   template<class DurationT>
   DurationT
@@ -143,6 +148,8 @@ public:
 
 private:
   rcl_duration_t rcl_duration_;
+
+  Duration() = default;
 };
 
 }  // namespace rclcpp
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -37,7 +37,7 @@ Duration::Duration(int32_t seconds, uint32_t nanoseconds)
   rcl_duration_.nanoseconds += nanoseconds;
 }
 
-Duration::Duration(int64_t nanoseconds)
+Duration::Duration(rcl_duration_value_t nanoseconds)
 {
   rcl_duration_.nanoseconds = nanoseconds;
 }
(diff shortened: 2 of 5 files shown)


wjwwood:
> Deprecate rclcpp::Duration::Duration(rcl_duration_value_t nanoseconds) in favor of rclcpp::Duration::from_nanoseconds(rcl_duration_value_t nanoseconds).
This is the approach taken by this drat PR.

Sounds reasonable to me.

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -143,6 +148,8 @@ class RCLCPP_PUBLIC Duration
 
 private:
   rcl_duration_t rcl_duration_;
+
+  Duration() = default;
```
wjwwood, line 152: Why is this needed?

wjwwood (approved):
lgtm

clalancette:
> Sounds reasonable to me.

Agreed, I think that is the best solution here.

rclcpp/include/rclcpp/duration.hpp
ivanpauno: To implement `Duration::from_seconds`, `Duration::from_nanoseconds` without reusing the deprecated constructor (see [here](https://github.com/ros2/rclcpp/blob/64f1a3e4a3585023b8959a92bb3a58bbc4640244/rclcpp/src/rclcpp/duration.cpp#L249-L263)).
I could also make use of `Duration::Duration(std::chrono::nanoseconds)` instead of adding the private default constructor.

wjwwood: I see. I don't feel strongly about it.

ivanpauno:
CI building all dependencies to catch downstream warnings:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12931/)

fujitatomoya (approved):
this makes sense!

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12934/)

Downstream PRs:
- https://github.com/ros2/geometry2/pull/340
- https://github.com/ros-perception/laser_geometry/pull/72

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -38,10 +38,14 @@ class RCLCPP_PUBLIC Duration
    */
   Duration(int32_t seconds, uint32_t nanoseconds);
 
-  // This constructor matches any numeric value - ints or floats.
+  /// Construct duration from the specified nanoseconds.
+  [[deprecated(
+    "Use Duration::from_nanoseconds instead or std::chrono literals. For example:"
+    "rclcpp::Duration::from_nanoseconds(int64_variable);"
+    "rclcpp::Duration(0ns);")]]
```
clalancette, line 45: Nit:

```suggestion
    "rclcpp::Duration(std::chrono::nanoseconds(0));")]]
```

ivanpauno: `0ns` is valid if you bring `namespace std::chrono_literals` into scope.
If I cannot do `using namespace std::chrono_literals` for some reason (e.g. polluting a header file), I would rather use `rclcpp::Duration::from_nanoseconds(0)` than `rclcpp::Duration(std::chrono::nanoseconds(0))`.

If you think this is confusing for cpp users that don't know about `std::chrono_literals` I can take the suggestion.

clalancette: > If I cannot do `using namespace std::chrono_literals` for some reason (e.g. polluting a header file), I would rather use `rclcpp::Duration::from_nanoseconds(0)` than `rclcpp::Duration(std::chrono::nanoseconds(0))`.
> 
> If you think this is confusing for cpp users that don't know about `std::chrono_literals` I can take the suggestion.

It's not that it is confusing, just that the sentence above explicitly talks about `std::chrono` namespace.  Just for the sake of consistency, I would keep using that namespace.  If users know about the `std::chrono_literals`, then of course they can just use that as well.

But its a nit.  I'll leave it to you whether to fix it or not.

ivanpauno: Ah, I meant to write `std::chrono_literals` in the beginning of the error message.
See https://github.com/ros2/rclcpp/pull/1432/commits/fa216a67719b2c598f9a2e54a516938403bf98f4.

ivanpauno:
Remaining CI jobs:

* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7890/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10658/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12890/)

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -249,7 +249,17 @@ Duration::to_rmw_time() const
...
 Duration::from_seconds(double seconds)
 {
-  return Duration(static_cast<int64_t>(RCL_S_TO_NS(seconds)));
+  Duration ret;
+  ret.rcl_duration_.nanoseconds = static_cast<int64_t>(RCL_S_TO_NS(seconds));
+  return ret;
+}
+
+Duration
+Duration::from_nanoseconds(rcl_duration_value_t seconds)
```
jacobperron, line 258: Tiny nitpick: I would have the variable name match the units
```suggestion
Duration::from_nanoseconds(rcl_duration_value_t nanoseconds)
```

jacobperron (approved):

rclcpp/src/rclcpp/duration.cpp
ivanpauno: thanks

PR #1434, opened 2020-10-31
Use std::unordered_map to improve (average) lookup time

whoan:


clalancette:
Can you please explain how this helped in your application?  I'm sceptical of doing micro-optimizations like this unless there is a shown problem that it addresses.

whoan:
Hi @clalancette . Thanks for looking at this PR. It does not address any actual problem. Looking at the code I saw that the callbacks do not need to be sorted, so I used an `unordered_map` as an optimization. The map is usually implemented as a Red-black tree so current [insertion](https://github.com/ros2/rclcpp/blob/9daf6c2efc59e47f311523b7943c10ef0b16216d/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp#L195) and [look up](https://github.com/ros2/rclcpp/blob/9daf6c2efc59e47f311523b7943c10ef0b16216d/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp#L437) will be O(log n) and we can do it in O(1) (average) with an `unordered_map` as it is implemented as a hash table.

If you still think it's not worth it, feel free to close, I will understand. Thanks again for your fast answer.

fujitatomoya (commented):
I am not convinced with this fix, with using `std::unordered_map` could be optimization but maybe it is too small for now.

reason that i have,

- `cb_map_` only has 6 entries at most which are https://github.com/ros2/rcl_interfaces/blob/7a6ebaa2064151a48dd189cd514fc7db5d260047/lifecycle_msgs/msg/State.msg#L26-L51.
- Insertion / Deletion only happens at initialization / finalization time. (tree re-balance is unlikely to happen)
- For small `N`, a sorted vector is often the faster container, and has less memory footprint.

whoan:
@fujitatomoya I wasn't aware of points 1 and 2, so I agree, it's not worth it.
Regarding point 3, it's true is has less memory footprint, and in addition, it has better locality of reference, which can be helpful if different callbacks need to be accessed over a short period of time.

I am closing this one. Thanks for reviewing.

fujitatomoya:
@whoan thanks for the comments and bring this up!

PR #1406, opened 2020-10-14
[rclcpp_components] Add warnings

audrow:
This PR enables `-Wformat=2`, `-Wconversion`, `-Wshadow`, `-Wsign-conversion`, `-Wold-style-cast`, and `-Wcast-qual` in `rclcpp_components`. This PR relies on using gtest v1.10.0, see https://github.com/ament/googletest/pull/8.

PR #1405, opened 2020-10-14
[rclcpp_action] Add warnings

audrow:
This PR enables `-Wformat=2`, `-Wconversion`, `-Wshadow`, `-Wsign-conversion`, and `-Wcast-qual` in `rclcpp_action`. This PR relies on using gtest v1.10.0, see https://github.com/ament/googletest/pull/8.

PR #1442, opened 2020-11-05
Fix string literal warnings

audrow:
This PR fixes the `-Wformat-nonliteral` warnings seen in #1406 and #1405 by changing the generated macros in `rclcpp` for logging. This approach will force changes to several tests, specifically in [`test_logging.cpp`](https://github.com/ros2/rclcpp/blob/d9d67e68131bb0c815dd7494846da390b6c12648/rclcpp/test/rclcpp/test_logging.cpp).

I've made this a draft PR to make it easier to discuss the changes and tradeoffs of this approach and to make a spot for recording CI runs.

### DIFF
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -151,10 +151,9 @@ def get_rclcpp_suffix_from_features(features):
 @[ end if]@
       logger.get_name(), \
 @[ if 'stream' not in feature_combination]@
-      rclcpp::get_c_string(RCLCPP_FIRST_ARG(__VA_ARGS__, "")), \
-        RCLCPP_ALL_BUT_FIRST_ARGS(__VA_ARGS__,"")); \
+      __VA_ARGS__); \
 @[ else]@
-      "%s", rclcpp::get_c_string(ss.str())); \
+      "%s", ss.str()); \
 @[ end if]@
   } while (0)
 
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -153,7 +153,7 @@ def get_rclcpp_suffix_from_features(features):
 @[ if 'stream' not in feature_combination]@
       __VA_ARGS__); \
 @[ else]@
-      "%s", ss.str()); \
+      "%s", ss.str().c_str()); \
 @[ end if]@
   } while (0)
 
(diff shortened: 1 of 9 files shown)


audrow:
I ran tests for packages above `rclcpp` on macOS to see what additional warnings these changes create:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10734/)

clalancette:
Just to recap what we discussed in person:

We decided to try to implement this in 2 parts.  In order to get the benefit of the format literal error checking, we are going to remove `rclcpp::get_c_string` from the `const char *` version of the macro.  We'll also try to attempt to keep it working for the `std::string` version, but with a warning (either in code or in documentation).  Then we'll convert all of our existing internal uses of the macro to use the `const char *` version.  This keeps it so that all of downstream doesn't have to change right away, but also gives us better warning/error checking.

@audrow Let me know if I summarized the conversation properly.

audrow:
@clalancette That sounds right to me.

audrow:
I think we are going to proceed with the solution proposed in this PR.

We have tried several other approaches in hopes of keeping some of the convenience of passing strings into the logging macros directly:
* Using the preprocessor (macros) to put different versions of the logging macro depending on if a `std::string` or `char *` format string was passed. This didn't work because I wasn't able to find how to do typechecking with macros. 
* Using templated logging functions to handle `std::string` and `char *`  format arguments. This didn't work the logging function is in rcutils, which uses C, and C doesn't support templating.

Extending the second approach, I tried having templated methods in C++ that would pass the format argument to the C logging macros (which place the format string literal in the C logging function `rcutils_log` after expansion). This got me back into the same warning as this PR is attempting to fix, because the variable in the C++ templated function would be passed to the C logging function instead of a string literal, giving a nonliteral warning. Perhaps this approach could solve the problem if there is a way to pass string literals to functions inside of functions without the string literal being assigned to a variable. I explored this option (including using rvalues), but I wasn't able to figure out a solution.

audrow:
Here is a rerun using ros2/examples#295, ros2/rviz#633, and ros2/demos#476.
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10969/)

audrow:
Here's CI running testing on all downstream packages:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13293/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8224/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11022/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13335/)

clalancette (approved):

PR #1449, opened 2020-11-10
Fix test crashes on CentOS 7

clalancette:
CentOS7 has been having test failures for a while; see https://ci.ros2.org/view/nightly/job/nightly_linux-centos_debug/589/#showFailuresLink for the latest example.

The problem ends up being that the mocking_utils `inject_on_return` doesn't work on CentOS7 (https://github.com/ros2/rclcpp/blob/06465ba827da0ce018199fb0f80d15ffe6fc26fd/rclcpp/test/mocking_utils/patch.hpp#L550).  This PR addresses that by changing to `patch_and_return` (which does work on CentOS7), and doing a slight class refactoring so we can stop abusing `inject_on_return` to do partial initialization.  In my testing on CentOS7, this fixes all of the current crashes we are seeing (in rclcpp and rclcpp_action, at least).

### DIFF
--- a/rclcpp/include/rclcpp/graph_listener.hpp
+++ b/rclcpp/include/rclcpp/graph_listener.hpp
@@ -160,6 +160,8 @@ protected:
   void
   run_loop();
 
+  void init_wait_set();
+
 private:
   RCLCPP_DISABLE_COPY(GraphListener)
 
--- a/rclcpp/include/rclcpp/graph_listener.hpp
+++ b/rclcpp/include/rclcpp/graph_listener.hpp
@@ -163,6 +163,9 @@ protected:
   RCLCPP_PUBLIC
   void init_wait_set();
 
+  RCLCPP_PUBLIC
+  void cleanup_wait_set();
+
 private:
   RCLCPP_DISABLE_COPY(GraphListener)
 
--- a/rclcpp/include/rclcpp/graph_listener.hpp
+++ b/rclcpp/include/rclcpp/graph_listener.hpp
@@ -160,6 +160,7 @@ protected:
   void
   run_loop();
 
+  RCLCPP_PUBLIC
   void init_wait_set();
 
 private:
--- a/rclcpp/test/rclcpp/test_graph_listener.cpp
+++ b/rclcpp/test/rclcpp/test_graph_listener.cpp
@@ -129,6 +129,11 @@ public:
   {
     this->init_wait_set();
   }
+
+  void mock_cleanup_wait_set()
+  {
+    this->cleanup_wait_set();
+  }
 };
 
 /* Errors running graph protected methods */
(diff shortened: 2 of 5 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12996/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7948/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10723/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12976/)
* CentOS [Build Status](http://ci.ros2.org/job/ci_linux-centos/86/)

rclcpp/include/rclcpp/graph_listener.hpp
```diff
@@ -160,6 +160,8 @@ class GraphListener : public std::enable_shared_from_this<GraphListener>
   void
   run_loop();
 
+  void init_wait_set();
```
brawner, line 163: `RCLCPP_PUBLIC`?

rclcpp/test/rclcpp/test_graph_listener.cpp
```diff
@@ -268,15 +261,15 @@ TEST_F(TestGraphListener, test_errors_graph_listener_add_remove_node) {
 /* Shutdown errors */
 TEST_F(TestGraphListener, test_graph_listener_shutdown_wait_fini_error_nothrow) {
   graph_listener()->start_if_not_started();
-  auto mock_wait_set_fini = mocking_utils::inject_on_return(
+  auto mock_wait_set_fini = mocking_utils::patch_and_return(
```
brawner, line 264: By switching these to `patch_and_return` you're not finalizing any of these objects which is probably leaking memory in these tests. 

One possible option would be to call `patch` with a custom lambda that calls the `fini` function, but returns `RCL_RET_ERROR`. Would that work on centos, or is that why it's failing to begin with?

rclcpp/include/rclcpp/graph_listener.hpp
clalancette: Good call, see 9f210213c2be2cb30ad7956da6bcbf5573db551f .  Windows CI failed for other reasons, but it probably would have failed here too eventually.

rclcpp/test/rclcpp/test_graph_listener.cpp
clalancette: I'll give it a shot.  I think that is going to fail as well on CentOS, but we'll see.

hidmic: > Would that work on centos, or is that why it's failing to begin with?

It'll fail. The sort of global PLT optimization CentOS seems to be doing is what's preventing `inject_on_return` to work to being with.

brawner: These tests here and below actually passed in the example you provided. What trouble were they giving you?

brawner: After some discussion with @clalancette, it appears these leaks may not be avoidable.

brawner (approved):

rclcpp/test/rclcpp/test_graph_listener.cpp
clalancette: Yeah, it failed.  See 84b5289, where I take a different approach.  There are still leaks, but it is not worse than before.

clalancette:
Here's another CI for the latest code:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13014/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7964/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10740/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12995/)
* CentOS [Build Status](http://ci.ros2.org/job/ci_linux-centos/87/)

clalancette:
The macOS failure also happens in the nightly repeated tests: https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/2153/#showFailuresLink

The 13 MSBuild warnings also happen on the latest Windows nightly: https://ci.ros2.org/view/nightly/job/nightly_win_rel/1749/

So I'm going to go ahead and merge this.  Thanks for the reviews.

Issue #1354, opened 2020-09-30
Add describe_parameters API to parameter_client


brawner:
Both of the async/sync versions of the ParameterClient do not have API for calling the describe_parameters service, but it seems like it would be a useful addition. The ParameterClient does have an `rclcpp::Client` for the describe_parameters service so this should be straightforward to implement.

fujitatomoya:
CC: @brawner @clalancette 

close this since https://github.com/ros2/rclcpp/pull/1453 is merged.

fujitatomoya:
@brawner 

i thought i opened this one...could you close this?

brawner:
Closed via #1453

PR #1453, opened 2020-11-12
support describe_parameters methods to parameter client.

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/1354

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -15,6 +15,8 @@
 #ifndef RCLCPP__PARAMETER_CLIENT_HPP_
 #define RCLCPP__PARAMETER_CLIENT_HPP_
 
+#include <functional>
+#include <future>
 #include <memory>
 #include <string>
 #include <utility>
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -121,6 +121,14 @@ public:
     > callback = nullptr);
 
   RCLCPP_PUBLIC
+  std::shared_future<std::vector<rcl_interfaces::msg::ParameterDescriptor>>
+  describe_parameters(
+    const std::vector<std::string> & names,
+    std::function<
+      void(std::shared_future<std::vector<rcl_interfaces::msg::ParameterDescriptor>>)
+    > callback = nullptr);
+
+  RCLCPP_PUBLIC
   std::shared_future<std::vector<rclcpp::ParameterType>>
   get_parameter_types(
     const std::vector<std::string> & names,
@@ -377,6 +385,10 @@ public:
         })
     );
   }
+
+  RCLCPP_PUBLIC
+  std::vector<rcl_interfaces::msg::ParameterDescriptor>
+  describe_parameters(const std::vector<std::string> & parameter_names);
 
   RCLCPP_PUBLIC
   std::vector<rclcpp::ParameterType>
--- a/rclcpp/src/rclcpp/parameter_client.cpp
+++ b/rclcpp/src/rclcpp/parameter_client.cpp
@@ -144,6 +144,41 @@ AsyncParametersClient::get_parameters(
   return future_result;
 }
 
+std::shared_future<std::vector<rcl_interfaces::msg::ParameterDescriptor>>
+AsyncParametersClient::describe_parameters(
+  const std::vector<std::string> & names,
+  std::function<
+    void(std::shared_future<std::vector<rcl_interfaces::msg::ParameterDescriptor>>)
+  > callback)
+{
+  auto promise_result =
+    std::make_shared<std::promise<std::vector<rcl_interfaces::msg::ParameterDescriptor>>>();
+  auto future_result = promise_result->get_future().share();
+
+  auto request = std::make_shared<rcl_interfaces::srv::DescribeParameters::Request>();
+  request->names = names;
+
+  describe_parameters_client_->async_send_request(
+    request,
+    [promise_result, future_result, callback](
+      rclcpp::Client<rcl_interfaces::srv::DescribeParameters>::SharedFuture cb_f)
+    {
+      std::vector<rcl_interfaces::msg::ParameterDescriptor> descriptors;
+      auto & descs = cb_f.get()->descriptors;
+      for (auto & desc : descs) {
+        descriptors.push_back(
+          static_cast<rcl_interfaces::msg::ParameterDescriptor>(desc));
+      }
+      promise_result->set_value(descriptors);
+      if (callback != nullptr) {
+        callback(future_result);
+      }
+    }
+  );
+
+  return future_result;
+}
+
 std::shared_future<std::vector<rclcpp::ParameterType>>
 AsyncParametersClient::get_parameter_types(
   const std::vector<std::string> & names,
@@ -335,6 +370,22 @@ SyncParametersClient::has_parameter(const std::string & parameter_name)
   names.push_back(parameter_name);
   auto vars = list_parameters(names, 1);
   return vars.names.size() > 0;
+}
+
+std::vector<rcl_interfaces::msg::ParameterDescriptor>
+SyncParametersClient::describe_parameters(const std::vector<std::string> & parameter_names)
+{
+  auto f = async_parameters_client_->describe_parameters(parameter_names);
+
+  using rclcpp::executors::spin_node_until_future_complete;
+  if (
+    spin_node_until_future_complete(
+      *executor_, node_base_interface_,
+      f) == rclcpp::FutureReturnCode::SUCCESS)
+  {
+    return f.get();
+  }
+  return std::vector<rcl_interfaces::msg::ParameterDescriptor>();
 }
 
 std::vector<rclcpp::ParameterType>
--- a/rclcpp/src/rclcpp/parameter_client.cpp
+++ b/rclcpp/src/rclcpp/parameter_client.cpp
@@ -164,10 +164,9 @@ AsyncParametersClient::describe_parameters(
       rclcpp::Client<rcl_interfaces::srv::DescribeParameters>::SharedFuture cb_f)
     {
       std::vector<rcl_interfaces::msg::ParameterDescriptor> descriptors;
-      auto & descs = cb_f.get()->descriptors;
-      for (auto & desc : descs) {
-        descriptors.push_back(
-          static_cast<rcl_interfaces::msg::ParameterDescriptor>(desc));
+      std::vector<rcl_interfaces::msg::ParameterDescriptor> & descs = cb_f.get()->descriptors;
+      for (rcl_interfaces::msg::ParameterDescriptor & desc : descs) {
+        descriptors.push_back(desc);
       }
       promise_result->set_value(descriptors);
       if (callback != nullptr) {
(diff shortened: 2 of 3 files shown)


fujitatomoya:
@clalancette @brawner 

could you take a look when you got time?

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -120,6 +120,14 @@ class AsyncParametersClient
       void(std::shared_future<std::vector<rclcpp::Parameter>>)
     > callback = nullptr);
 
+  RCLCPP_PUBLIC
+  std::shared_future<std::vector<rcl_interfaces::msg::ParameterDescriptor>>
+  describe_parameters(
+    const std::vector<std::string> & names,
+    std::function<
+      void(std::shared_future<std::vector<rcl_interfaces::msg::ParameterDescriptor>>)
+    > callback = nullptr);
```
clalancette, line 129: Add includes to the top of this file for:

```
#include <functional>
#include <future>
```

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -144,6 +144,41 @@ AsyncParametersClient::get_parameters(
...
+  describe_parameters_client_->async_send_request(
+    request,
+    [promise_result, future_result, callback](
+      rclcpp::Client<rcl_interfaces::srv::DescribeParameters>::SharedFuture cb_f)
+    {
+      std::vector<rcl_interfaces::msg::ParameterDescriptor> descriptors;
+      auto & descs = cb_f.get()->descriptors;
+      for (auto & desc : descs) {
+        descriptors.push_back(
+          static_cast<rcl_interfaces::msg::ParameterDescriptor>(desc));
```
clalancette, line 170: Can we not use `auto` here?  I know that we use it elsewhere in this file, but it is really hard for me to tell what the types of all of this is (and hence why the `static_cast` is needed).  This would be much easier to read if we used the real types.

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -337,6 +372,22 @@ SyncParametersClient::has_parameter(const std::string & parameter_name)
...
+  auto f = async_parameters_client_->describe_parameters(parameter_names);
+
+  using rclcpp::executors::spin_node_until_future_complete;
+  if (
+    spin_node_until_future_complete(
+      *executor_, node_base_interface_,
+      f) == rclcpp::FutureReturnCode::SUCCESS)
+  {
+    return f.get();
+  }
```
clalancette, line 387: This is purely subjective, but I think this is kind of ugly.  What if instead we did:

```
rclcpp::FutureReturnCode future = spin_node_until_future_complete(*executor_, node_base_interface_, f);
if (future == rclcpp::FutureReturnCode::SUCCESS) {
  return f.get();
}
```

rclcpp/test/rclcpp/test_parameter_client.cpp
```diff
@@ -291,3 +291,43 @@ TEST_F(TestParameterClient, sync_parameter_get_parameter_types) {
...
+/*
+  Coverage for async describe_parameters
+ */
+TEST_F(TestParameterClient, async_parameter_describe_parameters) {
+  auto asynchronous_client = std::make_shared<rclcpp::AsyncParametersClient>(node);
+  bool callback_called = false;
+  auto callback = [&callback_called](
+    std::shared_future<std::vector<rcl_interfaces::msg::ParameterDescriptor>> result)
+    {
+      // We expect the result to be empty since we tried to get a parameter that didn't exist.
```
clalancette, line 304: Can we also add a test here to check when some parameters *do* exist?  That would cover more of the code, I think.

Also, what happens if some of the parameters exist, and some don't?  It might be worth adding a test for that as well.

rclcpp/test/rclcpp/test_parameter_client.cpp
```diff
@@ -291,3 +291,43 @@ TEST_F(TestParameterClient, sync_parameter_get_parameter_types) {
...
+/*
+  Coverage for sync describe_parameters
+ */
+TEST_F(TestParameterClient, sync_parameter_describe_parameters) {
+  node->declare_parameter("foo", 4);
+  auto synchronous_client = std::make_shared<rclcpp::SyncParametersClient>(node);
+  std::vector<std::string> names{"foo"};
+  std::vector<rcl_interfaces::msg::ParameterDescriptor> parameter_types =
+    synchronous_client->describe_parameters(names);
+  ASSERT_EQ(1u, parameter_types.size());
```
clalancette, line 327: And similarly, can we add a test for the `SyncParameterClient` where the parameters don't exist, and also if some exist and some don't?

clalancette (changes requested):

rclcpp/src/rclcpp/parameter_client.cpp
fujitatomoya: agree with you.

fujitatomoya:
@clalancette thanks for checking on this. I've addressed your comments, could you take a look?

clalancette (changes requested):
Overall, the code is looking good.  Thanks for the updates.

One test that is still missing is what happens if some of the requested parameters exist and some don't.  Something like:

```
node->declare_parameter("foo", 4);
node->declare_parameter("bar", "this is bar");
future = async_client->describe_parameters({"foo", "baz", "bar"}, callback);
```

What happens in this case?  Is it an error?  Do we only get the ones back that have been declared?  It would be good to have a test to tell us what to expect.

fujitatomoya:
> It would be good to have a test to tell us what to expect.

totally, i will add them to make sure.

fujitatomoya:
@clalancette 

I've added the test cases that you mentioned https://github.com/ros2/rclcpp/pull/1453#pullrequestreview-531420868

clalancette:
> I've added the test cases that you mentioned [#1453 (review)](https://github.com/ros2/rclcpp/pull/1453#pullrequestreview-531420868)

Great, thank you.  Now that we have the test cases, I'm wondering about the interface for the user.

In particular, if the user asks to get parameters that don't exist, they'll get back an empty vector (either through the callback or via the sync return value).  I guess that they *can* detect that an error occurred (by comparing the size of the list they passed in to the one they get back), but it is hard for them to tell how they failed.  What if we instead failed the future in this case?  At least that would give the user an indication that something went wrong.

fujitatomoya:
I believe that is why [allow_undeclared_](https://github.com/ros2/rclcpp/blob/de353f9e45ca2ef191708e557f217eda843ab46a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp#L197) is to change this behavior. with `allow_undeclared_ is true`, it will return descriptors as much as possible. In case of the parameters don't exist, it will fill in `rcl_interfaces::msg::ParameterDescriptor default_description` instead of throwing exception. And user application can see that `do_not_exist_parameter` returns with type.PARAMETER_NOT_SET.

I see what your concern is, but i am not sure if we could change `allow_undeclared_` into true as default. personally I would use that true as default, but there would be some other opinion.

at least, i think `allow_undeclared_ is true` cases should be tested here, i will add them.

fujitatomoya:
sorry i accidentally closes, never mind.

fujitatomoya:
@clalancette 

> at least, i think allow_undeclared_ is true cases should be tested here, i will add them.

done, could you check when you got time? thanks!

clalancette:
> at least, i think `allow_undeclared_ is true` cases should be tested here, i will add them.

Thanks, those are good tests to have.

But I'm still wondering about the interface in the default, `allow_undeclared` = false scenario.  In that case, if the user asks for a parameter that doesn't exist, the only indication that they've done something not allowed is that they get a vector of 0 size back.  While that is *a* signal, it is pretty subtle.

Let me ask a related question which may help to clarify the situation to me.  In the current `AsyncParametersClient` API, what happens if `allow_undeclared` = false, and the user calls `get_parameters`, `get_parameter_types`, or `set_parameters` on a list the contains parameters that don't exist?  What is the behavior?

fujitatomoya:
> Let me ask a related question which may help to clarify the situation to me. In the current AsyncParametersClient API, what happens if allow_undeclared = false, and the user calls get_parameters, get_parameter_types, or set_parameters on a list the contains parameters that don't exist? What is the behavior?

good question, it seems inconsistent behavior for user application...

- get_parameters

it returns the same size with request. and in case of non-exist parameters, it returns with NOT_SET.

https://github.com/ros2/rclcpp/blob/a7db1dcca219b69569e6da63eb6331bfec553cff/rclcpp/src/rclcpp/parameter_service.cpp#L44-L50

- get_parameter_types

same behavior with describe_parameters. it returns a vector of 0 size.

https://github.com/ros2/rclcpp/blob/a7db1dcca219b69569e6da63eb6331bfec553cff/rclcpp/src/rclcpp/parameter_service.cpp#L69-L71

- set_parameters

it returns the same size with request. and in case of undeclared parameters, it fills in `successful = false` with exception message. set_parameters process one-by-one, since there's no way to return a partial result.

https://github.com/ros2/rclcpp/blob/a7db1dcca219b69569e6da63eb6331bfec553cff/rclcpp/src/rclcpp/parameter_service.cpp#L90-L93

fujitatomoya:
@clalancette 

about the inconsistent behavior, i will create new issue and do follow-up. how does it sound to you? or would you want to have discussion in this thread?

fujitatomoya:
@clalancette 

> about the inconsistent behavior, i will create new issue and do follow-up. how does it sound to you? or would you want to have discussion in this thread?

https://github.com/ros2/rclcpp/issues/1482 is created as follow-up, besides there is inconsistency between rclpy and rclcpp. i would need to have discussion for the behavior.

fujitatomoya:
@clalancette friendly ping

clalancette:
Apologies, this fell completely off my radar.  Let me respond to some of your previous questions:

> about the inconsistent behavior, i will create new issue and do follow-up. how does it sound to you?

Yes, that sounds great.  I see you've already done this in #1482, thanks for doing that.

> besides there is inconsistency between rclpy and rclcpp

That's unfortunate too.  But let's follow up on all of that in #1482 as well.

So that leaves us with this PR.  I think what I missed earlier is that this PR isn't the one that actually defines the behavior of the code when non-existent parameters are passed.  This PR just passes along what the underlying `describe_parameters` does.  So from that perspective, this is OK.  Thanks for the patience, I'm going to do one more review pass here.

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -144,6 +144,40 @@ AsyncParametersClient::get_parameters(
...
+    request,
+    [promise_result, future_result, callback](
+      rclcpp::Client<rcl_interfaces::srv::DescribeParameters>::SharedFuture cb_f)
+    {
+      std::vector<rcl_interfaces::msg::ParameterDescriptor> descriptors;
+      std::vector<rcl_interfaces::msg::ParameterDescriptor> & descs = cb_f.get()->descriptors;
+      for (rcl_interfaces::msg::ParameterDescriptor & desc : descs) {
+        descriptors.push_back(desc);
+      }
+      promise_result->set_value(descriptors);
```
clalancette, line 171: Looking at this again, do we actually need to do this?  Isn't this just copying the data from one vector into another?  Instead, can we just do:

```
promise_result->set_value(cb_f.get()->descriptors);
```

(one thing I don't know about is the lifetime of the objects, so that bears thinking about)

clalancette (commented):
I have one question inline.  Once we answer that, I'll be happy with this.

rclcpp/src/rclcpp/parameter_client.cpp
fujitatomoya: Nice catch 👍 thanks for the suggestion!

fujitatomoya:
@clalancette 

all comments are resolved, requesting final check.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13262/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8192/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10977/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13291/)

PR #1457, opened 2020-11-15
Add Topic Statistics Options to be configured via Parameters

InitialDuan:
**What:** Implemented the feature in https://github.com/ros-tooling/roadmap/issues/1 of Meta-Ticket https://github.com/ros2/ros2/issues/917

**Why:** Instead requiring a user to manually enable and configure Topic Statistics via NodeOptions, allow them to enable and configure Topic Statistics through Parameters. It would greatly simplify the ability to use the Topic Statistics feature so that users don't have to modify their subscription options in code.

The following should be configurable at runtime:
- enable / disable
- publish period
- published statistics topic

**How:** For any subscriber, run 
```
ros2 run package_name executable_name [--ros-args -p stats_enabled:="a_bool_value"] [-p stats_period:="a_double_value"] [-p stats_topic:="a_string_value"]
```

**Examples with Nodes in [Topic Statistics Tutorial](https://index.ros.org/doc/ros2/Tutorials/Topics/Topic-Statistics-Tutorial/):** 
```
ros2 run cpp_pubsub listener --ros-args -p stats_enabled:=true -p stats_period:=0.5 -p stats_topic:=new_statistics_topic
ros2 run cpp_pubsub listener_with_topic_statistics --ros-args -p stats_enabled:=false
```

**Note:** 

1. The run-time configuration through Parameters has prioirty over code-configuration via NodeOptions. If the Parameters is not given, the default NodeOptions in code will be used if there is.
2. The type double is used for publish-period since I want to make the period smaller als 1 also possible.

 
**Feedback wanted:** It's my first time to try to contribute an open-source project. Any Feedback is appreciated! @dabonnie

### DIFF
--- a/rclcpp/include/rclcpp/create_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_subscription.hpp
@@ -122,8 +122,14 @@ create_subscription(
       rclcpp::topic_statistics::SubscriptionTopicStatistics<CallbackMessageT>
       >(node_topics->get_node_base_interface()->get_name(), publisher);
 
-    auto sub_call_back = [subscription_topic_stats]() {
-        subscription_topic_stats->publish_message();
+    std::weak_ptr<
+      rclcpp::topic_statistics::SubscriptionTopicStatistics<CallbackMessageT>
+    > weak_subscription_topic_stats(subscription_topic_stats);
+    auto sub_call_back = [weak_subscription_topic_stats]() {
+        auto subscription_topic_stats = weak_subscription_topic_stats.lock();
+        if (subscription_topic_stats) {
+          subscription_topic_stats->publish_message();
+        }
       };
 
     auto node_timer_interface = node_topics->get_node_timers_interface();
--- a/rclcpp/include/rclcpp/create_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_subscription.hpp
@@ -100,50 +100,66 @@ create_subscription(
   
   using rclcpp::node_interfaces::get_node_parameters_interface; //added
   auto node_parameters = get_node_parameters_interface(std::forward<NodeT>(node)); //added
-  node_parameters->declare_parameter("enable_statistics", rclcpp::ParameterValue(false)); //added  
   
-  if (rclcpp::detail::resolve_enable_topic_statistics(options,*node_topics->get_node_base_interface())||
-  node_parameters->get_parameter("enable_statistics").as_bool()) //added
-  {
-    if (options.topic_stats_options.publish_period <= std::chrono::milliseconds(0)) {
-      throw std::invalid_argument(
-              "topic_stats_options.publish_period must be greater than 0, specified value of " +
-              std::to_string(options.topic_stats_options.publish_period.count()) +
-              " ms");
-    }
+  node_parameters->declare_parameter("stats_enabled", rclcpp::ParameterValue()); //added  
+  rclcpp::Parameter stats_enabled; //added
+  node_parameters->get_parameter("stats_enabled", stats_enabled); //added
+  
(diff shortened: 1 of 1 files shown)


InitialDuan:
About Jenkins build failure: I think the stderr is comming from when including the new header files in my create_subscription.hpp
``` 
#include "rclcpp/node_interfaces/get_node_parameters_interface.hpp" //Line 29
#include "rclcpp/node_interfaces/node_parameters_interface.hpp" //Line 32
```
Does including header files need any other steps? How could it be solved?

clalancette (changes requested):
There's a whole slew of style problems in here; that's why the PR job is failing.  Please make sure to run the tests locally before opening up a PR (`colcon test --packages-select rclcpp` will help you here).

While I understand the want for this, it seems kind of weird to me to have a feature that can both be configured in the `NodeOptions` and changed via parameters.  I'm wondering if we should deprecate the ability to set it via `NodeOptions` and just allow it to be set via parameters.  But I haven't thought through all of the possibilities, so maybe others can chime in why we might want both.

InitialDuan:
> There's a whole slew of style problems in here; that's why the PR job is failing. Please make sure to run the tests locally before opening up a PR (`colcon test --packages-select rclcpp` will help you here).

ok, locally runing ```colcon test --packages-select rclcpp```  just shows the build fails without any error information, but ```colcon build``` shows the same error messages as Jenkins. 

But I still don't get it, I just included and used ```get_node_parameters_interface``` in the same way as the ```get_node_topics_interface``` used in the original create_subscription.hpp file, and building other packages which includes ```rclcpp/create_subscription.hpp``` works properly. Can you point out what kind of style problem exactly?

clalancette:
Looking at the PR job output, it doesn't look like the `#include` is what is failing, it looks like the instantiation is failing.  My guess is that something with the types being passed to `get_node_parameters_interface` is wrong, though I'm not sure about that.  It needs further investigation.

fujitatomoya (commented):
externally and runtime configuration sounds good to me. I have a couple of questions though.

```
ros2 run package_name executable_name [--ros-args -p stats_enabled:="a_bool_value"] ...
```

- what if there are multiple nodes inside the process space? it enables statistics for all nodes?

> The run-time configuration through Parameters has prioirty over code-configuration via NodeOptions.

runtime configuration prevails NodeOptions seems to be reasonable. using parameter at runtime, it can be configurable for each node?



to think about the actual use case for application development, we would do the following.

- start all nodes w/o statistics.
- something wrong / debugging...
- enable statistics on specific nodes to make sure data delivery.

with that use case, external setting via parameter(or probably service) after node starts could be useful, but i don't know if it is necessary to add argument parameter.

InitialDuan:
> what if there are multiple nodes inside the process space? it enables statistics for all nodes?

For now, the statistics has to be enabled node by node, an enable / disable switch for system-wide Topic Statistics would be one of the next steps according to Meta-Ticket https://github.com/ros2/ros2/issues/917.

> runtime configuration prevails NodeOptions seems to be reasonable. using parameter at runtime, it can be configurable for each node?
> 
Yes, the things goes like before, for each subscriber, when it calls the create_subscription() function, the statistics will be published by itstelf if enabled.

> to think about the actual use case for application development, we would do the following.
> 
> * start all nodes w/o statistics.
> * something wrong / debugging...
> * enable statistics on specific nodes to make sure data delivery.
> 
> with that use case, external setting via parameter(or probably service) after node starts could be useful, but i don't know if it is necessary to add argument parameter.

Yes, I have also read your proposals in Meta-Ticket https://github.com/ros2/ros2/issues/917. What I pushed is just to realize enabling Topic Statistics Options to be configured via Parameters as the frist step.

clalancette:
@InitialDuan Friendly ping.  This needs to be rebased onto the latest, and also needs to have all commits signed for the DCO bot.

PR #1461, opened 2020-11-16
[rclcpp_lifecycle] Change uint8_t iterator variables to size_t

brawner:
This is a small theoretical issue I found. The type of `states_size` and `transitions_size` is `unsigned int`, but the for loop int type is `uint8_t`. While `default_state_machine` has only 11 states and 25 transitions, looping with a small integer type that is compared to a larger type could lead to an infinite loop. While `size_t` here is large enough to avoid problems, I'd also be happy to use also `unsigned int`.

### DIFF
--- a/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
+++ b/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
@@ -265,7 +265,7 @@ public:
       throw std::runtime_error(
               "Can't get available states. State machine is not initialized.");
     }
-    for (size_t i = 0; i < state_machine_.transition_map.states_size; ++i) {
+    for (unsigned int i = 0; i < state_machine_.transition_map.states_size; ++i) {
       lifecycle_msgs::msg::State state;
       state.id = static_cast<uint8_t>(state_machine_.transition_map.states[i].id);
       state.label = static_cast<std::string>(state_machine_.transition_map.states[i].label);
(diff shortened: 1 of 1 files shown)


brawner:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13074/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8019/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10796/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13054/)

clalancette:
> I'd also be happy to use also unsigned int.

Given that [states_size](https://github.com/ros2/rcl/blob/df5c8d0f0a697158a10ef57bd3b3074d168961a7/rcl_lifecycle/include/rcl_lifecycle/data_types.h#L63) is defined as an `unsigned int`, I suggest we stick with that type here.  It just ensures that we are comparing the same type always.  I don't think this change has any practical difference on the platforms we support, but it makes it easy to tell that we are doing the correct thing.

brawner:
Sounds good, fixed.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13088/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8027/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10805/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13065/)

clalancette (approved):

PR #1465, opened 2020-11-17
Qos configurability (take 2)

ivanpauno:
Reapply #1408.

Fixes:
- Windows visibility issues.
- Windows ABI issues.
  Structs with public members that are from `std` generate a lot of warnings on Windows, so `QosOverridingOptions` is now a `class`.
- Fixes bug in `check_if_stringified_policy_is_null`.

Cosmetic changes:
- `QosOverridingOptions::with_default_policies(...)` instead of `QosOverridingOptions{QosOverridingOptions::kDefaultPolicies, ...}`

I will try to run CI in the correct branch this time :joy:

### DIFF
--- a/rclcpp/include/rclcpp/create_publisher.hpp
+++ b/rclcpp/include/rclcpp/create_publisher.hpp
@@ -54,7 +54,7 @@ create_publisher(
 )
 {
   auto node_topics_interface = rclcpp::node_interfaces::get_node_topics_interface(node_topics);
-  const rclcpp::QoS & actual_qos = options.qos_overriding_options.policy_kinds.size() ?
+  const rclcpp::QoS & actual_qos = options.qos_overriding_options.get_policy_kinds().size() ?
     rclcpp::detail::declare_qos_parameters(
     options.qos_overriding_options, node_parameters,
     node_topics_interface->resolve_topic_name(topic_name),
--- a/rclcpp/include/rclcpp/create_publisher.hpp
+++ b/rclcpp/include/rclcpp/create_publisher.hpp
@@ -17,6 +17,7 @@
 
 #include <memory>
 #include <string>
+#include <utility>
 
 #include "rclcpp/node_interfaces/get_node_topics_interface.hpp"
 #include "rclcpp/node_interfaces/node_topics_interface.hpp"
@@ -24,10 +25,56 @@
 #include "rclcpp/publisher_factory.hpp"
 #include "rclcpp/publisher_options.hpp"
 #include "rclcpp/qos.hpp"
+#include "rclcpp/qos_overriding_options.hpp"
+#include "rclcpp/detail/qos_parameters.hpp"
+
 #include "rmw/qos_profiles.h"
 
 namespace rclcpp
 {
+
+namespace detail
+{
+/// Create and return a publisher of the given MessageT type.
+template<
+  typename MessageT,
+  typename AllocatorT = std::allocator<void>,
+  typename PublisherT = rclcpp::Publisher<MessageT, AllocatorT>,
+  typename NodeParametersT,
+  typename NodeTopicsT>
+std::shared_ptr<PublisherT>
+create_publisher(
+  NodeParametersT & node_parameters,
+  NodeTopicsT & node_topics,
+  const std::string & topic_name,
+  const rclcpp::QoS & qos,
+  const rclcpp::PublisherOptionsWithAllocator<AllocatorT> & options = (
+    rclcpp::PublisherOptionsWithAllocator<AllocatorT>()
+  )
+)
+{
+  auto node_topics_interface = rclcpp::node_interfaces::get_node_topics_interface(node_topics);
+  const rclcpp::QoS & actual_qos = options.qos_overriding_options.policy_kinds.size() ?
+    rclcpp::detail::declare_qos_parameters(
+    options.qos_overriding_options, node_parameters,
+    node_topics_interface->resolve_topic_name(topic_name),
+    qos, rclcpp::detail::PublisherQosParametersTraits{}) :
+    qos;
+
+  // Create the publisher.
+  auto pub = node_topics_interface->create_publisher(
+    topic_name,
+    rclcpp::create_publisher_factory<MessageT, AllocatorT, PublisherT>(options),
+    actual_qos
+  );
+
+  // Add the publisher to the node topics interface.
+  node_topics_interface->add_publisher(pub, options.callback_group);
+
+  return std::dynamic_pointer_cast<PublisherT>(pub);
+}
+}  // namespace detail
+
 
 /// Create and return a publisher of the given MessageT type.
 /**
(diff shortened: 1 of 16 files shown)


ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13083/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8025/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10803/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13063/)

rclcpp/include/rclcpp/qos_overriding_options.hpp
```diff
@@ -0,0 +1,154 @@
...
+ * ```
+ */
+class QosOverridingOptions
+{
+public:
+  /// Default constructor, no overrides allowed.
+  RCLCPP_PUBLIC
+  QosOverridingOptions() = default;
+
+  /// Construct passing a list of qos policies that and a verification callback.
```
jacobperron, line 96:
```suggestion
  /// Construct passing a list of QoS policies and a verification callback.
```

rclcpp/include/rclcpp/qos_overriding_options.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+  RCLCPP_PUBLIC
+  const std::vector<QosPolicyKind> &
+  get_policy_kinds() const;
+
+  RCLCPP_PUBLIC
+  const QosCallback &
+  get_validation_callback() const;
+
+  /// Construct passing a list of qos policies that and a verification callback.
```
jacobperron, line 132:
```suggestion
  /// Construct passing a list of QoS policies and a verification callback.
```

rclcpp/include/rclcpp/qos_overriding_options.hpp
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * Same that `QosOverridingOptions` constrctor, but only declares the default policies:
```
jacobperron, line 134:
```suggestion
   * Same as `QosOverridingOptions` constructor, but only declares the default policies:
```

jacobperron (approved):
Other than some typos in comments, LGTM

rclcpp/include/rclcpp/create_publisher.hpp
```diff
@@ -17,54 +17,108 @@
...
 
   return std::dynamic_pointer_cast<PublisherT>(pub);
 }
+}  // namespace detail
+
+
+/// Create and return a publisher of the given MessageT type.
+/**
+ * The NodeT type only needs to have a method called get_node_topics_interface()
+ * which returns a shared_ptr to a NodeTopicsInterface.
```
wjwwood, line 82: I think this needs to be updated.

rclcpp/include/rclcpp/qos_overriding_options.hpp
```diff
@@ -0,0 +1,154 @@
...
+ *      /my/topic/name:
+ *        publisher:  # publisher without provided id
+ *          reliability: reliable
+ *          depth: 100
+ *        publisher_my_id:  # publisher with `id="my_id"
+ *          reliability: reliable
+ *          depth: 10
+ * ```
+ */
+class QosOverridingOptions
```
jacobperron, line 89: I just noticed that we're not consistent with the case used for the name `Qos`. In rclcpp, when we write CamelCase there's a mix of capitalizing the "s" and leaving it lowercase. I.e. `QoS` vs `Qos`. We should probably be consistent, though I'm not sure which is the preferred casing.

I'm pointing it out here since this PR introduces several instances of using a lowercase "s", `Qos`. Though there were mixed conventions even before this PR.

jacobperron: In rclpy, we seem to be more consistent in using `QoS`, with the exception of changes introduced by this feature.

wjwwood: My preference is `QoS`, but according to the google style guide, only the first letter of an acronym is supposed to be capitalized, so it would be `RosThing` rather than `ROSThing`, which I don't care for, but maybe that's where this is coming from.

https://google.github.io/styleguide/cppguide.html#General_Naming_Rules
> For the purposes of the naming rules below, a "word" is anything that you would write in English without internal spaces. This includes abbreviations, such as acronyms and initialisms. For names written in mixed case (also sometimes referred to as "camel case" or "Pascal case"), in which the first letter of each word is capitalized, prefer to capitalize abbreviations as single words, e.g., StartRpc() rather than StartRPC().

I definitely prefer QoS though, because otherwise it's ambiguous as to where or not the `o` would be capitalized if the acronym were to be taken out of the class name.

jacobperron: I also prefer `QoS` (and `ROSThing` over `RosThing`), I think simply because that's how I would write it in other contexts. But I'd like see consistency at the very least.

Also, I think in this case it will be less disruptive to users if we change the APIs introduced in PR to match most of the other code using `QoS`, since this is something new not yet released into a ROS distro (except Rolling).

wjwwood: Agreed.

ivanpauno: I think it's ok to modify all instances of `QosOverridingOptions` are modified to `QOSOverridingOptions` or `QoSOverridingOptions`.

We already have some instances of `QOS` in rclcpp (see [here](https://github.com/ros2/rclcpp/blob/dc7e95aaa524f758ff8ccf903698065420f9cc48/rclcpp/include/rclcpp/qos_event.hpp#L84)), so I prefer that one.

I find `QoS` weird, e.g. a CamelCase to snake case converter is going to get it wrong (qo_s).
I also find QOSOverridingOptions hard to read, but better than QoSOverridingOptions.
e.g. most CamelCase to snake_case converters will get the first one right `qos_overriding_options`, whereas the second one would be `qo_s_overriding_options`.

jacobperron: > I find QoS weird, e.g. a CamelCase to snake case converter is going to get it wrong (qo_s).

That's a reasonable argument.

Though, if we're choosing between `Qos` and `QOS`, I'd rather us just follow the google style guide ([as documented](https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#id3)) and pick `Qos`, since refactoring will be disruptive either way.

wjwwood: > I find QoS weird, e.g. a CamelCase to snake case converter is going to get it wrong (qo_s).

I don't feel that QoS is weird as that is the initialization outside of the class name. Whether or not a camel case to snake case converter is smart enough to handle this isn't really important imo. We only use it with the Messages to message header conversion as far as I know and even that was a bad idea in my mind. The convert is never going to be quite right. There are already other cases it fails for, there was an issue about it... but it's something like `Point3D` becomes `point_3_d`. These names are inherently contextual.

> Though, if we're choosing between Qos and QOS, I'd rather us just follow the google style guide (as documented) and pick Qos, since refactoring will be disruptive either way.

But refactoring new code is less disruptive than existing code. If we're gonna do this I think we should list all the uses of QOS, QoS, and Qos and consider which would be most disruptive for users.

clalancette: > But refactoring new code is less disruptive than existing code. If we're gonna do this I think we should list all the uses of QOS, QoS, and Qos and consider which would be most disruptive for users.

I agree whole-heartedly with this.  I think consistency with what is already there is more important than any particular spelling.

PR #1467, opened 2020-11-18
Add getters to rclcpp::qos and rclcpp::*Policy enum classes

ivanpauno:
This will be useful in things like https://github.com/ros2/demos/pull/474.

### DIFF
--- a/rclcpp/include/rclcpp/qos.hpp
+++ b/rclcpp/include/rclcpp/qos.hpp
@@ -30,6 +30,38 @@ namespace rclcpp
 RCLCPP_PUBLIC
 std::string qos_policy_name_from_kind(rmw_qos_policy_kind_t policy_kind);
 
+enum class HistoryPolicy
+{
+  KeepLast = RMW_QOS_POLICY_HISTORY_KEEP_LAST,
+  KeepAll = RMW_QOS_POLICY_HISTORY_KEEP_ALL,
+  SystemDefault = RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT,
+  Unknown = RMW_QOS_POLICY_HISTORY_UNKNOWN,
+};
+
+enum class ReliabilityPolicy
+{
+  BestEffort = RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT,
+  Reliable = RMW_QOS_POLICY_RELIABILITY_RELIABLE,
+  SystemDefault = RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT,
+  Unknown = RMW_QOS_POLICY_RELIABILITY_UNKNOWN,
+};
+
+enum class DurabilityPolicy
+{
+  Volatile = RMW_QOS_POLICY_DURABILITY_VOLATILE,
+  TransientLocal = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
+  SystemDefault = RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT,
+  Unknown = RMW_QOS_POLICY_DURABILITY_UNKNOWN,
+};
+
+enum class LivelinessPolicy
+{
+  Automatic = RMW_QOS_POLICY_LIVELINESS_AUTOMATIC,
+  ManualByTopic = RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_TOPIC,
+  SystemDefault = RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT,
+  Unknown = RMW_QOS_POLICY_LIVELINESS_UNKNOWN,
+};
+
 /// QoS initialization values, cannot be created directly, use KeepAll or KeepLast instead.
 struct RCLCPP_PUBLIC QoSInitialization
 {
@@ -84,6 +116,10 @@ public:
 
   /// Set the history policy.
   QoS &
+  history(HistoryPolicy history);
+
+  /// Set the history policy.
+  QoS &
   history(rmw_qos_history_policy_t history);
 
   /// Set the history to keep last.
@@ -98,6 +134,10 @@ public:
   QoS &
   reliability(rmw_qos_reliability_policy_t reliability);
 
+  /// Set the reliability setting.
+  QoS &
+  reliability(ReliabilityPolicy reliability);
+
   /// Set the reliability setting to reliable.
   QoS &
   reliable();
@@ -109,6 +149,10 @@ public:
   /// Set the durability setting.
   QoS &
   durability(rmw_qos_durability_policy_t durability);
+
+  /// Set the durability setting.
+  QoS &
+  durability(DurabilityPolicy durability);
 
   /// Set the durability setting to volatile.
   /**
@@ -141,6 +185,10 @@ public:
   QoS &
   liveliness(rmw_qos_liveliness_policy_t liveliness);
 
+  /// Set the liveliness setting.
+  QoS &
+  liveliness(LivelinessPolicy liveliness);
+
   /// Set the liveliness_lease_duration setting.
   QoS &
   liveliness_lease_duration(rmw_time_t liveliness_lease_duration);
@@ -153,16 +201,16 @@ public:
   QoS &
   avoid_ros_namespace_conventions(bool avoid_ros_namespace_conventions);
 
-  rmw_qos_history_policy_t
+  HistoryPolicy
   get_history() const;
 
   size_t
   get_depth() const;
 
-  rmw_qos_reliability_policy_t
+  ReliabilityPolicy
   get_reliability() const;
 
-  rmw_qos_durability_policy_t
+  DurabilityPolicy
   get_durability() const;
 
   rclcpp::Duration
@@ -171,7 +219,7 @@ public:
   rclcpp::Duration
   get_lifespan() const;
 
-  rmw_qos_liveliness_policy_t
+  LivelinessPolicy
   get_liveliness() const;
 
   rclcpp::Duration
--- a/rclcpp/include/rclcpp/qos.hpp
+++ b/rclcpp/include/rclcpp/qos.hpp
@@ -153,6 +153,33 @@ public:
   QoS &
   avoid_ros_namespace_conventions(bool avoid_ros_namespace_conventions);
 
+  rmw_qos_history_policy_t
+  get_history() const;
+
+  size_t
+  get_depth() const;
+
+  rmw_qos_reliability_policy_t
+  get_reliability() const;
+
+  rmw_qos_durability_policy_t
+  get_durability() const;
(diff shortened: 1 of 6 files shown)


rclcpp/include/rclcpp/qos.hpp
```diff
@@ -153,6 +201,33 @@ class RCLCPP_PUBLIC QoS
   QoS &
   avoid_ros_namespace_conventions(bool avoid_ros_namespace_conventions);
 
+  HistoryPolicy
+  get_history() const;
```
wjwwood, line 205: Why not:

```c++
HistoryPolicy
history() const;
```

That would follow the style set already by `NodeOptions`, e.g.:

https://github.com/ros2/rclcpp/blob/ba1056fe63590e4253cf362f766fce46be860086/rclcpp/include/rclcpp/node_options.hpp#L100-L114

Also, all the accessors need at least a minimal docstring brief.

wjwwood (changes requested):

rclcpp/include/rclcpp/qos.hpp
ivanpauno: I avoided that because I felt the difference between `reliable()` being a setter and `reliability()` being a getter is subtle, but it's true it's more consistent with other places.

wjwwood: Well, let's not introduce inconsistency. Let's either keep it consistent with what is there or update all of them. Personally I think the existing style is fine, but if others agree with you, I'm fine with changing it.

ivanpauno: > Well, let's not introduce inconsistency. Let's either keep it consistent with what is there or update all of them. Personally I think the existing style is fine, but if others agree with you, I'm fine with changing it.

Yeah, I agree inconsistency isn't nice, done in https://github.com/ros2/rclcpp/pull/1467/commits/18c71adfcb46e4cc6a6299e59cb1db30b280bbdc.

My preference for the explicit `get_` prefix is probably a non-native speaker issue, my mind has problems to process the difference between the noun and the adjective (reliability/reliable) at first glance.

wjwwood (approved):

wjwwood:
lgtm, thanks for iterating!

wjwwood:
CI (fastrtps only):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13189/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8120/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10904/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13203/)

wjwwood:
I believe the cmake warnings on Windows are being addressed separately. They show up on https://ci.ros2.org/view/nightly/job/nightly_win_rel/1768/cmake/.

How should we test for coverage? Do we have that documented somewhere yet? @ivanpauno or @brawner maybe?

brawner:
Coverage can be measured locally with instructions at https://index.ros.org/doc/ros2/Contributing/ROS-2-On-boarding-Guide/#how-to-measure-coverage-locally-using-lcov-ubuntu

It's not simple for rclcpp though because there are many directories. You mostly want to make sure that generated html files show that you've covered all code you introduced. 

All the cmake warnings have been resolved, except zstd. It's been patched but not released.

fujitatomoya (approved):

wjwwood:
> All the cmake warnings have been resolved, except zstd. It's been patched but not released.

I don't understand this comment, so should we not see cmake warnings right now then? What is `zstd`?

clalancette:
> I don't understand this comment, so should we not see cmake warnings right now then? What is `zstd`?

If you look at the [latest nightly](https://ci.ros2.org/view/nightly/job/nightly_win_rel/1769/), you'll see that there are still 2 CMake warnings: one for zstd_vendor, and one for Fast-CDR.  zstd_vendor is a package that rosbag2 uses for compression; I think @brawner fixed the CMake problem with that package in the upstream code, but there has been no release with the fix.  We still have to fix Fast-CDR; I'll look at that today.

In the CI run above, there are 5 CMake warnings on Windows.  That includes the 2 that are still "live".  The other 3 look to be problems that were fixed after this CI build was launched, so I'd say that this is good to go.

wjwwood:
Thanks for the explanation @clalancette.

I haven't had time to run the coverage locally, should we wait for that from me or someone else? (we really need an automated job for that) Or just merge it? @ivanpauno?

ivanpauno:
> I haven't had time to run the coverage locally, should we wait for that from me or someone else? (we really need an automated job for that) Or just merge it? @ivanpauno?

I'm checking coverage locally.
I have found the workflow using [colcon-lcov-result](https://github.com/colcon/colcon-lcov-result) more or less easy to run.
I guess that does exactly the same of what's described in the `contributing guide`.

rclcpp/include/rclcpp/qos.hpp
```diff
@@ -98,6 +134,10 @@ class RCLCPP_PUBLIC QoS
   QoS &
   reliability(rmw_qos_reliability_policy_t reliability);
 
+  /// Set the reliability setting.
+  QoS &
+  reliability(ReliabilityPolicy reliability);
```
ivanpauno, line 139: needs coverage

rclcpp/include/rclcpp/qos.hpp
```diff
@@ -110,6 +150,10 @@ class RCLCPP_PUBLIC QoS
   QoS &
   durability(rmw_qos_durability_policy_t durability);
 
+  /// Set the durability setting.
+  QoS &
+  durability(DurabilityPolicy durability);
```
ivanpauno, line 155: needs coverage

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -246,6 +246,26 @@ Duration::to_rmw_time() const
   return result;
 }
 
+Duration
+Duration::from_rmw_time(rmw_time_t duration)
+{
+  Duration ret;
+  constexpr rcl_duration_value_t limit = std::numeric_limits<rcl_duration_value_t>::max();
+  if (duration.sec > limit || duration.nsec > limit) {
```
ivanpauno, line 254: needs coverage

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -246,6 +246,26 @@ Duration::to_rmw_time() const
...
+    // saturate if will overflow
+    ret.rcl_duration_.nanoseconds = limit;
+    return ret;
+  }
+  uint64_t total_ns = RCL_S_TO_NS(duration.sec) + duration.nsec;
+  if (total_ns < duration.nsec) {
```
ivanpauno, line 260: needs coverage

ivanpauno (commented):
adding notes about what needs extra tests, so I don't forget ..

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13246/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8177/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10961/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13279/)

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -246,6 +246,26 @@ Duration::to_rmw_time() const
   return result;
 }
 
+Duration
+Duration::from_rmw_time(rmw_time_t duration)
```
wjwwood, line 250: Why does the logic here differ so much from `Duration::Duration(const builtin_interfaces::msg::Duration & duration_msg)`? It seems like they do the same thing in principle.

rclcpp/src/rclcpp/duration.cpp
wjwwood, line 254: Should it be something like:

```suggestion
  if (RCL_S_TO_NS(duration.sec) > limit || duration.nsec > limit) {
```

Just because the value of `.sec` could be much smaller than `limit` but still overflow `ret.rcl_duration_.nanoseconds`.

rclcpp/test/rclcpp/test_duration.cpp
```diff
@@ -148,6 +148,34 @@ TEST_F(TestDuration, from_seconds) {
     rclcpp::Duration::from_seconds(-1.5));
 }
 
+TEST_F(TestDuration, from_rmw_time) {
+  constexpr auto max_rcl_duration = std::numeric_limits<rcl_duration_value_t>::max();
+  {
+    rmw_time_t rmw_duration{};
+    rmw_duration.sec = max_rcl_duration + 1uLL;
```
wjwwood, line 155: Seems like there is a value of `rmw_duration.sec` that is less than `max_rcl_duration` which will still overflow the duration. This is related to my other comment. Like `rmw_duration.sec = max_rcl_duration / 10uLL;` will likely cause the problem because `RCL_NS_TO_S` will multiply it by much more.

wjwwood (commented):
mostly lgtm, but I had some concerns about integer logic, but maybe I'm misreading it.

rclcpp/src/rclcpp/duration.cpp
ivanpauno: The logic is the "same", but considering that in `builtin_interfaces::msg::Duration` seconds is an `int32_t` and nanoseconds an `uint32_t`, the result represented as nanoseconds always fits in an `int64_t` (`rcl_duration_value_t`).

Whereas `rmw_time_t` seconds and nanoseconds are both a `uint64_t`, and the result in nanoseconds might not fit in `int64_t`.

wjwwood: Makes sense. 👍

rclcpp/src/rclcpp/duration.cpp
ivanpauno: thanks, solved in https://github.com/ros2/rclcpp/pull/1467/commits/655f76e14c95876951e268c7c3d63bc185ede1a1

rclcpp/test/rclcpp/test_duration.cpp
ivanpauno: https://github.com/ros2/rclcpp/pull/1467/commits/655f76e14c95876951e268c7c3d63bc185ede1a1 should cover this case correctly now.

wjwwood (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13250/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8180/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10963/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13281/)

Issue #1468, opened 2020-11-19
SEGV caused by order of destruction of Node sub-interfaces


guru-florida:
## Bug report
The SEGV is caused due to some sub-iface destructor code expecting living references to the node base or other interface but it was already released earlier in the Node destructor. Memory corruption occurs and eventually malloc checks abort later in the execution, usually somewhere in malloc_consolidate, FastRTPS library or waitfor collection resizing.

The destructor for the Node class is empty. Sure, all the sub-interfaces are shared_ptr and thus their destructors are called but many of these sub-interfaces reference each other. For example, all of the sub-interfaces reference at least the NodeBaseInterface so it must be released last. Furthermore, clock and time_source sub ifaces reference many others so those other dependencies should also stay alive until the clock and time_source is released. 

**NodeParametersInterface**
For example, the NodeParametersInterface contains parameter service calls that captures a weak reference to the base interface. During destruction it de-references the weak reference, finds it is null and spits out this error:
```
[ERROR] [1605744975.811485921] [rclcpp]: Error in destruction of rcl service handle: the Node Handle was destructed too early. You will leak memory
```
NodeParametersInterface at least uses as weakptr to base iface and thus fails gracefully (with a leak) but I expect some of the other sub ifaces may not. I see many examples in my searches of similar crash backtraces as I've encountered, explained below.

- Operating System:
  - Linux/Ubuntu 20.04  (probably all OSs)
- Installation type:
  - latest apt binaries
- Version or commit hash:
  - foxy, latest
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
The nature of these types of mem corruption can be hard to pin down since the debugger/malloc doesnt trip until much later in the execution. I encountered this error while creating and destroying many Nodes at runtime. Specifically, I wrote a node to do episodic parameter tuning/training for a humanoid robot in Gazebo simulation. For each episode, ROS nodes are created then torn-down (cycled). The SEGV would occur somewhere between episode 1 and ~15 and always occur immediately after a tear-down. It got progressively worse as I added more nodes to simulation that are cycled per episode. I would say any such program that continuously creates and destroys nodes will replicate this issue, such as the one in [Issue #447](https://github.com/ros2/rclcpp/issues/447) (I will test this and report).

#### Related Issues
[...]

PR #1469, opened 2020-11-19
Fix SEGV caused by order of destruction of Node sub-interfaces

guru-florida:
I think this is high impact but may have gone somewhat unnoticed since most ROS setups probably only tear-down Nodes on termination. Please see description and related issues in #1468.

> SEGV is caused due to some sub-iface destructor code expecting living references to the node base or other interface but it was already released earlier in the Node destructor. Memory corruption occurs and eventually malloc checks abort later in the execution, usually somewhere in malloc_consolidate, FastRTPS library or waitfor collection resizing.

I have based this fix to foxy branch as I think it should be applied there and then brought forward to master. I can rebase to master though if you want.

This fix explicitly releases the shared_ptrs to sub interfaces in the reverse order of creation. This ensures that the heavy dependencies between the sub-interfaces exist during tear-down. Especially the base interface, but also true of many others such as topics, waitables, graph, services, clock and logging interfaces.

### DIFF
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -180,7 +180,7 @@ public:
     std::weak_ptr<rcl_node_t> weak_node_handle(node_handle_);
     // rcl does the static memory allocation here
     service_handle_ = std::shared_ptr<rcl_service_t>(
-      new rcl_service_t, [weak_node_handle](rcl_service_t * service)
+      new rcl_service_t, [weak_node_handle, service_name](rcl_service_t * service)
       {
         auto handle = weak_node_handle.lock();
         if (handle) {
@@ -192,10 +192,10 @@ public:
             rcl_reset_error();
           }
         } else {
-          RCLCPP_ERROR(
+          RCLCPP_ERROR_STREAM(
             rclcpp::get_logger("rclcpp"),
-            "Error in destruction of rcl service handle: "
-            "the Node Handle was destructed too early. You will leak memory");
+            "Error in destruction of rcl service handle " << service_name <<
+              ": the Node Handle was destructed too early. You will leak memory");
         }
         delete service;
       });
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -185,7 +185,18 @@ Node::Node(
 }
 
 Node::~Node()
-{}
+{
+  // release sub-interfaces in an order that allows them to consult with node_base during tear-down
+  node_waitables_.reset();
+  node_time_source_.reset();
+  node_parameters_.reset();
+  node_clock_.reset();
+  node_services_.reset();
+  node_topics_.reset();
+  node_timers_.reset();
+  node_logging_.reset();
+  node_graph_.reset();
+}
 
 const char *
 Node::get_name() const
--- a/rclcpp_lifecycle/src/lifecycle_node.cpp
+++ b/rclcpp_lifecycle/src/lifecycle_node.cpp
@@ -128,7 +128,18 @@ LifecycleNode::LifecycleNode(
 }
 
 LifecycleNode::~LifecycleNode()
-{}
+{
+  // release sub-interfaces in an order that allows them to consult with node_base during tear-down
+  node_waitables_.reset();
+  node_time_source_.reset();
+  node_parameters_.reset();
+  node_clock_.reset();
+  node_services_.reset();
+  node_topics_.reset();
+  node_timers_.reset();
+  node_logging_.reset();
+  node_graph_.reset();
+}
 
 const char *
 LifecycleNode::get_name() const


clalancette:
> I have based this fix to foxy branch as I think it should be applied there and then brought forward to master. I can rebase to master though if you want.

Our usual development flow is to apply patches to master, and then backport to stable releases as appropriate.  So if you don't mind, rebasing on master and then retargeting this PR to master would be the right way to go here.  Thanks.

guru-florida:
@cottsay Can you take a quick look at my failed check? It is coming up as a syntax error in benchmark/benchmark_lifecycle_node.cpp and the PERFORMANCE_TEST_FIXTURE_PAUSE_MEASUREMENTS macro. This code seems to be a recent commit you made last night. It's not code I affected, though this seems to have passed in your PR so I am not sure what is going on.

cottsay:
@brawner, it looks like #1462 took out PR builds on this repo. Please prioritize a rolling release of `performance_test_fixture` to get things going again.

guru-florida:
Thanks @cottsay ....and my apologies I should have addressed that request to @brawner and not yourself. oops. I'll rebase my changes to a commit before 1462 and force push.

cottsay:
>  I'll rebase my changes to a commit before 1462 and force push.

The problem isn't in this PR at all. The dependencies that are installed for PR builds are the latest released packages, and the referenced PR takes a dependency on unreleased features of `performance_test_fixture`. No PR builds will pass until either that dependency is dropped or the new features are available in the latest released version of `performance_test_fixture`.

There isn't anything we can do here until that problem is resolved.

clalancette:
I *think* this should be fixed now.  I'm going to request a rebuild and see what happens.

clalancette:

ivanpauno:
This approach seems ok-ish, but I wonder why interfaces keep a raw pointer of other interfaces (example [here](https://github.com/ros2/rclcpp/blob/504b082d8be8c14076db631807b98a4dc2b3234a/rclcpp/include/rclcpp/node_interfaces/node_topics.hpp#L91)) instead of a shared pointer.

e.g.:
```cpp
auto node = std::make_shared<rclcpp::Node>("my_node");
std::shared_ptr<rclcpp::NodeTopicsInterface> topics_interface = node->get_node_topics_interface();
node.reset();

// Use after free when using `topics_interface` here, because though `topics_interface` is still a valid shared pointer the base interface raw pointer stored in it is not more valid.
```

But maybe this PR is a good first step into fixing the larger ownership issue, @wjwwood what do you think?

wjwwood:
I'm not sure why they use raw pointers, probably because I assumed at the time that the interfaces were only ever owned by the node and so the node could enforce the order of creation and destruction.

I agree that this patch, while it may work, isn't really the right solution if we're already using shared pointers to store the interfaces. If we were using unique pointers (or raw pointers) to store them, then yes I think that would be ok. But I think a better patch would be to change interfaces so that they store shared pointers to each other, preventing issues with code like the example @ivanpauno posed. 

It may be that we have to use a patch like this for foxy to preserve ABI, but galactic we probably should upgrade to storing shared pointers between interfaces (assuming there are no issues with that, like circular references).

guru-florida:
(stating the obvious:) Just using sharedptrs everywhere isnt a silver bullet. Just like we're still left with resource/lifetime issues on GC languages. Take the [service.hpp example](https://github.com/ros2/rclcpp/blob/234b5e423b1e0f094c22be66214070312825017d/rclcpp/include/rclcpp/service.hpp#L180) that led me to the issue, it uses a weakptr which is great because it can detect if the underlying node interface is gone without explicitly holding onto it...however, if it's gone it can no longer properly release its own resources by calling rcl_service_fini function. IMO switching to sharedptr is good, weakptr probably better, but we probably still need to resolve tear-down of shared resources in a more mindful way. I wish I had an actual proposal but I think this is a tough nut. Hopefully we can come up with a better solution in Galactic and so I am all in favor of this PR being temporary.

wjwwood:
I agree the shared pointers aren't a silver bullet, but if we continue to return a shared pointer from `get_X_interface()` then we don't have much choice, since users can extend the lifetime of the interface object as @ivanpauno's example showed and the node class can no longer ensure the destruction happens in the right order.

So if we move away from shared pointer, which is an option, we can, but we need to also deprecate getting shared ownership of the interfaces from the node.

Then the question becomes, what does that cost us in terms of usability. There are probably cases where being able to get a shared pointer to the interfaces from the node is really useful. So maybe keeping them shared but fixing the ownership, e.g. so one interface keeps its dependent ones alive, is the best option.

We'll just have to see.

ivanpauno:
>  Just using sharedptrs everywhere isnt a silver bullet. Just like we're still left with resource/lifetime issues on GC languages. Take the service.hpp example that led me to the issue, it uses a weakptr which is great because it can detect if the underlying node interface is gone without explicitly holding onto it...however, if it's gone it can no longer properly release its own resources by calling rcl_service_fini function. IMO switching to sharedptr is good, weakptr probably better, but we probably still need to resolve tear-down of shared resources in a more mindful way

+1
Shared pointers are definitely not a silver bullet.

> Then the question becomes, what does that cost us in terms of usability. There are probably cases where being able to get a shared pointer to the interfaces from the node is really useful. So maybe keeping them shared but fixing the ownership, e.g. so one interface keeps its dependent ones alive, is the best option.

:+1: for the moment that seems like the best solution.

ivanpauno:
@guru-florida friendly ping

guru-florida:
Hi @ivanpauno. I agree with you and @wjwwood. I am hesitant to change anything in the node API like moving away from shared_ptrs, too much to break. We can't be responsible for how the external nodes hold the sub-interface ptrs, but we can correct the teardown internally and hopefully those external ptrs will keep the required sub-interfaces alive until teardown is possible. Over the next week, I can review the sub-interfaces teardown and present a summary of sub-iface inter-dependencies. I would like to get your guys' input on that before writing code...we should tread carefully here.

ivanpauno:
> Hi @ivanpauno. I agree with you and @wjwwood. I am hesitant to change anything in the node API like moving away from shared_ptrs, too much to break. We can't be responsible for how the external nodes hold the sub-interface ptrs, but we can correct the teardown internally and hopefully those external ptrs will keep the required sub-interfaces alive until teardown is possible. Over the next week, I can review the sub-interfaces teardown and present a summary of sub-iface inter-dependencies. I would like to get your guys' input on that before writing code...we should tread carefully here.

Sounds good, for the moment I think that merging this fix is fine.

ivanpauno (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13682/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8561/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11399/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13751/)

ivanpauno:
@guru-florida could you rebase?

guru-florida:
@ivanpauno I've rebased.   I added a small code change to the ServiceInterface destructor that was generating the memory leak log entries. It now prints the name of the service that was causing the leaks. I hope this will help other devs track down their leaks as it's what helped me do so. I feel it's related, but if you want me to separate out this commit I can.

Build Server is failing ATM. Looks like master is failing on Jenkins since Build 914. I don't understand why it's failing, the  test_graph_services_by_node passes when I run the tests locally.

fujitatomoya (approved):
I think it would be nice to have a fix which is dedicated to the issue, but i am okay to go.

fujitatomoya:
CI: (retry of https://github.com/ros2/rclcpp/pull/1469#issuecomment-780695360)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13775/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8640/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11483/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13852/)

wjwwood:
@ivanpauno @fujitatomoya did we consider this for backport?

ivanpauno:
@mergifyio backport foxy

wjwwood:
I'll take that as a "no, and we should be able to" 🤣

ivanpauno:
> I'll take that as a "no, and we should be able to" rofl

Yes, I forgot to backport.
I've opened a PR to apply the same change to foxy.

Galactic already has this patch applied and Dashing/Eloquent are EOL, so a backport to Foxy is the only thing missing.

Issue #1460, opened 2020-11-16
[rclcpp_lifecycle] The get_available_transitions method and get_available_transitions service produce different results


brawner:
I noticed this while writing up the benchmarks, but calling `rclcpp_lifecycle::LifecycleNode::get_available_transitions()` will return all 25 transitions, but calling the `get_available_transitions` service will only return the available transitions for the current state. In order to get all 25 transitions from a service, you need to call the `get_transition_graph` service.

Perhaps it would make sense to add a `get_transition_graph` method to the lifecycle node and update `get_available_transitions` to only return the available transitions of the current state so that it matches the similarly named services?

Thoughts maintainers/author? @wjwwood @ivanpauno @mabelzhang @Karsten1987

wjwwood:
Makes sense to me to make it consistent between the two.

fujitatomoya:
@brawner @wjwwood 

could you review https://github.com/ros2/rclcpp/pull/1472 when you got time?

fujitatomoya:
@brawner 

i think that we can close this.

brawner:
Resolved by #1472, thanks @fujitatomoya

PR #1472, opened 2020-11-22
add LifecycleNode::get_transition_graph along with service.

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/1460

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
@@ -702,9 +702,9 @@ public:
   std::vector<Transition>
   get_available_transitions();
 
-  /// Return a list with the available transitions.
+  /// Return a list with all the transitions.
   /**
-   * \return list with the available transitions.
+   * \return list with all the transitions in the transition graph.
    */
   RCLCPP_LIFECYCLE_PUBLIC
   std::vector<Transition>
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
@@ -694,13 +694,21 @@ public:
   std::vector<State>
   get_available_states();
 
+  /// Return a list with the current available transitions.
+  /**
+   * \return list with the current available transitions.
+   */
+  RCLCPP_LIFECYCLE_PUBLIC
+  std::vector<Transition>
+  get_available_transitions();
+
   /// Return a list with the available transitions.
   /**
    * \return list with the available transitions.
    */
   RCLCPP_LIFECYCLE_PUBLIC
   std::vector<Transition>
-  get_available_transitions();
+  get_transition_graph();
 
   /// Trigger the specified transition.
   /*
--- a/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
+++ b/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
@@ -352,6 +352,18 @@ public:
   get_available_transitions()
   {
     std::vector<Transition> transitions;
+    transitions.reserve(state_machine_.current_state->valid_transition_size);
+
+    for (unsigned int i = 0; i < state_machine_.current_state->valid_transition_size; ++i) {
+      transitions.emplace_back(&state_machine_.current_state->valid_transitions[i]);
+    }
+    return transitions;
+  }
+
+  std::vector<Transition>
+  get_transition_graph()
+  {
+    std::vector<Transition> transitions;
     transitions.reserve(state_machine_.transition_map.transitions_size);
 
     for (unsigned int i = 0; i < state_machine_.transition_map.transitions_size; ++i) {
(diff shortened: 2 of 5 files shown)


rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -694,13 +694,21 @@ class LifecycleNode : public node_interfaces::LifecycleNodeInterface,
...
 
+  /// Return a list with the current available transitions.
+  /**
+   * \return list with the current available transitions.
+   */
+  RCLCPP_LIFECYCLE_PUBLIC
+  std::vector<Transition>
+  get_available_transitions();
+
   /// Return a list with the available transitions.
```
brawner, line 705: I suggest updating this comment to clarify it returns all the transitions in the transition graph.

rclcpp_lifecycle/test/benchmark/benchmark_lifecycle_node.cpp
```diff
@@ -131,6 +131,19 @@ BENCHMARK_F(BenchmarkLifecycleNode, get_available_transitions)(benchmark::State
   }
 }
 
+BENCHMARK_F(BenchmarkLifecycleNode, get_transition_graph)(benchmark::State & state) {
+  for (auto _ : state) {
+    constexpr size_t expected_transitions = 25u;
+    const auto & transitions = node->get_transition_graph();
+    if (transitions.size() != expected_transitions) {
+      const std::string msg = std::to_string(transitions.size());
```
brawner, line 139: I know you're just mirroring the other benchmarks (which I think I wrote), but I think this error message could be a little more detailed. Maybe something like:
```suggestion
      const std::string msg = std::string("Expected number of transitions did not match actual: ") + std::to_string(transitions.size());
```

rclcpp_lifecycle/test/test_state_machine_info.cpp
```diff
@@ -54,8 +54,30 @@ TEST_F(TestStateMachineInfo, available_states) {
 
 TEST_F(TestStateMachineInfo, available_transitions) {
   auto test_node = std::make_shared<rclcpp_lifecycle::LifecycleNode>("testnode");
-  std::vector<rclcpp_lifecycle::Transition> available_transitions =
+  std::vector<rclcpp_lifecycle::Transition> current_available_transitions =
     test_node->get_available_transitions();
+  EXPECT_EQ(2u, current_available_transitions.size());
```
brawner, line 59: I think a better test of `get_available_transitions` would be to check the available transitions from a couple of different states and ensure they match the expected transitions by id

brawner (commented):
I think it looks pretty good. Thanks for taking care of this. I just had a few comments.

rclcpp_lifecycle/test/test_state_machine_info.cpp
fujitatomoya: okay, sgtm. i would add more concrete test cases for `get_available_transitions `.

fujitatomoya: more test scenarios added along with each primary state.

fujitatomoya:
@brawner 

requesting review one more time, thanks in advance!

brawner (approved):

brawner:
Building with benchmarks on and testing `--packages-select rclcpp_lifecycle`. With benchmarks on, there should be warnings on aarch64 and windows that osrf_testing_tools_cpp cannot be found.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13164/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8095/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10880/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13172/)

brawner:
Benchmark results seem reasonable to me:
```
BenchmarkLifecycleNode/get_available_states                       1915 ns         1915 ns       365023 heap_allocations=1
BenchmarkLifecycleNode/get_available_transitions                  1522 ns         1522 ns       456442 heap_allocations=1
BenchmarkLifecycleNode/get_transition_graph                       2561 ns         2561 ns       273973 heap_allocations=1
```

brawner:
The ci jobs all passed here. The warnings on aarch64 are the expected for benchmarks, and the ones on windows have been resolved. @wjwwood did you have any more comments or do you think this is good to merge?

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -350,6 +350,18 @@ class LifecycleNode::LifecycleNodeInterfaceImpl
 
   std::vector<Transition>
   get_available_transitions()
+  {
+    std::vector<Transition> transitions;
+    transitions.reserve(state_machine_.current_state->valid_transition_size);
+
+    for (unsigned int i = 0; i < state_machine_.current_state->valid_transition_size; ++i) {
```
wjwwood, line 357: I know this is because the data structure in C uses `unsigned int`, but really we should be using `size_t` here and there. Nothing to do for now, just a comment...

wjwwood (approved):
lgtm

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
brawner: Totally agreed

Issue #1392, opened 2020-10-08
Streamline creation of subscriptions in callback groups


schornakj:
#### Feature description
The motivation behind this is that I would like a convenient way to create a subscription associated with a callback group within a node's initializer list. It is straightforward to do this for actions and services, since `rclcpp::create_service`, `rclcpp_action::create_server`, etc., allow me to pass a callback group as an argument. However, the `rclcpp::create_subscription` function takes a `SubscriptionOptions` struct as an argument, and the `SubscriptionOptions` struct does not provide a constructor that takes a `rclcpp::CallbackGroup::SharedPtr`. This means that I have to pass an instance of a `CallbackGroup` to my `SubscriptionOptions` in the body of the node's constructor and then create the subscription using those `SubscriptionOptions`.

#### Implementation considerations
A simple approach would be to add a constructor overload to `SubscriptionOptions` that just takes `rclcpp::CallbackGroup::SharedPtr` as an argument. As an extension, `rclcpp::create_subscription` could be overloaded to take a callback group as an argument, similar to the equivalent functions that create clients and servers.

ivanpauno:
> A simple approach would be to add a constructor overload to SubscriptionOptions that just takes rclcpp::CallbackGroup::SharedPtr as an argument. As an extension, rclcpp::create_subscription could be overloaded to take a callback group as an argument, similar to the equivalent functions that create clients and servers.

I prefer the first option, but I think both are ok. PRs are welcomed.
Thanks for opening the issue @schornakj.

ivanpauno:
Another option is to use the parameter idiom (similar to `NodeOptions`) in `SubscriptionOptions`, that's maybe better than having lots of different constructors.

y-okumura-isp:
I think the following API is used in this use case. 
But it is usefull to use parameter idiom, so I sent PR above.
By the way, I dont know "parameter ideom" well, and I believe it is type of [method chaining](https://en.wikipedia.org/wiki/Method_chaining). Please tell me if I misunderstand.

```
  explicit SubscriptionOptionsWithAllocator(
    const SubscriptionOptionsBase & subscription_options_base)
```

Thank you.

ivanpauno:
> By the way, I dont know "parameter ideom" well, and I believe it is type of method chaining. Please tell me if I misunderstand.

Yeap, the "named parameter idiom" and "method chaining" is the same thing.
Thanks for opening a PR @y-okumura-isp !

PR #1481, opened 2020-11-26
Add SubscriptionOptionsBase::set_callback_group(#1392)

y-okumura-isp:
https://github.com/ros2/rclcpp/issues/1392

### DIFF
--- a/rclcpp/include/rclcpp/subscription_options.hpp
+++ b/rclcpp/include/rclcpp/subscription_options.hpp
@@ -77,6 +77,7 @@ struct SubscriptionOptionsBase
   QosOverridingOptions qos_overriding_options;
 
   /// Set use_default_callbacks.
+  RCLCPP_PUBLIC
   SubscriptionOptionsBase &
   set_use_default_callbacks(bool uses)
   {
@@ -85,6 +86,7 @@ struct SubscriptionOptionsBase
   }
 
   /// Set ignore_local_publications.
+  RCLCPP_PUBLIC
   SubscriptionOptionsBase &
   set_ignore_local_publications(bool ignores)
   {
@@ -93,6 +95,7 @@ struct SubscriptionOptionsBase
   }
 
   /// Set callback group.
+  RCLCPP_PUBLIC
   SubscriptionOptionsBase &
   set_callback_group(rclcpp::CallbackGroup::SharedPtr group)
   {
@@ -101,6 +104,7 @@ struct SubscriptionOptionsBase
   }
 
   /// Set intra_process_comm.
+  RCLCPP_PUBLIC
   SubscriptionOptionsBase &
   set_use_intra_process_comm(bool uses)
   {
--- a/rclcpp/include/rclcpp/subscription_options.hpp
+++ b/rclcpp/include/rclcpp/subscription_options.hpp
@@ -76,11 +76,67 @@ struct SubscriptionOptionsBase
 
   QosOverridingOptions qos_overriding_options;
 
+  /// Enable use_default_callbacks setting.
+  SubscriptionOptionsBase &
+  enable_use_default_callbacks()
+  {
+    use_default_callbacks = true;
+    return *this;
+  }
+
+  /// Enable use_default_callbacks setting.
+  SubscriptionOptionsBase &
+  disable_use_default_callbacks()
+  {
+    use_default_callbacks = false;
+    return *this;
+  }
+
+  /// Enable ignore_local_publications setting.
+  SubscriptionOptionsBase &
+  enable_ignore_local_publications()
+  {
+    ignore_local_publications = true;
+    return *this;
+  }
+
+  /// Disable ignore_local_publications setting.
+  SubscriptionOptionsBase &
+  disable_ignore_local_publications()
+  {
+    ignore_local_publications = false;
+    return *this;
+  }
+
   /// Set callback group, parameter idiom style.
   SubscriptionOptionsBase &
   set_callback_group(rclcpp::CallbackGroup::SharedPtr group)
   {
     callback_group = group;
+    return *this;
+  }
+
+  /// Enable intra_process_comm setting.
+  SubscriptionOptionsBase &
+  enable_use_intra_process_comm()
+  {
+    use_intra_process_comm = IntraProcessSetting::Enable;
+    return *this;
+  }
+
+  /// Disable intra_process_comm setting.
+  SubscriptionOptionsBase &
+  disable_use_intra_process_comm()
+  {
+    use_intra_process_comm = IntraProcessSetting::Disable;
+    return *this;
+  }
+
+  /// Clear intra_process_comm setting
+  SubscriptionOptionsBase &
+  clear_use_intra_process_comm()
+  {
+    use_intra_process_comm = IntraProcessSetting::NodeDefault;
     return *this;
   }
 };
--- a/rclcpp/include/rclcpp/subscription_options.hpp
+++ b/rclcpp/include/rclcpp/subscription_options.hpp
@@ -76,39 +76,23 @@ struct SubscriptionOptionsBase
 
   QosOverridingOptions qos_overriding_options;
 
-  /// Enable use_default_callbacks setting.
+  /// Set use_default_callbacks.
   SubscriptionOptionsBase &
-  enable_use_default_callbacks()
+  set_use_default_callbacks(bool uses)
   {
-    use_default_callbacks = true;
+    use_default_callbacks = uses;
     return *this;
   }
 
-  /// Enable use_default_callbacks setting.
+  /// Set ignore_local_publications.
   SubscriptionOptionsBase &
-  disable_use_default_callbacks()
+  set_ignore_local_publications(bool ignores)
   {
-    use_default_callbacks = false;
+    ignore_local_publications = ignores;
     return *this;
   }
 
-  /// Enable ignore_local_publications setting.
-  SubscriptionOptionsBase &
-  enable_ignore_local_publications()
-  {
-    ignore_local_publications = true;
-    return *this;
-  }
-
-  /// Disable ignore_local_publications setting.
-  SubscriptionOptionsBase &
-  disable_ignore_local_publications()
-  {
-    ignore_local_publications = false;
-    return *this;
-  }
-
-  /// Set callback group, parameter idiom style.
+  /// Set callback group.
   SubscriptionOptionsBase &
   set_callback_group(rclcpp::CallbackGroup::SharedPtr group)
   {
@@ -116,27 +100,11 @@ struct SubscriptionOptionsBase
     return *this;
   }
 
-  /// Enable intra_process_comm setting.
+  /// Set intra_process_comm.
   SubscriptionOptionsBase &
-  enable_use_intra_process_comm()
+  set_use_intra_process_comm(bool uses)
   {
-    use_intra_process_comm = IntraProcessSetting::Enable;
-    return *this;
-  }
-
-  /// Disable intra_process_comm setting.
-  SubscriptionOptionsBase &
-  disable_use_intra_process_comm()
-  {
-    use_intra_process_comm = IntraProcessSetting::Disable;
-    return *this;
-  }
-
-  /// Clear intra_process_comm setting
-  SubscriptionOptionsBase &
-  clear_use_intra_process_comm()
-  {
-    use_intra_process_comm = IntraProcessSetting::NodeDefault;
+    use_intra_process_comm = uses ? IntraProcessSetting::Enable : IntraProcessSetting::Disable;
     return *this;
   }
 };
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
Could you elaborate why we need this method as setter? I think that It probably would be nice to have more description why this is needed and what the problem is.

y-okumura-isp:
Thank you for reviewing.
This is related with #1392.

The motivation of this issues is "more convenient way to create a subscription associated with a callback group".

> Another option is to use the parameter idiom (similar to NodeOptions) in SubscriptionOptions, that's maybe better than having lots of different constructors.

I partially agree with this opinion.
I feel parameter idiom is not so convenient, but it is reasonable to manipulate XxxOptions in a similar way.
Additionally, in my point of view, I like setter in user code for validation or encapsulation
(but we probably don't need validation because `struct` or `enum` are used properly) .

I think it is better to add other setters than `callback_group` if this PR woud be  accepted.
But I added only this setter to start discussion.

Thank you.

fujitatomoya:
Ah, i didn't see https://github.com/ros2/rclcpp/issues/1392, that was oversight on my part. thanks!

rclcpp/include/rclcpp/subscription_options.hpp
```diff
@@ -75,6 +75,14 @@ struct SubscriptionOptionsBase
   TopicStatisticsOptions topic_stats_options;
 
   QosOverridingOptions qos_overriding_options;
+
+  /// Set callback group, parameter idiom style.
+  SubscriptionOptionsBase &
+  set_callback_group(rclcpp::CallbackGroup::SharedPtr group)
```
ivanpauno, line 81: If we only add one method, this is not being of much help.
It would be gread to add methods enabling chaining of all options.

rclcpp/test/rclcpp/test_subscription_options.cpp
```diff
@@ -54,6 +54,16 @@ class TestSubscriptionOptions : public ::testing::Test
...
 };
 
+TEST_F(TestSubscriptionOptions, subscription_options_constructor) {
+  rclcpp::SubscriptionOptions option1;
+  EXPECT_EQ(option1.callback_group, nullptr);
+
+  auto group = std::make_shared<rclcpp::CallbackGroup>(rclcpp::CallbackGroupType::Reentrant);
+  rclcpp::SubscriptionOptions option2;
+  option2.set_callback_group(group);
+  EXPECT_EQ(option2.callback_group, group);
```
ivanpauno, line 64:
```suggestion
  auto group = std::make_shared<rclcpp::CallbackGroup>(rclcpp::CallbackGroupType::Reentrant);
  auto option = rclcpp::SubscriptionOptions().set_callback_group(group);
  EXPECT_EQ(option.callback_group, group);
```

I deleted `option1`, as it didn't see to be adding much to the test.
I've also made a one line example of how the method is useful, though as commented above it would be better to fully enable method chaining in this PR.

ivanpauno (changes requested):

rclcpp/test/rclcpp/test_subscription_options.cpp
y-okumura-isp: Thank you for reviewing.
I wanted comments before writing a lot.

> It would be gread to add methods enabling chaining of all options.

I'll be glad to do it!

y-okumura-isp:
@ivanpauno I updated my PR, so would you review again?
I mainly added boolean value setters.
I thought people use lambda function for set callback functions or write some code block rather than setter to set `struct` or `class` variables.

Additionally, functions are implemented in hpp file.
Is is better to create subscription_options.cpp? (now this file does not exit).

Thank you.

rclcpp/include/rclcpp/subscription_options.hpp
```diff
@@ -75,6 +75,70 @@ struct SubscriptionOptionsBase
   TopicStatisticsOptions topic_stats_options;
 
   QosOverridingOptions qos_overriding_options;
+
+  /// Enable use_default_callbacks setting.
+  SubscriptionOptionsBase &
+  enable_use_default_callbacks()
```
ivanpauno, line 81: maybe `set_use_default_callbacks(bool)` is a better idea.
@wjwwood @mabelzhang any preference?

Same comments apply elsewhere.

rclcpp/include/rclcpp/subscription_options.hpp
```diff
@@ -75,6 +75,70 @@ struct SubscriptionOptionsBase
...
+  SubscriptionOptionsBase &
+  disable_ignore_local_publications()
+  {
+    ignore_local_publications = false;
+    return *this;
+  }
+
+  /// Set callback group, parameter idiom style.
+  SubscriptionOptionsBase &
+  set_callback_group(rclcpp::CallbackGroup::SharedPtr group)
```
ivanpauno, line 113: Note to other reviewers: we need a `set_` prefix, because we already have public members with the same name.

Should we deprecate the public members so this struct can be converted to a class in the future?
Another alternative is to apply method chaining in a builder class:

```cpp
SubscriptionOption::Builder().callback_group(group).enable_intra_process_comm().build();
```

I don't have a strong preference

rclcpp/include/rclcpp/subscription_options.hpp
ivanpauno: Anyways, I would prefer `enable_default_callbacks` than `enable_use_default_callbacks`.

ivanpauno (commented):
Let's wait to other reviewers opinion, but overall this seems like a good addition to me

ivanpauno:
> Is is better to create subscription_options.cpp? (now this file does not exit).

I don't have an strong preference for this kind of one liners, keeping them in the header file sounds ok to me.
(note: you need to add RCLCPP_PUBLIC macros to avoid visibility issues on Windows)

rclcpp/include/rclcpp/subscription_options.hpp
wjwwood: I think that given the existing member access, the builder option is best.

---

If we would prefer to change it, we could have a two tick-tock cycle to phase it out:

- ROS G:
  - new class `rclcpp::future::SubscriptionOptionsBase` with named parameter idiom properties
  - deprecate use of `rclcpp::SubscriptionOptionBase`
- ROS H:
  - move `rclcpp::future::SubscriptionOptionBase` to `rclcpp::SubscriptionOptionBase`, removing old version
  - deprecate `rclcpp::future::SubscriptionOptionBase`
- ROS I:
  - remove `rclcpp::future::SubscriptionOptionBase`

The benefit would be that we have consistency with other option-like structures in rclcpp, but the downside is the churn for developers (though we could hide it for many developers which won't touch subscription options).

rclcpp/include/rclcpp/subscription_options.hpp
wjwwood: Since this is a binary option using true/false, I think we should do `set_use_default_callbacks(bool)` as you suggested.

Same for `ignore_local_publications`.

rclcpp/include/rclcpp/subscription_options.hpp
```diff
@@ -75,6 +75,70 @@ struct SubscriptionOptionsBase
...
+  SubscriptionOptionsBase &
+  disable_use_intra_process_comm()
+  {
+    use_intra_process_comm = IntraProcessSetting::Disable;
+    return *this;
+  }
+
+  /// Clear intra_process_comm setting
+  SubscriptionOptionsBase &
+  clear_use_intra_process_comm()
```
wjwwood, line 137: I'm not sure this is needed often enough to need its own method. This is the default state of the subscriptions options, so if you're using this function you've changed the default and you're changing it back... I find it unlikely that this scenario overlaps with one-liners, but maybe I'm missing some obvious case.

Normally, I wouldn't care if this is unlikely to be used, but if we eliminated it then we could make this `set_use_intra_process_comm(bool)` instead.

The middle ground is `set_use_intra_process_comm(IntraProcessSetting)`, but that looses some of the brevity. Though honestly, I prefer that to the current set of three methods with awkward sounding names. But it is subjective.

rclcpp/include/rclcpp/subscription_options.hpp
y-okumura-isp: Thank you for comments.

I choose enable/disable because `use_intra_process_comm` is not bool.
But as in comments below, it looks OK to treat it as if this is bool because one-liners are the target.

In 5ffeb9b, I think we can unify `set_some(bool)`.

rclcpp/include/rclcpp/subscription_options.hpp
y-okumura-isp: (Almost same comment as above)
Thank you for comments.

It looks OK to treat it as if this is bool because one-liners are the target.

In 5ffeb9b, I think we can unify `set_some(bool)`.

y-okumura-isp:
> (note: you need to add RCLCPP_PUBLIC macros to avoid visibility issues on Windows)

Sorry, I'm not familiar with Windows.
I saw some header files, but not every public function had RCLCPP_PUBLIC annotation.
Is it OK to add this macro to all setter function we have introduced?

y-okumura-isp:
I studied RCLCPP_PUBLIC as I hadn't used it much before..  In this case, we need to let newly introduced APIs public (right?).

@ivanpauno @wjwwood 
I updated my PR. Would you please review again?

Thank you.

rclcpp/include/rclcpp/subscription_options.hpp
```diff
@@ -75,6 +75,42 @@ struct SubscriptionOptionsBase
...
   QosOverridingOptions qos_overriding_options;
+
+  /// Set use_default_callbacks.
+  RCLCPP_PUBLIC
+  SubscriptionOptionsBase &
+  set_use_default_callbacks(bool uses)
+  {
+    use_default_callbacks = uses;
+    return *this;
+  }
```
wjwwood, line 86: Sorry I missed this before, but we really try to avoid implementation in headers, in order to avoid ODR violations and also to decrease compile times.

Since this class is not templated, I'd say it should have a `.cpp` file (it didn't have one before because it didn't have any methods) and these should be implemented there.

rclcpp/include/rclcpp/subscription_options.hpp
wjwwood: @ivanpauno what should we do here? builder I guess? Should we aim for the deprecation cycle?

@y-okumura-isp that will require moving these new functions out of this class and into a builder class instead.

wjwwood (changes requested):

rclcpp/include/rclcpp/subscription_options.hpp
y-okumura-isp: Sorry, I should have discuss API before update PR.

`SubscriptionOptionsBase` has many parameters which are `class` or `struct`, for example, `SubscriptionEventCallbacks event_callbacks` or `TopicStatisticsOptions topic_stats_options` and so on.
So I think Builder could set boolean options but we would set other parameters manually like below, right?

```
auto sub_option = SubscriptionOption::Builder().callback_group(group).use_intra_process_comm(true).build();

// we can set complex options directly
sub_option.event_callbacks.deadline_callback = [](auto) {};

// Could we overwrite parameters which Builder set? or hide them?
sub_option.use_default_callbacks = false;

// we no longer need setter functions
// sub_option.set_callback_group(group);   // no longer need
```

If above code is right, I think even if we would introduce Builder, existing API of SubscriptionOptionsBase would not change.
Would we need two tick-tock cycle in such a case?

If I misunderstand, please tell me.

Thank you.

ivanpauno: > So I think Builder could set boolean options but we would set other parameters manually like below, right?

I would make everything available in the builder methods, e.g.:

```cpp
auto sub_option = SubscriptionOption::Builder().callback_group(group).use_intra_process_comm(true).deadline_callback([](auto) {}).build();
```

> existing API of SubscriptionOptionsBase would not change.

That's fine, in the future we could deprecate direct access, but I think that having both options is ok.

> Would we need two tick-tock cycle in such a case?

That's needed if direct access to the fields is deprecated, which I would avoid at the moment.

ivanpauno:
@y-okumura-isp friendly ping, were you able to make progress in this PR?

rclcpp/include/rclcpp/subscription_options.hpp
y-okumura-isp: @ivanpauno Thank you for reply! I I'm sorry I missed it.

> That's fine, in the future we could deprecate direct access, but I think that having both options is ok.
> > Would we need two tick-tock cycle in such a case?
> That's needed if direct access to the fields is deprecated, which I would avoid at the moment.

I agree. I would upadte PR.

y-okumura-isp:
@ivanpauno Thank you for ping.

> @y-okumura-isp friendly ping, were you able to make progress in this PR?

I missed your comment. I'm going to update PR.

ivanpauno:
@y-okumura-isp friendly ping, are you planning to move forward with this?

clalancette:
Given the age of this, I'm going to close it out.  Feel free to reopen if you are going to work on it and respond to the reviews.

Issue #1475, opened 2020-11-25
test_publisher run_event_handler test fails with CycloneDDS


sloretz:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu Focal
- Installation type:
  - From source
- Version or commit hash:
  - master
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
RMW_IMPLEMENTATION=rmw_cyclonedds_cpp build/rclcpp/test/rclcpp/test_publisher
```

#### Expected behavior

All tests should pass

#### Actual behavior

```
[ RUN      ] TestPublisher.run_event_handlers
1606261998.534775 [0] test_publi: using network interface enp3s0 (udp/192.168.1.150) selected arbitrarily from: enp3s0, docker0
[WARN] [1606261998.541156630] [ns.my_node]: New subscription discovered on topic '/ns/topic', requesting incompatible QoS. No messages will be sent to it. Last incompatible policy: INVALID_QOS_POLICY
/home/sloretz/bigssd/ros2/src/ros2/rclcpp/rclcpp/test/rclcpp/test_publisher.cpp:509: Failure
Expected: handler->execute(data) throws an exception of type std::runtime_error.
  Actual: it throws nothing.
[  FAILED  ] TestPublisher.run_event_handlers (11 ms)
```

#### Additional information

The test started failing since  ros2/rclcpp#1241, and reverting to the commit before resolves the test failure, but it's not clear to me if that PR introduced a bug or exposed an existing bug. Before the PR the test seemed to run the event handlers `execute()` method without any expectations. That PR changed the test to expect no events on a newly created publisher. When run with `rmw_cyclonedds_cpp` there is an event.

[...]

PR #1497, opened 2020-12-21
Fix wrong test expectations

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1475.

The test was passing with fastrtps, because it does nothing in that case :smile: (fastrtps doesn't implement the incompatible qos event).

### DIFF
--- a/rclcpp/test/rclcpp/test_publisher.cpp
+++ b/rclcpp/test/rclcpp/test_publisher.cpp
@@ -506,6 +506,6 @@ TEST_F(TestPublisher, run_event_handlers) {
 
   for (const auto & handler : publisher->get_event_handlers()) {
     std::shared_ptr<void> data = handler->take_data();
-    EXPECT_THROW(handler->execute(data), std::runtime_error);
+    handler->execute(data);
   }
 }


ivanpauno:
CI, only cyclonedds, only linux:

* Before [Build Status](http://ci.ros2.org/job/ci_linux/13309/)
* After [Build Status](http://ci.ros2.org/job/ci_linux/13308/)

rclcpp/test/rclcpp/test_publisher.cpp
```diff
@@ -506,6 +506,6 @@ TEST_F(TestPublisher, run_event_handlers) {
 
   for (const auto & handler : publisher->get_event_handlers()) {
     std::shared_ptr<void> data = handler->take_data();
-    EXPECT_THROW(handler->execute(data), std::runtime_error);
```
sloretz, line 509: If the test was passing before for FastRTPS, does that mean it was throwing an exception here, or that there were no event handlers and this loop was never entered?

ivanpauno: > there were no event handlers and this loop was never entered?

Exactly that.
The only "default" event handler is the "incompatible qos" event, which is not implemented in fastrtps.

ivanpauno:
Test is passing with cyclonedds, another run with only fastrtps and up to rclcpp:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13310/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8241/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11041/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13351/)

sloretz (approved):
LGTM with Green CI

PR #1501, opened 2021-01-02
Make the default callback group of a node configurable via `NodeOptions`

jdlangs:
This adds an option to the `NodeOptions` data structure to specify the type of the default callback group when constructing a node. Currently, it is hardcoded to be mutually exclusive. The intention is to make things easier for developers who would prefer a reentrant group to do things like call a service from within a callback and wait for the response (along with a `MultiThreadedExecutor`). This enables them to not have to worry about creating or keeping track of other callback groups.

This was prompted from the discussion in https://github.com/ros2/ros2_documentation/issues/901. I'll be happy to incorporate any feedback or thoughts.

Thanks,
Josh


wjwwood:
The issue I see with changing the default is that there are things which use the default callback group that assume it is mutually exclusive but are not created by the user. For example, services used with the ROS parameters interface (which use services) assumes that the group they are added to is mutually exclusive.

My opinion is that it is better to be explicit and create a reentrant callback group if that's what you want to use.

I could see having something integrated into the "subnode" concept, where you can specify a different default callback group for the subnode without replacing the original node's default callback group. This would allow existing "built-in" in things to continue to use the existing mutually exclusive default callback group, but users could do something like this:

```c++
auto node = ...;
// with a better name... maybe
auto my_subnode = node->create_sub_node_with_different_default_callback_group(callback_group_type);

auto thing = my_subnode->create_xyz(...);
function_that_takes_node_to_create_things(my_subnode, ...);
```

Or as a class:

```c++
class MyNode : public Node
{
public:
  MyNode() : Node(...)
  : subnode_(this->create_sub_node_with_different_default_callback_group(callback_group_type)
  {
    this->service_ = this->subnode_->create_service(...);
  }
};
```

That would let you change the default behavior for "user created things", without interfering with built-in functions of the node that rely on the default callback group.

That being said, the subnode concept needs some work done to it so it doesn't have to be reimplemented in each new kind of node class, e.g. LifecycleNode.

----

So I'm interested in the subnode option, but I do not care for this approach of a NodeOption, preferring the alternative of just being explicit and creating a custom callback group if you want something different.

We could work on making the workaround more streamlined too, if it is too verbose.

I would propose to close this pull request in favor of other options.

CC @clalancette @ivanpauno @mabelzhang

jdlangs:
Thanks for the detailed analysis of the proposal and the brainstorming of alternatives!

My initial reaction is that the problem is not changing the type of the default but that components are making (brittle) assumptions about what the default is going to be. If, for example, someone decided to fork rclcpp and just change the hardcoded default, how would they know it is going to create problems? It would seem better that your example of the parameter internals either become agnostic to the default type or if they need a mutually exclusive group, it should be created and owned explicitly instead of relying on the default.

I can think of a few other options for reducing the verbosity and the subnode proposal is interesting too. Giving the option to the developer to easily change the default feels like the cleanest option to me though so if you're willing to indulge a bit more conversation to see if it's possible I would definitely appreciate it.

wjwwood:
> My initial reaction is that the problem is not changing the type of the default but that components are making (brittle) assumptions about what the default is going to be. If, for example, someone decided to fork rclcpp and just change the hardcoded default, how would they know it is going to create problems? It would seem better that your example of the parameter internals either become agnostic to the default type or if they need a mutually exclusive group, it should be created and owned explicitly instead of relying on the default.

Well it's only being made brittle by the fact that you're changing something that is known to be static.

We could have them create their own callback group and therefore ensure what they get, but to be honest that's already what's happening, it just so happens they share that callback group with each other (the built-ins do I mean) as well as with the user.

However, I think that part of the reason for using the single mutually exclusive callback group that is shared between the node's built-in functions and the user, is so that by default nothing is run concurrently within a node where user code is likely to share resources.

For example, when a parameter is being set (locally or remotely), the user can add a hook to validate those changes and either accept or reject them. Consider the possibility that the validation hook (written by the user) accesses a shared resource in the node which is also modified in a timer or subscription callback that the user also wrote. If they don't share a mutually exclusive callback group then the user needs to synchronize access to this shared resource. This could easily be overlooked and cause crashes or if synchronized by the user with a mutex or something could cause unnecessary resource contention or even dead-lock (one of the reasons callback groups were created in the first place was to avoid these two issues).

This is a contrived example for sure, but I think the safest default is to have everything in a single mutually exclusive callback group, and let users deviate from that explicitly. I'd only recommend that to expert users, as they then have to be very careful about multi-threading issues.

----

A few other ideas for convenience that would be an alternative to this proposal:

You could have a "unique_callback_group_option" when creating things, so you can somewhat concisely express the desire for the new thing you're creating to have its own callback group. The new thing could own the callback group and therefore alleviate you (the user) from having to manage the callback group's lifetime. I think this could be done without creating a circular reference:

```c++
auto node = ...;
auto sub = node->create_subscription(
  "topic", 10, [](const auto & msg) { /* ... */ }, rclcpp::SubscriptionOptions().use_unique_callback_group());
```

Or something like that, assuming no other subscription options need to be set. The "unique" callback group could be mutually exclusive by default, allowing it to be called concurrently with other callbacks, but not with itself. And we could have a similar option for having its own reentrant callback group which would additionally allow the callback to be called concurrently with itself.

wjwwood:
> However, I think that part of the reason for using the single mutually exclusive callback group that is shared between the node's built-in functions and the user, is so that by default nothing is run concurrently within a node where user code is likely to share resources.

I should also say, that this has been a positively remarked on thing for ROS 1 and it is ROS 1's default behavior that we're trying to emulate here with the default callback group.

We are (or have been) in some sense explicitly choosing to favor safety for more novice users over convenience for more expert users, which is a theme in ROS to be honest, and not necessarily a bad one. That doesn't mean we don't want to make it as nice to user as possible for expert use cases, but I think this is where my internal push back on this change (especially to make reentrant the default) comes from.

jdlangs:
I understand the point of view of keeping things safe by default and that being able to change the default easily might be too tempting to solve these issues without fully understanding the implications.

I'll continue to think over alternate API changes and will maybe open another PR in the future.

PR #1502, opened 2021-01-02
Use std compliant non-method std::filesystem::exists function

jdlangs:
I was looking into updating `rcpputils::fs` to be more compliant with `std::filesystem` for an eventual C++17 migration and found the current functionality includes a `path::exists()` method which [doesn't exist](https://en.cppreference.com/w/cpp/filesystem/path) in `std::filesystem`. This simply changes usage to the non-member function and can be merged regardless of any future changes to `rcpputils`.

### DIFF
--- a/rclcpp_components/test/test_component_manager.cpp
+++ b/rclcpp_components/test/test_component_manager.cpp
@@ -51,9 +51,10 @@ TEST_F(TestComponentManager, get_component_resources_valid)
   EXPECT_EQ("test_rclcpp_components::TestComponentBar", resources[1].first);
   EXPECT_EQ("test_rclcpp_components::TestComponentNoNode", resources[2].first);
 
-  EXPECT_TRUE(rcpputils::fs::path(resources[0].second).exists());
-  EXPECT_TRUE(rcpputils::fs::path(resources[1].second).exists());
-  EXPECT_TRUE(rcpputils::fs::path(resources[2].second).exists());
+  namespace fs = rcpputils::fs;
+  EXPECT_TRUE(fs::exists(fs::path(resources[0].second)));
+  EXPECT_TRUE(fs::exists(fs::path(resources[1].second)));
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
lgtm

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13348/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8276/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11078/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13400/)

clalancette:
All right, CI looks good, so merging.  Thanks for the contribution.

jdlangs:
Thanks!

I'm not familiar with the procedure for getting stuff like this backported; what are the usual steps? Am I expected to take the lead on it?

clalancette:
> I'm not familiar with the procedure for getting stuff like this backported; what are the usual steps? Am I expected to take the lead on it?

That's most expedient, yes.  Though I'm not entirely sure this makes sense to backport.  First, it is only in a test.  And second, we won't be switching Foxy to C++17, so the fact that it isn't 100% compliant with `std::filesystem` doesn't matter as much.  What's your thinking here?

jdlangs:
It's really just so I don't have to keep around a forked version if combined with a modified `rcpputils` (since I'll have foxy versions as my baseline). But yeah, it's not a huge improvement since the `std::filesystem` stuff can't be backported anyways.

Is the process just opening a PR against foxy with this commit cherry-picked on to a new branch?

clalancette:
> Is the process just opening a PR against foxy with this commit cherry-picked on to a new branch?

Yeah, that should do it for this PR.

jdlangs:
Ok thanks; if it looks like it will save some effort I'll go ahead with that, otherwise I won't bother you guys.

PR #1503, opened 2021-01-05
Make sure to wait for graph change events in test_node_graph.

clalancette:
While debugging something else, I came across this fairly rare
flake in the test_node_graph tests.  Essentially, it may take
some time for node changes to propagate into the graph.  If
the NodeGraph client asks for data before the changes are
in the graph, they may get something they don't expect.  The
fix is to wait for an event on the NodeGraph to happen, and
then look for the data you are interested in.

Before this change, I could get the flake to happen on a loaded
aarch64 system fairly regularly.  After this change, I can no
longer make the flake happen on a loaded system.

@eboasson FYI, this should fix the aarch64 failures we are seeing in the tests in https://github.com/ros2/rmw/pull/293 .

### DIFF
--- a/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
+++ b/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
@@ -14,6 +14,7 @@
 
 #include <gtest/gtest.h>
 
+#include <functional>
 #include <memory>
 #include <string>
 #include <utility>
@@ -67,45 +68,45 @@ protected:
 
   const rclcpp::node_interfaces::NodeGraph * node_graph() const {return node_graph_;}
 
-  size_t get_num_topics()
+  size_t get_num_graph_things(std::function<size_t()> predicate)
   {
     constexpr std::chrono::milliseconds timeout(100);
 
-    int tries = 0;
-    size_t num_topics = 0;
-    while (num_topics < 1u && tries++ < 5) {
+    size_t tries = 0;
+    size_t num_things = 0;
+    while (tries++ < 5) {
+      num_things = predicate();
+      if (num_things >= 1) {
+        break;
+      }
+
       auto event = node()->get_graph_event();
       EXPECT_NO_THROW(node()->wait_for_graph_change(event, timeout));
-
-      auto topic_names_and_types = node_graph()->get_topic_names_and_types();
-      num_topics = topic_names_and_types.size();
     }
 
     // Run notify_graph_change() here to remove the graph_event users we created
     // in the loop above.
     node_graph_->notify_graph_change();
 
-    return num_topics;
+    return num_things;
+  }
+
+  size_t get_num_topics()
+  {
+    return get_num_graph_things(
+      [this]() -> size_t {
+        auto topic_names_and_types = node_graph()->get_topic_names_and_types();
+        return topic_names_and_types.size();
+      });
   }
 
   size_t get_num_services()
   {
-    constexpr std::chrono::milliseconds timeout(100);
-
-    int tries = 0;
-    size_t num_services = 0;
-    while (num_services < 1u && tries++ < 5) {
-      auto event = node()->get_graph_event();
-      EXPECT_NO_THROW(node()->wait_for_graph_change(event, timeout));
-
-      auto service_names_and_types = node_graph()->get_service_names_and_types();
-      num_services = service_names_and_types.size();
-    }
-
-    // Run notify_graph_change() here to remove the graph_event users we created
-    // in the loop above.
-    node_graph_->notify_graph_change();
-    return num_services;
+    return get_num_graph_things(
+      [this]() -> size_t {
+        auto service_names_and_types = node_graph()->get_service_names_and_types();
+        return service_names_and_types.size();
+      });
   }
 
 private:
@@ -347,17 +348,12 @@ TEST_F(TestNodeGraph, get_info_by_topic)
 
   EXPECT_EQ(0u, node_graph()->get_publishers_info_by_topic("topic", true).size());
 
-  constexpr std::chrono::milliseconds timeout(100);
-  int tries = 0;
-  size_t num_publishers = 0;
   std::vector<rclcpp::TopicEndpointInfo> publishers;
-  while (num_publishers < 1u && tries++ < 5) {
-    auto event = node()->get_graph_event();
-    EXPECT_NO_THROW(node()->wait_for_graph_change(event, timeout));
-
-    publishers = node_graph()->get_publishers_info_by_topic("topic", false);
-    num_publishers = publishers.size();
-  }
+  size_t num_publishers = get_num_graph_things(
+    [this, &publishers]() {
+      publishers = node_graph()->get_publishers_info_by_topic("topic", false);
+      return publishers.size();
+    });
   ASSERT_EQ(1u, num_publishers);
 
   auto publisher_endpoint_info = publishers[0];
--- a/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
+++ b/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
@@ -17,6 +17,7 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "rcl/graph.h"
 #include "rcl/node_options.h"
--- a/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
+++ b/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
@@ -66,6 +66,47 @@ protected:
 
   const rclcpp::node_interfaces::NodeGraph * node_graph() const {return node_graph_;}
 
+  size_t get_num_topics()
+  {
+    constexpr std::chrono::milliseconds timeout(100);
+
+    int tries = 0;
+    size_t num_topics = 0;
+    while (num_topics < 1u && tries++ < 5) {
+      auto event = node()->get_graph_event();
+      EXPECT_NO_THROW(node()->wait_for_graph_change(event, timeout));
+
+      auto topic_names_and_types = node_graph()->get_topic_names_and_types();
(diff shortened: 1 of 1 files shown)


tfoote (approved):

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -66,18 +66,57 @@ class TestNodeGraph : public ::testing::Test
...
+      auto event = node()->get_graph_event();
+      EXPECT_NO_THROW(node()->wait_for_graph_change(event, timeout));
+
+      auto topic_names_and_types = node_graph()->get_topic_names_and_types();
+      num_topics = topic_names_and_types.size();
+    }
+
+    // Run notify_graph_change() here to remove the graph_event users we created
+    // in the loop above.
+    node_graph_->notify_graph_change();
```
hidmic, line 85: @clalancette I'm not sure I follow this bit -- or at least I find it counter-intuitive.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -66,18 +66,57 @@ class TestNodeGraph : public ::testing::Test
...
+    constexpr std::chrono::milliseconds timeout(100);
+
+    int tries = 0;
+    size_t num_topics = 0;
+    while (num_topics < 1u && tries++ < 5) {
+      auto event = node()->get_graph_event();
+      EXPECT_NO_THROW(node()->wait_for_graph_change(event, timeout));
+
+      auto topic_names_and_types = node_graph()->get_topic_names_and_types();
+      num_topics = topic_names_and_types.size();
```
hidmic, line 80: @clalancette meta: having some form of `wait_for_graph()` function taking the node graph, a predicate on that node graph, and a timeout would be neat.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
clalancette: >  I'm not sure I follow this bit -- or at least I find it counter-intuitive.

I agree that it is counter-intuitive, though it is the way current `NodeGraph` API is setup.

A call to [node->get_graph_event](https://github.com/ros2/rclcpp/blob/b9ef1fedfa409c966658a747d653f84cce7357c3/rclcpp/src/rclcpp/node.cpp#L386) ends up calling [node_graph->get_graph_event](https://github.com/ros2/rclcpp/blob/b9ef1fedfa409c966658a747d653f84cce7357c3/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp#L368).  There, a new shared point to `rclcpp::Event` is created, a weak reference to that `Event` is stored, the `graph_users_count_` is incremented, and the shared pointer is returned.  However, there is no explicit decrement of `graph_users_count_`.  Instead, when `notify_graph_change` is called, it loops through the weak pointers, attempting to lock them.  If it finds one that it cannot lock anymore (because it has been destroyed), then it removes that weak pointer from the list and recalculates the size of `graph_users_count`.  That way you'll get the "correct" number of events when you later call `count_graph_users`.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
clalancette: Yeah, true.  I did consider it, but I wasn't sure there were enough uses in here to warrant it.  But if you think it is clearer, I can do that.  Thoughts?

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -66,18 +67,57 @@ class TestNodeGraph : public ::testing::Test
 
   const rclcpp::node_interfaces::NodeGraph * node_graph() const {return node_graph_;}
 
+  size_t get_num_topics()
+  {
+    constexpr std::chrono::milliseconds timeout(100);
+
+    int tries = 0;
```
wjwwood, line 74: nitpick: `size_t`?

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
hidmic: I see. It's out of scope for this PR, but it'd be nice to have a way to explicitly drop the event instead of doing this.

hidmic (approved):

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -66,18 +67,57 @@ class TestNodeGraph : public ::testing::Test
...
+    size_t num_topics = 0;
+    while (num_topics < 1u && tries++ < 5) {
+      auto event = node()->get_graph_event();
+      EXPECT_NO_THROW(node()->wait_for_graph_change(event, timeout));
+
+      auto topic_names_and_types = node_graph()->get_topic_names_and_types();
+      num_topics = topic_names_and_types.size();
```
wjwwood, line 81: Doesn't this mean that you will always wait 100ms in the event that all the graph events occurred before you get to this point?

Maybe it should be more of a while true, with a break when the `num_topics` meets the threshold? And then make that check before waiting?

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
wjwwood: This also seems like an abuse of this function to me. If you need to prune old events then maybe we should have a function for that, but why does that matter?

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
clalancette: Yeah, I guess that is more correct.  I'll switch it.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
clalancette: > Doesn't this mean that you will always wait 100ms in the event that all the graph events occurred before you get to this point?

Yep, exactly.  In practice I didn't see this happening, and 100ms was short enough that I didn't think it would be too much of an issue if it did (the test would just be a bit slower).  But I can change it to a different loop to avoid the sleep completely.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
wjwwood: To be specific, I confused why clearing the old events is important.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
wjwwood: Well, you don't _have_ to change it, I was just wondering if it was intentional. Especially because it must have worked for fastrtps before this, so in that case we've didn't need the sleep or the retries I guess?

clalancette: > Especially because it must have worked for fastrtps before this, so in that case we didn't need the sleep or the retries I guess?

Yeah, I'm currently thinking that there is some detail of CycloneDDS that makes the graph change just a bit slower.  In practice, this was an extreme flake even with CycloneDDS; I couldn't get it happen with 2 hours of testing.  I could only reproduce it with a lot of load on the machine.

I'm going to go ahead and make the change anyway.  It is a little nicer, and while I'm in there I'll also make the change that @hidmic suggested of using a predicate.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
clalancette: > To be specific, I confused why clearing the old events is important.

Mostly because of this check: https://github.com/ros2/rclcpp/blob/b9ef1fedfa409c966658a747d653f84cce7357c3/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp#L94 .  At that point, we can't really know how many graph events it took to get the previous data, so we really need to clean up.  Alternatively, we could remove that check, but I feel like it is a good thing to test.

clalancette:
All right, a6f6a74 should fix the problems pointed out by @wjwwood , and also uses a predicate system as suggested by @hidmic.  Please take another look.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13349/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8277/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11079/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13401/)

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
wjwwood: I don't even understand what that check is doing? Maybe just getting coverage on that function? The value it returns isn't really important I think.

I mean if we ever added something to the node that happen to use the graph events, this would break because there would be more users. I think that change could just be changed to >= 1, if needed?

wjwwood (commented):
lgtm, except I'd get rid of the `notify_graph_change()` and instead change the "count users" check to be less fragile.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
clalancette: > I mean if we ever added something to the node that happen to use the graph events, this would break because there would be more users. I think that change could just be changed to >= 1, if needed?

That's true, we could do that.

Here's what I'm going to do.  To take advantage of the CI that already successfully completed, I'm going to merge this PR as-is.  That fixes one of our current problems in https://github.com/ros2/rmw/pull/293 .  I'll do an immediate follow-up to change that fragile `count_graph_users` test and remove the `notify_graph_change`, which should satisfy @wjwwood 's concerns.

PR #1504, opened 2021-01-06
Make the test of graph users more robust.

clalancette:
In particular, we don't know for sure that the graph will only
have 1 user; we just know that it should have *at least* one
user.  So change that check to be more robust.  This also
lets us get rid of the slightly strange call to 'notify_graph_change'
that was being used to cleanup graph users.

This is a followup to #1503


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13350/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8278/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11080/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13402/)

wjwwood (approved):

PR #1505, opened 2021-01-06
Allow to add/remove nodes thread safely in rclcpp::Executor 

ivanpauno:
This PR makes `Executor::add_node()` and `Executor::remove_node()` thread safe.
`memory_strategy_mutex_` was renamed to `mutex_` and now it's used to protect all internal collections.
Thread safety annotations were added mainly because of documentation purposes, we're not running the thread sanitizer in CI.

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -179,7 +179,6 @@ Executor::get_automatically_added_callback_groups_from_nodes()
 void
 Executor::add_callback_groups_from_nodes_associated_to_executor()
 {
-  std::lock_guard<std::mutex> guard{mutex_};
   for (auto & weak_node : weak_nodes_) {
     auto node = weak_node.lock();
     if (node) {
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -736,7 +736,6 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
 
   rcl_ret_t status =
     rcl_wait(&wait_set_, std::chrono::duration_cast<std::chrono::nanoseconds>(timeout).count());
-
   if (status == RCL_RET_WAIT_SET_EMPTY) {
     RCUTILS_LOG_WARN_NAMED(
       "rclcpp",
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -753,7 +753,8 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
 
 rclcpp::node_interfaces::NodeBaseInterface::SharedPtr
 Executor::get_node_by_group(
-  const rclcpp::memory_strategy::MemoryStrategy::WeakCallbackGroupsToNodesMap & weak_groups_to_nodes,
+  const rclcpp::memory_strategy::MemoryStrategy::WeakCallbackGroupsToNodesMap &
+  weak_groups_to_nodes,
   rclcpp::CallbackGroup::SharedPtr group)
 {
   if (!group) {
@@ -812,7 +813,8 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
 bool
 Executor::get_next_ready_executable_from_map(
   AnyExecutable & any_executable,
-  const rclcpp::memory_strategy::MemoryStrategy::WeakCallbackGroupsToNodesMap & weak_groups_to_nodes)
+  const rclcpp::memory_strategy::MemoryStrategy::WeakCallbackGroupsToNodesMap &
+  weak_groups_to_nodes)
 {
   bool success = false;
   std::lock_guard<std::mutex> guard{mutex_};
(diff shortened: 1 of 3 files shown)


ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13351/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8279/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11081/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13403/)

ivanpauno:
build: `--packages-up-to rclcpp`
test:  `--retest-until-fail 100 --packages-select rclcpp --ctest-args -R test_executor`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13352/)

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -730,6 +736,7 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
 
   rcl_ret_t status =
     rcl_wait(&wait_set_, std::chrono::duration_cast<std::chrono::nanoseconds>(timeout).count());
+
```
jacobperron, line 739: nit: revert this unrelated whitespace change.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -741,6 +748,7 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
 
   // check the null handles in the wait set and remove them from the handles in memory strategy
   // for callback-based entities
+  std::lock_guard<std::mutex> guard(mutex_);
   memory_strategy_->remove_null_handles(&wait_set_);
```
jacobperron, line 752: It's possible that `memory_strategy_` is modified during `rcl_wait()`. Should we be concerned with that? I don't know enough about the executor implementation to be sure.

jacobperron (commented):
This seems to fix the contrived test of repeatedly adding/removing nodes to the executor, but it does not resolve https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1169. But, perhaps the issue isn't not related to thread-safety in the executor.

rclcpp/src/rclcpp/executor.cpp
ivanpauno: > Should we be concerned with that?

`rcl_wait()` cannot modify `memory_strategy_`.
The reason to not hold `mutex_` locked while calling `rcl_wait()` is that in the following situation:
- Thread 1: `executor.spin();`
- Thread 2: `executor.add_node(node);`

Thread 2 might wait long if there happens to not be any event awakening the executor.

ivanpauno:
> This seems to fix the contrived test of repeatedly adding/removing nodes to the executor, but it does not resolve [ros-simulation/gazebo_ros_pkgs#1169](https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1169). But, perhaps the issue isn't not related to thread-safety in the executor.

Was https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1197 solving ros-simulation/gazebo_ros_pkgs#1169?

rclcpp/src/rclcpp/executor.cpp
ivanpauno:
```suggestion
```

ivanpauno: Fixed in https://github.com/ros2/rclcpp/pull/1505/commits/44524078585283875a7a7cf434f3f601e5f98af6.

rclcpp/src/rclcpp/executor.cpp
jacobperron: But while thread 1 is spinning, thread 2 can modify `memory_strategy_`, right? And then we are using the result of `rcl_wait()` to update `memory_strategy_` here.

jacobperron:
> Was ros-simulation/gazebo_ros_pkgs#1197 solving ros-simulation/gazebo_ros_pkgs#1169?

It appears to fix the issue, but I'm not sure where the bug is exactly. I'll try getting a better trace as to what's happening.

rclcpp/src/rclcpp/executor.cpp
ivanpauno: Good question, yes but that's ok.

`remove_node()`, `add_node()` only add/remove a `guard_condition` to `memory_strategy_`, and `remove_null_handles()`, `get_next_*()` methods aren't affected by that change.

There's a problem if you add or remove other entities (i.e. not a guard condition, e.g. a subscription) to `memory_strategy_` in the middle, but that only happens at the start of `wait_for_work()` with the mutex taken.

jacobperron:
I've posted a backtrace here (using this branch): https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1169#issuecomment-755658147

jacobperron (commented):
LGTM after fixing the lint errors. It'd be good to get @wjwwood to take a look too.

fujitatomoya (approved):

Issue #1506, opened 2021-01-06
Make the lifecycle services optional


Karsten1987:
There is currently no way to not have services attached to the lifecycle node. That prevents an easy setup of a lifecycle management system as each loaded lifecycle node can be individually controlled and thus does not have to go through the management system.
This is a precise use case we see at ros control, where each controller is a lifecycle node but should solely be controlled via the controller manager - not the controller itself - in order to guarantee a proper cleanup and resource management.

see https://github.com/ros-controls/ros2_control/issues/266

Karsten1987:
* Linux [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=13357)](http://ci.ros2.org/job/ci_linux/13357/)
* Linux-aarch64 [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=8284)](http://ci.ros2.org/job/ci_linux-aarch64/8284/)
* macOS [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=11086)](http://ci.ros2.org/job/ci_osx/11086/)
* Windows [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=13408)](http://ci.ros2.org/job/ci_windows/13408/)

Karsten1987:
CI once more:

* Linux [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=13774)](http://ci.ros2.org/job/ci_linux/13774/)
* Linux-aarch64 [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=8639)](http://ci.ros2.org/job/ci_linux-aarch64/8639/)
* macOS [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=11482)](http://ci.ros2.org/job/ci_osx/11482/)
* Windows [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=13851)](http://ci.ros2.org/job/ci_windows/13851/)

Karsten1987:
@fujitatomoya @clalancette What's the final verdict here? We'd like to get this in before the API freeze in order to move forward with ros-control in Galactic. 
It looks like the rcl related PR (ros2/rcl#882) got approved, the rclcpp PR (#1507) is lacking an approval/review.

fujitatomoya:
@Karsten1987 

[...]

PR #1507, opened 2021-01-06
make rcl_lifecyle_com_interface optional in lifecycle nodes

Karsten1987:
connects to #1506

### DIFF
--- a/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
+++ b/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
@@ -75,7 +75,7 @@ public:
   }
 
   void
-  init()
+  init(bool enable_communication_interface = true)
   {
     rcl_node_t * node_handle = node_base_interface_->get_rcl_node_handle();
     const rcl_node_options_t * node_options =
@@ -96,6 +96,7 @@ public:
       rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableStatesSrv>(),
       rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableTransitionsSrv>(),
       rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableTransitionsSrv>(),
+      enable_communication_interface,
       true,
       &node_options->allocator);
     if (ret != RCL_RET_OK) {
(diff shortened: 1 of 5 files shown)


rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -96,6 +96,7 @@ class LifecycleNode::LifecycleNodeInterfaceImpl
       rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableStatesSrv>(),
       rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableTransitionsSrv>(),
       rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableTransitionsSrv>(),
+      enable_communication_interface,
```
fujitatomoya, line 99: if this is false, publication on topic named `/<node_name>/transition_event` will not be initialized. but rclcpp_lifecycle will try to publish `lifecycle_msgs__msg__TransitionEvent` once `change_state` is called. i may be missing something, but i thnink this will cause transition error.

Karsten1987: very well seen! I think it makes sense to enable the event publisher in any case and only disable the actual services.
I'll change that accordingly in the rcl scope.

fujitatomoya: > I think it makes sense to enable the event publisher in any case and only disable the actual services.

So do I.

clalancette:
Same comment as https://github.com/ros2/rcl/pull/882#issuecomment-773302509

rclcpp_lifecycle/test/mocking_utils/patch.hpp
```diff
@@ -256,6 +256,30 @@ struct PatchTraits<ID, ReturnT(
     mock_type, ReturnT, ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5, ArgT6, ArgT7, ArgT8, ArgT9);
 };
 
+/// Traits specialization for void(ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5, ArgT6, ArgT7)
```
fujitatomoya, line 259: nitpick:

```suggestion
/// Traits specialization for ReturnT(ArgT0, ArgT1, ArgT2, ArgT3, ArgT4, ArgT5, ArgT6, ArgT7, ArgT8, ArgT9, ArgT10)
```

fujitatomoya (changes requested):

fujitatomoya:

fujitatomoya (approved):
lgtm

fujitatomoya:
@Karsten1987 there is DCO error, could you check that?

PR #1508, opened 2021-01-07
Better documentation for the QoS class

nnmm:
When I learned ROS2, this was one of the things that annoyed me, so I hope to spare others the experience. There is a tutorial, but I think it's best to keep documentation close to the code.

BTW, I think it would be great to link to hpp & cpp files in Doxygen. That would make having sparse documentation much more bearable.

### DIFF
--- a/rclcpp/include/rclcpp/qos.hpp
+++ b/rclcpp/include/rclcpp/qos.hpp
@@ -102,7 +102,7 @@ struct RCLCPP_PUBLIC KeepLast : public rclcpp::QoSInitialization
 class RCLCPP_PUBLIC QoS
 {
 public:
-  /// Constructor which allows creating a QoS by specifying only the history policy and history depth.
+  /// Create a QoS by specifying only the history policy and history depth.
   /**
    * When using the default initial profile, the defaults will include:
    *
--- a/rclcpp/include/rclcpp/qos.hpp
+++ b/rclcpp/include/rclcpp/qos.hpp
@@ -91,22 +91,30 @@ struct RCLCPP_PUBLIC KeepLast : public rclcpp::QoSInitialization
 
 /// Encapsulation of Quality of Service settings.
 /**
- * Quality of Service settings contains parameters for the underlying middleware, typically DDS,
- * to tune communication between nodes. See
- * <a href="https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/">
- About Quality of Service Settings</a> for an explanation.
+ * Quality of Service settings control the behavior of publishers, subscriptions,
+ * and other entities, and includes things like how data is sent or resent,
+ * how data is buffered on the publishing and subscribing side, and other things.
+ * See:
+ *   <a href="https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings">
+ *     https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings
+ *   </a>
  */
 class RCLCPP_PUBLIC QoS
 {
 public:
   /// Constructor which allows creating a QoS by specifying only the history policy and history depth.
   /**
-   * When using the default initial profile, this uses \link rclcpp::ReliabilityPolicy::Reliable
-   * ReliabilityPolicy::Reliable\endlink and \link rclcpp::DurabilityPolicy::Volatile
-   * DurabilityPolicy::Volatile\endlink. See the global variable rmw_qos_profile_default for a full list.
+   * When using the default initial profile, the defaults will include:
+   *
+   *   - \link rclcpp::ReliabilityPolicy::Reliable ReliabilityPolicy::Reliable\endlink
+   *   - \link rclcpp::DurabilityPolicy::Volatile DurabilityPolicy::Volatile\endlink
+   *
+   * See rmw_qos_profile_default for a full list of default settings.
+   * If some other rmw_qos_profile_t is passed to initial_profile, then the defaults will derive from
+   * that profile instead.
    *
    * \param[in] qos_initialization Specifies history policy and history depth.
-   * \param[in] initial_profile Specifies other settings.
+   * \param[in] initial_profile The rmw_qos_profile_t instance on which to base the default settings.
    */
   explicit
   QoS(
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/qos.hpp
```diff
@@ -90,18 +90,36 @@ struct RCLCPP_PUBLIC KeepLast : public rclcpp::QoSInitialization
 };
 
 /// Encapsulation of Quality of Service settings.
+/**
+ * Quality of Service settings contains parameters for the underlying middleware, typically DDS,
+ * to tune communication between nodes. See
+ * <a href="https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/">
+ About Quality of Service Settings</a> for an explanation.
+ */
```
wjwwood, line 98:
```suggestion
/**
 * Quality of Service settings control the behavior of publishers, subscriptions,
 * and other entities, and includes things like how data is sent or resent,
 * how data is buffered on the publishing and subscribing side, and other things.
 * See:
 *   <a href="https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings">
 *     https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings
 *   </a>
 */
```

We should avoid mentioning DDS or any underlying middleware specifics at this level of documentation.

Also, I prefer if we don't hind url's behind anchors. The url is human readable and more transparent if it is exposed directly (you know where you're going without mouse hovering).

rclcpp/include/rclcpp/qos.hpp
```diff
@@ -90,18 +90,36 @@ struct RCLCPP_PUBLIC KeepLast : public rclcpp::QoSInitialization
...
-  /// Constructor which allows you to construct a QoS by giving the only required settings.
+  /// Constructor which allows creating a QoS by specifying only the history policy and history depth.
+  /**
+   * When using the default initial profile, this uses \link rclcpp::ReliabilityPolicy::Reliable
+   * ReliabilityPolicy::Reliable\endlink and \link rclcpp::DurabilityPolicy::Volatile
+   * DurabilityPolicy::Volatile\endlink. See the global variable rmw_qos_profile_default for a full list.
+   *
+   * \param[in] qos_initialization Specifies history policy and history depth.
+   * \param[in] initial_profile Specifies other settings.
+   */
```
wjwwood, line 110:
```suggestion
  /// Constructor which allows creating a QoS by specifying only the history policy and history depth.
  /**
   * When using the default initial profile, the defaults will include:
   *
   *   - \link rclcpp::ReliabilityPolicy::Reliable ReliabilityPolicy::Reliable\endlink
   *   - \link rclcpp::DurabilityPolicy::Volatile DurabilityPolicy::Volatile\endlink
   *
   * See rmw_qos_profile_default for a full list of default settings.
   * If some other rmw_qos_profile_t is passed to initial_profile, then the defaults will derive from
   * that profile instead.
   *
   * \param[in] qos_initialization Specifies history policy and history depth.
   * \param[in] initial_profile The rmw_qos_profile_t instance on which to base the default settings.
   */
```

Just cleaning it up a bit.

Whenever possible we should try to avoid leaning on rmw or rcl in the rclcpp documentation. Ideally rclcpp's documentation would be self describing, but in some cases like this referring to rmw is counter productive.

rclcpp/include/rclcpp/qos.hpp
```diff
@@ -90,18 +90,36 @@ struct RCLCPP_PUBLIC KeepLast : public rclcpp::QoSInitialization
...
   QoS(
     const QoSInitialization & qos_initialization,
     const rmw_qos_profile_t & initial_profile = rmw_qos_profile_default);
 
   /// Conversion constructor to ease construction in the common case of just specifying depth.
   /**
-   * Convenience constructor, equivalent to QoS(KeepLast(history_depth)).
+   * This is a convenience constructor that calls
+   * QoS(const QoSInitialization&, const rmw_qos_profile_t &) with the first argument set
+   * to KeepLast(history_depth) and the second argument equal to the default.
```
wjwwood, line 120:
```suggestion
   * This is a convenience constructor that calls QoS(KeepLast(history_depth)).
```

I actually don't see that linking to the constructor is that use here, since it's pretty obvious what is happening, and the original comment was much more concise.

rclcpp/include/rclcpp/qos.hpp
```diff
@@ -90,18 +90,36 @@ struct RCLCPP_PUBLIC KeepLast : public rclcpp::QoSInitialization
...
+   *
+   * \param[in] history_depth How many messages can be queued up by the node.
```
wjwwood, line 122:
```suggestion
   * \param[in] history_depth How many messages can be queued when publishing
   *   with a Publisher, or how many messages can be queued before being replaced
   *   by a Subscription.
```

The node doesn't queue anything, and these settings do not influence a node, technically.

wjwwood (changes requested):
Thanks for the pull request. I had some suggestions. Please consider taking them, in part or whole, and I can re-review it after that.

nnmm:
Thanks for your comments @wjwwood!

wjwwood (dismissed):
lgtm, thanks again!

rclcpp/include/rclcpp/qos.hpp
```diff
@@ -90,18 +90,44 @@ struct RCLCPP_PUBLIC KeepLast : public rclcpp::QoSInitialization
...
+ * See:
+ *   <a href="https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings">
+ *     https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings
+ *   </a>
+ */
 class RCLCPP_PUBLIC QoS
 {
 public:
-  /// Constructor which allows you to construct a QoS by giving the only required settings.
+  /// Constructor which allows creating a QoS by specifying only the history policy and history depth.
```
wjwwood, line 105: Unfortunately, this line is too long, according to the linter failure.

```suggestion
  /// Create a QoS by specifying only the history policy and history depth.
```

Maybe just this, it follows the imperative style which we tend to prefer.

wjwwood (approved):

PR #1509, opened 2021-01-07
Add get_logging_directory method to rclcpp::Logger

ivanpauno:


rclcpp/include/rclcpp/logger.hpp
```diff
@@ -161,6 +162,10 @@ class Logger
   RCLCPP_PUBLIC
   void
   set_level(Level level);
+
+  RCLCPP_PUBLIC
+  rcpputils::fs::path
+  get_logging_directory() const;
```
wjwwood, line 168: What's the thinking behind making this a class method rather than a free function or a static class method?

jacobperron, line 168: It's probably worth adding some API docs similar to [rcl_get_logging_directory](https://github.com/ros2/rcl_logging/blob/a1edc81af3a7ac87fc63ce36bd890e7000dba7b7/rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h#L95-L103)

ivanpauno: I was going to make it a static class method, but maybe we wanted in the future to support different loggers to use different directories.
If we're never going to support that, a static method sounds better.

rclcpp/src/rclcpp/logger.cpp
```diff
@@ -64,4 +66,15 @@ Logger::set_level(Level level)
   }
 }
 
+rcpputils::fs::path
+Logger::get_logging_directory() const
+{
+  char * log_dir = NULL;
+  auto allocator = rcutils_get_default_allocator();
+  rcl_logging_get_logging_directory(allocator, &log_dir);
```
jacobperron, line 74: Consider throwing an exception if this call fails.

ivanpauno: Thanks, done in https://github.com/ros2/rclcpp/pull/1509/commits/1b6c60f52776d92f7966bd4210138cea76067333.

rclcpp/include/rclcpp/logger.hpp
ivanpauno: Added in https://github.com/ros2/rclcpp/pull/1509/commits/1b6c60f52776d92f7966bd4210138cea76067333.

ivanpauno: Modified to a `static` method in https://github.com/ros2/rclcpp/pull/1509/commits/1b6c60f52776d92f7966bd4210138cea76067333.

wjwwood: Well... why not a free function? I guess I'm trying to understand why this is the right namespace for it. It doesn't exactly map to the rcl function's style. That would be something more like `rclcpp::Logging::get_logging_directory()`, which I am not advocating for.

I actually think a free function would make more sense. Sorry for the back and forth.

rclcpp/include/rclcpp/logger.hpp
```diff
@@ -161,6 +162,16 @@ class Logger
   RCLCPP_PUBLIC
   void
   set_level(Level level);
+
+  /// Get the current logging directory.
+  /**
+   * For more details of how the logging directory is determined,
+   * see \ref rcl_logging_get_logging_directory.
```
wjwwood, line 169: Also, it would be nice to document the return value and possible exceptions thrown.

ivanpauno: See https://github.com/ros2/rclcpp/pull/1509/commits/3c42fa6c3485a7cdfd72969b7757c0a29c1b065d.

rclcpp/include/rclcpp/logger.hpp
ivanpauno: Sounds good to me, see https://github.com/ros2/rclcpp/pull/1509/commits/3c42fa6c3485a7cdfd72969b7757c0a29c1b065d.

The new function is defined in `rclcpp/logger.hpp`, but you can also include [rclcpp/logging.hpp](https://github.com/ros2/rclcpp/blob/master/rclcpp/resource/logging.hpp.em) and it will be available.

wjwwood (approved):

jacobperron (approved):
LGTM with green CI

fujitatomoya (approved):

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13374/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8298/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11100/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13425/)

jacobperron (approved):

PR #1514, opened 2021-01-13
get_parameters_service_ should return empty if allow_undeclared_ is false

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/1510

### DIFF
--- a/rclcpp/src/rclcpp/parameter_service.cpp
+++ b/rclcpp/src/rclcpp/parameter_service.cpp
@@ -47,7 +47,7 @@ ParameterService::ParameterService(
           response->values.push_back(param.get_value_message());
         }
       } catch (const rclcpp::exceptions::ParameterNotDeclaredException & ex) {
-        RCLCPP_WARN(rclcpp::get_logger("rclcpp"), "Failed to get parameters: %s", ex.what());
+        RCLCPP_DEBUG(rclcpp::get_logger("rclcpp"), "Failed to get parameters: %s", ex.what());
       }
     },
     qos_profile, nullptr);
@@ -68,7 +68,7 @@ ParameterService::ParameterService(
             return static_cast<rclcpp::ParameterType>(type);
           });
       } catch (const rclcpp::exceptions::ParameterNotDeclaredException & ex) {
-        RCLCPP_WARN(rclcpp::get_logger("rclcpp"), "Failed to get parameter types: %s", ex.what());
+        RCLCPP_DEBUG(rclcpp::get_logger("rclcpp"), "Failed to get parameter types: %s", ex.what());
       }
     },
     qos_profile, nullptr);
@@ -89,7 +89,7 @@ ParameterService::ParameterService(
           result = node_params->set_parameters_atomically(
             {rclcpp::Parameter::from_parameter_msg(p)});
         } catch (const rclcpp::exceptions::ParameterNotDeclaredException & ex) {
-          RCLCPP_WARN(rclcpp::get_logger("rclcpp"), "Failed to set parameter: %s", ex.what());
+          RCLCPP_DEBUG(rclcpp::get_logger("rclcpp"), "Failed to set parameter: %s", ex.what());
           result.successful = false;
           result.reason = ex.what();
         }
@@ -117,7 +117,7 @@ ParameterService::ParameterService(
         auto result = node_params->set_parameters_atomically(pvariants);
         response->result = result;
       } catch (const rclcpp::exceptions::ParameterNotDeclaredException & ex) {
-        RCLCPP_WARN(
+        RCLCPP_DEBUG(
           rclcpp::get_logger("rclcpp"), "Failed to set parameters atomically: %s", ex.what());
         response->result.successful = false;
         response->result.reason = "One or more parameters were not declared before setting";
@@ -137,7 +137,7 @@ ParameterService::ParameterService(
         auto descriptors = node_params->describe_parameters(request->names);
         response->descriptors = descriptors;
       } catch (const rclcpp::exceptions::ParameterNotDeclaredException & ex) {
-        RCLCPP_WARN(rclcpp::get_logger("rclcpp"), "Failed to describe parameters: %s", ex.what());
+        RCLCPP_DEBUG(rclcpp::get_logger("rclcpp"), "Failed to describe parameters: %s", ex.what());
       }
     },
     qos_profile, nullptr);
--- a/rclcpp/src/rclcpp/parameter_service.cpp
+++ b/rclcpp/src/rclcpp/parameter_service.cpp
@@ -41,12 +41,13 @@ ParameterService::ParameterService(
       const std::shared_ptr<rcl_interfaces::srv::GetParameters::Request> request,
       std::shared_ptr<rcl_interfaces::srv::GetParameters::Response> response)
     {
-      for (const auto & name : request->names) {
-        // Default construct param to NOT_SET
-        rclcpp::Parameter param;
-        node_params->get_parameter(name, param);
-        // push back NOT_SET when get_parameter() call fails
-        response->values.push_back(param.get_value_message());
+      try {
+        auto parameters = node_params->get_parameters(request->names);
+        for (const auto param : parameters) {
+          response->values.push_back(param.get_value_message());
+        }
+      } catch (const rclcpp::exceptions::ParameterNotDeclaredException & ex) {
+        RCLCPP_WARN(rclcpp::get_logger("rclcpp"), "Failed to get parameters: %s", ex.what());
       }
     },
     qos_profile, nullptr);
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/parameter_service.cpp
```diff
@@ -41,12 +41,13 @@ ParameterService::ParameterService(
...
-        node_params->get_parameter(name, param);
-        // push back NOT_SET when get_parameter() call fails
-        response->values.push_back(param.get_value_message());
+      try {
+        auto parameters = node_params->get_parameters(request->names);
+        for (const auto param : parameters) {
+          response->values.push_back(param.get_value_message());
+        }
+      } catch (const rclcpp::exceptions::ParameterNotDeclaredException & ex) {
+        RCLCPP_WARN(rclcpp::get_logger("rclcpp"), "Failed to get parameters: %s", ex.what());
```
ivanpauno, line 50: why logging a warning when an external request used the wrong name?
IMO, this should be ignored silently.

ivanpauno (changes requested):
If you now request:

```
ros2 service call /talker/get_parameters rcl_interfaces/srv/GetParameters "names: [use_sim_time, not_exist]"
```

and the answer is:

```
rcl_interfaces.srv.GetParameters_Response(values=[rcl_interfaces.msg.ParameterValue(type=1, bool_value=False, integer_value=0, double_value=0.0, string_value='', byte_array_value=[], bool_array_value=[], integer_array_value=[], double_array_value=[], string_array_value=[])])
```

How do you know if the given value corresponds to `use_sim_time` or `not_exist`?

fujitatomoya:
> How do you know if the given value corresponds to use_sim_time or not_exist?

in this case of https://github.com/ros2/rclcpp/pull/1514#pullrequestreview-567284794, with this PR it will return empty.

```
root@5c7a784cba7f:~/ros2_ws/colcon_ws# ros2 service call /talker/get_parameters rcl_interfaces/srv/GetParameters "names: [use_sim_time, not_exist]"
requester: making request: rcl_interfaces.srv.GetParameters_Request(names=['use_sim_time', 'not_exist'])

response:
rcl_interfaces.srv.GetParameters_Response(values=[])
```

this behavior is consistent with `get_parameter_types` and `describe_parameters` when allow_undeclare_parameter is false(default).

rclcpp/src/rclcpp/parameter_service.cpp
fujitatomoya: i think that it would be nice to show warning on the node that's been requested with `non-exist` parameter name when allow_undeclare_parameter is false on that node.

Barry-Xu-2018:
I discussed with @iuhilnehc-ynos. 

Share our thought.

From user perspective, it is better to reply something even if non-existent parameters is included in request parameters.
If a user request many parameters (only one parameter doesn't exist), return empty may make user confused what does service happen. 
So we think current behavior of `get_parameter_types` and `describe_parameters` are not good.   

If the parameter exist, related information should be returned to the caller.  
If the parameter doesn't exist, should return initialized default information (type is 'Not Set') to the caller.

fujitatomoya:
@Barry-Xu-2018 

thanks for sharing opinion. i thought that as well, that would be a direction we would take also.
but at the moment, i was thinking that user can see the parameters using `list_parameters` then request the operation via services. besides, i expect that is what the `allow_undeclare_parameter` flag is for.

rclcpp/src/rclcpp/parameter_service.cpp
ivanpauno: A warning is logged if the node is malfunctioning.

In this case, the node received a "bad service request" so in the response it should be encoded that the "request is bad".
But there's nothing malfunctioning, somebody else sent a "bad request", so no need to log a warn.

ivanpauno:
> in this case of #1514 (review), with this PR it will return empty.

ah, so if one name is wrong the result is completely empty (?).
I guess that's ok, particularly if it's consistent with the other service calls, but I'm not sure what was the intention.

rclcpp/src/rclcpp/parameter_service.cpp
fujitatomoya: that is correct, it makes sense. i will change these warnings into debug information.

fujitatomoya:
i pushed a new commit which is https://github.com/fujitatomoya/rclcpp/commit/0b31523f23b1a349db03bc1e4d7d510416f59eff for this PR. but it does not appear to be here, i am not sure why.

ivanpauno:
> i pushed a new commit which is [fujitatomoya@0b31523](https://github.com/fujitatomoya/rclcpp/commit/0b31523f23b1a349db03bc1e4d7d510416f59eff) for this PR. but it does not appear to be here, i am not sure why.

I can see it in the PR, sometimes github responds slowly ...

wjwwood:
I see the issue you guys have identified. But I wonder if, in addition to returning empty, we should also return a string that explains why? My thinking is that returning empty is not the same as throwing an exceptions (which is what the C++ API does), and if try to map that to the data centric response message I think we should have a `failure_reason` string field which is empty when it succeeds.

Aside from that, the change in behavior which is that when `allow_undeclared = false` and a non existing parameter is requested, nothing is returned (ideally in the future with a reason why as I suggested), and when `allow_undeclared = true` and a non existing parameter is requested, they are all returned with the non existing ones type as "not set".

ivanpauno (approved):

ivanpauno:
>  But I wonder if, in addition to returning empty, we should also return a string that explains why?

That sounds fine, but couldn't the same be argued for the "describe parameters" and "get parameters types" services?
IMHO, this is at least making things consistent.

Note to self: review the tests

hidmic (approved):
Looks reasonable to me, though it needs CI.

hidmic:
> My thinking is that returning empty is not the same as throwing an exceptions (which is what the C++ API does), and if try to map that to the data centric response message I think we should have a `failure_reason` string field which is empty when it succeeds.

I agree with not conflating return values and errors. But I think we can (and probably should) do better than that. I'd suggest we propagate exceptions implicitly. Implicit fields (like goal IDs in action messages) could carry exception type and reason as strings. The receiver can then throw a generic exception or map the exception type to a concrete type.

It's an ABI breaking change, but so is manually adding custom fields to message definitions.

ivanpauno:
Running CI, including downstream packages tests just in case:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13993/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8817/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11687/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14071/)

wjwwood:
> But I think we can (and probably should) do better than that. I'd suggest we propagate exceptions implicitly. Implicit fields (like goal IDs in action messages) could carry exception type and reason as strings. The receiver can then throw a generic exception or map the exception type to a concrete type.

That sounds nice, but it's a lot more work. As it involves changes in all the rmw implementations (and each is a bit different in how they implement services) as well as necessary changes in rmw/rcl/rclcpp/rclpy to expose the exception path.

fujitatomoya:
thanks for leaving the comment.

> > But I wonder if, in addition to returning empty, we should also return a string that explains why?

> That sounds fine, but couldn't the same be argued for the "describe parameters" and "get parameters types" services?
IMHO, this is at least making things consistent.

yes, that is the purpose of this PR. probably we can discuss on how to return the failure with explanation string as next step.

fujitatomoya:
CI is unstable because [system_test/test_rclcpp](https://github.com/ros2/system_tests/tree/master/test_rclcpp) needs to be updated accordingly.

fujitatomoya:

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14005/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8827/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11698/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14086/)

fujitatomoya:
CI: https://github.com/ros2/system_tests/pull/466#issuecomment-803687289

fujitatomoya:
CI green, this must go with https://github.com/ros2/system_tests/pull/466 which waits for at least one review.

ivanpauno:

hidmic:
> That sounds nice, but it's a lot more work. As it involves changes in all the rmw implementations (and each is a bit different in how they implement services) as well as necessary changes in rmw/rcl/rclcpp/rclpy to expose the exception path.

It is certainly more work. Though not that much if we do the same we did for actions i.e. tweak `rosidl` generators to add an extra field and just use it in client libraries. It'd be transparent to both `rmw` and `rcl`. It's not the prettiest approach, but it's  simple.

ivanpauno:
Going in, I will merge this and https://github.com/ros2/system_tests/pull/466 and then re-release `rclcpp` to make the `system_tests` PR job pass.

fujitatomoya:
@ivanpauno thanks 👍

PR #1517, opened 2021-01-15
Change to using unique_ptrs for DummyExecutor.

clalancette:
clang static analysis suggested that there was a possible memory
leak here, and it is right.  The test is expecting to throw during
the constructor, in which case the memory would be automatically
freed.  However, if the test did *not* throw for some reason,
we would leak the memory.  Switch to using a unique_ptr here
which will free the memory in all cases at the end of the scope.

### DIFF
--- a/rclcpp/test/rclcpp/test_executor.cpp
+++ b/rclcpp/test/rclcpp/test_executor.cpp
@@ -122,14 +122,14 @@ TEST_F(TestExecutor, constructor_bad_guard_condition_init) {
   auto mock = mocking_utils::patch_and_return(
     "lib:rclcpp", rcl_guard_condition_init, RCL_RET_ERROR);
   EXPECT_THROW(
-    std::unique_ptr<DummyExecutor>(new DummyExecutor()),
+    static_cast<void>(std::make_unique<DummyExecutor>()),
     rclcpp::exceptions::RCLError);
 }
 
 TEST_F(TestExecutor, constructor_bad_wait_set_init) {
   auto mock = mocking_utils::patch_and_return("lib:rclcpp", rcl_wait_set_init, RCL_RET_ERROR);
   RCLCPP_EXPECT_THROW_EQ(
-    std::unique_ptr<DummyExecutor>(new DummyExecutor()),
+    static_cast<void>(std::make_unique<DummyExecutor>()),
     std::runtime_error("Failed to create wait set in Executor constructor: error not set"));
 }
 
--- a/rclcpp/test/rclcpp/test_executor.cpp
+++ b/rclcpp/test/rclcpp/test_executor.cpp
@@ -122,14 +122,14 @@ TEST_F(TestExecutor, constructor_bad_guard_condition_init) {
   auto mock = mocking_utils::patch_and_return(
     "lib:rclcpp", rcl_guard_condition_init, RCL_RET_ERROR);
   EXPECT_THROW(
-    new DummyExecutor(),
+    std::make_unique<DummyExecutor>(),
     rclcpp::exceptions::RCLError);
 }
 
 TEST_F(TestExecutor, constructor_bad_wait_set_init) {
   auto mock = mocking_utils::patch_and_return("lib:rclcpp", rcl_wait_set_init, RCL_RET_ERROR);
   RCLCPP_EXPECT_THROW_EQ(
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13440/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8351/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11159/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13494/)

clalancette:
It turns out that Windows was cranky with that change because `std::make_unique` is marked as `[[nodiscard]]`.  Instead, use `std::unique_ptr`.  Here's another CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13441/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8352/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11160/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13495/)

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -122,14 +122,14 @@ TEST_F(TestExecutor, constructor_bad_guard_condition_init) {
   auto mock = mocking_utils::patch_and_return(
     "lib:rclcpp", rcl_guard_condition_init, RCL_RET_ERROR);
   EXPECT_THROW(
-    std::make_unique<DummyExecutor>(),
+    std::unique_ptr<DummyExecutor>(new DummyExecutor()),
```
wjwwood, line 125: You could have just done `auto throw_away = ...` instead. In fact this can be dangerous too, since make_unique (and make_shared) are carefully crafted to be exception safe.

But in this instance it's probably fine to just use `new`.

wjwwood: Or maybe `(void)` before it would be fine.

mm318: Should do it in a more C++ way:
```
static_cast<void>(std::make_unique<DummyExecutor>());
```
(Taken from https://stackoverflow.com/a/53581769)

clalancette: Changed now.

wjwwood (approved):

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -122,14 +122,14 @@ TEST_F(TestExecutor, constructor_bad_guard_condition_init) {
...
-    new DummyExecutor(),
+    static_cast<void>(std::make_unique<DummyExecutor>()),
     rclcpp::exceptions::RCLError);
 }
 
 TEST_F(TestExecutor, constructor_bad_wait_set_init) {
   auto mock = mocking_utils::patch_and_return("lib:rclcpp", rcl_wait_set_init, RCL_RET_ERROR);
   RCLCPP_EXPECT_THROW_EQ(
-    new DummyExecutor(),
+    static_cast<void>(std::make_unique<DummyExecutor>()),
```
wjwwood, line 132: I don't think we need to do anything with this, but why is this a pointer in the first place? Couldn't it be on the stack? Unlike some entities in rclcpp, the executor need not be a smart pointer.

wjwwood: It would probably end up longer anyways, because we'd need to "use" the variable we create to avoid unused variable warnings.

clalancette: I honestly don't remember why I wrote it this way originally.  I guess I could move it to the stack, but like you say it would probably be longer.  So I'll leave it for now.

clalancette:
One more CI with the latest changes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13449/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8360/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11173/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13512/)

PR #1520, opened 2021-01-19
Fix error when using IPC with StaticSingleThreadExecutor

mauropasse:
Currently trying to use IPC with the **StaticSingleThreadExecutor** yields:
```
  what():  'data' is empty
Aborted (core dumped)
```
This is because the new API `execute(data)` for waitables requires `data` to be non `nullptr`. This PR makes the waitable to take the data (the message itself, on a intra-process subscription) before using it as the `execute` parameter on the static executor.

Taking the oportunity also fixing here a `cppcheck` which was failing:
```
[static_executor_entities_collector.cpp:327]: (error: deallocuse) Dereferencing 'node_ptr' after it is deallocated / released
```

### DIFF
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -323,8 +323,7 @@ StaticExecutorEntitiesCollector::add_callback_group(
     throw std::runtime_error("Callback group was already added to executor.");
   }
   if (is_new_node) {
-    rclcpp::node_interfaces::NodeBaseInterface::WeakPtr node_weak_ptr(node_ptr);
-    weak_nodes_to_guard_conditions_[node_weak_ptr] = node_ptr->get_notify_guard_condition();
+    weak_nodes_to_guard_conditions_[node_ptr] = node_ptr->get_notify_guard_condition();
     return true;
   }
   return false;
--- a/rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
@@ -175,9 +175,10 @@ StaticSingleThreadedExecutor::execute_ready_executables()
   }
   // Execute all the ready waitables
   for (size_t i = 0; i < entities_collector_->get_number_of_waitables(); ++i) {
-    if (entities_collector_->get_waitable(i)->is_ready(&wait_set_)) {
-      std::shared_ptr<void> shared_ptr;
-      entities_collector_->get_waitable(i)->execute(shared_ptr);
+    auto waitable = entities_collector_->get_waitable(i);
+    if (waitable->is_ready(&wait_set_)) {
+      auto data = waitable->take_data();
+      waitable->execute(data);
     }
   }
 }


rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -175,9 +175,10 @@ StaticSingleThreadedExecutor::execute_ready_executables()
   }
   // Execute all the ready waitables
   for (size_t i = 0; i < entities_collector_->get_number_of_waitables(); ++i) {
-    if (entities_collector_->get_waitable(i)->is_ready(&wait_set_)) {
-      std::shared_ptr<void> shared_ptr;
-      entities_collector_->get_waitable(i)->execute(shared_ptr);
+    auto waitable = entities_collector_->get_waitable(i);
+    if (waitable->is_ready(&wait_set_)) {
+      auto data = waitable->take_data();
+      waitable->execute(data);
```
wjwwood, line 181: I'm wondering why we missed this in the first case, maybe you could think of a test to prevent this from regressing?

wjwwood (commented):
@ivanpauno I'd like you to have a look at these changes too.

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
```diff
@@ -323,8 +323,7 @@ StaticExecutorEntitiesCollector::add_callback_group(
     throw std::runtime_error("Callback group was already added to executor.");
   }
   if (is_new_node) {
-    rclcpp::node_interfaces::NodeBaseInterface::WeakPtr node_weak_ptr(node_ptr);
-    weak_nodes_to_guard_conditions_[node_weak_ptr] = node_ptr->get_notify_guard_condition();
+    weak_nodes_to_guard_conditions_[node_ptr] = node_ptr->get_notify_guard_condition();
```
ivanpauno, line 326: this looks like a false positive, right?
the change looks fine though

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
ivanpauno: I have no idea why we missed this in https://github.com/ros2/rclcpp/pull/1241.
Probably, the static single threaded executor is not being tested as much as the others (?).

ivanpauno (approved):
LGTM!
A regression test would be nice, but I think it's fine to add that in a follow up.

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13450/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8361/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11174/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13513/)

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
mauropasse: Yes, seemed like a false positive for me too.

ivanpauno:
Going in, thanks for the fix @mauropasse !

PR #1436, opened 2020-11-05
Enforce static parameter types

ivanpauno:
See https://github.com/ros2/rcl_interfaces/pull/115.

PR #1522, opened 2021-01-21
Enforce static parameter types

ivanpauno:
Alternative to https://github.com/ros2/rclcpp/pull/1436.

The following declarations will be valid:

```cpp
// declares an int parameter that cannot change of type
node->declare_parameter("my_param", 5);

// declares an int parameter that cannot change of type, but no default value is provided.
// if no param override is provided, it will start with type PARAMETER_NOT_SET
node->declare_parameter("my_param", rclcpp::PARAMETER_INTEGER);

// declares an int parameter that cannot change of type with a descriptor
node->declare_parameter("my_param", 5, descriptor);

// declares an int parameter that cannot change of type, no default value provided, and descriptor.
node->declare_parameter("my_param", rclcpp::PARAMETER_INTEGER, descriptor);

// dynamically typed parameter, no default value
node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET);

// dynamically typed parameter, no default value, descriptor
node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET, descriptor);

// dynamically typed parameter, default value
node->declare_parameter("my_param", 5, rclcpp::ParameterDescriptor{false});

// dynamically typed parameter, default value, descriptor
node->declare_parameter("my_param", 5, {descriptor, false});
```

The `rcl_interfaces::msg::ParameterDescriptor` wrapper I wrote (rclcpp::ParameterDescriptor) is probably not a great idea.
If we delete that we can make the last two lines look like:

```cpp
// dynamically typed parameter, default value
node->declare_parameter("my_param", 5, rcl_interfaces::msg::ParameterDescriptor{}, false, true);  // name, default, descriptor, ignore_overrides, dynamically typed

// dynamically typed parameter, default value, descriptor
node->declare_parameter("my_param", 5, descriptor, false, true);  // name, default, descriptor, ignore_overrides, dynamically typed
```

I don't mind either way.

The only case this PR will break:

```cpp
node->declare_parameter("my_param");
```

I can make that work like `node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET);` and deprecate it.
The issue is cases like `node->declare_parameter("my_param", rclcpp::ParameterValue{});`, which typically happens on wrappers over rclcpp API using default values. That will start throwing an exception.


rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -727,45 +734,53 @@ TEST_F(TestNode, declare_parameter_with_cli_overrides) {
...
-    rclcpp::ParameterValue value = node->declare_parameter("parameter_bool");
+    rclcpp::ParameterValue value = node->declare_parameter(
+      "parameter_bool", rclcpp::ParameterType::PARAMETER_BOOL);
     EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_BOOL);
     EXPECT_EQ(value.get<bool>(), true);
   }
   {
-    rclcpp::ParameterValue value = node->declare_parameter("parameter_int");
+    rclcpp::ParameterValue value = node->declare_parameter(
+      "parameter_int", rclcpp::ParameterType::PARAMETER_INTEGER);
```
jacobperron, line 744: I think we also need a test case for declaring an unset, typed parameter without an override. When I try this locally, I get a runtime exception. E.g.

    node->declare_parameter("my_param", rclcpp::PARAMETER_INTEGER);

without an override result in

```
terminate called after throwing an instance of 'rclcpp::exceptions::InvalidParameterTypeException'
  what():  parameter 'my_param' has invalid type: Wrong parameter type, parameter {my_param} is of type {not set}, setting it to {not set} is not allowed.
```

I expected that this case shouldn't produce an exception.

jacobperron (commented):
> The rcl_interfaces::msg::ParameterDescriptor wrapper I wrote (rclcpp::ParameterDescriptor) is probably not a great idea.

I guess removing it makes for slightly less code to maintain? I'm okay with adding the extra parameter the `declare_parameter` for the case of dynamically typed variables with a default value, since I expect it is a rare use-case. I also don't feel strongly either way.

+1 for keeping

    node->declare_parameter("my_param")

and have it behave like 

    node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET)

I don't know if we need to deprecate it :shrug: 

> The issue is cases like node->declare_parameter("my_param", rclcpp::ParameterValue{});, which typically happens on wrappers over rclcpp API using default values. That will start throwing an exception.

Hm, yeah this case seems tricky. On the surface of it, I would say that it should result in a dynamically typed parameter since it looks a lot like `node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET)`. I'm not sure how easy it is to handle this case though.

ivanpauno:
> Hm, yeah this case seems tricky. On the surface of it, I would say that it should result in a dynamically typed parameter since it looks a lot like node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET). I'm not sure how easy it is to handle this case though.

The problem is that I would expect most of the code written like `node->declare_parameter(name, rclcpp::ParameterValue{})` to not want a dynamically typed parameter. e.g.: https://github.com/ros-planning/navigation2/blob/8d6cd2878bae389d932982e759acb760f3291f63/nav2_util/include/nav2_util/node_utils.hpp#L84-L94

So I would like to at least deprecate that option, so people start using `node->declare_parameter(name, rclcpp::PARAMETER_*)`, to choose a specific type or dynamic typing is explicitly.

But the problem is that I don't have a way to deprecate that, more than detecting it at runtime and logging a warning (and I strongly prefer compile warnings in cpp).

So, we either break that strange use case (which I bet not much people is using), or deprecate it with a runtime warning.
(if someone figures out how to detect that case at compile time, that would be great)

rclcpp/test/rclcpp/test_node.cpp
ivanpauno: Thanks!

I knew I didn't test this case and I had a mental note about it, will fix.

ivanpauno:
>I guess removing it makes for slightly less code to maintain? I'm okay with adding the extra parameter to declare_parameter for the case of dynamically typed variables with a default value, since I expect it is a rare use-case. I also don't feel strongly either way.
>
> +1 for keeping
> 
> ```
> node->declare_parameter("my_param")
> ```

Sounds good!

jacobperron:
> The problem is that I would expect most of the code written like `node->declare_parameter(name, rclcpp::ParameterValue{})` to not want a dynamically typed parameter

To be fair, anyone using code written like this already has to deal with it being a dynamically typed value since that's the current behaviour, right? Allowing this syntax to continue to behave the same seems reasonable to me. In any case, I agree that most people will want to switch to the statically typed APIs. We should make sure to document it well and announce the feature as part of Galactic. I'm okay with adding some deprecation warnings if we really don't want people to be using the old API (though I don't see much wrong with them).

mjeronimo:
Of the list in the first comment, these are the two that I have trouble with:

```
// dynamically typed parameter, no default value
node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET);

// dynamically typed parameter, no default value, descriptor
node->declare_parameter("my_param", rclcpp::PARAMETER_NOT_SET, descriptor);

```
I find using PARAMETER_NOT_SET to mean dynamically typed is a bit confusing.

ivanpauno:
> I find using PARAMETER_NOT_SET to mean dynamically typed is a bit confusing.

+1.
I can extend the enumeration and make PARAMETER_NOT_SET and PARAMETER_DYNAMIC_TYPE different.

extra note: I (ab)used the existing PARAMETER_NOT_SET enumeration, because it doesn't make sense to use that value to specify a type (the `allowed_type` field in the descriptor).
At the same time, it doesn't make sense to use `PARAMETER_DYNAMIC_TYPE` to specify the current type of a parameter (currently, the `type` field of the descriptor is **updated** each time the parameter changes of type, it's not only being set when declaring).

ivanpauno:
> I can extend the enumeration and make PARAMETER_NOT_SET and PARAMETER_DYNAMIC_TYPE different.

Done!

jacobperron:
> I can extend the enumeration and make PARAMETER_NOT_SET and PARAMETER_DYNAMIC_TYPE different.

This raises questions for me. Does the type `PARAMETER_DYNAMIC_TYPE` get changed to the actual type when an override is provided? With this new dynamic type, do we have a use case for "NOT_SET" still (either the type is dynamic or it is static and must have a valid type)? It seems like we're just renaming the enum.

After the latest discussion, it seems it would be simpler (more clear semantically) to introduce a single field boolean `dynamically_typed` and not allow users to declare a statically typed parameter without a default.

mjeronimo:
> 
> After the latest discussion, it seems it would be simpler (more clear semantically) to introduce a single field boolean `dynamically_typed` and not allow users to declare a statically typed parameter without a default.

Yes, at the message/data structure level, I think having the type, value, and whether the type can dynamically change is the cleanest approach. At the API level, when defining parameters, we can either require that the user supplies an initial value for statically-typed parameters or have well-defined initial values if the API allows them not to provide explicit initial values.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -364,54 +400,144 @@ __declare_parameter_common(
...
     ignore_override);
 
   // If it failed to be set, then throw an exception.
   if (!result.successful) {
+    constexpr const char type_error_msg_start[] = "Wrong parameter type";
+    if (
+      0u == std::strncmp(
+        result.reason.c_str(), type_error_msg_start, sizeof(type_error_msg_start) - 1))
+    {
+      // TODO(ivanpauno): Refactor the logic so we don't need the above `strncmp` and we can
```
jacobperron, line 462: I don't have any immediate thoughts about how to do this, so I guess it's okay to do in a follow-up PR.

If you don't plan to tackle it in the near-term, can you create an issue for visibility after this PR is merged?

jacobperron (commented):
Overall, LGTM. Some minor comments.

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
```diff
@@ -45,6 +45,17 @@ struct OnSetParametersCallbackHandle
   OnParametersSetCallbackType callback;
 };
 
+#define RCLCPP_INTERNAL_NODE_PARAMETERS_INTERFACE_DEPTRECATE_DECLARE \
```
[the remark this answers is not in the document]
ivanpauno: Done, see https://github.com/ros2/rclcpp/pull/1522/commits/81deafe4cb4a840bd8b7f3ca8262145af47c27b4.

rclcpp/include/rclcpp/parameter_value.hpp
```diff
@@ -32,6 +32,8 @@ namespace rclcpp
 enum ParameterType : uint8_t
 {
   PARAMETER_NOT_SET = rcl_interfaces::msg::ParameterType::PARAMETER_NOT_SET,
+  /// Special value, only used to declare a dynamically typed parameter.
+  /// Not valid as a parameter type.
```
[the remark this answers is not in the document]
ivanpauno:
```suggestion
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
ivanpauno: > If you don't plan to tackle it in the near-term, can you create an issue for visibility after this PR is merged?

Sure, I think the current code is okayish though.

jacobperron (approved):
LGTM. It be good for @mjeronimo to take another look.

Also, we should add a release note for Galactic.

mjeronimo (approved):
Looks good to me.

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13690/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8569/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11407/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13761/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13696/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8573/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11412/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13767/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -305,16 +305,58 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * \param[in] name The name of the parameter.
+   * \param[in] type Desired type of the parameter, which will enforced at runtime.
+   * \param[in] parameter_descriptor An optional, custom description for
+   *   the parameter.
+   * \param[in] ignore_override When `true`, the parameter override is ignored.
+   *    Default to `false`.
+   * \return A const reference to the value of the parameter.
+   * \throws Same as the previous overload taking a default value.
+   * \throws rclcpp::exceptions::InvalidParameterTypeException
+   *   if an override is not provided or the provided override is of the wrong type.
```
wjwwood, line 334: Wait so you're gonna throw `rclcpp::exceptions::InvalidParameterTypeException` when the user doesn't provide an override? Shouldn't we have a more precise error?

I guess the idea is that no override results in `NOT_SET` which will not match the ParameterType given? It's just surprising.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -305,16 +305,58 @@ class Node : public std::enable_shared_from_this<Node>
...
+
+  [[deprecated(
+    "declare_parameter() with only a name is deprecated and will be deleted in the future.\n" \
+    "If you want to declare a parameter that won't change type without a default value use:\n" \
+    "`node->declare_parameter<ParameterT>(name)`, where e.g. ParameterT=int64_t.\n\n" \
+    "If you want to declare a parameter that can dynamically change type use:\n" \
+    "```\n" \
+    "rcl_interfaces::msg::ParameterDescriptor descriptor;\n" \
+    "descriptor.dynamic_typing = true;\n" \
+    "node->declare_parameter(name, rclcpp::ParameterValue{}, descriptor);" \
```
wjwwood, line 353: Why not this?

```suggestion
    "node->declare_parameter(name, rclcpp::ParameterValue{}, descriptor);\n" \
```

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -172,6 +172,23 @@ Node::declare_parameter(
...
+  bool ignore_override)
+{
+  // get advantage of parameter value template magic to get
+  // the correct rclcpp::ParameterType from ParameterT
+  rclcpp::ParameterValue value{ParameterT{}};
+  return this->declare_parameter(
+    name,
+    value.get_type(),
+    parameter_descriptor,
+    ignore_override).get<ParameterT>();
```
wjwwood, line 189: This is easier to read imo:

```suggestion
  return this->declare_parameter(
    name,
    value.get_type(),
    parameter_descriptor,
    ignore_override
  ).get<ParameterT>();
```

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
```diff
@@ -45,6 +45,17 @@ struct OnSetParametersCallbackHandle
...
 
+#define RCLCPP_INTERNAL_NODE_PARAMETERS_INTERFACE_DEPRECATE_DECLARE \
+  "declare_parameter() with only a name is deprecated and will be deleted in the future.\n" \
+  "If you want to declare a parameter that won't change type without a default value use:\n" \
+  "`node_params->declare_parameter(name, type)`, with e.g. type=rclcpp::PARAMETER_INTEGER.\n\n" \
+  "If you want to declare a parameter that can dynamically change type use:\n" \
+  "```\n" \
+  "rcl_interfaces::msg::ParameterDescriptor descriptor;\n" \
+  "descriptor.dynamic_typing = true;\n" \
+  "node_params->declare_parameter(name, rclcpp::ParameterValue{}, descriptor);" \
```
wjwwood, line 56:
```suggestion
  "node_params->declare_parameter(name, rclcpp::ParameterValue{}, descriptor);\n" \
```

wjwwood (commented):
What about `rclcpp_lifecyle::LifecycleNode`?

https://github.com/ros2/rclcpp/blob/ab743392e4daf546eb0de704936b5f14ef6e0417/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp#L277-L287

I think we need a note added here:

https://index.ros.org/doc/ros2/Releases/Release-Galactic-Geochelone/#changes-since-the-foxy-release

rclcpp/include/rclcpp/node.hpp
ivanpauno: > Shouldn't we have a more precise error?

Sounds reasonable to me.
I'm not sure about naming though, does `rclcpp::exceptions::NoOverrideProvided` sounds good?

> I guess the idea is that no override results in NOT_SET which will not match the ParameterType given? It's just surprising.

Yes, that was the idea.
`undeclare_parameter` will also throw a `InvalidParameterTypeException` exception, but with a better error message than this case.
I'm not sure if we also need a different exception for that one too https://github.com/ros2/rclcpp/blob/0e9be57f3cb9941fc96a606f7ada9a0481ee0627/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L560-L563.

rclcpp/include/rclcpp/node.hpp
ivanpauno: ups, thanks!

ivanpauno:
> What about rclcpp_lifecyle::LifecycleNode?

Good question, updated in https://github.com/ros2/rclcpp/pull/1522/commits/89597aa9155733b38d25f278c31470af367cf3b7.

ivanpauno:
Another round of CI (previous build warnings should be solved):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13714/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8586/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11425/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13786/)

rclcpp/include/rclcpp/node.hpp
wjwwood: Hmm, I suppose the error type doesn't matter that much as long as the error message is clear, except if there's a case where they want to catch one but not the other error case.

In general, I'm in favor of more, and more specific, exception types.

> does rclcpp::exceptions::NoOverrideProvided sounds good?

`NoParameterOverrideProvided`? I don't feel strongly about the name. Just that the use of the existing exception type was a bit weird. Even if it was kind of correct.

wjwwood (approved):
lgtm, with one pending discussion about exception names.

rclcpp/include/rclcpp/node.hpp
ivanpauno: Fixed in https://github.com/ros2/rclcpp/pull/1522/commits/89a734e1d664ec0b2c5043fc997f126d43256acf

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13718/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8589/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11430/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13790/)

ivanpauno:
Windows once again, after fixing compile warnings (see https://github.com/ros2/rclcpp/pull/1522/commits/8f0e6382ab1c0b58928072a6a08b841eaa41de85):

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13791/)

wjwwood (approved):

ivanpauno:
@ros-pull-request-builder retest this please.

ivanpauno:

ivanpauno:
The Rpr checker failures aren't related to this and are happening in other places

ros-discourse:

https://discourse.ros.org/t/feedback-required-should-parameters-be-able-to-change-of-type/18319/2

PR #1535, opened 2021-01-29
clear statistics after window reset (#1531)

### LINKED ISSUE #1531: topic statistics results include measurement values out of window
## Bug report

topic statistics results don't reset by window updating so that the results include values out of a window.
For instance, this behavior can cause steady-state results hidden if the maximum value is recorded at the start-up phase.

I think statistics results should be calculated within a window.

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - from source
- Version or commit hash:
  - https://github.com/ros2/rclcpp/commit/83af414811b806ae614344e6cdedeaf88b8f9499
  - https://github.com/ros-tooling/libstatistics_collector/commit/b629461b74860bac5aae1d80e091684f249dc339
- DDS implementation:
  - Fast-DDS
  - rclcpp

#### Steps to reproduce issue

```bash
$ cd ~/ros2_ws/
$ git clone -b check_statistics_transition_state https://github.com/hsgwa/intra_process_ping_latency_test.git src/intra_process_ping_latency_test
$ colcon build --packages-select intra_process_ping_latency_test
$ . ~/ros2_ws/install/local_setup.bash
$ ros2 run intra_process_ping_latency_test ping_test

$ # in other terminal
$ PYTHONUNBUFFERED=1 ros2 topic echo /topic_statistics | grep message_age -A 18 --line-buffered | grep "data_type: 3" -A 1 --line-buffered | grep "data:"
# print out message_age minimum values
```
Here, ping_test adjusts message_age=1s for first 10s to imitate transition state.
After that, message_age fall down to about 1 ms order value.
message_age_timeseries

#### Expected behavior

```bash
$ PYTHONUNBUFFERED=1 ros2 topic echo /topic_statistics | grep message_age -A 18 --line-buffered | grep "data_type: 3" -A 1 --line-buffered | grep "data:"
  data: 1000.388805 # Transition state. message_age are adjusted to 1000 ms
  data: 1000.388805
  data: 0.791165  # 10 s later
  data: 0.361127 # max value doesn't include transition state.
  data: 0.361127
```

#### Actual behavior

```bash
$ PYTHONUNBUFFERED=1 ros2 topic echo /topic_statistics | grep message_age -A 18 --line-buffered | grep "data_type: 3" -A 1 --line-buffered | grep "data:"
  data: 1001.457474 # Transition state. message_age are adjusted to 1000 ms
  data: 1001.457474
  data: 1001.457474
  data: 1001.457474  # 10 s later
  data: 1001.457474  # max value includes transition state.
  data: 1001.457474
  data: 1001.457474
  data: 1001.457474
```
#### Additional information
The above outputs omit window_start and window_stop though, every window is updated to prevent overlap.

```
$ PYTHONUNBUFFERED=1 ros2 topic echo /topic_statistics | grep message_age -A 18 --line-buffered  | grep window -A 2
window_start:
  sec: 1611812308
  nanosec: 560635188
window_stop:
  sec: 1611812309
  nanosec: 560659761
--
window_start:
  sec: 1611812309
  nanosec: 560659761
window_stop:
  sec: 1611812310
  nanosec: 560712642
```

From the source code, it seems that resetting the statistics here will solve this issue.
https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp#L154

hsgwa:
Closes #1531 

Signed-off-by: hsgwa <hasegawa@isp.co.jp>

### DIFF
--- a/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
@@ -127,7 +127,7 @@ public:
   /**
    * This method acquires a lock to prevent race conditions to collectors list.
    */
-  virtual void publish_message()
+  virtual void publish_and_reset_message()
   {
     std::vector<MetricsMessage> msgs;
     rclcpp::Time window_end{get_current_nanoseconds_since_epoch()};
--- a/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
@@ -152,6 +152,12 @@ public:
       publisher_->publish(msg);
     }
     window_start_ = window_end;
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
+      for (auto & collector : subscriber_statistics_collectors_) {
+        collector->ClearCurrentMeasurements();
+      }
+    }
   }
 
 protected:
--- a/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
@@ -127,7 +127,7 @@ public:
   /**
    * This method acquires a lock to prevent race conditions to collectors list.
    */
-  virtual void publish_message_and_reset()
+  virtual void publish_message_and_reset_measurements()
   {
     std::vector<MetricsMessage> msgs;
     rclcpp::Time window_end{get_current_nanoseconds_since_epoch()};
--- a/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
@@ -136,6 +136,7 @@ public:
       std::lock_guard<std::mutex> lock(mutex_);
       for (auto & collector : subscriber_statistics_collectors_) {
         const auto collected_stats = collector->GetStatisticsResults();
+        collector->ClearCurrentMeasurements();
 
         auto message = libstatistics_collector::collector::GenerateStatisticMessage(
           node_name_,
@@ -152,12 +153,6 @@ public:
       publisher_->publish(msg);
     }
     window_start_ = window_end;
-    {
-      std::lock_guard<std::mutex> lock(mutex_);
-      for (auto & collector : subscriber_statistics_collectors_) {
-        collector->ClearCurrentMeasurements();
-      }
-    }
   }
 
 protected:
--- a/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
+++ b/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
@@ -127,7 +127,7 @@ public:
   /**
    * This method acquires a lock to prevent race conditions to collectors list.
    */
-  virtual void publish_and_reset_message()
+  virtual void publish_message_and_reset()
   {
     std::vector<MetricsMessage> msgs;
     rclcpp::Time window_end{get_current_nanoseconds_since_epoch()};
--- a/rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
+++ b/rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
@@ -556,7 +556,7 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_stats_include_window
     switch (type) {
       case StatisticDataType::STATISTICS_DATA_TYPE_MINIMUM:
       case StatisticDataType::STATISTICS_DATA_TYPE_MAXIMUM:
-        EXPECT_GT(stats_point.data, message_age_offset);
+        EXPECT_GE(stats_point.data, message_age_offset);
         break;
       default:
         break;
--- a/rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
+++ b/rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
@@ -57,6 +57,9 @@ constexpr const std::chrono::seconds kTestDuration{10};
 constexpr const uint64_t kNumExpectedMessages{8};
 constexpr const uint64_t kNumExpectedMessageAgeMessages{kNumExpectedMessages / 2};
 constexpr const uint64_t kNumExpectedMessagePeriodMessages{kNumExpectedMessages / 2};
+constexpr const std::chrono::seconds kTransitionDuration{
+  std::chrono::seconds{1} *kNumExpectedMessages / 2 / 2};
+constexpr const std::chrono::seconds kInitialMessageAgeOffset{std::chrono::seconds{1}};
 }  // namespace
 
 using rclcpp::msg::MessageWithHeader;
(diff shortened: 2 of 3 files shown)


ivanpauno:
@dabonnie @emersonknapp can you provide feedback here? Thanks!

mm318:
If I recall correctly, the clear should be called in the callback, such as at https://github.com/ros2/rclcpp/blob/900be20a5a75e4cff05b6e5fba5bd682b5cfb181/rclcpp/include/rclcpp/create_subscription.hpp#L104 because it is up to the user to decide if the statistics is a something like a rolling average or not.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
```diff
@@ -152,6 +152,12 @@ class SubscriptionTopicStatistics
       publisher_->publish(msg);
     }
     window_start_ = window_end;
+    {
```
dabonnie, line 155: Could you please add a unit test as well? Could be simple to check if a max or min is now outside of a window (see the existing test fixtures [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp)).

dabonnie (changes requested):
I agree with this change and the bug filed by @hsgwa (nice catch!). However, I think this change would greatly benefit from a unit test and updating the documentation on the ROS 2 Wiki to make the behavior clear.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
```diff
@@ -152,6 +152,12 @@ class SubscriptionTopicStatistics
...
+      std::lock_guard<std::mutex> lock(mutex_);
+      for (auto & collector : subscriber_statistics_collectors_) {
+        collector->ClearCurrentMeasurements();
```
fujitatomoya, line 158: What about calling this right after `statistics_msgs::msg::MetricsMessage` is created for each collector? i think it would be better for each collector to have less time window to reset.

mm318:
I'm still not convinced that `publish_message()` should semantically imply clearing measurements. I would refactor the new block of code into a `clear_measurements()` member function, and call that from the timer call back.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
mm318: This is a good point.

I'm just trying to think how to do it while satisfying https://github.com/ros2/rclcpp/pull/1535#issuecomment-770429078. Maybe wrap the `publish()` in an async call or rename this member function to `publish_message_and_reset_measurements()`

fujitatomoya:
> because it is up to the user to decide if the statistics is a something like a rolling average or not.

i think currently it only supports moving average described [how-statistics-are-calculated](https://index.ros.org/doc/ros2/Concepts/About-Topic-Statistics/#how-statistics-are-calculated), so it would be okay call clear in `publish_message` for now.

> I'm still not convinced that publish_message() should semantically imply clearing measurements. I would refactor the new block of code into a clear_measurements() member function, and call that from the timer call back.

but this makes also sense to me.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
hsgwa: OK.
I'll try to add a unit test as well.

hsgwa:
> I would refactor the new block of code into a clear_measurements() member function, and call that from the timer call back.

That make sense.
I understood that role of publish_message() is too large.
What about create new blocks such as "clear_measurement" and "clear_window"?

> because it is up to the user to decide if the statistics is a something like a rolling average or not.

I was thinking statistics should be calculated within window because average is documented as "moving" average.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
hsgwa: True.
The ClearCurrentMeasurements() is likely to be after GetStatisticsResults() at the earliest.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
```diff
@@ -127,7 +127,7 @@ class SubscriptionTopicStatistics
   /**
    * This method acquires a lock to prevent race conditions to collectors list.
    */
-  virtual void publish_message()
+  virtual void publish_and_reset_message()
```
mm318, line 130: Nitpick: should be `publish_message_and_reset()`.

mm318 (commented):
Would you also be able to add a unit test, as **dabonnie** mentioned?

hsgwa:
Sorry for the delay.
I've been interrupted by other tasks, but I'm working on this.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
hsgwa: Thanks.
I'd appreciate it if you could point out even the smallest details.

rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp
hsgwa: I've added a unit test.
Would you please check if it is the test you were expecting?

rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -476,3 +522,58 @@ TEST_F(TestSubscriptionTopicStatisticsFixture, test_receive_stats_for_message_wi
...
+
+  rclcpp::executors::SingleThreadedExecutor ex;
+  ex.add_node(msg_with_header_publisher);
+  ex.add_node(statistics_listener);
+  ex.add_node(msg_with_header_subscriber);
+
+  // Spin and get future
+  ex.spin_until_future_complete(statistics_listener->GetFuture(), kTestDuration);
+
+  const auto received_messages = statistics_listener->GetReceivedMessages();
```
dabonnie, line 548: Check that the number received is not zero or check against `kNumExpectedMessages`?

dabonnie (changes requested):
Overall LGTM with the addition of one minor check in the test. @wjwwood could you please review?

rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
hsgwa: Thanks.
I added this check.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13612/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8495/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11328/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13673/)

rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -57,6 +57,9 @@ constexpr const std::chrono::seconds kTestDuration{10};
 constexpr const uint64_t kNumExpectedMessages{8};
 constexpr const uint64_t kNumExpectedMessageAgeMessages{kNumExpectedMessages / 2};
 constexpr const uint64_t kNumExpectedMessagePeriodMessages{kNumExpectedMessages / 2};
+constexpr const std::chrono::seconds kTransitionDuration{
```
mm318, line 60: I think the naming of this constant is too specific to the context of https://github.com/ros2/rclcpp/issues/1531. Let's name it to something like `kUnstableMessageAgeWindowDuration`.

rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -57,6 +57,9 @@ constexpr const std::chrono::seconds kTestDuration{10};
...
+  std::chrono::seconds{1} *kNumExpectedMessages / 2 / 2};
```
mm318, line 61: Not sure why this is defined as this expression. This can be any duration longer than a collection window (100 ms), right?

How about using `kTestDuration / 2`? And add a comment that this can just be any duration that's between the duration of a collection window (100 ms) and the duration of the test (1 s).

rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
```diff
@@ -57,6 +57,9 @@ constexpr const std::chrono::seconds kTestDuration{10};
...
+constexpr const std::chrono::seconds kInitialMessageAgeOffset{std::chrono::seconds{1}};
```
mm318, line 62: Similarly, probably rename to `kUnstableMessageAgeOffset`.

rclcpp/test/rclcpp/topic_statistics/test_subscription_topic_statistics.cpp
hsgwa: Thank you for pointing it out!

> This can be any duration longer than a collection window (100 ms), right?

Theoretically yes.
(collection window is 1s for default)
But actually, there is a slight discrepancy between the timing of the message_age switch and the window reset.
So, the minimum value is collection window * 2.

I think kTestDuration is rather kTestTimeout.
The actual test time is window duration * window number.
I modified related variable names. Is it clear for you?

mm318: Ah, right. I was confusing `TransitionMessageStampPublisher`'s `publish_period` with the collection window.

It is very clear now. Thank you!

mm318 (approved):
Re-running CI after some updates were made to this pull request:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13616/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8498/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11331/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13676/)

PR #1551, opened 2021-02-11
Fix UBSAN warnings in any_subscription_callback.

clalancette:
When running the rclcpp tests under UBSAN, it complained that
the AnySubscriptionCallback tests were using an uninitialized
allocator.

Reviewing the code there, it also looks like the AnySubscriptionCallback
constructor wasn't checking for a null allocator.

Fix this by doing 3 things:

1.  Add a check for a null allocator in the AnySubscriptionCallback
constructor.
2.  Add a new test to test_any_subscription_callback that tests the new
nullptr handling.
3.  Fix the test fixture to initialize the allocator before trying to
call the AnySubscriptionCallback constructor.

### DIFF
--- a/rclcpp/test/rclcpp/test_any_subscription_callback.cpp
+++ b/rclcpp/test/rclcpp/test_any_subscription_callback.cpp
@@ -26,7 +26,8 @@ class TestAnySubscriptionCallback : public ::testing::Test
 {
 public:
   TestAnySubscriptionCallback()
-  : any_subscription_callback_(allocator_) {}
+  : allocator_(std::make_shared<std::allocator<void>>()),
+    any_subscription_callback_(allocator_) {}
   void SetUp()
   {
     msg_shared_ptr_ = std::make_shared<test_msgs::msg::Empty>();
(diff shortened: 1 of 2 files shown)


rclcpp/test/rclcpp/test_any_subscription_callback.cpp
```diff
@@ -45,6 +46,20 @@ class TestAnySubscriptionCallback : public ::testing::Test
...
+  // We need to wrap this in a function because `EXPECT_THROW` is a macro, and thinks
+  // that the comma in here splits macro arguments, not the template arguments.
+  rclcpp::AnySubscriptionCallback<
+    test_msgs::msg::Empty, std::allocator<void>> any_subscription_callback_(nullptr);
+}
+
+TEST(AnySubscription, null_allocator) {
+  EXPECT_THROW(
+    construct_with_null_allocator(),
+    std::runtime_error);
```
fujitatomoya, line 60: just curious if the following works? probably bit nicer instead of introducing `construct_with_null_allocator`?

```suggestion
    {
      rclcpp::AnySubscriptionCallback<
        test_msgs::msg::Empty, std::allocator<void>> any_subscription_callback_(nullptr);
    }, std::runtime_error);
```

fujitatomoya (approved):
basically LGTM, but one minor comment.

rclcpp/test/rclcpp/test_any_subscription_callback.cpp
clalancette: It ends up having the same problem.  I think the macro parser is dumb and just assumes that the next comma in there is splitting arguments, regardless what is there in front.

I could probably do something with token pasting, but in the end I thought that having a separate function was cleaner-looking.

But thanks anyway for the idea and the review.  I'm going to go ahead and run CI on this.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13651/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8531/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11367/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13716/)

clalancette:
All right, CI is green and I have an approval.  Going ahead and merging, thanks for the review.

Issue #1540, opened 2021-02-03
Unify clock behavior between use_sim_time true/false


doisyg:
## Feature request
It is my understanding that one can get very different behavior when asking for the node time when `use_sim_time` is `true` and the simulation time is published on the `/clock` topic, compared to when it is `false`.
See for instance this code that blocks forever in the timer_callback when `use_sim_time` is `true` and that behave properly when `use_sim_time` is `false`:

```
#include <rclcpp/rclcpp.hpp>
#include <chrono>
#include <std_msgs/msg/string.hpp>

using namespace std::chrono_literals;

class TimeTest: public rclcpp::Node
{
  public:
    TimeTest()
      : Node("time_test_node")
    {
      this->set_parameter(rclcpp::Parameter("use_sim_time", true));
      timer = this->create_wall_timer(2s, std::bind(&TimeTest::timer_callback, this));
    }

  private:
    rclcpp::TimerBase::SharedPtr timer;

    void timer_callback()
    {
      rclcpp::Time start_time = this->get_clock()->now();
      bool is_time_out = false;
      while(!is_time_out){
        rclcpp::Time time_now = this->get_clock()->now();
        rclcpp::Duration time_spent = time_now -start_time;
        RCLCPP_INFO_STREAM(this->get_logger(), " now (s): " << time_now.seconds());
        rclcpp::sleep_for(100ms);
        is_time_out = time_spent.seconds() > 1.0;
      }
    }
};

int main(int argc, char **argv)
{
  rclcpp::init(argc, argv);
  rclcpp::spin(std::make_shared<TimeTest>());
  rclcpp::shutdown();
  return 0;
}
```

It would be really useful to have the same behavior in both case.

#### Feature description
The node clock should be updated in a similar way with `use_sim_time` `true`/`false`, i.e no clock freeze specific to `use_sim_time` `true`

#### Implementation considerations
I guess the clock freeze `with use_sim_time` is `true` in callbacks because a node is by default mono-threaded and hence the simulated clock cannot be updated (the internal callback of the `/clock` topic cannot be served in parallel of another callback). 
One solution could be to add a specific thread to the node to serve the `/clock` topic and update the node clock when `use_sim_time` is `true` (in a transparent way for the user of the node).

wjwwood:
I think this should be "won't fix", as you shouldn't be blocking in a callback.

[...]

Issue #1454, opened 2020-11-13
Executor quick destruction after spin hangs


v-lopez:
## Bug report

**Required Info:**

- Operating System: Ubuntu 20.04
- Installation type: Binaries
- Version or commit hash: 5.0.0-1focal.20201007.211227
- DDS implementation: Fast DDS
- Client library (if applicable): rclcpp

#### Steps to reproduce issue
``` 
TEST(Foo, bar) {
  auto node = std::make_shared<rclcpp::Node>("test");
  auto executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
  executor->add_node(node);
  auto executor_spin_future = std::async(
    std::launch::async, [&executor]() -> void {
      executor->spin();
    });
  //  std::this_thread::sleep_for(50ms);
  executor->cancel();
}
```

#### Expected behavior
That the test ends.

#### Actual behavior
Very frequently, the test hangs on the future destruction, waiting for thread join();
Because the thread is stuck on `executor->spin()` (wait_for_work exactly)

#### Additional information
Uncommenting the `sleep_for` fixes the issue on my machine, so there seems to be a race condition.
In our use case, we have test than end faster than that time and trigger this issue, we can add the sleep as a workaround but it's ugly: https://github.com/ros-controls/ros2_control/pull/234/files/833b1661209cc12a9d6f9ef45c87f4072e641aa7#diff-28c9aa0318c48cffb66c294bbf17394b6f9295bb8b9fb83d2cac539699f7e354R130

ivanpauno:
The semantics of `cancel` is strange:

- If `spin` is already running, it will return. But you can spin again afterwards, and those will continue blocking.
- A `cancel` that happens while the executor is not spinning doesn't have any effect (the first `wait` of the executor will be awaken immediately because a guard condition was triggered by `cancel`, but `spin` won't return and will continue doing work).
- `rclpy` Executor doesn't have a `cancel` method, but rather a [shutdown one](https://github.com/ros2/rclpy/blob/25c2d02c73a4f32b955d6809ae4a4edd1272fff7/rclpy/rclpy/executors.py#L204). `spin()` calls after shutdown happened will not do work.

I'm not sure what is the intended semantics of `cancel`, but its current behavior doesn't sound quite useful (it will always be racy).

> Moreover, a `cancel` that happens while the executor is not spinning doesn't have any effect.

Maybe instead of that, the first `spin` after a cancel should not do work.
[...]

PR #1556, opened 2021-02-23
Clock subscription callback group spins in its own thread

anaelle-sw:
Good morning,

this pull request is motivated by the #1540 issue. In short, we would like the node's clock behavior to be consistent, independently from the `use_sim_time` parameter value. Currently, if using simulation time, the node's clock time won't be updated within callbacks. 

To always keep node's clock time up to date, we added a callback group to the existing clock subscription's options and make this callback group spins in its own thread, in [`TimeSource::create_clock_sub` method](https://github.com/wyca-robotics/rclcpp/blob/10ae308afd4194aab65b0dc9fd30d118425e272e/rclcpp/src/rclcpp/time_source.cpp#L249). The thread is cleanly finished in the [class destructor](https://github.com/wyca-robotics/rclcpp/blob/10ae308afd4194aab65b0dc9fd30d118425e272e/rclcpp/src/rclcpp/time_source.cpp#L182).

### DIFF
--- a/rclcpp/include/rclcpp/node_options.hpp
+++ b/rclcpp/include/rclcpp/node_options.hpp
@@ -258,6 +258,21 @@ public:
   NodeOptions &
   clock_qos(const rclcpp::QoS & clock_qos);
 
+
+  /// Return the use_clock_thread flag.
+  RCLCPP_PUBLIC
+  bool
+  use_clock_thread() const;
+
+  /// Set the use_clock_thread flag, return this for parameter idiom.
+  /**
+   * If true, a dedicated thread will be used to subscribe to "/clock"
+   * topic, using a callback group
+   */
+  RCLCPP_PUBLIC
+  NodeOptions &
+  use_clock_thread(bool use_clock_thread);
+
   /// Return a reference to the parameter_event_qos QoS.
   RCLCPP_PUBLIC
   const rclcpp::QoS &
@@ -383,6 +398,8 @@ private:
   bool start_parameter_event_publisher_ {true};
 
   rclcpp::QoS clock_qos_ = rclcpp::ClockQoS();
+
+  bool use_clock_thread_ {true};
 
   rclcpp::QoS parameter_event_qos_ = rclcpp::ParameterEventsQoS(
     rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_parameter_events)
--- a/rclcpp/include/rclcpp/node_options.hpp
+++ b/rclcpp/include/rclcpp/node_options.hpp
@@ -47,6 +47,7 @@ public:
    *   - start_parameter_services = true
    *   - start_parameter_event_publisher = true
    *   - clock_qos = rclcpp::ClockQoS()
+   *   - use_clock_thread = true
    *   - rosout_qos = rclcpp::RosoutQoS()
    *   - parameter_event_qos = rclcpp::ParameterEventQoS
    *     - with history setting and depth from rmw_qos_profile_parameter_events
@@ -266,8 +267,7 @@ public:
 
   /// Set the use_clock_thread flag, return this for parameter idiom.
   /**
-   * If true, a dedicated thread will be used to subscribe to "/clock"
-   * topic, using a callback group
+   * If true, a dedicated thread will be used to subscribe to "/clock" topic.
    */
   RCLCPP_PUBLIC
   NodeOptions &
--- a/rclcpp/include/rclcpp/time_source.hpp
+++ b/rclcpp/include/rclcpp/time_source.hpp
@@ -124,6 +124,11 @@ public:
   RCLCPP_PUBLIC
   ~TimeSource();
 
+protected:
+  // Dedicated thread for clock subscription.
+  bool use_clock_thread_;
+  std::thread clock_executor_thread_;
+
 private:
   // Preserve the node reference
   rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_;
@@ -140,9 +145,6 @@ private:
   // QoS of the clock subscription.
   rclcpp::QoS qos_;
 
-  // Options to use dedicated thread for clock subscription.
-  bool use_clock_thread_;
-
   // The subscription for the clock callback
   using MessageT = rosgraph_msgs::msg::Clock;
   using Alloc = std::allocator<void>;
@@ -151,7 +153,6 @@ private:
   std::mutex clock_sub_lock_;
   rclcpp::CallbackGroup::SharedPtr clock_callback_group_;
   rclcpp::executors::SingleThreadedExecutor clock_executor_;
-  std::thread clock_executor_thread_;
 
   // The clock callback itself
   void clock_cb(const rosgraph_msgs::msg::Clock::SharedPtr msg);
--- a/rclcpp/include/rclcpp/time_source.hpp
+++ b/rclcpp/include/rclcpp/time_source.hpp
@@ -153,6 +153,7 @@ private:
   std::mutex clock_sub_lock_;
   rclcpp::CallbackGroup::SharedPtr clock_callback_group_;
   rclcpp::executors::SingleThreadedExecutor clock_executor_;
+  bool is_clock_executor_cancelled_;
 
   // The clock callback itself
   void clock_cb(const rosgraph_msgs::msg::Clock::SharedPtr msg);
--- a/rclcpp/include/rclcpp/time_source.hpp
+++ b/rclcpp/include/rclcpp/time_source.hpp
@@ -58,7 +58,9 @@ public:
    * \param qos QoS that will be used when creating a `/clock` subscription.
    */
   RCLCPP_PUBLIC
-  explicit TimeSource(rclcpp::Node::SharedPtr node, const rclcpp::QoS & qos = rclcpp::ClockQoS());
+  explicit TimeSource(rclcpp::Node::SharedPtr node,
+                      const rclcpp::QoS & qos = rclcpp::ClockQoS(),
+                      bool use_clock_thread = true);
 
   /// Empty constructor
   /**
@@ -67,7 +69,8 @@ public:
    * \param qos QoS that will be used when creating a `/clock` subscription.
    */
   RCLCPP_PUBLIC
-  explicit TimeSource(const rclcpp::QoS & qos = rclcpp::ClockQoS());
+  explicit TimeSource(const rclcpp::QoS & qos = rclcpp::ClockQoS(),
+                     bool use_clock_thread = true);
 
   /// Attack node to the time source.
   /**
@@ -134,6 +137,9 @@ private:
 
   // QoS of the clock subscription.
   rclcpp::QoS qos_;
+
+  // Options to use dedicated thread for clock subscription.
+  bool use_clock_thread_;
 
   // The subscription for the clock callback
   using MessageT = rosgraph_msgs::msg::Clock;
--- a/rclcpp/include/rclcpp/time_source.hpp
+++ b/rclcpp/include/rclcpp/time_source.hpp
@@ -153,7 +153,7 @@ private:
   std::mutex clock_sub_lock_;
   rclcpp::CallbackGroup::SharedPtr clock_callback_group_;
   rclcpp::executors::SingleThreadedExecutor clock_executor_;
-  bool is_clock_executor_cancelled_;
+  std::promise<void> cancel_clock_executor_promise_;
 
   // The clock callback itself
   void clock_cb(const rosgraph_msgs::msg::Clock::SharedPtr msg);
--- a/rclcpp/include/rclcpp/time_source.hpp
+++ b/rclcpp/include/rclcpp/time_source.hpp
@@ -25,6 +25,7 @@
 #include "rcl_interfaces/msg/parameter_event.hpp"
 
 #include "rclcpp/node.hpp"
+#include "rclcpp/executors.hpp"
 #include "rclcpp/node_interfaces/node_parameters_interface.hpp"
 
 
@@ -140,6 +141,9 @@ private:
   using SubscriptionT = rclcpp::Subscription<MessageT, Alloc>;
   std::shared_ptr<SubscriptionT> clock_subscription_{nullptr};
   std::mutex clock_sub_lock_;
+  rclcpp::CallbackGroup::SharedPtr clock_callback_group_;
+  rclcpp::executors::SingleThreadedExecutor clock_executor_;
+  std::thread clock_executor_thread_;
 
   // The clock callback itself
   void clock_cb(const rosgraph_msgs::msg::Clock::SharedPtr msg);
--- a/rclcpp/include/rclcpp/time_source.hpp
+++ b/rclcpp/include/rclcpp/time_source.hpp
@@ -59,9 +59,9 @@ public:
    */
   RCLCPP_PUBLIC
   explicit TimeSource(
-      rclcpp::Node::SharedPtr node,
-      const rclcpp::QoS & qos = rclcpp::ClockQoS(),
-      bool use_clock_thread = true);
+    rclcpp::Node::SharedPtr node,
+    const rclcpp::QoS & qos = rclcpp::ClockQoS(),
+    bool use_clock_thread = true);
 
   /// Empty constructor
   /**
@@ -71,8 +71,8 @@ public:
    */
   RCLCPP_PUBLIC
   explicit TimeSource(
-      const rclcpp::QoS & qos = rclcpp::ClockQoS(),
-      bool use_clock_thread = true);
+    const rclcpp::QoS & qos = rclcpp::ClockQoS(),
+    bool use_clock_thread = true);
 
   /// Attack node to the time source.
   /**
--- a/rclcpp/include/rclcpp/time_source.hpp
+++ b/rclcpp/include/rclcpp/time_source.hpp
@@ -58,9 +58,10 @@ public:
    * \param qos QoS that will be used when creating a `/clock` subscription.
    */
   RCLCPP_PUBLIC
-  explicit TimeSource(rclcpp::Node::SharedPtr node,
-                      const rclcpp::QoS & qos = rclcpp::ClockQoS(),
-                      bool use_clock_thread = true);
+  explicit TimeSource(
+      rclcpp::Node::SharedPtr node,
+      const rclcpp::QoS & qos = rclcpp::ClockQoS(),
+      bool use_clock_thread = true);
 
   /// Empty constructor
   /**
@@ -69,8 +70,9 @@ public:
    * \param qos QoS that will be used when creating a `/clock` subscription.
    */
   RCLCPP_PUBLIC
-  explicit TimeSource(const rclcpp::QoS & qos = rclcpp::ClockQoS(),
-                     bool use_clock_thread = true);
+  explicit TimeSource(
+      const rclcpp::QoS & qos = rclcpp::ClockQoS(),
+      bool use_clock_thread = true);
 
   /// Attack node to the time source.
   /**
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -261,9 +261,11 @@ void TimeSource::create_clock_sub()
     options
   );
 
-  clock_executor_thread_ = std::thread([this]() {
-    clock_executor_.add_callback_group(clock_callback_group_, node_base_);
-    clock_executor_.spin();}
+  clock_executor_thread_ = std::thread(
+    [this]() {
+        clock_executor_.add_callback_group(clock_callback_group_, node_base_);
+        clock_executor_.spin();
+    }
   );
 }
 
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -271,19 +271,17 @@ void TimeSource::create_clock_sub()
     std::bind(&TimeSource::clock_cb, this, std::placeholders::_1),
     options
   );
-
-  clock_executor_thread_ = std::thread(
-    [this]() {
-      clock_executor_.add_callback_group(clock_callback_group_, node_base_);
-      clock_executor_.spin();
-    }
-  );
 }
 
 void TimeSource::destroy_clock_sub()
 {
   std::lock_guard<std::mutex> guard(clock_sub_lock_);
   clock_subscription_.reset();
+  if (clock_executor_thread_.joinable()) {
+    clock_executor_.cancel();
+    clock_executor_thread_.join();
+    clock_executor_.remove_callback_group(clock_callback_group_);
+  }
 }
 
 void TimeSource::on_parameter_event(const rcl_interfaces::msg::ParameterEvent::SharedPtr event)
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -179,7 +179,7 @@ TimeSource::~TimeSource()
     this->detachNode();
   }
 
-  if (clock_subscription_) {
+  if (clock_executor_thread_.joinable()) {
     clock_executor_.cancel();
     clock_executor_thread_.join();
   }
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -254,6 +254,14 @@ void TimeSource::create_clock_sub()
     );
   }
   options.callback_group = clock_callback_group_;
+  if (!clock_executor_thread_.joinable()) {
+    clock_executor_thread_ = std::thread(
+      [this]() {
+        clock_executor_.add_callback_group(clock_callback_group_, node_base_);
+        clock_executor_.spin();
+      }
+    );
+  }
 
   clock_subscription_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(
     node_parameters_,
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -252,12 +252,10 @@ void TimeSource::create_clock_sub()
     });
 
   if (use_clock_thread_) {
-    if (clock_callback_group_ == nullptr) {
-      clock_callback_group_ = node_base_->create_callback_group(
-        rclcpp::CallbackGroupType::MutuallyExclusive,
-        false
-      );
-    }
+    clock_callback_group_ = node_base_->create_callback_group(
+      rclcpp::CallbackGroupType::MutuallyExclusive,
+      false
+    );
     options.callback_group = clock_callback_group_;
     if (!clock_executor_thread_.joinable()) {
       clock_executor_thread_ = std::thread(
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -246,7 +246,10 @@ void TimeSource::create_clock_sub()
       rclcpp::QosPolicyKind::History,
       rclcpp::QosPolicyKind::Reliability,
     });
-  clock_callback_group_ = node_base_->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive, false);
+  clock_callback_group_ = node_base_->create_callback_group(
+    rclcpp::CallbackGroupType::MutuallyExclusive,
+    false
+  );
   options.callback_group = clock_callback_group_;
 
   clock_subscription_ = rclcpp::create_subscription<rosgraph_msgs::msg::Clock>(
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -149,6 +149,10 @@ void TimeSource::detachNode()
   sim_time_cb_handler_.reset();
   node_parameters_.reset();
   disable_ros_time();
+  if (clock_executor_thread_.joinable()) {
+    clock_executor_.cancel();
+    clock_executor_thread_.join();
+  }
 }
 
 void TimeSource::attachClock(std::shared_ptr<rclcpp::Clock> clock)
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -55,6 +55,7 @@ TimeSource::TimeSource(
 
 void TimeSource::attachNode(rclcpp::Node::SharedPtr node)
 {
+  use_clock_thread_ = node->get_node_options().use_clock_thread();
   attachNode(
     node->get_node_base_interface(),
     node->get_node_topics_interface(),
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -34,9 +34,9 @@ namespace rclcpp
 {
 
 TimeSource::TimeSource(
-    std::shared_ptr<rclcpp::Node> node,
-    const rclcpp::QoS & qos,
-    bool use_clock_thread)
+  std::shared_ptr<rclcpp::Node> node,
+  const rclcpp::QoS & qos,
+  bool use_clock_thread)
 : logger_(rclcpp::get_logger("rclcpp")),
   qos_(qos),
   use_clock_thread_(use_clock_thread)
@@ -45,8 +45,8 @@ TimeSource::TimeSource(
 }
 
 TimeSource::TimeSource(
-    const rclcpp::QoS & qos,
-    bool use_clock_thread)
+  const rclcpp::QoS & qos,
+  bool use_clock_thread)
 : logger_(rclcpp::get_logger("rclcpp")),
   qos_(qos),
   use_clock_thread_(use_clock_thread)
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -37,9 +37,9 @@ TimeSource::TimeSource(
   std::shared_ptr<rclcpp::Node> node,
   const rclcpp::QoS & qos,
   bool use_clock_thread)
-: logger_(rclcpp::get_logger("rclcpp")),
-  qos_(qos),
-  use_clock_thread_(use_clock_thread)
+: use_clock_thread_(use_clock_thread),
+  logger_(rclcpp::get_logger("rclcpp")),
+  qos_(qos)
 {
   this->attachNode(node);
 }
@@ -47,9 +47,9 @@ TimeSource::TimeSource(
 TimeSource::TimeSource(
   const rclcpp::QoS & qos,
   bool use_clock_thread)
-: logger_(rclcpp::get_logger("rclcpp")),
-  qos_(qos),
-  use_clock_thread_(use_clock_thread)
+: use_clock_thread_(use_clock_thread),
+  logger_(rclcpp::get_logger("rclcpp")),
+  qos_(qos)
 {
 }
 
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -33,16 +33,21 @@
 namespace rclcpp
 {
 
-TimeSource::TimeSource(std::shared_ptr<rclcpp::Node> node, const rclcpp::QoS & qos)
+TimeSource::TimeSource(std::shared_ptr<rclcpp::Node> node,
+                       const rclcpp::QoS & qos,
+                       bool use_clock_thread)
 : logger_(rclcpp::get_logger("rclcpp")),
-  qos_(qos)
+  qos_(qos),
+  use_clock_thread_(use_clock_thread)
 {
   this->attachNode(node);
 }
 
-TimeSource::TimeSource(const rclcpp::QoS & qos)
+TimeSource::TimeSource(const rclcpp::QoS & qos,
+                       bool use_clock_thread)
 : logger_(rclcpp::get_logger("rclcpp")),
-  qos_(qos)
+  qos_(qos),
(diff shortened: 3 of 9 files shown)


ivanpauno:
The approach sounds reasonable to me, but I'm not sure if we want to do this or not.
@wjwwood what do you think?

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -241,6 +246,11 @@ void TimeSource::create_clock_sub()
       rclcpp::QosPolicyKind::History,
       rclcpp::QosPolicyKind::Reliability,
     });
+  clock_callback_group_ = node_base_->create_callback_group(
```
fujitatomoya, line 249: does this need to be initialized every time `use_sim_time` becomes true? i think this should be initialized somewhere else just once and store it as in option here.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -178,6 +178,11 @@ TimeSource::~TimeSource()
   {
     this->detachNode();
   }
+
+  if (clock_subscription_) {
+    clock_executor_.cancel();
+    clock_executor_thread_.join();
```
fujitatomoya, line 184: i think `TimeSource::destroy_clock_sub` should be modified accordingly, for example, user can turn on -> off -> on -> off...

fujitatomoya (changes requested):

rclcpp/src/rclcpp/time_source.cpp
anaelle-sw: It is initialized in `create_clock_sub` method, which [automatically returns if the subscription to clock is already created](https://github.com/wyca-robotics/rclcpp/blob/2c6024c8f49cf8c766c8a180cbcda05589b57f2d/rclcpp/src/rclcpp/time_source.cpp#L236). 
In one hand, if the subscription is reset and initialized again, I agree that the callback group shouldn't be initialized again. But the callback group is not useful if `use_sim_time = false`. So it can be initialized only once, but may be not in the class constructor, since it might remain unused.
A solution would be to test in the `create_clock_sub` whether `clock_callback_group_` has already been initialized:
```suggestion
  if (clock_callback_group_ == nullptr) {
    clock_callback_group_ = node_base_->create_callback_group(
      rclcpp::CallbackGroupType::MutuallyExclusive,
      false
    );
  }
```
What do you think about this?

anaelle-sw: I also added [a test on `clock_executor_thread_.joinable()`](https://github.com/wyca-robotics/rclcpp/blob/a82efabf75f0fa78523985eda775815168ed6f18/rclcpp/src/rclcpp/time_source.cpp#L257) to make sure the thread has ended before starting a new one.

rclcpp/src/rclcpp/time_source.cpp
anaelle-sw: Ok, it is done [here](https://github.com/wyca-robotics/rclcpp/blob/a82efabf75f0fa78523985eda775815168ed6f18/rclcpp/src/rclcpp/time_source.cpp#L280)

fujitatomoya (commented):
i am okay with this approach.

anaelle-sw:
Is there anything blocking the merge? 
If it is about the failed checks, I don't really get how they are related to the PR commits. Could you help on solving these? Thanks

fujitatomoya:
@wjwwood @clalancette friendly ping, could you check on this when you got time?

clalancette:

wjwwood (commented):
This looks ok to me, but I do think we should make this optional, like in NodeOptions, having a way to control whether it is in a thread or just part of the node.

fujitatomoya:
> having a way to control whether it is in a thread or just part of the node.

agree. and also adding test would be better.

fujitatomoya:
@anaelle-sw would you address https://github.com/ros2/rclcpp/pull/1556#issuecomment-796332127?

anaelle-sw:
I added the options ` bool use_clock_thread` to `rclcpp::NodeOptions`. Few notes about it: 
- The name may no be perfectly chosen, I can change it if you have any better idea. 
- This option is set to `true` by default, as we consider the inconsistency between simulation and real time behaviors as an issue to be fixed. What do you think about it?
- This option is implemented the same way `rclcpp::QoS qos_clock` option is. Is it ok for you?



I will try to write some tests as you asked for. I think we can test this scenario:
 - 3 nodes: node A has `use_clock_thread` set to false, node B has `use_clock_thread` set to true, node C is regularly publishing on `/clock` topic
 - Only node C is spinning
 - Get `rclcpp::Node::now()` from nodes A and B, sleep for few seconds, then get both `rclcpp::Node::now()` again after sleep
 - Time from node A should be the same before and after sleep, and time from node B should have been updated.
 Do you think about other behaviors to test?

fujitatomoya:
I am okay with the implementation. probably another review can be used.

> This option is set to true by default

this behavior only happens when `use_sim_time` is enabled, and at least i think this should be true by default to address the original issue. so that application which does not know the implementation details can avoid the problem easily.

> Time from node A should be the same before and after sleep, and time from node B should have been updated.

i am not sure what we are testing here, but probably clock testing is not required. for me, above looks like we are testing simulation time and system time, not exactly about this fix.

how about the following? that would be something i would check.

- `use_clock_thread` option can be set via `NodeOptions` to `TimeSource`
- check `clock_executor_thread_` status with enabling/disabling `use_sime_time` parameter.
- sleeping in timer callback does not get frozen with `use_clock_thread` is default/true. (making sure original issue is fixed)

anaelle-sw:
> how about the following? that would be something i would check.
> 
>     * `use_clock_thread` option can be set via `NodeOptions` to `TimeSource`
> 
>     * check `clock_executor_thread_` status with enabling/disabling `use_sime_time` parameter.
> 
>     * sleeping in timer callback does not get frozen with `use_clock_thread` is default/true. (making sure original issue is fixed)

I added three tests in `test_time_source.cpp` to check each of these points. 
For the two first ones, I moved `use_clock_thread_` and `clock_executor_thread_` class attributes to `protected` (instead of `private`). I don't know if there is another way to test it.
The code to test the third point is a bit laborious for the momenyt, maybe there is a easier and clearer way to do it. Do you have any suggestions about it?

fujitatomoya (changes requested):
overall, i am okay with this. just a couple of minor comments for test and could you also check the warnings from https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/972/

rclcpp/test/rclcpp/test_time_source.cpp
```diff
@@ -516,3 +516,147 @@ TEST_F(TestTimeSource, no_pre_jump_callback) {
...
+};
+
+TEST_F(TestTimeSource, use_clock_thread_set_by_node_options) {
+  auto node_ = std::make_shared<rclcpp::Node>(
+        "clock_thread_testing_node",
+        rclcpp::NodeOptions().use_clock_thread(false));
+
+  TestTimeSourceClockOptions ts(node_);
+
+  ASSERT_FALSE(ts.GetUseClockThreadOption());
```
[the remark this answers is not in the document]
anaelle-sw: Done [here](https://github.com/ros2/rclcpp/blob/3f2cc72b8c23d32f7a873c72a2e328dcbeb5a265/rclcpp/test/rclcpp/test_time_source.cpp#L542): 
```c
  auto default_node_ = std::make_shared<rclcpp::Node>(
    "default_option_node");

  auto clock_thread_node_ = std::make_shared<rclcpp::Node>(
    "clock_thread_node",
    rclcpp::NodeOptions().use_clock_thread(true));

  auto no_clock_thread_node_ = std::make_shared<rclcpp::Node>(
    "no_clock_thread_node",
    rclcpp::NodeOptions().use_clock_thread(false));

  // Test value of use_clock_thread_ TimeSource attribute
  // when the different nodes are attached

  ClockThreadTestingTimeSource ts;

  ts.attachNode(default_node_);
  ASSERT_TRUE(ts.GetUseClockThreadOption());
  ts.detachNode();

  ts.attachNode(clock_thread_node_);
  ASSERT_TRUE(ts.GetUseClockThreadOption());
  ts.detachNode();

  ts.attachNode(no_clock_thread_node_);
  ASSERT_FALSE(ts.GetUseClockThreadOption());
  ts.detachNode();
```

rclcpp/test/rclcpp/test_time_source.cpp
```diff
@@ -516,3 +516,147 @@ TEST_F(TestTimeSource, no_pre_jump_callback) {
...
+}
+
+TEST_F(TestTimeSource, check_clock_thread_status) {
+  TestTimeSourceClockOptions ts;
```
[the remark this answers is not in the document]
anaelle-sw: Done [here](https://github.com/ros2/rclcpp/blob/3f2cc72b8c23d32f7a873c72a2e328dcbeb5a265/rclcpp/test/rclcpp/test_time_source.cpp#L575):
```c
  // Tests for node with use_clock_thread option forced to false
  auto no_clock_thread_node_ = std::make_shared<rclcpp::Node>(
    "no_clock_thread_node",
    rclcpp::NodeOptions().use_clock_thread(false));

  no_clock_thread_node_->set_parameter(rclcpp::Parameter("use_sim_time", true));
  ts.attachNode(no_clock_thread_node_);
  ASSERT_FALSE(ts.IsClockThreadJoinable());
  ts.detachNode();

  no_clock_thread_node_->set_parameter(rclcpp::Parameter("use_sim_time", false));
  ts.attachNode(no_clock_thread_node_);
  ASSERT_FALSE(ts.IsClockThreadJoinable());
  ts.detachNode();
```

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -141,6 +149,10 @@ void TimeSource::detachNode()
   sim_time_cb_handler_.reset();
   node_parameters_.reset();
   disable_ros_time();
+  if (clock_executor_thread_.joinable()) {
+    clock_executor_.cancel();
+    clock_executor_thread_.join();
+  }
```
rpaaron, line 155: Looks like if the destructor runs before before the `clock_executor_thread` manages to call `spin`, this will block for ever.
Seems unlikely, but there are several bug reports for this behaviour already in other areas.
Perhaps set a flag after `spin`, and keep calling `cancel` until the flag is true?

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -178,6 +190,11 @@ TimeSource::~TimeSource()
   {
     this->detachNode();
   }
+
+  if (clock_executor_thread_.joinable()) {
```
fujitatomoya, line 194: Sorry i just noticed that this is actually done in `TimeSource::detachNode`, so this statement should be deleted.

rclcpp/src/rclcpp/time_source.cpp
fujitatomoya: which is https://github.com/ros2/rclcpp/issues/1454, you are right. there's been discussion on that, it probably would take some time to support that. in the meantime, we'd want to have atomic internal flag which indicate spinning or not?

rclcpp/src/rclcpp/time_source.cpp
anaelle-sw: You're right, I forgot to remove it from destructor. Actually, this statement is also in function `destroy_clock_sub()`:
```c
void TimeSource::destroy_clock_sub()
{
  std::lock_guard<std::mutex> guard(clock_sub_lock_);
  clock_subscription_.reset();
  if (clock_executor_thread_.joinable()) {
    clock_executor_.cancel();
    clock_executor_thread_.join();
    clock_executor_.remove_callback_group(clock_callback_group_);
  }
}
```
and function `detachNode()` actually just replicates what's in `destroy_clock_sub()`. 
So maybe the cleaner way is to:
- keep  `destroy_clock_sub()` as it currently is
- call  `destroy_clock_sub()` from `detachNode()` (and remove redundant statements)
- still call `detachNode()` from destructor  (and remove redundant statements)

Are you ok with this?

rclcpp/src/rclcpp/time_source.cpp
anaelle-sw: I have tested to attach and detach several times in a raw a Node to a TimeSource, checking if the thread is running in between:
```c
auto node = std::make_shared<rclcpp::Node>("test_node");
node->set_parameter(rclcpp::Parameter("use_sim_time", true));

for(int i = 0; i < 10; i++) {
  ts.attachNode(node);
  std::cout << "[Node attached] Clock thread is running? : "  <<
               ts.IsClockThreadJoinable() << std::endl;
  ts.detachNode();
  std::cout << "[Node detached] Clock thread is running? : "  <<
               ts.IsClockThreadJoinable() << std::endl;
  std::cout << std::endl;
}
```
I didn't notice any unexpected behavior: the thread is running when the node is attached, and is not otherwise, and nothing is blocking.

fujitatomoya: it is unlikely to happen because of context switches and race condition. but in theory that happens, and if that happens it will hang up eventually. as far as i see, this needs to be taken care by atomic flag or something, at least https://github.com/ros2/rclcpp/issues/1454 is addressed.

anaelle-sw:
I encountered an issue concerning the clock subscription's callback group.
If you attach a node A requiring a dedicated clock thread, then a node B also requiring a dedicated clock thread, just like this example: 
```
auto node_1 = std::make_shared<rclcpp::Node>("node_1");
node_1->set_parameter(rclcpp::Parameter("use_sim_time", true));
ts.attachNode(node_1);
ts.detachNode();

auto node_2 = std::make_shared<rclcpp::Node>("node_2");
node_2->set_parameter(rclcpp::Parameter("use_sim_time", true));
ts.attachNode(node_2);
```
well the last line makes the application crash with error:
```c
  what():  Cannot create subscription, callback group not in node.
```


What happens is that when attaching `node_1` to the TimeSource , the `clock_callback_group` is initialized [here](https://github.com/ros2/rclcpp/blob/719c4b2ee91aee6fc1117883e95e2a32ccc3e3cf/rclcpp/src/rclcpp/time_source.cpp#L263) this way:
```c
if (clock_callback_group_ == nullptr) {
  clock_callback_group_ = node_base_->create_callback_group(
    rclcpp::CallbackGroupType::MutuallyExclusive,
    false
  );
}
```
So when attaching `node_2` to the TimeSource, the `clock_callback_group` is a ready initialized (so different from `nullptr`) and it will be directly passed to the clock subscription.
But the problem is that the TimeSource `node_base_` attribute changed in the meantime, so the callback group is not "valid" anymore because it is linked to the old node base. and so it cannot be passed to the subscription, which is created for the new node...

The two possible solutions I see for this are either:

1 - We don't check if `clock_callback_group` is already initialized, but always re-initialize it when creating a new clock subscription. It may be a loss of optimization in two cases: if `use_sim_time` parameter is set to `true`, then `false`, then `true` again; and when we attach, detach, and attach again the very same node that requires a dedicated clock thread. Actually, it just consists in reverting the changes related [this review](https://github.com/ros2/rclcpp/pull/1556#discussion_r583311498)

or

2 - We can keep the old node base value in an attribute, and before re-initializing the callback group we check if the old node base is different from the new one. 

I find the second solution less clear than the first one, but it would allow to keep the callback group initialization optimization. But in this case, shouldn't we optimize the subscription initialization too? 
What do you think of this?

By the way, I have tested all other possible `attachNode` sequences: node not requiring sub to clock, to node requiring sub clock but no dedicated thread, to node requiring both, etc...
For now, the only problematic sequence is from a node requiring sub clock and dedicated thread to another different node also requiring  sub clock and dedicated thread. 
Should I add this test to `test_time_source.cpp`?

rclcpp/src/rclcpp/time_source.cpp
anaelle-sw: Ok, it is done [here](https://github.com/ros2/rclcpp/pull/1556/commits/aa2e216e2043f5e23ea2c366cc5bae9deb6bf816).

rpaaron: because `is_clock_executor_cancelled_` is accessed by different threads, I think you'll need to make it `atomic<bool>`
Also, maybe you want a small sleep or yield that destructor, after `cancel` to allow the other thread to run and set the boolean.

And just for reference, there is another approach to dealing with this issue described here https://github.com/ros2/rclcpp/issues/1454#issuecomment-799695171

fujitatomoya: > I think you'll need to make it atomic<bool>

Just FYI, it is something like https://github.com/ros2/rclcpp/blob/35c89c8afcac89a43872f0bf5b763ff324c82b05/rclcpp/include/rclcpp/executor.hpp#L524

rclcpp/src/rclcpp/time_source.cpp
fujitatomoya: i think that this is correct to do. i believe that `destroy_clock_sub` should be called when `detachNode` is called, since `Node` related objects will be reset. (this is a bug in current implementation.)

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -255,6 +283,13 @@ void TimeSource::create_clock_sub()
 void TimeSource::destroy_clock_sub()
 {
   std::lock_guard<std::mutex> guard(clock_sub_lock_);
+  if (clock_executor_thread_.joinable()) {
+    while (!is_clock_executor_cancelled_) {
+      clock_executor_.cancel();
+    }
```
fujitatomoya, line 289: This still has the same problem if `clock_executor_.cancel()` is callbed before `clock_executor_.spin()`. after all, to support the race condition, i think https://github.com/ros2/rclcpp/issues/1454#issuecomment-799695171 is better to give the responsibility to the `Executor` for this.

fujitatomoya:
> We don't check if clock_callback_group is already initialized, but always re-initialize it when creating a new clock subscription.

i am convinced to do this way. sorry for taking back my comment before, but if we do this, we can keep the behavior simple and i guess cost is pretty small.

fujitatomoya:
> Should I add this test to test_time_source.cpp?

thanks for asking this 👍 i think basically the more unit tests there are, the better quality we can keep.

on the other hand, we would not want to have complications for maintenance perspective. but i believe that it is always worth to discuss, and probably we can keep the unit tests simple for more coverage.

rclcpp/src/rclcpp/time_source.cpp
rpaaron: @fujitatomoya you mean it will block forever if `create_clock_sub` is never called? yeah, is `create_clock_sub` guaranteed to be called? if so, its ok? if not, but `create_clock_sub`, if its called, is guaranteed to be called in the same thread as the destructor, then this flag can be initially set to true, an set to false only when `create_clock_sub` is called.
If that doesn't work, will need more sophisticated synchronisation machinery... which is https://github.com/ros2/rclcpp/issues/1454#issuecomment-799695171, looks like that would be a nice solution

fujitatomoya: Ah, i see what you mean, thanks for pointing that out. this should be working eventually but it might keep calling `Executor::cancel()` and it consumes CPU for nothing. as you mentioned, probably we would want more sophisticated synchronization.

anaelle-sw:
> > We don't check if clock_callback_group is already initialized, but always re-initialize it when creating a new clock subscription.
> 
> i am convinced to do this way. sorry for taking back my comment before, but if we do this, we can keep the behavior simple and i guess cost is pretty small.

Ok, no problem. It is done [in this commit](https://github.com/ros2/rclcpp/pull/1556/commits/d2e8d6d7500c7006483e772e36b193ba4e58d677)




> > Should I add this test to test_time_source.cpp?
> 
> thanks for asking this +1 i think basically the more unit tests there are, the better quality we can keep.
> 
> on the other hand, we would not want to have complications for maintenance perspective. but i believe that it is always worth to discuss, and probably we can keep the unit tests simple for more coverage.

If ever needed, I have kept these tests in our own repo. By the way, I can confirm that this commit fixed it, it is now possible to  do this: 

> ```c
> auto node_1 = std::make_shared<rclcpp::Node>("node_1");
> node_1->set_parameter(rclcpp::Parameter("use_sim_time", true));
> ts.attachNode(node_1);
> ts.detachNode();
> 
> auto node_2 = std::make_shared<rclcpp::Node>("node_2");
> node_2->set_parameter(rclcpp::Parameter("use_sim_time", true));
> ts.attachNode(node_2);
>```

without any crash.

rclcpp/src/rclcpp/time_source.cpp
anaelle-sw: I applied the method you recommended [here](https://github.com/ros2/rclcpp/pull/1556/commits/41eba461681f9ea8a819020dd6e7721687e8401f)

fujitatomoya (approved):
lgtm

fujitatomoya:
https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/991/testReport/ failure is unrelated.

@anaelle-sw 

there are 5 GCC warnings, could you address those https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/991/gcc/new/ ?

anaelle-sw:
> https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/991/testReport/ failure is unrelated.

Actually, yesterday I tried to figure out what was causing this. It seems that it is directly related to the executor cancellation. I mean after calling the `promise.set_value()`, if `executor.cancel()` is also called in `destroy_clock_sub` function, the tests are not throwing this error anymore. I tried many different combination of this, with setting the future and/or the promise outside the clock thread, or having the future as a class attribute... The only way to not having the error `The test did not generate a result file` is to call `executor.cancel()`.
I don't really understand why, because I tested the executor status after calling `destroy_clock_sub` or `detachNode`, and it is never spinning even when not calling `executor.cancel()`, which is what is expected. Do you have any idea on why this error is happening? Do you think it is ok to commit the version with calling `executor.cancel()` anyway?

anaelle-sw:
> there are 5 GCC warnings, could you address those https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/991/gcc/new/ ?

Sure, sorry I forgot to do this earlier.  Done [here](https://github.com/ros2/rclcpp/pull/1556/commits/3a08a199ab92b67ee24f9985b3160b1e7e93e087)

fujitatomoya:

fujitatomoya:
yeah, i can reproduce the following test failure on my local environment.

```
build/rclcpp/Testing/20210318-2329/Test.xml: 92 tests, 0 errors, 1 failure, 7 skipped
- test_time_source
  <<< failure message
    -- run_test.py: extra environment variables to append:
     - LD_LIBRARY_PATH+=/root/ros2_ws/colcon_ws/build/rclcpp/test/rclcpp
    -- run_test.py: invoking following command in '/root/ros2_ws/colcon_ws/build/rclcpp/test/rclcpp':
     - /root/ros2_ws/colcon_ws/build/rclcpp/test/rclcpp/test_time_source --gtest_output=xml:/root/ros2_ws/colcon_ws/build/rclcpp/test_results/rclcpp/test_time_source.gtest.xml
    Running main() from /root/ros2_ws/colcon_ws/install/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
    [==========] Running 14 tests from 1 test suite.
    [----------] 14 tests from TestTimeSource
    [ RUN      ] TestTimeSource.detachUnattached
    [       OK ] TestTimeSource.detachUnattached (11 ms)
    [ RUN      ] TestTimeSource.reattach
    [       OK ] TestTimeSource.reattach (9 ms)
    [ RUN      ] TestTimeSource.ROS_time_valid_attach_detach
    [       OK ] TestTimeSource.ROS_time_valid_attach_detach (14 ms)
    [ RUN      ] TestTimeSource.ROS_time_valid_wall_time
    [       OK ] TestTimeSource.ROS_time_valid_wall_time (10 ms)
    [ RUN      ] TestTimeSource.ROS_time_valid_sim_time
  >>>
build/rclcpp/test_results/rclcpp/test_time_source.gtest.xml: 1 test, 1 error, 0 failures, 0 skipped
- rclcpp test_time_source.gtest.missing_result
  <<< error message
    The test did not generate a result file.
  >>>

Summary: 1955 tests, 1 error, 1 failure, 14 skipped

```

fujitatomoya:
> Do you have any idea on why this error is happening? Do you think it is ok to commit the version with calling executor.cancel() anyway?

error is happening because clock thread is in

https://github.com/ros2/rclcpp/blob/7bfda87d32bab7e9caa843b2ededa18e427fd00c/rclcpp/include/rclcpp/executor.hpp#L359

after setting `promise.set_value()` to complete the future, it needs to call `clock_executor_.cancel()` to trigger the condition.

```
diff --git a/rclcpp/src/rclcpp/time_source.cpp b/rclcpp/src/rclcpp/time_source.cpp
index 11b396d..9f86428 100644
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -283,6 +283,7 @@ void TimeSource::destroy_clock_sub()
   std::lock_guard<std::mutex> guard(clock_sub_lock_);
   if (clock_executor_thread_.joinable()) {
     cancel_clock_executor_promise_.set_value();
+    clock_executor_.cancel();
     clock_executor_thread_.join();
     clock_executor_.remove_callback_group(clock_callback_group_);
   }
```

anaelle-sw:
> after setting `promise.set_value()` to complete the future, it needs to call `clock_executor_.cancel()` to trigger the condition.

Thanks for the explanation. Last [commit](https://github.com/ros2/rclcpp/pull/1556/commits/5d095d770b49b12728dc7f93bdebe77d64579276) fixes it. Now the tests are not throwing any error on my local setup.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14002/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8824/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11695/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14082/)

anaelle-sw:
Do you have any idea on how to solve this? :
Windows [Build Status](http://ci.ros2.org/job/ci_windows/14082/)

fujitatomoya:
No i am not sure about this. @clalancette @wjwwood any idea?

clalancette:
I think you got unlucky twice.  I kicked it off again:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14119/)

anaelle-sw:
I have re-based the PR branch onto ros2/rclcpp master branch. 
Anything blocking from merging?

fujitatomoya:
CI again (after rebase just in case):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14089/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8907/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11782/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14183/)

fujitatomoya:
@wjwwood @ivanpauno @clalancette i think we could use another review, when you got time!

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -258,6 +258,21 @@ class NodeOptions
...
+
+  /// Return the use_clock_thread flag.
+  RCLCPP_PUBLIC
+  bool
+  use_clock_thread() const;
+
+  /// Set the use_clock_thread flag, return this for parameter idiom.
+  /**
+   * If true, a dedicated thread will be used to subscribe to "/clock"
+   * topic, using a callback group
```
wjwwood, line 270: Maybe?

```suggestion
   * If true, a dedicated thread will be used to subscribe to "/clock"
   * topic, using a callback group and executor dedicated to this task.
```

Seemed like an incomplete thought, and also no punctuation.

wjwwood: Or just

```suggestion
   * If true, a dedicated thread will be used to subscribe to "/clock"
   * topic.
```

Since the details (beyond it being in a thread) maybe don't matter so much?

rclcpp/include/rclcpp/time_source.hpp
```diff
@@ -118,6 +124,11 @@ class TimeSource
   RCLCPP_PUBLIC
   ~TimeSource();
 
+protected:
+  // Dedicated thread for clock subscription.
+  bool use_clock_thread_;
+  std::thread clock_executor_thread_;
```
wjwwood, line 130: Just curious why is this protected while other new things are private? It's not a problem, just curious if it was done intentionally for some reason.

rclcpp/include/rclcpp/node_options.hpp
wjwwood: Also, this should be mentioned in a higher doc block as well, around line 49 of this file.

rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -242,6 +250,24 @@ void TimeSource::create_clock_sub()
...
+      false
+    );
+    options.callback_group = clock_callback_group_;
+    if (!clock_executor_thread_.joinable()) {
+      clock_executor_thread_ = std::thread(
+        [this]() {
+          cancel_clock_executor_promise_ = std::promise<void>{};
+          auto future = cancel_clock_executor_promise_.get_future();
+          clock_executor_.add_callback_group(clock_callback_group_, node_base_);
+          clock_executor_.spin_until_future_complete(future);
```
wjwwood, line 265: I'm wondering if the future is necessary? You already cancel the executor below, so that should interrupt any spinning. You could just do something like this:

```c++
while (clock_not_canceled_) {
  clock_executor_.spin_once();
}

// And in the destroy_clock_sub() method...
clock_not_canceled_ = false;
clock_executor_.cancel();
clock_executor_thread_.join();
// ... and so on
```

It's not much different, and this code is certainly not wrong, but the future seems unnecessary to me.

Just a question, not a request to change it.

rclcpp/test/rclcpp/test_time_source.cpp
```diff
@@ -516,3 +516,229 @@ TEST_F(TestTimeSource, no_pre_jump_callback) {
   EXPECT_EQ(1, cbo.last_postcallback_id_);
   EXPECT_EQ(1, cbo.post_callback_calls_);
 }
+
+// A TimeSource-inheriting class
+// that allows to access to TimeSource protected attributes
```
wjwwood, line 521:
```suggestion
// that allows access to TimeSource protected attributes
```

rclcpp/include/rclcpp/time_source.hpp
wjwwood: Nevermind, I see this is for testing.

wjwwood (commented):
I had some nitpicks and questions, but generally lgtm. Should be good to go with just one round of iteration.

rclcpp/src/rclcpp/time_source.cpp
anaelle-sw: The reason is that the class destructor maybe called before `executor.spin()` (in `create_clock_sub` function). In this case, since `executor.cancel()` only cancels the executor **already** spinning, the executor will be always spinning once `executor.spin()` is reached. 
This [conversation](https://github.com/ros2/rclcpp/pull/1556#discussion_r594604635) explains the issue. 
And [this other one](https://github.com/ros2/rclcpp/pull/1556#discussion_r595632418) explains why the design with`spin_until_future_complete` is better that just using a flag to get executor state.

rclcpp/test/rclcpp/test_time_source.cpp
anaelle-sw: Done in [last commit](https://github.com/ros2/rclcpp/pull/1556/commits/e16e18430eceaa5647a8b9e3e0447ff367e2973e)

rclcpp/include/rclcpp/node_options.hpp
anaelle-sw: Done in [last commit](https://github.com/ros2/rclcpp/pull/1556/commits/e16e18430eceaa5647a8b9e3e0447ff367e2973e)

anaelle-sw:
I have re-based again the PR branch onto master.

anaelle-sw:
I don't undertand why [the tests failed](https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/1052/), as I only changed the some comments since the last time the tests passed.
When I run the tests on my local setup, only `test_parameter_event_handler` (`SameParameterDifferentNode`) returns failure...

wjwwood (approved):

wjwwood:
I'm not sure, they could be flaky. The cppcheck failure is a timeout... so maybe master + your code just put it over the time limit. I can run it again.

fujitatomoya:
CI: (build with `--packages-above-and-dependencies rclcpp` and test with `--packages-above rclcpp`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14109/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8917/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11798/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14205/)

fujitatomoya:
@anaelle-sw 

could you check https://ci.ros2.org/job/ci_windows/14205/msbuild/new/, this one is related to the PR.

Note,
- https://ci.ros2.org/job/ci_osx/11798/clang/new/ is unrelated.
- https://ci.ros2.org/job/ci_windows/14205/cmake/new/ is unrelated.

thanks

anaelle-sw:
> could you check https://ci.ros2.org/job/ci_windows/14205/msbuild/new/, this one is related to the PR.

It should be good now ([last commit](https://github.com/ros2/rclcpp/pull/1556/commits/fdd2390e2d6b2eac9dbd0af157f459e08db7036f))

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14111/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8919/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11800/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14207/)

fujitatomoya:
- https://ci.ros2.org/job/ci_linux/14111/
- https://ci.ros2.org/job/ci_osx/11800/
- https://ci.ros2.org/job/ci_windows/14207/

those are unrelated to the fix, i will merge this into the mainline.

@anaelle-sw thanks for integrating and contribution 👍

jacobperron:
This PR breaks the [domain_bridge](), which does not use a default `rclcpp::Context`. Here's a complete example which worked before this PR and now results in a runtime error (and segfault):

```c++
#include <memory>
#include <rclcpp/rclcpp.hpp>

int main(void) {
  auto context = std::make_shared<rclcpp::Context>();
  context->init(0, nullptr);
  rclcpp::NodeOptions options;
  options.context(context);
  auto node = std::make_shared<rclcpp::Node>("my_node", options);

  return 0;
}
```

Gives the following error:

```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLInvalidArgument'
  what():  context argument is null, at /home/jacob/ws/ros/latest/src/ros2/rcl/rcl/src/rcl/guard_condition.c:65
[1]    541774 abort (core dumped)  ./build/foo/foo
```

rclcpp/include/rclcpp/time_source.hpp
```diff
@@ -140,6 +151,9 @@ class TimeSource
   using SubscriptionT = rclcpp::Subscription<MessageT, Alloc>;
   std::shared_ptr<SubscriptionT> clock_subscription_{nullptr};
   std::mutex clock_sub_lock_;
+  rclcpp::CallbackGroup::SharedPtr clock_callback_group_;
+  rclcpp::executors::SingleThreadedExecutor clock_executor_;
```
jacobperron, line 155: I think this is the bug. The executor is assuming that the default context was initialized (i.e. `rclcpp::init()` is called).

fujitatomoya: thanks, I see, this is regression. it is expecting default context has been initialized instead core dump. Do we want to revert this fix for now? i think yes, since it's been busy for next release (test, stability and so on) , probably we can come back on this after taking care of this issue?

clalancette: Yeah, I'd say let's revert it for now unless we can fix it quickly.

fujitatomoya: i guess that having clock_executor_ shared_ptr and construct the object based on `node_base_` in `TimeSource`. let me try to give it a shot see if we can fix this quickly.

fujitatomoya: https://github.com/ros2/rclcpp/pull/1617 can fix https://github.com/ros2/rclcpp/pull/1556#issuecomment-812256630

hidmic:
@anaelle-sw I suspect this patch introduced the regression that's causing `test_time_source` to timeout every now and then https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/lastCompletedBuild/testReport/projectroot.test/rclcpp/test_time_source/. My gut feeling says it's because [setting a promise from within the clock thread](https://github.com/ros2/rclcpp/pull/1556/files#diff-eb4b7173ae370e64c5fc7ff4532391d514a0fccbc62168e2bb0bf936817e4894R262) is racing with [setting its value](https://github.com/ros2/rclcpp/pull/1556/files#diff-eb4b7173ae370e64c5fc7ff4532391d514a0fccbc62168e2bb0bf936817e4894R285) on destruction.

PR #1568, opened 2021-03-03
Document design decisions that were made for statically typed parameters

ivanpauno:
Document some of the decisions that were made in https://github.com/ros2/rclcpp/pull/1522, https://github.com/ros2/rclpy/pull/683.

There's also a question in the document.

### DIFF
--- a/rclcpp/doc/notes_on_statically_typed_parameters.md
+++ b/rclcpp/doc/notes_on_statically_typed_parameters.md
@@ -106,3 +106,35 @@ node->declare_parameter("my_param", rclcpp::ParameterValue{});  // not valid, wi
 ```py
 node.declare_parameter("my_param", None);  # not valid, will raise error
 ```
+
+## Possible improvements
+
+### Easier way to declare dynamically typed parameters
+
+Declaring a dynamically typed parameter in `rclcpp` could be considered to be a bit verbose:
+
+```cpp
+rcl_interfaces::msg::ParameterDescriptor descriptor;
+descriptor.dynamic_typing = true;
+
+node->declare_parameter(name, rclcpp::ParameterValue{}, descriptor);
+```
+
+the following ways could be supported to make it simpler:
+
+```cpp
+node->declare_parameter(name, rclcpp::PARAMETER_DYNAMIC);
+node->declare_parameter(name, default_value, rclcpp::PARAMETER_DYNAMIC);
+```
+
+or alternatively:
+
+```cpp
+node->declare_parameter(name, default_value, rclcpp::ParameterDescriptor{}.dynamic_typing());
+```
+
+For `rclpy`, there's already a short way to do it:
+
+```py
+node.declare_parameter(name, default_value, rclpy.ParameterDescriptor(dynamic_typing=true));
+```
--- a/rclcpp/doc/notes_on_statically_typed_parameters.md
+++ b/rclcpp/doc/notes_on_statically_typed_parameters.md
@@ -14,7 +14,7 @@ $ ros2 param get /listener use_sim_time
 String value is: not_a_boolean
 ```
 
-For most use cases, having a static parameter types is enough.
+For most use cases, having static parameter types is enough.
 This notes document some of the decisions that were made when implementing static parameter types enforcement in:
 
 * https://github.com/ros2/rclcpp/pull/1522
--- a/rclcpp/doc/notes_on_statically_typed_parameters.md
+++ b/rclcpp/doc/notes_on_statically_typed_parameters.md
@@ -56,7 +56,7 @@ In the current implementation that isn't enforce, but it might make sense to do 
 ## Declaring a parameter without a default value
 
 There might be cases were a default value does not make sense and the user must always provide an override.
-In those cases, the parameter type can be specified explicetly:
+In those cases, the parameter type can be specified explicitly:
 
 ```cpp
 // method signature
--- a/rclcpp/doc/notes_on_statically_typed_parameters.md
+++ b/rclcpp/doc/notes_on_statically_typed_parameters.md
@@ -1,0 +1,111 @@
+# Notes on statically typed parameters
+
+## Introduction
+
+Up to ROS 2 Foxy, all parameters could be changed of type anytime, except the user installed a parameter callback to reject that change.
+This could generate confusing errors, for example:
+
+```
+$ ros2 run demo_nodes_py listener &
+$ ros2 param set /listener use_sim_time not_a_boolean
+[ERROR] [1614712713.233733147] [listener]: use_sim_time parameter set to something besides a bool
+Set parameter successful
+$ ros2 param get /listener use_sim_time
+String value is: not_a_boolean
+```
+
+For most use cases, having a static parameter types is enough.
+This notes document some of the decisions that were made when implementing static parameter types enforcement in:
+
+* https://github.com/ros2/rclcpp/pull/1522
+* https://github.com/ros2/rclpy/pull/683
+
+## Allowing dynamically typed parameters
+
+There might be some scenarios were dynamic typing is desired, so a `dynamic_typing` field was added to the [parameter descriptor](https://github.com/ros2/rcl_interfaces/blob/09b5ed93a733adb9deddc47f9a4a8c6e9f584667/rcl_interfaces/msg/ParameterDescriptor.msg#L25).
+It defaults to `false`.
+
+For example:
+
+```cpp
+rcl_interfaces::msg::ParameterDescriptor descriptor;
+descriptor.dynamic_typing = true;
+
+node->declare_parameter("dynamically_typed_parameter", rclcpp::ParameterValue{}, descriptor);
+```
+
+```py
+rcl_interfaces.msg.ParameterDescriptor descriptor;
+descriptor.dynamic_typing = True;
+
+node.declare_parameter("dynamically_typed_parameter", None, descriptor);
+```
+
+## How is the parameter type specified?
+
+The parameter type will be inferred from the default value provided when declaring it.
+
+## Statically typed parameters when allowing undeclared parameters
+
+When undeclared parameters are allowed, the descriptor of undeclared parameters will have the dynamic typing field set.
+That's because in that case there is no declaration, and it doesn't make much sense to enforce the type of the first value set.
+
+TBD: Should parameters that were declared with the dynamic typing option off enforce that the parameter doesn't change the type even if allow undeclared parameters is set?
+In the current implementation that isn't enforce, but it might make sense to do it for parameters that were declared.
+
+## Declaring a parameter without a default value
+
+There might be cases were a default value does not make sense and the user must always provide an override.
+In those cases, the parameter type can be specified explicetly:
+
+```cpp
+// method signature
+template<typename T>
+Node::declare_parameter<T>(std::string name, rcl_interfaces::msg::ParameterDescriptor = rcl_interfaces::msg::ParameterDescriptor{});
+// or alternatively
+Node::declare_parameter(std::string name, rclcpp::ParameterType type, rcl_interfaces::msg::ParameterDescriptor = rcl_interfaces::msg::ParameterDescriptor{});
+
+// examples
+node->declare_paramter<int64_t>("my_integer_parameter");  // declare an integer parameter
+node->declare_paramter("another_integer_parameter", rclcpp::ParameterType::PARAMETER_INTEGER);  // another way to do the same
+```
+
+```py
+# method signature
+Node.declare_parameter(name: str, param_type: rclpy.Parameter.Type, descriptor: rcl_interfaces.msg.ParameterDescriptor = rcl_interfaces.msg.ParameterDescriptor())
+
+# example
+node.declare_paramter('my_integer_parameter', rclpy.Parameter.Type.INTEGER);  # declare an integer parameter
+```
+
+If the parameter is optional, e.g.: only needed depending on the value of another parameter, the recommended approach is to conditionally declare the parameter:
+
+```cpp
+auto mode = node->declare_parameter("mode", "modeA");  // "mode" parameter is an string
+if (mode == "modeB") {
+    node->declare_parameter<int64_t>("param_needed_when_mode_b");  // when "modeB", the user must provide "param_needed_when_mode_b"
+}
+```
+
+## Other migration notes
+
+Declaring a parameter with only a name is deprecated:
+
+```cpp
+node->declare_parameter("my_param");  // this generates a build warning
+```
+
+```py
+node.declare_parameter("my_param");  # this generates a python user warning
+```
+
+Before, you could initialize a parameter with the "NOT SET" value (in cpp `rclcpp::ParameterValue{}`, in python `None`).
+Now this will throw an exception in both cases:
+
+```cpp
+node->declare_parameter("my_param", rclcpp::ParameterValue{});  // not valid, will throw exception
+```
+
+```py
+node.declare_parameter("my_param", None);  # not valid, will raise error
+```
--- a/rclcpp/doc/notes_on_statically_typed_parameters.md
+++ b/rclcpp/doc/notes_on_statically_typed_parameters.md
@@ -47,7 +47,8 @@ The parameter type will be inferred from the default value provided when declari
 
 ## Statically typed parameters when allowing undeclared parameters
 
-When undeclared parameters are allowed and a parameter is set without a previous declaration, the parameter will be dynamically typed. This is consistent with other allowed behaviors when undeclared parameters are allowed, e.g. trying to get an undeclared parameter returns "NOT_SET".
+When undeclared parameters are allowed and a parameter is set without a previous declaration, the parameter will be dynamically typed.
+This is consistent with other allowed behaviors when undeclared parameters are allowed, e.g. trying to get an undeclared parameter returns "NOT_SET".
 Parameter declarations will remain special and dynamic or static typing will be used based on the parameter descriptor (default to static).
 
 ## Declaring a parameter without a default value
--- a/rclcpp/doc/notes_on_statically_typed_parameters.md
+++ b/rclcpp/doc/notes_on_statically_typed_parameters.md
@@ -47,11 +47,8 @@ The parameter type will be inferred from the default value provided when declari
 
 ## Statically typed parameters when allowing undeclared parameters
 
-When undeclared parameters are allowed, the descriptor of undeclared parameters will have the dynamic typing field set.
-That's because in that case there is no declaration, and it doesn't make much sense to enforce the type of the first value set.
-
-TBD: Should parameters that were declared with the dynamic typing option off enforce that the parameter doesn't change the type even if allow undeclared parameters is set?
-In the current implementation that isn't enforce, but it might make sense to do it for parameters that were declared.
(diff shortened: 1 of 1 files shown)


rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+## How is the parameter type specified?
+
+The parameter type will be inferred from the default value provided when declaring it.
+
+## Statically typed parameters when allowing undeclared parameters
+
+When undeclared parameters are allowed, the descriptor of undeclared parameters will have the dynamic typing field set.
+That's because in that case there is no declaration, and it doesn't make much sense to enforce the type of the first value set.
```
fujitatomoya, line 51: i think this makes sense to me.

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,111 @@
...
+
+TBD: Should parameters that were declared with the dynamic typing option off enforce that the parameter doesn't change the type even if allow undeclared parameters is set?
+In the current implementation that isn't enforce, but it might make sense to do it for parameters that were declared.
+
+## Declaring a parameter without a default value
+
+There might be cases were a default value does not make sense and the user must always provide an override.
+In those cases, the parameter type can be specified explicetly:
```
fujitatomoya, line 59: typo?

```suggestion
In those cases, the parameter type can be specified explicitly:
```

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: yeah, my question is that if you actually made a declaration e.g. `node->declare_parameter("my_int", 5)`, should an static type being enforced even if `allow_undeclared_parameters == true`?

wjwwood: I could be wrong, but when `allow_undeclared_parameters` is true I think `get_parameter` for something that was never set or declared will return `NOT_SET`. If that's the case, that would be a reason to justify this beyond "it doesn't make much sense" which is pretty vague.

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,111 @@
...
+
+The parameter type will be inferred from the default value provided when declaring it.
+
+## Statically typed parameters when allowing undeclared parameters
+
+When undeclared parameters are allowed, the descriptor of undeclared parameters will have the dynamic typing field set.
+That's because in that case there is no declaration, and it doesn't make much sense to enforce the type of the first value set.
+
+TBD: Should parameters that were declared with the dynamic typing option off enforce that the parameter doesn't change the type even if allow undeclared parameters is set?
+In the current implementation that isn't enforce, but it might make sense to do it for parameters that were declared.
```
wjwwood, line 54: I don't feel strongly about this, but even when `allow_undeclared_parameters` is true, declare will set meta data about the parameter and fail if it was declared previously, so I think declare remains special. So I guess it would make sense to make it static unless otherwise specified in the declare call. Put another way, if you only ever call get/set, then making it dynamic makes sense to me. If you call declare, it would make sense to follow what happens when `allow_undeclared_parameters` is false.

I don't feel strongly one way or the other though. As long as it is documented.

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,111 @@
...
+
+```py
+# method signature
+Node.declare_parameter(name: str, param_type: rclpy.Parameter.Type, descriptor: rcl_interfaces.msg.ParameterDescriptor = rcl_interfaces.msg.ParameterDescriptor())
+
+# example
+node.declare_paramter('my_integer_parameter', rclpy.Parameter.Type.INTEGER);  # declare an integer parameter
+```
+
+If the parameter is optional, e.g.: only needed depending on the value of another parameter, the recommended approach is to conditionally declare the parameter:
```
wjwwood, line 81: I know this was discussed a lot already, and implemented, so I'm just weighing my opinion here I don't mean to say it needs to change, but I don't care for how this make "requires override" implicit. I would have much rather preferred an explicit setting for "requires override" and if no type is given making the parameter dynamic and initialized to `NOT_SET`.

In the case that the user provides a name and a type but no default, then I'd say it should either be an error or it should default to the equivalent of 0, e.g. `0`, `0.0`, `""`, `false`, etc..

What I don't care for is, if a user asks "how do I make a parameter require an external override?" and the answer is "declare it by name and type but don't give a default value" then my question as a user would be "why?" and the answer depends on you understand quite a lot about how the parameter system works. I just seems like neat trick rather than a more intentional answer like "mark it as requires override". Also when reading code, it's hard to pick out the declare call (potentially amongst many) that omits a default value, whereas an explicit setting could be seen more easily and searched for.

That's just my personal prespective. But I trust you guys discussed it and had reason for coming to this solution.

That being said, this workaround isn't bad.

wjwwood: Also...

>  If the parameter is optional

We should be careful, I think, to be specific in our terminology.

The logical "If..." following this section (in my opinion) is "If the parameter override is optional ..." (because in the examples above we're making the parameter override required).

However, this case you've described here also makes sense, I think, but I would more accurately describe it as "If the parameter may be unused, but when used requires a parameter override, then you could ..."

It's a subtle difference, but I think saying that the "parameter is optional" is inaccurate, imo it is being conditionally used, which is different from being optional. Optional implies it could be required (the opposite) and _parameters_ cannot be required, only parameter overrides can be.

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,111 @@
...
+```cpp
+node->declare_parameter("my_param");  // this generates a build warning
+```
+
+```py
+node.declare_parameter("my_param");  # this generates a python user warning
+```
+
+Before, you could initialize a parameter with the "NOT SET" value (in cpp `rclcpp::ParameterValue{}`, in python `None`).
+Now this will throw an exception in both cases:
```
wjwwood, line 103: Again, I wasn't able to follow the whole discussion, but why can `None`/default value not be a stand in for "dynamically typed and initialized to None"?

I mean it's not the end of the world because they can say "declare as dynamically typed and default to NOT_SET" explicitly, but that just a bit tedious.

Again, not trying to say it should change, but usually in the "Alternatives" section of a REP you don't just say what you decided to do, but also why you decided that (summary of the development discussions), what you didn't choose, and why you didn't choose them. That could be helpful here too.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: > I could be wrong, but when allow_undeclared_parameters is true I think get_parameter for something that was never set or declared will return NOT_SET. If that's the case, that would be a reason to justify this beyond "it doesn't make much sense" which is pretty vague.

That sounds fine to me

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: > Put another way, if you only ever call get/set, then making it dynamic makes sense to me. If you call declare, it would make sense to follow what happens when allow_undeclared_parameters is false.

:+1:
I will document that here and meanwhile fix the behavior in `rclcpp` and `rclpy`.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: > I know this was discussed a lot already, and implemented, so I'm just weighing my opinion here I don't mean to say it needs to change

I should have written this before working on the implementation, so I don't mind doing any modification.

> I would have much rather preferred an explicit setting for "requires override" and if no type is given making the parameter dynamic and initialized to NOT_SET

Do you mean that `node->declare_parameter("my_param")` should declare a dynamic parameter initialized to "NOT_SET"? It's doing that, but deprecated, so maybe we can un-deprecate it (?).

I deprecated that one to force people to think "did I really want a dynamically typed parameter?", but maybe that was too much.

> In the case that the user provides a name and a type but no default, then I'd say it should either be an error or it should default to the equivalent of 0, e.g. 0, 0.0, "", false, etc..
>
> What I don't care for is, if a user asks "how do I make a parameter require an external override?" and the answer is "declare it by name and type but don't give a default value" then my question as a user would be "why?" and the answer depends on you understand quite a lot about how the parameter system works. I just seems like neat trick rather than a more intentional answer like "mark it as requires override". Also when reading code, it's hard to pick out the declare call (potentially amongst many) that omits a default value, whereas an explicit setting could be seen more easily and searched for.

`node->declare_parameter(name, parameter_type)` was only added for the case you want a statically typed parameter and don't want to have a default value, which seems like a valid use case.
If I would want a parameter defaulted to `0`, `0.0`, etc; I would rather use `node->declare_parameter(name, 0)`.
I agree that maybe it's not clear that an override is required, and it might be found confusing.

I'm not sure how to make this less confusing, maybe the problem is naming (?).
I'm open to suggestions.

> However, this case you've described here also makes sense, I think, but I would more accurately describe it as "If the parameter may be unused, but when used requires a parameter override, then you could ..."

That suggestion sounds fine to me

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: > Again, I wasn't able to follow the whole discussion, but why can None/default value not be a stand in for "dynamically typed and initialized to None"?

I think I made this change to make people think if they really wanted a dynamically typed parameter, but not throwing an exception and instead making it dynamically typed also sounds reasonable to me (and maybe it will avoid breaking downstream packages).

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: > However, this case you've described here also makes sense, I think, but I would more accurately describe it as "If the parameter may be unused, but when used requires a parameter override, then you could ..."

Addressed in https://github.com/ros2/rclcpp/pull/1568/commits/f245d42a871e4f33bc075cfd9f7d8994c05476c8.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: Fixed in https://github.com/ros2/rclcpp/pull/1568/commits/f245d42a871e4f33bc075cfd9f7d8994c05476c8.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: Addressed in https://github.com/ros2/rclcpp/pull/1568/commits/f245d42a871e4f33bc075cfd9f7d8994c05476c8.

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,108 @@
+# Notes on statically typed parameters
+
+## Introduction
+
+Up to ROS 2 Foxy, all parameters could be changed of type anytime, except the user installed a parameter callback to reject that change.
```
jacobperron, line 5:
```suggestion
Until ROS 2 Foxy, all parameters could change type anytime, except if the user installed a parameter callback to reject a change.
```

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,108 @@
...
+```
+$ ros2 run demo_nodes_py listener &
+$ ros2 param set /listener use_sim_time not_a_boolean
+[ERROR] [1614712713.233733147] [listener]: use_sim_time parameter set to something besides a bool
+Set parameter successful
+$ ros2 param get /listener use_sim_time
+String value is: not_a_boolean
+```
+
+For most use cases, having a static parameter types is enough.
```
jacobperron, line 17:
```suggestion
For most use cases, having static parameter types is enough.
```

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,108 @@
...
+This notes document some of the decisions that were made when implementing static parameter types enforcement in:
```
jacobperron, line 18:
```suggestion
This article documents some of the decisions that were made when implementing static parameter types enforcement in:
```

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,108 @@
...
+
+* https://github.com/ros2/rclcpp/pull/1522
+* https://github.com/ros2/rclpy/pull/683
+
+## Allowing dynamically typed parameters
+
+There might be some scenarios were dynamic typing is desired, so a `dynamic_typing` field was added to the [parameter descriptor](https://github.com/ros2/rcl_interfaces/blob/09b5ed93a733adb9deddc47f9a4a8c6e9f584667/rcl_interfaces/msg/ParameterDescriptor.msg#L25).
```
jacobperron, line 25:
```suggestion
There might be some scenarios where dynamic typing is desired, so a `dynamic_typing` field was added to the [parameter descriptor](https://github.com/ros2/rcl_interfaces/blob/09b5ed93a733adb9deddc47f9a4a8c6e9f584667/rcl_interfaces/msg/ParameterDescriptor.msg#L25).
```

rclcpp/doc/notes_on_statically_typed_parameters.md
jacobperron: Deprecating this particular signature makes sense to me, since I think we want to encourage people to use statically typed parameters by default. Maybe we should document this rationale here.

rclcpp/doc/notes_on_statically_typed_parameters.md
wjwwood: @ivanpauno what happens if you do this:

```c++
// allow_undeclared_parameters = true
node->set_parameter("foo", ...);
node->declare_parameter("foo", ...);
```

I guess the `declare_parameter` fails because it is already declared? I guess that's fine right?

ivanpauno: > I guess the declare_parameter fails because it is already declared?

I guess the same, but I will double check.

>  I guess that's fine right?

Sounds fine to me, I think that was the behavior before these last changes to enforce static parameter types.

rclcpp/doc/notes_on_statically_typed_parameters.md
wjwwood: > Do you mean that node->declare_parameter("my_param") should declare a dynamic parameter initialized to "NOT_SET"? It's doing that, but deprecated, so maybe we can un-deprecate it (?).

> I deprecated that one to force people to think "did I really want a dynamically typed parameter?", but maybe that was too much.

I think that's what I meant, yes. I don't feel strongly about it. I think that our users will probably want something convenient, but I'm ok with them having to opt into dynamic typing personally.

For the other half, i.e. being explicit when asking for a required override, I think that's more important, because it's hard to know, looking at the source whether or not a given parameter is dynamic (unless you know what you're looking for).

> node->declare_parameter(name, parameter_type) was only added for the case you want a statically typed parameter and don't want to have a default value, which seems like a valid use case.

So are there two use cases here?

- "you want a statically typed parameter and don't want to have a default value", and
- you want to require an override

I guess it makes sense that if you require an override a default value is meaningless. So they are related, but I wonder if the first ever makes sense? Why would you do that unless you wanted to require an override? I guess you wouldn't which leads you to the current behavior which is to say "if you do the first item above, it implies the second".

I dunno, I'm just typing-out-loud here :)

Maybe having side by side examples of each case would help, do we have documentation like this or an example where it's like "and this is how you do X with parameters, and this is how you do Y with parameters, etc..." basically like a parameters howto list or "cookbook"?

rclcpp/doc/notes_on_statically_typed_parameters.md
wjwwood: I'm fine with deprecating and then throwing, and therefore requiring explicitly opting into a dynamic parameter.

As I said in the other thread, I just think (based on past experience) people will want the more convenient way of getting dynamically typed parameters.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: > I guess the same, but I will double check.

https://github.com/ros2/rclcpp/pull/1575 has added test for that case.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: > As I said in the other thread, I just think (based on past experience) people will want the more convenient way of getting dynamically typed parameters.

I think that could be added, something like:

```cpp
node->declare_parameter("parameter_name", rclcpp::PARAMETER_DYNAMIC);
node->declare_parameter("parameter_name", default_value, rclcpp::PARAMETER_DYNAMIC);
```

or maybe something to streamline the parameter descriptor construction:

```cpp
node->declare_parameter("parameter_name", default_value, rclcpp::ParameterDescriptor{}.dynamic_typing());
```

wjwwood: Those look nicer.

ivanpauno: I've documented that as a possible improvement, see https://github.com/ros2/rclcpp/pull/1568/commits/46a773e5640648b7ffae96824233ea6f1c13e415

rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -0,0 +1,140 @@
...
+node.declare_parameter("dynamically_typed_parameter", None, descriptor);
+```
+
+## How is the parameter type specified?
+
+The parameter type will be inferred from the default value provided when declaring it.
+
+## Statically typed parameters when allowing undeclared parameters
+
+When undeclared parameters are allowed and a parameter is set without a previous declaration, the parameter will be dynamically typed. This is consistent with other allowed behaviors when undeclared parameters are allowed, e.g. trying to get an undeclared parameter returns "NOT_SET".
```
jacobperron, line 50: nit: one sentence per line

jacobperron (approved):
LGTM, thanks for documenting :bow:

jacobperron:
Please also add a release note about this for Galactic, perhaps referring to this document.

rclcpp/doc/notes_on_statically_typed_parameters.md
wjwwood: > So the other option is to allow "uninitialized statically typed parameters" which weren't supported because I didn't see their usefulness, though I might have overlooked it.

I don't think we need that, but again, I don't feel strongly about it.

> IMO, it should be the "required type", even if the parameter is uninitialized at the moment.

I agree with that.

wjwwood (approved):
I think all of my feedback was addressed.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno:
```suggestion
When undeclared parameters are allowed and a parameter is set without a previous declaration, the parameter will be dynamically typed.
This is consistent with other allowed behaviors when undeclared parameters are allowed, e.g. trying to get an undeclared parameter returns "NOT_SET".
```

ivanpauno:
Thanks for the reviews @jacobperron @wjwwood !!!

PR #1571, opened 2021-03-04
Quiet clang memory leak warning on "DoNotOptimize".

clalancette:
When building rclcpp under clang static analysis, it complains
that the "DoNotOptimize" function from Google benchmark can
cause a memory leak.  I can't see how this is possible, so I
can only assume that the inline assembly that is used to implement
"DoNotOptimize" is causing a false positive.  Just quite the
warning here when building under clang static analysis.


wjwwood (approved):

clalancette:
CI (only up to rclcpp):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13809/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8675/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11518/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13896/)

PR #1574, opened 2021-03-05
Add support for rmw_connextdds

asorbini:
This PR removes all references to `rmw_connext_cpp`, so that it may be replaced by `rmw_connextdds`.

Changes mostly involved getting rid of several comments about "long running tests", and re-enabling a test in `test_allocator_memory_strategy.cpp`.

The PR also includes a fix for `test_multi_threaded_executor.cpp`, to avoid an implicit conversion from integer to double which was causing the test to fail.

See [rticommunity/rmw_connextdds #9](https://github.com/rticommunity/rmw_connextdds/issues/9) for a list of related PRs, and an overview of all the changes required to replace [ros2/rmw_connext](https://github.com/ros2/rmw_connext) (`rmw_connext_cpp`) with [rticommunity/rmw_connextdds](https://github.com/rticommunity/rmw_connextdds) in the ROS2 source tree.

### DIFF
--- a/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
+++ b/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
@@ -86,10 +86,13 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
         double diff = static_cast<double>(std::abs((now - last).nanoseconds())) / 1.0e9;
         last = now;
 
-        if (diff < PERIOD - TOLERANCE) {
+        double diff_exp =
+            static_cast<double>(PERIOD) - static_cast<double>(TOLERANCE);
+        if (diff < diff_exp) {
           executor.cancel();
-          ASSERT_GT(diff, PERIOD - TOLERANCE);
+          ASSERT_GT(diff, diff_exp);
         }
+
       }
     };
 
--- a/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
+++ b/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
@@ -87,7 +87,7 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
         last = now;
 
         double diff_exp =
-            static_cast<double>(PERIOD) - static_cast<double>(TOLERANCE);
+          static_cast<double>(PERIOD) - static_cast<double>(TOLERANCE);
         if (diff < diff_exp) {
           executor.cancel();
           ASSERT_GT(diff, diff_exp);
--- a/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
+++ b/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
@@ -86,11 +86,9 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
         double diff = static_cast<double>(std::abs((now - last).nanoseconds())) / 1.0e9;
         last = now;
 
-        double diff_exp =
-          static_cast<double>(PERIOD) - static_cast<double>(TOLERANCE);
-        if (diff < diff_exp) {
+        if (diff < PERIOD - TOLERANCE) {
           executor.cancel();
-          ASSERT_GT(diff, diff_exp);
+          ASSERT_GT(diff, PERIOD - TOLERANCE);
         }
       }
     };
--- a/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
+++ b/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
@@ -92,7 +92,6 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
           executor.cancel();
           ASSERT_GT(diff, diff_exp);
         }
-
       }
     };
 
--- a/rclcpp/test/rclcpp/strategies/test_allocator_memory_strategy.cpp
+++ b/rclcpp/test/rclcpp/strategies/test_allocator_memory_strategy.cpp
@@ -511,7 +511,10 @@ TEST_F(TestAllocatorMemoryStrategy, number_of_entities_with_subscription) {
   RclWaitSetSizes expected_sizes = {};
   expected_sizes.size_of_subscriptions = 1;
   const std::string implementation_identifier = rmw_get_implementation_identifier();
-  if (implementation_identifier == "rmw_cyclonedds_cpp") {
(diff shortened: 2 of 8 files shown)


rclcpp/test/rclcpp/strategies/test_allocator_memory_strategy.cpp
```diff
@@ -511,9 +511,7 @@ TEST_F(TestAllocatorMemoryStrategy, number_of_entities_with_subscription) {
   RclWaitSetSizes expected_sizes = {};
   expected_sizes.size_of_subscriptions = 1;
   const std::string implementation_identifier = rmw_get_implementation_identifier();
-  if (implementation_identifier == "rmw_connext_cpp" ||
-    implementation_identifier == "rmw_cyclonedds_cpp")
-  {
+  if (implementation_identifier == "rmw_cyclonedds_cpp") {
     // For connext, a subscription will also add an event and waitable
```
clalancette, line 515:
```suggestion
    // For cyclonedds, a subscription will also add an event and waitable
```

clalancette (commented):
One minor thing to fix in here.

rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
```diff
@@ -86,9 +86,11 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
...
         last = now;
 
-        if (diff < PERIOD - TOLERANCE) {
+        double diff_exp =
+          static_cast<double>(PERIOD) - static_cast<double>(TOLERANCE);
+        if (diff < diff_exp) {
           executor.cancel();
-          ASSERT_GT(diff, PERIOD - TOLERANCE);
+          ASSERT_GT(diff, diff_exp);
         }
```
fujitatomoya, line 94: although this is okay to change, but may i ask why? did we have an problem? as far as i see, PERIOD and TOLERANCE are declared with `constexpr double`, which will be used as double at runtime in this case.

ivanpauno: +1
The changes to this file doesn't make sense to me

rclcpp/test/rclcpp/strategies/test_allocator_memory_strategy.cpp
```diff
@@ -511,9 +511,7 @@ TEST_F(TestAllocatorMemoryStrategy, number_of_entities_with_subscription) {
   RclWaitSetSizes expected_sizes = {};
   expected_sizes.size_of_subscriptions = 1;
   const std::string implementation_identifier = rmw_get_implementation_identifier();
-  if (implementation_identifier == "rmw_connext_cpp" ||
-    implementation_identifier == "rmw_cyclonedds_cpp")
-  {
+  if (implementation_identifier == "rmw_cyclonedds_cpp") {
```
ivanpauno, line 514: IMO, keeping `rmw_connext_cpp` here doesn't make any harm.
Not removing support for `rmw_connext_cpp` immediately will make the transition easier IMO

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
```diff
@@ -30,10 +30,6 @@
 #include "rclcpp/executor.hpp"
 #include "rclcpp/rclcpp.hpp"
 
-// Note: This is a long running test with rmw_connext_cpp, if you change this file, please check
-// that this test can complete fully, or adjust the timeout as necessary.
-// See https://github.com/ros2/rmw_connext/issues/325 for resolution]
```
ivanpauno, line 35: does the test run faster with the new `rmw_connextdds` implementation?

rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
asorbini: You are both absolutely right: this change doesn't make any sense :) I chalk it all up to a little burnout and a great desire to see all tests passing. I made this change when the test failed on a local test machine I was using, and convinced myself I had "fixed it" when instead it was a flaky test on that machine, and it just happened to pass after I made the change (but later failed during another run over the weekend).

I'll revert the change. I have also never seen this test fail on the official build machines so I'm leaning towards blaming the failure on the test machine being less powerful than then one on ci.ros2.org (and thus needing some greater tolerance and/or period). I haven't validated this hypothesis though, but I can dig further.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
asorbini: I think the "long running" aspect was caused by the long time required, by default, to shut down a DomainParticipant. For `rmw_connextdds`, I followed the advice on the [linked thread](https://github.com/ros2/rmw_connext/issues/325), and reduced the shutdown period to 50ms like it was done for [rmw_connext_cpp](https://github.com/ros2/rmw_connext/pull/473). In fact, I think these comments might not apply to `rmw_connext_cpp` either, since that change. 

Just to make sure, I tried to manually run the test with `rmw_connextdds` and `rmw_cyclonedds_cpp`, and they ran in 20 and 12 seconds respectively. Connext certainly has some noticeable finalization overhead, but I think the two times can be considered "in the same ballpark".

ivanpauno: Makes sense, thanks for explaining!

rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
asorbini: Changes reverted by [84c9fcd](https://github.com/ros2/rclcpp/pull/1574/commits/84c9fcd605a9d4a8a0e4314a85aeeb8c77dec31d).

asorbini:
@ivanpauno I agree that it would be harmless to keep exceptions for `rmw_connext_cpp`.

The same is true for other PRs related to this one which made changes to tests:

 - [ros2/system_tests#463](https://github.com/ros2/system_tests/pull/463)
 - [ros2/rosbag2#671](https://github.com/ros2/rosbag2/pull/671)
 - [ros2/rclpy#698](https://github.com/ros2/rclpy/pull/698)
 - [ros2/rcl#895](https://github.com/ros2/rcl/pull/895)).
 
If we decide to go with keeping support for `rmw_connext_cpp` along side `rmw_connextdds` to allow for an incremental transition period, then I can update all the PRs to restore any exception for `rmw_connext_cpp` (and `rmw_connext_dynamic_cpp`) that I had removed/replaced.

clalancette (approved):

Issue #491, opened 2018-06-05
Add the ability to have asynchronous service callbacks


clalancette:
## Feature request

Add in an API to do asynchronous service callbacks for service servers.

#### Feature description

It is currently possible in ROS2 to have a service callback in a thread by using separate callback groups and using a multi-threaded executor.  However, it would probably be nice to allow for an API where the user can specify that their service is asynchronous.  During the initial call to the service, the service would kick off a thread (or whatever else is needed to do the work), then return some kind of promise.  When the service eventually completes, it would signal the client that it has completed the promise.  I'm not entirely sure what the API for this would look like currently, but @wjwwood may have some ideas here.

mkhansen-intel:
What is the status of this feature request. I'm looking at how to do this for implementing Actions, and don't want to duplicate.

wjwwood:
> What is the status of this feature request.

Nothing has been done about this feature request as far as I'm aware.

Basically we just need a Service server callback that doesn't require that you return the response from the callback and instead allows you to send the response when ever and from where ever you want.

mkhansen-intel:
I'm not sure where to ask this, here or on Discourse; would it be possible to add another callback group type that is "concurrent non-reentrant"? Right now there are only "MutuallyExclusive" and "Reentrant" types. It would be nice to be able to have multiple callbacks in the same group able to execute concurrently, but not pre-empt themselves (non-reentrant).

wjwwood:
> "concurrent non-reentrant"

Perhaps the name should be changed, but I don't see a need for anything other than the two callback groups we have at the moment.

If you consider two timers, A and B, then you can make a sort of table:

|                        | Mutually Exclusive                                                      | Reentrant                                                         |
|------------------------|-------------------------------------------------------------------------|--------------------------------------------------------------------|
| A and B in same CG     | A and B will not be called at the same time as each other or themselves | A and B may be called at the same time as each other or themselves |
| A and B in separate CG | A and B will be called concurrently with each other but not themselves  | A and B may be called at the same time as each other or themselves |

The only case not served by the current design that I'm aware of is "A and B may be called at the same time as themselves, but not each other", which was deemed uncommon enough that we didn't need to support it. We'd either have to allow nesting callback groups (explicitly decided not to do that originally) or make some other change if we wanted to support that, but I don't think we need to do so.

[...]

PR #1576, opened 2021-03-09
basic async service

rpaaron:
#491 

naive prevention of `handle_request` from calling `send_response`.  Instead let the client do it, when the result is available.

### DIFF
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -30,6 +30,9 @@ namespace rclcpp
 {
 
 template<typename ServiceT>
+class Service;
+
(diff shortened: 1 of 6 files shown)


rpaaron:
example client would look something like:
```c++
auto service = node.create_service_async<AddTwoInts>(
    "/add_two_ints",
    [&](const std::shared_ptr<rmw_request_id_t> request_header,
        const std::shared_ptr<AddTwoInts::Request> request,
        const std::shared_ptr<AddTwoInts::Response> response)
    {
        std::thread t([=]() {
            /* something that takes a long time */
            response->sum = request->a + request->b;
            service->send_response(request, response);
        });
        t.join();
    }
});

PR #1579, opened 2021-03-12
Add RMW listener APIs

irobot-ros:
This PR replaces https://github.com/ros2/rclcpp/pull/1416 and contains the changes necessary in order to add support for RMW listener APIs in rclcpp.
The events executor will be re-proposed as a follow-up PR

@mauropasse @wjwwood @ivanpauno


irobot-ros (commented):
@wjwwood Thank you for the updates, I really like the new style and naming however I have some doubts for what concerns `Waitable`s.

I think I understand why you removed the `set_listener_callback` API from the base Waitable class.
A callback should be entirely specified by the context calling the set-callback function and the rclcpp entity should be responsible only of forwarding it as-is to the underlying RMW entities.
This means that there is no space for a generic "set callback" function in the base Waitable, but rather that the derived-from-Waitable classes should define all the specific APIs they need.
And in theory I definitely agree with this approach.

However, looking at the more practical side, this is not how rclcpp currently works.
I think that the majority of the users wouldn't really care to set their own callbacks if there was an efficient implementation of the executor class.
But an executor has only access to callback groups and can grab only the base waitable classes.

What's your idea for implementing this?
I don't really see a solution that does not involve exposing all the individual entities that make up a node

wjwwood:
> I think I understand why you removed the set_listener_callback API from the base Waitable class.
A callback should be entirely specified by the context calling the set-callback function and the rclcpp entity should be responsible only of forwarding it as-is to the underlying RMW entities.
This means that there is no space for a generic "set callback" function in the base Waitable, but rather that the derived-from-Waitable classes should define all the specific APIs they need.
And in theory I definitely agree with this approach.

This was my thinking. I didn't think a base virtual class would be useful because a waitable can be made up of multiple entities and so a single callback doesn't really make sense. Like what happens if it is a subscription and qos event? If you had to generalize that to a single callback, would you combine the events or provide two arguments? How would an executor know for each waitable, how many sources of callbacks there could be?

But you're right that we need something in the base waitable class that we can use generically. I'll keep thinking about it.

rclcpp/src/rclcpp/waitable.cpp
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stdexcept>
```
audrow, line 15: Probably not needed.

audrow (approved):

alsora:
> 
> Like what happens if it is a subscription and qos event? If you had to generalize that to a single callback, would you combine the events or provide two arguments? How would an executor know for each waitable, how many sources of callbacks there could be?

We have been looking into this in order to implement support for actions.

I think that the executor shouldn't know about how many entities a Waitable is made of. 

The Waitable could provide a generic `set_on_event_callback()` API. Calling this could result in setting the same user-provided-callback to all the eentities the Waitable is made of, however a different token should be associated to each of them, as if we were binding it as an extra argument. 

So, from the executor point of view there would simply be an event for Waitable X. 
By inspecting the event, the Waitable should be able to retrieve the token and understand which of its entities generated the event.

wjwwood:
> So, from the executor point of view there would simply be an event for Waitable X.
By inspecting the event, the Waitable should be able to retrieve the token and understand which of its entities generated the event.

Ok, that makes sense to me, but how will it know which entity within the waitable generated the event for each call to the callback? Or is that not necessary?

Like I imagine something like this:

- Waitable has a subscription and a QoS event
- Waitable receives message, `on_event` callback called, executor adds entry in event queue
- event queue: [WaitableEvent]
- Waitable receives QoS event, `on_event` callback called, executor adds entry in event queue
- event queue: [WaitableEvent, WaitableEvent]
- executor pops off the event queue, looks to Waitable, how does the waitable know to do a subscription or qos event (both available)? Or does that not matter?
- executor pops off the event queue, looks to Waitable, it executes the remaining activity (which ever one was not done last time)

So, it seems like you don't have enough information in a generic (virtual method on Waitable base class) callback to properly inject things into the event queue? But maybe this was a problem before? I don't know...

----

Either way, we agree there needs to be some kind of callback for Waitable that lets the event executor be notified when something in the Waitable is ready. Should that callback just be `std::function<void ()>`?

----

Another question I had last night but forgot to mention, what is the relationship between these callbacks and using items in wait sets? Can you set this callback and use these items in a wait set like normal at the same time? If not, should we enforce that?

alsora:
> Ok, that makes sense to me, but how will it know which entity within the waitable generated the event for each call to the callback? Or is that not necessary?

We implemented this by having the Waitable pass down to the RMW entity two separate objects:
 - a callback function pointer
 - a void pointer to "user data" 
 - a void pointer to "entity data" 

The key part is that while the "user data" would be specified by the application, the entity data would be specified by the Waitable itself.

Considering the Events Executor and a Waitable with a QoS Event and a Subscription, those elements would be:
```
user_data = {
   executor_ptr, // this is a pointer to the executor because we used a static member function as callback where the ptr would be casted to "this".
   EVENT_TYPE_WAITABLE, // Enum to tell executor that this event has  been generated by a Waitable
   waitable_id // Id of the particular waitable that generated the event, e.g. to retrieve it inside the entities collector
};

entity_data = {
  entity_id // id of the particular entity within the waitable that generated the event (i.e. QoS or Subscription)
};
```

The executor wouldn't care about the entity_data (it's just a void pointer from its point of view and it does not know how to cast it) and it would just forward it to the Waitable `execute()` method.
The solution is definitely not a nice C++ implementation, but in the end I think this is sort of what  we need.

> 
> Another question I had last night but forgot to mention, what is the relationship between these callbacks and using items in wait sets? Can you set this callback and use these items in a wait set like normal at the same time? If not, should we enforce that?

In the RMW layer, invoking the listener callback is an additional operation beside standard waitset usage.
However, I think that in the rclcpp layer we should enforce that  you can only use either one or the other, as otherwise that would cause problems similar to have an entity in multiple executors.

mauropasse:
> We implemented this by having the Waitable pass down to the RMW entity the following, separate, objects:
> 
> a callback function pointer
> a void pointer to "user data"
> a void pointer to "entity data"

Just to clarify, for actions support we pass down to the RMW just the `callback` and `user_data`, where `user_data` is a struct containing `entity_gen_id` (the entity which generated the event, like a service part of an action) and `execution_id` (the entity which will execute the event, like the Action waitable).
Then, when the event is executed, it'll be called `action_waitable->execute(entity_gen_id)` so the action waitable will know who generated the event and how to execute, based on it.

rclcpp/src/rclcpp/qos_event.cpp
```diff
@@ -69,17 +69,18 @@ QOSEventHandlerBase::is_ready(rcl_wait_set_t * wait_set)
...
-  rcl_ret_t ret = rcl_event_set_listener_callback(
+  rcl_ret_t ret = rcl_event_set_callback(
     &event_handle_,
     callback,
     user_data);
 
   if (RCL_RET_OK != ret) {
-    throw std::runtime_error("Couldn't set listener callback to QOSEventHandlerBase");
+    using rclcpp::exceptions::throw_from_rcl_error;
+    throw_from_rcl_error(ret, "failed to set the on new message callback for subscription");
```
mauropasse, line 83: wrong error message?

mauropasse: Our approach for un-setting the callbacks was passing `callback` and `user_data` as `nullptr`, where the RMW started counting the `unread_events` again, as when the callback is still not set like before adding a node to the executor. 
Now we get an error here,  where `rcl_event_set_callback` detects it is a `nullptr`, when the app is being finalized.
If passing `nullptr` is an error, we should have another way to "unset" the callbacks.

wjwwood: Ok, I'll look at that.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -171,13 +229,20 @@ class ClientBase
   const rcl_node_t *
   get_rcl_node_handle() const;
 
+  RCLCPP_PUBLIC
+  void
+  set_on_new_response_callback(rcl_event_callback_t callback, const void * user_data);
```
mauropasse, line 234: why is `set_on_new_response_callback` (and the other new ones) protected?

wjwwood: Because the `set_on_new_response_callback(std::function<void(size_t)> callback)` signature is the public one?

Does it need to be public?

fujitatomoya (approved):
overall, looks good to me.

irobot-ros:
@wjwwood I put up a short proposal for handling waitables,
https://gist.github.com/irobot-ros/e99c81b0ca2ade2b24016ef64aa2d0b0

rclcpp/include/rclcpp/client.hpp
```diff
@@ -150,6 +151,69 @@ class ClientBase
...
+   * callback was set.
+   *
+   * Since this callback is called from the middleware, you should aim to make
+   * it fast and not blocking.
+   * If you need to do a lot of work or wait for some other event, you should
+   * spin it off to another thread, otherwise you risk blocking the middleware.
+   *
+   * Calling it again will clear any previously set callback.
+   *
+   * This function is thread-safe.
```
irobot-ros, line 168: @wjwwood I was implementing a way to clear the previously set callbacks and I noticed that the `set_*` functions are not thread safe as mentioned here.

If two threads call `set_on_new_response_callback` concurrently, it may happen that one thread modifies the permanent storage `on_new_response_callback_` while this is being used by the RMW layer.
For example it could happen with the following execution order:
 - thread 1 `set_on_new_response_callback(trampoline, new_callback)`
 - thread 2  `set_on_new_response_callback(trampoline, new_callback)`
 - thread 1 `on_new_response_callback_ = new_callback;`
 - thread 1 `set_on_new_response_callback(trampoline, on_new_response_callback_)` and return 
  - thread 2 `on_new_response_callback_ = new_callback;` ---> this is where problems may arise

wjwwood: You're right. We'll have to add a mutex to protect the `on_new_response_callback_` sequence. I wrote the documentation before I changed that code in order to resolve the issue that changing `on_new_response_callback_` would make the old value invalid while it is still set as the callback, allowing for a small window (before ` set_on_new_response_callback` is called) where rcl/rmw could call an invalid callback.

irobot-ros: Ok, I;m going to add some commits, I will take care of that

irobot-ros:
Last commits address the following items:
 - restore listener APIs for waitables (with support for future waitables made of multiple entities, e.g. actions)
 - add mutex to make the new APIs thread safe
 - add APIs to publisher and subscription to allow users to set listeners for the waitables contained in those entities (e.g. ipc subscription, qos handlers)
 - add unit tests
 - resolved conflicts

irobot-ros:
Hi, any update on this PR review?
Besides being blocked by https://github.com/eclipse-cyclonedds/cyclonedds/pull/699, can we consider the work here completed?

wjwwood:
I need to do another review, but also at the moment this cannot be merged due to merge conflicts, which will need to be resolved.

wjwwood:
> Besides being blocked by eclipse-cyclonedds/cyclonedds#699

We will need to ensure this is merged before or a the same time as this pull request to avoid introducing failing tests.

alsora:
@wjwwood friendly ping.
Can we please proceed with these PRs?

alsora:
@wjwwood @ivanpauno can you please trigger a CI build including all the related PRs?
rcl: https://github.com/ros2/rcl/pull/839
rmw: https://github.com/ros2/rmw/pull/286
rmw_connextdds: https://github.com/ros2/rmw_connextdds/pull/44
rmw_cyclonedds: https://github.com/ros2/rmw_cyclonedds/pull/256
rmw_fastrpts: https://github.com/ros2/rmw_fastrtps/pull/468
rmw_implementation: https://github.com/ros2/rmw_implementation/pull/161


at least we can make sure that there are no failures when someone will ever review it

alsora:
I resolved conflicts.
Can someone please trigger a full CI build?

alsora:
@wjwwood @ivanpauno @fujitatomoya friendly ping for running a CI build including all the related PRs.
Is there anything else we should do? Assuming we get green CI can we finally merge this work?

rcl: ros2/rcl#839
rmw: ros2/rmw#286
rmw_connextdds: ros2/rmw_connextdds#44
rmw_cyclonedds: ros2/rmw_cyclonedds#256
rmw_fastrpts: ros2/rmw_fastrtps#468
rmw_implementation: ros2/rmw_implementation#161

alsora:
@wjwwood  thanks for addressing the conflicts.
Is there any test/anything else we should address before proceeding with this PR?

alsora:
CI using this repos file https://gist.githubusercontent.com/irobot-ros/4cfa8ae3d0f1f5955559ac03c65e4e11/raw/373568f96542e32629c27b6273c059dd560dc6ef/rmw-api.repos

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16122/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10752/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16453/)

wjwwood:
Another CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16165/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10793/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16505/)

alsora:
Fixed linter errors and started new CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16169/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10797/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16510/)

alsora:
@wjwwood can you double check the above run? It says everything is unstable, but there are no test failures that I can see and all the failures from your run should have been addressed.

clalancette:
> @wjwwood can you double check the above run? It says everything is unstable, but there are no test failures that I can see and all the failures from your run should have been addressed.

There's a CMake warning here: https://ci.ros2.org/job/ci_linux/16169/cmake/ .  This almost certainly has to do with #1886 .  You can just rebase this onto master, and that should fix it.

alsora:
Merged to get latest CMake changes and triggered a new build.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16174/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10801/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16515/)

rclcpp/include/rclcpp/waitable.hpp
```diff
@@ -200,6 +201,45 @@ class Waitable
...
+   * by the custom classes who inherit from rclcpp::Waitable if they want to use it.
+   *
+   * \sa rclcpp::Waitable::clear_on_ready_callback
+   *
+   * \param[in] callback functor to be called when the waitable becomes ready
+   */
+  RCLCPP_PUBLIC
+  virtual
+  void
+  set_on_ready_callback(std::function<void(size_t, int)> callback);
```
wjwwood, line 231: TODO, perhaps for myself, I really don't like using a bare `int` here, rather than `size_t` or maybe `int64_t`. Maybe we should just use a typed enum like `enum class EntityType : size_t {...};`, see: https://en.cppreference.com/w/cpp/language/enum

wjwwood (commented):
I made another review, fixed a few things, and left with a few questions:

- what is the behavior of the on new message callback when using intra-process and inter-process? Is the event triggered twice? Is that ok?
- Do we need tests for the `rclcpp_action` changes?

In general, I think this is good to merge, just want to follow up on a few particulars and run tests again (after fixing a few small things and also having to rebase some of the pull requests).

alsora:
Started a new build since the previous one was not including recent changes to test_qos_event.cpp

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16235/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10860/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16591/)

wjwwood (approved):

JanStaschulat:
> The events executor will be re-proposed as a follow-up PR

What is the status of the follow-up PR for the events executor? I would be available for code review.

alsora:
Hi @JanStaschulat, thank you!
We did a slight change of plans: we will soon release a stand-alone package containing the events executor.
This package will be compatible with Humble and Rolling and we will release it also as binary.

After that we will open PRs for rclcpp.

PR #1584, opened 2021-03-19
Guard against integer overflow in duration conversion

jacobperron:
Guard against overflow when converting from rclcpp::Duration to builtin_interfaces::msg::Duration,
which is a unsigned to signed conversion.

---

I discovered this bug when making a copy of a `rclcpp::QoS` object with default duration values. Since default durations are max int64 values, they end up overflowing during the copy which uses `rclcpp::Duration::from_rmw_time()` internally.

We can easily see the issue with this reproduction, where we would expect `foo` and `baz` to be equivalent, but they are not:

```c++
  auto foo = rclcpp::Duration::from_nanoseconds(9223372036854775807u);
  auto bar = foo.to_rmw_time();
  rclcpp::Duration baz = rclcpp::Duration::from_rmw_time(bar);

  std::cerr << "foo: " << foo.nanoseconds() << std::endl;
  std::cerr << "baz: " << baz.nanoseconds() << std::endl;
```

### DIFF
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -67,21 +67,27 @@ Duration::operator builtin_interfaces::msg::Duration() const
 {
   builtin_interfaces::msg::Duration msg_duration;
   constexpr rcl_duration_value_t kDivisor = RCL_S_TO_NS(1);
-  constexpr int32_t limit_s = std::numeric_limits<int32_t>::max();
-  constexpr uint32_t limit_ns = std::numeric_limits<uint32_t>::max();
+  constexpr int32_t max_s = std::numeric_limits<int32_t>::max();
+  constexpr int32_t min_s = std::numeric_limits<int32_t>::min();
+  constexpr uint32_t max_ns = std::numeric_limits<uint32_t>::max();
   const auto result = std::div(rcl_duration_.nanoseconds, kDivisor);
   if (result.rem >= 0) {
     // saturate if we will overflow
-    if (result.quot > limit_s) {
-      msg_duration.sec = limit_s;
-      msg_duration.nanosec = limit_ns;
+    if (result.quot > max_s) {
+      msg_duration.sec = max_s;
+      msg_duration.nanosec = max_ns;
     } else {
       msg_duration.sec = static_cast<int32_t>(result.quot);
       msg_duration.nanosec = static_cast<uint32_t>(result.rem);
     }
   } else {
-    msg_duration.sec = static_cast<int32_t>(result.quot - 1);
-    msg_duration.nanosec = static_cast<uint32_t>(kDivisor + result.rem);
+    if (result.quot < min_s) {
+      msg_duration.sec = min_s;
+      msg_duration.nanosec = max_ns;
+    } else {
+      msg_duration.sec = static_cast<int32_t>(result.quot - 1);
+      msg_duration.nanosec = static_cast<uint32_t>(kDivisor + result.rem);
+    }
   }
   return msg_duration;
 }
@@ -251,13 +257,8 @@ Duration::to_rmw_time() const
   rmw_time_t result;
   constexpr rcl_duration_value_t kDivisor = RCL_S_TO_NS(1);
   const auto div_result = std::div(rcl_duration_.nanoseconds, kDivisor);
-  if (div_result.rem >= 0) {
-    result.sec = static_cast<uint64_t>(div_result.quot);
-    result.nsec = static_cast<uint64_t>(div_result.rem);
-  } else {
-    result.sec = static_cast<uint64_t>(div_result.quot - 1);
-    result.nsec = static_cast<uint64_t>(kDivisor + div_result.rem);
-  }
+  result.sec = static_cast<uint64_t>(div_result.quot);
+  result.nsec = static_cast<uint64_t>(div_result.rem);
 
   return result;
 }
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/duration.cpp
```diff
@@ -238,11 +246,19 @@ Duration::to_rmw_time() const
     throw std::runtime_error("rmw_time_t cannot be negative");
   }
 
-  // reuse conversion logic from msg creation
-  builtin_interfaces::msg::Duration msg = *this;
+  // Purposefully avoid using the same message creation logic to avoid possible overflow
+  // converting from uint64_t to int32_t, then back to uint64_t
```
fujitatomoya, line 250: rcutils_duration_value_t is `int64_t`, im not mistaken?

```suggestion
  // converting from int64_t to int32_t, then back to uint64_t
```

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -67,10 +67,18 @@ Duration::operator builtin_interfaces::msg::Duration() const
...
   builtin_interfaces::msg::Duration msg_duration;
   constexpr rcl_duration_value_t kDivisor = RCL_S_TO_NS(1);
+  constexpr std::int32_t limit_s = std::numeric_limits<std::int32_t>::max();
+  constexpr std::uint32_t limit_ns = std::numeric_limits<std::uint32_t>::max();
   const auto result = std::div(rcl_duration_.nanoseconds, kDivisor);
   if (result.rem >= 0) {
-    msg_duration.sec = static_cast<std::int32_t>(result.quot);
-    msg_duration.nanosec = static_cast<std::uint32_t>(result.rem);
+    // saturate if we will overflow
+    if (result.quot > limit_s) {
```
ivanpauno, line 75: should we check if `quot < min_limit` as well?
We're almost not using negative durations, but IIUC those are valid

rclcpp/src/rclcpp/duration.cpp
ivanpauno: we could also write a template for a common implementation, but it doesn't seem to be worth it

ivanpauno (approved):

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -67,10 +67,18 @@ Duration::operator builtin_interfaces::msg::Duration() const
 {
   builtin_interfaces::msg::Duration msg_duration;
   constexpr rcl_duration_value_t kDivisor = RCL_S_TO_NS(1);
+  constexpr std::int32_t limit_s = std::numeric_limits<std::int32_t>::max();
+  constexpr std::uint32_t limit_ns = std::numeric_limits<std::uint32_t>::max();
```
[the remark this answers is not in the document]
jacobperron: I don't have a preference; just copied this from the existing code already in this file.

rclcpp/src/rclcpp/duration.cpp
jacobperron: Updated the comment 05c0169

I realized the logic here could be simplified further since we guarantee that there are no negative values (throwing exception if there are), so I didn't write a template.

rclcpp/src/rclcpp/duration.cpp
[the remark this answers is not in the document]
jacobperron: Updated 6bf4ea1

rclcpp/src/rclcpp/duration.cpp
jacobperron: Yes, it's probably good to cover the negative case too. see e4636b1

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14003/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8825/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11696/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14083/)

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -67,13 +67,27 @@ Duration::operator builtin_interfaces::msg::Duration() const
...
   } else {
-    msg_duration.sec = static_cast<std::int32_t>(result.quot - 1);
-    msg_duration.nanosec = static_cast<std::uint32_t>(kDivisor + result.rem);
+    if (result.quot < min_s) {
+      msg_duration.sec = min_s;
+      msg_duration.nanosec = max_ns;
+    } else {
+      msg_duration.sec = static_cast<int32_t>(result.quot - 1);
+      msg_duration.nanosec = static_cast<uint32_t>(kDivisor + result.rem);
+    }
```
ivanpauno, line 90:
```suggestion
    if (result.quot <= min_s) {
      msg_duration.sec = min_s;
      msg_duration.nanosec = 0;
    } else {
      msg_duration.sec = static_cast<int32_t>(result.quot - 1);
      msg_duration.nanosec = static_cast<uint32_t>(kDivisor + result.rem);
    }
```

The `<=` is because of the `- 1` applied afterwards.
The reminder nanoseconds is always added (`original_nanoseconds = sec * kDivisor + reminder_nanosec`), so it should saturate to `0` here.

jacobperron: Good point. It also makes sense that the lower bound of the unsigned value is `0`. See b621885

jacobperron:
Testing packages above, and including, rclcpp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14016/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8838/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11710/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14101/)

ivanpauno (approved):

PR #1588, opened 2021-03-19
Document alternative for declaring an optional parameter

jacobperron:
While we could conditionally declare a parameter by using a second
parameter, we can also just catch the expected exception in the
case that no parameter override is provided.

I think this makes for a better user experience.

### DIFF
--- a/rclcpp/doc/notes_on_statically_typed_parameters.md
+++ b/rclcpp/doc/notes_on_statically_typed_parameters.md
@@ -85,6 +85,16 @@ if (mode == "modeB") {
 }
 ```
 
+Or, wrap the declaration in a try-catch:
+
+```cpp
+try {
+  node->declare_parameter<int64_t>("param_that_is_optional");
+} catch (const rclcpp::exceptions::NoParameterOverrideProvided &) {
+  // a parameter value was not provided
+}
+```
+
 ## Other migration notes
 
 Declaring a parameter with only a name is deprecated:


rclcpp/doc/notes_on_statically_typed_parameters.md
```diff
@@ -85,6 +85,16 @@ if (mode == "modeB") {
...
 
+Or, wrap the declaration in a try-catch:
+
+```cpp
+try {
+  node->declare_parameter<int64_t>("param_that_is_optional");
+} catch (const rclcpp::exceptions::NoParameterOverrideProvided &) {
+  // a parameter value was not provided
+}
+```
```
wjwwood, line 96: I think to be equivalent to the above, wouldn't it need to be like this:

```suggestion
Or, wrap the declaration in a try-catch:

```cpp
auto mode = node->declare_parameter("mode", "modeA");
try {
  node->declare_parameter<int64_t>("param_that_is_optional");
} catch (const rclcpp::exceptions::NoParameterOverrideProvided &) {
  // a parameter value was not provided
  if (mode == "modeB") {
    // parameter and parameter override was required
    throw;
  }
}
```
```

What you've described here is more like "conditionally declare a parameter if an override is given", which is like a subset of https://github.com/ros2/rclcpp/blob/7bfda87d32bab7e9caa843b2ededa18e427fd00c/rclcpp/include/rclcpp/node_options.hpp#L329-L343.

Also, I'm kind of concerned with the conditionally declared parameters. I'm not saying they shouldn't be allowed, but if I were making a system, I'd prefer the parameters to be the same no matter what, and if one was not provided (and that was ok) it should be "not set" or something. I'm thinking about "dumping" parameters from a node in various scenarios. It's maybe undesirable to have different number and order of parameters in the dump depending on the circumstances, because then if a parameter is missing from the dump, it's hard to know if it was specified or if the dump maybe came from a difference version of the program that didn't have the parameter yet/any longer. But that's just a architectural decision for the developer of the node.

Comment (jacobperron) [reply rclcpp/doc/notes_on_statically_typed_parameters.md:96]: Maybe what we need is a representation for a *value* that is "not set"; we only allow the *type* to be "not set".

Comment (wjwwood) [reply rclcpp/doc/notes_on_statically_typed_parameters.md:96]: Yeah, I think this is essentially what @ivanpauno was talking about with "uninitialized parameters". It can have a type, say `int64_t`, but the value is either an `int64_t` or `uninitialized`.

**Comment by jacobperron** [comment]
Somewhat related, I ran into another scenario where I have a generated list of names and types for which I want to declare parameters. The problem is that I don't want to require users to provide overrides at runtime, rather I expect values may be provided during runtime (e.g. with `ros2 param set`), well after the node starts. The only workaround I found was to conditionally call `declare_parameter` based on the parameter type, e.g.:

```c++
void declare_parameter_with_default(const std::string & name, const rclcpp::ParameterType & type)
{
  switch (type) {
    case rclcpp::ParameterType::PARAMETER_INTEGER:
      parameters_interface_->declare_parameter(name, rclcpp::ParameterValue(0));
      break;
    case rclcpp::ParameterType::PARAMETER_DOUBLE:
      parameters_interface_->declare_parameter(name, rclcpp::ParameterValue(0.0));
      break;
    case rclcpp::ParameterType::PARAMETER_STRING:
      parameters_interface_->declare_parameter(name, rclcpp::ParameterValue(""));
      break;
    case rclcpp::ParameterType::PARAMETER_INTEGER_ARRAY:
      parameters_interface_->declare_parameter(name, rclcpp::ParameterValue(std::vector<int>()));
      break;
    case rclcpp::ParameterType::PARAMETER_DOUBLE_ARRAY:
      parameters_interface_->declare_parameter(
        name, rclcpp::ParameterValue(std::vector<double>()));
      break;
    default:
      throw std::runtime_error("Unexpected parameter type when declaring parameters");
  }   
}
```

I would rather replace the above code with:

```c++
parameters_interface_->declare_parameter(name, type);
```

but it throws an exception if the user does not provide an override.

wjwwood:
That use case makes sense to me @jacobperron

ivanpauno:
Yes, that last use case sounds like a good case for having an "uninitialized" state for statically typed parameters.

rclcpp/doc/notes_on_statically_typed_parameters.md
ivanpauno: I like @wjwwood suggestion [here](https://github.com/ros2/rclcpp/pull/1588#discussion_r597995432).
If not, the try/catch block isn't completely equivalent.

> Also, I'm kind of concerned with the conditionally declared parameters. I'm not saying they shouldn't be allowed, but if I were making a system, I'd prefer the parameters to be the same no matter what, and if one was not provided (and that was ok) it should be "not set" or something. I'm thinking about "dumping" parameters from a node in various scenarios.

That sounds fair.

jacobperron:
I've opened a ticket for allowing uninitialized parameters: https://github.com/ros2/rclcpp/issues/1649

jacobperron:
In retrospect, I agree that we shouldn't encourage the conditional declaration of parameters, so I'm going to close this ticket. I'd much rather see #1649 resolved, supporting "uninitialized" parameters.

Issue #1587, opened 2021-03-19
Avoid calling user parameter callbacks for "qos_overrides" parameter changes


jacobperron:
## Feature request

This is less of an explicit request, and more a question about how we should interact with user-defined parameter callbacks.

#### Feature description

If a user registers a callback with  `rclcpp::Node::add_on_set_parameters_callback()`, should the callback be triggered for any `qos_overrides*` parameters that are internally created by the node (since https://github.com/ros2/rclcpp/pull/1465)?

For example, it is somewhat common to see user callbacks that look like this:

```c++
  auto result = rcl_interfaces::msg::SetParametersResult();
  result.successful = true;

  for (auto parameter : parameters) {
    if (parameter.get_name().compare("user_defined_param") == 0) {
      // do something with the value here ...
    } else {
      string_result << "The parameter " << parameter.get_name() << " is not reconfigurable.\n";
      result.successful = false;
    }
  }
```

Since https://github.com/ros2/rclcpp/pull/1465, user code doing something like the code above will fail (the node terminates the process) if QoS overrides are enabled, because the parameter validation callback is not aware of the `qos_overrides` parameters declared by the nodes implementation.

Possible fixes:

1. Document this pitfall well and provide an example of handling the new "qos_overrides" parameters like:
```diff
  auto result = rcl_interfaces::msg::SetParametersResult();
  result.successful = true;

  for (auto parameter : parameters) {
    if (parameter.get_name().compare("user_defined_param") == 0) {
      // do something with the value here ...
+   else if (parameter.name.rfind("qos_overrides", 0) == 0) {
+     // Ignore changes to QoS override parameters here.
+     // Optionally, we could validate QoS values here, but we also have a more convenient mechanism for that when enabling QoS overrides
    } else {
      string_result << "The parameter " << parameter.get_name() << " is not reconfigurable.\n";
      result.successful = false;
    }
  }
```

2. Document that rejecting unknown parameters is an anti-pattern
Just because one user callback doesn't know how to deal with a parameter doesn't mean that it should reject it. It's possible that another registered callback is expecting said parameter. If one callback rejects a change, then it may prevent other callbacks from getting a chance to handle the change. We would then recommend that the original example above be changed to look like:
```diff
  auto result = rcl_interfaces::msg::SetParametersResult();
  result.successful = true;

  for (auto parameter : parameters) {
    if (parameter.get_name().compare("user_defined_param") == 0) {
      // do something with the value here ...
+  }
-   } else {
[...]

PR #1590, opened 2021-03-22
Document misuse of parameters callback

jacobperron:
Related to https://github.com/ros2/rclcpp/issues/1587

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -866,6 +866,29 @@ public:
    * When a callback returns a not successful result, the remaining callbacks aren't called.
    * The order of the callback is the reverse from the registration order.
    *
+   * It is considered bad practice to reject changes for "unknown" parameters as this prevents
+   * other parts of the node (that may be aware of these parameters) from handling them.
+   * For example, **do not write**:
+   *
+   * ```cpp
+   * rcl_interfaces::msg::SetParametersResult
+   * my_callback(const std::vector<rclcpp::Parameter> & parameters)
+   * {
+   *   rcl_interfaces::msg::SetParametersResult result;
+   *   result.successful = true;
+   *   for (const auto & parameter : parameters) {
+   *     if (parameter.get_name().compare("some_expected_parameter") == 0) {
+   *       // Handle change to parameter 'some_expected_parameter'
+   *     } else {
+   *       // Do not do this
+   *       result.successful = false;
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -866,6 +866,29 @@ class Node : public std::enable_shared_from_this<Node>
    * When a callback returns a not successful result, the remaining callbacks aren't called.
    * The order of the callback is the reverse from the registration order.
    *
+   * It is considered bad practice to reject changes for "unknown" parameters as this prevents
+   * other parts of the node (that may be aware of these parameters) from handling them.
```
clalancette, line 870: It might be a good idea to just move this part of the note up above (where the initial parameter callback example is), and just drop the "do not do this" code.  That will shorten this already lengthy comment, and not put bad code in here that people will blindly copy-n-paste.

clalancette (changes requested):
Feedback inline.

I think it is also worthwhile to add a sentence to the [Galactic release notes](https://docs.ros.org/en/rolling/Releases/Release-Galactic-Geochelone.html) to emphasize this point.  While this is bad practice before Galactic, it would still work in many circumstances.  Now it is probably going to be a hard failure.

clalancette:
Oh, there is a tutorial around parameter callbacks, but it is way outdated: https://docs.ros.org/en/rolling/Tutorials/Monitoring-For-Parameter-Changes-CPP.html .  It would be good to update that, and also add the note about not rejecting unknown parameters there.

wjwwood:
@clalancette I think that tutorial is slightly different. It is only monitoring the parameter events and cannot reject a change, as this callback can.

clalancette:
> @clalancette I think that tutorial is slightly different. It is only monitoring the parameter events and cannot reject a change, as this callback can.

Oh, you are right.  I misunderstood that article.  Never mind about that one, the rest of my comment stands.

rclcpp/include/rclcpp/node.hpp
jacobperron: See a056e3c

fujitatomoya (approved):

clalancette (approved):

jacobperron:
Galactic release note: https://github.com/ros2/ros2_documentation/pull/1267

jacobperron:
rclpy equivalent: https://github.com/ros2/rclpy/pull/734

Issue #1586, opened 2021-03-19
Create load_parameters and delete_parameters methods on rclcpp::[A]syncParametersClient


BriceRenaudeau:
#### Feature description
It would be great to add these two methods to rclcpp::[A]syncParametersClient:

1. `load_parameters` to do the same as `ros2 param load [node] [yaml_file]` 
2. `delete_parameters` to do the same as `ros2 param delete [node] [parameters]`

#### Implementation considerations
Right now, I am using a system call to execute these two commands.

BriceRenaudeau:
To delete a parameter one can use :
```c++
// without value, it sets the parameter to NOT_SET, behave like delete
auto set_results = parameter_client->set_parameters({rclcpp::Parameter(param)});
```

wjwwood:
Makes sense to me, we only have them implemented in `rclpy`. Bonus points for considering if working it into `rcl` so `rclpy` and `rclcpp` can share the same implementation of these two functions makes sense.

BriceRenaudeau:
Linked issue #1029

sloretz:
Closing as this appears to have been resolved by #1596

PR #1596, opened 2021-03-24
Create load_parameters and delete_parameters methods

BriceRenaudeau:
RP to implement #1586

### DIFF
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -157,22 +157,24 @@ public:
       void(std::shared_future<rcl_interfaces::msg::SetParametersResult>)
     > callback = nullptr);
 
+  /// Delete several parameters at once.
   /**
-   * @brief delete_parameters allows to delete several parameters
-   * like "ros2 param delete" would do
-   * @param parameters_names vector of parameters names
-   * @return the future of the set_parameter service used to delete the parameters
+   * This functions behave like command-line tool `ros2 param delete` would.
+   *
+   * \param parameters_names vector of parameters names
+   * \return the future of the set_parameter service used to delete the parameters
    */
   RCLCPP_PUBLIC
   std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>
   delete_parameters(
     const std::vector<std::string> & parameters_names);
 
+  /// Load parameters from yaml file.
   /**
-   * @brief load_parameters allows to load parameters in a node
-   * like "ros2 param load" would do
-   * @param yaml_filename the full name of the yaml file
-   * @return the future of the set_parameter service used to load the parameters
+   * This functions behave like command-line tool `ros2 param load` would.
+   *
+   * \param yaml_filename the full name of the yaml file
+   * \return the future of the set_parameter service used to load the parameters
    */
   RCLCPP_PUBLIC
   std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>
@@ -469,12 +471,13 @@ public:
     );
   }
 
+  /// Delete several parameters at once.
   /**
-   * @brief delete_parameters allows to delete several parameters
-   * like "ros2 param delete" would do
-   * @param parameters_names vector of parameters names
-   * @param timeout for the spin used to make it synchronous
-   * @return the result of the set_parameter service used to delete the parameters
+   * This functions behave like command-line tool `ros2 param delete` would.
+   *
+   * \param parameters_names vector of parameters names
+   * \param timeout for the spin used to make it synchronous
+   * \return the future of the set_parameter service used to delete the parameters
    */
   template<typename RepT = int64_t, typename RatioT = std::milli>
   std::vector<rcl_interfaces::msg::SetParametersResult>
@@ -488,12 +491,13 @@ public:
     );
   }
 
+  /// Load parameters from yaml file.
   /**
-   * @brief load_parameters allows to load parameters in a node
-   * like "ros2 param load" would do
-   * @param yaml_filename the full name of the yaml file
-   * @param timeout for the spin used to make it synchronous
-   * @return the result of the set_parameter service used to load the parameters
+   * This functions behave like command-line tool `ros2 param load` would.
+   *
+   * \param yaml_filename the full name of the yaml file
+   * \param timeout for the spin used to make it synchronous
+   * \return the future of the set_parameter service used to load the parameters
    */
   template<typename RepT = int64_t, typename RatioT = std::milli>
   std::vector<rcl_interfaces::msg::SetParametersResult>
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -181,6 +181,18 @@ public:
   load_parameters(
     const std::string & yaml_filename);
 
+  /// Load parameters from parameter map.
+  /**
+   * This functions filter the parameters to be set.
+   *
+   * \param yaml_filename the full name of the yaml file
+   * \return the future of the set_parameter service used to load the parameters
+   * \throw InvalidParametersException if there is no parameter to set
+   */
+  RCLCPP_PUBLIC
+  std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>
+  load_parameters(const rclcpp::ParameterMap & parameter_map);
+
   RCLCPP_PUBLIC
   std::shared_future<rcl_interfaces::msg::ListParametersResult>
   list_parameters(
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -159,7 +159,7 @@ public:
 
   /// Delete several parameters at once.
   /**
-   * This functions behave like command-line tool `ros2 param delete` would.
+   * This function behaves like command-line tool `ros2 param delete` would.
    *
    * \param parameters_names vector of parameters names
    * \return the future of the set_parameter service used to delete the parameters
@@ -171,7 +171,7 @@ public:
 
   /// Load parameters from yaml file.
   /**
-   * This functions behave like command-line tool `ros2 param load` would.
+   * This function behaves like command-line tool `ros2 param load` would.
    *
    * \param yaml_filename the full name of the yaml file
    * \return the future of the set_parameter service used to load the parameters
@@ -183,7 +183,7 @@ public:
 
   /// Load parameters from parameter map.
   /**
-   * This functions filter the parameters to be set.
+   * This function filters the parameters to be set based on the node name.
    *
    * \param yaml_filename the full name of the yaml file
    * \return the future of the set_parameter service used to load the parameters
@@ -485,7 +485,7 @@ public:
 
   /// Delete several parameters at once.
   /**
-   * This functions behave like command-line tool `ros2 param delete` would.
+   * This function behaves like command-line tool `ros2 param delete` would.
    *
    * \param parameters_names vector of parameters names
    * \param timeout for the spin used to make it synchronous
@@ -505,7 +505,7 @@ public:
 
   /// Load parameters from yaml file.
   /**
-   * This functions behave like command-line tool `ros2 param load` would.
+   * This function behaves like command-line tool `ros2 param load` would.
    *
    * \param yaml_filename the full name of the yaml file
    * \param timeout for the spin used to make it synchronous
--- a/rclcpp/include/rclcpp/parameter_client.hpp
+++ b/rclcpp/include/rclcpp/parameter_client.hpp
@@ -155,6 +155,14 @@ public:
     > callback = nullptr);
 
   RCLCPP_PUBLIC
+  std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>
+  delete_parameters(
+    const std::vector<std::string> & parameters_names,
+    std::function<
+      void(std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>)
+    > callback = nullptr);
+
+  RCLCPP_PUBLIC
   std::shared_future<rcl_interfaces::msg::ListParametersResult>
   list_parameters(
     const std::vector<std::string> & prefixes,
@@ -445,6 +453,18 @@ public:
   }
 
   template<typename RepT = int64_t, typename RatioT = std::milli>
+  std::vector<rcl_interfaces::msg::SetParametersResult>
+  delete_parameters(
+    const std::vector<std::string> & parameters_names,
+    std::chrono::duration<RepT, RatioT> timeout = std::chrono::duration<RepT, RatioT>(-1))
+  {
+    return delete_parameters(
+      parameters_names,
+      std::chrono::duration_cast<std::chrono::nanoseconds>(timeout)
+    );
+  }
+
+  template<typename RepT = int64_t, typename RatioT = std::milli>
   rcl_interfaces::msg::ListParametersResult
   list_parameters(
     const std::vector<std::string> & parameter_prefixes,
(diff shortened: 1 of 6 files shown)


rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -273,6 +273,58 @@ AsyncParametersClient::set_parameters_atomically(
...
+  }
+
+  // create list of parameters to set
+  rclcpp::ParameterMap parameter_map = rclcpp::parameter_map_from(rcl_parameters);
+  std::vector<rclcpp::Parameter> parameters;
+  std::string fully_qualified_name =
+    node_topics_interface_->get_node_base_interface()->get_fully_qualified_name();
+
+  // only add remote_node parameters
+  for (auto & params : parameter_map) {
```
wjwwood, line 308: nitpick: `const auto & ...`

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -273,6 +273,58 @@ AsyncParametersClient::set_parameters_atomically(
...
+    std::string node_name = params.first;
+    if (node_name == fully_qualified_name ||
+      node_name == "/**" ||
+      (node_name.substr(node_name.find("/*/") + 3) == remote_node_name_))
```
wjwwood, line 312: Can you explain the logic of this?

The first case is like "if it matches the name of _this_ node", the second case is "if it is a wild card" (makes sense), and the third is "if the name of the _remote_ node matches it, allowing for `/*/` at the beginning"?

What's confusing to me is that it seems like one condition applies to this node, and the last one applies to a remote node (also the logic there is kind of confusing if you don't have expert knowledge of the wildcard syntax).

So I guess I'm checking my understanding and asking for some comments for each condition that describe what they're doing (even if it seems obvious to you).

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -273,6 +273,58 @@ AsyncParametersClient::set_parameters_atomically(
...
+    {
+      for (auto & param : params.second) {
```
wjwwood, line 314: nitpick: `const auto & ...`

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -273,6 +273,58 @@ AsyncParametersClient::set_parameters_atomically(
...
+        parameters.push_back(param);
+      }
+    }
+  }
+
+  if (parameters.size() == 0) {
+    throw std::runtime_error("No valid parameters from yaml file");
```
wjwwood, line 321: Should this be an exception or maybe should it silently pass if there are no parameter matching the target node? Either way this is the kind of thing that needs to be documented in the header.

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -420,6 +472,42 @@ SyncParametersClient::set_parameters(
   return std::vector<rcl_interfaces::msg::SetParametersResult>();
 }
 
+std::vector<rcl_interfaces::msg::SetParametersResult>
+SyncParametersClient::delete_parameters(
+  const std::vector<std::string> & parameters_names,
+  std::chrono::nanoseconds timeout)
+{
+  auto f = async_parameters_client_->delete_parameters(parameters_names);
```
wjwwood, line 480: This is just a suggestion, it is fine as-is, but if you had a method on the async parameter client that prepared the arguments for `set_parameters()` and returned them, you could get them here and call the sync parameter client's `set_parameters` rather than duplicating the logic in these two new methods.

wjwwood (changes requested):

wjwwood:
Thanks for opening a pull request for this!

rclcpp/src/rclcpp/parameter_client.cpp
```diff
@@ -273,6 +273,58 @@ AsyncParametersClient::set_parameters_atomically(
...
+    parameters.push_back(rclcpp::Parameter(name));
+  }
+  auto future_result = set_parameters(parameters);
+
+  return future_result;
+}
+
+std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>
+AsyncParametersClient::load_parameters(
+  const std::string & yaml_filename)
```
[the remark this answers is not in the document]
BriceRenaudeau: I really want to keep the `load_parameters` that takes a yaml filename but exposing `parameter_map_from_yaml_file(const std::string & yaml_filename)` is a great tool.
I will try to do it.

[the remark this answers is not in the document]
wjwwood: You can keep it, just add another one that takes a `rclcpp::ParameterMap` as well. Something like this:

```c++
// in rclcpp/parameter_map.hpp/cpp
rclcpp::ParameterMap
parameter_map_from_yaml_file(const std::string & yaml_filename)
{
  // ...
}

// in rclcpp/parameter_client.hpp/cpp
std::shared_future<std::vector<rcl_interfaces::msg::SetParameterResult>>
AsyncParameterClient::load_parameters(rclcpp::ParameterMap parameter_map)
{
  // ...
}

std::shared_future<std::vector<rcl_interfaces::msg::SetParameterResult>>
AsyncParameterClient::load_parameters(const std::string & yaml_filename)
{
  rclcpp::ParameterMap parameter_map = rclcpp::parameter_map_from_yaml_file(yaml_filename);
  return this->load_parameters(parameter_map);
}
```

Does that make sense?

rclcpp/src/rclcpp/parameter_client.cpp
BriceRenaudeau: I mixed the two nodes...
Want we want to do is to check witch parameters are for the target node `remote_node_name_`.
To do so we have to check the yaml `node_name` there is 3 options:

1. namespace/node_name 
2. `/**` wild card
3. `/*/` node_name (any namespace)

I made a mistake on the first it should be `node_name == remote_node_name_`.
The second is ok.
The third must compare what is after `/*/` and the remote_node_name_ without namespace!

Hence to fit 1. we need `remote_node_name_` to be fully qualified and to fit 3. we need to extract the node name from it.

Another issue is when the ParameterClient is a client to its own node, remote_node_name_ isn't the fully_qualified_name.

https://github.com/ros2/rclcpp/blob/0088e35052270eca4ccbb0f0c2325f4030c39baf/rclcpp/src/rclcpp/parameter_client.cpp#L42-L46

To fix it, one way is to break the API by adding the remote_node_namespace param. The other is to use `remote_node_name_ = node_base_interface->get_fully_qualified_name()` that may change the behavior of some implementations.

What is your opinion?

rclcpp/src/rclcpp/parameter_client.cpp
BriceRenaudeau: To pass silently, one has to fix the crash that happens if set_parameters is called with an empty vector.

rclcpp/src/rclcpp/parameter_client.cpp
wjwwood: > I made a mistake on the first it should be node_name == remote_node_name_.

👍 I thought so, but I just wanted to double check.

> The other is to use remote_node_name_ = node_base_interface->get_fully_qualified_name() that may change the behavior of some implementations.

I'd go with this. I don't think it should be an issue, because the behavior before seems wrong to me, so anyone trying to use it with namespaces may have not gotten it working at all.

I'd prefer this over adding an explicit node namespace argument, because I believe the intention was always that the node name given to the client was meant to be the fully qualified name (something we should have documented and tested, but apparently haven't).

That's my impression, but I didn't think deeply about it. You may have to try it and see.

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -154,6 +157,28 @@ class AsyncParametersClient
       void(std::shared_future<rcl_interfaces::msg::SetParametersResult>)
     > callback = nullptr);
 
+  /**
+   * @brief delete_parameters allows to delete several parameters
+   * like "ros2 param delete" would do
+   * @param parameters_names vector of parameters names
+   * @return the future of the set_parameter service used to delete the parameters
+   */
```
wjwwood, line 165: Please put the brief on a separate line with `///` and use `\` rather than `@`, in order to match this project's style. Also try to use imperative syntax in the brief (as I do below), and please use complete sentences. Putting it all together, e.g.:

```suggestion
  /// Delete several parameters at once.
  /**
   * This functions like command-line tool `ros2 param delete` would.
   *
   * \param parameters_names vector of parameters names
   * \return the future of the set_parameter service used to delete the parameters
   */
```

rclcpp/src/rclcpp/parameter_client.cpp
wjwwood: Maybe I don't understand, but could you just not call `set_parameters` at all in this case? I guess then you wouldn't know what to return... In that case I guess throwing an exception here is ok.

rclcpp/src/rclcpp/parameter_client.cpp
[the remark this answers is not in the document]
BriceRenaudeau: Yes. I will implement this suggestion.

rclcpp/src/rclcpp/parameter_client.cpp
BriceRenaudeau: using a duplicate method seems to be the standard in the file.

rclcpp/src/rclcpp/parameter_client.cpp
BriceRenaudeau: Done

rclcpp/src/rclcpp/parameter_client.cpp
wjwwood: That's fine, just a suggestion.

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -154,6 +157,42 @@ class AsyncParametersClient
       void(std::shared_future<rcl_interfaces::msg::SetParametersResult>)
     > callback = nullptr);
 
+  /// Delete several parameters at once.
+  /**
+   * This functions behave like command-line tool `ros2 param delete` would.
```
wjwwood, line 162: nitpick

```suggestion
   * This function behaves like command-line tool `ros2 param delete` would.
```

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -154,6 +157,42 @@ class AsyncParametersClient
...
+   * \return the future of the set_parameter service used to delete the parameters
+   */
+  RCLCPP_PUBLIC
+  std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>
+  delete_parameters(
+    const std::vector<std::string> & parameters_names);
+
+  /// Load parameters from yaml file.
+  /**
+   * This functions behave like command-line tool `ros2 param load` would.
```
wjwwood, line 174: nitpick:

```suggestion
   * This function behaves like command-line tool `ros2 param load` would.
```

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -154,6 +157,42 @@ class AsyncParametersClient
...
+   * \return the future of the set_parameter service used to load the parameters
+   */
+  RCLCPP_PUBLIC
+  std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>
+  load_parameters(
+    const std::string & yaml_filename);
+
+  /// Load parameters from parameter map.
+  /**
+   * This functions filter the parameters to be set.
```
wjwwood, line 186: I don't really understand this line. Maybe this?

```suggestion
   * This function filters the parameters to be set based on the node name.
```

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -444,6 +483,46 @@ class SyncParametersClient
     );
   }
 
+  /// Delete several parameters at once.
+  /**
+   * This functions behave like command-line tool `ros2 param delete` would.
```
wjwwood, line 488:
```suggestion
   * This function behaves like command-line tool `ros2 param delete` would.
```

rclcpp/include/rclcpp/parameter_client.hpp
```diff
@@ -444,6 +483,46 @@ class SyncParametersClient
...
+  {
+    return delete_parameters(
+      parameters_names,
+      std::chrono::duration_cast<std::chrono::nanoseconds>(timeout)
+    );
+  }
+
+  /// Load parameters from yaml file.
+  /**
+   * This functions behave like command-line tool `ros2 param load` would.
```
wjwwood, line 508:
```suggestion
   * This function behaves like command-line tool `ros2 param load` would.
```

rclcpp/src/rclcpp/parameter_client.cpp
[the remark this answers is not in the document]
BriceRenaudeau: Done

BriceRenaudeau:
I will need help to fix the failing checks. I don't know why they are failing, they don't seem related to my modifications.

rclcpp/test/rclcpp/test_parameter_client.cpp
```diff
@@ -861,3 +861,92 @@ TEST_F(TestParameterClient, sync_parameter_describe_parameters_allow_undeclared)
...
+    load_node, load_future, std::chrono::milliseconds(100));
+  ASSERT_EQ(result_code, rclcpp::FutureReturnCode::SUCCESS);
+  ASSERT_EQ(load_future.get()[0].successful, true);
+  // list parameters
+  auto list_parameters = asynchronous_client->list_parameters({}, 3);
+  rclcpp::spin_until_future_complete(
+    load_node, list_parameters, std::chrono::milliseconds(100));
+  ASSERT_EQ(
+    list_parameters.get().names.size(),
+    5);
```
wjwwood, line 931: This is one of the errors:

```
00:32:38 In file included from /tmp/ws/src/rclcpp/rclcpp/test/rclcpp/test_parameter_client.cpp:15:
00:32:38 /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest.h: In instantiation of ‘testing::AssertionResult testing::internal::CmpHelperEQ(const char*, const char*, const T1&, const T2&) [with T1 = long unsigned int; T2 = int]’:
00:32:38 /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest.h:1554:23:   required from ‘static testing::AssertionResult testing::internal::EqHelper::Compare(const char*, const char*, const T1&, const T2&) [with T1 = long unsigned int; T2 = int; typename std::enable_if<((! std::is_integral<_Tp>::value) || (! std::is_pointer<_Dp>::value))>::type* <anonymous> = 0]’
00:32:38 /tmp/ws/src/rclcpp/rclcpp/test/rclcpp/test_parameter_client.cpp:929:3:   required from here
00:32:38 /opt/ros/rolling/src/gtest_vendor/include/gtest/gtest.h:1527:11: warning: comparison of integer expressions of different signedness: ‘const long unsigned int’ and ‘const int’ [-Wsign-compare]
00:32:38  1527 |   if (lhs == rhs) {
00:32:38       |       ~~~~^~~~~~
```

wjwwood (changes requested):
The test failures look like flaky tests (can be ignored), but the signed comparison thing will have to be fixed.

BriceRenaudeau:
The 2 errors left are not related to this PR.

wjwwood:

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14230/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9028/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11921/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14334/)

wjwwood:
The compiler warnings are due to not being rebased. They all stem from the recent deprecation/change to `tracetools`'s function `get_symbol`.

wjwwood:
And Rpr will not pass until some releases are done, it is missing `rcl_wait_for_subscribers` which is from a different pull request.

PR #1557, opened 2021-02-23
[rclcpp] Type Adaptation feature

wjwwood:
This feature, a.k.a. Type Masquerading, is still a WIP.

----

Related:

- https://github.com/ros-infrastructure/rep/pull/303
- ros2/examples#300
- ros2/demos#482
- https://github.com/ros2/demos/pull/497

----

This is part of the Type Masquerading project.

- [ ] Add H-turtle release note (see, https://github.com/ros2/demos/pull/497)

PR #1598, opened 2021-03-26
refactor AnySubscriptionCallback and add/deprecate callback signatures

wjwwood:
This pull request refactors `AnySubscriptionCallback` to use C++17's new `std::variant`.
This has the benefit of making the "dispatch" functions evaluate what to do at compile time using `constexpr`, avoiding runtime checks for which callback is set and therefore should be called.

Also refactoring it into the variant has side-effects needed for @audrow's work on supporting the new `TypeAdapter`s with `Subscription`s.

Additionally, I added some new subscription callback signatures and deprecated one other.
I'm wanting a RFC (from anyone, but also @ros2/team) on these changes, specifically:

- added a `const MessageT &` versions
- removed redundant `const` statements on pass-by-value versions
  - for example:
      `std::function<void (const std::shared_ptr<const MessageT>)>` ->
      `std::function<void (std::shared_ptr<const MessageT>)>`
- added a `const std::shared_ptr<const MessageT> &` versions
- ~~deprecated the `std::shared_ptr<MessageT>` versions~~
  - We decided to not deprecate this right now, but discourage it in documentation and deprecate later (h-turtle?)

Details on each change below...

Here's an exhaustive list of the callback signatures now:

- `std::function<void (const MessageT &)>` (new)
- `std::function<void (const MessageT &, const rclcpp::MessageInfo &)>` (new)
- `std::function<void (std::unique_ptr<MessageT, MessageDeleter>)>`
- `std::function<void (std::unique_ptr<MessageT, MessageDeleter>, const rclcpp::MessageInfo &)>`
- `std::function<void (std::shared_ptr<const MessageT>)>` (removed redundant `const`)
- `std::function<void (std::shared_ptr<const MessageT>, const rclcpp::MessageInfo &)>` (removed redundant `const`)
- `std::function<void (const std::shared_ptr<const MessageT> &)>` (new)
- `std::function<void (const std::shared_ptr<const MessageT> &, const rclcpp::MessageInfo &)>` (new)
- `std::function<void (std::shared_ptr<MessageT>)>` (~~deprecated~~ we decided to push this until later)
- `std::function<void (std::shared_ptr<MessageT>, const rclcpp::MessageInfo &)>` (~~deprecated~~ we decided to push this until later)

----

I think the new `const MessageT &` should be preferred unless you are doing something with the `unique_ptr` signature (trying to get zero-copy).
I actually don't know what the use case for receiving a `shared_ptr` is over `const &`, to be honest.
@mjcarroll didn't we discuss this at some point when refactoring the `publish()` signatures? Maybe we have notes?

----

I removed the redundant `const` statements, because they don't really make sense, and the compiler doesn't use them to differentiate types.
For example, if you put `std::function<void (const std::shared_ptr<const MessageT>)>` and `std::function<void (std::shared_ptr<const MessageT>)>` in a `std::variant` it will result in a ambiguous call when you try to assign one of those into the variant.
This example demonstrates this:

```c++
#include <functional>
#include <memory>
#include <variant>

int main() {
  using variant_type = std::variant<
    // comment out either of these to get it to work
    std::function<void (std::shared_ptr<const int>)>,
    std::function<void (const std::shared_ptr<const int>)>
  >;

  {
    std::function<void (std::shared_ptr<const int>)> cb1 =
      [](std::shared_ptr<const int>) {};
    variant_type v = cb1;
  }

  {
    std::function<void (const std::shared_ptr<const int>)> cb2 =
      [](const std::shared_ptr<const int>) {};
    variant_type v = cb2;
  }
}
```

Removing the `const` statements will not affect users who have them in their callbacks right now.

----

I added `const shared_ptr<...> &` versions because I felt it better reflected that the C++ Core Guidelines suggests, since the user is indicating they `"might" retain refcount`:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-owner

But I'm looking for feedback on that decision too.

Edit: we're keeping this one and the `shared_ptr<const MessageT>` one

----

I proposed to deprecate the `shared_ptr<MessageT>` versions of the callbacks, because to do it safely it will require that we make a unique copy of the shared pointer for each subscription that asks for that signature. Otherwise one callback could mutate MessageT before or during another callback that is reading it. Also, if the user really wants a `shared_ptr<MessageT>` at the end of the day, they can make their own from `const MessageT &` or `unique_ptr<MessageT>`.

But this deprecation will likely hit many people and that might not be what we want, even if the required change is small, e.g.:

```diff
--- void my_callback(std::shared_ptr<MessageT> msg)
+++ (clipboard)
@@ -1 +1 @@
-void my_callback(std::shared_ptr<MessageT> msg)
+void my_callback(std::shared_ptr<const MessageT> msg)
```

Edit: we decided to delay this until after Galactic

----

I still need to fix all the deprecation warnings (~~here~~ and in downstream packages) and add a few tests, but I wanted to open this now to unblock @audrow on https://github.com/ros2/rclcpp/pull/1557 and to get feedback from @ros2/team.

### DIFF
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -20,7 +20,7 @@
 #include <stdexcept>
 #include <type_traits>
 #include <utility>
-#include <variant>
+#include <variant>  // NOLINT[build/include_order]
 
 #include "tracetools/tracetools.h"
 #include "tracetools/utils.hpp"
@@ -141,7 +141,7 @@ private:
 
 public:
   explicit
-  AnySubscriptionCallback(const AllocatorT & allocator = AllocatorT())
+  AnySubscriptionCallback(const AllocatorT & allocator = AllocatorT())  // NOLINT[runtime/explicit]
   {
     message_allocator_ = allocator;
     allocator::set_allocator_for_deleter(&message_deleter_, &message_allocator_);
@@ -149,7 +149,7 @@ public:
 
   [[deprecated("use AnySubscriptionCallback(const AllocatorT & allocator) instead")]]
   explicit
-  AnySubscriptionCallback(std::shared_ptr<AllocatorT> allocator)
+  AnySubscriptionCallback(std::shared_ptr<AllocatorT> allocator)  // NOLINT[runtime/explicit]
   {
     if (allocator == nullptr) {
       throw std::runtime_error("invalid allocator");
@@ -246,15 +246,17 @@ public:
           callback(create_unique_ptr_from_shared_ptr_message(message));
         } else if constexpr (std::is_same_v<T, UniquePtrWithInfoCallback>) {
           callback(create_unique_ptr_from_shared_ptr_message(message), message_info);
-        } else if constexpr (
+        } else if constexpr (  // NOLINT[readability/braces]
           std::is_same_v<T, SharedConstPtrCallback>||
           std::is_same_v<T, ConstRefSharedConstPtrCallback>||
-          std::is_same_v<T, SharedPtrCallback>) {
+          std::is_same_v<T, SharedPtrCallback>)
+        {
           callback(message);
-        } else if constexpr (
+        } else if constexpr (  // NOLINT[readability/braces]
           std::is_same_v<T, SharedConstPtrWithInfoCallback>||
           std::is_same_v<T, ConstRefSharedConstPtrWithInfoCallback>||
-          std::is_same_v<T, SharedPtrWithInfoCallback>) {
+          std::is_same_v<T, SharedPtrWithInfoCallback>)
+        {
           callback(message, message_info);
         }
       }, callback_variant_);
@@ -275,7 +277,7 @@ public:
           std::is_same_v<T, SharedConstPtrCallback>||
           std::is_same_v<T, ConstRefSharedConstPtrCallback>) {
           callback(message);
-        } else if constexpr (
+        } else if constexpr (  // NOLINT[readability/braces]
           std::is_same_v<T, SharedConstPtrWithInfoCallback>||
           std::is_same_v<T, ConstRefSharedConstPtrWithInfoCallback>) {
           callback(message, message_info);
@@ -302,7 +304,7 @@ public:
           std::is_same_v<T, SharedConstPtrCallback>||
           std::is_same_v<T, ConstRefSharedConstPtrCallback>) {
           callback(std::move(message));
-        } else if constexpr (
+        } else if constexpr (  // NOLINT[readability/braces]
           std::is_same_v<T, SharedConstPtrWithInfoCallback>||
           std::is_same_v<T, ConstRefSharedConstPtrWithInfoCallback>) {
           callback(std::move(message), message_info);
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -219,6 +219,14 @@ public:
     const rclcpp::MessageInfo & message_info)
   {
     TRACEPOINT(callback_start, static_cast<const void *>(this), false);
+    // Check if the variant is "unset", throw if it is.
+    if (callback_variant_.index() == 0) {
+      if (std::get<0>(callback_variant_) == nullptr) {
+        // This can happen if it is default initialized, or if it is assigned nullptr.
+        throw std::runtime_error("dispatch called on an unset AnySubscriptionCallback");
+      }
+    }
+    // Dispatch.
     std::visit(
       [&message, &message_info, this](auto && callback) {
         using T = std::decay_t<decltype(callback)>;
@@ -263,6 +271,14 @@ public:
     const rclcpp::MessageInfo & message_info)
   {
     TRACEPOINT(callback_start, static_cast<const void *>(this), true);
+    // Check if the variant is "unset", throw if it is.
+    if (callback_variant_.index() == 0) {
+      if (std::get<0>(callback_variant_) == nullptr) {
+        // This can happen if it is default initialized, or if it is assigned nullptr.
+        throw std::runtime_error("dispatch called on an unset AnySubscriptionCallback");
+      }
+    }
+    // Dispatch.
     std::visit(
       [&message, &message_info](auto && callback) {
         using T = std::decay_t<decltype(callback)>;
@@ -290,6 +306,14 @@ public:
     const rclcpp::MessageInfo & message_info)
   {
     TRACEPOINT(callback_start, static_cast<const void *>(this), true);
+    // Check if the variant is "unset", throw if it is.
+    if (callback_variant_.index() == 0) {
+      if (std::get<0>(callback_variant_) == nullptr) {
+        // This can happen if it is default initialized, or if it is assigned nullptr.
+        throw std::runtime_error("dispatch called on an unset AnySubscriptionCallback");
+      }
+    }
+    // Dispatch.
     std::visit(
       [&message, &message_info](auto && callback) {
         using T = std::decay_t<decltype(callback)>;
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -96,21 +96,13 @@ struct AnySubscriptionCallbackHelper
 
 template<
   typename MessageT,
-  typename AllocatorT
+  typename AllocatorT = std::allocator<void>
 >
 class AnySubscriptionCallback
 {
 private:
   using HelperT = typename rclcpp::detail::AnySubscriptionCallbackHelper<MessageT, AllocatorT>;
   using MessageDeleterHelper = rclcpp::detail::MessageDeleterHelper<MessageT, AllocatorT>;
-
-  // using variant_type = HelperT::variant_type;
-
-  // TODO(wjwwood): switch to inheriting from std::variant (i.e. HelperT::variant_type) once
-  // inheriting from std::variant is realistic (maybe C++23?), see:
-  //   http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2162r0.html
-  // For now, compose the variant into this class as a private attribute.
-  typename HelperT::variant_type callback_variant_;
 
   using MessageAllocTraits = typename MessageDeleterHelper::MessageAllocTraits;
   using MessageAlloc = typename MessageDeleterHelper::MessageAlloc;
@@ -162,7 +154,7 @@ public:
    * unique_ptr callback signatures when using them with lambda functions.
    */
   template<typename CallbackT>
-  void
+  AnySubscriptionCallback<MessageT, AllocatorT>
   set(CallbackT callback)
   {
     // Use the SubscriptionCallbackTypeHelper to determine the actual type of
@@ -192,6 +184,9 @@ public:
       // Otherwise just assign it.
       callback_variant_ = static_cast<typename scbth::callback_type>(callback);
     }
+
+    // Return copy of self for easier testing, normally will be compiled out.
+    return *this;
   }
 
   /// Function for shared_ptr to non-const MessageT, which is deprecated.
@@ -345,7 +340,25 @@ public:
 #endif  // TRACETOOLS_DISABLED
   }
 
+  typename HelperT::variant_type &
+  get_variant()
+  {
+    return callback_variant_;
+  }
+
+  const typename HelperT::variant_type &
+  get_variant() const
+  {
+    return callback_variant_;
+  }
+
 private:
+  // TODO(wjwwood): switch to inheriting from std::variant (i.e. HelperT::variant_type) once
+  // inheriting from std::variant is realistic (maybe C++23?), see:
+  //   http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2162r0.html
+  // For now, compose the variant into this class as a private attribute.
+  typename HelperT::variant_type callback_variant_;
+
   MessageAlloc message_allocator_;
   MessageDeleter message_deleter_;
 };
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -234,6 +234,8 @@ public:
           MessageAllocTraits::construct(message_allocator_, ptr, *message);
           return std::unique_ptr<MessageT, MessageDeleter>(ptr, message_deleter_);
         };
+        // Avoid an unused variable warning in constexpr-if cases below where it is not used.
+        (void)create_unique_ptr_from_shared_ptr_message;
 
         if constexpr (std::is_same_v<T, ConstRefCallback>) {
           callback(*message);
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -213,6 +213,14 @@ public:
     callback_variant_ = callback;
   }
 
+  std::unique_ptr<MessageT, MessageDeleter>
+  create_unique_ptr_from_shared_ptr_message(const std::shared_ptr<const MessageT> & message)
+  {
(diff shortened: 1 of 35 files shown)


rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -15,256 +15,338 @@
...
   using UniquePtrWithInfoCallback =
-    std::function<void (MessageUniquePtr, const rclcpp::MessageInfo &)>;
+    std::function<void (std::unique_ptr<MessageT, MessageDeleter>, const rclcpp::MessageInfo &)>;
+
+  // TODO(wjwwood): should we drop this one, because you can achieve it easily
+  //   with the const shared_ptr & version?
+  using SharedConstPtrCallback =
+    std::function<void (std::shared_ptr<const MessageT>)>;
+  using SharedConstPtrWithInfoCallback =
+    std::function<void (std::shared_ptr<const MessageT>, const rclcpp::MessageInfo &)>;
```
wjwwood, line 69: RFC

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -15,256 +15,338 @@
...
+
+  // TODO(wjwwood): should we drop this one, because you should prefer the
+  //   shared_ptr one (if you intend to actuall share ownership) or the
+  //   const MessageT & one if you just want to read the message?
+  using ConstRefSharedConstPtrCallback =
+    std::function<void (const std::shared_ptr<const MessageT> &)>;
+  using ConstRefSharedConstPtrWithInfoCallback =
+    std::function<void (const std::shared_ptr<const MessageT> &, const rclcpp::MessageInfo &)>;
```
wjwwood, line 77: RFC

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -15,256 +15,338 @@
...
+
+  // Deprecated signatures:
+  using SharedPtrCallback =
+    std::function<void (std::shared_ptr<MessageT>)>;
+  using SharedPtrWithInfoCallback =
+    std::function<void (std::shared_ptr<MessageT>, const rclcpp::MessageInfo &)>;
```
wjwwood, line 83: RFC

rclcpp/include/rclcpp/detail/subscription_callback_type_helper.hpp
```diff
@@ -0,0 +1,163 @@
...
+ *     std::function<void (std::shared_ptr<int>)> cb = [](std::shared_ptr<int>){};
+ *     f(cb);
+ *   }
+ *
+ * If this program ever starts working in a future version of C++, this class
+ * may become redundant.
+ *
+ * This helper works by using SFINAE with rclcpp::function_traits::same_arguments<>
+ * to narrow down the exact std::function<> type for the given CallbackT.
+ */
```
wjwwood, line 60: @mjcarroll FYI, since you mentioned something similar was at issue in ignition?

rclcpp/include/rclcpp/any_subscription_callback.hpp
ivanpauno: > should we drop this one, because you should prefer the
> shared_ptr one (if you intend to actuall share ownership) or the
> const MessageT & one if you just want to read the message?

:+1:

rclcpp/include/rclcpp/any_subscription_callback.hpp
ivanpauno: :+1:

audrow (approved):
Looks good to me with green CI, and I agree with the design decisions.

rclcpp/include/rclcpp/any_subscription_callback.hpp
ivanpauno: > added a const std::shared_ptr<const MessageT> & versions
> deprecated the std::shared_ptr<MessageT> versions

Reading the post again, I see that the proposal was to add the `const shared_ptr<T> &` signature and deprecate the one that passes a shared pointer by value.

I don't mind much really.
Most of the cases the user will want either:

- To always take ownership, where passing a `std::shared_ptr<T>` by value makes more sense.
- To not take ownership at all, where passing a `const & T` makes sense.

There's a slight chance that someone will want to write a callback where they sometimes take ownership and sometimes not (depending on some condition), where `const std::shared_ptr<T> &` makes more sense.

Having the three options seems like a lot, so maybe `const std::shared_ptr<T> &` is better than `std::shared_ptr<T>` (?).

Karsten1987: if we really only differentiate between having full ownership (read & write the incoming message) vs not taking ownership (read-only) then why do we need a smart pointer aka heap allocation in the first place? 
We could support a `const T &` for read-only and `T &&` for read-write. 
I am just throwing it in the mix to see whether we'd need a heap-allocation at all.

ivanpauno: >  read-only and T && for read-write

In that case you only have access until the end of the callback, if it's allocated in the heap it's lifetime can be extended.
`T &&` would be somehow equivalent to the current `std::unique_ptr<T>` signatures.

>  heap allocation in the first place?

In the case of `const T &` or `T &&` heap allocation can be avoided, though for really large messages you can get a stack overflow (e.g. https://github.com/ros2/performance_test/blob/57791ae74169a7eca61f72541a35cf824a8e3b09/performance_test/src/msg/Array8m.msg#L1).
Though that's weird, most "large messages" use "sequences" (heap allocated) and not "arrays", so their size in the stack isn't big.

i.e. I would always heap allocate the messages independently of the signature.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacobperron: If we are going to keep `const shared_ptr<const <MessageT> &`, then I think we should definitely keep this one too.
This signature makes the intent that a refcount "will" be kept versus "might" be kept.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacobperron: I don't see much value in keeping this signature. Like @ivanpauno says, it seems like there a small chance that someone will want to conditionally take ownership in the callback, in which case the signature makes sense.

In any case, we should clearly document when/why users should pick a signature. E.g. prefer `const & T`, unless you want to retain a copy, etc.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacobperron: :+1:  It'll be disruptive, but sounds like a safer direction to move in.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: Maybe, but on the other hand, it doesn't matter to the middleware which is happening, it would boil down to something like:

```c++
shared_ptr<const MessageT> message_to_be_delivered;
if (shared_ptr<const MessageT> || const shared_ptr<const Message T> &) {
  user_callback(message_to_be_delivered);
}
```

Also, just because the user has a callback is using `shared_ptr<const MessageT>` doesn't mean they'll keep it beyond the scope of the function, and if they're not necessarily going to do that, `const shared_ptr<const MessageT> &` is slight more efficient (in theory).

But thanks for giving you preference, I'll take into consideration.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: My personal preference is to keep only one or the other, but I actually am leaning towards the `const shared_ptr<const MessageT> &`, since the user could always do this:

```c++
void callback(const shared_ptr<const MessageT> & msg_ref) {
  shared_ptr<const MessageT> msg = msg_ref;
}
```

And allowing the user to use `shared_ptr<const MessageT>` in their signature doesn't tell the middleware anything or let it optimize anything (that I can foresee). So it basically comes do to (for me), the `const &` one is more flexible (case where you might want to retain ownership) and the plain one is easier to write.

Keep the feedback coming :)

>Having the three options seems like a lot, so maybe const std::shared_ptr<T> & is better than std::shared_ptr<T> (?).

I'll also note that it doesn't cost us that much to keep all the signatures if they have some "correct" purpose.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: We could keep the signature and just make a copy for each subscription that asks for this, but I think it adds to the clutter and just isn't necessary. I'd prefer to put the allocation of the copy onto the user so they can control it. Making those copies would cause us to do the allocation and it might not be trivial for the user to influence how that happens.

rclcpp/include/rclcpp/any_subscription_callback.hpp
jacobperron: I figured we were introducing the `const shared_ptr &` signature to better express intent of the user. If I see a user callback accepting a shared_ptr then my first thought would be that they are going to keep a reference. Otherwise, they should use a different signature.

I don't feel too strongly about if we keep this one or not, but it would be nice to document why someone should prefer one signature over another.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: > 👍 It'll be disruptive, but sounds like a safer direction to move in.

It will be disruptive, I think quite a lot of people are using this right now. I haven't addressed all the deprecation warnings yet just because I wanted to see how we landed on this signature.

Luckily we've had the `shared_ptr<const MessageT>` for a while so people can use that in older versions if they have one branch for several distributions. Otherwise, I'd say we should consider backporting some new alternatives to this one.

wjwwood: I'm running a CI job (linux only, only cyclone) to see what deprecation warnings come up:

[Build Status](https://ci.ros2.org/job/ci_linux/14123/)

clalancette:
I talked this over with @wjwwood , and here are some of my thoughts:

> I removed the redundant `const` statements, because they don't really make sense, and the compiler doesn't use them to differentiate types.

:+1: to this

> I added const shared_ptr<...> & versions because I felt it better reflected that the C++ Core Guidelines suggests, since the user is indicating they "might" retain refcount:

:+1: to this

> I proposed to deprecate the shared_ptr<MessageT> versions of the callbacks, because to do it safely it will require that we make a unique copy of the shared pointer for each subscription that asks for that signature.

As William explained, this one is redundant with the `const std::shared_ptr<T> &` signature.  But the cost of keeping it is low, and the potential number of impacted downstream users is high, so I'd suggest just keeping the `std::shared_ptr<T>` signatures.  However, it can have a performance impact in some cases, so I'll suggest we update the documentation to recommend against its use in normal circumstances.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: So it failed because we don't have a way to propagate the C++17 standard being selected in cmake. I'm working on a way to address that, but you could browse the warnings from `rclcpp` which built.

rclcpp/include/rclcpp/any_subscription_callback.hpp
ivanpauno: I agree with everything in the last comment and [here](https://github.com/ros2/rclcpp/pull/1598#issuecomment-809683513).

`const shared_ptr<> &` is a bit more flexible, but also keeping the `shared_ptr` by value signature is low cost and might express purpose more clearly in some cases. So in order to avoid breaking code, keeping both sounds fine :+1:.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: Though I would have preferred to push the deprecation forward, I'm going to drop the deprecation from this pull request, due to concerns about how close it is to the Galactic API freeze made by other maintainers.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: We decided to keep both the `shared_ptr<const MessageT>` and `const shared_ptr<const MessageT> &` signatures.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: We decided to keep both the `shared_ptr<const MessageT>` and `const shared_ptr<const MessageT> &` signatures.

rclcpp/include/rclcpp/any_subscription_callback.hpp
wjwwood: We decided not to deprecate this until after galactic, but I will be discouraging its use in the doc updates I am going to make, and I will "fix" all the places it would have been a warning that I can. I already did it for this repository.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14183/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8980/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11865/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14285/)

audrow:
Here's a new CI run. It's a rebuild of the previous one but will use https://github.com/ament/googletest/pull/15 and https://github.com/ros-visualization/rqt/pull/246 (but not https://github.com/ros2/rclcpp/pull/1613).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14208/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9006/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11901/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14311/)

audrow (approved):

wjwwood:

wjwwood:

audrow:
Reran after refactoring CMakeLists to use `ament_target_dependencies`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14221/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9021/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11912/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14322/)

wjwwood:
Ok, I think this is ready to merge, but let me explain.

> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/14221/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9021/)
> * macOS [Build Status](http://ci.ros2.org/job/ci_osx/11912/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/14322/)

so these tests that @audrow ran were good except for two things:

- cmake linter error and
- many c++17 deprecation warnings on windows

@audrow fixed the linter issue and reran up to that package with these which are good:

> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/14226/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9025/)
> * macOS [Build Status](http://ci.ros2.org/job/ci_osx/11917/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/14330/) (only failure is cmake warning in cyclonedds)

And I disabled the c++ deprecation warnings on Windows for now, with a plan to follow up later, then retested windows:

> * Second Windows (after suppressing c++17 warnings) [Build Status](https://ci.ros2.org/job/ci_windows/14325/) (only failures were cmake linter warning that was fixed, and cyclonedds cmake warnings)

Therefore I think we've addressed all the concerns, so I'm gonna merge all of these and then run a smoke test CI job to ensure I got everything merged.

wjwwood:
Follow up issues:

- https://github.com/ros2/ros2/issues/1123
- https://github.com/ros2/rclcpp/issues/1619

PR #1606, opened 2021-03-30
Fix flaky lifecycle node tests

### LINKED ISSUE #1605: Flaky test_lifecycle_node tests
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - source
- Version or commit hash:
  - 63e79223f98867ba00c059eb612438ca9461a2f7
- DDS implementation:
  - CycloneDDS
  - rclcpp

#### Steps to reproduce issue

Induce some stress in one terminal
```
stress -c 32
```

In another terminal run the test
```
./build/rclcpp_lifecycle/test_lifecycle_node
```

#### Expected behavior

Test passes.

#### Actual behavior

Test frequently fails at different locations.

jacobperron:
Apparently, the topics and services that LifecycleNode provides are not
available immediately after creating a node.

Fix flaky tests by accounting for some delay between the LifecycleNode
constructor and queries about its topics and services.

Fixes #1605

### DIFF
--- a/rclcpp_lifecycle/test/test_lifecycle_node.cpp
+++ b/rclcpp_lifecycle/test/test_lifecycle_node.cpp
@@ -34,6 +34,84 @@
 using lifecycle_msgs::msg::State;
 using lifecycle_msgs::msg::Transition;
 
+static
+bool wait_for_event(
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node,
+  std::function<bool()> predicate,
+  std::chrono::nanoseconds timeout,
+  std::chrono::nanoseconds sleep_period)
+{
+  auto start = std::chrono::steady_clock::now();
+  std::chrono::microseconds time_slept(0);
+
+  while (!predicate() &&
+    time_slept < std::chrono::duration_cast<std::chrono::microseconds>(timeout))
+  {
+    rclcpp::Event::SharedPtr graph_event = node->get_graph_event();
+    node->wait_for_graph_change(graph_event, sleep_period);
+    time_slept = std::chrono::duration_cast<std::chrono::microseconds>(
+      std::chrono::steady_clock::now() - start);
+  }
+  return predicate();
+}
+
+static
+bool wait_for_topic(
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node,
+  const std::string & topic,
+  std::chrono::nanoseconds timeout = std::chrono::seconds(3),
+  std::chrono::nanoseconds sleep_period = std::chrono::milliseconds(100))
+{
+  return wait_for_event(
+      node,
+      [node, topic]()
+      {
+        auto topic_names_and_types = node->get_topic_names_and_types();
+        return topic_names_and_types.end() != topic_names_and_types.find(topic);
+      },
+      timeout,
+      sleep_period);
+}
+
+static
+bool wait_for_service(
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node,
+  const std::string & service,
+  std::chrono::nanoseconds timeout = std::chrono::seconds(3),
+  std::chrono::nanoseconds sleep_period = std::chrono::milliseconds(100))
+{
+  return wait_for_event(
+      node,
+      [node, service]()
+      {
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
lgtm

rclcpp_lifecycle/test/test_lifecycle_node.cpp
```diff
@@ -34,6 +34,84 @@
...
+
+  while (!predicate() &&
+    time_slept < std::chrono::duration_cast<std::chrono::microseconds>(timeout))
+  {
+    rclcpp::Event::SharedPtr graph_event = node->get_graph_event();
+    node->wait_for_graph_change(graph_event, sleep_period);
+    time_slept = std::chrono::duration_cast<std::chrono::microseconds>(
+      std::chrono::steady_clock::now() - start);
+  }
+  return predicate();
```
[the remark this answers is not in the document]
jacobperron: Applied suggestion in 2c5545e

rclcpp_lifecycle/test/test_lifecycle_node.cpp
```diff
@@ -34,6 +34,85 @@
...
+  }
+  return predicate_result;
+}
+
+static
+bool wait_for_topic(
+  std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node,
+  const std::string & topic,
+  std::chrono::nanoseconds timeout = std::chrono::seconds(3),
+  std::chrono::nanoseconds sleep_period = std::chrono::milliseconds(100))
```
Karsten1987, line 64: while these timeouts usually work, they are still prone to flakiness IMO. That is, we might increase or decrease these over time. It's a very minor nit, but you could pull these out into constants in case we have to deal with them later on again.

jacobperron: I opted for parameters to the function so they could be adapted individually if necessary, though it's more likely we'd just update these defaults if we needed to bump the timeout. I think something else is very wrong if a node can't query it's own topics/services within 3 seconds :fearful:

jacobperron: I guess a resource constrained machine could cause us to exceed the timeout.

jacobperron: See 71c392a

Karsten1987: I think you have it in line 103/104 again.

Karsten1987 (approved):

jacobperron:
I just noticed that this `rclcpp` is also flaky due to a similar pattern of creating a ROS entity and then immediately checking that it exists:

https://github.com/ros2/rclcpp/blob/f986ca3edce8447740e51f21cb7114cba3e2666f/rclcpp/test/rclcpp/test_node.cpp#L2558-L2561

Seems like this pattern is not reliable since the switch to CycloneDDS. I'm not sure we guarantee anywhere that the graph query API will contain entities that were created right before the query, even if they were created by the same node making the query.

eboasson:
@clalancette I checked, the reason is that the Cyclone RMW layer updates the graph layer asynchronously from the discovery thread. There's no particular reason the create/destroy subscription and publisher operations can't update the graph synchronously, one would just have to make sure to filter out any local entities in the discovery thread.

clalancette:
> @clalancette I checked, the reason is that the Cyclone RMW layer updates the graph layer asynchronously from the discovery thread. There's no particular reason the create/destroy subscription and publisher operations can't update the graph synchronously, one would just have to make sure to filter out any local entities in the discovery thread.

Thanks for looking into it, it is appreciated.  So it sounds this isn't unexpected, it is just different from what we had before.  I guess in order to be robust to different implementations, the tests really shouldn't expect that the topics and services are immediately available.  @jacobperron does that make sense to you?

jacobperron:
Makes sense to me. I'm working on adding API to `rcl` to make this kind of testing pattern easier, e.g. `rcl_wait_for_publishers()`. Stay tuned...

jacobperron:
> Stay tuned...

https://github.com/ros2/rcl/pull/907

PR #1609, opened 2021-03-30
Initialize integers in test_parameter_event_handler.cpp to avoid undefined behavior

ivanpauno:
Likely fixes https://ci.ros2.org/view/nightly/job/nightly_linux_release/1870/testReport/junit/projectroot.test/rclcpp/test_parameter_event_handler/

I have only see that once, it's probably flaky due to the variables not being initialized.


clalancette:
> I have only see that once, it's probably flaky due to the variables not being initialized.

Actually, it is happening quite a lot, so it will be good to fix this one.

jacobperron (approved):
I can reliable see this failing locally.

This PR fixes it.

fujitatomoya (approved):

ivanpauno:
Rpr failure is unrelated

PR #1612, opened 2021-04-01
Use rclcpp::guard_condition

mauropasse:
This PR makes **rclcpp** to not handle directly `rcl_guard_condition_t`, but to use the `rclcpp::GuardCondition` class which takes care of creation/init/delete of it, which makes the code cleaner and less error prone. 

The work being done in https://github.com/ros2/rmw/pull/286 assigns listener callbacks to RMW entities, and originally also assigned a callback to the RMW guard conditions, since many events are generated from the triggering of a guard condition. The listener callback for guard conditions is now removed, so this PR sets the ground for a possible implementation where the guard condition callback is called directly from `rclcpp::GuardCondition` (instead of the originall call from RMW guard conditions).

### DIFF
--- a/rclcpp/include/rclcpp/detail/add_guard_condition_to_rcl_wait_set.hpp
+++ b/rclcpp/include/rclcpp/detail/add_guard_condition_to_rcl_wait_set.hpp
@@ -1,0 +1,40 @@
+// Copyright 2021 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__DETAIL__ADD_GUARD_CONDITION_TO_RCL_WAIT_SET_HPP_
+#define RCLCPP__DETAIL__ADD_GUARD_CONDITION_TO_RCL_WAIT_SET_HPP_
+
+#include "rclcpp/guard_condition.hpp"
+
+namespace rclcpp
+{
+namespace detail
+{
+
+/// Adds the guard condition to a waitset
+/**
+ * This function is thread-safe.
+ * \param[in] wait_set reference to a wait set where to add the guard condition
+ * \param[in] guard_condition reference to the guard_condition to be added
+ */
+RCLCPP_PUBLIC
+void
+add_guard_condition_to_rcl_wait_set(
+  rcl_wait_set_t & wait_set,
+  const rclcpp::GuardCondition & guard_condition);
+
+}  // namespace detail
+}  // namespace rclcpp
+
+#endif  // RCLCPP__DETAIL__ADD_GUARD_CONDITION_TO_RCL_WAIT_SET_HPP_
--- a/rclcpp/include/rclcpp/detail/add_guard_condition_to_rcl_wait_set.hpp
+++ b/rclcpp/include/rclcpp/detail/add_guard_condition_to_rcl_wait_set.hpp
@@ -24,7 +24,6 @@ namespace detail
 
 /// Adds the guard condition to a waitset
 /**
- * This function is thread-safe.
  * \param[in] wait_set reference to a wait set where to add the guard condition
  * \param[in] guard_condition reference to the guard_condition to be added
  */
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -525,7 +525,7 @@ protected:
   std::atomic_bool spinning;
 
   /// Guard condition for signaling the rmw layer to wake up for special events.
-  rcl_guard_condition_t interrupt_guard_condition_ = rcl_get_zero_initialized_guard_condition();
+  rclcpp::GuardCondition interrupt_guard_condition_;
 
   std::shared_ptr<rclcpp::GuardCondition> shutdown_guard_condition_;
 
@@ -549,7 +549,7 @@ protected:
   spin_once_impl(std::chrono::nanoseconds timeout);
 
   typedef std::map<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr,
-      const rcl_guard_condition_t *,
+      const rclcpp::GuardCondition *,
       std::owner_less<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr>>
     WeakNodesToGuardConditionsMap;
 
--- a/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
+++ b/rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
@@ -102,7 +102,7 @@ public:
    * \throws std::runtime_error if it couldn't add guard condition to wait set
    */
   RCLCPP_PUBLIC
-  bool
+  void
   add_to_wait_set(rcl_wait_set_t * wait_set) override;
 
   RCLCPP_PUBLIC
@@ -328,7 +328,7 @@ private:
   WeakCallbackGroupsToNodesMap weak_groups_to_nodes_associated_with_executor_;
 
   typedef std::map<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr,
-      const rcl_guard_condition_t *,
+      const rclcpp::GuardCondition *,
       std::owner_less<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr>>
     WeakNodesToGuardConditionsMap;
   WeakNodesToGuardConditionsMap weak_nodes_to_guard_conditions_;
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
@@ -25,6 +25,7 @@
 
 #include "rcl/error_handling.h"
 
+#include "rclcpp/guard_condition.hpp"
 #include "rclcpp/qos.hpp"
 #include "rclcpp/type_support_decl.hpp"
 #include "rclcpp/waitable.hpp"
@@ -41,9 +42,10 @@ public:
 
   RCLCPP_PUBLIC
   SubscriptionIntraProcessBase(
+    rclcpp::Context::SharedPtr context,
     const std::string & topic_name,
     const rclcpp::QoS & qos_profile)
-  : topic_name_(topic_name), qos_profile_(qos_profile)
+  : gc_(context), topic_name_(topic_name), qos_profile_(qos_profile)
   {}
 
   virtual ~SubscriptionIntraProcessBase() = default;
@@ -53,7 +55,7 @@ public:
   get_number_of_ready_guard_conditions() {return 1;}
 
   RCLCPP_PUBLIC
-  bool
+  void
   add_to_wait_set(rcl_wait_set_t * wait_set);
 
   virtual bool
@@ -79,7 +81,7 @@ public:
 
 protected:
   std::recursive_mutex reentrant_mutex_;
-  rcl_guard_condition_t gc_;
+  rclcpp::GuardCondition gc_;
 
 private:
   virtual void
--- a/rclcpp/include/rclcpp/guard_condition.hpp
+++ b/rclcpp/include/rclcpp/guard_condition.hpp
@@ -47,7 +47,9 @@ public:
   RCLCPP_PUBLIC
   explicit GuardCondition(
     rclcpp::Context::SharedPtr context =
-    rclcpp::contexts::get_global_default_context());
+    rclcpp::contexts::get_global_default_context(),
+    rcl_guard_condition_options_t guard_condition_options =
+    rcl_guard_condition_get_default_options());
 
   RCLCPP_PUBLIC
   virtual
@@ -57,6 +59,11 @@ public:
   RCLCPP_PUBLIC
   rclcpp::Context::SharedPtr
   get_context() const;
+
+  /// Return the underlying rcl guard condition structure.
+  RCLCPP_PUBLIC
+  rcl_guard_condition_t &
+  get_rcl_guard_condition();
 
   /// Return the underlying rcl guard condition structure.
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/memory_strategy.hpp
@@ -64,9 +64,11 @@ public:
   virtual void clear_handles() = 0;
   virtual void remove_null_handles(rcl_wait_set_t * wait_set) = 0;
 
-  virtual void add_guard_condition(const rcl_guard_condition_t * guard_condition) = 0;
+  virtual void
+  add_guard_condition(const rclcpp::GuardCondition & guard_condition) = 0;
 
-  virtual void remove_guard_condition(const rcl_guard_condition_t * guard_condition) = 0;
+  virtual void
+  remove_guard_condition(const rclcpp::GuardCondition & guard_condition) = 0;
 
   virtual void
   get_next_subscription(
--- a/rclcpp/include/rclcpp/memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/memory_strategy.hpp
@@ -68,7 +68,7 @@ public:
   add_guard_condition(const rclcpp::GuardCondition & guard_condition) = 0;
 
   virtual void
-  remove_guard_condition(const rclcpp::GuardCondition & guard_condition) = 0;
+  remove_guard_condition(const rclcpp::GuardCondition * guard_condition) = 0;
 
   virtual void
   get_next_subscription(
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -114,12 +114,8 @@ public:
   get_associated_with_executor_atomic() override;
 
   RCLCPP_PUBLIC
-  rcl_guard_condition_t *
+  rclcpp::GuardCondition &
   get_notify_guard_condition() override;
-
-  RCLCPP_PUBLIC
-  std::unique_lock<std::recursive_mutex>
-  acquire_notify_guard_condition_lock() const override;
 
   RCLCPP_PUBLIC
   bool
@@ -149,7 +145,7 @@ private:
 
   /// Guard condition for notifying the Executor of changes to this node.
   mutable std::recursive_mutex notify_guard_condition_mutex_;
-  rcl_guard_condition_t notify_guard_condition_ = rcl_get_zero_initialized_guard_condition();
+  rclcpp::GuardCondition notify_guard_condition_;
   bool notify_guard_condition_is_valid_;
 };
 
--- a/rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
@@ -24,6 +24,7 @@
 
 #include "rclcpp/callback_group.hpp"
 #include "rclcpp/context.hpp"
+#include "rclcpp/guard_condition.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/visibility_control.hpp"
 
@@ -142,23 +143,16 @@ public:
   std::atomic_bool &
   get_associated_with_executor_atomic() = 0;
 
-  /// Return guard condition that should be notified when the internal node state changes.
+  /// Return a guard condition that should be notified when the internal node state changes.
   /**
    * For example, this should be notified when a publisher is added or removed.
    *
-   * \return the rcl_guard_condition_t if it is valid, else nullptr
+   * \return the GuardCondition if it is valid, else thow runtime error
    */
   RCLCPP_PUBLIC
   virtual
-  rcl_guard_condition_t *
+  rclcpp::GuardCondition &
   get_notify_guard_condition() = 0;
-
-  /// Acquire and return a scoped lock that protects the notify guard condition.
-  /** This should be used when triggering the notify guard condition. */
-  RCLCPP_PUBLIC
-  virtual
-  std::unique_lock<std::recursive_mutex>
-  acquire_notify_guard_condition_lock() const = 0;
 
   /// Return the default preference for using intra process communication.
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
@@ -72,10 +72,10 @@ public:
     guard_conditions_.push_back(&guard_condition);
   }
 
-  void remove_guard_condition(const rclcpp::GuardCondition & guard_condition) override
+  void remove_guard_condition(const rclcpp::GuardCondition * guard_condition) override
   {
     for (auto it = guard_conditions_.begin(); it != guard_conditions_.end(); ++it) {
-      if (*it == &guard_condition) {
+      if (*it == guard_condition) {
         guard_conditions_.erase(it);
         break;
       }
--- a/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
+++ b/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
@@ -21,6 +21,7 @@
 #include "rcl/allocator.h"
 
 #include "rclcpp/allocator/allocator_common.hpp"
+#include "rclcpp/detail/add_guard_condition_to_rcl_wait_set.hpp"
 #include "rclcpp/memory_strategy.hpp"
 #include "rclcpp/node.hpp"
 #include "rclcpp/visibility_control.hpp"
@@ -61,20 +62,20 @@ public:
     allocator_ = std::make_shared<VoidAlloc>();
   }
 
-  void add_guard_condition(const rcl_guard_condition_t * guard_condition) override
+  void add_guard_condition(const rclcpp::GuardCondition & guard_condition) override
   {
     for (const auto & existing_guard_condition : guard_conditions_) {
-      if (existing_guard_condition == guard_condition) {
+      if (existing_guard_condition == &guard_condition) {
         return;
       }
     }
-    guard_conditions_.push_back(guard_condition);
+    guard_conditions_.push_back(&guard_condition);
   }
 
-  void remove_guard_condition(const rcl_guard_condition_t * guard_condition) override
+  void remove_guard_condition(const rclcpp::GuardCondition & guard_condition) override
   {
     for (auto it = guard_conditions_.begin(); it != guard_conditions_.end(); ++it) {
-      if (*it == guard_condition) {
+      if (*it == &guard_condition) {
         guard_conditions_.erase(it);
         break;
       }
@@ -240,22 +241,11 @@ public:
     }
 
     for (auto guard_condition : guard_conditions_) {
-      if (rcl_wait_set_add_guard_condition(wait_set, guard_condition, NULL) != RCL_RET_OK) {
-        RCUTILS_LOG_ERROR_NAMED(
-          "rclcpp",
-          "Couldn't add guard_condition to wait set: %s",
-          rcl_get_error_string().str);
-        return false;
-      }
+      detail::add_guard_condition_to_rcl_wait_set(*wait_set, *guard_condition);
     }
 
     for (auto waitable : waitable_handles_) {
-      if (!waitable->add_to_wait_set(wait_set)) {
-        RCUTILS_LOG_ERROR_NAMED(
-          "rclcpp",
-          "Couldn't add waitable to wait set: %s", rcl_get_error_string().str);
-        return false;
-      }
+      waitable->add_to_wait_set(wait_set);
     }
     return true;
   }
@@ -509,7 +499,7 @@ private:
   using VectorRebind =
     std::vector<T, typename std::allocator_traits<Alloc>::template rebind_alloc<T>>;
 
-  VectorRebind<const rcl_guard_condition_t *> guard_conditions_;
+  VectorRebind<const rclcpp::GuardCondition *> guard_conditions_;
 
   VectorRebind<std::shared_ptr<const rcl_subscription_t>> subscription_handles_;
   VectorRebind<std::shared_ptr<const rcl_service_t>> service_handles_;
--- a/rclcpp/include/rclcpp/waitable.hpp
+++ b/rclcpp/include/rclcpp/waitable.hpp
@@ -103,12 +103,11 @@ public:
   /// Add the Waitable to a wait set.
   /**
    * \param[in] wait_set A handle to the wait set to add the Waitable to.
-   * \return `true` if the Waitable is added successfully, `false` otherwise.
    * \throws rclcpp::execptions::RCLError from rcl_wait_set_add_*()
    */
   RCLCPP_PUBLIC
   virtual
-  bool
+  void
   add_to_wait_set(rcl_wait_set_t * wait_set) = 0;
 
   /// Check if the Waitable is ready.
@@ -145,8 +144,7 @@ public:
    * ```cpp
    * // ... create a wait set and a Waitable
    * // Add the Waitable to the wait set
-   * bool add_ret = waitable.add_to_wait_set(wait_set);
-   * // ... error handling
+   * waitable.add_to_wait_set(wait_set);
    * // Wait
    * rcl_ret_t wait_ret = rcl_wait(wait_set);
    * // ... error handling
@@ -172,8 +170,7 @@ public:
    * ```cpp
    * // ... create a wait set and a Waitable
    * // Add the Waitable to the wait set
-   * bool add_ret = waitable.add_to_wait_set(wait_set);
-   * // ... error handling
+   * waitable.add_to_wait_set(wait_set);
    * // Wait
    * rcl_ret_t wait_ret = rcl_wait(wait_set);
    * // ... error handling
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -45,18 +45,12 @@ using rclcpp::FutureReturnCode;
 
 Executor::Executor(const rclcpp::ExecutorOptions & options)
 : spinning(false),
+  interrupt_guard_condition_(options.context),
   shutdown_guard_condition_(std::make_shared<rclcpp::GuardCondition>(options.context)),
   memory_strategy_(options.memory_strategy)
 {
   // Store the context for later use.
   context_ = options.context;
-
-  rcl_guard_condition_options_t guard_condition_options = rcl_guard_condition_get_default_options();
-  rcl_ret_t ret = rcl_guard_condition_init(
-    &interrupt_guard_condition_, context_->get_rcl_context().get(), guard_condition_options);
-  if (RCL_RET_OK != ret) {
-    throw_from_rcl_error(ret, "Failed to create interrupt guard condition in Executor constructor");
-  }
 
   shutdown_callback_handle_ = context_->add_on_shutdown_callback(
     [weak_gc = std::weak_ptr<rclcpp::GuardCondition>{shutdown_guard_condition_}]() {
@@ -68,13 +62,13 @@ Executor::Executor(const rclcpp::ExecutorOptions & options)
 
   // The number of guard conditions is always at least 2: 1 for the ctrl-c guard cond,
   // and one for the executor's guard cond (interrupt_guard_condition_)
-  memory_strategy_->add_guard_condition(&shutdown_guard_condition_->get_rcl_guard_condition());
+  memory_strategy_->add_guard_condition(*shutdown_guard_condition_.get());
 
   // Put the executor's guard condition in
-  memory_strategy_->add_guard_condition(&interrupt_guard_condition_);
+  memory_strategy_->add_guard_condition(interrupt_guard_condition_);
   rcl_allocator_t allocator = memory_strategy_->get_allocator();
 
-  ret = rcl_wait_set_init(
+  rcl_ret_t ret = rcl_wait_set_init(
     &wait_set_,
     0, 2, 0, 0, 0, 0,
     context_->get_rcl_context().get(),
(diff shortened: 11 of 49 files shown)


rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
```diff
@@ -191,17 +191,17 @@ TYPED_TEST(TestAddCallbackGroupsToExecutor, add_callback_groups_after_add_node_t
   rclcpp::CallbackGroup::SharedPtr cb_grp = node->create_callback_group(
     rclcpp::CallbackGroupType::MutuallyExclusive);
   rclcpp::TimerBase::SharedPtr timer_ = node->create_wall_timer(
-    2s, timer_callback, cb_grp);
+    10ms, timer_callback, cb_grp);
```
irobot-ros, line 194: why this change?

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -2556,6 +2556,7 @@ TEST_F(TestNode, get_publishers_subscriptions_info_by_topic) {
   };
   rclcpp::QoS qos = rclcpp::QoS(qos_initialization, rmw_qos_profile_default);
   auto publisher = node->create_publisher<test_msgs::msg::BasicTypes>(topic_name, qos);
+  std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
irobot-ros, line 2559: this should be all synchronous, why is necessary to wait?

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
mauropasse: This test takes 4 seconds for each executor (16 seconds when testing 4 different executors) due the period of the timer, which doesn't have any impact on the output of the test.

rclcpp/test/rclcpp/test_node.cpp
mauropasse: I was getting random failures in this test. I didn't have it anymore when adding this sleep.

clalancette: We've seen this elsewhere. The problem is that depending on the backend DDS implementation, creating a publisher and then immediately trying to get info on it may or may not work. That's because the backend may create the topic asynchronously.

You'll probably want a solution similar to #1606.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -104,8 +104,8 @@ class NodeBase : public NodeBaseInterface
   get_associated_with_executor_atomic() override;
 
   RCLCPP_PUBLIC
-  rcl_guard_condition_t *
-  get_notify_guard_condition() override;
+  rclcpp::GuardCondition::SharedPtr
+  get_notify_guard_condition() const override;
```
wjwwood, line 108: If you add `const` the return type should be `const` too, either `shared_ptr<const rclcpp::GuardConditioon>` or, as I suggested, `const rclcpp::GuardCondition &`.

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
```diff
@@ -138,12 +139,12 @@ class NodeBaseInterface
...
    *
-   * \return the rcl_guard_condition_t if it is valid, else nullptr
+   * \return the GuardCondition::SharedPtr if it is valid, else nullptr
    */
   RCLCPP_PUBLIC
   virtual
-  rcl_guard_condition_t *
-  get_notify_guard_condition() = 0;
+  rclcpp::GuardCondition::SharedPtr
+  get_notify_guard_condition() const = 0;
```
wjwwood, line 147: Same comments apply here as from `NodeBase`.

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
```diff
@@ -170,6 +171,11 @@ class NodeBaseInterface
   std::string
   resolve_topic_or_service_name(
     const std::string & name, bool is_service, bool only_expand = false) const = 0;
+
+  RCLCPP_PUBLIC
+  virtual
+  void
+  trigger_notify_guard_condition() const = 0;
```
wjwwood, line 178: This needs documentation.

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -504,7 +504,7 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
   using VectorRebind =
     std::vector<T, typename std::allocator_traits<Alloc>::template rebind_alloc<T>>;
 
-  VectorRebind<const rcl_guard_condition_t *> guard_conditions_;
+  VectorRebind<GuardCondition::SharedPtr> guard_conditions_;
```
wjwwood, line 507: Why does this need to be a shared pointer now? Previously it did not share ownership, is that now necessary?

rclcpp/src/rclcpp/guard_condition.cpp
```diff
@@ -77,4 +78,15 @@ GuardCondition::exchange_in_use_by_wait_set_state(bool in_use_state)
...
 }
 
+bool
+GuardCondition::add_to_wait_set(rcl_wait_set_t * wait_set)
+{
+  rcl_ret_t ret = rcl_wait_set_add_guard_condition(wait_set, &this->rcl_guard_condition_, NULL);
+  if (RCL_RET_OK != ret) {
+    rclcpp::exceptions::throw_from_rcl_error(
+      ret, "failed to add guard condition to wait set");
+    return false;
```
wjwwood, line 88: This line of code is never reached... It seems odd (maybe an accident) to mix return type and exceptions.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
wjwwood: I don't know why they were set to this before hand, but it might make the tests flaky. It really shouldn't, but I'm just throwing it out there, because it does happen if the test isn't constructed properly.

rclcpp/test/rclcpp/test_node.cpp
wjwwood: Agreed, I'd say remove this as it is (AFAIK) unrelated to this change, and also because it isn't a scalable solution.

wjwwood (changes requested):
This pull request is definitely an improvement in a good direction.

I had concern that we're just using shared pointer everywhere, when we don't really need to, and I pointed that out in places.

Also, anywhere the API is changed in a way that would break existing code calling it in one of our `.hpp` files, we need to consider deprecating it rather than just changing it. It isn't always possible, desirable, or necessary, but we need to think about each case and justify it if we don't deprecate things.

mauropasse:
Last commit:
- rebased master.
- Now using `rclcpp::GuardCondition` instead of shared pointers of it.
- Added documentation to `GuardCondition::add_to_wait_set` but not sure if is enough.
- Reverted changes in tests not related to the use of `rclcpp::GuardCondition`.
- Created new api to get the node's `rclcpp::GuardCondition` (I had to slightly modify the original `NodeBase::get_notify_guard_condition`)
Let me know if I'm missing something

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -129,15 +117,9 @@ Executor::~Executor()
...
-  if (rcl_guard_condition_fini(&interrupt_guard_condition_) != RCL_RET_OK) {
-    RCUTILS_LOG_ERROR_NAMED(
-      "rclcpp",
-      "failed to destroy guard condition: %s", rcl_get_error_string().str);
-    rcl_reset_error();
-  }
   // Remove and release the sigint guard condition
-  memory_strategy_->remove_guard_condition(&shutdown_guard_condition_->get_rcl_guard_condition());
+  memory_strategy_->remove_guard_condition(shutdown_guard_condition_.get());
+  memory_strategy_->remove_guard_condition(&interrupt_guard_condition_);
```
mauropasse, line 122: This line is new, before the `interrupt_guard_condition_` wasn't being removed from the `memory_strategy_`

irobot-ros:
Hi, any updates on this PR?
We think that we have addressed the review comments

rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
```diff
@@ -39,8 +40,10 @@ class SubscriptionIntraProcessBase : public rclcpp::Waitable
   RCLCPP_SMART_PTR_ALIASES_ONLY(SubscriptionIntraProcessBase)
 
   RCLCPP_PUBLIC
-  SubscriptionIntraProcessBase(const std::string & topic_name, rmw_qos_profile_t qos_profile)
-  : topic_name_(topic_name), qos_profile_(qos_profile)
+  SubscriptionIntraProcessBase(
+    rclcpp::Context::SharedPtr context,
+    const std::string & topic_name, rmw_qos_profile_t qos_profile)
```
wjwwood, line 45:
```suggestion
    const std::string & topic_name,
    rmw_qos_profile_t qos_profile)
```

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -104,9 +104,13 @@ class NodeBase : public NodeBaseInterface
   get_associated_with_executor_atomic() override;
 
   RCLCPP_PUBLIC
-  rcl_guard_condition_t *
+  const rcl_guard_condition_t *
```
wjwwood, line 107: Why is this needed? This technically breaks API.

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
```diff
@@ -142,9 +143,20 @@ class NodeBaseInterface
    */
   RCLCPP_PUBLIC
   virtual
-  rcl_guard_condition_t *
+  const rcl_guard_condition_t *
```
wjwwood, line 146: Same here, it shouldn't be necessary if you can return a non-const pointer to the rclcpp guard condition.

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
```diff
@@ -142,9 +143,20 @@ class NodeBaseInterface
...
   get_notify_guard_condition() = 0;
 
+  /// Return guard condition that should be notified when the internal node state changes.
```
wjwwood, line 149:
```suggestion
  /// Return a guard condition that should be notified when the internal node state changes.
```

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -548,7 +548,7 @@ class Executor
   spin_once_impl(std::chrono::nanoseconds timeout);
 
   typedef std::map<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr,
-      const rcl_guard_condition_t *,
+      const rclcpp::GuardCondition *,
```
wjwwood, line 551: As long as we're breaking API, let's change this to a reference (where ever it makes sense):

```suggestion
      const rclcpp::GuardCondition &,
```

This prevents the possibility of passing a nullptr into these methods that took pointers to guard conditions in the past. It puts the responsibility onto the caller of the methods to check this.

I think this is way more idiomatic C++, and I'd prefer to do it now, as we upgrade to C++ classes from C structs, instead of doing it in two steps, breaking API each time.

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
```diff
@@ -65,7 +65,7 @@ class StaticExecutorEntitiesCollector final
   init(
     rcl_wait_set_t * p_wait_set,
     rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy,
-    rcl_guard_condition_t * executor_guard_condition);
+    rclcpp::GuardCondition * executor_guard_condition);
```
wjwwood, line 68: Another example of this here, but there are several more, so I won't point them each out.

wjwwood:
This needs to be rebased, due to merge conflicts, and I left some comments with requested changes, but otherwise it looks really close. Thanks for following up on it.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
mauropasse: I removed this change in dd4ab15. For that I had to add a non-const version of `get_rcl_guard_condition` API from `GuardCondition` class

rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
mauropasse: Done in b363b21

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
mauropasse: Done in b363b21

rclcpp/include/rclcpp/executor.hpp
mauropasse: With my last commit, the API is not being broken anymore. Anyway I tried this change, but having a reference in a map seems not behave nicely, as references must be init directly and can't be copied, I was getting a bunch of strange errors.

wjwwood: Ah sorry, maybe the map should hold the pointer still, but in places where we pass by pointer, we could pass by reference instead (only in function/method interfaces).

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
wjwwood: For example, changing this one, but not the map type, would be good I think.

rclcpp/include/rclcpp/executor.hpp
wjwwood: > the API is not being broken anymore

Also, we are still breaking API since we're changing the type from `rcl_guard_condition_t` to `rclcpp::GuardCondition` in several places in the API. Which is fine, but anyway it is breaking.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -107,6 +107,10 @@ class NodeBase : public NodeBaseInterface
   rcl_guard_condition_t *
   get_notify_guard_condition() override;
 
+  RCLCPP_PUBLIC
+  rclcpp::GuardCondition *
```
wjwwood, line 111: This is another case where I think we should use a reference rather than a pointer, since it should never be nullptr, right?

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -107,6 +107,10 @@ class NodeBase : public NodeBaseInterface
...
+  get_notify_rclcpp_guard_condition() override;
```
wjwwood, line 112: Normally, I'd agree with adding a new function with the new return type, but since we're changing (in several places) public API to take an rclcpp guard condition instead of an rcl one, maybe we should just update the existing `get_notify_guard_condition()` to return the rclcpp guard condition instead.

At the very least we should deprecate `rcl_guard_condition_t * get_notify_guard_condition()` if we don't just change the return type.

Instead, users can get the rclcpp guard condition and then get the rcl one from that, if they still need it.

rclcpp/include/rclcpp/guard_condition.hpp
```diff
@@ -89,6 +95,15 @@ class GuardCondition
...
   exchange_in_use_by_wait_set_state(bool in_use_state);
 
+  /// Adds the guard condition to a waitset
+  /**
+   * This function is thread-safe.
+   * \param[in] wait_set pointer to a wait set where to add the guard condition
+   */
+  RCLCPP_PUBLIC
+  void
+  add_to_wait_set(rcl_wait_set_t * wait_set) const;
```
wjwwood, line 105: I'm reluctant to add new public API that uses the rcl type here. Ideally, we'd update it to work with the `rclcpp::WaitSet`, as I've said elsewhere.

Since that it out of scope, if we still want this function, I'd like to put it into a free function, in the detail namespace, so it is not a public API, and therefore we can remove it later when we roll out use of the rclcpp WaitSet.

So something like this:

```c++
// rclcpp/include/rclcpp/detail/add_guard_condition_to_rcl_wait_set.hpp
namespace rclcpp
{
namespace detail
{
RCLCPP_PUBLIC
void
add_guard_condition_to_rcl_wait_set(rcl_wait_set_t & wait_set, const rclcpp::GuardCondition & guard_condition);
}  // namespace detail
}  // namespace rclcpp

File: rclcpp/include/rclcpp/memory_strategy.hpp [hunk 18]
```diff
@@ -64,9 +64,11 @@ class RCLCPP_PUBLIC MemoryStrategy
   virtual void clear_handles() = 0;
   virtual void remove_null_handles(rcl_wait_set_t * wait_set) = 0;
 
-  virtual void add_guard_condition(const rcl_guard_condition_t * guard_condition) = 0;
+  virtual void
+  add_guard_condition(const rclcpp::GuardCondition * guard_condition) = 0;
```
wjwwood, line 68: This is another place where we could take a reference instead of a pointer.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -310,3 +297,10 @@ NodeBase::resolve_topic_or_service_name(
   allocator.deallocate(output_cstr, allocator.state);
   return output;
 }
+
+void
+NodeBase::trigger_notify_guard_condition()
+{
+  std::unique_lock<std::recursive_mutex> lock(notify_guard_condition_mutex_);
+  notify_guard_condition_.trigger();
+}
```
wjwwood, line 306: With the introduction of this function, maybe we don't need the `acquire_notify_guard_condition_lock` any longer and should deprecate it.

I can't find why we had it this way, it seemed intended to be used like this:

```c++
  {
    auto notify_condition_lock = node_base_->acquire_notify_guard_condition_lock();
    // some other "critical section" stuff
    rcl_ret_t ret = rcl_trigger_guard_condition(node_base_->get_notify_guard_condition());
    if (RCL_RET_OK != ret) {
      throw_from_rcl_error(ret, "failed to trigger notify guard condition");
    }
  }
```

But it doesn't appear to be used like that anywhere. At least not anymore.

And now I'm wondering if the lock is needed at all. I think triggering a guard condition should be thread-safe...

rclcpp/test/rclcpp/node_interfaces/test_node_services.cpp
```diff
@@ -103,7 +103,7 @@ TEST_F(TestNodeService, add_service_rcl_trigger_guard_condition_error)
     "lib:rclcpp", rcl_trigger_guard_condition, RCL_RET_ERROR);
   RCLCPP_EXPECT_THROW_EQ(
     node_services->add_service(service, callback_group),
-    std::runtime_error("Failed to notify wait set on service creation: error not set"));
```
wjwwood, line 106: This seems like a step backwards... this is, perhaps, another reason to not introduce the `trigger_notify_guard_condition()` method?

Or it should have an error message argument to augment this error message. Or maybe all the callers should catch and rethrow. Either way, it would be better if these errors didn't change in this way.

rclcpp/include/rclcpp/executor.hpp
mauropasse: Done in ccc78c5

rclcpp/include/rclcpp/executors/static_executor_entities_collector.hpp
mauropasse: Done in ccc78c5

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
mauropasse: Done in ccc78c5

rclcpp/include/rclcpp/memory_strategy.hpp
mauropasse: Done in ccc78c5

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
mauropasse: > maybe we should just update the existing get_notify_guard_condition() to return the rclcpp guard condition instead.

Done in 2b2cda5

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mauropasse: > With the introduction of this function, maybe we don't need the acquire_notify_guard_condition_lock any longer and should deprecate it.

> I'm wondering if the lock is needed at all. I think triggering a guard condition should be thread-safe

I also don't think the lock is needed, as triggering the guard condition is thread-safe.
Done in 82629b7

rclcpp/include/rclcpp/guard_condition.hpp
mauropasse: Done in 0d6e8a9

rclcpp/test/rclcpp/node_interfaces/test_node_services.cpp
mauropasse: I'm not sure of the correct or best way to implement the augment of the error message. Should be passed as argument when calling `guard_condition->trigger("some error message")` ?

Regarding the introduction of `trigger_notify_guard_condition`, I can always remove it, so first we have to get the node's notify guard condition and then trigger it.

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
```diff
@@ -61,24 +62,24 @@ class AllocatorMemoryStrategy : public memory_strategy::MemoryStrategy
...
-  void remove_guard_condition(const rcl_guard_condition_t * guard_condition) override
+  void add_guard_condition(const rclcpp::GuardCondition::SharedPtr guard_condition) override
   {
-    for (auto it = guard_conditions_.begin(); it != guard_conditions_.end(); ++it) {
-      if (*it == guard_condition) {
-        guard_conditions_.erase(it);
-        break;
-      }
-    }
+    add_guard_condition(guard_condition.get());
```
wjwwood, line 72: Since this method is not retaining the ownership of the shared pointer (or otherwise using it), can we just get rid of these signatures that take the shared pointer and leave just the ones that take a reference?

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -699,9 +668,9 @@ Executor::wait_for_work(std::chrono::nanoseconds timeout)
           invalid_group_ptrs.push_back(weak_group_ptr);
           auto node_guard_pair = weak_nodes_to_guard_conditions_.find(weak_node_ptr);
           if (node_guard_pair != weak_nodes_to_guard_conditions_.end()) {
-            auto guard_condition = node_guard_pair->second;
+            const auto & guard_condition = node_guard_pair->second;
             weak_nodes_to_guard_conditions_.erase(weak_node_ptr);
-            memory_strategy_->remove_guard_condition(guard_condition);
+            memory_strategy_->remove_guard_condition(*guard_condition);
```
wjwwood, line 673: Is it possible to have this be nullptr unexpectedly? Should we check it?

rclcpp/src/rclcpp/subscription_intra_process_base.cpp
```diff
@@ -21,8 +22,9 @@ SubscriptionIntraProcessBase::add_to_wait_set(rcl_wait_set_t * wait_set)
 {
   std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
 
-  rcl_ret_t ret = rcl_wait_set_add_guard_condition(wait_set, &gc_, NULL);
-  return RCL_RET_OK == ret;
+  detail::add_guard_condition_to_rcl_wait_set(*wait_set, gc_);
+
+  return true;
```
wjwwood, line 27: Should we just change this to not return anything?

rclcpp/test/rclcpp/executors/test_executors.cpp
```diff
@@ -352,40 +354,18 @@ TYPED_TEST(TestExecutors, testSpinUntilFutureCompleteWithTimeout) {
...
-  }
+  TestWaitable() = default;
 
   bool
   add_to_wait_set(rcl_wait_set_t * wait_set) override
   {
-    rcl_ret_t ret = rcl_wait_set_add_guard_condition(wait_set, &gc_, NULL);
-    return RCL_RET_OK == ret;
+    rclcpp::detail::add_guard_condition_to_rcl_wait_set(*wait_set, gc_);
+    return true;
```
wjwwood, line 363: Same here, should we just not return anything?

rclcpp/test/rclcpp/node_interfaces/test_node_services.cpp
wjwwood: I would say we should catch the throw from trigger and rethrow with more information.

wjwwood (commented):
A few more comments and suggestions, but this is really close!

Thanks for your patience in iterating on this.

rclcpp/include/rclcpp/guard_condition.hpp
```diff
@@ -59,6 +61,10 @@ class GuardCondition
   get_context() const;
 
   /// Return the underlying rcl guard condition structure.
+  RCLCPP_PUBLIC
+  rcl_guard_condition_t &
+  get_rcl_guard_condition();
+
   RCLCPP_PUBLIC
```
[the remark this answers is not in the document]
mauropasse: Done in 7a4b2a1

rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp
mauropasse: Done in 7a4b2a1

rclcpp/src/rclcpp/executor.cpp
mauropasse: Here we're in the situation in which a node could be expired: `weak_node_ptr.expired()`
If that the case the node's guard condition has been destructed, so the dereference `*guard_condition` is wrong since is now a dangling pointer. It could be `nullptr` or not, either way is dangerous.
The memory strategy should hold weak pointers of the guard condition, or have a map `{WeakNode, GuardCondition}` as the executor does, or share the same map. 
So the node's GuardCondition should be created as SharedPtr, and stored as WeakPtr in executor/memory_strategy. 
What do you think?

irobot-ros: What about storing only WeakNodes? I assume that regardless of the approach, before using the guard condition we will still have to check that the node has not gone out of scope, so at that point it may be worth directly calling `get_notify_guard_condition` every time

rclcpp/src/rclcpp/subscription_intra_process_base.cpp
mauropasse: Done in 34f8ed0

rclcpp/test/rclcpp/executors/test_executors.cpp
mauropasse: Done in 34f8ed0

rclcpp/test/rclcpp/node_interfaces/test_node_services.cpp
mauropasse: Done in 765257c

rclcpp/src/rclcpp/executor.cpp
mauropasse: Commit 9da50f1 fixes this problem (accessing an out-of-scope `GuardCondition` pointer) by restoring `remove_guard_condition` original way of taking directly a pointer to the guard condition to remove.
This way we also avoid to take the address of a `GuardCondition` inside the API, which might have gone out of scope.
I have rebased this PR with master. I think all the comments have been addressed now.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -265,20 +242,14 @@ NodeBase::get_associated_with_executor_atomic()
...
 }
 
-rcl_guard_condition_t *
+rclcpp::GuardCondition &
 NodeBase::get_notify_guard_condition()
 {
   std::lock_guard<std::recursive_mutex> notify_condition_lock(notify_guard_condition_mutex_);
   if (!notify_guard_condition_is_valid_) {
-    return nullptr;
+    throw std::runtime_error("Trying to get invalid notify guard condition");
```
wjwwood, line 250:
```suggestion
    throw std::runtime_error("failed to get notify guard condition because it is invalid");
```

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -348,12 +348,12 @@ NodeGraph::notify_graph_change()
...
-    rcl_ret_t ret = rcl_trigger_guard_condition(node_base_->get_notify_guard_condition());
-    if (RCL_RET_OK != ret) {
-      throw_from_rcl_error(ret, "failed to trigger notify guard condition");
-    }
+  auto & node_gc = node_base_->get_notify_guard_condition();
+  try {
+    node_gc.trigger();
+  } catch (const rclcpp::exceptions::RCLError & ex) {
+    throw std::runtime_error(
+            std::string("Failed to notify wait set on graph change: ") + ex.what());
```
wjwwood, line 356:
```suggestion
            std::string("failed to notify wait set on graph change: ") + ex.what());
```

rclcpp/src/rclcpp/node_interfaces/node_services.cpp
```diff
@@ -41,14 +41,12 @@ NodeServices::add_service(
...
-              std::string("Failed to notify wait set on service creation: ") +
-              rmw_get_error_string().str
-      );
-    }
+  auto & node_gc = node_base_->get_notify_guard_condition();
+  try {
+    node_gc.trigger();
+  } catch (const rclcpp::exceptions::RCLError & ex) {
+    throw std::runtime_error(
+            std::string("Failed to notify wait set on service creation: ") + ex.what());
```
wjwwood, line 49: I know it was like this before, but the predominate style elsewhere is lower case:

```suggestion
            std::string("failed to notify wait set on service creation: ") + ex.what());
```

rclcpp/src/rclcpp/node_interfaces/node_services.cpp
```diff
@@ -68,14 +66,12 @@ NodeServices::add_client(
...
-              std::string("Failed to notify wait set on client creation: ") +
-              rmw_get_error_string().str
-      );
-    }
+  auto & node_gc = node_base_->get_notify_guard_condition();
+  try {
+    node_gc.trigger();
+  } catch (const rclcpp::exceptions::RCLError & ex) {
+    throw std::runtime_error(
+            std::string("Failed to notify wait set on client creation: ") + ex.what());
```
wjwwood, line 74:
```suggestion
            std::string("failed to notify wait set on client creation: ") + ex.what());
```

rclcpp/src/rclcpp/node_interfaces/node_timers.cpp
```diff
@@ -41,11 +41,15 @@ NodeTimers::add_timer(
...
-  if (rcl_trigger_guard_condition(node_base_->get_notify_guard_condition()) != RCL_RET_OK) {
+
+  auto & node_gc = node_base_->get_notify_guard_condition();
+  try {
+    node_gc.trigger();
+  } catch (const rclcpp::exceptions::RCLError & ex) {
     throw std::runtime_error(
-            std::string("Failed to notify wait set on timer creation: ") +
-            rmw_get_error_string().str);
+            std::string("Failed to notify wait set on timer creation: ") + ex.what());
```
wjwwood, line 50:
```suggestion
            std::string("failed to notify wait set on timer creation: ") + ex.what());
```

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
```diff
@@ -60,13 +60,12 @@ NodeTopics::add_publisher(
...
-      throw std::runtime_error(
-              std::string("Failed to notify wait set on publisher creation: ") +
-              rmw_get_error_string().str);
-    }
+  auto & node_gc = node_base_->get_notify_guard_condition();
+  try {
+    node_gc.trigger();
+  } catch (const rclcpp::exceptions::RCLError & ex) {
+    throw std::runtime_error(
+            std::string("Failed to notify wait set on publisher creation: ") + ex.what());
```
wjwwood, line 68:
```suggestion
            std::string("failed to notify wait set on publisher creation: ") + ex.what());
```

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
```diff
@@ -108,13 +107,12 @@ NodeTopics::add_subscription(
...
-    if (ret != RCL_RET_OK) {
-      using rclcpp::exceptions::throw_from_rcl_error;
-      throw_from_rcl_error(ret, "failed to notify wait set on subscription creation");
-    }
+  auto & node_gc = node_base_->get_notify_guard_condition();
+  try {
+    node_gc.trigger();
+  } catch (const rclcpp::exceptions::RCLError & ex) {
+    throw std::runtime_error(
+            std::string("Failed to notify wait set on subscription creation: ") + ex.what());
```
wjwwood, line 115:
```suggestion
            std::string("failed to notify wait set on subscription creation: ") + ex.what());
```

rclcpp/src/rclcpp/node_interfaces/node_waitables.cpp
```diff
@@ -41,14 +41,12 @@ NodeWaitables::add_waitable(
...
-              std::string("Failed to notify wait set on waitable creation: ") +
-              rmw_get_error_string().str
-      );
-    }
+  auto & node_gc = node_base_->get_notify_guard_condition();
+  try {
+    node_gc.trigger();
+  } catch (const rclcpp::exceptions::RCLError & ex) {
+    throw std::runtime_error(
+            std::string("Failed to notify wait set on waitable creation: ") + ex.what());
```
wjwwood, line 49:
```suggestion
            std::string("failed to notify wait set on waitable creation: ") + ex.what());
```

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -326,9 +326,9 @@ TEST_F(TestNodeGraph, notify_graph_change_rcl_error)
 {
   auto mock = mocking_utils::patch_and_return(
     "lib:rclcpp", rcl_trigger_guard_condition, RCL_RET_ERROR);
-  EXPECT_THROW(
+  RCLCPP_EXPECT_THROW_EQ(
     node()->get_node_graph_interface()->notify_graph_change(),
-    rclcpp::exceptions::RCLError);
+    std::runtime_error("Failed to notify wait set on graph change: error not set"));
```
wjwwood, line 331: This seems odd... why is it "error not set"?

wjwwood: I guess it was like that before... but I'm not sure.

wjwwood (commented):
Sorry for the long delay. Had a few more comments/questions. But the changes that you made in response to my last review looked good.

rclcpp/include/rclcpp/detail/add_guard_condition_to_rcl_wait_set.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include "rclcpp/guard_condition.hpp"
+
+namespace rclcpp
+{
+namespace detail
+{
+
+/// Adds the guard condition to a waitset
+/**
+ * This function is thread-safe.
```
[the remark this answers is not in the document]
mauropasse: Yes you are right, there could be data races on the `wait_set`. I've removed the statement.

rclcpp/src/rclcpp/node_interfaces/node_waitables.cpp
mauropasse: Done in f99baec

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
mauropasse: Done in f99baec

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
mauropasse: Done in f99baec

rclcpp/src/rclcpp/node_interfaces/node_timers.cpp
mauropasse: Done in f99baec

rclcpp/src/rclcpp/node_interfaces/node_services.cpp
mauropasse: Done in f99baec

rclcpp/src/rclcpp/node_interfaces/node_services.cpp
mauropasse: Done in f99baec

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
mauropasse: Done in f99baec

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mauropasse: Done in f99baec

mauropasse:
@wjwwood I think I fixed the test, this PR is needed by `test_rclcpp`: https://github.com/ros2/system_tests/pull/488

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -114,13 +114,9 @@ class NodeBase : public NodeBaseInterface
   get_associated_with_executor_atomic() override;
 
   RCLCPP_PUBLIC
-  rcl_guard_condition_t *
+  rclcpp::GuardCondition &
   get_notify_guard_condition() override;
```
wjwwood, line 118: This API also should be mentioned in the notice we put in the Humble release page. It is even less likely to be used outside of the core, but still it should be mentioned.

wjwwood (approved):
lgtm, just need to sort out what to do/document for the API  breaks.

I'll run CI again.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15546/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10230/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13213/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15757/)

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
mauropasse: Done in https://github.com/ros2/ros2_documentation/pull/2086

rclcpp/include/rclcpp/waitable.hpp
```diff
@@ -103,12 +103,11 @@ class Waitable
...
   /**
    * \param[in] wait_set A handle to the wait set to add the Waitable to.
-   * \return `true` if the Waitable is added successfully, `false` otherwise.
    * \throws rclcpp::execptions::RCLError from rcl_wait_set_add_*()
    */
   RCLCPP_PUBLIC
   virtual
-  bool
+  void
   add_to_wait_set(rcl_wait_set_t * wait_set) = 0;
```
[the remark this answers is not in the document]
mauropasse: Done in https://github.com/ros2/ros2_documentation/pull/2086

mauropasse:
Just in case the `rclcpp_test` errors I see in the CI about `test_two_service_calls__rmw_*`, are not related with changes from this PR. I can reproduce them locally using the master branch.
Anything missing from this PR @wjwwood ?

wjwwood:
@mergifyio rebase

PR #1622, opened 2021-04-06
remove deprecated code which was deprecated in foxy and should be removed in galactic

wjwwood:
I went through all mentions of `deprecated` in rclcpp and checked if they existed on the foxy branch and generally investigated each.

Since they already had deprecation warnings, downstream stuff really should not be broken, but CI will tell.

Someone please check my work.


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14258/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9050/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11942/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14358/)

cottsay:
Is the PR build failure expected?

wjwwood:
Yes, it's because master rclcpp depends on unreleased changes in rcl/rmw.

Karsten1987 (approved):
makes sense to me - given green CI

rclcpp/test/rclcpp/test_init.cpp
```diff
@@ -1,66 +0,0 @@
-// Copyright 2018 Open Source Robotics Foundation, Inc.
```
Karsten1987, line 1: I didn't really dig into this, but do we have an equivalent test for this?

audrow (approved):
LGTM!

rclcpp/test/rclcpp/test_init.cpp
wjwwood: I assume so, our coverage should have taken care of that. `rclcpp::ok()` is the alternative.

wjwwood: `rclcpp::ok()` is covered in this file:

https://github.com/ros2/rclcpp/blob/cba6f20988758128377751de85dbc8351aeebefc/rclcpp/test/rclcpp/test_utilities.cpp

Issue #1008, opened 2020-03-01
test_multi_threaded_executor timer_over_take is flaky due to timer jitter


brawner:
This seems to be a reoccurring flaky test, as a test introduced in #383 has been relaxed a couple of times (#501, and #907), but the test may still fail on different platforms. The measured period between timer callbacks is supposed to be 0.1s, but the linux test result below shows that the period could be less than half of that. While the test is not meant to measure timer jitter, the results suggest that the timer jitter cannot be reliably bounded.

See:
https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1780/testReport/(root)/projectroot/test_multi_threaded_executor/
https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/1866/testReport/(root)/projectroot/test_multi_threaded_executor/
https://ci.ros2.org/view/nightly/job/nightly_windows-container_repeated/6/testReport/(root)/projectroot/test_multi_threaded_executor/

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2020-03-18/13313/1

chapulina:
I think this may be a duplicate of #751.

This test is still flaky one year later. Here's the failure 

```
[ RUN      ] TestMultiThreadedExecutor.timer_over_take
/Users/osrf/jenkins-agent/workspace/nightly_osx_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp:91: Failure
Expected: (diff) > (PERIOD - TOLERANCE), actual: 0.732793 vs 0.75
[  FAILED  ] TestMultiThreadedExecutor.timer_over_take (4201 ms)
```

On #751, there was a suggestion to "remove the upper bound condition on the test", which I think may refer exactly to the offending expectation, and @clalancette agreed at the time (2 years ago!) that it was a good idea. 

---

https://github.com/osrf/buildfarmer/issues/161

clalancette:
Actually, I don't think we have the upper bound anymore.  I'm going to close out #751.

[...]

PR #1624, opened 2021-04-06
Mark test_multi_threaded_executor as xfail

ivanpauno:
Related to https://github.com/ros2/rclcpp/issues/1008.

This test has been failing since a while ago, and doesn't seem to be easy to solve.
Marking it as xfail until a solution is found to avoid confusions.

---

What I think is happening is that the executor is being slow to execute the timer callback sometimes, and then the measured period is sometimes smaller than expected.

e.g.: period = 1000ms
timer ready at t1
callback executed at t1 + 500ms (executor delayed to schedule the callback)
timer ready again at t1 + 1000ms
callback executed at t1 + 1100ms
measured period of 1100ms - 500ms = 600ms

I remember experiencing something like that a lot ago, though I should try again to confirm.

### DIFF
--- a/rclcpp/test/rclcpp/CMakeLists.txt
+++ b/rclcpp/test/rclcpp/CMakeLists.txt
@@ -593,6 +593,7 @@ ament_add_gtest(test_multi_threaded_executor executors/test_multi_threaded_execu
 if(TARGET test_multi_threaded_executor)
   ament_target_dependencies(test_multi_threaded_executor
     "rcl")
+  ament_add_test_label(test_multi_threaded_executor xfail)
   target_link_libraries(test_multi_threaded_executor ${PROJECT_NAME})
 endif()
 


hidmic (approved):
> What I think is happening is that the executor is being slow to execute the timer callback sometimes, and then the measured period is sometimes smaller than expected.

That's quite plausible. I'm not sure we can do anything to prevent that, though perhaps we can reduce that latency. A test that's not this sensitive to time would be best e.g. using larger periods or asserts on timer period statistics.

clalancette:
I've been reluctant to add more xfail tests, because right now is the period where we are trying to get rid of them.

What I'll say here is that if we don't solve this issue during the Galactic stabilization period, then I'm OK with marking it as xfail.  But I do think this is one of the issues we should take a deeper look on in the next couple of weeks.  So I'd say hold off on merging this PR until then.

fujitatomoya:
> callback executed at t1 + 500ms (executor delayed to schedule the callback)

this also should be detected as error with upper bound?

https://github.com/ros2/rclcpp/blob/41fedb7beb123dee01b9e54e3579270203cd9404/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp#L74-L76

i do not really see that is `reasonable`... i think it would be nice to detect the problem in the 1st place.

but this is meant to be, based on https://github.com/ros2/rclcpp/issues/904 and https://github.com/ros2/rclcpp/pull/907 to relax the test failure.

> I remember experiencing something like that a lot ago, though I should try again to confirm.

👍 it would be nice to see if the problem is with executor or cyclic accuracy.

fujitatomoya (approved):

clalancette:
@ivanpauno Excellent, this is some great investigation.  Thanks for digging in here.

> More info: The "timer call" is [done by the executor](https://github.com/ros2/rclcpp/blob/41fedb7beb123dee01b9e54e3579270203cd9404/rclcpp/include/rclcpp/timer.hpp#L201), for some reason the executor is like "blocked" for a time and executes the timer way off the mark.

So I think the real question to answer here is why the executor is blocked.  There are 2 separate reasons I can think of:

1. The executor itself is blocked internally.  If this is the case, it seems like a bug in the executor that we should probably investigate.
2. The OS is just really slow to reschedule the process back in sometimes.  If this is the case, then it isn't the fault of any of our code and we can fix the test up to be more robust to this kind of problem.

Unfortunately, I'm not quite sure how to tell if it is case 1 or case 2.

ivanpauno:
> So I think the real question to answer here is why the executor is blocked. There are 2 separate reasons I can think of:

A third option: the timers don't really notify the waitset, we calculate the sleep duration based on the next call time of the timers in `rcl_wait()` and then call `rmw_wait()` with that. Maybe that calculation is wrong, or a clock type mismatch is causing it to take longer (I haven't debugged that yet).
The test is using the "steady clock" and I'm almost sure `rmw_wait()` uses the "system clock", I remember to have seen an issue about system clocks adjustments ...

fujitatomoya:
> The test is using the "steady clock" and I'm almost sure rmw_wait() uses the "system clock", I remember to have seen an issue about system clocks adjustments ...

https://github.com/ros2/ros2/issues/1080, as far as i know, cyclone and fastdds uses "system clock" only.

ivanpauno:
Closing in favor of https://github.com/ros2/rclcpp/pull/1628.

clalancette:
> Closing in favor of #1628.

I'm assuming you actually meant to close this based on this comment.  If not, feel free to reopen :).

PR #1595, opened 2021-03-24
Remove rmw_connext_cpp references.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

PR #1625, opened 2021-04-09
Restore test exception for Connext

asorbini:
This PR restores an exception for Connext in test `TestAllocatorMemoryStrategy::number_of_entities_with_subscription` which was [deleted](https://github.com/ros2/rclcpp/pull/1595/files#diff-4e1d578c35ec70e95e1a61099f536a316415483223a17061108a1a4e3f379a5bL514) in the process of replacing `rmw_connext_cpp` with `rmw_connextdds`.

The fix will enable `rmw_connextdds` to pass the test, as discovered while working on [rmw_connextdds#22](https://github.com/ros2/rmw_connextdds/pull/22).

### DIFF
--- a/rclcpp/test/rclcpp/strategies/test_allocator_memory_strategy.cpp
+++ b/rclcpp/test/rclcpp/strategies/test_allocator_memory_strategy.cpp
@@ -511,8 +511,10 @@ TEST_F(TestAllocatorMemoryStrategy, number_of_entities_with_subscription) {
   RclWaitSetSizes expected_sizes = {};
   expected_sizes.size_of_subscriptions = 1;
   const std::string implementation_identifier = rmw_get_implementation_identifier();
-  if (implementation_identifier == "rmw_cyclonedds_cpp") {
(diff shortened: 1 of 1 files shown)


clalancette:
Sorry, I don't fully understand this PR.  How is this test passing on Connext today, and how will this change fix it?  Or does this require ros2/rmw_connextdds#22 ?

asorbini:
This test only runs with the default RMW implementation, thus I had actually failed to run it in the past for `rmw_connextdds`.

My previous recommendation to remove this exception was wrong, because based on the wrong assumption that I had been running this test (and not seen it fail) as part of the CI plans, when instead that was never the case .

clalancette (approved):

asorbini:
See [this comment](https://github.com/ros2/rmw_connextdds/pull/22#issuecomment-816797969) for CI results including this PR

ivanpauno (approved):
LGTM!

ivanpauno:
`test_allocator_memory_strategy` passed in the linked CI jobs, going in!

PR #1626, opened 2021-04-09
Do not test the value of the history policy when testing the get_publishers/subscriptions_info_by_topic() methods

ivanpauno:
The history qos policy of endpoints is not necessarely shared during discovery.
cyclonedds does seem to share it, but fastrtps/connext don't and it's not guaranteed by the standard.

Test the value of the `reliability` policy instead of `history`, which is shared during discovery by all the implementations.

See https://github.com/ros2/rmw_connextdds/pull/22#discussion_r610672597.

### DIFF
--- a/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
+++ b/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
@@ -364,28 +364,10 @@ TEST_F(TestNodeGraph, get_info_by_topic)
   EXPECT_EQ(rclcpp::EndpointType::Publisher, const_publisher_endpoint_info.endpoint_type());
 
   rclcpp::QoS actual_qos = publisher_endpoint_info.qos_profile();
-  switch (actual_qos.get_rmw_qos_profile().history) {
-    case RMW_QOS_POLICY_HISTORY_KEEP_LAST:
-      EXPECT_EQ(1u, actual_qos.get_rmw_qos_profile().depth);
-      break;
-    case RMW_QOS_POLICY_HISTORY_UNKNOWN:
-      EXPECT_EQ(0u, actual_qos.get_rmw_qos_profile().depth);
-      break;
-    default:
-      ADD_FAILURE() << "unexpected history";
-  }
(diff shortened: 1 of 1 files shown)


jacobperron (approved):

hidmic:
I'd suggest we update the `rmw` API contracts accordingly before doing this. Nowhere it says that QoS profiles may be partially filled. And that's assuming there's no use case that'd require all the information. I'm not against changing our API to match the implementations, but dropping or restricting features because it's not easy to implement on top of X doesn't sound like a good rationale.

wjwwood:
I kind of agree with @hidmic. We should at least document this. I could imagine non-DDS vendors also not communicating this, but in that case you could argue they should bridge that gap with a side-channel of information.

ivanpauno:
I opened https://github.com/ros2/rmw/pull/308 adding documentation in `rmw`.
We can merge this independently of that PR, because checking the value of `reliability` or `history` doesn't affect the point of the test.

fujitatomoya (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14354/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9134/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12034/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14453/)

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14355/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9135/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12035/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14454/)

ivanpauno:
`test_node_graph` didn't fail in the jobs above.

All the other failures are unrelated, see https://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/293/#showFailuresLink (https://github.com/ros2/rmw_connextdds/pull/22 will solve most of them).

PR #1627, opened 2021-04-12
Check first parameter type and range before calling the user validation callbacks

ivanpauno:
Based on @jacobperron comment:

```cpp
node->declare_parameter("my_int", 42);
node->add_on_set_parameters_callback([](const std::vector<rclcpp::Parameter> & params) {
  for (const auto & param : params) {
    if (param.get_name() == "my_int") {
      param.as_int();
    }
  }
});
```

This could throw as the parameter type was checked after calling the user provided "on set" parameters callback.
Inver the order so parameter type is checked first (integer/floating point range will also be checked before user callbacks now).

This is also consistent with the current `rclpy` behavior: https://github.com/ros2/rclpy/blob/e938f92933a6919b8f277b17e5cd7deb18d75da1/rclpy/rclpy/node.py#L795-L811.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -300,12 +300,13 @@ __set_parameters_atomically_common(
   bool allow_undeclared = false)
 {
   // Check if the value being set complies with the descriptor.
-  result = __check_parameters(parameter_infos, parameters, allow_undeclared);
+  rcl_interfaces::msg::SetParametersResult result = __check_parameters(
+    parameter_infos, parameters, allow_undeclared);
   if (!result.successful) {
     return result;
   }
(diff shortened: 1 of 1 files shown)


ivanpauno:
I'm qualifying this as a bug, because it was inconsistent with `rclpy` and confusing.
Thus, I think it's worth mergin it after the feature/API freeze.

jacobperron (approved):
I'm okay with this change for Galactic. It's not touching the library's API/ABI.

wjwwood (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14350/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9131/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12030/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14449/)

PR #1516, opened 2021-01-15
Allow timers to keep up the intended rate in MultiThreadedExecutor

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1487.

Something in between https://github.com/ros2/rclcpp/pull/621 and https://github.com/ros2/rclcpp/pull/1168.
The whole idea of the `scheduled_timers_` container seems wrong, but this works around the issue.

PR #1628, opened 2021-04-12
Use a different implementation of mutex two priorities

### LINKED ISSUE #1008: test_multi_threaded_executor timer_over_take is flaky due to timer jitter

ivanpauno:
Fixes #1008 

The naive implementation added in https://github.com/ros2/rclcpp/pull/1516 doesn't work as well as I expected.
Even though only one "low priority thread" can be waiting in the "data mutex", it's still pretty likely to find the following behavior:

- low priority thread releases data mutex and then the "barrier" mutex
- another low priority thread releases takes the barrier mutex and then also takes the "data" mutex, even when there was a "high priority" thread waiting.

i.e.: the "high priority" take isn't always as "high priority" as we wanted.

Investigating https://github.com/ros2/rclcpp/issues/1008, I realized that the problem was that the executor was taking too long to schedule a timer callback (see investigation in https://github.com/ros2/rclcpp/pull/1624).
After some further investigation, I realized that the problem is that one of the "worker threads" is taking too long to remove the timer from the scheduled timers map, so a new timer callback can't be scheduled:

https://github.com/ros2/rclcpp/blob/a62287bf8d1eab0f146394ff0efdd4da44d4ef34/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L109-L113

Esentially, that's the same problem we observed in https://github.com/ros2/rclcpp/pull/1516, but the "MutexTwoPriorities" implementation wasn't good enough.

With this patch, I cannot reproduce the issue locally anymore.

---

Steps to reproduce the issue (on linux):

```basj
stress --cpu 8 &  # use number of cores instead of 8
for i in {1..100}; do ./test_multi_threaded_executor; if [ $? -ne 0 ]; then break;fi;clear; done
```

That should sometimes fails with the old implementation, and shouldn't fail at all with the new one.

---

The new implementation of "MutexTwoPriorities" is way trickier than the previous one, it should be reviewed with care.

https://www.modernescpp.com/index.php/c-core-guidelines-be-aware-of-the-traps-of-condition-variables

### DIFF
--- a/rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
+++ b/rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
@@ -35,7 +35,7 @@ HighPriorityLockable::lock()
   if (parent_.data_taken_) {
     ++parent_.hp_waiting_count_;
     while (parent_.data_taken_) {
-      parent_.cv_.wait(guard);
+      parent_.hp_cv_.wait(guard);
     }
     --parent_.hp_waiting_count_;
   }
@@ -45,16 +45,17 @@ HighPriorityLockable::lock()
 void
 HighPriorityLockable::unlock()
 {
+  bool notify_lp{false};
   {
     std::lock_guard<std::mutex> guard{parent_.cv_mutex_};
     parent_.data_taken_ = false;
+    notify_lp = 0u == parent_.hp_waiting_count_;
   }
-  // We need to notify_all(), and not only one,
-  // because the notified thread might be low priority
-  // when a high priority thread is still waiting.
-  // In that case, the low priority thread will go to sleep again,
-  // release the mutex, and the high priority thread can continue.
-  parent_.cv_.notify_all();
+  if (notify_lp) {
+    parent_.lp_cv_.notify_one();
+  } else {
+    parent_.hp_cv_.notify_one();
+  }
 }
 
 LowPriorityLockable::LowPriorityLockable(MutexTwoPriorities & parent)
@@ -66,7 +67,7 @@ LowPriorityLockable::lock()
 {
   std::unique_lock<std::mutex> guard{parent_.cv_mutex_};
   while (parent_.data_taken_ || parent_.hp_waiting_count_) {
-    parent_.cv_.wait(guard);
+    parent_.lp_cv_.wait(guard);
   }
   parent_.data_taken_ = true;
 }
@@ -74,13 +75,17 @@ LowPriorityLockable::lock()
 void
 LowPriorityLockable::unlock()
 {
+  bool notify_lp{false};
   {
     std::lock_guard<std::mutex> guard{parent_.cv_mutex_};
     parent_.data_taken_ = false;
+    notify_lp = 0u == parent_.hp_waiting_count_;
   }
-  // See comment in HighPriorityLockable::unlock() to understand
-  // why not using notify_one().
-  parent_.cv_.notify_all();
+  if (notify_lp) {
+    parent_.lp_cv_.notify_one();
+  } else {
+    parent_.hp_cv_.notify_one();
+  }
 }
 
 HighPriorityLockable
--- a/rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
+++ b/rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
@@ -31,16 +31,12 @@ HighPriorityLockable::HighPriorityLockable(MutexTwoPriorities & parent)
 void
 HighPriorityLockable::lock()
 {
-  bool did_aument_count{false};
   std::unique_lock<std::mutex> guard{parent_.cv_mutex_};
   if (parent_.data_taken_) {
     ++parent_.hp_waiting_count_;
-    did_aument_count = true;
-  }
-  while (parent_.data_taken_) {
-    parent_.cv_.wait(guard);
-  }
-  if (did_aument_count) {
+    while (parent_.data_taken_) {
+      parent_.cv_.wait(guard);
+    }
     --parent_.hp_waiting_count_;
   }
   parent_.data_taken_ = true;
--- a/rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
+++ b/rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
@@ -31,13 +31,34 @@ HighPriorityLockable::HighPriorityLockable(MutexTwoPriorities & parent)
 void
 HighPriorityLockable::lock()
 {
-  parent_.data_.lock();
+  bool did_aument_count{false};
+  std::unique_lock<std::mutex> guard{parent_.cv_mutex_};
+  if (parent_.data_taken_) {
+    ++parent_.hp_waiting_count_;
+    did_aument_count = true;
+  }
+  while (parent_.data_taken_) {
+    parent_.cv_.wait(guard);
+  }
+  if (did_aument_count) {
+    --parent_.hp_waiting_count_;
+  }
+  parent_.data_taken_ = true;
 }
 
 void
 HighPriorityLockable::unlock()
 {
-  parent_.data_.unlock();
+  {
+    std::lock_guard<std::mutex> guard{parent_.cv_mutex_};
+    parent_.data_taken_ = false;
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
```diff
@@ -31,13 +31,34 @@ HighPriorityLockable::HighPriorityLockable(MutexTwoPriorities & parent)
...
+  {
+    std::lock_guard<std::mutex> guard{parent_.cv_mutex_};
+    parent_.data_taken_ = false;
+  }
+  // We need to notify_all(), and not only one,
+  // because the notified thread might be low priority
+  // when a high priority thread is still waiting.
+  // In that case, the low priority thread will go to sleep again,
+  // release the mutex, and the high priority thread can continue.
+  parent_.cv_.notify_all();
```
hidmic, line 61: @ivanpauno I know this is tricky code, and that you've carefully thought this true. Having said that, wouldn't having one condition variable per priority, and notifying one on each, boost efficiency a bit?

rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
```diff
@@ -47,16 +68,23 @@ LowPriorityLockable::LowPriorityLockable(MutexTwoPriorities & parent)
 void
 LowPriorityLockable::lock()
 {
-  std::unique_lock<std::mutex> barrier_guard{parent_.barrier_};
-  parent_.data_.lock();
-  barrier_guard.release();
+  std::unique_lock<std::mutex> guard{parent_.cv_mutex_};
+  while (parent_.data_taken_ || parent_.hp_waiting_count_) {
```
fujitatomoya, line 72:
```suggestion
  while (parent_.hp_waiting_count_ || parent_.data_taken_) {
```

high priority should be checked 1st? since that means that there is at least one high priority thread in wait.

fujitatomoya:
sharing verification result on my environment (AMD Ryzen 9 3900X 12-Core Processor)

- https://github.com/ros2/ros2/commit/e6dffe9962013cce4210420a78530f62d8a74ea4 with `stress --cpu 24`

i CAN reproduce the issue.

```
Running main() from /root/ros2_ws/colcon_ws/install/gtest_vendor/src/gtest_vendor/src/gtest_main.cc
[==========] Running 1 test from 1 test suite.
[----------] 1 test from TestMultiThreadedExecutor
[ RUN      ] TestMultiThreadedExecutor.timer_over_take
/root/ros2_ws/colcon_ws/src/ros2/rclcpp/rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp:91: Failure
Expected: (diff) > (PERIOD - TOLERANCE), actual: 0.383995 vs 0.75
[  FAILED  ] TestMultiThreadedExecutor.timer_over_take (6240 ms)
[----------] 1 test from TestMultiThreadedExecutor (6240 ms total)

[==========] 1 test from 1 test suite ran. (6262 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestMultiThreadedExecutor.timer_over_take
```

- https://github.com/ros2/rclcpp/tree/ivanpauno/improve-mutex-two-priorities with `stress --cpu 24`

i CANNOT see the issue with at least 100 loops.

rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
ivanpauno: Yes, I thought about that when writting the comment as well, but it was late and I didn't have much brain power left :smile: .
I will give it a try now.

rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
ivanpauno: The order is irrelevant.
These are two boolean flags, so short-circuit behavior doesn't matter.

rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
```diff
@@ -31,13 +31,34 @@ HighPriorityLockable::HighPriorityLockable(MutexTwoPriorities & parent)
...
+  if (parent_.data_taken_) {
+    ++parent_.hp_waiting_count_;
+    did_aument_count = true;
+  }
+  while (parent_.data_taken_) {
+    parent_.cv_.wait(guard);
+  }
+  if (did_aument_count) {
+    --parent_.hp_waiting_count_;
+  }
```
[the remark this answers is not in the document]
ivanpauno: Pretty much, thanks for the suggestion!!

Fixed in https://github.com/ros2/rclcpp/pull/1628/commits/91ce2e7911fb804be97ef18adcb3f8bc7aa73851.

rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1628/commits/909bca576ac3b2d4b3d8bd108efc21eda42ca618.
I'm almost sure the code is still correct, though the version with one conditional variable is more trivial to follow.
PTAL

rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
```diff
@@ -70,21 +67,25 @@ LowPriorityLockable::lock()
...
 void
 LowPriorityLockable::unlock()
 {
+  bool notify_lp{false};
   {
     std::lock_guard<std::mutex> guard{parent_.cv_mutex_};
     parent_.data_taken_ = false;
+    notify_lp = 0u == parent_.hp_waiting_count_;
+  }
+  if (notify_lp) {
```
hidmic, line 84: @ivanpauno oh, didn't consider a conditional notification. It seems safe though.

hidmic (approved):

rclcpp/src/rclcpp/detail/mutex_two_priorities.cpp
ivanpauno: Yeah, I really though a lot about that part of the code, but given that `notify_lp` is calculated with the mutex taken this should be safe.

ivanpauno:
CI, testing downstream packages:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14365/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9144/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12044/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14463/)

CI, testing `test_multi_threaded_executor` with `--retest-until-fail 100`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14366/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9145/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12045/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14464/)

clalancette:
> i CANNOT see the issue with at least 100 loops.

Same here; before this PR, I could reproduce the issue on Linux with stress on the machine.  After this PR, I can't reproduce it anymore.  I'll do a review of this next, but fantastic job @ivanpauno

clalancette (approved):

PR #1629, opened 2021-04-13
Avoid returning loan when none was obtained

MiguelCompany:
While testing ros2/rmw_fastrtps#523 I saw the following error message several times during the execution of `test_communication`

```
[ERROR] [rclcpp]: rcl_return_loaned_message_from_subscription() failed for subscription on topic '/test_message_BasicTypes': loaned_message argument is null
```

I discovered that `Executor::execute_subscription` was trying to return a loaned message, even when none was obtained (i.e. there was no data available)

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -610,16 +610,18 @@ Executor::execute_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
         return true;
       },
       [&]() {subscription->handle_loaned_message(loaned_msg, message_info);});
-    rcl_ret_t ret = rcl_return_loaned_message_from_subscription(
-      subscription->get_subscription_handle().get(),
-      loaned_msg);
-    if (RCL_RET_OK != ret) {
-      RCLCPP_ERROR(
-        rclcpp::get_logger("rclcpp"),
-        "rcl_return_loaned_message_from_subscription() failed for subscription on topic '%s': %s",
-        subscription->get_topic_name(), rcl_get_error_string().str);
+    if (nullptr != loaned_msg) {
+      rcl_ret_t ret = rcl_return_loaned_message_from_subscription(
+        subscription->get_subscription_handle().get(),
+        loaned_msg);
+      if (RCL_RET_OK != ret) {
+        RCLCPP_ERROR(
+          rclcpp::get_logger("rclcpp"),
+          "rcl_return_loaned_message_from_subscription() failed for subscription on topic '%s': %s",
+          subscription->get_topic_name(), rcl_get_error_string().str);
+      }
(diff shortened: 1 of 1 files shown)


clalancette (commented):
It is definitely a bug to attempt returning a loaned message when we failed to take it in the first place.  Thanks for bringing this up.

I have two larger questions about this scenario, though.

1. Assume that `rcl_take_loaned_message` failed, and returned `RCL_RET_SUBSCRIPTION_TAKE_FAILED`.  In that case, `loaned_msg` will be `nullptr`, and we'll silently not do *any* work for that subscription.  That seems like a problem.
2. Relatedly, if taking the message via a loan fails, should we fall back to taking a copy?  I can see some arguments for and against this, I'm not quite sure which way we should go.

MiguelCompany:
> 1. Assume that `rcl_take_loaned_message` failed, and returned `RCL_RET_SUBSCRIPTION_TAKE_FAILED`.  In that case, `loaned_msg` will be `nullptr`, and we'll silently not do _any_ work for that subscription.  That seems like a problem.

It is explained [here](https://github.com/ros2/rclcpp/blob/085f161230769b57a0e4a711bb45a09c94b69b69/rclcpp/src/rclcpp/executor.cpp#L554-L564) that it may be ok for the middleware to interrupt wait spuriously. Sometimes, a liveliness event may wake-up the wait with no valid data to be taken (i.e. take will return only a SampleInfo with no associated data). Note that this also applies to non-loaned messages.

> 2. Relatedly, if taking the message via a loan fails, should we fall back to taking a copy?  I can see some arguments for and against this, I'm not quite sure which way we should go.

No. As far as I can see, `RCL_RET_SUBSCRIPTION_TAKE_FAILED` will only be returned when the `rmw` returns `RMW_RET_OK` and `taken` is false.

clalancette:
Makes sense, thanks for the explanations.  This looks good to me, so I'm going to approve and run CI on this, but I'd appreciate another review by an rclcpp maintainer.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14371/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9148/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12048/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14467/)

wjwwood (approved):

fujitatomoya (approved):
lgtm

clalancette:
The macOS failures is the well-known one in `test_multi_threaded_executor`, so isn't because of this PR (and incidentally, should be fixed elsewhere).  Given the green CI elsewhere and all the approvals, merging.  Thanks for the fix.

Issue #1524, opened 2021-01-23
rclcpp::spin_some(node)  Occasionally crash.


BaiCai1990:
## Bug report

**Required Info:**

- Operating System:
   Debian 10 buster
- Installation type:
  from source
- Version or commit hash:
  ref: refs/heads/master
- DDS implementation:
  Fast-RTPS
- Client library (if applicable):
  rclcpp

#### Steps to reproduce issue

```
--This issue is occationally occurs. Blow is the code.
rclcpp::WallRate main_loop(100);     
while(rclcpp::ok)
{
  rclcpp::spin_some(ros_node_);
  current_state = GetCurrentState();
  switch(current_state){
  case IDLE:{
    rclcpp::WallRate r(100);
    while(rclcpp::ok() && GetCurrentState() == IDLE){
    rclcpp::spin_some(ros_node_);                               //!!!!!!!!!!!!!!!!!!!!!!!! **crashed here**!!!!
    // other code
    r.sleep();
    }
    break;
  }
  case WALK:{
    rclcpp::WallRate r(100);
    while(rclcpp::ok() && GetCurrentState() == WALK){
      rclcpp::spin_some(ros_node_);
      // other code
      r.sleep();
    }
    break;
  }

  main_loop.sleep();
}
```

#### Expected behavior
run normal
#### Actual behavior
When state switch from WALK to IDLE, the program will  occationally crash in rclcpp::spin_some()
#### Additional information
In WALK state the loop is time-consumming than IDLE state

```
#0  0x00007ffff22a2704 in eprosima::fastrtps::rtps::StatefulReader::send_acknack(eprosima::fastrtps::rtps::WriterProxy const*, eprosima::fastrtps::rtps::RTPSMessageSenderInterface const&, bool) ()
   from /opt/ros2_foxy/install/fastrtps/lib/libfastrtps.so.2
#1  0x00007ffff229fc7c in std::_Function_handler<bool (), eprosima::fastrtps::rtps::WriterProxy::WriterProxy(eprosima::fastrtps::rtps::StatefulReader*, eprosima::fastrtps::rtps::RemoteLocatorsAllocationAttributes const&, eprosima::fastrtps::ResourceLimitedContainerConfig const&)::{lambda()#1}>::_M_invoke(std::_Any_data const&) ()
   from /opt/ros2_foxy/install/fastrtps/lib/libfastrtps.so.2
[...]

PR #1632, opened 2021-04-15
fix segmentation fault when adding multiple executors simultaneously

DensoADAS:
While creating multiple SingleThreadedExecturors in different thread I ran in crashes.
By adding a lock for "on_shutdown_callbacks_" in "on_shutdown" which is called in the constructor of SingleThreadedExecturors I solved the issue.

### DIFF
--- a/rclcpp/src/rclcpp/context.cpp
+++ b/rclcpp/src/rclcpp/context.cpp
@@ -344,6 +344,7 @@ Context::shutdown(const std::string & reason)
 rclcpp::Context::OnShutdownCallback
 Context::on_shutdown(OnShutdownCallback callback)
 {
+  std::lock_guard<std::mutex> lock(on_shutdown_callbacks_mutex_);
   on_shutdown_callbacks_.push_back(callback);
   return callback;
 }
--- a/rclcpp/test/rclcpp/test_node.cpp
+++ b/rclcpp/test/rclcpp/test_node.cpp
@@ -2556,6 +2556,7 @@ TEST_F(TestNode, get_publishers_subscriptions_info_by_topic) {
   };
   rclcpp::QoS qos = rclcpp::QoS(qos_initialization, rmw_qos_profile_default);
   auto publisher = node->create_publisher<test_msgs::msg::BasicTypes>(topic_name, qos);
+  std::this_thread::sleep_for(std::chrono::milliseconds(30));
   // List should have one item
   auto publisher_list = node->get_publishers_info_by_topic(fq_topic_name);
   ASSERT_EQ(publisher_list.size(), (size_t)1);


rclcpp/src/rclcpp/context.cpp
```diff
@@ -344,6 +344,7 @@ Context::shutdown(const std::string & reason)
 rclcpp::Context::OnShutdownCallback
 Context::on_shutdown(OnShutdownCallback callback)
 {
+  std::lock_guard<std::mutex> lock(on_shutdown_callbacks_mutex_);
```
fujitatomoya, line 347: probably not only adding mutex here, but also all access to `on_shutdown_callbacks_`?

fujitatomoya (commented):
I am not sure if this is related to https://github.com/ros2/rclcpp/issues/1524, but `on_shutdown_callbacks_` should be protected by `on_shutdown_callbacks_mutex_`.

rclcpp/src/rclcpp/context.cpp
DensoADAS: I think this is not possible without changing the API as the reference is returned

ivanpauno: We should deprecate in the future those getters, returning a reference to an internal container isn't great.

I think that the mutex should also be taken [here](https://github.com/ros2/rclcpp/blob/6eb34a84e6009a73860bdf54418f34c4f81b2969/rclcpp/src/rclcpp/context.cpp#L318) before calling all the callbacks:

```cpp
  {
    std::lock_guard<std::mutex> lock(on_shutdown_callbacks_mutex_);
    for (const auto & callback : on_shutdown_callbacks_) {
      callback();
    }
  }
```

wjwwood: > We should deprecate in the future those getters, returning a reference to an internal container isn't great.

In the general case, I don't agree with this, but in this particular case, I agree.

We could return a copy (since they should be trivially copyable) and lock that operation.

Or we could just drop the accessors. I think I added them just because having hidden state like that can bite you later if you need access to it and have to add an API to get it, or work around it by keeping track of them outside this class (duplicating the effort). So I'd lean towards returning a copy and protecting that with the mutex.

wjwwood:
@DensoADAS can you fix the commit so it passes our DCO check?

wjwwood:
I can help address the comments, but I cannot fix the DCO thing.

ivanpauno:
@DensoADAS friendly ping, could you sign the commit so the DCO check passes? Thanks!

DensoADAS:
Hi, just saw the mail. Do I have to sign off with a different mail/name?

clalancette:
The DCO bot can be a bit picky at times.  Right now it says this:

```
Expected "Joshua Hampp j.hampp@denso-adas.de", but got "Joshua Hampp j.hampp@eu.denso.com".
```

I think it is unhappy because the author (Joshua Hampp <j.hampp@denso-adas.de>) doesn't match the sign-off line (Joshua Hampp <j.hampp@eu.denso.com>).

ivanpauno:
Superseded by https://github.com/ros2/rclcpp/pull/1639, which fixed another thing + this.

ivanpauno:
Thanks @DensoADAS for the contribution!

Issue #1599, opened 2021-03-27
node crashes while received a cancel request if the state is SUCCEEDED


KavenYau:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries or source 
- Version or commit hash:
  - 47f21dab3d2c01710c8ea04784e970eec4a92e15
- DDS implementation:
  -  Fast-RTPS
- Client library (if applicable):
  - rclcpp (rclcpp_action)

#### Steps to reproduce issue

- I commented the code `std::lock_guard<std::recursive_mutex> lock(update_mutex_);` in this file https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/simple_action_server.hpp#L125, because it would cause deadlock,  although this issue #1518 claims that it is fixed. ( due to there are two threads tring to acquire the the mutex `update_mutex_`, https://github.com/ros-planning/navigation2/blob/main/nav2_util/include/nav2_util/simple_action_server.hpp#L384 )
```
  rclcpp_action::CancelResponse handle_cancel(
    const std::shared_ptr<rclcpp_action::ServerGoalHandle<ActionT>>/*handle*/)
  {
    // std::lock_guard<std::recursive_mutex> lock(update_mutex_);
    debug_msg("Received request for goal cancellation");
    return rclcpp_action::CancelResponse::ACCEPT;
  }
```
- then run navigation2 as usual, send goal and cancel goal
- then sometimes will crush and print this infomation
```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  goal_handle attempted invalid transition from state SUCCEEDED with event CANCEL_GOAL, at /tmp/binarydeb/ros-foxy-rcl-action-1.1.10/src/rcl_action/goal_handle.c:95
```

So far as I am aware, https://github.com/ros2/rcl/blob/foxy/rcl_action/src/rcl_action/goal_state_machine.c#L82 shows that the state SUCCEEDED can not be transformed to any state.  https://github.com/ros2/rclcpp/blob/foxy/rclcpp_action/src/server_goal_handle.cpp#L81 the `_cancel_goal()` function will be called internally only. so if the `handle_cancel_` user callback return `rclcpp_action::CancelResponse::ACCEPT` https://github.com/ros2/rclcpp/blob/foxy/rclcpp_action/include/rclcpp_action/server.hpp#L363, and it crushes. 
I can't catch the exception because it occurs internally.

[...]

PR #1635, opened 2021-04-15
Fix action server deadlock issue that caused by other mutexes locked in CancelCallback

KavenYau:
Resolves the deadlock issue mentioned in https://github.com/ros-planning/navigation2/issues/2273 and https://github.com/ros-planning/navigation2/pull/2304

The unit test which I added could also reproduce the issue.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/server.hpp
+++ b/rclcpp_action/include/rclcpp_action/server.hpp
@@ -356,16 +356,20 @@ protected:
   CancelResponse
   call_handle_cancel_callback(const GoalUUID & uuid) override
   {
-    std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+    std::shared_ptr<ServerGoalHandle<ActionT>> goal_handle;
+    {
+      std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+      auto element = goal_handles_.find(uuid);
+      if (element != goal_handles_.end()) {
+        goal_handle = element->second.lock();
+      }
+    }
+
     CancelResponse resp = CancelResponse::REJECT;
-    auto element = goal_handles_.find(uuid);
-    if (element != goal_handles_.end()) {
-      std::shared_ptr<ServerGoalHandle<ActionT>> goal_handle = element->second.lock();
-      if (goal_handle) {
-        resp = handle_cancel_(goal_handle);
-        if (CancelResponse::ACCEPT == resp) {
-          goal_handle->_cancel_goal();
-        }
+    if (goal_handle) {
+      resp = handle_cancel_(goal_handle);
+      if (CancelResponse::ACCEPT == resp) {
+        goal_handle->_cancel_goal();
       }
     }
     return resp;
--- a/rclcpp_action/test/test_server.cpp
+++ b/rclcpp_action/test/test_server.cpp
@@ -1234,10 +1234,13 @@ public:
         this->TryLockFor(lock, std::chrono::milliseconds(1000));
         return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
       },
-      [this](std::shared_ptr<GoalHandle>) {
+      [this](std::shared_ptr<GoalHandle> handle) {
         // instead of making a deadlock, check if it can acquire the lock in a second
         std::unique_lock<std::recursive_timed_mutex> lock(server_mutex_, std::defer_lock);
         this->TryLockFor(lock, std::chrono::milliseconds(1000));
+        if (!handle->is_active()) {
+          return rclcpp_action::CancelResponse::REJECT;
+        }
         return rclcpp_action::CancelResponse::ACCEPT;
       },
       [this](std::shared_ptr<GoalHandle> handle) {
@@ -1304,5 +1307,14 @@ TEST_F(TestDeadlockServer, deadlock_while_canceled)
   std::thread t(&TestDeadlockServer::GoalCanceled, this);
   rclcpp::sleep_for(std::chrono::milliseconds(50));
   send_goal_request(node_, uuid2_);  // deadlock here
+  t.join();
+}
+
+TEST_F(TestDeadlockServer, deadlock_while_succeed_and_canceled)
+{
+  send_goal_request(node_, uuid1_);
+  std::thread t(&TestDeadlockServer::GoalSucceeded, this);
+  rclcpp::sleep_for(std::chrono::milliseconds(50));
+  send_cancel_request(node_, uuid1_);
   t.join();
 }
--- a/rclcpp_action/test/test_server.cpp
+++ b/rclcpp_action/test/test_server.cpp
@@ -1238,6 +1238,7 @@ public:
         // instead of making a deadlock, check if it can acquire the lock in a second
         std::unique_lock<std::recursive_timed_mutex> lock(server_mutex_, std::defer_lock);
         this->TryLockFor(lock, std::chrono::milliseconds(1000));
+        // TODO(KavenYau): this check may become obsolete with https://github.com/ros2/rclcpp/issues/1599
         if (!handle->is_active()) {
           return rclcpp_action::CancelResponse::REJECT;
         }


KavenYau:
It's also related to #1599.
```c++
        if (handle->is_active()) return rclcpp_action::CancelResponse::ACCEPT;
        else return rclcpp_action::CancelResponse::REJECT;
```
if did not add the if/else condition to CancelCallback to check whether the goal active, it would throw exception.

SteveMacenski (approved):

SteveMacenski:
@clalancette I leave it to you if you think this bug patch should make it into Galactic, but it would be very nice if it did since it impacts a number of Nav2 users :smile: 

Love it, thanks for the patch @KavenYau! Hopefully this is the last-last race condition issue in the action server we'll need to fix :smile:

clalancette:
So the good news is that as it stands, this patch is ABI compatible.  So we could do it in a Galactic patch release.

That said, this would be a good fix to have (and is one of the reasons to do testing at this point).  So I've put it on the Galactic project board to look at.  That doesn't guarantee we'll get it in, but we'll definitely consider it.

SteveMacenski:
Let me know if I can help speed things up - its overall a pretty simple change so I'd hope a maintainer here could review it shortly

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -356,16 +356,20 @@ class Server : public ServerBase, public std::enable_shared_from_this<Server<Act
...
-      std::shared_ptr<ServerGoalHandle<ActionT>> goal_handle = element->second.lock();
-      if (goal_handle) {
-        resp = handle_cancel_(goal_handle);
-        if (CancelResponse::ACCEPT == resp) {
-          goal_handle->_cancel_goal();
-        }
+    if (goal_handle) {
+      resp = handle_cancel_(goal_handle);
+      if (CancelResponse::ACCEPT == resp) {
+        goal_handle->_cancel_goal();
```
wjwwood, line 372: I was concerned that this introduced a race here, allowing the goal to be finished asynchronously, after it was found but before we call the user callback on this line. But it seems that was always the expectation, since the user callback can return `CancelResponse::ACCEPT` or `CancelResponse::REJECT` (though that might only be one reason for it.

This is why you need the new code in the test:

```c++
        if (!handle->is_active()) {
          return rclcpp_action::CancelResponse::REJECT;
        }
```

However, now I'm wondering about `goal_handle->_cancel_goal();`, what if this happens:

- user callback checks `if (!handle->is_active())` which fails
- user returns `CancelResponse::ACCEPT`
- (⚡) context switch, goal finishes (terminates? enters terminal state? moves to aborted/succeeded state? not sure on the terminology)
- switch back to original thread, call `goal_handle->_cancel_goal()` because `CancelResponse::ACCEPT == resp`
- now we're canceling a finished goal

It looks like `_cancel_goal` doesn't care if the state changed under us:

https://github.com/ros2/rclcpp/blob/1037822a63330495dcf0de5e8f20544375a5f116/rclcpp_action/src/server_goal_handle.cpp#L80-L88

Though `rcl` does:

https://github.com/ros2/rcl/blob/414fdbd479b83f2b8af8235ad1fa3aeba2b0027f/rcl_action/src/rcl_action/goal_handle.c#L92-L99

Maybe we should be calling `rcl_action_goal_handle_is_cancelable()` before assigning it, or catch that rcl error?

Also, it should be mentioned to users (maybe already documented) that even though their cancel goal callback is called, and they return `CancelResponse::ACCEPT`, the resulting state may not be `canceled`...

@jacobperron any insight? I'm unfortunately not as versed in the design/implementation of actions as I think I need to be to judge this.

See also: https://design.ros2.org/articles/actions.html

wjwwood (commented):
I'm concerned that this will require all cancel callbacks for servers to check the `is_active()` state... which isn't ideal, imo.

Also, I'm concerned we could be allowing a goal to change states from SUCCEEDED or ABORTED to CANCELED erroneously.

I don't feel confident approving this change myself, I think we need more eyes on this.

rclcpp_action/include/rclcpp_action/server.hpp
KavenYau: @wjwwood Thanks for looking into it.

Yes, this patch could only fix the deadlock issue.

> However, now I'm wondering about goal_handle->_cancel_goal();, what if this happens:
user callback checks if (!handle->is_active()) which fails
user returns CancelResponse::ACCEPT
(⚡) context switch, goal finishes (terminates? enters terminal state? moves to aborted/succeeded state? not sure on the terminology)
switch back to original thread, call goal_handle->_cancel_goal() because CancelResponse::ACCEPT == resp
now we're canceling a finished goal

Actually this case I had mentioned in https://github.com/ros2/rclcpp/issues/1599#issuecomment-814145969, but maybe I didn't explain it clearly enough .

> Maybe we should be calling rcl_action_goal_handle_is_cancelable() before assigning it, or catch that rcl error?

> Also, it should be mentioned to users (maybe already documented) that even though their cancel goal callback is called, and they return CancelResponse::ACCEPT, the resulting state may not be canceled...

I think catching the exception internally is a straightforward way to resolve it, then the users have not to add `if (!handle->is_active())` to there CancelCallback.

But @fujitatomoya thought it was a strange behavior. https://github.com/ros2/rclcpp/issues/1599#issuecomment-810283121. So far I have no other solotions to fix this issue.

fujitatomoya: removing mutex before calling user callback looks okay to me 👍 

> I think catching the exception internally is a straightforward way to resolve it

i believe the point for this discussion is how it appears to be for user application. currently user application does receive exception if this happens. which i think it seems to be okay since user application gets this corner case as an exception. (probably exception message and adding check `rcl_action_goal_handle_is_cancelable` would be better.) i mean basically it does what needs to be done. 

i just do not have good feeling that system conceals this exception from user application, which is suggested on https://github.com/ros2/rclcpp/issues/1599.

KavenYau: I see.

> even though their cancel goal callback is called, and they return CancelResponse::ACCEPT, the resulting state may not be canceled..

In addition, I wonder make sure if you agree with this solution?

And how we fix it? catching the exception like below code? or doing some changes like this patch https://github.com/KavenYau/rclcpp/commit/3ba7b2ef01449929b6c976bdb9f21c886dff3464? or other better suggestions?
```c++
try {
  goal_handle->_cancel_goal();
} catch (const rclcpp::exceptions::RCLError & ex) {
  return CancelResponse::REJECT;
}
```

jacobperron: > Maybe we should be calling rcl_action_goal_handle_is_cancelable() before assigning it, or catch that rcl error?

It looks like an issue unrelated to this change. I think that it would be nice to handle this race between transitioning to canceling and some terminal state (succeeded or aborted) instead of throwing an exception.

> Also, it should be mentioned to users (maybe already documented) that even though their cancel goal callback is called, and they return CancelResponse::ACCEPT, the resulting state may not be canceled...

Indeed, by design *accepting* a cancel request only tries to transition the goal to the **canceling** state; it does not mean the goal is immediately transitioned to the **canceled** state. I don't know if this is documented in rclcpp_action very well, but I think the design doc captures it?

jacobperron (commented):
The change proposed in this PR looks okay to me. Correct me if I'm wrong, but I think the race related to transitioning to canceling and transitioning to a terminal state is not a new bug being introduced (it already exists).

KavenYau:
Is there any progress on this？ :grin:

rclcpp_action/include/rclcpp_action/server.hpp
fujitatomoya: > It looks like an issue unrelated to this change.

i think so too, this fix is unrelated.

> However, it's probably better to address this in a separate PR.

agree, issue is already opened. https://github.com/ros2/rclcpp/issues/1599

> Indeed, by design accepting a cancel request only tries to transition the goal to the canceling state; it does not mean the goal is immediately transitioned to the canceled state. I don't know if this is documented in rclcpp_action very well, but I think the design doc captures it?

i did check the action design, and i see `The list of goals in the response indicates which goals will be attempted to be canceled.` as you mentioned. besides, 



it described clearly that after accepting `cancel request` that still possible transit into `succeeded`.

so i think my assumption is wrong, it should not throw exception in this case.

fujitatomoya (approved):
this fix itself looks good to me.

fujitatomoya:
@jacobperron @wjwwood it would be probably better if either of you take a look again? and about `exception problem`, i think we can discuss that on https://github.com/ros2/rclcpp/issues/1599?

rclcpp_action/include/rclcpp_action/server.hpp
wjwwood: I think we can follow up on this in https://github.com/ros2/rclcpp/issues/1599

rclcpp_action/test/test_server.cpp
```diff
@@ -1234,10 +1234,13 @@ class TestDeadlockServer : public TestServer
...
         return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
       },
-      [this](std::shared_ptr<GoalHandle>) {
+      [this](std::shared_ptr<GoalHandle> handle) {
         // instead of making a deadlock, check if it can acquire the lock in a second
         std::unique_lock<std::recursive_timed_mutex> lock(server_mutex_, std::defer_lock);
         this->TryLockFor(lock, std::chrono::milliseconds(1000));
+        if (!handle->is_active()) {
+          return rclcpp_action::CancelResponse::REJECT;
+        }
```
wjwwood, line 1243: Should we have a TODO here mentioning the issue @fujitatomoya created?

```suggestion
        # TODO(KavenYau): this check may become obsolete with https://github.com/ros2/rclcpp/issues/1599
        if (!handle->is_active()) {
          return rclcpp_action::CancelResponse::REJECT;
        }
```

Or is the thinking that we'll need it always?

wjwwood (commented):
changes lgtm

rclcpp_action/test/test_server.cpp
jacobperron: I think we should patch things so that users don't have to check if the goal is active. IIUC, there's still potentially a race doing this in general, since the goal state may change between checking `is_active()` and returning from the user-callback.

jacobperron (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14451/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9222/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12122/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/14572/)

fujitatomoya:
it seems windows job got stuck to clear out docker containers?

```
00:00:09.572 C:\J\workspace\ci_windows>powershell -Command "if ($(docker ps -q) -ne $null) { docker stop $(docker ps -q)}" 
...
```

jacobperron:
I've retriggered the Windows build.

fujitatomoya:
@jacobperron thanks :+1: 

all green, i think this can be merged.

clalancette:
Should we consider backporting this one to Galactic?

SteveMacenski:
I think so :smile:

KavenYau:
Should / Could I backport this to Foxy / Galactic?

clalancette:
> Should I backport this to Foxy / Galactic?

Definitely Galactic.  I haven't looked at the code in Foxy, but if it applies there and doesn't break ABI, then it can be backported to Foxy as well.

jacobperron:
I've already added this to the Foxy project board.

KavenYau:
> I've already added this to the Foxy project board.

Dose it mean that I don't need to make a backporting PR for it because it's already in plan?

clalancette:
> Dose it mean that I don't need to make a backporting PR for it because it's already in plan?

It still needs someone to do the work for both Galactic and Foxy, so if you are willing to do it, it would be appreciated.

ivanpauno:
@Mergifyio backport galactic

Issue #1638, opened 2021-04-19
Possible memory leak in spin_some()


BriceRenaudeau:
## Bug report

When using `rclcpp::spin_some(node);`, the node resources increase forever.

**Required Info:**

- Operating System:
  -  Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - Rolling
- DDS implementation:
  - Fast-RTPS or CycloneDDS
- Client library (if applicable):
  - rclcpp
  
#### Steps to reproduce issue
Compile and execute this simple node.
```c++
#include <rclcpp/rclcpp.hpp>
int main(int argc, char * argv[])
{
  rclcpp::init(argc, argv);
  auto node = std::make_shared<rclcpp::Node>("foo");
  while(rclcpp::ok())
  {
    rclcpp::spin_some(node);
  }

  rclcpp::shutdown();
  return 0;
}
```
Using `htop` you will see the resources of the process increase.

#### Expected behavior

The node resources should be stable.

#### Actual behavior

The node resources increase until the OS kills the node because of memory storage.
`Out of memory: Killed process`

#### Additional information

It can be easily bypassed by creating an executor out of the loop and spinning it inside.

clalancette:
> * Version or commit hash:
>   
>   * release version

Can you be more specific on what version you are using?  Is it Foxy, is it Rolling?

clalancette:
OK, yeah, confirmed.  It is definitely leaking in the latest from-source build as well.

fujitatomoya:
confirmed this happens with https://github.com/ros2/ros2/commit/2c3dcb37040365ee5f09eede264cd05910134076. (cyclonedds and fastdds)

```
(gdb) info proc mappings
process 11010
Mapped address spaces:
...<snip>
      0x555555561000     0x55555ead0000  0x956f000        0x0 [heap]

(gdb) whatis 0x555555561000
type = long
```

fujitatomoya:
@Barry-Xu-2018 @iuhilnehc-ynos i guess this is high priority, we would want to take a look.

Barry-Xu-2018:
> @Barry-Xu-2018 @iuhilnehc-ynos i guess this is high priority, we would want to take a look.

Okay. We check it now.

Barry-Xu-2018:
The cause is related to below codes

[...]

PR #1639, opened 2021-04-21
Use OnShutdown callback handle instead of OnShutdown callback

Barry-Xu-2018:
1. Add remove_on_shutdown_callback() for the context
2. on_shutdown_callback is removed by remove_on_shutdown_callback() while
   the destructor of executor is executed.


Barry-Xu-2018:
Related to #1638

rclcpp/include/rclcpp/context.hpp
```diff
@@ -196,29 +203,38 @@ class Context : public std::enable_shared_from_this<Context>
...
   RCLCPP_PUBLIC
   virtual
-  OnShutdownCallback
-  on_shutdown(OnShutdownCallback callback);
+  OnShutdownCallbackHandle::SharedPtr
+  on_shutdown(OnShutdownCallbackHandle::OnShutdownCallbackType callback);
+
+  /// Remove a on_shutdown callback handle registered with on_shutdown.
+  /**
+   * \param[in] callback_handle the on shutdown callback handle to be removed.
```
clalancette, line 215:
```suggestion
   * \param[in] callback_handle the handle to be removed
```

rclcpp/src/rclcpp/context.cpp
```diff
@@ -341,20 +348,38 @@ Context::shutdown(const std::string & reason)
...
 }
 
-rclcpp::Context::OnShutdownCallback
-Context::on_shutdown(OnShutdownCallback callback)
+rclcpp::OnShutdownCallbackHandle::SharedPtr
+Context::on_shutdown(OnShutdownCallbackHandle::OnShutdownCallbackType callback)
 {
-  on_shutdown_callbacks_.push_back(callback);
-  return callback;
+  std::lock_guard<std::mutex> lock(on_shutdown_callbacks_mutex_);
```
clalancette, line 354: Relatively minor, but this can be moved down to right above the `on_shutdown_callbacks_.push_back()`.  That is, we don't need to be holding this lock while we are allocating the handle and setting the callback.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -341,20 +348,38 @@ Context::shutdown(const std::string & reason)
...
+  auto it = std::find_if(
+    on_shutdown_callbacks_.begin(),
+    on_shutdown_callbacks_.end(),
+    [callback_handle](const auto & weak_handle) {
+      return callback_handle.get() == weak_handle.lock().get();
+    });
+  if (it != on_shutdown_callbacks_.end()) {
+    on_shutdown_callbacks_.erase(it);
+  }
```
clalancette, line 373: I have to say that I don't love this; if there are a lot of handles in the list, this becomes a bottleneck.  But I guess until we prove that this is a performance problem, we can leave it as-is.

clalancette (commented):
I've left a few minor comments.  I really want @ivanpauno and @wjwwood to review this.

rclcpp/include/rclcpp/context.hpp
```diff
@@ -196,29 +203,38 @@ class Context : public std::enable_shared_from_this<Context>
...
    * \param[in] callback the on shutdown callback to be registered
-   * \return the callback passed, for convenience when storing a passed lambda
+   * \return the created shared pointer of callback handler
    */
   RCLCPP_PUBLIC
   virtual
-  OnShutdownCallback
-  on_shutdown(OnShutdownCallback callback);
+  OnShutdownCallbackHandle::SharedPtr
+  on_shutdown(OnShutdownCallbackHandle::OnShutdownCallbackType callback);
```
ivanpauno, line 211: The caller now needs to retain the return value to ensure that the callback was still valid, which wasn't the case before.
It's better to return an identifier that can be passed to remove but that you don't need to retain if you want to keep the callback forever, so previous code isn't broken.

rclcpp/src/rclcpp/context.cpp
ivanpauno: :+1:
It doesn't matter much though, as I don't think that removing a shutdown callback is performance critical.

The underlying container could be a `std::unordered_set`, or something like that.

ivanpauno: I think that one reason to remove the `get_on_shutdown_callbacks()` methods is to not expose the underlying container type.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -341,20 +348,38 @@ Context::shutdown(const std::string & reason)
...
 
-rclcpp::Context::OnShutdownCallback
-Context::on_shutdown(OnShutdownCallback callback)
+rclcpp::OnShutdownCallbackHandle::SharedPtr
+Context::on_shutdown(OnShutdownCallbackHandle::OnShutdownCallbackType callback)
 {
-  on_shutdown_callbacks_.push_back(callback);
-  return callback;
+  std::lock_guard<std::mutex> lock(on_shutdown_callbacks_mutex_);
+  auto handle = std::make_shared<OnShutdownCallbackHandle>();
```
ivanpauno, line 355: If we keep the idea of returning an scoped handle, this could have a custom deleter that automatically calls `remove_on_shutdown_callback()` (leveraging this->weak_from_this()).

rclcpp/include/rclcpp/context.hpp
ivanpauno: i.e. I prefer either the idea of returning an iterator to the handle (which breaks ABI if you change the container type), or what @wjwwood mentioned about using a `std::unordered_map` with an identifier using as a key (incremented each time a new callback is added).

Barry-Xu-2018:
@clalancette @ivanpauno @fujitatomoya 

Thanks for your review comments.     
I find this version has another problem.  
Now shutdown callback handle should be hold by who calls Context::on_shutdown().  
For example, Executor add below code to hold created shutdown callback handler.
```
  /// shutdown callback handle registered to Context
  rclcpp::OnShutdownCallbackHandle::SharedPtr shutdown_callback_handle_;
```

In existing codes, there are 2 positions which call Context::on_shutdown().

1. 
https://github.com/ros2/rclcpp/blob/bff6916e8f3f4d0cb56ebe62a5958bfa5a586977/rclcpp/src/rclcpp/graph_listener.cpp#L93-L100

Like the change of Executor, graph_listener can hold the handle.

2. 
https://github.com/ros2/rclcpp/blob/bff6916e8f3f4d0cb56ebe62a5958bfa5a586977/rclcpp/src/rclcpp/utilities.cpp#L180-L188

Here is a problem.  How to hold the handle if not change the interface of rclcpp::on_shutdown() (We think caller should hold handle.) ? 

Another way is to hold shutdown callback handle in Context side instead of caller side.  I think it is reasonable that the caller hold shutdown callback handle. Context should not care whether caller is destructed.  
Please share your thought.

rclcpp/src/rclcpp/context.cpp
Barry-Xu-2018: Yeah. I follow the code style   
https://github.com/ros2/rclcpp/blob/b8df9347a10bff383fdc8890c5b1bfdf16430d8f/rclcpp/src/rclcpp/parameter_event_handler.cpp#L40-L56

But the performance is a problem. I will change to use `std::unordered_set`.

rclcpp/include/rclcpp/context.hpp
Barry-Xu-2018: > The caller now needs to retain the return value to ensure that the callback was still valid, which wasn't the case before.

Yeah. It has a problem. (Refer to my comments https://github.com/ros2/rclcpp/pull/1639#issuecomment-824654837)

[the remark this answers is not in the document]
Barry-Xu-2018: I am not sure whether it is acceptable at this time ?
@ivanpauno @clalancette What do you think ?

ivanpauno:
> We think caller should hold handle

I would rather don't make that a requirement as [commented above](https://github.com/ros2/rclcpp/pull/1639#discussion_r617531470).
We don't control all the external packages that call `on_shutdown`, so I would rather avoid that change.

wjwwood:
I would say, do not make the handle scoped, i.e. hold a shared pointer to the handle in the context, not a weak pointer.

Then we could change `rclcpp::on_shutdown()` to return the handle and existing calls would simply ignore (implicitly discard) the handle, which is fine because they never intend to remove it. In other cases, you can change the code to keep the handle and use it to remove the callback later (not relying on RAII to have the destruction of the handle remove the callback).

In that case we're treating the address in the shared pointer given to the user as the "id" which can be used to remove the callback later. We could also use an integer, but that has the roll over issue, and the allocator ensures an address in a shared pointer is only used once at a time (but can be reused later), so I think that's better.

ivanpauno:
> Like the change of Executor, graph_listener

Yes, GraphListener can do the same than Executor.

> rclcpp::on_shutdown()

We need the same change than in `Context::on_shutdown()` here, return a handle and add a remove function.

Barry-Xu-2018:
@ivanpauno @clalancette 

Please help to review again.

rclcpp/include/rclcpp/context.hpp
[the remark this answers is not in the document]
wjwwood: You can add the new function, then deprecate this one in h-turtle. I would try to keep the fix as narrow as possible for galactic.

rclcpp/include/rclcpp/context.hpp
```diff
@@ -177,8 +187,6 @@ class Context : public std::enable_shared_from_this<Context>
   bool
   shutdown(const std::string & reason);
 
-  using OnShutdownCallback = std::function<void ()>;
```
wjwwood, line 180: Let's keep this here in case someone was using it:

```c++
using OnShutdownCallback = OnShutdownCallbackHandle::OnShutdownCallbackType;
```

rclcpp/include/rclcpp/context.hpp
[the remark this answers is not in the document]
wjwwood: To be clear, I think you could add the new function _now_, and just have one call the other. Then deprecate `on_shutdown` in h-turtle, maybe.

rclcpp/include/rclcpp/context.hpp
```diff
@@ -196,29 +204,38 @@ class Context : public std::enable_shared_from_this<Context>
...
+   * \return the created shared pointer of callback handler
    */
   RCLCPP_PUBLIC
   virtual
-  OnShutdownCallback
-  on_shutdown(OnShutdownCallback callback);
+  OnShutdownCallbackHandle::WeakPtr
+  on_shutdown(OnShutdownCallbackHandle::OnShutdownCallbackType callback);
+
+  /// Remove a on_shutdown callback handle registered with on_shutdown.
```
wjwwood, line 214:
```suggestion
  /// Remove an on_shutdown callback handle registered with on_shutdown.
```

rclcpp/include/rclcpp/context.hpp
```diff
@@ -196,29 +204,38 @@ class Context : public std::enable_shared_from_this<Context>
...
+  /**
+   * \param[in] callback_handle the handle to be removed.
+   */
+  RCLCPP_PUBLIC
+  virtual
+  void
+  remove_on_shutdown_callback(const OnShutdownCallbackHandle::WeakPtr callback_handle);
```
wjwwood, line 221: Either remove the (redundant `const` on a by-value argument):

```suggestion
  remove_on_shutdown_callback(OnShutdownCallbackHandle::WeakPtr callback_handle);
```

Or make it a const-reference:

```suggestion
  remove_on_shutdown_callback(const OnShutdownCallbackHandle::WeakPtr & callback_handle);
```

wjwwood: I'd prefer the latter, myself, but either is fine.

rclcpp/include/rclcpp/context.hpp
```diff
@@ -196,29 +204,38 @@ class Context : public std::enable_shared_from_this<Context>
...
+  /// Return the shutdown callback handles
   /**
    * Using the returned reference is not thread-safe with calls that modify
-   * the list of "on shutdown" callbacks, i.e. on_shutdown().
+   * the list of "on shutdown" callback handles, i.e. on_shutdown().
    */
   RCLCPP_PUBLIC
-  std::vector<OnShutdownCallback> &
+  std::unordered_set<OnShutdownCallbackHandle::SharedPtr> &
   get_on_shutdown_callbacks();
```
wjwwood, line 239: Management of this structure is handled with `add/remove` functions, so I don't think we need read/write access to it. I'd just remove this signature. Users can make a copy with the const version of this method if they need a mutable copy of this structure.

rclcpp/src/rclcpp/context.cpp
```diff
@@ -341,20 +346,35 @@ Context::shutdown(const std::string & reason)
...
+void
+Context::remove_on_shutdown_callback(const OnShutdownCallbackHandle::WeakPtr callback_handle)
 {
-  on_shutdown_callbacks_.push_back(callback);
-  return callback;
+  auto callback_handle_shared_ptr = callback_handle.lock();
+  if (callback_handle_shared_ptr) {
+    std::lock_guard<std::mutex> lock(on_shutdown_callbacks_mutex_);
+    on_shutdown_callbacks_.erase(callback_handle_shared_ptr);
+  }
```
wjwwood, line 368: Is it possible that between locking the weak ptr and acquiring the lock, this pointer could have been removed already?

I suppose it doesn't matter since `erase` won't care if it removed anything or not.

rclcpp/include/rclcpp/context.hpp
```diff
@@ -47,6 +48,15 @@ class ContextAlreadyInitialized : public std::runtime_error
...
 
+struct OnShutdownCallbackHandle
+{
+  RCLCPP_SMART_PTR_DEFINITIONS(OnShutdownCallbackHandle)
+
+  using OnShutdownCallbackType = std::function<void ()>;
+
+  OnShutdownCallbackType callback;
+};
+
```
ivanpauno, line 59: To avoid leaking implementation details, I would define `OnShutdownCallback`/`OnShutdownCallbackHandle` like:

```cpp
using OnShutdownCallbackType = std::function<void ()>;
class OnShutdownCallbackHandle
{
  friend rclcpp::Context;
  std::weak_ptr<OnShutdownCallbackType> handle_;
};
```

So in the future `OnShutdownCallbackHandle` can be something different than a std::weak_ptr.
They can be defined nested in `rclcpp::Context` or directly in the `rclcpp` namespace, I don't mind strongly.

rclcpp/include/rclcpp/context.hpp
```diff
@@ -196,29 +204,38 @@ class Context : public std::enable_shared_from_this<Context>
...
+  /// Return the shutdown callback handles as const.
   /**
    * Using the returned reference is not thread-safe with calls that modify
-   * the list of "on shutdown" callbacks, i.e. on_shutdown().
+   * the list of "on shutdown" callback handles, i.e. on_shutdown().
    */
   RCLCPP_PUBLIC
-  const std::vector<OnShutdownCallback> &
+  const std::unordered_set<OnShutdownCallbackHandle::SharedPtr> &
   get_on_shutdown_callbacks() const;
```
ivanpauno, line 230: The problem of this is that we will need to change the return type each time we change the internal type.

Maybe the return type could be `std::vector<OnShutdownCallback>`, which is a copy of the internal callbacks.
We could also have one that return the "handles", or the return type could be `std::vector<std::pair<OnShutdownCallbackHandle, OnShutdownCallback>>`.

rclcpp/include/rclcpp/context.hpp
ivanpauno, line 239: if we start to return a copy, this overload can be deleted

rclcpp/include/rclcpp/context.hpp
Barry-Xu-2018: I want to declare const-reference. But miss `&`. I will fix it.

rclcpp/include/rclcpp/context.hpp
Barry-Xu-2018: Okay.

rclcpp/src/rclcpp/context.cpp
Barry-Xu-2018: > Is it possible that between locking the weak ptr and acquiring the lock, this pointer could have been removed already?

Yes. This scenario exists.

> I suppose it doesn't matter since erase won't care if it removed anything or not.

Yes. But I tend to modify it. `on_shutdown_callbacks_mutex_` should be locked at first.

rclcpp/include/rclcpp/context.hpp
Barry-Xu-2018: > The problem of this is that we will need to change the return type each time we change the internal type.

Yes. This is a problem. 

> Maybe the return type could be std::vector<OnShutdownCallback>, which is a copy of the internal callbacks.

Yes. This returned type is the same as the previous.

rclcpp/include/rclcpp/context.hpp
Barry-Xu-2018: Thanks.  
But I think we cannot use `std::weak_ptr<OnShutdownCallbackType> handle_;`. We should hold handle in Context.

Barry-Xu-2018:
@wjwwood @ivanpauno 

Please help to review again.

rclcpp/include/rclcpp/context.hpp
ivanpauno: The `Context` container for the callbacks is `std::unordered_set<std::shared_ptr<OnShutdownCallbackType>>`, in the same fashion than now.

You can seamlessly convert between `OnShutdownCallbackHandle` and `std::shared_ptr<OnShutdownCallbackType>` internally.

e.g.
```cpp
OnShutdownCallbackHandle
on_shutdown(OnShutdownCallbackType callback) {
  auto callback_shared_ptr = std::make_shared<OnShutdownCallbackType>(callback);
  on_shutdown_callbacks_.emplace(callback_shared_ptr);
  OnShutdownCallbackHandle handle;
  handle.handle_ = callback_shared_ptr;
  return handle;
}

bool
remove_on_shutdown_callback(OnShutdownCallbackHandle handle) {
  auto callback_shared_ptr = handle.handle_.lock();
  if (callback_shared_ptr == nullptr) {
    return false;
  }
  return on_shutdown_callbacks_.erase(callback_shared_ptr) == 1;
}
```

wjwwood:
> @wjwwood Now I made a new interface add_on_shutdown_callback instead of on_shutdown (Now on_shutdown is marked as deprecated)

I suggested that you do *not* deprecate it for galactic, but now that we've forked for galactic and master is rolling/h-turtle again, we can do that.

> Whether should I use new add_on_shutdown_callback to update existing codes ?

Yes, we cannot merge it without addressing the deprecation warnings for `rclcpp` and any downstream code in https://github.com/ros2/ros2/blob/master/ros2.repos.

rclcpp/include/rclcpp/context.hpp
```diff
@@ -198,27 +213,51 @@ class Context : public std::enable_shared_from_this<Context>
    * \param[in] callback the on shutdown callback to be registered
    * \return the callback passed, for convenience when storing a passed lambda
    */
+  [[deprecated("use add_on_shutdown_callback(OnShutdownCallback) instead")]]
```
wjwwood, line 216: We might want to keep this out for now, merge this, back port it to galactic, and then make a new pull request that adds this and fixes the warnings on master.

wjwwood (approved):
changes lgtm

Barry-Xu-2018:
> Yes, we cannot merge it without addressing the deprecation warnings for rclcpp and any downstream code in https://github.com/ros2/ros2/blob/master/ros2.repos.

Okay. 

> We might want to keep this out for now, merge this, back port it to galactic, and then make a new pull request that adds this and fixes the warnings on master.

Okay. I will remove `[[deprecated()]` for this fixing.   
And make new commit to add `[[deprecated()]` and fixing build warnings.

Barry-Xu-2018:
I do rebase.

Barry-Xu-2018:
@wjwwood 

> We might want to keep this out for now, merge this, back port it to galactic

Done.   
I submit https://github.com/ros2/rclcpp/pull/1639/commits/16be0943e97de4aecdad6167452eafec4288bd07 to remove deprecated declaration.

rclcpp/include/rclcpp/context.hpp
ivanpauno: Sorry, I think I didn't explain my idea correctly.

The point of my comment was to make the return type **exactly** `OnShutdownCallbackHandle` and not `OnShutdownCallbackHandle::WeakPtr`, so the fact that we're using a `std::weak_ptr` as a handle and internally storing a `shared_ptr` is completly transparet for the user.
In the future if we use another approach and `OnShutdownCallbackHandle` is an integer (or whatever other thing), the API doesn't change and we only break ABI.

rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -119,13 +119,6 @@ TEST(TestUtilities, test_context_basic_access) {
   EXPECT_EQ(std::string{""}, context1->shutdown_reason());
 }
 
-TEST(TestUtilities, test_context_basic_access_const_methods) {
-  auto context1 = std::make_shared<const rclcpp::contexts::DefaultContext>();
-
-  EXPECT_NE(nullptr, context1->get_init_options().get_rcl_init_options());
-  EXPECT_EQ(0u, context1->get_on_shutdown_callbacks().size());
-}
```
ivanpauno, line 127: I think we can keep this test, right?

ivanpauno (approved):
I only have two minor comments left (the previous one is [here](https://github.com/ros2/rclcpp/pull/1639#discussion_r622194302)).

rclcpp/include/rclcpp/context.hpp
Barry-Xu-2018: Thanks for further explanation.  
Get it.  
I will update codes.

rclcpp/test/rclcpp/test_utilities.cpp
Barry-Xu-2018: Yeah. I should keep get_rcl_init_options() test.

ivanpauno (approved):
LGTM!

rclcpp/test/rclcpp/test_utilities.cpp
```diff
@@ -123,7 +123,6 @@ TEST(TestUtilities, test_context_basic_access_const_methods) {
   auto context1 = std::make_shared<const rclcpp::contexts::DefaultContext>();
 
   EXPECT_NE(nullptr, context1->get_init_options().get_rcl_init_options());
-  EXPECT_EQ(0u, context1->get_on_shutdown_callbacks().size());
```
ivanpauno: I have readded this line in https://github.com/ros2/rclcpp/pull/1639/commits/63244220c405a42916f4f74395f47515d5b92733, modified const qualifiers to make it work in https://github.com/ros2/rclcpp/pull/1639/commits/b27f448f7067e17a4aafa42e69af04003c652125.

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14512/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9270/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12178/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14612/)

ivanpauno:
@Mergifyio backport galactic foxy

ivanpauno:
Uff wrong tab, I wanted to backport another PR :man_facepalming:.

@wjwwood could you take another look to this one?

ivanpauno:
Thanks for the contribution @Barry-Xu-2018 !!

Issue #1611, opened 2021-03-31
Callback not reached, using callback group and Fast-RTPS


anaelle-sw:
## Bug report

**Required Info:**

- Operating System:  
  - Ubuntu 20.04
- Installation type: 
  - Binaries
- Version or commit hash:
  - [8.1.0](https://github.com/ros2/rclcpp/releases/tag/8.1.0) ROS2 Rolling
  - **Edit**: Also reproduced on latest release: [8.2.0](https://github.com/ros2/rclcpp/releases/tag/8.2.0) ROS2 Rolling
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Hello, 
we are currently using CycloneDDS as RWM implementation, but we would need to use Fast-RTPS.
Using Fast-RTPS changes our ROS2 application behavior. I found out it is related to the callback groups, which we use a lot.
I tried to isolate the issue with 3 nodes that are quite simple:
- A service client
```c
#include <chrono>
#include <functional>
#include <memory>
#include <string>

#include <rclcpp/rclcpp.hpp>
#include <std_msgs/msg/bool.hpp>
#include <std_srvs/srv/trigger.hpp>

class CallbackGroupSrvClient: public rclcpp::Node {
public:
  // Constructor
  CallbackGroupSrvClient()
  : Node("callback_goup_srv_client"){

    // Init callback group
    callback_group_ = this->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive, false);
    callback_group_executor_thread_ = std::thread([this]() {
      callback_group_executor_.add_callback_group(callback_group_, this->get_node_base_interface());
      callback_group_executor_.spin();
    });

    // Init subscription
    subscription_ = this->create_subscription<std_msgs::msg::Bool>(
          "topic_bool",
          rclcpp::QoS(1),
          std::bind(
            &CallbackGroupSrvClient::sub_callback,
            this,
            std::placeholders::_1
            )
          );

    // Create service client and link it to callback group
    srv_client_ = this->create_client<std_srvs::srv::Trigger>(
          "trigger_srv",
          rmw_qos_profile_services_default,
          callback_group_
          );
  }

  // Destructor
  ~CallbackGroupSrvClient(){
    callback_group_executor_.cancel();
    callback_group_executor_thread_.join();
  }

private:
  // Methods
  void sub_callback(const std_msgs::msg::Bool::SharedPtr ){
[...]

PR #1640, opened 2021-04-25
Fix bug that a callback not reached

iuhilnehc-ynos:
a subscriber on a new executor with a callback group triggered to receive a message

related to https://github.com/ros2/rclcpp/issues/1611

### DIFF
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -176,13 +176,17 @@ public:
 
   /// Create and return the notify guard condition.
   RCLCPP_PUBLIC
-  std::shared_ptr<rclcpp::GuardCondition>
-  create_notify_guard_condition(
-    const rclcpp::Context::SharedPtr context_ptr);
+  rclcpp::GuardCondition::SharedPtr
+  create_notify_guard_condition(const rclcpp::Context::SharedPtr context_ptr);
 
-  /// Return the notify guard condition created before.
+  /// Destroy the notify guard condition.
   RCLCPP_PUBLIC
-  std::weak_ptr<rclcpp::GuardCondition>
+  void
+  destroy_notify_guard_condition();
+
+  /// Return the notify guard condition.
+  RCLCPP_PUBLIC
+  rclcpp::GuardCondition::SharedPtr
   get_notify_guard_condition();
 
 protected:
@@ -229,6 +233,7 @@ protected:
   const bool automatically_add_to_executor_with_node_;
   // defer the creation of the guard condition
   std::shared_ptr<rclcpp::GuardCondition> notify_guard_condition_ = nullptr;
+  std::mutex notify_guard_condition_mutex_;
 
 private:
   template<typename TypeT, typename Function>
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -177,7 +177,7 @@ public:
 
   /// Return the notify guard condition.
   RCLCPP_PUBLIC
-  rclcpp::GuardCondition::SharedPtr
+  rclcpp::GuardCondition &
   get_notify_guard_condition();
 
 protected:
@@ -222,7 +222,7 @@ protected:
   std::vector<rclcpp::Waitable::WeakPtr> waitable_ptrs_;
   std::atomic_bool can_be_taken_from_;
   const bool automatically_add_to_executor_with_node_;
-  rclcpp::GuardCondition::SharedPtr notify_guard_condition_;
+  rclcpp::GuardCondition notify_guard_condition_;
 
 private:
   template<typename TypeT, typename Function>
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -16,12 +16,12 @@
 #define RCLCPP__CALLBACK_GROUP_HPP_
 
 #include <atomic>
-#include <memory>
 #include <mutex>
 #include <string>
 #include <vector>
 
 #include "rclcpp/client.hpp"
+#include "rclcpp/context.hpp"
 #include "rclcpp/guard_condition.hpp"
 #include "rclcpp/publisher_base.hpp"
 #include "rclcpp/service.hpp"
@@ -29,17 +29,6 @@
 #include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 #include "rclcpp/waitable.hpp"
-
-namespace rclcpp
-{
-namespace node_interfaces
-{
-
-/// forward declaration.
-class NodeBaseInterface;
-
-}
-}
 
 namespace rclcpp
 {
@@ -188,7 +177,7 @@ public:
   RCLCPP_PUBLIC
   std::shared_ptr<rclcpp::GuardCondition>
   create_notify_guard_condition(
-    const std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface> nodebase_ptr);
+    const rclcpp::Context::SharedPtr context_ptr);
 
   /// Return the notify guard condition created before.
   RCLCPP_PUBLIC
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -21,6 +21,8 @@
 #include <vector>
 
 #include "rclcpp/client.hpp"
+#include "rclcpp/context.hpp"
+#include "rclcpp/guard_condition.hpp"
 #include "rclcpp/publisher_base.hpp"
 #include "rclcpp/service.hpp"
 #include "rclcpp/subscription_base.hpp"
@@ -85,6 +87,7 @@ public:
    * group, or after, is irrelevant; the callback group will be automatically
    * added to the executor in either case.
    *
+   * \param[in] context_ptr Shared pointer to the context.
    * \param[in] group_type The type of the callback group.
    * \param[in] automatically_add_to_executor_with_node A boolean that
    *   determines whether a callback group is automatically added to an executor
@@ -92,6 +95,7 @@ public:
    */
   RCLCPP_PUBLIC
   explicit CallbackGroup(
+    rclcpp::Context::SharedPtr context_ptr,
     CallbackGroupType group_type,
     bool automatically_add_to_executor_with_node = true);
 
@@ -171,6 +175,11 @@ public:
   bool
   automatically_add_to_executor_with_node() const;
 
+  /// Return the notify guard condition.
+  RCLCPP_PUBLIC
+  rclcpp::GuardCondition::SharedPtr
+  get_notify_guard_condition();
+
 protected:
   RCLCPP_DISABLE_COPY(CallbackGroup)
 
@@ -213,6 +222,7 @@ protected:
   std::vector<rclcpp::Waitable::WeakPtr> waitable_ptrs_;
   std::atomic_bool can_be_taken_from_;
   const bool automatically_add_to_executor_with_node_;
+  rclcpp::GuardCondition::SharedPtr notify_guard_condition_;
 
 private:
   template<typename TypeT, typename Function>
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -16,6 +16,7 @@
 #define RCLCPP__CALLBACK_GROUP_HPP_
 
 #include <atomic>
+#include <memory>
 #include <mutex>
 #include <string>
 #include <vector>
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -16,9 +16,9 @@
 #define RCLCPP__CALLBACK_GROUP_HPP_
 
 #include <atomic>
+#include <functional>
 #include <memory>
 #include <mutex>
-#include <string>
 #include <vector>
 
 #include "rclcpp/client.hpp"
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -98,6 +98,10 @@ public:
     CallbackGroupType group_type,
     bool automatically_add_to_executor_with_node = true);
 
+  /// Default destructor.
+  RCLCPP_PUBLIC
+  ~CallbackGroup();
+
   template<typename Function>
   rclcpp::SubscriptionBase::SharedPtr
   find_subscription_ptrs_if(Function func) const
@@ -174,20 +178,20 @@ public:
   bool
   automatically_add_to_executor_with_node() const;
 
-  /// Create and return the notify guard condition.
+  /// Defer creating the notify guard condition and return it.
   RCLCPP_PUBLIC
   rclcpp::GuardCondition::SharedPtr
   create_notify_guard_condition(const rclcpp::Context::SharedPtr context_ptr);
 
-  /// Destroy the notify guard condition.
-  RCLCPP_PUBLIC
-  void
-  destroy_notify_guard_condition();
-
-  /// Return the notify guard condition.
+  /// Get the notify guard condition.
   RCLCPP_PUBLIC
   rclcpp::GuardCondition::SharedPtr
   get_notify_guard_condition();
+
+  /// Trigger the notify guard condition.
+  RCLCPP_PUBLIC
+  void
+  trigger_notify_guard_condition();
 
 protected:
   RCLCPP_DISABLE_COPY(CallbackGroup)
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -16,12 +16,12 @@
 #define RCLCPP__CALLBACK_GROUP_HPP_
 
 #include <atomic>
+#include <memory>
 #include <mutex>
 #include <string>
 #include <vector>
 
 #include "rclcpp/client.hpp"
-#include "rclcpp/context.hpp"
 #include "rclcpp/guard_condition.hpp"
 #include "rclcpp/publisher_base.hpp"
 #include "rclcpp/service.hpp"
@@ -29,6 +29,17 @@
 #include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 #include "rclcpp/waitable.hpp"
+
+namespace rclcpp
+{
+namespace node_interfaces
+{
+
+/// forward declaration.
+class NodeBaseInterface;
+
+}
+}
 
 namespace rclcpp
 {
@@ -87,7 +98,6 @@ public:
    * group, or after, is irrelevant; the callback group will be automatically
    * added to the executor in either case.
    *
-   * \param[in] context_ptr Shared pointer to the context.
    * \param[in] group_type The type of the callback group.
    * \param[in] automatically_add_to_executor_with_node A boolean that
    *   determines whether a callback group is automatically added to an executor
@@ -95,7 +105,6 @@ public:
    */
   RCLCPP_PUBLIC
   explicit CallbackGroup(
-    rclcpp::Context::SharedPtr context_ptr,
     CallbackGroupType group_type,
     bool automatically_add_to_executor_with_node = true);
 
@@ -175,9 +184,15 @@ public:
   bool
   automatically_add_to_executor_with_node() const;
 
-  /// Return the notify guard condition.
+  /// Create and return the notify guard condition.
   RCLCPP_PUBLIC
-  rclcpp::GuardCondition &
+  std::shared_ptr<rclcpp::GuardCondition>
+  create_notify_guard_condition(
+    const std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface> nodebase_ptr);
+
+  /// Return the notify guard condition created before.
+  RCLCPP_PUBLIC
+  std::weak_ptr<rclcpp::GuardCondition>
   get_notify_guard_condition();
 
 protected:
@@ -222,7 +237,8 @@ protected:
   std::vector<rclcpp::Waitable::WeakPtr> waitable_ptrs_;
   std::atomic_bool can_be_taken_from_;
   const bool automatically_add_to_executor_with_node_;
-  rclcpp::GuardCondition notify_guard_condition_;
+  // defer the creation of the guard condition
+  std::shared_ptr<rclcpp::GuardCondition> notify_guard_condition_ = nullptr;
 
 private:
   template<typename TypeT, typename Function>
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -181,12 +181,7 @@ public:
   /// Defer creating the notify guard condition and return it.
   RCLCPP_PUBLIC
   rclcpp::GuardCondition::SharedPtr
-  create_notify_guard_condition(const rclcpp::Context::SharedPtr context_ptr);
-
-  /// Get the notify guard condition.
-  RCLCPP_PUBLIC
-  rclcpp::GuardCondition::SharedPtr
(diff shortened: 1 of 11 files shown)


iuhilnehc-ynos:
this test case expects test result as

```
The following tests FAILED:
	 26 - test_add_callback_groups_to_executor__rmw_fastrtps_cpp (Failed)
	 27 - test_add_callback_groups_to_executor__rmw_fastrtps_dynamic_cpp (Failed)
```

after using https://github.com/ros2/rmw_fastrtps/pull/527, there is no failure.

wjwwood (approved):

fujitatomoya (approved):

asorbini:
Before merging this test, can we discuss whether attaching a node to two different executors should be allowed by the API? This isn't currently a requirement AFAIK, so making this change is not trivial at all.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
```diff
@@ -276,6 +276,68 @@ TYPED_TEST(TestAddCallbackGroupsToExecutor, one_node_many_callback_groups_many_e
...
+      options.callback_group = cb_grp;
+      auto subscription =
+        node->create_subscription<test_msgs::msg::Empty>("topic_name", qos, sub_callback, options);
+
+      // create a publisher to send data
+      auto publisher =
+        node->create_publisher<test_msgs::msg::Empty>("topic_name", qos);
+      publisher->publish(test_msgs::msg::Empty());
+      // wait for message to be published
+      std::this_thread::sleep_for(100ms);
```
wjwwood, line 324: When `subscription` and `publisher` go out of scope they will be removed from the graph, and the 100ms sleep is unlikely to be enough all the time, i.e. this will be flaky.

Is it important to the test that these be in the scope of the timer callback? If not, then move them out, if so, then maybe this isn't an issue with the middleware, but how you're using the API.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
```diff
@@ -276,6 +276,68 @@ TYPED_TEST(TestAddCallbackGroupsToExecutor, one_node_many_callback_groups_many_e
...
+    };
+
+  // timer on default callback group
+  timer = node->create_wall_timer(1s, timer_callback);
+  auto start = std::chrono::steady_clock::now();
+  while (!received_message &&
+    (std::chrono::steady_clock::now() - start) < 3s)
```
wjwwood, line 331: This time is also likely to be flaky in the long term. There's really no value in not setting it to a higher value.

You could also use a future and `spin_until_future_complete`.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
```diff
@@ -276,6 +276,68 @@ TYPED_TEST(TestAddCallbackGroupsToExecutor, one_node_many_callback_groups_many_e
...
+  {
+    rclcpp::spin_some(node);
```
wjwwood, line 333: I don't think it's important for the test, but this method will add and remove the node from an executor each time this is called, so it would probably be better to create an executor, add the node, and either spin in a thread or use spin until future complete with a timeout, rather than using one of the `rclcpp::spin*` functions.

wjwwood (changes requested):
I reexamined the test to check what @asorbini was asking about, and I think I found some flaws, but the general idea (i.e. using more than one executor per node via `add_callback_group()` and having a callback group wake an executor when something is added to it in order to consider new items) is a supported use case based on the API.

It might be that this is a bug in rclcpp and not the middlewares however. I need to look more deeply at the root cause of the issue. But I don't have time at the moment.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
iuhilnehc-ynos: > Is it important to the test that these be in the scope of the timer callback?

I want to use the simplest source code to emulate the reported example of this issue.
timer callback ---> service callback
create a subscription on the service callback
there is a remote publisher to publish a message to the subscription,
that's all to reproduce this issue.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
iuhilnehc-ynos: This kind of test code already existed in the `rclcpp` test, is it a problem here?

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
iuhilnehc-ynos: `100ms` might be caused this test case to be flaky, I'll update it.

wjwwood: > I want to use the simplest source code to emulate the reported example of this issue.

Yes, but do you understand that not keeping these in scope might cause issues? I don't think (I could be wrong) that is what the original problem was surrounding.

Also, sleeping in the timer callback will block anything else (either something you set up or functions of the node that are implementation details) from happening via callback (for a single threaded executor) so you should never do that if you can avoid it.

iuhilnehc-ynos: Please refer to https://github.com/ros2/rclcpp/issues/1611#issue-846709308 reported by user,

```
  void srv_callback(const std_srvs::srv::Trigger::Request::SharedPtr, std_srvs::srv::Trigger::Response::SharedPtr res){
    RCLCPP_INFO_STREAM(this->get_logger(), "Received request");
    new_msg_ = false;

    // Create subscription
    auto sub_options_ = rclcpp::SubscriptionOptions();
    sub_options_.callback_group = callback_group_;
    auto subscription = this->create_subscription<std_msgs::msg::Bool>(
          "topic_bool",
          rclcpp::QoS(1),
          std::bind(
            &CallbackGroupSrvServer::sub_callback,
            this,
            std::placeholders::_1),
          sub_options_
          );

    // Wait for new_msg
    while (rclcpp::ok()
           && new_msg_ == false) {
      rclcpp::sleep_for(std::chrono::seconds(1));
      RCLCPP_INFO_STREAM(this->get_logger(), "Sleep...");
    }

    res->message = "";
    res->success = true;
    return;
  }
```

`srv_callback` running on the main thread, and the `CallbackGroupSrvServer::sub_callback` running on the `callback_group_executor_thread_` thread, so it seems it's a normal case. (publishing message can run on the main thread), so I think it's not a problem.

iuhilnehc-ynos: And the user said `which we use a lot.`

iuhilnehc-ynos: I think that we'd better use the same case (but using three applications is unnecessary) to reproduce the issue by adding a regression test,

iuhilnehc-ynos: I don't want to argue about whether it is a good design or not. I mean It's just a use case somebody might use.

wjwwood: > I don't want to argue about whether it is a good design or not. I mean It's just a use case somebody might use.

Yes, but it is an anti-pattern (two of them):

- not keeping the pub/sub in scope and assuming some short period of time is long enough for the message to arrive
- blocking in a service/timer (or anything) callback

We could warn or prevent these if it were possible, but we cannot prevent the user from doing so or know when to warn and when not to.

This is a test, not some random user code, so we should avoid these issues. And doing so we should figure out what about the user reported code is the source of the problem. Is it the blocking in the callback? Is it the pub/sub going out of scope? Is it a bug in rclcpp? It's hard to tell without eliminating the obviously bad things to do.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
wjwwood: No, but it's not ideal either. The tests are pretty scattered in terms of age and quality, so they're not necessarily the best example for best practices.

Like I said, it's probably not affecting the test, but using a future and `spin_until_future_complete` is better. Also, since this test is to do with how nodes, callback groups, and executors interact, it's probably better to not add/remove the node to the executor in a loop...

iuhilnehc-ynos:
This kind of test source code seems not good (adding it into rclcpp might make other users wrongly use it in the future), close it.

fujitatomoya:
i do not know why we need to close this. after `rclcpp` is fixed, we can add this kind of test?

wjwwood:
> i do not know why we need to close this. after rclcpp is fixed, we can add this kind of test?

I don't think we needed to close it, but I do think the things I pointed out needed to be addressed.

I do not think, however, that we should have tests that copy user submitted code by including anti-patterns or otherwise not-recommended practices. Instead, the test should (ideally) test one potential issue at a time. It can be distilled from a user reported issue, but again it should be cleaned up and written as properly as possible.

fujitatomoya:
yeah okay that makes sense to me.

iuhilnehc-ynos:
reopen it after not using some anti-pattern source code.

iuhilnehc-ynos:
@wjwwood 

I have updated `rclcpp` source code by following https://github.com/ros2/rclcpp/issues/1611#issuecomment-843726301 if I understood correctly. 
Could you help to review it again?

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -225,18 +224,17 @@ Executor::add_callback_group_to_map(
...
-  if (is_new_node) {
-    rclcpp::node_interfaces::NodeBaseInterface::WeakPtr node_weak_ptr(node_ptr);
-    weak_nodes_to_guard_conditions_[node_weak_ptr] = node_ptr->get_notify_guard_condition();
-    if (notify) {
-      // Interrupt waiting to handle new node
-      rcl_ret_t ret = rcl_trigger_guard_condition(&interrupt_guard_condition_);
-      if (ret != RCL_RET_OK) {
-        throw_from_rcl_error(ret, "Failed to trigger guard condition on callback group add");
-      }
+  weak_groups_to_guard_conditions_[weak_group_ptr] =
```
fujitatomoya, line 227: just to confirm, this does not need to check if `weak_group_ptr` is already in the map, since it already checks with `get_associated_with_executor_atomic` above?

fujitatomoya (commented):
overall, i think it looks good.

rclcpp/src/rclcpp/executor.cpp
iuhilnehc-ynos: Yes

fujitatomoya:
@wjwwood friendly ping.

SteveMacenski:
Noting that this is causing a bug in the Humble migration and this fixes it for an issue in Nav2 linked above @clalancette

clalancette:
It looks like this needs a rebase, and another review for the latest changes.  Once that is done, we can hopefully merge it in for Rolling.

Backporting it to Humble is going to be tricky, as it changes ABI.  We'll have to consider another way to do that.

iuhilnehc-ynos:
@wjwwood friendly ping.

SteveMacenski:
Also a friendly ping :-) @clalancette

wjwwood:
Sorry for dropping this one.

So, what I was going to say last time I reviewed this, but then got distracted and didn't follow up (again sorry), is that this breaks API.

I think we should perhaps do a few things differently here:

- deprecate the existing API signature, and when it is used use the https://docs.ros2.org/galactic/api/rclcpp/namespacerclcpp_1_1contexts.html#a18a2108b667f7e582db2c9ef216870a6 function
  - This lets us tick-tock, as well as back port with minimal changes
- add new constructor with context argument (this pr already does this)
- consider adding a method to the node class (or a free function takes the node as an argument) which creates the callback group for you, so that users can create a callback group with a node rather than a context, in which case it uses the context from the node
  -  this is nice because then the user a) doesn't have to know what a context is, and b) the user isn't responsible for ensuring the same context is used for the callback group and the node (which is required)

Also that last point brings up the issue that we assume that the context used with the callback group and the entities later associated with the callback groups all use the context. We should probably assert that is the case somewhere.

wjwwood (changes requested):
This lgtm, except for the breaking API without deprecation, which I mentioned in my most recent comment. I think if we address that, then we can move forward with this.

rclcpp/src/rclcpp/callback_group.cpp
```diff
@@ -14,17 +14,20 @@
...
 #include "rclcpp/callback_group.hpp"
 
+#include <memory>
 #include <vector>
 
 using rclcpp::CallbackGroup;
 using rclcpp::CallbackGroupType;
 
 CallbackGroup::CallbackGroup(
+  rclcpp::Context::SharedPtr context_ptr,
```
wjwwood, line 24: Specifically this change is breaking API.

wjwwood:
Yet another option is that we could defer the creation of the guard condition for each callback group until they are added to an executor. At that time we could pass the context from the executor to the callback group, or even create the guard condition in the executor and give the guard condition to the callback group (dependency injection).

This approach prevents the callback group and executor's context from being different (which is not allowed) and it would also be possible to do without changing API, and maybe be easier to do without breaking ABI for the back ports.

iuhilnehc-ynos:
> Yet another option is that we could defer the creation of the guard condition for each callback group until they are added to an executor. 

> It needs some more thought, but I actually like this approach better.

Thanks for the guide. I'll dig into it to see if it works.

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17029/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11589/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17448/)

iuhilnehc-ynos:
fix cpplint issue and rebased as system_tests needs the latest cpplint, re-CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17029/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11590/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17449/)

iuhilnehc-ynos:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17031/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11592/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17451/)

rclcpp/src/rclcpp/callback_group.cpp
iuhilnehc-ynos: removed it because of deferring to create the guard condition in the callback gruop

iuhilnehc-ynos:
rerun CI for windows:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17457/)

iuhilnehc-ynos:
cc: @wjwwood

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -171,6 +174,17 @@ class CallbackGroup
   bool
   automatically_add_to_executor_with_node() const;
 
+  /// Create and return the notify guard condition.
+  RCLCPP_PUBLIC
+  std::shared_ptr<rclcpp::GuardCondition>
+  create_notify_guard_condition(
+    const rclcpp::Context::SharedPtr context_ptr);
```
wjwwood, line 181: nitpick:

```suggestion
  create_notify_guard_condition(const rclcpp::Context::SharedPtr context_ptr);
```

rclcpp/src/rclcpp/callback_group.cpp
```diff
@@ -97,6 +95,22 @@ CallbackGroup::automatically_add_to_executor_with_node() const
   return automatically_add_to_executor_with_node_;
 }
 
+std::shared_ptr<rclcpp::GuardCondition>
+CallbackGroup::create_notify_guard_condition(
+  const rclcpp::Context::SharedPtr context_ptr)
+{
+  notify_guard_condition_ =
+    std::make_unique<rclcpp::GuardCondition>(context_ptr);
```
wjwwood, line 103: It is technically ok as-is, but I'm curious why you used `make_unique` instead of `make_shared`?

wjwwood, line 103: I don't love that this implicitly resets the notify guard condition if it was already set. It sounds like a subtle bug waiting to happen.

I kind of think that we should either have an interface like this:

- `create_notify_guard_condition()` -> throws if already has one
- `destroy_notify_guard_condition()` -> throws if not already set
- `get_notify_guard_condition()` -> could return nullptr

Or something like this:

- `exchange_notify_guard_condition()` -> returns previous pointer value, could be nullptr
  - the guard condition would be created by the executor and passed in here
- `get_notify_guard_condition()` -> can return nullptr
  - this could store/return a `weak_ptr`

Both of these are more explicit about changing controlling the lifecycle of the shared pointer.

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -171,6 +174,17 @@ class CallbackGroup
...
+
+  /// Return the notify guard condition created before.
+  RCLCPP_PUBLIC
+  std::weak_ptr<rclcpp::GuardCondition>
+  get_notify_guard_condition();
```
wjwwood, line 186: This is weird to me because the guard condition is stored as a shared_ptr, so why return a weak_ptr?

wjwwood (changes requested):

wjwwood:
Thanks for iterating!

rclcpp/include/rclcpp/callback_group.hpp
iuhilnehc-ynos: I think you asked me to remove the `const` as the constructor of `rclcpp::GuardCondition` uses a non-const context.

rclcpp/src/rclcpp/callback_group.cpp
iuhilnehc-ynos: Sorry. I used the `unique_ptr` for the member at the beginning. For some reason, I changed to `shared_ptr` for it but forgot to update this code.

iuhilnehc-ynos: The function is not intended to create the guard condition multiple times in the callback group. I'll update from `create_notify_guard_condition` to `get_notify_guard_condition`, and also add a check statement.

rclcpp/include/rclcpp/callback_group.hpp
iuhilnehc-ynos: Because this get_notify_guard_condition could return nullptr as the creation function is a deferred function. (I don't know what kind of this case could be.)

- use shared_ptr

```cpp
auto gc = get_notify_guard_condition();
if (gc)
   gc->trigger();
```

- use weak_ptr

```cpp
if (auto gc = get_notify_guard_condition().lock() )
   gc->trigger();
```

so I decided to use the weak_ptr for this function.

rclcpp/src/rclcpp/callback_group.cpp
iuhilnehc-ynos: Sorry, I didn't think over your suggestions before.
I think the first one you suggested is good. Thank you for the detailed guidance.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -679,6 +690,9 @@ void
 Executor::wait_for_work(std::chrono::nanoseconds timeout)
 {
   TRACEPOINT(rclcpp_executor_wait_for_work, timeout.count());
+  // To keep the guard conditions alive as they could be destroyed when the executor removes nodes.
+  // The lifetime of these guard conditions should be guaranteed during rcl_wait().
+  std::vector<GuardCondition::SharedPtr> alive_guard_condition_ptrs;
```
iuhilnehc-ynos, line 695: @wjwwood 

Is it worth introducing this code not to break the CallbackGroup API and make it defer to create the guard condition in the callback group?
I was worried about adding extra code in this kind of the main loop. I appreciate it if you have any suggestions.

iuhilnehc-ynos:
retry CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17090/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11641/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17509/)

iuhilnehc-ynos:
CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17092/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11643/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17511/)

SteveMacenski:
This all passed now!

rclcpp/src/rclcpp/executor.cpp
fujitatomoya: @wjwwood friendly ping.

fujitatomoya:
@wjwwood could you take an another look on your previous comments? i believe this has some good impact for ROS 2 system and user application.

alsora:
We have been cherry-picking this PR in our Humble branch and it fixed the problem with callback groups

alsora (approved):

clalancette (changes requested):
I rebased this PR and made a few minor changes (just to the include files).

Overall, I think the idea behind this PR is good, but I will point out that it is *not* thread-safe.  In particular, the lock around `get_notify_guard_condition` is useless; it is only held while manipulating the shared_ptr, but is then dropped when we return the object.  Anything we do with that object afterwards is not thread-safe.

What I think we should do instead is to make sure we do any operations on the callback-group guard condition with the lock held.  That would require us to add some additional APIs (like a new `trigger` method), modify some of the existing ones (change `destroy_notify_guard_condition` to only destroy the guard condition if it was created and not error out), and otherwise clean some things up.  With those things done, we should be able to make this thread-safe and move forward with this.

iuhilnehc-ynos:
One callback group can only have one [notify_guard_condition_](https://github.com/ros2/rclcpp/blob/8dd0ec85a70412f707439dda4843ad0f91d7ddd0/rclcpp/include/rclcpp/callback_group.hpp#L235), the [notify_guard_condition_mutex_](https://github.com/ros2/rclcpp/blob/8dd0ec85a70412f707439dda4843ad0f91d7ddd0/rclcpp/include/rclcpp/callback_group.hpp#L236) is jsut for these [three operations](https://github.com/ros2/rclcpp/blob/8dd0ec85a70412f707439dda4843ad0f91d7ddd0/rclcpp/src/rclcpp/callback_group.cpp#L110-L139), and the callback group can be [used in only one executor](https://github.com/ros2/rclcpp/blob/8dd0ec85a70412f707439dda4843ad0f91d7ddd0/rclcpp/src/rclcpp/executor.cpp#L208-L212).

so it seems unnecessary to add extra mutex to make the callback-group guard condition thread-safe.

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17533/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12068/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18088/)

iuhilnehc-ynos:
retry CI:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18098/)

fujitatomoya:
@iuhilnehc-ynos thanks for the effort on this, i will take a look at this today. I think this is very helpful for user application.

fujitatomoya (approved):
lgtm

fujitatomoya:
dependent and above CI all green, i will go ahead to merge this.

fujitatomoya:
@iuhilnehc-ynos i think we can backport this to humble, right? it would be really nice to backport since this affects user application behavior.

fujitatomoya:
@Mergify backport this to humble

Just create backport PR so that we will not forget, let's give it a couple of weeks to make sure there is no degression, then we can merge the backport PR to humble.

PR #1641, opened 2021-04-26
Returns CancelResponse::REJECT while goal handle failed to transit to CANCELING state

KavenYau:
For fixing #1599. Some conversations in https://github.com/ros2/rclcpp/pull/1635 are helpful to this PR.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/server.hpp
+++ b/rclcpp_action/include/rclcpp_action/server.hpp
@@ -372,6 +372,9 @@ protected:
         try {
           goal_handle->_cancel_goal();
         } catch (const rclcpp::exceptions::RCLError & ex) {
+          RCLCPP_DEBUG(
+            rclcpp::get_logger("rclcpp_action"),
+            "Failed to cancel goal in call_handle_cancel_callback: %s", ex.what());
           return CancelResponse::REJECT;
         }
       }
--- a/rclcpp_action/include/rclcpp_action/server.hpp
+++ b/rclcpp_action/include/rclcpp_action/server.hpp
@@ -369,7 +369,11 @@ protected:
     if (goal_handle) {
       resp = handle_cancel_(goal_handle);
       if (CancelResponse::ACCEPT == resp) {
-        goal_handle->_cancel_goal();
+        try {
+          goal_handle->_cancel_goal();
+        } catch (const rclcpp::exceptions::RCLError & ex) {
+          return CancelResponse::REJECT;
+        }
       }
     }
     return resp;
--- a/rclcpp_action/test/test_server.cpp
+++ b/rclcpp_action/test/test_server.cpp
@@ -1234,14 +1234,10 @@ public:
         this->TryLockFor(lock, std::chrono::milliseconds(1000));
         return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE;
       },
-      [this](std::shared_ptr<GoalHandle> handle) {
+      [this](std::shared_ptr<GoalHandle>) {
         // instead of making a deadlock, check if it can acquire the lock in a second
         std::unique_lock<std::recursive_timed_mutex> lock(server_mutex_, std::defer_lock);
         this->TryLockFor(lock, std::chrono::milliseconds(1000));
-        // TODO(KavenYau): this check may become obsolete with https://github.com/ros2/rclcpp/issues/1599
-        if (!handle->is_active()) {
-          return rclcpp_action::CancelResponse::REJECT;
-        }
         return rclcpp_action::CancelResponse::ACCEPT;
       },
       [this](std::shared_ptr<GoalHandle> handle) {
@@ -1316,6 +1312,9 @@ TEST_F(TestDeadlockServer, deadlock_while_succeed_and_canceled)
   send_goal_request(node_, uuid1_);
   std::thread t(&TestDeadlockServer::GoalSucceeded, this);
   rclcpp::sleep_for(std::chrono::milliseconds(50));
-  send_cancel_request(node_, uuid1_);
+  auto response_ptr = send_cancel_request(node_, uuid1_);
+
+  // current goal handle is not cancelable, so it returns ERROR_REJECTED
+  EXPECT_EQ(CancelResponse::ERROR_REJECTED, response_ptr->return_code);
   t.join();
 }


rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -369,7 +369,9 @@ class Server : public ServerBase, public std::enable_shared_from_this<Server<Act
     if (goal_handle) {
       resp = handle_cancel_(goal_handle);
       if (CancelResponse::ACCEPT == resp) {
-        goal_handle->_cancel_goal();
+        if (!goal_handle->_cancel_goal()) {
+          return CancelResponse::REJECT;
+        }
```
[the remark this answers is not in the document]
KavenYau: Sure, using `try & catch` statement would be less change. I do agree with you.

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
```diff
@@ -87,10 +87,11 @@ class ServerGoalHandleBase
   void
   _succeed();
 
+  /// Transition the goal to canceling state if it never reached a terminal state.
   /// \internal
   RCLCPP_ACTION_PUBLIC
-  void
-  _cancel_goal();
+  bool
+  _cancel_goal() noexcept;
```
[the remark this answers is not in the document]
KavenYau: You're right, I have reset and resubmited a commit.

fujitatomoya (approved):
looks good to me. i guess we could do backport for Galactic/Foxy w/o breaking ABI.

fujitatomoya:
@ivanpauno @wjwwood @jacobperron could either of you take a look for another review when you got time?

rclcpp_action/test/test_server.cpp
```diff
@@ -1316,6 +1312,9 @@ TEST_F(TestDeadlockServer, deadlock_while_succeed_and_canceled)
   send_goal_request(node_, uuid1_);
   std::thread t(&TestDeadlockServer::GoalSucceeded, this);
   rclcpp::sleep_for(std::chrono::milliseconds(50));
```
ivanpauno, line 1314: why are we calling `sleep_for()`? isn't this extremly flaky?

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -369,7 +369,11 @@ class Server : public ServerBase, public std::enable_shared_from_this<Server<Act
     if (goal_handle) {
       resp = handle_cancel_(goal_handle);
       if (CancelResponse::ACCEPT == resp) {
-        goal_handle->_cancel_goal();
+        try {
+          goal_handle->_cancel_goal();
+        } catch (const rclcpp::exceptions::RCLError & ex) {
+          return CancelResponse::REJECT;
```
jacobperron, line 375: It might be nice to log a debug message here.

rclcpp_action/test/test_server.cpp
KavenYau: The aim is to ensure that this thread `std::thread t(&TestDeadlockServer::GoalSucceeded, this);` could be running first.

rclcpp_action/include/rclcpp_action/server.hpp
KavenYau: :+1:  It's a good suggestion. I have pushed a commit for it.

rclcpp_action/test/test_server.cpp
ivanpauno: My comment is because `sleep_for` cannot ensure that.

KavenYau: I do agree with you that `sleep_for` cannot ensure that, but I think it works in most of the time. I just modified base on original test case.

Do you think we should improve `TestDeadlockServer` test cases in this PR?

ivanpauno: > I just modified base on original test case

That's true, sounds fair to keep it for the moment.

rclcpp_action/include/rclcpp_action/server.hpp
```diff
@@ -369,7 +369,11 @@ class Server : public ServerBase, public std::enable_shared_from_this<Server<Act
     if (goal_handle) {
       resp = handle_cancel_(goal_handle);
       if (CancelResponse::ACCEPT == resp) {
-        goal_handle->_cancel_goal();
+        try {
+          goal_handle->_cancel_goal();
+        } catch (const rclcpp::exceptions::RCLError & ex) {
```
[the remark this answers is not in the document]
KavenYau: According to this file https://github.com/ros2/rclcpp/blob/master/rclcpp_action/src/server_goal_handle.cpp, I see that all errors of transiting state failed are throwing `RCLError` exception. Is it better to regard them as the same error? What do you think?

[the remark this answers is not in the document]
ivanpauno: > And if you are talking about adding a new exception to clarify this type of error. I am not sure if it would affect users‘ implementation of rclcpp_action because they should catch another exception?

Yes, I meant that.
`rcl_action_update_goal_state` has two return values, one is the one [we want to catch](https://github.com/ros2/rcl/blob/414fdbd479b83f2b8af8235ad1fa3aeba2b0027f/rcl_action/src/rcl_action/goal_handle.c#L99), the [other one](https://github.com/ros2/rcl/blob/414fdbd479b83f2b8af8235ad1fa3aeba2b0027f/rcl_action/src/rcl_action/goal_handle.c#L90) shouldn't be catched.

It would be good to have a more specific exception for the one we want to catch.
@jacobperron what do you think?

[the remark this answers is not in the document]
jacobperron: I agree that it would be nicer to have more specific exceptions. I also agree that we shouldn't expect a null goal handle. However, it may not be API compatible to change exception types, so I suggest we do that for H-turtle. If we leave this PR as it is, then we can backport this change to Galactic (and maybe Foxy).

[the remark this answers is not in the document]
ivanpauno: we are already doing similar stuff in other places (e.g. [here](https://github.com/ros2/rclcpp/blob/1037822a63330495dcf0de5e8f20544375a5f116/rclcpp_action/src/server_goal_handle.cpp#L120-L122)), so I'm ok with it.

> However, it may not be API compatible to change exception types

If the new exception is a subclass of `RCLError`, it will be both API and ABI compatible.

ivanpauno (approved):

ivanpauno:
Adding it to the galactic board, I guess we want to backport this.

ivanpauno:
@Mergifyio backport galactic

Issue #1339, opened 2020-09-28
Publisher and subscription options pass dangling pointer to get_rcl_allocator


stevewolter:
In rclcpp/include/rclcpp/subscription_options.hpp and rclcpp/include/rclcpp/publisher_options.hpp, the conversion to RCL options ("to_rcl_publisher_options") passes a reference to a C++ allocator that goes out of scope at the end of the function:

    auto message_alloc = std::make_shared<MessageAllocatorT>(*this->get_allocator().get());
    result.allocator = get_rcl_allocator(*message_alloc);

This could cause unexpected problems for users of custom allocators when they reference the allocator in an overload of get_rcl_allocator. Standard ROS code is not affected, because neither std::allocator nor tlsf_allocator carry internal state. This is hard to fix, because neither rcl_publisher_options_t nor rcl_allocator_t have destructors (so they can't just create their own reference to the shared_ptr).

PR #1643, opened 2021-04-27
use dynamic_pointer_cast to detect allocator mismatch in intra process manager

wjwwood:
We believe this is the root cause of https://github.com/ros2/demos/issues/501

PR #1647, opened 2021-04-29
Keep custom allocator in publisher and subscription options alive.

hidmic:
Also, enforce an allocator bound to void is used to avoid surprises. Connected to #1643.

I used `mutable` and kept `MessageT` as template parameter to keep changes local (i.e. not force a change all the way up the call stack due template signature and `const`-ness changes).

CI up to `rclcpp` and `demo_nodes_cpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14524/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9281/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12190/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14627/)

though I don't think we can easily write a regression test for this (i.e. in a way that's accurate and cross-platform). 

Fixes https://github.com/ros2/rclcpp/issues/1339

### DIFF
--- a/rclcpp/include/rclcpp/publisher_options.hpp
+++ b/rclcpp/include/rclcpp/publisher_options.hpp
@@ -104,15 +104,18 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   get_allocator() const
   {
     if (!this->allocator) {
-      if constexpr (std::is_same_v<Allocator, std::allocator<void>>) {
-        auto g_instance = std::make_shared<std::allocator<void>>();
-        return g_instance;
-      } else {
-        throw std::runtime_error("allocator is nullptr");
+      if (!allocator_storage_) {
+        allocator_storage_ = std::make_shared<Allocator>();
       }
+      return allocator_storage_;
     }
     return this->allocator;
   }
+
+private:
+  // This is a temporal workaround, to make sure that get_allocator()
+  // always returns a copy of the same allocator.
+  mutable std::shared_ptr<Allocator> allocator_storage_;
 };
 
 using PublisherOptions = PublisherOptionsWithAllocator<std::allocator<void>>;
--- a/rclcpp/include/rclcpp/publisher_options.hpp
+++ b/rclcpp/include/rclcpp/publisher_options.hpp
@@ -17,6 +17,7 @@
 
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include "rcl/publisher.h"
@@ -65,11 +66,11 @@ template<typename Allocator>
 struct PublisherOptionsWithAllocator : public PublisherOptionsBase
 {
   static_assert(
-    std::is_void<typename std::allocator_traits<Allocator>::value_type>::value,
+    std::is_void_v<typename std::allocator_traits<Allocator>::value_type>,
     "Publisher allocator value type must be void");
 
   /// Optional custom allocator.
-  mutable std::shared_ptr<Allocator> allocator = nullptr;
+  std::shared_ptr<Allocator> allocator = nullptr;
 
   PublisherOptionsWithAllocator<Allocator>() {}
 
@@ -103,7 +104,12 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   get_allocator() const
   {
     if (!this->allocator) {
-      this->allocator = std::make_shared<Allocator>();
+      if constexpr (std::is_same_v<Allocator, std::allocator<void>>) {
+        auto g_instance = std::make_shared<std::allocator<void>>();
+        return g_instance;
+      } else {
+        throw std::runtime_error("allocator is nullptr");
+      }
     }
     return this->allocator;
   }
--- a/rclcpp/include/rclcpp/publisher_options.hpp
+++ b/rclcpp/include/rclcpp/publisher_options.hpp
@@ -64,8 +64,12 @@ struct PublisherOptionsBase
 template<typename Allocator>
 struct PublisherOptionsWithAllocator : public PublisherOptionsBase
 {
+  static_assert(
+    std::is_void<typename std::allocator_traits<Allocator>::value_type>::value,
+    "Publisher allocator value type must be void");
+
   /// Optional custom allocator.
-  std::shared_ptr<Allocator> allocator = nullptr;
+  mutable std::shared_ptr<Allocator> allocator = nullptr;
 
   PublisherOptionsWithAllocator<Allocator>() {}
 
@@ -80,10 +84,7 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   to_rcl_publisher_options(const rclcpp::QoS & qos) const
   {
     rcl_publisher_options_t result = rcl_publisher_get_default_options();
-    using AllocatorTraits = std::allocator_traits<Allocator>;
-    using MessageAllocatorT = typename AllocatorTraits::template rebind_alloc<MessageT>;
-    auto message_alloc = std::make_shared<MessageAllocatorT>(*this->get_allocator().get());
-    result.allocator = rclcpp::allocator::get_rcl_allocator<MessageT>(*message_alloc);
+    result.allocator = rclcpp::allocator::get_rcl_allocator<void>(*this->get_allocator());
     result.qos = qos.get_rmw_qos_profile();
     result.rmw_publisher_options.require_unique_network_flow_endpoints =
       this->require_unique_network_flow_endpoints;
@@ -102,7 +103,7 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   get_allocator() const
   {
     if (!this->allocator) {
-      return std::make_shared<Allocator>();
+      this->allocator = std::make_shared<Allocator>();
     }
     return this->allocator;
   }
--- a/rclcpp/include/rclcpp/subscription_options.hpp
+++ b/rclcpp/include/rclcpp/subscription_options.hpp
@@ -126,20 +126,22 @@ struct SubscriptionOptionsWithAllocator : public SubscriptionOptionsBase
     return result;
   }
 
-  /// Get the allocator, creating one if needed.
   std::shared_ptr<Allocator>
   get_allocator() const
   {
     if (!this->allocator) {
-      if constexpr (std::is_same_v<Allocator, std::allocator<void>>) {
-        auto g_instance = std::make_shared<std::allocator<void>>();
-        return g_instance;
-      } else {
-        throw std::runtime_error("allocator is nullptr");
+      if (!allocator_storage_) {
+        allocator_storage_ = std::make_shared<Allocator>();
       }
+      return allocator_storage_;
     }
     return this->allocator;
   }
+
+private:
+  // This is a temporal workaround, to make sure that get_allocator()
+  // always returns a copy of the same allocator.
+  mutable std::shared_ptr<Allocator> allocator_storage_;
 };
 
 using SubscriptionOptions = SubscriptionOptionsWithAllocator<std::allocator<void>>;
--- a/rclcpp/include/rclcpp/subscription_options.hpp
+++ b/rclcpp/include/rclcpp/subscription_options.hpp
@@ -18,6 +18,7 @@
 #include <chrono>
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include "rclcpp/callback_group.hpp"
@@ -87,11 +88,11 @@ template<typename Allocator>
 struct SubscriptionOptionsWithAllocator : public SubscriptionOptionsBase
 {
   static_assert(
-    std::is_void<typename std::allocator_traits<Allocator>::value_type>::value,
+    std::is_void_v<typename std::allocator_traits<Allocator>::value_type>,
     "Subscription allocator value type must be void");
 
   /// Optional custom allocator.
-  mutable std::shared_ptr<Allocator> allocator = nullptr;
+  std::shared_ptr<Allocator> allocator = nullptr;
 
   SubscriptionOptionsWithAllocator<Allocator>() {}
 
@@ -130,7 +131,12 @@ struct SubscriptionOptionsWithAllocator : public SubscriptionOptionsBase
   get_allocator() const
   {
     if (!this->allocator) {
-      this->allocator = std::make_shared<Allocator>();
+      if constexpr (std::is_same_v<Allocator, std::allocator<void>>) {
+        auto g_instance = std::make_shared<std::allocator<void>>();
+        return g_instance;
+      } else {
+        throw std::runtime_error("allocator is nullptr");
+      }
     }
     return this->allocator;
   }
--- a/rclcpp/include/rclcpp/subscription_options.hpp
+++ b/rclcpp/include/rclcpp/subscription_options.hpp
@@ -86,8 +86,12 @@ struct SubscriptionOptionsBase
 template<typename Allocator>
 struct SubscriptionOptionsWithAllocator : public SubscriptionOptionsBase
 {
+  static_assert(
+    std::is_void<typename std::allocator_traits<Allocator>::value_type>::value,
+    "Subscription allocator value type must be void");
+
   /// Optional custom allocator.
-  std::shared_ptr<Allocator> allocator = nullptr;
+  mutable std::shared_ptr<Allocator> allocator = nullptr;
 
   SubscriptionOptionsWithAllocator<Allocator>() {}
 
@@ -107,10 +111,7 @@ struct SubscriptionOptionsWithAllocator : public SubscriptionOptionsBase
   to_rcl_subscription_options(const rclcpp::QoS & qos) const
   {
     rcl_subscription_options_t result = rcl_subscription_get_default_options();
-    using AllocatorTraits = std::allocator_traits<Allocator>;
-    using MessageAllocatorT = typename AllocatorTraits::template rebind_alloc<MessageT>;
-    auto message_alloc = std::make_shared<MessageAllocatorT>(*this->get_allocator().get());
-    result.allocator = allocator::get_rcl_allocator<MessageT>(*message_alloc);
+    result.allocator = allocator::get_rcl_allocator<void>(*this->get_allocator());
     result.qos = qos.get_rmw_qos_profile();
     result.rmw_subscription_options.ignore_local_publications = this->ignore_local_publications;
     result.rmw_subscription_options.require_unique_network_flow_endpoints =
@@ -129,7 +130,7 @@ struct SubscriptionOptionsWithAllocator : public SubscriptionOptionsBase
   get_allocator() const
   {
     if (!this->allocator) {
-      return std::make_shared<Allocator>();
+      this->allocator = std::make_shared<Allocator>();
     }
     return this->allocator;
   }
--- a/rclcpp/test/rclcpp/test_intra_process_manager_with_allocators.cpp
+++ b/rclcpp/test/rclcpp/test_intra_process_manager_with_allocators.cpp
@@ -18,6 +18,7 @@
 #include <list>
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <utility>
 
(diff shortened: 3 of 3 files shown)


rclcpp/include/rclcpp/publisher_options.hpp
```diff
@@ -64,8 +64,12 @@ struct PublisherOptionsBase
 template<typename Allocator>
 struct PublisherOptionsWithAllocator : public PublisherOptionsBase
 {
+  static_assert(
+    std::is_void<typename std::allocator_traits<Allocator>::value_type>::value,
+    "Publisher allocator value type must be void");
```
ivanpauno, line 69: why?
could't it be rebinded to `void`?

rclcpp/include/rclcpp/publisher_options.hpp
```diff
@@ -64,8 +64,12 @@ struct PublisherOptionsBase
...
+
   /// Optional custom allocator.
-  std::shared_ptr<Allocator> allocator = nullptr;
+  mutable std::shared_ptr<Allocator> allocator = nullptr;
```
ivanpauno, line 72: mmmm, this is not really desired.
I can't come up with a good solution though

rclcpp/include/rclcpp/subscription_options.hpp
```diff
@@ -86,8 +86,12 @@ struct SubscriptionOptionsBase
 template<typename Allocator>
 struct SubscriptionOptionsWithAllocator : public SubscriptionOptionsBase
 {
+  static_assert(
+    std::is_void<typename std::allocator_traits<Allocator>::value_type>::value,
+    "Subscription allocator value type must be void");
```
ivanpauno, line 91: same comment here

hidmic:
Argh, the same warning is happening in the demo itself -.-

rclcpp/include/rclcpp/publisher_options.hpp
```diff
@@ -102,7 +103,7 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   get_allocator() const
   {
     if (!this->allocator) {
-      return std::make_shared<Allocator>();
+      this->allocator = std::make_shared<Allocator>();
```
wjwwood, line 106: I'm not sure this is such a good idea. This would be a weird side-effect of `to_rcl_publisher_options`.

Instead, maybe should leave this as-is (since this methood's purpose is to provide an allocator, even if one is not set, but not to ensure it stays alive). And then we can have logic in `to_rcl_publisher_options` like this:

```c++
if (this->allocator) {
  result.allocator = rclcpp::allocator::get_rcl_allocator<void>(*this->get_allocator());
} else {
  result.allocator = rcl_get_default_allocator();
}
```

Or perhaps:

```c++
if (this->allocator) {
  result.allocator = rclcpp::allocator::get_rcl_allocator<void>(*this->get_allocator());
} else {
  static std::allocator<void> default_std_allocator;
  result.allocator = rclcpp::allocator::get_rcl_allocator<void>(default_std_allocator);
}
```

rclcpp/include/rclcpp/publisher_options.hpp
hidmic: It can be, but it's unnecessarily surprising. There's no setter API, `allocator` is a plain member-field. So a non-`void` allocator instance set will never be used, and we're forced to keep the allocator that is actually used separately. 

Also, what would be the point of allowing custom allocators of arbitrary type T? What's the use of accepting e.g. `std::allocator<std::vector<double>>`?

rclcpp/include/rclcpp/publisher_options.hpp
hidmic: Yeah, this is awful, but it's the smallest change possible.

rclcpp/test/rclcpp/test_intra_process_manager_with_allocators.cpp
```diff
@@ -57,7 +58,10 @@ struct MyAllocator
       return nullptr;
     }
     num_allocs++;
-    return static_cast<T *>(std::malloc(size * sizeof(T)));
+    // Use sizeof(char) in place for sizeof(void)
+    constexpr size_t value_size = sizeof(
+      typename std::conditional<!std::is_void<T>::value, T, char>::type);
```
wjwwood, line 63: I guess this is fine. I spent some time looking at libcxx's code and I couldn't find anything in the `void` specialization that avoids `sizeof(void)`, but I think that's technically a GCC extension. I wondered if `uint8_t` is better than `char` here (even though they are functionally the same), and I was actually looking for what they did when I couldn't find logic similar to this in their `allocate` function for `std::allocator`.

rclcpp/include/rclcpp/publisher_options.hpp
ivanpauno: Sounds good, I guess this doesn't break any existing code ...

rclcpp/test/rclcpp/test_intra_process_manager_with_allocators.cpp
wjwwood: I was actually hoping to find something on the `std::allocator` class or the `std::allocator_traits` that we could use instead, e.g. something like `std::allocator<T>::element_size`, but it doesn't exist.

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: I'm personally fine with this check. If anyone was passing something different before, it wasn't doing what they expected anyways.

rclcpp/include/rclcpp/publisher_options.hpp
ivanpauno: unrelated question, does `std::shared_ptr<Allocator>` makes sense at all?

IIUC allocators copy constructors must not throw and must "maintain the same resources" as the original one.
This also apply when constructing from a rebinded version.

From [cppreference](https://en.cppreference.com/w/cpp/named_req/Allocator):

Declaration | Effect | Requirements
-- | -- | --
A a1(a); <br> A a1 = a; | Copy-constructs a1 such that a1 == a. (Note: Every Allocator also satisfies CopyConstructible.) | Does not throw exceptions.
A a(b) | Constructs a such that B(a)==b and A(b)==a. (Note: This implies that all allocators related by rebind maintain each other's resources, such as memory pools.) | Does not throw exceptions.

i.e.: If the allocator implementation is stateful, it seems to already be a requirement that the internal state may be shared by difference allocator instances.

ivanpauno: > ```c++
> if (this->allocator) {
>  result.allocator = rclcpp::allocator::get_rcl_allocator<void>(*this->get_allocator());
>} else {
>  result.allocator = rcl_get_default_allocator();
>}
>```

The problem with this options is that IDK what to do in places like here:

https://github.com/ros2/rclcpp/blob/79c2dd8e8b3e73366c59474b91271a2fd57954bc/rclcpp/include/rclcpp/subscription_factory.hpp#L93

if `allocator` is `nullptr`, it doesn't sound like a good idea to create a new instance because that might involve creating new resources in custom allocators (though that sounds valid for `std::allocator`).

> Or perhaps:
> 
> ```c++
> if (this->allocator) {
>   result.allocator = rclcpp::allocator::get_rcl_allocator<void>(*this->get_allocator());
>} else {
>  static std::allocator<void> default_std_allocator;
>  result.allocator = rclcpp::allocator::get_rcl_allocator<void>(default_std_allocator);
>}
>```

That one sounds valid for a `std::allocator`, but it doesn't for custom ones (because the static allocator might be using resources unnecessarily).

Maybe I can specialize the method for `std::allocator<void>`, which returns an static instance of it.
For custom allocators, an exception is thrown if `nullptr`.

ivanpauno: See https://github.com/ros2/rclcpp/pull/1647/commits/6ff1246a3de528ee829628453adac269a2b25709.

(addressing comments here because Mich is OoO and wanted to get this one in)

rclcpp/include/rclcpp/publisher_options.hpp
ivanpauno: Resolving this one, continuing discussion in https://github.com/ros2/rclcpp/pull/1647#discussion_r623376683.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14532/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9288/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12197/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14638/)

rclcpp/include/rclcpp/publisher_options.hpp
```diff
@@ -102,7 +104,12 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   get_allocator() const
   {
     if (!this->allocator) {
-      return std::make_shared<Allocator>();
+      if constexpr (std::is_same_v<Allocator, std::allocator<void>>) {
+        auto g_instance = std::make_shared<std::allocator<void>>();
+        return g_instance;
+      } else {
+        throw std::runtime_error("allocator is nullptr");
```
wjwwood, line 111: Why can we not make an allocator if it isn't std::allocator?

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: > unrelated question, does std::shared_ptr<Allocator> makes sense at all?

No, see: https://github.com/ros2/rclcpp/issues/849#issuecomment-539758462

I wanted to remove all use of `shared_ptr<AllocatorT>`, but haven't been able to justify the churn caused by it. We should do it though, and I have been doing it as I go, where I can.

> if allocator is nullptr, it doesn't sound like a good idea to create a new instance because that might involve creating new resources in custom allocators (though that sounds valid for std::allocator).

I don't think that's an issue. If the allocator is nullptr, but the template argument is a custom allocator, then the caller has to know we'll create one for them. It's no different than the std::vector constructor, for example:

https://en.cppreference.com/w/cpp/container/vector/vector

Notice that the ones that take the allocator as an argument have a default value of `const Allocator& alloc = Allocator()`, so us creating one when it is needed and the member is nullptr, really isn't different.

Long term, maybe having a constructor for the options that optionally takes an allocator is the right thing to do, e.g.:

```c++
PublisherOptionsWithAllocator(const Allocator & allocator = Allocator())
: allocator(std::make_shared<Allocator>(allocator)) {}
```

Later we'd want to change the member of the options to be a non-`shared_ptr`.

ivanpauno: > I don't think that's an issue. If the allocator is nullptr, but the template argument is a custom allocator, then the caller has to know we'll create one for them. It's no different than the std::vector constructor, for example:

The difference is that we call `*Options.get_allocator()` multiple times, not only once.
So when `options.allocator = nullptr`, we're going to create multiple new instances of the allocator type instead of only one and copies of that one when `options.allocator != nullptr`.

The version using a `static` works if the compiler would guarantee that they are initialized lazely, but that's not guaranteed and if `options.allocator != nullptr` we may create two instances of the allocator (which might matter if the allocator is stateful).

ivanpauno: > The difference is that we call *Options.get_allocator() multiple times, not only once.

Actually, I think we might be calling these methods only once.
I will double check on Monday ...

rclcpp/include/rclcpp/publisher_options.hpp
ivanpauno: With stateless allocators, a new instance is the same than a copy of an old instance.
A custom allocator might be stateful, and thus a new instance and a copy of an old instance might not be the same.

I wondered about the possibility of `get_allocator()` being called multiple times, and retrieving a difference instance each time, which would be unexpected.

I didn't like the `static` variable alternative, because it might use resources innecesarly if `options.allocator != nullptr` and the allocator is stateful (and statics are not guaranteed to be initialized lazely).

hidmic: I think Ivan's right. I don't see a solution for stateful custom allocators other than throwing like this. Attempting lazy initialization has its own set of issues, as discussed above, and won't do for allocators that are not default-constructible (AFAIK no allocator is required to).

I will say though that using `if constexpr` prevents this patch from being backported to Foxy.

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: If we're calling it only once, I'd avoid throwing. If we're calling more than once I'd advocate for a second storage for the "default", construct it lazily, then return it from then on.

Either way, I think default constructing one when the user doesn't provide one is ok, it's just the current architecture leaves us in a weird place, as I said above about the constructor.

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: Well, I think it comes down to whether or not we return multiple copies. I agree that's a problem, but if we return only one new instance and reuse it from then on, then I don't think there's a problem in making one on demand.

rclcpp/include/rclcpp/publisher_options.hpp
ivanpauno: > second storage

Does adding a new private member variable for that sound ok? Seems weird, but I'm ok with it.
Besides that, the only other options I see is adding a `mutable` keyword to the `allocator` member as @hidmic tried before and a `static`.

> Either way, I think default constructing one when the user doesn't provide one is ok, it's just the current architecture leaves us in a weird place, as I said above about the constructor.

---

Defualt constructing one seems okay to me.
The only problem is that I guess that most code using allocators is currently doing:

```cpp
PublisherOptionsWithAllocator options;
options.allocator = <expression>;
```

and that would now default construct an allocator + destroy that one and set `options.allocator` to <expression>, which may be undesirable.

Long term, we might want to add:

```cpp
  PublisherOptions(const & Allocator alloc = Allocator());
```

and make the `allocator` member variable private (`get_allocator()` will be the getter).
It will break all code that uses allocators now, but that's not too much code and I think it's better to break code hard if we want to change the way allocators are initialized in `PublisherOptions`/`SubscriptionOptions`.
We could probably take advantage of the breakage there and also change `std::shared_ptr<Allocator>` to `Allocator`.

rclcpp/include/rclcpp/publisher_options.hpp
hidmic: >  I don't think there's a problem in making one on demand. 

Assuming you can i.e. that `Allocator` is default constructible, which need not be.

That's why I'd say we keep this patch simple, throwing or failing compilation for allocators that are not default constructible like Ivan did, and returning copies for the rest. That we can backport. For H-Turtle we should revamp all of this.

ivanpauno: > and statics are not guaranteed to be initialized lazely

Sorry, this is actually wrong, statics are guaranteed to be lazely initialized (see https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables).

It would still be different of what `std::vector` do, which is to use a new `Allocator()` for each `std::vector<Allocator>` that didn't explicitly provide one, instead of sharing the same static one.
But maybe that's better than throwing (?).

We could also use an extra private field for the extra storage (see first part of [this comment](https://github.com/ros2/rclcpp/pull/1647#discussion_r625739061), it would only introduce an ABI break (so we cannot backport to Foxy, but I think we can still do that in galactic).

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: > Does adding a new private member variable for that sound ok? Seems weird, but I'm ok with it.

I don't think it's an issue, to be honest. Like I said, this is a workaround, long term we should use a constructor argument and get away from shared ptr too.

> Defualt constructing one seems okay to me.
The only problem is that I guess that most code using allocators is currently doing:

Yeah, I'd avoid that, and go for the private default, and have `get_allocator()` return one or the other.

> Long term, we might want to add:

Agreed.

rclcpp/include/rclcpp/publisher_options.hpp
wjwwood: > We could also use an extra private field for the extra storage (see first part of this comment, it would only introduce an ABI break (so we cannot backport to Foxy, but I think we can still do that in galactic).

That'd be my vote... Another reason we should really do the pimpl everywhere so we have options in the future.

ivanpauno: Fixed in https://github.com/ros2/rclcpp/pull/1647/commits/fe13993311aa5f0ce43b60039503a019b84353f4.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14556/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9313/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12221/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14678/)

hidmic (commented):
LGTM ! I can't approve my own PR :sweat_smile:

wjwwood (approved):

wjwwood:
Thanks @ivanpauno and @hidmic!

rclcpp/test/rclcpp/test_intra_process_manager_with_allocators.cpp
hidmic: Hmm, the `tlsf_cpp` allocator in `realtime_support` is showing the same issue https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/778/console#console-section-24.

rclcpp/include/rclcpp/publisher_options.hpp
```diff
@@ -80,10 +85,7 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   to_rcl_publisher_options(const rclcpp::QoS & qos) const
   {
     rcl_publisher_options_t result = rcl_publisher_get_default_options();
-    using AllocatorTraits = std::allocator_traits<Allocator>;
-    using MessageAllocatorT = typename AllocatorTraits::template rebind_alloc<MessageT>;
-    auto message_alloc = std::make_shared<MessageAllocatorT>(*this->get_allocator().get());
-    result.allocator = rclcpp::allocator::get_rcl_allocator<MessageT>(*message_alloc);
+    result.allocator = rclcpp::allocator::get_rcl_allocator<void>(*this->get_allocator());
```
hidmic, line 88: Hmm, I think this was a mistake. From `cppreference`:

> The explicit specialization for void lacks the member typedefs reference, const_reference, size_type and difference_type. This specialization declares no member functions.

It appears `void` allocators are not supposed to be used for anything but type-erasure. If everything complies with that i.e. no code attempts to call `allocate` on a `void` allocator then you can afford not making the explicit specialization like the test allocator, the allocator tutorial and now the tlsf allocator do...

ivanpauno: can you fix that one?

rclcpp/test/rclcpp/test_intra_process_manager_with_allocators.cpp
clalancette: There is also a new warning in all builds, likely related to this: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1593/gcc/new/

rclcpp/include/rclcpp/publisher_options.hpp
hidmic: #1657 is up.

rclcpp/test/rclcpp/test_intra_process_manager_with_allocators.cpp
hidmic: Same root cause.

PR #1657, opened 2021-05-05
Do not attempt to use void allocators for memory allocation.

hidmic:
Void allocators are not supposed to even have `allocate` nor `deallocate` methods (see [cppreference](https://en.cppreference.com/w/cpp/memory/allocator)). `allocator::get_rcl_allocator` has an [specialization](https://github.com/ros2/rclcpp/blob/1fc2d58799a6d4530522be5c236c70be53455e60/rclcpp/include/rclcpp/allocator/allocator_common.hpp#L85-L94) to bypass itself if `Allocator` is `std::allocator<void>`, but it doesn't cover custom void allocators (which shouldn't have `allocate` nor `deallocate` methods to begin with).

This patch amends this, keeping a rebound allocator for byte-sized memory blocks around for publisher and subscription options.

Follow-up after 1fc2d58799a6d4530522be5c236c70be53455e60.

### DIFF
--- a/rclcpp/include/rclcpp/publisher_options.hpp
+++ b/rclcpp/include/rclcpp/publisher_options.hpp
@@ -85,7 +85,7 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   to_rcl_publisher_options(const rclcpp::QoS & qos) const
   {
     rcl_publisher_options_t result = rcl_publisher_get_default_options();
-    result.allocator = rclcpp::allocator::get_rcl_allocator<void>(*this->get_allocator());
+    result.allocator = this->get_rcl_allocator();
     result.qos = qos.get_rmw_qos_profile();
     result.rmw_publisher_options.require_unique_network_flow_endpoints =
       this->require_unique_network_flow_endpoints;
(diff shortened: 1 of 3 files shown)


hidmic:
I know this is even uglier. Other ideas are welcome. Sorry for not realizing before.

hidmic:
Running Linux CI with `clang` up to `rclcpp` and `tlsf_cpp`: [Build Status](https://ci.ros2.org/job/ci_linux/14562/)

hidmic:
Worth noting, `std::allocator<void>` is deprecated in C++17 and gone in C++20 (people complaining in 3, 2, 1...https://groups.google.com/a/isocpp.org/g/std-discussion/c/L0IolOWBKPA).

rclcpp/include/rclcpp/publisher_options.hpp
```diff
@@ -113,9 +113,26 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
...
+    return rclcpp::allocator::get_rcl_allocator<char>(*plain_allocator_storage_);
+  }
+
   // This is a temporal workaround, to make sure that get_allocator()
   // always returns a copy of the same allocator.
   mutable std::shared_ptr<Allocator> allocator_storage_;
+
+  // This is a temporal workaround, to keep the plain allocator that backs
+  // up the rcl allocator returned in rcl_publisher_options_t alive.
+  mutable std::shared_ptr<PlainAllocator> plain_allocator_storage_;
```
ivanpauno, line 135: can this be

```suggestion
  mutable std::optional<PlainAllocator> plain_allocator_storage_;
```

?

hidmic:
Regular CI up to `rclcpp` and `tlsf_cpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14563/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9321/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12228/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14689/)

rclcpp/include/rclcpp/publisher_options.hpp
hidmic: Yeah, `std::shared_ptr` makes no sense in this context. Perhaps `std::unique_ptr`, so we can backport to Foxy?

hidmic: Well, neither `std::shared_ptr` nor `std::optional` work in a situation like:

> it'd be easy to fetch rcl_publisher_options_t from some PublisherOptions, then copy those options, then drop the original

Arguably, that's bogus code (lifetimes are not properly managed). Alas, https://github.com/ros2/rclcpp/blob/1fc2d58799a6d4530522be5c236c70be53455e60/rclcpp/include/rclcpp/subscription.hpp#L115-L118. `std::shared_ptr` would still work in that case, but only because the execution order is favorable.

wjwwood (approved):

wjwwood:
I'll assume @ivanpauno would approve given the result of the discussion on his comment. Merging.

fujitatomoya:
@Mergifyio backport galactic foxy

PR #1658, opened 2021-05-05
Fix returning invalid namespace if sub_namespace is empty

### LINKED ISSUE #1656: Node::get_effective_namespace returns invalid namespace if sub-namespace is empty
Hi everyone

I tested this on foxy but as you can see from: https://github.com/ros2/rclcpp/blob/1fc2d58799a6d4530522be5c236c70be53455e60/rclcpp/src/rclcpp/node.cpp#L81-L95

it returns an namespace with a slash at the end in both cases.  `rclcpp::expand_topic_or_service_name` will then throw an exception it represents an invalid namespace format (e.g., '/' at the end are not allowed).

Is this still the case? Or rclcpp behaving wrong in this case. But since we usually specify namespaces without slashes almost everywhere I can recall using it I think even if it were permitted we should return the namespace without the slash.

BR, Markus

devrite:
Fixes #1656 . Instead of returning the node namespace with a slash (unless it was passed with one) we do not add one and rather return the unmodified namespace if the sub_namespace is empty.


wjwwood:
@devrite thanks for taking this on! Can you come up with a regression test for this fix? That would help show what it is addressing and also prevent it from breaking in the future again.

devrite:
@wjwwood I will try to add one.

BR, Markus

fujitatomoya (commented):
@devrite as i mentioned https://github.com/ros2/rclcpp/issues/1656#issuecomment-833323599, this will change the behavior of `node->create_sub_node("")`. (before is `InvalidNamespaceError` but with this patch it succeeds.) i am not sure if this is okay to go, i'd like to hear opinion from others.

devrite:
@fujitatomoya If I look at the interface documentation of `get_subnamespce()` it shall only return an empty string for the original note though it does not say how this behaves for a sub_node derived from a sub_node. In both cases it would not return an empty string anyway but in the latter would return the same. Following the documentation the latter case would not violate it, but I think it is not intended, since before it threw an exception when calling it and the examples do not show an empty sub_namespace node I think it is not intended.

The cause is the function `extend_effective_namespace` it those make the assumption the original sub_namespace has been validated, but does not explicitly validate the extending one, which caused same later error to throw. 

See the documentation part: 
https://github.com/ros2/rclcpp/blob/0659d829cec5a39c6ad861d597b4aa9aee224a25/rclcpp/include/rclcpp/node.hpp#L1172-L1269
as well as the extending function:

https://github.com/ros2/rclcpp/blob/0659d829cec5a39c6ad861d597b4aa9aee224a25/rclcpp/src/rclcpp/node.cpp#L52-L79

and the call to the sub_node constructor used by `create_sub_node()`:

https://github.com/ros2/rclcpp/blob/0659d829cec5a39c6ad861d597b4aa9aee224a25/rclcpp/src/rclcpp/node.cpp#L214-L227

devrite:
To sum up in my opinion it should fail, but on the other hand nothing blocks me to create an additional normal node in the same namespace. So I am not sure if there are any more side effects to sub_node as there would be to a normal node. But for this I am not deep enough inside the structure of a node: I also would like some comment on this.

devrite:
@wjwwood I added a regerssion test thanks to @fujitatomoya  pointing me to the right location. We still need to decide between the two behaviors:
* Old: Empty sub_namespace for sub_nodes not allowed (not documented though)
* New: Allowed, no error (any side effects?)

devrite:
I just quick inspected the code that fails but I do not see any reason why it should. As far as I can tell it does not involve any sub_namespaces. Any thoughts?

fujitatomoya:

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -162,6 +162,11 @@ TEST_F(TestNode, get_name_and_namespace) {
 }
 
 TEST_F(TestNode, subnode_get_name_and_namespace) {
+  {
+    auto node = std::make_shared<rclcpp::Node>("my_node", "ns");
+    EXPECT_STREQ("", node->get_sub_namespace().c_str());
+    EXPECT_STREQ("/ns", node->get_effective_namespace().c_str());
```
wjwwood, line 168: So does this test block fail without your patch?

I was expecting a new test case that fails without the patch but passes with the new patch. It seems to me that this should already be working since you're not even creating a new sub node?

wjwwood:
Before commenting on what I think the behavior should be, I want to understand the use case through new test cases. I'd also like to see the documentation updated to clarify this corner case.

rclcpp/test/rclcpp/test_node.cpp
devrite: It fails without the patch. As the only case when the sub_namespace was empty, so far, was when it was a normal node.

It does not throw or anything just the effective namespace for a normal node contains a trailing slash. That is why the second line will fail without the patch.

Currently the patch will allow sub_nodes with an empty sub_namespace,, ie. one that resides in the original namespace. This failed before, because of some checks that were done on the effective namespace, which previously contained the trailing slash and was considered invalid.

So if we want the old behaviour we would have to check for the empty sub_namespace explicitly now.

devrite:
Regarding the doc. The only thing, depending on what we decide, is to state it explicitly if an empty sub_namespace is allowed or not. Currently there is no note disallowing it or allowing it according to the doc. It just failed for sub_nodes with an empty sub_namespace so far. Although again I do not know whether there are any side effects (internal impl?) or not if we allow it for sub_nodes.

devrite:
Any update on how I should proceed? New behavior or old?

devrite:
@wjwwood I added another regression test and implemented the old behavior, this means:

* The bug of #1656 by creating the correct effective namespace (ie., if the sub namespace is empty the original namespace is returned)
* In order to maintain the old behaviour and empty sub-namespace for sub-nodes has been disallowed and will throw an exception

So we can just fix the issue itself for now without introducing new behavior and can postpone any discussion whether we want sub-nodes with an empty sub-namespace.

wjwwood (approved):
Ok, this makes sense to me now. Thanks for iterating on it. lgtm

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14953/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9693/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12629/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15116/)

devrite:
@wjwwood  Will check locally. With https://gist.githubusercontent.com/wjwwood/e679536d05c5ddb7fcbdf1608bca39da/raw/8b03ea9bd67d1692e23f29982061ba5a53e731e1/empty_sub_namespace.repos

But from a quick inspection on the linux build it fails in a service test to compile. Since I changed nothing in the headers this should have no influence.

devrite:
Got it. This error was just because my branch was based on an older master commit. After rebasing everything built fine and the test in rclcpp and test_rclcpp succeeded locally.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14955/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9695/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12632/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15118/)

devrite:
I test locall using the ros:rolling-ros-base-focal image (with git, compiler, etc. installed) and executing `rosdep update && rosdep install --from-paths ./ --ignore-src -q -y` after importing the repos.

@fujitatomoya Just checked. It fails to build image_transport_tutorials due to not being able to find cv_bridge. I could reproduce the same issue locally under the following condistions:

* The dependency is not installed
* Is installed but the env is not sourced (/opt/ros) if it is allowed to use installed deps

If it is intended to be a pure ws build (minimal external deps in /opt/ros and not to be installed via rosdep):
* Add vision_opencv to the repository list

```
  ros-perception/vision_opencv:
    type: git
    url: https://github.com/ros-perception/vision_opencv.git
    version: ros2
```

But the failing package image_transport_tutorial is lacking a library dependency and will still fail to build:
```
diff --git a/image_transport_tutorial/CMakeLists.txt b/image_transport_tutorial/CMakeLists.txt
index c88520a..f6ba3ca 100644
--- a/image_transport_tutorial/CMakeLists.txt
+++ b/image_transport_tutorial/CMakeLists.txt
@@ -49,6 +49,7 @@ ament_target_dependencies(my_subscriber
 # add the plugin example
 add_library(resized_plugins src/manifest.cpp src/resized_publisher.cpp src/resized_subscriber.cpp)
 ament_target_dependencies(resized_plugins
+  "cv_bridge"
   "image_transport"
   "OpenCV")
```

See https://github.com/ros-perception/image_common/pull/202

Rest of the build seems good.

devrite:
Changes in image_common have been reverted, so everything should build for now.

fujitatomoya:
@devrite thanks for checking on that 👍 i will run another CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14970/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9711/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12650/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15134/)

fujitatomoya:

PR #1668, opened 2021-05-12
[service] Don't use a weak_ptr to avoid leaking

ivanpauno:
Fixes https://github.com/ros2/rmw_connextdds/issues/40.

I don't see why we were using a weak_ptr, and that causes unexpected bugs when something external (e.g. the executor) is holding the rcl service handle but not the rcl node handle.

-----

Ownership diagram:

parent =====> child
weak_parent -----> weak_child

```
                                   (through a rclcpp::Node weak_ptr)
rclcpp::Executor --------------------------------------------
                                 |                          |
                                 |                          |
                                 V  (captured by deleter)   |
rclcpp::Service =======> rcl_service handle=======||        |
      ||                                          ||        |
      ||                                          VV        V
      ||=======================================> rcl_node handle
```

Based on the diagram above, I don't see any problem with keeping a shared_ptr instead of a weak_ptr in the deleter.
I will run some tests with valgrind just in case.

### DIFF
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -180,22 +180,14 @@ public:
     std::weak_ptr<rcl_node_t> weak_node_handle(node_handle_);
     // rcl does the static memory allocation here
     service_handle_ = std::shared_ptr<rcl_service_t>(
-      new rcl_service_t, [weak_node_handle, service_name](rcl_service_t * service)
+      new rcl_service_t, [handle = node_handle_, service_name](rcl_service_t * service)
       {
-        auto handle = weak_node_handle.lock();
-        if (handle) {
-          if (rcl_service_fini(service, handle.get()) != RCL_RET_OK) {
-            RCLCPP_ERROR(
-              rclcpp::get_node_logger(handle.get()).get_child("rclcpp"),
-              "Error in destruction of rcl service handle: %s",
-              rcl_get_error_string().str);
-            rcl_reset_error();
-          }
-        } else {
-          RCLCPP_ERROR_STREAM(
-            rclcpp::get_logger("rclcpp"),
-            "Error in destruction of rcl service handle " << service_name <<
-              ": the Node Handle was destructed too early. You will leak memory");
+        if (rcl_service_fini(service, handle.get()) != RCL_RET_OK) {
+          RCLCPP_ERROR(
+            rclcpp::get_node_logger(handle.get()).get_child("rclcpp"),
+            "Error in destruction of rcl service handle: %s",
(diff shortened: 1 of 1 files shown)


ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14614/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9369/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12280/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14748/)

ivanpauno:
> I will run some tests with valgrind just in case.

I checked (using `test_service` in `rclcpp`), there aren't new leaks after this change (though we have some preexisting ones :disappointed: )

rclcpp/include/rclcpp/service.hpp
```diff
@@ -180,22 +180,14 @@ class Service : public ServiceBase
     std::weak_ptr<rcl_node_t> weak_node_handle(node_handle_);
```
hidmic, line 180: @ivanpauno unused?

hidmic (approved):
Hmm, I see this solves the problem by extending the node's lifetime, but I'm not very enthusiastic about allowing a service to outlive its parent node. 

That's probably not an easy fix though :/

ivanpauno:
> Hmm, I see this solves the problem by extending the node's lifetime, but I'm not very enthusiastic about allowing a service to outlive its parent node.

That was already the case

https://github.com/ros2/rclcpp/blob/cb820ffed9ad0b1cd69583e4992ea087a5686402/rclcpp/include/rclcpp/service.hpp#L135

rclcpp/include/rclcpp/service.hpp
ivanpauno:
```suggestion
```

hidmic:
> That was already the case

Yeah, not asking for a fix here. Just saying, I'm not convinced it should be happening in a well-formed program to begin with.

ivanpauno:
> Yeah, not asking for a fix here. Just saying, I'm not convinced it should be happening in a well-formed program to begin with.

The program is definetely well-formed.
I also don't see good alternatives to the current approach.

ivanpauno:
@Mergifyio backport foxy galactic

wjwwood:
This is fine. I think it would be better to have a structure that has the shared pointer to the service and the node, but capturing it in the deleter is effectively the same thing, and it is backwards compatible. So on balance this is good.

I'm not sure why we were using a weak pointer before... the node---->callback_group---->service are all weak pointers, so it should be fine to back non-weak back references.

PR #1673, opened 2021-05-13
Declare parameters uninitialized

### LINKED ISSUE #1649: Allow declaring parameters with uninitialized values
#### Feature description

There are situations where we want to declare a parameter, but not initialize it until later.

For example, it would be nice to declare a parameter, with a type like this without it throwing an exception because the user did not provide an override at runtime:

```c++
node->declare_parameter(name, type);
```

We could then later set the value of the parameter programmatically from the node itself (e.g. with `set_parameter`) or let the user set it later with a service call (e.g. `ros2 param set`).

After declaring a parameter without an initial value and if there is no user-provided override, then I would expect the parameter to have a value of "UNSET", and trying to access a value might throw an exception.

jacobperron:
Fixes #1649

Allow declaring parameters without an initial value or override.
This was possible prior to Galactic, but was made impossible since we started enforcing the types of parameters in Galactic.

---

Opening for early feedback. I still need to test this more thoroughly for my use-case and add documentation.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -351,8 +351,6 @@ __declare_parameter_common(
 
   // If there is no initial value, then skip initialization
   if (initial_value->get_type() == rclcpp::PARAMETER_NOT_SET) {
-    // descriptor type should be set
-    assert(parameter_descriptor.type != rclcpp::PARAMETER_NOT_SET);
     // Add declared parameters to storage (without a value)
     parameter_infos[name].descriptor.name = name;
     parameter_infos[name].descriptor.type = parameter_descriptor.type;
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -809,14 +809,20 @@ NodeParameters::get_parameters(const std::vector<std::string> & names) const
 rclcpp::Parameter
 NodeParameters::get_parameter(const std::string & name) const
 {
-  rclcpp::Parameter parameter;
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
 
-  if (get_parameter(name, parameter)) {
-    return parameter;
+  auto param_iter = parameters_.find(name);
+  if (
+    parameters_.end() != param_iter &&
+    param_iter->second.value.get_type() != rclcpp::ParameterType::PARAMETER_NOT_SET)
+  {
+    return rclcpp::Parameter{name, param_iter->second.value};
   } else if (this->allow_undeclared_) {
-    return parameter;
-  } else {
+    return rclcpp::Parameter{};
+  } else if (parameters_.end() == param_iter) {
     throw rclcpp::exceptions::ParameterNotDeclaredException(name);
+  } else {
+    throw rclcpp::exceptions::NoParameterOverrideProvided(name);
   }
 }
 
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -827,7 +827,7 @@ NodeParameters::get_parameter(const std::string & name) const
   } else if (parameters_.end() == param_iter) {
     throw rclcpp::exceptions::ParameterNotDeclaredException(name);
   } else {
-    throw rclcpp::exceptions::NoParameterOverrideProvided(name);
+    throw rclcpp::exceptions::ParameterUninitializedException(name);
   }
 }
 
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -353,7 +353,11 @@ __declare_parameter_common(
   if (initial_value->get_type() == rclcpp::PARAMETER_NOT_SET) {
     // Add declared parameters to storage (without a value)
     parameter_infos[name].descriptor.name = name;
-    parameter_infos[name].descriptor.type = parameter_descriptor.type;
+    if (parameter_descriptor.dynamic_typing) {
+      parameter_infos[name].descriptor.type = rclcpp::PARAMETER_NOT_SET;
+    } else {
+      parameter_infos[name].descriptor.type = parameter_descriptor.type;
+    }
     parameters_out[name] = parameter_infos.at(name);
     rcl_interfaces::msg::SetParametersResult result;
     result.successful = true;
@@ -814,7 +818,8 @@ NodeParameters::get_parameter(const std::string & name) const
   auto param_iter = parameters_.find(name);
   if (
     parameters_.end() != param_iter &&
-    param_iter->second.value.get_type() != rclcpp::ParameterType::PARAMETER_NOT_SET)
+    (param_iter->second.value.get_type() != rclcpp::ParameterType::PARAMETER_NOT_SET ||
+      param_iter->second.descriptor.dynamic_typing))
   {
     return rclcpp::Parameter{name, param_iter->second.value};
   } else if (this->allow_undeclared_) {
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -349,6 +349,19 @@ __declare_parameter_common(
     initial_value = &overrides_it->second;
   }
 
+  // If there is no initial value, then skip initialization
+  if (initial_value->get_type() == rclcpp::PARAMETER_NOT_SET) {
+    // descriptor type should be set
+    assert(parameter_descriptor.type != rclcpp::PARAMETER_NOT_SET);
+    // Add declared parameters to storage (without a value)
+    parameter_infos[name].descriptor.name = name;
+    parameter_infos[name].descriptor.type = parameter_descriptor.type;
+    parameters_out[name] = parameter_infos.at(name);
+    rcl_interfaces::msg::SetParametersResult result;
+    result.successful = true;
+    return result;
+  }
+
   // Check with the user's callback to see if the initial value can be set.
   std::vector<rclcpp::Parameter> parameter_wrappers {rclcpp::Parameter(name, *initial_value)};
   // This function also takes care of default vs initial value.
@@ -411,14 +424,6 @@ declare_parameter_helper(
       };
     }
     parameter_descriptor.type = static_cast<uint8_t>(type);
-  }
-
-  if (
-    rclcpp::PARAMETER_NOT_SET == default_value.get_type() &&
-    overrides.find(name) == overrides.end() &&
-    parameter_descriptor.dynamic_typing == false)
-  {
-    throw rclcpp::exceptions::NoParameterOverrideProvided(name);
   }
 
   rcl_interfaces::msg::ParameterEvent parameter_event;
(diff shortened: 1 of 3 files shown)


rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -349,6 +349,19 @@ __declare_parameter_common(
     initial_value = &overrides_it->second;
   }
 
+  // If there is no initial value, then skip initialization
+  if (initial_value->get_type() == rclcpp::PARAMETER_NOT_SET) {
+    // descriptor type should be set
+    assert(parameter_descriptor.type != rclcpp::PARAMETER_NOT_SET);
```
ivanpauno, line 355: is this the case if `dynamic_typing == false`?

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -334,11 +334,17 @@ TEST_F(TestNode, declare_parameter_with_no_initial_values) {
   {
     rcl_interfaces::msg::ParameterDescriptor descriptor;
     descriptor.dynamic_typing = true;
-    // no default, no initial
+    // no default, no initial, dynamic typing
     rclcpp::ParameterValue value = node->declare_parameter(
       "parameter"_unq, rclcpp::ParameterValue{}, descriptor);
```
ivanpauno, line 339: e.g. I think that the previous `assert` will comply here in debug mode

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -334,11 +334,17 @@ TEST_F(TestNode, declare_parameter_with_no_initial_values) {
...
     EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_NOT_SET);
   }
+  {
+    // no default, no initial, static typing
+    rclcpp::ParameterValue value = node->declare_parameter(
+      "parameter"_unq, rclcpp::ParameterType::PARAMETER_INTEGER);
+    EXPECT_EQ(value.get_type(), rclcpp::PARAMETER_NOT_SET);
+  }
```
ivanpauno, line 347: it would be good to test what happens when you try to get the parameter.

IMO, if static typing is specified and we allow them to be uninitialized, the first get should throw an error if it's still uninitialized.
Does that make sense to you?

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -2798,9 +2804,8 @@ TEST_F(TestNode, static_and_dynamic_typing) {
     EXPECT_EQ("hello!", param);
   }
   {
-    EXPECT_THROW(
-      node->declare_parameter("integer_override_not_given", rclcpp::PARAMETER_INTEGER),
-      rclcpp::exceptions::NoParameterOverrideProvided);
+    auto param = node->declare_parameter("integer_override_not_given", rclcpp::PARAMETER_INTEGER);
+    EXPECT_EQ(rclcpp::PARAMETER_NOT_SET, param.get_type());
```
ivanpauno, line 2808: is this any different to the test case added in L342?

ivanpauno (commented):
Thanks for working on this!

I completely forgot about this.
I only have some minimal comments

rclcpp/test/rclcpp/test_node.cpp
jacobperron: No, I can remove the case I added and leave this one

ivanpauno: I don't mind much, but I think we should remove one of them to avoid duplication.

wjwwood (commented):
Nothing to add beyond what @ivanpauno said.

fujitatomoya (commented):
i am also good to go with @ivanpauno 's comments.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
jacobperron: Good catch! It is not always true, I missed this other instance where [`__declare_parameters_common()`](https://github.com/ros2/rclcpp/blob/8764151063988b4bec0680d857e648c17c2f9e47/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L656) is called. Compiling in debug mode causes a test failure :+1: 

The assertion is not necessary; removed in 24d2641

rclcpp/test/rclcpp/test_node.cpp
jacobperron: We get an exception if you try to access a parameter that is uninitialized (static or dynamic). I've added tests and changed the code to throw a more accurate exceptions (01e5ae2).

I'm still not sure if we want to throw an exception if `dynamic_typing=true` or return an unset parameter. I think it's easy to implement the latter, but I went with throwing an exception for consistency.

rclcpp/test/rclcpp/test_node.cpp
jacobperron: Removed duplicate test in 01e5ae2

jacobperron:
I've addressed the feedback. PTAL :slightly_smiling_face:

rclcpp/test/rclcpp/test_node.cpp
ivanpauno: > I'm still not sure if we want to throw an exception if dynamic_typing=true or return an unset parameter. I think it's easy to implement the latter, but I went with throwing an exception for consistency.

mmm, in that case we should return the unset parameter IMO.
That was the behavior in previous releases for dynamic typing, and I would rather keep it so people can exactly get the previous behavior by using `dynamic_typing=false`.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -349,6 +349,17 @@ __declare_parameter_common(
     initial_value = &overrides_it->second;
   }
 
+  // If there is no initial value, then skip initialization
+  if (initial_value->get_type() == rclcpp::PARAMETER_NOT_SET) {
+    // Add declared parameters to storage (without a value)
+    parameter_infos[name].descriptor.name = name;
+    parameter_infos[name].descriptor.type = parameter_descriptor.type;
```
ivanpauno, line 356:
```suggestion
    parameter_infos[name].descriptor.type = parameter_descriptor.dynamic_typing ? rclcpp::PARAMETER_NOT_SET: parameter_descriptor.type;
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -806,14 +809,20 @@ NodeParameters::get_parameters(const std::vector<std::string> & names) const
...
 {
-  rclcpp::Parameter parameter;
+  std::lock_guard<std::recursive_mutex> lock(mutex_);
 
-  if (get_parameter(name, parameter)) {
-    return parameter;
+  auto param_iter = parameters_.find(name);
+  if (
+    parameters_.end() != param_iter &&
+    param_iter->second.value.get_type() != rclcpp::ParameterType::PARAMETER_NOT_SET)
```
ivanpauno, line 817:
```suggestion
    (param_iter->second.value.get_type() != rclcpp::ParameterType::PARAMETER_NOT_SET ||
      param_iter->second.descriptor.dynamic_typing == true))
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -806,14 +809,20 @@ NodeParameters::get_parameters(const std::vector<std::string> & names) const
...
+  {
+    return rclcpp::Parameter{name, param_iter->second.value};
   } else if (this->allow_undeclared_) {
-    return parameter;
-  } else {
+    return rclcpp::Parameter{};
+  } else if (parameters_.end() == param_iter) {
     throw rclcpp::exceptions::ParameterNotDeclaredException(name);
+  } else {
+    throw rclcpp::exceptions::NoParameterOverrideProvided(name);
```
ivanpauno, line 825: This should maybe be `ParameterUninitializedException`.
It's not exactly `NoParameterOverrideProvided`, because this is cause by an override not being provided and the user not setting the parameter before getting it.

If we change this, `NoParameterOverrideProvided` wouldn't be thrown anymore and should be deprecated.

rclcpp/test/rclcpp/test_node.cpp
ivanpauno: Another thing to test, you can declare an uninitialized statically typed parameter, then set it, then get it.
That should be valid without any exception thrown

ivanpauno (commented):
A few extra comments

rclcpp/test/rclcpp/test_node.cpp
jacobperron: I suppose, though we technically the dynamic vs. static typing feature is new in Galactic, so I don't think there's many people using it yet. Though, I don't mind making the behavior such that it returns NOT_SET

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
jacobperron: If we want to try and get this in Galactic, do you think it is okay to deprecate it? I can add the new exception in any case.

jacobperron: hmm.. `NoParameterOverrideProvided` was only added for the dynamic v. static type feature. If we're willing to change API just before the Galactic release then we can avoid a deprecation cycle :sweat_smile:

ivanpauno: > hmm.. NoParameterOverrideProvided was only added for the dynamic v. static type feature. If we're willing to change API just before the Galactic release then we can avoid a deprecation cycle sweat_smile

:+1:

rclcpp/test/rclcpp/test_node.cpp
jacobperron: Test added in 344149c
Not longer throwing exception in the dynamic typing case: 1ea6ea4

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
jacobperron: I applied this with a written out if-else block since the ternary operator caused a compiler warning (1ea6ea4)

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
jacobperron: Added `ParameterUninitializedException` (0d97e32).
Removed `NoParameterOverrideProvided` (b32f19d). I'll check for any downstream usage.

ivanpauno (approved):
LGTM!

If we're going to make these changes in `galactic`, it would be good to first implement the same changes in `rclpy` so we're sure we can complete the implementation in both.

There are some minimal API changes here, so I would like to see full CI as well (build --packakes-above-and-dependencies, test --packages-above rclcpp).

wjwwood (approved):

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14655/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9408/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12319/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14799/)

jacobperron:
I've discovered that navigation2 references `NoParameterOverrideProvided` (introduced in https://github.com/ros-planning/navigation2/pull/2239/). I've opened a PR to fix it; I don't think it should be too disruptive: https://github.com/ros-planning/navigation2/pull/2347

@SteveMacenski Heads up. Our plan is to first release this into Rolling (today), address any fallout, and then backport to Galactic before the week is out.

SteveMacenski:
Thanks for the heads up

jacobperron:
rclpy equivalent: https://github.com/ros2/rclpy/pull/798

PR #1674, opened 2021-05-13
fix syntax issue with gcc

wjwwood:
required for https://github.com/ros2/demos/pull/482

I tried a few solutions from the referenced SO post about this, but for various reasons, none of them work for us. So we'll just have to document that this macro must be used outside of any namespace to work properly.


audrow (approved):
Looks good. The comment is clear.

audrow:
CI can be found here: https://github.com/ros2/demos/pull/482#issuecomment-841449048

PR #1677, opened 2021-05-17
Fix occasionally missing goal result caused by race condition

KavenYau:
Fix occasionally missing goal result when a action finished quickly (one thread enter to `execute_result_request_received` and the other enter to `publish_result`).

### DIFF
--- a/rclcpp_action/src/server.cpp
+++ b/rclcpp_action/src/server.cpp
@@ -500,9 +500,13 @@ ServerBase::execute_result_request_received(std::shared_ptr<void> & data)
     result_response = create_result_response(action_msgs::msg::GoalStatus::STATUS_UNKNOWN);
   } else {
     // Goal exists, check if a result is already available
+    std::lock_guard<std::recursive_mutex> lock(pimpl_->unordered_map_mutex_);
     auto iter = pimpl_->goal_results_.find(uuid);
     if (iter != pimpl_->goal_results_.end()) {
       result_response = iter->second;
+    } else {
+      // Store the request so it can be responded to later
+      pimpl_->result_requests_[uuid].push_back(request_header);
     }
   }
 
@@ -514,10 +518,6 @@ ServerBase::execute_result_request_received(std::shared_ptr<void> & data)
     if (RCL_RET_OK != rcl_ret) {
       rclcpp::exceptions::throw_from_rcl_error(rcl_ret);
     }
-  } else {
-    // Store the request so it can be responded to later
-    std::lock_guard<std::recursive_mutex> lock(pimpl_->unordered_map_mutex_);
-    pimpl_->result_requests_[uuid].push_back(request_header);
   }
   data.reset();
 }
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
this makes sense to me.

rclcpp_action/src/server.cpp
```diff
@@ -627,19 +627,19 @@ ServerBase::publish_result(const GoalUUID & uuid, std::shared_ptr<void> result_m
   }
 
   {
-    std::lock_guard<std::recursive_mutex> lock(pimpl_->unordered_map_mutex_);
+    std::lock_guard<std::recursive_mutex> unordered_map_lock(pimpl_->unordered_map_mutex_);
```
ivanpauno, line 630: This seems fine to me, but it can introduce a ptential deadlock in the future.
I would rather:

- Document in which order we're currently taking the `unordered_map_mutex_`, `action_server_reentrant_mutex_` mutexes, to avoid someone doing it in the opposite order later.
- Take `action_server_reentrant_mutex_` out of the if/loop together with  `unordered_map_mutex_` using `std::scoped_lock`.

rclcpp_action/src/server.cpp
```diff
@@ -627,19 +627,19 @@ ServerBase::publish_result(const GoalUUID & uuid, std::shared_ptr<void> result_m
...
-      std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
-      rcl_ret_t ret = rcl_action_send_result_response(
-        pimpl_->action_server_.get(), &request_header, result_msg.get());
-      if (RCL_RET_OK != ret) {
-        rclcpp::exceptions::throw_from_rcl_error(ret);
+    // if there are clients who already asked for the result, send it to them
+    auto iter = pimpl_->result_requests_.find(uuid);
+    if (iter != pimpl_->result_requests_.end()) {
+      for (auto & request_header : iter->second) {
+        std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
```
ivanpauno, line 637: the code is old, but it makes more sense to take the mutex out of the for loop.

rclcpp_action/src/server.cpp
```diff
@@ -500,9 +500,13 @@ ServerBase::execute_result_request_received(std::shared_ptr<void> & data)
     result_response = create_result_response(action_msgs::msg::GoalStatus::STATUS_UNKNOWN);
   } else {
     // Goal exists, check if a result is already available
+    std::lock_guard<std::recursive_mutex> lock(pimpl_->unordered_map_mutex_);
```
ivanpauno, line 503: nit:

```suggestion
    std::lock_guard lock(pimpl_->unordered_map_mutex_);
```

rclcpp_action/src/server.cpp
ivanpauno, line 630: nit: 

```suggestion
    std::lock_guard unordered_map_lock(pimpl_->unordered_map_mutex_);
```

KavenYau: With using `std::scoped_lock.`, then if could not be backported to `foxy`? Because `foxy` is compiled with C++14.

KavenYau: This two changes are inconsistent with other `std::lock_guard` in `server.cpp`.

rclcpp_action/src/server.cpp
KavenYau: Do you mean like this?
```c++
    auto iter = pimpl_->result_requests_.find(uuid);
    if (iter != pimpl_->result_requests_.end()) {
      std::lock_guard<std::recursive_mutex> lock(pimpl_->action_server_reentrant_mutex_);
      for (auto & request_header : iter->second) {
        rcl_ret_t ret = rcl_action_send_result_response(
          pimpl_->action_server_.get(), &request_header, result_msg.get());
        if (RCL_RET_OK != ret) {
        }
      }
    }
```

rclcpp_action/src/server.cpp
ivanpauno: Sounds good, but please add documentation explaining what's the current order we're taking the mutexes.

rclcpp_action/src/server.cpp
ivanpauno: yes, that would be find.

rclcpp_action/src/server.cpp
ivanpauno: Before, we were using c++14 and this was not possible, so we're going to find inconsistencies.
The idea is to use the less verbose version in new code.

In this case, a backport PR will need to be modified to be compliant with c++14 again.
I think that's fine, but it's also fine if you want to keep the verbose version here.

rclcpp_action/src/server.cpp
KavenYau: OK, that was applied.

rclcpp_action/src/server.cpp
KavenYau: I added a NOTE comment to this block scope.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14668/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9420/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12331/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14811/)

ivanpauno (approved):

ivanpauno:
Failures are unrelated

fujitatomoya:
@ivanpauno we would want this to backport to foxy and galactic? what you think?

SteveMacenski:
I would like that :smile:

fujitatomoya:
backpork PRs are open for foxy and galactic.

- https://github.com/ros2/rclcpp/pull/1682
- https://github.com/ros2/rclcpp/pull/1683

PR #1680, opened 2021-05-19
Enable all QoS override policies

audrow:
This makes it so that all QoS options can be overridden.


audrow:
Closed after a discussion where we decided that it probably isn't best to expose all QoS overrides for our use-case: https://github.com/ros-perception/image_pipeline/pull/651#discussion_r635608091. Perhaps this could be merged in if there is a reason for it to be included.

PR #1686, opened 2021-06-03
adding node name to the executor runtime_error

SteveMacenski:
I'm trying to debug an issue in rviz2 where it crashes with this warning and it would be really nice to know the name of the specific node that's giving me trouble. I half expect to just see "rviz" which won't be very helpful, but this isn't the first time I've run into the hunt for "which node am I?".

In general, when new software is ported to ROS2 (and folks are learning about spinning mechanisms in ROS2), this error comes up **alot** and it would be really helpful in debugging to know where to look when launching a large system.

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -268,7 +268,8 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
   // If the node already has an executor
   std::atomic_bool & has_executor = node_ptr->get_associated_with_executor_atomic();
   if (has_executor.exchange(true)) {
-    throw std::runtime_error("Node has already been added to an executor.");
+    throw std::runtime_error(
+      std::string("Node (%s) has already been added to an executor.", node_ptr->get_name()));
   }
   std::lock_guard<std::mutex> guard{mutex_};
   for (auto & weak_group : node_ptr->get_callback_groups()) {
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -269,9 +269,8 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
   std::atomic_bool & has_executor = node_ptr->get_associated_with_executor_atomic();
   if (has_executor.exchange(true)) {
     throw std::runtime_error(
-            std::string(
-              "Node (%s) has already been added to an executor.",
-              node_ptr->get_fully_qualified_name()));
+            std::string("Node '") + node_ptr->get_fully_qualified_name() +
+            "' has already been added to an executor.");
   }
   std::lock_guard<std::mutex> guard{mutex_};
   for (auto & weak_group : node_ptr->get_callback_groups()) {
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Build failure looks like from the keys issue

fujitatomoya:
agree. i think user application is likely to have this warning in the begging, and as you mentioned it should tell explicitly which node is the problem.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -268,7 +268,8 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
   // If the node already has an executor
   std::atomic_bool & has_executor = node_ptr->get_associated_with_executor_atomic();
   if (has_executor.exchange(true)) {
-    throw std::runtime_error("Node has already been added to an executor.");
+    throw std::runtime_error(
+      std::string("Node (%s) has already been added to an executor.", node_ptr->get_name()));
```
fujitatomoya, line 272: how about telling fully qualified name including namespace?

```suggestion
      std::string("Node (%s) has already been added to an executor.", node_ptr->get_fully_qualified_name()));
```

SteveMacenski:
Thanks, that would be better, its been updated!

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -268,7 +268,10 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
   // If the node already has an executor
   std::atomic_bool & has_executor = node_ptr->get_associated_with_executor_atomic();
   if (has_executor.exchange(true)) {
-    throw std::runtime_error("Node has already been added to an executor.");
+    throw std::runtime_error(
+            std::string(
+              "Node (%s) has already been added to an executor.",
+              node_ptr->get_fully_qualified_name()));
```
clalancette, line 274: This is mixing printf-style string in with normal strings, and is causing the tests to fail.  I think this wants to be something like:

```suggestion
            std::string("Node '") + node_ptr->get_fully_qualified_name() + "' has already been added to an executor.");
```

(that particular line will probably fail linting, so you'll also have to fix that up)

SteveMacenski:
The failure is

```
10:51:49 83: /tmp/ws/src/rclcpp/rclcpp/test/rclcpp/executors/test_executors.cpp:170: Failure
10:51:49 83: Expected: executor2.add_node(this->node) throws an exception of type std::runtime_error.
10:51:49 83:   Actual: it throws a different type.
```

From here: https://github.com/ros2/rclcpp/blob/master/rclcpp/test/rclcpp/executors/test_executors.cpp#L170. Why would changing the input string format change the type of exception being interpreted?

clalancette:
> ```
> 
> ```
> 
> 
>     
>   
> 
> From here: https://github.com/ros2/rclcpp/blob/master/rclcpp/test/rclcpp/executors/test_executors.cpp#L170. Why would changing the input string format change the type of exception being interpreted? I made the requested change, but that's pretty bizarre.

It's because the previous code was invalid.  That previous code was causing `std::string` to throw a `std::length_error` exception when instantiating the error string.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14730/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9481/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12395/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14875/)

fujitatomoya (approved):

fujitatomoya:
http://ci.ros2.org/job/ci_linux/14730/ is failed with know issue which is https://github.com/ros2/ros2cli/issues/630. it is unrelated to this fix. I will go ahead to merge this.

SteveMacenski:
Thanks!

PR #1690, opened 2021-06-09
Fix bug: Add node while spinning

mauropasse:
This is a fix for a `StaticSingleThreadedExecutor` bug, which fails when adding a new node while the executor is spinning.
This happens due the `spin()` thread tries to add the new node's guard condition to the waitset, but the waitset has not been resized yet (so fails).

0. `StaticSingleThreadedExecutor` created and spinning in its own thread.
1. Node is added to executor, guard condition stored:
https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp#L330
2. Spin thread tries to add the new guard condition to a _not-yet-resized_ wait set
https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp#L270
3. Resize of the wait set happens when some guard condition of the executor has been triggered (this can happen after 2)
https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp#L110

The workaround in this PR is adding the new nodes' guard condition AFTER the wait set has been resized, so it can accept the guard condition.

Wasn't there a test for adding nodes while executor spinning? Couldn't find it.

### DIFF
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -61,8 +61,7 @@ StaticExecutorEntitiesCollector::~StaticExecutorEntitiesCollector()
 void
 StaticExecutorEntitiesCollector::init(
   rcl_wait_set_t * p_wait_set,
-  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy,
-  rcl_guard_condition_t * executor_guard_condition)
+  rclcpp::memory_strategy::MemoryStrategy::SharedPtr memory_strategy)
 {
   // Empty initialize executable list
   exec_list_ = rclcpp::experimental::ExecutableList();
@@ -73,9 +72,6 @@ StaticExecutorEntitiesCollector::init(
     throw std::runtime_error("Received NULL memory strategy in executor waitable.");
   }
   memory_strategy_ = memory_strategy;
-
-  // Add executor's guard condition
-  memory_strategy_->add_guard_condition(executor_guard_condition);
 
   // Get memory strategy and executable list. Prepare wait_set_
   std::shared_ptr<void> shared_ptr;
@@ -108,6 +104,14 @@ StaticExecutorEntitiesCollector::execute(std::shared_ptr<void> & data)
   fill_executable_list();
   // Resize the wait_set_ based on memory_strategy handles (rcl_wait_set_resize)
   prepare_wait_set();
+  // Add new nodes guard conditions to map
+  std::lock_guard<std::mutex> guard{new_nodes_mutex_};
+  for (const auto & weak_node : new_nodes_) {
+    if (auto node_ptr = weak_node.lock()) {
+      weak_nodes_to_guard_conditions_[node_ptr] = node_ptr->get_notify_guard_condition();
+    }
+  }
+  new_nodes_.clear();
 }
 
 void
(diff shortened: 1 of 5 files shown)


fujitatomoya:
@mauropasse 

> which fails when adding a new node while the executor is spinning.

are we allowed to do that with `StaticSingleThreadedExecutor`?

https://github.com/ros2/rclcpp/blob/7d8b2690ff188672bfe3703d8b3c074b87303931/rclcpp/include/rclcpp/executors/static_single_threaded_executor.hpp#L44-L47

mauropasse:
The idea with the `StaticSingleThreadedExecutor` was an improvement in performance by not rebuilding the executable list at every iteration, but it still should handle dynamic situations as nodes/pubs/subs being added at runtime. That's the whole point of having an `execute` API in the entities collector, which rebuilds everything when something changes. I haven't seen this situation before, on Foxy wasn't happening, don't know why

KavenYau:
What's the manifestation of this issue? Throwing exception?

mauropasse:
> @KavenYau What's the manifestation of this issue? Throwing exception? 


```
  what():  Executor waitable: couldn't add guard condition to wait set
Aborted (core dumped)
```

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
```diff
@@ -327,7 +334,7 @@ StaticExecutorEntitiesCollector::add_callback_group(
     throw std::runtime_error("Callback group was already added to executor.");
   }
   if (is_new_node) {
-    weak_nodes_to_guard_conditions_[node_ptr] = node_ptr->get_notify_guard_condition();
+    new_nodes_.push_back(node_ptr);
```
wjwwood, line 337: `new_nodes_` is updated here, and read/cleared above, is there a race in this case or are these two sections synchronized already?

mauropasse: Sections were not synchronized. Added a mutex in b8db146

Also the executor's guard condition was being added twice to the memory strategy, changed that in 41febe3

fujitatomoya:
@mauropasse 

> but it still should handle dynamic situations as nodes/pubs/subs being added at runtime.

okay, i see the point. and it should not fail or core dump.

there is conflict against the baseline, could you rebase it? and then i can start the CI.

btw, test would be ideal.

wjwwood (approved):
The changes lgtm now, needs a rebase though.

fujitatomoya (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15065/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9797/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12750/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15232/)

mauropasse:
```
==> Contents of `00-ros2.repos`:
ros2/rclcpp:
     type: git
     url: https://github.com/mauropasse/rclcpp.git
     version: fix-static-executor-add-node-bug

20:32:53 === src/ros2/rclcpp (git) ===
20:32:53 Could not checkout ref 'fix-static-executor-add-node-bug': fatal: invalid reference: fix-static-executor-add-node-bug
```
My branch is called `mauro/fix-static-executor-add-node-bug`, seems that CI is missing the `mauro/` ?

fujitatomoya:
My bad. thanks for pointing that out.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15070/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9802/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12755/)
* Windows [Build Status]

mauropasse:
No problem. CI passing now!

PR #1692, opened 2021-06-11
Use a different mechanism to avoid timers being scheduled multiple times by the MultiThreadedExecutor

ivanpauno:
I haven't seen issues related to this since https://github.com/ros2/rclcpp/pull/1516, but I think that's still a fragile approach.
This takes a simular approach to https://github.com/ros2/rclcpp/pull/1241 for timers:

- rcl_timer_call() is called when getting the timer executable.
  This is done in a mutually exclusive fashion, thus preventing a different thread from taking the same timer again.
- execute() will only execute the timer callback, as `rcl_timer_call()` was called before.

I think this approach is simpler and I expect it to work better, though I want to do some more testing to confirm that before merging.


ivanpauno:
Testing packages: rclcpp, test_rclcpp
`--retest-until-fail 10` to see if timer tests are more fragile now or not

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14739/)

ivanpauno:
@wjwwood friendly ping

wjwwood (approved):
This makes sense to me, but it's hard for me to say whether or not it is 100% robust.

It is simpler, which is good, and the retest seems clean, so I'm ok with it if you are @ivanpauno.

fujitatomoya (approved):
looks good to me 👍

ivanpauno:
Rebased, CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14919/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9661/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12591/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15078/)

mauropasse:
Hi @ivanpauno . These changes have broken the `StaticSingleThreadedExecutor`.
Before, the `next_call_time` for a timer, was computed on `rcl_timer_call` when executing the timer callback.
https://github.com/ros2/rclcpp/blob/0d6d9e67782bcd1de06af875459e4abb2b91ab86/rclcpp/include/rclcpp/timer.hpp#L198-L201
Now, the new api `call()` which is now in charge to call `rcl_timer_call()` is used by the `get_next_timer()` API, which is not used on the `StaticSingleThreadedExecutor`. 
So now the timers are always ready, the next time call is never reset, and the callbacks are executed using 100% CPU.
A greedy fix for this is calling `call()` here:
https://github.com/ros2/rclcpp/blob/d5f3d35fbe3127e30d4ee9d40f57c42584eeb830/rclcpp/include/rclcpp/timer.hpp#L225-L231
But you probably know better if this is a good fix. If you confirm it I can open a PR with the fix.
I'm surprised the tests didn't catch this problem :(

ivanpauno:
Hi @mauropasse, sorry I forgot about the `StaticSingleThreadedExecutor`.
I would rather just add a call to timer.call() before this line:

https://github.com/ros2/rclcpp/blob/d5f3d35fbe3127e30d4ee9d40f57c42584eeb830/rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp#L240

Feel free to add me as a reviewer if you open a PR.

PR #1693, opened 2021-06-13
Add arguments to configure number of threads

wep21:
This PR enables to configure number of threads in `component_container_mt` as below.
```
    container = ComposableNodeContainer(
            name='my_container',
            namespace='',
            package='rclcpp_components',
            executable='component_container_mt',
            composable_node_descriptions=[component],
            output='screen',
            arguments=['-t', '5'],
    )
```

### DIFF
--- a/rclcpp_components/src/component_container_mt.cpp
+++ b/rclcpp_components/src/component_container_mt.cpp
@@ -21,8 +21,23 @@
 int main(int argc, char * argv[])
 {
   /// Component container with a multi-threaded executor.
-  rclcpp::init(argc, argv);
-  auto exec = std::make_shared<rclcpp::executors::MultiThreadedExecutor>();
+  auto vargv = rclcpp::init_and_remove_ros_arguments(argc, argv);
+  size_t number_of_threads{0};
+  rclcpp::Logger logger{rclcpp::get_logger("component_container_mt")};
+
+  if (vargv.size() == 3 && (vargv.at(1) == "--thread-num" || vargv.at(1) == "-t")) {
+    try {
+      number_of_threads = static_cast<size_t>(std::stoi(vargv.at(2)));
+      RCLCPP_INFO_STREAM(logger, "number of threads: " << number_of_threads);
+    } catch (const std::invalid_argument & ex) {
+      RCLCPP_ERROR_STREAM(logger, ex.what());
+    } catch (const std::out_of_range & ex) {
+      RCLCPP_ERROR_STREAM(logger, ex.what());
+    }
+  }
+
+  auto exec = std::make_shared<rclcpp::executors::MultiThreadedExecutor>(
+    rclcpp::ExecutorOptions(), number_of_threads);
   auto node = std::make_shared<rclcpp_components::ComponentManager>(exec);
   exec->add_node(node);
   exec->spin();
--- a/rclcpp_components/src/component_container_mt.cpp
+++ b/rclcpp_components/src/component_container_mt.cpp
@@ -29,12 +29,12 @@ int main(int argc, char * argv[])
     for (auto itr = vargv.begin() + 1; itr != vargv.end(); ++itr) {
       if (*itr == "--help" || *itr == "-h") {
         RCLCPP_INFO_STREAM(logger,
-          "Usage: component_container_mt --thread-num <number of thread>" << std::endl);
+          "Usage: component_container_mt --thread-num <number of thread>");
         return 0;
       } else if (*itr == "--thread-num" || *itr == "-t") {
         if (itr == vargv.end() - 1) {
           RCLCPP_ERROR_STREAM(logger, "Empty thread number" << std::endl
-            << "Usage: component_container_mt --thread-num <number of thread>" << std::endl);
+            << "Usage: component_container_mt --thread-num <number of thread>");
           return 0;
         }
         try {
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
I am not really inclined to have this option. instead, how about providing documents how to support the custom component container? because this kind of extra option and arguments would be required and dependent on use cases?

fujitatomoya:
CC: @clalancette @ivanpauno what would you think?

wep21:
@fujitatomoya
> this kind of extra option and arguments would be required and dependent on use cases?

I think it's better for `component_container_mt` to have the same option or arguments as `MultiThreadedExecutor`.

rclcpp_components/src/component_container_mt.cpp
```diff
@@ -21,8 +21,27 @@
...
   /// Component container with a multi-threaded executor.
-  rclcpp::init(argc, argv);
-  auto exec = std::make_shared<rclcpp::executors::MultiThreadedExecutor>();
+  auto vargv = rclcpp::init_and_remove_ros_arguments(argc, argv);
+  size_t number_of_threads{0};
+  rclcpp::Logger logger{rclcpp::get_logger("component_container_mt")};
+
+  if (!vargv.empty()) {
+    for (auto itr = vargv.begin(); itr != vargv.end() - 1; ++itr) {
+      if (*itr == "--thread-num" || *itr == "-t") {
```
ivanpauno, line 30: arguments that are not `--thread-num/-t` are being skipped, we should instead fail when finding them

ivanpauno (commented):
Adding this options sounds reasonable to me.
It would also be great to provide a `--help` option.

rclcpp_components/src/component_container_mt.cpp
```diff
@@ -21,8 +21,27 @@
...
+        try {
+          number_of_threads = static_cast<size_t>(std::stoi(*(itr + 1)));
```
fujitatomoya, line 32: what if user accidentally have `--thread-num -10`, it will try to create so many threads in `MultiThreadedExecutor`. i think that case should be avoided.

rclcpp_components/src/component_container_mt.cpp
wep21: FIxed at https://github.com/ros2/rclcpp/pull/1693/commits/e0110cc9c77618db584a9f62809a7e093084df35

rclcpp_components/src/component_container_mt.cpp
wep21: Fixed at https://github.com/ros2/rclcpp/pull/1693/commits/9e89651ba95fc4076309c104786edf69cf65fd56

rclcpp_components/src/component_container_mt.cpp
```diff
@@ -21,8 +21,52 @@
...
+        return 0;
+      } else if (*itr == "--thread-num" || *itr == "-t") {
+        if (itr == vargv.end() - 1) {
+          RCLCPP_ERROR_STREAM(
+            logger, "Empty thread number" << std::endl <<
+              "Usage: component_container_mt --thread-num <number of thread>");
+          return 0;
+        }
+        try {
+          const auto arg_number_of_threads{std::stoi(*(itr + 1))};
```
ivanpauno, line 43: `std::stoull()` seems more appropriate

```suggestion
          const auto arg_number_of_threads{std::stoull(*(itr + 1))};
```

ivanpauno (approved):
@wep21 could you add tests?

Maybe, you can move the function doing argument processing to anoter cpp file and unit test that.

rclcpp_components/src/component_container_mt.cpp
wep21: @ivanpauno
`stoull` cannot handle a minus value which is converted to a huge number. How do you think?

ivanpauno: that's interesting ....
I thought that would throw `std::invalid_argument{}`, but the cpp standard library always manages to surprise me.

ivanpauno: resolving this comment, only adding unit tests is pending then

wjwwood:
Have we considered using ROS parameters to configure the number of threads instead? The container is already a node (so that it can provide ROS Services). It's slightly more tedious on the command line, but from launch it should be somewhat straightforward (may require changes to `ComposableNodeContainer`, but I didn't check).

wep21:
@wjwwood For current implementation, it is difficult to pass ros parameter to executor argument because `ComponentManager` node constructor obtains executor from argument.

wjwwood:
> @wjwwood For current implementation, it is difficult to pass ros parameter to executor argument because ComponentManager node constructor obtains executor from argument.

Interesting, but we could/maybe should change that. There's no strong reason that the executor is needed in the constructor as far as I know. It could be passed in via a method that is called after the constructor.

Also, we could look at the parameter overrides before creating the node to extract the setting, though the API to do so is a bit cumbersome if I remember correctly, but this is also an opportunity to fix that up while we do it.

Anyway, I still think custom argument parsing here is fragile and redundant.

ivanpauno:
Using a parameter sounds fine to me :+1:

ivanpauno:
Replaced by https://github.com/ros2/rclcpp/pull/1708, closing

PR #1702, opened 2021-06-25
Add hook to generate node options in ComponentManager

rebecca-butler:
This PR adds a new container that creates an instance of the domain bridge on startup. The user must pass in a YAML config file specifying the domain IDs and topics to be bridged.

This PR also adds an extra argument to specify domain IDs when loading in components.

Fixes ros2/domain_bridge#31


rclcpp_components/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ find_package(ament_cmake_ros REQUIRED)
 find_package(ament_index_cpp REQUIRED)
 find_package(class_loader REQUIRED)
 find_package(composition_interfaces REQUIRED)
+find_package(domain_bridge REQUIRED)
```
clalancette, line 20: Do we really want to make `domain_bridge` a dependency of `rclcpp_components` like this?  It means bringing `domain_bridge` into the ROS 2 core, so we'd have to add it to the package.xml here, and also to https://github.com/ros2/ros2/blob/master/ros2.repos .  But we should seriously consider if that is what we really want to do; it means that `domain_bridge` becomes release-critical (i.e. we cannot release without it).

I'd much rather see us try to do something like this inside of the `domain_bridge` package instead, and keep it out of the ROS 2 core.

wjwwood: Came here to ask this too. I don't think we want it as a dependency of `rclcpp_components`.

> I'd much rather see us try to do something like this inside of the domain_bridge package instead, and keep it out of the ROS 2 core.

👍

jacobperron: +1 I think this executable should live in the domain_bridge package.

rebecca-butler: Makes sense, thanks for your comments. I've moved `domain_bridge_container.cpp` into `domain_bridge`, removing the need for the dependency.

rclcpp_components/src/component_manager.cpp
```diff
@@ -178,6 +178,17 @@ ComponentManager::OnLoadNode(
                     "Extra component argument 'use_intra_process_comms' must be a boolean");
           }
           options.use_intra_process_comms(extra_argument.get_value<bool>());
+        } else if (extra_argument.get_name() == "domain") {
```
jacobperron, line 181: I think this makes the argument a bit more descriptive:
```suggestion
        } else if (extra_argument.get_name() == "domain_id") {
```

jacobperron:
I don't see anything wrong in letting users configure the domain ID for individual components; but I'd like someone else to take a look, since I'm not sure if there could be any unintended consequences. The alternative is to only allow this in the domain bridge specific container.

wjwwood:
I'm not sure it makes sense, because the components are nodes, but the domain id (AFAIK) can only be changed on the context level, as that is where it maps to a participant. @ivanpauno might have more precise input.

rclcpp_components/src/component_manager.cpp
```diff
@@ -178,6 +178,17 @@ ComponentManager::OnLoadNode(
...
+          if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_INTEGER) {
+            throw ComponentManagerException(
+                    "Extra component argument 'domain' must be an integer");
+          }
+          std::size_t domain_id = extra_argument.get_value<std::size_t>();
```
mjeronimo, line 186: At this point, the extra_argument is an integer (a signed type) and is then being returned as a size_t (unsigned). Negative values would get through the first check and then would be converted to large positive values. Depending on how careful one wants to be, there could be an additional semantic check on the domain_id's value before the conversion.

rebecca-butler:
During our offline discussion, it was decided that the functionality for setting the domain ID should not live in `rclcpp_components` as its use case is specific to the domain bridge. I've added a new component manager class to the `domain_bridge` package in the linked PR.

I was originally going to add setters/getters to the existing component manager so the derived class could have its own implementation of `OnLoadNode` (9f7cedc). After doing this, I realized that the derived class doesn't need to modify most of `OnLoadNode`, only the part where the node options are set. I thought it might be simpler to have a `SetNodeOptions` function that the derived class can easily redefine. This cuts down on code duplication and doesn't require exposing any private members.

rclcpp_components/src/component_manager.cpp
rebecca-butler: I fixed this when I moved the code to the linked PR.

rclcpp_components/src/component_manager.cpp
rebecca-butler: Thanks for pointing that out, I've addressed it in: https://github.com/ros2/domain_bridge/pull/32/commits/c3bbf8e94b3a48ecb30568f3c0728002b3d9a6c8

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -131,6 +131,20 @@ class ComponentManager : public rclcpp::Node
   create_component_factory(const ComponentResource & resource);
 
 protected:
+  /// Set node options for componenet
```
jacobperron, line 134:
```suggestion
  /// Set node options for a component
```

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -131,6 +131,20 @@ class ComponentManager : public rclcpp::Node
...
+   * \param remap_rules node remap rules
+   * \param request information with the node to load
+   * \return node options
+   */
+  RCLCPP_COMPONENTS_PUBLIC
+  virtual rclcpp::NodeOptions
+  SetNodeOptions(
+    std::vector<rclcpp::Parameter> parameters,
+    std::vector<std::string> remap_rules,
+    const std::shared_ptr<LoadNode::Request> request);
```
jacobperron, line 146: I like the idea of a hook to generate node options that children can override.

A couple suggestions:

1. I would rename this method to `GetNodeOptions` or `CreateNodeOptions`, since it's not really setting anything on the component manager.
2. I would only pass a single `request` argument, since the parameters and remap information is already contained there in the request.
```suggestion
  CreateNodeOptions(const std::shared_ptr<LoadNode::Request> request);
```

rclcpp_components/include/rclcpp_components/component_manager.hpp
jacobperron: Or, if we rename the method:
```suggestion
  /// Create node options for loaded component
```

rebecca-butler: Fixed in 5db2a22

rclcpp_components/include/rclcpp_components/component_manager.hpp
rebecca-butler: I've renamed this method to `CreateNodeOptions` in 5db2a22 and removed the `parameters` and `remap_rules` arguments.

jacobperron (approved):
LGTM

rebecca-butler:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14847/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9597/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12514/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15006/)

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -131,6 +131,15 @@ class ComponentManager : public rclcpp::Node
...
 
 protected:
+  /// Create node options for loaded component
+  /*
+   * \param request information with the node to load
+   * \return node options
+   */
+  RCLCPP_COMPONENTS_PUBLIC
+  virtual rclcpp::NodeOptions
+  CreateNodeOptions(const std::shared_ptr<LoadNode::Request> request);
```
wjwwood, line 141: I missed this when skimming before, but can we please change this to be `create_node_options`? We don't use CamelCase for methods in core ROS 2 packages like this one.

wjwwood: Well, I just saw that more of these methods (protected ones only) are CamelCase... so I guess this can be left as-is for now... we really need to change them though.

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -131,6 +131,15 @@ class ComponentManager : public rclcpp::Node
   create_component_factory(const ComponentResource & resource);
 
 protected:
+  /// Create node options for loaded component
+  /*
```
wjwwood, line 135: Also, this should be `/**`, but this whole file is wrong...

PR #1705, opened 2021-07-01
wait for message

Karsten1987:
Introducing "wait_for_message". Similar to ROS1, given either a node & topic or a previously instantiated subscription object, the function waits for the next incoming message until either received or timed out.
I could further see some more overlays to this explicitly taking some QOS parameters, even though one could achieve this as-is by initializing the subscription explicitly. 

I am using it within some ros control functionality and it'd be great to have this backported to existing distros.

### DIFF
--- a/rclcpp/include/rclcpp/wait_for_message.hpp
+++ b/rclcpp/include/rclcpp/wait_for_message.hpp
@@ -15,6 +15,9 @@
 #ifndef RCLCPP__WAIT_FOR_MESSAGE_HPP_
 #define RCLCPP__WAIT_FOR_MESSAGE_HPP_
 
+#include <memory>
+#include <string>
+
 #include "rclcpp/node.hpp"
 #include "rclcpp/visibility_control.hpp"
 #include "rclcpp/wait_set.hpp"
@@ -37,19 +40,19 @@ bool wait_for_message(
   std::shared_ptr<rclcpp::Subscription<MsgT>> subscription,
   std::chrono::duration<Rep, Period> time_to_wait = std::chrono::duration<Rep, Period>(-1))
 {
-    rclcpp::WaitSet wait_set;
-    wait_set.add_subscription(subscription);
-    auto ret = wait_set.wait(time_to_wait);
-    if (ret.kind() != rclcpp::WaitResultKind::Ready) {
-      return false;
-    }
+  rclcpp::WaitSet wait_set;
+  wait_set.add_subscription(subscription);
+  auto ret = wait_set.wait(time_to_wait);
+  if (ret.kind() != rclcpp::WaitResultKind::Ready) {
+    return false;
+  }
 
-    rclcpp::MessageInfo info;
-    if (!subscription->take(out, info)) {
-      return false;
-    }
+  rclcpp::MessageInfo info;
+  if (!subscription->take(out, info)) {
+    return false;
+  }
 
-    return true;
+  return true;
 }
 
 /// Wait for the next incoming message.
--- a/rclcpp/include/rclcpp/wait_for_message.hpp
+++ b/rclcpp/include/rclcpp/wait_for_message.hpp
@@ -1,0 +1,78 @@
+// Copyright 2021 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__WAIT_FOR_MESSAGE_HPP_
+#define RCLCPP__WAIT_FOR_MESSAGE_HPP_
+
+#include "rclcpp/node.hpp"
+#include "rclcpp/visibility_control.hpp"
+#include "rclcpp/wait_set.hpp"
+
+namespace rclcpp
+{
+/// Wait for the next incoming message.
+/**
+ * Given an already initialized subscription,
+ * wait for the next incoming message to arrive before the specified timeout.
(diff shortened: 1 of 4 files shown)


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14834/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9584/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12500/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14991/)

rclcpp/include/rclcpp/wait_for_message.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * \param[out] out is the message to be filled when a new message is arriving.
+ * \param[in] subscription shared pointer to a previously initialized subscription.
+ * \param[in] time_to_wait parameter specifying the timeout before returning.
+ * \return true if a message was successfully received, false otherwise.
+ */
+template<class MsgT, class Rep = int64_t, class Period = std::milli>
+bool wait_for_message(
+  MsgT & out,
+  std::shared_ptr<rclcpp::Subscription<MsgT>> subscription,
+  std::chrono::duration<Rep, Period> time_to_wait = std::chrono::duration<Rep, Period>(-1))
```
wjwwood, line 41: One issue here is that the `rclcpp::WaitSet` does not add a guard condition that is triggered on shutdown. So if you use `-1` this will actually block forever, until a message is received, even if you get a SIGINT or shutdown is called somewhere.

So you'd either want to add that yourself, or use an executor. I know you want to take the message without a callback/callback_group, so the wait set is more ideal. But you'll need to add the guard condition yourself. You can see how executors do it:

https://github.com/ros2/rclcpp/blob/e9e398d2d49de2ecc3148d425f8659d9db74918c/rclcpp/src/rclcpp/executor.cpp#L59-L65

Karsten1987: I've added a guard condition and hooked it up to the shutdown callbacks of the context. I am deducing the context from the incoming node - hope this is the right approach.

I couldn't really find another way of detecting whether it was indeed the GC which triggered the waitset to return so I've manually added a `shutdown_requested` flag. Let me know if there's a better way for it.

Karsten1987:
@wjwwood just a friendly ping.

rclcpp/include/rclcpp/wait_for_message.hpp
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+  rclcpp::WaitSet wait_set;
+  wait_set.add_subscription(subscription);
+  wait_set.add_guard_condition(gc);
+  auto ret = wait_set.wait(time_to_wait);
+  if (ret.kind() != rclcpp::WaitResultKind::Ready) {
+    return false;
+  }
+  if (shutdown_requested) {
```
wjwwood, line 63: You could also check here if the gc was set, rather than having a separate variable which is set in the shutdown callback.

rclcpp/include/rclcpp/wait_for_message.hpp
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Given an already initialized subscription,
+ * wait for the next incoming message to arrive before the specified timeout.
+ *
+ * \param[out] out is the message to be filled when a new message is arriving.
+ * \param[in] subscription shared pointer to a previously initialized subscription.
+ * \param[in] context shared pointer to a context to watch for SIGINT requests.
+ * \param[in] time_to_wait parameter specifying the timeout before returning.
+ * \return true if a message was successfully received, false otherwise.
+ */
```
wjwwood, line 37: You should mention something about the behavior when shutdown is called here.

wjwwood (approved):
lgtm, with one non-blocking suggestion on how it could be simplified.

fujitatomoya (approved):
lgtm

fujitatomoya:
@Karsten1987 quick question, these APIs are expected to use just to check the next message is ready for topic, right? user might use this API to get the message every time, but recreating subscription every time does not look good to me. just checking.

rclcpp/include/rclcpp/wait_for_message.hpp
Karsten1987: how do I do this? Can you point me to some examples?

Karsten1987:
@fujitatomoya you're right - the idea is to block until a single message arrived (or shutdown/timeout happened). The typical use case is e.g. to wait for a message on a 'latched' topic such as `/robot_description` where you need the info of the message to initialize part of your system.

> user might use this API to get the message every time, but recreating subscription every time does not look good to me

Agreed, but even then the overlay function can be used which allows you to pass in an already created subscription to the wait function. So you can explicitly wait for a new message on an existing subscription.

rclcpp/include/rclcpp/wait_for_message.hpp
wjwwood: Here's an example from the not quiet yet merged wait set examples:

https://github.com/ros2/examples/pull/315/files#diff-9f3dc30bbe08c799d6cc2ffa4f8ef065efd3755610c60f077ffa74b543c1f435R50-R77

Karsten1987: Wouldn't the size of guard conditions within a waitset still be the same whether or not they've triggered? This backs up what I am seeing here that the number of guard conditions is always `>0` when the waitset is ready - independent of whether the GC fired or not.

wjwwood: I don't understand the question. The size doesn't change, no, but that's not what the logic checks?

Karsten1987: My bad, I misread the example.

Karsten1987:
Thanks for the reviews. I think I've address all comments and take the approvals to merge once the CI below turns out to be green:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14944/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9686/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12621/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15109/)

Issue #1707, opened 2021-07-01
Service server: Extend API to be able to defer a response


ivanpauno:
#### Feature description

Currently, when creating a service one must provide a callback that takes a request and returns a response.
In some cases, the response cannot be calculated immediately.
It would be useful to also be able to create a service server only providing a callback that takes a request (the request message and the request id), but defers to send a response.
In that case, user code is responsible of sending a response later.

#### Implementation considerations

I'm currently using a [hack](https://github.com/ros2/domain_bridge/blob/ab5d3ab13e009eacd98ad11c848da639e06fc59a/include/domain_bridge/service_bridge_impl.inc#L84-L121) in the domain bridge to work around the issue.
The use case is bridging a service between two domains.

In a service callback we want to send the request to another server in a different domain.
A response cannot immediately be sent, because we cannot block in the service callback waiting for the other service to respond.
If we could defer to send a response, we would only need to capture the "request id" in an "async_send_request" callback to the original server, and use in that callback that "request id" and response to complete the bridge.
See [here](https://github.com/ros2/domain_bridge/blob/ab5d3ab13e009eacd98ad11c848da639e06fc59a/include/domain_bridge/service_bridge_impl.inc#L193-L211).

wjwwood:
Is this the same as https://github.com/ros2/rclcpp/issues/491?

ivanpauno:
> Is this the same as #491?

Yes, it's the same.
My idea was to extend `Service` constructor so that it also accepts a callback with signature

```cpp
void (
      const std::shared_ptr<rmw_request_id_t>,
      const std::shared_ptr<typename ServiceT::Request>);
```

instead of adding an `AsyncService` class.

@wjwwood what do you think?

wjwwood:
> Yes, it's the same.

Let's close one or the other then.

wjwwood:
> instead of adding an AsyncService class.

Yes, I think this is preferable.

PR #1709, opened 2021-07-08
Support to defer to send a response in services

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1707, https://github.com/ros2/rclcpp/issues/491.

TODO: add tests

### DIFF
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -15,10 +15,12 @@
 #ifndef RCLCPP__ANY_SERVICE_CALLBACK_HPP_
 #define RCLCPP__ANY_SERVICE_CALLBACK_HPP_
 
+#include <variant>
 #include <functional>
 #include <memory>
 #include <stdexcept>
 #include <type_traits>
+#include <utility>
 
 #include "rclcpp/function_traits.hpp"
 #include "rclcpp/visibility_control.hpp"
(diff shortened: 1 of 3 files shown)


rclcpp/include/rclcpp/any_service_callback.hpp
```diff
@@ -29,93 +31,116 @@
...
+      std::shared_ptr<typename ServiceT::Response>
+    )>;
+  using SharedPtrDeferResponseCallback = std::function<
+    void (
+      std::shared_ptr<rmw_request_id_t>,
+      std::shared_ptr<typename ServiceT::Request>
+    )>;
+
+  std::variant<
+    std::monostate,
```
wjwwood, line 140: Is the monostate needed if the callbacks can be default initialized?

wjwwood (approved):
lgtm, with one question.

rclcpp/include/rclcpp/any_service_callback.hpp
ivanpauno: It's not needed.
The approach I used is to use `std::monostate` as the default, and to check that the callbacks passed to `set()` as an argument are not null.
e.g. if `callback_` is in the `SharedPtrCallback` variant, I don't need to check if the callback is actually `nullptr` or not, as that was checked before.

I think that `std::monostate` just use an extra value of the variant discriminant, so implementation wise the difference doesn't matter much.

ivanpauno:
> Is the TODO done? I guess not, only tests have been updated, but not new ones added.

Not yet, I will add some tests before running CI.

ivanpauno:
All client/service tests were in `test_rclcpp`, so I have opened https://github.com/ros2/system_tests/pull/478.

I forgot that in most cases you actuall also want to receive a `std::shared_ptr<Service>` in the callback, for e.g. doing things like this:

https://github.com/ivanpauno/domain_bridge/blob/5377aaf95a1b812c8d1df52663a1e1ca5feb6ca3/include/domain_bridge/service_bridge_impl.inc#L181-L195

For being able to do this, we would need to add `std::enable_shared_from_this` as a base class of `Service`, @wjwwood do you think that would be okay?
You can also work around that problem by creating a simpler wrapper class, similar of what I did in the `test_rclcpp` PR, but having direct support for that in rclcpp is maybe not a bad idea.

wjwwood:
Well, the user can always capture the service in a lambda or use `std::bind`. They also have to keep the service in scope anyway, so I don't think we need to do it for them.

That being said, we can have the enable shared from this, but maybe we should avoid it if we can.

ivanpauno:
> Well, the user can always capture the service in a lambda or use std::bind. They also have to keep the service in scope anyway, so I don't think we need to do it for them.

The problem is that they can't because they have to pass the lambda to `create_service()`, which is the method returning the `service` object, so there's a chicken and egg problem.

wjwwood:
> The problem is that they can't because they have to pass the lambda to create_service(), which is the method returning the service object, so there's a chicken and egg problem.

Ah, I see. Hmm, in that case I suppose it would be better to do as you say.

wjwwood (commented):
lgtm, waiting to approve until you make the change you mentioned.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14903/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9646/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12572/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15058/)

wjwwood (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14911/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9653/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12581/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15068/)

ivanpauno:
Windows CI again:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15069/)

rclcpp/test/rclcpp/test_any_service_callback.cpp
```diff
@@ -73,6 +74,36 @@ TEST_F(TestAnyServiceCallback, set_and_dispatch_header) {
 
   any_service_callback_.set(callback_with_header);
   EXPECT_NO_THROW(
-    any_service_callback_.dispatch(request_header_, request_, response_));
+    EXPECT_NE(nullptr, any_service_callback_.dispatch(nullptr, request_header_, request_)));
+  EXPECT_EQ(callback_with_header_calls, 1);
+}
+
+TEST_F(TestAnyServiceCallback, set_and_dispatch_defered) {
```
aboseley, line 81: It would be great if an example was added to ros/examples for this new feature

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/deferrable-canceleable-lifecycle-transitions/32318/1

rclcpp/test/rclcpp/test_any_service_callback.cpp
tgroechel: For anyone coming back to this later, I threw together an example of async services here: https://github.com/tgroechel/lifecycle_prac/blob/main/src/async_srv.cpp#L70-L108

Haven't compiled it yet but should give the idea (I patched this together while doing deferrable lifecycle transitions)

tgroechel: And here is an example (also not compiled) leveraging the `SharedFutureResponse` in order to call a service within a service: https://github.com/tgroechel/lifecycle_prac/blob/main/src/async_srv.cpp#L10-L69C1

Issue #1706, opened 2021-07-01
Publishing right before shutdown due to SIGINT


erikschuitema:
#### Feature description
Sometimes it is useful to publish messages right before shutting down, for example, a (lifecycle) state change message.

If a node shuts down due to a SIGINT signal and uses the default signal handler, the global context is shutdown as well. If `rclcpp::spin(node)` is used inside `main()` and SIGINT ends it, any code following it cannot publish because of this.

The callbacks installed with `rclcpp::on_shutdown()` are called _after_ the shutdown, so this is also not the place to publish.

Currently, the only possible way I could find is to install a signal handler before `rclcpp::init()`, which will be stored as the 'old' signal handler inside the `rclcpp::SignalHandler` singleton when it installs the default handler. This 'old' handler does get called before actually shutting down.

Since signal handlers are fairly limited and typically require global variables to do anything useful, it would be a nice feature if callbacks could be installed that are called before a shutdown. 

#### Implementation considerations
One option would be to have `rclcpp::on_shutdown()` accept an option to specify that the callback be called either before or after shutting down.
This implies that the callback is executed before/after any call to `rclcpp::shutdown()`, instead of only after SIGINT, but this could actually be an advantage.

wjwwood:
So, having a callback before shutdown is pretty straightforward, we would just need to add a second loop through the callbacks before shutdown here:

https://github.com/ros2/rclcpp/blob/e9e398d2d49de2ecc3148d425f8659d9db74918c/rclcpp/src/rclcpp/context.cpp#L310-L323

As you suggested, with either a flag to `on_shutdown()` or something similar we could separate pre- and post- shutdown callbacks.

----

However, I'm not sure a pre-shutdown callback will be sufficient. If you publish and then shutdown and then (presumably) continue by shutting the process down immediately, you may never send the published message, because publishing may be done asynchronously or not depending on the middleware implementation. This is why people are working on this issue: https://github.com/ros2/rmw/issues/295 ("Add ability to confirm reader receive all sent messages"). You could use something like this in an "on shutdown" callback, but you wouldn't want to block for too long as you'd be holding up all the other callbacks and the shutdown of other contexts.

wjwwood:
But either way a feature to do callbacks pre-shutdown might be generally useful.

erikschuitema:
[...]

PR #1714, opened 2021-07-14
Add the interface for pre-shutdown callback

Barry-Xu-2018:
This commit is to add the process for pre-shutdown callback.   
Mentioned at https://github.com/ros2/rclcpp/issues/1706#issuecomment-872613908  

This implementation follow existing interfaces as below:
add_on_shutdown_callback  
remove_on_shutdown_callback   
get_on_shutdown_callbacks   

Add interfaces as below: 
add_pre_shutdown_callback
remove_pre_shutdown_callback
get_pre_shutdown_callbacks

If add new interface for post_shutdown, current implementation have to introduce new interfaces.  
I think we can use one interface for different type of shutdown callback.  
e.g.
```c++
  enum class ShutdownType
  {
    pre_shutdown,
    on_shutdown
  };

  add_shutdown_callback(
    ShutdownType shutdown_type,
    ShutdownCallback callback);
```
But this will obsolete current interface.   
Could you share your opinion ?

### DIFF
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -200,7 +200,7 @@ public:
    * to last step in shutdown().
    *
    * When shutdown occurs due to the signal handler, these callbacks are run
-   * asynchronoulsy in the dedicated singal handling thread.
+   * asynchronously in the dedicated singal handling thread.
    *
    * Also, shutdown() may be called from the destructor of this function.
    * Therefore, it is not safe to throw exceptions from these callbacks.
@@ -224,7 +224,7 @@ public:
    * to last step in shutdown().
    *
    * When shutdown occurs due to the signal handler, these callbacks are run
-   * asynchronously in the dedicated singal handling thread.
+   * asynchronously in the dedicated signal handling thread.
    *
    * Also, shutdown() may be called from the destructor of this function.
    * Therefore, it is not safe to throw exceptions from these callbacks.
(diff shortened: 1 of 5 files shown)


fujitatomoya (commented):
Does it sound asymmetrical for `OnShutdownCallbackType` and `PreShutdownCallbackHandle`? We would just want to name `PostShutdownCallbackType`, although we`d want to keep the current interfaces.

Barry-Xu-2018:
> Does it sound asymmetrical for OnShutdownCallbackType and PreShutdownCallbackHandle? We would just want to name PostShutdownCallbackType, although we`d want to keep the current interfaces.

I want to say how to deal with while new added interfaces are similar to existing interface.  
> If add new interface for post_shutdown, current implementation have to introduce new interfaces.
This description is unsuitable. post_shutdown is  on_shutdown. 

I think we can keep the same interface for pre-shutdown and post-shutdown (on-shutdown).   
e.g.  
```
add_shutdown_callback(
    ShutdownType shutdown_type,
    ShutdownCallback callback);

remove_shutdown_callback(
    ShutdownType shutdown_type,
    ...) 

get_shutdown_callbacks(
   ShutdownType shutdown_type,
   ...)
```
But this will obsolete current existing interface step by step.   

If adopt current implementation of this commit, it will add new 3 interfaces. But I think it is not a good way. So want to discuss this.

Barry-Xu-2018:
@wjwwood @ivanpauno  
Do you any opinion about the implementation of pre-shutdown callbacks ?

fujitatomoya:
https://ci.ros2.org/job/ci_windows/15745/testReport/junit/launch_testing_ros.test.examples/set_param_launch_test/set_param_launch_test/ is unrelated to the fix. i will go ahead to merge.

PR #1715, opened 2021-07-15
Change log level for lifecycle_publisher

alsora:
De-activating a lifecycle publisher while the function that was invoking `publish` is still running floods the log of useless warning messages.
This requires to add a boolean check around the publish call, thus making useless the choice of a lifecycle publisher over a regular publisher.


wjwwood:
I think having a visible warning might still be worth having, @Karsten1987 what do you think?

We have logging macros like "throttled", "once", and even ones that have arbitrary predicates, e.g.:

https://docs.ros2.org/foxy/api/rclcpp/logging_8hpp.html#a451bee77c253ec72f4984bb577ff818a

Maybe we could warn throttled, and debug every time? (or something like that)

alsora:
Nice, I didn't know about those macros.
Ideally I would like a warn log only once but controlled by a flag that gets reset after a transition to the active state.
We can then also have debug logs all the time

wjwwood:
Either way works for me, you can have your own flag to control when using this macro and a custom function or expression.

fujitatomoya (commented):
I am not really inclined to have this dedicated flag to control print or not to break the API. @alsora do you have particular reason to use `RCLCPP_WARN_FUNCTION`?

how about,

- RCLCPP_DEBUG, then user can control the logging.
- RCLCPP_WARN_ONCE only once state is deactivated, then reset that once it becomes activated internally w/o changing API?

alsora:
@fujitatomoya why is this breaking the API?
My first implementation was using RCLCPP_DEBUG, I then changed it after the other review comments.

fujitatomoya:
@alsora sorry my bad, i though new parameter for the method, but just a constructor.

> My first implementation was using RCLCPP_DEBUG

okay, i am good with this one too.

wjwwood (approved):
lgtm

Just a note, this won't be backport-able since it will break ABI.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14980/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9721/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12661/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15144/)

wjwwood:
Looks like it doesn't compile.

wjwwood:
Looks like it tries to `*function` and that doesn't work for lambda...  Not sure what the right thing to do there is.

wjwwood:
We can't coerce a lambda that captures into a function pointer... Unfortunately. So I'm not actually sure how to fix this...

I think we need a new kind of macro that takes a function _and_ a `void *` state argument. We could wrap that up so lambdas with captures work in the rclcpp macro, but at the bottom I think we need a new macro in rcutils that takes a function and a state to be passed to that function each time it is called.

alsora:
My bad, I only built the code without unit tests, but this file is just a header so it didn't actually got built.
I went back to use `RCLCPP_WARN` and added an helper function that manages the logging logic

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -130,7 +124,29 @@ class LifecyclePublisher : public LifecyclePublisherInterface,
   }
 
 private:
+  /**
+   * @brief Helper function that logs a message saying that publisher can't publish
+   * because it's not enabled.
+   */
```
wjwwood, line 130:
```suggestion
  /// Helper function that logs a message saying that publisher can't publish because it's not enabled.
```

wjwwood: We don't use `@` but instead use `\` in our doxygen blocks, and we prefer an implicit "brief" avoiding the command.

wjwwood: By the way, my suggestion is too long, so either shorten the message or do multi-line like this:

```suggestion
   /// Helper function that logs a message saying that publisher can't publish
   /// because it's not enabled.
   /**
    * Anything else you have to say/document.
    */
```

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15026/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9763/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12716/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15188/)

wjwwood:
The cmake warning is probably due to this needing to be rebased. I'll wait for feedback on doc changes, and if you can rebase it at the same time I'd appreciate that.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15076/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9808/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12761/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15243/)

PR #1717, opened 2021-07-16
to create a sublogger while getting child of Logger

iuhilnehc-ynos:
use new APIs based on https://github.com/ros2/rcl/pull/921

### DIFF
--- a/rclcpp/include/rclcpp/logger.hpp
+++ b/rclcpp/include/rclcpp/logger.hpp
@@ -152,9 +152,6 @@ public:
    * For example, ```get_logger('abc').get_child('def')``` will return a logger
    * with name `abc.def`.
    *
-   * It is recommended to use a temporary variable to store the return logger
-   * instead of calling it every time.
-   *
    * \param[in] suffix the child logger's suffix
    * \return a logger with the fully-qualified name including the suffix, or
    * \return a dummy logger if this logger is invalid (e.g. because logging is
--- a/rclcpp/include/rclcpp/logger.hpp
+++ b/rclcpp/include/rclcpp/logger.hpp
@@ -17,6 +17,7 @@
 
 #include <memory>
 #include <string>
+#include <utility>
 
 #include "rclcpp/visibility_control.hpp"
 
@@ -122,6 +123,7 @@ private:
   : name_(new std::string(name)) {}
 
   std::shared_ptr<const std::string> name_;
+  std::shared_ptr<std::pair<std::string, std::string>> logger_sublogger_pairname_ = nullptr;
 
 public:
   RCLCPP_PUBLIC
@@ -150,6 +152,9 @@ public:
    * For example, ```get_logger('abc').get_child('def')``` will return a logger
    * with name `abc.def`.
    *
+   * It is recommended to use a temporary variable to store the return logger
+   * instead of calling it every time.
+   *
    * \param[in] suffix the child logger's suffix
    * \return a logger with the fully-qualified name including the suffix, or
    * \return a dummy logger if this logger is invalid (e.g. because logging is
@@ -157,13 +162,7 @@ public:
    */
   RCLCPP_PUBLIC
   Logger
-  get_child(const std::string & suffix)
-  {
-    if (!name_) {
-      return Logger();
-    }
-    return Logger(*name_ + "." + suffix);
-  }
+  get_child(const std::string & suffix);
 
   /// Set level for current logger.
   /**
--- a/rclcpp/src/rclcpp/logger.cpp
+++ b/rclcpp/src/rclcpp/logger.cpp
@@ -12,13 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
 #include <string>
+#include <utility>
 
 #include "rcl_logging_interface/rcl_logging_interface.h"
+#include "rcl/logging_rosout.h"
 
 #include "rclcpp/exceptions.hpp"
 #include "rclcpp/logger.hpp"
 #include "rclcpp/logging.hpp"
+
+#include "./logging_mutex.hpp"
 
 namespace rclcpp
 {
@@ -60,6 +65,46 @@ get_logging_directory()
   std::string path{log_dir};
   allocator.deallocate(log_dir, allocator.state);
   return path;
+}
+
+Logger
+Logger::get_child(const std::string & suffix)
+{
+  if (!name_) {
+    return Logger();
+  }
+
+  rcl_ret_t rcl_ret = RCL_RET_OK;
+  std::shared_ptr<std::recursive_mutex> logging_mutex;
+  logging_mutex = get_global_logging_mutex();
+  {
+    std::lock_guard<std::recursive_mutex> guard(*logging_mutex);
+    rcl_ret = rcl_logging_rosout_add_sublogger((*name_).c_str(), suffix.c_str());
+    if (RCL_RET_OK != rcl_ret && RCL_RET_SUBLOGGER_ALREADY_EXIST != rcl_ret) {
+      exceptions::throw_from_rcl_error(
+        rcl_ret, "failed to call rcl_logging_rosout_add_sublogger",
+        rcutils_get_error_state(), rcutils_reset_error);
+    }
+  }
+
+  Logger logger(*name_ + "." + suffix);
+  if (RCL_RET_OK == rcl_ret) {
+    logger.logger_sublogger_pairname_.reset(
+      new std::pair<std::string, std::string>({*name_, suffix}),
+      [logging_mutex](std::pair<std::string, std::string> * logger_sublogger_pairname_ptr) {
+        std::lock_guard<std::recursive_mutex> guard(*logging_mutex);
+        rcl_ret_t rcl_ret = rcl_logging_rosout_remove_sublogger(
+          logger_sublogger_pairname_ptr->first.c_str(),
+          logger_sublogger_pairname_ptr->second.c_str());
+        delete logger_sublogger_pairname_ptr;
+        if (RCL_RET_OK != rcl_ret) {
+          exceptions::throw_from_rcl_error(
+            rcl_ret, "failed to call rcl_logging_rosout_remove_sublogger",
+            rcutils_get_error_state(), rcutils_reset_error);
+        }
+      });
+  }
+  return logger;
 }
 
 void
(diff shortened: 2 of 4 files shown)


rclcpp/include/rclcpp/logger.hpp
```diff
@@ -150,20 +152,17 @@ class Logger
    * For example, ```get_logger('abc').get_child('def')``` will return a logger
    * with name `abc.def`.
    *
+   * It is recommended to use a temporary variable to store the return logger
+   * instead of calling it every time.
```
fujitatomoya, line 156: I think it would be okay to call this every time, if sub-logger is already allocated, it just returns it from rcl. or do you want to mention this because of race condition to support multi-thread use case?

fujitatomoya (approved):
a minor comment, but LGTM.

rclcpp/include/rclcpp/logger.hpp
iuhilnehc-ynos: I hope to be able to call it every time, but currently, it seems to be ineffective.
Consider the following situation,

```cpp
RCLCPP_INFO( node.get_child("child"), "a child message")
```

The `node.get_child("child")` will return a temporary variable `Logger` after creating a publisher using `rcl_logging_rosout_add_sublogger`. After publishing "a child message" by the created publisher, the temporary variable `Logger` will be immediately out of its scope, which means the publisher will be finalized. The problem is that it may happen that some external subscriptions might not receive this message.

I have two thoughts about fixing this issue:
1. not to destroy the publisher after Logger is destructed, make the publisher destroyed only if the corresponding node is destructed.
2. to destroy the publisher after Logger is destructed, but calling `rcl_publisher_wait_for_all_acked` for the `rosout` publisher before it's finalized.

It seems I was stuck here because I have no idea about which method could be accepted. If there are any suggestions, I'd appreciate that.

iuhilnehc-ynos: > The problem is that it may happen that some external subscriptions might not receive this message.

I tested again and found that this problem not only occurs when the `rosout` publisher is destroyed immediately after publishing the message but also occurs when the message is sent before the external subscription information is collected on the DDS.

iuhilnehc-ynos: @clalancette @fujitatomoya 

Creating a new publisher for each subordinate logger might not be the correct solution.

fujitatomoya: Ah, right. you are right. it can be gone right away.

fujitatomoya: How about destroy child when parent is destructed?

iuhilnehc-ynos: > but also occurs when the message is sent before the external subscription information is collected on the DDS.

I think this subordinate logger can be just an item sharing the parent entry containing the existing publisher, we can't create a new publisher for this subordinate logger. I have updated https://github.com/ros2/rcl/pull/921, would you mind reviewing it again?

I'll remove this comment. `It is recommended to use a temporary variable to store the return logger ...`

iuhilnehc-ynos: > it can be gone right away

Yes. Besides this, the publisher can't be used right away after it is newly created.

> How about destroy child when parent is destructed?

It only solves the former case (occurs when the rosout publisher is destroyed immediately), but not for the other case (the message is sent before the external subscription information is collected on the DDS.)

iuhilnehc-ynos: Sorry about these comments above, which should be discussed on https://github.com/ros2/rcl/pull/921.

fujitatomoya: > I think this subordinate logger can be just an item sharing the parent entry containing the existing publisher, we can't create a new publisher for this subordinate logger.

this sounds good instead of having the publisher for each one of them.

> but not for the other case (the message is sent before the external subscription information is collected on the DDS.)

why, i though that using parent publisher solved this problem too, because it should've been create before the subordinate logger?

fujitatomoya: @iuhilnehc-ynos friendly ping. just checking if i am missing something.

iuhilnehc-ynos: @fujitatomoya 

> i though that using parent publisher solved this problem too

Yes, using the parent publisher can solve this problem. I commented at [rcl](https://github.com/ros2/rcl/pull/921#discussion_r674481487) and had updated the [source code](https://github.com/ros2/rcl/pull/921/commits/f7a7a06b094ff8eaeb34d191067103087277a485).

And I also removed this comment of `rclcpp`.

fujitatomoya:
@iuhilnehc-ynos can you start CI with https://github.com/ros2/rcl/pull/921?

iuhilnehc-ynos:
@fujitatomoya 

I need to re-check if the implementations are good or not, after that, I'll trigger the CI.

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17781/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12325/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18424/)

fujitatomoya:
@clalancette @wjwwood i am good to go with this change, could you do another review?

this needs to be aligned with https://github.com/ros2/rcl/pull/921

iuhilnehc-ynos:
The [URL for Windows](https://github.com/ros2/rclcpp/pull/1717#issuecomment-1341958787) by https://ci.ros2.org/job/ci_launcher/11246/console seems incorrect (Maybe it's because rebuild the ci_launcher before https://ci.ros2.org/job/ci_launcher/11242/console, but the `windows` in https://ci.ros2.org/job/ci_launcher/11242/console didn't trigger at that time.). 

After checking the commit ID in Windows [Build Status](http://ci.ros2.org/job/ci_windows/18424/) that belongs to https://ci.ros2.org/job/ci_launcher/11242/console, although it is correct, I'd like to re-run CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17792/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12338/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18443/)

fujitatomoya:
@clalancette could you also review this? this depends on https://github.com/ros2/rcl/pull/921, they must be merged together.

rclcpp/include/rclcpp/logger.hpp
```diff
@@ -122,6 +123,7 @@ class Logger
   : name_(new std::string(name)) {}
 
   std::shared_ptr<const std::string> name_;
+  std::shared_ptr<std::pair<std::string, std::string>> logger_sublogger_pairname_ = nullptr;
```
clalancette, line 126: Do we even need to store this pair?  That is, we already have the `name_` above, so wouldn't it be sufficient to just store the `suffix` here?

clalancette (changes requested):
I've got one thing that I think can be simplified; otherwise, this looks pretty good to me.

rclcpp/include/rclcpp/logger.hpp
iuhilnehc-ynos: Yes, we need it because the `name_` in the Logger is a full name.
I think that it's better to keep the same parameters for `rcl_logging_rosout_remove_sublogger` and `rcl_logging_rosout_add_sublogger`.

clalancette: Oh, I see.  I misunderstood how this works.  OK, now I understand, that makes sense.

rclcpp/src/rclcpp/logger.cpp
```diff
@@ -62,6 +67,46 @@ get_logging_directory()
...
+    std::lock_guard<std::recursive_mutex> guard(*logging_mutex);
+    rcl_ret = rcl_logging_rosout_add_sublogger((*name_).c_str(), suffix.c_str());
+    if (RCL_RET_OK != rcl_ret) {
+      exceptions::throw_from_rcl_error(
+        rcl_ret, "failed to call rcl_logging_rosout_add_sublogger",
+        rcutils_get_error_state(), rcutils_reset_error);
+    }
+  }
+
+  Logger logger(*name_ + "." + suffix);
```
clalancette, line 90: This is a minor point, but I think we should use `RCUTILS_LOGGING_SEPARATOR_STRING` here.

rclcpp/test/rclcpp/CMakeLists.txt
```diff
@@ -726,6 +726,23 @@ if(TARGET test_rosout_qos)
...
+    target_link_libraries(test_rosout_subscription${target_suffix} ${PROJECT_NAME})
+    ament_target_dependencies(test_rosout_subscription${target_suffix}
+      "rcl_interfaces"
+      "rmw"
+      "rosidl_runtime_cpp"
+      "rosidl_typesupport_cpp"
+    )
+  endif()
+endfunction()
+call_for_each_rmw_implementation(test_rosout_subscription_for_rmw_implementation)
```
clalancette, line 744: Honestly, I don't think we need to test this for each RMW implementation.  That is, the point of this test is to test the rosout subscription, not to test out how well the RMWs work.  I would just let this one run with whatever is currently configured to be the default RMW.

clalancette (changes requested):
Two small fixes here, then I think this will be good.

rclcpp/src/rclcpp/logger.cpp
iuhilnehc-ynos: Thank you for pointing this.

rclcpp/test/rclcpp/CMakeLists.txt
iuhilnehc-ynos: Updated in https://github.com/ros2/rclcpp/pull/1717/commits/48e13f1b687604a4ecd41ebf3d6c06207ba21caf

clalancette (approved):

iuhilnehc-ynos:

iuhilnehc-ynos:

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18094/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12626/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18738/)

clalancette:
The single failed test on Windows is a known flake, so going ahead and merging this.

PR #1723, opened 2021-07-22
Add in callback_groups_for_each.

clalancette:
The main reason to add this method in is to make accesses to the
callback_groups_ vector thread-safe.  By having a
callback_groups_for_each that accepts a std::function, we can
just have the callers give us the callback they are interested
in, and we can take care of the locking.

The rest of this fairly large PR is cleaning up all of the places
that use get_callback_groups() to instead use
callback_groups_for_each().

This is the solution to the problem identified in https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1296 .  In short, when using a thread to deal with `use_sim_time`, it is the case that the callbacks_group_ vector can be accessed by two threads simultaneously.  In certain circumstances (particularly startup time), this can lead to a crash.  By doing proper locking, we avoid the crash.

There are two other things that I'd like to discuss in relation to this:

1. Should we remove the now-unsafe `get_callback_groups()` method?  I have not done that here, but I really think we should.  It is not currently possible to access the `callback_groups_` vector in a thread-safe way.  Alternatively we could deprecate it, but given that there is really no safe way to use it I'd rather just remove it.
2. When adding the `callback_groups_for_each()` method, I replicated the previous structure where there is a method on both the `NodeBase` and `Node` objects.  The latter is a thin wrapper around the former.  But it seems redundant; `Node` has a public method to get the `NodeBase`, and `NodeBase` has the public method `callback_groups_for_each()`.  So anyplace that currently does `node->callback_groups_for_each()` could be trivially changed to `node->node_base()->callback_groups_for_each()`.  I'd opt for removing the method from the `Node` class, but I'm interested in what others think

Questions, comments, and reviews are all very welcome.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -148,11 +148,6 @@ public:
     rclcpp::CallbackGroupType group_type,
     bool automatically_add_to_executor_with_node = true);
 
-  /// Return the list of callback groups in the node.
-  RCLCPP_PUBLIC
-  const std::vector<rclcpp::CallbackGroup::WeakPtr> &
-  get_callback_groups() const;
-
   /// Iterate over the callback groups in the node, calling the given function on each valid one.
   /**
    * This method is called in a thread-safe way, and also makes sure to only call the given
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -153,6 +153,17 @@ public:
   const std::vector<rclcpp::CallbackGroup::WeakPtr> &
   get_callback_groups() const;
 
+  /// Iterate over the callback groups in the node, calling the given function on each valid one.
+  /**
+   * This method is called in a thread-safe way, and also makes sure to only call the given
+   * function on those items that are still valid.
+   *
+   * \param[in] func The callback function to call on each valid callback group.
+   */
+  RCLCPP_PUBLIC
+  void
+  for_each_callback_group(const node_interfaces::NodeBaseInterface::CallbackGroupFunction & func);
+
   /// Create and return a Publisher.
   /**
    * The rclcpp::QoS has several convenient constructors, including a
--- a/rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
@@ -128,6 +128,20 @@ public:
   const std::vector<rclcpp::CallbackGroup::WeakPtr> &
   get_callback_groups() const = 0;
 
+  using CallbackGroupFunction = std::function<void (rclcpp::CallbackGroup::SharedPtr)>;
+
+  /// Iterate over the stored callback groups, calling the given function on each valid one.
+  /**
+   * This method is called in a thread-safe way, and also makes sure to only call the given
+   * function on those items that are still valid.
+   *
+   * \param[in] func The callback function to call on each valid callback group.
+   */
+  RCLCPP_PUBLIC
+  virtual
+  void
+  for_each_callback_group(const CallbackGroupFunction & func) = 0;
+
   /// Return the atomic bool which is used to ensure only one executor is used.
   RCLCPP_PUBLIC
   virtual
--- a/rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
@@ -122,12 +122,6 @@ public:
   bool
   callback_group_in_node(rclcpp::CallbackGroup::SharedPtr group) = 0;
 
-  /// Return list of callback groups associated with this node.
-  RCLCPP_PUBLIC
-  virtual
-  const std::vector<rclcpp::CallbackGroup::WeakPtr> &
-  get_callback_groups() const = 0;
-
   using CallbackGroupFunction = std::function<void (rclcpp::CallbackGroup::SharedPtr)>;
 
   /// Iterate over the stored callback groups, calling the given function on each valid one.
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -190,14 +190,12 @@ Executor::add_callback_groups_from_nodes_associated_to_executor()
   for (auto & weak_node : weak_nodes_) {
     auto node = weak_node.lock();
     if (node) {
-      auto group_ptrs = node->get_callback_groups();
-      std::for_each(
-        group_ptrs.begin(), group_ptrs.end(),
-        [this, node](rclcpp::CallbackGroup::WeakPtr group_ptr)
+      node->for_each_callback_group(
+        [this, node](rclcpp::CallbackGroup::SharedPtr shared_group_ptr)
         {
-          auto shared_group_ptr = group_ptr.lock();
-          if (shared_group_ptr && shared_group_ptr->automatically_add_to_executor_with_node() &&
-          !shared_group_ptr->get_associated_with_executor_atomic().load())
+          if (
+            shared_group_ptr->automatically_add_to_executor_with_node() &&
+            !shared_group_ptr->get_associated_with_executor_atomic().load())
           {
             this->add_callback_group_to_map(
               shared_group_ptr,
@@ -273,18 +271,20 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
             "' has already been added to an executor.");
   }
   std::lock_guard<std::mutex> guard{mutex_};
-  for (auto & weak_group : node_ptr->get_callback_groups()) {
-    auto group_ptr = weak_group.lock();
-    if (group_ptr != nullptr && !group_ptr->get_associated_with_executor_atomic().load() &&
-      group_ptr->automatically_add_to_executor_with_node())
+  node_ptr->for_each_callback_group(
+    [this, node_ptr, notify](rclcpp::CallbackGroup::SharedPtr group_ptr)
     {
-      this->add_callback_group_to_map(
-        group_ptr,
-        node_ptr,
-        weak_groups_to_nodes_associated_with_executor_,
-        notify);
-    }
-  }
+      if (!group_ptr->get_associated_with_executor_atomic().load() &&
+      group_ptr->automatically_add_to_executor_with_node())
+      {
+        this->add_callback_group_to_map(
+          group_ptr,
+          node_ptr,
+          weak_groups_to_nodes_associated_with_executor_,
+          notify);
+      }
+    });
+
   weak_nodes_.push_back(node_ptr);
 }
 
--- a/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
+++ b/rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
@@ -290,18 +290,20 @@ StaticExecutorEntitiesCollector::add_node(
   if (has_executor.exchange(true)) {
     throw std::runtime_error("Node has already been added to an executor.");
   }
-  for (const auto & weak_group : node_ptr->get_callback_groups()) {
-    auto group_ptr = weak_group.lock();
-    if (group_ptr != nullptr && !group_ptr->get_associated_with_executor_atomic().load() &&
-      group_ptr->automatically_add_to_executor_with_node())
+  node_ptr->for_each_callback_group(
+    [this, node_ptr, &is_new_node](rclcpp::CallbackGroup::SharedPtr group_ptr)
     {
-      is_new_node = (add_callback_group(
+      if (
+        !group_ptr->get_associated_with_executor_atomic().load() &&
+        group_ptr->automatically_add_to_executor_with_node())
+      {
+        is_new_node = (add_callback_group(
           group_ptr,
           node_ptr,
           weak_groups_to_nodes_associated_with_executor_) ||
         is_new_node);
-    }
-  }
+      }
+    });
   weak_nodes_.push_back(node_ptr);
   return is_new_node;
 }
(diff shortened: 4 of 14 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14918/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9660/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12590/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15077/)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -153,6 +153,11 @@ class Node : public std::enable_shared_from_this<Node>
   const std::vector<rclcpp::CallbackGroup::WeakPtr> &
   get_callback_groups() const;
 
+  /// Iterate over the callback groups in the node, calling func on each valid one.
+  RCLCPP_PUBLIC
+  void
+  callback_groups_iterate(const node_interfaces::NodeBaseInterface::CallbackGroupCallback & func);
```
ivanpauno, line 159: +1 to `callback_groups_for_each()`, iterate() sounds like it will return an iterator.
Alternative: `for_each_callback_group()`, it is easier to read IMO.

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -2733,13 +2733,33 @@ TEST_F(TestNode, get_publishers_subscriptions_info_by_topic) {
 
 TEST_F(TestNode, callback_groups) {
   auto node = std::make_shared<rclcpp::Node>("node", "ns");
-  size_t num_callback_groups_in_basic_node = node->get_callback_groups().size();
```
ivanpauno, line 2736: we could still use `get_callback_groups().size()` here.
If we want to delete `get_callback_groups()`, which is a good idea IMO, we could add a `callback_groups_size()` method.

ivanpauno (approved):
LGTM with green CI.

I have only left some minor comments

rclcpp/include/rclcpp/node.hpp
clalancette: LOL.  I totally meant this to be `callback_groups_for_each`; I just forgot to commit that change :).  Will update.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -240,11 +238,13 @@ NodeBase::get_default_callback_group()
 bool
 NodeBase::callback_group_in_node(rclcpp::CallbackGroup::SharedPtr group)
 {
+  std::lock_guard<std::mutex> lock(callback_groups_mutex_);
   bool group_belongs_to_this_node = false;
   for (auto & weak_group : this->callback_groups_) {
     auto cur_group = weak_group.lock();
     if (cur_group && (cur_group == group)) {
       group_belongs_to_this_node = true;
+      break;
```
ivanpauno, line 247:
```suggestion
      return true;
```

rclcpp/test/rclcpp/test_node.cpp
clalancette: Yeah, we could.  Right now we only seem to get the `size` of the callback group in tests, so I'm OK with the more verbose way of doing it.  In the future if we decide we need it in more places, we can add it in.

ivanpauno:
(a bit unrelated)

Why are we storing the "callback groups" that we don't need to [automatically add to the executor](https://github.com/ros2/rclcpp/blob/41fedb7beb123dee01b9e54e3579270203cd9404/rclcpp/include/rclcpp/callback_group.hpp#L96) in the node? It seems to be unnecessary.
In that case, it feels that we don't even need to call something like `node->create_callback_group()`, but we should be able to construct the `CallbackGroup` directly.

rclcpp/include/rclcpp/node.hpp
clalancette: Renamed now :).

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
clalancette: I've also fixed this up in the latest code.

clalancette:
> Why are we storing the "callback groups" that we don't need to [automatically add to the executor](https://github.com/ros2/rclcpp/blob/41fedb7beb123dee01b9e54e3579270203cd9404/rclcpp/include/rclcpp/callback_group.hpp#L96) in the node? It seems to be unnecessary.

I'm not sure I totally understand the question.  As far as I can tell, we need to store each callback group we create so that we can later iterate over them, either via `callback_groups_for_each` or for `callback_group_in_node`.

Or did you have a different way in mind to find and iterate over the callback groups?

ivanpauno:
> Or did you have a different way in mind to find and iterate over the callback groups?

AFAIS, in all places where we iterate over callback groups we're skipping the ones that should not be automatically added to the executor, e.g. https://github.com/ros2/rclcpp/blob/551dbcf3472ef555e1d5380e0346585866098a71/rclcpp/src/rclcpp/executor.cpp#L196.
It seems that the only reason that we had for storing a callback groups in `Node` was for adding callback groups to the executor, so I don't see why we also need to store the ones we want to handle separately.

adityapande-1995 (approved):

clalancette:
> It seems that the only reason that we had for storing a callback groups in `Node` was for adding callback groups to the executor, so I don't see why we also need to store the ones we want to handle separately.
> 
> edit: The `callback_group_in_node()` check should be simply deleted in that case.

Ah, I see what you mean.

I agree that we should maybe re-examine how this is done.  That said, I don't necessarily want to do it in here; this PR is already large, and if we decide to go ahead and remove `get_callback_groups`, it is going to get even larger.

What I'll do is to open a separate issue to address that, and we can talk about it there.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -153,6 +153,11 @@ class Node : public std::enable_shared_from_this<Node>
   const std::vector<rclcpp::CallbackGroup::WeakPtr> &
   get_callback_groups() const;
 
+  /// Iterate over the callback groups in the node, calling func on each valid one.
+  RCLCPP_PUBLIC
+  void
+  callback_groups_for_each(const node_interfaces::NodeBaseInterface::CallbackGroupCallback & func);
```
wjwwood, line 159: nitpick: the name sounds weird to me, what about `for_each_callback_group()`? That's _closer_ to being a verb/action than the current name. The current name makes me think like the C api's where it is a `<class_name_method_name>` pattern and implies `callback_groups` is a class, but it isn't.

Another, more complicated solution would be to create a new container like `CallbackGroups` or something which has a `for_each()` method on it.

That all being said, it's just bikeshedding.

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
```diff
@@ -128,6 +128,13 @@ class NodeBaseInterface
   const std::vector<rclcpp::CallbackGroup::WeakPtr> &
   get_callback_groups() const = 0;
 
+  using CallbackGroupCallback = std::function<void (rclcpp::CallbackGroup::SharedPtr)>;
```
wjwwood, line 131: nitpick: another naming thing, when I think of "callback" (especially in the context of ROS) I think of asynchronous and persistent, but that's purely subjective. This is, however, just a function-like thing used in the scope of the method, so I'd have done `CallbackGroupFunction` or something like that. `std::for_each` calls it a "unary function", so it also aligns with that better too:

https://en.cppreference.com/w/cpp/algorithm/for_each

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
```diff
@@ -128,6 +128,13 @@ class NodeBaseInterface
...
+
+  RCLCPP_PUBLIC
```
wjwwood, line 133: Needs docstring

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -273,18 +270,20 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
...
-      this->add_callback_group_to_map(
-        group_ptr,
-        node_ptr,
-        weak_groups_to_nodes_associated_with_executor_,
-        notify);
-    }
-  }
+      if (!group_ptr->get_associated_with_executor_atomic().load() &&
+      group_ptr->automatically_add_to_executor_with_node())
+      {
```
wjwwood, line 278: nitpick: this formatting is confusing to read, I'd suggest:

```suggestion
      if (
        !group_ptr->get_associated_with_executor_atomic().load() &&
        group_ptr->automatically_add_to_executor_with_node()
      ) {
```

rclcpp/src/rclcpp/executors/static_executor_entities_collector.cpp
```diff
@@ -290,18 +290,19 @@ StaticExecutorEntitiesCollector::add_node(
...
-    auto group_ptr = weak_group.lock();
-    if (group_ptr != nullptr && !group_ptr->get_associated_with_executor_atomic().load() &&
-      group_ptr->automatically_add_to_executor_with_node())
+  node_ptr->callback_groups_for_each(
+    [this, node_ptr, &is_new_node](rclcpp::CallbackGroup::SharedPtr group_ptr)
     {
-      is_new_node = (add_callback_group(
+      if (!group_ptr->get_associated_with_executor_atomic().load() &&
+      group_ptr->automatically_add_to_executor_with_node())
+      {
```
wjwwood, line 298: Same:

```suggestion
      if (
        !group_ptr->get_associated_with_executor_atomic().load() &&
        group_ptr->automatically_add_to_executor_with_node()
      ) {
```

rclcpp/src/rclcpp/executor.cpp
wjwwood: And, I know there's other code in this style already committed, but I wanted to point it out on new changes, as a nitpick only.

rclcpp/include/rclcpp/node.hpp
wjwwood: Also, either the name of the method, or the documentation should probably mention the purpose of this method, i.e. that it provides thread-safety by atomically iterating over the callbacks groups (for the whole iteration, as opposed to locking in each loop) and synchronizing with adding and/or deleting callback groups.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -256,6 +254,17 @@ NodeBase::get_callback_groups() const
...
 
+void NodeBase::callback_groups_for_each(const CallbackGroupCallback & func)
+{
+  std::lock_guard<std::mutex> lock(callback_groups_mutex_);
+  for (rclcpp::CallbackGroup::WeakPtr & weak_group : this->callback_groups_) {
+    rclcpp::CallbackGroup::SharedPtr cur_group = weak_group.lock();
+    if (cur_group) {
+      func(cur_group);
+    }
+  }
```
wjwwood, line 265: nitpick: I always suggest avoiding any kind of abbreviation because it makes reading code harder for people who don't speak english, i.e. `cur_group` probably doesn't translate automatically very well. So either `current_group` or just `group`:

```suggestion
  for (rclcpp::CallbackGroup::WeakPtr & weak_group : this->callback_groups_) {
    rclcpp::CallbackGroup::SharedPtr group = weak_group.lock();
    if (group) {
      func(group);
    }
  }
```

I actually feel the same way about `func` and would use `function` myself, but that one is common enough that it probably won't confuse anyone.

rclcpp/test/rclcpp/test_node.cpp
wjwwood: I also think it would be a good idea to get rid of `get_callback_groups()` in its current form, because a) it would force users of it that are outside of our code base to consider if they're using it in a thread-unsafe way and b) because we could provide a const method that returns a copy of the callback groups vector instead, which could be used for `size()` and other things, which I would prefer over a dedicated `get_callback_groups_size()`.

wjwwood:
I also think we might not need to store the callback groups that aren't automatically added to an executor, and that they could even be constructed directly.

wjwwood:
The only thing that might be affected is that notifying any executor that the callback group is associated with that something has been added/removed (the executor needs to interrupt itself in this case to rebuild the wait set), sometimes goes through the node, but again that might be only if the `automatically_add_to_executor_with_node` is true.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
clalancette: Yeah, makes sense.  This was actually just copy-n-pasted from `callback_group_in_node` above, but I'll change it to just "group".  I'm going to leave "func" as-is, since that one is pretty clear.

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
clalancette: Yeah, I didn't love the name `CallbackGroupCallback` either.  I've opted for `CallbackGroupFunction`, so as not to make it too long.  Let me know what you think.

rclcpp/include/rclcpp/node.hpp
clalancette: > nitpick: the name sounds weird to me, what about `for_each_callback_group()`

Ivan also mentioned this, so I'll take that as 2 votes for `for_each_callback_group()`.  I'll change to that instead.

clalancette: > Also, either the name of the method, or the documentation should probably mention the purpose of this method, i.e. that it provides thread-safety by atomically iterating over the callbacks groups (for the whole iteration, as opposed to locking in each loop) and synchronizing with adding and/or deleting callback groups.

Yeah, I'll mention that when I write the documentation.

clalancette:
All right.  I've responded to all of the inline comments, I believe.  I also rebased this onto the latest to get #1726 so windows CI will pass.  Finally, I've removed `get_callback_groups` in a separate commit.  That change also requires a [small PR](https://github.com/ros2/examples/pull/320) to the examples repository, but otherwise requires no further changes in the core.

clalancette:
Full CI (including https://github.com/ros2/examples/pull/320):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14928/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9670/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12602/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15092/)

ivanpauno (approved):

clalancette:
The failing test on Windows is flakey; it also failed on Windows in https://ci.ros2.org/view/nightly/job/nightly_win_rep/2312.  So this is otherwise green.  Since I also have approvals, I'm going to go ahead and merge this one along with https://github.com/ros2/examples/pull/320 .

wjwwood:
👍

PR #1726, opened 2021-07-22
Fix windows CI

ivanpauno:
https://github.com/ros2/rclcpp/pull/1709 broke `rclcpp_components` on Windows nighties, e.g.: https://ci.ros2.org/view/nightly/job/nightly_win_rel/2002/.
I think this is a bug on Windows implementation of `std::bind`, as the previous code really should work.

This does some `if constexpr` magic using function traits to workaround the issue.
I'm surprised we didn't hit a similar issue when using `std::variant` in `AnySubscriptionCallback` implementation, I have no idea of what's the difference.

### DIFF
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -75,7 +75,43 @@ public:
     if (!callback) {
       throw std::invalid_argument("AnyServiceCallback::set(): callback cannot be nullptr");
     }
-    callback_ = std::forward<CallbackT>(callback);
+    // Workaround Windows issue with std::bind
+    if constexpr (
+      rclcpp::function_traits::same_arguments<
+        CallbackT,
+        SharedPtrCallback
+      >::value)
+    {
+      // one would think that the following works:
+      // callback_.emplace<SharedPtrCallback>(callback);
+      // but it does not for some reason.
+      callback_.emplace<1>(callback);
+    } else if constexpr (
+      rclcpp::function_traits::same_arguments<
+        CallbackT,
+        SharedPtrWithRequestHeaderCallback
+      >::value)
+    {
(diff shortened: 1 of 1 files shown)


ivanpauno:
Windows CI:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15084/)

ivanpauno:
Better, run all platforms:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14925/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9667/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12599/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15085/)

ivanpauno:
After searching a bit, I found:

- https://stackoverflow.com/questions/55095155/socket-io-client-stdbind-error-c2338-tuple-index-out-of-bounds
- https://developercommunity.visualstudio.com/t/stdis-convertible-is-broken-for-stdbind-functors/317991

I guess Windows `std::_Binder` implementation is not as SFINAE friendly as it should be.
Maybe gcc is too permissive, I'm not sure.

wjwwood (approved):
lgtm

Yeah, I mean that's why we have `function_traits` in rclcpp.

ivanpauno:
> Yeah, I mean that's why we have function_traits in rclcpp.

I thought that `std::variant::operator=` was going to do all the magic.
It actually did on Linux, so I'm not sure whatever gcc was too permissive or visual studio has a bug.

wjwwood:
> It actually did on Linux, so I'm not sure whatever gcc was too permissive or visual studio has a bug.

Depends on who you ask 🙃

ivanpauno:
CI is green, going in!

Thanks for the review @wjwwood !!!

PR #1727, opened 2021-07-22
Use rcpputils/scope_exit.hpp and remove rclcpp/scope_exit.hpp

### LINKED ISSUE #1725: Remove RCLCPP_SCOPE_EXIT() and use RCPPUTILS_SCOPE_EXIT()
#### Feature description

A scope exit helper macro (`RCPPUTILS_SCOPE_EXIT()`) was added to `rcpputils` in https://github.com/ros2/rcpputils/pull/78, so we could use that here instead of `RCLCPP_SCOPE_EXIT()`. Same for the actual underlying class.

This was mentioned here https://github.com/ros2/rcpputils/pull/78#pullrequestreview-437794401, but I couldn't find an existing issue for it.

The one from `rcpputils` is cancelable while the one currently in `rclcpp` isn't, but I think that should be fine.

I will open a PR for this soon.

christophebedard:
Closes #1725

Requires https://github.com/ros2/system_tests/pull/479 to be merged before this, since some packages in system_tests use the `rclcpp/scope_exit.hpp` header.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>


rclcpp/include/rclcpp/macros.hpp
```diff
@@ -108,7 +108,4 @@
   __RCLCPP_UNIQUE_PTR_ALIAS(__VA_ARGS__) \
   __RCLCPP_MAKE_UNIQUE_DEFINITION(__VA_ARGS__)
 
-#define RCLCPP_STRING_JOIN(arg1, arg2) RCLCPP_DO_STRING_JOIN(arg1, arg2)
-#define RCLCPP_DO_STRING_JOIN(arg1, arg2) arg1 ## arg2
```
wjwwood, line 112: I'm not sure it's possible but deprecating these first would be nice. I don't know if you can get a conditional `#warning` into this, probably not. If not, maybe we should just leave it? I don't like having dead code lying around, but this is trivial and we can't know if someone is using this as it is currently part of the public API. It would also prevent us from backporting this at all (maybe we don't care about that).

rclcpp/include/rclcpp/scope_exit.hpp
```diff
@@ -1,52 +0,0 @@
...
-
-#include <functional>
-
-#include "rclcpp/macros.hpp"
-
-namespace rclcpp
-{
-
-template<typename Callable>
-struct ScopeExit
```
wjwwood, line 30: I think we should deprecate this header before removing it, if it is technically possible. We can, and have, removed things without first deprecating them (but most often out of technical necessity), and think this code is not widely used, but it is in the public API and we should at least try to do a deprecation first.

We can put a `#warning` in this header and possibly even just alias this class to the rcpputils version, assuming the APIs are the same. (also for the macro)

wjwwood (commented):
Changes lgtm, but I am recommending we do a deprecation first.

rclcpp/include/rclcpp/macros.hpp
christophebedard: reverted in 92da89e03244c2a99cf3d18748533e26d3408800

rclcpp/include/rclcpp/scope_exit.hpp
christophebedard: done in 68adbafe7566d76729ce3f405648e59ee8b68caf

To keep it simple I just kept the file and added a warning.

christophebedard: I found some files in system_tests indirectly including this header, so this PR will break that (and other instances of it), but in any case I think we're supposed to "include what we use." We'll see with CI.

christophebedard:
`--packages-above-and-dependencies rclcpp rclcpp_action rclcpp_lifecycle`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14929/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9671/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12603/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15093/)

fujitatomoya (approved):
lgtm

rclcpp/include/rclcpp/scope_exit.hpp
wjwwood: Ok, well another option is that we can wrap the `#warrning` in a conditional, and when including it from within rclcpp we can disable the warning, that way it only applies when external users include it directly.

Or we could just deprecate the `struct ScopeExit`, so people stop using it. I guess they might still be including it even if they stop using it, but then I feel less bad about the header disappearing in the next version.

wjwwood: > in any case I think we're supposed to "include what we use."

I guess you mean, if we change the rclcpp headers so this header is not implicitly included, and then that breaks someone that's their fault for not "including what they use", which I'm also ok with.

wjwwood: Do you want to do anything here or merge it as-is?

christophebedard: > I guess you mean, if we change the rclcpp headers so this header is not implicitly included, and then that breaks someone that's their fault for not "including what they use", which I'm also ok with.

yes exactly.

> Do you want to do anything here or merge it as-is?

Since you're okay with breaking the transitive include, I'd just stick with this for simplicity.

wjwwood (approved):

cottsay:
The ros1_bridge wasn't part of typical CI and doesn't appear to run in the nightly debug/release builds, but it is showing a new CMake warning as a result of this change.

Could someone please follow-up there with a fix? https://github.com/ros2/ros1_bridge/blob/eceda757c066bef36489d4a10ec414d797935081/src/dynamic_bridge.cpp#L39

christophebedard:
sure I can fix it!

Issue #1697, opened 2021-06-18
Calling a dead service causes a memory leak on the client side


umateusz:
## Bug report

If the service is stopped (due to an error or user intervention) and the client keeps trying to call it, it will leak memory. Unprocessed requests will be accumulated in pending_requests_ (https://github.com/ros2/rclcpp/blob/7d8b2690ff188672bfe3703d8b3c074b87303931/rclcpp/include/rclcpp/client.hpp#L379). Is there a way to prevent this e.g. request timeout? 

It should not be possible to leak memory on the client side. User cannot delete old requests, so they will hang until the end of the client's operation. Moreover, the problem will occur if the service does not keep up with the processing of client requests.

IMHO: Using spin_until_future_complete is not a solution as it cannot be called from a composable node and it cannot erase outdated entry from pending_request_ collection.

- Operating System: Ubuntu 20.04
- Installation type: binaries
- Version or commit hash: foxy
- DDS implementation: rmw_cyclonedds_cpp (should also be present for other implementations)
- Client library (if applicable): rclcpp
- 
#### Steps to reproduce issue

I prepared the problem example by modifying the standard (minimal_client, minimal_service) examples and placed it on the branch: https://github.com/umateusz/examples/tree/memory-leak-in-service-client

#### Client:
```
#include <chrono>
#include <cinttypes>
#include <memory>
#include "example_interfaces/srv/add_two_ints.hpp"
#include "rclcpp/rclcpp.hpp"

using AddTwoInts = example_interfaces::srv::AddTwoInts;
using namespace std::chrono_literals;

class ClientWithTimer : public rclcpp::Node
{
public:
  ClientWithTimer() : Node("client_with_timer")
  {
    client_ = create_client<AddTwoInts>("add_two_ints");
    while (!client_->wait_for_service(std::chrono::seconds(1))) {
      if (!rclcpp::ok()) {
        RCLCPP_ERROR(get_logger(), "client interrupted while waiting for service to appear.");
        rclcpp::shutdown();
        return;
      }
      RCLCPP_INFO(get_logger(), "waiting for service to appear...");
    }
    timer_ = create_wall_timer(500ms, std::bind(&ClientWithTimer::timer_callback, this));
  }

private:
  void timer_callback()
  {
    RCLCPP_INFO(this->get_logger(), "Hello, world!");
    auto request = std::make_shared<AddTwoInts::Request>();
    request->a = 41;
    request->b = 1;
    auto result_future = client_->async_send_request(
[...]

PR #1728, opened 2021-07-27
Updating client API to be able to remove pending requests

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1697.

This modifies a few things in the client API:

- Modifies the async_send_request() methods that take a callback as a second argument to not return a future anymore.
  Though this breaks API, nobody was actually using that in the ROS 2 core.
  If someone wants to have both a callback and a future, they can capture the future in a lambda and set it.
  - Alternative: create a `SharedFutureAndRequestID` class, similar to the `FutureAndRequestID` class introduced here.
  - I'm planning to deprecate the current callback signatures that accept a `Future`, and instead have the two listed below.
    That wasn't done here to avoid having to create simultaneous PRs in other repos.
    - void (SharedResponse)
    - void (SharedRequest, SharedResponse)
- Modifies the async_send_request() methods that take only one argument to return a `FutureAndRequestID`.
  - This is a future like class.
    Though it breaks ABI, it has some implicit conversions to `std::future` to not break API.
    The implicit conversion by value was only added to not break existing code and generates a deprecation warning.
- Adds a `remove_pending_request()` method, the return value of any of the async_send_request() overloads can be passed to it.

Examples:

```cpp
auto future = client->async_send_request(request);
auto result = executor->spin_until_complete(future, timeout);
if (result == timeout) {
  client->remove_request(future);
  // handle timeout
}
```

When using callbacks, you need to combine the client with a Timer to prune the old ones:

```cpp
  // this example supposes you're only waiting for one response at a time
  //
  // setup timer
  this->timer = this->create_wall_timer(prune_old_requests, [this] () {
    if (this->client->remove_request(this->last_request_id)) {
      // handle timeout
    }
  };
  this->timer.cancel();

  // when sending a request
  this->last_request_id = client->send_async_request(request, [this] (SharedResponse response) {
    this->handle_response(response);
    this->timer.cancel();
  }
  this->timer.reset();
```

### DIFF
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -448,7 +448,7 @@ private:
    *
    * If the callback is never called, because we never got a reply for the service server, remove_pending_request()
    * has to be called with the returned request id or prune_pending_requests().
-   * Not doing so will make the `Client` instance to use more memory each time a response is not
+   * Not doing so will make the `Client` instance use more memory each time a response is not
    * received from the service server.
    * In this case, it's convenient to setup a timer to cleanup the pending requests.
    *
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -17,12 +17,14 @@
 
 #include <atomic>
 #include <future>
-#include <map>
+#include <unordered_map>
 #include <memory>
+#include <optional>
 #include <sstream>
 #include <string>
 #include <tuple>
 #include <utility>
+#include <variant>
 
 #include "rcl/client.h"
 #include "rcl/error_handling.h"
@@ -178,6 +180,9 @@ template<typename ServiceT>
 class Client : public ClientBase
 {
 public:
+  using Request = typename ServiceT::Request;
+  using Response = typename ServiceT::Response;
+
   using SharedRequest = typename ServiceT::Request::SharedPtr;
   using SharedResponse = typename ServiceT::Response::SharedPtr;
 
@@ -187,6 +192,7 @@ public:
   using SharedPromise = std::shared_ptr<Promise>;
   using SharedPromiseWithRequest = std::shared_ptr<PromiseWithRequest>;
 
+  using Future = std::future<SharedResponse>;
   using SharedFuture = std::shared_future<SharedResponse>;
   using SharedFutureWithRequest = std::shared_future<std::pair<SharedRequest, SharedResponse>>;
 
@@ -194,6 +200,54 @@ public:
   using CallbackWithRequestType = std::function<void (SharedFutureWithRequest)>;
 
   RCLCPP_SMART_PTR_DEFINITIONS(Client)
+
+  class FutureAndRequestId {
+    public:
+      FutureAndRequestId(Future impl, int64_t req_id)
+      : impl_(std::move(impl)), req_id_(req_id)
+      {}
+
+      // implicit convertion to Future by reference for backwards compatibility
+      operator Future&() {return impl_;}
+      // also allow to implicitly convert to a future by value
+      // only for backwards comp
+      [[deprecated("FutureAndRequestId: use take_future() instead of an implicit conversion")]]
+      operator Future() {return impl_;}
+
+      Future
+      take_future() {
+        return impl_; // we're moving the future out
+      }
+
+      int64_t get_request_id() const {return req_id_;}
+
+      // rule of five
+      FutureAndRequestId(FutureAndRequestId && other) noexcept = default;
+      FutureAndRequestId(const FutureAndRequestId & other) = delete;
+      FutureAndRequestId & operator=(FutureAndRequestId && other) noexcept = default;
+      FutureAndRequestId & operator=(const FutureAndRequestId & other) = delete;
+      ~FutureAndRequestId() = default;
+
+      // delegate future like methods to the std::future impl_
+      SharedFuture share() noexcept {return impl_.share();}
+      SharedResponse get() {return impl_.get();}
+      bool valid() const noexcept {return impl_.valid();}
+      void wait() const {return impl_.wait();}
+
+      template<class Rep, class Period>
+      std::future_status wait_for(const std::chrono::duration<Rep,Period> & timeout_duration) const
+      {
+        return impl_.wait_for(timeout_duration);
+      }
+      template<class Clock, class Duration>
+      std::future_status wait_until(const std::chrono::time_point<Clock,Duration> & timeout_time) const
+      {
+        return impl_.wait_until(timeout_time);
+      }
+    private:
+      Future impl_;
+      int64_t req_id_;
+  };
 
   /// Default constructor.
   /**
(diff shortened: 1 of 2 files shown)


ivanpauno:
> When using callbacks, you need to combine the client with a Timer to prune the old ones:

I'm planning to write a more complete example of how to do this, but I would like first to get some early feedback.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -195,6 +201,88 @@ class Client : public ClientBase
...
+    // TODO(ivanpauno): Maybe, deprecate this in favor of get_future() (?)
+    operator Future &() {return impl_;}
+
+    /// Deprecated, use take_future() instead.
+    /**
+     * Allow implicit conversions to `std::future` by value.
+     * \deprecated
+     */
+    [[deprecated("FutureAndRequestId: use take_future() instead of an implicit conversion")]]
+    operator Future() {return impl_;}
```
ivanpauno, line 222: I think I actually don't need this two custom implicit conversions, but actually one to `SharedFuture` to avoid breaking pre-existing code.
I'm taking a look locally.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -195,6 +201,88 @@ class Client : public ClientBase
...
+    /// Move assignment.
+    FutureAndRequestId & operator=(FutureAndRequestId && other) noexcept = default;
+    /// Deleted copy assignment, each instance is a unique owner of the future.
+    FutureAndRequestId & operator=(const FutureAndRequestId & other) = delete;
+    /// Destructor.
+    ~FutureAndRequestId() = default;
+
+private:
+    Future impl_;
+    int64_t req_id_;
```
ivanpauno, line 283: I think we can make this two members public, this is simply a pair there's no invariant here.

The future like methods were added to avoid breaking pre-existing code.
It's not a bad idea to deprecate all of them though, that way we don't break pre-existing code but users get a warning and they take a look to `remove_pending_request()`.

ivanpauno:
> Though this breaks API, nobody was actually using that in the ROS 2 core.

I finally found some tests in `test_rclcpp` that were using that.
Though I still think it's weird, you typically either want a callback or a future, not both.

> If someone wants to have both a callback and a future, they can capture the future in a lambda and set it.

Note: this isn't that easy.
The issue is that `std::function` is a wrapper over a **copyable** callable.
A lambda that captures a promise is not copyable, so you cannot pass that to `async_send_request()`.
To solve that you need to wrap the promise to something that's actually copyable (e.g. a shared_ptr, a wrapper that actually moves on copy, etc).
We could also be delting that requirement from our api by having a `rcpputils::unique_function{}` implementation.

> Alternative: create a SharedFutureAndRequestID class, similar to the FutureAndRequestID class introduced here.

If anyone feels that's a better solution than breaking this use case, I can update the PR.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -292,34 +380,83 @@ class Client : public ClientBase
...
+    auto future = promise.get_future();
+    auto req_id = async_send_request_impl(
+      *request,
+      std::move(promise));
+    return FutureAndRequestId(std::move(future), req_id);
   }
 
+  /// Send a request to the service server and schedule a callback in the executor.
+  /**
+   * Similar to the previous overload, but a callback will automatically be called when a response is get.
```
jacobperron, line 447:
```suggestion
   * Similar to the previous overload, but a callback will automatically be called when a response is received.
```

rclcpp/include/rclcpp/client.hpp
```diff
@@ -292,34 +380,83 @@ class Client : public ClientBase
...
+   *
+   * If the callback is never called, because we never got a reply for the service server, remove_pending_request()
+   * has to be called with the returned request id or prune_pending_requests().
+   * Not doing so will make the `Client` instance to use more memory each time a response is not
```
jacobperron, line 451:
```suggestion
   * Not doing so will make the `Client` instance use more memory each time a response is not
```

jacobperron (commented):
> Though I still think it's weird, you typically either want a callback or a future, not both.

Is there a technical reason for not providing both, or is it just aesthetic? We could still maintain API stability and document that the user can (or should) use the returned future to prune the request if there's a timeout.

---

The approach looks okay to me. Have you thought about a way to handle pruning internally (so the user doesn't have to worry about it)? Alternatively, have you consider something like this: https://github.com/ros2/rclcpp/issues/1697#issuecomment-868829970 ?

ivanpauno:
> Is there a technical reason for not providing both, or is it just aesthetic?

It's mainly the second, but it's also performance.
Why to set a promise value that will be ignored by most users?

> We could still maintain API stability and document that the user can (or should) use the returned future to prune the request if there's a timeout.

We cannot remove the pending request with the future, that's why I created the `FutureAndRequestID` class.
But yes, I can create a `SharedFutureAndRequestID` to keep backwards compatibility.

> That wasn't done here to avoid having to create simultaneous PRs in other repos.
> void (SharedResponse)
> void (SharedRequest, SharedResponse)

If we later add this, we could in that case only return the request id and not a future.
I'm having to set a promise either way, as the current callback signature is taking a SharedResponse, so I will go ahead with the `SharedFutureAndRequestId` solution.

> The approach looks okay to me. Have you thought about a way to handle pruning requests internally (so the user doesn't have to worry about it)?

Yes but that doesn't fit well with the async_send_request API.
My idea is basically this https://github.com/ros2/rclcpp/issues/1697#issuecomment-868838247.
We have a timer for each pending request, and we cancel() the timer when we get a response.
It involves using more resources though (creating timers and adding them to the wait set / executor).
I will explore this solution a bit more.

> Alternatively, have you consider something like this: #1697 (comment) ?

Yes, I think that solution is interesting.
The issue is that we basically don't need the internal `pending_requests_` map in that case, and the behavior to handle a response is different.
As we're not storing the pending requests internally, we can only have one callback to handle all responses (not one per request send).
That means, we need a different class implementing the `ClientBase` interface.
I could add a class like that, but I preferred to first extend the current API.

---

Long term, I would rather have some simpler wrappers of the rcl handles that don't know anything about callbacks at all, and then on top of that we could build a Condition (similar as waitable) and a task concept.

```cpp
Subscription sub{node, topic_name, sub_opts};
Client cli{node, service_names, cli_opts};

executor.on_condition(sub.has_message_condition(), [&sub](TaskHandle handle) {
  // take message and do something with it
  // the TaskHandle can be used to e.g. cancel this task
  // or maybe more powerful things, like swapping the condition the task is waiting on.
});
client.send_request(request);
auto complete_token = executor.on_condition_once(
  cli.has_response_condition(),
  [&cli](TaskHandle handle) {
    // take and handle response here
  });
executor.spin_until_complete(complete_token, timeout);
```

Don't look into the details of the snippet above, but an approach like that has some benefits:

- It separates concerns better.
- You don't need to add everything to the executor.
  e.g. if a client is not waiting on a response, no need to add it to the internal wait set.
         In the same fashion, if you don't need to take data from a subscription for some time, you can avoid to have a task waiting on it.
- Maybe: can handle multithreaded approach without the need of callback groups.

ivanpauno:
I clicked the merge button in the wrong window :man_facepalming:.
I've reverted in https://github.com/ros2/rclcpp/pull/1733, I will open a new PR replacing this one.

PR #1741, opened 2021-08-03
Galactic: for_each_callback_group backport

adityapande-1995:
This PR aims takle https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1296 by backporting https://github.com/ros2/rclcpp/pull/1723 to galactic without breaking ABI. To do this, I added a new non virtual function in node_base class, and added static members to it to keep track of a global map of mutexes. This map stores the mutexes for each node_base instance. However, while using the new ```for_each_callback_group``` method, the pointer of type ```NodeBaseInterface``` has to be typecasted to ```NodeBase```. The global map of mutexes is written as a separate class and access to it is protected by an internal mutex.

# TODO 
* General code cleanup, linting issues
* Modify ```static_executor_entities_collector```
* Modify the corresponding test cases

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -70,7 +70,6 @@
 #include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 
-
 namespace rclcpp
 {
 
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -70,6 +70,7 @@
 #include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+
 namespace rclcpp
 {
 
@@ -152,6 +153,17 @@ public:
   RCLCPP_PUBLIC
   const std::vector<rclcpp::CallbackGroup::WeakPtr> &
   get_callback_groups() const;
+
+  /// Iterate over the callback groups in the node, calling the given function on each valid one.
+  /**
+   * This method is called in a thread-safe way, and also makes sure to only call the given
+   * function on those items that are still valid.
+   *
+   * \param[in] func The callback function to call on each valid callback group.
+   */
+  RCLCPP_PUBLIC
+  void
+  for_each_callback_group(const node_interfaces::NodeBaseInterface::CallbackGroupFunction & func);
 
   /// Create and return a Publisher.
   /**
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -35,6 +35,9 @@ namespace node_interfaces
 {
 
 class map_of_mutexes;
+void global_for_each_callback_group(
+  NodeBaseInterface * node_base_interface,
+  const NodeBaseInterface::CallbackGroupFunction & func);
 
 /// Implementation of the NodeBase part of the Node API.
 class NodeBase : public NodeBaseInterface
@@ -43,9 +46,6 @@ public:
   RCLCPP_SMART_PTR_ALIASES_ONLY(NodeBase)
 
   static map_of_mutexes map_object;
-
-  // Non virtual method
-  void for_each_callback_group(const CallbackGroupFunction & func);
 
   RCLCPP_PUBLIC
   NodeBase(
@@ -164,13 +164,14 @@ public:
   ~map_of_mutexes();
 
   // Methods need to be protected by internal mutex
-  void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase * nodebase);
+  void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
   std::shared_ptr<std::mutex>
-  get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase * nodebase);
-  void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase * nodebase);
+  get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
+  void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
 
   // Members
-  std::unordered_map<const rclcpp::node_interfaces::NodeBase *, std::shared_ptr<std::mutex>> data;
+  std::unordered_map<const rclcpp::node_interfaces::NodeBaseInterface *,
+    std::shared_ptr<std::mutex>> data;
   std::mutex internal_mutex;
 };
 
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -34,10 +34,25 @@ namespace rclcpp
 namespace node_interfaces
 {
 
-class map_of_mutexes;
 void global_for_each_callback_group(
   NodeBaseInterface * node_base_interface,
   const NodeBaseInterface::CallbackGroupFunction & func);
+
+// Class to hold the global map of mutexes
+class map_of_mutexes final
+{
+public:
+  // Methods need to be protected by internal mutex
+  void create_mutex_of_nodebase(const NodeBaseInterface * nodebase);
+  std::shared_ptr<std::mutex>
+  get_mutex_of_nodebase(const NodeBaseInterface * nodebase);
+  void delete_mutex_of_nodebase(const NodeBaseInterface * nodebase);
+
+private:
+  // Members
+  std::unordered_map<const NodeBaseInterface *, std::shared_ptr<std::mutex>> data;
+  std::mutex internal_mutex;
+};
 
 /// Implementation of the NodeBase part of the Node API.
 class NodeBase : public NodeBaseInterface
@@ -155,24 +170,5 @@ private:
 
 }  // namespace node_interfaces
 }  // namespace rclcpp
-
-// Class to hold the global map of mutexes
-class rclcpp::node_interfaces::map_of_mutexes
-{
-public:
-  map_of_mutexes();
-  ~map_of_mutexes();
-
-  // Methods need to be protected by internal mutex
-  void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
-  std::shared_ptr<std::mutex>
-  get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
-  void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
-
-  // Members
-  std::unordered_map<const rclcpp::node_interfaces::NodeBaseInterface *,
-    std::shared_ptr<std::mutex>> data;
-  std::mutex internal_mutex;
-};
 
 #endif  // RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -15,8 +15,6 @@
 #ifndef RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
 #define RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
 
-#include <atomic>
-#include <functional>
 #include <memory>
 #include <mutex>
 #include <string>
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -33,6 +33,7 @@ namespace rclcpp
 namespace node_interfaces
 {
 
+RCLCPP_PUBLIC
 void global_for_each_callback_group(
   NodeBaseInterface * node_base_interface,
   const NodeBaseInterface::CallbackGroupFunction & func);
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -163,18 +163,19 @@ private:
 // Class to hold the global map of mutexes
 class rclcpp::node_interfaces::map_of_mutexes
 {
-  public:
-    map_of_mutexes();
-    ~map_of_mutexes();
+public:
+  map_of_mutexes();
+  ~map_of_mutexes();
 
-    // Methods need to be protected by internal mutex
-    void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
-    std::shared_ptr<std::mutex> get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
-    void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+  // Methods need to be protected by internal mutex
+  void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase * nodebase);
+  std::shared_ptr<std::mutex>
+  get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase * nodebase);
+  void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase * nodebase);
 
-    // Members
-    std::unordered_map<const rclcpp::node_interfaces::NodeBase*, std::shared_ptr<std::mutex>> data;
-    std::mutex internal_mutex;
+  // Members
+  std::unordered_map<const rclcpp::node_interfaces::NodeBase *, std::shared_ptr<std::mutex>> data;
+  std::mutex internal_mutex;
 };
 
 #endif  // RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -15,26 +15,41 @@
 #ifndef RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
 #define RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
 
+#include <atomic>
+#include <functional>
 #include <memory>
+#include <mutex>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 #include "rcl/node.h"
+#include "rclcpp/callback_group.hpp"
 #include "rclcpp/context.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/visibility_control.hpp"
+
 
 namespace rclcpp
 {
 namespace node_interfaces
 {
 
+class map_of_mutexes;
+
 /// Implementation of the NodeBase part of the Node API.
 class NodeBase : public NodeBaseInterface
 {
 public:
   RCLCPP_SMART_PTR_ALIASES_ONLY(NodeBase)
+
+  static std::unique_ptr<map_of_mutexes> map_object_ptr;
+  static bool map_init_flag;
+  static std::mutex map_init_flag_mutex;
+
+  // Non virtual method
+  void for_each_callback_group(const CallbackGroupFunction & func);
 
   RCLCPP_PUBLIC
   NodeBase(
@@ -144,5 +159,22 @@ private:
 
 }  // namespace node_interfaces
 }  // namespace rclcpp
+
+// Class to hold the global map of mutexes
+class rclcpp::node_interfaces::map_of_mutexes
+{
+  public:
+    map_of_mutexes();
+    ~map_of_mutexes();
+
+    // Methods need to be protected by internal mutex
+    void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+    std::shared_ptr<std::mutex> get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+    void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+
+    // Members
+    std::unordered_map<const rclcpp::node_interfaces::NodeBase*, std::shared_ptr<std::mutex>> data;
+    std::mutex internal_mutex;
+};
 
 #endif  // RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -42,9 +42,7 @@ class NodeBase : public NodeBaseInterface
 public:
   RCLCPP_SMART_PTR_ALIASES_ONLY(NodeBase)
 
-  static std::unique_ptr<map_of_mutexes> map_object_ptr;
-  static bool map_init_flag;
-  static std::mutex map_init_flag_mutex;
+  static map_of_mutexes map_object;
 
   // Non virtual method
   void for_each_callback_group(const CallbackGroupFunction & func);
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -28,7 +28,6 @@
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/visibility_control.hpp"
 
-
 namespace rclcpp
 {
 namespace node_interfaces
@@ -49,9 +48,8 @@ public:
   void delete_mutex_of_nodebase(const NodeBaseInterface * nodebase);
 
 private:
-  // Members
-  std::unordered_map<const NodeBaseInterface *, std::shared_ptr<std::mutex>> data;
-  std::mutex internal_mutex;
+  std::unordered_map<const NodeBaseInterface *, std::shared_ptr<std::mutex>> data_;
+  std::mutex internal_mutex_;
 };
 
 /// Implementation of the NodeBase part of the Node API.
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -193,15 +193,15 @@ Executor::add_callback_groups_from_nodes_associated_to_executor()
     if (node) {
       auto node_base = std::dynamic_pointer_cast<rclcpp::node_interfaces::NodeBase>(node);
       node_base->for_each_callback_group(
-          [this,node](rclcpp::CallbackGroup::SharedPtr shared_group_ptr)
+        [this, node_base](rclcpp::CallbackGroup::SharedPtr shared_group_ptr)
         {
-         if(
-             shared_group_ptr->automatically_add_to_executor_with_node() &&
-             !shared_group_ptr->get_associated_with_executor_atomic().load())
+          if (
+            shared_group_ptr->automatically_add_to_executor_with_node() &&
+            !shared_group_ptr->get_associated_with_executor_atomic().load())
           {
             this->add_callback_group_to_map(
               shared_group_ptr,
-              node,
+              node_base,
               weak_groups_to_nodes_associated_with_executor_,
               true);
           }
@@ -273,20 +273,20 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
   std::lock_guard<std::mutex> guard{mutex_};
   auto node_base = std::dynamic_pointer_cast<rclcpp::node_interfaces::NodeBase>(node_ptr);
   node_base->for_each_callback_group(
-    [this, node_ptr, notify](rclcpp::CallbackGroup::SharedPtr group_ptr)
+    [this, node_base, notify](rclcpp::CallbackGroup::SharedPtr group_ptr)
     {
       if (!group_ptr->get_associated_with_executor_atomic().load() &&
       group_ptr->automatically_add_to_executor_with_node())
       {
         this->add_callback_group_to_map(
           group_ptr,
-          node_ptr,
+          node_base,
           weak_groups_to_nodes_associated_with_executor_,
           notify);
       }
     });
 
-  weak_nodes_.push_back(node_ptr);
+  weak_nodes_.push_back(node_base);
 }
 
 void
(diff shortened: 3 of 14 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -70,6 +70,8 @@
 #include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+typedef std::function<void (rclcpp::CallbackGroup::SharedPtr)> CallbackGroupFunction;
```
clalancette, line 73: We should remove this, and instead rely on the version that lives in node_base.hpp.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -153,6 +155,9 @@ class Node : public std::enable_shared_from_this<Node>
   const std::vector<rclcpp::CallbackGroup::WeakPtr> &
   get_callback_groups() const;
 
+  RCLCPP_PUBLIC
+  void for_each_callback_group(const CallbackGroupFunction & func);
```
clalancette, line 159: We may as well make this one as much like the one on the `master` branch as possible.  That means copying over the style and the docblock as well (the only thing that should be different in the end is the exact namespace that `CallbackGroupFunction` is in).

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1310,9 +1315,9 @@ class Node : public std::enable_shared_from_this<Node>
   const std::string sub_namespace_;
   const std::string effective_namespace_;
 };
-
 }  // namespace rclcpp
 
+
```
clalancette, line 1320: Both of these whitespace changes are unnecessary here.  Just leave them as-is.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -18,13 +18,26 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <unordered_map>
+#include <mutex>
```
clalancette, line 22: Alphabetize these with the ones above.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -18,13 +18,26 @@
...
+#include <iostream>
```
clalancette, line 23: Remove this one, as it isn't needed.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -18,13 +18,26 @@
...
 #include "rclcpp/context.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+#if 1
+#define DPRINT(x) std::cout << "DPRINT: " << x << std::endl;
+#else
+#define DPRINT(x)
+#endif
```
clalancette, line 35: Remove this debug code.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -18,13 +18,26 @@
...
+
+typedef std::function<void (rclcpp::CallbackGroup::SharedPtr)> CallbackGroupFunction;
+
+class map_of_mutexes;
```
clalancette, line 39: These should both be in the namespace.  I would suggest putting them into the `rclcpp::node_interfaces` namespace, but other namespaces under `rclcpp` would work as well.

Also, make sure to `#include <functional>` at the top of this file to get `std::function`.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -145,4 +164,22 @@ class NodeBase : public NodeBaseInterface
...
+
+    // Methods need to be protected by internal mutex
+    void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+    std::shared_ptr<std::mutex> get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+    void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+
+    // Members
+    std::unordered_map<const rclcpp::node_interfaces::NodeBase* , std::shared_ptr<std::mutex> > data;
+    std::mutex internal_mutex;
+
```
clalancette, line 182: Unnecessary blank space.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -145,4 +164,22 @@ class NodeBase : public NodeBaseInterface
...
+    map_of_mutexes();
+    ~map_of_mutexes();
+
+    // Methods need to be protected by internal mutex
+    void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+    std::shared_ptr<std::mutex> get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+    void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBase* nodebase);
+
+    // Members
+    std::unordered_map<const rclcpp::node_interfaces::NodeBase* , std::shared_ptr<std::mutex> > data;
```
clalancette, line 180:
```suggestion
    std::unordered_map<const rclcpp::node_interfaces::NodeBase*, std::shared_ptr<std::mutex>> data;
```

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -268,21 +268,26 @@ Executor::add_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_pt
   // If the node already has an executor
   std::atomic_bool & has_executor = node_ptr->get_associated_with_executor_atomic();
   if (has_executor.exchange(true)) {
-    throw std::runtime_error("Node has already been added to an executor.");
+    throw std::runtime_error(
+            std::string("Node '") + node_ptr->get_fully_qualified_name() +
+            "' has already been added to an executor.");
```
clalancette, line 273: This is an unrelated change, so let's not put it into this PR.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -190,14 +191,13 @@ Executor::add_callback_groups_from_nodes_associated_to_executor()
   for (auto & weak_node : weak_nodes_) {
     auto node = weak_node.lock();
     if (node) {
-      auto group_ptrs = node->get_callback_groups();
-      std::for_each(
-        group_ptrs.begin(), group_ptrs.end(),
-        [this, node](rclcpp::CallbackGroup::WeakPtr group_ptr)
+      auto node_base_type = std::dynamic_pointer_cast<rclcpp::node_interfaces::NodeBase>(node);
```
clalancette, line 194: This is a bit of bikeshedding, but I feel like `node_base` is a better name for this variable.  This is the instantiation of a type, not the type itself.  Same goes a few times below.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -46,6 +46,20 @@ NodeBase::NodeBase(
   associated_with_executor_(false),
   notify_guard_condition_is_valid_(false)
 {
+
+  // Initialize map_of_mutex object if this is the first NodeBase instance
+  if (this->count_of_instances == 0){
+    this->map_object_ptr = std::make_unique<map_of_mutexes>();
+    this->count_of_instances++;
+    DPRINT("NB constructor- Made a new mm object !");
```
clalancette, line 54: Remove this debug print.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -46,6 +46,20 @@ NodeBase::NodeBase(
...
+  }
+  else{
+    DPRINT("NB constructor- mm object already exists")
+  }
```
clalancette, line 58: Remove this debug code.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -46,6 +46,20 @@ NodeBase::NodeBase(
   associated_with_executor_(false),
   notify_guard_condition_is_valid_(false)
 {
+
+  // Initialize map_of_mutex object if this is the first NodeBase instance
+  if (this->count_of_instances == 0){
```
clalancette, line 51:
```suggestion
  if (this->count_of_instances == 0) {
```

clalancette: Also, it's not clear to me that this is actually a count.  That is, it is incremented here, but it is never decremented.  This feels more like it is a boolean flag of "has this been initialized yet?".  There is also the secondary problem of what happens if two nodes are created on two different threads and race here.  It might be better to rely on C++ static initialization for this object, though that has some of its own problems.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -310,3 +327,64 @@ NodeBase::resolve_topic_or_service_name(
...
+
+// Initialize static variables
+std::unique_ptr<map_of_mutexes> NodeBase::map_object_ptr;
+int NodeBase::count_of_instances;
+
+
+//  ***************** Implementation of map_of_mutexes ***************
+map_of_mutexes::map_of_mutexes()
+{
+  DPRINT("mm constructor called")
```
clalancette, line 340: Remove this DPRINT and all of the ones below.

clalancette (changes requested):
I left a preliminary round of review.  I would also suggest rebasing all of the individual commits into a single commit, adding a useful commit message, and signing it properly (which will make the DCO bot happy).

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
adityapande-1995: I'll change that to a bool and add appropriate locking around it.

adityapande-1995:
Changes:
* Squashed commits, signed 
* Removed debug prints, macro
* removed ``typedef`` lines, ``CallbackGroupFunction`` now works the same way as in thr olling commit
* Changed variable name to ``node_base``
* ``map_of_mutexes`` now in ``rclcpp::node_interfaces`` namespace
* ``count_of_instances`` changed to boolean flag, protected by a mutex now
* Added docstring for ``for_each_callback_group`` in ``node.cpp`` and ``lifecycle_node.cpp``
* Removed blank spaces

TODO
* Modify tests
* Make changes to ``static_entities_collector``
* general cleanup, linting fixes, ament, uncrustify checks

adityapande-1995:
TODO :
- CI failures in ```test_node.cpp```
- Change test cases to use ``for_each_callback_group``

adityapande-1995:
TODO: Modify ```test_allocator_memory_strategy.cpp``` test to use ```for_each_callback_group```

rclcpp/include/rclcpp/node.hpp
```diff
@@ -70,6 +70,7 @@
 #include "rclcpp/timer.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+
```
ivanpauno, line 73: do we need this new line?

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -15,27 +15,42 @@
 #ifndef RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
 #define RCLCPP__NODE_INTERFACES__NODE_BASE_HPP_
 
+#include <atomic>
+#include <functional>
```
ivanpauno, line 19: do we need to include this header?

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -15,27 +15,42 @@
...
 
+class map_of_mutexes;
+
 /// Implementation of the NodeBase part of the Node API.
 class NodeBase : public NodeBaseInterface
 {
 public:
   RCLCPP_SMART_PTR_ALIASES_ONLY(NodeBase)
 
+  static std::unique_ptr<map_of_mutexes> map_object_ptr;
```
ivanpauno, line 47: why an unique_ptr?
Can't this just be an `static map_of_mutexes map_object;`

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -15,27 +15,42 @@
...
+  static bool map_init_flag;
+  static std::mutex map_init_flag_mutex;
```
ivanpauno, line 49: do we need to lazely initialize the map?
If yes, you can use std::call_once() and std::once_flag instead of doing it manually.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -605,3 +605,10 @@ Node::get_node_options() const
 {
   return this->node_options_;
 }
+
+void Node::for_each_callback_group(
+  const node_interfaces::NodeBaseInterface::CallbackGroupFunction & func)
+{
+  auto node_base = std::dynamic_pointer_cast<rclcpp::node_interfaces::NodeBase>(node_base_);
```
ivanpauno, line 612: this is pretty unsafe, we're not even checking if the resulting ptr is not null.
We don't even need to add a method to `NodeBaseInterface` though, we can just add a free function that works like a thread safe interface to iterate over `get_callback_groups()`.
Of course, that only works if all callers are using the new interface, but that's also a problem if we add a method instead of a free function.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
ivanpauno: I don't think we need to lazely initialize the map, it seems to me that it can be initialized before main as all other globals.

ivanpauno (commented):
I have a bunch of questions/comments.

rclcpp/include/rclcpp/node.hpp
adityapande-1995: will delete

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
adityapande-1995: will delete

rclcpp/src/rclcpp/node.cpp
adityapande-1995: Will add a check to make sure the pointer is not NULL.

I had originally planned for a free function, but I discussed with @clalancette  and he suggested we make a non-virtual method in ```NodeBase``` and typecast ```NodebaseInterface``` members and then use the ```for_each_callback_group``` to make the code is consistent between rolling and the backport. Effectively the free function would do the same, typecast NodebaseInterface to NodeBase and then implement the for_each routine.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
adityapande-1995: Will do

rclcpp/src/rclcpp/node.cpp
clalancette: > We don't even need to add a method to `NodeBaseInterface` though, we can just add a free function that works like a thread safe interface to iterate over `get_callback_groups()`.

Yeah, as @adityapande-1995 points out, I don't think it matters whether we make it a free function or a non-virtual member function.  Either case preserves ABI; making it a non-virtual member function is slightly nicer in that you get `this` for free (rather than having to pass it in).

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
clalancette: Yeah, I agree.  I think that will be better, and will also get rid of the need for the extra mutex while lazily initializing.

rclcpp/src/rclcpp/node.cpp
ivanpauno: yes, the only difference is that the free function alternative doesn't need dynamic casting, as it can receive a reference to the base class directly.

I don't think we're extending `NodeBaseInterface` in another place, so I don't mind much.

ivanpauno: example:

```cpp
void
for_each_calback_group_thread_safe(NodeBaseInterface & base_interface, FunctionCgT fn) {
  auto & mutex = get_cg_mutex_for_base_interface(&base_interface);
  std::lock_guard guard{mutex};
  for (const auto & cg : base_interface->get_callback_groups()) {
    fn(cg);
  }
}
```

In that case no dynamic cast is needed because [get_callback_groups](https://github.com/ros2/rclcpp/blob/e6733bf2c6cd40c8d8d26ec451042247577f0901/rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp#L131) is already a virtual method.

In that case, we will need the `NodeBaseInterface` constructor to register the mutex though (I think now `NodeBase` is doing that).

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
adityapande-1995: I think that was because ```this->map_object = map_of_mutexes()``` was not allowed as the class member ```std::mutex internal_mutex``` is copyable. To make the map_of_mutex static member ptr to an object, I'll have to change the ```internal_mutex``` as a pointer so that the copy constructor is valid.

clalancette: > I think that was because in `node_base` constructor, `this->map_object = map_of_mutexes()`

But we shouldn't need that at all.  Once we have `map_of_mutexes` in static storage, there is no need to do an assignment to another static within the class; just use the global that already exists.

rclcpp/src/rclcpp/node.cpp
clalancette: Ah, I see.  Yeah, that could work more nicely, leveraging the `get_callback_groups` that has to stay here.  @adityapande-1995 this is definitely worth exploring, as getting rid of the `dynamic_cast` will be a lot safer.

adityapande-1995: Will do

adityapande-1995: Added free function, removed the non virtual one and dynamic pointer casts

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -145,4 +156,23 @@ class NodeBase : public NodeBaseInterface
...
+  void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
+  std::shared_ptr<std::mutex>
+  get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
+  void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
+
+  // Members
+  std::unordered_map<const rclcpp::node_interfaces::NodeBaseInterface *,
+    std::shared_ptr<std::mutex>> data;
+  std::mutex internal_mutex;
+};
```
clalancette, line 176: I'll suggest just moving this class up above the `NodeBase` definition.  That way you can get rid of the forward-reference to it, and also just define it in the proper namespace.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -310,3 +316,51 @@ NodeBase::resolve_topic_or_service_name(
...
+rclcpp::node_interfaces::map_of_mutexes::~map_of_mutexes()
+{
+}
+
+// For each callback group free function implementation
+void rclcpp::node_interfaces::global_for_each_callback_group(
+  NodeBaseInterface * node_base_interface, const NodeBaseInterface::CallbackGroupFunction & func)
+{
+  auto mutex_ptr = NodeBase::map_object.get_mutex_of_nodebase(node_base_interface);
+  if (mutex_ptr) {
```
clalancette, line 356: When can this condition be false?  I think we need to do "something" if this is false, either printing out an error or throwing an exception (depending on how severe the situation is).

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -310,3 +316,51 @@ NodeBase::resolve_topic_or_service_name(
   allocator.deallocate(output_cstr, allocator.state);
   return output;
 }
+
+rclcpp::node_interfaces::map_of_mutexes NodeBase::map_object = map_of_mutexes();
+
+rclcpp::node_interfaces::map_of_mutexes::map_of_mutexes()
+{
+}
```
clalancette, line 324: Since this is the default constructor, we don't need to define this.  Just remove this (and the declaration in the class).

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -310,3 +316,51 @@ NodeBase::resolve_topic_or_service_name(
...
+void rclcpp::node_interfaces::map_of_mutexes::delete_mutex_of_nodebase(
+  const rclcpp::node_interfaces::NodeBaseInterface * nodebase)
+{
+  std::lock_guard<std::mutex> guard(this->internal_mutex);
+  this->data.erase(nodebase);
+}
+
+rclcpp::node_interfaces::map_of_mutexes::~map_of_mutexes()
+{
+}
```
clalancette, line 349: Same here; just remove the default destructor.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -145,4 +156,23 @@ class NodeBase : public NodeBaseInterface
 }  // namespace node_interfaces
 }  // namespace rclcpp
 
+// Class to hold the global map of mutexes
+class rclcpp::node_interfaces::map_of_mutexes
```
clalancette, line 160: Nit: this can be `final`.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -145,4 +156,23 @@ class NodeBase : public NodeBaseInterface
...
+  // Methods need to be protected by internal mutex
+  void create_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
+  std::shared_ptr<std::mutex>
+  get_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
+  void delete_mutex_of_nodebase(const rclcpp::node_interfaces::NodeBaseInterface * nodebase);
+
+  // Members
+  std::unordered_map<const rclcpp::node_interfaces::NodeBaseInterface *,
+    std::shared_ptr<std::mutex>> data;
+  std::mutex internal_mutex;
```
clalancette, line 175: I think these internal members can be `private`, no?

clalancette (changes requested):
Overall, I think this is getting close.  I've left a few more things to think about.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
ivanpauno: +1
I would also always stick to the rule of zero or rule of five, and never to the rule of three or any other intermediate solution.
See https://en.cppreference.com/w/cpp/language/rule_of_three#the_rule_of_three_five_zero#rule_of_zero, https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
adityapande-1995: This is probably not necessary, as the mutex is registered when a new ```NodeBase``` is created and deleted only when the ```NodeBase``` object is destroyed, so this should always return a non-NULL pointer. I've added a ```RCLCPP_ERROR```, though I think this check can be removed.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -16,26 +16,52 @@
...
 #include <vector>
 
 #include "rcl/node.h"
+#include "rclcpp/callback_group.hpp"
 #include "rclcpp/context.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+
```
clalancette, line 31: Nit: unnecessary whitespace change.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -16,26 +16,52 @@
...
+{
+public:
+  // Methods need to be protected by internal mutex
+  void create_mutex_of_nodebase(const NodeBaseInterface * nodebase);
+  std::shared_ptr<std::mutex>
+  get_mutex_of_nodebase(const NodeBaseInterface * nodebase);
+  void delete_mutex_of_nodebase(const NodeBaseInterface * nodebase);
+
+private:
+  // Members
```
clalancette, line 52: Nit: no need for this comment, it is pretty obvious.

rclcpp/include/rclcpp/node_interfaces/node_base.hpp
```diff
@@ -16,26 +16,52 @@
...
+  std::unordered_map<const NodeBaseInterface *, std::shared_ptr<std::mutex>> data;
+  std::mutex internal_mutex;
```
clalancette, line 54: Nit: all of our member variables end with `_`, so this should be:

```suggestion
  std::unordered_map<const NodeBaseInterface *, std::shared_ptr<std::mutex>> data_;
  std::mutex internal_mutex_;
```

(obviously you'll have to update the code as well)

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
clalancette: I'd say let's remove the check, then.

clalancette (changes requested):
I have a bunch of nits inline.  Once those are all fixed, I think this is good to go from my perspective, though I'd like to get @ivanpauno 's final thoughts before we run CI on this.

clalancette (approved):
This looks good to me with three things:

1. Green CI.
2. Running `abi-compliance-checker` before and after this PR to make sure we aren't changing ABI.
3. An approval from @ivanpauno and @cottsay

adityapande-1995:
CI Jobs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15059/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9792/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12745/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15226/)

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -46,6 +48,9 @@ NodeBase::NodeBase(
   associated_with_executor_(false),
   notify_guard_condition_is_valid_(false)
 {
+  // Generate a mutex for this instance of NodeBase
+  this->map_object.create_mutex_of_nodebase(this);
```
cottsay, line 52: Is it standard practice to refer to static members using the instance reference like this? It seems misleading to me.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -89,6 +89,7 @@
 #include "rclcpp_lifecycle/transition.hpp"
 #include "rclcpp_lifecycle/visibility_control.h"
 
+
```
cottsay, line 92: Any particular reason for this spurious newline?

cottsay (approved):
Looks good from an ABI compatibility perspective. I had a couple of other comments inline, but nothing critical.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
adityapande-1995: Changed to use ```NodeBase::``` instead of ```this->```

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
adityapande-1995: Removed

adityapande-1995:
CI Jobs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15063/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9795/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12748/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15229/)

adityapande-1995:
@cottsay How should we handle the unstable windows CI ? The failures are in ```rosbag2_transport``` which doesn't seem to be related to this PR.

cottsay:
> How should we handle the unstable windows CI ? The failures are in rosbag2_transport which don't seem to be related to this PR.

If you're confident that they aren't a regression caused by this PR, then I think it's OK to ignore the test failures. It might be worth triggering a baseline build to be sure, though.

adityapande-1995:
CI on raw galactic:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15072/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9804/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12757/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15239/)

ivanpauno (approved):
LGTM!

ivanpauno:
> @cottsay How should we handle the unstable windows CI ? The failures are in rosbag2_transport which don't seem to be related to this PR.

I can't tell if those failures are related or not, +1 to compare with a baseline build.

adityapande-1995:
> CI on galactic branch of rclcpp:
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/15072/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9804/)
> * macOS [Build Status](http://ci.ros2.org/job/ci_osx/12757/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/15239/)

So this CI was run by changing the branch of rclcpp to ```galactic```, i.e without my changes and it produced the same error in Windows. @cottsay @ivanpauno

ivanpauno:
> So this CI was run by changing the branch of rclcpp to galactic, i.e without my changes and it produced the same error in Windows. @cottsay @ivanpauno

LGTM!

cottsay (approved):
Thanks for double-checking!

adityapande-1995:
ABI check report

image

Issue #1743, opened 2021-08-04
Function template version of declare_parameter does not work with uint32_t and uint64_t


Bi0T1N:
## Bug report

**Required Info:**

- Operating System:
  - Linux Mint 20.1 (based on Ubuntu 20.04)
- Installation type:
  - ROS 2 Galactic Geochelone Debian packages for Ubuntu Focal
- Version or commit hash:
  - Latest Galactic Geochelone
- DDS implementation:
  - default one shipped with the Debian package -> Cyclone DDS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
The minimal (non)working example is based on the tutorial for [Using parameters in a class (C++)](https://docs.ros.org/en/galactic/Tutorials/Using-Parameters-In-A-Class-CPP.html):
```c++
#include <rclcpp/rclcpp.hpp>
#include <chrono>
#include <string>
#include <functional>

using namespace std::chrono_literals;

class ParametersClass: public rclcpp::Node
{
  public:
    ParametersClass()
      : Node("parameter_node")
    {
      this->declare_parameter<int8_t>("this_int_works1", 1);
      this->declare_parameter<int16_t>("this_int_works2", 1);
      this->declare_parameter<int32_t>("this_int_works3", 1);
      this->declare_parameter<int64_t>("this_int_works4", 1);

      this->declare_parameter<uint8_t>("this_uint_works1", 1);
      this->declare_parameter<uint16_t>("this_uint_works1", 1);
      this->declare_parameter<uint32_t>("this_uint_does_not_work1", 1);
      this->declare_parameter<uint64_t>("this_uint_does_not_work2", 1);
    }
};

int main(int argc, char** argv)
{
  rclcpp::init(argc, argv);
  rclcpp::spin(std::make_shared<ParametersClass>());
  rclcpp::shutdown();
  return 0;
}
```

```cmake
cmake_minimum_required(VERSION 3.8)
project(cpp_parameters)

if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wall -Wextra -Wpedantic)
endif()

# find dependencies
find_package(ament_cmake REQUIRED)
find_package(rclcpp REQUIRED)

add_executable(parameter_node src/cpp_parameters_node.cpp)
ament_target_dependencies(parameter_node rclcpp)

install(TARGETS
  parameter_node
[...]

PR #1744, opened 2021-08-04
Add tests for fixed-width integer function templates of declare_parameter

Bi0T1N:
this PR is related to #1743 but does not compile at the moment as discussed in the issue.

### DIFF
--- a/rclcpp/test/rclcpp/test_node.cpp
+++ b/rclcpp/test/rclcpp/test_node.cpp
@@ -2589,6 +2589,49 @@ TEST_F(TestNode, declare_parameter_with_vector) {
   }
 }
 
+// test declare parameter templates with fixed length integers
+TEST_F(TestNode, declare_parameter_fixed_len_integer_templates) {
+  auto node = std::make_shared<rclcpp::Node>(
+    "test_declare_parameter_fixed_len_integer_templates"_unq,
+    rclcpp::NodeOptions().allow_undeclared_parameters(true));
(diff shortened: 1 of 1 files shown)


rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -2589,6 +2589,49 @@ TEST_F(TestNode, declare_parameter_with_vector) {
...
+    auto name8 = "parameter"_unq;
+
+    node->declare_parameter<int8_t>(name1, 5);
+    node->declare_parameter<uint8_t>(name2, 5);
+    node->declare_parameter<int16_t>(name3, 5);
+    node->declare_parameter<uint16_t>(name4, 5);
+    node->declare_parameter<int32_t>(name5, 5);
+    node->declare_parameter<uint32_t>(name6, 5);
+    node->declare_parameter<int64_t>(name7, 5);
+    node->declare_parameter<uint64_t>(name8, 5);
```
wjwwood, line 2615: As I mentioned in the other issue, I don't think this case can/should work.

wjwwood:
I'm not sure what the state of this pull request is, if it is eclipsed by https://github.com/ros2/rclcpp/pull/1747/ or if it should be closed with https://github.com/ros2/rclcpp/pull/1759, or if part of this pull request could/should be salvaged, but the `uint64_t` instance should be removed.

For now, I'll close and we can reopen it if need be.

Issue #1749, opened 2021-08-17
Parameter redeclaration


livanov93:
## Bug report

**Required Info:**
`automatically_declare_parameters_from_overrides` [method](https://github.com/ros2/rclcpp/blob/1037822a63330495dcf0de5e8f20544375a5f116/rclcpp/include/rclcpp/node_options.hpp#L357) on the `NodeOptions` redeclares already declared paramters despite its description:

```
   * If true, automatically iterate through the node's parameter overrides and
   * implicitly declare any that have not already been declared.
   * Otherwise, parameters passed to the node's parameter_overrides, and/or the
   * global arguments (e.g. parameter overrides from a YAML file), which are
   * not explicitly declared will not appear on the node at all, even if
   * `allow_undeclared_parameters` is true.
   * Already declared parameters will not be re-declared, and parameters
   * declared in this way will use the default constructed ParameterDescriptor.
   
```

- Operating System:
  - Ubuntu 20.04.2 LTS
- Installation type:
  - binaries
- Version or commit hash:
  - foxy
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

Test package can be found [here](https://github.com/livanov93/test_params_pkg). Put the package in your ~/ros2_ws/src and build it from ~/ros2_ws/
```
colcon build --packages-select test_params_pkg
source install/setup.bash
ros2 run test_params_pkg test_params_node --ros-args -p declared_parameter:="Totally"

```

#### Expected behavior
```
declared_param: Totally
```
#### Actual behavior
```
terminate called after throwing an instance of 'rclcpp::exceptions::ParameterAlreadyDeclaredException'
  what():  parameter 'declared_parameter' has already been declared
```

wjwwood:
I think that if you trace the source of the exception, it will be this line:

https://github.com/livanov93/test_params_pkg/blob/6c856e28ec9b69850517172c5b25572f5e515d60/src/test_params_node.cpp#L12

So, it's actually the user code that is re-declaring the parameter, not this setting. But you have a point that it might not be possible to declare a parameter before the logic for `automatically_declare_parameters_from_overrides` manages to declare them.

So we might want to adjust the documentation, but this is expected behavior as far as I can see here.

You can see where the logic for `automatically_declare_parameters_from_overrides` avoids re-declaring here:

[...]

PR #1751, opened 2021-08-19
Improve documentation for auto declaration from overrides

livanov93:
PR makes clear what the `automatically_declare_parameters_from_overrides` flag does and emphasizes that the user should check if parameter already exists on the potential parameter declaration.

PR tackles #1749.

### DIFF
--- a/rclcpp/include/rclcpp/node_options.hpp
+++ b/rclcpp/include/rclcpp/node_options.hpp
@@ -350,7 +350,8 @@ public:
    * not explicitly declared will not appear on the node at all, even if
    * `allow_undeclared_parameters` is true.
    * Parameter declaration from overrides is done in the node's base constructor,
-   * so user takes care if some parameter is already (e.g. automatically) declared.
+   * so the user must take care to check if the parameter is already (e.g.
+   * automatically) declared before declaring it themselves.
    * Already declared parameters will not be re-declared, and parameters
    * declared in this way will use the default constructed ParameterDescriptor.
    */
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -349,6 +349,8 @@ class NodeOptions
    * global arguments (e.g. parameter overrides from a YAML file), which are
    * not explicitly declared will not appear on the node at all, even if
    * `allow_undeclared_parameters` is true.
+   * Parameter declaration from overrides is done in the node's base construcor,
```
fujitatomoya, line 352: nit,

```suggestion
   * Parameter declaration from overrides is done in the node's base constructor,
```

fujitatomoya (changes requested):

rclcpp/include/rclcpp/node_options.hpp
```diff
@@ -349,6 +349,8 @@ class NodeOptions
    * global arguments (e.g. parameter overrides from a YAML file), which are
    * not explicitly declared will not appear on the node at all, even if
    * `allow_undeclared_parameters` is true.
+   * Parameter declaration from overrides is done in the node's base constructor,
+   * so user takes care if some parameter is already (e.g. automatically) declared.
```
wjwwood, line 353:
```suggestion
   * Parameter declaration from overrides is done in the node's base constructor,
   * so the user must take care to check if the parameter is already (e.g.
   * automatically) declared before declaring it themselves.
```

fujitatomoya (approved):

wjwwood (approved):

PR #1754, opened 2021-08-24
Fix lifetime of context so it remains alive while its dependent node handles are still in use

mxgrey:
Currently it is possible for a `rclcpp::Context` to destruct while dependent objects (e.g. `rclcpp::Publisher`, `rclcpp::Subscription`) are still alive. This can lead to segmentation faults, exceptions, or other problems in the RMW layer.

This PR fixes that problem by tying the lifetime of contexts to the lifetimes of the node handles that depend on them. Since the objects that depend on nodes (like Publisher and Subscriber) hold a counted reference to their node handle, we just need to ensure that contexts remain alive as long as any of their dependent node handles are alive. This PR accomplishes that by distributing `std::shared_ptr<rcl_node_t>` instances that are actually an alias for a `std::shared_ptr<NodeHandleWithContext>` which bundles together a `std::shared_ptr<rclcpp::Context>` with a `rcl_node_t*`, and whose destructor will delete the `rcl_node_t*`.

By using an aliased `std::shared_ptr`, we get the benefit of the corrected lifetime without any modification to API or even ABI of `rclcpp`. It should be possible to squash this PR and then cherry-pick the commit onto other branches of rclcpp without any harmful effects.

I've also added a unit test that will result in a seg fault if it gets run without the fix provided by this PR.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -34,17 +34,23 @@ using rclcpp::node_interfaces::NodeBase;
 
 namespace
 {
-/// This class bundles together the lifetime of the rcl_node_t handle with the
-/// lifetime of the RCL context. This ensures that the context will remain alive
-/// for as long as the rcl_node_t handle is alive.
+/// A class to manage the lifetime of a node handle and its context
+/**
+ * This class bundles together the lifetime of the rcl_node_t handle with the
+ * lifetime of the RCL context. This ensures that the context will remain alive
+ * for as long as the rcl_node_t handle is alive.
+ */
 class NodeHandleWithContext
 {
 public:
-  /// The make function returns a std::shared_ptr<rcl_node_t> which is actually
-  /// an alias for a std::shared_ptr<NodeHandleWithContext>. There is no use for
-  /// accessing a NodeHandleWithContext instance directly, because its only job
-  /// is to couple together the lifetime of a context with the lifetime of a
-  /// node handle that depends on it.
+  /// Make an instance of a NodeHandleWithContext
+  /**
+   * The make function returns a std::shared_ptr<rcl_node_t> which is actually
+   * an alias for a std::shared_ptr<NodeHandleWithContext>. There is no use for
+   * accessing a NodeHandleWithContext instance directly, because its only job
+   * is to couple together the lifetime of a context with the lifetime of a
+   * node handle that depends on it.
+   */
   static std::shared_ptr<rcl_node_t>
   make(
     rclcpp::Context::SharedPtr context,
@@ -82,8 +88,9 @@ public:
   }
 
 private:
-  /// The constructor is private because this class is only meant to be aliased
-  /// into a std::shared_ptr<rcl_node_t>.
+  // The constructor is private because this class is only meant to be aliased
+  // into a std::shared_ptr<rcl_node_t> and should not be constructed any other
+  // way.
   NodeHandleWithContext(
     rclcpp::Context::SharedPtr context,
     std::shared_ptr<std::recursive_mutex> logging_mutex,
@@ -91,9 +98,7 @@ private:
   : context_(std::move(context)),
     logging_mutex_(std::move(logging_mutex)),
     node_handle_(node_handle)
-  {
-    // Do nothing
-  }
+  {}
 
   rclcpp::Context::SharedPtr context_;
   std::shared_ptr<std::recursive_mutex> logging_mutex_;
@@ -201,8 +206,7 @@ NodeBase::NodeBase(
     throw_from_rcl_error(ret, "failed to initialize rcl node");
   }
 
-  node_handle_ = NodeHandleWithContext::make(
-    context_, logging_mutex, rcl_node.release());
+  node_handle_ = NodeHandleWithContext::make(context_, logging_mutex, rcl_node.release());
 
   // Create the default callback group.
   using rclcpp::CallbackGroupType;
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -73,17 +73,16 @@ public:
   NodeHandleWithContext & operator=(NodeHandleWithContext &&) = delete;
 
   NodeHandleWithContext(NodeHandleWithContext && other)
-    : context_(std::move(other.context_)),
-      logging_mutex_(std::move(other.logging_mutex_)),
-      node_handle_(other.node_handle_)
+  : context_(std::move(other.context_)),
+    logging_mutex_(std::move(other.logging_mutex_)),
+    node_handle_(other.node_handle_)
   {
     other.node_handle_ = nullptr;
   }
 
   ~NodeHandleWithContext()
   {
-    if (!node_handle_)
-    {
+    if (!node_handle_) {
       // If the node_handle_ is null, then this object was moved-from. We don't
       // need to do any cleanup.
       return;
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -57,8 +57,8 @@ public:
     std::shared_ptr<std::recursive_mutex> logging_mutex,
     rcl_node_t * node_handle)
   {
-    auto aliased_ptr = std::shared_ptr<NodeHandleWithContext>(
-      new NodeHandleWithContext(
+    auto aliased_ptr = std::make_shared<NodeHandleWithContext>(
+      NodeHandleWithContext(
         std::move(context),
         std::move(logging_mutex),
         node_handle));
@@ -66,15 +66,29 @@ public:
     return std::shared_ptr<rcl_node_t>(std::move(aliased_ptr), node_handle);
   }
 
-  // This class should not be copied or moved. It should only exist in the
+  // This class should not be copied. It should only exist in the
   // std::shared_ptr that it was originally provided in.
   NodeHandleWithContext(const NodeHandleWithContext &) = delete;
-  NodeHandleWithContext(NodeHandleWithContext &&) = delete;
   NodeHandleWithContext & operator=(const NodeHandleWithContext &) = delete;
   NodeHandleWithContext & operator=(NodeHandleWithContext &&) = delete;
 
+  NodeHandleWithContext(NodeHandleWithContext && other)
+    : context_(std::move(other.context_)),
+      logging_mutex_(std::move(other.logging_mutex_)),
+      node_handle_(other.node_handle_)
+  {
+    other.node_handle_ = nullptr;
+  }
+
   ~NodeHandleWithContext()
   {
+    if (!node_handle_)
+    {
+      // If the node_handle_ is null, then this object was moved-from. We don't
+      // need to do any cleanup.
+      return;
+    }
+
     std::lock_guard<std::recursive_mutex> guard(*logging_mutex_);
     // TODO(ivanpauno): Instead of mutually excluding rcl_node_fini with the global logger mutex,
     // rcl_logging_rosout_fini_publisher_for_node could be decoupled from there and be called
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -60,6 +60,13 @@ public:
     return std::shared_ptr<rcl_node_t>(std::move(aliased_ptr), node_handle);
   }
 
+  // This class should not be copied or moved. It should only exist in the
+  // std::shared_ptr that it was originally provided in.
+  NodeHandleWithContext(const NodeHandleWithContext&) = delete;
+  NodeHandleWithContext(NodeHandleWithContext&&) = delete;
+  NodeHandleWithContext& operator=(const NodeHandleWithContext&) = delete;
+  NodeHandleWithContext& operator=(NodeHandleWithContext&&) = delete;
+
   ~NodeHandleWithContext()
   {
     std::lock_guard<std::recursive_mutex> guard(*logging_mutex_);
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -62,10 +62,10 @@ public:
 
   // This class should not be copied or moved. It should only exist in the
   // std::shared_ptr that it was originally provided in.
-  NodeHandleWithContext(const NodeHandleWithContext&) = delete;
-  NodeHandleWithContext(NodeHandleWithContext&&) = delete;
-  NodeHandleWithContext& operator=(const NodeHandleWithContext&) = delete;
-  NodeHandleWithContext& operator=(NodeHandleWithContext&&) = delete;
+  NodeHandleWithContext(const NodeHandleWithContext &) = delete;
+  NodeHandleWithContext(NodeHandleWithContext &&) = delete;
+  NodeHandleWithContext& operator=(const NodeHandleWithContext &) = delete;
+  NodeHandleWithContext& operator=(NodeHandleWithContext &&) = delete;
 
   ~NodeHandleWithContext()
   {
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -64,8 +64,8 @@ public:
   // std::shared_ptr that it was originally provided in.
   NodeHandleWithContext(const NodeHandleWithContext &) = delete;
   NodeHandleWithContext(NodeHandleWithContext &&) = delete;
-  NodeHandleWithContext& operator=(const NodeHandleWithContext &) = delete;
-  NodeHandleWithContext& operator=(NodeHandleWithContext &&) = delete;
+  NodeHandleWithContext & operator=(const NodeHandleWithContext &) = delete;
+  NodeHandleWithContext & operator=(NodeHandleWithContext &&) = delete;
 
   ~NodeHandleWithContext()
   {
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -15,6 +15,7 @@
 #include <string>
 #include <limits>
 #include <memory>
+#include <utility>
 #include <vector>
 
 #include "rclcpp/node_interfaces/node_base.hpp"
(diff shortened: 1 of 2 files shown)


fujitatomoya (changes requested):
this fix seems reasonable to me. could you address linter failure? https://build.ros2.org/job/Gpr__rclcpp__ubuntu_focal_amd64/27/

fujitatomoya:
CC: @ivanpauno

mxgrey:
@fujitatomoya thanks for pointing that out; the linter looks happy now!

clalancette:
@mxgrey Also, can you please retarget this fix to the master branch?  Once we get it on there, we can then backport it to Galactic and Foxy as appropriate.  Thanks.

mxgrey:
To target `master`, I had to force push a cherry-pick because of a multitude of conflicts, but I think it should be okay now.

ivanpauno:
The patch seems reasonable, but it's really not clear how we're expecting the ownership relationships to be.
Should a `rclcpp::Node` retain a `rclcpp::Context` or only a `rcl_context_t`? Should `rclcpp::NodeBase` be doing that?
Should a `rclcpp::Publisher` retain a `rcl_node_t`, a `rclcpp::NodeBase` or a `rclcpp::Node`?

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -31,6 +32,68 @@ using rclcpp::exceptions::throw_from_rcl_error;
...
+    auto aliased_ptr = std::shared_ptr<NodeHandleWithContext>(
+      new NodeHandleWithContext(
+        std::move(context),
+        std::move(logging_mutex),
+        node_handle));
+
+    return std::shared_ptr<rcl_node_t>(std::move(aliased_ptr), node_handle);
+  }
+
+  ~NodeHandleWithContext()
```
ivanpauno, line 63: Can you `= delete` copy assignment/constructors and `= default` move assignment/constructors  to comply with the rule of 5?
(it doesn't seem safe to have a copy constructor/assignment in this case)

mxgrey:
> it's really not clear how we're expecting the ownership relationships to be.

My rationale here is that anything which uses a living `rcl_node_t` definitely needs its corresponding `rcl_context_t` to be alive, and the lifetime of the `rcl_context_t` is managed by the `rclcpp::Context`. So taking those two facts together, as long as the `rcl_node_t` is alive, we should ensure that the `rclcpp::Context` that manages its `rcl_context_t` stays alive. We can most easily do that by coupling the `rclcpp::Context::SharedPtr` with the `std::shared_ptr<rcl_node_t>`.

My thought is that if we focus on managing the most primitive, direct level of rcl's dependencies, then the lifetimes should come out correctly. Although I guess by that logic we could perhaps have the `std::shared_ptr<rcl_node_t>` be coupled with the `std::shared_ptr<rcl_context_t>` instead of `rclcpp::Context::SharedPtr`. But I don't see much point in dwelling too much on that since (1) The current PR successfully fixes the problem that is known to exist, and (2) We can trivially switch that implementation detail in a future PR without impacting API or ABI.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mxgrey: The `NodeHandleWithContext` class should be inaccessible in general since it's hidden behind an aliased `std::shared_ptr`, but you're right that we should delete them anyway to make the intention of the class absolutely clear. Done here: c454258

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -31,6 +32,75 @@ using rclcpp::exceptions::throw_from_rcl_error;
 
 using rclcpp::node_interfaces::NodeBase;
 
+namespace
+{
+/// This class bundles together the lifetime of the rcl_node_t handle with the
+/// lifetime of the RCL context. This ensures that the context will remain alive
+/// for as long as the rcl_node_t handle is alive.
```
wjwwood, line 39: If these comments are meant to be doxygen, please use this style which is prevalent throughout the project:

```
/// Brief, single-line description.
/**
 * Everything else...
 */
```

If not, please just use `//`.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -31,6 +32,75 @@ using rclcpp::exceptions::throw_from_rcl_error;
...
+class NodeHandleWithContext
+{
+public:
+  /// The make function returns a std::shared_ptr<rcl_node_t> which is actually
+  /// an alias for a std::shared_ptr<NodeHandleWithContext>. There is no use for
+  /// accessing a NodeHandleWithContext instance directly, because its only job
+  /// is to couple together the lifetime of a context with the lifetime of a
+  /// node handle that depends on it.
```
wjwwood, line 47: Same here, and anywhere else.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -31,6 +32,75 @@ using rclcpp::exceptions::throw_from_rcl_error;
...
+  NodeHandleWithContext(
+    rclcpp::Context::SharedPtr context,
+    std::shared_ptr<std::recursive_mutex> logging_mutex,
+    rcl_node_t * node_handle)
+  : context_(std::move(context)),
+    logging_mutex_(std::move(logging_mutex)),
+    node_handle_(node_handle)
+  {
+    // Do nothing
+  }
```
wjwwood, line 96: nitpick: just simplify this:

```suggestion
  {}
```

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -31,6 +32,75 @@ using rclcpp::exceptions::throw_from_rcl_error;
...
+  make(
+    rclcpp::Context::SharedPtr context,
+    std::shared_ptr<std::recursive_mutex> logging_mutex,
+    rcl_node_t * node_handle)
+  {
+    auto aliased_ptr = std::shared_ptr<NodeHandleWithContext>(
+      new NodeHandleWithContext(
+        std::move(context),
+        std::move(logging_mutex),
+        node_handle));
```
wjwwood, line 58: Can you use `std::make_shared` here? I think that would be better, unless the aliased pointer requires two allocations as is done here.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -131,20 +201,8 @@ NodeBase::NodeBase(
...
-      // here directly.
-      if (rcl_node_fini(node) != RCL_RET_OK) {
-        RCUTILS_LOG_ERROR_NAMED(
-          "rclcpp",
-          "Error in destruction of rcl node handle: %s", rcl_get_error_string().str);
-      }
-      delete node;
-    });
+  node_handle_ = NodeHandleWithContext::make(
+    context_, logging_mutex, rcl_node.release());
```
wjwwood, line 205: nitpick:

```suggestion
  node_handle_ = NodeHandleWithContext::make(context_, logging_mutex, rcl_node.release());
```

wjwwood (approved):
A few nitpicks, but otherwise lgtm.

As for what the ownership relationships should be, I think the rclcpp objects cannot back-reference one another because that might cause circular references. Often the rclcpp Context refers to rclcpp Node and so on down to rclcpp pub/sub objects, this is needed for various rclcpp specific tasks.

Separately, we want to provide a "safe" interface that protects the user from use-after-free like errors due to destroying rclcpp objects in the wrong order (i.e. not in the reverse order of construction), even though we document you should not do this. To accomplish this, having a back-reference through the underlying rcl objects seems good, since they do not keep track of one another otherwise, but allows the rclcpp objects to go out of scope until the dependent rclcpp objects are done with them. Which is exactly what this pr does for context and node.

So I think rclcpp should depend down (context->node->pub/sub/etc) if at all, and to provide safety the rcl should depend up (pub/sub/etc->node->context), and this pull request is in that direction.

wjwwood:
I'll wait to address test until comments have been replied to.

fujitatomoya (approved):
lgtm

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mxgrey: We can't use `std::make_shared` here because `NodeHandleWithContext` has a private constructor, and that unfortunately prohibits `std::make_shared` from being able to call it. If you try to side-step that by making `std::make_shared` a friend function of the class, then suddenly anyone can construct a `NodeHandleWithContext` anywhere simply by calling `std::make_shared<NodeHandleWithContext>`.

The aliasing pointer doesn't require the two allocations (I assume you're referring to [1] the allocation of the reference-counting block and [2] the allocation of the `NodeHandleWithContext` instance), it's just a matter of whether we want the code to self-document the way this class is meant to be used. Personally I think ensuring correctness is worth the very small additional allocation, but I suppose it's ultimately a subjective value judgment.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mxgrey: Fixed in c5ff0f4

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mxgrey: Fixed in c5ff0f4

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mxgrey: Fixed in c5ff0f4

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mxgrey: Fixed in c5ff0f4

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mxgrey: As a side note, one method I sometimes use to side-step this limitation is to use move semantics, like this:

```
ClassWithPrivateConstructor my_instance;
return std::make_shared<ClassWithPrivateConstructor>(std::move(my_instance));
```

That can work if your class is movable, but we're explicitly disabling movability here because we have a destructor that does some special RAII work, and so we'd need to make a custom move constructor or else we could end up double-deleting.

ivanpauno: > We can't use std::make_shared here because NodeHandleWithContext has a private constructor, and that unfortunately prohibits std::make_shared from being able to call it

`NodeHandleWithContext` is declared and defined in a .cpp file and there's no reference to it in any header AFAIS, it doesn't seem to be a problem to make the constructor public IMO.
We can also mark the constructor with the `RCLCPP_LOCAL` attribute, so the symbol is hidden.

mxgrey: When possible I like to put constraints even inside implementations, as reminders to the old maintainers and clues to the new maintainers about how an implementation detail is meant to work. I don't mind making a custom move constructor for this class if it's considered important to have a single allocation.

mxgrey: With c8716ac we are now doing a single allocation for the `std::shared_ptr`.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15276/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9990/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12957/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15467/)

mxgrey:
@fujitatomoya , I'm not super familiar with the ROS buildfarm setup, but the only error I can find for the `ci_linux` appears to be this one:

```
run_ros2_batch.py: error: unrecognized arguments: ...
```

It looks to me like the `CI_ROS2_REPOS_URL` variable is meant to be a URL to a `.repos` file, but it's been filled in with the literal YAML text that's desired. The CI scripts don't seem to be set up for that type of usage.

@wjwwood can jump in to correct me if I'm wrong, but I'm guessing the way it's meant to be used is to upload the desired `.repos` file to a publicly accessible URL and then fill that parameter with the URL.

mxgrey:
Also I'm not sure if the Windows CI is actually using the branch of this PR. I don't see any mention of my repo in the console log.

clalancette:
Yeah, that repos URL is meant to be a URL, not a literal pasting of the repositories.  The fact that Windows ran at all with that configuration is actually a bug :).

fujitatomoya:
@mxgrey sorry, my bad. just re-started CI.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15278/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9992/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12959/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15470/)

fujitatomoya:
all green, waiting for the final check on https://github.com/ros2/rclcpp/pull/1754#discussion_r707765908. @wjwwood @ivanpauno

wjwwood (approved):
lgtm

sorry for dropping the ball on this review.

I probably would have just made the constructor public too, but this move constructor approach is fine as well.

iuhilnehc-ynos:
I know it is a fix without ABI breaking, but I don't like the separation of `rcl_node_init` and `rcl_node_fini` in two different locations in this PR. I just wondered if we could use the following fix,

```cpp
diff --git a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
index 8a19a6c9..d4b1236b 100644
--- a/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_base.hpp
@@ -34,7 +34,7 @@ namespace node_interfaces
 {
 
 /// Implementation of the NodeBase part of the Node API.
-class NodeBase : public NodeBaseInterface
+class NodeBase : public NodeBaseInterface, public std::enable_shared_from_this<NodeBase>
 {
 public:
   RCLCPP_SMART_PTR_ALIASES_ONLY(NodeBase)
diff --git a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
index 5240c8c8..19f3addf 100644
--- a/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_base.cpp
@@ -207,13 +207,13 @@ NodeBase::get_rcl_node_handle() const
 std::shared_ptr<rcl_node_t>
 NodeBase::get_shared_rcl_node_handle()
 {
-  return node_handle_;
+  return std::shared_ptr<rcl_node_t>(shared_from_this(), node_handle_.get());
 }
 
 std::shared_ptr<const rcl_node_t>
 NodeBase::get_shared_rcl_node_handle() const
 {
-  return node_handle_;
+  return std::shared_ptr<rcl_node_t>(shared_from_this(), node_handle_.get());
 }
 
```

mxgrey:
@iuhilnehc-ynos The change that you're recommending would result in circular `std::shared_ptr` dependencies for most ROS users.

The common paradigm for writing a ROS node is to inherit the `Node` class and then create publishers/subscribers/services/clients as members of your derived class. Those members are stored as `std::shared_ptr<T>`. If we give publishers, subscribers, etc, a shared reference to the base class of the node that is storing them then they end up with a circular lifetime dependency and none of them will ever be deleted.

I assume that's the reason those classes were given a reference to the underlying node handle rather than a reference to the node itself in the first place.

iuhilnehc-ynos:
@mxgrey

Thanks for your response. 

> The common paradigm for writing a ROS node is to inherit the `Node` class and then create publishers/subscribers/services/clients as members of your derived class. Those members are stored as `std::shared_ptr<T>`. 

Let's use the `talker` demo,
https://github.com/ros2/demos/blob/0cfdbe9d5c0f1a52629ff2348a243932bb4a949f/demo_nodes_cpp/src/topics/talker.cpp

> If we give publishers, subscribers, etc, a shared reference to the base class of the node that is storing them then they end up with a circular lifetime dependency and none of them will ever be deleted.

Could you tell me how to make the circular dependency? I don't know what kind of cases you told me about.
```cpp
node_ptr_ = this->get_node_base_interface()->get_shared_rcl_node_handle();   // or whatelse ?

std::shared_ptr<rcl_node_t> node_ptr_;
```

mxgrey:
Ah, I just realized that the `Node` class _contains_ a `NodeBase`; it doesn't _inherit_ a `NodeBase` which is an assumption I jumped to based on the class names. Now that I realize that, I take back what I said; there wouldn't be a circular dependency.

fujitatomoya:
@iuhilnehc-ynos i think https://github.com/ros2/rclcpp/pull/1754#issuecomment-963852021 sounds better for me, and we can solve the problem at the same time.
@ivanpauno @wjwwood @clalancette what do you think? are we missing something here?

ivanpauno:
> @iuhilnehc-ynos i think #1754 (comment) sounds better for me, and we can solve the problem at the same time.
@ivanpauno @wjwwood @clalancette what do you think? are we missing something here?

That patch seems reasonable and a bit more clean, I don't mind either way though.

fujitatomoya:
@iuhilnehc-ynos could you make PR based on your idea?

iuhilnehc-ynos:
@fujitatomoya 

[The patch](https://github.com/ros2/rclcpp/pull/1754#issuecomment-963852021) failed to pass some test cases,

```shell

[ RUN      ] TestMemoryStrategy.get_node_by_group
/home/chenlh/Projects/ROS2/ros2-master/src/ros2/rclcpp/rclcpp/test/rclcpp/test_memory_strategy.cpp:312: Failure
  nullptr
    Which is: (nullptr)
  memory_strategy()->get_node_by_group(callback_group, weak_groups_to_nodes)
    Which is: 16-byte object <30-2D C3-12 74-55 00-00 40-D4 C5-12 74-55 00-00>
[  FAILED  ] TestMemoryStrategy.get_node_by_group (11 ms)
[ RUN      ] TestMemoryStrategy.get_group_by_subscription
/home/chenlh/Projects/ROS2/ros2-master/src/ros2/rclcpp/rclcpp/test/rclcpp/test_memory_strategy.cpp:366: Failure
  nullptr
    Which is: (nullptr)
  memory_strategy()->get_group_by_subscription(subscription, weak_groups_to_nodes)
    Which is: 16-byte object <10-AD C7-12 74-55 00-00 00-AD C7-12 74-55 00-00>
[  FAILED  ] TestMemoryStrategy.get_group_by_subscription (10 ms)
```

I need to check them. If these failures can be fixed, I'll create a new PR.
NOTE: **I apologize if it's a problem with my patch.**

wjwwood:
That approach is fine with me as long as the tests pass. Off-hand I don't see why those failures are occurring with your patch @iuhilnehc-ynos.

iuhilnehc-ynos:
@wjwwood 

Thank you for noticing it.

@fujitatomoya 

IMO, there exists another potential issue(Besides subscription[[1](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L320), [2](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L346-L347), [3](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L361)],  other entities ([service](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L387-L405), [client](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L431-L444), [timer](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L470-L483), and [waitable](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L510-L521)) don't hold the callback_group  ).
Anyway, I'll create a PR to fix it and also use [the above patch](https://github.com/ros2/rclcpp/pull/1754#issuecomment-963852021).

Issue #1746, opened 2021-08-09
rclcpp_lifecycle::LifecycleNode::get_current_state is not thread safe


igorbanfi:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.4
- Installation type:
  - binaries
- Version or commit hash:
  - foxy
- DDS implementation:
  - fastDDS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

In code below get_current_state is called in 2 threads and after random number of iterations ``` Error in state! Internal state_handle is NULL. ``` is thrown when calling id() state.

The problem disappears if only one thread calls get_current_state function.

```
#include <chrono>
#include <iostream>
#include <memory>
#include <thread>
#include "lifecycle_msgs/msg/transition.hpp"
#include "rclcpp/rclcpp.hpp"
#include "rclcpp/publisher.hpp"
#include "rclcpp_lifecycle/lifecycle_node.hpp"
#include "rcutils/logging_macros.h"

/// Get current node state in a loop
void checkState(std::shared_ptr<rclcpp_lifecycle::LifecycleNode> node) {

   // infinite loop
   while (1) {
      // This call fails
      node->get_current_state().id();
   }
}

int main(int argc, char * argv[])
{
   setvbuf(stdout, NULL, _IONBF, BUFSIZ);

   rclcpp::init(argc, argv);

   std::shared_ptr<rclcpp_lifecycle::LifecycleNode> lc_node =
           std::make_shared<rclcpp_lifecycle::LifecycleNode>("lc_talker");

   lc_node->trigger_transition(lifecycle_msgs::msg::Transition::TRANSITION_CONFIGURE);

   // Created multiple threads
   std::thread thread_object_1(checkState, lc_node);
   std::thread thread_object_2(checkState, lc_node);

   // wait
   thread_object_1.join();

   rclcpp::shutdown();

   return 0;
}
```

devrite:
This is true. Current states queries from impl, sets to local var and returns it.

https://github.com/ros2/rclcpp/blob/b918bd4c25c2ec59bf6ef58eff04612902eb9330/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp#L335-L340

Solution (if it is safe to query the internal state from multiple threads): Do not set the local var, just return the constructed value.

fujitatomoya:
This is a bug, confirmed that core crash happens on mainline.

> Do not set the local var, just return the constructed value.

[...]

PR #1756, opened 2021-08-25
Bugfix 20210810 get current state

fujitatomoya:
fix https://github.com/ros2/rclcpp/issues/1746

### DIFF
--- a/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
+++ b/rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
@@ -143,6 +143,7 @@ private:
   node_interfaces::LifecycleNodeInterface::CallbackReturn
   execute_callback(unsigned int cb_id, const State & previous_state) const;
 
+  mutable std::recursive_mutex state_machine_mutex_;
   rcl_lifecycle_state_machine_t state_machine_;
   State current_state_;
   std::map<
--- a/rclcpp_lifecycle/src/state.cpp
+++ b/rclcpp_lifecycle/src/state.cpp
@@ -71,6 +71,7 @@ State::State(
   if (!rcl_lifecycle_state_handle) {
     throw std::runtime_error("rcl_lifecycle_state_handle is null");
   }
+
   state_handle_ = const_cast<rcl_lifecycle_state_t *>(rcl_lifecycle_state_handle);
 }
 
@@ -94,6 +95,8 @@ State::operator=(const State & rhs)
     return *this;
   }
 
+  // hold the lock until state_handle_ is reconstructed
+  std::lock_guard<std::recursive_mutex> lock(state_handle_mutex_);
   // reset all currently used resources
   reset();
 
@@ -129,6 +132,7 @@ State::operator=(const State & rhs)
 uint8_t
 State::id() const
 {
+  std::lock_guard<std::recursive_mutex> lock(state_handle_mutex_);
   if (!state_handle_) {
     throw std::runtime_error("Error in state! Internal state_handle is NULL.");
   }
@@ -138,6 +142,7 @@ State::id() const
 std::string
 State::label() const
 {
+  std::lock_guard<std::recursive_mutex> lock(state_handle_mutex_);
   if (!state_handle_) {
     throw std::runtime_error("Error in state! Internal state_handle is NULL.");
   }
@@ -147,6 +152,7 @@ State::label() const
 void
 State::reset() noexcept
 {
+  std::lock_guard<std::recursive_mutex> lock(state_handle_mutex_);
   if (!owns_rcl_state_handle_) {
     state_handle_ = nullptr;
   }
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp
@@ -15,6 +15,7 @@
 #ifndef RCLCPP_LIFECYCLE__STATE_HPP_
 #define RCLCPP_LIFECYCLE__STATE_HPP_
 
+#include <mutex>
 #include <string>
 
 #include "rcl_lifecycle/data_types.h"
@@ -92,6 +93,7 @@ protected:
 
   bool owns_rcl_state_handle_;
 
+  mutable std::recursive_mutex state_handle_mutex_;
   rcl_lifecycle_state_t * state_handle_;
 };
 
--- a/rclcpp_lifecycle/src/lifecycle_node_interface_impl.cpp
+++ b/rclcpp_lifecycle/src/lifecycle_node_interface_impl.cpp
@@ -14,6 +14,7 @@
 
 #include <functional>
 #include <memory>
+#include <mutex>
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -58,7 +59,11 @@ LifecycleNode::LifecycleNodeInterfaceImpl::LifecycleNodeInterfaceImpl(
 LifecycleNode::LifecycleNodeInterfaceImpl::~LifecycleNodeInterfaceImpl()
 {
   rcl_node_t * node_handle = node_base_interface_->get_rcl_node_handle();
-  auto ret = rcl_lifecycle_state_machine_fini(&state_machine_, node_handle);
+  rcl_ret_t ret;
+  {
+    std::lock_guard<std::recursive_mutex> lock(state_machine_mutex_);
+    ret = rcl_lifecycle_state_machine_fini(&state_machine_, node_handle);
+  }
   if (ret != RCL_RET_OK) {
     RCUTILS_LOG_FATAL_NAMED(
       "rclcpp_lifecycle",
@@ -72,7 +77,6 @@ LifecycleNode::LifecycleNodeInterfaceImpl::init(bool enable_communication_interf
   rcl_node_t * node_handle = node_base_interface_->get_rcl_node_handle();
   const rcl_node_options_t * node_options =
     rcl_node_get_options(node_base_interface_->get_rcl_node_handle());
-  state_machine_ = rcl_lifecycle_get_zero_initialized_state_machine();
   auto state_machine_options = rcl_lifecycle_get_default_state_machine_options();
   state_machine_options.enable_com_interface = enable_communication_interface;
   state_machine_options.allocator = node_options->allocator;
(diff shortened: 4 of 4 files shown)


rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -335,6 +365,7 @@ class LifecycleNode::LifecycleNodeInterfaceImpl
   const State &
   get_current_state()
   {
+    std::lock_guard<std::recursive_mutex> lock(state_machine_mutex_);
     current_state_ = State(state_machine_.current_state);
     return current_state_;
```
clalancette, line 370: As pointed out in the initial issue, we could also make this even safer by not storing the current_state and just using the lock.  That is, change this code to:

```
    std::lock_guard<std::recursive_mutex> lock(state_machine_mutex_);
    return State(state_machine_.current_state);
```

(and remove the `current_state_` member variable)

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -411,27 +451,32 @@ class LifecycleNode::LifecycleNodeInterfaceImpl
...
 
     // error handling ?!
     // TODO(karsten1987): iterate over possible ret value
     if (cb_return_code == node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR) {
       RCUTILS_LOG_WARN("Error occurred while doing error handling.");
 
-      auto error_cb_code = execute_callback(state_machine_.current_state->id, initial_state);
+      auto error_cb_code = execute_callback(current_state_id, initial_state);
       auto error_cb_label = get_label_for_return_code(error_cb_code);
+      std::lock_guard<std::recursive_mutex> lock(state_machine_mutex_);
```
clalancette, line 479: I think it would be better to keep holding the lock across the above and down into this code.  This minimizes the number of times we take and drop the lock in this function.

rclcpp_lifecycle/src/state.cpp
```diff
@@ -101,25 +108,29 @@ State::operator=(const State & rhs)
 
   // we don't own the handle, so we can return straight ahead
   if (!owns_rcl_state_handle_) {
+    std::lock_guard<std::recursive_mutex> lock(state_handle_mutex_);
     state_handle_ = rhs.state_handle_;
     return *this;
   }
 
   // we own the handle, so we have to deep-copy the rhs object
```
clalancette, line 116: I would actually refactor this to be:

```
  std::lock_guard<std::recursive_mutex> lock(state_handle_mutex_);
  if (!owns_rcl_state_handle_) {
    state_handle_ = rhs.state_handle_;
  } else {
     // do the deep-copy stuff here
  }
  return *this;
```

clalancette (changes requested):

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
Barry-Xu-2018: @fujitatomoya 

About this problem, I discussed with iuhilnehc-ynos last Friday.  If we want to resolve this completely, the current idea cannot avoid introducing thread and queue. But this make code more complex (may introduce new problem). Another way is to keep this solution, and we need to remind users to avoid changing the node state in different threads.

iuhilnehc-ynos: > for example,
> 
> 1. thread-1 calls `trigger_transition` and hold `state_machine_mutex_` internally.
> 2. context switch to thread-2 holds user-mutex, and call `trigger_transition`.
> 3. thread-2 is waiting on `state_machine_mutex_`.
> 4. context switch to thread-1, and user callback requires user-mutex.
> 5. deadlock

Sorry, I noticed that I made a mistake in https://github.com/ros2/rclcpp/pull/1756#discussion_r738962746 which still leads to a deadlock by following this example.

Back to this example, I think it seems to be the user's responsibility not to do these actions.
There is no reason to hold user-mutex to call `trigger_transition`  on step 2 if the user wants to hold the user-mutex inside user callback.

I am inclined to keep the original lock. Let's hear others' ideas.

fujitatomoya:
@clalancette @ivanpauno what do you think about this fix? there is still discussion needs to be done for future, but this can fix  https://github.com/ros2/rclcpp/issues/1746 (avoid crash and exception.)

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
ivanpauno: > i think it should return the pointer to current state, unless it does not really indicate the current state? cz it might be changed right after that... thinking i would like to keep the current implementation.

The issue is that the returned reference can be invalidated anytime, and an external caller doesn't have acces to the `state_machine_mutex_`.
I agree with @clalancette, returning a copy will avoid the race condition altogether.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -89,23 +92,29 @@ class LifecycleNode::LifecycleNodeInterfaceImpl
...
-      ROSIDL_GET_MSG_TYPE_SUPPORT(lifecycle_msgs, msg, TransitionEvent),
-      rosidl_typesupport_cpp::get_service_type_support_handle<ChangeStateSrv>(),
-      rosidl_typesupport_cpp::get_service_type_support_handle<GetStateSrv>(),
-      rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableStatesSrv>(),
-      rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableTransitionsSrv>(),
-      rosidl_typesupport_cpp::get_service_type_support_handle<GetAvailableTransitionsSrv>(),
-      &state_machine_options);
+    rcl_ret_t ret;
+    {
+      std::lock_guard<std::recursive_mutex> lock(state_machine_mutex_);
```
clalancette, line 97: Do we really need this scope for this lock?  It looks like we basically need to hold the lock until the end of the method, so I would suggest just taking the lock here and not dropping it until we leave the method.  That also means that re-taking the lock in `if (enable_communication_interface)` is no longer necessary.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
```diff
@@ -248,6 +260,7 @@ class LifecycleNode::LifecycleNodeInterfaceImpl
   {
     (void)header;
     (void)req;
+    std::lock_guard<std::recursive_mutex> lock(state_machine_mutex_);
     if (rcl_lifecycle_state_machine_is_initialized(&state_machine_) != RCL_RET_OK) {
       throw std::runtime_error(
               "Can't get state. State machine is not initialized.");
```
clalancette, line 266: I know you didn't add this, but this seems like a serious problem.  An outside process can call this service and potentially crash this node, which doesn't seem ideal.  I would say that we shouldn't fix it here, but we should make an issue to fix it later.  Would you mind doing that?

clalancette (changes requested):
I've left a few more comments, and we still have open the question on what to do with `get_current_state`.  But this is getting closer.

fujitatomoya:
@clalancette thanks, i will look into comments and get back to you.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
fujitatomoya: I mentioned before, i was hesitating to do this since this affects user interfaces for `LifecycleNode`, most of operation methods return `State &`. returning constructed value as a copy can be done, but just checking if we can actually do this now. or any good solution or idea to keep the user interface?

https://github.com/ros2/rclcpp/blob/152dbc6e38e18c18a5883b51c33162d237f891e0/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp#L789-L909

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
fujitatomoya: yeah, basically i agree with the idea you mentioned. but i think error state is much better than deadlock. and we cannot know how user application is developed. with all these possible conditions and we already have the similar code in action, i still think we need to release the lock during user callback.

https://github.com/ros2/rclcpp/blob/152dbc6e38e18c18a5883b51c33162d237f891e0/rclcpp_action/src/server.cpp#L411

fujitatomoya:
I am still requesting some help for a couple of questions. any comments will do good!

fujitatomoya:
@clalancette @iuhilnehc-ynos friendly ping.

iuhilnehc-ynos:
@fujitatomoya 

I don't know if I should continue to review it because returning a copy of State[[1](https://github.com/ros2/rclcpp/pull/1756/files#r708303467),[2](https://github.com/ros2/rclcpp/pull/1756/files#r740427063)] seems to be the consensus. 
To fix an issue with the correct way rather than a workaround might be better. I know it could affect many source codes or even other repositories.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
ivanpauno: I think you can change those methods to return `State` by value as well.
It's probably not going to break any external code.

You'll probably need though to add a copy constructor to `State`.
The `State` class internally stores a rcl handle https://github.com/ros2/rclcpp/blob/1688f05aead3968bcea6cab27b346260f633f0b0/rclcpp_lifecycle/include/rclcpp_lifecycle/state.hpp#L94
so maybe that needs mutual exclusion internally as well (?)
The fact that we're using shallow copies might be a bit of a problem.

rclcpp_lifecycle/src/lifecycle_node_interface_impl.hpp
clalancette: The more I look at `get_current_state`, the less sense it makes to me.  It really should be a `const` method, but it can't be because it modifies internal state of the class.  I think we should address that problem first before we go forward here.

Towards that end, I've opened up https://github.com/ros2/rclcpp/pull/2027 .  This doesn't actually fix the problem, but does a number of cleanups in this class and should make it easier for us to make that method const in a followup.

fujitatomoya: @clalancette thanks, i will look at another PR, then we can get back to this if necessary.

clalancette (approved):
With all of the latest fixes, I've now rebased this onto the latest.  I think this looks much better now and should fix the original problem.  That said, since I did substantial work on this, I'd like @fujitatomoya to take another look at this before we go ahead with it.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17532/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12066/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18085/)

fujitatomoya:
@clalancette thanks i will take a look at this soon.

roncapat:
@fujitatomoya is this backportable to humble?

fujitatomoya:
Unfortunately I do not think so, this breaks ABI so cannot land on humble.

PR #1764, opened 2021-09-02
Make sure to check for empty extension before accessing front.

clalancette:
The documentation for std::string::front() says that calling it
on an empty string is undefined behavior.  It actually seems to
work on all platforms except Windows Debug, where it throws an
error.  Make sure to check for empty first.

We also notice that there is no reason to check the
existing_sub_namespace for empty; the rest of the code handles
that fine.  So remove that check.

Finally, we add an anonymous namespace so that these functions do
not pollute the global namespace.

This should fix the Windows Debug test failure as seen in https://ci.ros2.org/view/nightly/job/nightly_win_deb/2099/testReport/junit/projectroot.test/rclcpp/test_node/

### DIFF
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -47,6 +47,9 @@ using rclcpp::Node;
 using rclcpp::NodeOptions;
 using rclcpp::exceptions::throw_from_rcl_error;
 
+namespace
+{
+
 RCLCPP_LOCAL
 std::string
 extend_sub_namespace(const std::string & existing_sub_namespace, const std::string & extension)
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15152/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9882/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12843/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15332/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/15333/)

rclcpp/src/rclcpp/node.cpp
```diff
@@ -47,25 +47,28 @@ using rclcpp::Node;
...
 RCLCPP_LOCAL
 std::string
 extend_sub_namespace(const std::string & existing_sub_namespace, const std::string & extension)
 {
   // Assumption is that the existing_sub_namespace does not need checking
   // because it would be checked already when it was set with this function.
 
-  // check if the new sub-namespace extension is absolute
-  if (extension.front() == '/') {
+  if (extension.empty()) {
```
fujitatomoya, line 60: getting rid of `existing_sub_namespace.empty()` actually changes the behavior between former and current, but i think it should be okay.

fujitatomoya (approved):
as you mentioned, `std::string::front` shall not be called on empty strings. LGTM.

rclcpp/src/rclcpp/node.cpp
clalancette: Yeah, that is true.  It changes it slightly, but I don't think in a way that most users will notice.

Previous to this PR there are 4 cases:

| | empty existing_sub_namespace | not empty existing namespace |
--- | --- | ---
| empty extension | throw exception |  new_subnamespace = existingnamespace (arguably buggy) |
| not empty extension | new subnamespace = extension | new subnamespace = existingnamespace + extension |

After this PR, the 4 cases look like:

| | empty existing_sub_namespace | not empty existing namespace |
--- | --- | ---
| empty extension | throw exception |  throw exception |
| not empty extension | new subnamespace = extension | new subnamespace = existingnamespace + extension |

clalancette:
The Windows Debug failure is unrelated to this code, and is being worked on separately in https://github.com/ros-perception/laser_geometry/pull/82

PR #1770, opened 2021-09-10
Fix dangerous std::bind capture in ParameterEventHandler implementation

gbiggs:
The implementation of the `ParameterEventHandler` object, in `parameter_event_handler.cpp`, has an instance of a `std::bind` capture that is dangerous. On line 178, the event subscription callback is set up like this:

```
  event_subscription_ = rclcpp::create_subscription<rcl_interfaces::msg::ParameterEvent>(
      node_topics, "/parameter_events", qos,
      std::bind(&ParameterEventHandler::event_callback, this, std::placeholders::_1));
```

This could lead to a race condition if the destruction of the `ParameterEventHandler` is interleaved with processing of a parameter event in another thread.

This PR fixes the potential race condition by moving the state that needs to be accessed in the `event_callback` function, and the function itself, into a private `struct`. The `ParameterEventHandler` class holds a shared pointer to this struct, and the `std::bind` is replaced with a lambda function that captures a copy of that shared pointer. This means that as long as the lambda exists, the lifetime of the state data used by the callback is protected.

For additional context on why the race condition could occur, see [this presentation](https://docs.google.com/presentation/d/1FPUOh85col-KyczN2fpxafwcvUJYT1bnStcrxYS5KFQ/edit#slide=id.p) from @mxgrey. Specifically [the slides from this point](https://docs.google.com/presentation/d/1FPUOh85col-KyczN2fpxafwcvUJYT1bnStcrxYS5KFQ/edit#slide=id.ge5a9a1248e_0_189) describe how to use lambda and bind captures safely.

Note that a lambda of the form `[this]() { ... }` is equivalent to `std::bind(..., this, ...)`. This is the pattern found in `ParameterEventHandler`.


ivanpauno (approved):
lgtm!

ivanpauno:
@gbiggs could you rebase this PR as well?

ivanpauno:
Check the uncrustify issues as well: https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/1368/testReport/junit/rclcpp/uncrustify/src_rclcpp_parameter_event_handler_cpp/.

gbiggs:
@ivanpauno I've fixed the formatting error and rebased. Can you please run CI?

Issue #1704, opened 2021-06-30
Shutdown cleanly on SIGTERM


emersonknapp:
#### Feature description

The `rclcpp` Context shuts down gracefully on a SIGINT, which is generally received via a Ctrl+C interaction. However, another reasonable way on Linux to try to shut down processes is via a SIGTERM - according to [GNU Documentation](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html):

> The SIGTERM signal is a generic signal used to cause program termination. Unlike SIGKILL, this signal can be blocked, handled, and ignored. It is the normal way to politely ask a program to terminate.

It seems reasonable to me to have `rclcpp` generate a clean shutdown on SIGTERM by default. This could be another `InitOptions` member like `shutdown_on_sigint`, defaulting to `true`.

#### Implementation considerations

Should be pretty easy, install signal handlers the same way we do for SIGINT.

wjwwood:
Seems reasonable to me, but it is a significant change in behavior which might affect how things with with say launch right now. So it would need to be thoroughly tested.

@ros2/team for feedback.

emersonknapp:
>  significant change in behavior

We could also consider default to false, in which case the default behavior would not change. Though I think I'd prefer to default to true.

clalancette:
:+1: to handling SIGTERM as well, though I agree with William that it needs thorough testing.  Messing with the signal handlers is always fragile.

PR #1771, opened 2021-09-10
Handle sigterm, in the same way we handle sigint

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1704.

Done:
- Cleanup previous code.
- Refactor code to be able to handle more than one signal type.
  This could be improved though, maybe storing a map from signal numbers to old signal handlers and using the same handler for all signals (?).
- Handle sigterm.

TODO:

- [ ] Add option, so users can decide if the want to install the rclcpp default sigint handler, sigterm handler, or both.
  Currently it's only possible to install both or nothing.
- [ ] Discuss what's the desired default: installing only a sigint handler (previous behavior), or both sigint and sigterm (current behavior of this PR).

### DIFF
--- a/rclcpp/src/rclcpp/signal_handler.cpp
+++ b/rclcpp/src/rclcpp/signal_handler.cpp
@@ -57,7 +57,7 @@ set_signal_handler(
     char error_string[1024];
     rcutils_strerror(error_string, sizeof(error_string));
     auto msg =
-      "Failed to set SIGINT signal handler (" + std::to_string(errno) + "): " + error_string;
+      "Failed to set signal handler (" + std::to_string(errno) + "): " + error_string;
     throw std::runtime_error(msg);
   }
   return old_signal_handler;
@@ -281,7 +281,7 @@ SignalHandler::signal_handler_common()
   instance.signal_received_.store(true);
   RCLCPP_DEBUG(
     get_logger(),
-    "signal_handler(): SIGINT received, notifying deferred signal handler");
+    "signal_handler(): notifying deferred signal handler");
   instance.notify_signal_handler();
 }
 
@@ -290,7 +290,7 @@ SignalHandler::deferred_signal_handler()
 {
   while (true) {
     if (signal_received_.exchange(false)) {
-      RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): SIGINT received, shutting down");
+      RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): shutting down");
       for (auto context_ptr : rclcpp::get_contexts()) {
         if (context_ptr->get_init_options().shutdown_on_sigint) {
           RCLCPP_DEBUG(
@@ -306,9 +306,9 @@ SignalHandler::deferred_signal_handler()
       RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): signal handling uninstalled");
       break;
     }
-    RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): waiting for SIGINT or uninstall");
+    RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): waiting for SIGINT/SIGTERM or uninstall");
     wait_for_signal();
-    RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): woken up due to SIGINT or uninstall");
+    RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): woken up due to SIGINT/SIGTERM or uninstall");
   }
 }
 
--- a/rclcpp/src/rclcpp/signal_handler.cpp
+++ b/rclcpp/src/rclcpp/signal_handler.cpp
@@ -35,6 +35,114 @@
 
 using rclcpp::SignalHandler;
 
+using rclcpp::detail::OneSignalHandler;
+
+
+static
+OneSignalHandler::signal_handler_type
+set_signal_handler(
+  int signal_value,
+  const OneSignalHandler::signal_handler_type & signal_handler)
+{
+  bool signal_handler_install_failed;
+  OneSignalHandler::signal_handler_type old_signal_handler;
+#if defined(RCLCPP_HAS_SIGACTION)
+  ssize_t ret = sigaction(signal_value, &signal_handler, &old_signal_handler);
+  signal_handler_install_failed = (ret == -1);
+#else
+  old_signal_handler = std::signal(signal_value, signal_handler);
+  signal_handler_install_failed = (old_signal_handler == SIG_ERR);
+#endif
+  if (signal_handler_install_failed) {
+    char error_string[1024];
+    rcutils_strerror(error_string, sizeof(error_string));
+    auto msg =
+      "Failed to set SIGINT signal handler (" + std::to_string(errno) + "): " + error_string;
+    throw std::runtime_error(msg);
+  }
+  return old_signal_handler;
+}
+
+
+OneSignalHandler::OneSignalHandler(int signum)
+: signum_{signum},
+  installed_{false}
+{}
+
+OneSignalHandler::~OneSignalHandler()
+{
+  this->uninstall();
+}
+
+bool
+OneSignalHandler::install(const OneSignalHandler::signal_handler_type & signal_handler)
+{
+  if (installed_) {
+    return false;
+  }
+  old_signal_handler_ = set_signal_handler(signum_, signal_handler);
+  installed_ = true;
+  return true;
+}
+
+bool
+OneSignalHandler::uninstall()
+{
+  if (!installed_) {
+    return false;
+  }
+  set_signal_handler(signum_, old_signal_handler_);
+  installed_ = false;
+  return true;
+}
+
+// Unfortunately macros (or duplicated code) are needed here,
+// as the signal handler must be a function pointer.
+#if defined(RCLCPP_HAS_SIGACTION)
+#define DEFINE_SIGNAL_HANDLER(SIGNAL_NAME_LOWER_CASE) \
+  void \
+  SignalHandler::signal_handler_ ## SIGNAL_NAME_LOWER_CASE( \
+    int signal_value, siginfo_t * siginfo, void * context) \
+  { \
+    RCLCPP_INFO(SignalHandler::get_logger(), "signal_handler(signal_value=%d)", signal_value); \
+ \
+    auto & instance = SignalHandler::get_global_signal_handler(); \
+ \
+    auto old_signal_handler = instance.get_ ## SIGNAL_NAME_LOWER_CASE ## _old_handler(); \
+    if (old_signal_handler.sa_flags & SA_SIGINFO) { \
+      if (old_signal_handler.sa_sigaction != NULL) { \
+        old_signal_handler.sa_sigaction(signal_value, siginfo, context); \
+      } \
+    } else { \
+      if ( \
+        old_signal_handler.sa_handler != NULL &&  /* Is set */ \
+        old_signal_handler.sa_handler != SIG_DFL &&  /* Is not default*/ \
+        old_signal_handler.sa_handler != SIG_IGN)  /* Is not ignored */ \
+      { \
+        old_signal_handler.sa_handler(signal_value); \
+      } \
+    } \
+ \
+    instance.signal_handler_common(); \
+  }
+#else
+#define DEFINE_SIGNAL_HANDLER(SIGNAL_NAME_LOWER_CASE) \
+  void \
+  SignalHandler::signal_handler_ ## SIGNAL_NAME_LOWER_CASE(int signal_value) \
+  { \
+    RCLCPP_INFO(SignalHandler::get_logger(), "signal_handler(signal_value=%d)", signal_value); \
+    auto & instance = SignalHandler::get_global_signal_handler(); \
+ \
+    auto old_signal_handler = instance.get_ ## SIGNAL_NAME_LOWER_CASE ## _old_handler(); \
+    if (old_signal_handler) { \
+      old_signal_handler(signal_value); \
+    } \
+ \
+    instance.signal_handler_common(); \
+  }
+#endif
+
+
 rclcpp::Logger &
 SignalHandler::get_logger()
 {
@@ -48,6 +156,21 @@ SignalHandler::get_global_signal_handler()
   return signal_handler;
 }
 
+#if defined(RCLCPP_HAS_SIGACTION)
+static
+void
+initialize_handler_argument(
+  void (* handler)(
+    int signal_value, siginfo_t * siginfo, void * context),
+  OneSignalHandler::signal_handler_type * out)
+{
+  memset(out, 0, sizeof(*out));
+  sigemptyset(&out->sa_mask);
+  out->sa_sigaction = handler;
+  out->sa_flags = SA_SIGINFO;
+}
+#endif
+
 bool
 SignalHandler::install()
 {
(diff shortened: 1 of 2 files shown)


ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15193/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9913/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12873/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15374/)

ros-discourse:

https://discourse.ros.org/t/change-in-rclcpp-rclpy-shutting-down-gracefully-on-sigterm/22333/1

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -161,82 +205,41 @@ SignalHandler::~SignalHandler()
...
-
 void
 SignalHandler::deferred_signal_handler()
 {
   while (true) {
     if (signal_received_.exchange(false)) {
-      RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): SIGINT received, shutting down");
+      RCLCPP_DEBUG(get_logger(), "deferred_signal_handler(): shutting down");
       for (auto context_ptr : rclcpp::get_contexts()) {
         if (context_ptr->get_init_options().shutdown_on_sigint) {
```
hidmic, line 244: @ivanpauno meta: do we need a `shutdown_on_sigterm` option?

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -53,7 +53,11 @@ namespace rclcpp
  */
 RCLCPP_PUBLIC
 void
-init(int argc, char const * const argv[], const InitOptions & init_options = InitOptions());
+init(
+  int argc,
+  char const * const argv[],
+  const InitOptions & init_options = InitOptions(),
+  bool install_sigterm_handler = true);
```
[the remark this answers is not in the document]
hidmic: Also, I wonder if we should even install a handler if the corresponding `shutdown_on_<signal>`  flag is false.

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -161,82 +205,41 @@ SignalHandler::~SignalHandler()
...
-
-  signal_handler_common();
-}
-#endif
-
 void
 SignalHandler::deferred_signal_handler()
 {
   while (true) {
     if (signal_received_.exchange(false)) {
```
hidmic, line 241: @ivanpauno meta: we could probably forward the signal number through an atomic integer instead.

hidmic (commented):
First pass.

sloretz:
Could this change hide shutdown issues where nodes fail to shut down in `launch_ros` tests? In [this `nightly_linux_repeated` job](https://ci.ros2.org/job/nightly_linux_repeated/2416) a test node `node_with_name_1-2` fails to terminate after 5 seconds of receving SIGINT. `launch` escalates that to SIGTERM, but the signal number being a disallowed exit code that causes the on-exit call to fail.

```
70: [ERROR] [node_with_name_1-2]: process[node_with_name_1-2] failed to terminate '5' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
```

```
70: ======================================================================
70: FAIL: test_10_nodes (test_rclcpp.TestNNodesAfterShutdown)
70: ----------------------------------------------------------------------
70: Traceback (most recent call last):
70:   File "/home/jenkins-agent/workspace/nightly_linux_repeated/ws/build/test_rclcpp/test_n_nodes__rmw_fastrtps_cpp_Release.py", line 61, in test_10_nodes
70:     launch_testing.asserts.assertExitCodes(
70:   File "/home/jenkins-agent/workspace/nightly_linux_repeated/ws/install/launch_testing/lib/python3.8/site-packages/launch_testing/asserts/assert_exit_codes.py", line 62, in assertExitCodes
70:     assert info.returncode in allowable_exit_codes, 'Proc {} exited with code {}'.format(
70: AssertionError: Proc node_with_name_1-2 exited with code -15
```

If handling SIGTERM means the exit code becomes 0 instead of -15, then does that mean failure to terminate on SIGINT would go unnoticed? OR, does it mean the escalation to SIGTERM by `launch_ros` would likely have no effect?

rclcpp/src/rclcpp/signal_handler.cpp
ivanpauno: you could get two signals before the deferred signal handler starts running, so it really doesn't help much.

rclcpp/include/rclcpp/utilities.hpp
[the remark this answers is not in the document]
ivanpauno: > @ivanpauno meta: shall we make this an init option (and put it in the InitOptions struct)?

`Context.init()` doesn't initialize signal handlers, while `rclcpp::init()` does.
Thus, I added an extra argument to init.

rclcpp/src/rclcpp/signal_handler.cpp
ivanpauno: Maybe, I don't mind either way.

ivanpauno:
> If handling SIGTERM means the exit code becomes 0 instead of -15, then does that mean failure to terminate on SIGINT would go unnoticed? OR, does it mean the escalation to SIGTERM by launch_ros would likely have no effect?

I would bet that the escalation to SIGTERM will have no effect, though I don't really know.
launch finally escalates to SIGKILL, so it wouldn't really be a problem in that case.

rclcpp/src/rclcpp/signal_handler.cpp
hidmic: Either that or rename `shutdown_on_sigint` to `shutdown_on_signal` if having two booleans is unnecessary.

rclcpp/src/rclcpp/signal_handler.cpp
hidmic: Fair point. In which case differentiated handling is not possible. Then, `shutdown_on_sigint` --> `shutdown_on_signal`.

rclcpp/include/rclcpp/utilities.hpp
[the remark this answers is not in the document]
hidmic: Good point.

rclcpp/src/rclcpp/signal_handler.cpp
ivanpauno: Renaming sounds fine to me, though it's an unnecessary breaking change.

rclcpp/src/rclcpp/signal_handler.hpp
```diff
@@ -84,35 +84,36 @@ class SignalHandler final
...
 
 #if defined(RCLCPP_HAS_SIGACTION)
-  /// Signal handler function.
+  /// Signal handler function for SIGINT.
   static
   void
   signal_handler(int signal_value, siginfo_t * siginfo, void * context);
 #else
-  /// Signal handler function.
+  /// Signal handler function for SIGINT.
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rclcpp/pull/1771/commits/b7d2cb5821b609ee6c41f8e1eaa06610cd3d3ef5

rclcpp/include/rclcpp/context.hpp
```diff
@@ -268,7 +268,7 @@ class Context : public std::enable_shared_from_this<Context>
    *
    *   - this context is shutdown()
    *   - this context is destructed (resulting in shutdown)
-   *   - this context has shutdown_on_sigint=true and SIGINT occurs (resulting in shutdown)
+   *   - this context has shutdown_on_signal=true and SIGINT occurs (resulting in shutdown)
```
hidmic, line 271: @ivanpauno nit: 
```suggestion
   *   - this context has shutdown_on_signal=true and SIGINT/SIGTERM occurs (resulting in shutdown)
```

rclcpp/src/rclcpp/signal_handler.cpp
hidmic: >  though it's an unnecessary breaking change.

I disagree, but perhaps others can chime in. @wjwwood ?

hidmic (commented):
Overall LGTM

hidmic (approved):
LGTM pending green CI (after latest commits)

rclcpp/src/rclcpp/signal_handler.cpp
wjwwood: Hmm, I don't mind, we could also alias them and deprecate the old one through a tick-tock cycle. I agree that "just changing it" it probably unnecessarily breaking.

rclcpp/include/rclcpp/utilities.hpp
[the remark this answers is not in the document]
wjwwood: This feels weird to me... what if someone wants to install the sigterm but not the sigint handler?

It seems like we need another configuration instead, which has a bool for each, and then it feels more natural to see "init options" and "signal handler options" as the two arguments to `init()` since it is "context.init" + "install_signal_handlers".

rclcpp/include/rclcpp/init_options.hpp
```diff
@@ -29,7 +29,7 @@ class InitOptions
 {
 public:
   /// If true, the context will be shutdown on SIGINT by the signal handler (if it was installed).
-  bool shutdown_on_sigint = true;
+  bool shutdown_on_signal = true;
```
wjwwood, line 32: The comment needs to be updated, as currently this would possible affect SIGTERM too.

But on that point, what if I wanted to shutdown on SIGTERM but not SIGINT? And on another Context shutdown on both.

I think, unfortunately, to do this right, each context needs the ability to shutdown on SIGTERM, SIGINT, both, or neither.

rclcpp/src/rclcpp/signal_handler.cpp
wjwwood: > you could get two signals before the deferred signal handler starts running, so it really doesn't help much.

The only important thing is that if you get sigterm and sigint you notice both (regardless of how many times they occur), so you could just have two atomic bools to check here, one for sigint and one for sigterm.

rclcpp/src/rclcpp/signal_handler.hpp
```diff
@@ -39,14 +39,14 @@
 namespace rclcpp
 {
 
-/// Responsible for manaaging the SIGINT signal handling.
+/// Responsible for manaaging the SIGINT/SIGTERM signal handling.
```
wjwwood, line 42:
```suggestion
/// Responsible for managing the SIGINT/SIGTERM signal handling.
```

wjwwood (changes requested):

rclcpp/include/rclcpp/utilities.hpp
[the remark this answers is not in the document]
ivanpauno: Adding a "signal handler options" sounds good to me, it could be an enum "install sigint/install sigterm/install both/ install none"

rclcpp/include/rclcpp/init_options.hpp
ivanpauno: We can add that, I'm not sure if it's really useful though.

rclcpp/src/rclcpp/signal_handler.hpp
ivanpauno: Fixed https://github.com/ros2/rclcpp/pull/1771/commits/a8055c9ddf39e5a0d84cabbdfa0083ce5c17ebc3.

rclcpp/src/rclcpp/signal_handler.cpp
ivanpauno: > The only important thing is that if you get sigterm and sigint you notice both (regardless of how many times they occur), so you could just have two atomic bools to check here, one for sigint and one for sigterm.

yes, I think that's possible.
It's a bit tricky though

rclcpp/include/rclcpp/utilities.hpp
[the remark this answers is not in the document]
ivanpauno: Done, see https://github.com/ros2/rclcpp/pull/1771/commits/b4d87ea24b63c2abfe412b0240dcfa8ed1c7ee69.

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -42,6 +42,20 @@ std::string to_string(T value)
 
 namespace rclcpp
 {
+
+/// Option to indicate what signal handlers should rclcpp install.
```
wjwwood, line 46:
```suggestion
/// Option to indicate which signal handlers rclcpp should install.
```

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -151,98 +214,57 @@ SignalHandler::~SignalHandler()
...
+  if (SIGINT == signum) {
+    return old_sigint_handler_;
+  } else if (SIGTERM == signum) {
+    return old_sigterm_handler_;
+  }
 #if defined(RCLCPP_HAS_SIGACTION)
-  ssize_t ret = sigaction(signal_value, &signal_handler, &old_signal_handler);
-  signal_handler_install_failed = (ret == -1);
+  SignalHandler::signal_handler_type ret;
+  memset(&ret, 0, sizeof(ret));
```
wjwwood, line 236: We probably shouldn't use a bare `memset` here, instead maybe the C11 variant or the Windows specific "safe" version:

https://en.cppreference.com/w/c/string/byte/memset

Ideally, we'd have our own util function to do the right thing in the right places, like we do for other C calls, like `snprintf` in `rcutils`.

The reason I mention this is that I'm currently fixing clang-tidy warnings and this is one of them (using normal `memset` rather than `memset_s`).

wjwwood: I know we were using it before, I just wanted to point it out. I'll have to change that after this is merged, which is fine.

wjwwood (commented):
I have two remaining concerns:

- the change away from static members
  - this part of the pull request seems unrelated to supporting sigterm, and honestly touching that code seems like a bad idea, given how sensitive it has been in the past, can you explain why it was needed to add support for sigterm?
- It is not possible to tell which signal handlers were installed after calling `install_signal_handers()`. You can tell that something was installed, but not what, which might be a problem in the future or not, but from a purely design perspective we're loosing access to information that the system has for no reason.

ivanpauno:
> It is not possible to tell which signal handlers were installed after calling install_signal_handers(). You can tell that something was installed, but not what, which might be a problem in the future or not, but from a purely design perspective we're loosing access to information that the system has for no reason.
the change away from static members
this part of the pull request seems unrelated to supporting sigterm, and honestly touching that code seems like a bad idea, given how sensitive it has been in the past, can you explain why it was needed to add support for sigterm?

Sounds good, I can add that.

> the change away from static members
> this part of the pull request seems unrelated to supporting sigterm, and honestly touching that code seems like a bad idea, given how sensitive it has been in the past, can you explain why it was needed to add support for sigterm?

I'm not sure if I actually needed that, I think I was trying to simplify the code before adding sigterm support.
I don't see much risk tbh, we moved from globals to locals with static lifetime, which are less risky: it's impossible to have initialization order issues with static locals (https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use).
Now we actually only have one object with static lifetime (the signal handler), instead of two interacting (signal handler and logger), so it's impossible to have construction/destruction order issues.
If you mind strongly I can revert that part.

rclcpp/src/rclcpp/signal_handler.cpp
ivanpauno: The c11 variant (memset_s) is not supported by linux (all the bound checked library functions are "optional" in the c11 standard).
Could you show me a code snippet of the approach you're using so I apply the same one here?

rclcpp/include/rclcpp/utilities.hpp
ivanpauno: Thanks! Fixed in https://github.com/ros2/rclcpp/pull/1771/commits/6c75ff7a28bcaef26b8e2df3426fb079a0172cc3.

ivanpauno:
> It is not possible to tell which signal handlers were installed after calling install_signal_handers(). You can tell that something was installed, but not what, which might be a problem in the future or not, but from a purely design perspective we're loosing access to information that the system has for no reason.
the change away from static members
this part of the pull request seems unrelated to supporting sigterm, and honestly touching that code seems like a bad idea, given how sensitive it has been in the past, can you explain why it was needed to add support for sigterm?

See https://github.com/ros2/rclcpp/pull/1771/commits/d23d20652f9c8bf74104e8abad5c07e7a923c5b9.

wjwwood:
> If you mind strongly I can revert that part.

I don't I just saw it as unnecessary and that code is sensitive. I'm not concerned about the new approach, only that it's working now and changing it doesn't benefit us much but does introduce risk of regression. That being said, it's already changed, and it appears to make sense, so that's fine. If it breaks we'll have to revert and fix it up, but that's also fine.

Zainskl:
@ivanpauno ros2 humble:
When we do sudo reboot, nodes doesn't run cleanup methods (defined inside or outside of myNode() class
Let say
class MyNode():
     code...
      def cleanup(self):
           do something....
or outside of block

def do_something(node):
     code ...

In main():
    rclpy.init(args=args)
    node = MyNode()
    try:
         rclpy.spin(node)
    except Exception as e:
          print(f"error {str(e)}")
    finally:
          do_something(node)
          node.cleanup()
          node.destroy_node()
          if rclpy.ok():
               rclpy.shutdown()


Do you know what's happening?

christophebedard:
If you're facing an issue, please create a new issue (and mention this PR), otherwise it is hard for us to keep track of comments on 4-year-old merged PRs.

Issue #1772, opened 2021-09-14
Support a bunch of SingleThreadedExecutor with dedicated thread for each component nodes


gezp:
#### Feature description

Hi, guys

I'm working on project [Reduce ROS2 Nodes and Determinism](https://navigation.ros.org/2021summerOfCode/projects/spinners.html) of OSPP2021 (under the mentor of Steve Macenski).

Currently, i'm trying to support composed bringup for Nav2 stack, and i have done some works about manually composed bringup (`Compile-time composition`) of Nav2, which performs better than Normal bringup , in addition, i find a problem that  `a large multi-threaded executor` consumes higher cpu(increase 30%-50%) than `a bunch of single-threaded executors` , you can find some details here: https://discourse.ros.org/t/nav2-composition/22175. 

for `Run-time composition` , i notice that `rclcpp_components` only provides `component_container` which use `SingleThreadedExecutor` for all component nodes in container and `component_container_mt` which use `MultiThreadedExecutor`.  

how about supporting a bunch of `SinglethreadedExecutor` for  `component_container` ?  in other word, create a `SingleThreadedExecutor` with dedicated thread for each component nodes.

Edited : clean some irrelevant topic.

SteveMacenski:
@gezp please propose the simplified design we chatted about to the maintainers. I'm sure just the parameter + simplified needs we directly have can be accommodated without much concern.

gezp:
@SteveMacenski , i have cleaned this ticket, and open a new ticket for another topic: https://github.com/ros2/rclcpp/issues/1774.

SteveMacenski:
Should this be closed then?

PR #1773, opened 2021-09-14
Support a bunch of SingleThreadedExecutor for component nodes

gezp:

related to https://github.com/ros2/rclcpp/issues/1772. make `component_container` more flexible.

With these changes, we could choose `SingleThreadedExecutor`,  `MultiThreadedExecutor`, or a bunch of `SingleThreadedExecutor` to spin component nodes by setting parameters (`use_multi_executors` and `use_multi_threads`) for `component_container` .

in addition, `component_container_mt` should be deprecated in the future.

### DIFF
--- a/rclcpp_components/include/rclcpp_components/component_manager.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager.hpp
@@ -93,13 +93,11 @@ public:
    * Initializes the component manager. It creates the services: load node, unload node
    * and list nodes.
    *
-   * \param executor the executor which will spin the node.
    * \param node_name the name of the node that the data originates from.
    * \param node_options additional options to control creation of the node.
    */
   RCLCPP_COMPONENTS_PUBLIC
   ComponentManager(
-    std::weak_ptr<rclcpp::Executor> executor,
     std::string node_name = "ComponentManager",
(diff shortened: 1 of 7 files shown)


SteveMacenski:
This seems premature, I would close this PR. Lets have some discussion in the ticket and propose an idea for discussion. There's alot of things in this PR that would need to be changed and its better to have a discussion on design before jumping into the implementation (e.g. you should not hijack the multithreaded container to make it a multi-single threaded container and add additional statically set arguments).

fujitatomoya:
@gezp (CC: @SteveMacenski )

thanks for opening PR.

> This seems premature, I would close this PR. Lets have some discussion in the ticket and propose an idea for discussion.

agree.

especially what i would like to know the performance difference between `SingleThreaded`, `Multiple-SingleThreaded` and `MultiThreaded` cases with CPU consumption including details about number of threads, statistics for each thread.
for example, if the same task and performance or output with the same number of threads, and `MultiThreaded` is consuming more CPU resource than `Multiple-SingleThreaded`, this could be something we need to improve, i guess.

gezp:
thank you for suggestions, i will close this PR, and continue discuss these in ticket(https://github.com/ros2/rclcpp/issues/1772).

Issue #978, opened 2020-02-03
Setting certain extra-arguments to components has no effect


fabmene:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Binaries
- Version or commit hash:
  - Eloquent
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Minimal component example:
``` 
#include "rclcpp/rclcpp.hpp"
class MyNode : public rclcpp::Node
{
public:
  explicit MyNode(const rclcpp::NodeOptions& options) : rclcpp::Node("my_node", options){
    setvbuf(stdout, NULL, _IONBF, BUFSIZ);
    if (options.use_intra_process_comms())
      RCLCPP_INFO(this->get_logger(), "Using IPC");
    if (options.allow_undeclared_parameters())
      RCLCPP_INFO(this->get_logger(), "Allowing undeclared params");
    if (options.use_global_arguments())
      RCLCPP_INFO(this->get_logger(), "Using global arguments");
};

#include "rclcpp_components/register_node_macro.hpp"
RCLCPP_COMPONENTS_REGISTER_NODE(my_pkg::MyNode)
```
Start a component container and load above component with extra arguments, either as launch, or via cli:
```
container = ComposableNodeContainer(
            node_name='my_container',
            node_namespace='',
            package='rclcpp_components',
            node_executable='component_container',
            composable_node_descriptions=[
                ComposableNode(
                    package='my_pkg',
                    node_plugin='my_node::MNode',
                    node_name='my_node1',
                    extra_arguments=[{'use_intra_process_comms': True}, \
                      {'allow_undeclared_parameters': True}, \
                      {'use_global_arguments': True}])
            ], 
            output='screen',
```
#### Expected behavior
```
[component_container-1] [INFO] [my_node]: Using IPC
[component_container-1] [INFO] [my_node]: Allowing undeclared params
[component_container-1] [INFO] [my_node]: Using global arguments
```

#### Actual behavior
```
[component_container-1] [INFO] [my_node]: Using IPC
```

#### Additional Information
Does not seem to depend on order or number of arguments.

ivanpauno:
If I'm not mistaken, it's supposed only to work with `use_intra_process_comms`.
Maybe @hidmic can clarify

hidmic:
[...]

PR #1776, opened 2021-09-16
add use_global_arguments for node options of component nodes

gezp:

related to https://github.com/ros2/rclcpp/issues/978#. 

i'm trying to implement composed bringup for Nav2, but it can't work as expected. the reason is that internal nodes (like `local_costmap`, `global_costmap`) in Nav2 couldn't  load parameters from param's file(`use_global_arguments` is setting `false`) https://github.com/ros2/rclcpp/blob/f30329fbec79cb699f41a37cc27c0c475de3edc1/rclcpp_components/src/component_manager.cpp#L164

according to @wjwwood opinion( https://github.com/ros2/rclcpp/issues/978#issuecomment-593718794 ),    `allow_undeclared_parameters` and `use_global_arguments` shouldn't be a parameter,  but the `use_global_arguments`  does have a legitimate use case here (for Nav2), and i think it's useful when we have a large params file to configure all nodes (or component nodes).

`use_global_arguments = true` is default behavior for all other situations so its strange that we cannot set it to true at all due to "its generally bad" advise for the component container. It creates a situations where component container nodes have a different behavior from everywhere else and it silently fails if a user is clever enough to know that that is the problem. 

I still offer a warning to tell people not to do it but gives the flexibility to get the behavior that is expected every where else from the component container if explicitly called out in the extra parameters.

some extra details here: https://github.com/ros2/rclcpp/issues/978#issuecomment-920398249

### DIFF
--- a/rclcpp_components/src/component_manager.cpp
+++ b/rclcpp_components/src/component_manager.cpp
@@ -160,6 +160,18 @@ ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> r
                 "Extra component argument 'use_intra_process_comms' must be a boolean");
       }
       options.use_intra_process_comms(extra_argument.get_value<bool>());
+    } else if (extra_argument.get_name() == "use_global_arguments") {
+      if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
+        throw ComponentManagerException(
+                "Extra component argument 'use_global_arguments' must be a boolean");
+      }
+      options.use_global_arguments(extra_argument.get_value<bool>());
+      if (extra_argument.get_value<bool>()) {
+        RCLCPP_WARN(
+          get_logger(), "use_global_arguments is true by default in nodes, but is not "
+          "recommended in a component manager. If true, this will cause this node's behavior "
+          "to be influenced by global arguments, not only those targeted at this node.");
+      }
     }
   }
 
--- a/rclcpp_components/src/component_manager.cpp
+++ b/rclcpp_components/src/component_manager.cpp
@@ -160,15 +160,15 @@ ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> r
                 "Extra component argument 'use_intra_process_comms' must be a boolean");
       }
       options.use_intra_process_comms(extra_argument.get_value<bool>());
-    } else if (extra_argument.get_name() == "use_global_arguments") {
+    } else if (extra_argument.get_name() == "forward_global_arguments") {
       if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
         throw ComponentManagerException(
-                "Extra component argument 'use_global_arguments' must be a boolean");
+                "Extra component argument 'forward_global_arguments' must be a boolean");
       }
       options.use_global_arguments(extra_argument.get_value<bool>());
       if (extra_argument.get_value<bool>()) {
         RCLCPP_WARN(
-          get_logger(), "use_global_arguments is true by default in nodes, but is not "
+          get_logger(), "forward_global_arguments is true by default in nodes, but is not "
           "recommended in a component manager. If true, this will cause this node's behavior "
           "to be influenced by global arguments, not only those targeted at this node.");
       }
--- a/rclcpp_components/test/test_component_manager_api.cpp
+++ b/rclcpp_components/test/test_component_manager_api.cpp
@@ -211,6 +211,50 @@ TEST_F(TestComponentManager, components_api)
     EXPECT_EQ(result->unique_id, 0u);
   }
 
+  {
+    // use_global_arguments
+    auto request = std::make_shared<composition_interfaces::srv::LoadNode::Request>();
+    request->package_name = "rclcpp_components";
+    request->plugin_name = "test_rclcpp_components::TestComponentFoo";
+    request->node_name = "test_component_global_arguments";
+    rclcpp::Parameter use_global_arguments("use_global_arguments",
+      rclcpp::ParameterValue(true));
+    request->extra_arguments.push_back(use_global_arguments.to_parameter_msg());
+
+    auto future = composition_client->async_send_request(request);
+    auto ret = exec->spin_until_future_complete(future, 5s);  // Wait for the result.
+    auto result = future.get();
+    EXPECT_EQ(ret, rclcpp::FutureReturnCode::SUCCESS);
+    EXPECT_EQ(result->success, true);
+    EXPECT_EQ(result->error_message, "");
+    std::cout << result->full_node_name << std::endl;
+    EXPECT_EQ(result->full_node_name, "/test_component_global_arguments");
+    EXPECT_EQ(result->unique_id, 7u);
+  }
+
+  {
+    // use_global_arguments is not a bool type parameter
+    auto request = std::make_shared<composition_interfaces::srv::LoadNode::Request>();
+    request->package_name = "rclcpp_components";
+    request->plugin_name = "test_rclcpp_components::TestComponentFoo";
+    request->node_name = "test_component_global_arguments_str";
+
+    rclcpp::Parameter use_global_arguments("use_global_arguments",
+      rclcpp::ParameterValue("hello"));
+    request->extra_arguments.push_back(use_global_arguments.to_parameter_msg());
+
+    auto future = composition_client->async_send_request(request);
+    auto ret = exec->spin_until_future_complete(future, 5s);  // Wait for the result.
+    auto result = future.get();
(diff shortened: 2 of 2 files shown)


rclcpp_components/src/component_manager.cpp
```diff
@@ -160,6 +159,16 @@ ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> r
...
       options.use_intra_process_comms(extra_argument.get_value<bool>());
+    } else if (extra_argument.get_name() == "use_global_arguments") {
+      if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
+        throw ComponentManagerException(
+                "Extra component argument 'use_global_arguments' must be a boolean");
+      }
+      options.use_global_arguments(extra_argument.get_value<bool>());
+      RCLCPP_WARN(
+        get_logger(),
+        "This behavior is not recommended due to potential side-effect "
```
SteveMacenski, line 170: A bit more detail / context here may be good.

fujitatomoya: if the custom node constructor overwrites these options with default value (rejects specified options), these options cannot be changed or effective.

```
  SomeCustomNode(
    rclcpp::NodeOptions options
  )
  : Node(
      "custom_node",
      options.allow_undeclared_parameters(true).
      automatically_declare_parameters_from_overrides(true))
  {
  ...<snip>    
  }
```

i think that should be documented here? i might be missing something...

i mean if we can change the option for the node, that something we can do as user? and if that leads to side-effect or anything, that is the responsibility for node implementation, because it takes the arguments w/o check?

fujitatomoya (commented):
I think this extension makes sense for use case of `ComponentManager`.

rclcpp_components/src/component_manager.cpp
fujitatomoya: maybe it can take some explanation from

https://github.com/ros2/rclcpp/blob/fa3a6fa597c5d40b2fce46e5fb95f541e62076e7/rclcpp/include/rclcpp/node_options.hpp#L152-L156

SteveMacenski: Would `use_global_arguments is true by default in nodes, but is not recommended in a component manager. If true, this will cause this node's behavior to be influenced by global arguments, not only those targeted at this node.` work for you?

gezp: > use_global_arguments is true by default in nodes, but is not recommended in a component manager. 

`use_global_arguments ` is true by default for `NodeOptions` ,but false in component manager currently, i think we should keep this default value `false` for component manager, what do you think about ? @SteveMacenski

SteveMacenski: Yes, we are not changing the default. This warning is just telling people that `true` is default in general, but recommended not to be in this situation.

SteveMacenski (approved):
@fujitatomoya any other changes required?

rclcpp_components/src/component_manager.cpp
```diff
@@ -160,6 +160,16 @@ ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> r
...
+    } else if (extra_argument.get_name() == "use_global_arguments") {
+      if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
+        throw ComponentManagerException(
+                "Extra component argument 'use_global_arguments' must be a boolean");
+      }
+      options.use_global_arguments(extra_argument.get_value<bool>());
+      RCLCPP_WARN(
+        get_logger(), "use_global_arguments is true by default in nodes, but is not "
+        "recommended in a component manager. If true, this will cause this node's behavior "
+        "to be influenced by global arguments, not only those targeted at this node.");
```
fujitatomoya, line 172: we would want to print the warning message only when `use_global_arguments` is actually `true`? (not with `use_global_arguments` is `false` by user)

fujitatomoya:
@ivanpauno @hidmic @wjwwood there's been discussion on this via https://github.com/ros2/rclcpp/issues/978, what would you think? i think extending this option makes sense for user application.

rclcpp_components/src/component_manager.cpp
SteveMacenski: Good call

fujitatomoya (approved):
im good to go.

rclcpp_components/src/component_manager.cpp
```diff
@@ -160,6 +160,18 @@ ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> r
                 "Extra component argument 'use_intra_process_comms' must be a boolean");
       }
       options.use_intra_process_comms(extra_argument.get_value<bool>());
+    } else if (extra_argument.get_name() == "use_global_arguments") {
```
hidmic, line 163: @gezp meta: thinking about your use case, I wonder if it would make sense for the `component_manager` node to expose, e.g. as a parameter, the default value of `use_global_arguments` for the nodes it composes. Seems convenient if you're already passing a large parameter file to it.

hidmic (commented):
@gezp I'm onboard with this change. This patch is missing a test though.

rclcpp_components/src/component_manager.cpp
gezp: i agree with you, for my use case (use a large parameter file for all components), it seems better to use a parameter of  `component_manager` to configure all components `use_global_arguments`, which we needn't set this for components one by one in launch file. 
I'm not good at naming parameter, could you give me some suggestions ? @hidmic . how about  `use_global_arguments`  (i think it's bad, because it seems that we set `component_manager` nodes to use global arguments, but we want to set this for components)  or `component_use_global_arguments`?

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-10-28/22947/1

SteveMacenski:
@fujitatomoya @hidmic can we run CI again to merge it?

fujitatomoya:
CI retry:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15623/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10304/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13291/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15840/)

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-11-18/23209/1

rclcpp_components/test/test_component_manager_api.cpp
```diff
@@ -211,6 +211,50 @@ TEST_F(TestComponentManager, components_api)
...
+
+    auto future = composition_client->async_send_request(request);
+    auto ret = exec->spin_until_future_complete(future, 5s);  // Wait for the result.
+    auto result = future.get();
+    EXPECT_EQ(ret, rclcpp::FutureReturnCode::SUCCESS);
+    EXPECT_EQ(result->success, true);
+    EXPECT_EQ(result->error_message, "");
+    std::cout << result->full_node_name << std::endl;
+    EXPECT_EQ(result->full_node_name, "/test_component_global_arguments");
+    EXPECT_EQ(result->unique_id, 7u);
```
hidmic, line 232: @gezp nit: this tests that the manager did not fail but not that it did anything about `use_global_arguments`.

rclcpp_components/src/component_manager.cpp
hidmic: @gezp perhaps `forward_global_arguments`? I'm not very creative either. At any rate, I won't block this patch on it.

hidmic (approved):
Sorry for the delay (I was out for some time).

Overall makes sense to me. Tests could use some improvement though.

SteveMacenski:
@gezp can you make the testing related changes so we can get this merged in and get dynamic composition in Nav2? :partying_face:

rclcpp_components/test/test_component_manager_api.cpp
gezp: yes, `use_global_arguments`  is set true in `NodeOption` , but this component didn't use any global arguments, i don't know how to test for this.

SteveMacenski:
@gezp I think you just need to add the requested test and they can re-review

hidmic:
> @hidmic is this ready to review again?

Argh. This fell through the cracks again. Sorry about that.

---

Last CI (hopefully):


* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15938/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10611/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16260/)

Issue #1774, opened 2021-09-15
use dedicated executor for component nodes 


gezp:
#### Feature description

Hi, guys

I'm working on project [Reduce ROS2 Nodes and Determinism](https://navigation.ros.org/2021summerOfCode/projects/spinners.html) of OSPP2021 (under the mentor of Steve Macenski).

Currently, i'm trying to support composed bringup for Nav2 stack, and i have done some works about manually composed bringup (`Compile-time composition`) of Nav2, which performs better than Normal bringup , in addition, i find a problem that  `a large multi-threaded executor` consumes higher cpu(increase 30%-50%) than `a bunch of single-threaded executors` , you can find some details here: https://discourse.ros.org/t/nav2-composition/22175. 

for `Run-time composition` , i notice that `rclcpp_components` only provides `component_container` which use `SingleThreadedExecutor` for all component nodes in container and `component_container_mt` which use `MultiThreadedExecutor`.  

how about supporting a bunch of `SinglethreadedExecutor` for  `component_container` ?  in other word, create a `SingleThreadedExecutor` with dedicated thread for each component nodes.

gezp:
with this feature, we can merge `component_container` and `component_container_mt`,  and we can select `SinglethreadedExecutor` or`MultiThreadedExecutor`  for component nodes by parameters in `ComponentManager`

SteveMacenski:
This should be a parameterized option with the defaults for the current behavior of a single single-threaded executor. If a parameter, for instance `dedicated_thread` is set to `true`, when a new component is added, it is added to its own executor instead of a single one for all components. 

I don't think this is clear that that is the objective best solution, but it is a useful option to make available based on the needs and discussions in https://discourse.ros.org/t/nav2-composition/22175. In that ticket, we bring up analysis showing that N-single threaded executors are far more computationally efficient than a single N-multithreaded executor. We'd like to be able to have this N-single threaded container option available to us so we can do dynamic composed bring up in Nav2 rather than having to manually compose them in a process. That's not very flexible and configurable and afterall Nav2 is a framework rather than a standalone solution.

We do not suggest any changes to the multi-threaded executor container, nor are we suggesting to merge them into a single component manager, that would be feature creep and it makes sense for those to be separate executables. 

What we do suggest is the following:
- Add a parameter, such as `dedicated_thread` for the container to load the components into a single executor if false and into individual dedicated executor threads if true. Default to false for current behavior. 
[...]

PR #1781, opened 2021-09-23
support dedicated executor for component nodes

gezp:

Related to https://github.com/ros2/rclcpp/issues/1774

it offer an “executor per component” behavior for container, we can use `use_dedicated_thread` parameter to enable this.

### DIFF
--- a/rclcpp_components/CMakeLists.txt
+++ b/rclcpp_components/CMakeLists.txt
@@ -61,9 +61,20 @@ ament_target_dependencies(component_container_mt
   "rclcpp"
 )
 
+add_executable(
+  component_container_isolated
+  src/component_container_isolated.cpp
+)
+target_link_libraries(component_container_isolated component_manager)
+ament_target_dependencies(component_container_isolated
+  "rclcpp"
+)
+
+
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   target_link_libraries(component_container "stdc++fs")
   target_link_libraries(component_container_mt "stdc++fs")
+  target_link_libraries(component_container_isolated "stdc++fs")
 endif()
 
 if(BUILD_TESTING)
@@ -135,7 +146,7 @@ install(
 
 # Install executables
 install(
-  TARGETS component_container component_container_mt
+  TARGETS component_container component_container_mt component_container_isolated
   RUNTIME DESTINATION lib/${PROJECT_NAME}
 )
 
--- a/rclcpp_components/include/rclcpp_components/component_manager.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager.hpp
@@ -140,6 +140,22 @@ protected:
   virtual rclcpp::NodeOptions
   create_node_options(const std::shared_ptr<LoadNode::Request> request);
 
+  /// Add component node to executor model, it's invoked in on_load_node()
+  /**
+   * \param node_id  node_id of loaded component node in node_wrappers_
+   */
+  RCLCPP_COMPONENTS_PUBLIC
+  virtual void
+  add_node_to_executor(uint64_t node_id);
+
+  /// Remove component node from executor model, it's invoked in on_unload_node()
+  /**
+   * \param node_id  node_id of loaded component node in node_wrappers_
+   */
+  RCLCPP_COMPONENTS_PUBLIC
+  virtual void
+  remove_node_from_executor(uint64_t node_id);
+
   /// Service callback to load a new node in the component
   /**
    * This function allows to add parameters, remap rules, a specific node, name a namespace
@@ -231,7 +247,7 @@ protected:
     on_list_nodes(request_header, request, response);
   }
 
-private:
+protected:
   std::weak_ptr<rclcpp::Executor> executor_;
 
   uint64_t unique_id_ {1};
--- a/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
@@ -1,0 +1,99 @@
+// Copyright 2021 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+#ifndef RCLCPP_COMPONENTS__COMPONENT_MANAGER_ISOLATED_HPP__
+#define RCLCPP_COMPONENTS__COMPONENT_MANAGER_ISOLATED_HPP__
+
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+#include <unordered_map>
+
+#include "rclcpp_components/component_manager.hpp"
+
+
+namespace rclcpp_components
+{
+/// ComponentManagerIsolated uses dedicated single-threaded executors for each components.
+template<typename ExecutorT = rclcpp::executors::SingleThreadedExecutor>
+class ComponentManagerIsolated : public rclcpp_components::ComponentManager
+{
+  using rclcpp_components::ComponentManager::ComponentManager;
+
+  struct DedicatedExecutorWrapper
+  {
+    std::shared_ptr<rclcpp::Executor> executor;
+    std::thread thread;
+    std::promise<void> promise;
+  };
+
+public:
+  ~ComponentManagerIsolated()
+  {
+    if (node_wrappers_.size()) {
+      for (auto & executor_wrapper : dedicated_executor_wrappers_) {
+        executor_wrapper.second.promise.set_value();
+        executor_wrapper.second.executor->cancel();
+        executor_wrapper.second.thread.join();
+      }
+      node_wrappers_.clear();
+    }
+  }
+
+protected:
+  /// Add component node to executor model, it's invoked in on_load_node()
+  /**
+   * \param node_id  node_id of loaded component node in node_wrappers_
+   */
+  RCLCPP_COMPONENTS_PUBLIC
+  void
+  add_node_to_executor(uint64_t node_id) override
+  {
+    DedicatedExecutorWrapper executor_wrapper;
+    auto exec = std::make_shared<ExecutorT>();
+    exec->add_node(node_wrappers_[node_id].get_node_base_interface());
+    executor_wrapper.executor = exec;
+    executor_wrapper.thread = std::thread(
+      [exec, cancel_token = executor_wrapper.promise.get_future()]() {
+        exec->spin_until_future_complete(cancel_token);
+      });
+    dedicated_executor_wrappers_[node_id] = std::move(executor_wrapper);
+  }
+  /// Remove component node from executor model, it's invoked in on_unload_node()
+  /**
+   * \param node_id  node_id of loaded component node in node_wrappers_
+   */
+  RCLCPP_COMPONENTS_PUBLIC
+  void
+  remove_node_from_executor(uint64_t node_id) override
+  {
+    auto executor_wrapper = dedicated_executor_wrappers_.find(node_id);
+    if (executor_wrapper != dedicated_executor_wrappers_.end()) {
+      executor_wrapper->second.promise.set_value();
+      executor_wrapper->second.executor->cancel();
+      executor_wrapper->second.thread.join();
+      dedicated_executor_wrappers_.erase(executor_wrapper);
+    }
+  }
+
+private:
+  std::unordered_map<uint64_t, DedicatedExecutorWrapper> dedicated_executor_wrappers_;
+};
+
+}  // namespace rclcpp_components
+
+#endif  // RCLCPP_COMPONENTS__COMPONENT_MANAGER_ISOLATED_HPP__
--- a/rclcpp_components/src/component_container_isolated.cpp
+++ b/rclcpp_components/src/component_container_isolated.cpp
@@ -1,0 +1,49 @@
+// Copyright 2021 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include <vector>
+#include <string>
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/utilities.hpp"
+#include "rclcpp_components/component_manager_isolated.hpp"
+
+int main(int argc, char * argv[])
+{
+  /// Component container with dedicated single-threaded executors for each components.
+  rclcpp::init(argc, argv);
+  // parse arguments
+  bool use_multi_threaded_executor{false};
+  std::vector<std::string> args = rclcpp::remove_ros_arguments(argc, argv);
+  for (auto & arg : args) {
+    if (arg == std::string("--use_multi_threaded_executor")) {
+      use_multi_threaded_executor = true;
+    }
+  }
+  // create executor and component manager
+  auto exec = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+  rclcpp::Node::SharedPtr node;
+  if (use_multi_threaded_executor) {
+    using ComponentManagerIsolated =
+      rclcpp_components::ComponentManagerIsolated<rclcpp::executors::MultiThreadedExecutor>;
+    node = std::make_shared<ComponentManagerIsolated>(exec);
+  } else {
+    using ComponentManagerIsolated =
+      rclcpp_components::ComponentManagerIsolated<rclcpp::executors::SingleThreadedExecutor>;
+    node = std::make_shared<ComponentManagerIsolated>(exec);
+  }
+  exec->add_node(node);
+  exec->spin();
+}
--- a/rclcpp_components/src/component_manager.cpp
+++ b/rclcpp_components/src/component_manager.cpp
@@ -167,6 +167,22 @@ ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> r
 }
 
 void
+ComponentManager::add_node_to_executor(uint64_t node_id)
+{
+  if (auto exec = executor_.lock()) {
+    exec->add_node(node_wrappers_[node_id].get_node_base_interface(), true);
+  }
+}
+
+void
+ComponentManager::remove_node_from_executor(uint64_t node_id)
+{
+  if (auto exec = executor_.lock()) {
+    exec->remove_node(node_wrappers_[node_id].get_node_base_interface());
+  }
+}
+
+void
 ComponentManager::on_load_node(
   const std::shared_ptr<rmw_request_id_t> request_header,
   const std::shared_ptr<LoadNode::Request> request,
@@ -214,10 +230,9 @@ ComponentManager::on_load_node(
         throw ComponentManagerException("Component constructor threw an exception");
       }
 
+      add_node_to_executor(node_id);
+
       auto node = node_wrappers_[node_id].get_node_base_interface();
-      if (auto exec = executor_.lock()) {
-        exec->add_node(node, true);
-      }
       response->full_node_name = node->get_fully_qualified_name();
       response->unique_id = node_id;
       response->success = true;
@@ -253,9 +268,7 @@ ComponentManager::on_unload_node(
     response->error_message = ss.str();
     RCLCPP_WARN(get_logger(), "%s", ss.str().c_str());
   } else {
-    if (auto exec = executor_.lock()) {
-      exec->remove_node(wrapper->second.get_node_base_interface());
-    }
+    remove_node_from_executor(request->unique_id);
     node_wrappers_.erase(wrapper);
     response->success = true;
   }
--- a/rclcpp_components/test/test_component_manager_api.cpp
+++ b/rclcpp_components/test/test_component_manager_api.cpp
@@ -22,6 +22,7 @@
 #include "composition_interfaces/srv/list_nodes.hpp"
 
 #include "rclcpp_components/component_manager.hpp"
+#include "rclcpp_components/component_manager_isolated.hpp"
 
 using namespace std::chrono_literals;
 
@@ -36,11 +37,18 @@ protected:
 
 // TODO(hidmic): split up tests once Node bring up/tear down races
 //               are solved https://github.com/ros2/rclcpp/issues/863
-TEST_F(TestComponentManager, components_api)
+void test_components_api(bool use_dedicated_executor)
 {
   auto exec = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
   auto node = rclcpp::Node::make_shared("test_component_manager");
-  auto manager = std::make_shared<rclcpp_components::ComponentManager>(exec);
+  std::shared_ptr<rclcpp_components::ComponentManager> manager;
+  if (use_dedicated_executor) {
+    using ComponentManagerIsolated =
+      rclcpp_components::ComponentManagerIsolated<rclcpp::executors::SingleThreadedExecutor>;
+    manager = std::make_shared<ComponentManagerIsolated>(exec);
+  } else {
+    manager = std::make_shared<rclcpp_components::ComponentManager>(exec);
+  }
 
   exec->add_node(manager);
   exec->add_node(node);
@@ -319,5 +327,17 @@ TEST_F(TestComponentManager, components_api)
       EXPECT_EQ(result_unique_ids[3], 5u);
       EXPECT_EQ(result_unique_ids[4], 6u);
     }
+  }
+}
+
+TEST_F(TestComponentManager, components_api)
+{
+  {
+    SCOPED_TRACE("ComponentManager");
+    test_components_api(false);
+  }
+  {
+    SCOPED_TRACE("ComponentManagerIsolated");
+    test_components_api(true);
   }
 }


rclcpp_components/src/component_manager.cpp
```diff
@@ -215,9 +227,17 @@ ComponentManager::on_load_node(
...
+        auto exec = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+        exec->add_node(node);
+        executors_[node_id] = exec;
+        executor_threads_[node_id] = std::make_shared<std::thread>([exec]() {exec->spin();});
+      } else {
+        if (auto exec = executor_.lock()) {
+          exec->add_node(node, true);
+        }
       }
+
```
fujitatomoya, line 240: Is this new line is required to add? if not, we would probably want to avoid the unnecessary change.

fujitatomoya (changes requested):
probably we would want to discuss on implementation details, and test would be ideal too.

rclcpp_components/src/component_manager.cpp
```diff
@@ -37,6 +37,9 @@ ComponentManager::ComponentManager(
 : Node(std::move(node_name), node_options),
   executor_(executor)
 {
+  declare_parameter("use_dedicated_thread", use_dedicated_thread_);
```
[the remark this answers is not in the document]
gezp: @fujitatomoya do you mean that we could use `request->extra_arguments` to choose `Conatiner's Executor` or `Dedicated SingleThreadedExecutor`  that node should be loaded on at runtime? so we needn't global parameter `use_dedicated_thread`.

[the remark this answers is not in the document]
gezp: i don't consider to change `use_dedicated_thread` at runtime,  the parameter `use_dedicated_thread` is only  initialized at constructor, which won't change at runtime. i don't understand the user case you mentioned.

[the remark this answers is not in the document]
fujitatomoya: I was thinking both cases make sense,

- `use_dedicated_thread` is constructor option. once this option is chosen, any nodes that are loaded will have dedicated `SingleThreadedExecutor`.
- `use_dedicated_thread` parameter, so that user can have full control either loading nodes on dedicated `SingleThreadedExecutor` or the executor that ComponentManager spins on. (control where to load the nodes.)

I would like to hear from other opinion.

iuhilnehc-ynos:
I think it's better to add an item `use_dedicated_thread` in https://github.com/ros2/rcl_interfaces/blob/9a44202bbaf76e19fde21da0e90d8afbe89bd263/composition_interfaces/srv/LoadNode.srv.

fujitatomoya:
@iuhilnehc-ynos i thought that too, i think that also makes sense.

gezp:
@iuhilnehc-ynos  is it necessary to add a new item `use_dedicated_thread ` in `LoadNode.srv` ?  what do you think about using `extra_arguments` in  `LoadNode.srv`  to load parameter `use_dedicated_thread `  like `use_intra_process_comms` ?

iuhilnehc-ynos:
@gezp 

> is it necessary to add a new item `use_dedicated_thread ` in `LoadNode.srv` ? what do you think about using `extra_arguments` in `LoadNode.srv` to load parameter `use_dedicated_thread ` like `use_intra_process_comms` (https://github.com/ros2/rclcpp/blob/master/rclcpp_components/src/component_manager.cpp#L155-L163) ?

From my point of view, I think that `use_dedicated_thread` is different from `use_intra_process_comms`, the former is just used inside `component_manager` to create a special thread, to set it with a parameter of Node seems a bit strange, the latter `use_intra_process_comms` is an option of Node, which could be a parameter.

I think that the user does not want to get  `use_dedicated_thread` info by `ros2 param list a_node`.

What do you think?

iuhilnehc-ynos:
In other words, if a user sets a parameter `use_dedicated_thread` with `true` for a node, and then adds the node into a SingleThreadedExecutor/MultiThreadedExecutor, does it run in a dedicated thread?

iuhilnehc-ynos:
@gezp

Sorry, I made a mistake. After reconsideration, I think it's OK to use `extra_arguments`.

gezp:
according to suggestions from @fujitatomoya @iuhilnehc-ynos , i load `use_dedicated_thread` parameter at runtime, in other word,  use `request->extra_arguments` to check if use dedicated thread for loaded node. please review again!

SteveMacenski:
@fujitatomoya @iuhilnehc-ynos Can you take another look? Zhengpeng made the requested changes:
- Node loaded per-node
- Uses `extra_arguments` in `LoadNode.srv` to specify
- Removes global parameter

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-10-28/22947/1

rclcpp_components/test/test_component_manager_api.cpp
```diff
@@ -211,6 +211,71 @@ TEST_F(TestComponentManager, components_api)
...
+      rclcpp::ParameterValue(true));
+    request->extra_arguments.push_back(use_dedicated_thread.to_parameter_msg());
+
+    auto future = composition_client->async_send_request(request);
+    auto ret = exec->spin_until_future_complete(future, 5s);  // Wait for the result.
+    auto result = future.get();
+    EXPECT_EQ(ret, rclcpp::FutureReturnCode::SUCCESS);
+    EXPECT_EQ(result->success, true);
+    EXPECT_EQ(result->error_message, "");
+    std::cout << result->full_node_name << std::endl;
```
fujitatomoya, line 251: nitpick: is this needed?

fujitatomoya (approved):
lgtm

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15571/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10255/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13239/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15790/)

rclcpp_components/test/test_component_manager_api.cpp
gezp: the way of unloading nodes which use dedicated thread is different of normal nodes, so i added this test code, what do you think about  it? @fujitatomoya .if we needn't it , i can remove it.
https://github.com/ros2/rclcpp/blob/0716a84f32bca9535cc690ae72bf78433ea61a78/rclcpp_components/src/component_manager.cpp#L282-L293

gezp:
i rebased this PR, please retrigger CI. @fujitatomoya

rclcpp_components/test/test_component_manager_api.cpp
fujitatomoya: i mean test case is needed but `std::cout << result->full_node_name << std::endl;`? sorry i wasnt clear before.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15573/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10257/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13241/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15793/)

rclcpp_components/test/test_component_manager_api.cpp
```diff
@@ -211,6 +211,71 @@ TEST_F(TestComponentManager, components_api)
...
```
iuhilnehc-ynos, line 230: remove this line of `std::cout`

iuhilnehc-ynos (commented):
@gezp 

Some minor comments, could you please check them?

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -233,6 +233,8 @@ class ComponentManager : public rclcpp::Node
 
 private:
   std::weak_ptr<rclcpp::Executor> executor_;
+  std::map<uint64_t, std::shared_ptr<rclcpp::Executor>> executors_;
```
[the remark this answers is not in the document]
gezp: i agree with you that 'dedicated_executors_'  is a better name, and we can't use `std::unique_ptr` for it , because it's also used in `executor_threads_`
https://github.com/ros2/rclcpp/blob/e626f978d7e838835d13663e8cf4534a7753ee3c/rclcpp_components/src/component_manager.cpp#L238-L241

rclcpp_components/src/component_manager.cpp
```diff
@@ -188,6 +196,17 @@ ComponentManager::on_load_node(
...
+      for (const auto & a : request->extra_arguments) {
+        const rclcpp::Parameter extra_argument = rclcpp::Parameter::from_parameter_msg(a);
+        if (extra_argument.get_name() == "use_dedicated_thread") {
+          if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
+            throw ComponentManagerException(
+                    "Extra component argument 'use_dedicated_thread' must be a boolean");
+          }
+          use_dedicated_thread = extra_argument.get_value<bool>();
+        }
+      }
```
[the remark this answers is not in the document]
gezp: yes, this is a duplicated code, but i think `std::map<std::string, rclcpp::Parameter>` may be a better choice to avoid multiple O(n) search, but it will consume more memory, what do you think about?

gezp:
thanks for your suggestions @fujitatomoya @iuhilnehc-ynos , i updated this PR.

rclcpp_components/include/rclcpp_components/component_manager.hpp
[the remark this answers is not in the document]
iuhilnehc-ynos: To pass the raw pointer (`executors_[node_id].get()`) into the lambda could work.

such as,
```cpp
        auto exec = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
        exec->add_node(node);
        executors_[node_id] = std::move(exec);
        executor_threads_[node_id] = std::make_shared<std::thread>([executor=executors_[node_id].get()]() {executor->spin();});
```

rclcpp_components/src/component_manager.cpp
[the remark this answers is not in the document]
iuhilnehc-ynos: > i think `std::map<std::string, rclcpp::Parameter>` may be a better choice to avoid multiple O(n) search, but it will consume more memory, what do you think about?

Yeah, I agree with you.

iuhilnehc-ynos (approved):
LGTM

fujitatomoya:
CI(Retry):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15615/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10296/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13283/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15831/)

SteveMacenski:
> CI(Retry):

Love that green :heart:

fujitatomoya:
@ivanpauno @wjwwood @jacobperron requesting maintainer review on this.

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -233,6 +233,8 @@ class ComponentManager : public rclcpp::Node
 
 private:
   std::weak_ptr<rclcpp::Executor> executor_;
+  std::map<uint64_t, std::shared_ptr<rclcpp::Executor>> dedicated_executors_;
+  std::map<uint64_t, std::thread> dedicated_threads_;
```
ivanpauno, line 237: do the maps need to be ordered?
It doesn't look like they need, use `std::unordered_map` in that case

ivanpauno (commented):
Adding this option seems like too much configurability to me.
I would rather have an specialized component manager in another repo for this use case or just use manual composition when needed.

I understand this can be handy though, so I'm happy with it if another maintainer approves.

I have left some minor comments in the implementation.

rclcpp_components/src/component_manager.cpp
```diff
@@ -54,9 +73,19 @@ ComponentManager::~ComponentManager()
...
       for (auto & wrapper : node_wrappers_) {
+        // if node is in dedicated executors, ignore it.
+        if (dedicated_executors_.find(wrapper.first) != dedicated_executors_.end()) {
+          continue;
+        }
         exec->remove_node(wrapper.second.get_node_base_interface());
       }
     }
+    for (auto & exec : dedicated_executors_) {
+      exec.second->cancel();
```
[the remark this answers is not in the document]
SteveMacenski: https://github.com/ros2/rclcpp/blob/5fcb7892f70bd741cf86b019fc489f3a81a60510/rclcpp_components/src/component_manager.cpp#L257-L258

It is added to `dedicated_executors_` right before being spun, so it should always have been called `spin` to be on that map. I suppose the only way to make it more bulletproof is if those 2 lines are swapped so that its only added to `dedicated_executors_` _after_ its been spinning.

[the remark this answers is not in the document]
ivanpauno: Swapping the lines doesn't help.

The issue is that though the thread is already spawned, you don't actually know when there will be a context switch and `spin()` will be actually called.
In an extremely unlucky scenario, the thread is spawn and an unload request is received before `spin()` is called, resulting in the cancellation request being lost.

the "right way" to do it is:

```cpp
std::promise<void> prom;
dedicated_threads_[node_id] = std::thread([exec, cancel_token=prom->get_future()]() {exec->spin_until_future_complete(cancel_token);});
dedicated_promises_[node_id] = std::move(promise);
```

```
-      dedicated_executor->second->cancel();
+      dedicated_promise->second->set_value();
```

(the code will need some more changes than that, that's just the idea)

there's no race possible in that case, as a future/promise channel is stateful.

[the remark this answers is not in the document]
SteveMacenski: Thanks for the specific suggestion!

jacobperron (commented):
I tend to agree with @ivanpauno; rather than introducing a new option, I think it would be better to create a different component container that uses this dedicated executor approach (e.g. then you don't necessarily need the extra argument `use_dedicated_thread`).

Technically, we can create a new executable where we pass a different executor to `ComponentManager` or create an extension of `ComponentManager` to use multiple single-threaded executors (like in this PR).

For example, we ended up creating a [custom container in the domain_bridge](https://github.com/ros2/domain_bridge/blob/348fd60576b841e2ed5fb24353b9205d738aad99/include/domain_bridge/component_manager.hpp) that extends `ComponentManager` to add a new option for setting the domain ID.

SteveMacenski:
Would this help to get on a call? As you can imagine, @gezp and I have been trying to get this in for 2 months and have responded to all requested changes by those with approver rights. It's frustrating when it finally has 2 approvals and is largely ready for merging (even run CI jobs) that we're getting this push back.

 This is not such a big thing to add as a configurable setting, I don't think. Any separate container would pretty much just look exactly like this PR and it seems silly to have that much boilerplate copied over. This is enabling running a given node on request to load its own executor vs a shared executor due to known performance issues with the multi-threaded executor preventing us and others from using it in any meaningful application. 

It has broad need in Nav2 and other projects as we've discussed in several venues already (TSC, [Discourse](https://discourse.ros.org/t/nav2-composition/22175), issue tracker, etc). I don't see how a configuration option is a bridge too far, most things in ROS are configurable.

fujitatomoya:
I am with @SteveMacenski on this.

- this option is not too much complicated, and generally i believe this is useful.
- i brought this issue to ROS 2 MW WG Meeting 2021-09-22, refer to https://docs.google.com/document/d/1pdPdcBk2YcR0IpmslDN07WGbhNU3WIaz-UBh2sDW_P0/edit#heading=h.xulamadrng2l. and there were no denials to have this option.
- seems that i am the one making this mess and confusion...sorry @gezp @SteveMacenski . but i am not sure what else i can/should do. (probably we should just wait for maintainer check 1st of all?)

i think we surely need to discuss on this contribution process in community.

rclcpp_components/src/component_manager.cpp
```diff
@@ -152,17 +181,6 @@ ComponentManager::create_node_options(const std::shared_ptr<LoadNode::Request> r
...
-    const rclcpp::Parameter extra_argument = rclcpp::Parameter::from_parameter_msg(a);
-    if (extra_argument.get_name() == "use_intra_process_comms") {
-      if (extra_argument.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
-        throw ComponentManagerException(
-                "Extra component argument 'use_intra_process_comms' must be a boolean");
-      }
-      options.use_intra_process_comms(extra_argument.get_value<bool>());
-    }
-  }
-
```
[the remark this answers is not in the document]
gezp: thanks for pointing this problem,  i agree with you that subclasses could be affected.  `use_intra_process_comms` and `use_dedicated_thread` both are also extra argument, so i wanted to avoid duplicated code by using map to search extra argument, but i didn't notice this problem.

rclcpp_components/src/component_manager.cpp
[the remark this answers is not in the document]
gezp: thanks for more details about how to implement, it's helpful to me.

SteveMacenski:
> i think we surely need to discuss on this contribution process in community.

Certainly, but I also don't want to let that block this work if we can get it pushed through. There should be a larger discussion about what gets into these core stacks. 

<details>
  <summary>Click to expand!</summary>

I've had issues where I was told "just make/host your own!" when the aim was to solve a general problem that nontrivial number of people are running into, or adding in a new feature that other people find generally valuable but is found to be DOA by maintainers. 

The point of something like ROS 2 is to have alot of these tools built into it to build applications. I can understand the desire not to increase maintenance burden, but often PRs are rejected fixing or adding new foundational capabilities that are actually really needed to build actual robot systems and not POC demos. Then we have people upvoting them or other trusted community members reviewing and giving their approvals just to be squashed months later by core maintainers. It's heartbreaking and even hitting a point where **I** question the value of even trying to contribute to `rclcpp` / `tf2` / `rviz2` / etc. 

I'd put this in the category, since I've heard now from several companies that manually compose their systems because this isn't available and the other 2 options are non-functional. Nav2 using manual composition in the way we do with individual executors for each node wasn't something we came up with out of thin air, it was the suggestion from another company that ran into the same issue, so we wanted to produce a generalized solution since its a clear and distinct need. 
</details>

ivanpauno:
> Would this help to get on a call? As you can imagine, @gezp and I have been trying to get this in for 2 months and have responded to all requested changes by those with approver rights. It's frustrating when it finally has 2 approvals and is largely ready for merging (even run CI jobs) that we're getting push back saying its DOA.

Yes, I agree that's not ideal.

> This is not such a big thing to add as a configurable setting, I don't think. Any separate container would pretty much just look exactly like this PR and it seems silly to have that much boilerplate copied over. It would even be sillier if Nav2 needed a specific nav2_component_container because the defaults don't work well enough out of the box for a modest application like Nav2

Yes, I think that the main issue here is that the multithreaded executor performance is really bad, and spawning multiple single threaded executors is working better.
If the multithreaded would work correctly, we would probably not need this (this approach can still be useful in some cases, but that's more niche).

> This is enabling running a given node on request to load its own executor vs a shared executor due to known performance issues with the multi-threaded executor preventing us and others from using it in any meaningful application. It has broad need in Nav2 and other projects as we've discussed in several venues already (TSC, Discourse, issue tracker, etc). I don't see that this configuration option is a bridge too far, most things in ROS are configurable shrug .

The thing I find a bit too far is that you can choose for each node whatever it's added to a separate executor or not.
I would be more happy to only having three options: one single threaded executor, one multithreaded executor, one single threaded executor for each node.

ivanpauno:
> i brought this issue to ROS 2 MW WG Meeting 2021-09-22, refer to https://docs.google.com/document/d/1pdPdcBk2YcR0IpmslDN07WGbhNU3WIaz-UBh2sDW_P0/edit#heading=h.xulamadrng2l. and there were no denials to have this option.

Ok, if there was concensous there accepting this sounds fine.

ivanpauno:
> The thing I find a bit too far is that you can choose for each node whatever it's added to a separate executor or not.

Instead of modifying the `ComponentManager` class, I would rather see a `OneThreadPerNodeExecutor` that's passed to the current component manager implementation.
I think that's a bit more clean, though unfortunately `Executor` is not a pure virtual class, which might make implementing `OneThreadPerNodeExecutor`a bit more complex.

i.e.: I think we can merge this with the open comments addressed ([this](https://github.com/ros2/rclcpp/pull/1781#discussion_r750626276) and [this](https://github.com/ros2/rclcpp/pull/1781#discussion_r751695276)).

fujitatomoya:
@ivanpauno thanks for sharing your thought!

> If the multithreaded would work correctly, we would probably not need this (this approach can still be useful in some cases, but that's more niche).

I thought this too, agree.

> Instead of modifying the ComponentManager class, I would rather see a OneThreadPerNodeExecutor that's passed to the current component manager implementation.

sounds good, so that we can keep `ComponentManager` simple.

> I think that's a bit more clean, though unfortunately Executor is not a pure virtual class, which might make implementing OneThreadPerNodeExecutora bit more complex.

also agree.

jacobperron:
Since we're working around an issue with the multi-threaded executor, it makes more sense to me to just replace it with the dedicated single-threaded executor approach. I.e. don't bother offer an extra argument to opt-in to the new implementation, rather, just change the implementation. Thoughts?

jacobperron:
What I'm really try to figure out is if/how we can make the default behavior better instead of adding a option to opt-in.

SteveMacenski:
> Yes, I think that the main issue here is that the multithreaded executor performance is really bad, and spawning multiple single threaded executors is working better.

That is true, but also I like that each one acts independently, even if it had the same performance characteristics as N single threaded executors. I don't have to worry about async issues of `N` nodes in a common thread pool who each may have multiple subscriptions/services/actions that otherwise would never be called at the same time.  This is more isolated and I'd think a better solution to having 10, 20, 30 nodes in the same process doing arbitrary things. Personally, this would be my preference outside of resource limited situations.

>  I would be more happy to only having three options: one single threaded executor, one multithreaded executor, one single threaded executor for each node.

I'd be perfectly happy with that as well. I believe some amount of rearchitecture of the core container code would be necessary to entirely support cleanly. (Opinion) The most obvious solution to me is we'd have to change the constructor from being given an executor into the entire class being templated by the `ExecutorT` to spawn multiple. Then have a new manager class for overriding some functions for spawning multiple on request (which then could work with any future executor types as well). The actual code changes I could do in a day. If that sounds good to you and we want to file a ticket around it, I could commit to doing that early in 2022.

I don't think a new executor type would be as good, but I also don't have a design idea in mind for how that would work, you guys are definitely more familiar with that part of the code. The biggest concern I would have is if that's the right place to solve the problem with the separation of concerns being in component context, not execution models of arbitrary nodes.  

> Since we're working around an issue with the multi-threaded executor, it makes more sense to me to just replace it with the dedicated single-threaded executor approach. I.e. don't bother offering an extra argument to opt-in to the new implementation, rather, just change the implementation. Thoughts?

I think that would require more community discussion to see if/how people are using this today and if that would be majorly disruptive. I think the only people that would not want this are people on very resource constrained systems in more simplistic applications, but that's still a tangible niche. I think that part of the design discussion could be had in parallel to the work I suggested above?

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-11-18/23209/1

jacobperron:
Apologies, I was unaware of the existing consensus from the WG.

I'm fine with having an option to change the execution model, however it seems odd to me that the option is per node being added to the container. It seems more straightforward if the option was for the container process itself. I.e. when starting the container process we can select to either use the multi-threaded executor or use a set of dedicated single-threaded executors. I think moving the option to the container adds a bit of encapsulation that makes changing behavior in the future easier to do.

SteveMacenski:
So what are the next steps here? I see the options as:
- Merge this as is and continue the design discussion for an "isolated" executor container (1 executor per component) & the "centralized" executor container (current behavior) in another ticket which I'd plan on implementing
- Wait on this PR to be updated with that 
- Merge this as is and shrug off the quirk, I can see some value in having this "mixed" executor container, but I agree its probably a niche of a niche and users of it will probably be doing all "isolated" or all "centralized". 

Whatever we do, I'd just want buy in between @jacobperron and @ivanpauno on that action so that once its executed we can move forward.

ivanpauno:
I would prefer to see this PR updated to having an "isolate executor container".
i.e. having an option for the container process instead of one per node.

> I can see some value in having this "mixed" executor container

That one sounds to niche to me.
There might also be an use case of wanting to add some callback groups of a node to one executor and some other callback groups of the same node to another executor, etc.
Currently that can only be done by using manual composition.

jacobperron:
I agree with Ivan. Let's update this PR to move the option for the new behavior to the container process.

@gezp @SteveMacenski Thanks for iterating :slightly_smiling_face:

gezp:
>  I would prefer to see this PR updated to having an "isolate executor container"

i also agree with you @ivanpauno , it's more clear. i want to ask some suggestions about implementation, do you mean we need pass a bool argument `use_dedicated_executor` to `ComponentManager` constructor, then create a new type container? or just load `use_dedicated_executor` from ros parameter of container?

SteveMacenski:
Its a half baked idea, but I think this would work out -- let me know what you guys think.

- Make the `ComponentManager`  templated by `ExecutorT` rather than taking in an instance in the constructor
- Make the default `ComponentManager` instantiate a single `ExecutorT` that is shared by the `N` components
- Make a new `ComponentManagerIsolated` (?) I'm terrible at naming, don't use that, but just consistent with nomenclature I used before above. This should derive from `ComponentManager`
- Overload the functions `on_load_node` and `on_unload_node` to instantiate the individual `ExecutorT` instances per node
- At this point, this is where it gets fuzzy. I'm look over the copy-pasted code I pulled out form `on_load_node` and `on_unload_node` and see how much of it I could abstract out into separate functions within `ComponentManager` to call so that they share most of the same code and this manager only has the minimum amount of duplicated code as reasonably, and cleanly, possible
- Create the component container instances to support the additional 2 `ComponentManagerIsolated`'s for Single and Multithreaded executors.

ivanpauno:
> do you mean we need pass a bool argument use_dedicated_executor to ComponentManager constructor, then create a new type container? or just load use_dedicated_executor from ros parameter of container?

Currently we have a `component_container` and a `component_container_mt` executable, so I think that adding a third executable `component_container_isolated` (name can be bikesheded later) makes sense.
I wouldn't use a ros parameter.

ivanpauno:
> Its a 3/4 baked idea, but I think this would work out -- let me know what you guys think.

I haven't looked too much into the details, but that seems to make sense.

> Create the component container instances to support the additional 2 ComponentManagerIsolated's for Single and Multithreaded executors.

We could create both, but I think that the one using the single threaded executor is the most useful.

SteveMacenski:
> We could create both, but I think that the one using the single threaded executor is the most useful.

Agreed, I figure it better to be 'complete' though and its only an extra 6 lines of code to provide it out of the box initially.

fujitatomoya:
> Currently we have a component_container and a component_container_mt executable, so I think that adding a third executable component_container_isolated (name can be bikesheded later) makes sense.

I am also good to go with this.

gezp:
i use a simple way to implement isolate executor container, which only support dedicated single threaded executor. 

accoding to @SteveMacenski suggestions, `ComponentManagerIsolated` derived from `ComponentManager` maybe a better choice to support for single and multi threaded executors, but there are some duplicated code, i'm not sure whether to use this approach, i would like to hear your opinions.

in addition, i use `promise` and `spin_until_future_complete()` instead of `executor->cancel()`, but it can't work corrently, i don't know why. so i use both of  `prom.set_value()` and `executor->cancel()` to cancel executor. could you help me to solve this problem? @ivanpauno

SteveMacenski:
> ComponentManagerIsolated derived from ComponentManager maybe a better choice to support for single and multi threaded executors, but there are some duplicated code, i'm not sure whether to use this approach, i would like to hear your opinions.

Its hard to tell without seeing it. It sounds like you would only need to change `on_load_node` and `on_unload_node` which are both virtual so you can override them. `on_unload_node` looks like a trivial amount of code (< 15 functional lines) and is all executor-model-specific so I think that's fine to copy paste over and modify for the `N` executors. 

`on_load_node` seems like a bit more code. The functional part that would need to be changed looks like https://github.com/ros2/rclcpp/blob/master/rclcpp_components/src/component_manager.cpp#L217-L220. You could easily create a new, virtual function called something like `add_node_to_executor()` or something who's implementation is just these lines. Then in your implementation, you can override it to set it to use the `N` executors. 

I think that sounds the cleanest to me. You'll obviously need a little bit of code in the constructor / destructor as well, but little to any duplication if you call the base class constructor / destructors' to handle their own resources. 

What do you think?

gezp:
i updated PR.
* create `ComponentManagerIsolated` derived from `ComponentManager` ,  and override `on_load_node` and `on_unload_node`.
* create a new type of container `component_container_isolated` by using `ComponentManagerIsolated`.

need  `ComponentManagerIsolated`  be templated by `ExecutorT` to support both of single and multi threaded executors？then, a new type of container `component_container_mt_isolated` should be created to support multi threaded executors, right?

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
```diff
@@ -0,0 +1,61 @@
...
+  void
+  on_unload_node(
+    const std::shared_ptr<rmw_request_id_t> request_header,
+    const std::shared_ptr<UnloadNode::Request> request,
+    std::shared_ptr<UnloadNode::Response> response) override;
+
+private:
+  std::unordered_map<uint64_t, std::shared_ptr<rclcpp::Executor>> dedicated_executors_;
+  std::unordered_map<uint64_t, std::thread> dedicated_threads_;
+  std::unordered_map<uint64_t, std::promise<void>> dedicated_promises_;
```
ivanpauno, line 56: Maybe you can create a custom struct for the three values (executor, thread, promise) and create only one unordered map.
You could also use a tuple instead of defining a custom struct (defining a custom struct makes the code more readable than using tuples).

rclcpp_components/src/component_manager_isolated.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+    for (auto & t : dedicated_threads_) {
+      t.second.join();
+    }
+    node_wrappers_.clear();
+  }
+}
+
+void
+ComponentManagerIsolated::on_load_node(
```
ivanpauno, line 41: Some of the code here is repeated from the base class and could be shared with it.
Same in on_unload_node().

Anyways the approach seems reasonable!

ivanpauno: Feel free to ignore this comment

rclcpp_components/test/test_component_manager_api.cpp
```diff
@@ -321,3 +327,9 @@ TEST_F(TestComponentManager, components_api)
     }
   }
 }
+
+TEST_F(TestComponentManager, components_api)
+{
+  test_components_api(false);
```
ivanpauno, line 333: It's nicer to use a parameterized test for this, see https://github.com/google/googletest/blob/main/docs/advanced.md#value-parameterized-tests.

If you feel that's an overkill here, add [scoped traces](https://github.com/google/googletest/blob/main/docs/advanced.md#adding-traces-to-assertions) so it's more clear where the errors come from.
You could also return an [AssertionResult](https://github.com/google/googletest/blob/main/docs/advanced.md#adding-traces-to-assertions) to avoid [this pitfall](https://github.com/google/googletest/blob/main/docs/advanced.md#propagating-fatal-failures).

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
```diff
@@ -0,0 +1,61 @@
...
+{
+  using rclcpp_components::ComponentManager::ComponentManager;
+
+public:
+  ~ComponentManagerIsolated();
+
+protected:
+  RCLCPP_COMPONENTS_PUBLIC
+  void
+  on_load_node(
```
SteveMacenski, line 42: You'll want to add documentation here like you see in the other .hpp files for auto-doc generation

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
SteveMacenski: Agreed. A custom struct is the cleanest way

SteveMacenski:
> need ComponentManagerIsolated be templated by ExecutorT to support both of single and multi threaded executors

I would like that, yes. I think that would make this future proof. This may require making the normal `ComponentManager` to also be templated and remove the constructor argument, which I think is also a good move. You can simply instantiate `ExecutorT` in the constructor -- that way we can use the same API for everything.

> then, a new type of container component_container_mt_isolated should be created to support multi threaded executors, right?

:+1:

rclcpp_components/src/component_manager_isolated.cpp
```diff
@@ -0,0 +1,153 @@
...
+  const std::shared_ptr<rmw_request_id_t> request_header,
+  const std::shared_ptr<LoadNode::Request> request,
+  std::shared_ptr<LoadNode::Response> response)
+{
+  (void) request_header;
```
SteveMacenski, line 46: This entire code block L46-L87 and L103-114 are exactly the same as in the other container. 

What you could do instead is what I suggested above. Put the things in L88-L102 into a separate, virtual function with the implementation that is overrided. Then you won't even need to override or implement `on_load_node`. This is really good not to duplicate code + makes it so any change to the main container are automatically applied here. 

If you did that too, it would resolve Ivan's comment above on duplication as well, because you wouldn't have `on_load_node` at all.

SteveMacenski (changes requested):

rclcpp_components/src/component_container_isolated.cpp
```diff
@@ -0,0 +1,33 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
jacobperron: Technically, the copyright can remain OSRF, and there may be (legal) reasons to prefer leaving it as-is for both the contributor and OSRF. But, there's precedence for allowing additional copyright holders in ROS 2 source code. I would defer to @clalancette and/or @tfoote for clarity.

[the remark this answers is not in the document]
wjwwood: It's certainly not required to put your own name there, but it's also totally fine as long as the changes to the file are non-trivial or it's a new file (which is the case here).

So, putting yourself there is ok, and leaving it as-is is also ok. :)

gezp:
i updated PR.
*  `ComponentManagerIsolated` is templated by `ExecutorT` to support both of single and multi threaded executors
*  add a new type of container `component_container_mt_isolated`
*  add new virtual functions  `add_node_to_executor()` and `remove_node_from_executor()` to avoid duplicated code.

please review again @SteveMacenski @ivanpauno .

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+#include <unordered_map>
+
+#include "rclcpp_components/component_manager.hpp"
+
+
```
SteveMacenski, line 28: Its up to core maintainers (@ivanpauno ) but I would rather the other containers follow suit with the `ExecutorT` so that you can remove the constructor field. The template for type + executor constructor argument are redundant. You can have the same effect by removing all constructor arguments from the existing `component_manager.hpp` and making that also templated, which constructs its internal executor by instantiating `ExecutorT()`.

SteveMacenski: Also, needs the function documentation inline (see component_manager.hpp, you added there, just add here too)

SteveMacenski:
This looks really good to me. This has my stamp of approval after the inline `hpp` method docs are added

rclcpp_components/src/component_container_isolated.cpp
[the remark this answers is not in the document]
clalancette: > So, putting yourself there is ok, and leaving it as-is is also ok. :)

Yes, exactly.  Since this is a non-trivial amount of new work, it is perfectly acceptable to make yourself the copyright holder.  In that case, you would change the line to be `// Copyright 2021 <my name here>`.  On the other hand, it is also perfectly fine to "assign" the copyright to the Open Source Robotics Foundation, which gives the foundation most of the legal power over the code.  Totally up to you which way you want to go.

(the usual disclaimer of "I am not a lawyer" and "this is not legal advice" applies here)

rclcpp_components/CMakeLists.txt
```diff
@@ -61,9 +61,29 @@ ament_target_dependencies(component_container_mt
...
+  component_container_isolated
+  src/component_container_isolated.cpp
+)
+target_link_libraries(component_container_isolated component_manager)
+ament_target_dependencies(component_container_isolated
+  "rclcpp"
+)
+
+add_executable(
+  component_container_mt_isolated
```
ivanpauno, line 74: Though adding this is a couple lines of code, I think it's better to not add it to avoid confusion.

My reasons are:
- The more options, the more confusing which one is the best.
- The multithreaded executor doesn't execute anything in parallel, except if multiple nodes were added to it or if there're a node with more than one callback group (most nodes are just using the default callback group only).
  In this case, we would be launching one executor per node, which means that if the node is only using the default callback group the `MultithreadedExecutor` will perform worse than the single threaded one (multiple threads doing completely mutually exclusive work).
- Even in the case of nodes having multiple callback groups, creating more than one multithreaded executor will end up in more threads that cores, which usually doesn't perform well either.

ivanpauno: There's a fourth reason:

- The multithreaded executor is only taking advantage of the multiple threads in `spin()`, and not in the other methods like `spin_until_future_complete()`.

Though that's something we should improve ...

ivanpauno (approved):
LGTM!

I would prefer to see the `component_container_mt_isolated` removed before merging.
See my other comment.

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
ivanpauno: Yes, we're not taking any advantage of the template parameter here.
We can delete it or be consistent in the base class, I would just remove it as it doesn't seem to be needed.

(with the template parameter we could've not added the new virtual methods, but both approaches are fine)

rclcpp_components/CMakeLists.txt
SteveMacenski: >  a node with more than one callback group

There are definitely real examples of more complex nodes that do have multiple callback groups that could make use of this multi-threaded executor container where each can have an isolated executor. It's the same rationale for why we want the single threaded version of this -- we want them to be in the same process but be `isolated` from each other to prevent unintended consequences of callback groups that were not designed to operate at the same time, operating at the same time. But an application that can leverage this container type is designed for internally multiple callback groups being executed so it makes perfect sense. 

Some some simplistic application, sure, but isn't the point of ROS 2 to be supporting real, deployed, complex applications that ROS 1 could not? This is part of that, IMO. 

I do not believe this is the point of too much confusion. I'd be confused as a user if the multi-threaded version of this didn't exist - because it would make me ask "why not when there's a non-isolated `mt` version?". 

The only reason I think that we should not have an `mt-isolated` container is if we depreciate the normal `mt` container as well so there is no `mt` component containers at all.

wjwwood: > It's the same rationale for why we want the single threaded version of this -- we want them to be in the same process but be isolated from each other to prevent unintended consequences of callback groups that were not designed to operate at the same time, operating at the same time.

Sorry, I haven't been able to follow everything here, but I was skimming and saw this, and it stood out to me.

If two callbacks were designed to not be run at the same time, then they should be in the same mutually-exclusive callback group. If they're in different nodes, then something really bad has gone wrong in the design. Anything else and they could be run at the same time.

A node that by design requires a dedicated single-threaded executor to address this kind of issue is, in my opinion, a bad solution. If that is a requirement, then the node should instead create a thread and have its own single-threaded executor for those sensitive callbacks. That way it is not negatively impacted by which executor it is added to in a container. The downside to that approach is scaling, having every node have its own thread and executor will not scale as well, but used judiciously it's fine.

Now if the issue is just congestion, like two callbacks from two different nodes can in principle share a multi-threaded executor, but in practice one starves the callbacks of the other, then this approach can be fine I think. So I'm not opposed to having this as an option.

To reduce confusion (and again I haven't reviewed the changes so I don't know how feasible this is), it might be better to simply have CLI options (or ros param options) for the single container to control which executors it supports.

SteveMacenski: >  If they're in different nodes, then something really bad has gone wrong in the design. 

I don't disagree. That was probably not a good illustration. The congestion example you provide is a much better example of why one might want to isolate components with different multi-threaded executors -- so that one doesn't hog the entire pool. 

> To reduce confusion (and again I haven't reviewed the changes so I don't know how feasible this is), it might be better to simply have CLI options (or ros param options) for the single container to control which executors it supports.

I'd be fine with that. Then, shouldn't we also apply that same option to the normal component container for consistency and remove the existing `mt`?

gezp:
according to suggestions from @ivanpauno , @wjwwood , @SteveMacenski , i remove `component_container_mt_isolated`,  i also provide a  CLI options `--use_multi_threaded_executor`  to support  isolated `multi_threaded_executor`.

rclcpp_components/CMakeLists.txt
ivanpauno: > Sure... but that's clearly subject to later improvements - but my argument above still applies: ROS 2 is built for real commercial applications that are going to be complex.

My proposal of not having a "multithreaded" option in the isolated container is because it's a really unusual approach in any real application.
The most usual approaches are:

- Single threaded event loop. This might be combined with a thread pool of workers to handle blocking tasks (usually, the event loop has some way to interact with tasks in the thread pool to make the combination nice). e.g. python asyncio runtime
- One single threaded event loop per core. This might also be combined with a thread pool of workers to handle blocking tasks. e.g. boost asio can be configured this way, some rust runtimes, etc
- One multithreaded event loop. This might be combined with a thread pool of workers as well. e.g. boost::asio can be configured this way, default behavior of most rust runtimes, etc.

The "single threaded executor" is kind of equivalent to a single threaded event loop.
The "multi threaded executor" is kind of equivalent to a multi threaded event loop.
The "isolated executor per node" is kind of equivalent to having one event loop per core. This approach doesn't scale well if you have many nodes (because you will have much more threads than cores).

The "isolated multithreaded executor per node" is not equivalent to anything you usually see in event loop runtimes.
It really doesn't scale well, as you create as many threads as cores per each multithreaded executor (by default), and you  easily end up having a lot of threads.

---

It's only 5 extra lines of codes and the discussion doesn't seem to be making much progress, so I will accept this as-is.

ivanpauno (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15731/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10413/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13396/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15976/)

ivanpauno:
@gezp this seems to need a rebase to make CI pass, thanks!

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15732/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10414/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13397/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15977/)

ivanpauno:
> I approve this, though I would feel better if the normal container also contained the multi-threaded option and removed the separate container, but I leave it to maintainers.

That seems fine to me.
We could also have only one executable with two options (single/multi threaded, isolated/normal).

We should check the impact of removing the executable, I guess some launch files will break.
If that's the case, we should probably hava a deprecation cycle before removing the existing `component_container_mt` executable.

I would do that in a new PR instead of here, as deprecating/removing an executable will likely require changes in other core repos, which will block this PR further.

ivanpauno:
Thanks @gezp for the contribution!!

Blast545:
:man_farmer: I think this PR broke our Windows CI builds:
See: https://ci.ros2.org/job/ci_windows/15992/
https://ci.ros2.org/view/nightly/job/nightly_win_deb/2198/

The error:
<details>

```
03:30:50   Building Custom Rule C:/ci/ws/src/ros2/rclcpp/rclcpp_components/CMakeLists.txt

03:30:50   test_component_manager.cpp

03:30:50   test_component_manager.vcxproj -> C:\ci\ws\build\rclcpp_components\Debug\test_component_manager.exe

03:30:50   Building Custom Rule C:/ci/ws/src/ros2/rclcpp/rclcpp_components/CMakeLists.txt

03:30:50   test_component_manager_api.cpp

03:30:50 test_component_manager_api.obj : error LNK2001: unresolved external symbol "protected: virtual void __cdecl rclcpp_components::ComponentManagerIsolated<class rclcpp::executors::SingleThreadedExecutor>::add_node_to_executor(unsigned __int64)" (?add_node_to_executor@?$ComponentManagerIsolated@VSingleThreadedExecutor@executors@rclcpp@@@rclcpp_components@@MEAAX_K@Z) [C:\ci\ws\build\rclcpp_components\test_component_manager_api.vcxproj]

03:30:50 test_component_manager_api.obj : error LNK2001: unresolved external symbol "protected: virtual void __cdecl rclcpp_components::ComponentManagerIsolated<class rclcpp::executors::SingleThreadedExecutor>::remove_node_from_executor(unsigned __int64)" (?remove_node_from_executor@?$ComponentManagerIsolated@VSingleThreadedExecutor@executors@rclcpp@@@rclcpp_components@@MEAAX_K@Z) [C:\ci\ws\build\rclcpp_components\test_component_manager_api.vcxproj]

03:30:50 C:\ci\ws\build\rclcpp_components\Debug\test_component_manager_api.exe : fatal error LNK1120: 2 unresolved externals [C:\ci\ws\build\rclcpp_components\test_component_manager_api.vcxproj]

03:30:50 ---
03:30:50 Failed   <<< rclcpp_components [1min 47s, exited with code 1]
```
</details>

This PR was modifying `rclcpp_components` so CI should have been run up to that package instead of `rclcpp`.
Can you guys take a look to the problem? @gezp @ivanpauno

Issue #1190, opened 2020-06-23
Modify Parameters On Set Parameter Callback


threeal:
### Feature Description

Currently the [rclcpp::Node::add_on_set_parameter_callback()](http://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Node.html#a0c3ad2ca5c56eb54af0227f04e943679) only support callback of `rcl_interfaces::msg::SetParametersResult callback(const std::vector<rclcpp::Parameter> &parameters)`. I see that there is a point to make the parameters constant, so the `/node/set_parameters` service could return the result of the set parameters to be either success or not.

But considered the following scenario, i have a parameter that only accept the double value of `0.0` to `100.0`. if i set the parameter with double value between `0.0` to `100.0`, then it would be fine. but what if i set the parameter to be `101.0`, using the current `on_set_parameter_callback()` behavior, then i could only reject it. wouldn't it be better if i could accept it in condition that the parameter value is clamped to `100.0`? or what if i set the parameter using integer value of `50`? when i accept it, the parameter would simply changed from double to integer, and it would be considered bad. But i shouldn't reject it either. Wouldn't it be better if i could accept it in condition that the parameter type to be casted to double value?.

### Implementation Considerations

Add a new callback type for `rclcpp::Node::add_on_set_parameter_callback()` with the type of `rcl_interfaces::msg::SetParametersResult callback(std::vector<rclcpp::Parameter> &parameters)`.

For note, I'm already tried to solve the problem with const casting the `parameters` to `std::vector<rclcpp::Parameter> &`. It worked, but kinda be a tricky solution and a bad habit as the `parameters` was implicitly stated to be constant.

clalancette:
> But considered the following scenario, i have a parameter that only accept the double value of `0.0` to `100.0`. if i set the parameter with double value between `0.0` to `100.0`, then it would be fine. but what if i set the parameter to be `101.0`, using the current `on_set_parameter_callback()` behavior, then i could only reject it. wouldn't it be better if i could accept it in condition that the parameter value is clamped to `100.0`?

I'm not following what feature you are looking for here.  Can you explain more?

With the current callback, it is entirely possible to clamp a value as you set it.  You just use a `std::clamp` as you set the value in the callback.  As for setting the range between 0 and 100.0; to me, it seems non-sensical to set a range, and then to ignore that range and clamp it.  That is, if you want to accept *any* value and clamp it, then don't set a range.

[...]

PR #1789, opened 2021-09-28
Set parameters from inside callbacks 

adityapande-1995:
This PR aims to add a way to set parameters from inside another parameter's callback. 

## Current approach : 
* Maintain a separate vector container for callbacks passed using the ``allowRecursion`` boolean flag to ``add_on_set_parameters`` method.
* Maintain a loop counter and a context tracker to allow only recursive container callbacks to be called when ``set_parameters_atomically`` is called from inside another parameter's callback.

## TODO for current approach : 
* Manage loop counter and recursive access to ``add_on_set_parameters``, ``remove_on_set_parameters``, ``undeclare_parameters``
* Clean up code, remove debug prints

## Concerns
* Complicated locking and overall approach 
* Thread safety
* Order in which callbacks are run can be difficult to understand for the user, which may lead to unexpected behavior of parameter actions

## Suggested alternatives
* Have a new action called ``update_parameters`` which would update the parameters without triggering their callbacks. This would not solve the overall problem exactly, but would be much simpler to debug and understand.
* Maybe have this action inside ``get_parameters`` ?

## Original model :

image

## Current approach : 

image

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -929,7 +929,7 @@ public:
   RCLCPP_PUBLIC
   RCUTILS_WARN_UNUSED
   OnSetParametersCallbackHandle::SharedPtr
-  add_on_set_parameters_callback(OnParametersSetCallbackType callback);
+  add_on_set_parameters_callback(OnParametersSetCallbackType callback, bool allowRecursion = false);
 
   /// Remove a callback registered with `add_on_set_parameters_callback`.
   /**
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
@@ -39,6 +39,8 @@
 #include "rclcpp/publisher.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+#include <iostream>
+
 namespace rclcpp
 {
 namespace node_interfaces
@@ -61,6 +63,7 @@ public:
   explicit ParameterMutationRecursionGuard(bool & allow_mod)
   : allow_modification_(allow_mod)
   {
+    std::cout << "Locking mutation guard" << std::endl;
     if (!allow_modification_) {
       throw rclcpp::exceptions::ParameterModifiedInCallbackException(
               "cannot set or declare a parameter, or change the callback from within set callback");
@@ -71,6 +74,7 @@ public:
 
   ~ParameterMutationRecursionGuard()
   {
+    std::cout << "Unlocking mutation guard" << std::endl;
     allow_modification_ = true;
   }
 
@@ -193,7 +197,7 @@ public:
   RCLCPP_PUBLIC
   RCUTILS_WARN_UNUSED
   OnSetParametersCallbackHandle::SharedPtr
-  add_on_set_parameters_callback(OnParametersSetCallbackType callback) override;
+  add_on_set_parameters_callback(OnParametersSetCallbackType callback, bool allowRecursion) override;
 
   RCLCPP_PUBLIC
   void
@@ -218,6 +222,8 @@ private:
   OnParametersSetCallbackType on_parameters_set_callback_ = nullptr;
 
   CallbacksContainerType on_parameters_set_callback_container_;
+
+  CallbacksContainerType on_parameters_set_callback_container_recursive_;
 
   std::map<std::string, ParameterInfo> parameters_;
 
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
@@ -57,29 +57,48 @@ struct ParameterInfo
 };
 
 // Internal RAII-style guard for mutation recursion
+// class ParameterMutationRecursionGuard
+// {
+// public:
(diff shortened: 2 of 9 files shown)


adityapande-1995:
Example usage : 
```
#include "rclcpp/rclcpp.hpp"
#include <iostream>
class MyNode : public rclcpp::Node
{
public:
    MyNode() : Node("my_node_name")
    {
        auto param_change_callback_1 =
        [this](std::vector<rclcpp::Parameter> parameters) {
          auto result = rcl_interfaces::msg::SetParametersResult();
          result.successful = true;

          for (const auto & parameter : parameters) {
            std::string param_name = parameter.get_name();
            std::cout << "cb1 -- dwnstream cb param name:" << param_name <<std::endl;
            if (param_name == "my_int2") {
              std::cout << "dwnstream: Trying to set my_int1" << std::endl;
              //this->set_parameter(rclcpp::Parameter("my_int1", 5.0));
            }

        }
        return result;
      };
        
        auto param_change_callback_2 =
        [this](std::vector<rclcpp::Parameter> parameters) {
          auto result = rcl_interfaces::msg::SetParametersResult();
          result.successful = true;

          for (const auto & parameter : parameters) {
            std::string param_name = parameter.get_name();
            std::cout << "cb2 -- dwnstream cb param name:" << param_name <<std::endl;
        }
        return result;
      };

      this->callback_handle_1 = this->add_on_set_parameters_callback(param_change_callback_1);
      this->callback_handle_2 = this->add_on_set_parameters_callback(param_change_callback_2, true);

      this->declare_parameter("my_int1", 0);
      this->declare_parameter("my_int2", 0);
    }
private:
    rclcpp::Node::OnSetParametersCallbackHandle::SharedPtr callback_handle_1;
    rclcpp::Node::OnSetParametersCallbackHandle::SharedPtr callback_handle_2;
};


int main(int argc, char **argv)
{
    rclcpp::init(argc, argv);
    auto node = std::make_shared<MyNode>();
    rclcpp::spin(node);
    rclcpp::shutdown();
    return 0;
}
```

rclcpp/include/rclcpp/node.hpp
```diff
@@ -929,7 +929,7 @@ class Node : public std::enable_shared_from_this<Node>
   RCLCPP_PUBLIC
   RCUTILS_WARN_UNUSED
   OnSetParametersCallbackHandle::SharedPtr
-  add_on_set_parameters_callback(OnParametersSetCallbackType callback);
+  add_on_set_parameters_callback(OnParametersSetCallbackType callback, bool allowRecursion = false);
```
ivanpauno, line 932: I don't like the idea of having an `allowRecursion`, we should either always support recursion or not at all.
nit

```suggestion
  add_on_set_parameters_callback(OnParametersSetCallbackType callback, bool allow_recursion = false);
```

ivanpauno (changes requested):
Could you explain what is the current behavior when calling `set_parameter()` from a parameter callback and how does this PR solve the issue?

I took a quick look and the approach taken here seems to be fairly complicated.

ivanpauno:
> Could you explain what is the current behavior when calling set_parameter() from a parameter callback and how does this PR solve the issue?

Is the current issue iterator invalidation here?

https://github.com/ros2/rclcpp/blob/9e445bdb910503cd16f3ccf57382e01169738cc3/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L274-L284

adityapande-1995:
@ivanpauno curently, if you call ``set_parameter()`` from inside a callback, the ``ParameterMutationRecursionGuard`` does not allow this, and throws a runtime exception saying that this operation is not allowed from inside a callback. As discussed with @wjwwood and @clalancette , we want to enable this functionality, to make sort of "meta-parameters" that can trigger other parameter actions. Of course this can lead to an infinite recursion (where parameters can call each other in a circular manner) and users can mess up things easily. Thus we wanted an explicit flag, or a new callback group container that would hold such callbacks which can be used from inside other callbacks. I've also added a ``loop_counter`` to keep track of the recursion depth and exit if we exceed it.

We **could** consider removing the  ``ParameterMutationRecursionGuard`` entirely, but the users would have no way of explicitly knowing that they are risking infinite recursion.

ivanpauno:
The "on set parameter" callbacks doesn't seem to be a good for this though.
IIUC they are supposed to only be checking if the parameter is going to be accepted or rejected, and not to have side-effects.
The issue with having side-effects in the callback is that the parameter might end up not being set, because another callback rejected it.

> As discussed with @wjwwood and @clalancette , we want to enable this functionality, to make sort of "meta-parameters" that can trigger other parameter actions

Could you give an example of where this is needed? That might help me to understand the issue in a better way.

adityapande-1995:
After a discussion with @jacobperron @clalancette @audrow @mjeronimo @ivanpauno and looking at https://github.com/ros2/rclcpp/issues/1190, https://github.com/ros2/rclcpp/issues/609, it was decided that https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1312 would use Parameter Event handler to reflect the true values of the slip parameters, and the ideal solution in ``rclcpp`` would be to have a self callback in the node, which would trigger when a parameter is changed. This should **not** be a validation callback.

Another major issue with this PR was that the callbacks are meant to be validation callbacks, and setting parameters inside them is not a valid use case. If one of the ``set_parameter`` action in the entire chain fails, then first callback that started the entire chain would fail as well.

Issue #731, opened 2019-05-21
fixup parameters and sub-nodes


wjwwood:
This is a follow up from https://github.com/ros2/rclcpp/pull/495#issuecomment-485906580.

Currently how parameters and "sub-nodes" interact is not documented and is not consistent or tested. That needs to be documented, audited, and tested.

mikeferguson:
So, is this ticket basically tracking that sub nodes do not actually use the sub_namespace in the name of the parameter? For example:

```
auto subnode = parent->create_sub_node("controller_name");
std::string controller_type = subnode->declare_parameter<std::string>("type", "");
```

I would expect the parameter to be "controller_name/type", but instead "ros2 param list" shows only a parameter named "type"? Is this expected behavior? Is this a bug?

fujitatomoya:
I am not sure about this either, I'd like to know.

> I would expect the parameter to be "controller_name/type", but instead "ros2 param list" shows only a parameter named "type"?

currently, yes. only `type` is declared w/o namespace.

I think that this is bug, it should be extended with sub_namespace?

wjwwood:
Yeah @mikeferguson I consider it a bug, or at least I think it will be surprising to most users.

The issue (why I didn't just fix it on the spot) is that the way the node class is separated into parts complicates this. The subnode feature was added to `Node`, but not to one of the components, meaning it's hard to use from the `NodeParameters` class. This is just my recollection of the issue, but maybe it's easier to fix now, I don't know off hand without trying to do it.

jlack1987:
@wjwwood could you recommend an acceptable way of fixing this or point me in the right direction? I'm very interested in fixing this issue and am willing to work on it

jlack1987:
I'm willing to spend some time attempting to fix this issue, but i'd appreciate it if I could get a more involved devs input on a suggested fix/implementation. I can just run off and implement a fix, but I don't want to end up spending a lot of time on a fix and it not being an acceptable implementation

fujitatomoya:
just sharing the update. we had a quick chat on this in MW WG. we consider this is bug, all entities of sub-nodes should be extended and propagated with sub-nodes namespace. (topics, services and parameters) as @wjwwood mentioned https://github.com/ros2/rclcpp/issues/731#issue-446871224 these should be tested well.

jlack1987:
[...]

Issue #1778, opened 2021-09-16
Spinning node does not also spin sub nodes


jlack1987:
## Bug report
I'm assuming this is a bug so i'm filing as so but if this is intended behavior then perhaps this is a feature request.
**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - foxy
- DDS implementation:
  - whatever is default
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Here's some test code i've been playing with. It's nothing great but it should show the behavior. With this what I expect is for each sub node to have its own topics/services/params namespaced by the name of the sub node but that's not what happens.

```
#include <chrono>
#include <string>

#include <rclcpp/rclcpp.hpp>
#include <std_msgs/msg/int32.hpp>
#include <std_srvs/srv/trigger.hpp>

using namespace std::chrono_literals;

class IDoRosStuff
{
  public:
    IDoRosStuff(const rclcpp::Node::SharedPtr& node) : n(node), i(0)
    {
    }

    void init()
    {
        pub_a = n->create_publisher<std_msgs::msg::Int32>("~/a", rclcpp::SystemDefaultsQoS());
        pub_b = n->create_publisher<std_msgs::msg::Int32>("~/b", rclcpp::SystemDefaultsQoS());

        srv_a = n->create_service<std_srvs::srv::Trigger>("~/a", std::bind(&IDoRosStuff::serviceACb, this, std::placeholders::_1, std::placeholders::_2));
        srv_b = n->create_service<std_srvs::srv::Trigger>("~/b", std::bind(&IDoRosStuff::serviceBCb, this, std::placeholders::_1, std::placeholders::_2));
    }

    void run()
    {
        std_msgs::msg::Int32 msg;
        msg.data = i;
        pub_a->publish(msg);
        pub_b->publish(msg);
        i++;
    }

    rclcpp::Parameter param_a, param_b, param_c;
    rclcpp::Publisher<std_msgs::msg::Int32>::SharedPtr pub_a, pub_b;
    rclcpp::Service<std_srvs::srv::Trigger>::SharedPtr srv_a, srv_b;
    rclcpp::Node::SharedPtr n;
    unsigned int i;

    void serviceACb(const std::shared_ptr<std_srvs::srv::Trigger::Request> req, std::shared_ptr<std_srvs::srv::Trigger::Response> resp)
    {
        (void)req;
        (void)resp;
[...]

PR #1796, opened 2021-10-08
Modify sub node copy constructor to create new node parameters object

jlack1987:
My attempt at fixing https://github.com/ros2/rclcpp/issues/731. The tldr of the change here is when a sub node is created, the pointer to the new node creates a new pointer to a `NodeParameters` object instead of being given the "parent" nodes pointer. This results in the following behavior change from parameters,

- Parameters are scoped into the node/sub node
- When a callback is registered via the `add_on_set_parameters_callback` call, it will only be triggered when a parameter registered to that particular node or sub node is modified, i.e. if `node_a` has a sub node `sub_node_a` and a parameter of `sub_node_a` is modified, callbacks registered to `node_a` will not be called
- Node options are still past down to sub nodes, so parameter relevant options such as `allow_undeclared_parameters` will apply to any sub nodes

I feel like the fix to this issue can't be this simple; however, I have been unable to find any test failures or problems that result from this change. I did add unit tests that failed prior to making this modification and then pass after the change.

Signed-off-by: Jordan Lack <jlack@houstonmechatronics.com>

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -523,6 +523,8 @@ public:
    *   has not been declared.
    * \throws rclcpp::exceptions::ParameterImmutableException if the parameter
    *   was create as read_only (immutable).
+   * \throws rclcpp::exceptions::InvalidParameterTypeException if the parameter
+   *   was created as statically typed
    */
   RCLCPP_PUBLIC
   void
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -236,7 +236,22 @@ Node::Node(
   node_topics_(other.node_topics_),
   node_services_(other.node_services_),
   node_clock_(other.node_clock_),
-  node_parameters_(other.node_parameters_),
+  node_parameters_(std::make_shared<rclcpp::node_interfaces::NodeParameters>(
+      other.node_base_,
+      other.node_logging_,
+      other.node_topics_,
+      other.node_services_,
+      other.node_clock_,
+      other.node_options_.parameter_overrides(),
+      other.node_options_.start_parameter_services(),
+      other.node_options_.start_parameter_event_publisher(),
+      // This is needed in order to apply parameter overrides to the qos profile provided in
+      // options.
+      get_parameter_events_qos(*other.node_base_, other.node_options_),
+      other.node_options_.parameter_event_publisher_options(),
+      other.node_options_.allow_undeclared_parameters(),
+      other.node_options_.automatically_declare_parameters_from_overrides()
+    )),
   node_time_source_(other.node_time_source_),
   node_waitables_(other.node_waitables_),
   node_options_(other.node_options_),
--- a/rclcpp/test/rclcpp/test_node.cpp
+++ b/rclcpp/test/rclcpp/test_node.cpp
@@ -211,6 +211,75 @@ TEST_F(TestNode, subnode_get_name_and_namespace) {
     }, rclcpp::exceptions::NameValidationError);
   }
 }
+
+TEST_F(TestNode, subnode_params) {
+  // create a node, a sub node, and a sub node of the sub node
+  auto node = std::make_shared<rclcpp::Node>("my_node", "ns");
+  auto subnode = node->create_sub_node("sub_ns");
+  auto subsubnode = subnode->create_sub_node("subsub_ns");
+
+  std::string name = "a";
+  ASSERT_NO_THROW(
+  {
+    // declare/set a parameter with the same name in each node, they
+    // should not be considered duplicates and thus should not throw
+    node->declare_parameter(name, 3.);
+    subnode->declare_parameter(name, 2.);
+    subsubnode->declare_parameter(name, 1.);
+  });
+
+  // retrieve and check the values
+  EXPECT_EQ(node->get_parameter(name).as_double(), 3.);
+  EXPECT_EQ(subnode->get_parameter(name).as_double(), 2.);
+  EXPECT_EQ(subsubnode->get_parameter(name).as_double(), 1.);
+
+  ASSERT_NO_THROW(
+  {
+    // change the values
+    node->set_parameter({name, 6.});
+    subnode->set_parameter({name, 5.});
+    subsubnode->set_parameter({name, 4.});
+  });
+
+  // check once more
+  EXPECT_EQ(node->get_parameter(name).as_double(), 6.);
+  EXPECT_EQ(subnode->get_parameter(name).as_double(), 5.);
+  EXPECT_EQ(subsubnode->get_parameter(name).as_double(), 4.);
+
+  // variable to keep track of how many times the callback is hit
+  int count = 0;
+  auto on_set_parameters =
+    [&count](const std::vector<rclcpp::Parameter> & parameters) {
+      // void this so there's no unused-variable warning
+      (void)parameters;
+      rcl_interfaces::msg::SetParametersResult result;
+      result.successful = true;
+      count++;
+      return result;
+    };
+
+  auto node_handler = node->add_on_set_parameters_callback(on_set_parameters);
+  auto subnode_handler = subnode->add_on_set_parameters_callback(on_set_parameters);
+  auto subsubnode_handler = subsubnode->add_on_set_parameters_callback(on_set_parameters);
+
+  subsubnode->set_parameter({name, -1.});
+  // subsubnode is the only one that had its parameter modified, so
+  // the callback should only have been called once, incrementing
+  // count by one
+  EXPECT_EQ(count, 1);
+  EXPECT_EQ(subsubnode->get_parameter(name).as_double(), -1);
+
+  subnode->set_parameter({name, -2.});
+  // again this should only increment count by one
+  EXPECT_EQ(count, 2);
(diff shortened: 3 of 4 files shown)


aprotyas:
You've probably tried this, but do you observe the expected behavior from https://github.com/ros2/rclcpp/issues/1778?

jlack1987:
@fujitatomoya topics and services stay the same here(i'll write some more unit tests to show that today). And as far as the comments you referenced [here](https://github.com/ros2/rclcpp/issues/731#issuecomment-640277329), the short answer is that it kinda sorta fixes it :stuck_out_tongue_winking_eye:.

It does fix the behavior as far as the interaction with the param server from the users frame of reference goes, but the way it fixes it does not match how topics/services are handled via simple namespace extension; however, after getting into the code I personally don't believe appending/prepending namespaces to the params is a good/clean way to fix that. I'll try to explain my reasoning on that. First though if you haven't please check out [this simple test](https://github.com/jlack1987/rclcpp/blob/feature/731/rclcpp/test/rclcpp/test_node.cpp#L226) to see how the user interacts with parameters from sub-nodes.

Currently(on released code and `master` params are stored with the `NodeParameters` pointer owned by a `Node`, and when sub nodes are created they are simply given the parent nodes pointer to the `NodeParameters` which really breaks parameter interaction from subnodes because multiple subnodes then can't create parameters of the same name. One option is to prepend the name of the parameter with the sub nodes namespace; however this imo isn't a great solution for parameters because when the `on_set_parameters_callback` is called, the parameter name would be `/ns1/ns2/.../name`. Yeah we could string split that  suppose and only send the last bit, but that seems kinda sloppy to me bc the name you're sending to the callback isn't the actual stored name of the parameter; however, if you all prefer that implementation I can try that instead.

The way I did it was to create a new `NodeParameters` pointer for sub nodes, which means that sub node parameters aren't accessible from the parent node nor any of the other sub nodes even if they share a common parent. Imo this simplifies the interaction from the user though in that the parameter name is simply what you give it and there's no internal parameter name expansion or crunching each time the user tries to interact with it. It's possible though there's some fallout that I haven't seen yet from this implementation, but the tests pass so I haven't found that yet :smile:

aprotyas:
> sub node parameters aren't accessible from the parent node nor any of the other sub nodes even if they share a common parent

I'm pretty sure there isn't a "correct" answer to this since #731 says `how parameters and "sub-nodes" interact is not documented and is not consistent`, but is there (should there be?) a need for sub-nodes to be able to access their parent nodes' parameters? I can make a case either way mentally, but I'm not sure what is the desired behavior here.

clalancette:
> I'm currently migrating a large code base from ros1->ros2 and I either have to fix the param behavior for sub nodes in ros2 or not use ros2 params sob so i'm very invested in fixing this however you all believe it should be done.

I think it is clear that subnodes are currently under-specified, so one other way out of this conundrum is to not use subnodes.  That is, you may not want to get bogged down in fully specifying subnodes at this point, but you definitely want to use parameters; they are the right way forward.  If you describe what you are using subnodes for, maybe we can suggest another way to do things.

Of course, if you *want* to fully specify subnodes, then that is great too.  I think at that point we'd want to move this to a design discussion instead.

jlack1987:
@fujitatomoya I went ahead and implemented most of the "extend namespace internally" implementation so I could display the behavior and why I think it's not the best approach. Best way to look is to look at my test. So first I make a sub node and a sub node on that sub node [here](https://github.com/jlack1987/rclcpp/blob/feature/extend-param-namespace/rclcpp/test/rclcpp/test_node.cpp#L284). 

The end behavior is largely the same between the two implementations until we get to [adding on_set_parameters_callback callbacks] to the nodes. Looking then at the [next line](https://github.com/jlack1987/rclcpp/blob/feature/extend-param-namespace/rclcpp/test/rclcpp/test_node.cpp#L334) where I set subsubnode's parameter. When that is set, it calls the callback for both the other subnode *and* the parent nodes callbacks and notifies them that `/sub_ns/subsub_ns/a` has changed.

While this may seem like a minor annoyance for this simple test, for a large node that has dozens of sub nodes that all have sub nodes of their own and so on, it would mean possibly hundreds of callbacks would be hit every time any one of those subnodes parameters changes and in almost all of those callbacks they're probably being notified of a parameter change on a parameter that they have no knowledge even exists bc it was created in another sub node somewhere else in the tree of nodes/subnodes. In addition, the name of the parameter that changed would have to have the entire sub namespace prepended to it when the callback is called otherwise there'd be no way of differentiating between a parameter "a" that exists in multiple sub nodes. I hope this makes sense. Let me know what you think!

fujitatomoya:
@jlack1987 yeah, i think that can be a problem for user experience. thanks for the explanation.

fujitatomoya:
btw, this fix changes the current behavior for `ros2 param list`. Having different `NodeParameters` from parent node, it conceals subordinate node's parameter via `ros2 param list`.

fujitatomoya:
> but it seems that I can't make sub node params show up with ros2 param list

for allowing us to do that, i think eventually we need to create `NodeBase` (calling rcl_init_node) for subordinate node, which makes me think what's the difference between node and subordinate node ??? 🤔 ...

> But making it such that sub nodes don't share the same NodeBase pointer would mean that calling spin() on the parent node wouldn't spin the sub nodes

this is true.

we can have discussion in MW WG, if you'd like to bring the topic. https://docs.ros.org/en/rolling/Governance.html#middleware, always welcome!

wjwwood:
Sorry, I need to catch on this (long) thread to make a more thoughtful response, but my general feedback after skimming, is that a "sub-node" should completely be syntactic sugar on top of a Node, and to that end, it shouldn't require any additional entities at the rcl level (like additional node objects etc...). Additionally, I think it shouldn't require new instances of `NodeParameter` or anything like that... It should simply modify the incoming arguments when interacting with parameters, by adding implicit namespace unless the argument is absolute already.

Any other issues, like number of callbacks to process when there are lots of parameters, or separation of concern when listing parameters, should be address some other way other than having multiple places where we can store parameters within a node.

This is all my professional opinion about it, but I could be wrong and/or convinced otherwise. Hopefully I can find more time to iterate with you guys on it soon.

wjwwood:
Added "more-information-needed" as a stand-in for "needs more discussion".

jlack1987:
Yeah I agree with you @wjwwood, initially I thought there was a simple solution to the problem and implemented said solution but upon further digging it became clear that solving it is much more difficult than I thought. My only workaround right now is a painful one but it's all I can come up with, and it's to name parameters with their full namespace in the param file. Something like,

```
my_node:
  ros__parameters:
    /sub1/sub/sub3/param1: 1
    /sub1/sub/sub5/param1: 1
    /sub1/sub/sub3/sub4/param2: 3
```

Obviously this isn't great bc I have hundreds of parameters and some very long namespaces but I don't see any other solution.

jlack1987:
Also, feel free to close this MR whenever as it's not a complete solution

fujitatomoya:
> "sub-node" should completely be syntactic sugar on top of a Node

i still think this is the way it is. it just takes care of namespace on top of parent node.

jlack1987:
> "sub-node" should completely be syntactic sugar on top of a Node

>> i still think this is the way it is. it just takes care of namespace on top of parent node

You are correct I think with the exception of parameters. With respect to sub nodes, the behavior of parameters is not intuitive and not the way they "should" behave in my opinion. I know ros2 is different from ros1, but there's a really nice mechanic from ros1 that we're obviously trying to mimic here and that's
```
// the ros1 way
ros::NodeHandle node("ns");
ros::NodeHandle sub_node(node, "sub_ns"); // topics, services, parameters are all namespaced here

// the ros2 way
rclcpp::Node::SharedPtr node("my_node");
rclcpp::Node::SharedPtr sub_node = node->create_sub_node("sub_ns"); // topics and services are namespaced, but parameters are not
```
and the current state of things does give the desired mechanics for topics and services, but not for parameters.

tylerjw:
For context, I did not know about subnodes until I read this thread.  

I've also been struggling with the idea of how to namespace parameters for different subsystems in a node and then handle reconfiguring those subsystems when a parameter is dynamically updated.  The approach I've been working on so far is to have a single object that I pass around to each subsystem to handle the dynamic reconfiguring of subsystems.  Each subsystem then registers callbacks for parameter updates to this dynamic reconfigure object.  This dynamic reconfigure object has the one `on_set_parameters_callback` callback in my node and handles splitting the names of the parameters into namespaces and calling the appropriate subsystem's callback so it can then reconfigure itself.  If there was a pattern built into ros2 for handling subsystems and their parameters with callbacks for changes it would be much nicer than implementing that functionality myself.

One large downside to my approach is that each subsystem has to have a namespace it is initialized with and uses for declaring and getting parameters.  It also uses this namespace to register a callback for when any parameter in its namespace is changed.

So far I have not had to deal with sub-subsystems.  I've done this so far by just having one level of namespaces for subsystems.  Having features built into ROS for handling parameters of subsystems within a node would be really nice.

aprotyas:
@fujitatomoya do you know if this was discussed in a MW WG meeting?

fujitatomoya:
i remember that we had a quick talk but not really discussed, probably it would be nice to bring this back to topic.

jlack1987:
Gonna glose this MR as the approach is ultimately flawed. I have linked to this PR though in the issue for which this PR attempts to fix so others can find their way here and see the discussion which I think could be useful to others.

PR #1375, opened 2020-10-05
Add clock qos to node options

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1318.

I will open a similar PR in `rclpy`.

(edit) Edited title, as this is allowing to configure the clock qos and not the rosout qos.

PR #1801, opened 2021-10-13
Fix unused QoS profile for clock subscription and make ClockQoS the default

nnmm:
This is a followup to the discussion in https://github.com/ros2/rclcpp/pull/1375.


ivanpauno:
the tests that failed seem to be unrelated, @nnmm could you rebase so we run CI again?

PR #1802, opened 2021-10-13
Suppress clang dead-store warnings in the benchmarks.

clalancette:
clang static analysis complains that there are dead stores in
most of the benchmark tests, which is technically correct.
We use an idiom like:

for (auto _ : state) {
}

And never access _.  Silence clang here by doing (void)_;
all of the places this is seen.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15459/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10153/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13137/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15673/)

PR #1807, opened 2021-10-21
Galactic-backport: handle sigterm in the same way we handle sigint

j-rivero:
Mostly a backport of #1771 but the default behavior for handling SIGINT is not enabled to preserve backwards compatibility.

The PR was done in three steps:
 * Backport #1771 7c1bd17
 * Turn off the default for SIGINT 03267bf

In a final step, I realized that adding a new parameter we will be kiling the ABI of `rclcpp::init` which sounds to me not a nice thing to do. I tried to avoid it by keeping the current signature of `rclcpp::init,`making it to call a new signature that includes the `SignalHandlerOptions` with the option of only enabling SigTerm. Hopefully the change in e0dad49 will make the `rclcpp::init` options in the same way but keeping ABI stable.

### DIFF
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -75,7 +75,7 @@ public:
    * Every context which is constructed is added to a global vector of contexts,
    * which is used by the signal handler to conditionally shutdown each context
    * on SIGINT.
-   * See the shutdown_on_sigint option in the InitOptions class.
+   * See the shutdown_on_signal option in the InitOptions class.
    */
   RCLCPP_PUBLIC
   Context();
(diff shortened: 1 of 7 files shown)


ivanpauno (approved):

ivanpauno:
@j-rivero is this ready to be merged or are we waiting for an rclpy equivalent to be ready?

j-rivero:
> @j-rivero is this ready to be merged or are we waiting for an rclpy equivalent to be ready?

rclpy equivalent would require a good amount of work. This is ready here if we support the fact of having rclcpp features not present in rclpy.

ivanpauno:
> rclpy equivalent would require a good amount of work. This is ready here if we support the fact of having rclcpp features not present in rclpy.

That sounds fine to me, particurlarly considering that the sigterm handler is disabled by default.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10273/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13257/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15809/)

ivanpauno:
@j-rivero I think that CI is failing because of this change https://github.com/ros2/system_tests/pull/485/files#diff-26d995793afc3a671b4deb7b1a2cc8a07e4a1231c21cb612de8d58b67e5a6b3a.
The test is expecting for some particular output in stderr that is never printed.
I think we can update the PR to use the old wording, to avoid creating a PR in system_tests as well.

clalancette:
Closing this one as Galactic is now EOL.

PR #1793, opened 2021-10-01
Revert "Add Clock::sleep_until method (#1748)"

Blast545:
This reverts commit 81df5843f321b08db621c17502ff94193ffd3657.

Attempts to address `nighly_osx_debug` test regressions.

PR #1748, opened 2021-08-12
Add Clock::sleep_until method

emersonknapp:
Part of #1730 
Related to ros2/rosbag2#694

Add a `Clock::sleep_until` method. Handle all 3 clock types, and edge cases for shutdown and enabling/disabling ROS time.

Adds test cases for enable/disable ROStime, and `rclcpp::shutdown` - which needed special interrupts. Tests ROS time via `/clock` messages, and has basic testing for steady/system times, though those are much simpler cases.

PR #1814, opened 2021-11-03
Re-add Clock::sleep_until

sloretz:
This revives #1748 which was reverted in #1793 with a couple changes.

Originally the PR caused 5 test failures on OSX. The tests themselves passed, but the return code of the test processes would either be -11,

```
1: -- run_test.py: return code -11
```

or it had an error message and a return code of -6.

```
1: libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
1: -- run_test.py: return code -6
```

All the failing tests created `Clock` instances but never initialized the global default context. The cause of the failures probably had something to do with `Clock` removing its shutdown callback from the global default context in its destructor.

This ties into another problem with the PR: the `Clock::sleep_until` would wake on shutdown of the global default context, but it can't be assumed that context is the one being used. This PR makes `Clock::sleep_until()` take a context as the second argument, and has it default to the global default context for convenience.

This new API [overlaps with `Context::sleep_for()`](https://github.com/ros2/rclcpp/blob/94264320b4a2fde8999b895824d2c85cf0e4d2ac/rclcpp/src/rclcpp/context.cpp#L493-L495). Maybe that method should be deprecated?

I haven't yet reviewed a lot of the other code, so I'm leaving this as a draft while I do that.

### DIFF
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -19,6 +19,7 @@
 #include <memory>
 #include <mutex>
 
+#include "rclcpp/contexts/default_context.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/time.hpp"
 #include "rclcpp/visibility_control.hpp"
@@ -87,7 +88,9 @@ public:
    *     false. There is not a consistent choice of sleeping time when the time source changes,
    *     so this is up to the caller to call again if needed.
    *
+   * \throws std::runtime_error if the context is invalid
    * \param until absolute time according to current clock type to sleep until.
+   * \param context the rclcpp context the clock should use to check that ROS is still initialized.
    * \return true immediately if `until` is in the past
    * \return true when the time `until` is reached
    * \return false if time cannot be reached reliably, for example from shutdown or a change
@@ -95,7 +98,9 @@ public:
    */
   RCLCPP_PUBLIC
   bool
-  sleep_until(Time until);
+  sleep_until(
+    Time until,
+    Context::SharedPtr context = contexts::get_global_default_context());
 
   /**
    * Returns the clock of the type `RCL_ROS_TIME` is active.
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -89,6 +89,7 @@ public:
    *     so this is up to the caller to call again if needed.
    *
    * \throws std::runtime_error if the context is invalid
+   * \throws std::runtime_error if `until` has a different clock type from this clock
    * \param until absolute time according to current clock type to sleep until.
    * \param context the rclcpp context the clock should use to check that ROS is still initialized.
    * \return true immediately if `until` is in the past
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -88,14 +88,14 @@ public:
    *     false. There is not a consistent choice of sleeping time when the time source changes,
    *     so this is up to the caller to call again if needed.
    *
-   * \throws std::runtime_error if the context is invalid
-   * \throws std::runtime_error if `until` has a different clock type from this clock
    * \param until absolute time according to current clock type to sleep until.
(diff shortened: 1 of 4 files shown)


sloretz:
Checking of OSX test failures fixed.

CI OSX Build type Debug build: `--packages-above-and-dependencies rclcpp` test: `--packages-above rclcpp`  [Build Status](https://ci.ros2.org/job/ci_osx/13221/)

sloretz:
Ready for review

CI (build: `--packages-above-and-dependencies rclcpp` test: `--packages-above rclcpp`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15558/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10242/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13226/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15772/)

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -78,6 +79,30 @@ class Clock
...
+   * Sleep until a specified Time, according to clock type.
+   *
+   * Notes for RCL_ROS_TIME clock type:
+   *   - Can sleep forever if ros time is active and received clock never reaches `until`
+   *   - If ROS time enabled state changes during the sleep, this method will immediately return
+   *     false. There is not a consistent choice of sleeping time when the time source changes,
+   *     so this is up to the caller to call again if needed.
+   *
+   * \throws std::runtime_error if the context is invalid
+   * \throws std::runtime_error if `until` has a different clock type from this clock
```
clalancette, line 92: Nit, but I think it is more common in our doxygen comments to have `throws` after `return`.

clalancette (changes requested):
> This new API overlaps with Context::sleep_for().

Is that really true?  `sleep_until` and `sleep_for` are different (though related) concepts.  So I think having them both is fine.  Did you have something else in mind?

Besides the minor thing pointed out inline, I think it would be good to have tests for a non-global context, as well as a test with an invalid context.

sloretz:
> Is that really true? sleep_until and sleep_for are different (though related) concepts. So I think having them both is fine. Did you have something else in mind?

Ah I'm getting ahead of myself. A `Clock::sleep_for()` PR is coming next.

rclcpp/include/rclcpp/clock.hpp
sloretz: Put `\throws` at the end of the doc block in 27b36681

sloretz:
Rebuild of CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15563/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10247/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13231/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15778/)

clalancette (approved):

sloretz:
The windows job appears to have hung in `test_launch_ros`. @Blast545 is there a ticket open for this already? I wasn't able to find one.

Here's a new Windows job: [Build Status](https://ci.ros2.org/job/ci_windows/15785/)

sloretz:
Another Windows Hang - rerun [Build Status](https://ci.ros2.org/job/ci_windows/15786/)

Blast545:
@sloretz In the buildfarm I've only seen that hang only a couple of times in the past weeks:

https://ci.ros2.org/job/nightly_win_deb/2151/
https://ci.ros2.org/view/nightly/job/nightly_win_rel/2101/

I thought it was related to https://github.com/ros2/launch_ros/issues/280 which seems to be addressed after https://github.com/ros2/launch_ros/pull/281. I'll open a ticket, it's a flaky hang.

Blast545:
@sloretz I opened an issue https://github.com/ros2/launch_ros/issues/285 with the hang situation. The code does not seem related to the hang, however, it seems odd to me that you got it twice for 3 tests, as it hasn't appeared that often in the buildfarm.

Another rerun JIC [Build Status](https://ci.ros2.org/job/ci_windows/15787/)

PR #1815, opened 2021-11-04
min_forward & min_backward thresholds must not be disabled

sloretz:
Required by ros2/rcl#948
Part of fixing ros2/rcl#947

### DIFF
--- a/rclcpp/test/rclcpp/test_time_source.cpp
+++ b/rclcpp/test/rclcpp/test_time_source.cpp
@@ -304,8 +304,8 @@ public:
 TEST_F(TestTimeSource, callbacks) {
   CallbackObject cbo;
   rcl_jump_threshold_t jump_threshold;
-  jump_threshold.min_forward.nanoseconds = 0;
-  jump_threshold.min_backward.nanoseconds = 0;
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):

fujitatomoya (changes requested):
probably we are missing the following?

https://github.com/ros2/rclcpp/blob/94264320b4a2fde8999b895824d2c85cf0e4d2ac/rclcpp/test/rclcpp/test_time_source.cpp#L493-L494

sloretz:
> probably we are missing the following?

I think that test is fine as is. It doesn't need the time threshold enabled because the callback is triggered by enabling ROS time.

https://github.com/ros2/rclcpp/blob/746f1a80b0aa587e0fc2f6bd110be73540c10117/rclcpp/test/rclcpp/test_time_source.cpp#L495

https://github.com/ros2/rclcpp/blob/746f1a80b0aa587e0fc2f6bd110be73540c10117/rclcpp/test/rclcpp/test_time_source.cpp#L510-L512

fujitatomoya:
> because the callback is triggered by enabling ROS time.

right, thanks!

fujitatomoya (approved):

Issue #1762, opened 2021-09-02
Allow to undeclare statically typed or read-only parameters from host node


bpwilcox:
I had been following the discussions and eventual implementation of the design for enabling statically typed parameters. While I am glad to see this enhancement, one area I see lacking sufficient discussion in the design conversation and the current documentation is the behavior of undeclaring parameters. Let me first start by illustrating a simple user scenario:

There is a node which dynamically loads and unloads plugins, each of which declares its own parameters on the host node. These parameters would appropriately be either read-only or statically typed in order to protect from any undesired modifications. Those parameters will be accessible by other nodes in the system, and when the plugins are dynamically unloaded, we expect that clients of this node should have awareness that the plugins and hence their parameters are gone (e.g. it is not fetching values from a no longer active plugin).  

Currently, a parameter which is either read-only or is statically typed (the default) [cannot be undeclared](https://github.com/ros2/rclcpp/blob/ecb81ef2c35a3e605b7c501ace4ef1baf4755bf9/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L555). My initial impression was that this behavior was uninituitive, because to undeclare a parameter seems a different use case than attempting to modify it. Even further, setting a parameter to `rclcpp::PARAMETER_NOT_SET` seems like an _additional_ use case, especially now that [uninitialized parameters are allowed](https://github.com/ros2/rclcpp/pull/1673) again. The main counterpoint I can see is that a parameter could be undeclared and subsequently declared again with a different type or different value (in case of read-only).

[...]

PR #1824, opened 2021-11-11
add APIs to force undeclaring parameters

alsora:
This PR addresses https://github.com/ros2/rclcpp/issues/1762

It adds new APIs to the `node_parameters` interface that allow to bypass `read_only` and `dynamic_typing` restrictions.
These APIs are used by the node that owns the parameters, while parameter servers (which control changes coming from external nodes) still use old APIs.

### DIFF
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
@@ -110,16 +110,7 @@ public:
   RCLCPP_PUBLIC
   virtual
   void
-  undeclare_parameter(const std::string & name) = 0;
-
-  /// Undeclare a parameter ignoring "read_only" and "dynamic_typing" flags.
-  /**
-   * \sa rclcpp::Node::undeclare_parameter
-   */
-  RCLCPP_PUBLIC
-  virtual
-  void
-  force_undeclare_parameter(const std::string & name) = 0;
+  undeclare_parameter(const std::string & name, bool force) = 0;
 
   /// Return true if the parameter has been declared, otherwise false.
   /**
@@ -137,16 +128,7 @@ public:
   RCLCPP_PUBLIC
   virtual
   std::vector<rcl_interfaces::msg::SetParametersResult>
-  set_parameters(const std::vector<rclcpp::Parameter> & parameters) = 0;
-
-  /// Set one or more parameters, one at a time, ignoring "read_only" and "dynamic_typing" flags.
-  /**
-   * \sa rclcpp::Node::set_parameters
-   */
-  RCLCPP_PUBLIC
-  virtual
-  std::vector<rcl_interfaces::msg::SetParametersResult>
-  force_set_parameters(const std::vector<rclcpp::Parameter> & parameters) = 0;
+  set_parameters(const std::vector<rclcpp::Parameter> & parameters, bool force) = 0;
 
   /// Set one or more parameters, all at once.
   /**
@@ -155,16 +137,7 @@ public:
   RCLCPP_PUBLIC
   virtual
   rcl_interfaces::msg::SetParametersResult
-  set_parameters_atomically(const std::vector<rclcpp::Parameter> & parameters) = 0;
-
-  /// Set one or more parameters, all at once, ignoring "read_only" and "dynamic_typing" flags.
-  /**
-   * \sa rclcpp::Node::set_parameters_atomically
-   */
-  RCLCPP_PUBLIC
-  virtual
-  rcl_interfaces::msg::SetParametersResult
-  force_set_parameters_atomically(const std::vector<rclcpp::Parameter> & parameters) = 0;
+  set_parameters_atomically(const std::vector<rclcpp::Parameter> & parameters, bool force) = 0;
 
   /// Get descriptions of parameters given their names.
   /*
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
@@ -128,7 +128,7 @@ public:
   RCLCPP_PUBLIC
   virtual
   std::vector<rcl_interfaces::msg::SetParametersResult>
-  set_parameters(const std::vector<rclcpp::Parameter> & parameters, bool force) = 0;
+  set_parameters(const std::vector<rclcpp::Parameter> & parameters) = 0;
 
   /// Set one or more parameters, all at once.
   /**
@@ -137,7 +137,7 @@ public:
   RCLCPP_PUBLIC
   virtual
   rcl_interfaces::msg::SetParametersResult
-  set_parameters_atomically(const std::vector<rclcpp::Parameter> & parameters, bool force) = 0;
+  set_parameters_atomically(const std::vector<rclcpp::Parameter> & parameters) = 0;
 
   /// Get descriptions of parameters given their names.
   /*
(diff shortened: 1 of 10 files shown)


rclcpp/src/rclcpp/node.cpp
```diff
@@ -359,7 +359,7 @@ Node::declare_parameter(
 void
 Node::undeclare_parameter(const std::string & name)
 {
-  this->node_parameters_->undeclare_parameter(name);
+  this->node_parameters_->force_undeclare_parameter(name);
```
gbiggs, line 362: Should this also be passing a boolean `false`? The local node can use `force_undeclare_parameter` to ensure it's undeclared, but having a non-forced one would catch programmer errors.

rclcpp/src/rclcpp/node.cpp
```diff
@@ -377,13 +377,13 @@ Node::set_parameter(const rclcpp::Parameter & parameter)
 std::vector<rcl_interfaces::msg::SetParametersResult>
 Node::set_parameters(const std::vector<rclcpp::Parameter> & parameters)
 {
-  return node_parameters_->set_parameters(parameters);
+  return node_parameters_->force_set_parameters(parameters);
```
gbiggs, line 380: Should this also be passing a boolean `false`?

rclcpp/src/rclcpp/node.cpp
```diff
@@ -377,13 +377,13 @@ Node::set_parameter(const rclcpp::Parameter & parameter)
...
 }
 
 rcl_interfaces::msg::SetParametersResult
 Node::set_parameters_atomically(const std::vector<rclcpp::Parameter> & parameters)
 {
-  return node_parameters_->set_parameters_atomically(parameters);
+  return node_parameters_->force_set_parameters_atomically(parameters);
```
gbiggs, line 386: Should this also be passing a boolean `false`?

rclcpp/src/rclcpp/node.cpp
alsora: I'm not sure I understand.
Are you suggesting to expose both the forced and non-forced methods as part of the node interface? such that users can choose which one to use?
(and then internally use a flag)

tylerjw (approved):
I read through this and it seems to make sense and fulfill the requirements of the lined issue.  If I understand correctly what this does is allow the host node to set and undeclare parameters even if they are read-only and/or dynamically typed.

rclcpp/src/rclcpp/node.cpp
gbiggs: Yes. I wouldn't say it's a hard requirement, but I think it would be better for users to default to using the unforced one unless they know what they're doing.

alsora: 240012a should address this.
I also added the changes to the lifecycle nodes, since I forgot before.

fujitatomoya:

alsora:
I thought about that, but my conclusion was that as long as you support undeclaring a parameter, then everything else is already automatically possible.

Given a read_only parameter I could undeclare it and the declare it back again with a different value.
For consumers of this parameter the result would be the same as if the parameter was set to a new value directly.

alsora:
Hi, any update on this?
Can you clarify why un-declaring a read-only parameter and then declaring it again with a different value is better than allowing the node to change it?
IMO, by requiring to use 2 operations we have the risk of an external consumer to fail to read the parameter if this operation ends up happening between the undeclaration and redeclaration

fujitatomoya:
@alsora thanks for the comment!

I am just saying `can overwrite the read-only value` does not make sense. if it is allowed, what is that supposed to mean `read-only`? (for me, it sounds like updating contents in the read-only file system... but we can mount/unmount the file system.)

> external consumer to fail to read the parameter if this operation ends up happening between the undeclaration and redeclaration

i think this is still okay, remote nodes can see this parameter events. event for modifying `read-only` parameter would be confusing, i guess.

that is just my opinion, i would like to hear from others.

alsora:
Following your file-system example, the way I was seeing a read-only parameter is: this parameter is read only for everyone except its owner who created it. Or the node having sudo powers.

I think that it's a common situation to have a node that can autonomously decide to transition between different "modes".
Other nodes may be interested in knowing which mode is the node currently running and they could query that by reading the parameter.

But maybe in this situation a ROS service `get_mode` could work better?

fujitatomoya:
> this parameter is read only for everyone except its owner who created it.

okay, that is `remote-read-only`. which sounds good.
if that is the spec, probably we just do not need the concept of `force`, since owner can have all access.

alsora:
Well, the concept of force was exactly added to allow owners to have this kind of access.
I'm not sure how to do it differently (unless we add a `remote_read_only` to the descriptor).

In my first implementation I did not expose a force API and I just did that under the hood, but I think it has a better semantic meaning now.

alsora:
I moved the force "set" parameter functionality to a separate PR so we can discuss whether we want it or not.
https://github.com/ros2/rclcpp/pull/1850

fujitatomoya:
@alsora separating PR works for me, thanks 👍

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
```diff
@@ -137,8 +137,7 @@ class NodeParametersInterface
   RCLCPP_PUBLIC
   virtual
   rcl_interfaces::msg::SetParametersResult
-  set_parameters_atomically(
-    const std::vector<rclcpp::Parameter> & parameters) = 0;
+  set_parameters_atomically(const std::vector<rclcpp::Parameter> & parameters) = 0;
```
fujitatomoya, line 140: do we really need this change?

fujitatomoya (approved):
lgtm with minor comment

fujitatomoya:
@alsora btw, why dont we add some note like https://github.com/ros2/rcl_interfaces/pull/134? i think it would be useful to user.

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
alsora: No we don't.
Just a leftover from reverting one of the commits.

alsora: Fixed in 741fb10

fujitatomoya:
@gbiggs @wjwwood friendly ping. any comments for current code?

alsora:
Hi, any updates on this PR? I think all comments have been addressed

alsora:
@gbiggs @wjwwood friendly ping

wjwwood:
Sorry for the delay, I'm starting to catch up with my correspondences now.

I'm unfortunately not sure how I feel about this change any longer, see: https://github.com/ros2/rclcpp/pull/1850#issuecomment-1042449401

I would either like to be convinced otherwise or be overruled, _or_ see these changes on this pull request:

- make the methods to force undeclare protected
- prevent undeclare/re-declare
  - that way at least the readers of the parameter only have to assume it may disappear, but never that it can change
  - I'm aware that this might not be sufficient to satisfy @bpwilcox's original use case as described here: https://github.com/ros2/rclcpp/issues/1762#issue-986652664 because the unloaded plugins might be recreated, therefore a less exploitive use of "undeclare/re-declare different"

Like I said, I can be convinced otherwise or overruled, if there is consensus without me.

alsora:
Closing as this will need to be rewritten using a different approach (i.e. adding a new "immutable" attribute type)

PR #1832, opened 2021-11-30
make node base sharable

iuhilnehc-ynos:
It's intent to use another way to fix https://github.com/ros2/rclcpp/pull/1754 to prevent the separation of `rcl_node_init` and `rcl_node_fini`.

<del>and changing the creation behavior of subscription that not holding callback group to keep the consistency of other entities to make the test passed.</del>

### DIFF
--- a/rclcpp/include/rclcpp/subscription_factory.hpp
+++ b/rclcpp/include/rclcpp/subscription_factory.hpp
@@ -88,11 +88,7 @@ create_subscription_factory(
   subscription_topic_stats = nullptr
 )
 {
-  rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> options_copy = options;
-  // reset callback group as the Subscription is unnecessary to hold it
-  options_copy.callback_group.reset();
-
-  auto allocator = options_copy.get_allocator();
+  auto allocator = options.get_allocator();
 
   using rclcpp::AnySubscriptionCallback;
   AnySubscriptionCallback<MessageT, AllocatorT> any_subscription_callback(*allocator);
@@ -100,8 +96,7 @@ create_subscription_factory(
 
   SubscriptionFactory factory {
     // factory function that creates a MessageT specific SubscriptionT
-    [options = std::move(options_copy), msg_mem_strat, any_subscription_callback,
-      subscription_topic_stats](
+    [options, msg_mem_strat, any_subscription_callback, subscription_topic_stats](
       rclcpp::node_interfaces::NodeBaseInterface * node_base,
       const std::string & topic_name,
       const rclcpp::QoS & qos
--- a/rclcpp/include/rclcpp/subscription_factory.hpp
+++ b/rclcpp/include/rclcpp/subscription_factory.hpp
@@ -88,7 +88,11 @@ create_subscription_factory(
   subscription_topic_stats = nullptr
 )
 {
-  auto allocator = options.get_allocator();
+  rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> options_copy = options;
+  // reset callback group as the Subscription is unnecessary to hold it
+  options_copy.callback_group.reset();
+
+  auto allocator = options_copy.get_allocator();
 
   using rclcpp::AnySubscriptionCallback;
   AnySubscriptionCallback<MessageT, AllocatorT> any_subscription_callback(*allocator);
@@ -96,7 +100,8 @@ create_subscription_factory(
 
   SubscriptionFactory factory {
     // factory function that creates a MessageT specific SubscriptionT
-    [options, msg_mem_strat, any_subscription_callback, subscription_topic_stats](
+    [options = std::move(options_copy), msg_mem_strat, any_subscription_callback,
+      subscription_topic_stats](
       rclcpp::node_interfaces::NodeBaseInterface * node_base,
       const std::string & topic_name,
       const rclcpp::QoS & qos
--- a/rclcpp/test/rclcpp/test_memory_strategy.cpp
+++ b/rclcpp/test/rclcpp/test_memory_strategy.cpp
@@ -319,6 +319,7 @@ TEST_F(TestMemoryStrategy, get_node_by_group) {
 TEST_F(TestMemoryStrategy, get_group_by_subscription) {
   WeakCallbackGroupsToNodesMap weak_groups_to_nodes;
   rclcpp::SubscriptionBase::SharedPtr subscription = nullptr;
+  rclcpp::CallbackGroup::SharedPtr callback_group = nullptr;
   {
     auto node = std::make_shared<rclcpp::Node>("node", "ns");
     node->for_each_callback_group(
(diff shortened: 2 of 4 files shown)


iuhilnehc-ynos:

rclcpp/include/rclcpp/subscription_factory.hpp
```diff
@@ -88,15 +88,20 @@ create_subscription_factory(
   subscription_topic_stats = nullptr
 )
 {
-  auto allocator = options.get_allocator();
+  rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> options_copy = options;
+  // reset callback group as the Subscription is unnecessary to hold it
+  options_copy.callback_group.reset();
```
ivanpauno, line 93: why do we need to do this?

iuhilnehc-ynos: Because I want to keep consistent behavior with other entities. 
I found other entities not to hold the callback group, refer to https://github.com/ros2/rclcpp/pull/1754#issuecomment-982423135, so I think it's better not to hold a callback group in the subscription.
(Next time I will create a new issue to confirm if it's a bug when I encounter this situation.)

iuhilnehc-ynos: Is it a correct comment?
https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L361 

IMO, [This](https://github.com/ros2/rclcpp/blob/87d754b2190edf53a3f7b9f95812e5a80313f649/rclcpp/test/rclcpp/test_memory_strategy.cpp#L317) seems just for a coverage test. Not to consider whether it's suitable compared with other entities.

I am not sure if there is some special case that must hold the callback group in the subscription.

ivanpauno: > Because I want to make the subscription keep consistent behavior with other entities.
I found other entities not to hold the callback group, refer to #1754 (comment), so I think it's better not to hold a callback group in the subscription.

That sounds fine, but it doesn't seem to be needed in this PR (which seems to be a replacement of https://github.com/ros2/rclcpp/pull/1754).
I would rather see this done in another PR if it's not needed here.

ivanpauno: > I am not sure if there is some special case that must hold the callback group in the subscription.

Could you explain how this change affected that test?
I see there's a lifetime issue, but it isn't to clear to me what exactly the problem is.

iuhilnehc-ynos: > I would rather see this done in another PR if it's not needed here.

OK. I'll create a new PR.

> Could you explain how this change affected that test?

Yes, it's my pleasure. Let me try to explain it.

```patch
 std::shared_ptr<const rcl_node_t>
 NodeBase::get_shared_rcl_node_handle() const
 {
-  return node_handle_;
+  return std::shared_ptr<const rcl_node_t>(shared_from_this(), node_handle_.get());
 }
```

After making the `NodeBase` sharable with rcl_node_t, it means the dependency chain will be `SubscriptionBase -> rcl_node_t -> NodeBase`. Even if the `Node` goes out of scope, we can also use `memory_strategy()->get_group_by_subscription(subscription, weak_groups_to_nodes)` to get the `CallbackGroup` pointer because the `callback_group` and `NodeBase` are alive. So I need to make the `callback_group` out of life to make the test `TEST_F(TestMemoryStrategy, get_group_by_subscription)` passed.

I know this patch extends a bit more lifetime of `NodeBase`.

ivanpauno: > Even if the Node goes out of scope, we can also use memory_strategy()->get_group_by_subscription(subscription, weak_groups_to_nodes) to get the CallbackGroup pointer because the callback_group and NodeBase are alive

IMO, the test makes no sense.
It's testing stuff that's not clearly defined.

I would just update the tests without updating anything else, I would prefer to discuss consitency with other cases in a new PR.

> OK. I'll create a new PR.

:+1: thanks

fujitatomoya (commented):
looks good to me once @ivanpauno 's comment is addressed.

rclcpp/test/rclcpp/test_memory_strategy.cpp
```diff
@@ -307,6 +307,8 @@ TEST_F(TestMemoryStrategy, get_node_by_group) {
     EXPECT_EQ(
       node_handle,
       memory_strategy()->get_node_by_group(callback_group, weak_groups_to_nodes));
+    // Clear the handles to not hold NodeBase.
+    memory_strategy()->clear_handles();
```
ivanpauno, line 311: is this line needed or can we delete it without modifying anything else?
If we cannot delete it, could you explain what changes?

Sorry for making all these questions, but there are many objects with interdependent lifetimes and it's hard to follow why this addition is needed.

ivanpauno (approved):
I have one extra question, otherwise lgtm.

rclcpp/test/rclcpp/test_memory_strategy.cpp
iuhilnehc-ynos: > is this line needed or can we delete it without modifying anything else?
> If we cannot delete it, could you explain what changes?

This line can be removed, but I need to update 
https://github.com/ros2/rclcpp/blob/0b890eea1d83e6ff2bc6ca8df7f192b1151e6599/rclcpp/test/rclcpp/test_memory_strategy.cpp#L314-L316 to expect getting the callback_group as the NodeBase(AllocatorMemoryStrategy-> VectorRebind<std::shared_ptr<const rcl_subscription_t>> -> NodeBase, as we know, some subscriptions(e.g. /parameter_events) and services are created when a Node created) is still alive even if the Node goes out of scope.

What I did is try to not update the expectation of a test before. I can remove that line and update the expectation. But there might be a new class derived from the `memory_strategy::MemoryStrategy` and not using the `std::shared_ptr<const rcl_subscriptions_t>` to keep the entity in the future, the expectation could be failed. So I'd like to make use of calling this virtual member function `MemoryStrategy::clear_handles()` to clean `NodeBase` holder, and then to expect a `nullptr` in the test.

What do you think about it?

ivanpauno: I see, I think the change looks good.

ivanpauno (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15721/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10403/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13386/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15966/)

iuhilnehc-ynos:
rebased from master branch and then re-run CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15722/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10404/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13387/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15967/)

PR #1833, opened 2021-12-03
not to hold callback group in subscription

iuhilnehc-ynos:
keep consistency with other cases to not hold callback group in subscription

https://github.com/ros2/rclcpp/pull/1832#discussion_r761027390
<details><summary> The above link can jump but not show the correct location. Copy comments</summary>

> Because I want to make the subscription keep consistent behavior with other entities.
I found other entities not to hold the callback group, refer to #1754 (comment), so I think it's better not to hold a callback group in the subscription.

That sounds fine, but it doesn't seem to be needed in this PR (which seems to be a replacement of https://github.com/ros2/rclcpp/pull/1754).
I would rather see this done in another PR if it's not needed here.

</details>

NOTE:
the callback group for subscription is needed (kept) here, 
https://github.com/ros2/rclcpp/blob/536df11ee0fdac3e4753754d11052068983c4d5c/rclcpp/include/rclcpp/create_subscription.hpp#L137
not in the subscription
https://github.com/ros2/rclcpp/blob/536df11ee0fdac3e4753754d11052068983c4d5c/rclcpp/include/rclcpp/subscription.hpp#L146

### DIFF
--- a/rclcpp/include/rclcpp/create_generic_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_generic_subscription.hpp
@@ -54,7 +54,8 @@ std::shared_ptr<GenericSubscription> create_generic_subscription(
   std::function<void(std::shared_ptr<rclcpp::SerializedMessage>)> callback,
   const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
     rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
-  )
+  ),
+  rclcpp::CallbackGroup::SharedPtr group = nullptr
 )
 {
   auto ts_lib = rclcpp::get_typesupport_library(
@@ -69,7 +70,7 @@ std::shared_ptr<GenericSubscription> create_generic_subscription(
     callback,
     options);
 
-  topics_interface->add_subscription(subscription, options.callback_group);
+  topics_interface->add_subscription(subscription, group);
 
   return subscription;
 }
--- a/rclcpp/include/rclcpp/create_generic_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_generic_subscription.hpp
@@ -41,9 +41,10 @@ namespace rclcpp
  * \param topic_type Topic type
  * \param qos %QoS settings
  * \param callback Callback for new messages of serialized form
- * \param options %Publisher options.
- * Not all publisher options are currently respected, the only relevant options for this
- * publisher are `event_callbacks`, `use_default_callbacks`, and `%callback_group`.
+ * \param options %Subscription options.
+ * Not all subscription options are currently respected, the only relevant options for this
+ * subscription are `event_callbacks` and `use_default_callbacks`.
+ * \param group Callback group to execute this subscription's callback.
  */
 template<typename AllocatorT = std::allocator<void>>
 std::shared_ptr<GenericSubscription> create_generic_subscription(
--- a/rclcpp/include/rclcpp/create_generic_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_generic_subscription.hpp
@@ -70,6 +70,22 @@ std::shared_ptr<GenericSubscription> create_generic_subscription(
     callback,
     options);
 
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#else
+# pragma warning(push)
+# pragma warning(disable: 4996)
+#endif
+  if (group == nullptr) {
+    group = options.callback_group;
+  }
+#ifndef _WIN32
+# pragma GCC diagnostic pop
+#else
+# pragma warning(pop)
+#endif
+
   topics_interface->add_subscription(subscription, group);
 
   return subscription;
--- a/rclcpp/include/rclcpp/create_generic_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_generic_subscription.hpp
@@ -44,7 +44,7 @@ namespace rclcpp
  * \param options %Subscription options.
  * Not all subscription options are currently respected, the only relevant options for this
  * subscription are `event_callbacks` and `use_default_callbacks`.
- * \param group Callback group to execute this subscription's callback.
+ * \param callback_group Callback group to execute this subscription's callback.
  */
 template<typename AllocatorT = std::allocator<void>>
 std::shared_ptr<GenericSubscription> create_generic_subscription(
@@ -56,7 +56,7 @@ std::shared_ptr<GenericSubscription> create_generic_subscription(
   const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
     rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
   ),
-  rclcpp::CallbackGroup::SharedPtr group = nullptr
+  rclcpp::CallbackGroup::SharedPtr callback_group = nullptr
 )
 {
   auto ts_lib = rclcpp::get_typesupport_library(
@@ -78,8 +78,8 @@ std::shared_ptr<GenericSubscription> create_generic_subscription(
 # pragma warning(push)
 # pragma warning(disable: 4996)
 #endif
-  if (group == nullptr) {
-    group = options.callback_group;
+  if (callback_group == nullptr) {
+    callback_group = options.callback_group;
   }
 #ifndef _WIN32
 # pragma GCC diagnostic pop
@@ -87,7 +87,7 @@ std::shared_ptr<GenericSubscription> create_generic_subscription(
 # pragma warning(pop)
 #endif
 
-  topics_interface->add_subscription(subscription, group);
+  topics_interface->add_subscription(subscription, callback_group);
 
   return subscription;
 }
--- a/rclcpp/include/rclcpp/create_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_subscription.hpp
@@ -62,7 +62,7 @@ create_subscription(
   const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
     rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
   ),
-  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  rclcpp::CallbackGroup::SharedPtr callback_group = nullptr,
   typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (
     MessageMemoryStrategyT::create_default()
   )
@@ -112,7 +112,7 @@ create_subscription(
       std::chrono::duration_cast<std::chrono::nanoseconds>(
         options.topic_stats_options.publish_period),
       sub_call_back,
-      group,
+      callback_group,
       node_topics_interface->get_node_base_interface(),
       node_timer_interface
     );
@@ -143,8 +143,8 @@ create_subscription(
 # pragma warning(push)
 # pragma warning(disable: 4996)
 #endif
-  if (group == nullptr) {
-    group = options.callback_group;
+  if (callback_group == nullptr) {
+    callback_group = options.callback_group;
   }
 #ifndef _WIN32
 # pragma GCC diagnostic pop
@@ -152,7 +152,7 @@ create_subscription(
 # pragma warning(pop)
 #endif
 
-  node_topics_interface->add_subscription(sub, group);
+  node_topics_interface->add_subscription(sub, callback_group);
 
   return std::dynamic_pointer_cast<SubscriptionT>(sub);
 }
@@ -230,7 +230,7 @@ create_subscription(
  * \param qos
  * \param callback
  * \param options
- * \param group
+ * \param callback_group
  * \param msg_mem_strat
  * \return the created subscription
  * \throws std::invalid_argument if topic statistics is enabled and the publish period is
@@ -252,7 +252,7 @@ create_subscription(
   const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
     rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
   ),
-  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  rclcpp::CallbackGroup::SharedPtr callback_group = nullptr,
   typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (
     MessageMemoryStrategyT::create_default()
   )
@@ -260,7 +260,8 @@ create_subscription(
 {
   return rclcpp::detail::create_subscription<
     MessageT, CallbackT, AllocatorT, SubscriptionT, MessageMemoryStrategyT>(
-    node, node, topic_name, qos, std::forward<CallbackT>(callback), options, group, msg_mem_strat);
+    node, node, topic_name, qos, std::forward<CallbackT>(callback), options, callback_group,
+    msg_mem_strat);
 }
 
 /// Create and return a subscription of the given MessageT type.
@@ -310,7 +311,7 @@ create_subscription(
   const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
     rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
   ),
-  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  rclcpp::CallbackGroup::SharedPtr callback_group = nullptr,
   typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (
     MessageMemoryStrategyT::create_default()
   )
@@ -319,7 +320,7 @@ create_subscription(
   return rclcpp::detail::create_subscription<
     MessageT, CallbackT, AllocatorT, SubscriptionT, MessageMemoryStrategyT>(
     node_parameters, node_topics, topic_name, qos,
-    std::forward<CallbackT>(callback), options, group, msg_mem_strat);
+    std::forward<CallbackT>(callback), options, callback_group, msg_mem_strat);
 }
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/create_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_subscription.hpp
@@ -191,7 +191,6 @@ template<
   typename SubscriptionT = rclcpp::Subscription<MessageT, AllocatorT>,
   typename MessageMemoryStrategyT = typename SubscriptionT::MessageMemoryStrategyType,
   typename NodeT>
-[[deprecated("use another overloaded method create_subscription instead")]]
 typename std::shared_ptr<SubscriptionT>
 create_subscription(
   NodeT & node,
@@ -273,14 +272,13 @@ template<
   typename SubscriptionT = rclcpp::Subscription<MessageT, AllocatorT>,
   typename MessageMemoryStrategyT = typename SubscriptionT::MessageMemoryStrategyType>
 typename std::shared_ptr<SubscriptionT>
-[[deprecated("use another overloaded method create_subscription instead")]]
 create_subscription(
   rclcpp::node_interfaces::NodeParametersInterface::SharedPtr & node_parameters,
   rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr & node_topics,
   const std::string & topic_name,
   const rclcpp::QoS & qos,
   CallbackT && callback,
-  const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>&options,
+  const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options,
   typename MessageMemoryStrategyT::SharedPtr msg_mem_strat
 )
 {
--- a/rclcpp/include/rclcpp/create_subscription.hpp
+++ b/rclcpp/include/rclcpp/create_subscription.hpp
@@ -62,6 +62,7 @@ create_subscription(
   const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
     rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
   ),
+  rclcpp::CallbackGroup::SharedPtr group = nullptr,
   typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (
     MessageMemoryStrategyT::create_default()
   )
@@ -111,7 +112,7 @@ create_subscription(
       std::chrono::duration_cast<std::chrono::nanoseconds>(
         options.topic_stats_options.publish_period),
       sub_call_back,
-      options.callback_group,
+      group,
       node_topics_interface->get_node_base_interface(),
       node_timer_interface
     );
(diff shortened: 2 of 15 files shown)


alsora:
Woudln't be a better approach to remove the callback group field from the `SubscriptionOptionsWithAllocator` structure?
Then we can have `create_subscription` function take the callback group as a separate argument.

I think that this makes it much clearer because:
 - you get the same approach as other entities
 - in `create_subscription.hpp` you don't have to be careful at which one to use among `options` and `options_copy`
 - in the subscription code you don't have an empty/unused field

iuhilnehc-ynos:
@alsora

Thank you for your response.

> Woudln't be a better approach to remove the callback group field from the `SubscriptionOptionsWithAllocator` structure?
> Then we can have `create_subscription` function take the callback group as a separate argument.

Yes, make it as other entities did. It's ideal.
(Or make it as weak_ptr, or add a new structure kept in the subscription, etc.)

From my point of view, I just considered not breaking ABI because it affects so many other repositories.

I'd like to hear others' opinions about your recommendation. After that, I'll update it.

alsora:
Ok, IMO we shouldn't be afraid to break API/ABI to have clean and consistent code.
Especially considering that this goes to master and the next release is in 6 months.

If you end up following my suggestion, you will have to update the documentation to mention this breaking changes.

clalancette:
> Ok, IMO we shouldn't be afraid to break API/ABI to have clean and consistent code.

Yes, agreed.  ABI, in particular, we don't guarantee between releases at all, and we always rebuild all downstream packages.  If we are going to break API, we should discuss it a little further and also probably have a deprecation cycle, but we can do that as well.

All of that is to say is that on Rolling, we should do what we need to make the code "best", and then worry about API/ABI more fully if we try to backport it to a released version.

iuhilnehc-ynos:
@clalancette

> Yes, agreed. ABI, in particular, we don't guarantee between releases at all, and we always rebuild all downstream packages. If we are going to break API, we should discuss it a little further and also probably have a deprecation cycle, but we can do that as well.
> 
> All of that is to say is that on Rolling, we should do what we need to make the code "best", and then worry about API/ABI more fully if we try to backport it to a released version.

Thank you, got it. To separate the `callback_group` will be my final choice if there is no other way.

After re-consideration, I think to separate the `callback_group` from `SubscriptionOptions` seems unreasonable because it's an option of `SubscriptionOptions` for users.

What about saving the necessary variable in the subscription, what do you think? @wjwwood 
From
https://github.com/ros2/rclcpp/blob/0775e2f6e7b53b5b5cab1702e4f3fa1206f3b59b/rclcpp/include/rclcpp/subscription.hpp#L396-L401
to
```patch
    * It is important to save a copy of this so that the rmw payload which it
    * may contain is kept alive for the duration of the subscription.
    */
-  const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> options_;
+  std::shared_ptr<rclcpp::detail::RMWImplementationSpecificSubscriptionPayload>
+    rmw_implementation_payload_;
```
Notice that `GenericSubscription` does not copy the options.

alsora:
> After re-consideration, I think to separate the callback_group from SubscriptionOptions seems unreasonable because it's an option of SubscriptionOptions for users.

This is a rather arbitrary consideration IMO. There isn't a clear distinction of what should go into that data-structure and what not.

Anyhow, I'm ok with your proposal of only storing the rmw_payload rather than the full options.
This assumes that after construction we don't need anymore the options for "regular operations".

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -173,6 +174,57 @@ template<
   typename SubscriptionT = rclcpp::Subscription<MessageT, AllocatorT>,
   typename MessageMemoryStrategyT = typename SubscriptionT::MessageMemoryStrategyType,
   typename NodeT>
+[[deprecated("use another overloaded method create_subscription instead")]]
```
alsora, line 177: we should mention that the purpose of the deprecation is to have a separate argument for the callback group

rclcpp/test/rclcpp/test_memory_strategy.cpp
```diff
@@ -362,12 +356,13 @@ TEST_F(TestMemoryStrategy, get_group_by_subscription) {
         memory_strategy()->get_group_by_subscription(subscription, weak_groups_to_nodes));
     }  // callback_group goes out of scope
     EXPECT_EQ(
-      callback_group,
+      nullptr,
       memory_strategy()->get_group_by_subscription(subscription, weak_groups_to_nodes));
   }  // Node goes out of scope
-  // NodeBase(SubscriptionBase->rcl_node_t->NodeBase) is still alive.
+  // NodeBase(SubscriptionBase->rcl_node_t->NodeBase) is still alive,
```
alsora, line 362: The comment above says "node goes out of scope", why this comment says that node is still alive?

iuhilnehc-ynos: This comment says the NodeBase is still alive, not Node, because the SubscriptionBase holds the rcl_node_t which holds the NodeBase.

rclcpp/include/rclcpp/create_subscription.hpp
iuhilnehc-ynos: I think the message in the deprecated can be either the reason or suggestion, what about adding the reason  [here](https://github.com/ros2/rclcpp/blob/126eeee3aaecb94e088fff24ee24a3743f898755/rclcpp/include/rclcpp/create_subscription.hpp#L146-L153)?

iuhilnehc-ynos: I'll follow your suggestion, what about
```suggestion
[[deprecated(
  "To Keep consistency with other cases(Client, Service, Timer, etc) not hold callback_group,\n" \
  "the callback_group in the SubscriptionOptionsBase will be removed in the future since the\n" \
  "SubscriptionOptions is held in the Subscription and then the callback_group is depend\n" \
  "on the lifetime of Subscription.\n" \
  "Use\n" \
  "  template<\n" \
  "    typename MessageT,\n" \
  "    typename CallbackT,\n" \
  "    typename AllocatorT = std::allocator<void>,\n" \
  "    typename SubscriptionT = rclcpp::Subscription<MessageT, AllocatorT>,\n" \
  "    typename MessageMemoryStrategyT = typename SubscriptionT::MessageMemoryStrategyType\n" \
  "    typename NodeT>\n" \
  "  typename std::shared_ptr<SubscriptionT>\n" \
  "  create_subscription(\n" \
  "    NodeT & node,\n" \
  "    const std::string & topic_name,\n" \
  "    const rclcpp::QoS & qos,\n" \
  "    CallbackT && callback,\n" \
  "    const SubscriptionOptionsWithAllocator<AllocatorT> & options =(\n" \
  "      SubscriptionOptionsWithAllocator<AllocatorT>()\n" \
  "    ),\n" \
  "    rclcpp::CallbackGroup::SharedPtr group = nullptr,\n" \
  "    typename MessageMemoryStrategyT::SharedPtr msg_mem_strat = (\n" \
  "      MessageMemoryStrategyT::create_default()\n" \
  "    )\n" \
  "  );\n" \
  "instead"
)]]
```

fujitatomoya (commented):
https://github.com/ros2/rclcpp/blob/ee20dd31e676c4803e2393bc86ce76f0a21f9b73/rclcpp/include/rclcpp/parameter_client.hpp#L214-L215

this also needs to be updated accordingly?

the same fix needs to be applied to `rclcpp::PublisherOptionsBase` as follow-up? other entities have dedicated argument for `rclcpp::CallbackGroup`, if i am not missed.

fujitatomoya:
note for myself, this must be tested with `--packages-above-and-dependencies rclcpp rclcpp_lifecycle` in CI.

iuhilnehc-ynos:
> https://github.com/ros2/rclcpp/blob/ee20dd31e676c4803e2393bc86ce76f0a21f9b73/rclcpp/include/rclcpp/parameter_client.hpp#L214-L215
> 
> this also needs to be updated accordingly?

Thank you, @fujitatomoya. I'll update it.

> the same fix needs to be applied to `rclcpp::PublisherOptionsBase` as follow-up? other entities have dedicated argument for `rclcpp::CallbackGroup`, if i am not missed.

Yes, I'll update it in the next PR if this one can be merged.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15817/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10495/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13481/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16102/)

iuhilnehc-ynos:
rerun CI from https://github.com/ros2/rclcpp/pull/1833#issuecomment-999052020

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15825/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10503/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13489/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16110/)

The downstream repositories have the deprecated warning message as we expected.
(NOTE: I can update them after it's merged. If someone would like to fix them, I'd appreciate it.)

fujitatomoya:
> The downstream repositories have the deprecated warning message as we expected.
> (NOTE: I can update them after it's merged. If someone would like to fix them, I'd appreciate it.)

(downstream packages are in ros2 repository, so we should address them at the same time, i guess.) anyway, all done.

New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15878/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10555/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13544/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16181/)

fujitatomoya:
CI(retry):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15907/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10582/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13571/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16229/)

iuhilnehc-ynos:
A failure test case in the CI. Log as below,
```
[ RUN      ] TestTimeSource.clock_sleep_until_with_ros_time_basic
  what():  Resource deadlock avoided
```
I need to reproduce it and confirm if it was introduced by this PR.

Barry-Xu-2018:
```
[ RUN      ] TestTimeSource.clock_sleep_until_with_ros_time_basic
  what():  Resource deadlock avoided
```

This problem isn't introduced by this PR.   
I can reproduce this on master branch.  
A new issue #1861 is created for this issue.

fujitatomoya:
@clalancette 

this PR has to go with the following PRs.

- https://github.com/ros2/examples/pull/335
- https://github.com/ros2/geometry2/pull/492
- https://github.com/ros2/system_tests/pull/492

(we do not have permission above repos.)

everything seems to be okay with https://github.com/ros2/rclcpp/pull/1833#issuecomment-10075 including dependencies.

Could you give us (@fujitatomoya @iuhilnehc-ynos @Barry-Xu-2018 ) the access to the above repos? or we could wait for the maintainers response.

ralph-lange:
I tested and reviewed the connected PR https://github.com/ros2/examples/pull/335. I would like to add that I support this API change: All fields of the `SubscriptionOptions` are about the communication, i.e., oriented to the interface of the node whereas the callback group is more a node internal aspect and related to execution management. Therefore, it is only logical to make the callback group a separate parameter of `create_subscription`.

fujitatomoya:
@clalancette can you merge this one with following dependent repositories? (I do not have access to merge the following repos.)

- https://github.com/ros2/examples/pull/335
- https://github.com/ros2/geometry2/pull/492
- https://github.com/ros2/system_tests/pull/492

all 4 PRs should be merged at the same time.

rclcpp/include/rclcpp/create_generic_subscription.hpp
```diff
@@ -54,7 +55,8 @@ std::shared_ptr<GenericSubscription> create_generic_subscription(
   std::function<void(std::shared_ptr<rclcpp::SerializedMessage>)> callback,
   const rclcpp::SubscriptionOptionsWithAllocator<AllocatorT> & options = (
     rclcpp::SubscriptionOptionsWithAllocator<AllocatorT>()
-  )
+  ),
+  rclcpp::CallbackGroup::SharedPtr group = nullptr
```
wjwwood, line 59: This variable should be `callback_group`, not just `group`, in my opinion.

wjwwood: And everywhere else you added this argument.

wjwwood (commented):
So, if the issue is that the subscription is keeping the callback group in scope via the subscription options, have we discussed either:

- storing the callback group in the subscription options as a weak pointer (maybe ok), or
- resetting the callback group pointer for the subscription option after storing it in the subscription (probably not ok), or
- not storing the options in the subscription class (not so desirable, but maybe ok)

I was just wondering because I haven't been able to keep up on the conversation completely. Sorry if it was already discussed.

rclcpp/include/rclcpp/create_generic_subscription.hpp
iuhilnehc-ynos: > storing the callback group in the subscription options as a weak pointer (maybe ok)

Yes, this is an alternative way.
To keep the same method with calling `create_wall_timer`, `create_client` and `create_service` by `callback_group`, I think to put the `callback_group` in the function argument could be acceptable as well.

> This variable should be callback_group, not just group, in my opinion.

OK. In addition to updating it imported by this PR, I'll update other variable names (e.g. [1](https://github.com/ros2/rclcpp/blob/16914e31a15417d5751c4cb611f591bbaa458eca/rclcpp/include/rclcpp/node.hpp#L241), [2](https://github.com/ros2/rclcpp/blob/16914e31a15417d5751c4cb611f591bbaa458eca/rclcpp/include/rclcpp/node.hpp#L255), etc.) that should be `callback_group`.

iuhilnehc-ynos:
Rebased and updated the variable name `group` imported by this PR.

rclcpp/include/rclcpp/create_generic_subscription.hpp
iuhilnehc-ynos: Updated in https://github.com/ros2/rclcpp/pull/1833/commits/a9601ed095b99190a022d9683c37a7591501abf9

alsora (approved):

fujitatomoya:
@clalancette could you take a look at previous comment https://github.com/ros2/rclcpp/pull/1833#issuecomment-1053173876?

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16662/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11248/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17057/)

clalancette:
@fujitatomoya I can go take a look and merge them all in, but do you know why CI is failing in https://github.com/ros2/rclcpp/pull/1833#issuecomment-1120088136 ?

fujitatomoya:
@clalancette test repo file is old, i will rebase the repo file and restart the CI against master branch.

fujitatomoya:
CI(retry with https://gist.githubusercontent.com/fujitatomoya/bb7f9773ee426b740bf9af56f48d2598/raw/d48ac2c96b48f37e2916ca365a97635328929441/ros2.repos):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16667/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11253/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17062/)

wjwwood (changes requested):
Maybe I missed the reason why, but unless there is a reason for it, the new signatures added to support the old argument order need a deprecation warning in code.

fujitatomoya:
@iuhilnehc-ynos besides @wjwwood previous comments, i think we need to rebase this. i will do the dependent packages accordingly.

rclcpp/include/rclcpp/create_subscription.hpp
```diff
@@ -150,6 +168,8 @@ create_subscription(
  * NodeT must also have a method called get_node_parameters_interface()
  * which returns a shared_ptr to a NodeParametersInterface.
  *
+ * \deprecated use another overloaded method create_subscription instead.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: template can't be deprecated.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -201,6 +201,8 @@ class Node : public std::enable_shared_from_this<Node>
 
   /// Create and return a Subscription.
   /**
+   * \deprecated use another overloaded method create_subscription instead.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: template can't be deprecated.

rclcpp/include/rclcpp/create_subscription.hpp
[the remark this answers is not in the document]
wjwwood: Are you sure? I can seem to do it in this example:

https://godbolt.org/z/5hbbfzsxn

I know there are some corner cases where it doesn't work ideally or as well, but can you be specific? In general templates _can_ be deprecated.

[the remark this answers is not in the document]
iuhilnehc-ynos: Ahhh, I used the wrong syntax before, 
```
    template<class T>
-    [[deprecated("use create_thing(string, bool) instead")]]
    void
 +   [[deprecated("use create_thing(string, bool) instead")]]
    create_thing(bool option, std::string name)
    {

    }
```
and I got the wrong answer [here](https://stackoverflow.com/questions/25984196/why-cant-templates-be-deprecated-with-deprecated).

Sorry, I'll update them.

iuhilnehc-ynos:
retry CI based on https://github.com/ros2/rclcpp/pull/1833#issuecomment-1121347305

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16671/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11257/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17066/)

iuhilnehc-ynos:
The failed test can't be reproduced in my local env(ubuntu20, ubuntu22).

`colcon test --packages-select rclcpp --ctest-args -R test_publisher_subscription_count_api`

fujitatomoya:
there are 3 more repositories to address this deprecation. i will do that later.

fujitatomoya:
CI (https://gist.githubusercontent.com/fujitatomoya/275ad49b0cdb2603b04a2f87a3022ce7/raw/e8b389df88e75cd2264c8572a14ba710958d5028/ros2.repos):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16677/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11263/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17074/)

fujitatomoya:
windows failure https://ci.ros2.org/job/ci_windows/17074/ is unrelated.

@wjwwood @clalancette 

I think this is ready to merge, the followings must go with this.

- https://github.com/ros2/examples/pull/335
- https://github.com/ros2/geometry2/pull/492
- https://github.com/ros2/system_tests/pull/492
- https://github.com/ros2/realtime_support/pull/115
- https://github.com/ros2/demos/pull/563

wjwwood (changes requested):
The changes lgtm, with a few exceptions that I commented on.

Also we need to document this. I've asked folks where it should be documented, we're in a bit of a strange state at the moment where we don't want to put it in humble, but we don't have an I-turtle release page yet so we cannot put it under the "changes since the last release" section there yet. I'll update when I have an idea where to put it.

---

Finally, after re-reviewing this and seeing the changes (that we realized were needed after properly adding the deprecation warnings) and after my comments here (https://github.com/ros2/realtime_support/pull/115#pullrequestreview-977761778), I'm questioning if this is the right direction to take.

It's true that this makes subscriptions "more like" the other entities, but I think that's less to do with them being right and subscription being wrong and more to do with the fact that subscriptions were the first to get their own options structure due to the number and complexity of them.

So I think a better alternative might be to instead change the option classes to add (yet another) layer to them. That would allow for us to fix this problem without new signatures of `create_subscription()` and without any deprecations (maybe). What I was imagining was a new base class for the subscription (and publisher) options which did not contain problematic options (from a lifespan perspective) and the subscription would hold on to that base class while the user would use the full derived class to set up the options.

Also, it occurs to me that really the callback group and msg mem strat are executor specific, which is related to a shift in class layout that needs to happen to make wait set based use of our entities more natural, so really we should have like `ExecutableSubscriptionOptions` inherit from `SubscriptionOptionsWithAllocator` inherit from `SubscriptionOptionsBase`, with a user-friendly `SubscriptionOptions` which is just an alias to the `ExecutableSubscriptionOptions`.

Something like this sketch:

```c++
// everything except the allocator and the callback group/msg mem strat
class SubscriptionOptionsBase;
// all the normal subscription options and the allocator
template<typename AllocatorT>
class SubscriptionOptionsWithAllocator<AllocatorT>
: public SubscriptionOptionsBase;
// all the normal subscription options plus things needed
// to use a subscription with an executor
template<typename AllocatorT>
class ExecutableSubscriptionOptions<AllocatorT>
: public SubscriptionOptionsWithAllocator<AllocatorT>;

{
  // user code
  SubscriptionOptions so;
  so.callback_group = custom_callback_group;
  so.message_memory_strategy = custom_mms;
  auto sub = node->create_subscription<...>("...", 10, my_callback, so);
  // could also support, but maybe deprecate (would affect less users):
  // node->create_subscription<...>("...", 10, my_callback, so, custom_mms);

  SubscriptionOptionsBase after = sub->get_subscription_options();
}
```

I won't block this pr on this suggestion, but I do think it's the better way to solve the problem. As we add more options the awkwardness of passing nullptr's for unused settings will just get worse.

iuhilnehc-ynos:
@wjwwood 
To hold a callback group in the subscription is not a big deal. Considering this PR is not good enough, I decide to close it rather than merge it.

@fujitatomoya
Thanks for your help. I am sorry about [these PRs](https://github.com/ros2/rclcpp/pull/1833#issuecomment-1123141254).

I'll create another PR if I can use an elegant method to fix this issue in the future.

fujitatomoya:
> As we add more options the awkwardness of passing nullptr's for unused settings will just get worse.

agree on this.

> Thanks for your help. I am sorry about https://github.com/ros2/rclcpp/pull/1833#issuecomment-1123141254.

no problem. those are just adjustments to suppress the warnings.

maybe we can come back on this issue with different approach.

PR #1840, opened 2021-12-08
Revert "create component_container_isolated (#1781)"

Blast545:
Opening this to test if indeed #1781  is causing the windows regression.

This reverts commit 321c74c2b3ae0336528f561a039fd554a28f4543.


Blast545:
[Build Status](https://ci.ros2.org/job/ci_windows/15997/)

SteveMacenski:
@Blast545 can you fix the issue rather than simply reverting it? I think the major issue (if my understanding is correct) is that implementations need to be moved into the header or the cpp file needs to create instantiations of each ExecutorT type (e.x. https://github.com/ros-planning/navigation2/blob/main/nav2_smac_planner/src/a_star.cpp#L426-L429). Probably the first is best since these functions are so simple.

clalancette:
> @Blast545 can you fix the issue rather than simply reverting it?

Blash545's role is generally in keeping the buildfarm green.  If the fix is easy enough, that can mean fixing it forward, but if it is involved it generally means a revert.

That is to say, someone involved with the original PR (or at least involved with rclcpp) really needs to look at it and fix it.  I would generally jump on these things, but I don't really have time for the rest of this week, and we can't leave Windows broken for days.  So if @ivanpauno or @gezp can't look at it today, I think a revert is in order.

SteveMacenski:
Fair enough. I don't have a Windows environment to test in, so what I said above would be my best guess as to the solution, but I couldn't know until running CI. I can submit a PR in a couple of hours for this if that would be helpful -- else we can wait for @gezp

clalancette:
> Fair enough. I don't have a Windows environment to test in (nor have I ever developed in Windows either), so what I said above would be my best guess as to the solution, but I couldn't know until running CI. I can submit a PR in a couple of hours for this if that would be helpful -- else we can wait for @gezp / revert.

That's fine.  I would say we can go one more day with this, so let's wait for @gezp .  If we don't have a fix forward by tomorrow morning Pacific time, we'll do a revert.

gezp:
i think i solved this issue, please see more details here https://github.com/ros2/rclcpp/pull/1843, i have tested on my local win10 environment, it could be built  successfully.

clalancette:
Closing, as this was fixed forward by #1843

PR #1849, opened 2021-12-20
Add non transform capabilities for intra-process

gonzodepedro:
Signed-off-by: Gonzalo de Pedro <gonzalo@depedro.com.ar>


rclcpp/test/rclcpp/test_publisher_with_type_adapter.cpp
```diff
@@ -166,17 +166,19 @@ TEST_F(TestPublisher, various_creation_signatures) {
 
 /*
  * Testing that conversion errors are passed up.
+ * TODO. This test was removed when implementing non convering intra-proccess
+ * publishing. It may be fixed and added again later.
```
clalancette, line 170: There is quite a good reason why we have to disable this test now, though it is slightly subtle.

Previous to this PR, anytime we used the TypeAdapt functionality in the rclcpp intra-process manager, we always converted immediately from a TypeAdapt type (i.e. `std::string`) to a ROS message type (i.e. `std_msgs::msg::String`).  Because of that, we always knew immediately if a TypeAdapted type was viable immediately on publish (regardless of whether there were subscribers), and hence this test made sense.

After this PR, we only convert to a ROS message type if one of the subscribers requires us to do that.  This is more efficient since we save a conversion to and from the custom type if the publisher and subscriber are of the same custom type and are using the intra-process manager to communicate.  However, this means that in the absence of any subscribers, we no longer attempt the conversion so this never throws.

There may be a way to keep this test with some additional fixes, but we haven't found a way to do that yet.  Another way to fix this is to add some subscribers of the ROS message type, which will cause the conversion to happen.

rclcpp/include/rclcpp/publisher.hpp
```diff
@@ -319,13 +319,28 @@ class Publisher : public PublisherBase
...
+    bool inter_process_publish_needed =
+      get_subscription_count() > get_intra_process_subscription_count();
+
+    if (inter_process_publish_needed) {
+      auto shared_msg = this->do_intra_process_publish_and_return_shared(std::move(msg));
+      // TODO(clalancette): This is unnecessarily doing an additional conversion
+      // that may have already been done in do_intra_process_publish_and_return_shared().
+      // We should just reuse that effort.
+      ROSMessageType ros_msg;
+      rclcpp::TypeAdapter<MessageT>::convert_to_ros_message(*msg, ros_msg);
```
clalancette, line 339: While doing some testing, I realized that this is actually buggy.  The problem is that we have already `std::move` the msg in `do_intra_process_publish_and_return_shared` above, which means that we cannot safely access it here.  This code should instead be:

```
      ROSMessageType ros_msg;
      // TODO(clalancette): This is unnecessarily doing an additional conversion
      // that may have already been done in do_intra_process_publish_and_return_shared().
      // We should just reuse that effort.
      rclcpp::TypeAdapter<MessageT>::convert_to_ros_message(*msg, ros_msg);
      auto shared_msg = this->do_intra_process_publish_and_return_shared(std::move(msg));
```

clalancette:
For context, the point of this PR is to enhance [TypeAdaptation](https://ros.org/reps/rep-2007.html) so that it avoids unnecessary conversions.  In particular, if a Publisher and Subscription are both in the same context, using intra-process communications, and use the same custom TypeAdapted type, then we can just place the published data directly from the publisher into the Subscription queue for delivery without doing any conversions.  If any of several other situations happen (Publisher is ROS message, Subscription is custom, Publisher is custom, Subscription is ROS message, etc), then we do the conversion as necessary.

This has two primary benefits, both related to performance:
1. We can avoid doing the (possibly) expensive and unnecessary to a ROS type on the publisher, and then from the ROS type on the subscription.
2. We can use a custom type to hold onto custom "handles", allowing the Publisher to pass those handles directly to the Subscription.  Those handles might be handles to hardware accelerators, which means that the system can use the hardware accelerator to do the conversion without synchronizing with the CPU to publish to the next node.

clalancette:
Here's preliminary CI to see where we stand:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15809/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10487/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13471/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16090/)

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -203,22 +205,18 @@ class IntraProcessManager
       // None of the buffers require ownership, so we promote the pointer
       std::shared_ptr<MessageT> msg = std::move(message);
 
-      this->template add_shared_msg_to_buffers<MessageT, Alloc, Deleter>(
+      this->template add_shared_msg_to_buffers<MessageT, Alloc, Deleter, ROSMessageType>(
         msg, sub_ids.take_shared_subscriptions);
     } else if (!sub_ids.take_ownership_subscriptions.empty() && // NOLINT
       sub_ids.take_shared_subscriptions.size() <= 1)
     {
-      // There is at maximum 1 buffer that does not require ownership.
```
alsora, line 211: why removing this comment? I think it still applies

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -352,17 +362,46 @@ class IntraProcessManager
       auto subscription_base = subscription_it->second.lock();
       if (subscription_base) {
         auto subscription = std::dynamic_pointer_cast<
-          rclcpp::experimental::SubscriptionIntraProcessBuffer<MessageT, Alloc, Deleter>
+          rclcpp::experimental::SubscriptionIntraProcessBuffer<PublishedType,
+          PublishedTypeAllocator, PublishedTypeDeleter, ROSMessageType>
           >(subscription_base);
         if (nullptr == subscription) {
```
alsora, line 368: this code is very complex to read due the abundance of nested conditions.
I recommend to re-organize it to reduce nesting, for example by inverting many checks and completely resolving some code paths as soon as possible.

For example:
```
if (!subscription_base) {
  subscriptions_.erase(id);
  continue;
}

auto subscription = std::dynamic...
if (subscription) {
  subscription->provide_intra_process_data(message);
  continue;
}

auto ros_message_subscription = std::dynamic...

```

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -390,28 +439,64 @@ class IntraProcessManager
...
-          // If this is the last subscription, give up ownership
-          subscription->provide_intra_process_message(std::move(message));
+          auto ros_message_subscription = std::dynamic_pointer_cast<
+            rclcpp::experimental::ROSMessageIntraProcessBuffer<ROSMessageType,
+            ROSMessageTypeAllocator, ROSMessageTypeDeleter>
+            >(subscription_base);
+
+          if (nullptr == ros_message_subscription) {
+            throw std::runtime_error(
+                    "--failed to dynamic cast SubscriptionIntraProcessBase to "
```
alsora, line 453: nitpick: remove copy pasted `--`

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -352,17 +362,46 @@ class IntraProcessManager
...
-                  "allocator types, which is not supported");
+          auto ros_message_subscription = std::dynamic_pointer_cast<
+            rclcpp::experimental::ROSMessageIntraProcessBuffer<ROSMessageType,
+            ROSMessageTypeAllocator, ROSMessageTypeDeleter>
+            >(subscription_base);
+
+          if (nullptr == ros_message_subscription) {
+            throw std::runtime_error(
+                    "failed to dynamic cast SubscriptionIntraProcessBase to "
+                    "SubscriptionIntraProcessBuffer<MessageT, Alloc, Deleter>, which "
```
alsora, line 377: this is not the correct failure.
We definitely failed to do that,but we also failed the `ros_message_subscription` cast.

The same applies to the runtime error in the following function

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -390,28 +439,64 @@ class IntraProcessManager
...
+              auto ptr = ros_message_alloc.allocate(1);
+              ros_message_alloc.construct(ptr);
+              ROSMessageTypeDeleter deleter;
+              allocator::set_allocator_for_deleter(&deleter, &allocator);
+              rclcpp::TypeAdapter<MessageT>::convert_to_ros_message(*message, *ptr);
+              auto ros_msg = std::unique_ptr<ROSMessageType, ROSMessageTypeDeleter>(ptr, deleter);
+              ros_message_subscription->provide_intra_process_message(std::move(ros_msg));
+            } else {
+              if constexpr (std::is_same<MessageT, ROSMessageType>::value) {
+                if (std::next(it) == subscription_ids.end()) {
```
alsora, line 469: this code block is almost identically copied 3 times within the function.
it may be better to make it a lambda or a standalone function

rclcpp/include/rclcpp/experimental/subscription_intra_process_buffer.hpp
```diff
@@ -15,50 +15,61 @@
...
-  typename Deleter = std::default_delete<MessageT>
+  typename SubscribedType,
+  typename Alloc = std::allocator<SubscribedType>,
+  typename Deleter = std::default_delete<SubscribedType>,
+  /// MessageT::ros_message_type if MessageT is a TypeAdapter,
+  /// otherwise just MessageT.
+  typename ROSMessageType = SubscribedType
 >
-class SubscriptionIntraProcessBuffer : public SubscriptionIntraProcessBase
+class SubscriptionIntraProcessBuffer : public ROSMessageIntraProcessBuffer<ROSMessageType,
```
alsora, line 47: I'm not sure I understand the logic with the hierarchy in the Subscription IPC class.
We have that:

`SubscriptionIntraProcess` is a
`SubscriptionIntraProcessBuffer` is a
`ROSMessageIntraProcessBuffer` is a
`SubscriptionIntraProcessBase` is a
`Waitable`

Do we really need all these layers?

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
```diff
@@ -14,9 +14,18 @@
 
 #include "rclcpp/node_interfaces/node_topics.hpp"
 
+#include <stdexcept>
```
alsora, line 17: throughout the PR there are a lot of changes to unrelated files where you fix some header include issues.
This PR is already very complex, so I would suggest to undo these changes and open a separate PR for them

alsora:
I just gave a quick look at this from a style point of view.
I need to take a little bit more time to comment on the actual implementation, but my first impression seems good to me.

However, I want to point out that I'm a little bit worried about the increase in complexity caused by adding these new capabilities.
I wonder how will we implement the next feature (e.g. intra-process support for loaned messages), i.e. if we will keep having to exponentially grow all those if-else conditions in various parts of the codebase.

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
clalancette: That's my fault, I noticed them while we were developing them and just fixed them.  I'm fine with pulling them out into a separate PR.

clalancette:
(force-pushed to resolve DCO)

rclcpp/test/rclcpp/test_subscription_publisher_with_same_type_adapter.cpp
```diff
@@ -0,0 +1,516 @@
...
+  executor.spin_once(std::chrono::milliseconds(0));
+  int i = 0;
+  while (!is_received && i < g_max_loops) {
+    printf("spin_node_once() - callback (1) expected - try %d/%d\n", ++i, g_max_loops);
+    executor.spin_once(g_sleep_per_loop);
+  }
+}
+
+/*
+ * Testing that subscriber receives type adapted types and ROS message types with intra proccess communications.
```
audrow, line 95: It would be nice to summarize all of the cases that are checked in this test.

audrow (approved):
A few nitpicks but the code and tests look good to me. LGTM with green CI for all packages above rclcpp.

clalancette:
All right, with the latest fixes on macOS, let's see what CI here looks like:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15835/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10512/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13500/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16120/)

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -476,6 +476,21 @@ class AnySubscriptionCallback
...
+  {
+    if constexpr (rclcpp::TypeAdapter<MessageT>::is_specialized::value) {
+      auto ptr = ROSMessageTypeAllocatorTraits::allocate(ros_message_type_allocator_, 1);
+      ROSMessageTypeAllocatorTraits::construct(ros_message_type_allocator_, ptr);
+      rclcpp::TypeAdapter<MessageT>::convert_to_ros_message(msg, *ptr);
+      return std::unique_ptr<ROSMessageType, ROSMessageTypeDeleter>(ptr, ros_message_type_deleter_);
+    } else {
+      throw std::runtime_error(
+              "convert_custom_type_to_ros_message_unique_ptr "
+              "unexpectedly called without TypeAdapter");
```
ivanpauno, line 490: I would rather show an error at build time:

```cpp
// helper added somewhere
namespace detail
{
template<class> inline constexpr bool always_false_v = false;
}
```

and then:

```suggestion
      static_assert(
          rclcpp::detail::always_false_v<ROSMessageType>,
          "convert_custom_type_to_ros_message_unique_ptr() "
          "unexpectedly called without specialized TypeAdapter");
```

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -677,65 +692,89 @@ class AnySubscriptionCallback
...
         {
-          callback(convert_ros_message_to_custom_type_unique_ptr(*message), message_info);
+          callback(message, message_info);
         }
         // conditions for ros message type
-        else if constexpr (std::is_same_v<T, ConstRefROSMessageCallback>) {  // NOLINT
-          callback(*message);
-        } else if constexpr (std::is_same_v<T, ConstRefWithInfoROSMessageCallback>) {
-          callback(*message, message_info);
+        else if constexpr (std::is_same_v<T, ConstRefROSMessageCallback>) {  // NOLINT[readability/braces]
```
ivanpauno, line 728: this case cannot happen, right?

Now for intraprocess communication, the intraprocess buffer always stores instances of `SubscribedType`, and that will match the type specified in the callback.
We don't need a special code path for this case AFAIS.

I also don't think the [possible types](https://github.com/ros2/rclcpp/blob/02e0d118a0f65205f7e8e25bf6b79e16955e66ca/rclcpp/include/rclcpp/any_subscription_callback.hpp#L56) should have two specializations for ROS messages and type adapters, the possible callbacks are the same in both cases and the argument is always a kind of reference/pointer to SubscribedType.

In the only case we need to aware of the "ROS message type" is in interprocess dispatch.

gonzodepedro: Even if the buffer is of SubscribedType, the callback may require a ROSMessage.

ivanpauno: > Even if the buffer is of SubscribedType, the callback may require a ROSMessage.

How?
We're creating a buffer type based on the callback type.

i.e. if your callback is a `smart_ptr<ROSMessage>` or a `ROSMessage &` the buffer will also be a smart pointer to a ROSMessage.

Thus, all those code paths seem unnecessary, and we do have a lot of code paths!

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
clalancette: This should be fixed in the latest commits.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
clalancette: https://github.com/ros2/rclcpp/pull/1849/commits/1779cd820a78763c867ad8183166c7025e5285fa refactors this in the way you suggested, by getting out early whenever we can.  I'm going to mark this conversation as resolved, since it is much easier to read now.  But feel free to comment again and unresolve the conversation if you disagree.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
clalancette: We're now down to 2 copies, and they aren't exactly the same.  I'm going to mark this as resolved for now, but feel free to comment and unresolve if you disagree with the resolution.

rclcpp/src/rclcpp/node_interfaces/node_topics.cpp
clalancette: I've now removed the unrelated changes, and I'll submit them separately from this PR.

rclcpp/include/rclcpp/experimental/subscription_intra_process_buffer.hpp
clalancette: > Do we really need all these layers?

I agree that this is confusing, but I think we do need all of these layers.

So at the base, the whole thing needs to be a `Waitable` because it needs to be waited on by the executors.  We need the `SubscriptionIntraProcessBase` as a non-templated type so that we can store essentially untyped subscriptions in the `IntraProcessManager` (https://github.com/ros2/rclcpp/blob/c8c4851b53de49cc1fe2f3fc3d36ef3e2df8ffee/rclcpp/include/rclcpp/experimental/intra_process_manager.hpp#L319-L320).  The split between `ROSMessageIntraProcessBuffer` and `SubscriptionIntraProcessBuffer` is how the subscriber ultimately communicates whether it is a ROS type or not to the `IntraProcessManager`.  So we need both of those layers to distinguish between those two cases.  And finally, we need `SubscriptionIntraProcess` as the thing that the `Subscription` class interacts with.

I'll note that this hierarchy was largely here already; the only thing this PR does is to add in the `ROSMessageIntraProcessBuffer`.

One thing we *might* be able to do is collapse the `SubscriptionIntraProcess` and `SubscriptionIntraProcessBuffer` classes into one.  Neither of them does too much on its own, and it isn't entirely clear why they are separated.  If you'd like me to try to do that, I'll give that a whirl.  I think we need the rest of the layers.

rclcpp/include/rclcpp/any_subscription_callback.hpp
ivanpauno: I don't think that example makes sense, specifying `StringTypeAdapter` is not really useful if the callback is expecting a `msg::String`.
i.e. what's the benefit of doing that instead of just using `create_subscription<msg::String>`?
If there's no benefit, we shouldn't support it.
It adds complexity because you need to keep track of more combinations in the implementation.

I also think we cannot really allow that kind of code if we want the compiler to be able to infer the correct `Subscription` type from the callback type in `Node::create_subscription()`.

i.e. being able to write 
```cpp
auto sub = node->create_subscription(topic_name, 1, callback);  // subscription type inferred from callback type
```

I think there's no way to infer it if both a `Subscription<msg::String>` and a `Subscription<StringTypeAdapter>` accept a `msg::String` callback. Though I might be wrong, c++ templates and specialization rules are complex and I cannot really remember them.
I have also not kept track of what's currently possible, I have a few questions:
- is it possible to mix callback and subscription type as the example above?
- is automatic template parameter inference still working for `create_subscription()`? IIUC, that used to work (*).

If the answer to both those question is yes, my only reason to not support that is that I don't see the benefit and it adds complexity.
I guess we can ignore that, but if we later want to not support that to reduce complexity it will be an API breaking change.


(*) If we currently support code as the example above, I guess it will only break if no type is passed and a type adapter is available, though it will still work if there's no type adapter for that message type.
I can also imagine a long and impossible to understand compiler error, which is also something to check and consider.

ivanpauno: > is automatic template parameter inference still working for create_subscription()? IIUC, that used to work (*).

Sorry, my memory is bad, I think we never supported this.
My concern about breaking template parameter inference can be ignored.

The only part I think that it's still valid from my last comment is that supporting that example seems unnecessary, but considering that we already do support it, I think we can move forward.

rclcpp/include/rclcpp/any_subscription_callback.hpp
```diff
@@ -485,9 +485,10 @@ class AnySubscriptionCallback
       rclcpp::TypeAdapter<MessageT>::convert_to_ros_message(msg, *ptr);
       return std::unique_ptr<ROSMessageType, ROSMessageTypeDeleter>(ptr, ros_message_type_deleter_);
     } else {
-      throw std::runtime_error(
-              "convert_custom_type_to_ros_message_unique_ptr "
-              "unexpectedly called without TypeAdapter");
+      static_assert(
```
audrow, line 488: Nice, I like making this compile time rather than run time.

rclcpp/include/rclcpp/experimental/intra_process_manager.hpp
```diff
@@ -397,72 +428,85 @@ class IntraProcessManager
...
-                Deleter deleter = message.get_deleter();
-                auto ptr = MessageAllocTraits::allocate(allocator, 1);
-                MessageAllocTraits::construct(allocator, ptr, *message);
-                copy_message = MessageUniquePtr(ptr, deleter);
-
-                ros_message_subscription->provide_intra_process_message(std::move(copy_message));
-              }
-            }
-          }
+      if (subscription_base == nullptr) {
```
audrow, line 447: Nice, all this is much cleaner.

rclcpp/include/rclcpp/experimental/subscription_intra_process_buffer.hpp
audrow: > I'll note that this hierarchy was largely here already; the only thing this PR does is to add in the `ROSMessageIntraProcessBuffer`.

Do you think that it makes sense to rename `ROSMessageIntraProcessBuffer` to something like `SubscriptionROSMessageIntraProcessBuffer`, or something similar? It is just a little odd to me that it's in the hierarchy but doesn't start with `Subscription`, like the other classes after waitable.

>  If you'd like me to try to do that, I'll give that a whirl. I think we need the rest of the layers.

Maybe make an issue and possibly do it in another PR? This isn't that high priority for me.

audrow (approved):
Looks good to me - nice job with the tests and in refactoring!

rclcpp/include/rclcpp/any_subscription_callback.hpp
clalancette: > The only part I think that it's still valid from my last comment is that supporting that example seems unnecessary, but considering that we already do support it, I think we can move forward.

I talked to @audrow and @wjwwood a little bit about this, and part of the idea was so that you could potentially have multiple TypeAdapters for the same types together.  Passing the correct TypeAdapter type to the subscription would let the compiler know which of them to choose.  That said, I'll let @wjwwood comment here, as he has the best long-term vision for this.

Regardless, I think I'm going to move forward here and address the (minor) points raised by @audrow , and then run CI on the lot again.

rclcpp/include/rclcpp/experimental/subscription_intra_process_buffer.hpp
clalancette: > Do you think that it makes sense to rename ROSMessageIntraProcessBuffer to something like SubscriptionROSMessageIntraProcessBuffer, or something similar? It is just a little odd to me that it's in the hierarchy but doesn't start with Subscription, like the other classes after waitable.

That makes sense to me.  I'll go ahead and rename it (probably to `SubscriptionROSMsgIntraProcessBuffer`, just to shorten it a bit).

clalancette:
CI on the latest:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15901/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10576/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13565/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16219/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16220/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/172/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/10/)

clalancette:
All right, CI is in really good shape, and I have an approval.  I'm going to go ahead and merge this one, if there are any lingering things we want to fix to clean it up and/or reduce complexity, we can do it in a follow-up PR.

rclcpp/include/rclcpp/any_subscription_callback.hpp
ivanpauno: > Regardless, I think I'm going to move forward here and address the (minor) points raised by @audrow , and then run CI on the lot again.

Moving forward sounds good.

> Passing the correct TypeAdapter type to the subscription would let the compiler know which of them to choose

That's fine.
My comment is about the need to support `Subscription<TypeAdapterT>(..., CallbackUsingRosMessageT callback, ...)`, i.e. allowing to pass a callback that takes some smart pointer or referece to a ROS message type while asking for a `TypeAdapterT` subscription.
In practice, if the user replace `TypeAdapterT` with `RosMessageT` the implementation will do behind the scenes the exact same thing.
Allowing that adds a lot of code paths in the dispatching logic, but it doesn't seem to add any value.

Putting it another way, you always want the intraprocess buffer to match the type expected by the callback (as in that way it's possible to avoid the most copies).
I assumed that the type expected by the callback was always the `Subscription<...>` template parameter type, but that doesn't seem to actually be the case.
The intraprocess buffer type can currently be different to the callback type, as it's chosen basen on the subscription type, which is suboptimal.
As commented in the previous paragrah, there doesn't seem to be a real reason to allow mixing a `RosMessageT` callback type with a `TypeAdapterT` subscription.

```cpp
// this two subscriptions will dispatch intraprocess and interprocess messages in the exact same way
// their type is different for the cpp type system, but that doesn't seem to have any usefulness.
// my proposal is to not support creating sub2, only sub1
auto callback = [] (msg::MyMsg::SharedPtr msg) {...};
rclcpp::Subscription<msg::MyMsg> sub1 = node->create_subscription<msg::MyMsg>(..., callback);
rclcpp::Subscription<MyTypeAdaptedToMsg> sub2 = node->create_subscription<MyTypeAdaptedToMsg>(..., callback);
```

clalancette: I've opened up https://github.com/ros2/rclcpp/issues/1860 to track this (and one other issue) for the future.

Issue #1837, opened 2021-12-07
Support loaned message for serialized data


Barry-Xu-2018:
rosbag2 sends serialized data by rclcpp::GenericPublisher.   
Now there is no interface to use loaned message for serialized data (DDS not support it).   
In order to use current loaned message implementation, whether we can add new interface to rclcpp::GenericPublisher ?   

e.g.
```c++  
void GenericPublisher::try_publish_loaned_msg(const rclcpp::SerializedMessage & message)
{
    if (this->can_loan_messages()) {
      rcl_borrow_loaned_message(.., loaned_msg);
      rmw_deserialize(message.get_rcl_serialized_message(), .., loaned_msg);
      rcl_publish_loaned_message(..., loaned_msg, null);
    } else {
      publish(message); // Call original publish()
    }
}
```
Any thoughts?

fujitatomoya:
@ivanpauno @wjwwood @jacobperron 
CC: @emersonknapp 

this is `ros2 bag play` to enable `LoanedMessage` for zero copy enhancement. would you share your thoughts about this?

emersonknapp:
I agree with the approach you are taking. I think that this new interface would be very useful, and it's good to put it in `rclcpp` so that others could use it too. It will be important to document that deserialization happens within this function, so the user should be aware of that for performance.

fujitatomoya:
@emersonknapp thanks for the comment.

> It will be important to document that deserialization happens within this function, so the user should be aware of that for performance.

agree 👍

fujitatomoya:
@ivanpauno @wjwwood friendly ping, what would you think about this?

fujitatomoya:
@ivanpauno @wjwwood @jacobperron friendly ping.

jacobperron:
IMO, I would just throw an error if the publisher doesn't support loaned messages. Otherwise, it's not clear exactly what the call will do unless the user calls `can_loan_messages()`. I imagine usage would look like this:

```c++
if (publisher->can_loan_messages()) {
    publisher->publish_loaned_message(message);
} else {
    publisher->publish(message);
}
```

It just seems more clear from the outside what is happening. Than having the `can_loan_messages()` check inside the publish method.

We could consider naming the method such that it's clear that deserializing is happening. E.g. `publish_deserialized_and_loaned()`

wjwwood:
I agree with @jacobperron, have it throw and let callers do the try if they want.

PR #1856, opened 2022-01-06
Add publish by loaned message in GenericPublisher

Barry-Xu-2018:
The implementation for #1837

### DIFF
--- a/rclcpp/include/rclcpp/generic_publisher.hpp
+++ b/rclcpp/include/rclcpp/generic_publisher.hpp
@@ -36,21 +36,6 @@
 namespace rclcpp
 {
 
-class TypeSupportInterface
-{
-protected:
-  TypeSupportInterface(
-    std::shared_ptr<rcpputils::SharedLibrary> ts_lib,
-    const std::string & topic_type)
-  : ts_lib_(ts_lib),
-    type_support_(rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib)) {}
-
-  // The type support library should stay loaded, so it is stored in the GenericPublisher
-  std::shared_ptr<rcpputils::SharedLibrary> ts_lib_;
-
-  const rosidl_message_type_support_t * type_support_;
-};
-
 /// %Publisher for serialized messages whose type is not known at compile time.
 /**
  * Since the type is not known at compile time, this is not a template, and the dynamic library
@@ -58,7 +43,7 @@ protected:
  *
  * It does not support intra-process handling.
  */
-class GenericPublisher : public TypeSupportInterface, public rclcpp::PublisherBase
+class GenericPublisher : public rclcpp::PublisherBase
 {
 public:
   // cppcheck-suppress unknownMacro
@@ -89,12 +74,12 @@ public:
     const std::string & topic_type,
     const rclcpp::QoS & qos,
     const rclcpp::PublisherOptionsWithAllocator<AllocatorT> & options)
-  : TypeSupportInterface(ts_lib, topic_type),
-    PublisherBase(
+  : rclcpp::PublisherBase(
       node_base,
       topic_name,
-      *type_support_,
-      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos))
+      *rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib),
+      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos)),
+    ts_lib_(ts_lib)
   {
     // This is unfortunately duplicated with the code in publisher.hpp.
     // TODO(nnmm): Deduplicate by moving this into PublisherBase.
@@ -143,6 +128,9 @@ public:
   void publish_loaned_msg(const rclcpp::SerializedMessage & message);
 
 private:
+  // The type support library should stay loaded, so it is stored in the GenericPublisher
+  std::shared_ptr<rcpputils::SharedLibrary> ts_lib_;
+
   void * borrow_loaned_message();
   void deserialize_message(
     const rmw_serialized_message_t & serialized_message,
--- a/rclcpp/include/rclcpp/generic_publisher.hpp
+++ b/rclcpp/include/rclcpp/generic_publisher.hpp
@@ -31,8 +31,25 @@
 #include "rclcpp/typesupport_helpers.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+#include "rmw/rmw.h"
+
 namespace rclcpp
 {
+
+class TypeSupportInterface
+{
+protected:
+  TypeSupportInterface(
+    std::shared_ptr<rcpputils::SharedLibrary> ts_lib,
+    const std::string & topic_type)
+  : ts_lib_(ts_lib),
+    type_support_(rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib)) {}
+
+  // The type support library should stay loaded, so it is stored in the GenericPublisher
+  std::shared_ptr<rcpputils::SharedLibrary> ts_lib_;
+
+  const rosidl_message_type_support_t * type_support_;
+};
 
 /// %Publisher for serialized messages whose type is not known at compile time.
 /**
@@ -41,7 +58,7 @@ namespace rclcpp
  *
  * It does not support intra-process handling.
  */
-class GenericPublisher : public rclcpp::PublisherBase
+class GenericPublisher : public TypeSupportInterface, public rclcpp::PublisherBase
 {
 public:
   // cppcheck-suppress unknownMacro
@@ -72,12 +89,12 @@ public:
     const std::string & topic_type,
     const rclcpp::QoS & qos,
     const rclcpp::PublisherOptionsWithAllocator<AllocatorT> & options)
-  : rclcpp::PublisherBase(
+  : TypeSupportInterface(ts_lib, topic_type),
+    PublisherBase(
       node_base,
       topic_name,
-      *rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib),
-      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos)),
-    ts_lib_(ts_lib)
+      *type_support_,
+      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos))
   {
     // This is unfortunately duplicated with the code in publisher.hpp.
     // TODO(nnmm): Deduplicate by moving this into PublisherBase.
@@ -116,9 +133,15 @@ public:
   RCLCPP_PUBLIC
   void publish(const rclcpp::SerializedMessage & message);
 
+  /// Publish a rclcpp::SerializedMessage via loaned message.
+  void publish_loaned_msg(const rclcpp::SerializedMessage & message);
+
 private:
-  // The type support library should stay loaded, so it is stored in the GenericPublisher
-  std::shared_ptr<rcpputils::SharedLibrary> ts_lib_;
+  void * borrow_loaned_message();
+  void deserialize_message(
+    const rmw_serialized_message_t & serialized_message,
+    void * deserialized_msg);
+  void publish_loaned_message(void * loaned_message);
 };
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/generic_publisher.hpp
+++ b/rclcpp/include/rclcpp/generic_publisher.hpp
@@ -133,7 +133,12 @@ public:
   RCLCPP_PUBLIC
   void publish(const rclcpp::SerializedMessage & message);
 
-  /// Publish a rclcpp::SerializedMessage via loaned message.
+  /**
+   * Publish a rclcpp::SerializedMessage via loaned message after de-serialization.
+   *
+   * \param message a serialized message
+   * \throws anything rclcpp::exceptions::throw_from_rcl_error can show
+   */
   void publish_loaned_msg(const rclcpp::SerializedMessage & message);
 
 private:
--- a/rclcpp/include/rclcpp/generic_publisher.hpp
+++ b/rclcpp/include/rclcpp/generic_publisher.hpp
@@ -125,7 +125,7 @@ public:
    * \throws anything rclcpp::exceptions::throw_from_rcl_error can show
    */
   RCLCPP_PUBLIC
-  void publish_loaned_msg(const rclcpp::SerializedMessage & message);
+  void publish_as_loaned_msg(const rclcpp::SerializedMessage & message);
 
 private:
   // The type support library should stay loaded, so it is stored in the GenericPublisher
--- a/rclcpp/include/rclcpp/generic_publisher.hpp
+++ b/rclcpp/include/rclcpp/generic_publisher.hpp
@@ -139,6 +139,7 @@ public:
    * \param message a serialized message
    * \throws anything rclcpp::exceptions::throw_from_rcl_error can show
    */
+  RCLCPP_PUBLIC
   void publish_loaned_msg(const rclcpp::SerializedMessage & message);
 
 private:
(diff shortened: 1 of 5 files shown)


fujitatomoya:
CC: @iuhilnehc-ynos can you do review when you have time?

rclcpp/src/rclcpp/generic_publisher.cpp
```diff
@@ -31,4 +31,47 @@ void GenericPublisher::publish(const rclcpp::SerializedMessage & message)
...
 }
 
+void GenericPublisher::publish_loaned_msg(const rclcpp::SerializedMessage & message)
+{
+  if (this->can_loan_messages()) {
+    auto loaned_message = borrow_loaned_message();
+    deserialize_message(message.get_rcl_serialized_message(), loaned_message);
+    publish_loaned_message(loaned_message);
+  } else {
+    throw std::runtime_error("failed to loan messages");
```
fujitatomoya, line 41: more direct message would be suitable to tell use that loan message cannot be handled with current middleware?

```suggestion
    throw std::runtime_error("current middleware cannot handle loan messages");
```

rclcpp/src/rclcpp/generic_publisher.cpp
```diff
@@ -31,4 +31,47 @@ void GenericPublisher::publish(const rclcpp::SerializedMessage & message)
...
+  }
+}
+
+void * GenericPublisher::borrow_loaned_message()
+{
+  void * loaned_message = nullptr;
+  auto return_code = rcl_borrow_loaned_message(
```
fujitatomoya, line 48: `can_loan_messages` should be issued before here to check if the feature is enabled on this publisher? because in that case, `failed to borrow loaned msg` is not really reasonable for user application?

rclcpp/src/rclcpp/generic_publisher.cpp
```diff
@@ -31,4 +31,47 @@ void GenericPublisher::publish(const rclcpp::SerializedMessage & message)
...
+{
+  auto return_code = rmw_deserialize(&serialized_message, type_support_, deserialized_msg);
+  if (return_code != RMW_RET_OK) {
+    rclcpp::exceptions::throw_from_rcl_error(return_code, "failed to deserialize msg");
+  }
+}
+
+void GenericPublisher::publish_loaned_message(void * loaned_message)
+{
+  auto return_code = rcl_publish_loaned_message(
```
fujitatomoya, line 69: same, `can_loan_messages` should be issued before here to check if the feature is enabled on this publisher?

fujitatomoya (changes requested):

rclcpp/include/rclcpp/generic_publisher.hpp
```diff
@@ -117,9 +134,15 @@ class GenericPublisher : public rclcpp::PublisherBase
   RCLCPP_PUBLIC
   void publish(const rclcpp::SerializedMessage & message);
 
+  /// Publish a rclcpp::SerializedMessage via loaned message.
```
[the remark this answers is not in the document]
iuhilnehc-ynos: I think that exception information in the docstring is also important for users.

rclcpp/src/rclcpp/generic_publisher.cpp
```diff
@@ -31,4 +31,47 @@ void GenericPublisher::publish(const rclcpp::SerializedMessage & message)
   }
 }
 
+void GenericPublisher::publish_loaned_msg(const rclcpp::SerializedMessage & message)
+{
+  if (this->can_loan_messages()) {
```
iuhilnehc-ynos, line 36: I think that calling the `can_loan_messages` outside is the user's responsibility.
https://github.com/ros2/rclcpp/issues/1837#issuecomment-1006142083

iuhilnehc-ynos (changes requested):
some minor comments

rclcpp/include/rclcpp/generic_publisher.hpp
```diff
@@ -31,17 +31,34 @@
 #include "rclcpp/typesupport_helpers.hpp"
 #include "rclcpp/visibility_control.hpp"
 
+#include "rmw/rmw.h"
+
 namespace rclcpp
 {
 
+class TypeSupportInterface
```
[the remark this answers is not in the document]
Barry-Xu-2018: I checked current implementation of GenericSubscription.    
It defined the interface for loaned message.  
https://github.com/ros2/rclcpp/blob/6b321edb4fabcc57f666f77a4dc091e95948070b/rclcpp/include/rclcpp/generic_subscription.hpp#L146-L149   

`MessageInfo` is relevant to  `rmw_message_info_s`  
```c++
typedef struct RMW_PUBLIC_TYPE rmw_message_info_s
{
  rmw_time_point_value_t source_timestamp;
  rmw_time_point_value_t received_timestamp;
  rmw_gid_t publisher_gid;

  /// Whether this message is from intra_process communication or not
  bool from_intra_process;
} rmw_message_info_t;
```

It doesn't involve type info. So I don't move `TypeSupportInterface` to common header file at this time.

rclcpp/include/rclcpp/generic_publisher.hpp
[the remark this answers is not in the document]
Barry-Xu-2018: > I think that exception information in the docstring is also important for users.  

I only follow annotation style of `publish` function.   
I agree with your opinion. This information is useful and important for users.

rclcpp/src/rclcpp/generic_publisher.cpp
Barry-Xu-2018: Yes. I will check return code to show correct reason.

rclcpp/src/rclcpp/generic_publisher.cpp
Barry-Xu-2018: Yeah. Not call can_loan_messages() to check if this feature is enabled.   
Even if the user doesn't call can_loan_messages(), rcl function will return error code.

Barry-Xu-2018:
@fujitatomoya @iuhilnehc-ynos please review again

fujitatomoya (approved):

rclcpp/test/rclcpp/test_generic_pubsub.cpp
```diff
@@ -147,14 +171,59 @@ TEST_F(RclcppGenericNodeFixture, publisher_and_subscriber_work)
...
+
+    auto subscribed_messages = subscriber_future_.get();
+    EXPECT_THAT(subscribed_messages, SizeIs(Not(0)));
+    EXPECT_EQ(subscribed_messages[0], test_messages[0]);
+  } else {
+    ASSERT_THROW(
+    {
+      publisher->publish_loaned_msg(
+        serialize_message<int64_t, test_msgs::msg::BasicTypes>(test_messages[0]));
+    }, rclcpp::exceptions::RCLError);
```
iuhilnehc-ynos, line 223: Can it be tested? BTW, Do you think we need to add the failure case for `publish_loaned_msg` with `test_msgs::msg::Strings`

iuhilnehc-ynos (commented):
LGTM for the implementation, but some test cases might be changed.

iuhilnehc-ynos (approved):
LGTM

fujitatomoya:
@wjwwood @ivanpauno could you take a look at this?

rclcpp/test/rclcpp/test_generic_pubsub.cpp
Barry-Xu-2018: > Can it be tested?  
Yes. Currently we should do configuration to enable shared memory feature for cyclonedds and fastdds.
By default, test always execute this code.   
For the success case for calling publish_loaned_message, we have to do test in local environment (different dds need to different configuration).

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16168/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10796/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16509/)

fujitatomoya:
@Barry-Xu-2018 could you check the build failure?

Barry-Xu-2018:
> @Barry-Xu-2018 could you check the build failure?

I check failure log for Linux/Linux-aarch64.
Build is failed while building `quality_of_service_demo_cpp` package.

Barry-Xu-2018:
Do rebase

fujitatomoya:
let's give it a shot once.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16172/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10799/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16513/)

rclcpp/include/rclcpp/generic_publisher.hpp
```diff
@@ -31,17 +31,34 @@
...
+
 /// %Publisher for serialized messages whose type is not known at compile time.
 /**
  * Since the type is not known at compile time, this is not a template, and the dynamic library
  * containing type support information has to be identified and loaded based on the type name.
  *
  * It does not support intra-process handling.
  */
-class GenericPublisher : public rclcpp::PublisherBase
+class GenericPublisher : public TypeSupportInterface, public rclcpp::PublisherBase
```
wjwwood, line 61: Why make this a new class and inherit from it? It doesn't appear to be used anywhere else.

And even if it is, I think we should prefer composition over inheritance in this case, i.e. have `TypeSupportInterface` be a member of `GenericPublisher` rather than a parent class to it.

wjwwood:
Is there a corresponding pull request where this is being used (like in `rosbag2`)?

fujitatomoya:
@wjwwood thanks for checking on this.

originally we are motivated to use this for rosbag2, you can see https://github.com/ros2/rosbag2/issues/914.

rclcpp/include/rclcpp/generic_publisher.hpp
Barry-Xu-2018: @wjwwood  

Thanks for your comments.  

> Why make this a new class and inherit from it? It doesn't appear to be used anywhere else.

In original code, rosidl_message_type_support_t is only used  in `PublisherBase::PublisherBase` and not stored. But `GenericPublisher` need this for the new interface.  
I don't want to change `PublisherBase`. So I consider changing in GenericPublisher.  

> And even if it is, I think we should prefer composition over inheritance in this case, i.e. have TypeSupportInterface be a member of GenericPublisher rather than a parent class to it.  

For this case, composition cannot be used.   
We need to save `rosidl_message_type_support_t` before calling the Constructor of PublisherBase.   
What I want as below
```c++
GenericPublisher(
    std::shared_ptr<rcpputils::SharedLibrary> ts_lib,
    const std::string & topic_type,
    const rclcpp::QoS & qos,
    const rclcpp::PublisherOptionsWithAllocator<AllocatorT> & options)
  : type_support_handle_(rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib)),
    PublisherBase(
      node_base,
      topic_name,
      *type_support_handle_,
      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos))
```
Actually, passed `type_support_handle_` isn't initialized by `rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib)`.

Use simple code to explain
```c++
class Base {
public:
  Base(int base_data):
  base_data_(base_data)
  {
    std::cout << "Base: base_data_=" << base_data_ << std::endl;
  }
private:
  int base_data_;
};

class A : public Base {
public:
  A(int a, int b):
  a_data_(a+b),
  Base(a_data_) {
    std::cout << "A: a_data_= " << a_data_ << std::endl;
  }
private:
  int a_data_;
};

int main() {
  A a(10, 5);

  return 0;
}
```
We want to get "base_data_ == a_data_".  
In fact, output is
```
Base: base_data_=32764
A: a_data_= 15
```
So even if I use composition, I cannot pass correct `type support handle` to PublisherBase.  
I have to choose inheritance.   

I think there is another way.  
I can add a new protected variable in PublisherBase to store `rosidl_message_type_support_t`. But this will modify  PublisherBase slightly.   
What do you think ?

Barry-Xu-2018: Friendly ping @wjwwood

Barry-Xu-2018:
[f4af264](https://github.com/ros2/rclcpp/pull/1856/commits/f4af264f8441f26cf76be762c2cd09f33b95c692) is used to avoid 
 inheritance. But this will modify `PublisherBase`.    
 @wjwwood  please help to review this change.

Barry-Xu-2018:
Rebased

Barry-Xu-2018:
The failure of `Rpr__rclcpp__ubuntu_jammy_amd64 `  is as below.  rcl in build environment is old.  

```
17:25:19 [ 23%] Building CXX object CMakeFiles/rclcpp.dir/src/rclcpp/any_executable.cpp.o
17:25:19 In file included from /tmp/ws/src/rclcpp/rclcpp/include/rclcpp/callback_group.hpp:23,
17:25:19                  from /tmp/ws/src/rclcpp/rclcpp/include/rclcpp/any_executable.hpp:20,
17:25:19                  from /tmp/ws/src/rclcpp/rclcpp/src/rclcpp/any_executable.cpp:15:
17:25:19 /tmp/ws/src/rclcpp/rclcpp/include/rclcpp/client.hpp:33:10: fatal error: rcl/event_callback.h: No such file or directory
17:25:19    33 | #include "rcl/event_callback.h"
17:25:19       |          ^~~~~~~~~~~~~~~~~~~~~~
17:25:19 compilation terminated.
```

clalancette:
> The failure of `Rpr__rclcpp__ubuntu_jammy_amd64 ` is as below. rcl in build environment is old.

Due to general problems with the transition to Ubuntu Jammy, Rpr jobs don't work at the moment.  Running CI on https://ci.ros2.org is the recommended workflow until this gets sorted out in Jammy.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16326/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10946/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16698/)

rclcpp/include/rclcpp/generic_publisher.hpp
fujitatomoya: @wjwwood can you check on this?

fujitatomoya:
@alsora @SteveMacenski if you have time, could you take a look at this?

rclcpp/src/rclcpp/generic_publisher.cpp
```diff
@@ -31,4 +31,49 @@ void GenericPublisher::publish(const rclcpp::SerializedMessage & message)
   }
 }
 
+void GenericPublisher::publish_loaned_msg(const rclcpp::SerializedMessage & message)
```
SteveMacenski, line 34: Since this isn't actually publishing a loaned message, instead its publishing a given message **as** as loaned message internally, I think this API would need to be `publish_msg_as_loaned` or `publish_as_loaned_msg` to make the distinction that this function does not publish a pre-borrowed message its being given as input, but transfers a given message into the loaned set of memory locally.

Barry-Xu-2018: Yes. I will update code to use publish_as_loaned_msg.

Barry-Xu-2018:
@SteveMacenski please review again.  

BTW, the failed tests in Rpr__rclcpp__ubuntu_jammy_amd64 are  
[projectroot.test.rclcpp.test_timer](https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/59/testReport/junit/projectroot.test/rclcpp/test_timer/)
[rclcpp.TestTimer.test_is_canceled_reset](https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/59/testReport/junit/rclcpp/TestTimer/test_is_canceled_reset/)
They are unrelated to this change.

SteveMacenski (approved):
LGTM

Barry-Xu-2018:
@fujitatomoya  friendly ping.   
Please look at  SteveMacensk' comments.   
If you also think so, please help to run CI again.

PR #1857, opened 2022-01-07
Cleanup includes.

clalancette:
Remove ones that aren't needed, and add in ones that are
actually needed in the respective files.

### DIFF
--- a/rclcpp/include/rclcpp/experimental/buffers/intra_process_buffer.hpp
+++ b/rclcpp/include/rclcpp/experimental/buffers/intra_process_buffer.hpp
@@ -16,6 +16,7 @@
 #define RCLCPP__EXPERIMENTAL__BUFFERS__INTRA_PROCESS_BUFFER_HPP_
 
 #include <memory>
+#include <stdexcept>
 #include <type_traits>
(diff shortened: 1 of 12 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15909/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10584/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16232/)

clalancette:

sloretz (approved):
Was this PR made with a tool like `include-what-you-use`?

I didn't double check every single added and removed include, but based on the changes just being include directives and the fact that all packages above built, this change LGTM

wjwwood (approved):

clalancette:
> Was this PR made with a tool like `include-what-you-use`?

No, I didn't know that tool existed :).  I just did it by hand (I was looking in most of these files for other reasons anyway, and I noticed that the includes needed some love).

I'll take a look at `include-what-you-use`, maybe it would be something interesting to integrate into CI.

PR #1859, opened 2022-01-07
Fix include order and relative paths for cpplint

jacobperron:
Relates to https://github.com/ament/ament_lint/pull/324

### DIFF
--- a/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
@@ -72,8 +72,9 @@ MultiThreadedExecutor::get_number_of_threads()
 }
 
 void
-MultiThreadedExecutor::run(size_t)
+MultiThreadedExecutor::run(size_t this_thread_number)
 {
+  (void)this_thread_number;
   while (rclcpp::ok(this->context_) && spinning.load()) {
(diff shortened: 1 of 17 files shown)


rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -72,8 +72,9 @@ MultiThreadedExecutor::get_number_of_threads()
 }
 
 void
-MultiThreadedExecutor::run(size_t)
+MultiThreadedExecutor::run(size_t this_thread_number)
```
jacobperron, line 75: This was a false positive from cpplint. It was interpreting this as a C-style cast...

fujitatomoya: i see that you are just trying to fix the cpplint error. but do we actually need this argument?

clalancette: We don't, but I would argue that putting the argument here and then using `(void)this_thread_number;` (as below) is more idiomatic in this codebase.

clalancette (approved):
Looks good with green CI.

jacobperron:
I've updated includes to use double-quotes for backwards compatibility with older versions of cpplint (12feaaf).

PR #1862, opened 2022-01-11
Rename stringstream in macros to a more unique name

smith-doug:
Renames the std::stringstream in the logging stream macros from ss to something more unique.

### DIFF
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -141,8 +141,8 @@ def get_rclcpp_suffix_from_features(features):
     }; \
 @[ end if] \
 @[ if 'stream' in feature_combination]@
-    std::stringstream ss; \
-    ss << @(stream_arg); \
+    std::stringstream __rclcpp_stream_ss__; \
+    __rclcpp_stream_ss__ << @(stream_arg); \
 @[ end if]@
     RCUTILS_LOG_@(severity)@(get_suffix_from_features(feature_combination))_NAMED( \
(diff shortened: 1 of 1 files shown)


aprotyas (changes requested):
The C++ standard reserves identifiers that start with a double underscore, regardless of scope. As such, `__rclcpp_stream_ss__` is not a valid identifier for the `stringstream` here.

Also, what is the motivation behind this PR? Is the `ss` identifier causing problems anywhere?

smith-doug:
I tend to use a stringstream named ss for quick things.  I put ss.str() in the args for a RCLCPP_INFO_STREAM and had to scratch my head for a bit when it was outputting incorrectly.

clalancette (approved):
I was thinking that maybe we should add a test for this, but the only thing we could test for is that `ss` works now, and that is just arbitrary.  If someone happens to use `rclcpp_stream_ss_` in their code, then they will hit the same issue.  However, that is far less likely, so I think this fix is an improvement.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15924/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10599/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16248/)

aprotyas (approved):

clalancette:
Windows test failures are known, so this looks good.  Thanks for the PR!

Aposhian:
This issue was biting me on galactic, with a repro here: https://github.com/Aposhian/repro-rclcpp-error-stream. I was about to file an issue.

Issue #1846, opened 2021-12-14
Allow LifecycleNode to automatically transition Lifecycle-enabled ROS2 entities


bpwilcox:
This issue is a question/enhancement regarding the behavior of lifecycle nodes and their counterpart lifecycle entities (`LifecyclePublisher` and future lifecycle-enabled versions of subscriptions, services, action servers, etc.)

First, from the managed nodes design [doc](https://design.ros2.org/articles/node_lifecycle.html):

> Primary State: Active
> This is the main state of the node’s life cycle. While in this state, the node performs any processing, responds to service requests, reads and processes data, produces output, etc.

Similarly with other state guidelines in the documentation. the design outlines the expected behavior of a node in particular states. For the active state, all ROS communication is expected to be active and processing data. As it relates to ROS2 entities such as publishers, subscriptions, services, etc., this implies that their behavior would be in alignment with the lifecycle state of the node. To date, there only exists one lifecycle entity, the [LifecyclePublisher](https://github.com/ros2/rclcpp/blob/master/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp), however with the introduction of a `LifecycleService` in #1838 / #1836, and the potential for lifecycle versions of subscriptions, action servers and more, the code is beginning to align better with the design documentation.

As of today, a LifecyclePublisher must be manually managed to activate/deactivate its state in a node. With the addition of new lifecycle entities to match the intended design, this management will only grow in scale/complexity as users will have to micromanage all entities for lifecycle transitions. **My proposal in this issue** is that instead of requiring user management of lifecycle entity activation/deactivation, **allow the lifecycle node to automatically transition the state of its lifecycle entities (e.g. lifecycle publishers, subscriptions, services, etc.).** Ignoring implementation considerations for now, this feature would ensure that **lifecycle entity states (active/inactive) align with the lifecycle node**, which already seems to be the intended design according to the documentation. This distinctly distinguishes lifecycle entities from the standard ones, since developers who don't anticipate to align their entities with lifecycle states would likely be well served by standard ones instead.

### Benefits
- Reduce user micromanaging of lifecycle entities as components scale. 
- Default behavior to that of intended best practice as recommended by managed node design documentation
- Implicit insight into state of lifecycle entities via query of lifecycle node state (e.g. an inactive lifecycle node also implies that lifecycle entities are inactive)

[...]

PR #1863, opened 2022-01-13
Automatically transition lifecycle entities when node transitions

ivanpauno:
Fixes https://github.com/ros2/rclcpp/issues/1846.
Similar to what was done in rclpy: https://github.com/ros2/rclpy/pull/865.
Demo: https://github.com/ros2/demos/pull/552.

### DIFF
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
@@ -88,6 +88,13 @@ public:
     rclcpp::Publisher<MessageT, Alloc>::publish(msg);
   }
 
+  void
+  on_activate() override
+  {
+    SimpleManagedEntity::on_activate();
+    should_log_ = true;
+  }
+
 private:
   /// LifecyclePublisher log helper function
   /**
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
@@ -19,35 +19,20 @@
 #include <string>
 #include <utility>
 
+#include "rclcpp/logging.hpp"
 #include "rclcpp/publisher.hpp"
 
-#include "rclcpp/logging.hpp"
+#include "rclcpp_lifecycle/managed_entity.hpp"
+
 
 namespace rclcpp_lifecycle
 {
-/// base class with only
-/**
- * pure virtual functions. A managed
- * node can then deactivate or activate
- * the publishing.
- * It is more a convenient interface class
- * than a necessary base class.
- */
-class LifecyclePublisherInterface
-{
-public:
-  virtual ~LifecyclePublisherInterface() {}
-  virtual void on_activate() = 0;
-  virtual void on_deactivate() = 0;
-  virtual bool is_activated() = 0;
-};
-
 /// brief child class of rclcpp Publisher class.
 /**
  * Overrides all publisher functions to check for enabled/disabled state.
  */
 template<typename MessageT, typename Alloc = std::allocator<void>>
-class LifecyclePublisher : public LifecyclePublisherInterface,
+class LifecyclePublisher : public SimpleManagedEntity,
   public rclcpp::Publisher<MessageT, Alloc>
 {
 public:
@@ -64,7 +49,6 @@ public:
     const rclcpp::QoS & qos,
     const rclcpp::PublisherOptionsWithAllocator<Alloc> & options)
   : rclcpp::Publisher<MessageT, Alloc>(node_base, topic, qos, options),
-    enabled_(false),
     should_log_(true),
     logger_(rclcpp::get_logger("LifecyclePublisher"))
   {
@@ -81,7 +65,7 @@ public:
   virtual void
   publish(std::unique_ptr<MessageT, MessageDeleter> msg)
   {
-    if (!enabled_) {
+    if (!this->is_activated()) {
       log_publisher_not_enabled();
       return;
     }
@@ -97,30 +81,11 @@ public:
   virtual void
   publish(const MessageT & msg)
   {
-    if (!enabled_) {
+    if (!this->is_activated()) {
       log_publisher_not_enabled();
       return;
     }
     rclcpp::Publisher<MessageT, Alloc>::publish(msg);
-  }
-
-  virtual void
-  on_activate()
-  {
-    enabled_ = true;
-  }
-
-  virtual void
-  on_deactivate()
-  {
-    enabled_ = false;
-    should_log_ = true;
-  }
-
-  virtual bool
-  is_activated()
-  {
-    return enabled_;
   }
 
 private:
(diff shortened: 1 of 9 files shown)


rclcpp_lifecycle/include/rclcpp_lifecycle/managed_entity.hpp
```diff
@@ -0,0 +1,46 @@
...
+#ifndef RCLCPP_LIFECYCLE__MANAGED_ENTITY_HPP_
+#define RCLCPP_LIFECYCLE__MANAGED_ENTITY_HPP_
+
+#include <atomic>
+
+namespace rclcpp_lifecycle
+{
+
+///  Base class for lifecycle entities, like `LifecyclePublisher`.
+class ManagedEntityInterface
```
ivanpauno, line 24: The interface in rclpy is a bit different (see [here](https://github.com/ros2/rclpy/blob/38095981409f000c733b4c53037aa31ab0a76279/rclpy/rclpy/lifecycle/managed_entity.py#L23-L47)):

- on_activate()/on_deactivate() receives the previous state as a parameter and can fail.
  - I think in most of the cases on_activate()/on_deactivate() shouldn't fail, the on_shutdown()/on_configure() transitions should be doing the fallible work.
  - Receiving the previous state is not important for this transitions, as it's always 'inactive' for `on_activate()` and 'activated'  for `on_deactivate()`.
- The rclpy interface allows you to also provide `on_shutdown()`/`on_configure()`/`on_error()` transitions.
  - LifecyclePublisher is not using it, though entities could lazely create resources on `on_configure()` and destroy them on `on_shutdown()` if those methods were available.
  - The default on_configure/on_shutdown() implementation in Node might be less useful, as it's not clear what to do if some of the callbacks fail (see [this](https://github.com/ros2/rclpy/blob/38095981409f000c733b4c53037aa31ab0a76279/rclpy/rclpy/lifecycle/node.py#L188-L189) and [this](https://github.com/ros2/rclpy/blob/38095981409f000c733b4c53037aa31ab0a76279/rclpy/rclpy/lifecycle/node.py#L197-L198)).
    In practice, I think people usually want to manually write on_configure()/on_shutdown().

We could either modify this PR to do exactly what rclpy does or reduce rclpy flexibility to match more closely what this PR is adding.
I prefer a bit this less flexible version, which is what we actually seem to need.

rclcpp_lifecycle/include/rclcpp_lifecycle/managed_entity.hpp
```diff
@@ -0,0 +1,46 @@
...
+class SimpleManagedEntity : public ManagedEntityInterface
+{
+public:
+  ~SimpleManagedEntity() override = default;
+  void on_activate() override;
+  void on_deactivate() override;
+  bool is_activated() const;
+
+private:
+  std::atomic<bool> enabled_ = false;
```
jacobperron, line 42: nitpick: I would use consistent terminology
```suggestion
  std::atomic<bool> activated_ = false;
```

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
```diff
@@ -97,32 +81,13 @@ class LifecyclePublisher : public LifecyclePublisherInterface,
...
-  on_activate()
-  {
-    enabled_ = true;
-  }
-
-  virtual void
-  on_deactivate()
-  {
-    enabled_ = false;
-    should_log_ = true;
```
jacobperron, line 117: I think this breaks the previous logging behavior; we no longer set `should_log_` to true after a deactivation. TBH, I'm not sure if this logging behavior is necessary though; I guess it could be useful for debugging, but it also seems like it could be annoying in some scenarios :shrug:

rclcpp_lifecycle/include/rclcpp_lifecycle/managed_entity.hpp
jacobperron: > I think in most of the cases on_activate()/on_deactivate() shouldn't fail

Having less failure points sounds nice to me. I haven't thought through scenarios where someone may want to fail during activation/deactivation, but we could add it if the need arises. 

> The default on_configure/on_shutdown() implementation in Node might be less useful, as it's not clear what to do if some of the callbacks fail

If one or more of the callbacks fail, I would think we would want to transition all entities to the `unconfigued` state (as indicated by the design doc diagram). What do you think?

I agree that in most cases people will probably provide their own on_configure/on_shutdown logic. I think that it still might be nice if we call the transitions on all managed entities though. E.g. call any user-defined callback for the node and then, if that is successful, call the transition callbacks for managed entities. Or do you see something wrong with that behavior?

ivanpauno: > If one or more of the callbacks fail, I would think we would want to transition all entities to the unconfigued state (as indicated by the design doc diagram). What do you think?

I think we could do that.
The logic would be:
- configure
  - Try to transition all managed entities to configured
  - If transition fails (TRANSITION_FAILURE), shut down all the managed entities that already transitioned to go back to the unconfigured state.
  - If any callback errors (TRANSITION_ERROR), then we immediately return error without doing any more transitions.
    This applies when a on_configure callback errors, or when an on_shutdown callback errors after a previous failure.
- shutdown
  - Similar, but on failures we do nothing, and continue trying to move the rest of the entities to the shutdown state.

The only problem I see is that the logic is a bit complex and might not be actually useful in practice (we don't have any lifecycle entity that do work on configure/shutdown currently).

> I agree that in most cases people will probably provide their own on_configure/on_shutdown logic. I think that it still might be nice if we call the transitions on all managed entities though. E.g. call any user-defined callback for the node and then, if that is successful, call the transition callbacks for managed entities. Or do you see something wrong with that behavior?

I think this has been resolved in the discussion [here](https://github.com/ros2/demos/pull/552#discussion_r784217616).

jacobperron: Fair enough. I'm fine keeping the implementation simple for now; we can add complexity if the need arises.

rclcpp_lifecycle/include/rclcpp_lifecycle/managed_entity.hpp
ivanpauno: Done in https://github.com/ros2/rclcpp/pull/1863/commits/b2c73ec207aa417d8a029adc68dd3d0fe2cd5493

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_publisher.hpp
ivanpauno: Fixed in https://github.com/ros2/rclcpp/pull/1863/commits/469f5d74b09a6dd2e73c733d323252730d89ca4e

jacobperron (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15951/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10624/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16273/)

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16276/)

Issue #1861, opened 2022-01-11
Get exception "Resource deadlock avoided" while running TestTimeSource.clock_sleep_until_with_ros_time_basic


Barry-Xu-2018:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - source
- Version or commit hash:
  - 6b321edb4fabcc57f666f77a4dc091e95948070b
- DDS implementation:
  - cyclonedds
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

Find this issue in CI execution  https://build.ros2.org/job/Rpr__rclcpp__ubuntu_focal_amd64/1636/testReport/projectroot.test/rclcpp/test_time_source/

The below steps also can reproduce this problem  
```
$ source install/setup.bash
$ build/rclcpp/test/rclcpp/test_time_source --gtest_filter=*clock_sleep_until_with_ros_time_basic --gtest_repeat=10000

...
Repeating all tests (iteration 9304) . . .

Note: Google Test filter = *clock_sleep_until_with_ros_time_basic*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from TestTimeSource
[ RUN      ] TestTimeSource.clock_sleep_until_with_ros_time_basic
terminate called after throwing an instance of 'std::system_error'
  what():  Resource deadlock avoided
Aborted (core dumped)
```

#### Expected behavior

No exception

#### Actual behavior

Unexpected exception occurs

fujitatomoya:
Does this happen with rmw_fastrtps with mainline?

Barry-Xu-2018:
> Does this happen with rmw_fastrtps with mainline?

Also happen after checking.

Barry-Xu-2018:
Backstrace info based on coredump file   
[gdb_coredump.log](https://github.com/ros2/rclcpp/files/7844930/gdb_coredump.log)

Barry-Xu-2018:
After checking codes, I find the possible cause

In test codes
https://github.com/ros2/rclcpp/blob/80f93d1dbbe8398524ceec637f50ce2e3f1e9103/rclcpp/test/rclcpp/test_time_source.cpp#L751-L755

NodeState will create a thread (new executor) to  run a subscriber to receive clock message. 

[...]

PR #1865, opened 2022-01-14
Fix potential resource deadlock of NodeSate while destorying timeSource

Barry-Xu-2018:
Address issue #1861

### DIFF
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -591,6 +591,8 @@ bool TimeSource::clock_thread_is_joinable()
 
 TimeSource::~TimeSource()
 {
+  // Make sure clock subscriber thread is terminated at this time
+  node_state_->detachNode();
 }
 
 }  // namespace rclcpp


rclcpp/src/rclcpp/time_source.cpp
```diff
@@ -591,6 +591,8 @@ bool TimeSource::clock_thread_is_joinable()
 
 TimeSource::~TimeSource()
 {
+  // Make sure clock subscriber thread is terminated at this time
+  node_state_->detachNode();
```
ralph-lange, line 595:
```suggestion
  node_state_.reset();
```
For consistency with `TimeSource::detachNode()`, I propose to rely on NodeState's dtor, which will call `NodeState::detachNode()`.

clalancette: In my testing, using `node_state_->detachNode()` fixes the problem, while `node_state_.reset()` does not.  I think it has to do with the lifetime of objects, specifically when passing the pointers into lambda captures.

That said, you are right, this shouldn't be necessary at all.  In point of fact, looking at this code it is unnecessarily complicated.  We really don't need to use shared and weak pointers for most of these things; if we manage the ClocksState as part of the NodeState, then we can just use normal object lifetimes.  I made a set of changes that do this on this branch: https://github.com/clalancette/rclcpp/tree/clalancette/time-source-cleanups

It's a bigger change than this, but it cleans up the code pretty nicely and it solves the problem for me.  @Barry-Xu-2018 can you take a look and let me know what you think?

Barry-Xu-2018: @ralph-lange  

Thanks for your comments.   
I find a bug which is relevant to NodeState's dtor.   (Refer to https://github.com/ros2/rclcpp/issues/1861#issuecomment-1012963738). At this case, after TimeSource is destroyed (Normally, internal NodeState should be destroyed too), NodeState is alive.  So I add ` node_state_->detachNode()` to make sure clock subscriber thread must be terminated at this time.

Barry-Xu-2018: @clalancette 

> We really don't need to use shared and weak pointers for most of these things; if we manage the ClocksState as part of the NodeState, then we can just use normal object lifetimes. I made a set of changes that do this on this branch: https://github.com/clalancette/rclcpp/tree/clalancette/time-source-cleanups

Completely agree.  I also think my solution isn't good solution.   

> It's a bigger change than this, but it cleans up the code pretty nicely and it solves the problem for me.
> @Barry-Xu-2018 can you take a look and let me know what you think?

I quickly read your code.   
I have one comments in detachNode().
https://github.com/clalancette/rclcpp/blob/a68d7519147f9ae38416d8a61dffc8e39ead0c1a/rclcpp/src/rclcpp/time_source.cpp#L282-L294
I think `clocks_state_.disable_ros_time(); ` should be moved behind `destroy_clock_sub();`.   
It is because we not sure whether clock_cb() is executed at disabling ros time.  
```c++
  void clock_cb(std::shared_ptr<const rosgraph_msgs::msg::Clock> msg)
  {
    if (!clocks_state_.is_ros_time_active() && SET_TRUE == this->parameter_state_) {
      clocks_state_.enable_ros_time();
    }
    // Cache the last message in case a new clock is attached.
    clocks_state_.cache_last_msg(msg);
   ...
```

ralph-lange: @Barry-Xu-2018, @clalancette Thank you very much for the additional explanation.
@clalancette I've reviewed your bigger change and like it very much. The structure (object lifetimes, pointers) is much cleaner now. First, I had some doubts whether passing the this pointer of NodeState (instead of a shared_ptr) to the lambda expression for the subscription may cause new problems because of the risk that the `clock_executor_` still accesses the the object while it is already being cleaned up. But of course this cannot happen if `destroy_clock_sub()` is called as first step of clean up. I propose to add a prominent note in `detachNode` or `~NodeState` to ensure that this is not changed accidentally in the future.

clalancette:
Closing in favor of #1867.

PR #1871, opened 2022-01-20
add is_spinning() method to executor base class

alsora:
This PR addresses a problem I recently noticed when you start an executor in a separate thread.
For example:

```
auto executor = std::make_shared<SingleThreadedExecutor>();
std::thread executor_thread([&](){
    executor->spin();
});

//// When I reach this line I don't have any way to know whether the executor is already spinning or not.
executor->cancel(); // This call may be uneffective as it may get executed before the executor starts to spin
/// Similarly, if I have to do an operation that requires the executor to be spinning
```

There are hacky solutions to this, for example I added a node with a one shot timer with period 0 to the executor such that I can guarantee that after that timer triggers the executor is definitely spinning.
However this looks very bad and it could be done in a much nicer way thanks to the addition of this new API.

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -401,6 +401,15 @@ public:
   void
   set_memory_strategy(memory_strategy::MemoryStrategy::SharedPtr memory_strategy);
 
+  /// Returns true if the executor is currently spinning.
+  /**
+   * This function can be called asynchronously from any thread.
+   * \return True if the executor is currently spinning.
+   */
+  RCLCPP_PUBLIC
+  bool
+  is_spinning();
+
 protected:
   RCLCPP_PUBLIC
   void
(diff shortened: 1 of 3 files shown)


fujitatomoya (approved):
i think this is useful.

alsora:
CI run 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15983/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10653/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16305/)

SteveMacenski (approved):

SteveMacenski:
I will make use of this almost immediately in Nav2, I've had this concern myself.

alsora:
Merging because Windows test failures are unrelated (I probably should have limited the test scope to only rclcpp and above, the test failures are in rclpy and rosbag2_transport)

PR #1873, opened 2022-01-20
typo fix.

fujitatomoya:

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -913,7 +913,7 @@ Executor::get_next_executable(AnyExecutable & any_executable, std::chrono::nanos
   return success;
 }
 
-// Returns true iff the weak_groups_to_nodes map has node_ptr as the value in any of its entry.
+// Returns true if the weak_groups_to_nodes map has node_ptr as the value in any of its entry.
 bool
 Executor::has_node(
   const rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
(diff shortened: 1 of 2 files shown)


rclcpp/src/rclcpp/executor.cpp
```diff
@@ -913,7 +913,7 @@ Executor::get_next_executable(AnyExecutable & any_executable, std::chrono::nanos
   return success;
 }
 
-// Returns true iff the weak_groups_to_nodes map has node_ptr as the value in any of its entry.
+// Returns true if the weak_groups_to_nodes map has node_ptr as the value in any of its entry.
```
fujitatomoya, line 916: Do we need a `s` in `Returns` ? since i am not native speaker, not sure... 😅

alsora:
Is this really a typo?
`iff` could be a short hand for `if and only if`, however I don't think we need this kind of language in doxygen documentation.

fujitatomoya:
> iff could be a short hand for if and only if

thanks! that is something i did not know 😅 if this is not typo, we can just go ahead to close this one!

clalancette:
> > 
> 
> thanks! that is something i did not know sweat_smile if this is not typo, we can just go ahead to close this one!

I agree with Alberto in that this is likely "iff", meaning "if and only if".  That said, I think it would probably aid understanding (and avoid this ambiguity) by replacing it with "if and only if" where it is used.  @fujitatomoya If you go ahead and do that, I'm happy to approve.

fujitatomoya:
i do not think that is necessary to fix, thanks for checking!

Issue #1821, opened 2021-11-10
spin until timeout


briansoe66:
## Feature request

I propose a spin until timeout function: 

    void spin_until_timeout(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr, 
                            std::chrono::duration< TimeRepT, TimeT > timeout)

This is useful for writing unit tests. Currently, I use spin_until_future_complete(), with a placeholder future, based on code in benchmark_service.cpp

    constexpr char empty_service_name[] = "empty_service";
    auto client = node->create_client<test_msgs::srv::Empty>(empty_service_name);
    auto shared_request = std::make_shared<test_msgs::srv::Empty::Request>();
    auto future = client->async_send_request(shared_request);
    spin_until_future_complete(node, future, std::chrono::seconds(1));

fujitatomoya:
@hliberacki it's been a while, are you still working on this? I think the following PRs need to be rebased and verified once again before merge.

* https://github.com/ros2/rclpy/pull/919
* https://github.com/ros2/rclcpp/pull/1957
* https://github.com/ros2/system_tests/pull/505
* https://github.com/ros2/ros2cli/pull/704
* https://github.com/ros2/launch_ros/pull/306
* https://github.com/ros2/geometry2/pull/512
* https://github.com/ros2/examples/pull/339
* https://github.com/ros2/demos/pull/556
* https://github.com/ros-visualization/interactive_markers/pull/90
* https://gitlab.com/ros-tracing/ros2_tracing/-/merge_requests/291
* https://github.com/ros-visualization/rqt_reconfigure/pull/111

hliberacki:
@fujitatomoya Yes I'm, sorry I did not have enough capacity lately to do it. AFAIK what is pending was a review of my change in Rclpy repo. If I recall correctly @wjwwood found some issues there, yet those were not reviewed after my fixes.

Nevertheless I will rebase my other MR and see what has to be changed in current codebase. And when Rclpy will be correct, we can proceed to summarize this those changes all together.

Once again sorry for the delays on that

fujitatomoya:
repo file for these PRs included, https://gist.githubusercontent.com/fujitatomoya/9656dc65ca96a02690370b4695bd65c8/raw/c6a694c597cb7b1ce43960ba59ba939c4a35544c/ros2.repos

fujitatomoya:
@hliberacki Just FYI.

[...]

Issue #1712, opened 2021-07-12
[API Rename] spin_until_future_complete to spin_until_complete


SteveMacenski:
Hi,

I was hoping to get some feedback on the idea of changing `spin_until_future_complete` to `spin_until_complete`. When writing application code with actions / services, I run into extending lines often while spinning to see if the server was completed. This is it of itself isn't a big deal, but I wonder if we can't shorten it a bit. I'm not sure what the "future" part of the API provides. The input is a future, but we don't have `publish_message(msg)` we just have `publish(msg)`. 

So I propose changing `spin_until_future_complete(future)` to `spin_until_complete(future)` which I think semantically makes more sense and also opens the opportunity in future redesigns to use other mechanics other than `futures` to spin under the same API without adding a few different `spin_until_XYZ_complete()`. This certainly isn't an earth-shattering suggestion, but if you think this would be agreeable, I'd be happy to help implement.

wjwwood:
I don't feel strongly about it. The rationale for removing the `future` part makes sense to me, but I also don't feel like it is a huge drawback compared to iterating on the API. So if you want to propose a pull request that adds the new name and deprecates the old one (then ideally make sure we remove it next release), I think I'd be ok with that.

Maybe get feedback from @ivanpauno or @mabelzhang first before opening something.

fujitatomoya:
im good to go with this. less redundancy makes sense to me.

SteveMacenski:
OK, I'll wait for the feedback from William's other suggestions before getting into it.

I've done a quick overview of the main repos and see C++ instances of `spin_until_future_complete` in:
- rclcpp
- rclc

For python its
- rclpy  
- ros2cli

So overall, it actually seems like it may be pretty trivial to make that conversion. I assume you'd like me to do Python at the same time?

clalancette:
While I don't care hugely what the name is, I honestly don't think the churn here is worth it.  In order to get this in, we are going to have to:

1. Add a newly named API.
2. Deprecate the old API.
3. Fix all of the usage of the deprecated name in the core, and then eventually in all downstream packages.
4. Leave both the deprecated API and the new one in place for Humble.
5. Remove the deprecated API in I-Turtle.

It just doesn't seem to be worth it to me.  Maybe it will be worth it in the future if we add `spin_until_XYZ_complete`, but I wouldn't speculatively change it because of something that might or might not happen in the future.

But I'm not a maintainer here, this is just my 2 cents.

[...]

PR #1874, opened 2022-01-21
Add support for spin_until_timeout (#1821)

hliberacki:
Introduce executors new spin_for method, replace spin_until_future_complete with spin_until_complete. (#1821)
    
    * Introduce spin_for method.
    * Introduce spin_until_complete.
    * Deprecate spin_until_future_complete.
    * Replace usage of deprecated method.
    * Update unit-tests.
    
    Signed-off-by: Hubert Liberacki <hliberacki@gmail.com>

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -26,6 +26,7 @@
 #include <mutex>
 #include <string>
 #include <vector>
+#include <type_traits>
 
 #include "rcl/guard_condition.h"
 #include "rcl/wait.h"
@@ -319,6 +320,51 @@ public:
   virtual void
   spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1));
 
+  /// Spin (blocking) until the condition is complete, it times out waiting,
+  /// or rclcpp is interrupted.
+  /**
+   * \param[in] future The condition which can be callable or future type to wait on.
+   *   If this function returns SUCCESS, the future can be
+   *   accessed without blocking (though it may still throw an exception).
+   * \param[in] timeout Optional timeout parameter, which gets passed to Executor::spin_node_once.
+   *   `-1` is block forever, `0` is non-blocking.
+   *   If the time spent inside the blocking loop exceeds this timeout, return a TIMEOUT return
+   *   code.
+   * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
+   */
+  template<typename ConditionT, typename DurationT = std::chrono::milliseconds>
+  FutureReturnCode
+  spin_until_complete(
+    const ConditionT & condition,
+    DurationT timeout = DurationT(-1))
+  {
+    if constexpr (std::is_invocable_v<ConditionT>) {
+      using RetT = std::invoke_result_t<ConditionT>;
+      static_assert(
+        std::is_same_v<bool, RetT>,
+        "Conditional callable has to return boolean type");
+      return spin_until_complete_impl(condition, timeout);
+    } else {
+      auto check_future = [&condition]() {
+          return condition.wait_for(std::chrono::seconds(0)) ==
+                 std::future_status::ready;
+        };
+      return spin_until_complete_impl(check_future, timeout);
+    }
+  }
+
+  /// Spin (blocking) for at least the given amount of duration.
+  /**
+   * \param[in] duration gets passed to Executor::spin_node_once,
+   *   spins the executor for given duration.
+   */
+  template<typename DurationT>
+  void
+  spin_for(DurationT duration)
+  {
+    (void)spin_until_complete([]() {return false;}, duration);
+  }
+
   /// Spin (blocking) until the future is complete, it times out waiting, or rclcpp is interrupted.
   /**
    * \param[in] future The future to wait on. If this function returns SUCCESS, the future can be
@@ -330,57 +376,13 @@ public:
    * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
    */
   template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+  [[deprecated("use spin_until_complete(const ConditionT & condition, DurationT timeout) instead")]]
   FutureReturnCode
   spin_until_future_complete(
     const FutureT & future,
     std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
   {
-    // TODO(wjwwood): does not work recursively; can't call spin_node_until_future_complete
-    // inside a callback executed by an executor.
-
-    // Check the future before entering the while loop.
-    // If the future is already complete, don't try to spin.
-    std::future_status status = future.wait_for(std::chrono::seconds(0));
-    if (status == std::future_status::ready) {
-      return FutureReturnCode::SUCCESS;
-    }
-
-    auto end_time = std::chrono::steady_clock::now();
-    std::chrono::nanoseconds timeout_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
-      timeout);
-    if (timeout_ns > std::chrono::nanoseconds::zero()) {
-      end_time += timeout_ns;
-    }
-    std::chrono::nanoseconds timeout_left = timeout_ns;
-
-    if (spinning.exchange(true)) {
-      throw std::runtime_error("spin_until_future_complete() called while already spinning");
-    }
-    RCPPUTILS_SCOPE_EXIT(this->spinning.store(false); );
-    while (rclcpp::ok(this->context_) && spinning.load()) {
-      // Do one item of work.
-      spin_once_impl(timeout_left);
-
-      // Check if the future is set, return SUCCESS if it is.
-      status = future.wait_for(std::chrono::seconds(0));
-      if (status == std::future_status::ready) {
-        return FutureReturnCode::SUCCESS;
-      }
-      // If the original timeout is < 0, then this is blocking, never TIMEOUT.
-      if (timeout_ns < std::chrono::nanoseconds::zero()) {
-        continue;
-      }
-      // Otherwise check if we still have time to wait, return TIMEOUT if not.
-      auto now = std::chrono::steady_clock::now();
-      if (now >= end_time) {
-        return FutureReturnCode::TIMEOUT;
-      }
-      // Subtract the elapsed time from the original timeout.
-      timeout_left = std::chrono::duration_cast<std::chrono::nanoseconds>(end_time - now);
-    }
-
-    // The future did not complete before ok() returned false, return INTERRUPTED.
-    return FutureReturnCode::INTERRUPTED;
+    return spin_until_complete(future, timeout);
   }
 
   /// Cancel any running spin* function, causing it to return.
@@ -560,6 +562,55 @@ protected:
   virtual void
   spin_once_impl(std::chrono::nanoseconds timeout);
 
+protected:
+  // Implementation details, used by spin_until_complete and spin_for.
+  // Previouse implementation of spin_until_future_complete.
+  template<typename ConditionT, typename DurationT>
+  FutureReturnCode
+  spin_until_complete_impl(ConditionT condition, DurationT timeout)
+  {
+    auto end_time = std::chrono::steady_clock::now();
+    std::chrono::nanoseconds timeout_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+      timeout);
+    if (timeout_ns > std::chrono::nanoseconds::zero()) {
+      end_time += timeout_ns;
+    }
+    std::chrono::nanoseconds timeout_left = timeout_ns;
+
+    // Preliminary check, finish if conditon is done already.
+    if (condition()) {
+      return FutureReturnCode::SUCCESS;
+    }
+
+    if (spinning.exchange(true)) {
+      throw std::runtime_error("spin_until_complete() called while already spinning");
+    }
+    RCPPUTILS_SCOPE_EXIT(this->spinning.store(false); );
+    while (rclcpp::ok(this->context_) && spinning.load()) {
+      // Do one item of work.
+      spin_once_impl(timeout_left);
+
+      if (condition()) {
+        return FutureReturnCode::SUCCESS;
+      }
+      // If the original timeout is < 0, then this is blocking, never TIMEOUT.
+      if (timeout_ns < std::chrono::nanoseconds::zero()) {
+        continue;
+      }
+      // Otherwise check if we still have time to wait, return TIMEOUT if not.
+      auto now = std::chrono::steady_clock::now();
+      if (now >= end_time) {
+        return FutureReturnCode::TIMEOUT;
+      }
+      // Subtract the elapsed time from the original timeout.
+      timeout_left = std::chrono::duration_cast<std::chrono::nanoseconds>(end_time - now);
+    }
+
+    // The condition did not pass before ok() returned false, return INTERRUPTED.
+    return FutureReturnCode::INTERRUPTED;
+  }
+
+public:
   typedef std::map<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr,
       const rclcpp::GuardCondition *,
       std::owner_less<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr>>
--- a/rclcpp/include/rclcpp/executors.hpp
+++ b/rclcpp/include/rclcpp/executors.hpp
@@ -54,6 +54,50 @@ namespace executors
 using rclcpp::executors::MultiThreadedExecutor;
 using rclcpp::executors::SingleThreadedExecutor;
 
+/// Spin (blocking) until the conditon is complete, it times out waiting, or rclcpp is interrupted.
+/**
+ * \param[in] executor The executor which will spin the node.
+ * \param[in] node_ptr The node to spin.
+ * \param[in] condition The callable or future to wait on. If `SUCCESS`, the condition is safe to
+ *   access after this function
+ * \param[in] timeout Optional timeout parameter, which gets passed to
+ *   Executor::spin_node_once.
+ *   `-1` is block forever, `0` is non-blocking.
+ *   If the time spent inside the blocking loop exceeds this timeout, return a `TIMEOUT` return code.
+ * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
+ */
+template<typename ConditionT, typename DurationT = std::chrono::milliseconds>
+rclcpp::FutureReturnCode
+spin_node_until_complete(
+  rclcpp::Executor & executor,
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
+  const ConditionT & condition,
+  DurationT timeout = DurationT(-1))
+{
+  // TODO(wjwwood): does not work recursively; can't call spin_node_until_complete
+  // inside a callback executed by an executor.
+  executor.add_node(node_ptr);
+  auto retcode = executor.spin_until_complete(condition, timeout);
+  executor.remove_node(node_ptr);
+  return retcode;
+}
+
+template<typename NodeT = rclcpp::Node, typename ConditionT,
+  typename DurationT = std::chrono::milliseconds>
+rclcpp::FutureReturnCode
+spin_node_until_complete(
+  rclcpp::Executor & executor,
+  std::shared_ptr<NodeT> node_ptr,
+  const ConditionT & condition,
+  DurationT timeout = DurationT(-1))
+{
+  return rclcpp::executors::spin_node_until_complete(
+    executor,
+    node_ptr->get_node_base_interface(),
+    condition,
+    timeout);
+}
+
 /// Spin (blocking) until the future is complete, it times out waiting, or rclcpp is interrupted.
 /**
  * \param[in] executor The executor which will spin the node.
@@ -67,6 +111,10 @@ using rclcpp::executors::SingleThreadedExecutor;
  * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
  */
 template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+[[deprecated(
+  "use spin_node_until_complete(Executor &, node_interfaces::NodeBaseInterface::SharedPtr, "
+  "const ConditionT &, DurationT) instead"
+)]]
 rclcpp::FutureReturnCode
 spin_node_until_future_complete(
   rclcpp::Executor & executor,
@@ -74,16 +122,15 @@ spin_node_until_future_complete(
   const FutureT & future,
   std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
 {
-  // TODO(wjwwood): does not work recursively; can't call spin_node_until_future_complete
-  // inside a callback executed by an executor.
-  executor.add_node(node_ptr);
-  auto retcode = executor.spin_until_future_complete(future, timeout);
-  executor.remove_node(node_ptr);
-  return retcode;
+  return spin_until_complete(executor, node_ptr, future, timeout);
 }
 
 template<typename NodeT = rclcpp::Node, typename FutureT, typename TimeRepT = int64_t,
   typename TimeT = std::milli>
+[[deprecated(
+  "use spin_node_until_complete(Executor &, std::shared_ptr<NodeT>, "
+  "const ConditionT &, DurationT) instead"
+)]]
 rclcpp::FutureReturnCode
 spin_node_until_future_complete(
   rclcpp::Executor & executor,
@@ -91,7 +138,7 @@ spin_node_until_future_complete(
   const FutureT & future,
   std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
 {
-  return rclcpp::executors::spin_node_until_future_complete(
+  return rclcpp::executors::spin_node_until_complete(
     executor,
     node_ptr->get_node_base_interface(),
     future,
@@ -100,7 +147,33 @@ spin_node_until_future_complete(
 
 }  // namespace executors
 
+template<typename ConditionT, typename DurationT = std::chrono::milliseconds>
+rclcpp::FutureReturnCode
+spin_until_complete(
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
+  const ConditionT & condition,
+  DurationT timeout = DurationT(-1))
+{
+  rclcpp::executors::SingleThreadedExecutor executor;
+  return executors::spin_node_until_complete<ConditionT>(executor, node_ptr, condition, timeout);
+}
+
+template<typename NodeT = rclcpp::Node, typename ConditionT,
+  typename DurationT = std::chrono::milliseconds>
+rclcpp::FutureReturnCode
+spin_until_complete(
+  std::shared_ptr<NodeT> node_ptr,
+  const ConditionT & condition,
+  DurationT timeout = DurationT(-1))
+{
+  return rclcpp::spin_until_complete(node_ptr->get_node_base_interface(), condition, timeout);
+}
+
 template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+[[deprecated(
+  "use spin_until_complete(node_interfaces::NodeBaseInterface::SharedPtr, "
+  "const ConditionT &,DurationT) instead"
+)]]
 rclcpp::FutureReturnCode
 spin_until_future_complete(
   rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr,
@@ -108,18 +181,22 @@ spin_until_future_complete(
   std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
 {
   rclcpp::executors::SingleThreadedExecutor executor;
-  return executors::spin_node_until_future_complete<FutureT>(executor, node_ptr, future, timeout);
+  return executors::spin_node_until_complete<FutureT>(executor, node_ptr, future, timeout);
 }
 
 template<typename NodeT = rclcpp::Node, typename FutureT, typename TimeRepT = int64_t,
   typename TimeT = std::milli>
+[[deprecated(
+  "use spin_until_complete(std::shared_ptr<NodeT>, const ConditionT &, "
+  "DurationT) instead"
+)]]
 rclcpp::FutureReturnCode
 spin_until_future_complete(
   std::shared_ptr<NodeT> node_ptr,
   const FutureT & future,
   std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
 {
-  return rclcpp::spin_until_future_complete(node_ptr->get_node_base_interface(), future, timeout);
+  return rclcpp::spin_until_complete(node_ptr->get_node_base_interface(), future, timeout);
 }
 
 }  // namespace rclcpp
--- a/rclcpp/test/rclcpp/executors/test_executors.cpp
+++ b/rclcpp/test/rclcpp/executors/test_executors.cpp
@@ -222,7 +222,7 @@ TYPED_TEST(TestExecutors, spinWhileAlreadySpinning) {
 }
 
 // Check executor exits immediately if future is complete.
-TYPED_TEST(TestExecutors, testSpinUntilFutureComplete) {
+TYPED_TEST(TestExecutors, testSpinUntilCompleteFuture) {
   using ExecutorType = TypeParam;
   ExecutorType executor;
   executor.add_node(this->node);
@@ -232,11 +232,30 @@ TYPED_TEST(TestExecutors, testSpinUntilFutureComplete) {
   std::future<bool> future = promise.get_future();
   promise.set_value(true);
 
-  // spin_until_future_complete is expected to exit immediately, but would block up until its
+  // spin_until_complete is expected to exit immediately, but would block up until its
   // timeout if the future is not checked before spin_once_impl.
   auto start = std::chrono::steady_clock::now();
   auto shared_future = future.share();
-  auto ret = executor.spin_until_future_complete(shared_future, 1s);
+  auto ret = executor.spin_until_complete(shared_future, 1s);
+  executor.remove_node(this->node, true);
+  // Check it didn't reach timeout
+  EXPECT_GT(500ms, (std::chrono::steady_clock::now() - start));
+  EXPECT_EQ(rclcpp::FutureReturnCode::SUCCESS, ret);
+}
+
+// Check executor exits immediately if future is complete.
+TYPED_TEST(TestExecutors, testSpinUntilCompleteCallable) {
+  using ExecutorType = TypeParam;
+  ExecutorType executor;
+  executor.add_node(this->node);
+
+  // test success of an immediately completed condition
+  auto condition = []() {return true;};
+
+  // spin_until_complete is expected to exit immediately, but would block up until its
+  // timeout if the future is not checked before spin_once_impl.
+  auto start = std::chrono::steady_clock::now();
+  auto ret = executor.spin_until_complete(condition, 1s);
   executor.remove_node(this->node, true);
   // Check it didn't reach timeout
   EXPECT_GT(500ms, (std::chrono::steady_clock::now() - start));
@@ -244,7 +263,7 @@ TYPED_TEST(TestExecutors, testSpinUntilFutureComplete) {
 }
 
 // Same test, but uses a shared future.
-TYPED_TEST(TestExecutors, testSpinUntilSharedFutureComplete) {
+TYPED_TEST(TestExecutors, testSpinUntilCompleteSharedFuture) {
   using ExecutorType = TypeParam;
   ExecutorType executor;
   executor.add_node(this->node);
@@ -254,11 +273,11 @@ TYPED_TEST(TestExecutors, testSpinUntilSharedFutureComplete) {
   std::future<bool> future = promise.get_future();
   promise.set_value(true);
 
-  // spin_until_future_complete is expected to exit immediately, but would block up until its
+  // spin_until_complete is expected to exit immediately, but would block up until its
   // timeout if the future is not checked before spin_once_impl.
   auto shared_future = future.share();
   auto start = std::chrono::steady_clock::now();
-  auto ret = executor.spin_until_future_complete(shared_future, 1s);
+  auto ret = executor.spin_until_complete(shared_future, 1s);
   executor.remove_node(this->node, true);
 
   // Check it didn't reach timeout
@@ -267,7 +286,7 @@ TYPED_TEST(TestExecutors, testSpinUntilSharedFutureComplete) {
 }
 
 // For a longer running future that should require several iterations of spin_once
-TYPED_TEST(TestExecutors, testSpinUntilFutureCompleteNoTimeout) {
+TYPED_TEST(TestExecutors, testSpinUntilCompleteNoTimeout) {
   using ExecutorType = TypeParam;
   ExecutorType executor;
   executor.add_node(this->node);
@@ -286,7 +305,7 @@ TYPED_TEST(TestExecutors, testSpinUntilFutureCompleteNoTimeout) {
 
   // Timeout set to negative for no timeout.
   std::thread spinner([&]() {
-      auto ret = executor.spin_until_future_complete(future, -1s);
+      auto ret = executor.spin_until_complete(future, -1s);
       EXPECT_EQ(rclcpp::FutureReturnCode::SUCCESS, ret);
       executor.remove_node(this->node, true);
       executor.cancel();
@@ -312,15 +331,15 @@ TYPED_TEST(TestExecutors, testSpinUntilFutureCompleteNoTimeout) {
   spinner.join();
 }
 
-// Check spin_until_future_complete timeout works as expected
-TYPED_TEST(TestExecutors, testSpinUntilFutureCompleteWithTimeout) {
+// Check spin_until_complete timeout works as expected
+TYPED_TEST(TestExecutors, testSpinUntilCompleteWithTimeout) {
   using ExecutorType = TypeParam;
   ExecutorType executor;
   executor.add_node(this->node);
 
   bool spin_exited = false;
 
-  // Needs to run longer than spin_until_future_complete's timeout.
+  // Needs to run longer than spin_until_complete's timeout.
   std::future<void> future = std::async(
     std::launch::async,
     [&spin_exited]() {
@@ -332,7 +351,7 @@ TYPED_TEST(TestExecutors, testSpinUntilFutureCompleteWithTimeout) {
 
   // Short timeout
   std::thread spinner([&]() {
-      auto ret = executor.spin_until_future_complete(future, 1ms);
+      auto ret = executor.spin_until_complete(future, 1ms);
       EXPECT_EQ(rclcpp::FutureReturnCode::TIMEOUT, ret);
       executor.remove_node(this->node, true);
       spin_exited = true;
@@ -482,8 +501,8 @@ TYPED_TEST(TestExecutors, spinSome) {
   spinner.join();
 }
 
-// Check spin_node_until_future_complete with node base pointer
-TYPED_TEST(TestExecutors, testSpinNodeUntilFutureCompleteNodeBasePtr) {
+// Check spin_node_until_complete with node base pointer
+TYPED_TEST(TestExecutors, testSpinNodeUntilCompleteNodeBasePtr) {
   using ExecutorType = TypeParam;
   ExecutorType executor;
 
@@ -492,13 +511,13 @@ TYPED_TEST(TestExecutors, testSpinNodeUntilFutureCompleteNodeBasePtr) {
   promise.set_value(true);
 
   auto shared_future = future.share();
-  auto ret = rclcpp::executors::spin_node_until_future_complete(
+  auto ret = rclcpp::executors::spin_node_until_complete(
     executor, this->node->get_node_base_interface(), shared_future, 1s);
   EXPECT_EQ(rclcpp::FutureReturnCode::SUCCESS, ret);
 }
 
-// Check spin_node_until_future_complete with node pointer
-TYPED_TEST(TestExecutors, testSpinNodeUntilFutureCompleteNodePtr) {
+// Check spin_node_until_complete with node pointer
+TYPED_TEST(TestExecutors, testSpinNodeUntilCompleteNodePtr) {
   using ExecutorType = TypeParam;
   ExecutorType executor;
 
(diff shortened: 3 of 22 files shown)


rclcpp/include/rclcpp/executor.hpp
```diff
@@ -381,6 +383,47 @@ class Executor
...
+      timeout);
+    if (timeout_ns <= std::chrono::nanoseconds::zero()) {
+      // No work to be done, timeout is reached.
+      return;
+    }
+
+    end_time += timeout_ns;
+    std::chrono::nanoseconds timeout_left = timeout_ns;
+
+    RCPPUTILS_SCOPE_EXIT(this->spinning.store(false); );
```
alsora, line 411: this must be kept as close as possible to `if (spinning.exchange(true)) {`.
in the current code, if `timeout_ns <= std::chrono::nanoseconds::zero()` is verified, the node will return without setting `spinning` back to false

alsora:
I can see value in this new utility.
However, the implementation is almost a copy-paste of `spin_until_future_complete`.

What if we make this new function more generic, by taking as an extra argument a function that signals that we need to exit early.
Then, `spin_until_future_complete` would just invoke your new function passing as "early_exit" argument something like this
```
      // Check if the future is set, return SUCCESS if it is.
      if (status == std::future_status::ready) {
        return FutureReturnCode::SUCCESS;
      }
```

hliberacki:
Thanks for the feedback.

You are right, it is basically the simplified version of `spin_until_future_complete`.

I did it as it is, only because it's my first change and I did not want to break anything - but my initial idea was to make one generic method.

I'll make it more generic.

hliberacki:
@alsora I was also thinking about API stability, because this could be easily changed to `spin_until_complete` and then define the completion predicate as an lambda or simply template parameter, which is already there. Yet again I wasn't sure how much API can change within versions.

alsora:
I think we wouldn't break any API as we would keep `spin_until_future_complete` with the existing signature and just change its internal implementation to invoke this more generic `spin_until_complete`.

SteveMacenski:
Interestingly, I suggested a similar API change awhile back https://github.com/ros2/rclcpp/issues/1712. The reason this hasn't started yet is due to the lack of desire for churn in the API without a cause. Once we change something more significant or support non-futures, I would jump on that opportunity to make the change in name to coincide with a necessary change.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -381,6 +383,47 @@ class Executor
     return FutureReturnCode::INTERRUPTED;
   }
 
+  /// Spin (blocking) until it times out, or rclcpp is interrupted.
+  /**
+   * \param[in] timeout Timeout parameter, which gets passed to Executor::spin_node_once.
+   * \throws std::runtime_error if spinning is already taking place.
+   */
+  template<typename TimeRepT = int64_t, typename TimeT = std::milli>
+  void spin_until_timeout(
```
SteveMacenski, line 392: May I suggest rather than adding a new API, just adding a timeout field to `spin()`? Spinning without the default field will be indefinite, but if  given the timeout field, would use it and return. 

Semantically, I think that would be cleaner to use as an end-user & involve less changes to the API at the same time. 

`spin_until_XYZ_complete` I think makes sense when you're waiting for something ill-defined timeframe to complete, but when you have a specific timeout, that feels more like an argument to the main `spin` than a separate method.

hliberacki: Yeah you are right, it does sound more consistant.

rclcpp/test/rclcpp/executors/test_executors.cpp
```diff
@@ -221,6 +221,56 @@ TYPED_TEST(TestExecutors, spinWhileAlreadySpinning) {
...
+  using ExecutorType = TypeParam;
+  ExecutorType executor;
+  executor.add_node(this->node);
+
+  // block intil timeout is fulfilled.
+  auto start = std::chrono::steady_clock::now();
+  executor.spin_until_timeout(2s);
+  executor.remove_node(this->node, true);
+  // Check if timeout was fulfilled
+  EXPECT_GT(2s, (std::chrono::steady_clock::now() - start));
```
fujitatomoya, line 235: how about checking if it returns in specific time window?

e.g) `EXPECT_LT(3s, (std::chrono::steady_clock::now() - start));`

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -344,16 +344,18 @@ class Executor
     }
 
     auto end_time = std::chrono::steady_clock::now();
+
+    if (spinning.exchange(true)) {
+      throw std::runtime_error("spin_until_future_complete() called while already spinning");
+    }
+
```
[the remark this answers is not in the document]
hliberacki: Yes, but I wasn't aware of the fact that you and @alsora mentioned. I will restore the old version.

rclcpp/include/rclcpp/executor.hpp
ivanpauno: I think `spin_for` is a better name for this, it matches better the standard `sleep_for()`.

ivanpauno (commented):
I'm not sure if it's a good idea to continue adding more flavors of `spin_*()`, but the implementation lgtm!

alsora:
Hi @hliberacki 
it looks like there are still some open comments in this PR.

In particular:
 - correctly handle setting and un-setting `is_spinning_` 
 - whether to merge this new function into an existing one or not

hliberacki:
> Hi @hliberacki it looks like there are still some open comments in this PR.
> 
> In particular:
> 
>     * correctly handle setting and un-setting `is_spinning_`
> 
>     * whether to merge this new function into an existing one or not

Hi @alsora,

Yeah sorry for that it's pending for that long - I had some fire in my day-to-day project. 
I'm getting back to this PR, to make it done.

At the moment I would need to know which way would be the best fit, your first comment https://github.com/ros2/rclcpp/pull/1874#issuecomment-1018515145 - which is significantly easier or @SteveMacenski comment https://github.com/ros2/rclcpp/pull/1874#discussion_r789979122 to extend `spin()` with timeout argument.

Even though I find @SteveMacenski solution more robust - it will be a lot of changes because I need to change most of `spin` calls in the project (both in Python and C++) - which is a lot ;). Making it a default argument, since doesn't drastically change the amount of work. 

Some part of the work is already done, but still, I would need to fix tests - parts of python codebase and probably nodes which I did not take look into.

I could also take your suggestion and we can create a second ticket - to handle altering `spin` API (adding timeout).

How do you feel about it?

And again sorry for being idle that long - not gonna happen again :)

alsora:
No worries!
Just keep in mind that if we want this feature to be available in ROS 2 H-release, it will need to be merged before the 4th of April.

About whether to integrate this into `spin` or into `spin_until_..` I can see pro and cons to both approaches.
I would personally prefer to leave `spin` untouched.

The `spin()` function is the default one, that we recommend most people to use (e.g. create a main that loads a node and make it spin forever).
For this reason, I would like for this function to be as fast and as simple as possible and to encourage the good pattern of setup a ROS node, make it spin and forget about it.

Then, next to this, I can see a variety of other "spin functionalities" that can be provided by one or more functions.
All these alternatives would follow the paradigm of "spin until X happens", where X would be:
 - a timeout
 - no more work available
 - a future is ready
 - etc

hliberacki:
> No worries! Just keep in mind that if we want this feature to be available in ROS 2 H-release, it will need to be merged before the 4th of April.
> 
> About whether to integrate this into `spin` or into `spin_until_..` I can see pro and cons to both approaches. I would personally prefer to leave `spin` untouched.
> 
> The `spin()` function is the default one, that we recommend most people to use (e.g. create a main that loads a node and make it spin forever). For this reason, I would like for this function to be as fast and as simple as possible and to encourage the good pattern of setup a ROS node, make it spin and forget about it.
> 
> Then, next to this, I can see a variety of other "spin functionalities" that can be provided by one or more functions. All these alternatives would follow the paradigm of "spin until X happens", where X would be:
> 
>     * a timeout
> 
>     * no more work available
> 
>     * a future is ready
> 
>     * etc
> 
> 
> In theory all these "spin until X" could be done using a single, configurable function.

Ok, it seems a better fit for the moment. I will take care of that this week

SteveMacenski:
> Even though I find @SteveMacenski solution more robust - it will be a lot of changes because I need to change most of spin calls in the project (both in Python and C++) - which is a lot ;)

Think about it this way, every time anyone uses the usual `spin()` API you can think to yourself that you made an impact there! I would think of changing a bunch of stuff as a big positive. Plus you brought up a good point and its a compliment that I want to interact with it every day :wink:

SteveMacenski:
Definitely!

ivanpauno:
> Do spin(timeout)

Would this be a new overload of the current `spin()`?
That sounds fine to me, though I like a bit more having a differently named method.

> Create a new spin_until_complete()

I would rather not see the word "until" in the name for this overload.
I think it's common in cpp that "until" methods receive a timepoint, and "for" methods receive a duration/period.
So in this case I would expect the name to be `spin_for()` (if I'm understanding the implementation correctly).
e.g.: std::this_thread::sleep_for() vs std::this_thread::sleep_until()
e.g.: [asio::io_context::run_until()](https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/reference/io_context/run_until.html) vs [asio::io_context::run_for()](https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/reference/io_context/run_for.html)

ivanpauno (commented):
LGTM with the minor naming issue resolved

SteveMacenski:
I don't have a strong preference for `spin_for` vs `spin_until`, though I have a slight preference for `until` but I think that's mostly just because that's the phrasing we have now. That's not a good technical reason so I'll yield to `for` if others agree

hliberacki:
> > Even though I find @SteveMacenski solution more robust - it will be a lot of changes because I need to change most of spin calls in the project (both in Python and C++) - which is a lot ;)
> 
> Think about it this way, every time anyone uses the usual `spin()` API you can think to yourself that you made an impact there! I would think of changing a bunch of stuff as a big positive. Plus you brought up a good point and its a compliment that I want to interact with it every day 😉
> 
> I know its unfair that different maintainers are telling you different things -- just to see if my summarization of this conversation is correct to see if we can derive some sense of consensus
> 
> * I suggested `spin(timeout)` and got 👍 from @fujitatomoya and @hliberacki since we didn't like the `spin_until_timeout` API.
> * @ivanpauno didn't like to keep adding more of the `spin_...` apis
> * @alsora doesn't like changing `spin()` since its the most used and perfers to create a single named function `spin_XYZ` that can be overloaded for use with futures, timeouts, etc as needed
> 
> It sounds to me like the 2 options to make the most people happy
> 
> * Do `spin(timeout)`
> * Create a new `spin_until_complete()` API that we change all existing `spin_until_future_complete`s to & add the timeout option. Which is interestingly the same idea I brought up about a year ago in [[API Rename] spin_until_future_complete to spin_until_complete #1712](https://github.com/ros2/rclcpp/issues/1712)
> 
> I actually prefer the second idea, which would make me, Alberto, and Ivan happy

For starters thanks for the feedback @SteveMacenski . I guess following your previous idea - https://github.com/ros2/rclcpp/issues/1712 would be a good tradeoff between overloading `spin` and creating `spin_until_timeout` - which I originally did.

> I know its unfair that different maintainers are telling you different things -- just to see if my summarization of this >conversation is correct to see if we can derive some sense of consensus

That's not a problem at all, I am just still a newbie Ros wise so it's hard for me to pick the best approach. 

I will then implement as it's described in #1712. As per naming, to me for and until sounds equaly good.

hliberacki:
Since the most concern was about the end API - I thought that it might be beneficial that I just push how I've updated the public API executors - from one of my first commits. I've done far more than this single change- but this is still a draft that I will push as a complete result. I'd like to only know whether you feel ok with API approach (`executor.hpp`). Then I would not need to update all the unit tests if the API is not fine. I am about to push more implementation this weekend. @SteveMacenski it this what you meant while you were describing #1712?

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -560,6 +608,46 @@ class Executor
   virtual void
   spin_once_impl(std::chrono::nanoseconds timeout);
 
+private:
```
hliberacki, line 611: Not sure whether this shall be in the private section - will change it if it's done differently.

hliberacki:
> @hliberacki is this PR still draft? you are still working on this as PoC?

I'm still working on that. i have changed status from open to draft to mark that the patch that i currently committed is just a question about the the API ans not an actual final solution - to avoid confusion.

i have done far more and intensified my work since yesterday - so this is definitely on going (at the moment I'm extending unit tests for executors). but since I'm doing it which take a bit of time I'd like to confirm how the revievers feel about the API itself so i would not polish thing (document, unit test) that has to be changed. thats why I've pushed it in that state, hopefully it makes sense :)

also I know that there is a release in the begging of April therefore i wanted to have some very basic feedback before the weekend, because i would like to finalize it during the weekend or be close to do it.

SteveMacenski:
> @SteveMacenski it this what you meant while you were describing https://github.com/ros2/rclcpp/issues/1712?

It would have to instead replace the API as well for all instances of `spin_until_future_complete` so that that API does not exist anymore (in reality, adds a deprecation warning but calls the new API in its wake). That way there are only 3 APIs: `spin` `spin_once` and `spin_until_complete`

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -319,6 +320,53 @@ class Executor
   virtual void
   spin_once(std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1));
 
+  // TODO(hliberacki): Add documentation
+  template<typename FutureT, typename DurationT = std::chrono::milliseconds,
+    // typename std::enable_if_t<
```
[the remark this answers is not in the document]
SteveMacenski: I believe Humble targets C++17, so I see no reason why not to use any C++17 features you like, but I am definitely not the right person to ask. Someone at OR would have to jump in to let us know if we should be trying to minimize its use to help in C++14 compatibility for other important users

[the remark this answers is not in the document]
ivanpauno: > typename std::enable_if_t<!std::is_invocable_v<FutureT>, bool> = true

It'll be hard to understand in the future why `std::is_invocable_v` was used.
I would rather have the more accurate condition (even if it doesn't look pretty) or two overloads.

rclcpp/include/rclcpp/executor.hpp
SteveMacenski: I prefer `protected` so that it can be messed with by a derived class

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -319,6 +320,53 @@ class Executor
...
+      "Conditional callable has to return boolean type");
+    return spin_until_complete_impl(condition, timeout);
+  }
+  // TODO(hliberacki): Add documentation
+  /// spin (blocking) for given amount of time
+  template<typename DurationT>
+  void
+  spin_for(DurationT timeout)
+  {
+    return spin_until_complete([]() {return false;}, timeout);
```
ivanpauno, line 367: I guess the compiler can be clever and optimize the implementation well enough, but maybe having a direct implementation is better.

ivanpauno (approved):

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
hliberacki: @ivanpauno At the moment i've decided to use `if constexpr` and reduce second check whether the `Condition` is a of type `Future` or `SharedFuture`. I did it because realize that there are more use cases, like using `src/ros2/rclcpp/rclcpp/include/rclcpp/client.hpp` from`src/ros2/rclcpp/rclcpp/include/rclcpp/client.hpp`. 

If you still would like to  have more grain granularity - then I would suggest to create Concept like requirement of the API for the given condition. So it would have to have `wait_for(timeout)` method which returns `enum class future_status`. I did not introduce it for now, on purpose because I am not sure whether it will be valuable to have that type_trait which is a bit hard to read. But If needed I can introduce it - no problem with that.

rclcpp/include/rclcpp/executor.hpp
hliberacki: @ivanpauno Sorry maybe it's the hour (past midnight in my time-zone ;) ) But I don't get that comment. I'd like to reimplement the logic here as well? I thought it might be a waste of time since it's exactly the same logic the difference is only calling this lambda which always returns true - so as you've said probably most compilers will just create a constant value there. 

But sure I can just provide direct implementation here as well if needed. Should not take more than a few minutes.

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -294,6 +294,41 @@ TEST_F(TestExecutor, spin_some_elapsed) {
...
+  bool timer_called = false;
+  auto timer =
+    node->create_wall_timer(
+    std::chrono::milliseconds(10), [&]() {
+      timer_called = true;
+    });
+  dummy.add_node(node);
+  dummy.spin_for(std::chrono::milliseconds(5));
+
+  ASSERT_FALSE(timer_called);
```
alsora, line 329: this test seems very prone to sporadic failures.
I would really like that unit-test didn't depend on wall time, but currently they do.
this means that if it takes more than 5ms for the executor to start spinning (due to a CPU spike in the CI server) this test will fail.

we should at least use larger intervals: for example a timer with 1s period.

alsora:
@hliberacki I did a quick look at the PR and it seems great!
Starting a CI job

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16464/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11054/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16816/)

hliberacki:
> @hliberacki I did a quick look at the PR and it seems great! Starting a CI job
> 
>     * Linux [Build Status](http://ci.ros2.org/job/ci_linux/16464/)
> 
>     * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11054/)
> 
>     * Windows [Build Status](http://ci.ros2.org/job/ci_windows/16816/)

Thanks :)

I still need to update all usages of deprecated API. I will push PR to those nodes. Not sure whether GitHub has depended on pull requests available (I Guess not?) then we would not be able to have the runners complete - since there is circular dependency ;).

I could move the deprecated attribute into separate PR, and update all usages. When It's done then we could accept the second PR - so that there would be no failures. 

Currently, I'm also updating `rclpy` and trying to fix one test which failed after my changes - yet I need to investigate why because locally I have zero failures, which is odd, and also I haven't changed anything in timers implementation.

alsora:
Yes, I see that the job I started before failed due to some other repos using the now deprecated APIs.
The ROS 2 CI allows to specify a full list of branches/repositories to build, so you can prepare all the necessary PRs (e.g. rclcpp, rclpy, etc).

We can test all the PRs together with the CI and then manually merge all the PRs at the same time

hliberacki:
I wasn't aware of that. I'll try to create that package today (Most are done, yet I will be available around late evening) so tomorrow it would be possible to test it as a whole.

hliberacki:
@alsora It took me a bit, but I think all repositories which use `spin_until_future_complete`  are updated. I've added a comment to each PR, so they are linked with this one.

Current checks fail for the timer test (single test) - but I can't reproduce it on my local setup. Could you rerun the CI with all dependent repositories, so we can see whether it might magically fix itself ;)?

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16484/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11073/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16854/)

I hope to have correctly selected all repos =)

rclcpp/test/rclcpp/test_executor.cpp
fujitatomoya: agree, this could be flaky.

fujitatomoya (approved):
lgtm with minor comments.

rclcpp/test/rclcpp/test_executor.cpp
hliberacki: Done - I guess for the negative test where `spinning` takes less time than the `timer` - it should not be flaky (broken) anymore.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -560,6 +561,48 @@ class Executor
   virtual void
   spin_once_impl(std::chrono::nanoseconds timeout);
 
+protected:
+  // Implementation details, used by spin_until_complete and spin_for.
+  // Previouse implementation of spin_until_future_complete.
+  template<typename Condition, typename DurationT>
+  FutureReturnCode spin_until_complete_impl(Condition condition, DurationT timeout)
+  {
+    auto end_time = std::chrono::steady_clock::now();
```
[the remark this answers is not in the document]
hliberacki: Done - I've added that after the start timestamp is calculated - since the execution of callable can take more than the single fetching status of the future. I did not want to spoil the timeout here.

hliberacki:
@alsora @fujitatomoya Everything shall be fixed right now (hope so :))

hliberacki:
@alsora I think the last CI did not include `rqt_reconfigure` with my changes. I can see it used dashing branch ->
```
19:06:27   ros-visualization/rqt_reconfigure:
19:06:27     type: git
19:06:27     url: https://github.com/ros-visualization/rqt_reconfigure.git
19:06:27     version: dashing
```

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16510/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11097/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16882/)

hliberacki:
@alsora sorry for the number of patches/changes - I've just updated the last thing which made the build `unstable` in `rclpy` ->
```
Double quotes found but single quotes preferred:
            raise TypeError("Condition has to be of Future or Callable type")
```            
It was minor but all of the things shall be gone now

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16511/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11098/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16883/)

hliberacki:
@alsora @fujitatomoya I can see that the Windows machine is not stable. From what I can see tests that are failing are not directly connected with this PR - but with `rosbag`. I did a quick research and it seems that those tests might be flaky. I am just wondering whether this PR introduced them - if you think that they did or you have no knowledge about it I will do my best to fix them.

Here is why I think they might be flaky ->
* https://github.com/ros2/rosbag2/issues/933
* https://github.com/ros2/rosbag2/issues/501
* https://githubhot.com/repo/ros2/rosbag2/issues/732

Between CIs [16882](http://ci.ros2.org/job/ci_windows/16882/) and [16883](http://ci.ros2.org/job/ci_windows/16883/), there was only a single change introduced - double quotes replaced with single quotes in a single python file, yet fewer tests failed. 

Moreover, I can see that some of those tests are sporadically failing from build to build. 

What do you think about it?

alsora (approved):

alsora:
Yes, I agree.
PR is good to go for me.

@wjwwood @ivanpauno @fujitatomoya it looks like all associated PRs were just waiting for a green CI.
Can someone of you take care of merging all of them? I only have merge permissions in rclcpp.

hliberacki:
@clalancette I am not going to opt for or against :) since I am a newbie. Regarding transitioning a downstream application after this change - I took navigation2 https://github.com/ros-planning/navigation2

> a) It still compiles (with deprecation warnings)
I did not build it (yet) but compiling the repo will end up in a warning for each usage of the old API. So here it depends on the project setup whether `warnings as errors` are turned on. If they are, there will be a compilation error.

> b) How difficult it is to fix the package up
I've "greped" the whole repository - from what I can see I was able to find 118 usages of `spin_until_future_complete` in the whole repository. To fix those we would just need to change those calls (names only) `spin_until_future_complete` to `spin_until_complete`. Api is the same, we can either pass future as before or a Callable so there is nothing more.

Single pros of this change would be, that you can run `spin_for` instead of
```
std::promise<void> p;
std::future<void> f;
auto result = spin_until_future_complete(f, timeout); // Future will never be done, it's a place holder
```

clalancette:
> I've "greped" the whole repository

Just to be clear; what exactly did you look through?  If you looked through https://github.com/ros2/ros2/blob/master/ros2.repos , you've only looked at the core and not the entire ecosystem.

hliberacki:
@clalancette by the above i only meant a single repository as a example of making it correct after this PR. I've checked Nav2, https://github.com/ros-planning/navigation2 (as you suggested) and those numbers comes only from that. but I'm aware that this is just a tip of iceberg.

SteveMacenski:
I would opt again since its LTS and its so close to the deadline. But I’m fine with either if someone feels a rush

hliberacki:
@clalancette @alsora Are we ready to merge the changes? If so I would rebase it, and resolve conflicts.

clalancette:
> @clalancette @alsora Are we ready to merge the changes? If so I would rebase it, and resolve conflicts.

Yep, the Rolling branches are open for business again, so please go ahead and rebase.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -319,6 +320,50 @@ class Executor
...
+   * \param[in] future The condition which can be callable or future type to wait on.
+   *   If this function returns SUCCESS, the future can be
+   *   accessed without blocking (though it may still throw an exception).
+   * \param[in] timeout Optional timeout parameter, which gets passed to Executor::spin_node_once.
+   *   `-1` is block forever, `0` is non-blocking.
+   *   If the time spent inside the blocking loop exceeds this timeout, return a TIMEOUT return
+   *   code.
+   * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
+   */
+  template<typename Condition, typename DurationT = std::chrono::milliseconds>
```
wjwwood, line 335: nitpick: use `ConditionT`

We've been somewhat inconsistent on this, but the trailing `T` is preferable as it makes it easier to search through the project, especially if in the future we have an actual class called `Condition` (which is very possible).

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -319,6 +320,50 @@ class Executor
...
+    DurationT timeout = DurationT(-1))
+  {
+    if constexpr (std::is_invocable_v<Condition>) {
+      using RetT = std::invoke_result_t<Condition>;
+      static_assert(
+        std::is_same_v<bool, RetT>,
+        "Conditional callable has to return boolean type");
+      return spin_until_complete_impl(condition, timeout);
+    } else {
+      auto checkFuture = [&condition]() {
```
wjwwood, line 347:
```suggestion
      auto check_future = [&condition]() {
```

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -319,6 +320,50 @@ class Executor
...
+          return condition.wait_for(std::chrono::seconds(0)) ==
+                 std::future_status::ready;
+        };
+      return spin_until_complete_impl(checkFuture, timeout);
+    }
+  }
+
+  /// spin (blocking) for given amount of duration.
```
wjwwood, line 355:
```suggestion
  /// Spin (blocking) for given amount of duration.
```

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -319,6 +320,50 @@ class Executor
...
+   *   If this function returns SUCCESS, the future can be
+   *   accessed without blocking (though it may still throw an exception).
+   * \param[in] timeout Optional timeout parameter, which gets passed to Executor::spin_node_once.
+   *   `-1` is block forever, `0` is non-blocking.
+   *   If the time spent inside the blocking loop exceeds this timeout, return a TIMEOUT return
+   *   code.
+   * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
+   */
+  template<typename Condition, typename DurationT = std::chrono::milliseconds>
+  FutureReturnCode spin_until_complete(
```
wjwwood, line 336:
```suggestion
  FutureReturnCode
  spin_until_complete(
```

Following the style of the rest of the methods in the class.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -319,6 +320,50 @@ class Executor
...
+  /**
+   * \param[in] duration gets passed to Executor::spin_node_once,
+   *   spins the executor for given duration.
+   */
+  template<typename DurationT>
+  void
+  spin_for(DurationT duration)
+  {
+    (void)spin_until_complete([]() {return false;}, duration);
+  }
```
wjwwood, line 365: I'm concerned that the semantics of this method are misleading slightly, and that we need to at least document it more clearly or even consider a name change.

It's inaccurate to say that we will spin for the given duration because we cannot guarantee how long we will spin (i.e. block) for. We can only control how long we wait for work to be available.

Consider this timeline:

- enter `spin_for(1s)`
- wait 0.5 seconds for something to be ready
- execute the callback for that thing for 1 second
- return from `spin_for(1s)` after roughly 1.5 seconds

In my opinion, we should at the very least clarify this in the documentation, and more Ideally pick a name and terminology that reflects what's really happening.

For example, but not necessarily the best answer, you could say `max_wait_time` instead of `duration` or `spin_for_at_least()` instead of `spin_for()`.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -330,57 +375,13 @@ class Executor
    * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
    */
   template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+  [[deprecated("use spin_until_complete(const Condition & condition, DurationT timeout) instead")]]
```
wjwwood, line 378: If you take the `ConditionT` recommendation, don't forget to update this too.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -560,6 +561,54 @@ class Executor
   virtual void
   spin_once_impl(std::chrono::nanoseconds timeout);
 
+protected:
+  // Implementation details, used by spin_until_complete and spin_for.
+  // Previouse implementation of spin_until_future_complete.
+  template<typename Condition, typename DurationT>
+  FutureReturnCode spin_until_complete_impl(Condition condition, DurationT timeout)
```
wjwwood, line 568:
```suggestion
  FutureReturnCode
  spin_until_complete_impl(Condition condition, DurationT timeout)
```

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -560,6 +561,54 @@ class Executor
   virtual void
   spin_once_impl(std::chrono::nanoseconds timeout);
 
+protected:
+  // Implementation details, used by spin_until_complete and spin_for.
+  // Previouse implementation of spin_until_future_complete.
+  template<typename Condition, typename DurationT>
```
wjwwood, line 567: Here too with the `ConditionT` recommendation.

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -67,31 +111,31 @@ using rclcpp::executors::SingleThreadedExecutor;
  * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
  */
 template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+[[deprecated("use spin_node_until_complete(Executor &,"
+                                           "node_interfaces::NodeBaseInterface::SharedPtr,"
+                                           "const Condition &, DurationT) instead")]]
```
wjwwood, line 116: nitpick: Typically we do not match indent like this, we indent everything or nothing, here's my suggestion:

```suggestion
[[deprecated(
  "use spin_node_until_complete(Executor &, node_interfaces::NodeBaseInterface::SharedPtr, "
  "const Condition &, DurationT) instead"
)]]
```

Any easy rule of thumb is "only indent a multiple of 2 (or 4 for Python), and never more than a single indent if you can help it".

This indents 43 spaces a weird, odd-numbered number of spaces, and wastes a lot of space for no reason.

Also consider if you changed the name of the function, you'd have to adjust the indentation of all of the lines to match, making future diffs more difficult.

Also, your line breaks lost a separating single whitespace between arguments that should be there.

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -67,31 +111,31 @@ using rclcpp::executors::SingleThreadedExecutor;
...
-  auto retcode = executor.spin_until_future_complete(future, timeout);
-  executor.remove_node(node_ptr);
-  return retcode;
+  return spin_until_complete(executor, node_ptr, future, timeout);
 }
 
 template<typename NodeT = rclcpp::Node, typename FutureT, typename TimeRepT = int64_t,
   typename TimeT = std::milli>
+[[deprecated("use spin_node_until_complete(Executor &, std::shared_ptr<NodeT>,"
+                                           "const Condition &, DurationT) instead")]]
```
wjwwood, line 130:
```suggestion
[[deprecated(
  "use spin_node_until_complete(Executor &, std::shared_ptr<NodeT>, "
  "const Condition &, DurationT) instead"
)]]
```

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -100,26 +144,52 @@ spin_node_until_future_complete(
...
+  std::shared_ptr<NodeT> node_ptr,
+  const Condition & condition,
+  DurationT timeout = DurationT(-1))
+{
+  return rclcpp::spin_until_complete(node_ptr->get_node_base_interface(), condition, timeout);
+}
+
 template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+[[deprecated("use spin_until_complete(node_interfaces::NodeBaseInterface::SharedPtr,"
+                                      "const Condition &,DurationT) instead")]]
```
wjwwood, line 171:
```suggestion
[[deprecated(
  "use spin_until_complete(node_interfaces::NodeBaseInterface::SharedPtr, "
  "const Condition &,DurationT) instead"
)]]
```

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -100,26 +144,52 @@ spin_node_until_future_complete(
...
 {
   rclcpp::executors::SingleThreadedExecutor executor;
-  return executors::spin_node_until_future_complete<FutureT>(executor, node_ptr, future, timeout);
+  return executors::spin_node_until_complete<FutureT>(executor, node_ptr, future, timeout);
 }
 
 template<typename NodeT = rclcpp::Node, typename FutureT, typename TimeRepT = int64_t,
   typename TimeT = std::milli>
+[[deprecated("use spin_until_complete(std::shared_ptr<NodeT>, const Condition &,"
+                                      "DurationT) instead")]]
```
wjwwood, line 185:
```suggestion
[[deprecated(
  "use spin_until_complete(std::shared_ptr<NodeT>, const Condition &, "
  "DurationT) instead"
)]]
```

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -294,6 +294,41 @@ TEST_F(TestExecutor, spin_some_elapsed) {
...
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+      timer_called = true;
+    });
+  dummy.add_node(node);
+  // Wait for the wall timer to have expired.
+  std::this_thread::sleep_for(std::chrono::milliseconds(50));
+  dummy.spin_for(std::chrono::seconds(1));
+
+  ASSERT_TRUE(timer_called);
+}
```
wjwwood, line 314: This kind of timing test is actually very fragile since we don't use real-time operating systems on our buildfarm machines.

I would not be surprised to see this become a flakey test at some point.

---

To help avoid that, consider:

- using `0` for the timer's duration (rather than 10ms)
- do not sleep in the callback for the timer (not sure what the purpose of that is)
- do not sleep to wait for the timer to expire, check `timer->time_until_trigger() <= 0` in a busy-loop (can sleep inside the busy loop)

---

Taking a step back, what is being tested here? That `spin_for` calls callbacks that were ready when it is entered? If so, maybe use a duration of `0` for `sleep_for()`? As this should work even for non-blocking spins.

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -294,6 +294,41 @@ TEST_F(TestExecutor, spin_some_elapsed) {
...
+  auto timer =
+    node->create_wall_timer(
+    std::chrono::seconds(1), [&]() {
+      timer_called = true;
+    });
+  dummy.add_node(node);
+  dummy.spin_for(std::chrono::milliseconds(5));
+
+  ASSERT_FALSE(timer_called);
+}
```
wjwwood, line 330: Again, I think this is at risk of being flakey. This one is harder because there's no getting around they timing nature of this test, and I do think it has some value, but what I would say is the timing tolerance is not important, so crank it up, for example use a timer period of like 10s to help avoid lag spikes from causing this to fail.

wjwwood: I now see that was discussed before. I think we should crank it up even more than 1s, but generally I agree with the sporadic failure part.

rclcpp/test/rclcpp/test_timer.cpp
```diff
@@ -111,13 +111,13 @@ TEST_F(TestTimer, test_is_canceled_reset)
 
   // reset shouldn't affect state (not canceled yet)
   timer->reset();
-  EXPECT_LE(timer->time_until_trigger().count(), std::chrono::nanoseconds::max().count());
+  EXPECT_LE(timer->time_until_trigger().count(), std::chrono::milliseconds::max().count());
```
wjwwood, line 114: Why this change?

wjwwood (changes requested):

wjwwood:
Sorry I didn't have time to review this until now, but I had some feedback, some of which I think should be addressed before merging.

rclcpp/include/rclcpp/executor.hpp
wjwwood: This is probably the most significant concern I have.

wjwwood: It might be sufficient to use phrasing like the `std::this_thread::sleep_for()` function uses:

> Blocks the execution of the current thread for at least the specified sleep_duration.
>
> This function may block for longer than sleep_duration due to scheduling or resource contention delays.

https://en.cppreference.com/w/cpp/thread/sleep_for

hliberacki: Sorry for the late reply.

I had the same concern while producing this code. Making it more apparent in a description section would be a must - I've kind of considered that as a user's responsibility to be aware of it, but now I think it might be wrong to do so. 

I was also considering a slight improvement, where we could execute the inner lambda within a separate thread.
But I was thinking about it for some time and I'm not sure whether it's the correct approach - I'm not sure if this would not bring additional overhead. 

I thought that could have an additional parameter or second variant where we directly specify whether the call spawns thread inside or does it in the same thread. Then we could state 2 things, the first approach would not exceed the timeout but lambda could be terminated before it finishes and there will be an additional thread, if you'd rather not then call is without a thread but then you might exceed the timeout. 

I think you've suggested something similar in #1916.

what do you think about it?

wjwwood: > I was also considering a slight improvement, where we could execute the inner lambda within a separate thread.
But I was thinking about it for some time and I'm not sure whether it's the correct approach - I'm not sure if this would not bring additional overhead.

👎 The callback should be executed in the thread that called spin, if it's a single threaded executor.

> lambda could be terminated before it finishes

I do not believe there is a way to do that.

> I thought that could have an additional parameter or second variant where we directly specify whether the call spawns thread inside or does it in the same thread. Then we could state 2 things, the first approach would not exceed the timeout but lambda could be terminated before it finishes and there will be an additional thread, if you'd rather not then call is without a thread but then you might exceed the timeout.

I think that's unnecessarily complicated for this function. I'd prefer to let the user do that or maybe have a separate function/class to help them do that.

I think it's just important to ensure the user understands the blocking behavior and that the documentation and/or the name of the function and parameters reflect that.

hliberacki:
sorry for being idle, taking care of closing (fixing code review) this issue(s) right now.

fujitatomoya:
@hliberacki there are also merge conflicts need to be addressed. once it is ready, i will start the CI.

rclcpp/test/rclcpp/test_timer.cpp
hliberacki: That was a mistake on my side - I don't recall why this was even needed. Restored

hliberacki:
> @hliberacki there are also merge conflicts need to be addressed. once it is ready, i will start the CI.

@fujitatomoya I am also updating all the related issues and checking whether any new PR is needed. I will let you know when the whole pipeline can be run. This should not take to long (hope so)

hliberacki:
@fujitatomoya looks like all is updated/rebased and fixed. From what I can see this PR is no longer needed https://github.com/ros-visualization/interactive_markers/pull/90. I will close it later on

alsora:
```
* Linux [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=16792)](http://ci.ros2.org/job/ci_linux/16792/)
* Linux-aarch64 [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=11373)](http://ci.ros2.org/job/ci_linux-aarch64/11373/)
* Windows [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=17202)](http://ci.ros2.org/job/ci_windows/17202/)
```

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16820/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11398/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17226/)

hliberacki:
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/16820/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11398/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/17226/)

Not sure why linux build was not triggered, but as for now Linux-aarch64 is passing. @wjwwood could you please take a look whether the condition of this PR is acceptable for you?

alsora:
Usually it means that the jenkins machines are busy.. we can try wait a couple hours to see if it starts otherwise I can restart that single job.
Anyhow the fact that linux-aarch64 is green is a great sign =)
We should be good to go!

hliberacki:
> Usually it means that the jenkins machines are busy.. we can try wait a couple hours to see if it starts otherwise I can restart that single job. Anyhow the fact that linux-aarch64 is green is a great sign =) We should be good to go!

I can see that there were 4 Windows warnings - but they are present for the last few builds - and not related to my changes, shall we ignore them in the scope of this PR?

alsora:
Yes, the PR is good to go for me.
@wjwwood do you consider your requested changes to be addressed?
If so, would you mind merging the whole thing?

Blast545:
:man_farmer: This seems to have introduced several build regressions on the linux buildfarm, it's failing on `system_tests`, can you take a look? @hliberacki

See:
https://build.ros2.org/view/Rci/job/Rci__nightly-release_ubuntu_jammy_amd64/118/console
https://ci.ros2.org/view/nightly/job/nightly_linux_debug/2338/console

I'm opening a revert PR to be sure this is causing the problem and merge before EOD if we don't find another way to address it. @clalancette

clalancette:
We definitely missed merging at least https://github.com/ros2/system_tests/pull/499 , there may be others.  I'll go see if I can track them down.

hliberacki:
@clalancette most of that (all?) has already been reviewed, so it shall be rather straight forward.

hliberacki:
> So the complete list that needs to be merged alongside this is:
> 
> * https://github.com/ros2/rclpy/pull/919
> * https://github.com/ros2/system_tests/pull/499
> * https://github.com/ros2/ros2cli/pull/704
> * https://github.com/ros2/launch_ros/pull/306
> * https://github.com/ros2/geometry2/pull/512
> * https://github.com/ros2/examples/pull/339
> * https://github.com/ros2/demos/pull/556
> * https://github.com/ros-visualization/interactive_markers/pull/90
> * https://gitlab.com/ros-tracing/ros2_tracing/-/merge_requests/291
> * https://github.com/ros-visualization/rqt_reconfigure/pull/111
> 
> I'm not at all sure all of this churn was worth a pretty subjective name change, but here we are.
> 
> So we can either revert this (which is the easiest thing to do), or we can merge forward all of the rest of the bits.  The problem with merging forward is that not all of the PRs are approved, so someone is going to have to go through, make sure that they are reviewed and approved, and then finish merging them up.
> 
> @fujitatomoya @wjwwood @SteveMacenski @alsora Can you go through and review all of the dependent PRs here, and then, assuming they are good, merge the rest?  If we can't get it done by the end of the day, then @Blast545 let's revert.

unfortunately replying on my phone, so tis hard to check the list myself but
all of the related PRs are linked with this PR - they are listed in the comment section.

wjwwood:
Ah this is my fault, I should have merged the corresponding prs, I was on my phone and didn't see the cross-references. In retrospect it's obvious that this was required.

christophebedard:
> * https://gitlab.com/ros-tracing/ros2_tracing/-/merge_requests/291

Should I proceed with merging this, or wait?

wjwwood:
So, @clalancette and I discussed it and I'm going to revert this.

The `rclpy` pr is not ready and it blocks the merge of the `system_tests` pr because that pr changes both C++ and Python uses of this function, but since this pr was merged prematurely by me, `system_tests` is emitting deprecation warnings (currently escalated to errors, but we're changing that in https://github.com/ros2/system_tests/pull/503).

I could have split the `system_tests` pr to be C++ and Python separately, but I don't have time to do that right this second, and so I'm going to revert this for now until we can do this properly and all at once.

hliberacki:
> So, @clalancette and I discussed it and I'm going to revert this.
> 
> The `rclpy` pr is not ready and it blocks the merge of the `system_tests` pr because that pr changes both C++ and Python uses of this function, but since this pr was merged prematurely by me, `system_tests` is emitting deprecation warnings (currently escalated to errors, but we're changing that in https://github.com/ros2/system_tests/pull/503).
> 
> I could have split the `system_tests` pr to be C++ and Python separately, but I don't have time to do that right this second, and so I'm going to revert this for now until we can do this properly and all at once.

@wjwwood thanks for taking care of that. i will try to fix your findings in rclpy as soon as possible.

SteveMacenski:
> Can you go through and review all of the dependent PRs here

Can do

SteveMacenski:
Something that also should probably be updated is documentation on docs.ros.org. Migration guide mention for Iron + update the tutorials. I took a look and its a pretty trivial number of changes. I'll do that now

SteveMacenski:
https://github.com/ros2/ros2_documentation/pull/2798 Docs PR

Ryanf55:
Hey there, it seems like this didn't get in to `humble`. Looking at the examples in `humble`, it's not clear to me how to handle timeouts in services in `humble`. Would anyone be able to assist with a demo on how to not block an application if a service server fails to return? I can open up a separate issue if needed.

fujitatomoya:
@clalancette @wjwwood @alsora any objection for taking these PRs? i think this is useful. you can see the entire PR list here, https://github.com/ros2/rclcpp/pull/1874#issuecomment-1165688522

alsora:
Yes, I agree that this is a very useful feature to have.

PR #1875, opened 2022-01-21
Deal with rosidl_runtime_cpp::Vector in C++ messages.

hidmic:
Precisely what the title says. Connected to https://github.com/ros2/rosidl/pull/647. 

Interoperability with `std::vector<bool>` is kept in all cases but when an explicit `std::vector<bool>` reference is requested e.g. `const std::vector<bool> & param = rclcpp::ParameterValue{...}.get<std::vector<bool>>();` will no longer compile.

### DIFF
--- a/rclcpp/include/rclcpp/parameter.hpp
+++ b/rclcpp/include/rclcpp/parameter.hpp
@@ -164,12 +164,12 @@ public:
   const std::vector<uint8_t> &
   as_byte_array() const;
 
-  /// Get value of parameter as bool array (vector<bool>).
+  /// Get value of parameter as bool array (rosidl_runtime_cpp::Vector<bool>).
   /**
    * \throws rclcpp::ParameterTypeException if the type doesn't match
    */
   RCLCPP_PUBLIC
-  const std::vector<bool> &
+  const rosidl_runtime_cpp::Vector<bool> &
   as_bool_array() const;
 
   /// Get value of parameter as integer array (vector<int64_t>).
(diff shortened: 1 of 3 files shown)


hidmic:
Tagging @ros2/team for awareness. I know this has the potential to break downstream code, but not doing this means no C++ introspection API for bool sequences. Also, `std::vector<bool>` drops several guarantees too that regular `std::vector<T>` has (see https://en.cppreference.com/w/cpp/container/vector_bool).

hidmic:
Full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15993/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10662/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16316/)

ivanpauno:
> Tagging @ros2/team for awareness. I know this has the potential to break downstream code, but not doing this means no C++ introspection API for bool sequences. Also, note that std::vector<bool> drops several guarantees that regular std::vector<T> has (see https://en.cppreference.com/w/cpp/container/vector_bool).

IIUC we didn't have support for `bool field[]` in the generated C and CPP code at all (it was being skipped).
Using a custom vector type to avoid the `std::vector<bool>` caveats sound like a good idea to me :+1:.

ivanpauno:
> IIUC we didn't have support for bool field[] in the generated C and CPP code at all (it was being skipped).

or was it already working for other typesupports and only not working for the introspection typesupport?

hidmic:
> or was it already working for other typesupports and only not working for the introspection typesupport?

@ivanpauno IIUC it was working everywhere but for introspection typesupports. Judging by https://github.com/ros2/rosidl/pull/647, introspection typesupports had not been touched in a long while.

PR #1881, opened 2022-02-01
Use spin() in component_manager_isolated.hpp

alsora:
Replace `spin_until_future_complete` with `spin` in `component_manager_isolated.hpp`.

`spin_until_future_complete()` is more inefficient as it needs to check the state of the future and the timeout after every work iteration.

This PR addresses the race-condition issues mentioned in https://github.com/ros2/rclcpp/pull/1781 by using the recently added `is_spinning()` API

Tests on the navigation stack show 10% less CPU usage when using `spin()` rather than `spin_until_future_complete()`

FYI @SteveMacenski @gezp @ivanpauno

### DIFF
--- a/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
@@ -96,7 +96,7 @@ private:
     // spin in a separate thread, i.e. the time gap between when the executor is created and when
     // it starts to spin is small (although it's not negligible).
 
-    while(!executor_wrapper.executor->is_spinning()) {
+    while (!executor_wrapper.executor->is_spinning()) {
       // This is an arbitrarily small delay to avoid busy looping
       rclcpp::sleep_for(std::chrono::milliseconds(1));
     }
--- a/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
@@ -38,7 +38,6 @@ class ComponentManagerIsolated : public rclcpp_components::ComponentManager
   {
     std::shared_ptr<rclcpp::Executor> executor;
     std::thread thread;
-    std::promise<void> promise;
   };
 
 public:
@@ -46,9 +45,7 @@ public:
   {
     if (node_wrappers_.size()) {
       for (auto & executor_wrapper : dedicated_executor_wrappers_) {
-        executor_wrapper.second.promise.set_value();
-        executor_wrapper.second.executor->cancel();
-        executor_wrapper.second.thread.join();
+        cancel_executor(executor_wrapper.second);
       }
       node_wrappers_.clear();
     }
@@ -67,8 +64,8 @@ protected:
     exec->add_node(node_wrappers_[node_id].get_node_base_interface());
     executor_wrapper.executor = exec;
     executor_wrapper.thread = std::thread(
-      [exec, cancel_token = executor_wrapper.promise.get_future()]() {
-        exec->spin_until_future_complete(cancel_token);
+      [exec]() {
+        exec->spin();
       });
     dedicated_executor_wrappers_[node_id] = std::move(executor_wrapper);
   }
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

fujitatomoya:
@alsora 

> Tests on the navigation stack show 10% less CPU usage when using spin() rather than spin_until_future_complete()

sounds great, would you mind sharing the platform you benchmarked this?

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
```diff
@@ -67,8 +64,8 @@ class ComponentManagerIsolated : public rclcpp_components::ComponentManager
     exec->add_node(node_wrappers_[node_id].get_node_base_interface());
     executor_wrapper.executor = exec;
     executor_wrapper.thread = std::thread(
-      [exec, cancel_token = executor_wrapper.promise.get_future()]() {
-        exec->spin_until_future_complete(cancel_token);
+      [exec]() {
+        exec->spin();
```
fujitatomoya, line 68: this is also bug fix to take advantage of `rclcpp::executors::MultiThreadedExecutor` with `--use_multi_threaded_executor` option? before this fix, it is not multi-threaded to execute the executables.

alsora:
@fujitatomoya the tests have been carried out by @gezp, he can definitely provide you more details but I think the 10% less cpu was on a raspberry pi 4, while on x86 the improvement was ~5%

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
alsora: I didn't realize that there was this option. I can have a look in a follow up PR.

fujitatomoya:
Rasp4 is one of the common benchmark platform we have too, this is good information. thanks!

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
```diff
@@ -81,14 +78,36 @@ class ComponentManagerIsolated : public rclcpp_components::ComponentManager
...
+    // We can't immediately call the cancel() API on an executor because if it is not
+    // already spinning, this operation will have no effect.
+    // We rely on the assumption that this class creates executors and then immediately makes them
+    // spin in a separate thread, i.e. the time gap between when the executor is created and when
+    // it starts to spin is small (although it's not negligible).
+
+    while (!executor_wrapper.executor->is_spinning()) {
+      // This is an arbitrarily small delay to avoid busy looping
+      rclcpp::sleep_for(std::chrono::milliseconds(1));
+    }
```
ivanpauno, line 102: I would love to have something better than this, but it also doesn't look too bad.
The only problem I see if the spinning thread crashes.

ivanpauno (approved):

ivanpauno:
LGTM with green CI

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
alsora: Yes, I agree that it's not elegant, but at the same time it's also very simple.
What about adding a timeout here?

There isn't currently any try/catch around spin, so if that thread crashes, the whole app will crash

alsora:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16071/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10723/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16408/)

alsora:
Hi, can someone give a look at the windows CI above? It says that there are more than 70 test failures.

They all seem unrelated tests to what I'm doing here and looking at the logs it may also seem an infrastructure issue, with some tests unable to start.
Do you have any insight?

ivanpauno:
Same failures are happening on nightlies, e.g.: https://ci.ros2.org/view/nightly/job/nightly_win_rel/2207/testReport/junit/(root)/projectroot/test_action_client/.

PR #1886, opened 2022-02-14
Use ament_generate_version_header

sloretz:
Part of https://github.com/ament/ament_cmake/pull/377 - this uses `ament_generate_version_header` instead of `ament_cmake_gen_version_h` to generate the version header. The installed header is identical.


sloretz:
CI here: https://github.com/ament/ament_cmake/pull/377#issuecomment-1039466580

wjwwood (approved):

PR #1891, opened 2022-02-24
create a base class for executors entities collectors

alsora:

This PR splits the static executor entities collector class into a base and a derived class.
This allows to reuse most of the functionalities when creating new executors.

Both classes are moved to a `detail` subfolder.

### DIFF
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -53,9 +53,10 @@ set(${PROJECT_NAME}_SRCS
   src/rclcpp/executable_list.cpp
   src/rclcpp/executor.cpp
   src/rclcpp/executors.cpp
+  src/rclcpp/executors/detail/entities_collector.cpp
+  src/rclcpp/executors/detail/static_executor_entities_collector.cpp
   src/rclcpp/executors/multi_threaded_executor.cpp
   src/rclcpp/executors/single_threaded_executor.cpp
-  src/rclcpp/executors/static_executor_entities_collector.cpp
   src/rclcpp/executors/static_single_threaded_executor.cpp
   src/rclcpp/expand_topic_or_service_name.cpp
   src/rclcpp/future_return_code.cpp
--- a/rclcpp/include/rclcpp/executors/detail/entities_collector.hpp
+++ b/rclcpp/include/rclcpp/executors/detail/entities_collector.hpp
@@ -1,0 +1,199 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__EXECUTORS__DETAIL__ENTITIES_COLLECTOR_HPP_
+#define RCLCPP__EXECUTORS__DETAIL__ENTITIES_COLLECTOR_HPP_
+
+#include <chrono>
+#include <list>
+#include <map>
+#include <memory>
+#include <vector>
+
+#include "rcl/guard_condition.h"
+#include "rcl/wait.h"
+
+#include "rclcpp/callback_group.hpp"
+#include "rclcpp/macros.hpp"
+#include "rclcpp/node_interfaces/node_base_interface.hpp"
+#include "rclcpp/visibility_control.hpp"
+#include "rclcpp/waitable.hpp"
+
+namespace rclcpp
+{
+namespace executors
+{
+namespace detail
+{
+typedef std::map<rclcpp::CallbackGroup::WeakPtr,
+    rclcpp::node_interfaces::NodeBaseInterface::WeakPtr,
+    std::owner_less<rclcpp::CallbackGroup::WeakPtr>> WeakCallbackGroupsToNodesMap;
+
+class EntitiesCollector
+  : public rclcpp::Waitable
+{
+public:
+  RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(EntitiesCollector)
+
+  // Constructor
+  RCLCPP_PUBLIC
+  EntitiesCollector() = default;
+
+  // Destructor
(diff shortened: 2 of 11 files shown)


alsora:
Starting CI since github actions is not up to date
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16256/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10877/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16615/)

alsora:
New CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16271/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10892/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16636/)

alsora:
Running again windows.
Failures don't seem to be related.

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16674/)

clalancette:
> Failures don't seem to be related.

The previous Windows failure is due to a bug I introduced into the CI scripts that have since been fixed.  Your next attempt should be a lot more successful :).

alsora:
Hi, can I get a review here?

alsora:
Hi, can we get a review here?

ivanpauno:
> Hi, can we get a review here?

I'm starting a review, but it'll take some time until I complete it.

> This PR splits the static executor entities collector class into a base and a derived class.
> This allows to reuse most of the functionalities when creating new executors.

Do you have any concrete example of this?

alsora:
IMO every executor should use an "entities collector" object, as refreshing the list of entities at every iteration is a huge waste of CPU.
This should be added also to the `SingleThreadedExecutor` allowing to finally get rid of the `StaticSingleThreadedExecutor`.

The `EventsExecutor` will need an entities collector, and it will be implement those virtual functions in a different way since it does not rely on a waitset.
Similarly I expect that other people interested in implementing/designing new executors may want to use this feature

alsora:
Running CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16884/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17285/)

ivanpauno:
> IMO every executor should use an "entities collector" object, as refreshing the list of entities at every iteration is a huge waste of CPU.

Agreed.
I think we should clean up all the code related to executors ("entities collector", "memory strategy", etc), and keep only one implementation that works like the current "static executor".

I'm asking for a concrete example of another extension because I don't see how the proposed interfaces in the base class are really useful as an extension point.
Some questions:

- Why not an abstract base class?
- Why were [these methods](https://github.com/ros2/rclcpp/blob/a4311b69c8944f5310ee1e86a755d8f597f88a07/rclcpp/include/rclcpp/executors/detail/entities_collector.hpp#L146-L164) chosen as an extension point?
- Why the base class has some [provided behavior](https://github.com/ros2/rclcpp/blob/a4311b69c8944f5310ee1e86a755d8f597f88a07/rclcpp/include/rclcpp/executors/detail/entities_collector.hpp#L186-L192)? How other executors take advantage of that?

i.e. I need some motivation of the proposed API, as it doesn't seem immediately obvious how other executors would take advantage of it.

alsora:
See here an example of an additional derived class, used for the events-executor: https://github.com/irobot-ros/events-executor/blob/main/irobot_events_executor/src/rclcpp/executors/events_executor/events_executor_entities_collector.cpp

>  - Why not an abstract base class?
>  - Why were [these methods](https://github.com/ros2/rclcpp/blob/a4311b69c8944f5310ee1e86a755d8f597f88a07/rclcpp/include/rclcpp/executors/detail/entities_collector.hpp#L146-L164) chosen as an extension point?

The entity collector functionalities can be essentially split into two phases.
An entity collector must 1) compute the list of entities from a node or callback group and 2) apply some operations on this list to "connect" the entities  with the executor.
The first item is agnostic with respect to the particular executor implementation, as it only depends on the implementation of nodes and  callback groups.
The main purpose of this PR is to avoid duplicating this first part across different executors, so an abstract base class is not an solution.
The second  part is executor-specific and it's the one that classes derived from the entities collector base must implement through the pure virtual methods you linked above.

>  - Why the base class has some [provided behavior](https://github.com/ros2/rclcpp/blob/a4311b69c8944f5310ee1e86a755d8f597f88a07/rclcpp/include/rclcpp/executors/detail/entities_collector.hpp#L186-L192)? How other executors take advantage of that?

These members of the base class are the "output" of the phase 1) mentioned above. Collected entities end up there and can then be accessed by the derived class for its executor-specific phase 2).

ivanpauno:
@alsora I looked at the code here and how it's used in https://github.com/irobot-ros/events-executor/blob/main/irobot_events_executor/src/rclcpp/executors/events_executor/events_executor_entities_collector.cpp.
The abstractions are a bit tough to understand and follow, this is partially fault of the current code.
I think we should clean things up if we want this to be maintainable in the future.

Some comments:

- The inteface the executor uses to interact with the entities collector is different for the "static executor" that for the "events executor", e.g. [this](https://github.com/ros2/rclcpp/blob/a4311b69c8944f5310ee1e86a755d8f597f88a07/rclcpp/include/rclcpp/executors/detail/static_executor_entities_collector.hpp#L178) vs [this](https://github.com/irobot-ros/events-executor/blob/c49b672f9596c7e28759d80d89334628d04f41e7/irobot_events_executor/src/rclcpp/executors/events_executor/events_executor_entities_collector.cpp#L336).
  The only methods in common are https://github.com/ros2/rclcpp/blob/a4311b69c8944f5310ee1e86a755d8f597f88a07/rclcpp/include/rclcpp/executors/detail/entities_collector.hpp#L70-L143.
  IMO, the "entities collector" is doing too much.
  We probably need to split some of its functionality in a few classes and use composition instead of inheritance.
- Most of the common logic in the "entities collector" base class is to collect "nodes" and "callback groups".
  That's later used in derived classes to iterate over all entities.
  It would be nices to hava an API to interate over entities, instead of using protected members in the derived classes.
  If not, it seems we're leaking too many details, and implementors of derived class have to basically read the implementation of the base class to understand how everything works.
- How the "static entities collector" interacts with "memory statregy" is also a bit confusing.
  Is memory strategy needed at all? should the "executable list" be filled directly?

I don't have much more time to give more design thoughts on this, but I think it would be better if we see first a proposal of some interfaces and how they interact and later an implementation.

ivanpauno (changes requested):
see above comment

ivanpauno:
@wjwwood maybe you're interested on providing feedback here as well

PR #1893, opened 2022-02-28
time_until_trigger returns max time if timer is cancelled

mauropasse:
This PR changes `rclcpp::TimerBase::time_until_trigger` to return `std::chrono::nanoseconds::max()` if the timer is canceled, since is does not seem correct to return the `rcl_timer_get_time_until_next_call` if the timer won't be called at that time.

### DIFF
--- a/rclcpp/src/rclcpp/timer.cpp
+++ b/rclcpp/src/rclcpp/timer.cpp
@@ -119,7 +119,9 @@ TimerBase::time_until_trigger()
   int64_t time_until_next_call = 0;
   rcl_ret_t ret = rcl_timer_get_time_until_next_call(
     timer_handle_.get(), &time_until_next_call);
-  if (ret != RCL_RET_OK) {
+  if (ret == RCL_RET_TIMER_CANCELED) {
+    return std::chrono::nanoseconds::max();
+  } else if (ret != RCL_RET_OK) {
     rclcpp::exceptions::throw_from_rcl_error(ret, "Timer could not get time until next call");
   }
   return std::chrono::nanoseconds(time_until_next_call);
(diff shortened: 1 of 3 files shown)


alsora:
Starting CI since github action is not up to date
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16258/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10879/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16617/)

mauropasse:
I'm adding a test for this, FYI

fujitatomoya (commented):
How about addressing this problem in `rcl` implementation instead?

- maintenance cost effective.
- client frontends can rely on just `rcl` and be consistent behavior.

what about updating `rcl_timer_get_time_until_next_call` with `RCL_RET_TIMER_CANCELED`?

mauropasse:
Thanks @fujitatomoya for your review. I applied your suggestion on https://github.com/ros2/rcl/pull/963

rclcpp/src/rclcpp/timer.cpp
```diff
@@ -119,7 +119,9 @@ TimerBase::time_until_trigger()
   int64_t time_until_next_call = 0;
   rcl_ret_t ret = rcl_timer_get_time_until_next_call(
     timer_handle_.get(), &time_until_next_call);
-  if (ret != RCL_RET_OK) {
+  if (ret == RCL_RET_TIMER_CANCELED) {
+    return std::chrono::nanoseconds::max();
```
fujitatomoya, line 123: probably we would want to document this? when timer is canceled, it will return that max()? instead, it just appears to be the time for the next trigger?

fujitatomoya (approved):

PR #1896, opened 2022-03-08
Micro-optimizations in rclcpp

clalancette:
While doing some profiling of the rclcpp code, I noticed a few different places where we were unnecessarily doing work.  This PR has 3 commits that address each of these problems.

1. The first patch adds a new method to the `CallbackGroup` class that collects all of the pointers at once.  This avoids the overhead of taking and dropping the mutex inside of `CallbackGroup` over and over.  I incidentally don't love the name `collect_all_ptrs`, so I'm happy to rename that to something else.
2. The second patch changes the API of the `MemoryStrategy` class to pass the shared_ptrs as const-ref in the APIs.  Since the default implementation doesn't need to hold an additional reference, this reduces overhead for creating and destroying the shared_ptr class quite a bit.  If a subclass were written that *does* need to hold onto an additional reference, they can explicitly do that by invoking the shared_ptr copy constructor.
3. The last patch changes a bunch of loops to use const-ref to shared_ptr instead of copying the shared_ptr.  As with the second point, this reduces overhead in creating and destroying the shared_ptr's that we don't need to hold a reference to.

There are more details about each of these points in the individual commits.  In the program I was profiling, the changes here (along with a couple of other minor changes to rcl and rmw_fastrtps which will be coming up) reduced overhead by about 20%.

### DIFF
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -138,6 +138,14 @@ public:
   const CallbackGroupType &
   type() const;
 
+  RCLCPP_PUBLIC
+  void collect_all_ptrs(
+    std::function<void(const rclcpp::SubscriptionBase::SharedPtr &)> sub_func,
+    std::function<void(const rclcpp::ServiceBase::SharedPtr &)> service_func,
+    std::function<void(const rclcpp::ClientBase::SharedPtr &)> client_func,
+    std::function<void(const rclcpp::TimerBase::SharedPtr &)> timer_func,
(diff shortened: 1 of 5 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16298/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10918/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16665/)

fujitatomoya (approved):

clalancette:
Thanks @fujitatomoya !

PR #1897, opened 2022-03-09
Fix a bunch more rosdoc2 issues in rclcpp.

clalancette:
There are a smattering of problems in here:

1.  We weren't properly using PREDEFINE for all of our macros.
2.  The Doxyfiles were referencing tag files that may not exist
(this will be handled by rosdoc2 automatically).
3.  The C++ parser in doxygen can't handle "friend classname",
but can handle "friend class classname"; it shouldn't matter
one way or the other to the compiler.
4.  There were a couple of parameters that were not documented.
5.  The doxygen C++ parser can't handle decltype in all situations.
6.  There was a structure using a C-style declaration.

This patch fixes all of the above issues.  We still aren't totally
clean on a rosdoc2 build, but we are a lot closer.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/types.hpp
+++ b/rclcpp_action/include/rclcpp_action/types.hpp
@@ -17,6 +17,7 @@
 
 #include <array>
 #include <climits>
+#include <functional>
 #include <string>
 
 #include "rcl_action/types.h"
--- a/rclcpp_action/include/rclcpp_action/types.hpp
+++ b/rclcpp_action/include/rclcpp_action/types.hpp
@@ -15,8 +15,8 @@
 #ifndef RCLCPP_ACTION__TYPES_HPP_
 #define RCLCPP_ACTION__TYPES_HPP_
 
+#include <array>
 #include <climits>
-#include <functional>
 #include <string>
 
 #include "rcl_action/types.h"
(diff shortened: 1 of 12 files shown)


clalancette:
@hidmic FYI

rclcpp_action/include/rclcpp_action/types.hpp
```diff
@@ -15,8 +15,8 @@
 #ifndef RCLCPP_ACTION__TYPES_HPP_
 #define RCLCPP_ACTION__TYPES_HPP_
 
+#include <array>
 #include <climits>
-#include <functional>
```
aprotyas, line 19: `std::hash<T>` transitively gets included here, but shouldn't this header stay to explicitly include the template?

clalancette: Oh, I see.  I didn't realize that `std::less` and `std::hash` are included by putting them into namespace `std`.  Technically, we don't actually need to have the `#include` in here, since we are generating the code right here and we don't depend on anything else.  But I guess it would be nicer to include the template specialization for `std::hash` and `std::less`, so I'll add that back.  Thanks for the heads up.

clalancette: Fixed in 1492cd9

aprotyas (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16312/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10932/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16682/)

PR #1908, opened 2022-03-27
[NodeParameters] Set name in param info pre-check

### LINKED ISSUE #1906: broken InvalidParameterValueException message
## Bug report

When setting a parameter in the command line, which is outside of the limits, I get an error message that is not correctly formatted:
```
terminate called after throwing an instance of 'rclcpp::exceptions::InvalidParameterValueException'
  what():  parameter 'ExposureTime' could not be set: Parameter {} doesn't comply with integer range.
```

I assume that `{}` is supposed to be something else, e.g. contain the value.

**Required Info:**

- Operating System: Ubuntu 20.04
- Installation type: binary
- Version or commit hash: galactic
- Client library (if applicable): rclcpp

#### Steps to reproduce issue
1. declare a parameter with bounds: `declare_parameter("ExposureTime", default_value, param_descr);` where `param_descr` is a `ParameterDescriptor` with integer bounds, e.g. (400 ... 20000).
2. initialise the parameter outside of the bounds, e.g. `--ros-args -p ExposureTime:=300`

#### Expected behavior
The exception should contain a useful error message.

#### Actual behavior
`Parameter {} doesn't comply with integer range.` does not make sense to me.

aprotyas:
In `__check_parameters`, there are function-scoped parameter
descriptor structures generated by searching the `ParameterInfo`
map provided from other declaratory functions. This PR populates
this temporary structure with the parameter name information.
This way, the change does not show in other data structures, but the
name field is also not empty when used to generate error messages in
`__check_parameter_value_in_range`.

See this [demo node](https://github.com/aprotyas/ros2_playground/tree/master/rclcpp_1906) for testing purposes.

Fixes #1906.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -298,6 +298,11 @@ __set_parameters_atomically_common(
   const OnParametersSetCallbackType & callback,
   bool allow_undeclared = false)
 {
+  // Set the parameter name before checks, for more desciprtive error messages
+  for (size_t i = 0; i < parameters.size(); ++i) {
+    const std::string & name = parameters[i].get_name();
+    parameter_infos[name].descriptor.name = parameters[i].get_name();
+  }
   // Check if the value being set complies with the descriptor.
   rcl_interfaces::msg::SetParametersResult result = __check_parameters(
     parameter_infos, parameters, allow_undeclared);
@@ -314,7 +319,6 @@ __set_parameters_atomically_common(
   if (result.successful) {
     for (size_t i = 0; i < parameters.size(); ++i) {
       const std::string & name = parameters[i].get_name();
-      parameter_infos[name].descriptor.name = parameters[i].get_name();
       parameter_infos[name].descriptor.type = parameters[i].get_type();
       parameter_infos[name].value = parameters[i].get_parameter_value();
     }
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -236,7 +236,9 @@ __check_parameters(
     } else {
       descriptor = parameter_infos[name].descriptor;
     }
-    descriptor.name = name;
+    if (descriptor.name.empty()) {
+      descriptor.name = name;
+    }
     const auto new_type = parameter.get_type();
     const auto specified_type = static_cast<rclcpp::ParameterType>(descriptor.type);
     result.successful = descriptor.dynamic_typing || specified_type == new_type;
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -236,6 +236,7 @@ __check_parameters(
     } else {
       descriptor = parameter_infos[name].descriptor;
     }
+    descriptor.name = name;
     const auto new_type = parameter.get_type();
     const auto specified_type = static_cast<rclcpp::ParameterType>(descriptor.type);
     result.successful = descriptor.dynamic_typing || specified_type == new_type;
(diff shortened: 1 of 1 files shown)


aprotyas:
I'm unsure how benign this change is given it triggers a couple of parameter test failures in my local setup, but I'd like to get input from maintainers about said failures.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -298,6 +298,11 @@ __set_parameters_atomically_common(
   const OnParametersSetCallbackType & callback,
   bool allow_undeclared = false)
 {
+  // Set the parameter name before checks, for more desciprtive error messages
+  for (size_t i = 0; i < parameters.size(); ++i) {
+    const std::string & name = parameters[i].get_name();
+    parameter_infos[name].descriptor.name = parameters[i].get_name();
```
fujitatomoya, line 304: see https://github.com/ros2/rclcpp/issues/1906#issuecomment-1081089117. i am not sure which tests are actually failing though.

aprotyas: Thanks for context about the ParameterInfo structure - do you have a guess why this would cause tests to fail though?

aprotyas: Dropping this approach and favoring 746d93b instead.

fujitatomoya: > Thanks for context about the ParameterInfo structure - do you have a guess why this would cause tests to fail though?

no, not really. i have never tried this patch with my local environment yet.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -236,6 +236,7 @@ __check_parameters(
     } else {
       descriptor = parameter_infos[name].descriptor;
     }
+    descriptor.name = name;
```
fujitatomoya, line 239: this will change the behavior to overwrite the `descriptor.name` from user application.

before, can be `what():  parameter 'foo' could not be set: Parameter {bar} doesn't comply with integer range.`, if user put the descriptor name into `bar`. but after this PR, it will be `what():  parameter 'foo' could not be set: Parameter {foo} doesn't comply with integer range.`. so it cannot print the descriptor name.

probably we would want to store actual parameter name in here if `descriptor.name` is empty? or maybe overwriting would be okay? i would like to hear from other people on this.

aprotyas: Good point! I forgot that this descriptor is provided by the user.

> or maybe overwriting would be okay?

Maybe. It's unclear to me what the expectation with the parameter descriptors are. Do they need to be fully populated by the user? What if the user provides a different parameter name? I don't see any code that checks if there is a discrepancy between the parameter name and the name provided in the descriptor, for example.

aprotyas: In `declare_parameter(const std::string&)`, it looks like an empty descriptor is provided to satisfy the more complete `declare_parameter` signature. https://github.com/ros2/rclcpp/blob/eac006317692b7eff3aed671eadbff3d72cca098/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L467-L473

That leads me to believe that the `name` argument provided to `declare_parameter` should be the only source of ground truth in this situation, i.e. the `name` argument is the parameter argument, and what is found in the parameter descriptor's name field is irrelevant. I don't have enough context about descriptors to assert this though. Thoughts?

fujitatomoya: i might be wrong...

> Do they need to be fully populated by the user?

AFAIK, i do not think so, this is optional and user can/may put some parts only.

> What if the user provides a different parameter name?

in this case, actual parameter name will be used instead of descriptor name. the same behavior goes to type.

i think that this behavior is intentional, so that user application can use descriptor (for instance of range declaration) on some other parameters as well.

> probably we would want to store actual parameter name in here if descriptor.name is empty?

i am currently inclined to have this one.

- we can keep the user descriptor name if that exists.
- we do not change current behavior.
- we can address the original issue https://github.com/ros2/rclcpp/issues/1906

i am open to hear from other opinion 👂

fujitatomoya (approved):

fujitatomoya:
I am not sure why windows CI is failing, retry.

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16823/)

aprotyas:
Trying one more time before looking further:
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/16826/)

fujitatomoya:
I believe https://ci.ros2.org/job/ci_windows/16826/cmake/new/NORMAL/ is unrelated to the fix.

Issue #1917, opened 2022-04-07
GuardCondition should have a callback that gets handled by the executor


sloretz:
## Feature request

Discovered while looking at ros2/rclcpp#1916

#### Feature description

I think it would be useful to have a callback that gets run by the executor after the guard condition is triggered.
`rclpy` [already has this feature](https://github.com/ros2/rclpy/blob/c704619fc9666c5403a7d0373eb4795c94b721b1/rclpy/rclpy/guard_condition.py#L21).

This could be used to do work in the executor. One could use it to complete a future to work around ros2/rclcpp#1916.

You can sort of do this now with timers with a 0 second period by canceling them in the callback. The timer callback won't run again until it's reset, so resetting is sort of like calling `trigger()`).

#### Implementation considerations

The `GuardCondition` class has a callback that gets called when it's triggered, but this isn't suitable for this feature.
https://github.com/ros2/rclcpp/blob/85a7046ac39bfc0d787ed2bbe62574424e31b871/rclcpp/include/rclcpp/guard_condition.hpp#L112

This callback replaces the trigger implementation so that the `rcl_trigger_guard_condition` method never gets called, and the executor isn't woken. A different callback would be needed to implement this feature.

https://github.com/ros2/rclcpp/blob/85a7046ac39bfc0d787ed2bbe62574424e31b871/rclcpp/src/rclcpp/guard_condition.cpp#L79-L87

alsora:
I think that this is an error in the `guard_condition.cpp` code.

Calling `rcl_trigger_guard_condition` should be independent from whether there is an `on_trigger_callback_` or not.

wjwwood:
> This could be used to do work in the executor. One could use it to complete a future to work around https://github.com/ros2/rclcpp/issues/1916.

I'd rather see a `Executor::call_soon(<function or lambda>)` kind of function. (Ab)using guard conditions to accomplish this is no different really than doing the same with zero duration, one-off timers.

That being said, if users are to use a guard condition effectively, they will need to have an associated callback.

PR #1923, opened 2022-04-29
Always trigger guard condition waitset

alsora:
Ttrigger guard condition waitset regardless of whether a trigger callback is present
This should address https://github.com/ros2/rclcpp/issues/1917

Also minor cleanup at the class (remove unneded lock and better use of if-else)

### DIFF
--- a/rclcpp/src/rclcpp/guard_condition.cpp
+++ b/rclcpp/src/rclcpp/guard_condition.cpp
@@ -100,7 +100,7 @@ void
 GuardCondition::add_to_wait_set(rcl_wait_set_t * wait_set)
 {
   std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
-  
+
   if (exchange_in_use_by_wait_set_state(true)) {
     if (wait_set != wait_set_) {
       throw std::runtime_error("guard condition has already been added to a wait set.");
--- a/rclcpp/src/rclcpp/guard_condition.cpp
+++ b/rclcpp/src/rclcpp/guard_condition.cpp
@@ -74,16 +74,19 @@ GuardCondition::get_rcl_guard_condition() const
 void
 GuardCondition::trigger()
 {
-  std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
+  rcl_ret_t ret = rcl_trigger_guard_condition(&rcl_guard_condition_);
+  if (RCL_RET_OK != ret) {
+    rclcpp::exceptions::throw_from_rcl_error(ret);
+  }
 
-  if (on_trigger_callback_) {
-    on_trigger_callback_(1);
-  } else {
-    rcl_ret_t ret = rcl_trigger_guard_condition(&rcl_guard_condition_);
-    if (RCL_RET_OK != ret) {
-      rclcpp::exceptions::throw_from_rcl_error(ret);
+  {
+    std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
+
+    if (on_trigger_callback_) {
+      on_trigger_callback_(1);
+    } else {
+      unread_count_++;
     }
-    unread_count_++;
   }
 }
 
@@ -96,8 +99,6 @@ GuardCondition::exchange_in_use_by_wait_set_state(bool in_use_state)
 void
 GuardCondition::add_to_wait_set(rcl_wait_set_t * wait_set)
 {
-  std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
-
   if (exchange_in_use_by_wait_set_state(true)) {
     if (wait_set != wait_set_) {
       throw std::runtime_error("guard condition has already been added to a wait set.");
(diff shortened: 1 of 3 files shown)


rclcpp/src/rclcpp/guard_condition.cpp
```diff
@@ -96,8 +99,6 @@ GuardCondition::exchange_in_use_by_wait_set_state(bool in_use_state)
 void
 GuardCondition::add_to_wait_set(rcl_wait_set_t * wait_set)
 {
-  std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
-
```
fujitatomoya, line 100: How about the following MT case?

1. Thread-A comes with `wait_set-X`.
2. Context switch to thread-B right after thread-A stores `wait_set-X` to `wait_set_`.
3. Thread-B also comes with `wait_set-X`.
4. Thread-B and Thread-A might call `rcl_wait_set_add_guard_condition` at the same time.

as far as i see the doc of `rcl_wait_set_add_subscription`, these functions are not thread-safe?

alsora: I'm not sure.
The other waitables (e.g. actions, intra-process sub) do not use a mutex there.
Also the executor does not seem to use a mutex when adding entities (subscriptions, services, etc) to waitsets.

wjwwood: It's undefined behavior to add an item to two different wait sets, and it's also undefined behavior to try to wait on a single wait set from two different threads.

There is a thread-safe `rclcpp::WaitSet` that protects this case from happening.

alsora: It looks to me that the `add_to_wait_set`  function is already preventing to add an item to two different wait-sets.

This is the current code.
The atomic `in_use_by_wait_set_` should be enough to avoid the undefined behavior, without need for a mutex.

```
void
GuardCondition::add_to_wait_set(rcl_wait_set_t * wait_set)
{
  if (exchange_in_use_by_wait_set_state(true)) {
    if (wait_set != wait_set_) {
      throw std::runtime_error("guard condition has already been added to a wait set.");
    }
  } else {
    wait_set_ = wait_set;
  }

  rcl_ret_t ret = rcl_wait_set_add_guard_condition(wait_set, &this->rcl_guard_condition_, NULL);
  if (RCL_RET_OK != ret) {
      ret, "failed to add guard condition to wait set");
  }
}
```

alsora: It looks to me that the `add_to_wait_set`  function is already preventing to add an item to two different wait-sets.

This is the current code.
The atomic `in_use_by_wait_set_` should be enough to avoid the undefined behavior, without need for a mutex.

```
void
GuardCondition::add_to_wait_set(rcl_wait_set_t * wait_set)
{
  if (exchange_in_use_by_wait_set_state(true)) {
    if (wait_set != wait_set_) {
      throw std::runtime_error("guard condition has already been added to a wait set.");
    }
  } else {
    wait_set_ = wait_set;
  }

  rcl_ret_t ret = rcl_wait_set_add_guard_condition(wait_set, &this->rcl_guard_condition_, NULL);
  if (RCL_RET_OK != ret) {
      ret, "failed to add guard condition to wait set");
  }
}
```

alsora:
I restored the mutex to prevent race condition that could cause problems.

rclcpp/src/rclcpp/guard_condition.cpp
```diff
@@ -97,7 +100,7 @@ void
 GuardCondition::add_to_wait_set(rcl_wait_set_t * wait_set)
 {
   std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
-
+
```
fujitatomoya, line 103: nit: unnecessary difference.

fujitatomoya (approved):
lgtm with minor nit

wjwwood (commented):
Looks reasonable.

It seems like a test to avoid this regressing in the future would be good to have.

Also, if it's not already clear from the docstring what should happen in this corner case, it should be clarified there.

alsora:
Added unit-tests.

Running full CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16719/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11299/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17113/)

wjwwood:
This doc block should mention how it interacts with the "on trigger callback":

https://github.com/ros2/rclcpp/blob/02802bcc385c3d6d814add93618f3388d44adec7/rclcpp/include/rclcpp/guard_condition.hpp#L75-L85

And, I guess we missed it, but we should have some documentation on the "on trigger callback" method as well (it currently has none):

https://github.com/ros2/rclcpp/blob/02802bcc385c3d6d814add93618f3388d44adec7/rclcpp/include/rclcpp/guard_condition.hpp#L110-L112

alsora:
@wjwwood @fujitatomoya anything else to do here?

wjwwood (approved):
lgtm

Sorry for not re-reviewing it. I don't get notifications for pushed commits, so a comment is helpful to bump a re-review.

PR #1947, opened 2022-06-06
Support pre-set and post-set parameter callbacks in addition to on-set-parameter-callback.

deepanshubansal01:
This PR allows supporting `add_pre_set_parameter_callback` and `add_post_set_parameter_callback` in addition to the existing `add_on_set_parameter_callback`  for Node API. 

**Design document:**  https://github.com/ros2/rclcpp/blob/deepanshu/local-param-changed-callback-support/rclcpp/doc/proposed_node_parameter_callbacks.md
**Demo node:** https://github.com/ros2/demos/pull/565
**rclpy ticket:** https://github.com/ros2/rclpy/pull/966
**Related discussion:** https://github.com/ros2/rclcpp/issues/1190, https://github.com/ros2/rclcpp/issues/609

cc: @adityapande-1995 @jacobperron 

Signed-off-by: deepanshu <deepanshubansal01@gmail.com>


rclcpp/include/rclcpp/node.hpp
```diff
@@ -933,6 +938,11 @@ class Node : public std::enable_shared_from_this<Node>
   OnSetParametersCallbackHandle::SharedPtr
   add_on_set_parameters_callback(OnParametersSetCallbackType callback);
 
+  RCLCPP_PUBLIC
+      RCUTILS_WARN_UNUSED
+  OnSetLocalParametersCallbackHandle::SharedPtr
+  add_local_parameters_callback(OnLocalParametersSetCallbackType callback);
```
adityapande-1995, line 944: Similar t the nomenclature of ``add_n_set_parameters_callback``, we should have this as ``add_post_set_parameters_callback``

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
```diff
@@ -200,6 +210,8 @@ class NodeParameters : public NodeParametersInterface
 
   CallbacksContainerType on_parameters_set_callback_container_;
 
+  std::list<OnSetLocalParametersCallbackHandle::WeakPtr> on_local_parameters_set_callback_container_;
```
adityapande-1995, line 213: same here and elsewhere, ``post_set_parameter_callback_container_``.

adityapande-1995 (commented):
Is this what happens currently when you chain parameters like this ? (Should nehave like this ideally I believe)
```
param A (set successfully) --> param B (fails to set) --> param C (should not be triggered) 
```

rclcpp/include/rclcpp/node.hpp
adityapande-1995: ``OnSetLocalParametersCallbackHandle`` can be changed to ``PostSetParametersCallbackHandle``

jacobperron: IMO, we should consider moving away from the current naming scheme. I think "on_set_parameters" is confusing, and should probably be deprecated. I also think the term "local" is kind of redundant since both callbacks in question deal with parameters local to the node. It's only when a user uses ParameterEventsHandler (which is separate from the node) that they can get parameter events for other nodes.

Changing the name of the existing functions will result in more churn, but I think it's worth taking the time now to reflect on what we want the API to look like.

It's just my opinion, but I would ultimately like to see:

- `add_parameters_validation_callback`, as the new name for the validation feature.
- `add_parameters_set_callback` (or `add_on_parameters_set_callback`), as the name for the new feature.

adityapande-1995:
> > Is this what happens currently when you chain parameters like this ? (Should behave like this ideally I believe)
> > ```
> > param A (set successfully) --> param B (fails to set) --> param C (should not be triggered) 
> > ```
> 
> Is param B and param C being set inside post callback?

Yup, So I meant A's callback sets B, B's callback sets C.

deepanshubansal01:
> > > Is this what happens currently when you chain parameters like this ? (Should behave like this ideally I believe)
> > > ```
> > > param A (set successfully) --> param B (fails to set) --> param C (should not be triggered) 
> > > ```
> > 
> > 
> > Is param B and param C being set inside post callback?
> 
> Yup, So I meant A's callback sets B, B's callback sets C.

I will check this to confirm the behaviour.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -587,6 +587,26 @@ __find_parameter_by_name(
...
+      if (nullptr != shared_handle) {
+        shared_handle->callback(parameters);
+        it++;
+      } else {
+        it = on_local_parameters_set_callback_container_.erase(it);
+      }
+    }
+  }
+
+  return result;
```
adityapande-1995, line 607: One more concern here that might be worth discussing : When we set a parameter in the command line say using
``ros2 param set /some_node some_parameter False``, I want it to return after the validation call ends. According to the current implementation, it will block until the new "pose-set" callbacks have returned, @deepanshubansal01 can correct me if I'm wrong.

Ideally I would want it to return after the validation call is done. @jacobperron what do you think of am async implementation ?

rclcpp/include/rclcpp/node.hpp
clalancette: > It's just my opinion, but I would ultimately like to see:
>
>    add_parameters_validation_callback, as the new name for the validation feature.
>    add_parameters_set_callback (or add_on_parameters_set_callback), as the name for the new feature.

If I were designing both of these APIs from the ground-up, I would agree with you.  But I'm actually somewhat reluctant to make this API change now because:

1. We already renamed the existing API twice before.  For Eloquent, we renamed the original `register_param_change_callback` to `set_on_parameters_set_callback`, and then in Galactic we changed it to `add_on_set_parameters_callback`.  While we can keep renaming it, we incur costs on the downstream consumers every time we do it.
2. `add_on_parameters_set_callback` is too similar to `add_on_set_parameters_callback`, even though they have different semantics.

I'm not totally sure what to do about this, but I thought I would point it out.  At the very least I think the second one should have a different name.

deepanshubansal01:
> > > > Is this what happens currently when you chain parameters like this ? (Should behave like this ideally I believe)
> > > > ```
> > > > param A (set successfully) --> param B (fails to set) --> param C (should not be triggered) 
> > > > ```
> > > 
> > > 
> > > Is param B and param C being set inside post callback?
> > 
> > 
> > Yup, So I meant A's callback sets B, B's callback sets C.
> 
> I will check this to confirm the behaviour.

deepanshubansal01:
> Member

I checked, this is indeed the behaviour right now, so should be good.

deepanshubansal01:
> > > > > Is this what happens currently when you chain parameters like this ? (Should behave like this ideally I believe)
> > > > > ```
> > > > > param A (set successfully) --> param B (fails to set) --> param C (should not be triggered) 
> > > > > ```
> > > > 
> > > > 
> > > > Is param B and param C being set inside post callback?
> > > 
> > > 
> > > Yup, So I meant A's callback sets B, B's callback sets C.
> > 
> > 
> > I will check this to confirm the behaviour.

I checked, this is indeed the behaviour right now, so should be good.

adityapande-1995:
Some related discussion here : https://github.com/ros2/rclcpp/issues/1190, https://github.com/ros2/rclcpp/issues/609

jacobperron:
To recap from our discussion at the ROS 2 meeting:

- Our current plan is to introduce a "pre set parameters" callback, that can return a modified list of parameters that is then forwarded to the current "on set parameters" callback for validation. This should satisfy the use-case of updating additional parameters conditioned on changes to a parameter.
- The post callback is still valuable as a place to cause side-effects based on parameter changes, but we will avoid subverting the parameter recursion lock (which we no longer need since we can use the pre-callback to request changes to parameters based on another).

deepanshubansal01:
Not sure but should we do something about `add_pre_set_parameters_callback`, `add_on_set_parameters_callback`, `add_post_set_parameters_callback` having almost similar code except that they are using different callback containers.

deepanshubansal01:
One more thing which we should probably address is that since the `pre_set_parameter_callback` will be mainly used to modify the` rclcpp::Parameter` list, this might result in a different behaviour where the `set_parameters_atomically` should have been called **non-atomically** but is being called **atomically** now.

We can probably do something about this by not calling the registered `pre_set_parameter_callback` inside the `set_parameters_atomically` but outside it. After this we can change if `set_parameters_atomically` needs to be called atomically or non atomically.

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,21 @@
+# Proposed node parameters callback Design
+
+## Introduction:
+
+ The original requirement came in **gazeb_ros_pkgs** for setting individual wheel slip params based on global wheel slip value [link to original issue](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1365). The main idea was to set some parameters using ```set_parameter``` API once the a given param was set successfully. The requirement let to the discussion of supporting registered callbacks once the parameters are set and have been requested by users before. Some related discussion here [#609](https://github.com/ros2/rclcpp/issues/609) [#1789](https://github.com/ros2/rclcpp/pull/1789).
```
jacobperron, line 5: style nitpick: please use one sentence per line to help keep future diffs easier to read (see our [style guide](https://docs.ros.org/en/rolling/Contributing/Code-Style-Language-Versions.html#markdown-restructured-text-docblocks))

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,21 @@
...
+
+In the current Node API, the ```add_on_set_parameters_callback``` is used for doing any validation for parameters before setting the parameters successfully. Once the parameters are validated ```ParameterEventHandler``` object is used to publish any changes to node parameters on **/parameter_events** topic, which can be subscribed by any node to see the changed parameter. Note that the **/parameter_events** is a topic for all nodes on the network, and we have to rely on executors to process those.
```
jacobperron, line 7: I think only one set of backticks are necessary for inline formatting in markdown,
```suggestion
In the current Node API, the `add_on_set_parameters_callback` is used for doing any validation for parameters before setting the parameters successfully. Once the parameters are validated `ParameterEventHandler` object is used to publish any changes to node parameters on **/parameter_events** topic, which can be subscribed by any node to see the changed parameter. Note that the **/parameter_events** is a topic for all nodes on the network, and we have to rely on executors to process those.
```

This can be updated throughout the document.

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,21 @@
...
+
+ The validation callback is often abused to trigger side-effects in the code, for instance updating class attributes even before parameter has been set successfully. Instead of relying on **/parameter_events** it would be good to support an internal node interface API for registering callbacks for successful parameter set ```add_post_set_parameters_callback```.
+
+ We can use the proposed ```add_post_set_parameters_callback``` for setting some param but this might result in unpredictable behaviour because of infinite recusion and due to the params being set atomically. In order to get around this we propose adding another registered callback ```add_pre_set_parameters_callback``` which will be triggered before the validation callbacks and can be used to modify the parameter list as required.
```
jacobperron, line 11: This paragraph seems like it belongs in the next section instead of the intro, as it is describing the design.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -859,11 +859,26 @@ class Node : public std::enable_shared_from_this<Node>
...
 
+  using PostSetParametersCallbackHandle =
+      rclcpp::node_interfaces::PostSetParametersCallbackHandle;
+  using PostSetParametersCallbackType =
+      rclcpp::node_interfaces::NodeParametersInterface::PostSetParametersCallbackType;
+
+  RCLCPP_PUBLIC
+  RCUTILS_WARN_UNUSED
+  PreSetParametersCallbackHandle::SharedPtr
+  add_pre_set_parameters_callback(PreSetParametersCallbackType callback);
```
jacobperron, line 880: Please add a detailed docblock describing this method (similar to `add_on_set_parameters_callback`).

rclcpp/include/rclcpp/node.hpp
```diff
@@ -933,6 +948,11 @@ class Node : public std::enable_shared_from_this<Node>
   OnSetParametersCallbackHandle::SharedPtr
   add_on_set_parameters_callback(OnParametersSetCallbackType callback);
 
+  RCLCPP_PUBLIC
+  RCUTILS_WARN_UNUSED
+  PostSetParametersCallbackHandle::SharedPtr
+  add_post_set_parameters_callback(PostSetParametersCallbackType callback);
```
jacobperron, line 954: Same here; add some documentation.

rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
```diff
@@ -205,6 +241,15 @@ class NodeParametersInterface
...
   remove_on_set_parameters_callback(const OnSetParametersCallbackHandle * const handler) = 0;
 
+  /// Remove a callback registered with `add_post_set_parameters_callback`.
+  /**
+   * \sa rclcpp::Node::remove_post_set_parameters_callback
+   */
+  RCLCPP_PUBLIC
+  virtual
+  void
+  remove_post_set_parameters_callback(const PostSetParametersCallbackHandle * const handler) = 0;
```
jacobperron, line 251: I think we should have a remove method for the pre set parameters callback too.

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
```diff
@@ -171,15 +171,29 @@ class NodeParameters : public NodeParametersInterface
...
+  PreSetParametersCallbackHandle::SharedPtr
+  add_pre_set_parameters_callback(PreSetParametersCallbackType callback) override;
+
   RCLCPP_PUBLIC
   RCUTILS_WARN_UNUSED
   OnSetParametersCallbackHandle::SharedPtr
   add_on_set_parameters_callback(OnParametersSetCallbackType callback) override;
 
+  RCLCPP_PUBLIC
+      RCUTILS_WARN_UNUSED
```
jacobperron, line 185: whitespace looks wrong here.

rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
```diff
@@ -137,7 +137,7 @@ class NodeParameters : public NodeParametersInterface
   RCLCPP_PUBLIC
   rcl_interfaces::msg::SetParametersResult
   set_parameters_atomically(
-    const std::vector<rclcpp::Parameter> & parameters) override;
+      const std::vector<rclcpp::Parameter> & parameters) override;
```
jacobperron, line 140: I'd revert this whitespace change unless it is necessary.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -933,6 +948,11 @@ class Node : public std::enable_shared_from_this<Node>
...
+
   /// Remove a callback registered with `add_on_set_parameters_callback`.
```
jacobperron, line 956: We should expose remove methods for the pre and post callbacks here too.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -596,12 +604,27 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
 
   rcl_interfaces::msg::SetParametersResult result;
 
+  // check for any pre-set parameter registered callbacks
```
jacobperron, line 607: Is there an issue if we instead moved the pre/post callbacks into `__set_parameters_atomically_common`?
It seems more intuitive to me to process those callbacks there, since technically that's where parameters are ultimately being set. As is currently implemented, pre/post callbacks won't be called when parameters are declared, but the validation callbacks do get called. Is there a reason for this asymmetry?

jacobperron (commented):
Looks reasonable to me. Besides my comments below, we should add some tests for the new callbacks.

jacobperron:
> One more thing which we should probably address is that since the pre_set_parameter_callback will be mainly used to modify the rclcpp::Parameter list, this might result in a different behaviour where the set_parameters_atomically should have been called non-atomically but is being called atomically now.

IMO, it seems like a very narrow use case to want to amend the parameter list to include some that should be set non-atomically. We could maintain two different lists of "pre parameter" callbacks, one for atomic and one for non-atomic setting, but that seems like overkill to me. If the need arises (or others feel strongly about it) I think it would be easy to create an additional API after the fact :shrug: 

It's probably worth mentioning that the current pre set callbacks will set the returned parameter list atomically in the design doc and API documentation.

deepanshubansal01:
Based on the discussion on where should the **pre and post set parameter** callback _**live in the code**_, and whether we should support these callbacks for declare parameters. 

The `pre set parameter` callback should ideally be called at the start of any of the `set_parameter*` API. This is because since the pre set parameter will likely modify the list of parameters we should do all the validation checks that are currently in place. For instance if the name is empty or if undeclared params are allowed(this is valid in case the modified list contains a param which has not been declared before). 

Hence the current API design calls the` pre set param` callbacks in the beginning of `set_parameters_atomically` function and not in `__set_parameters_atomically_common`. https://github.com/ros2/rclcpp/blob/f0dd1679017c2e9caecc11dc63186cc8171e96e9/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L665-L676

For the `post set parameter `callback it does make sense to have them inside `__set_parameters_atomically_common` since once the params are set successfully we can call the `post set params` callback there only. https://github.com/ros2/rclcpp/blob/f0dd1679017c2e9caecc11dc63186cc8171e96e9/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L381-L382

Regarding whether we should support pre and post callbacks for declare parameters, I guess having post callback makes sense for declare parameter(and the current changes does that) but I am not sure about pre set parameter callback for declare parameter. Since the pre set parameter callback is designed to make changes to list of params and `declare_parameter` API declares + validates a single param as opposed to `set_parameters_atomically` which can validate and set multiple params.

jacobperron:
> I guess having post callback makes sense for declare parameter but I am not sure about pre set parameter callback for declare parameter

Sounds good to me :+1:

deepanshubansal01:
The PR is ready for review @jacobperron @adityapande-1995 @ivanpauno now. Thanks

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
+# Proposed node parameters callback Design
+
+## Introduction:
+
+The original requirement came in **gazeb_ros_pkgs** for setting individual wheel slip params based on global wheel slip value [link to original issue](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1365).
```
jacobperron, line 5:
```suggestion
The original requirement came in **gazebo_ros_pkgs** for setting individual wheel slip parameters based on global wheel slip value [link to original issue](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1365). 
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
+# Proposed node parameters callback Design
+
+## Introduction:
+
+The original requirement came in **gazeb_ros_pkgs** for setting individual wheel slip params based on global wheel slip value [link to original issue](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1365). 
+The main idea was to set some parameters using `set_parameter` API once a given param was set successfully.
```
jacobperron, line 6:
```suggestion
The main requirement is to set one or more parameters after another parameter is set successfully. 
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
+# Proposed node parameters callback Design
+
+## Introduction:
+
+The original requirement came in **gazeb_ros_pkgs** for setting individual wheel slip params based on global wheel slip value [link to original issue](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1365). 
+The main idea was to set some parameters using `set_parameter` API once a given param was set successfully. 
+The requirement let to the discussion of supporting registered callbacks once the parameters are set and have been requested by users before.
```
jacobperron, line 7:
```suggestion
Additionally, it would be nice if users could be notified locally (via a callback) when parameters have been set successfully (i.e. post validation).
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+
+In the current Node API, the `add_on_set_parameters_callback` is used for doing any validation for parameters before setting the parameters successfully and once the parameters are validated `ParameterEventHandler` object is used to publish any
```
jacobperron, line 9:
```suggestion
With the current parameters API, the `add_on_set_parameters_callback` is intended for validation of parameter values before they are set, it should **not** cause any side-effects.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
+# Proposed node parameters callback Design
+
+## Introduction:
+
+The original requirement came in **gazeb_ros_pkgs** for setting individual wheel slip params based on global wheel slip value [link to original issue](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1365). 
+The main idea was to set some parameters using `set_parameter` API once a given param was set successfully. 
+The requirement let to the discussion of supporting registered callbacks once the parameters are set and have been requested by users before.
+
+In the current Node API, the `add_on_set_parameters_callback` is used for doing any validation for parameters before setting the parameters successfully and once the parameters are validated `ParameterEventHandler` object is used to publish any 
+changes to node parameters on **/parameter_events** topic, which can be subscribed by any node to see the changed parameter(Note that the **/parameter_events** is a topic for all nodes on the network, and we have to rely on executors to process those).
```
jacobperron, line 10:
```suggestion
There is also the `ParameterEventHandler` that publishes changes to node parameters on `/parameter_events` topic for external nodes to see.
Though the node could subscribe to the `/parameter_events` topic to be notified of changes to its own parameters, it is less than ideal since there is a delay caused by waiting for an executor to process the callback.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+
+We propose adding a `PostSetParametersCallbackHandle` for successful parameter set similar to OnSetParametersCallbackHandle for parameter validation,
```
jacobperron, line 12:
```suggestion
We propose adding a `PostSetParametersCallbackHandle` for successful parameter set similar to `OnSetParametersCallbackHandle` for parameter validation.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+The main idea was to set some parameters using `set_parameter` API once a given param was set successfully. 
+The requirement let to the discussion of supporting registered callbacks once the parameters are set and have been requested by users before.
+
+In the current Node API, the `add_on_set_parameters_callback` is used for doing any validation for parameters before setting the parameters successfully and once the parameters are validated `ParameterEventHandler` object is used to publish any 
+changes to node parameters on **/parameter_events** topic, which can be subscribed by any node to see the changed parameter(Note that the **/parameter_events** is a topic for all nodes on the network, and we have to rely on executors to process those).
+
+We propose adding a `PostSetParametersCallbackHandle` for successful parameter set similar to OnSetParametersCallbackHandle for parameter validation, 
+additionally we propose adding a `PreSetParametersCallbackHandle` which can be used to modify the parameters being set
```
jacobperron, line 13:
```suggestion
Additionally, we propose adding a `PreSetParametersCallbackHandle` useful for modifying the list of parameters being set.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+
+Some related discussion here [#609](https://github.com/ros2/rclcpp/issues/609) [#1789](https://github.com/ros2/rclcpp/pull/1789).
```
jacobperron, line 15:
```suggestion
Related discussion can be found in [#609](https://github.com/ros2/rclcpp/issues/609) and [#1789](https://github.com/ros2/rclcpp/pull/1789).
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Proposed Architecture
+
+The validation callback is often abused to trigger side effects in the code, for instance updating class attributes even before parameter has been set successfully.
```
jacobperron, line 19:
```suggestion
The validation callback is often abused to trigger side effects in the code, for instance updating class attributes even before a parameter has been set successfully. 
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+We propose adding a `PostSetParametersCallbackHandle` for successful parameter set similar to OnSetParametersCallbackHandle for parameter validation, 
+additionally we propose adding a `PreSetParametersCallbackHandle` which can be used to modify the parameters being set
+
+Some related discussion here [#609](https://github.com/ros2/rclcpp/issues/609) [#1789](https://github.com/ros2/rclcpp/pull/1789).
+
+## Proposed Architecture
+
+The validation callback is often abused to trigger side effects in the code, for instance updating class attributes even before parameter has been set successfully. 
+Instead of relying on **/parameter_events** it would be good to support an internal node interface API for registering callbacks for successful parameter set 
+`add_post_set_parameters_callback`.
```
jacobperron, line 21:
```suggestion
Instead of relying on the `/parameter_events` topic to be notified of parameter changes, users can register a callback with a new API, `add_post_set_parameters_callback`.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+
+We can use the proposed `add_post_set_parameters_callback` for setting some param but this might result in unpredictable behaviour because of infinite recursion
```
jacobperron, line 23:
```suggestion
It is possible to use the proposed `add_post_set_parameters_callback` for setting additional parameters, but this might result in infinite recursion and does not allow those additional parameters to be set atomically with the original parameter(s) changed.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Proposed Architecture
+
+The validation callback is often abused to trigger side effects in the code, for instance updating class attributes even before parameter has been set successfully. 
+Instead of relying on **/parameter_events** it would be good to support an internal node interface API for registering callbacks for successful parameter set 
+`add_post_set_parameters_callback`.
+
+We can use the proposed `add_post_set_parameters_callback` for setting some param but this might result in unpredictable behaviour because of infinite recursion 
+and due to the params being set atomically. In order to get around this we propose adding another registered callback `add_pre_set_parameters_callback` which 
+will be triggered before the validation callbacks and can be used to modify the parameter list as required.
```
jacobperron, line 25:
```suggestion
To workaround these issues, we propose adding a "pre set" callback type that can be registered with `add_pre_set_parameters_callback`, which will be triggered before the validation callbacks and can be used to modify the parameter list.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+
+![Desgin API](https://github.com/ros2/rclcpp/blob/deepanshu/local-param-changed-callback-support/rclcpp/doc/param_callback_design.png?raw=true)
+
+## Alternatives
+
+* The **/parameter_events** topic can be used to monitor any changes to node parameters and use ```set_parameter``` API if required to set any further parameters. However, it seems weird that we have to rely on network to notify nodes of their own parameter changes. Therefore, it makes sense to support a callback which gets triggered after the parameter is changed, so that it can trigger some more events.
```
jacobperron, line 31:
```suggestion
* Users could call `set_parameter` while processing a message from the `/parameter_events` topic, However, there is extra overhead in having to create subscription (as noted earlier).
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+* The ```add_on_set_parameters_callback``` can be used to ```set_paramters```, however the ```ParameterMutationRecursionGuard``` will not allow this since the ```add_on_set_parameters_callback``` is supposed to be used for validation purposes only.
```
jacobperron, line 32: I don't think the recursion guard is a strong reason to avoid this approach, since it is in our control to remove or patch the recursion guard to allow for some recursion depth before failing.
```suggestion
* Users could call `set_parameter` inside the "on set" parameters callback, however it is not well-defined how side-effects should handle cases where parameter validation fails.
```

rclcpp/doc/proposed_node_parameter_callbacks.md
```diff
@@ -0,0 +1,33 @@
...
+* The ```add_post_set_parameters_callback```  can be used to set parameters but this might result in different behaviours depending on whether the params were set atomically or not. Also, we might enter an infinite recursion depending on how params are being set.
```
jacobperron, line 33: Following my previous comment, I wouldn't rule out this use-case. It's possible we could relax the recursion guard in the future.
```suggestion
```

jacobperron (commented):
I've left wording suggestions for for the design doc.
I'm reviewing the code next.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -305,18 +305,53 @@ __check_parameters(
...
+using PostSetParametersCallbackType =
+    rclcpp::node_interfaces::NodeParametersInterface::PostSetParametersCallbackType;
+using PostSetParametersCallbackHandle =
+    rclcpp::node_interfaces::PostSetParametersCallbackHandle;
+using PostSetCallbacksHandleContainer =
+    rclcpp::node_interfaces::NodeParameters::PostSetCallbacksHandleContainer;
+
+RCLCPP_LOCAL
+void
+__call_pre_set_parameters_callbacks(
```
jacobperron, line 331: I think a double underscore prefix is technically reserved by the standard for compilers. I would simply remove the underscore prefix.

I think this function can be made `static` too.

I realize this is just following the convention of existing functions in this file, so I'm okay with leaving it as-is for now.

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -305,18 +305,53 @@ __check_parameters(
...
+    std::vector<rclcpp::Parameter> & parameters,
+    PreSetCallbacksHandleContainer & callback_container)
+{
+  if(!callback_container.empty()){
```
jacobperron, line 335: IMO, returning early makes the code more readable as it avoids nested indentation.
```suggestion
  if (callback_container.empty()) {
    return;
  }
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -441,11 +503,13 @@ declare_parameter_helper(
...
   std::map<std::string, rclcpp::node_interfaces::ParameterInfo> & parameters,
   const std::map<std::string, rclcpp::ParameterValue> & overrides,
-  CallbacksContainerType & callback_container,
+  OnSetCallbacksHandleContainer & on_set_callback_container,
+  PostSetCallbacksHandleContainer & post_set_callback_container,
   rclcpp::Publisher<rcl_interfaces::msg::ParameterEvent> * events_publisher,
   const std::string & combined_name,
   rclcpp::node_interfaces::NodeClockInterface & node_clock)
 {
+
```
jacobperron, line 512:
```suggestion
```

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
```diff
@@ -633,12 +700,27 @@ NodeParameters::set_parameters_atomically(const std::vector<rclcpp::Parameter> &
...
+  if (parameters_after_pre_set_callback.empty())
+  {
+    result.successful = false;
+    result.reason = "parameter list cannot be empty, this might be due to "
+                    "pre_set_parameters_callback modifying the original parameters list.";
+    return result;
+  }
+
+  // Check if any of the parameters are read-only,empty_post_set_callback_container or if any parameters are not
```
jacobperron, line 718: A typo I think
```suggestion
  // Check if any of the parameters are read-only, or if any parameters are not
```

rclcpp/test/rclcpp/node_interfaces/test_node_parameters.cpp
```diff
@@ -197,5 +197,124 @@ TEST_F(TestNodeParameters, add_remove_parameters_callback) {
...
+
+  rcl_interfaces::msg::ParameterDescriptor param2_descriptor;
+  param1_descriptor.name = "param2";
+  param1_descriptor.type = rcl_interfaces::msg::ParameterType::PARAMETER_DOUBLE;
+  param1_descriptor.read_only = false;
+
+  node_parameters->declare_parameter(
+      "param1", rclcpp::ParameterValue(0.0), param1_descriptor, false);
+  node_parameters->declare_parameter(
+      "param2", rclcpp::ParameterValue(0.0), param2_descriptor, false);
```
jacobperron, line 217: We could also test that declaring a parameter does not trigger the pre-set callbacks by moving these after registering a the first callback and checking that their values are unchanged.

rclcpp/test/rclcpp/node_interfaces/test_node_parameters.cpp
```diff
@@ -197,5 +197,124 @@ TEST_F(TestNodeParameters, add_remove_parameters_callback) {
...
+  param2_descriptor.type = rcl_interfaces::msg::ParameterType::PARAMETER_DOUBLE;
+  param2_descriptor.read_only = false;
+
+  node_parameters->declare_parameter(
+      "param1", rclcpp::ParameterValue(0.0),param1_descriptor, false);
+  node_parameters->declare_parameter(
+      "param2", rclcpp::ParameterValue(0.0),param2_descriptor, false);
+
+  double variable_tracking_param1_internally = node_parameters->get_parameter("param1").get_value<double>();
+  double variable_tracking_param2_internally = node_parameters->get_parameter("param2").get_value<double>();
```
jacobperron, line 286: I think the following is equivalent:
```suggestion
   double variable_tracking_param1_internally = node_parameters->declare_parameter(
      "param1", rclcpp::ParameterValue(0.0),param1_descriptor, false);
  double variable_tracking_param2_internally = node_parameters->declare_parameter(
      "param2", rclcpp::ParameterValue(0.0),param2_descriptor, false);
```

jacobperron: However, it would also be nice to test that declaring parameters also triggers the callback.

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -1444,6 +1444,38 @@ TEST_F(TestNode, set_parameter_undeclared_parameters_not_allowed) {
...
+    // parameter has not been declared before will throw
+    auto name1 = "parameter"_unq;
+    auto name2 = "parameter"_unq;
+    auto default_value = 0;  // default value of name1 param
+
+    rcl_interfaces::msg::ParameterDescriptor descriptor;
+    descriptor.type = rclcpp::PARAMETER_INTEGER;
+
+    // declare name1 parameter only
+    node->declare_parameter(name1, default_value, descriptor);
```
jacobperron, line 1458: I think the descriptor is optional (the fact that it's an integer can be inferred by the default value).
```suggestion
    // declare name1 parameter only
    node->declare_parameter(name1, default_value);
```

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -1481,6 +1513,39 @@ TEST_F(TestNode, set_parameter_undeclared_parameters_allowed) {
...
+    // parameter has not been declared will not throw if undeclared
+    // parameters are allowed
+    auto name1 = "parameter"_unq;
+    auto name2 = "parameter"_unq;
+
+    rcl_interfaces::msg::ParameterDescriptor descriptor;
+    descriptor.type = rclcpp::PARAMETER_INTEGER;
+
+    // declare name1 parameter only
+    node->declare_parameter(name1, 0, descriptor);
```
jacobperron, line 1527:
```suggestion
    // declare name1 parameter only
    node->declare_parameter(name1, 0);
```

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -1625,6 +1690,50 @@ TEST_F(TestNode, set_parameters_undeclared_parameters_not_allowed) {
...
+    auto name2 = "parameter"_unq;
+    auto name3 = "parameter"_unq;
+    auto default_value = 0;
+
+    rcl_interfaces::msg::ParameterDescriptor descriptor;
+    descriptor.type = rclcpp::PARAMETER_INTEGER;
+
+    // declare name1 and name2 parameter only
+    node->declare_parameter(name1, default_value, descriptor);
+    node->declare_parameter(name2, default_value, descriptor);
```
jacobperron, line 1710:
```suggestion
    // declare name1 and name2 parameter only
    node->declare_parameter(name1, default_value);
    node->declare_parameter(name2, default_value);
```

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -1673,6 +1782,48 @@ TEST_F(TestNode, set_parameters_undeclared_parameters_allowed) {
...
```
jacobperron, line 1799:
```suggestion
    // declare name1 and name2 parameter only
    node->declare_parameter(name1, default_value);
    node->declare_parameter(name2, default_value);
```

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -1815,6 +1966,49 @@ TEST_F(TestNode, set_parameters_atomically_undeclared_parameters_not_allowed) {
...
```
jacobperron, line 1985:
```suggestion
    // declare name1 and name2 parameter only
    node->declare_parameter(name1, default_value);
    node->declare_parameter(name2, default_value);
```

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -1815,6 +1966,49 @@ TEST_F(TestNode, set_parameters_atomically_undeclared_parameters_not_allowed) {
...
+          for(const auto&param: parameters){
+            if(param.get_name() == name2){
+              parameters.emplace_back(rclcpp::Parameter(name3, 3));
+            }
+          }
+        };
+
+    auto handler = node->add_pre_set_parameters_callback(pre_set_parameters);
+    EXPECT_THROW(node->set_parameters_atomically({rclcpp::Parameter(name1, 1),
+                                                            rclcpp::Parameter(name2, 2)}),
```
jacobperron, line 2000: indentation looks off here.

rclcpp/test/rclcpp/test_node.cpp
```diff
@@ -1903,6 +2097,47 @@ TEST_F(TestNode, set_parameters_atomically_undeclared_parameters_allowed) {
...
+    auto name2 = "parameter"_unq;
+    auto name3 = "parameter"_unq;
+    auto default_value = 0;
+
+    rcl_interfaces::msg::ParameterDescriptor descriptor;
+    descriptor.type = rclcpp::PARAMETER_INTEGER;
+
+    // declare name1 and name2 parameter only
+    node->declare_parameter(name1, default_value, descriptor);
+    node->declare_parameter(name2, default_value, descriptor);
```
jacobperron, line 2114:
```suggestion
    // declare name1 and name2 parameter only
    node->declare_parameter(name1, default_value);
    node->declare_parameter(name2, default_value);
```

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -510,8 +510,8 @@ class LifecycleNode : public node_interfaces::LifecycleNodeInterface,
 
   using OnSetParametersCallbackHandle =
     rclcpp::node_interfaces::OnSetParametersCallbackHandle;
-  using OnParametersSetCallbackType =
```
jacobperron, line 513: deprecated the old name

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -521,7 +521,7 @@ class LifecycleNode : public node_interfaces::LifecycleNodeInterface,
   RCUTILS_WARN_UNUSED
   rclcpp_lifecycle::LifecycleNode::OnSetParametersCallbackHandle::SharedPtr
   add_on_set_parameters_callback(
```
jacobperron, line 523: I think it might be worth exposing the methods for adding/removing the new pre and post callbacks to the lifecycle node as well.

jacobperron (commented):
The code looks good to me. My comments are mainly suggestions in the docs.

I think there are a lot of linter issues to resolve too.

deepanshubansal01:
> The code looks good to me. My comments are mainly suggestions in the docs.
> 
> I think there are a lot of linter issues to resolve too.

Thanks a lot for the review. I will look into the suggestions that you mentioned.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
deepanshubansal01: yeah, make sense. Thanks!

rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
deepanshubansal01: makes sense.

rclcpp/test/rclcpp/node_interfaces/test_node_parameters.cpp
deepanshubansal01: Done

deepanshubansal01:
I have made changes to the code based on review suggestions, I have also updated documentation for existing Node API's based on pre and post callbacks. I can use a review on that specific commit https://github.com/ros2/rclcpp/pull/1947/commits/d0bc26b6e869c285d71a6322e4c117ca1f40b6a8

Thanks!

jacobperron (approved):
A few aesthetic things, otherwise LGTM!

adityapande-1995:
CI :

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16979/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11544/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17386/)

jacobperron:
The same Windows test failure occurred on the most recent nightly: https://ci.ros2.org/view/nightly/job/nightly_win_rep/lastBuild/testReport/junit/launch_testing_examples.launch_testing_examples/check_multiple_nodes_launch_test/launch_testing_examples_check_multiple_nodes_launch_test/

jacobperron (approved):

adityapande-1995 (approved):
Looks good !

PR #1950, opened 2022-06-07
Fix subscription.is_serialized() for callbacks with message info

ivanpauno:
I think this was broken when support for type adaption was added.
Currently if you create a serialized subscription with a callback taking both the message and the message info, the executor throws an exception the first time a message is dispatched.

This fixes the issue.

### DIFF
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -954,7 +954,8 @@ public:
       std::holds_alternative<ConstRefSerializedMessageWithInfoCallback>(callback_variant_) ||
       std::holds_alternative<UniquePtrSerializedMessageWithInfoCallback>(callback_variant_) ||
       std::holds_alternative<SharedConstPtrSerializedMessageWithInfoCallback>(callback_variant_) ||
-      std::holds_alternative<ConstRefSharedConstPtrSerializedMessageWithInfoCallback>(callback_variant_) ||
+      std::holds_alternative<ConstRefSharedConstPtrSerializedMessageWithInfoCallback>(
+      callback_variant_) ||
       std::holds_alternative<SharedPtrSerializedMessageWithInfoCallback>(callback_variant_);
   }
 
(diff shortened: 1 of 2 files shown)


clalancette:
Yikes, we probably didn't notice because there wasn't a test.

Can you add tests that exercise this?

fujitatomoya (approved):

ivanpauno:
> Can you add tests that exercise this?

I have added some tests for `AnySubscriptionCallback` covering this case.
See https://github.com/ros2/rclcpp/pull/1950/commits/199d36b1726d3b8db33caa9798e25346e6a01388.

I would also like to have a test with a publisher and a subscription for each of the callbacks, and make sure the message arrives.
We don't currently have tests like that in `rclcpp`, I think that we added all of them in `test_rclcpp` (`system_tests` repo), so I could add one doing something like that there.

alsora (approved):

clalancette:
> I would also like to have a test with a publisher and a subscription for each of the callbacks, and make sure the message arrives.
> We don't currently have tests like that in `rclcpp`, I think that we added all of them in `test_rclcpp` (`system_tests` repo), so I could add one doing something like that there.

Yeah, we typically have tests like that in `test_rclcpp` or `test_communication`.  I think it would be fine if you wanted to add some tests there, though I would be careful about adding all combinations; I don't want to explode our CI times.

clalancette (approved):
Looks good to me with green CI (I'd be sure to run CI with `--packages-above rclcpp`)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16821/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11399/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17227/)

ivanpauno:
windows issues are unrelated (happening in nightlies as well https://ci.ros2.org/view/nightly/job/nightly_win_rel/2329/msbuild/new/).
Going in!

PR #1951, opened 2022-06-08
test adjustment for LoanedMessage.

fujitatomoya:
depends https://github.com/ros2/rmw_fastrtps/pull/568.
since [that](https://github.com/ros2/rmw_fastrtps/pull/568) will enable rmw_fastrtps zero copy in default, this fix is required to pass the CI.

### DIFF
--- a/rclcpp/test/rclcpp/test_publisher.cpp
+++ b/rclcpp/test/rclcpp/test_publisher.cpp
@@ -25,6 +25,8 @@
 #include "rclcpp/exceptions.hpp"
 #include "rclcpp/rclcpp.hpp"
 
+#include "rcutils/env.h"
+
 #include "../mocking_utils/patch.hpp"
 #include "../utils/rclcpp_gtest_macros.hpp"
 
@@ -413,10 +415,18 @@ TEST_F(TestPublisher, intra_process_publish_failures) {
 
   {
     rclcpp::LoanedMessage<test_msgs::msg::Empty> loaned_msg(*publisher, allocator);
-    loaned_msg.release();
+    auto msg = loaned_msg.release();  // this will unmanage the ownership of the message
     RCLCPP_EXPECT_THROW_EQ(
       publisher->publish(std::move(loaned_msg)),
       std::runtime_error("loaned message is not valid"));
+    // if the message is actually loaned from the middleware but not be published,
+    // it is user responsibility to return the message to the middleware manually
+    if (publisher->can_loan_messages()) {
+      ASSERT_EQ(
+        RCL_RET_OK,
+        rcl_return_loaned_message_from_publisher(
+          publisher->get_publisher_handle().get(), msg.get()));
+    }
   }
   RCLCPP_EXPECT_THROW_EQ(
     node->create_publisher<test_msgs::msg::Empty>(
@@ -485,6 +495,11 @@ public:
 TEST_F(TestPublisher, do_loaned_message_publish_error) {
   initialize();
   using PublisherT = TestPublisherProtectedMethods<test_msgs::msg::Empty, std::allocator<void>>;
+  // This test only passes when message is allocated on heap, not middleware.
+  // Since `do_loaned_message_publish()` will fail, there is no way to return the message
+  // to the middleware.
+  // This eventually fails to destroy publisher handle in the implementation.
+  ASSERT_TRUE(rcutils_set_env("ROS_DISABLE_LOANED_MESSAGES", "1"));
   auto publisher =
     node->create_publisher<test_msgs::msg::Empty, std::allocator<void>, PublisherT>("topic", 10);
 


fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16787/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11368/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17196/)

rclcpp/test/rclcpp/test_publisher.cpp
```diff
@@ -413,10 +415,18 @@ TEST_F(TestPublisher, intra_process_publish_failures) {
...
   {
     rclcpp::LoanedMessage<test_msgs::msg::Empty> loaned_msg(*publisher, allocator);
-    loaned_msg.release();
+    auto msg = loaned_msg.release();  // this will unmanage the ownership of the message
     RCLCPP_EXPECT_THROW_EQ(
       publisher->publish(std::move(loaned_msg)),
       std::runtime_error("loaned message is not valid"));
+    // if the message is actually loaned from the middleware but not be published,
+    // it is user responsibility to return the message to the middleware manually
+    if (publisher->can_loan_messages()) {
```
fujitatomoya, line 424: if this is not issued, rmw_fastrtps Fast-DDS will fail to destroy DDS writer since the allocated memory in Fast-DDS has not returned yet.

rclcpp/test/rclcpp/test_publisher.cpp
```diff
@@ -485,6 +495,11 @@ class TestPublisherProtectedMethods : public rclcpp::Publisher<MessageT, Allocat
 TEST_F(TestPublisher, do_loaned_message_publish_error) {
   initialize();
   using PublisherT = TestPublisherProtectedMethods<test_msgs::msg::Empty, std::allocator<void>>;
+  // This test only passes when message is allocated on heap, not middleware.
+  // Since `do_loaned_message_publish()` will fail, there is no way to return the message
+  // to the middleware.
+  // This eventually fails to destroy publisher handle in the implementation.
+  ASSERT_TRUE(rcutils_set_env("ROS_DISABLE_LOANED_MESSAGES", "1"));
```
fujitatomoya, line 502: just using heap memory instead of middleware memory to test the API, so that middleware can destroy publisher implementation(in this case Fast-DDS DDS writer).

fujitatomoya: yeah, that was basically a way i was thinking about.
maybe no, since we lost the ownership for the data, there is no way to return the memory to the middle-ware.
besides this test case actually does not call rmw implementation.

ivanpauno: > maybe no, since we lost the ownership for the data, there is no way to return the memory to the middle-ware.

If we're transferring the ownership of the borrowed message, like we're doing here, then the implementation should automatically do the cleanup on error.
If not, it's impossible to handle the error case correctly, which is a bug.

We should fix the bug IMO, instead of working it arround in the tests.

fujitatomoya: > then the implementation should

they should. but in this particular test case, implementation cannot be aware of this, since `rcl_publish_loaned_message` is patched here. (implementation is NOT called.)

https://github.com/ros2/rclcpp/blob/8e6a6fb32d8d6a818b483660e326f2c5313b64ae/rclcpp/test/rclcpp/test_publisher.cpp#L496-L497

besides calling `do_loaned_message_publish` with heap memory is not suitable for test case. `do_loaned_message_publish` is technically only issued if `can_loan_messages()` is true.

https://github.com/ros2/rclcpp/blob/8e6a6fb32d8d6a818b483660e326f2c5313b64ae/rclcpp/include/rclcpp/publisher.hpp#L413-L421

but using heap memory to test the particular API would be okay. so using `ROS_DISABLE_LOANED_MESSAGES` to pass the test would be acceptable for test coverage.
Or we could remove this test case entirely?

ivanpauno: > since rcl_publish_loaned_message is patched here

It's not super clear in the documentation though if `rcl_publish_loaned_message()` deallocates the memory when a failure happens or not.
That should be well documented IMO.
Though, it's true that's not related with the test, which I didn't notice before.

fujitatomoya: > couldn't you use [patch()](https://github.com/ros2/rclcpp/blob/8e6a6fb32d8d6a818b483660e326f2c5313b64ae/rclcpp/test/mocking_utils/patch.hpp#L532) instead of patch_and_return() to fix the issue?

we can do that, but it needs to have `if` statement to call middleware or not probably using `can_loan_message`. `rcl_publish_loaned_message` is not supposed to be called w/o `can_loan_message` is true.

I think using `ROS_DISABLE_LOANED_MESSAGES` is more suitable for this API test.

ivanpauno (approved):

fujitatomoya:
@ivanpauno thanks for the review, i am going to merge this into master. this does not affect the mainline, but required before https://github.com/ros2/rmw_fastrtps/pull/568

PR #1956, opened 2022-06-24
Revert "Add support for spin_until_timeout (#1821)"

wjwwood:
Reverts ros2/rclcpp#1874

See the reason here: https://github.com/ros2/rclcpp/pull/1874#issuecomment-1165811614


clalancette (approved):
This looks good to me once the Rpr job comes back green.

I'm happy with just the Rpr job because nothing else was merged since #1951 a couple of days ago, so we have pretty good confidence in that.

The only other thing I'll request when merging this is that we open a "Revert Revert" PR so we can get that ready to go once the dependent PRs are ready.

Issue #1968, opened 2022-07-09
shared_futures set by spin_until_future_complete invalid after first get()


fmrico:
## Bug report

**Required Info:**

- Operating System: Ubuntu 22.04
  
- Installation type: binaries

- Version or commit hash: Humble
 
- DDS implementation: Fast DDS

- Client library (if applicable): rclcpp

#### Steps to reproduce issue

In Galaxy, code like this was correct
``` 
  auto request = std::make_shared<plansys2_msgs::srv::GetPlan::Request>();
  request->domain = domain;
  request->problem = problem;

  auto future_result = get_plan_client_->async_send_request(request);

  if (rclcpp::spin_until_future_complete(node_, future_result, std::chrono::seconds(15)) !=
    rclcpp::FutureReturnCode::SUCCESS)
  {
    return {};
  }

  if (future_result.get()->success) {
    return future_result.get()->plan;
```

But in Humble it seems like the data pointed by the returned `shared_future` becomes invalid as soon as `get()` is called for the first time, producing segmentation faults in  subsequent calls to `get()`,  fixing it in this way:

``` 
  auto request = std::make_shared<plansys2_msgs::srv::GetPlan::Request>();
  request->domain = domain;
  request->problem = problem;

  auto future_result = get_plan_client_->async_send_request(request);

  if (rclcpp::spin_until_future_complete(node_, future_result, std::chrono::seconds(15)) !=
    rclcpp::FutureReturnCode::SUCCESS)
  {
    return {};
  }

  auto result = *future_result.get();

  if (result.success) {
    return result.plan;
```

fujitatomoya:
@fmrico i am not sure what is the problem here, could you elaborate a bit?

https://github.com/fujitatomoya/ros2_test_prover/blob/master/prover_rclcpp/src/rclcpp_1968.cpp works w/o any error with https://github.com/ros2/ros2/commit/28a83a843324918293ccd0d106117cf2fa4e33a6

```bash
root@tomoyafujita:~/ros2_ws/colcon_ws# ros2 run prover_rclcpp rclcpp_1968
Client: async_send_request call !!!
[INFO] [1657592176.287793495] [client]: Result = 3
[INFO] [1657592176.287849726] [client]: Result = 3
root@tomoyafujita:~/ros2_ws/colcon_ws# ros2 run prover_rclcpp rclcpp_1968
Client: async_send_request call !!!
[INFO] [1657592184.355723904] [client]: Result = 3
[...]

PR #1974, opened 2022-07-14
replace future with shared_future

iuhilnehc-ynos:
#1968 
It's not mandatory. I would like to keep consistent with the other two methods.

Some other repositories are as follows,

https://github.com/ros2/demos/pull/569
https://gitlab.com/ros-tracing/ros2_tracing/-/merge_requests/298

### DIFF
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -578,7 +578,7 @@ public:
 
   /// Send a request to the service server.
   /**
-   * This method returns a `FutureAndRequestId` instance
+   * This method returns a `SharedFutureAndRequestId` instance
    * that can be passed to Executor::spin_until_future_complete() to
    * wait until it has been completed.
    *
@@ -602,17 +602,17 @@ public:
    * ```
    *
    * \param[in] request request to be send.
-   * \return a FutureAndRequestId instance.
+   * \return a SharedFutureAndRequestId instance.
    */
-  FutureAndRequestId
+  SharedFutureAndRequestId
   async_send_request(SharedRequest request)
   {
     Promise promise;
-    auto future = promise.get_future();
+    auto future = promise.get_future().share();
     auto req_id = async_send_request_impl(
       *request,
       std::move(promise));
-    return FutureAndRequestId(std::move(future), req_id);
+    return SharedFutureAndRequestId(std::move(future), req_id);
   }
 
   /// Send a request to the service server and schedule a callback in the executor.


iuhilnehc-ynos:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16973/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11536/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17376/)

iuhilnehc-ynos:
Update [[demos]](https://github.com/ros2/demos/pull/569) project. Rerun CI,

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16974/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17377/)

iuhilnehc-ynos:
It seems there still exist other related repositories that need to be updated.
https://gitlab.com/ros-tracing/ros2_tracing
https://github.com/ros2/system_tests.git

iuhilnehc-ynos:
For me, I prefer to use the updated source code.

```patch
     // We send the service request for asking the current
     // state of the lc_talker node.
-    auto future_result = client_get_state_->async_send_request(request).future.share();
+    auto future_result = client_get_state_->async_send_request(request);

    // Let's wait until we have the answer from the node.
    // If the request times out, we return an unknown state.
    auto future_status = wait_for_result(future_result, time_out);

``` 

So I'll continue to update other repositories.

iuhilnehc-ynos:
Updated [ros2_tracing](https://gitlab.com/ros-tracing/ros2_tracing/-/merge_requests/298), run single CI in `Linux-aarch64` successfully

* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11539/)

And then re-run CI again,

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16975/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11540/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17378/)

ivanpauno:
I don't think this is a good idea.

Consistency doesn't really matter in this case IMO, because you never use the return value in the other cases.
Thye pass a callback as an argument, you really don't need to the returned future (for some reason, we defined the methods like that at some point). And the only reason why that was originally a `shared_future` is because the callbacks receive them as an argument as well https://github.com/ros2/rclcpp/blob/621586ecb0244a6cd60dbeb50adebe7e5206765f/rclcpp/include/rclcpp/client.hpp#L390-L391 

That also doesn't make sense, as they should be getting a reference/pointer to the response directly (there's nothing to be waited when the callback is called, so no need to receive the future as an argument).

IMO, we should clean up the other signatures, instead of modifying the one that seems to make sense for consitency.

ivanpauno (changes requested):
see above comment

iuhilnehc-ynos:
As the requested issue is closed, I'll close this PR.
This PR is not going to `clean up the other signatures`.

Issue #1259, opened 2020-08-04
[api review] Calling Syntax and Keeping Node-like Class APIs in Sync


wjwwood:
This is a result of the API review done in March 2020:

https://github.com/ros2/rclcpp/blob/master/rclcpp/doc/api_review_march_2020.md#calling-syntax-and-keeping-node-like-class-apis-in-sync

----

### Summary of the Issue

Currently, much of the API is exposed via the `rclcpp::Node` class, and due to the nature of the current architecture there is a lot of repeated code to expose these methods and then call the implementations which are in other classes like `rclcpp::node_interfaces::NodeTopics`, for example.

Also, we have other versions of the class `rclcpp::Node` with different semantics and interfaces, like `rclcpp_lifecycle::LifecycleNode`, and we have been having trouble keeping the interface provided there up to date with how things are done in `rclcpp::Node`. Since `LifecycleNode` has a different API from `Node` in some important cases, it does not just inherit from `Node`.

There are two main proposals (as I see it) to try and address this issue, either (a) break up the functionality in `Node` so that it is in separate classes and make `Node` multiple inherit from those classes, and then `LifecycleNode` could selectively inherit from those as well, or (b) change our calling convention from `node->do_thing(...)` to be `do_thing(node, ...)`.

For (a) which commonly referred to as the [Policy Based Design Pattern](https://en.wikipedia.org/wiki/Modern_C%2B%2B_Design#Policy-based_design), we'd be reversing previous design decisions which we discussed at length where we decided to use composition over inheritance for various reasons.
One of the reasons was testing, with the theory that having simpler separate interfaces we could more easily mock them as needed for testing.
The testing goal would still be met, either by keeping the "node_interface" classes as-is or by mocking the classes that node would multiple inherit from, however it's harder to indicate that a function needs a class that multiple inherits from several classes but not others.
Also having interdependency between the classes which are inherited from is a bit complicated in this design pattern.

For (b), we would be changing how we recommend all code be written (not a trivial thing to do at all), because example code like `auto pub = node->create_publsiher(...);` would be come `auto pub = create_publisher(node, ...);`.
This has some distinct advantages, however, in that it allows us to write these functions, like `create_publisher(node, ...)`, so that the node argument can be any class that meets the criteria of the function.
[...]

Issue #465, opened 2018-04-26
Extend Time implementation into Timer and Rate objects


tfoote:
The Timer and Rate need to use a clock, with a coupled time source to accurately sleep.

 *  [ ] Remove internal Rate sleep method, and externalize it with a valid clock or pass in the clock.
 *  [ ] Provide a sleep method for the Timer class.

vinnamkim:
@tfoote Is there any progress on this issue? or Anyone who working on this issue?

In my opinion, to solve this issue, it seems to implement `clock.sleep_for()` and `clock.sleep_until()` such as https://github.com/ros2/ros2/issues/530#issue-339636054 and https://design.ros2.org/articles/clock_and_time.html#public-api. 

To implement clock.sleep, [Clock](https://github.com/ros2/rclcpp/blob/c51b28420f354b890f3605fcf33651d0b3598a45/rclcpp/src/rclcpp/clock.cpp#L41) would be stuck in a while loop until the target time is coming. In the case of the simulation time, [Clock](https://github.com/ros2/rclcpp/blob/c51b28420f354b890f3605fcf33651d0b3598a45/rclcpp/src/rclcpp/clock.cpp#L41) doesn't subscribe [clock_msg](https://github.com/ros2/rclcpp/blob/c93beb5d164162fd0806641c7032b834b37de935/rclcpp/src/rclcpp/time_source.cpp#L178) by itself. The solution would be the followings.

1. [TimeSource](https://github.com/ros2/rclcpp/blob/c93beb5d164162fd0806641c7032b834b37de935/rclcpp/src/rclcpp/time_source.cpp#L35) class has it's own thread to spin clock_msg callback function likewise [ROS1](https://github.com/ros/ros_comm/blob/6b9efd56d6882d1c017152ba11a5780a1496b30b/clients/roscpp/src/libros/init.cpp#L397)
[...]

PR #1975, opened 2022-07-22
Add a `create_timer` method to `Node` and `LifecycleNode` classes

asymingt:
## Problem
Right now the `Node` and `LifecycleNode` classes only provide a method for creating a wall timer, even though the underlying functions for creating a generic timer (which respects ROS time via `/clock`) are available. This is a problem because those that write node implementations -- and especially those that come from a ROS1 background -- default to using `{Node, LifecycleNode}::create_wall_timer` not understanding the implications this decision has on timer behavior in replay, and especially when `ros2 bag play --rate` is used. The existing method of creating the correct timer type is opaque: it requires knowledge of free-functions in `rclcpp`.

Relates to issues:
- https://github.com/ros2/rclcpp/issues/465
- https://github.com/ros2/rclcpp/issues/1259

Relates to pull requests:
- https://github.com/ros2/rclcpp/pull/1982

## Fix
This PR adds the piping to allow a timer that respects `/clock` when `use_sim_time: True` node parameter is set. While it is possible to do this manually with `rclcpp::create_timer` it's much less convenient. An analogous function [already exists](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/node.py#L1485-L1514) in `rclpy`.

## Progress
- [x] Add the `create_timer` method to `Node` class.
- [x] Take care of edge cases in `chrono::duration` casts.
- [x] Add unit tests to `create_timer`, `timer_source` and `timer`.
- [x] Debug `TestTimeSource.check_sim_time_updated_in_callback_if_use_clock_thread`.
- [x] Update`LifecycleNode` to also support `create_timer`.
- [x] Refactor `create_timer` to drop raw pointers and have common error checks.

### DIFF
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -23,6 +23,7 @@
 
 #include "rclcpp/duration.hpp"
 #include "rclcpp/node_interfaces/get_node_base_interface.hpp"
+#include "rclcpp/node_interfaces/get_node_clock_interface.hpp"
 #include "rclcpp/node_interfaces/get_node_timers_interface.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/node_interfaces/node_clock_interface.hpp"
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -31,47 +31,8 @@
 
 namespace rclcpp
 {
-/// Create a timer with a given clock
-/// \internal
-template<typename CallbackT>
-typename rclcpp::TimerBase::SharedPtr
-create_timer(
-  std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
-  std::shared_ptr<node_interfaces::NodeTimersInterface> node_timers,
-  rclcpp::Clock::SharedPtr clock,
-  rclcpp::Duration period,
-  CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+namespace detail
 {
-  auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
-    clock,
-    period.to_chrono<std::chrono::nanoseconds>(),
-    std::forward<CallbackT>(callback),
-    node_base->get_context());
-
-  node_timers->add_timer(timer, group);
-  return timer;
-}
-
-/// Create a timer with a given clock
-template<typename NodeT, typename CallbackT>
-typename rclcpp::TimerBase::SharedPtr
-create_timer(
-  NodeT node,
-  rclcpp::Clock::SharedPtr clock,
-  rclcpp::Duration period,
-  CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
-{
-  return create_timer(
-    rclcpp::node_interfaces::get_node_base_interface(node),
-    rclcpp::node_interfaces::get_node_timers_interface(node),
-    clock,
-    period,
-    std::forward<CallbackT>(callback),
-    group);
-}
-
 /// Perform a safe cast to a timer period in nanoseconds
 /**
  *
@@ -117,6 +78,92 @@ safe_cast_to_period_in_ns(std::chrono::duration<DurationRepT, DurationT> period)
 
   return period_ns;
 }
+}
+
+/// Create a timer with a given clock
+/// \internal
+template<typename CallbackT>
+typename rclcpp::TimerBase::SharedPtr
+create_timer(
+  std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
+  std::shared_ptr<node_interfaces::NodeTimersInterface> node_timers,
+  rclcpp::Clock::SharedPtr clock,
+  rclcpp::Duration period,
+  CallbackT && callback,
+  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+{
+  return create_timer(
+    node_base.get(),
+    node_timers.get(),
+    clock,
+    period.to_chrono<std::chrono::nanoseconds>(),
+    std::forward<CallbackT>(callback),
+    group);
+}
+
+/// Create a timer with a given clock
+template<typename NodeT, typename CallbackT>
+typename rclcpp::TimerBase::SharedPtr
+create_timer(
+  NodeT node,
+  rclcpp::Clock::SharedPtr clock,
+  rclcpp::Duration period,
+  CallbackT && callback,
+  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+{
+  return create_timer(
+    clock,
+    period.to_chrono<std::chrono::nanoseconds>(),
+    std::forward<CallbackT>(callback),
+    group,
+    rclcpp::node_interfaces::get_node_base_interface(node).get(),
+    rclcpp::node_interfaces::get_node_timers_interface(node).get());
+}
+
+/// Convenience method to create a general timer with node resources.
+/**
+ *
+ * \tparam DurationRepT
+ * \tparam DurationT
+ * \tparam CallbackT
+ * \param clock clock to be used
+ * \param period period to execute callback. This duration must be 0 <= period < nanoseconds::max()
+ * \param callback callback to execute via the timer period
+ * \param group callback group
+ * \param node_base node base interface
+ * \param node_timers node timer interface
+ * \return shared pointer to a generic timer
+ * \throws std::invalid_argument if either clock, node_base or node_timers
+ * are nullptr, or period is negative or too large
+ */
+template<typename DurationRepT, typename DurationT, typename CallbackT>
+typename rclcpp::GenericTimer<CallbackT>::SharedPtr
+create_timer(
+  rclcpp::Clock::SharedPtr clock,
+  std::chrono::duration<DurationRepT, DurationT> period,
+  CallbackT callback,
+  rclcpp::CallbackGroup::SharedPtr group,
+  node_interfaces::NodeBaseInterface * node_base,
+  node_interfaces::NodeTimersInterface * node_timers)
+{
+  if (clock == nullptr) {
+    throw std::invalid_argument{"clock cannot be null"};
+  }
+  if (node_base == nullptr) {
+    throw std::invalid_argument{"input node_base cannot be null"};
+  }
+  if (node_timers == nullptr) {
+    throw std::invalid_argument{"input node_timers cannot be null"};
+  }
+
+  const std::chrono::nanoseconds period_ns = detail::safe_cast_to_period_in_ns(period);
+
+  // Add a new generic timer.
+  auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
+    std::move(clock), period_ns, std::move(callback), node_base->get_context());
+  node_timers->add_timer(timer, group);
+  return timer;
+}
 
 /// Convenience method to create a wall timer with node resources.
 /**
@@ -150,7 +197,7 @@ create_wall_timer(
     throw std::invalid_argument{"input node_timers cannot be null"};
   }
 
-  const std::chrono::nanoseconds period_ns = safe_cast_to_period_in_ns(period);
+  const std::chrono::nanoseconds period_ns = detail::safe_cast_to_period_in_ns(period);
 
   // Add a new wall timer.
   auto timer = rclcpp::WallTimer<CallbackT>::make_shared(
@@ -158,54 +205,6 @@ create_wall_timer(
   node_timers->add_timer(timer, group);
   return timer;
 }
-
-/// Convenience method to create a general timer with node resources.
-/**
- *
- * \tparam DurationRepT
- * \tparam DurationT
- * \tparam CallbackT
- * \param period period to execute callback. This duration must be 0 <= period < nanoseconds::max()
- * \param callback callback to execute via the timer period
- * \param group callback group
- * \param node_base node base interface
- * \param node_timers node timer interface
- * \return shared pointer to a generic timer
- * \throws std::invalid_argument if either node_base, node_timers or node_clock
- * are null, or period is negative or too large
- */
-template<typename DurationRepT, typename DurationT, typename CallbackT>
-typename rclcpp::GenericTimer<CallbackT>::SharedPtr
-create_timer(
-  std::chrono::duration<DurationRepT, DurationT> period,
-  CallbackT callback,
-  rclcpp::CallbackGroup::SharedPtr group,
-  node_interfaces::NodeBaseInterface * node_base,
-  node_interfaces::NodeTimersInterface * node_timers,
-  node_interfaces::NodeClockInterface * node_clock)
-{
-  if (node_base == nullptr) {
-    throw std::invalid_argument{"input node_base cannot be null"};
-  }
-
-  if (node_timers == nullptr) {
-    throw std::invalid_argument{"input node_timers cannot be null"};
-  }
-
-  if (node_clock == nullptr) {
-    throw std::invalid_argument{"input node_clock cannot be null"};
-  }
-
-  const std::chrono::nanoseconds period_ns = safe_cast_to_period_in_ns(period);
-
-  // Add a new generic timer.
-  auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
-    node_clock->get_clock(), period_ns, std::move(callback), node_base->get_context());
-  node_timers->add_timer(timer, group);
-  return timer;
-}
-
-
 }  // namespace rclcpp
 
 #endif  // RCLCPP__CREATE_TIMER_HPP_
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -78,7 +78,7 @@ safe_cast_to_period_in_ns(std::chrono::duration<DurationRepT, DurationT> period)
 
   return period_ns;
 }
-}
+}  // namespace detail
 
 /// Create a timer with a given clock
 /// \internal
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -25,6 +25,7 @@
 #include "rclcpp/node_interfaces/get_node_base_interface.hpp"
 #include "rclcpp/node_interfaces/get_node_timers_interface.hpp"
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
+#include "rclcpp/node_interfaces/node_clock_interface.hpp"
 #include "rclcpp/node_interfaces/node_timers_interface.hpp"
 
 namespace rclcpp
@@ -70,38 +71,19 @@ create_timer(
     group);
 }
 
-/// Convenience method to create a timer with node resources.
+/// Perform a safe cast to a timer period in nanoseconds
 /**
  *
  * \tparam DurationRepT
  * \tparam DurationT
- * \tparam CallbackT
  * \param period period to execute callback. This duration must be 0 <= period < nanoseconds::max()
- * \param callback callback to execute via the timer period
- * \param group
- * \param node_base
- * \param node_timers
- * \return
- * \throws std::invalid argument if either node_base or node_timers
- * are null, or period is negative or too large
+ * \return period, expressed as chrono::duration::nanoseconds
+ * \throws std::invalid_argument if period is negative or too large
  */
-template<typename DurationRepT, typename DurationT, typename CallbackT>
-typename rclcpp::WallTimer<CallbackT>::SharedPtr
-create_wall_timer(
-  std::chrono::duration<DurationRepT, DurationT> period,
-  CallbackT callback,
-  rclcpp::CallbackGroup::SharedPtr group,
-  node_interfaces::NodeBaseInterface * node_base,
-  node_interfaces::NodeTimersInterface * node_timers)
+template<typename DurationRepT, typename DurationT>
+std::chrono::nanoseconds
+safe_cast_to_period_in_ns(std::chrono::duration<DurationRepT, DurationT> period)
 {
-  if (node_base == nullptr) {
-    throw std::invalid_argument{"input node_base cannot be null"};
-  }
-
-  if (node_timers == nullptr) {
-    throw std::invalid_argument{"input node_timers cannot be null"};
-  }
-
   if (period < std::chrono::duration<DurationRepT, DurationT>::zero()) {
     throw std::invalid_argument{"timer period cannot be negative"};
   }
@@ -132,11 +114,96 @@ create_wall_timer(
             "Casting timer period to nanoseconds resulted in integer overflow."};
   }
 
+  return period_ns;
+}
+
+/// Convenience method to create a wall timer with node resources.
+/**
+ *
+ * \tparam DurationRepT
+ * \tparam DurationT
+ * \tparam CallbackT
+ * \param period period to execute callback. This duration must be 0 <= period < nanoseconds::max()
+ * \param callback callback to execute via the timer period
+ * \param group callback group
+ * \param node_base node base interface
+ * \param node_timers node timer interface
+ * \return shared pointer to a wall timer
+ * \throws std::invalid_argument if either node_base or node_timers
+ * are null, or period is negative or too large
+ */
+template<typename DurationRepT, typename DurationT, typename CallbackT>
+typename rclcpp::WallTimer<CallbackT>::SharedPtr
+create_wall_timer(
+  std::chrono::duration<DurationRepT, DurationT> period,
+  CallbackT callback,
+  rclcpp::CallbackGroup::SharedPtr group,
+  node_interfaces::NodeBaseInterface * node_base,
+  node_interfaces::NodeTimersInterface * node_timers)
+{
+  if (node_base == nullptr) {
+    throw std::invalid_argument{"input node_base cannot be null"};
+  }
+
+  if (node_timers == nullptr) {
+    throw std::invalid_argument{"input node_timers cannot be null"};
+  }
+
+  const std::chrono::nanoseconds period_ns = safe_cast_to_period_in_ns(period);
+
+  // Add a new wall timer.
   auto timer = rclcpp::WallTimer<CallbackT>::make_shared(
     period_ns, std::move(callback), node_base->get_context());
   node_timers->add_timer(timer, group);
   return timer;
 }
+
+/// Convenience method to create a general timer with node resources.
+/**
+ *
+ * \tparam DurationRepT
+ * \tparam DurationT
+ * \tparam CallbackT
+ * \param period period to execute callback. This duration must be 0 <= period < nanoseconds::max()
+ * \param callback callback to execute via the timer period
+ * \param group callback group
+ * \param node_base node base interface
+ * \param node_timers node timer interface
+ * \return shared pointer to a generic timer
+ * \throws std::invalid_argument if either node_base, node_timers or node_clock
+ * are null, or period is negative or too large
+ */
+template<typename DurationRepT, typename DurationT, typename CallbackT>
+typename rclcpp::GenericTimer<CallbackT>::SharedPtr
+create_timer(
+  std::chrono::duration<DurationRepT, DurationT> period,
+  CallbackT callback,
+  rclcpp::CallbackGroup::SharedPtr group,
+  node_interfaces::NodeBaseInterface * node_base,
+  node_interfaces::NodeTimersInterface * node_timers,
+  node_interfaces::NodeClockInterface * node_clock)
+{
+  if (node_base == nullptr) {
+    throw std::invalid_argument{"input node_base cannot be null"};
+  }
+
+  if (node_timers == nullptr) {
+    throw std::invalid_argument{"input node_timers cannot be null"};
+  }
+
+  if (node_clock == nullptr) {
+    throw std::invalid_argument{"input node_clock cannot be null"};
+  }
+
+  const std::chrono::nanoseconds period_ns = safe_cast_to_period_in_ns(period);
+
+  // Add a new generic timer.
+  auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
+    node_clock->get_clock(), period_ns, std::move(callback), node_base->get_context());
+  node_timers->add_timer(timer, group);
+  return timer;
+}
+
 
 }  // namespace rclcpp
 
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -34,7 +34,7 @@ namespace rclcpp
 /// Create a timer with a given clock
 /// \internal
 template<typename CallbackT>
-typename rclcpp::GenericTimer<CallbackT>::SharedPtr
+typename rclcpp::TimerBase::SharedPtr
 create_timer(
   std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
   std::shared_ptr<node_interfaces::NodeTimersInterface> node_timers,
@@ -43,37 +43,19 @@ create_timer(
   CallbackT && callback,
   rclcpp::CallbackGroup::SharedPtr group = nullptr)
 {
-  if (node_base == nullptr) {
-    throw std::invalid_argument{"input node_base cannot be null"};
-  }
-
-  if (node_timers == nullptr) {
-    throw std::invalid_argument{"input node_timers cannot be null"};
-  }
-
-  // If clock is not specified, then we want a wall timer by default.
-  if (clock == nullptr) {
-    auto timer = rclcpp::WallTimer<CallbackT>::make_shared(
-      period.to_chrono<std::chrono::nanoseconds>(),
-      std::forward<CallbackT>(callback),
-      node_base->get_context());
-    node_timers->add_timer(timer, group);
-    return timer;
-  }
-
-  // This is a timer that can be driven by any clock source.
   auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
     clock,
     period.to_chrono<std::chrono::nanoseconds>(),
     std::forward<CallbackT>(callback),
     node_base->get_context());
+
   node_timers->add_timer(timer, group);
   return timer;
 }
 
 /// Create a timer with a given clock
 template<typename NodeT, typename CallbackT>
-typename rclcpp::GenericTimer<CallbackT>::SharedPtr
+typename rclcpp::TimerBase::SharedPtr
 create_timer(
   NodeT node,
   rclcpp::Clock::SharedPtr clock,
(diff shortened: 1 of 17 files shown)


jacobperron:
Make sure to also sign-off your commits to satisfy the DCO check.

jacobperron (commented):
Looks good to me. Please add a test exercising the new method, then I'll trigger CI.

asymingt:
> Looks good to me. Please add a test exercising the new method, then I'll trigger CI.

Done. Please take a look at the unit tests, and particularly the `EXPECT_FALSE(timer->is_steady()` for the `GENERIC_TIMER` implementation.

Unit tests pass locally:

```txt
83: [==========] Running 16 tests from 1 test suite.
83: [----------] Global test environment set-up.
83: [----------] 16 tests from PerTimerType/TestTimer
83: [ RUN      ] PerTimerType/TestTimer.test_simple_cancel/wall_timer
83: [       OK ] PerTimerType/TestTimer.test_simple_cancel/wall_timer (17 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_simple_cancel/generic_timer
83: [       OK ] PerTimerType/TestTimer.test_simple_cancel/generic_timer (5 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_is_canceled_reset/wall_timer
83: [       OK ] PerTimerType/TestTimer.test_is_canceled_reset/wall_timer (5 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_is_canceled_reset/generic_timer
83: [       OK ] PerTimerType/TestTimer.test_is_canceled_reset/generic_timer (5 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_run_cancel_executor/wall_timer
83: [       OK ] PerTimerType/TestTimer.test_run_cancel_executor/wall_timer (105 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_run_cancel_executor/generic_timer
83: [       OK ] PerTimerType/TestTimer.test_run_cancel_executor/generic_timer (105 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_run_cancel_timer/wall_timer
83: [       OK ] PerTimerType/TestTimer.test_run_cancel_timer/wall_timer (105 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_run_cancel_timer/generic_timer
83: [       OK ] PerTimerType/TestTimer.test_run_cancel_timer/generic_timer (105 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_bad_arguments/wall_timer
83: [ERROR] [1658525415.784172078] []: Failed to fini rcl clock.
83: [       OK ] PerTimerType/TestTimer.test_bad_arguments/wall_timer (5 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_bad_arguments/generic_timer
83: [ERROR] [1658525415.788818007] []: Failed to fini rcl clock.
83: [       OK ] PerTimerType/TestTimer.test_bad_arguments/generic_timer (5 ms)
83: [ RUN      ] PerTimerType/TestTimer.callback_with_timer/wall_timer
83: [       OK ] PerTimerType/TestTimer.callback_with_timer/wall_timer (7 ms)
83: [ RUN      ] PerTimerType/TestTimer.callback_with_timer/generic_timer
83: [       OK ] PerTimerType/TestTimer.callback_with_timer/generic_timer (6 ms)
83: [ RUN      ] PerTimerType/TestTimer.callback_with_period_zero/wall_timer
83: [       OK ] PerTimerType/TestTimer.callback_with_period_zero/wall_timer (5 ms)
83: [ RUN      ] PerTimerType/TestTimer.callback_with_period_zero/generic_timer
83: [       OK ] PerTimerType/TestTimer.callback_with_period_zero/generic_timer (4 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_failures_with_exceptions/wall_timer
83: [ERROR] [1658525415.815430083] [rclcpp]: Failed to clean up rcl timer handle: error not set
83: [       OK ] PerTimerType/TestTimer.test_failures_with_exceptions/wall_timer (4 ms)
83: [ RUN      ] PerTimerType/TestTimer.test_failures_with_exceptions/generic_timer
83: [ERROR] [1658525415.820125821] [rclcpp]: Failed to clean up rcl timer handle: error not set
83: [       OK ] PerTimerType/TestTimer.test_failures_with_exceptions/generic_timer (5 ms)
83: [----------] 16 tests from PerTimerType/TestTimer (493 ms total)
83: 
83: [----------] Global test environment tear-down
83: [==========] 16 tests from 1 test suite ran. (493 ms total)
83: [  PASSED  ] 16 tests.
83: -- run_test.py: return code 0    
83: -- run_test.py: inject classname prefix into gtest result file 
The following tests passed:
        test_timer

...

100% tests passed, 0 tests failed out of 1

Label Time Summary:
gtest    =   0.68 sec*proc (1 test)

Total Test time (real) =   0.68 sec
Finished <<< rclcpp [0.94s]

Summary: 1 package finished [1.76s]
```

jacobperron:
>  To me it seems logical that being driven by /clock might end up making the time source non-monotonic. Does this violate any assumption made by the timer class?

Correct, if we are using a ROS clock (ie. potentially relying on the /clock topic) then we can't assume time will be be monotonically increasing. I don't think this violates any assumptions of the timer class, but I can double-check.

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -132,12 +114,97 @@ create_wall_timer(
...
+ * \param group callback group
+ * \param node_base node base interface
+ * \param node_timers node timer interface
+ * \return shared pointer to a generic timer
+ * \throws std::invalid_argument if either node_base, node_timers or node_clock
+ * are null, or period is negative or too large
+ */
+template<typename DurationRepT, typename DurationT, typename CallbackT>
+typename rclcpp::GenericTimer<CallbackT>::SharedPtr
+create_timer(
```
fujitatomoya, line 178: I think this needs to be tested in `rclcpp/test/rclcpp/test_create_timer.cpp`?

fujitatomoya (approved):
lgtm with one question.

rclcpp/include/rclcpp/create_timer.hpp
asymingt: Looks like [simply including this header](https://github.com/ros2/rclcpp/pull/1975/commits/4594e88330dbfc3ce98e81f9f10c5d02c3737742#diff-cb1652cc3b1eff7b2611c2b78644d80d73b7e7d4185649e20a88926de473af2dR26) has triggered a CI build failure, which might be beyond my ability to solve...

clalancette: > Looks like [including this header](https://github.com/ros2/rclcpp/pull/1975/commits/4594e88330dbfc3ce98e81f9f10c5d02c3737742#diff-cb1652cc3b1eff7b2611c2b78644d80d73b7e7d4185649e20a88926de473af2dR26) has triggered a CI failure, which might be beyond my ability to solve. I can take a look but I might need @jacobperron to step in.

It looks unlikely to be because of including the header.  Probably something in here is causing an endless loop, or at least a very slow loop, so that it hits the timeout.

asymingt: Oh, that's interesting. Thanks, @clalancette.  I had completely misinterpreted the `[ERROR] [1659380009.202064912] [rclcpp]: failed to remove clock` as being the source of the issue. It seems too coincidental that the test took exactly one minute and zero seconds to complete. Perhaps my additional tests pushed it over some 60s internal threshold. For what it's worth, the tests pass locally so perhaps just bumping the timeout might fix the issue.

clalancette: It's a little hard to see, but if you search for `test_time_source` in https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/215/consoleFull (for example), you can see it currently takes ~18 seconds.  Going beyond 60 seconds here is a huge increase, so I'd definitely say that there is something weird going on in this PR.

asymingt: OK -- Now that I can see a full console log delta between builds pre- and post- my changes, it's likely that this PR is interacting negatively with simulated time in replay -- or its potentially surfacing a pre-existing issue. This agrees with local rosbag2 replay behavior I'm seeing, and might point to a larger issue with node timers driven by the `/clock` topic. I'll try and dig deeper to work out why.

asymingt: @clalancette do you know if [this line of code](https://github.com/ros2/rclcpp/blob/4594e88330dbfc3ce98e81f9f10c5d02c3737742/rclcpp/test/rclcpp/test_time_source.cpp#L675) is correct? To me it looks like nanoseconds is being incremented and will overflow. Should it not rather be something like this:

```c++
pub_time_ += rclcpp::Duration(0, 1000000);
clock_msg_.clock = pub_time_;
clock_pub_->publish(clock_msg_);
```

clalancette: Off-hand, I'm not sure.  That said, I think your change makes sense, so I'd give that a shot.

asymingt: All failed tests happened on [agent-3af8c62f](https://build.ros2.org/computer/agent-3af8c62f). I think this CI worker needs to be spun down, as it's causing issues.

Blast545:

rclcpp/test/rclcpp/test_time_source.cpp
```diff
@@ -695,7 +694,7 @@ class ClockThreadTestingNode : public rclcpp::Node
     this->set_parameter(rclcpp::Parameter("use_sim_time", true));
 
     // Create a 100ms timer
-    timer_ = this->create_wall_timer(
+    timer_ = this->create_timer(
```
asymingt, line 697: We should consider reverting this, even though the reversion should be a no-op. I get that this timer sets up the periodic publishing of `/clock` to drive peer nodes. For this reason a wall timer makes more sense. However, my understanding is that since the parameter`use_sim_time` is not set on this node, it should fall back to a wall timer. Thus, both `create_timer` and `create_wall_timer` are equivalent here.

asymingt:
I suspect that the nondeterminism in this PR has to do with discovery of the `/clock` topic periodically failing for some reason. If you consider the `/clock` topic is produced by some `ClockSourceNode` for consumption by `ClockSinkNode` (which has `use_sim_time: True` set as a local parameter. Internally if `ClockSinkNode` uses `this->create_timer` the resulting timer will only be advanced when `/clock` is received. However, it seems like the DDS discovery mechanism sometimes fails to find `/clock` and the timer spins indefinitely at `ros::Time(0)`. I'm not sure why this only happens on some machines and not others, but I have seen it happen on my local machine.

asymingt:
> @clalancette and @jacobperron -- I suspect that the nondeterminism in this PR has to do with discovery of the `/clock` topic periodically failing for some reason. If you consider the `/clock` topic is produced by some `ClockSourceNode` for consumption by `ClockSinkNode` (which has `use_sim_time: True` set as a local parameter). Internally if `ClockSinkNode` uses `this->create_timer` the resulting timer will only be advanced when `/clock` is received). However, it seems like the DDS discovery mechanism sometimes fails to find `/clock` and the timer spins indefinitely at `ros::Time(0)`. I'm not sure why this only happens on some machines and not others, but I have seen it happen on my local machine.

Update: the problem is completely fixed locally when I set `rclcpp::NodeOptions().use_clock_thread(false)` when instantiating my sink node. Perhaps there is a race condition with the clock thread implementation.

rclcpp/test/rclcpp/test_time_source.cpp
jacobperron: If I understand the tests correctly, we want this to be a wall timer so I would just revert it to be safe.

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -132,12 +115,97 @@ create_wall_timer(
...
+    throw std::invalid_argument{"input node_base cannot be null"};
+  }
+
+  if (node_timers == nullptr) {
+    throw std::invalid_argument{"input node_timers cannot be null"};
+  }
+
+  if (node_clock == nullptr) {
+    throw std::invalid_argument{"input node_clock cannot be null"};
+  }
```
ivanpauno, line 197: can you add this checks to the `create_timer` function in L38 and reuse it?
Thanks!

It would also be nice to reorder so you first have all `create_timer()` overloads and then you have `create_wall_timer()`

ivanpauno (commented):
I haven't looked at the tests yet, but code looks good.
I only have a minor comment.

rclcpp/test/rclcpp/test_time_source.cpp
ivanpauno: yes, using `create_wall_timer()` here makes more sense I think

rclcpp/test/rclcpp/test_timer.cpp
```diff
@@ -44,21 +51,28 @@ class TestTimer : public ::testing::Test
 
     test_node = std::make_shared<rclcpp::Node>("test_timer_node");
 
-    timer = test_node->create_wall_timer(
-      100ms,
-      [this]() -> void
-      {
+    auto lambda_fun = [this]() -> void {
```
ivanpauno, line 54: I would use a more explicit name

```suggestion
    auto timer_callback = [this]() -> void {
```

ivanpauno:
> Update: the problem is completely fixed locally when I set rclcpp::NodeOptions().use_clock_thread(false) when instantiating my ClockSinkNode. Perhaps there is a race condition with the clock thread implementation.

Could you link what's the test that was causing issues?
I went through the previous CI jobs and I see different tests failing, but I'm not sure what's the one you're referring here.

rclcpp/test/rclcpp/test_time_source.cpp
asymingt: Can you double check this is what you want!

> We should consider reverting this, even though the reversion should be a no-op. I get that this timer sets up the periodic publishing of /clock to drive peer nodes' perception of time (which we call "ROS time"). For this reason a wall timer makes more sense. However, my understanding is that since the parameter use_sim_time is not set on this node, create_timer should simply fall back to a wall timer. Thus, both create_timer and create_wall_timer are equivalent here.

The above message I sent was wrong, as I incorrectly thought this change was to  the `SimClockPublisherNode` code, which it is not. If you look at how this code is used in the unit test , we have a `SimClockPublisherNode` that publishes time on `/clock` and a `ClockThreadTestingNode` that consumes this time. I'm fairly sure that we want a `create_timer` here to prove that a `/clock` driven sense of time has a callback within which you can make subsequent calls to `this->now()`. If we really wanted to test a wall timer, there would be no need to spin up a `SimClockPublisherNode` instance.

asymingt:
> > Update: the problem is completely fixed locally when I set rclcpp::NodeOptions().use_clock_thread(false) when instantiating my ClockSinkNode. Perhaps there is a race condition with the clock thread implementation.
> 
> Could you link what the test that was causing issues is? I went through the previous CI jobs and I see different tests failing, but I'm not sure what's the one you're referring here.

The last five commits are all empty, and in theory they should all pass. However, passing happens on only some CI workers (look at the CI build history). Locally, things pass every time. I can *occasionally* replicate the issue with test `TestTimeSource.check_sim_time_updated_in_callback_if_use_clock_thread`, but there's no single test I can show you repeatedly fails locally.

rclcpp/test/rclcpp/test_time_source.cpp
ivanpauno: Yes, that seems to be the case.
I'm not sure what was the test doing before as `create_wall_timer()` never used sim time.

ivanpauno:
> The last five commits are all empty, and in theory they should all pass. However, passing happens on only some CI workers (look at the CI build history). Locally, things pass every time. I can occasionally replicate the issue with test TestTimeSource.check_sim_time_updated_in_callback_if_use_clock_thread, but there's no single test I can show you repeatedly fails locally.

Okay, that's strange.
I will take a look and see if I can reproduce locally.

ivanpauno:
I'm taking a look to all the failures in https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/227/#showFailuresLink and https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/228/#showFailuresLink, and they don't seem to be related to your changes at all.

It seems that for some reason they're happening in the Rpr job but not in the nightlies.

About `test_time_source`, the tests are written in a way that can easily fail.
Increasing `1s` [here](https://github.com/ros2/rclcpp/blob/b3445cec5c9bcb186c5ec8a751f05db3d6bb79c5/rclcpp/test/rclcpp/test_time_source.cpp#L77) to a longer duration may fix the issue.

rclcpp/test/rclcpp/test_time_source.cpp
ivanpauno: [The comments](https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp/test/rclcpp/test_time_source.cpp#L739-L742) also say that this test should be running with `use_clock_thread=true`, and that's not the case.

When I modify the test to do what's supposed to do, it's super flaky.
IMO, it's fair to comment out the test and create a new issue for it.
@jacobperron any opinion?

rclcpp/include/rclcpp/create_timer.hpp
asymingt: This is unfortunately not trivial because it involves a function signature change. The reason is because there is a mismatch between argument types (some are raw pointers, and others are shared_ptr). What you have asked for requires a raw -> shared_ptr conversion, which is not possible.

asymingt: @ivanpauno  please take a look at commit https://github.com/ros2/rclcpp/pull/1975/commits/329ec5215c39b62cb395d4644751a25b5d84b86b. It's a fairly large change that I think achieves what you have asked for. Notably it:
- Removes raw pointers in favor of shared pointers.
- Forces the create_timer call to follow a sequence of function calls.
- Moves the argument nullptr checks to a common place.
It did require making changes to some other ROS2 components to avoid passing around a raw pointer from which a shared pointer cannot be derived.

asymingt:
> I'm taking a look to all the failures in https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/227/#showFailuresLink and https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/228/#showFailuresLink, and they don't seem to be related to your changes at all.
> 
> It seems that for some reason they're happening in the Rpr job but not in the nightlies.
> 
> About `test_time_source`, the tests are written in a way that can easily fail. Increasing `1s` [here](https://github.com/ros2/rclcpp/blob/b3445cec5c9bcb186c5ec8a751f05db3d6bb79c5/rclcpp/test/rclcpp/test_time_source.cpp#L77) to a longer duration may fix the issue.

I initially tried `10s` and that was too large, so I dropped it down to `2s` and this seems more stable.

asymingt:
@jacobperron and @ivanpauno - Please take a look at the last two commits and let me know if you are happy. I think I addressed all review comments, even though [this one](https://github.com/ros2/rclcpp/pull/1975#discussion_r942886593) was quite tricky. I am of the opinion that this PR is now ready to merge.

asymingt:

asymingt:

rclcpp/test/rclcpp/test_timer.cpp
asymingt: Fixed

ivanpauno:
Something I noticed while reading the tests, this line should say `>=` and not `==`:

https://github.com/ros2/rclcpp/blob/6e0ec4a77001765a045ab9abb0505ebb36e4c43f/rclcpp/test/rclcpp/test_time_source.cpp#L84

rclcpp/include/rclcpp/create_timer.hpp
ivanpauno: > This is unfortunately not trivial because it involves a function signature change. The reason is because there is a mismatch between argument types (some are raw pointers, and others are shared_ptr). If I understand your comment, what you have asked for requires a raw -> shared_ptr conversion, which is not possible.

Sorry, I didn't notice that wehn commenting, let me check your changes.

asymingt:
> Something I noticed while reading the tests, this line should say `>=` and not `==`:
> 
> https://github.com/ros2/rclcpp/blob/6e0ec4a77001765a045ab9abb0505ebb36e4c43f/rclcpp/test/rclcpp/test_time_source.cpp#L84

Fixed in https://github.com/ros2/rclcpp/pull/1975/commits/c7acee7434b9cddbc9eff86dde7a7af3cfdb2958

rclcpp/include/rclcpp/create_timer.hpp
ivanpauno: @asymingt sorry that my comment was confusing.
I simplified the changes in https://github.com/ros2/rclcpp/pull/1975/commits/5c129b8b17076cc9af55fc9b278ad283c6abb9b3 and https://github.com/ros2/rclcpp/pull/1975/commits/ff241d40333d31912f31570e915d0cb7665ea376

rclcpp/test/rclcpp/test_time_source.cpp
ivanpauno: I commented out the test as I think the problem is the test itself, see https://github.com/ros2/rclcpp/pull/1975/commits/009312185f959d93d0dde78b4a53d294aec68daa

rclcpp/include/rclcpp/create_timer.hpp
ivanpauno: I commented out one test (see https://github.com/ros2/rclcpp/pull/1975#discussion_r956348057), that isn't doing what it's supposed to do.
The other tests that are flaky, I was able to reproduce the failure both for this branch and for `rolling`, so the issue is unrelated to this work.

ivanpauno (approved):
LGTM!
I will wait for @jacobperron approval as I made some minor changes.

Thanks @asymingt !!

asymingt:
> LGTM! I will wait for @jacobperron approval as I made some minor changes.
> 
> Thanks @asymingt !!

Thanks for getting this one over the line, Ivan 👍

rclcpp/test/rclcpp/test_time_source.cpp
jacobperron: I thought the test was checking that the wall timer and simulation time updates were not interfering with each other. But, I'm not sure. I'm okay with disabling the flaky test. Maybe we can create a GitHub ticket to look into it (maybe it is just not a well-defined test).

jacobperron (approved):
LGTM, pending CI. One tiny nitpick.

rclcpp/include/rclcpp/node_interfaces/node_topics.hpp
```diff
@@ -15,6 +15,7 @@
 #ifndef RCLCPP__NODE_INTERFACES__NODE_TOPICS_HPP_
 #define RCLCPP__NODE_INTERFACES__NODE_TOPICS_HPP_
 
+#include <memory>
```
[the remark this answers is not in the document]
ivanpauno:
```suggestion
```

rclcpp/test/rclcpp/test_time_source.cpp
ivanpauno: Based on the comments, the test name, and the code; it seems that the test is supposed to be checking that the clock is updated when `use_clock_thread=True` even if you're blocked in a callback.
The part that's missing is that `use_clock_thread` was never actually enabled.
Enabling that doesn't solve the issue, as it seems something is not working well.

@anaelle-sw I think you added this feature, could you take a look at the issue?

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17184/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11726/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17607/)

ivanpauno:
I forgot to include rclcpp_lifecycle and test_rclcpp, running CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17187/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11728/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17609/)

sloretz:
I'm seeing a build failure in building tf2_ros that might be related to this change: https://ci.ros2.org/job/ci_linux-aarch64/11736/console#console-section-12

```
--- stderr: tf2_ros
09:46:40 In file included from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_subscription.hpp[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/node_impl.hpp[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/node.hpp:1[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/executors/single_threaded_executor.hpp[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/executors.hpp[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/rclcpp.hpp:[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/src/ros2/geometry2/tf2_ros/include/tf2_ros/create_timer_interface.h[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/src/ros2/geometry2/tf2_ros/include/tf2_ros/create_timer_ros.h[K,
09:46:40                  from [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/src/ros2/geometry2/tf2_ros/src/create_timer_ros.cpp[K:
09:46:40 /home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp: In instantiation of ‘[Krclcpp::TimerBase::SharedPtr rclcpp::create_timer(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeTimersInterface>, rclcpp::Clock::SharedPtr, rclcpp::Duration, CallbackT&&, rclcpp::CallbackGroup::SharedPtr) [with CallbackT = std::function<void()>; rclcpp::TimerBase::SharedPtr = std::shared_ptr<rclcpp::TimerBase>; rclcpp::Clock::SharedPtr = std::shared_ptr<rclcpp::Clock>; rclcpp::CallbackGroup::SharedPtr = std::shared_ptr<rclcpp::CallbackGrou[K’:
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/src/ros2/geometry2/tf2_ros/src/create_timer_ros.cpp:61:[K   required from here
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:95:[Kerro[Kno matching function for call t[Kcreate_timer(std::__shared_ptr<rclcpp::node_interfaces::NodeBaseInterface, __gnu_cxx::_S_atomic>::element_type*, std::__shared_ptr<rclcpp::node_interfaces::NodeTimersInterface, __gnu_cxx::_S_atomic>::element_type*, rclcpp::Clock::SharedPtr&, std::chrono::duration<long int, std::ratio<1, 1000000000> >, std::function<void()>, rclcpp::CallbackGroup::SharedPt[K’
09:46:40    95 |   return [Kcreate_tim[K
09:46:40       |          [K~~~~~~~~~~[K
09:46:40    96 | [K    node_base.get[K
09:46:40       |     [K~~~~~~~~~~~~~[K  
09:46:40    97 | [K    node_timers.get[K
09:46:40       |     [K~~~~~~~~~~~~~~~[K
09:46:40    98 | [K    clo[K
09:46:40       |     [K~~~[K            
09:46:40    99 | [K    period.to_chrono<std::chrono::nanoseconds>[K
09:46:40       |     [K~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
09:46:40   100 | [K    std::forward<CallbackT>(callbac[K
09:46:40       |     [K~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
09:46:40   101 | [K    gro[K;
09:46:40       |     [K~~~[K            
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:87[Knot[Kcandidate[Ktemplate<class CallbackT> rclcpp::TimerBase::SharedPtr rclcpp::create_timer(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>, std::shared_ptr<rclcpp::node_interfaces::NodeTimersInterface>, rclcpp::Clock::SharedPtr, rclcpp::Duration, CallbackT&&, rclcpp::CallbackGroup::SharedP[K’
09:46:40    87 | [Kcreate_ti[K(
09:46:40       | [K^~~~~~~~~[K
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:87[Knot[K  template argument deduction/substitution failed:
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:96:[Knot[K  cannot conver[Knode_base.std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>::<anonymous>.std::__shared_ptr<rclcpp::node_interfaces::NodeBaseInterface, __gnu_cxx::_S_atomic>::ge[K’ (typ[Kstd::__shared_ptr<rclcpp::node_interfaces::NodeBaseInterface, __gnu_cxx::_S_atomic>::element_ty[K’ {ak[Krclcpp::node_interfaces::NodeBaseInterfa[K’}) to typ[Kstd::shared_ptr<rclcpp::node_interfaces::NodeBaseInterfa[K’
09:46:40    96 |     [Knode_base.ge[K,
09:46:40       |     [K~~~~~~~~~~~~[K
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:141[Knot[Kcandidate[Ktemplate<class DurationRepT, class DurationT, class CallbackT> typename rclcpp::GenericTimer<CallbackT>::SharedPtr rclcpp::create_timer(rclcpp::Clock::SharedPtr, std::chrono::duration<_Rep1, _Period1>, CallbackT, rclcpp::CallbackGroup::SharedPtr, rclcpp::node_interfaces::NodeBaseInterface*, rclcpp::node_interfaces::NodeTimersInterfac[K’
09:46:40   141 | [Kcreate_ti[K(
09:46:40       | [K^~~~~~~~~[K
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:141[Knot[K  template argument deduction/substitution failed:
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:95:[Knot[K  mismatched type[Kstd::chrono::duration<_Rep1, _Perio[K’ an[Kstd::__shared_ptr<rclcpp::node_interfaces::NodeTimersInterface, __gnu_cxx::_S_atomic>::element_ty[K’ {ak[Krclcpp::node_interfaces::NodeTimersInterfa[K’}
09:46:40    95 |   return [Kcreate_tim[K
09:46:40       |          [K~~~~~~~~~~[K
09:46:40    96 | [K    node_base.get[K
09:46:40       |     [K~~~~~~~~~~~~~[K  
09:46:40    97 | [K    node_timers.get[K
09:46:40       |     [K~~~~~~~~~~~~~~~[K
09:46:40    98 | [K    clo[K
09:46:40       |     [K~~~[K            
09:46:40    99 | [K    period.to_chrono<std::chrono::nanoseconds>[K
09:46:40       |     [K~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
09:46:40   100 | [K    std::forward<CallbackT>(callbac[K
09:46:40       |     [K~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
09:46:40   101 | [K    gro[K;
09:46:40       |     [K~~~[K            
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:107[Knot[Kcandidate[Ktemplate<class NodeT, class CallbackT> rclcpp::TimerBase::SharedPtr rclcpp::create_timer(NodeT, rclcpp::Clock::SharedPtr, rclcpp::Duration, CallbackT&&, rclcpp::CallbackGroup::SharedP[K’
09:46:40   107 | [Kcreate_ti[K(
09:46:40       | [K^~~~~~~~~[K
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:107[Knot[K  template argument deduction/substitution failed:
09:46:40 [K/home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/rclcpp/include/rclcpp/rclcpp/create_timer.hpp:95:[Knot[K  candidate expects 5 arguments, 6 provided
09:46:40    95 |   return [Kcreate_tim[K
09:46:40       |          [K~~~~~~~~~~[K
09:46:40    96 | [K    node_base.get[K
09:46:40       |     [K~~~~~~~~~~~~~[K  
09:46:40    97 | [K    node_timers.get[K
09:46:40       |     [K~~~~~~~~~~~~~~~[K
09:46:40    98 | [K    clo[K
09:46:40       |     [K~~~[K            
09:46:40    99 | [K    period.to_chrono<std::chrono::nanoseconds>[K
09:46:40       |     [K~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
09:46:40   100 | [K    std::forward<CallbackT>(callbac[K
09:46:40       |     [K~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
09:46:40   101 | [K    gro[K;
09:46:40       |     [K~~~[K            
```

jacobperron:
We probably should have run with test args `--package-above rclcpp rclcpp_lifecycle`

sloretz:
Checking if it's this PR

build: `--packags-up-to tf2_ros`, test: `--packages-select tf2_ros` linux aarch64

With this PR: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/11737/)

WIthout this PR [repos file](https://gist.githubusercontent.com/sloretz/cdcec97436c787ac2178e6dd09a7e57f/raw/6854e82547468b2e702913a6346708271c0d2241/revert.repos) [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/11738/)

PR #1981, opened 2022-08-04
[docs] add note about callback lifetime for {on, post}_set_parameter_callback

ihasdapie:
I've been burned by ignoring the return value and so has a few others -- adding a note about the lifetime of the registered callback being tied to the returned callback handle further up the documentation should help others avoid making the same mistake.

Signed-off-by: Brian Chen <brian.chen@openrobotics.org>

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -1039,6 +1039,9 @@ public:
    * rcl_interfaces::msg::SetParametersResult to indicate whether or not the
    * parameter should be set or not, and if not why.
    *
+   * The back returns a shared pointer to the callback handle.
+   * The callback is valid as long as the smart pointer is alive.
+   *
    * For an example callback:
    *
    * ```cpp
@@ -1106,6 +1109,9 @@ public:
   /**
    * The callback is called when any of the `set_parameter*` or `declare_parameter*`
    * methods are successful.
+   *
+   * The back returns a shared pointer to the callback handle.
+   * The callback is valid as long as the smart pointer is alive.
    *
    * The callback takes a reference to a const vector of parameters that have been
    * set successfully.
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -991,6 +991,9 @@ public:
    * One of the use case of "pre set callback" can be updating additional parameters
    * conditioned on changes to a parameter.
    *
+   * Users should retain a copy of the returned shared pointer, as the callbak
+   * is valid only as long as the smart pointer is alive.
+   *
    * For an example callback:
    *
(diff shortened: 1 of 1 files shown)


rclcpp/include/rclcpp/node.hpp
```diff
@@ -1039,6 +1039,9 @@ class Node : public std::enable_shared_from_this<Node>
    * rcl_interfaces::msg::SetParametersResult to indicate whether or not the
    * parameter should be set or not, and if not why.
    *
+   * The back returns a shared pointer to the callback handle.
+   * The callback is valid as long as the smart pointer is alive.
+   *
```
fujitatomoya, line 1044: I think that the doc is already there. is this not enough?

https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp/include/rclcpp/node.hpp#L1097

ihasdapie: It's not the most obvious design so I figured that it might be good to make it _extra_ explicitly clear. It can be very confusing when the return value isn't persisted properly, as the user would just see the callback seemingly never get called. If they go put debug logging statements into rclcpp, the `post_set_callback_container_` mysteriously decreasing in size. Putting this detail at the beginning of the docstring might help with this.

 Taking a step back it would be more intuitive to tie the lifetime of the callback to the corresponding node to account for how most users won't scour scour the documentation and instead just assume the most likely behaviour. Then callbacks could be managed just with only the `add_{on, post}_set...` and `remove_{on, post}_set...` methods. Not sure if anyone's taking advantage of the behaviour of where the callback becomes invalidated when the smart pointer gets dropped.

jacobperron: I guess the current design is nice if the object that registered the callback goes out of scope, then we probably want to unregistered the callback to avoid potential undefined behavior.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1039,6 +1039,9 @@ class Node : public std::enable_shared_from_this<Node>
    * rcl_interfaces::msg::SetParametersResult to indicate whether or not the
    * parameter should be set or not, and if not why.
    *
+   * The back returns a shared pointer to the callback handle.
+   * The callback is valid as long as the smart pointer is alive.
```
jacobperron, line 1043:
```suggestion
   * Users should retain a copy of the returned shared pointer, otherwise the callback will be removed.
   * I.e. the callback is valid as long as the smart pointer is alive.
```

jacobperron (commented):
I think the same message should also be added to ` add_pre_set_parameters_callback`.

fujitatomoya (approved):
lgtm

PR #1678, opened 2021-05-17
Remove deprecated (in c++17 and newer) use of std::allocator<>::rebind

pettni:
Addresses #1655

PR #1984, opened 2022-08-09
[galactic backport] Remove use of std::allocator<>::rebind

connoranderson:
rebind is deprecated in c++17 and removed in c++20

backporting https://github.com/ros2/rclcpp/pull/1678 to galactic so it can build with c++20


fujitatomoya (approved):
[galactic targets c++17](https://docs.ros.org/en/galactic/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html#id2) and this is deprecated. So I think we should backport to galactic.

PR #1985, opened 2022-08-10
Service introspection

ihasdapie:
This PR is part of a prototype implementation for https://github.com/ros-infrastructure/rep/pull/360 https://github.com/ros2/ros2/issues/1285. 

Depends on: https://github.com/ros2/rcl/pull/997


rclcpp/include/rclcpp/create_client.hpp
```diff
@@ -44,15 +45,15 @@ create_client(
   std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
   std::shared_ptr<node_interfaces::NodeGraphInterface> node_graph,
   std::shared_ptr<node_interfaces::NodeServicesInterface> node_services,
+  std::shared_ptr<node_interfaces::NodeClockInterface> node_clock,
   const std::string & service_name,
-  const rclcpp::QoS & qos = rclcpp::ServicesQoS(),
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+  const rclcpp::QoS & qos,
+  rclcpp::CallbackGroup::SharedPtr group,
+  bool enable_service_introspection)
```
jacobperron, line 52: If we can, we should try and maintain backwards compatibility by leaving the old signature intact (and pass false for the service introspection flag and a nullptr for the clock interface.

rclcpp/include/rclcpp/create_client.hpp
```diff
@@ -63,12 +64,40 @@ create_client(
   std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
   std::shared_ptr<node_interfaces::NodeGraphInterface> node_graph,
   std::shared_ptr<node_interfaces::NodeServicesInterface> node_services,
+  std::shared_ptr<node_interfaces::NodeClockInterface> node_clock,
+  const std::string & service_name,
+  const rmw_qos_profile_t & qos_profile,
+  rclcpp::CallbackGroup::SharedPtr group,
+  bool enable_service_introspection)
```
jacobperron, line 71: While we're adding/modifying API, it might be a good time to consider introducing an options struct, analogous to those for publishers and [subscriptions](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/subscription_options.hpp#L39). This is where we could bundle together the optional parameters: clock interface, callback group, and service introspection flag.

Doing this now would avoid potential API breaks (or deprecation cycles) in the future. I could look at opening a separate PR for this and then we could build on top of it here (depending how much work it turns out to be).

Related issue: https://github.com/ros2/rclcpp/issues/1989

rclcpp/include/rclcpp/node.hpp
```diff
@@ -270,6 +271,22 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * \param[in] group Callback group to handle the reply to service calls.
+   * \return Shared pointer to the created client.
+   */
+  template<typename ServiceT>
+  typename Client<ServiceT>::SharedPtr
+  create_client(
+    const std::string & service_name,
+    const rclcpp::QoS & qos,
+    const rclcpp::QoS & service_event_publisher_qos,
+    rclcpp::CallbackGroup::SharedPtr group);
```
jacobperron, line 288: IMO, these three parameters should be optional (i.e. assign some default values here).

rclcpp/src/rclcpp/node_interfaces/node_service_introspection.cpp
```diff
@@ -0,0 +1,117 @@
...
+          for (auto srv = services_.begin(); srv != services_.end(); ++srv) {
+            if (srv->expired()) {
+              srv = services_.erase(srv);
+            } else {
+              ret = rcl_service_introspection_configure_server_service_events(
+                  srv->lock()->get_service_handle().get(),
+                  this->node_base_->get_rcl_node_handle(),
+                  param.get_value<bool>());
+              if (RCL_RET_OK != ret) {
+                throw std::runtime_error("Could not configure service introspection events");
```
jacobperron, line 62: The following will throw an exception with the error code and any error message set in rcl:
```suggestion
                rclcpp::exceptions::throw_from_rcl_error(ret);
```

rclcpp/src/rclcpp/node_options.cpp
```diff
@@ -227,6 +229,19 @@ NodeOptions::enable_topic_statistics() const
...
 
+bool
+NodeOptions::enable_service_introspection() const
+{
+  return this->enable_service_introspection_;
+}
+
+NodeOptions &
+NodeOptions::enable_service_introspection(bool enable_service_introspection) {
+  this->node_options_.reset();  // reset node options to make it be recreated on next access.
```
jacobperron, line 240: This is not necessary if we remove the option from the rcl struct (which I think we can do).

ihasdapie: I think that it makes more sense to keep the option in the rcl `rcl_node_options_t` struct as it keeps integration more consistent across client libraries

rclcpp/include/rclcpp/create_client.hpp
ihasdapie: This method is supposed to be an internal method -- the only place outside of `rclcpp` it is used appears to be in `interactive_markers` and a PR has been opened to reconcile that in https://github.com/ihasdapie/interactive_markers/commit/f0e6419c4056f24ee02908728573492af86cd41f. 

That being said we can definitely another method to keep the old signature -- it would just mean that we would need to figure out what kind of behaviour we want if the clock is NULL -- does service introspection just not work or are we ok with no timestamp?

rclcpp/include/rclcpp/create_client.hpp
ihasdapie: Yeah, I agree that it would be nice but a little outside of scope for this PR, though. We can open up a follow-up PR to introduce a ServiceOptions struct later on

rclcpp/src/rclcpp/node_options.cpp
ihasdapie: Actually, looking at it again I am mistaken. We can remove it from the rcl struct. This would mean that `rcl` supports enabling/disabling service introspection for the lifetime of the node on a per-client/service basis and (as it is currently) `rclpy` and `rclcpp` would make it such that enabling/disabling would work on  _all_ service/clients related to a node instead -- which is what is defined in the REP.

clalancette:
This has been substantially rewritten now, so I'm going to dismiss all of the review comments, and pull this out of draft.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -470,12 +473,12 @@ class Client : public ClientBase
   : ClientBase(node_base, node_graph)
   {
     using rosidl_typesupport_cpp::get_service_type_support_handle;
-    auto service_type_support_handle =
+    service_type_support_handle_ =
```
mjcarroll, line 476: I think you can put this in the initializer list now.

rclcpp/test/rclcpp/test_service_introspection.cpp
```diff
@@ -0,0 +1,339 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
```
mjcarroll, line 1: Copyright year

rclcpp/include/rclcpp/service.hpp
```diff
@@ -311,7 +313,7 @@ class Service
   : ServiceBase(node_handle), any_callback_(any_callback)
   {
     using rosidl_typesupport_cpp::get_service_type_support_handle;
-    auto service_type_support_handle = get_service_type_support_handle<ServiceT>();
+    service_type_support_handle_ = get_service_type_support_handle<ServiceT>();
```
mjcarroll, line 316: Same here

rclcpp/include/rclcpp/client.hpp
clalancette: Good call, fixed in 01aa7e0

rclcpp/test/rclcpp/test_service_introspection.cpp
clalancette: Fixed in 01aa7e0

rclcpp/include/rclcpp/service.hpp
clalancette: Fixed in 01aa7e0.  Actually, I also noticed another problem here which was that we weren't setting up the handle for all Service class constructors.  I also fixed that.

mjcarroll (approved):

clalancette:
CI for this is in https://github.com/ros2/ros2/issues/1285#issuecomment-1442499653

PR #1988, opened 2022-08-11
operator+= and operator-= for Duration

tylerjw:
Signed-off-by: Tyler Weaver <maybe@tylerjw.dev>

### DIFF
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -92,8 +92,12 @@ public:
   Duration
   operator+(const rclcpp::Duration & rhs) const;
 
+  Duratoin& operator+=(Const rclcpp::Duration & rhs);
+
   Duration
   operator-(const rclcpp::Duration & rhs) const;
+
+  Duration& operator-=(Const rclcpp::Duration & rhs);
 
   /// Get the maximum representable value.
   /**
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -92,12 +92,12 @@ public:
   Duration
   operator+(const rclcpp::Duration & rhs) const;
 
-  Duratoin& operator+=(Const rclcpp::Duration & rhs);
+  Duration & operator+=(const rclcpp::Duration & rhs);
 
   Duration
   operator-(const rclcpp::Duration & rhs) const;
 
-  Duration& operator-=(Const rclcpp::Duration & rhs);
+  Duration & operator-=(const rclcpp::Duration & rhs);
 
   /// Get the maximum representable value.
   /**
@@ -109,6 +109,27 @@ public:
 
   Duration
   operator*(double scale) const;
+
+  Duration &
+  operator*=(double scale);
+
+  Duration
+  operator*(const rclcpp::Duration & rhs) const;
+
+  Duration &
+  operator*=(const rclcpp::Duration & rhs);
+
+  Duration
+  operator/(double scale) const;
+
+  Duration &
+  operator/=(double scale);
+
+  Duration
+  operator/(const rclcpp::Duration & rhs) const;
+
+  Duration &
+  operator/=(const rclcpp::Duration & rhs);
 
   /// Get duration in nanosecods
   /**
--- a/rclcpp/include/rclcpp/duration.hpp
+++ b/rclcpp/include/rclcpp/duration.hpp
@@ -113,24 +113,6 @@ public:
   Duration &
   operator*=(double scale);
 
-  Duration
-  operator*(const rclcpp::Duration & rhs) const;
-
-  Duration &
-  operator*=(const rclcpp::Duration & rhs);
-
-  Duration
-  operator/(double scale) const;
-
-  Duration &
-  operator/=(double scale);
-
-  Duration
-  operator/(const rclcpp::Duration & rhs) const;
-
-  Duration &
-  operator/=(const rclcpp::Duration & rhs);
-
   /// Get duration in nanosecods
   /**
    * \return the duration in nanoseconds as a rcl_duration_value_t.
--- a/rclcpp/src/rclcpp/duration.cpp
+++ b/rclcpp/src/rclcpp/duration.cpp
@@ -161,8 +161,8 @@ Duration::operator+(const rclcpp::Duration & rhs) const
     rcl_duration_.nanoseconds + rhs.rcl_duration_.nanoseconds);
 }
 
-Duration&
-Duration::operator+=(const rclcpp::Duration & rhs) const
+Duration &
+Duration::operator+=(const rclcpp::Duration & rhs)
 {
   *this = *this + rhs;
   return *this;
@@ -197,8 +197,8 @@ Duration::operator-(const rclcpp::Duration & rhs) const
     rcl_duration_.nanoseconds - rhs.rcl_duration_.nanoseconds);
 }
 
-Duration&
-Duration::operator-=(const rclcpp::Duration & rhs) const
+Duration &
+Duration::operator-=(const rclcpp::Duration & rhs)
 {
   *this = *this - rhs;
   return *this;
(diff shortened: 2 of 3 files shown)


rclcpp/include/rclcpp/duration.hpp
```diff
@@ -92,9 +92,13 @@ class RCLCPP_PUBLIC Duration
   Duration
   operator+(const rclcpp::Duration & rhs) const;
 
+  Duratoin& operator+=(Const rclcpp::Duration & rhs);
```
fujitatomoya, line 95:
```suggestion
  Duration& operator+=(const rclcpp::Duration & rhs);
```

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -92,9 +92,13 @@ class RCLCPP_PUBLIC Duration
...
+
   Duration
   operator-(const rclcpp::Duration & rhs) const;
 
+  Duration& operator-=(Const rclcpp::Duration & rhs);
```
fujitatomoya, line 100:
```suggestion
  Duration& operator-=(const rclcpp::Duration & rhs);
```

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -161,6 +161,13 @@ Duration::operator+(const rclcpp::Duration & rhs) const
     rcl_duration_.nanoseconds + rhs.rcl_duration_.nanoseconds);
 }
 
+Duration&
+Duration::operator+=(const rclcpp::Duration & rhs) const
```
fujitatomoya, line 165:
```suggestion
Duration::operator+=(const rclcpp::Duration & rhs)
```

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -190,6 +197,13 @@ Duration::operator-(const rclcpp::Duration & rhs) const
     rcl_duration_.nanoseconds - rhs.rcl_duration_.nanoseconds);
 }
 
+Duration&
+Duration::operator-=(const rclcpp::Duration & rhs) const
```
fujitatomoya, line 201:
```suggestion
Duration::operator-=(const rclcpp::Duration & rhs)
```

fujitatomoya (changes requested):
adding operators makes sense.
but this PR cannot even be built, can you address build failure and add some test?

alsora:
LGTM with the changes requested by Tomoya

tylerjw:
I fixed the compile errors, added other operators that I noticed were missing, and added tests for it all.

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -106,6 +110,27 @@ class RCLCPP_PUBLIC Duration
...
   operator*(double scale) const;
 
+  Duration &
+  operator*=(double scale);
+
+  Duration
+  operator*(const rclcpp::Duration & rhs) const;
+
+  Duration &
+  operator*=(const rclcpp::Duration & rhs);
```
fujitatomoya, line 120: I do not really understand what this means and usecase, e.g) seconds = seconds * seconds. having scale is not enough?

rclcpp/include/rclcpp/duration.hpp
```diff
@@ -106,6 +110,27 @@ class RCLCPP_PUBLIC Duration
...
+  operator/(double scale) const;
+
+  Duration &
+  operator/=(double scale);
+
+  Duration
+  operator/(const rclcpp::Duration & rhs) const;
+
+  Duration &
+  operator/=(const rclcpp::Duration & rhs);
```
fujitatomoya, line 132: same here, it could return the scale instead?, yet i am not really sure the use case for the application.

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -222,6 +245,67 @@ Duration::operator*(double scale) const
...
+
+Duration
+Duration::operator/(double scale) const
+{
+  if (!std::isfinite(scale)) {
+    throw std::runtime_error("abnormal scale in rclcpp::Duration");
+  }
+  bounds_check_duration_scale(
+    this->rcl_duration_.nanoseconds,
+    1.0 / scale,
```
fujitatomoya, line 280: scale could be zero, that leads to zero division runtime exception?

rclcpp/src/rclcpp/duration.cpp
```diff
@@ -222,6 +245,67 @@ Duration::operator*(double scale) const
...
+
+Duration &
+Duration::operator*=(const rclcpp::Duration & rhs)
+{
+  *this = *this * rhs;
+  return *this;
+}
+
+Duration
+Duration::operator/(double scale) const
```
fujitatomoya, line 273: user can do the same thing via `Duration::operator*(double scale) const`? scale can be like 0.5 (division by 2).

fujitatomoya (commented):
@tylerjw thank you for your contribution. i got a few comments, could you share your thoughts? I was thinking that basically `operator -/+` extension would be enough.

rclcpp/include/rclcpp/duration.hpp
tylerjw: This PR was prompted because I was trying to refactor RobotTrajectory in MoveIt to use rclcpp::Duration objects to represent differences in time between states that previously were using `double` seconds to represent time. 

I remembered seeing division and having to do it with seconds to get it to compile. Looking back at the code though it is assigned into a double scale value.

Regardless, as you pointed out in your other comment, I grew the scope of the PR here making it more complex so I'm backing it back to much closer to the original intent. I removed the `operator/` stuff from this PR and the `seconds * seconds` stuff.

rclcpp/src/rclcpp/duration.cpp
tylerjw: This is a good catch, I removed division from this PR so it can be separate.

tylerjw:
@fujitatomoya to make this PR easier I reduced the scope to only `+=, -=, *=`. I will open a second PR for `/, /=`. Thinking about the units though you are correct that it makes no sense to have `seconds * seconds` unless we had a `seconds ^ 2` type we could assign into.

tylerjw:
As we have operators that work on `double` type scale values, would it make sense to also have integer type scale values operators? If we did, would `uint64_t` be the right type to use?

With the current implementation if someone writes:

```cpp
auto x = rclcpp::Duration::from_seconds(2.3) * 33554435;
```

The cast of of the integer value `33554435` to double would result in `33554432 +/- 4`. While there is a `long double` type that would make this situation better (would require larger numbers to result in loss of precision) if we offered operators that took a large integer scale value the user would receive more precision. I don't have any examples in my code of huge integer scale values being multiplied by a duration so maybe this is too much of an edge case to care about?

rclcpp/include/rclcpp/duration.hpp
tylerjw: This is a good point, seconds * seconds = seconds^2, not seconds. I don't have any examples of doing this in my code and we don't have a seconds^2 type. Removed this from the PR.

fujitatomoya:
> The cast of of the integer value 33554435 to double would result in 33554432 +/- 4. While there is a long double type that would make this situation better (would require larger numbers to result in loss of precision) if we offered operators that took a large integer scale value the user would receive more precision. I don't have any examples in my code of huge integer scale values being multiplied by a duration so maybe this is too much of an edge case to care about?

thanks for the information.
technically this is correct. depends on decimal digits of precision, it loses some scale when it implicitly cast the value to `double`.
beside, internally it casts the scale into `long double` to keep the precision.

https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp/src/rclcpp/duration.cpp#L219-L222

but as you suggested, probably it is too much... i would not have the case either.
IMO, we would want to keep the interface as it is, instead of having the high precision for uncommon case?.

fujitatomoya (approved):

alsora (approved):

v4hn (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17211/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11754/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17654/)

PR #2001, opened 2022-08-24
[Galactic backport] Do not attempt to use void allocators for memory allocation

sykwer:
We need to fix the type of allocator passed to a publisher and a subscriber, otherwise, extra memory space is allocated at various parts of `rcl`.
For instance, a memory area of size `sizeof(rcutils_string_map_impl_t) * sizeof(MessageT)` is allocated [here](https://github.com/ros2/rcutils/blob/galactic/src/string_map.c#L62).

This bug not only causes extra memory consumption, but also causes `std::bad_alloc` in the publisher (and subscriber) that handles the message type when a huge message (larger than a few hundred MB) such as static size array is defined.

backport  #1657

Signed-off-by: Takahiro Ishikawa <sykwer@gmail.com>

### DIFF
--- a/rclcpp/include/rclcpp/publisher_options.hpp
+++ b/rclcpp/include/rclcpp/publisher_options.hpp
@@ -80,10 +80,7 @@ struct PublisherOptionsWithAllocator : public PublisherOptionsBase
   to_rcl_publisher_options(const rclcpp::QoS & qos) const
   {
     rcl_publisher_options_t result = rcl_publisher_get_default_options();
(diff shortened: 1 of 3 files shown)


fujitatomoya:
@sykwer thanks for bringing this up, instead of having dedicated backport MR, can https://github.com/ros2/rclcpp/pull/1657 backport to foxy / galactic straightforward? let me try.

clalancette:
Given that Galactic is now EOL, closing this.

PR #2002, opened 2022-08-24
Do not clear entities callbacks on destruction

mauropasse:
Removing these not-necessary clearings of listener callbacks, since the objects are being destroyed anyway.

This also fixes a bug (use-after-free) happening during services destruction:
https://github.com/ros2/rclcpp/blob/3d69031d2a614cf02dc10ea5db153471ea32b1f2/rclcpp/src/rclcpp/service.cpp#L35-L38

On detail:
```
ServiceBase::~ServiceBase()
{
  // Here, implicitly, the deleter of service_handle_ was being called (*),
  // freeing the service listener (**):
  // ~service_handle_
  
  // Then, the following was called, which tries to access the freed mem (thus seg-fault) (***)
  clear_on_new_request_callback();
}
```
(*) https://github.com/ros2/rclcpp/blob/3d69031d2a614cf02dc10ea5db153471ea32b1f2/rclcpp/include/rclcpp/service.hpp#L316-L327

(**) https://github.com/ros2/rmw_fastrtps/blob/dbee45ec67104db51b96da9693466f7a14e2f000/rmw_fastrtps_shared_cpp/src/rmw_service.cpp#L132

(***) https://github.com/ros2/rmw_fastrtps/blob/dbee45ec67104db51b96da9693466f7a14e2f000/rmw_fastrtps_shared_cpp/src/rmw_service.cpp#L170

FYI @alsora 

### DIFF
--- a/rclcpp/src/rclcpp/service.cpp
+++ b/rclcpp/src/rclcpp/service.cpp
@@ -32,10 +32,6 @@ ServiceBase::ServiceBase(std::shared_ptr<rcl_node_t> node_handle)
   node_logger_(rclcpp::get_node_logger(node_handle_.get()))
 {}
 
-ServiceBase::~ServiceBase()
-{
-  clear_on_new_request_callback();
-}
 
 bool
 ServiceBase::take_type_erased_request(void * request_out, rmw_request_id_t & request_id_out)
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -130,7 +130,7 @@ public:
     rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph);
 
   RCLCPP_PUBLIC
-  virtual ~ClientBase();
+  virtual ~ClientBase() = default;
 
   /// Take the next response for this client as a type erased pointer.
   /**
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
@@ -52,7 +52,7 @@ public:
   {}
 
   RCLCPP_PUBLIC
-  virtual ~SubscriptionIntraProcessBase();
+  virtual ~SubscriptionIntraProcessBase() = default;
 
   RCLCPP_PUBLIC
   size_t
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -55,7 +55,7 @@ public:
   explicit ServiceBase(std::shared_ptr<rcl_node_t> node_handle);
 
   RCLCPP_PUBLIC
-  virtual ~ServiceBase();
+  virtual ~ServiceBase() = default;
 
   /// Return the name of the service.
   /** \return The name of the service. */
--- a/rclcpp/src/rclcpp/client.cpp
+++ b/rclcpp/src/rclcpp/client.cpp
@@ -65,13 +65,6 @@ ClientBase::ClientBase(
     });
 }
 
-ClientBase::~ClientBase()
-{
-  clear_on_new_response_callback();
-  // Make sure the client handle is destructed as early as possible and before the node handle
-  client_handle_.reset();
-}
-
 bool
 ClientBase::take_type_erased_response(void * response_out, rmw_request_id_t & request_header_out)
 {
--- a/rclcpp/src/rclcpp/publisher_base.cpp
+++ b/rclcpp/src/rclcpp/publisher_base.cpp
@@ -102,11 +102,6 @@ PublisherBase::PublisherBase(
 
 PublisherBase::~PublisherBase()
 {
-  for (const auto & pair : event_handlers_) {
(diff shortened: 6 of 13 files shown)


rclcpp/src/rclcpp/service.cpp
```diff
@@ -32,10 +32,6 @@ ServiceBase::ServiceBase(std::shared_ptr<rcl_node_t> node_handle)
   node_logger_(rclcpp::get_node_logger(node_handle_.get()))
 {}
 
-ServiceBase::~ServiceBase()
-{
-  clear_on_new_request_callback();
```
fujitatomoya, line 37: `std::shared_ptr<rcl_service_t> service_handle_` will be deleted after `clear_on_new_request_callback()`, right? All member variables are alive in the body of the destructor. So this is something like optimization?

mauropasse: Yes, usually that's the order. But I found a strange case with lifecycle nodes destruction, where it happens in the opposite order!
Anyway, even if the order were correct, IMO there is no real point in clearing those callbacks.

fujitatomoya: i think that this should not be called here, because it might generate the exception in dtor.

see,

https://github.com/ros2/rmw/blob/rolling/rmw/include/rmw/rmw.h#L3118-L3131

if the callback is NULL, as specification, it is supposed to return RMW_RET_INVALID_ARGUMENT.
and eventually rclcpp will generate `rclcpp::exceptions::throw_from_rcl_error`.

test is pass cz rmw_fastrtps always return RCL_RET_OK. (this is not correct behavior based on the doc.)

if we use RTI Coneext DDS, the test fails as following, because it returns RMW_RET_UNSUPPORTED.

```bash
[ RUN      ] TestService.on_new_request_callback
RTI Connext DDS Non-commercial license is for academic, research, evaluation and personal use only. USE FOR COMMERCIAL PURPOSES IS PROHIBITED. See RTI_LICENSE.TXT for terms. Download free tools at rti.com/ncl. License issued to Non-Commercial User license@rti.com For non-production use only.
Expires on 00-jan-00 See www.rti.com for more information.
[ERROR] [1661535719.598830456] [rmw_connextdds]: rmw_service_set_on_new_request_callback not implemented
unknown file: Failure
C++ exception with description "failed to set the on new request callback for service: rmw_service_set_on_new_request_callback not implemented, at /root/ros2_ws/colcon_ws/src/ros2/rmw_connextdds/rmw_connextdds_common/src/common/rmw_listener.cpp:45" thrown in the test body.
[  FAILED  ] TestService.on_new_request_callback (315 ms)
```

after all, i think after this PR is merged, we need to re-check the listener rmw implementation and extend test for all rmw tier 1 implementation. what do you think?

fujitatomoya (approved):

rclcpp/src/rclcpp/service.cpp
mauropasse: Good, you found a better reason for removing these calls! And yes I agree, tests should be extended for these calls to return the other possible values too. When I found some time I'll add them. You have any other ideas for testing?

alsora:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17193/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11734/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17615/)

rclcpp/src/rclcpp/service.cpp
fujitatomoya: I think we can manage the following with different PR,

- test all tier 1 rmw implementations with listener APIs for rclcpp (rclpy does not support?)
- bugfix rmw implementations if necessary, i think `rmw_connextdds` is the correct implementation.
- test would be required to adjust accordingly.

fujitatomoya:
@mauropasse can you check the CI failure? i think those two are unrelated, but need to double-check.

mauropasse:
@fujitatomoya I reverted the changes for clearing callbacks on `~QOSEventHandlerBase`.
This object is created when creating a Publisher/Subscriber, so it's destructor will be called before the rmw publisher/subscribers/etc are destroyed.
So the events callbacks need to be cleared on `QOSEventHandlerBase` destructor. No risk of `rmw` entities destroyed before, since they are created prior of this object.

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17329/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11866/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17817/)

mauropasse:
I'll add a comment about why this clearing is needed here, and not in the other entities, but I'll wait for all green CI.

fujitatomoya:
@alsora can you review this? i am good to go with it.

alsora:
Looks good to me.
We have been testing this PR for the last couple of days and it seems to fix the problem.
I'm going to go ahead and merge it.

fujitatomoya:
@alsora thanks!

Issue #1990, opened 2022-08-12
A timer that can be triggered a finite number of times.


Voldivh:
#### Feature description

There can be situations in which you would like to call a function a finite number of times within the (same period of time) and not to be called endlessly. For those kind of situations the proposal is to give the timers the option to select the number of times that the timer is triggered. An example of use can be a "One Shot Timer" ([which was implemented in ROS 1](http://docs.ros.org/en/noetic/api/roscpp/html/structros_1_1TimerOptions.html)). 

#### Implementation considerations

Given that use case stated above, it is possible to add the feature by keeping track of the number of times the timer is being triggered and canceling it after the number of callbacks is completed. An example of this implementation can be found on the following fork:

- https://github.com/Voldivh/rclcpp/tree/voldivh/callback_counter

alsora:
IMO a feature like this can be easily implement via a wrapper around the `Timer` class (and this is also how I would recommend to implement it, without modifying the existing classes.

However, I think that as long as something can be easily implemented by a wrapper, we should think twice before including it in rclcpp.
The main reason is to avoid bloating rclcpp with any possible extension/variation of the ROS 2 basic entities.

A different story instead is any feature that would require to modify the internals of rclcpp in order to be implemented.

fujitatomoya:
i am with @alsora on this, i think that user can handle this in the user callback internally.

tfoote:
I don't think that this gets anywhere near the definition of bloat. It requires two extra internal unsigned int and conditionally increments them. The constructor gets an additional default value.

rclcpp is designed to be the user friendly native developer interface for ROS in C++. Providing this feature is a core capability of a Timer. If we don't add it here, we'll need to make a parallel class. And I don't think that it's clearer or more discoverable to have a parallel class which would just replace a default argument. 

The real implementation of this is actually at the lower level in rcl. And the rclcpp Timer class is just making a nice c++ API on top of that. This does not change the underlying fundamentals of any of the system. 

If we do not integrate this into the core Time implementation what would you like to see a parallel class for the wrapper?

We're also likely to soon need to also provide Timers that don't autostart as well. Which in ROS 1 was also a default parameter. Which if we want to support the various combinations sounds like a potentially very complex situation compared to having two default parameters that the user can set.

doganulus:
It is a bloat because the feature is easily implementable by the user and apparently not a common case. It's not about the size of the implementation.

[...]

PR #2006, opened 2022-08-26
Timer that can be triggered a finite number of times.

Voldivh:
This PR addresses the #1990. Basically it gives the timers API the possibility to select the number of times the timer will be triggered and be canceled after that using a variable. This variable (`amount_of_callbacks`) have the default value of `-1` which indicates that the timer will be triggered infinite times until it is 'manually' cancelled, i.e., the same behavior the times have at the moment. For example, initializing the timer with `amount_of_callbacks = 1`, would create a one shot timer, which would be triggered once and then would be canceled.

### DIFF
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -39,13 +39,15 @@ create_timer(
   rclcpp::Clock::SharedPtr clock,
   rclcpp::Duration period,
   CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  unsigned int amount_of_callbacks = 0)
 {
   auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
     clock,
     period.to_chrono<std::chrono::nanoseconds>(),
     std::forward<CallbackT>(callback),
-    node_base->get_context());
+    node_base->get_context(),
+    amount_of_callbacks);
 
   node_timers->add_timer(timer, group);
   return timer;
@@ -59,7 +61,8 @@ create_timer(
   rclcpp::Clock::SharedPtr clock,
   rclcpp::Duration period,
   CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  unsigned int amount_of_callbacks = 0)
 {
   return create_timer(
     rclcpp::node_interfaces::get_node_base_interface(node),
@@ -67,7 +70,8 @@ create_timer(
     clock,
     period,
     std::forward<CallbackT>(callback),
-    group);
+    group,
+    amount_of_callbacks);
 }
 
 /// Convenience method to create a timer with node resources.
@@ -81,6 +85,7 @@ create_timer(
  * \param group
  * \param node_base
  * \param node_timers
+ * \param amount_of_callbacks Quantity of times the callback will be triggered.
  * \return
  * \throws std::invalid argument if either node_base or node_timers
  * are null, or period is negative or too large
@@ -92,7 +97,8 @@ create_wall_timer(
   CallbackT callback,
   rclcpp::CallbackGroup::SharedPtr group,
   node_interfaces::NodeBaseInterface * node_base,
-  node_interfaces::NodeTimersInterface * node_timers)
+  node_interfaces::NodeTimersInterface * node_timers,
+  unsigned int amount_of_callbacks = 0)
 {
   if (node_base == nullptr) {
     throw std::invalid_argument{"input node_base cannot be null"};
@@ -133,7 +139,7 @@ create_wall_timer(
   }
 
   auto timer = rclcpp::WallTimer<CallbackT>::make_shared(
-    period_ns, std::move(callback), node_base->get_context());
+    period_ns, std::move(callback), node_base->get_context(), amount_of_callbacks);
   node_timers->add_timer(timer, group);
   return timer;
 }
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -232,13 +232,15 @@ public:
    * \param[in] period Time interval between triggers of the callback.
    * \param[in] callback User-defined callback function.
    * \param[in] group Callback group to execute this timer's callback in.
+   * \param[in] amount_of_callbacks Quantity of times the callback will be triggered.
    */
   template<typename DurationRepT = int64_t, typename DurationT = std::milli, typename CallbackT>
   typename rclcpp::WallTimer<CallbackT>::SharedPtr
   create_wall_timer(
     std::chrono::duration<DurationRepT, DurationT> period,
     CallbackT callback,
-    rclcpp::CallbackGroup::SharedPtr group = nullptr);
+    rclcpp::CallbackGroup::SharedPtr group = nullptr,
+    unsigned int amount_of_callbacks = 0);
 
   /// Create and return a Client.
   /**
--- a/rclcpp/include/rclcpp/node_impl.hpp
+++ b/rclcpp/include/rclcpp/node_impl.hpp
@@ -110,14 +110,16 @@ typename rclcpp::WallTimer<CallbackT>::SharedPtr
 Node::create_wall_timer(
   std::chrono::duration<DurationRepT, DurationT> period,
   CallbackT callback,
-  rclcpp::CallbackGroup::SharedPtr group)
+  rclcpp::CallbackGroup::SharedPtr group,
+  unsigned int amount_of_callbacks)
 {
   return rclcpp::create_wall_timer(
     period,
     std::move(callback),
     group,
     this->node_base_.get(),
-    this->node_timers_.get());
+    this->node_timers_.get(),
+    amount_of_callbacks);
 }
 
 template<typename ServiceT>
--- a/rclcpp/include/rclcpp/timer.hpp
+++ b/rclcpp/include/rclcpp/timer.hpp
@@ -179,13 +179,16 @@ public:
    * \param[in] period The interval at which the timer fires.
    * \param[in] callback User-specified callback function.
    * \param[in] context custom context to be used.
+   * \param[in] amount_of_callbacks Quantity of times the callback will be triggered.
    */
   explicit GenericTimer(
     Clock::SharedPtr clock, std::chrono::nanoseconds period, FunctorT && callback,
-    rclcpp::Context::SharedPtr context
+    rclcpp::Context::SharedPtr context, unsigned int amount_of_callbacks = 0
   )
-  : TimerBase(clock, period, context), callback_(std::forward<FunctorT>(callback))
+  : TimerBase(clock, period, context), callback_(std::forward<FunctorT>(callback)),
+    amount_of_callbacks_(amount_of_callbacks)
   {
+    callbacks_called_ = 0;
     TRACEPOINT(
       rclcpp_timer_callback_added,
       static_cast<const void *>(get_timer_handle().get()),
@@ -229,6 +232,11 @@ public:
     TRACEPOINT(callback_start, reinterpret_cast<const void *>(&callback_), false);
     execute_callback_delegate<>();
     TRACEPOINT(callback_end, reinterpret_cast<const void *>(&callback_));
+    if (amount_of_callbacks_ != 0) {
+      if (amount_of_callbacks_ == ++callbacks_called_) {
+        cancel();
+      }
+    }
   }
 
   // void specialization
@@ -268,6 +276,8 @@ protected:
   RCLCPP_DISABLE_COPY(GenericTimer)
 
   FunctorT callback_;
+  unsigned int callbacks_called_;
+  unsigned int amount_of_callbacks_;
 };
 
 template<
@@ -287,13 +297,16 @@ public:
    * \param period The interval at which the timer fires
    * \param callback The callback function to execute every interval
    * \param context node context
+   * \param amount_of_callbacks Quantity of times the callback will be triggered.
    */
   WallTimer(
     std::chrono::nanoseconds period,
     FunctorT && callback,
-    rclcpp::Context::SharedPtr context)
+    rclcpp::Context::SharedPtr context,
+    unsigned int amount_of_callbacks = 0)
   : GenericTimer<FunctorT>(
-      std::make_shared<Clock>(RCL_STEADY_TIME), period, std::move(callback), context)
+      std::make_shared<Clock>(RCL_STEADY_TIME), period, std::move(callback), context,
+      amount_of_callbacks)
   {}
 
 protected:
--- a/rclcpp/test/rclcpp/test_create_timer.cpp
+++ b/rclcpp/test/rclcpp/test_create_timer.cpp
@@ -133,3 +133,32 @@ TEST(TestCreateTimer, timer_function_pointer)
 
   rclcpp::shutdown();
 }
+
+TEST(TestCreateTimer, timer_triggered_twice)
+{
+  rclcpp::init(0, nullptr);
+  auto node = std::make_shared<rclcpp::Node>("test_timer_triggered_twice");
+
+  std::atomic<int> callback_counter{0};
+
+  rclcpp::TimerBase::SharedPtr timer;
+  timer = rclcpp::create_timer(
+    node,
+    node->get_clock(),
+    rclcpp::Duration(0ms),
+    [&callback_counter]() {
+      callback_counter += 1;
+    }, nullptr, 2);
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(1, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(2, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_NE(callback_counter == 3);
+  ASSERT_TRUE(timer->is_canceled());
+
+  rclcpp::shutdown();
+}
--- a/rclcpp/test/rclcpp/test_timer.cpp
+++ b/rclcpp/test/rclcpp/test_timer.cpp
@@ -283,3 +283,21 @@ TEST_F(TestTimer, test_failures_with_exceptions)
       std::runtime_error("Timer could not get time until next call: error not set"));
   }
 }
+
+TEST_F(TestTimer, test_timer_triggered_twice) {
+  std::atomic<int> callback_counter{0};
+
+  rclcpp::TimerBase::SharedPtr timer_called_twice;
+  timer_called_twice = test_node->create_wall_timer(
+    0ms,
+    [&callback_counter]() {
+      callback_counter += 1;
+    }, nullptr, 1);
+
+  executor->spin();
+  ASSERT_EQ(1, callback_counter);
+
+  executor->spin();
+  ASSERT_NE(callback_counter == 2);
+  ASSERT_TRUE(timer_called_twice->is_canceled());
+}


alsora:
This implementation adds overhead to existing timers.
We should make sure that the added functionality does not affect timers that don't want to use it.

Voldivh:
> This implementation adds overhead to existing timers. We should make sure that the added functionality does not affect timers that don't want to use it.

I agree. I thought about doing the implementation in another class, however, that option would imply copying all the code from an existing class and just modify a couple of lines and would also reduce the discoverability of the feature. What would you recommend to do in this case so that we can make sure the feature doesn't affect the existing timers?

tfoote:
With adding generality there is always going to be overhead. There are different dimensions of overhead, there's pure runtime, there's memory footprint, there's development effort, and there's api complexity all of which are side effects of design decisions. 

For core C++ libraries often they can avoid runtime overhead by doing template meta-programming and other advanced features. Another alternative would be if we were to change this API to be a factory pattern with generators that leverage different potential backend implementations. I don't think that's something that we want to get into.

But if I look at this for what the effect to the user will be. It's increasing the memory footprint by 2 unsigned integers. And in each timer cycle there's one if statement checking if an unsigned integer is non-zero. This is a very low overhead for having a generic API available.  It keeps the API simple and compact. And does not increase the code complexity significantly. If we are to add another class with a parallel implmentation it will likely have a higher memory impact as both classes will be in parallel in the shared libraries and then potentially loaded into memory too, with a total extra space of much greater than 2 unsigned ints of space. 

And looking a little bit closer the user's callback is invoked before the if statement so to the user the only potential slowdown is that the CPU has to do approximately one extra cycle prior to recomputing the timeout and enqueuing itself back into the waitset. As such the user will not have any effect on their timing either delaying their callback, nor delaying their next callback. With the exception that they overran their last callback time buffer and then it will be approximately 1 cpu clock cycle later if the compiler has not optimized out the check that this code path will clearly not take.

I don't think that there's a way to do this more efficiently that's both simple and easy for the user to leverage following our established API and design patterns as well as not adding significant extra lines of code for ongoing maintenance and testing. As such I would suggest going with this current quite simple solution.

alsora:
@tfoote IMO the ROS 2 design should be rigorous about how new features are added.

This "callback counter" is a convenience utility and, as we discussed in the issue, it's something that users could very easily do from the application code.
It looks like several people think that this is a very important feature and, as such, it should be available out of the box.
I'm ok with that, but this should not affect users that don't want to use it.

This particular PR is adding an overhead that is probably negligible, but still doing an exception for this PR today, then for another PR tomorrow and so is not good.
There are a lot of on-going efforts to make ROS 2 more efficient, but this is possible only if performance is taken in serious consideration at all the layers.

My recommendation is to handle this via a new `OneShotTimer` class that simply has to store the additional counter and override the `execute_callback` method.

alsora:
@tfoote @Voldivh  sorry for the short notice, but would you be available to have a discussion on this in the ROS 2 Client Library Working Group? It's today (August 31th) at 8AM Pacific Time:  https://meet.google.com/pxe-ohqa-cjm

tfoote:
> This particular PR is adding an overhead that is probably negligible, but still doing an exception for this PR today, then for another PR tomorrow and so is not good.

Saying that there might be another PR tomorrow that might slow down performance is a reason to not take this one, is not really a valid critique of this PR.

> there are a lot of on-going efforts to make ROS 2 more efficient, but this is possible only if performance is taken in serious consideration at all the layers.

Indeed, have there ever been any metrics at issue or proposed which this change will effect adversely? What is a metric on the performance of this PR that is acceptable vs unacceptable? As this is in the non-zero case an unchanging variable I suspect that the compiler will likely even detect that as such and this if statement never get even evaluated.

Are there any known or perceived cases when this will ever become a bottleneck to computation that we will need to optimize? If not this  seems like a case of premature optimization for runtime overhead at the cost of many other aspects of the system.

Forcing things into separate classes and implementations has notable other deleterious effects such as slowing down developer/user velocity and increasing our support burden with the number of lines of code, tests and documentation to maintain. These are all real costs which will be born out by both us and users in the community and trades off one form of overhead for another.

> I'm ok with that, but this should not affect users that don't want to use it (this should apply to all features, but in particular to utilities).

This is a great design goal but it cannot be applied in the abstract on a single metric. All of the above costs I've outlined above also effect the entire community.

> it's also not a good design practice as there isn't a clear separation between the c++ core (i.e. a wrapper around rcl) and the all the stuff built on top of it.

I think that this is one of the fundamental area where we're looking at this differently. I see the rclcpp API as a user facing API that should be designed to optimize the development of the end user. To the end user there is absolutely no difference whether the code is implemented under the hood using rcl or locally implemented in rclcpp. If under the hood in the implementation you want to make a distinction as to how it's implemented that's great. But it should not be something the the user sees or cares about. If we were to extend the functionality that is implemented in rcl and change the backend of that feature in rclcpp, an rclcpp API user should not notice or care. rclcpp's public API  is an abstraction which should hide all of those underlying implementation details.

tfoote:
So I was thinking about this a bit last night. And I think that a better solution that can both keep the default API generic and fully featured and separately maintain a highly optimized path. More concretely we will have the default Timer with all the features and it will have slightly higher overhead, and separately we have the LowOverheadTimer(naming TBD, using this for now for clarity) which is highly optimized and has a less user centric API that developers who are worried about the runtime overhead can opt into the more optimized one. If we're worried about a single if statement I suspect that there's more optimizations that can be achieved by stripping out more of the layers that already exist.

This will then allow the majority of users who are not going to be effected by the overhead to have the familiar API with all the bells and whistles. And the small fraction of users who want/need to optimize things can choose to use the lower level interface which does not need to be as cleanly presented or fully featured which will allow even more runtime optimizations than we already offer.

alsora:
I like this proposal @tfoote!
This definitely goes into the direction that I'm pushing for the ROS 2 libraries: expose both user-friendly APIs as well as optimized bare-bone primitives.
The latter should also be used internally in the core library.

I really want to apologize if my comments before were sounding too strong, in hindsightI shouldn't have started this discussion on this very PR, but rather just started it separately.
However, I'm glad that we had some good discussions on these very important topics!

One of the features developed by iRobot is the Events Executor.
The main reason why this executor is more efficient than the default one is exactly because it skips all the layers, which currently are not zero-overhead wrappers.
There's a lot of possible optimizations in the ROS stack, to implement them it's really fundamental to have a clear design in mind that makes ROS usable both by people who don't care about performance and resource management (and are happy to delegate it to the ROS stack) and by people who on the other hand must have full control over the operations to run them on resource constrained hardware.

fujitatomoya:
@tfoote @alsora thanks for sharing your thoughts.

IMO, this feature is basic control for timer, this is acceptable and reasonable.
I think priority of `rclcpp` is what to provide `interface` for user application, and then we can review how efficiently we can implement?

> If we're worried about a single if statement I suspect that there's more optimizations that can be achieved by stripping out more of the layers that already exist.

This is so true. Is this something we really want to do as design driver? I am just confused that user client library driver...
If we are worried about single `if` statement, i think we have clear complexity threshold on specific platform as requirement, probably we would not even use `rclcpp`. (i think this is one of the driver of https://github.com/ros2/rclc.)

It will be really difficult to support new feature in `rclcpp`, if the performance comes 1st priority.
and what about the other client library like `rclpy`? the same driver should be applied as user client library?

Just a bit confused about where this is going to land.

thanks!

fujitatomoya:
CC: @ros2/team

clalancette:
> It will be really difficult to support new feature in `rclcpp`, if the performance comes 1st priority.

So given where we are in the lifecycle in ROS 2, my opinion here is that performance doesn't need to be 1st priority, but it definitely needs to be *a* priority.  We've heard a lot of grumbling from the community about how the performance of ROS 2 is worse than ROS 1 in a lot of ways.  While we have a lot of knobs that people can use to make the system more performant, the fact of the matter is that out-of-the-box, ROS 2 currently isn't great at this.  So any new features that go in need to be weighed against this fact.

That said, for this particular feature, I agree that the cost of an extra branch is likely going to be lost in the noise here.  There's an easy way to test; do some performance testing before and after this PR with timers.  If we can't measure the overhead (as I suspect), then I would be for this PR going in more-or-less as-is.

alsora:
Our analysis showed that one of the major CPU bottlenecks of ROS 2 is caused by how its abstraction layers are implemented.
Getting data from the user application to the DDS and viceversa is very expensive (these should be almost negligible, we should be just passing a pointer around).
The source of the overhead is not one big, expensive operation (i.e. something that would be easily detected if we had performance measurements in the PR process), but rather it comes from summing together a lot of if/else, checks and other stuff that on its own has a negligible impact on performance.

While developing the events executor, we did some tests where we invoked the DDS APIs as close as possible to the application layer.
This is obviously bad from an architecture point of view, but it improved the performance a lot.

The "fast-path" i.e. that sequence of operations that is invoked at very high frequency while an application is running has to be cleaned from all operations that aren't required.
We all love utilities and wrappers, but they should be implemented on top of efficient primitives, which are also accessible to users if they need them.

This PR is not a problem per se, and merging it will have no impact on performance for 99.9% of the users (me included), however it's a clear example of how the ROS 2 design does not follow the "you don't pay for what you don't use" principle in its fast-path
Users that want a timer that runs forever will have to check that if-condition every time, similarly, users with an executor with only a `rclcpp::Waitable` have to go through ~20 if-conditions every time the `Executor::get_next_ready_executable_from_map` function is invoked https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/executor.cpp#L826.
These are just examples and there are a lot of them unfortunately.

User-friendly, feature-rich APIs and efficient primitives are both equally important requirements for ROS 2, but they should be built one on top of the other rather than be mixed together.

Voldivh:
After reading all the discussion and the common ground reached regarding the software design I ran some simple test on my local host to check the overhead this PR would include. I basically created a timer with a callback and checked if the timer was being fired on time (according to the period) using `std::chrono::high_resolution_clock` as the watchdog of the timer. Also I kept looking at the general CPU and memory load of the process with the feature proposed in this PR and without it (i.e. rolling). Here are the results I got on my local host:

As for the timer being triggered on time, I only ran it for a few callbacks and all were triggered within the time period:
```
eloy@567080f97543:~/ros2_rolling_ws$ ros2 run testing_pkg timer 
Callbacks called: 100
Missed Deadlines: 0
```

As for the workload on the CPU and the memory usage, take a look at the **yellow row** from the next pictures:
#### Feature Branch
image

#### Rolling
image

As we expected, the CPU/memory consumption is basically the same for both cases. I know this was a simple test to run that does not show an actual benchmark test for the feature. However, I believe it shows that the overhead is so small that it doesn't affect the behavior of the timers.

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -229,6 +232,11 @@ class GenericTimer : public TimerBase
     TRACEPOINT(callback_start, reinterpret_cast<const void *>(&callback_), false);
     execute_callback_delegate<>();
     TRACEPOINT(callback_end, reinterpret_cast<const void *>(&callback_));
+    if (amount_of_callbacks_ != 0) {
+      if (amount_of_callbacks_ == ++callbacks_called_) {
```
fujitatomoya, line 236: `callbacks_called_` will never be reset after complete the `amount_of_callbacks_`?
in this case, user application must destroy this timer object to run the same timer again.
besides,if user cancels the timer along the way, user should be able to know how many times left when it reset(restart)?

Or, if `calcel` is called, it will reset `ccallbacks_called_`? this could be better user experience and consistent behavior?

rclcpp/test/rclcpp/test_create_timer.cpp
```diff
@@ -193,3 +193,32 @@ TEST(TestCreateTimer, timer_function_pointer)
...
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(1, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(2, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_NE(3, callback_counter);
+  ASSERT_TRUE(timer->is_canceled());
```
fujitatomoya, line 221: How about adding the test case to `reset` this timer object after this? I think user application might do that, and these cases need to be tested?

fujitatomoya (changes requested):

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -132,6 +136,7 @@ create_timer(
  * \param group callback group
  * \param node_base node base interface
  * \param node_timers node timer interface
+ * \param amount_of_callbacks Quantity of times the callback will be triggered.
```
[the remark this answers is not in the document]
Voldivh: Done.

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -90,15 +90,17 @@ create_timer(
   rclcpp::Clock::SharedPtr clock,
   rclcpp::Duration period,
   CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  unsigned int amount_of_callbacks = 0)
```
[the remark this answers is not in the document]
Voldivh: Done.

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -268,6 +276,8 @@ class GenericTimer : public TimerBase
   RCLCPP_DISABLE_COPY(GenericTimer)
 
   FunctorT callback_;
+  unsigned int callbacks_called_;
+  unsigned int amount_of_callbacks_;
```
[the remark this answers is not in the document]
Voldivh: I decided to go with uint_32_t. Done.

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -229,6 +232,11 @@ class GenericTimer : public TimerBase
     TRACEPOINT(callback_start, reinterpret_cast<const void *>(&callback_), false);
     execute_callback_delegate<>();
     TRACEPOINT(callback_end, reinterpret_cast<const void *>(&callback_));
+    if (amount_of_callbacks_ != 0) {
+      if (amount_of_callbacks_ == ++callbacks_called_) {
+        cancel();
```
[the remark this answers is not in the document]
Voldivh: Sorry to ask, this documentation where should be coded? On top of the `execute_callback()` method?

rclcpp/test/rclcpp/test_create_timer.cpp
Voldivh: I extended the test case previously coded to approach the behavior after using the `reset()` method.

rclcpp/include/rclcpp/timer.hpp
Voldivh: I see your point. At the moment what happens is that `callbacks_called_` is not being modified by the `reset()` method, that would imply that the timer would go to an ongoing state and then cancel itself due to the conditions. I drafted a simple workflow to ilustrate what is happening right now.
image

I think this actually doesn't make sense, thinking from the user point of view, what I would expect if I reset a "one-shot timer" for example is to get the same "one-shot timer" that triggers once. I think the workflow of the feature should be something like the following:
image

Taking that into account I would propose two possible solutions to get that behavior. One would be to reset the variable `callbacks_called_` after canceling the timer:
```
if (amount_of_callbacks_ != 0) {
    if (amount_of_callbacks_ <= ++callbacks_called_) {
      cancel();
      callbacks_called = 0;
    }
}
```
An the other option I would propose is to reset the value of `callbacks_called` on the `reset()` method, that would mean to modify that method to something like this:
```
void
TimerBase::reset()
{
  rcl_ret_t ret = rcl_timer_reset(timer_handle_.get());
  callbacks_called = 0;
  if (ret != RCL_RET_OK) {
    rclcpp::exceptions::throw_from_rcl_error(ret, "Couldn't reset timer");
  }
}
```

Let me know what you think regarding the behavior this feature should have after using `reset()` on the timer and how we could accomplish that.

rclcpp/include/rclcpp/timer.hpp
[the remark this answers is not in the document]
fujitatomoya: How about header file doc section?

rclcpp/include/rclcpp/timer.hpp
fujitatomoya: @Voldivh really appreciate the detailed flow diagram and explanation, this takes time but really good for understanding for everyone 🎉 

> reset the value of callbacks_called on the reset() method

i would take this proposal,

- `reset` makes more sense in general semantic? it actually `reset` the timer. (i may be wrong, I am not native English speaker... 😅  )
- probably it would be nice to keep the `state` after `cancel` for more extension for future, such as interface to let user application to access the `state` of the timer. (like how many times has callback been fired on this timer?)

rclcpp/include/rclcpp/node.hpp
```diff
@@ -232,13 +232,15 @@ class Node : public std::enable_shared_from_this<Node>
...
+   * \param[in] amount_of_callbacks Quantity of times the callback will be triggered.
    */
   template<typename DurationRepT = int64_t, typename DurationT = std::milli, typename CallbackT>
   typename rclcpp::WallTimer<CallbackT>::SharedPtr
   create_wall_timer(
     std::chrono::duration<DurationRepT, DurationT> period,
     CallbackT callback,
-    rclcpp::CallbackGroup::SharedPtr group = nullptr);
+    rclcpp::CallbackGroup::SharedPtr group = nullptr,
+    unsigned int amount_of_callbacks = 0);
```
alsora, line 243: this variable is of type `uint32_t` in all the other instances, please use the same type also here for consistency

Voldivh: Sure thing. Done.

rclcpp/include/rclcpp/timer.hpp
Voldivh: I had to modify add a member to the TimerBase class in order to do it. Please take a look.

rclcpp/include/rclcpp/timer.hpp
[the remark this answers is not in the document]
Voldivh: Done, Take a look and let me know what you think.

rclcpp/test/rclcpp/test_create_timer.cpp
```diff
@@ -193,3 +193,44 @@ TEST(TestCreateTimer, timer_function_pointer)
...
+    }, nullptr, 2);
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(1, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(2, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_NE(3, callback_counter);
```
petioptrv, line 220:
```suggestion
  ASSERT_EQ(2, callback_counter); // not incremented
```
I think the assertion here should also be that `callback_counter` is equal to 2. The current assertion would succeed if the counter were equal to 4, but that is not the expected bahaviour.

rclcpp/test/rclcpp/test_create_timer.cpp
```diff
@@ -193,3 +193,44 @@ TEST(TestCreateTimer, timer_function_pointer)
...
+  timer->reset();
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(3, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_EQ(4, callback_counter);
+
+  rclcpp::spin_some(node);
+  ASSERT_NE(5, callback_counter);
```
petioptrv, line 232:
```suggestion
  ASSERT_EQ(4, callback_counter); // not incremented
```
Same rationale here.

rclcpp/test/rclcpp/test_timer.cpp
```diff
@@ -334,3 +334,30 @@ INSTANTIATE_TEST_SUITE_P(
     return std::string("unknown");
   }
 );
+
+TEST_P(TestTimer, test_timer_triggered_twice) {
```
petioptrv, line 338:
```suggestion
TEST_P(TestTimer, test_timer_triggered_once) {
```

rclcpp/test/rclcpp/test_timer.cpp
```diff
@@ -334,3 +334,30 @@ INSTANTIATE_TEST_SUITE_P(
...
+    0ms,
+    [&callback_counter]() {
+      callback_counter += 1;
+    }, nullptr, 1);
+
+  executor->spin();
+  ASSERT_EQ(1, callback_counter);
+
+  executor->spin();
+  ASSERT_NE(2, callback_counter);
```
petioptrv, line 352:
```suggestion
  ASSERT_EQ(1, callback_counter); // not incremented
```
same rationale as above.

rclcpp/test/rclcpp/test_timer.cpp
```diff
@@ -334,3 +334,30 @@ INSTANTIATE_TEST_SUITE_P(
...
+  ASSERT_TRUE(timer_called_twice->is_canceled());
+
+  timer_called_twice->reset();
+
+  executor->spin();
+  ASSERT_EQ(2, callback_counter);
+
+  executor->spin();
+  ASSERT_NE(3, callback_counter);
```
petioptrv, line 361:
```suggestion
  ASSERT_EQ(2, callback_counter); // not incremented
```
same rationale as above.

rclcpp/test/rclcpp/test_create_timer.cpp
fujitatomoya: right, that is precise expectation here to check.

fujitatomoya (changes requested):

fujitatomoya:
@Voldivh sorry to be late to get back on this, can you take a look at latest comments?

Voldivh:
> @Voldivh sorry to be late to get back on this, can you take a look at latest comments?

Of course! I'll take a look.

rclcpp/test/rclcpp/test_timer.cpp
Voldivh: Done.

rclcpp/test/rclcpp/test_create_timer.cpp
Voldivh: I agree, this suggestion actually makes a lot more sense. Thanks and done.

fujitatomoya (approved):
lgtm

fujitatomoya:
@Voldivh could you address the conflicts?

Voldivh:
> @Voldivh could you address the conflicts?

Sure thing, will do.

fujitatomoya:
same here, build failed https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/906/console

```bash
10:51:52 /tmp/ws/src/rclcpp/rclcpp/include/rclcpp/subscription.hpp:208:5: error: there are no arguments to ‘TRACETOOLS_TRACEPOINT’ that depend on a template parameter, so a declaration of ‘TRACETOOLS_TRACEPOINT’ must be available [-fpermissive]
10:51:52   208 |     TRACETOOLS_TRACEPOINT(
10:51:52       |     ^~~~~~~~~~~~~~~~~~~~~
10:51:52 gmake[2]: *** [CMakeFiles/rclcpp.dir/build.make:181: CMakeFiles/rclcpp.dir/src/rclcpp/any_executable.cpp.o] Error 1
10:51:52 gmake[1]: *** [CMakeFiles/Makefile2:138: CMakeFiles/rclcpp.dir/all] Error 2
```

Voldivh:
> same here, build failed https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/906/console
> 
> ```shell
> 10:51:52 /tmp/ws/src/rclcpp/rclcpp/include/rclcpp/subscription.hpp:208:5: error: there are no arguments to ‘TRACETOOLS_TRACEPOINT’ that depend on a template parameter, so a declaration of ‘TRACETOOLS_TRACEPOINT’ must be available [-fpermissive]
> 10:51:52   208 |     TRACETOOLS_TRACEPOINT(
> 10:51:52       |     ^~~~~~~~~~~~~~~~~~~~~
> 10:51:52 gmake[2]: *** [CMakeFiles/rclcpp.dir/build.make:181: CMakeFiles/rclcpp.dir/src/rclcpp/any_executable.cpp.o] Error 1
> 10:51:52 gmake[1]: *** [CMakeFiles/Makefile2:138: CMakeFiles/rclcpp.dir/all] Error 2
> ```

Same question here with this [commit](https://github.com/ros2/rclcpp/commit/4efc05266b4fb19569e712648c092b32d3e8005e).

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18880/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13418/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19636/)

(Rpr job is supposed to fail.)

fujitatomoya:
@alsora could you review your comment?

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -91,7 +91,8 @@ create_timer(
   rclcpp::Duration period,
   CallbackT && callback,
   rclcpp::CallbackGroup::SharedPtr group = nullptr,
-  bool autostart = true)
+  bool autostart = true,
+  uint32_t amount_of_callbacks = 0)
```
clalancette, line 95:
```suggestion
  uint32_t number_of_callbacks = 0)
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -100,7 +101,8 @@ create_timer(
     group,
     node_base.get(),
     node_timers.get(),
-    autostart);
+    autostart,
+    amount_of_callbacks);
```
clalancette, line 105:
```suggestion
    number_of_callbacks);
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -112,7 +114,8 @@ create_timer(
   rclcpp::Duration period,
   CallbackT && callback,
   rclcpp::CallbackGroup::SharedPtr group = nullptr,
-  bool autostart = true)
+  bool autostart = true,
+  uint32_t amount_of_callbacks = 0)
```
clalancette, line 118:
```suggestion
  uint32_t number_of_callbacks = 0)
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -121,7 +124,8 @@ create_timer(
     group,
     rclcpp::node_interfaces::get_node_base_interface(node).get(),
     rclcpp::node_interfaces::get_node_timers_interface(node).get(),
-    autostart);
+    autostart,
+    amount_of_callbacks);
```
clalancette, line 128:
```suggestion
    number_of_callbacks);
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -137,6 +141,8 @@ create_timer(
  * \param node_base node base interface
  * \param node_timers node timer interface
  * \param autostart defines if the timer should start it's countdown on initialization or not.
+ * \param amount_of_callbacks Quantity of times the callback will be triggered. The default value is 0
+ * that means an infinite amount of callbacks i.e. the clock's default behavior.
```
clalancette, line 145:
```suggestion
 * \param number_of_callbacks number of times the callback will be triggered. The default value is 0
 * that means an infinite amount of callbacks i.e. the timer's default behavior.
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -150,7 +156,8 @@ create_timer(
   rclcpp::CallbackGroup::SharedPtr group,
   node_interfaces::NodeBaseInterface * node_base,
   node_interfaces::NodeTimersInterface * node_timers,
-  bool autostart = true)
+  bool autostart = true,
+  uint32_t amount_of_callbacks = 0)
```
clalancette, line 160:
```suggestion
  uint32_t number_of_callbacks = 0)
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -166,7 +173,8 @@ create_timer(
 
   // Add a new generic timer.
   auto timer = rclcpp::GenericTimer<CallbackT>::make_shared(
-    std::move(clock), period_ns, std::move(callback), node_base->get_context(), autostart);
+    std::move(clock), period_ns, std::move(callback), node_base->get_context(), autostart,
+    amount_of_callbacks);
```
clalancette, line 177:
```suggestion
    number_of_callbacks);
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -194,7 +202,8 @@ create_wall_timer(
   rclcpp::CallbackGroup::SharedPtr group,
   node_interfaces::NodeBaseInterface * node_base,
   node_interfaces::NodeTimersInterface * node_timers,
-  bool autostart = true)
+  bool autostart = true,
+  uint32_t amount_of_callbacks = 0)
```
clalancette, line 206:
```suggestion
  uint32_t number_of_callbacks = 0)
```

rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -208,7 +217,7 @@ create_wall_timer(
 
   // Add a new wall timer.
   auto timer = rclcpp::WallTimer<CallbackT>::make_shared(
-    period_ns, std::move(callback), node_base->get_context(), autostart);
+    period_ns, std::move(callback), node_base->get_context(), autostart, amount_of_callbacks);
```
clalancette, line 220:
```suggestion
    period_ns, std::move(callback), node_base->get_context(), autostart, number_of_callbacks);
```

rclcpp/include/rclcpp/node.hpp
```diff
@@ -233,14 +233,16 @@ class Node : public std::enable_shared_from_this<Node>
    * \param[in] callback User-defined callback function.
    * \param[in] group Callback group to execute this timer's callback in.
    * \param[in] autostart The state of the clock on initialization.
+   * \param[in] amount_of_callbacks Quantity of times the callback will be triggered.
```
clalancette, line 236:
```suggestion
   * \param[in] number_of_callbacks Number of times the callback will be triggered.
```

rclcpp/include/rclcpp/node.hpp
```diff
@@ -233,14 +233,16 @@ class Node : public std::enable_shared_from_this<Node>
...
    */
   template<typename DurationRepT = int64_t, typename DurationT = std::milli, typename CallbackT>
   typename rclcpp::WallTimer<CallbackT>::SharedPtr
   create_wall_timer(
     std::chrono::duration<DurationRepT, DurationT> period,
     CallbackT callback,
     rclcpp::CallbackGroup::SharedPtr group = nullptr,
-    bool autostart = true);
+    bool autostart = true,
+    uint32_t amount_of_callbacks = 0);
```
clalancette, line 245:
```suggestion
    uint32_t number_of_callbacks = 0);
```

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -111,15 +111,17 @@ Node::create_wall_timer(
   std::chrono::duration<DurationRepT, DurationT> period,
   CallbackT callback,
   rclcpp::CallbackGroup::SharedPtr group,
-  bool autostart)
+  bool autostart,
+  uint32_t amount_of_callbacks)
```
clalancette, line 115:
```suggestion
  uint32_t number_of_callbacks)
```

rclcpp/include/rclcpp/node_impl.hpp
```diff
@@ -111,15 +111,17 @@ Node::create_wall_timer(
...
 {
   return rclcpp::create_wall_timer(
     period,
     std::move(callback),
     group,
     this->node_base_.get(),
     this->node_timers_.get(),
-    autostart);
+    autostart,
+    amount_of_callbacks);
```
clalancette, line 124:
```suggestion
    number_of_callbacks);
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -202,7 +203,10 @@ class TimerBase
 using VoidCallbackType = std::function<void ()>;
 using TimerCallbackType = std::function<void (TimerBase &)>;
 
-/// Generic timer. Periodically executes a user-specified callback.
+/// Generic timer. Periodically executes a user-specified callback. The user
+/// can specify a finite amount of times that the callback will be executed
+/// and the timer will cancel itself after that. Otherwise, it will keep exe-
+/// cuting the callback on each period.
```
clalancette, line 209:
```suggestion
/// and the timer will cancel itself after that. Otherwise, it will keep
/// executing the callback on each period.
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -221,14 +225,18 @@ class GenericTimer : public TimerBase
    * \param[in] period The interval at which the timer fires.
    * \param[in] callback User-specified callback function.
    * \param[in] context custom context to be used.
-   * \param autostart timer state on initialization
+   * \param[in] autostart timer state on initialization
+   * \param[in] amount_of_callbacks Quantity of times the callback will be triggered.
```
clalancette, line 229:
```suggestion
   * \param[in] number_of_callbacks Number of times the callback will be triggered.
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -221,14 +225,18 @@ class GenericTimer : public TimerBase
...
    */
   explicit GenericTimer(
     Clock::SharedPtr clock, std::chrono::nanoseconds period, FunctorT && callback,
-    rclcpp::Context::SharedPtr context, bool autostart = true
+    rclcpp::Context::SharedPtr context, bool autostart = true, 
+    uint32_t amount_of_callbacks = 0
```
clalancette, line 234:
```suggestion
    uint32_t number_of_callbacks = 0
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -221,14 +225,18 @@ class GenericTimer : public TimerBase
...
   )
-  : TimerBase(clock, period, context, autostart), callback_(std::forward<FunctorT>(callback))
+  : TimerBase(clock, period, context, autostart), callback_(std::forward<FunctorT>(callback)),
+    amount_of_callbacks_(amount_of_callbacks)
```
clalancette, line 237:
```suggestion
    number_of_callbacks_(number_of_callbacks)
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -271,13 +279,23 @@ class GenericTimer : public TimerBase
...
+   * will always be triggered on each period.
    */
   void
   execute_callback() override
   {
     TRACETOOLS_TRACEPOINT(callback_start, reinterpret_cast<const void *>(&callback_), false);
     execute_callback_delegate<>();
     TRACETOOLS_TRACEPOINT(callback_end, reinterpret_cast<const void *>(&callback_));
+    if (amount_of_callbacks_ != 0) {
+      if (amount_of_callbacks_ <= ++callbacks_called_) {
```
clalancette, line 295:
```suggestion
    if (number_of_callbacks_ != 0) {
      if (number_of_callbacks_ <= ++callbacks_called_) {
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -317,6 +335,7 @@ class GenericTimer : public TimerBase
   RCLCPP_DISABLE_COPY(GenericTimer)
 
   FunctorT callback_;
+  uint32_t amount_of_callbacks_;
```
clalancette, line 338:
```suggestion
  uint32_t number_of_callbacks_;
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -337,14 +356,17 @@ class WallTimer : public GenericTimer<FunctorT>
    * \param callback The callback function to execute every interval
    * \param context node context
    * \param autostart timer state on initialization
+   * \param amount_of_callbacks Quantity of times the callback will be triggered.
```
clalancette, line 359:
```suggestion
   * \param number_of_callbacks Number of times the callback will be triggered.
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -337,14 +356,17 @@ class WallTimer : public GenericTimer<FunctorT>
...
    */
   WallTimer(
     std::chrono::nanoseconds period,
     FunctorT && callback,
     rclcpp::Context::SharedPtr context,
-    bool autostart = true)
+    bool autostart = true,
+    uint32_t amount_of_callbacks = 0)
```
clalancette, line 366:
```suggestion
    uint32_t number_of_callbacks = 0)
```

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -337,14 +356,17 @@ class WallTimer : public GenericTimer<FunctorT>
...
   : GenericTimer<FunctorT>(
-      std::make_shared<Clock>(RCL_STEADY_TIME), period, std::move(callback), context, autostart)
+      std::make_shared<Clock>(RCL_STEADY_TIME), period, std::move(callback), context, autostart,
+      amount_of_callbacks)
```
clalancette, line 369:
```suggestion
      number_of_callbacks)
```

rclcpp/test/rclcpp/test_create_timer.cpp
```diff
@@ -216,6 +256,6 @@ TEST(TestCreateTimer, timer_without_autostart)
   EXPECT_FALSE(timer->is_canceled());
 
   timer->cancel();
-
+
```
clalancette, line 259: Unnecessary whitespace change.

clalancette (changes requested):

Voldivh:
> as we talked before, this makes more sense if the canceled timer is removed from the `wait_set` for performance. AFAIK, the canceled timer is NOT removed from the `wait_set`, those timers are just skipped to execute the callback.
> 
> IMO, i think this PR is still useful for syntax sugar for application.

I took a closer look at the implementation of the cancel method and you are right, a canceled timer is not removed from the `wait_set` it is skipped after checking it's state (canceled or not canceled).

Taking that into account, I do feel that the PR would make more sense if it removes the timer from the `wait_set`. So far, I've only thought of destroying the timer object after the callbacks are completed, but, would be open to other ideas.

clalancette:
> Taking that into account, I do feel that the PR would make more sense if it removes the timer from the `wait_set`. So far, I've only thought of destroying the timer object after the callbacks are completed, but, would be open to other ideas.

I don't think we should destroy the timer object.  In particular, I can imagine that the user may want to retrigger the same timer multiple times, which they should be able to do by calling the `reset` method.  Destroying the object would prevent that.

In order to remove it from the waitset, I think we'd need to do something more clever where we track inside of the timer object whether we are currently in the waitset.  This would be `true` by default.  When we automatically expired via this mechanism, or `cancel` was called, we would remove ourselves from the waitset and set this as `false`.  The next time that `reset` was called, we would add ourselves back to the waitset and set this to `true`.

However, it's even more complicated than that, since at the `TimerBase` level we have no access to the executor.  So we'd somehow have to signal the executor to remove use from the waitset.  I'm not quite sure how we would do that.

Voldivh:
> > Taking that into account, I do feel that the PR would make more sense if it removes the timer from the `wait_set`. So far, I've only thought of destroying the timer object after the callbacks are completed, but, would be open to other ideas.
> 
> I don't think we should destroy the timer object. In particular, I can imagine that the user may want to retrigger the same timer multiple times, which they should be able to do by calling the `reset` method. Destroying the object would prevent that.
> 
> In order to remove it from the waitset, I think we'd need to do something more clever where we track inside of the timer object whether we are currently in the waitset. This would be `true` by default. When we automatically expired via this mechanism, or `cancel` was called, we would remove ourselves from the waitset and set this as `false`. The next time that `reset` was called, we would add ourselves back to the waitset and set this to `true`.
> 
> However, it's even more complicated than that, since at the `TimerBase` level we have no access to the executor. So we'd somehow have to signal the executor to remove use from the waitset. I'm not quite sure how we would do that.

Ok I see, as I feared, there is no easy way around it then. I'll take some time to look into it and get back to you.

PR #2019, opened 2022-10-04
check thread whether joinable before join

uupks:
If rclcpp is inited with rclcpp::SignalHandlerOptions::None,  signal_handler_thread will not be created.

Checking thread whether is joinable before join is a solution to this little bug.

### DIFF
--- a/rclcpp/src/rclcpp/signal_handler.cpp
+++ b/rclcpp/src/rclcpp/signal_handler.cpp
@@ -191,7 +191,9 @@ SignalHandler::uninstall()
     signal_handlers_options_ = SignalHandlerOptions::None;
     RCLCPP_DEBUG(get_logger(), "SignalHandler::uninstall(): notifying deferred signal handler");
     notify_signal_handler();
-    signal_handler_thread_.join();
+    if (signal_handler_thread_.joinable()) {
+      signal_handler_thread_.join();
+    }
     teardown_wait_for_signal();
   } catch (...) {
     installed_.exchange(true);


fujitatomoya (approved):
lgtm

fujitatomoya:
This is clearly bug, that could raise `std::system_error` with `Invalid argument`.
Test would be ideal.

fujitatomoya:
@clalancette could you take a look at this?

alsora (approved):
Looks correct to me.
+1 on adding a unit test

clalancette (approved):
This looks good to me, fixes an obvious bug.

While adding a test would be nice, it is not at all clear to me how to do that.  We don't have any existing tests for the `SignalHandler` class, and they would be tricky to write (particularly in a portable way).  So I'm fine with this going in as-is; I'll run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17485/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12012/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18033/)

clalancette:
The one failure on Windows is a known flake, so I'm going to go ahead with this.  Thanks for the contribution.

shuhaowu:
Is there a way to workaround this problem on Humble? The best way I've found is to override the signal handler with a custom one then call `rclcpp::shutdown`, but this hits this comment:

```
    // TODO(wjwwood): what happens if someone overrides our signal handler then calls uninstall?
    //   I think we need to assert that we're the current signal handler, and mitigate if not.
```

So the way to deal with this would be:

1. Register signal handler in rclcpp::init
2. Register your own signal handler which overrides the one in rclcpp::init
3. Let the custom signal handler handle the signal
4. Call rclcpp::shutdown, which will override the custom signal handler but it's OK at this point.

fujitatomoya:
@shuhaowu that could work, but straightforward way is to backport this to humble?

fujitatomoya:
@Mergifyio backport humble

PR #2023, opened 2022-10-07
optimize includes

julianreisch:
We removed unneeded includes where nothing is used from in the respective file. We made sure it compiles within ros2_foxy.


clalancette:
> We removed unneeded includes where nothing is used from in the respective file. We made sure it compiles within ros2_foxy.

Can you please sign-off the commit?  That will make the DCO bot happy.

Overall, this is a good change to have, and should cut down compilation times and memory usage a bit.  That said, if we only put this into Foxy, all of this will be lost on the next release.  My suggestion is to make a similar PR for Rolling first, and then we can consider this one for Foxy.

clalancette:
No response in a long time, so closing this out.  Do feel free to respond to my comments, and I'm happy to re-review.

Issue #2029, opened 2022-10-20
Possible deadlock on service which only happens if you have a single core (thus likely target or VM)


AlexisTM:
## Bug report

Calling a service client from a RO2 callback causes a deadlock.
The widely used solution is to make use of a Multithreaded Executor.
Yet, this will **fail** if the VM/target has only 1 core/1 core assigned as the default number of threads is the number of cores.

The biggest problem I see is for libraries requiring 2+ threads relying on the default ctor of the Multithreaded executor, which will **not work** for some targets. Having them needed to modify all third party executors to define the default number of threads to 2 instead of 0 will eventually cause problems. 

**Required Info:**

- Operating System:
  - Apertis Pro v2022 
- Installation type:
  - From source
- Version or commit hash:
  - 81d6f897a2a01d3a63354e9dc8cd4c8191ca789b
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
1- Create a single core VM
2- Have a multithreaded executor with default number of threads (0)
2- Use a service 
```

#### Expected behavior

We have multiple threads thus the deadlock is prevented.

#### Actual behavior

The Multithreaded executor uses a single thread as we have a single core on the machine. The application deadlocks.

#### Feature description
The Multithreaded executor defaults to 2+ when the number of threads is defined to 0.

#### Implementation considerations

When the number of threads is defined to 0, the Multithreaded Executor defaults to `std::max(nproc, 2);` number of threads.

Another solution is to default it to 2 to have a **reproducible behavior** thus developers defining the number of threads according to their needs instead of bumping to 128 threads per node for the latest thread-ripper and 1 on limited targets. 

When the speed is primordial and the number of threads is a feature, we could have an option such as the threads provided would be the **minimum number of threads** with a thread policy such as CONSERVATIVE or GREEDY. Node defined (the library maintainer knows better) or configuration defined (the library user knows better)

PR #2030, opened 2022-10-20
Set the minimum number of threads of the Multithreaded executor to 2

AlexisTM:
Signed-off-by: Alexis Paques <paa1ti@bosch.com>

Closes https://github.com/ros2/rclcpp/issues/2029

This ensures the number of threads of a **Multi-threaded executor** is at least 2 unless defined explicitly as 1 (why not use the SingleThreadedExecutor?)
Indeed, if we are calling a service from a callback and call the get function of the future directly, we effectively block until the service request resolves but it can't as the reply will not be processed, causing a deadlock. This is either solved by having a Multi-threaded executor and multiple callback groups for non time critical software (which has to wait for the result anyway).

The default number of threads is `std::thread::hardware_concurrency` which can be 1 for a VM or a small target. 
This means those targets (or VM) will **not** be able to receive the message of an otherwise correct node, while working on everybody else's machine.

The second motivation is to solve the misconception a Multithreaded executor _always_ has multiple thread, by making it have _always_ multiple treads by default.

### DIFF
--- a/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
@@ -21,6 +21,7 @@
 
 #include "rcpputils/scope_exit.hpp"
 
+#include "rclcpp/logging.hpp"
 #include "rclcpp/utilities.hpp"
 
 using rclcpp::executors::MultiThreadedExecutor;
@@ -36,9 +37,11 @@ MultiThreadedExecutor::MultiThreadedExecutor(
 {
   number_of_threads_ = number_of_threads
                            ? number_of_threads
-                           : std::max(std::thread::hardware_concurrency(), 2);
-  if (number_of_threads_ == 0) {
-    number_of_threads_ = 2;
+                           : std::max(std::thread::hardware_concurrency(), 2U);
+  if (number_of_threads_ == 1) {
+    RCLCPP_WARN(rclcpp::get_logger("rclcpp"),
+                "MultiThreadedExecutor executor used with a single thread.\n"
+                "Use SingleThreadedExecutor instead.");
   }
 }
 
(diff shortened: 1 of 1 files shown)


AlexisTM:
The use of `std::thread::hardware_concurrency` also bothers me. Indeed, the normal user will use the default Multi-threaded executor which would spin 128 threads on a Threadripper machine for each executor.

Instead, I would prefer having the application developer define the strict minimum amount of threads (with 2 by default) for reproducible results. Plus, for computation intensive software (reentrant data callbacks to start multiple long processing, maybe for CV), a user option to set the number of threads higher would be great (in that case as the number of cores).

alsora:
I would add a warning / error /something if a user tries to set the number of threads of this executor to 1.

Besides this, the change looks good to me.

AlexisTM:
The check for `if (number_of_threads_ == 0)` has been replaced for the check `if (number_of_threads_ == 1)` to output a warning.

The case of 0 is not possible anymore as if  `std::thread::hardware_concurrency()` returns 0, it will use 2 threads.

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -34,9 +35,13 @@ MultiThreadedExecutor::MultiThreadedExecutor(
   yield_before_execute_(yield_before_execute),
   next_exec_timeout_(next_exec_timeout)
 {
-  number_of_threads_ = number_of_threads ? number_of_threads : std::thread::hardware_concurrency();
-  if (number_of_threads_ == 0) {
-    number_of_threads_ = 1;
+  number_of_threads_ = number_of_threads
```
alsora, line 38: Minor: `number_of_threads` is a `size_t`, not a boolean, so prefer doing an explicit check with `number_of_threads > 0`

alsora (approved):

clalancette (approved):
This seems reasonable to me (though I'm going to update the title; we aren't setting the default to 2, we are setting the minimum to 2).  I'll run CI on it next.

clalancette:
CI is happy here, so going ahead with this.  Thanks for the contribution.

fujitatomoya (commented):
following up the review, lgtm

PR #2040, opened 2022-11-01
Implement validity checks for rclcpp::Clock

methylDragon:
This PR implements the analogues of the [isValid](https://docs.ros.org/en/latest/api/rostime/html/classros_1_1Time.html#a84cd12a499c70fb3eefce89cda2b6113) and [waitForValid](https://docs.ros.org/en/latest/api/rostime/html/classros_1_1Time.html#ab61825bb028f8c10c5e94aeca32bbfda) methods in rclcpp.

~~I also added the necessary tests.~~

~~Note: In this case, I added a special case for treating `RCL_CLOCK_UNINITIALIZED` as invalid as well, since there isn't an analogous clock type in the ROS 1 API.~~

Pinging @sloretz for visibility.

Edit: Waiting on:
- https://github.com/ros2/rcl/pull/1021

### DIFF
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -145,13 +145,8 @@ public:
    * uninitialized time (i.e. with zero time.)
    *
    * Note:
-   *     This notion of validity is different from the rcl notion of validity (rcl_clock_valid)!
-   *     A clock that is rcl_clock_valid is a clock that can be used to get a time (i.e. not
-   *     NULL, not RCL_CLOCK_UNINITIALIZED, or with a missing get_now() method.)
-   *
-   *     A clock that is valid in the sense codified by this method is a clock with non-zero time.
-   *     Consequently, a clock that is not rcl_clock_valid can never become valid, since it cannot
-   *     be used to obtain a time.
+   *     A valid clock must be both rcl_clock_valid and hold a time that is rcl_time_point_valid.
+   *     An invalid clock can potentially become valid if it is rcl_clock_valid.
    *
    * \return true if clock was or became valid
    */
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -138,6 +138,48 @@ public:
     Context::SharedPtr context = contexts::get_global_default_context());
 
   /**
+   * Check if the clock is valid.
+   *
+   * An invalid clock is either a clock that is RCL_CLOCK_UNINITIALIZED or is an RCL_ROS_TIME
+   * clock that is using ros time (ros_time_is_active()) but with time 0.
+   *
+   * \return true if clock was or became valid
+   */
+  RCLCPP_PUBLIC
+  bool
+  is_valid();
+
+  /**
+   * Wait for clock to become valid.
+   *
+   * \param timeout the maximum time to wait for.
+   * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
+   * \return true if clock was or became valid
+   */
+  RCLCPP_PUBLIC
+  bool
+  wait_for_valid(
+    Context::SharedPtr context=contexts::get_global_default_context(),
+    Duration wait_tick_ns=Duration(0, 1e7));
+
+  /**
+   * Wait for clock to become valid, with timeout.
+   *
+   * The timeout is waited in system time.
+   *
+   * \param timeout the maximum time to wait for.
+   * \param context the context to wait in.
+   * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
+   * \return true if the clock is active
+   */
+  RCLCPP_PUBLIC
+  bool
+  wait_for_valid(
+    const rclcpp::Duration& timeout,
+    Context::SharedPtr context=contexts::get_global_default_context(),
+    Duration wait_tick_ns=Duration(0, 1e7));
+
+  /**
    * Returns the clock of the type `RCL_ROS_TIME` is active.
    *
    * \return true if the clock is active
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -159,8 +159,8 @@ public:
   RCLCPP_PUBLIC
   bool
   wait_for_valid(
-    Context::SharedPtr context=contexts::get_global_default_context(),
-    Duration wait_tick_ns=Duration(0, 1e7));
+    Context::SharedPtr context = contexts::get_global_default_context(),
+    Duration wait_tick_ns = Duration(0, 1e7));
 
   /**
    * Wait for clock to become valid, with timeout.
@@ -175,9 +175,9 @@ public:
   RCLCPP_PUBLIC
   bool
   wait_for_valid(
-    const rclcpp::Duration& timeout,
-    Context::SharedPtr context=contexts::get_global_default_context(),
-    Duration wait_tick_ns=Duration(0, 1e7));
+    const rclcpp::Duration & timeout,
+    Context::SharedPtr context = contexts::get_global_default_context(),
+    Duration wait_tick_ns = Duration(0, 1e7));
 
   /**
    * Returns the clock of the type `RCL_ROS_TIME` is active.
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -140,8 +140,18 @@ public:
   /**
    * Check if the clock is valid.
    *
-   * An invalid clock is either a clock that is RCL_CLOCK_UNINITIALIZED or a clock with zero-time.
-   * In other words, an uninitialized clock.
+   * A valid clock is a clock that has a non-zero time.
+   * An invalid clock is an uninitialized clock (i.e. not rcl_clock_valid) or a clock with an
+   * uninitialized time (i.e. with zero time.)
+   *
+   * Note:
+   *     This notion of validity is different from the rcl notion of validity (rcl_clock_valid)!
+   *     A clock that is rcl_clock_valid is a clock that can be used to get a time (i.e. not
+   *     NULL, not RCL_CLOCK_UNINITIALIZED, or with a missing get_now() method.)
+   *
+   *     A clock that is valid in the sense codified by this method is a clock with non-zero time.
+   *     Consequently, a clock that is not rcl_clock_valid can never become valid, since it cannot
+   *     be used to obtain a time.
    *
    * \return true if clock was or became valid
    */
@@ -154,6 +164,7 @@ public:
    *
    * \param context the context to wait in.
    * \return true if clock was or became valid
+   * \throws std::runtime_error if the context is invalid or clock is not rcl_clock_valid
    */
   RCLCPP_PUBLIC
   bool
@@ -168,6 +179,7 @@ public:
    * \param context the context to wait in.
    * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
    * \return true if the clock is active
+   * \throws std::runtime_error if the context is invalid or clock is not rcl_clock_valid
    */
   RCLCPP_PUBLIC
   bool
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -140,8 +140,8 @@ public:
   /**
    * Check if the clock is valid.
    *
-   * An invalid clock is either a clock that is RCL_CLOCK_UNINITIALIZED or is an RCL_ROS_TIME
-   * clock that is using ros time (ros_time_is_active()) but with time 0.
+   * An invalid clock is either a clock that is RCL_CLOCK_UNINITIALIZED or a clock with zero-time.
+   * In other words, an uninitialized clock.
    *
    * \return true if clock was or became valid
    */
@@ -152,15 +152,12 @@ public:
   /**
    * Wait for clock to become valid.
    *
-   * \param timeout the maximum time to wait for.
-   * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
+   * \param context the context to wait in.
    * \return true if clock was or became valid
    */
   RCLCPP_PUBLIC
   bool
-  wait_for_valid(
-    Context::SharedPtr context = contexts::get_global_default_context(),
-    Duration wait_tick_ns = Duration(0, 1e7));
+  wait_for_valid(Context::SharedPtr context = contexts::get_global_default_context());
 
   /**
    * Wait for clock to become valid, with timeout.
@@ -177,7 +174,7 @@ public:
   wait_for_valid(
     const rclcpp::Duration & timeout,
     Context::SharedPtr context = contexts::get_global_default_context(),
-    Duration wait_tick_ns = Duration(0, 1e7));
+    Duration wait_tick_ns = Duration(0, static_cast<uint32_t>(1e7)));
 
   /**
    * Returns the clock of the type `RCL_ROS_TIME` is active.
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -138,38 +138,37 @@ public:
     Context::SharedPtr context = contexts::get_global_default_context());
 
   /**
-   * Check if the clock is valid.
+   * Check if the clock is started.
    *
-   * A valid clock is a clock that has a non-zero time.
-   * An invalid clock is an uninitialized clock (i.e. not rcl_clock_valid) or a clock with an
-   * uninitialized time (i.e. with zero time.)
+   * A started clock is a clock that reflects non-zero time.
+   * Typically a clock will be unstarted if it is using RCL_ROS_TIME with ROS time and
+   * nothing has been published on the clock topic yet.
    *
-   * Note:
-   *     A valid clock must be both rcl_clock_valid and hold a time that is rcl_time_point_valid.
-   *     An invalid clock can potentially become valid if it is rcl_clock_valid.
-   *
-   * \return true if clock was or became valid
+   * \return true if clock is started
+   * \throws std::runtime_error if the clock is not rcl_clock_valid
    */
   RCLCPP_PUBLIC
   bool
-  is_valid();
+  started();
 
   /**
-   * Wait for clock to become valid.
+   * Wait until clock to start.
    *
+   * \rclcpp::Clock::started
    * \param context the context to wait in
-   * \return true if clock was or became valid
+   * \return true if clock was already started or became started
    * \throws std::runtime_error if the context is invalid or clock is not rcl_clock_valid
    */
   RCLCPP_PUBLIC
   bool
-  wait_for_valid(Context::SharedPtr context = contexts::get_global_default_context());
+  wait_until_started(Context::SharedPtr context = contexts::get_global_default_context());
 
   /**
-   * Wait for clock to become valid, with timeout.
+   * Wait for clock to start, with timeout.
    *
-   * The timeout is waited in system time.
+   * The timeout is waited in steady time.
    *
+   * \rclcpp::Clock::started
    * \param timeout the maximum time to wait for.
    * \param context the context to wait in.
    * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
@@ -178,7 +177,7 @@ public:
    */
   RCLCPP_PUBLIC
   bool
-  wait_for_valid(
+  wait_until_started(
     const rclcpp::Duration & timeout,
     Context::SharedPtr context = contexts::get_global_default_context(),
     const rclcpp::Duration & wait_tick_ns = rclcpp::Duration(0, static_cast<uint32_t>(1e7)));
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -162,7 +162,7 @@ public:
   /**
    * Wait for clock to become valid.
    *
-   * \param context the context to wait in.
+   * \param context the context to wait in
    * \return true if clock was or became valid
    * \throws std::runtime_error if the context is invalid or clock is not rcl_clock_valid
    */
@@ -178,7 +178,7 @@ public:
    * \param timeout the maximum time to wait for.
    * \param context the context to wait in.
    * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
-   * \return true if the clock is active
+   * \return true if clock was or became valid
    * \throws std::runtime_error if the context is invalid or clock is not rcl_clock_valid
    */
   RCLCPP_PUBLIC
@@ -186,7 +186,7 @@ public:
   wait_for_valid(
     const rclcpp::Duration & timeout,
     Context::SharedPtr context = contexts::get_global_default_context(),
-    Duration wait_tick_ns = Duration(0, static_cast<uint32_t>(1e7)));
+    const rclcpp::Duration & wait_tick_ns = rclcpp::Duration(0, static_cast<uint32_t>(1e7)));
 
   /**
    * Returns the clock of the type `RCL_ROS_TIME` is active.
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -195,7 +195,7 @@ Clock::is_valid()
     case RCL_ROS_TIME:
     case RCL_STEADY_TIME:
     case RCL_SYSTEM_TIME:
-      return now().nanoseconds() > 0;
+      return now().is_valid();
 
     // By right we shouldn't even get to this block, but these cases are included for completeness
     case RCL_CLOCK_UNINITIALIZED:
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -183,48 +183,34 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
 }
 
 bool
-Clock::is_valid()
+Clock::started()
 {
-  // Checks for null pointer, missing get_now() implementation, and RCL_CLOCK_UNINITIALIZED
   if (!rcl_clock_valid(get_clock_handle())) {
-    RCUTILS_LOG_WARN("clock is uninitialized and will never become valid");
-    return false;
+    throw std::runtime_error("clock is not rcl_clock_valid");
   }
-
-  switch (get_clock_type()) {
-    case RCL_ROS_TIME:
-    case RCL_STEADY_TIME:
-    case RCL_SYSTEM_TIME:
-      return now().is_valid();
-
-    // By right we shouldn't even get to this block, but these cases are included for completeness
-    case RCL_CLOCK_UNINITIALIZED:
-    default:
-      return false;
-  }
+  return rcl_clock_time_started(get_clock_handle());
 }
 
 bool
-Clock::wait_for_valid(Context::SharedPtr context)
+Clock::wait_until_started(Context::SharedPtr context)
 {
   if (!context || !context->is_valid()) {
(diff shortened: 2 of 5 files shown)


methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17566/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12104/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18133/)

rclcpp/test/rclcpp/test_time.cpp
```diff
@@ -848,3 +849,217 @@ TEST_F(TestClockSleep, sleep_for_basic_ros) {
...
+  }
+}
+
+void trigger_clock_changes(
+  rclcpp::Node::SharedPtr node,
+  std::shared_ptr<rclcpp::Clock> clock,
+  bool expect_time_update = true,
+  bool zero = false)
+{
+  auto clock_pub = node->create_publisher<rosgraph_msgs::msg::Clock>("clock", 10);
```
sloretz, line 954: Exercising the time source logic and `use_sim_time` parameter doesn't need to be in the scope of this test. It might not even need a `Node` instance. I'd recommend using [get_clock_handle()](https://github.com/ros2/rclcpp/blob/edbfe1404b24d0bc85ff88e8ff1f006670788e46/rclcpp/include/rclcpp/clock.hpp#L154) and the rcl methods [rcl_enable_ros_time_override()](https://github.com/ros2/rcl/blob/0e41255e5522c8636a4691312009eae4920fe114/rcl/include/rcl/time.h#L534) and [rcl_set_ros_time_override()](https://github.com/ros2/rcl/blob/0e41255e5522c8636a4691312009eae4920fe114/rcl/include/rcl/time.h#L630) directly to change the time on the clock. That will make the test run faster.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -137,6 +137,48 @@ class Clock
     Duration rel_time,
     Context::SharedPtr context = contexts::get_global_default_context());
 
+  /**
+   * Check if the clock is valid.
+   *
+   * An invalid clock is either a clock that is RCL_CLOCK_UNINITIALIZED or is an RCL_ROS_TIME
```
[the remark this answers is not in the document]
methylDragon: Good catch, I'll update the docstring.

I was considering adding an `is_uninitialized` method, but decided against it to avoid splitting the userbase over which method to use (and prioritized `is_valid()` for consistency with the `wait_for_valid()` methods, which is both consistent with the ROS 1 names, and also imo a more intuitive name for initialized time.)

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17568/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12108/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18135/)

methylDragon:
Changes incorporated!

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -137,6 +137,46 @@ class Clock
...
+   */
+  RCLCPP_PUBLIC
+  bool
+  is_valid();
+
+  /**
+   * Wait for clock to become valid.
+   *
+   * \param timeout the maximum time to wait for.
+   * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
```
sloretz, line 156: nit, this param is gone.

rclcpp/test/rclcpp/test_time.cpp
```diff
@@ -848,3 +849,64 @@ TEST_F(TestClockSleep, sleep_for_basic_ros) {
   sleep_thread.join();
   EXPECT_TRUE(sleep_succeeded);
 }
+
+class TestClockValid : public ::testing::Test
+{
+protected:
+  static void SetUpTestCase()
+  {
+    rclcpp::shutdown();
```
sloretz, line 858: I'm surprised this is needed in setup since teardown calls it. What happens without it?

rclcpp/include/rclcpp/clock.hpp
methylDragon: Whoops

rclcpp/test/rclcpp/test_time.cpp
methylDragon: Ah, I was mimicking https://github.com/methylDragon/rclcpp/blob/ch3/clock-wait-for-valid/rclcpp/test/rclcpp/test_time.cpp#L461 

I don't think anything goes wrong if it gets removed though, but I put it in to be safe

rclcpp/include/rclcpp/clock.hpp
methylDragon: Sorry about that :bow: 

https://github.com/ros2/rclcpp/pull/2040/commits/937c2130e827a6216389e68a0ab9c08d3f3b5e62#diff-f8d8723006f022e91ac9a1e42f3622561b8e671d9508c7b17e3b4ea49b8902d6L156

rclcpp/test/rclcpp/test_time.cpp
methylDragon: Removed for my new test case, it seems necessary for the other one, for some reason..

https://github.com/ros2/rclcpp/compare/937c2130e827a6216389e68a0ab9c08d3f3b5e62..23271a6bd124f9345b86a82441810588d35b5de6

rclcpp/include/rclcpp/clock.hpp
methylDragon: Added the context param too! https://github.com/ros2/rclcpp/compare/937c2130e827a6216389e68a0ab9c08d3f3b5e62..23271a6bd124f9345b86a82441810588d35b5de6

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17570/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12110/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18137/)

sloretz (approved):
LGTM with green CI!

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,62 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+  }
+}
+
+bool
+Clock::wait_for_valid(Context::SharedPtr context)
+{
+  if (get_clock_type() == RCL_CLOCK_UNINITIALIZED) {
+    throw std::runtime_error(
+            "clock cannot be waited on as its clock_type is RCL_CLOCK_UNINITIALIZED");
+  }
```
fujitatomoya, line 206: What about the following?

```suggestion
  if (!is_valid()) {
    RCUTILS_LOG_ERROR("clock is not valid"); // this should be put in is_valid() method.
    return false;
  }
```

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,62 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+      return now().nanoseconds() > 0;
+
+    case RCL_CLOCK_UNINITIALIZED:
+    default:
+      return false;
+  }
+}
+
+bool
+Clock::wait_for_valid(Context::SharedPtr context)
```
fujitatomoya, line 201: Probably we need to check the context is valid, and if not throws exception?

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -137,6 +137,45 @@ class Clock
...
+
+  /**
+   * Wait for clock to become valid.
+   *
+   * \param context the context to wait in.
+   * \return true if clock was or became valid
+   */
+  RCLCPP_PUBLIC
+  bool
+  wait_for_valid(Context::SharedPtr context = contexts::get_global_default_context());
```
fujitatomoya, line 160: this throws exception with current code, if so probably doc needs to be described accordingly here.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -137,6 +137,45 @@ class Clock
...
+   * The timeout is waited in system time.
+   *
+   * \param timeout the maximum time to wait for.
+   * \param context the context to wait in.
+   * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
+   * \return true if the clock is active
+   */
+  RCLCPP_PUBLIC
+  bool
+  wait_for_valid(
```
fujitatomoya, line 174: same here, this throws exception.

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,62 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+{
+  if (get_clock_type() == RCL_CLOCK_UNINITIALIZED) {
+    throw std::runtime_error(
+            "clock cannot be waited on as its clock_type is RCL_CLOCK_UNINITIALIZED");
+  }
+
+  // Wait until the first valid time
+  return sleep_until(rclcpp::Time(0, 1, get_clock_type()), context);
```
fujitatomoya, line 209: Why do we need to call this? if this is the definition of valid or invalid, this should be called in `is_valid` method? this method is meant to loop forever to check if it becomes into valid?

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,62 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+
+  if (get_clock_type() == RCL_CLOCK_UNINITIALIZED) {
+    throw std::runtime_error(
+            "clock cannot be waited on as its clock_type is RCL_CLOCK_UNINITIALIZED");
+  }
+
+  Clock timeout_clock = Clock(RCL_SYSTEM_TIME);
+  Time start = timeout_clock.now();
+
+  while (!is_valid() && context->is_valid()) {
```
fujitatomoya, line 230: context pointer is already checked in the 1st place above?

```suggestion
  while (!is_valid()) {
```

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,62 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+    timeout_clock.sleep_for(Duration(wait_tick_ns));
+    if (timeout_clock.now() - start > timeout) {
+      return false;
+    }
+  }
+
+  return context->is_valid();
```
fujitatomoya, line 237: should we call `is_valid` instead? i think we are not checking if the context is valid or invalid?

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,62 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+  if (!context || !context->is_valid()) {
+    throw std::runtime_error("context cannot be slept with because it's invalid");
+  }
+
+  if (get_clock_type() == RCL_CLOCK_UNINITIALIZED) {
+    throw std::runtime_error(
+            "clock cannot be waited on as its clock_type is RCL_CLOCK_UNINITIALIZED");
+  }
+
+  Clock timeout_clock = Clock(RCL_SYSTEM_TIME);
```
fujitatomoya, line 227: probably we would want to use steady time instead? to avoid system time adjustment by NTP or user.

```suggestion
  Clock timeout_clock = Clock(RCL_STEADY_TIME);
```

fujitatomoya (changes requested):
@methylDragon thanks for the contribution.

Could you explain the actual problem that you meet to address with this PR? I am not really following the what needs to be done here, thanks in advance!

methylDragon:
> Member

Hey Tomoya, this PR is supposed to add `rclcpp` analogues of the `isValid` and `waitForValid` methods (in [ROS 1 time](http://docs.ros.org/en/diamondback/api/rostime/html/classros_1_1Time.html#a84cd12a499c70fb3eefce89cda2b6113)), used for checking if a clock is "valid"/"initialized" and so codifying that notion (from the [design docs](https://design.ros2.org/articles/clock_and_time.html).)

I'm upstreaming this change because it came up in one of my porting efforts.

rclcpp/src/rclcpp/clock.cpp
methylDragon: This is gonna be a little confusing, but the `rcl_clock_valid` notion of validity (validity of the internal clock handle) is different from the notion of validity for time sources (as defined in the [design docs](https://design.ros2.org/articles/clock_and_time.html))

The rcl notion is concerned with being able to get a time from a clock. But the notion of validity being put forward here is for when a non-zero time is obtained from the clock.

So we can't just fail if a clock isn't `is_valid()`, since in some cases (when the clock has time 0 (and so is `rcl_clock_valid()` but not `is_valid()`), we can wait for the clock **to become** valid.

That said, it makes sense to put an RCUTILS_LOG_ERROR (or an RCUTILS_LOG_WARN, at least) if the clock will never be valid (in the case of a missing now() implementation or a clock uninitialization.)

rclcpp/src/rclcpp/clock.cpp
methylDragon: If a clock is reflecting time 0, it is invalid. This waits until the clock becomes valid.

This is particularly useful if sim_time is active, since a time 0 will be invalid, and the simulation can remain paused.
In which case, the clock will wait until the simulation resumes and time moves forward into a valid, non-zero time

rclcpp/src/rclcpp/clock.cpp
methylDragon: The context can become invalid during the loop (e.g. when `rclcpp::shutdown()` is called.)

In that case, we should escape the loop for cleanup.

rclcpp/src/rclcpp/clock.cpp
methylDragon: Fair point, initially it was meant to check if we escaped the loop if the context became true or false, but it makes sense to just report the validity.

Likewise, I'll edit the `false` above to similarly report the validity.

rclcpp/src/rclcpp/clock.cpp
methylDragon: Good point! https://github.com/ros2/rclcpp/pull/2040/commits/a0091d0e5527fa45a09a77c70c98b8be82fe5b98

rclcpp/src/rclcpp/clock.cpp
methylDragon: https://github.com/ros2/rclcpp/pull/2040/commits/a0091d0e5527fa45a09a77c70c98b8be82fe5b98

rclcpp/src/rclcpp/clock.cpp
methylDragon: https://github.com/ros2/rclcpp/pull/2040/commits/a0091d0e5527fa45a09a77c70c98b8be82fe5b98

rclcpp/include/rclcpp/clock.hpp
methylDragon: https://github.com/ros2/rclcpp/pull/2040/commits/a0091d0e5527fa45a09a77c70c98b8be82fe5b98

rclcpp/include/rclcpp/clock.hpp
methylDragon: https://github.com/ros2/rclcpp/pull/2040/commits/a0091d0e5527fa45a09a77c70c98b8be82fe5b98

rclcpp/src/rclcpp/clock.cpp
methylDragon: Github didn't update, but the change is here:

https://github.com/ros2/rclcpp/pull/2040/commits/a0091d0e5527fa45a09a77c70c98b8be82fe5b98#diff-d574b4bf3206bd0cfb89e172528232189f5c5fcaecdf839a9a024da39ee52ee9R210

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,62 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
   return sleep_until(now() + rel_time, context);
 }
 
+bool
+Clock::is_valid()
```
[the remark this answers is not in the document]
methylDragon: Good point! `rcl_clock_valid` is more complete, since it checks for all cases where a time cannot be obtained from a clock.

https://github.com/ros2/rclcpp/pull/2040/commits/a0091d0e5527fa45a09a77c70c98b8be82fe5b98#diff-d574b4bf3206bd0cfb89e172528232189f5c5fcaecdf839a9a024da39ee52ee9R189

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17579/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12119/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18145/)

methylDragon:

fujitatomoya:
@methylDragon thank you for iterating with me.

> If time has not been set it will return zero if nothing has been received. A time value of zero should be considered an error meaning that time is uninitialized.

https://design.ros2.org/articles/clock_and_time.html, says above. thanks for pointing out the design.

> Notably, this notion of validity is separate from the `rcl` notion, since it also includes the fact that time 0 is **also invalid** (that is, it applies to the time reflected by the clock.)

this becomes my question that why these notions are different? by design, `rcl_clock_valid` should return `false` if the time is zero? and then we can call `rcl_clock_valid` from `rclcpp` and `rclpy`?

@clalancette @ivanpauno @wjwwood @sloretz what do you think?

clalancette:
> this becomes my question that why these notions are different?

Yes, this is a good question.

It turns out that ROS has been considering time==0 as invalid since practically forever.  It is not ideal, since time==0 is actually valid; it is just the start of time, not invalid time.  If we were totally redesigning this today, I think we might designate something like -1 as invalid, with 0 just being the start of time.  But if we did that now, we'd risk subtly breaking a lot of simulation applications that made some assumptions about past behavior.  So while it is not ideal, I think we should continue to treat time==0 as invalid.

> by design, `rcl_clock_valid` should return `false` if the time is zero? and then we can call `rcl_clock_valid` from `rclcpp` and `rclpy`?

But this is a good question; it would be nice if this policy were encoded in the `rcl` layer, and then just inherited by the client libraries.  @methylDragon Do you think that is feasible?

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,75 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+
+  if (!rcl_clock_valid(get_clock_handle())) {
+    throw std::runtime_error("clock cannot be waited on as it is not rcl_clock_valid");
+  }
+
+  Clock timeout_clock = Clock(RCL_STEADY_TIME);
+  Time start = timeout_clock.now();
+
+  while (!is_valid() && context->is_valid()) {  // Context check checks for rclcpp::shutdown()
+    timeout_clock.sleep_for(Duration(wait_tick_ns));
```
fujitatomoya, line 244: if `Duration(wait_tick_ns)` >> timeout, it will not be timed out with specified timeout Duration?
i think we need to add `if` statement to check if `wait_tick_ns` is less than `timeout`, and calculate for next sleep time based on how much time left until timeout.
test would be ideal.

fujitatomoya (changes requested):

fujitatomoya:
> It turns out that ROS has been considering time==0 as invalid since practically forever. It is not ideal, since time==0 is actually valid; it is just the start of time, not invalid time. If we were totally redesigning this today, I think we might designate something like -1 as invalid, with 0 just being the start of time.

agree on this. I was wondering the same thing, thanks for pointing this out.

rclcpp/src/rclcpp/clock.cpp
methylDragon: > But this is a good question; it would be nice if this policy were encoded in the `rcl` layer, and then just inherited by the client libraries. @methylDragon Do you think that is feasible?

It'll be nice to put this in the `rcl` layer, though I'm not too sure how to do it (well, more precisely, how to name the methods without causing API/ABI breaks...)
We need a distinct way to check if the members of a clock are initialized (currently what `rcl_clock_valid()` is doing), since that gets used by the client libraries (e.g. in the `ros_time_active()` method, which then feeds into the clock setting methods).

So if I were to add it into the `rcl` layer, I need to add a new method. The problem is:
- I either need to change the current `rcl_clock_valid()` and then have this new method be the `rcl_clock_valid()`, necessitating changes across every client library that uses the method. But I'm pretty sure this will break ABI/API.
- Or I need to add a new method, but then it'd be confusing because `rcl_clock_valid()` would not be encoding the notion of validity in the design docs... Is there a better method name for the new method I can use?

If we can't resolve the naming issue, I think it'd be more intuitive to place it in the client libraries? Or maybe I'm missing something :V

methylDragon: https://github.com/ros2/rclcpp/pull/2040/commits/6face35ae9b23405eb705169d767197f9101446c

methylDragon:
I just had an idea, I will create an `rcl_time_valid()` method in `rcl` to encode the notion of non-zero time validity, and use it alongside `rcl_clock_valid()` here.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -137,6 +137,57 @@ class Clock
...
+  /**
+   * Wait for clock to become valid.
+   *
+   * \param context the context to wait in
+   * \return true if clock was or became valid
+   * \throws std::runtime_error if the context is invalid or clock is not rcl_clock_valid
+   */
+  RCLCPP_PUBLIC
+  bool
+  wait_for_valid(Context::SharedPtr context = contexts::get_global_default_context());
```
ivanpauno, line 171: what's the use case of wait_for_valid()?
It seems like a strange feature to me

methylDragon: It'll block till the first valid time. In most cases this will be waiting for sim to start (and time to march forward)

methylDragon:

methylDragon:
Hey @fujitatomoya , with the merge of https://github.com/ros2/rcl/pull/1018 , the validity check is now codified in `rcl`! I think we can merge this PR once CI goes green, what do you think?

methylDragon:

methylDragon:

rclcpp/test/rclcpp/test_time.cpp
```diff
@@ -451,6 +452,18 @@ TEST_F(TestTime, test_overflow_underflow_throws) {
     std::underflow_error("addition leads to int64_t underflow"));
 }
 
+TEST_F(TestTime, validity) {
+  EXPECT_FALSE(rclcpp::Time(0, 0, RCL_CLOCK_UNINITIALIZED).is_valid());
+  EXPECT_FALSE(rclcpp::Time(0, 0, RCL_SYSTEM_TIME).is_valid());
+  EXPECT_FALSE(rclcpp::Time(0, 0, RCL_STEADY_TIME).is_valid());
+  EXPECT_FALSE(rclcpp::Time(0, 0, RCL_ROS_TIME).is_valid());
+
+  EXPECT_TRUE(rclcpp::Time(0, 1, RCL_CLOCK_UNINITIALIZED).is_valid());
```
fujitatomoya, line 461: i am inclined to make this FALSE, since it does not mean valid time because it is not initialized? what this time means valid to us? @ros2/team the same comment goes to https://github.com/ros2/rcl/pull/1018#pullrequestreview-1172904338. what does it mean valid for this time w/o initialization? for me, it is invalid...

methylDragon:
This PR has been updated to use https://github.com/ros2/rcl/pull/1021 instead!

It doesn't make sense for a non-zero time point to be invalid, so the names and implementations have been changed to make this PR about checking for whether a clock has **started** or not

methylDragon:

fujitatomoya:
@methylDragon thanks for iterating with patience, so this has been update and ready to review, right?

methylDragon:
> @methylDragon thanks for iterating with patience, so this has been update and ready to review, right?

Hey @fujitatomoya , this is ready, yep! Though it'll take awhile for `rcl` to get released so CI works...

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -182,6 +182,64 @@ Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
...
+    if (timeout < wait_tick_ns) {
+      timeout_clock.sleep_for(timeout);
+    } else {
+      timeout_clock.sleep_for(Duration(wait_tick_ns));
+    }
+
+    if (timeout_clock.now() - start > timeout) {
+      return started();
+    }
+  }
```
fujitatomoya, line 238: if `timeout` is 91s, `wait_tick_ns` is 10s, last sleep will be 10s that is 9 seconds off from the user expectation. i think it needs to calculate how much time left to wait for this case.

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -137,6 +137,51 @@ class Clock
...
+   * \param wait_tick_ns the time to wait between each iteration of the wait loop (in nanoseconds).
+   * \return true if clock was or became valid
+   * \throws std::runtime_error if the context is invalid or clock is not rcl_clock_valid
+   */
+  RCLCPP_PUBLIC
+  bool
+  wait_until_started(
+    const rclcpp::Duration & timeout,
+    Context::SharedPtr context = contexts::get_global_default_context(),
+    const rclcpp::Duration & wait_tick_ns = rclcpp::Duration(0, static_cast<uint32_t>(1e7)));
```
fujitatomoya, line 183: Do we need this argument? i though it could be user application responsibility. and have a single `wait_until_started` with timeout default 0 means wait forever?

fujitatomoya (changes requested):

rclcpp/include/rclcpp/clock.hpp
methylDragon: The wait ticks are meant to slow down the `while (!started() ...)` busy wait so it doesn't free spin, and it's the rate at which it checks if the clock has started (if the timeout has not passed).

There's another signature for this `wait_until_started` up there where no timeout being specified means it waits forever.

I'm not sure how else I should implement the wait :o

methylDragon: Hey @fujitatomoya , just giving you a tiny ping :>

rclcpp/src/rclcpp/clock.cpp
fujitatomoya: @methylDragon can you check this comment?

rclcpp/include/rclcpp/clock.hpp
fujitatomoya: yeah, okay if we have timeout argument, current implementation makes more sense. but i do not have strong opinion about this...

rclcpp/src/rclcpp/clock.cpp
methylDragon: Ah, right! I need to get to this, oops

methylDragon: https://github.com/ros2/rclcpp/pull/2040/commits/1df3675219280b460527de2cf3564b903f321c24 !

fujitatomoya (approved):

emersonknapp:
@methylDragon Can we backport this to Humble? I believe "new non-virtual methods on a class" is an ABI-stable change, so it seems to me like the answer is yes. The dependency https://github.com/ros2/rcl/pull/1021 only introduces new free functions, which I believe is also ABI stable

fujitatomoya:
> "new non-virtual methods on a class" is an ABI-stable change

yeah, correct. i believe backport should be fine. besides, this is something we add, there is no user application behavior change.

emersonknapp:
@mergifyio backport humble

methylDragon:
@Mergifyio backport humble

Issue #831, opened 2019-08-28
Proposal for NodeHandle concept that simplifies calling APIs with node-like objects


sloretz:
#### Feature description
This is an idea to simplify code that calls APIs using node interfaces. It proposes adding template classes `NodeHandle<>` and `NodeHandle<>::SharedPtr` who are given placeholder types that indicate the needed node interfaces.

##### When node interfaces are just needed for the lifetime of the call

`create_service()` needs a [`NodeBaseInterface` and `NodeServicesInterface`](https://github.com/ros2/rclcpp/blob/17841d6b7cff96efef87400fee2dcd63a6f10101/rclcpp/include/rclcpp/create_service.hpp#L34-L36). It doesn't need shared pointers to these interfaces; it just needs them for the duration of the call.

Before:

```c++
create_service(
  std::shared_ptr<node_interfaces::NodeBaseInterface> node_base,
  std::shared_ptr<node_interfaces::NodeServicesInterface> node_services,
  // ...

// User calls with
auto service = create_service(
  my_node_class->get_node_base_interface(),
  my_node_class->get_node_services_interface(),
  // ...
```

after

```c++
create_service(
     rclcpp::NodeHandle<Base, Services> node_handle,
     // ...

// User calls with
auto service = create_service(
  my_node_class,
  // ...
```

##### When node interfaces are needed long term

[`tf2_ros::TransformListener` needs the node base interface](https://github.com/ros2/geometry2/blob/4e519f3e77611a18f03398b6334e9cc8ccf49552/tf2_ros/include/tf2_ros/transform_listener.h#L104) for a long time, so it holds a shared pointer.

It currently does some logic in the header file and some in the cpp file, but the intent of this proposal is to enable moving all of it to the cpp file.

Before
```C++
  TransformListener(
    tf2::BufferCore & buffer,
    NodeT && node,
    // ...
    // in header creates subscription and calls initThread(node->get_node_base_interface()); to store shared pointer to node
```

After

```c++
  TransformListener(
    tf2::BufferCore & buffer,
    rclcpp::NodeHandle<Base, Subscription>::SharedPtr node_handle,
    // ...
    );  // Just declaration in header
```

#### Implementation considerations

Placeholder types should be created for each node interface. The purpose of these is to make the type of a `NodeHandle` more readable. In my opinion `rclcpp::NodeHandle<Base>` is easier to read than `rclcpp::NodeHandleBase<node_interfaces::NodeBaseInterface>`.

[...]

PR #2041, opened 2022-11-03
Implement Unified Node Interface (NodeInterfaces class)

methylDragon:
Finally a longstanding quibble about Nodes and LifecycleNodes and node interfaces has a potential resolution.

# Description
This PR introduces the concept of a `NodeInterfaces` class (that is, a unified adapter for any disparate collection of unique kinds of NodeInterfaces) that allows you to package interfaces into a single object. This is inspired by https://github.com/ros2/rclcpp/issues/831.

This allows you to write functions that take in a **single object** instead of a collection of node interfaces (or instead of a templated function...), and also supports both Nodes and LifecycleNodes, allowing you to quickly swap any place where you originally passed a Node with a NodeInterfaces. **This also means that the issue where Nodes and LifecycleNodes exist on separate inheritance trees can be mitigated a little by using this as a bridging class.**

What's more, the template parameters allow you to select which interfaces are made available, **and also can serve as type hints to downstream users** for which interfaces a function requires.

Additionally, you can aggregate interfaces from different nodes (or mocked interfaces) too, which isn't something you could trivially do if you made your function a template function that takes in a NodeT!

~~This PR also adds a new method for Nodes and LifecycleNodes to obtain their corresponding NodeInterfaces.~~

Tests have been implemented for ~~Nodes, LifecycleNodes, and~~ the NodeInterfaces template class itself.

# Example
This is best explained with an example.

```cpp
// Suppose we have some function:
void fn(NodeInterfaces<NodeBaseInterface, NodeClockInterface> interfaces);

// Then we can, explicitly:
rclcpp::Node node("some_node");
auto ni = NodeInterfaces<NodeBaseInterface, NodeClockInterface>(node);
fn(ni);

// But also:
fn(node);

// Subsetting a NodeInterfaces object also works!
auto ni_base = NodeInterfaces<NodeBaseInterface>(ni);

// Or aggregate them (you could aggregate interfaces from disparate node-likes)
auto ni_aggregated = NodeInterfaces<NodeBaseInterface, NodeClockInterface>(
  node->get_node_base_interface(),
  node->get_node_clock_interface()
)

// And then to access the interfaces:
// Get with get<>
auto base = ni.get<NodeBaseInterface>();

// Or the appropriate getter
auto clock = ni.get_clock_interface();
```

# Why is this useful?
(Adapted from https://github.com/ros2/rclcpp/issues/831)

Before you would have to...
```cpp
create_service(
  std::shared_ptr<node_interfaces::Base> node_base,
  std::shared_ptr<node_interfaces::Services> node_services,
  // ...

// User calls with
auto service = create_service(
  my_node_class->get_node_base_interface(),
  my_node_class->get_node_services_interface(),
...
```

But now you can:
```cpp
create_service(
   NodeInterfaces<Base, Services> node_handle,
   // ...

// User calls with
auto service = create_service(
  NodeInterfaces<Base, Services>(my_node_or_lifecycle_node_class),
  // ...
```

And just like that you can support both Nodes and LifecycleNodes!
This also serves as an easy drop-in replacement for any method that takes in an `rclcpp::Node` or similar shared pointer!

# Additional Notes
This uses C++14 compatible template metaprogramming constructs, for backportability. I'm pretty sure there are some C++17/C++20 constructs that would make the implementation more concise though.. For a future time.

EDIT: I realized that fold expressions are C++17... We'll have to find some workaround when backporting. (Actually so is is_same_v :man_facepalming: )


methylDragon:
Only thing I'm unsure of if if I should leave the NodeT constructor for `NodeHandle` explicit.
If it's not explicit we can rely on implicit conversion to allow passing in Node-like objects directly into methods that take in NodeHandle arguments without needing to explicitly instantiate the NodeHandle..

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17583/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12123/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18149/)

alsora:
I like the overall idea.
However I have a couple of questions from looking at your examples.

It seems like there are multiple methods to retrieve each interface (e.g. `base` and `get_node_base_interface`), why is that?

Can we implement it in a way such that users can customize it? I.e. If I create a new node interface, I would like to add it there

methylDragon:
> I like the overall idea. However I have a couple of questions from looking at your examples.
> 
> It seems like there are multiple methods to retrieve each interface (e.g. `base` and `get_node_base_interface`), why is that?
> 
> Can we implement it in a way such that users can customize it? I.e. If I create a new node interface, I would like to add it there

It's already implemented in such a way that you can attach a custom interface to the NodeHandle! Just use `set_node_XXX_interface(your_interface)` or `XXX(your_interface)`.

The multiple ways for retrieving the interface are just for convenience (or typing less...) I was basing it off the issue over this PR was inspired by.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1448,6 +1449,39 @@ class Node : public std::enable_shared_from_this<Node>
...
+   *     interfaces.
+   *   - ```node->get_node_handle<AllInterfaces>()``` will bind all interfaces.
+   *   - ```node->get_node_handle()``` will bind all interfaces.
+   *
+   * \return a NodeHandle bound with the Node's internal node interfaces.
+   */
+  template<typename ... InterfaceTs>
+  typename rclcpp::NodeHandle<InterfaceTs...>::SharedPtr
+  get_node_handle();
+
```
wjwwood, line 1484: It would be better (in my opinion) for the node class to know nothing about the new concept (`NodeHandle`). I would strive to remove functions like these. Instead a constructor for `NodeHandle` that can take any class that looks like `Node` or `LifecycleNode` would be better.

rclcpp/include/rclcpp/node_handle.hpp
```diff
@@ -0,0 +1,280 @@
...
+using BaseInterface = node_interfaces::NodeBaseInterface::SharedPtr;
+using ClockInterface = node_interfaces::NodeClockInterface::SharedPtr;
+using GraphInterface = node_interfaces::NodeGraphInterface::SharedPtr;
+using LoggingInterface = node_interfaces::NodeLoggingInterface::SharedPtr;
+using TimersInterface = node_interfaces::NodeTimersInterface::SharedPtr;
+using TopicsInterface = node_interfaces::NodeTopicsInterface::SharedPtr;
+using ServicesInterface = node_interfaces::NodeServicesInterface::SharedPtr;
+using WaitablesInterface = node_interfaces::NodeWaitablesInterface::SharedPtr;
+using ParametersInterface = node_interfaces::NodeParametersInterface::SharedPtr;
+using TimeSourceInterface = node_interfaces::NodeTimeSourceInterface::SharedPtr;
```
wjwwood, line 44: I think these renames are misleading because they don't imply that a `shared_ptr` is being used.

Actually, my preference would be to not have these aliases at all. In header files I prefer (and it's the prevailing style of `rclcpp`) to use full names and not create unnecessary aliases. In `_impl.hpp` files and `.cpp` files you can make any usings you want, but we should avoid them in headers, imo.

Also, if you do use them in headers, ideally do so in the scope of a function or if on a class, make them private.

rclcpp/include/rclcpp/node_handle.hpp
```diff
@@ -0,0 +1,280 @@
...
+struct AllInterfaces {};
```
wjwwood, line 45: I actually don't think we should have an all option, instead interfaces using this should explicitly specify which interfaces they need.

rclcpp/include/rclcpp/node_handle.hpp
wjwwood: And if the intention was to make shorthand for the users of this class, then I would recommend against that. The users can make shortcuts if they want.

rclcpp/include/rclcpp/node_handle.hpp
```diff
@@ -0,0 +1,280 @@
...
+  : base_(nullptr),
+    clock_(nullptr),
+    graph_(nullptr),
+    logging_(nullptr),
+    timers_(nullptr),
+    topics_(nullptr),
+    services_(nullptr),
+    waitables_(nullptr),
+    parameters_(nullptr),
+    time_source_(nullptr)
```
wjwwood, line 66: I think we should talk about this design. I have a few issues with it, which maybe we can improve on:

- it copies and stores the interface pointers itself, when instead we could just keep a pointer to the node class itself (which already has these pointers)
- this list is finite and has to be adjusted each time a new interface is added. I'm not sure we want to have that, and I think if we address the previous point we can avoid having to have an exhaustive list of interfaces in this class

rclcpp/include/rclcpp/node_handle.hpp
```diff
@@ -0,0 +1,280 @@
...
+  inline ParametersInterface parameters() {return parameters_;}
+
+  /// Set the bound NodeParametersInterface
+  inline void parameters(ParametersInterface interface) {parameters_ = interface;}
+
+  /// Return the bound NodeTimeSourceInterface
+  inline TimeSourceInterface time_source() {return time_source_;}
+
+  /// Set the bound NodeTimeSourceInterface
+  inline void time_source(TimeSourceInterface interface) {time_source_ = interface;}
```
wjwwood, line 193: I don't see the point of the setters. Why do we need them?

rclcpp/include/rclcpp/node_handle.hpp
```diff
@@ -0,0 +1,280 @@
...
+
+
+  // Alterate getters
```
wjwwood, line 196:
```suggestion
  // Alternate getters
```

rclcpp/include/rclcpp/node_handle.hpp
```diff
@@ -0,0 +1,280 @@
...
+  inline BaseInterface get_node_base_interface() {return base();}
+  inline ClockInterface get_node_clock_interface() {return clock();}
+  inline GraphInterface get_node_graph_interface() {return graph();}
+  inline LoggingInterface get_node_logging_interface() {return logging();}
+  inline TimersInterface get_node_timers_interface() {return timers();}
+  inline TopicsInterface get_node_topics_interface() {return topics();}
+  inline ServicesInterface get_node_services_interface() {return services();}
+  inline WaitablesInterface get_node_waitables_interface() {return waitables();}
+  inline ParametersInterface get_node_parameters_interface() {return parameters();}
+  inline TimeSourceInterface get_node_time_source_interface() {return time_source();}
```
wjwwood, line 206: Why do we need alternate getters?

Also why don't we have something like this instead:

```c++
node_handle.get<NodeBaseInterface>();
```

wjwwood (commented):
I'd also like to talk about the name of the new class `NodeHandle`. I think we should pick something else because really it's not a "handle" (we already have a `rcl_node_handle_t` which is actually a handle, i.e. an opaque object that is a stand in for another actual thing) and because we had a `NodeHandle` in ROS 1 which was something very different and that might cause confusion.

Ideally this whole class would be replaced with a `constraint` but we don't have access to that atm with C++17.

Some ideas from me:

- `NodeWithInterfaces<...>`
  - `my_func(NodeWithInterface<...>)` would read as "my_func takes a Node-like thing with the interfaces ..."
- `NodeWithRequiredInterfaces<...>`
- `NodeInterfaces<...>`
  - invokes the idea of a collection of interfaces, which may or may not be misleading depending on how we end up implementing this class

rclcpp/include/rclcpp/node.hpp
methylDragon: Hmm, I was thinking along the lines of the NodeHandle class being like an aggregate interface you can get, just as with the node interface getters.

But since a constructor taking in NodeT on NodeHandle already exists, I'm fine with removing the ones local to each Node/LifecycleNode class. They're more for convenience anyway.

rclcpp/include/rclcpp/node_handle.hpp
methylDragon: I was thinking of storing a node pointer to the node, but I decided against it for two reasons (though the latter is probably more significant than the former):
- It'll introduce an additional layer of function indirection
- We won't be able to aggregate node interfaces together (making them from parts)

I think the latter, aggregating reason is more important because it was one of the requirements I was trying to fulfill [from the discussion in the motivating issue](https://github.com/ros2/rclcpp/issues/831#issuecomment-530157701). But it's true that it means we'll need to adjust the NodeHandle class each time a new interface is added, though in my opinion it's a worthwhile trade-off since that shouldn't happen as often, I think?

rclcpp/include/rclcpp/node_handle.hpp
methylDragon: Will remove.

rclcpp/include/rclcpp/node_handle.hpp
methylDragon: It's to allow building of aggregate node handles (from parts), as suggested [from the discussion in the original issue](https://github.com/ros2/rclcpp/issues/831#issuecomment-530157701).

rclcpp/include/rclcpp/node_handle.hpp
methylDragon: One set is for convenience and the other set is to mimic the node interface getters. If I had to choose between the three (the two I mentioned earlier, and the `get<>()`), I think I'd go with the node interface getters, for parity with node-like types :o

rclcpp/include/rclcpp/node_handle.hpp
alsora: I agree with @methylDragon, we should allow this class to be constructed from generic objects, i.e not necessarily rclcpp::node or rclcpp_lifecycle::LifecycleNode

methylDragon:
> I'd also like to talk about the name of the new class `NodeHandle`. I think we should pick something else because really it's not a "handle" (we already have a `rcl_node_handle_t` which is actually a handle, i.e. an opaque object that is a stand in for another actual thing) and because we had a `NodeHandle` in ROS 1 which was something very different and that might cause confusion.
> 
> Ideally this whole class would be replaced with a `constraint` but we don't have access to that atm with C++17.
> 
> Some ideas from me:
> 
> * `NodeWithInterfaces<...>`
>   
>   * `my_func(NodeWithInterface<...>)` would read as "my_func takes a Node-like thing with the interfaces ..."
> * `NodeWithRequiredInterfaces<...>`
> * `NodeInterfaces<...>`
>   
>   * invokes the idea of a collection of interfaces, which may or may not be misleading depending on how we end up implementing this class

How about NodeInterfaceHandle? (Since it's a handler for node interfaces, no matter where they come from)

methylDragon:

ivanpauno (commented):
I think that something like this may be a good idea.
I have left one suggestion of how to implement it.

I think we have to thoroughly evaluate how this can be used in demos/examples, and double check if using the new "unified node interface" is beneficial.

ivanpauno:
> I'd also like to talk about the name of the new class NodeHandle. I think we should pick something else because really it's not a "handle" (we already have a rcl_node_handle_t which is actually a handle, i.e. an opaque object that is a stand in for another actual thing) and because we had a NodeHandle in ROS 1 which was something very different and that might cause confusion.

+1
Of the proposed options, `NodeInterfaces<...>` seems nice to me.

rclcpp/include/rclcpp/node_handle.hpp
methylDragon: I like this idea, it'll allow for users to create their own interfaces and bind them to the node handle as well.

Only thing is that this will also allow binding and inheritance from any possible class (as long as it has a constructor that takes nodeT). I think this is a fair enough tradeoff, since it also resolves @alsora 's

> Can we implement it in a way such that users can customize it? I.e. If I create a new node interface, I would like to add it there

And additionally gives me a good reason for introducing more terse type names for use with the template args.

---

I'll also move the implementation to the node_interfaces directory

methylDragon:
> > I'd also like to talk about the name of the new class NodeHandle. I think we should pick something else because really it's not a "handle" (we already have a rcl_node_handle_t which is actually a handle, i.e. an opaque object that is a stand in for another actual thing) and because we had a NodeHandle in ROS 1 which was something very different and that might cause confusion.
> 
> +1 Of the proposed options, `NodeInterfaces<...>` seems nice to me.

@ivanpauno 

Could I double-check that this +1 is over NodeInterfaceHandle? The multiple inheritance templated example you gave above used NodeInterfaceHandle :grimacing:. I'll change it to whichever you give the +1 to!

methylDragon:
I've updated the tests and moved everything to node_interfaces!

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17619/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18189/)

ivanpauno:
> Could I double-check that this +1 is over NodeInterfaceHandle (so +1 to NodeInterfaces even though NodeInterfaceHandle exists?) The multiple inheritance templated example you gave above used NodeInterfaceHandle grimacing.
> I'll change it to whichever you give the +1 to!

I kept the original naming in my example to avoid confusions.
I think that replacing `NodeHandle` with `NodeInterfaces` in my example is fine.
`NodeInterfaceHandle` seems a bit long IMHO.

> Also, what do you think about putting getters for this node interface aggregate onto Node and LifecycleNode?

I think it would be fine if the `NodeInterfaces` constructor is implicit.
i.e.
```cpp
// declaration
void my_func(NodeInterfaces<Base, Topics> node);

// usage
rclcpp::Node::SharedPtr node = std::make_shared<MyNodeDerivedClass>(...);
my_func(node);  // works thanks to implicit constructor
```

Having also a getter is fine though.

methylDragon:
> > Could I double-check that this +1 is over NodeInterfaceHandle (so +1 to NodeInterfaces even though NodeInterfaceHandle exists?) The multiple inheritance templated example you gave above used NodeInterfaceHandle grimacing.
> > I'll change it to whichever you give the +1 to!
> 
> I kept the original naming in my example to avoid confusions. I think that replacing `NodeHandle` with `NodeInterfaces` in my example is fine. `NodeInterfaceHandle` seems a bit long IMHO.
> 
> > Also, what do you think about putting getters for this node interface aggregate onto Node and LifecycleNode?
> 
> I think it would be fine if the `NodeInterfaces` constructor is implicit. i.e.
> 
> ```c++
> // declaration
> void my_func(NodeInterfaces<Base, Topics> node);
> 
> // usage
> rclcpp::Node::SharedPtr node = std::make_shared<MyNodeDerivedClass>(...);
> my_func(node);  // works thanks to implicit constructor
> ```
> 
> Having also a getter is fine though.

Alright, I'll rename it to `NodeInterfaces`!

Do you know how to allow an implicit constructor to get past the linters? Since the constructor takes a single argument, the linter screams that it should be explicit...

sloretz:
> Do you know how to allow an implicit constructor to get past the linters? Since the constructor takes a single argument, the linter screams that it should be explicit

Is it cpplint? I think `// NOLINT(runtime/explicit)` on the same line as the constructor would silence that.

methylDragon:
Class has been renamed! (The only thing that's iffy to me is it makes the include path `rclcpp/node_interfaces/node_interfaces.hpp`, which could potentially be misleading..)

methylDragon:
.. I have no idea how to fix that Windows linking error. All the involved objects are templated, so what gives :scream:

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1448,6 +1449,51 @@ class Node : public std::enable_shared_from_this<Node>
   rclcpp::node_interfaces::NodeTimeSourceInterface::SharedPtr
   get_node_time_source_interface();
 
+  /// Return a NodeInterfaces bound with the Node's internal node interfaces.
+  /**
+   * Specify which interfaces you want to bind using the template parameters by specifying
+   * interface support classes to use. Any unmentioned interfaces will be unavailable to bind.
+   *
+   * This method will return a NodeInterfaces with no bound interfaces. You must set them using
```
ivanpauno, line 1457: I think this is a bit strange, I would expect it to be the same as `NodeInterfaces<...>(node)`, I wouldn't expect it to work like the default constructor.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1448,6 +1449,51 @@ class Node : public std::enable_shared_from_this<Node>
...
+   *   - Templated constructor taking NodeT
+   *   - get_node_<interface_name>_interface()
+   *   - set_node_<interface_name>_interface()
+   *
+   * Usage example:
+   *   - ```NodeInterfaces<rclcpp::node_interfaces::Base>(node)```
+   *     will bind just the NodeBaseInterface.
+   *   - ```NodeInterfaces<
+   *          rclcpp::node_interfaces::Base, rclcpp::node_interfaces::Clock>(node)```
+   *     will bind both the NodeBaseInterface and NodeClockInterface.
```
ivanpauno, line 1486: I think that these examples should be in `NodeInterfaces` documentation and not here

rclcpp/include/rclcpp/node.hpp
ivanpauno: Actually, the implementation is what I expect.
I think either the documentation is outdated or not super clear

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1448,6 +1449,51 @@ class Node : public std::enable_shared_from_this<Node>
...
+   *   - Topics:     Supports NodeTopicsInterface
+   *   - Waitables:  Supports NodeWaitablesInterface
+   *
+   * Or you can define your own interface support classes!
+   *
+   * Each of the support classes should define:
+   *   - Default constructor
+   *   - Templated constructor taking NodeT
+   *   - get_node_<interface_name>_interface()
+   *   - set_node_<interface_name>_interface()
```
ivanpauno, line 1479: This also seems repeated with the documentation in `NodeInterfaces`, maybe just link there (?)

ivanpauno (commented):
LGTM!

I have some minor comments, I think it would be nice to try to deduplicate documentation.
Instead of duplicating, you can crosslink docs.

I think it would be nice to also use this in some examples, to see if it's making things easier to users.

rclcpp/include/rclcpp/node.hpp
methylDragon: https://github.com/ros2/rclcpp/pull/2041/commits/2d65b59c855d73b66ecbdf5638b59d408fc7f7b9

rclcpp/include/rclcpp/node.hpp
methylDragon: [ros2/rclcpp@`2d65b59` (#2041)](https://github.com/ros2/rclcpp/pull/2041/commits/2d65b59c855d73b66ecbdf5638b59d408fc7f7b9)

rclcpp/include/rclcpp/node.hpp
methylDragon: https://github.com/ros2/rclcpp/pull/2041/commits/2d65b59c855d73b66ecbdf5638b59d408fc7f7b9

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
```diff
@@ -0,0 +1,127 @@
...
+class NodeInterfaces
+  : public std::enable_shared_from_this<NodeInterfaces<InterfaceTs...>>, public InterfaceTs ...
+{
+  static_assert(0 != sizeof ...(InterfaceTs), "Template parameters must be populated!");
+
+public:
+  RCLCPP_SMART_PTR_DEFINITIONS(NodeInterfaces)
+
+  /// Create a new node handle with no bound node interfaces.
+  NodeInterfaces()
```
alsora, line 38: the comment seems to refer to the old name (`node handle`). this applies also to other comments below.

moreover, why do we need this constructor?

rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+class TestNodeInterfaces : public ::testing::Test
+{
+protected:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+
+
```
alsora, line 27: nit: remove double empty line

alsora:
Looks good to me.
I left a couple of minor comments about style.
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17631/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12179/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18206/)

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
methylDragon: This is necessary for creating an unbound object to then use to aggregate arbitrary sets of node interfaces using the setters!

So, e.g.

Make empty base and clock Interfaces object
Set base to some mock base object
Set clock to some node's clock object

Etc.

rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp
methylDragon: https://github.com/ros2/rclcpp/compare/1cac3b88b72d6ff02f8aeb71e788d0562a047408..aafce8ca04e6b761f7fcd3d7c77ca08965afea83

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
methylDragon: https://github.com/ros2/rclcpp/compare/1cac3b88b72d6ff02f8aeb71e788d0562a047408..aafce8ca04e6b761f7fcd3d7c77ca08965afea83

alsora (approved):
Looks good to me.

methylDragon:
I'm gonna change around a few methods for const correctness..

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17639/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12193/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18214/)

rclcpp/include/rclcpp/node.hpp
wjwwood: I still think remove them.

I really think Node should not have to know about them, because every future Node-like class will also have to add them to keep up with LifecycleNode and Node. It just doesn't scale and it doesn't follow encapsulation.

rclcpp/include/rclcpp/node_handle.hpp
wjwwood: I think those should be done via constructor arguments. You don't need to have setters to do this. I would still remove these.

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
wjwwood: As I said above, I would prefer no setters and just require people to pass aggregate parts via a constructor.

wjwwood (changes requested):
I left some new comments, which I think should be answered at least before merging.

I would also like to do a re-review quickly too.

Sorry for the slow responses.

rclcpp/include/rclcpp/node_handle.hpp
methylDragon: The current status of this is that the getters have all been removed on the aggregating `NodeInterfaces` class and are instead implemented by each of the support classes that the `NodeInterfaces` object inherits from (via multiple inheritance).

E.g. https://github.com/ros2/rclcpp/pull/2041/files#diff-10a9fcfcaf4778e4e836e88ad35ed243e3087b815bb2b8f46fc786042b3533f4R79

While it's true that the support classes create a bit of boilerplate, they're what's necessary to allow extensions to new types in the future (as well as supporting NodeTs with unconventional getter names for the interfaces (in which case, a new support class can be written by the user for that NodeT.))

Specifically, it allows users to specify:
- [How to get the interface from the passed in NodeT](https://github.com/ros2/rclcpp/pull/2041/files#diff-10a9fcfcaf4778e4e836e88ad35ed243e3087b815bb2b8f46fc786042b3533f4R53)
- [How users are expected to get the interface out from the NodeInterface](https://github.com/ros2/rclcpp/pull/2041/files#diff-10a9fcfcaf4778e4e836e88ad35ed243e3087b815bb2b8f46fc786042b3533f4R56)
- [And how to rebind the interface](https://github.com/ros2/rclcpp/pull/2041/files#diff-10a9fcfcaf4778e4e836e88ad35ed243e3087b815bb2b8f46fc786042b3533f4R59) (though this is probably getting removed) 

---

With the current status then, I'm not sure if a `get<>()` would be able to remove boilerplate compared to the current implementation or we need to constrain the support classes, which reduces extensibility.

In the former case we'll need a get<>() specialization on each support class to define how to access the interface that they store (which is really just renaming the current getter names on each support class..), and in the latter case we need to enforce that every support class has a consistent API (either a static getter name, a static protected/public member) so that we can remove the getters on each support class and have a single overarching one on the NodeInterface.

I think between the two options, I rather just leave it up to the user to define how they want to do their getters (but use the current getter naming convention for the current node interfaces.)

methylDragon: Another thing I was considering was using a tuple to store the interfaces, but that doesn't work because:
- There still needs to be a way to define how to obtain interfaces from a passed in NodeT in a variadic manner
- It means we can't aggregate more than one instance of an identical interface type (e.g. two NodeBaseInterfaces) (with support classes you could do it by defining two different support classes that both store NodeBaseInterfaces)

The only way to make tuples work is to remove the NodeT constructor, which I am very hesitant to do because it significantly cuts down on the amount of code a user has to write to pass around nodes/lifecycle nodes to a method.

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
methylDragon: Okay, I'll make this change, the getter question is still open though

methylDragon: I have it working locally, though I'd like to highlight that aggregating at construction means we're going to see some really janky looking formatting (as opposed to using setters).

```
    auto base_interfaces = std::make_shared<rclcpp::node_interfaces::NodeInterfaces<
          rclcpp::node_interfaces::Base,
          rclcpp::node_interfaces::Clock,
          rclcpp::node_interfaces::Graph
        >>(
      node->get_node_base_interface(),
      node->get_node_clock_interface(),
      node->get_node_graph_interface()
        );
```

rclcpp/include/rclcpp/node.hpp
methylDragon: Will remove!

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
wjwwood: I don't think making this class mutable just for style is the right approach, and you can make the formatting better, e.g.:

```c++
  using MyBaseNodeInterfaces = rclcpp::node_interfaces::NodeInterfaces<
    rclcpp::node_interfaces::Base,
    rclcpp::node_interfaces::Clock,
    rclcpp::node_interfaces::Graph
  >;
  auto base_interfaces = std::make_shared<MyBaseNodeInterfaces>(
    node->get_node_base_interface(),
    node->get_node_clock_interface(),
    node->get_node_graph_interface());
```

I ran that through `ament_uncrustify`.

wjwwood: If you still want to have a iterative approach to creating it, then I would say adding a "builder" pattern would be better than making it mutable, something like:

```c++
  auto base_interfaces_builder  = rclcpp::node_interfaces::NodeInterfacesBuilder<
    rclcpp::node_interfaces::Base,
    rclcpp::node_interfaces::Clock,
    rclcpp::node_interfaces::Graph
  >();
  auto base_interfaces = base_interfaces_builder
    .node_base(node->get_node_base_interface())
    .node_clock(node->get_node_clock_interface())
    .node_graph(node->get_node_graph_interface())
    .build();
```

Or some version of that. Though I think for the added amount of code to test and maintain in rclcpp it isn't worth it. However, I think it is preferable to making the class mutable, because if it is mutable it makes everything using it have to be more considerate, since it could change at any point, making thread-safe code more difficult. However, if you know it's immutable then you can just access it without much worry.

Also, if there were use cases for mutating it after creation then making it mutable would make more sense I think, but as far as I can imagine you'd only really use this when constructing it, which is exactly what the builder pattern is used for.

methylDragon: Fair point! The next iteration is ready!

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17736/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12277/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18379/)

methylDragon:
The windows instability seems to be unrelated

wjwwood:
I opened https://github.com/methylDragon/rclcpp/pull/1 to suggest a different approach for this pr.

methylDragon:
I'll be iterating on the changes incorporated from https://github.com/methylDragon/rclcpp/pull/1

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17826/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12393/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18495/)

rclcpp/include/rclcpp/detail/template_contains.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+struct test_template_contains
+{
+  static_assert(template_contains_v<int, int, double>, "failed");
+  static_assert(template_contains_v<double, int, double>, "failed");
+  static_assert(template_contains_v<int, int>, "failed");
+  static_assert(!template_contains_v<int>, "failed");
+  static_assert(!template_contains_v<int, float>, "failed");
+  static_assert(!template_contains_v<int, float, double>, "failed");
+};
```
ivanpauno, line 55: I would move this to a test file.
It seems unnecessary to have it here.

rclcpp/include/rclcpp/detail/template_unique.hpp
```diff
@@ -0,0 +1,63 @@
...
+{
+  enum { value = true };
+};
+
+namespace
+{
+
+struct test_template_unique
+{
+  static_assert(template_unique_v<int>, "failed");
```
ivanpauno, line 51: same

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+  /// NodeT::SharedPtr Constructor
+  // NOTE(CH3): We cannot dereference the shared_ptr and pass it, because otherwise a single arg
+  //            aggregate constructor call that passes a NodeInterface shared_ptr will fail
+  //            since it will match this constructor instead.
+  //
+  //            Instead, we will make do with doing the nullptr check here and enforce that
+  //            support classes implement support for a shared_ptr<NodeT> constructor.
+  template<typename NodeT>
+  NodeInterfaces(std::shared_ptr<NodeT> node)  // NOLINT(runtime/explicit)
```
ivanpauno, line 144: this one seems unnecessary, the `NodeT & node` alternative should be enough

methylDragon: I would like to keep it so it bakes in a nullptr check. I'll update the note to focus on that

methylDragon: https://github.com/ros2/rclcpp/pull/2041/commits/b99ebbbd77946659f4ddc7ab2953eea9bb4df492

rclcpp/include/rclcpp/detail/template_unique.hpp
methylDragon: https://github.com/ros2/rclcpp/pull/2041/commits/b99ebbbd77946659f4ddc7ab2953eea9bb4df492

rclcpp/include/rclcpp/detail/template_contains.hpp
methylDragon: https://github.com/ros2/rclcpp/pull/2041/commits/b99ebbbd77946659f4ddc7ab2953eea9bb4df492

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17835/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12403/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18502/)

methylDragon:
Re-running windows CI with parantheses around the fold expression (the only issue is cppcheck)

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18507/)

methylDragon:
I realized me adding the parentheses has a ridiculously minuscule chance of changing the behavior for linux machines, but I'm going to run CI just to be sure.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17841/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12409/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18509/)

rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * \param[in] node Node-like object from which to get the node interfaces
+   */
+  template<typename NodeT>
+  NodeInterfaces(NodeT & node)  // NOLINT(runtime/explicit)
+  : NodeInterfacesSupportsT(node)
+  {}
+
+  /// NodeT::SharedPtr Constructor
+  // NOTE(CH3): Have a special SharedPtr constructor to check for nullptr.
```
wjwwood, line 137: Do we need a note like this? If you want to clarify why this constructor exists or when it should be used, then I'd say just put it in the documentation. Either way, I think we should remove this note.

rclcpp/test/rclcpp/node_interfaces/detail/test_template_utils.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
wjwwood, line 1: I realize this might be based on another test file (copy-paste-edit), but personally, I think the copyright year should be 2022, since all the substantial content is new (and written this year).

rclcpp/test/rclcpp/CMakeLists.txt
```diff
@@ -262,6 +267,11 @@ ament_add_gtest(test_node_interfaces__node_waitables
 if(TARGET test_node_interfaces__node_waitables)
   target_link_libraries(test_node_interfaces__node_waitables ${PROJECT_NAME} mimick)
 endif()
+ament_add_gtest(test_node_interfaces__test_template_utils
+  node_interfaces/detail/test_template_utils.cpp)
```
wjwwood, line 271: Since this file isn't actually a gtest, is this ok? I'm actually not sure if it is or not. gtest command-line arguments, for example, will not work with this executable.

Maybe instead this should just be a target (not a gtest target) or maybe the `test_template_utils.cpp` should have the skeleton of a gtest main function and stuff in it.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -66,6 +66,7 @@
 #include "rclcpp/node_interfaces/node_base_interface.hpp"
 #include "rclcpp/node_interfaces/node_clock_interface.hpp"
 #include "rclcpp/node_interfaces/node_graph_interface.hpp"
+#include "rclcpp/node_interfaces/node_interfaces.hpp"
```
wjwwood, line 69: Is this needed?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node_impl.hpp
```diff
@@ -33,6 +33,7 @@
 #include "rclcpp/create_publisher.hpp"
 #include "rclcpp/create_service.hpp"
 #include "rclcpp/create_subscription.hpp"
+#include "rclcpp/node_interfaces/node_interfaces.hpp"
```
wjwwood, line 36: Is this needed?

wjwwood (commented):
A few small comments, but otherwise lgtm

rclcpp/test/rclcpp/CMakeLists.txt
```diff
@@ -262,6 +267,11 @@ ament_add_gtest(test_node_interfaces__node_waitables
 if(TARGET test_node_interfaces__node_waitables)
   target_link_libraries(test_node_interfaces__node_waitables ${PROJECT_NAME} mimick)
 endif()
+add_library(test_node_interfaces__test_template_utils  # Compile time test
+  node_interfaces/detail/test_template_utils.cpp)
+if(TARGET test_node_interfaces__test_template_utils)
+  target_link_libraries(test_node_interfaces__test_template_utils ${PROJECT_NAME})
+endif()
```
wjwwood, line 274: I think we need to check some other things that `ament_add_gtest()` does for you. For example, making sure it's not part of the "all" target and instead part of the "test" target, etc.

Maybe all we need is https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html and then add it to the `test` make target. I'm not sure what else might be needed. Might want to ask someone else on the ros team to review it.

wjwwood (approved):
lgtm, after checking on the compile-only test

rclcpp/test/rclcpp/CMakeLists.txt
methylDragon: Resolved with a no-op gtest

Crola1702:
Hey @methylDragon @wjwwood, this PR broke RHEL nightly builds when building rclcpp. Can you please take a look?:

* RHEL Debug: https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_debug/1371/consoleFull
* RHEL Release: https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_release/1370/consoleFull

Log output:

<details>

```
--- stderr: rclcpp
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp: In instantiation of ‘[Krclcpp::node_interfaces::NodeInterfaces<InterfaceTs>::NodeInterfaces(std::shared_ptr<_Yp>) [with NodeT = rclcpp::Node; InterfaceTs = {rclcpp::node_interfaces::NodeBaseInterface, rclcpp::node_interfaces::NodeGraphInterfac[K’:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp:47:[K   required from here
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp:153:[Kerro[Kuse of deleted functio[Krclcpp::Node::Node(const rclcpp::Nod[K’
   : NodeInterfaces([Knode ? *node : throw std::invalid_argument("given node pointer is nullpt[K)
                    [K~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/context.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/utilities.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/build/rclcpp/include/rclcpp/logging.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/client.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/callback_group.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp:1578:[Knot[Kdeclared here
   RCLCPP_DISABLE_COPY([KN[K)
                       [K^[K
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/macros.hpp:27[Knot[Kin definition of macr[KRCLCPP_DISABLE_C[K’
   [K__VA_ARG[K(const __VA_ARGS__ &) = delete; \
   [K^~~~~~~~[K
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp: In instantiation of ‘[Krclcpp::node_interfaces::NodeInterfaces<InterfaceTs>::NodeInterfaces(std::shared_ptr<_Yp>) [with NodeT = rclcpp::Node; InterfaceTs = {rclcpp::node_interfaces::NodeBaseInterfac[K’:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp:58:[K   required from here
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp:153:[Kerro[Kuse of deleted functio[Krclcpp::Node::Node(const rclcpp::Nod[K’
   : NodeInterfaces([Knode ? *node : throw std::invalid_argument("given node pointer is nullpt[K)
                    [K~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/context.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/utilities.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/build/rclcpp/include/rclcpp/logging.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/client.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/callback_group.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp:1578:[Knot[Kdeclared here
   RCLCPP_DISABLE_COPY([KN[K)
                       [K^[K
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/macros.hpp:27[Knot[Kin definition of macr[KRCLCPP_DISABLE_C[K’
   [K__VA_ARG[K(const __VA_ARGS__ &) = delete; \
   [K^~~~~~~~[K
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp: In instantiation of ‘[Krclcpp::node_interfaces::NodeInterfaces<InterfaceTs>::NodeInterfaces(std::shared_ptr<_Yp>) [with NodeT = rclcpp::Node; InterfaceTs = {rclcpp::node_interfaces::NodeBaseInterface, rclcpp::node_interfaces::NodeClockInterface, rclcpp::node_interfaces::NodeGraphInterface, rclcpp::node_interfaces::NodeLoggingInterface, rclcpp::node_interfaces::NodeTimersInterface, rclcpp::node_interfaces::NodeTopicsInterface, rclcpp::node_interfaces::NodeServicesInterface, rclcpp::node_interfaces::NodeWaitablesInterface, rclcpp::node_interfaces::NodeParametersInterface, rclcpp::node_interfaces::NodeTimeSourceInterfac[K’:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp:105:[K   required from here
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp:153:[Kerro[Kuse of deleted functio[Krclcpp::Node::Node(const rclcpp::Nod[K’
   : NodeInterfaces([Knode ? *node : throw std::invalid_argument("given node pointer is nullpt[K)
                    [K~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/context.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/utilities.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/build/rclcpp/include/rclcpp/logging.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/client.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/callback_group.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp:1578:[Knot[Kdeclared here
   RCLCPP_DISABLE_COPY([KN[K)
                       [K^[K
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/macros.hpp:27[Knot[Kin definition of macr[KRCLCPP_DISABLE_C[K’
   [K__VA_ARG[K(const __VA_ARGS__ &) = delete; \
   [K^~~~~~~~[K
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp: In instantiation of ‘[Krclcpp::node_interfaces::NodeInterfaces<InterfaceTs>::NodeInterfaces(std::shared_ptr<_Yp>) [with NodeT = rclcpp::Node; InterfaceTs = {rclcpp::node_interfaces::NodeBaseInterface, rclcpp::node_interfaces::NodeClockInterface, rclcpp::node_interfaces::NodeGraphInterface, rclcpp::node_interfaces::NodeLoggingInterface, rclcpp::node_interfaces::NodeParametersInterface, rclcpp::node_interfaces::NodeServicesInterface, rclcpp::node_interfaces::NodeTimeSourceInterface, rclcpp::node_interfaces::NodeTimersInterface, rclcpp::node_interfaces::NodeTopicsInterface, rclcpp::node_interfaces::NodeWaitablesInterfac[K’:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp:201:[K   required from here
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp:153:[Kerro[Kuse of deleted functio[Krclcpp::Node::Node(const rclcpp::Nod[K’
   : NodeInterfaces([Knode ? *node : throw std::invalid_argument("given node pointer is nullpt[K)
                    [K~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[K
In file included from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/context.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/utilities.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/build/rclcpp/include/rclcpp/logging.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/client.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/callback_group.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp[K,
                 from [K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp[K:
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/node.hpp:1578:[Knot[Kdeclared here
   RCLCPP_DISABLE_COPY([KN[K)
                       [K^[K
[K/home/jenkins-agent/workspace/nightly_linux-rhel_debug/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/macros.hpp:27[Knot[Kin definition of macr[KRCLCPP_DISABLE_C[K’
   [K__VA_ARG[K(const __VA_ARGS__ &) = delete; \
   [K^~~~~~~~[K
gmake[2]: *** [test/rclcpp/CMakeFiles/test_node_interfaces__node_interfaces.dir/build.make:76: test/rclcpp/CMakeFiles/test_node_interfaces__node_interfaces.dir/node_interfaces/test_node_interfaces.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2943: test/rclcpp/CMakeFiles/test_node_interfaces__node_interfaces.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
gmake: *** [Makefile:146: all] Error 2
```

</details>

I notice:
* [node_interfaces.hpp:153](https://github.com/ros2/rclcpp/pull/2041/files#diff-41454f18fae63765fd951ef505f09b77ee1efe5f3945ee6d6a472f481a0b9ad3R153): `use of deleted function`

methylDragon:
@Crola1702 I'm looking at it!

methylDragon:
> Hey @methylDragon @wjwwood, this PR broke RHEL nightly builds when building rclcpp. Can you please take a look?:
> 
> * RHEL Debug: https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_debug/1371/consoleFull
> * RHEL Release: https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_release/1370/consoleFull
> 
> Log output:
> 
> I notice:
> 
> * [node_interfaces.hpp:153](https://github.com/ros2/rclcpp/pull/2041/files#diff-41454f18fae63765fd951ef505f09b77ee1efe5f3945ee6d6a472f481a0b9ad3R153): `use of deleted function`

Noting that this was fixed in 
https://github.com/ros2/rclcpp/pull/2075

PR #2046, opened 2022-11-12
Make get_rcl_allocator a function family

tylerjw:
This is a rebase of 
- https://github.com/ros2/rclcpp/pull/1324

I don't understand this code but would really like to be able to use asan in my tests.

If someone could explain this it might be helpful. My best guess from reading the other PR is that C++ allocators just don't work for some reason (or this implementation of them does not) and instead we are using some struct from rcl anyway.

### DIFF
--- a/rclcpp/include/rclcpp/allocator/allocator_common.hpp
+++ b/rclcpp/include/rclcpp/allocator/allocator_common.hpp
@@ -30,14 +30,19 @@ namespace allocator
 template<typename T, typename Alloc>
 using AllocRebind = typename std::allocator_traits<Alloc>::template rebind_traits<T>;
 
-template<typename Alloc>
-void * retyped_allocate(size_t size, void * untyped_allocator)
+/// Return the equivalent rcl_allocator_t for the C++ standard allocator.
+/**
+ * This assumes that the user intent behind both allocators is the
+ * same: Using system malloc for allocation.
+ *
+ * If you're using a custom allocator in ROS, you'll need to provide
+ * your own overload for this function.
+ */
+template<typename T>
+rcl_allocator_t get_rcl_allocator(std::allocator<T> allocator)
 {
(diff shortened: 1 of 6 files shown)


tylerjw:
@clalancette could you review this and let me know if this was what you were expecting in a rebase of this? Do you understand why we have custom allocators? This seems like really error-prone code to be writing without sanitizers in CI.

rclcpp/include/rclcpp/allocator/allocator_common.hpp
```diff
@@ -30,14 +30,19 @@ namespace allocator
...
+template<typename T>
+rcl_allocator_t get_rcl_allocator(std::allocator<T> allocator)
 {
-  auto typed_allocator = static_cast<Alloc *>(untyped_allocator);
-  if (!typed_allocator) {
-    throw std::runtime_error("Received incorrect allocator type");
-  }
-  return std::allocator_traits<Alloc>::allocate(*typed_allocator, size);
+  (void)allocator;     // Remove warning
+  return rcl_get_default_allocator();
```
tylerjw, line 45: Does this seem useless, and should users call the underlying rcl function? Maybe this is needed for backward compatibility or something?

tylerjw:
@wjwwood have you had a chance to post your work somewhere?

haudren-woven:
I know that pings on PRs can be quite annoying, but I'd also love to see this fixed! @wjwwood could you maybe explain what the issue with the original code would be? It seems to be fairly straightforward in using the default rcl allocator. Is there a problem with that?

Pleune:
We currently have to maintain a branch with these fixed to allow using asan with rclcpp. These fixes work for us. @wjwwood, sorry again for a ping on this, but what are the issue you are seeing? Is there something we could attempt to fix to allow a merge here?

jmachowinski:
I am going to close this PR in favor for https://github.com/ros2/rclcpp/pull/3058 which will be merged in the near future.

PR #2048, opened 2022-11-18
Add in a warning for a KeepLast depth of 0.

clalancette:
It really doesn't make much sense to have a KeepLast depth of 0; no data could possibly be stored.  Indeed, the underlying DDS implementations don't actually support this.  It currently "works" because a KeepLast depth of 0 is assumed to be system default, so the RMW layer typically chooses "1" instead.  But this isn't something we should be encouraging users to do, so add a warning.

I'll note that I could easily be convinced to make this throw an exception instead.  That would be a much stronger warning to users not to do this, but it does break existing practice so has some risk of downstream breakage.  Happy to discuss it.

### DIFF
--- a/rclcpp/src/rclcpp/qos.cpp
+++ b/rclcpp/src/rclcpp/qos.cpp
@@ -84,6 +84,15 @@ QoS::QoS(
   const rmw_qos_profile_t & initial_profile)
 : rmw_qos_profile_(initial_profile)
 {
+  if (qos_initialization.history_policy == RMW_QOS_POLICY_HISTORY_KEEP_LAST &&
+    qos_initialization.depth == 0)
+  {
+    RCLCPP_WARN_ONCE(
+      rclcpp::get_logger(
+        "rclcpp"),
+      "A zero depth with KEEP_LAST doesn't make sense; no data could be stored."
+      "This will be interpreted as SYSTEM_DEFAULT");
+  }
   rmw_qos_profile_.history = qos_initialization.history_policy;
   rmw_qos_profile_.depth = qos_initialization.depth;
 }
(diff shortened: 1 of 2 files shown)


alsora (approved):
I usually think that user-specified settings that do not make sense should result in immediate crashes that are not possible to ignore, to make the robot fail as soon as possible.

However, here there's an additional complexity: what happens if someone changes the QoS at runtime using the qos_parameters_overrides?
In this case we should rather reject the change as we don't want people to be able to kill robots that are already running.
This probably is out of scope for the tiny change that you did here.

ivanpauno:
> However, here there's an additional complexity: what happens if someone changes the QoS at runtime using the qos_parameters_overrides?
In this case we should rather reject the change as we don't want people to be able to kill robots that are already running.

The parameters overrides are only read once at startup, they cannot be changed aferwards.

rclcpp/src/rclcpp/qos.cpp
```diff
@@ -67,7 +69,15 @@ KeepAll::KeepAll()
...
-{}
+{
+  if (depth == 0) {
+    RCLCPP_WARN_ONCE(
+      rclcpp::get_logger(
+        "rclcpp"),
+      "A zero depth with KEEP_LAST doesn't make sense; no data could be stored."
+      "This will be interpreted as SYSTEM_DEFAULT");
+  }
+}
```
ivanpauno, line 80: I would rather warn in `QoS::QoS()` and `QoS::keep_last()`, so we cover all the cases.

clalancette: Good point.  I've actually *added* it to both `QoS::QoS` and to `QoS::keep_last`, because it is possible to construct a `KeepLast` object without going through those APIs.

ivanpauno (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17719/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12265/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18345/)

mjcarroll (approved):

clalancette:
CI is clean here, so merging.

PR #2050, opened 2022-11-22
Mark rclcpp::Clock::now() as const

sloretz:
I think `now()` can be marked as `const`. All it does is [call rcl_clock_get_now()](https://github.com/ros2/rclcpp/blob/91bc312190e61c8915b0152e5c927c987fc22d6c/rclcpp/src/rclcpp/clock.cpp#L71-L80) which [doesn't look like it modifies the clock to me](https://github.com/ros2/rcl/blob/e4f7e1367dfda83d3e309397f9201642ebe38618/rcl/src/rcl/time.c#L268-L280).

* When `get_now` is `rcl_get_ros_time`: https://github.com/ros2/rcl/blob/e4f7e1367dfda83d3e309397f9201642ebe38618/rcl/src/rcl/time.c#L65-L73
* When `get_now` is `rcl_get_system_time`: https://github.com/ros2/rcl/blob/e4f7e1367dfda83d3e309397f9201642ebe38618/rcl/src/rcl/time.c#L44-L48
* When `get_now` is `rcl_get_steady_time`: https://github.com/ros2/rcl/blob/e4f7e1367dfda83d3e309397f9201642ebe38618/rcl/src/rcl/time.c#L36-L40

### DIFF
--- a/rclcpp/include/rclcpp/clock.hpp
+++ b/rclcpp/include/rclcpp/clock.hpp
@@ -77,7 +77,7 @@ public:
    */
   RCLCPP_PUBLIC
   Time
-  now();
+  now() const;
 
   /**
    * Sleep until a specified Time, according to clock type.
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -67,7 +67,7 @@ Clock::Clock(rcl_clock_type_t clock_type)
 Clock::~Clock() {}
 
 Time
-Clock::now()
+Clock::now() const
 {
   Time now(0, 0, impl_->rcl_clock_.type);
 


sloretz:
CI (build: `--packages-above-and-dependencies rclcpp` test: `--packages-select rclcpp`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17684/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12232/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18273/)

clalancette (approved):
Seems reasonable to me with green CI (though I think we should run tests with `--packages-above rclcpp`, just to make sure no downstream packages have a problem with this).

ivanpauno (approved):

clalancette:
> What might happen downstream at run time that needs to be tested?

Actually, that's a good point.  I missed that the build was `--packages-above-and-dependencies`, so that's good for me.  Sorry for the noise.

alsora (approved):

Timple:
Can this be backported to humble?

Would save me a workaround: https://github.com/nobleo/ros2_fmt_logger/blob/14e0cb5e0407e33aaecf88787296208b896a7f17/include/ros2_fmt_logger/ros2_fmt_logger.hpp#L572..L574

fujitatomoya:
@Timple 

this is ABI breaking change. application compiled against the old version would expect the non-const signature, so i think this would break the user space. i do not think we can backport this?

Timple:
Ah, then nevermind. I'll just wait for the EOL :smile:

Issue #2047, opened 2022-11-16
get_parameter variants with noexcept specifier


bvolkmer:
## Feature request
`Node::get_parameter()` variants, that throw no exceptions (2 functions args and all get_parameter_or) should be declared with the `noexcept` specifier.

#### Feature description
Functions that (as explicitly stated in the docs) throw no exception should be declared noexcept.
Static analyzers (like clang-tidy with [bugprone-exception-escape](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/exception-escape.html)) complain about possible exceptions in `get_parameter()` if the surrounding function should not throw exceptions.

fujitatomoya:
@bvolkmer it would be really appreciated if you can review https://github.com/ros2/rclcpp/pull/2052. thanks!

PR #2052, opened 2022-11-28
add noexcept to get_parameter methods.

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/2047

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -777,7 +777,7 @@ public:
    */
   RCLCPP_PUBLIC
   bool
-  get_parameter(const std::string & name, rclcpp::Parameter & parameter) const;
+  get_parameter(const std::string & name, rclcpp::Parameter & parameter) const noexcept;
 
   /// Get the value of a parameter by the given name, and return true if it was set.
   /**
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
@@ -162,7 +162,7 @@ public:
   bool
   get_parameter(
     const std::string & name,
-    rclcpp::Parameter & parameter) const override;
+    rclcpp::Parameter & parameter) const noexcept override;
 
   RCLCPP_PUBLIC
   bool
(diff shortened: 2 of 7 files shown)


fujitatomoya (commented):
> and all get_parameter_or

`get_parameter_or` uses template, that could possibly throw `rclcpp::ParameterTypeException`.

fujitatomoya:
@ivanpauno minor fix, could you take a look?

ivanpauno:
Why do we need this?
`noexcept` is a promise that we aren't going to throw exceptions from this method in the future (if not, we will break API).
I'm not sure if we should make that promise.

@wjwwood any opinion?

bvolkmer:
> Why do we need this? `noexcept` is a promise that we aren't going to throw exceptions from this method in the future (if not, we will break API). I'm not sure if we should make that promise.
> 
> @wjwwood any opinion?

But that exactly is the reason to use `noexcept`. If you're going to throw exceptions in the future, it will not silently compile, if someone expects it, not to. As exceptions are most probably a rare occasion (and thus hard to test), it is better to get warnings of a static analyzer after updating an API.

AFAIK, an unfulfilled `noexcept` will throw at most a warning in common compilers.

ivanpauno:
> AFAIK, an unfulfilled noexcept will throw at most a warning in common compilers.

Maybe, I havent' checked if a warning is shown or not. 
It's true that it compiles, and if in the future an exception is thrown I think `std::terminate()` is called.

But I'm not sure if these methods will need to throw an exception in the future or not, so I don't see the point of making a promise now.

e.g. if we add parameter name validation in the future we may need to throw an exception for invalid names in these methods.

fujitatomoya:
> But I'm not sure if these methods will need to throw an exception in the future or not, so I don't see the point of making a promise now.

me neither, i was thinking that with current implementation and code quality, this is worth to add.

ivanpauno:
could you explain what's the benefit of adding that?

fujitatomoya:
it just explicitly declares for user application that this method does(will) not throw exception, and provides safety freedom to the compiler to generator better code for performance(but this could be really small in this case, not really sure).

> if we add parameter name validation in the future we may need to throw an exception for invalid names in these methods.

this is totally correct, adding `noexcept` leads to fragile user interface. if we already have (or may have) such plan, i guess that we can skip adding `noexcept` at this moment.

PR #2055, opened 2022-12-03
Add alias libs for rclcpp all following form of rclcpp::X

RFRIEDM-Trimble:
Due to some requests in gscam2's upgrades to modern CMake, it would like to consume alias library targets for all its dependencies. This PR is to add them to rclcpp. 

Signed-off-by: Ryan Friedman <ryan_friedman@trimble.com>

### DIFF
--- a/rclcpp_components/CMakeLists.txt
+++ b/rclcpp_components/CMakeLists.txt
@@ -22,6 +22,7 @@ find_package(rcpputils REQUIRED)
 
 # Add an interface library that can be dependend upon by libraries who register components
 add_library(component INTERFACE)
+add_library(rclcpp::component ALIAS component)
 target_include_directories(component INTERFACE
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
   "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
@@ -34,6 +35,7 @@ add_library(
   SHARED
   src/component_manager.cpp
 )
+add_library(rclcpp::component_manager ALIAS component_manager)
 target_include_directories(component_manager PUBLIC
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
   "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
@@ -96,6 +98,7 @@ if(BUILD_TESTING)
 
   set(components "")
   add_library(test_component SHARED test/components/test_component.cpp)
+  add_library(rclcpp::test_component ALIAS test_component)
   target_link_libraries(test_component PRIVATE component)
   #rclcpp_components_register_nodes(test_component "test_rclcpp_components::TestComponent")
   set(components "${components}test_rclcpp_components::TestComponentFoo;$<TARGET_FILE:test_component>\n")
--- a/rclcpp/CMakeLists.txt
+++ b/rclcpp/CMakeLists.txt
@@ -179,6 +179,7 @@ foreach(interface_file ${interface_files})
 endforeach()
 
 add_library(${PROJECT_NAME} ${${PROJECT_NAME}_SRCS})
+add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_17)
 # TODO(wjwwood): address all deprecation warnings and then remove this
 if(WIN32)
--- a/rclcpp_lifecycle/CMakeLists.txt
+++ b/rclcpp_lifecycle/CMakeLists.txt
@@ -28,6 +28,7 @@ add_library(rclcpp_lifecycle
   src/state.cpp
   src/transition.cpp
 )
+add_library(rclcpp::rclcpp_lifecycle ALIAS ${PROJECT_NAME})
 target_include_directories(${PROJECT_NAME}
   PUBLIC
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"


rclcpp_components/CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ find_package(rcpputils REQUIRED)
 
 # Add an interface library that can be dependend upon by libraries who register components
 add_library(component INTERFACE)
+add_library(rclcpp::component ALIAS component)
+
```
fujitatomoya, line 26: can we avoid unnecessary change here?

fujitatomoya (commented):
lgtm with a nitpick

rclcpp_components/CMakeLists.txt
RFRIEDM-Trimble: Done, rebased without that.

fujitatomoya (approved):

fujitatomoya:
either @ivanpauno or @hidmic could you take a look? i would like to have another review.

rclcpp_components/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ find_package(rcpputils REQUIRED)
 
 # Add an interface library that can be dependend upon by libraries who register components
 add_library(component INTERFACE)
```
alsora, line 24: Should `rclcpp` be `${PROJECT_NAME} `?

rclcpp_components/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ find_package(rcpputils REQUIRED)
...
+add_library(rclcpp::component ALIAS component)
```
alsora, line 25: Should `rclcpp` be `${PROJECT_NAME} `?

RFRIEDM-Trimble: I had pondered this, and believe it is up to the project maintainers to define a convention. See this excerpt from Craig Scott's book, Professional CMake.

```
Target names need not be related to the project name. It is common to see tutorials and examples
use a variable for the project name and reuse that variable for the name of an executable target.

This only works for the most basic of projects and encourages a number of bad habits. Consider the
project name and executable name as being separate, even if initially they start out the same. Set
the project name directly rather than via a variable, choose a target name according to what the
target does rather than the project it is part of and assume the project will eventually need to define
more than one target. This reinforces better habits which will be important when working on more
complex multi-target projects.
```

Although it is mentioning executable, this practice is very common in ROS2 for libraries being named the same as the project.  I have not seen a convention for ALIAS libraries yet, other than 90% of ROS libraries name a single target the same as the project, which does not seem to be recommended. It also would break ABI to change existing library target names, however, it seems a  good time to pick convention for ALIAS libs.  As far as a naming convention, it ideally can handle the following scenarios. 

- A ROS2 repo with a single ROS package that has single target
- A ROS2 repo with a single ROS package that has multiple targets (executables, libraries, etc)
- A ROS2 repo with multiple ROS packages that have multiple target(s)

Let me know your ideas, and I can update this PR to match, as well as perhaps clarify the convention in tutorials/docs if accepted.

ivanpauno: This should be `rclcpp_components::component`.
That's also the exported target name

ivanpauno (changes requested):
The alias only applies to the local CMakeLists.txt, it cannot be used by dependencies.
We're already exporting target names namespaced with the project name, .e.g.:
- rclcpp::rclcpp
- rclcpp_components::component
- etc

You can already use those without this PR.

clalancette:
I have the same question as Ivan; why do we need this?  We already export the `rclcpp::rclcpp` targets to downstream (and we use it in a few places, so we know it works).  What does this add?

RFRIEDM-Trimble:
> The alias only applies to the local CMakeLists.txt, it cannot be used by dependencies. We're already exporting target names namespaced with the project name, .e.g.:
> 
> * rclcpp::rclcpp
> * rclcpp_components::component
> * etc
> 
> You can already use those without this PR.

I was not aware of this restriction having used alias libs before in CMake without ament that export automatically. Since this PR seems to provide no purpose, I will close it. [In my search, the ALIAS targets did not appear. ](https://github.com/ros2/rclcpp/search?q=ALIAS). Thank you for the info. 

If anyone thinks this is still worthwhile, please let me know, and I'll apply the suggestions and get it in.

ivanpauno:
@RFRIEDM-Trimble  I think you're confusing ALIAS targets with exported namespaced targets.
The alias is only local and cannot be exported.
It's sometimes defined with the same name than the exported target for convenience (that's particularly needed if something can be compiled by the cmake project or be already present in the system and find-packaged).

`ament_export_target()` is what makes targets available to other libraries https://github.com/ament/ament_cmake/blob/8454baf87487904c2be6f74e72c6e3f6506621e8/ament_cmake_export_targets/cmake/ament_cmake_export_targets_package_hook.cmake#L28-L33, and it's being used in rclcpp libraries.

RFRIEDM-Trimble:
> @RFRIEDM-Trimble I think you're confusing ALIAS targets with exported namespaced targets. The alias is only local and cannot be exported. It's sometimes defined with the same name than the exported target for convenience (that's particularly needed if something can be compiled by the cmake project or be already present in the system and find-packaged).
> 
> `ament_export_target()` is what makes targets available to other libraries https://github.com/ament/ament_cmake/blob/8454baf87487904c2be6f74e72c6e3f6506621e8/ament_cmake_export_targets/cmake/ament_cmake_export_targets_package_hook.cmake#L28-L33, and it's being used in rclcpp libraries.

Many thanks for the clarification. Yes, I was missing this point.  I tested if you try supplying a target that doesn't exist to ament_export_targets, and it fails with a descriptive error.

PR #2058, opened 2022-12-06
Fix url to API doc in readme

jrutgeer:
Changed README to point to the API doc via the rclcpp package info page at the ROS Index.

### DIFF
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -792,16 +792,14 @@ protected:
   async_send_request_impl(const Request & request, CallbackInfoVariant value)
   {
     int64_t sequence_number;
+    std::lock_guard<std::mutex> lock(pending_requests_mutex_);
     rcl_ret_t ret = rcl_send_request(get_client_handle().get(), &request, &sequence_number);
     if (RCL_RET_OK != ret) {
       rclcpp::exceptions::throw_from_rcl_error(ret, "failed to send request");
     }
-    {
-      std::lock_guard<std::mutex> lock(pending_requests_mutex_);
-      pending_requests_.try_emplace(
-        sequence_number,
-        std::make_pair(std::chrono::system_clock::now(), std::move(value)));
-    }
+    pending_requests_.try_emplace(
+      sequence_number,
+      std::make_pair(std::chrono::system_clock::now(), std::move(value)));
     return sequence_number;
   }
 


fujitatomoya (commented):
@jrutgeer thanks for the contribution, this PR surely fixes one link, which is good. but i can see there are many places use `docs.ros2.org`, so those probably need to be changed as well. besides we are having discussion on https://github.com/ros2/ros2/issues/1354.

README.md
```diff
@@ -8,7 +8,7 @@ rclcpp provides the standard C++ API for interacting with ROS 2.
 
 `#include "rclcpp/rclcpp.hpp"` allows use of the most common elements of the ROS 2 system.
 
-Visit the [rclcpp API documentation](http://docs.ros2.org/latest/api/rclcpp/) for a complete list of its main components.
+The link to the latest API documentation can be found on the rclcpp package info page, at the [ROS Index](https://index.ros.org/p/rclcpp/).
```
ivanpauno, line 11: Maybe link https://docs.ros.org/en/rolling/p/rclcpp/ directly?

```suggestion
Visit the [rclcpp API documentation](https://docs.ros.org/en/rolling/p/rclcpp/) for a complete list of its main components.
```

ivanpauno (commented):
what's wrong with the current link?

README.md
jrutgeer: @ivanpauno I considered that, but a link to the ROS Index has the added benefit of being instructional to new users (that in general API links for packages can be found through the ROS Index).
But that's up to personal opinion...  Main importance is that it's a correct link, as any new user is likely to search for rclcpp API docs before any other package's API, and google isn't helpful.

clalancette: Yeah, I have to agree with @jrutgeer here that showing people the general place to find API documentation is a good thing, so I would vote to go with the link to index.ros.org.

clalancette (approved):
Looks good to me, but I'll wait a bit longer before merging to get other opinions.

clalancette:
We've had no further reviews, so I'm happy to merge it as-is.

@jrutgeer I just realized that the DCO bot is unhappy because the commit does not have the `Signed-off-by:` line.  Can you please rebase, add that, and then force-push it?  Once that is done we can merge.

jrutgeer:
@clalancette I tried to do that, but I think I failed.
Also when I press compare above it seems that 18 files were changed??
I don´t know how come I could have had write access to this repository and not only my local fork?

clalancette:
> I don´t know how come I could have had write access to this repository and not only my local fork?

You only have access to your local fork.  Because you did a merge, you brought in changes that happened since your fork.

In this case, the easiest thing to do is to just make a brand-new commit with a Signed-off-by line, and then open a new PR.

PR #2059, opened 2022-12-08
Set explicitly callback type

mauropasse:
This is to fix seg-faults happening when compiling with `-O2` or `-O3` optimizations.
They happened because we did:

https://github.com/ros2/rclcpp/blob/338eed0c060ab58608fc6e44c6936bde2c4e061e/rclcpp/include/rclcpp/client.hpp#L289

assuming that `auto` was deduced to be a `std::function<void(size_t)>`. 

And then, to the trampoline, we say we were passing a function returning `void*` and with an arg of type `size_t`
https://github.com/ros2/rclcpp/blob/338eed0c060ab58608fc6e44c6936bde2c4e061e/rclcpp/include/rclcpp/client.hpp#L314-L316
But when compiling with optimizations, the `auto new_callback = ...` signature seems to change to other type. So then the `trampoline` function failed when did the cast back:
https://github.com/ros2/rclcpp/blob/338eed0c060ab58608fc6e44c6936bde2c4e061e/rclcpp/include/rclcpp/detail/cpp_callback_trampoline.hpp#L60

### DIFF
--- a/rclcpp/include/rclcpp/detail/cpp_callback_trampoline.hpp
+++ b/rclcpp/include/rclcpp/detail/cpp_callback_trampoline.hpp
@@ -41,6 +41,7 @@ namespace detail
  * so no exceptions should be thrown at this point, and doing so results in
  * undefined behavior.
  *
+ * \tparam UserDataRealT Declared type of the passed function
  * \tparam UserDataT Deduced type based on what is passed for user data,
  *   usually this type is either `void *` or `const void *`.
  * \tparam Args the arguments being passed to the callback
@@ -50,6 +51,7 @@ namespace detail
  * \returns whatever the callback returns, if anything
  */
 template<
+  typename UserDataRealT,
   typename UserDataT,
   typename ... Args,
   typename ReturnT = void
@@ -57,7 +59,7 @@ template<
 ReturnT
 cpp_callback_trampoline(UserDataT user_data, Args ... args) noexcept
 {
-  auto & actual_callback = *reinterpret_cast<const std::function<ReturnT(Args...)> *>(user_data);
+  auto & actual_callback = *static_cast<const UserDataRealT *>(user_data);
   return actual_callback(args ...);
 }
 
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -286,7 +286,7 @@ public:
               "is not callable.");
     }
 
-    auto new_callback =
+    std::function<void(size_t)> new_callback =
       [callback, this](size_t number_of_responses) {
         try {
           callback(number_of_responses);
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -312,7 +312,7 @@ public:
     // This two-step setting, prevents a gap where the old std::function has
     // been replaced but the middleware hasn't been told about the new one yet.
     set_on_new_response_callback(
-      rclcpp::detail::cpp_callback_trampoline<const void *, size_t>,
+      rclcpp::detail::cpp_callback_trampoline<decltype(new_callback), const void *, size_t>,
       static_cast<const void *>(&new_callback));
 
     // Store the std::function to keep it in scope, also overwrites the existing one.
@@ -320,7 +320,8 @@ public:
 
     // Set it again, now using the permanent storage.
     set_on_new_response_callback(
-      rclcpp::detail::cpp_callback_trampoline<const void *, size_t>,
+      rclcpp::detail::cpp_callback_trampoline<
+        decltype(on_new_response_callback_), const void *, size_t>,
       static_cast<const void *>(&on_new_response_callback_));
   }
 
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -286,7 +286,7 @@ public:
               "is not callable.");
     }
 
-    std::function<void(size_t)> new_callback =
+    auto new_callback =
       [callback, this](size_t number_of_responses) {
         try {
           callback(number_of_responses);
--- a/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
+++ b/rclcpp/include/rclcpp/experimental/subscription_intra_process_base.hpp
@@ -128,7 +128,7 @@ public:
     }
 
     // Note: we bind the int identifier argument to this waitable's entity types
-    std::function<void(size_t)> new_callback =
+    auto new_callback =
       [callback, this](size_t number_of_events) {
         try {
           callback(number_of_events, static_cast<int>(EntityType::Subscription));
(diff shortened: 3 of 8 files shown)


alsora (approved):

fujitatomoya (commented):
@mauropasse interesting.

> This is to fix seg-faults happening when compiling with -O2 or -O3 optimizations.

could you share the segmentation fault information detail? that might be helpful for other members.

mauropasse:
This is the backtrace. I think the important part is the `#2` where the trampoline does the cast back to `std::function<void(size_t)>`
```
#0  std::_Function_handler<void(long unsigned int, int), rclcpp::executors::EventsExecutorEntitiesCollector::create_waitable_callback(void*)::<lambda(size_t, int)> >::_M_invoke(const std::_Any_data &, unsigned long &&, int &&) (
    __functor=..., __args#0=@0x7fffffffdc80: 0, 
    __args#1=<error reading variable: Cannot access memory at address 0x0>)
    at /usr/include/c++/11/bits/std_function.h:290
#1  0x000055555557c959 in std::function<void (unsigned long)>::operator()(unsigned long) const (
    __args#0=<optimized out>, this=<optimized out>) at /usr/include/c++/11/bits/std_function.h:590
#2  rclcpp::detail::cpp_callback_trampoline<void const*, unsigned long, void> (user_data=<optimized out>)
    at /root/callback_issue/install/include/rclcpp/rclcpp/detail/cpp_callback_trampoline.hpp:61
#3  0x00007ffff74e9b7c in PubListener::set_on_new_event_callback(rmw_event_type_e, void const*, void (*)(void const*, unsigned long)) () from /opt/ros/humble/lib/librmw_fastrtps_shared_cpp.so
#4  0x00007ffff74ee44d in rmw_fastrtps_shared_cpp::__rmw_event_set_callback(rmw_event_s*, void (*)(void const*, unsigned long), void const*) () from /opt/ros/humble/lib/librmw_fastrtps_shared_cpp.so
#5  0x00007ffff7ee3472 in rclcpp::QOSEventHandlerBase::set_on_new_event_callback (this=this@entry=0x55555585cff0, 
    callback=callback@entry=0x55555557c930 <rclcpp::detail::cpp_callback_trampoline<void const*, unsigned long, void>(void const*, unsigned long)>, user_data=user_data@entry=0x7fffffffdda0)
    at /root/callback_issue/src/rclcpp/rclcpp/src/rclcpp/qos_event.cpp:79
#6  0x0000555555580a57 in rclcpp::QOSEventHandlerBase::set_on_ready_callback(std::function<void (unsigned long, int)>) (this=this@entry=0x55555585cff0, callback=...)
    at /root/callback_issue/install/include/rclcpp/rclcpp/qos_event.hpp:187

```

fujitatomoya:
> rclcpp::detail::cpp_callback_trampoline<void const*, unsigned long, void> (user_data=<optimized out>)

this seems correct to me.

> 0x000055555557c959 in std::function<void (unsigned long)>::operator()(unsigned long) const (

So does this line.

> __args#1=<error reading variable: Cannot access memory at address 0x0>)

callback function can be accessible, but cannot access 1st argument?

any thoughts? @iuhilnehc-ynos

mauropasse:
Check this example:  
https://onlinegdb.com/pdJ6u2gUs
```
  auto auto_function = [](size_t n){cout << n << endl;};
  std::function<void(size_t)> explicit_function = [](size_t n){cout << n << endl;};

  if (std::is_same<decltype(auto_function), decltype(explicit_function)>::value)
  {
    cout<<"Functions are same type" << endl;
  } else {
    cout<<"Functions are different type" << endl; # This is printed
  }
```

fujitatomoya:
@mauropasse thanks for the study.

```bash
(gdb) whatis auto_function
type = <lambda(size_t)>

(gdb) whatis explicit_function
type = std::function<void(long unsigned int)>

(gdb) ptype auto_function
type = struct <lambda(size_t)> {
}

(gdb) ptype explicit_function
type = class std::function<void(long unsigned int)> [with _Signature = void (unsigned long)]
        : public std::_Maybe_unary_or_binary_function<void, unsigned long>, private std::_Function_base {
  private:
    _Invoker_type _M_invoker;

  public:
    function(void);
    function(std::nullptr_t);
    function(const std::function<void(long unsigned int)> &);
    function(std::function<void(long unsigned int)> &&);
    std::function<void(long unsigned int)> & operator=(const std::function<void(long unsigned int)> &);
    std::function<void(long unsigned int)> & operator=(std::function<void(long unsigned int)> &&);
    std::function<void(long unsigned int)> & operator=(std::nullptr_t);
    void swap(std::function<void(long unsigned int)> &);
    operator bool(void) const;
    void operator()(unsigned long) const;
    const std::type_info & target_type(void) const;
    void function<main()::<lambda(size_t)> >(<lambda(size_t)>);

  private:
    typedef void (*_Invoker_type)(const std::_Any_data &, unsigned long &&);
}
```

reference: https://stackoverflow.com/questions/36030589/i-cannot-pass-lambda-as-stdfunction,

iuhilnehc-ynos:
This fix seems good to me, but I think it's just a workaround.

According to https://github.com/ros2/rclcpp/blob/2e39e0980365e08ff7d8e9a6f672aff78dc682cc/rclcpp/include/rclcpp/detail/cpp_callback_trampoline.hpp#L37-L38, I think that the original intention of `cpp_callback_trampoline` can deal with the lambda.

To cast the lambda to std::function by https://github.com/ros2/rclcpp/blob/2e39e0980365e08ff7d8e9a6f672aff78dc682cc/rclcpp/include/rclcpp/detail/cpp_callback_trampoline.hpp#L60 is incorrect.

mauropasse:
Thanks @fujitatomoya for the further investigation. Cool to see the function types from GDB.
@iuhilnehc-ynos I've reverted the first commit and applied your suggestion, so we still support what the documentation says. Thanks for that!

fujitatomoya (approved):
@iuhilnehc-ynos 's suggestion makes more sense to me, lgtm.

rclcpp/include/rclcpp/detail/cpp_callback_trampoline.hpp
```diff
@@ -50,14 +51,15 @@ namespace detail
...
+  typename UserDataRealT,
   typename UserDataT,
   typename ... Args,
   typename ReturnT = void
 >
 ReturnT
 cpp_callback_trampoline(UserDataT user_data, Args ... args) noexcept
 {
-  auto & actual_callback = *reinterpret_cast<const std::function<ReturnT(Args...)> *>(user_data);
+  auto & actual_callback = *static_cast<const UserDataRealT *>(user_data);
```
alsora, line 62: Rather than adding a new template argument (`UserDaraRealT`) could we use
```
static_cast<const decltype(UserDataT) *>
```
?

ivanpauno (approved):
lgtm!

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17817/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12384/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18482/)

Issue #1973, opened 2022-07-13
rclcpp fails to build with MSVC with the /permissive- flag on


AustinMooreResonantSciences:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - binaries
- Version or commit hash:
  - humble
- DDS implementation:
  - Fast-RTP
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Create a minimal ros2 project

with the following cpp file
```C++
#include "rclcpp/rclcpp.hpp"

int main() {
  return 0;
}
```
the follwing CMakeLists.txt
```
cmake_minimum_required(VERSION 3.18)
project(test1)

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

find_package(ament_cmake REQUIRED)
find_package(rclcpp REQUIRED)

add_executable(test1 test.cpp)

target_compile_options(test1 PUBLIC /permissive- /twoPhase-)
ament_target_dependencies(test1 rclcpp)

ament_package()
```
the following package.xml
```
<?xml version="1.0"?>
<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
<package format="3">
  <name>test1</name>
  <version>0.0.0</version>
  <description>NA</description>
  <maintainer email="austin.moore@resonantsciences.com">austin.moore</maintainer>
  <license>Proprietary</license>

  <buildtool_depend>ament_cmake</buildtool_depend>
  <depend>rclcpp</depend>
  <depend>std_msgs</depend>
  
  <buildtool_depend>rosidl_default_generators</buildtool_depend>
  <exec_depend>rosidl_default_runtime</exec_depend>
  <member_of_group>rosidl_interface_packages</member_of_group>
  
  <test_depend>ament_lint_auto</test_depend>
  <test_depend>ament_lint_common</test_depend>

  <export>
    <build_type>ament_cmake</build_type>
  </export>
</package>
```
#### Expected behavior
Configure the cmake project and build correctly with MSVC C++ conformance flag on "/permissive-"
#### Actual behavior
Fails to build with the following error due to MSVC C++ conformance flag.
```
C:\ros2\ros2-windows\include\rclcpp\rclcpp/service.hpp(320): error C2039: 'type': is not a member of 'std'
```
#### Additional information
[...]

PR #2063, opened 2022-12-12
Fix logging macros to build with msvc and cpp20

msz-rai:
When I was building my project (CMAKE_CXX_STANDARD 20) with rclcpp on Windows, it failed with the same error as described in https://github.com/ros2/rclcpp/issues/1973 (c++17 + [permissive flag](https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170) added to MSVC). The same flag is set implicitly with C++20 because is required for C++20 Modules support.

I was trying to investigate why the template in logging macro is not passing standards conformance, but I was not able to find it out. It seems to be an MSVC issue. A similar problem was reported in [Microsoft Developer Community](https://developercommunity.visualstudio.com/t/the-c2039-type-is-not-a-member-of-std-error-in-def/1562967) more than a year ago and has not been resolved yet.

I used helper type aliases from `type_traits` instead of member types. This change makes rclcpp build successfully with MSVC and C++20.

### DIFF
--- a/rclcpp/resource/logging.hpp.em
+++ b/rclcpp/resource/logging.hpp.em
@@ -124,7 +124,7 @@ def get_rclcpp_suffix_from_features(features):
 ) \
   do { \
     static_assert( \
-      ::std::is_same<typename std::remove_cv<typename std::remove_reference<decltype(logger)>::type>::type, \
+      ::std::is_same<typename std::remove_cv_t<typename std::remove_reference_t<decltype(logger)>>, \
       typename ::rclcpp::Logger>::value, \
       "First argument to logging macros must be an rclcpp::Logger"); \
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):
Though the change shouldn't be needed, it's okay to do it.
LGTM!

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17815/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12381/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18474/)

fujitatomoya:
@Mergifyio backport humble

PR #2066, opened 2022-12-15
Move event callback binding to PublisherBase and SubscriptionBase

methylDragon:
This de-duplicates some code

### DIFF
--- a/rclcpp/include/rclcpp/generic_publisher.hpp
+++ b/rclcpp/include/rclcpp/generic_publisher.hpp
@@ -78,38 +78,12 @@ public:
       node_base,
       topic_name,
       *rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib),
-      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos)),
+      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos),
+      // NOTE(methylDragon): Passing these args separately is necessary for event binding
+      options.event_callbacks,
+      options.use_default_callbacks),
     ts_lib_(ts_lib)
-  {
-    // This is unfortunately duplicated with the code in publisher.hpp.
-    // TODO(nnmm): Deduplicate by moving this into PublisherBase.
-    if (options.event_callbacks.deadline_callback) {
-      this->add_event_handler(
-        options.event_callbacks.deadline_callback,
-        RCL_PUBLISHER_OFFERED_DEADLINE_MISSED);
-    }
-    if (options.event_callbacks.liveliness_callback) {
-      this->add_event_handler(
-        options.event_callbacks.liveliness_callback,
-        RCL_PUBLISHER_LIVELINESS_LOST);
-    }
-    if (options.event_callbacks.incompatible_qos_callback) {
-      this->add_event_handler(
-        options.event_callbacks.incompatible_qos_callback,
-        RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
-    } else if (options.use_default_callbacks) {
-      // Register default callback when not specified
-      try {
-        this->add_event_handler(
-          [this](QOSOfferedIncompatibleQoSInfo & info) {
-            this->default_incompatible_qos_callback(info);
-          },
-          RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
-      } catch (UnsupportedEventTypeException & /*exc*/) {
-        // pass
-      }
-    }
-  }
+  {}
 
   RCLCPP_PUBLIC
   virtual ~GenericPublisher() = default;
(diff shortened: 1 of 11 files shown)


methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17832/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12400/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18499/)

methylDragon:
Rpr is happy, so I'm happy :grimacing:

rclcpp/include/rclcpp/generic_publisher.hpp
```diff
@@ -78,38 +78,12 @@ class GenericPublisher : public rclcpp::PublisherBase
       node_base,
       topic_name,
       *rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib),
-      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos)),
+      options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos),
+      // NOTE(methylDragon): Passing these args separately is necessary for event binding
+      options.event_callbacks,
+      options.use_default_callbacks),
```
fujitatomoya, line 84: directly off topic from this PR. but this `use_default_callbacks` flag should be declared in `PublisherEventCallbacks` and `SubscriptionEventCallbacks`? if so, we can just pass `options.event_callbacks` here?

as `PublisherBase` argument, this seems to be more general flag to use default callback for anything.

fujitatomoya (commented):
for code maintenance this duplication should be avoided, lgtm with minor comment.

fujitatomoya:
@ivanpauno @hidmic either of you can review on this?

rclcpp/include/rclcpp/generic_publisher.hpp
methylDragon: I think this one is kinda tough, because `use_default_callbacks` seems better fit to be part of the options struct than the event_callbacks struct, since it's a configuration for the pub/sub.

I'd have passed the entire options struct in, but that requires templating the base class, which is why I had to go with the compromise of just passing the event callbacks and the bool flag separately...

rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -569,6 +578,8 @@ class SubscriptionBase : public std::enable_shared_from_this<SubscriptionBase>
   uint64_t intra_process_subscription_id_;
   std::shared_ptr<rclcpp::experimental::SubscriptionIntraProcessBase> subscription_intra_process_;
 
+  const SubscriptionEventCallbacks event_callbacks_;
```
alsora, line 581: This is "duplicating" the callbacks object. 
The base class stores them in this variable, while the derived classes store the `options_` object (which also contains them).

methylDragon: This duplication is deliberate, the reason being that since the base class is now storing the event callbacks, there is no guarantee that derived classes will store their options object.

There are also cases where there are no derived classes, in which case the lifetime of the event_callbacks object is not guaranteed to be the same as the pub/sub bases. For those reasons I decided to duplicate them (it shouldn't be too significant either way, since it should be just function pointers).

alsora: Yes, the memory footprint is going to be very small.
However looking at the code I think that maybe the derived class shouldn't have an `options_` member at all.

But this is probably something that can be addressed in a separate PR

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -98,6 +101,8 @@ PublisherBase::PublisherBase(
     rmw_reset_error();
     throw std::runtime_error(msg);
   }
+
+  bind_event_callbacks(event_callbacks_, use_default_callbacks);
```
ivanpauno, line 105: It's a bad practice to call virtual methods for the constructor, as it will call the method defined here (see https://en.cppreference.com/w/cpp/language/virtual#During_construction_and_destruction).
Anyway, there's no need for the method to be `virtual` really, so I would just remove the qualifier.

ivanpauno (changes requested):

rclcpp/include/rclcpp/generic_publisher.hpp
fujitatomoya: > because use_default_callbacks seems better fit to be part of the options struct than the event_callbacks struct, since it's a configuration for the pub/sub.

i think this is an option for `event_callbacks`, not entire publisher or subscription. that is why i came up with this comment. it would be better if we name this option into `use_default_event_callbacks`?

rclcpp/src/rclcpp/publisher_base.cpp
fujitatomoya: 👍 this will call base class method every time, same thing goes to dtor.

ivanpauno (approved):
LGTM!

methylDragon:
One more run of CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17840/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12408/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18508/)

PR #2069, opened 2022-12-20
Hook up the inconsistent topic event inside of rclcpp

clalancette:
This PR is part of https://github.com/ros2/ros2/issues/1361 .  In particular, what this does is take the inconsistent topic event as it comes from the rmw/rcl layer, and allows the user to register for a callback to find that out.

Along the way, some refactoring was in order.  In particular, rclcpp assumed that all events were QoS events, which is not really the case.  This PR does some renaming of things to make it clear that these events are not just for QoS, but can be for other things as well.  With that done, the last patch in the series actually implements the inconsistent topic callback.

Part of https://github.com/ros2/ros2/issues/1361

Depends on https://github.com/ros2/rmw/pull/339
Depends on https://github.com/ros2/rcl/pull/1024


rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -145,21 +145,39 @@ PublisherBase::bind_event_callbacks(
...
+    incompatible_qos_cb = [this](QOSOfferedIncompatibleQoSInfo & info) {
+        this->default_incompatible_qos_callback(info);
+      };
+  }
+  try {
+    if (incompatible_qos_cb) {
+      this->add_event_handler(incompatible_qos_cb, RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
+    }
+  } catch (const UnsupportedEventTypeException & /*exc*/) {
+    // pass
```
ivanpauno, line 163: a debug message wouldn't be a bad idea

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -145,21 +145,39 @@ PublisherBase::bind_event_callbacks(
...
+    inconsistent_topic_cb = [this](InconsistentTopicInfo & info) {
+        this->default_inconsistent_topic_callback(info);
+      };
+  }
+  try {
+    if (inconsistent_topic_cb) {
+      this->add_event_handler(inconsistent_topic_cb, RCL_PUBLISHER_INCONSISTENT_TOPIC);
     }
+  } catch (UnsupportedEventTypeException & /*exc*/) {
+    // pass
```
ivanpauno, line 180: same

rclcpp/include/rclcpp/event_handler.hpp
```diff
@@ -0,0 +1,304 @@
...
+#include "rclcpp/waitable.hpp"
+
+namespace rclcpp
+{
+
+using QOSDeadlineRequestedInfo = rmw_requested_deadline_missed_status_t;
+using QOSDeadlineOfferedInfo = rmw_offered_deadline_missed_status_t;
+using QOSLivelinessChangedInfo = rmw_liveliness_changed_status_t;
+using QOSLivelinessLostInfo = rmw_liveliness_lost_status_t;
+using QOSMessageLostInfo = rmw_message_lost_status_t;
```
ivanpauno, line 45: it may be worth to rename this one as `MessageLostInfo`, as it's not qos related as well

ivanpauno (approved):
LGTM!

I have some minor comments

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -311,6 +329,17 @@ PublisherBase::default_incompatible_qos_callback(
...
 
+void
+PublisherBase::default_incompatible_type_callback(
+  rclcpp::IncompatibleTypeInfo & event) const
+{
+  (void)event;
+
+  RCLCPP_WARN(
+    rclcpp::get_logger(rcl_node_get_logger_name(rcl_node_handle_.get())),
+    "Incompatible type on topic '%s', no messages will be sent to it.", get_topic_name());
```
fujitatomoya, line 340: I want to know the behavior of this event.

If any incompatible type is detected in the topic, all endpoints can receive this event and stop sending / receiving messages? from this warning i would expect that behavior in DDS.

or only incompatible type endpoint will receive this event? and then the question is how we can decide on who is incompatible in the topic? Is that `1st comes 1st served` policy?

Some example conditions,

- publisher(type A) / subscription(type A) already communicating. then publisher/subscription(type B) joins and type B endpoints can only receive this event? or publisher(type A) / subscription(type A) also receives the event? and publisher(type A) / subscription(type A) also stops communicating?
- publisher(type A) exists on the topic. then subscription(type B) joins. subscription(type B) will receive the event? and what if after that publisher(type B) joins?

fujitatomoya (commented):
Implementation looks good to me, but i want to know the behavior of this event mechanism in actual use cases.

rclcpp/src/rclcpp/publisher_base.cpp
clalancette: Good call, fixed in d21e1cae5af762ee044893fc48fe32061d23ad49.

rclcpp/src/rclcpp/publisher_base.cpp
clalancette: Fixed in d21e1cae5af762ee044893fc48fe32061d23ad49

rclcpp/src/rclcpp/publisher_base.cpp
fujitatomoya: as event notification schema, i believe it is really clear from https://github.com/ros2/rmw/pull/339#discussion_r1085538475.

but (just checking) i think message communication will keep sending and receiving message if rmw implementation can do so only with compatible topic types? i think this is important behavior to know for user application.

clalancette: Oh, I see.  These are slightly different questions, sorry I didn't read them closer.

OK, let me go through them:

> publisher(type A) / subscription(type A) already communicating. then publisher/subscription(type B) joins and type B endpoints can only receive this event? or publisher(type A) / subscription(type A) also receives the event? and publisher(type A) / subscription(type A) also stops communicating?

In this case, if a publisher(type B) joins, then publisher(type B) and subscription(type A) both get an event.  publisher(type A) does *not* get an event.  If a subscription(type B) joins, then subscription(type B) and publisher(type A) both get an event.  In all cases, publisher(type A) and subscription(type A) keep communicating.

> * publisher(type A) exists on the topic. then subscription(type B) joins. subscription(type B) will receive the event? and what if after that publisher(type B) joins?

In this case, when subscription(type B) joins, both publisher(type A) and subscription(type B) get the event, and no communication happens.  When publisher(type B) joins, then subscription(type B) starts receiving data from publisher(type B) only.

Does that make sense?

fujitatomoya: it does. even inconsistent topic event comes up, endpoints can communicate each other with compatible type. I just wanted to check this just in case, thank you so much for the clarification.

fujitatomoya (approved):

PR #2077, opened 2023-01-03
Runtime Interface Reflection: rclcpp

methylDragon:
This PR is part of the runtime interface reflection subscription feature of REP-2011: https://github.com/ros2/ros2/issues/1374

# Description
This PR includes support for runtime type subscriptions, including:
- Edits to SubscriptionBase and executor code to allow for use of runtime type
  - This includes new pure virtual methods for SubscriptionBase that have also been implemented in all child classes
- A new RuntimeTypeSubscription subscription class that handles a subscription callback using ser_dynamic_data


rclcpp/src/rclcpp/dynamic_typesupport/dynamic_message_type_support.cpp
```diff
@@ -0,0 +1,465 @@
...
+  if (serialization_library_name.empty()) {
+    ret = rcl_dynamic_message_type_support_handle_create(nullptr, &description, &ts);
+  } else {
+    ret = rcl_dynamic_message_type_support_handle_create(
+      serialization_library_name.c_str(), &description, &ts);
+  }
+  if (ret != RCL_RET_OK) {
+    throw std::runtime_error("error initializing rosidl message type support");
+  }
+  if (!ts) {
```
emersonknapp, line 70: Can all these bad cases be true if `ret == RCL_RET_OK`?

rclcpp/include/rclcpp/dynamic_typesupport/dynamic_message_type_support.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+  std::shared_ptr<rosidl_message_type_support_t> rosidl_message_type_support_;
+
+private:
+  RCLCPP_PUBLIC
+  DynamicMessageTypeSupport();
+
+  RCLCPP_PUBLIC
+  void
+  manage_description_(rosidl_runtime_c__type_description__TypeDescription * description);
```
emersonknapp, line 179: `manage` is an overloaded term - this is really a `make_shared` function to put ownership of the object into a shared pointer, right?

rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -60,6 +63,14 @@ namespace experimental
 class IntraProcessManager;
 }  // namespace experimental
 
+enum class SubscriptionType : uint8_t
+{
+  ROS_MESSAGE = 1,  // take message as ROS message and handle as ROS message
+  SERIALIZED_MESSAGE = 2,  // take message as serialized and handle as serialized
+  DYNAMIC_MESSAGE_DIRECT = 3,  // take message as DynamicMessage and handle as DynamicMessage
+  DYNAMIC_MESSAGE_FROM_SERIALIZED = 4  // take message as serialized and handle as DynamicMessage
```
emersonknapp, line 71: The distinction here isn't  clear to me between dynamic message direct and serialized.

What does this wording pattern "take as X and handle as Y" mean? Does "handle" refer to the argument type for the callback?

rclcpp/include/rclcpp/dynamic_typesupport/dynamic_message_type_support.hpp
methylDragon: Yeah that's true.. I wanted it to be like a: "We are taking ownership/managing the lifetime of the pointer now)

It's a private method though, so I don't have any issues with changing the name

rclcpp/include/rclcpp/subscription_base.hpp
methylDragon: Yep, "handle" is for the callback.

The way it works with ROS is that a message has to be taken, then passed to a handle function that passes it to a callback. So take -> handle.

rclcpp/src/rclcpp/dynamic_typesupport/dynamic_message_type_support.cpp
methylDragon: :thinking: probably not.

I might still leave in the typesupport identifier check though

rclcpp/include/rclcpp/subscription_base.hpp
emersonknapp: What about a dynamic subscription that passes serialized data to the handler function? `GenericDynamicSubscription`, perhaps. It's what we would want in rosbag2 recording.

methylDragon: In that case you would want to use a serialized subscription (I.e. one that takes and handles serialized data), then use the DynamicMessage class to deserialize to into the DynamicMessage

You probably would need to subclass that subscription so you can instantiate a DynamicMessage and have it accessible inside the callback

methylDragon:
```
    // Take serialized and then convert to dynamic message
    case rclcpp::SubscriptionType::DYNAMIC_MESSAGE_FROM_SERIALIZED:
      {
        std::shared_ptr<SerializedMessage> serialized_msg =
          subscription->create_serialized_message();

        // NOTE(methylDragon): Is this clone necessary? If I'm following the pattern, it seems so.
        DynamicMessage::SharedPtr dynamic_message = subscription->create_dynamic_message();
        take_and_do_error_handling(
          "taking a serialized message from topic",
          subscription->get_topic_name(),
          [&]() {return subscription->take_serialized(*serialized_msg.get(), message_info);},
          [&]()
          {
            // === <<< THIS IS THE MOST IMPORTANT LINE!!! >>> ===
            bool ret = dynamic_message->deserialize(&serialized_msg->get_rcl_serialized_message());
            if (!ret) {
              throw_from_rcl_error(ret, "Couldn't convert serialized message to dynamic data!");
            }
            subscription->handle_dynamic_message(dynamic_message, message_info);
          }
        );
        subscription->return_serialized_message(serialized_msg);
        subscription->return_dynamic_message(dynamic_message);
        break;
      }
  }
```

methylDragon:
Closing this in favor of https://github.com/ros2/rclcpp/pull/2176

PR #2078, opened 2023-01-06
Avoid priority inversions in rclcpp

WideAwakeTN:
The content of this pull request ensures that rclcpp will not have priority inversions when using hard realtime scheduling, i.e. FIFO scheduling as defined in the POSIX standard. This is achieved by replacing std::mutex with rclcpp::PIMutex and std::recursive_mutex with rclcpp::RecursivePIMutex. To compile this pull request the following pull requests for rcutils and rcpputils are required:
https://github.com/ros2/rcpputils/pull/174
https://github.com/ros2/rcutils/pull/406

The code changes in this pull request should not effect the runtime behavior of existing ROS2 projects in a negative way. In case that a project uses ordinary fair/non-realtime thread scheduling it should behave as before.

In case this pull request gets accepted other ROS2 components can be made priority inversion safe in the manner which was demonstrated here. This will make ROS2 more suitable for hard realtime scenarios.

### DIFF
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -16,9 +16,7 @@
 #define RCLCPP__CALLBACK_GROUP_HPP_
 
 #include <atomic>
-#include <functional>
-#include <memory>
-#include <mutex>
+#include <string>
 #include <vector>
 
 #include "rclcpp/client.hpp"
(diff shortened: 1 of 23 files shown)


ros-discourse:

https://discourse.ros.org/t/avoiding-priority-inversions-includes-draft-pull-requests/29219/1

WideAwakeTN:
A ROS Discourse article has been created for this pull request:
https://discourse.ros.org/t/avoiding-priority-inversions-includes-draft-pull-requests/29219

PR #2079, opened 2023-01-09
Irobot/add dynamic network notify function

jakymiws:


fujitatomoya:
@jakymiws this looks really interesting, can you add more description or design how we want to implement this feature and expose to user client libraries? i think this could be really useful for edge devices and robots that move to different network and so on.

PR #2085, opened 2023-01-19
Improve component_manager_isolated shutdown

### LINKED ISSUE #2083: it takes too long to terminate when using component_container_isolated
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries (humble)
- Version or commit hash:
  - 
- DDS implementation:
  - cyclonedds
  - rclcpp

#### Steps to reproduce issue
I set `use_composition` to "True" to run the navigation stack nodes in one process and it worked.

after that, I typed `ctrl+c` to exit the process. but it takes too long to terminate(more than 10 seconds), and it does not terminate properly sometimes.

Below is part of the launch file I ran.

```
......

nav2_container = Node(
        condition=IfCondition(use_composition),
        package='rclcpp_components',
        executable='component_container_isolated',
        name=container_name,
        output='screen',
        respawn=use_respawn,
        respawn_delay=2.0,
        parameters=[configured_params_for_compositon, {'autostart': autostart}],
        remappings=remappings),
            
......

load_composable_nodes = LoadComposableNodes(
        condition=IfCondition(use_composition),
        target_container=container_name, #nav2_container
        composable_node_descriptions=[
            ComposableNode(
                package='nav2_controller',
                plugin='nav2_controller::ControllerServer',
                name='controller_server',
                parameters=[configured_params],
                remappings=[('cmd_vel', 'cmd_vel_nav')] + remappings),

            ComposableNode(
                package='nav2_smoother',
                plugin='nav2_smoother::SmootherServer',
                name='smoother_server',
                parameters=[configured_params],
                remappings=remappings),

            ComposableNode(
                package='nav2_planner',
                plugin='nav2_planner::PlannerServer',
                name='planner_server',
                parameters=[configured_params],
                remappings=remappings),

            ComposableNode(
                package='nav2_behaviors',
                plugin='behavior_server::BehaviorServer',
                name='behavior_server',
                parameters=[configured_params],
                remappings=remappings),

            ComposableNode(
                package='nav2_waypoint_follower',
                plugin='nav2_waypoint_follower::WaypointFollower',
                name='waypoint_follower',
                parameters=[configured_params],
                remappings=remappings),

            ComposableNode(
                package='nav2_bt_navigator',
                plugin='nav2_bt_navigator::BtNavigator',
                name='bt_navigator',
                parameters=[configured_params],
                remappings=remappings),

            ComposableNode(
                package='nav2_velocity_smoother',
                plugin='nav2_velocity_smoother::VelocitySmoother',
                name='velocity_smoother',
                parameters=[configured_params],
                remappings=[('cmd_vel', 'cmd_vel_nav'),
                            ('cmd_vel_smoothed', 'cmd_vel')] + remappings),

            ComposableNode(
                package='nav2_lifecycle_manager',
                plugin='nav2_lifecycle_manager::LifecycleManager',
                name='lifecycle_manager_navigation',
                parameters=[{'use_sim_time': use_sim_time,
                            'autostart': autostart,
                            'node_names': lifecycle_nodes}],
                remappings=remappings),
        ],
    )
    
.......
```

Below is the log when component container ends. (actually I use two container node, one is for navigation stack, and the other is for my custom node)

```
[ERROR] [component_container_isolated-13]: process[component_container_isolated-13] failed to terminate '5' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
[ERROR] [component_container_isolated-7]: process[component_container_isolated-7] failed to terminate '5' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
[INFO] [component_container_isolated-13]: sending signal 'SIGTERM' to process[component_container_isolated-13]
[INFO] [component_container_isolated-7]: sending signal 'SIGTERM' to process[component_container_isolated-7]
[component_container_isolated-13] [INFO 1673835779.167809326] [rclcpp]: signal_handler(signum=15) (./src/rclcpp/signal_handler.cpp:71)
[component_container_isolated-7] [INFO 1673835779.169543075] [rclcpp]: signal_handler(signum=15) (./src/rclcpp/signal_handler.cpp:71)
[ERROR] [component_container_isolated-13]: process[component_container_isolated-13] failed to terminate '10.0' seconds after receiving 'SIGTERM', escalating to 'SIGKILL'
[ERROR] [component_container_isolated-7]: process[component_container_isolated-7] failed to terminate '10.0' seconds after receiving 'SIGTERM', escalating to 'SIGKILL'
[INFO] [component_container_isolated-13]: sending signal 'SIGKILL' to process[component_container_isolated-13]
[INFO] [component_container_isolated-7]: sending signal 'SIGKILL' to process[component_container_isolated-7]
[ERROR] [component_container_isolated-7]: process has died [pid 22074, exit code -9, cmd '/opt/ros/humble/lib/rclcpp_components/component_container_isolated --ros-args -r __node:=perception_container -r __ns:=/'].
[ERROR] [component_container_isolated-13]: process has died [pid 22086, exit code -9, cmd '/opt/ros/humble/lib/rclcpp_components/component_container_isolated --ros-args -r __node:=nav_container -r __ns:=/ --params-file /tmp/tmpf4tgn9jt --params-file /tmp/launch_params_ei4e54qx -r /tf:=tf -r /tf_static:=tf_static'].

```

On the other hand, when using `component_container` instead of `component_container_isolated`, it is almost finished immediately.

Is it normal to take a long time to end when using `component_container_isolated`?

mjcarroll:
This eliminates a potential hang when the isolated container is being shutdown via the rclcpp::SignalHandler.

Previously, when the SignalHandler triggers the shutdown of the rest of the system, the ComponentManagerIsolated object would cancel and destroy the executor associated with each loaded component.

Part of the cancel process checks if the executor is currently spinning as there may have been a race condition where `cancel_executor` is called before the thread is created.  While I didn't personally observe this condition, the comment indicates that it may happen.

```
    while (!executor_wrapper.executor->is_spinning()) {
      // This is an arbitrarily small delay to avoid busy looping
      rclcpp::sleep_for(std::chrono::milliseconds(1));
    }
```

The issue is that there is a race where the executor is already shutdown/cancelled (so is_spinning is always false), causing an infinite loop.

This code introduces a guard atomic to check that the loop has been started before continuing to check the `is_spinning` as well as checking the overall ROS context for shutdown.

Closes #2083 

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
@@ -38,6 +38,16 @@ class ComponentManagerIsolated : public rclcpp_components::ComponentManager
   {
     std::shared_ptr<rclcpp::Executor> executor;
     std::thread thread;
+    std::atomic_bool thread_initialized;
+
+    /// Constructor for the wrapper.
+    /// This is necessary as atomic variables don't have copy/move operators
+    /// implemented so this structure is not copyable/movable by default
+    explicit DedicatedExecutorWrapper(std::shared_ptr<rclcpp::Executor> exec):
+      executor(exec),
+      thread_initialized(false)
+    {
+    }
   };
 
 public:
@@ -59,15 +69,19 @@ protected:
   void
   add_node_to_executor(uint64_t node_id) override
   {
-    DedicatedExecutorWrapper executor_wrapper;
     auto exec = std::make_shared<ExecutorT>();
     exec->add_node(node_wrappers_[node_id].get_node_base_interface());
-    executor_wrapper.executor = exec;
-    executor_wrapper.thread = std::thread(
-      [exec]() {
+
+    // Emplace rather than std::move since move operations are deleted for atomics
+    auto result = dedicated_executor_wrappers_.emplace(std::make_pair(node_id, exec));
+    DedicatedExecutorWrapper& wrapper = result.first->second;
+    wrapper.executor = exec;
+    auto& thread_initialized = wrapper.thread_initialized;
+    wrapper.thread = std::thread(
+      [exec, &thread_initialized]() {
+        thread_initialized = true;
         exec->spin();
       });
-    dedicated_executor_wrappers_[node_id] = std::move(executor_wrapper);
   }
   /// Remove component node from executor model, it's invoked in on_unload_node()
   /**
@@ -90,15 +104,22 @@ private:
    */
   void cancel_executor(DedicatedExecutorWrapper & executor_wrapper)
   {
-    // We can't immediately call the cancel() API on an executor because if it is not
-    // already spinning, this operation will have no effect.
-    // We rely on the assumption that this class creates executors and then immediately makes them
-    // spin in a separate thread, i.e. the time gap between when the executor is created and when
-    // it starts to spin is small (although it's not negligible).
+    // Verify that the executor thread has begun spinning.
+    // If it has not, then iterate until the thread starts to ensure
+    // that cancel will fully stop the execution
+    //
+    // This prevents a previous race condition that occur between the
+    // creation of the executor spin thread and cancelling an executor
 
-    while (!executor_wrapper.executor->is_spinning()) {
-      // This is an arbitrarily small delay to avoid busy looping
-      rclcpp::sleep_for(std::chrono::milliseconds(1));
+    if (!executor_wrapper.thread_initialized)
+    {
+      auto context = this->get_node_base_interface()->get_context();
+
+      // Guarantee that either the executor is spinning or we are shutting down.
+      while (!executor_wrapper.executor->is_spinning() && rclcpp::ok(context)) {
+        // This is an arbitrarily small delay to avoid busy looping
+        rclcpp::sleep_for(std::chrono::milliseconds(1));
+      }
     }
 
     // After the while loop we are sure that the executor is now spinning, so
--- a/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
+++ b/rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
@@ -43,8 +43,8 @@ class ComponentManagerIsolated : public rclcpp_components::ComponentManager
     /// Constructor for the wrapper.
     /// This is necessary as atomic variables don't have copy/move operators
     /// implemented so this structure is not copyable/movable by default
-    explicit DedicatedExecutorWrapper(std::shared_ptr<rclcpp::Executor> exec):
-      executor(exec),
+    explicit DedicatedExecutorWrapper(std::shared_ptr<rclcpp::Executor> exec)
+    : executor(exec),
       thread_initialized(false)
     {
     }
@@ -74,9 +74,9 @@ protected:
 
     // Emplace rather than std::move since move operations are deleted for atomics
     auto result = dedicated_executor_wrappers_.emplace(std::make_pair(node_id, exec));
-    DedicatedExecutorWrapper& wrapper = result.first->second;
+    DedicatedExecutorWrapper & wrapper = result.first->second;
     wrapper.executor = exec;
-    auto& thread_initialized = wrapper.thread_initialized;
+    auto & thread_initialized = wrapper.thread_initialized;
     wrapper.thread = std::thread(
       [exec, &thread_initialized]() {
         thread_initialized = true;
@@ -111,8 +111,7 @@ private:
     // This prevents a previous race condition that occur between the
     // creation of the executor spin thread and cancelling an executor
 
-    if (!executor_wrapper.thread_initialized)
-    {
+    if (!executor_wrapper.thread_initialized) {
       auto context = this->get_node_base_interface()->get_context();
 
       // Guarantee that either the executor is spinning or we are shutting down.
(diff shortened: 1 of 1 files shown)


mjcarroll:
@clalancette Now that I have this implemented, one open question is that if the `is_spinning` logic is still required with this atomic flag?

alsora (approved):

alsora:
@mjcarroll I think that the `is_spinning` check is still required and I'm actually not sure if you really need the atomic variable or the check on the context is sufficient.

Without the `is_spinning` there may be a data race where the thread starts (atomic becomes true) but then it's pre-empted before starting to spin.

mjcarroll:
> I think that the is_spinning check is still required and I'm actually not sure if you really need the atomic variable or the check on the context is sufficient.

Chris and I looked at this together, and we thought that the safest thing was to include both.  The context check by itself was enough to prevent it from hanging on my local computer, but that does not guarantee that it is completely safe.

The issue stems from a little bit of under-documentation in the way that the executor's cancel works.  The question is, should an executor be allowed to `spin` after `cancel` is called, and what is the state when `cancel` is called before the first `spin`.  In this case, this guard exists to prevent the `cancel` from being called before the first spin.  

A potential alternative would be to `spin_once`, set the `atomic`, and then `spin`, to guarantee that the `spin` has happened before the `cancel`.

The original fix here was to delay until the executor is spinning, but we ended up in a state where the executor is not spinning so the loop cannot be broken.

fujitatomoya (approved):
lgtm

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17971/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12508/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18613/)

rclcpp_components/include/rclcpp_components/component_manager_isolated.hpp
```diff
@@ -90,15 +104,21 @@ class ComponentManagerIsolated : public rclcpp_components::ComponentManager
...
-
-    while (!executor_wrapper.executor->is_spinning()) {
-      // This is an arbitrarily small delay to avoid busy looping
-      rclcpp::sleep_for(std::chrono::milliseconds(1));
+    // Verify that the executor thread has begun spinning.
+    // If it has not, then iterate until the thread starts to ensure
+    // that cancel will fully stop the execution
+    //
+    // This prevents a previous race condition that occur between the
+    // creation of the executor spin thread and cancelling an executor
```
clalancette, line 112:
```suggestion
    // If it has not, then wait until the thread starts to ensure
    // that cancel() will fully stop the execution
    //
    // This prevents a previous race condition that occurs between the
    // creation of the executor spin thread and cancelling an executor
```

clalancette (approved):
I've only left suggestions to clean up comments a bit.  Otherwise, this looks good to me.

fujitatomoya:
CI(Linux only):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17978/)

tonynajjar:
I spent 2 hours debugging this on humble before finally finding this fix. Thanks for the fix! Can you please backport to humble? I cherrypicked it locally and it worked well

fujitatomoya:
@mjcarroll @clalancette 

adding explicit default constructor should be fine to keep the ABI? i think that is okay but need to be sure...

clalancette:
> adding explicit default constructor should be fine to keep the ABI? i think that is okay but need to be sure...

I think that is OK, but the addition of the `atomic_bool` to the inner struct definitely breaks ABI.  Someone will have to find another way to backport this.

CHAIJZ888:
Hi @mjcarroll @clalancette,

I'm currently using ROS 2 Humble with component_container_isolated, and I'm still experiencing the shutdown hang originally reported in #2083.

The problem is consistently reproducible with Ctrl+C (SIGINT). The container does not exit cleanly and remains alive until the launch system escalates the signal. For example:

Plaintext
[ERROR] [component_container_isolated-X]: process failed to terminate '5.0' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
[ERROR] [component_container_isolated-X]: process failed to terminate '10.0' seconds after receiving 'SIGTERM', escalating to 'SIGKILL'
[ERROR] [component_container_isolated-X]: process has died ... exit code -9
I found #2085 and tested the patch locally against our Humble build. 

I understand the ABI concerns discussed previously: the addition of std::atomic_bool to the internal struct means the original PR #2085 cannot simply be backported to Humble.

Given that this issue still significantly impacts Humble users relying on component_container_isolated, could we consider exploring an ABI-compatible alternative or workaround that preserves the Humble ABI while fixing this executor shutdown race?

Thanks for your time and all the great work!

jmachowinski:
Feel free to open an PR.

CHAIJZ888:
> Feel free to open an PR.

Thanks @jmachowinski.

I opened #3282 with an ABI-compatible fix for Humble:

https://github.com/ros2/rclcpp/pull/3282

The change prevents `ComponentManagerIsolated` from waiting indefinitely
during context shutdown. It only modifies the `cancel_executor()` function
body and does not add data members or change class layout.

The PR also includes a regression test that reproduces the shutdown hang
against the original Humble implementation and passes with the proposed fix.

The PR is ready for review. I would appreciate any feedback on the
implementation and regression-test approach.

PR #2086, opened 2023-01-19
Add function to call through to dds layer to notify about new network interfaces

jakymiws:
Directly related to `irobot/add-dynamic-network-notify-function` in `jakymiws/rcl`


jakymiws:
Related to:
https://github.com/ros2/rcl/pull/1028
https://github.com/ros2/rmw_fastrtps/pull/662
https://github.com/ros2/rmw_implementation/pull/213
https://github.com/ros2/rmw/pull/344

rclcpp/include/rclcpp/utilities.hpp
```diff
@@ -330,6 +330,14 @@ RCLCPP_PUBLIC
 std::vector<const char *>
 get_c_vector_string(const std::vector<std::string> & strings_in);
 
+/// Call dds to notify about new network interfaces.
```
alsora, line 333: This layer shouldn't assume that the underlying middleware is a DDS.
Maybe the comment should say something like "Notify middleware to refresh available network interfaces?"

jakymiws: @alsora good point - will change that.

alsora:
@jakymiws we should also add a function implementation in the other TIER 1 RMW middelwares.
In this case we can return a NOT IMPLEMENTED error if the implementation does not support it yet

jakymiws:
> 

Done with cyclone-dds here: https://github.com/ros2/rmw_cyclonedds/pull/433 and connext-dds here: https://github.com/ros2/rmw_connextdds/pull/99

PR #2087, opened 2023-01-20
Allow != and == of Time objects with different sources

sloretz:
`rclcpp::Time` forbids comparisons between objects with different clock types. It raises an exception when that happens. This makes sense for `<`, `<=`, etc, but I think `==` and `!=` could be defined by saying `Time` objects with different clock types are never equal to each other.

I think this might simplify some of the logic for handling `Time` with different time sources here: https://github.com/locusrobotics/fuse/pull/307

@ros2/team thoughts?

### DIFF
--- a/rclcpp/include/rclcpp/time.hpp
+++ b/rclcpp/include/rclcpp/time.hpp
@@ -100,9 +100,6 @@ public:
   Time &
   operator=(const builtin_interfaces::msg::Time & time_msg);
 
-  /**
-   * \throws std::runtime_error if the time sources are different
-   */
   RCLCPP_PUBLIC
   bool
   operator==(const rclcpp::Time & rhs) const;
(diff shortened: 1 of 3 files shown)


wjwwood:
I agree with @alsora. However, I would be ok with a different comparison function (like a named one, e.g. `.clock_and_time_equal()` or something) being added as a method on Time to reduce duplication across code bases.

fujitatomoya (commented):
I am with @alsora , probably throwing exception is better for user application.

sloretz:
All feedback so far has pointed towards not allowing any kind of comparison between `Time` points with different clock types. Internally I received feedback that's pointed towards `Time` points for different clock types should have different types so that enforcement can happen at compile time. Given that direction, I'll close this PR.

Issue #2096, opened 2023-02-01
rclcpp::Context::shutdown docs imply on_shutdown callbacks are called in the order added, but implementation iterates an unordered_set


mmattb:
## Bug report

**Required Info:**

- Operating System:
  - All
- Installation type:
  - All
- Version or commit hash:
  - be8c5d01c683ea381ffb77f4fc246411d8d31639   (i.e. today's HEAD)
- DDS implementation:
  - All
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

This is a rough sketch of what we are trying in nav2.  I doubt anything this complicated is needed. Probably just making two arbitrary callbacks and registering them in different orders will repro the issue. You just need to beat the hash algorithm basically.

```
class MyChild : public rclcpp::LifecycleNode {
   MyChild() {
       register_my_rcl_preshutdown_callback();
   }

  ...
};

class MyDescendant : public MyChild
{
  MyDescendant() : MyChild(blah) {

    // Super constructor already ran by this point, so the MyDescendant
    // instance should have its preshutdown callback called before my_member's
    // preshutdown callback, but we are sometimes seeing them in reverse order!
    auto my_member = std::make_shared<MyChild>(blah);

  }

  ...
};
```

#### Actual behavior

In nav2 we are working on a fix which uses the preshutdown callback mechanism. A pair of callbacks are sometimes firing in the reverse order from which they were added.

It looks like both preshutdown callbacks and on shutdown callbacks use an unordered_set:

https://github.com/ros2/rclcpp/blob/37adc03c11171f67c213ee72f7ce43fdcbda7b9e/rclcpp/include/rclcpp/context.hpp#L379

And they are executed via an iteration of those sets, which I believe means they will execute in some arbitrary hash order?

https://github.com/ros2/rclcpp/blob/37adc03c11171f67c213ee72f7ce43fdcbda7b9e/rclcpp/src/rclcpp/context.cpp#L315

#### Expected behavior

But the docs state pretty clearly that callbacks will be executed in the order they are added:

https://github.com/ros2/rclcpp/blob/37adc03c11171f67c213ee72f7ce43fdcbda7b9e/rclcpp/include/rclcpp/context.hpp#L228

https://github.com/ros2/rclcpp/blob/37adc03c11171f67c213ee72f7ce43fdcbda7b9e/rclcpp/include/rclcpp/context.hpp#L178

[...]

PR #2097, opened 2023-02-02
Fix bug on the disorder of calling shutdown callback

Barry-Xu-2018:
Address #2096 

Signed-off-by: Barry Xu <barry.xu@sony.com>

### DIFF
--- a/rclcpp/include/rclcpp/context.hpp
+++ b/rclcpp/include/rclcpp/context.hpp
@@ -376,10 +376,10 @@ private:
   // attempt to acquire another sub context.
   std::recursive_mutex sub_contexts_mutex_;
 
-  std::unordered_set<std::shared_ptr<OnShutdownCallback>> on_shutdown_callbacks_;
+  std::vector<std::shared_ptr<OnShutdownCallback>> on_shutdown_callbacks_;
   mutable std::mutex on_shutdown_callbacks_mutex_;
 
-  std::unordered_set<std::shared_ptr<PreShutdownCallback>> pre_shutdown_callbacks_;
+  std::vector<std::shared_ptr<PreShutdownCallback>> pre_shutdown_callbacks_;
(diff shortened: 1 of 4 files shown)


alsora (approved):

fujitatomoya (approved):
lgtm! @Barry-Xu-2018 thanks for quick response.

fujitatomoya:
@clalancette @alsora thanks for the review, btw can we backport this to humble?

fujitatomoya:
I will start the CI.

clalancette:
> @clalancette @alsora thanks for the review, btw can we backport this to humble?

Hm.  So unfortunately, this breaks ABI because of the change of type from `std::unordered_set` to `std::vector`.  So this would not generally be eligible for backporting.  We'd have to consider another non-ABI breaking way to backport this, or explicitly make an exception here.

fujitatomoya:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18664/)

fujitatomoya:
@clalancette if the backport is hard requirement, we would probably consider other approach. I believe that breaking ABI/API policy would be much bigger problem.

Barry-Xu-2018:
@clalancette @fujitatomoya 

In order to keep ABI compatible, we cannot add new variable (except static variable) into class Context.   
Maybe use global variable as one solution, but it will make code ugly.   
```c++
struct ShutdownCallback {
  std::vector<weak_ptr> pre_callback;
  std::vector<weak_ptr> on_callback;
};

static std:map<Context*, ShutdownCallback> g_shutdownCallback;
static mutex s_shutdownCallback_mutex;
```

Change code on adding/removing/call shutdown callback to use this global variable.   

Now I not find better approach

clalancette:
I'm going to go ahead and merge this one in, since it is an obvious improvement for Rolling.  We can consider fixing Humble (either the documentation or with an ABI-compatible fix) separately.

PR #2104, opened 2023-02-15
Fix memory leak in tracetools::get_symbol()

christophebedard:
Part of https://github.com/ros2/ros2_tracing/issues/34

Requires https://github.com/ros2/ros2_tracing/pull/43

Requires https://github.com/ros2/ros2_tracing/pull/46 to help avoid memory allocations when the tracepoint isn't enabled.

### DIFF
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -191,10 +191,14 @@ public:
 #ifndef TRACETOOLS_DISABLED
     std::visit(
       [this](auto && arg) {
-        TRACEPOINT(
-          rclcpp_callback_register,
-          static_cast<const void *>(this),
-          tracetools::get_symbol(arg));
+        if (TRACEPOINT_ENABLED(rclcpp_callback_register)) {
+          char * symbol = tracetools::get_symbol(arg);
+          DO_TRACEPOINT(
+            rclcpp_callback_register,
+            static_cast<const void *>(this),
+            symbol);
+          std::free(symbol);
+        }
       }, callback_);
 #endif  // TRACETOOLS_DISABLED
   }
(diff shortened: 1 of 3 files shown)


christophebedard:
This is the simplest fix, but it could be improved. For example, we could first check if the tracepoint is actually enabled before calling `tracetools::get_symbol()` and allocating. This would require changes in `tracetools`, though. See https://github.com/ros2/ros2_tracing/pull/43#issuecomment-1429001945

alsora (approved):

alsora:
Isn't the macro sufficient to prevent using this code if tracetools is disabled?
Or do we support both "compile time disabled" and "runtime disabled" ?

mjcarroll:
> Or do we support both "compile time disabled" and "runtime disabled" ?

When building with tracepoints enabled, there will be some runtime associated here, even if the tracepoints are not actively being used.  I think this is more important as we are discussing building with tracepoints by default in the binary packages.

christophebedard:
In this case, this tracepoint is only triggered during initialization, so the overhead is fairly small. But, yes, being able to not do this kind of processing if a tracepoint is disabled at runtime would be much better.

Since that change (as described here https://github.com/ros2/ros2_tracing/pull/43#issuecomment-1429001945) requires modifications to `tracetools` and to potential downstream packages (including `rclcpp` for this exact instance), I could first add `TRACEPOINT_ENABLED(...)` and `DO_TRACEPOINT(...)` to `tracetools`. Then we'd only need to modify `rclcpp` once.

mjcarroll (approved):

christophebedard:
Testing `--packages-above tracetools test_tracetools rclcpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18089/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12621/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18733/)

Includes:

* https://github.com/ros2/ros2_tracing/pull/43
* This PR

mjcarroll (approved):
Another approval for good measure

christophebedard:
@mjcarroll thanks :laughing:

I can't merge this PR, so I'll let you merge it.

PR #2114, opened 2023-03-01
Revert "Fix memory leak in tracetools::get_symbol()"

mjcarroll:
Reverts ros2/rclcpp#2104


christophebedard:
We won't need to revert (see https://github.com/ros2/ros2_tracing/pull/48), so we can close this.

PR #2123, opened 2023-03-10
Make Rate to select the clock to work with

AlexeyMerzlyakov:
This is the continue of [#1373] [discussion](https://github.com/ros2/rclcpp/pull/1373#discussion_r1118830044) and related to the https://github.com/ros-planning/navigation2/issues/3325 issue in Nav2

The initial idea is to make `rclcpp::Rate` to work with ROS-time (currently it works only with system and steady times).

What was made:
* Add `rclcpp::Clock` pointer to the `rclcpp::Rate` constructor to allow developer to specify the clock to work with (by-default it is still `RCL_SYSTEM_TIME` if no clock is specified)
* Add `rclcpp::ROSRate` rate class, respective to ROS-time

Current implementation points:
* Getting rid of templates' structure allows developer to select the clock to work with in dynamics
* Renamed `rclcpp::GenericRate` -> to `rclcpp::Rate` and gave it the default clock to work with
* `rclcpp::Rate` and `rclcpp::WallRate` API remaining to be untouched as much as possible
* Although there are some changes in API: `rclcpp::Duration` instead of `std::chrono::nanoseconds` in durations and `Rate::get_type()` routine instead of previously used `Rate::is_steady()`
* New `rclcpp::ROSRate` class that working with ROS time

### DIFF
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -39,17 +39,19 @@ public:
 };
 
 template<class Clock = std::chrono::high_resolution_clock>
-class [[deprecated("use rclcpp::Rate class instead of GenericRate")]] GenericRate : public RateBase
+class [[deprecated("use rclcpp::Rate class instead of GenericRate")]] GenericRate
 {
 public:
   RCLCPP_SMART_PTR_DEFINITIONS(GenericRate)
 
   explicit GenericRate(double) {}
   explicit GenericRate(std::chrono::nanoseconds) {}
+  virtual ~GenericRate() {}
 
   virtual bool sleep() {return false;}
-  virtual rcl_clock_type_t get_type() const {return RCL_CLOCK_UNINITIALIZED;}
+  virtual bool is_steady() const {return false;}
   virtual void reset() {}
+  std::chrono::nanoseconds period() const {return std::chrono::seconds(1);}
 };
 
 class Rate : public RateBase
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -33,10 +33,19 @@ class RateBase
 public:
   RCLCPP_SMART_PTR_DEFINITIONS_NOT_COPYABLE(RateBase)
 
+  RCLCPP_PUBLIC
   virtual ~RateBase() {}
+
+  RCLCPP_PUBLIC
   virtual bool sleep() = 0;
+
+  RCLCPP_PUBLIC
   [[deprecated("use get_type() instead")]] virtual bool is_steady() const = 0;
+
+  RCLCPP_PUBLIC
   virtual rcl_clock_type_t get_type() const = 0;
+
+  RCLCPP_PUBLIC
   virtual void reset() = 0;
 };
 
@@ -120,32 +129,39 @@ private:
   std::chrono::time_point<Clock, ClockDurationNano> last_interval_;
 };
 
-class RCLCPP_PUBLIC Rate : public RateBase
+class Rate : public RateBase
 {
 public:
   RCLCPP_SMART_PTR_DEFINITIONS(Rate)
 
+  RCLCPP_PUBLIC
   explicit Rate(
     const double rate,
     Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME));
 
+  RCLCPP_PUBLIC
   explicit Rate(
     const Duration & period,
     Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME));
 
+  RCLCPP_PUBLIC
   virtual bool
   sleep();
 
+  RCLCPP_PUBLIC
   [[deprecated("use get_type() instead")]]
   virtual bool
   is_steady() const;
 
+  RCLCPP_PUBLIC
   virtual rcl_clock_type_t
   get_type() const;
 
+  RCLCPP_PUBLIC
   virtual void
   reset();
 
+  RCLCPP_PUBLIC
   Duration
   period() const;
 
@@ -157,17 +173,23 @@ private:
   Time last_interval_;
 };
 
-class RCLCPP_PUBLIC WallRate : public Rate
+class WallRate : public Rate
 {
 public:
+  RCLCPP_PUBLIC
   explicit WallRate(const double rate);
+
+  RCLCPP_PUBLIC
   explicit WallRate(const Duration & period);
 };
 
-class RCLCPP_PUBLIC ROSRate : public Rate
+class ROSRate : public Rate
 {
 public:
+  RCLCPP_PUBLIC
   explicit ROSRate(const double rate);
+
+  RCLCPP_PUBLIC
   explicit ROSRate(const Duration & period);
 };
 
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -131,7 +131,7 @@ public:
   : clock_(clock), period_(0, 0), last_interval_(clock_->now())
   {
     if (rate <= 0.0) {
-      throw std::invalid_argument{"rate must greater than 0"};
+      throw std::invalid_argument{"rate must be greater than 0"};
     }
     period_ = Duration::from_seconds(1.0 / rate);
   }
@@ -141,7 +141,7 @@ public:
   : clock_(clock), period_(period), last_interval_(clock_->now())
   {
     if (period <= Duration(0, 0)) {
-      throw std::invalid_argument{"period must greater than 0"};
+      throw std::invalid_argument{"period must be greater than 0"};
     }
   }
 
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -39,8 +39,9 @@ public:
   RCLCPP_PUBLIC
   virtual bool sleep() = 0;
 
+  [[deprecated("use get_type() instead")]]
   RCLCPP_PUBLIC
-  [[deprecated("use get_type() instead")]] virtual bool is_steady() const = 0;
+  virtual bool is_steady() const = 0;
 
   RCLCPP_PUBLIC
   virtual rcl_clock_type_t get_type() const = 0;
@@ -148,8 +149,8 @@ public:
   virtual bool
   sleep();
 
-  RCLCPP_PUBLIC
   [[deprecated("use get_type() instead")]]
+  RCLCPP_PUBLIC
   virtual bool
   is_steady() const;
 
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -127,23 +127,11 @@ public:
 
   explicit Rate(
     const double rate,
-    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME))
-  : clock_(clock), period_(0, 0), last_interval_(clock_->now())
-  {
-    if (rate <= 0.0) {
-      throw std::invalid_argument{"rate must be greater than 0"};
-    }
-    period_ = Duration::from_seconds(1.0 / rate);
-  }
+    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME));
+
   explicit Rate(
     const Duration & period,
-    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME))
-  : clock_(clock), period_(period), last_interval_(clock_->now())
-  {
-    if (period <= Duration(0, 0)) {
-      throw std::invalid_argument{"period must be greater than 0"};
-    }
-  }
+    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME));
 
   virtual bool
   sleep();
@@ -172,25 +160,15 @@ private:
 class WallRate : public Rate
 {
 public:
-  explicit WallRate(const double rate)
-  : Rate(rate, std::make_shared<Clock>(RCL_STEADY_TIME))
-  {}
-
-  explicit WallRate(const Duration & period)
-  : Rate(period, std::make_shared<Clock>(RCL_STEADY_TIME))
-  {}
+  explicit WallRate(const double rate);
+  explicit WallRate(const Duration & period);
 };
 
 class ROSRate : public Rate
 {
 public:
-  explicit ROSRate(const double rate)
-  : Rate(rate, std::make_shared<Clock>(RCL_ROS_TIME))
-  {}
-
-  explicit ROSRate(const Duration & period)
-  : Rate(period, std::make_shared<Clock>(RCL_ROS_TIME))
-  {}
+  explicit ROSRate(const double rate);
+  explicit ROSRate(const Duration & period);
 };
 
 }  // namespace rclcpp
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -35,7 +35,7 @@ public:
 
   virtual ~RateBase() {}
   virtual bool sleep() = 0;
-  virtual bool is_steady() const = 0;
+  [[deprecated("use get_type() instead")]] virtual bool is_steady() const = 0;
   virtual rcl_clock_type_t get_type() const = 0;
   virtual void reset() = 0;
 };
@@ -45,7 +45,7 @@ using std::chrono::duration_cast;
 using std::chrono::nanoseconds;
 
 template<class Clock = std::chrono::high_resolution_clock>
-class [[deprecated("use rclcpp::Rate class instead of GenericRate")]] GenericRate
+class [[deprecated("use rclcpp::Rate class instead of GenericRate")]] GenericRate : public RateBase
 {
 public:
   RCLCPP_SMART_PTR_DEFINITIONS(GenericRate)
@@ -56,8 +56,6 @@ public:
   explicit GenericRate(std::chrono::nanoseconds period)
   : period_(period), last_interval_(Clock::now())
   {}
-
-  virtual ~GenericRate() {}
 
   virtual bool
   sleep()
@@ -91,10 +89,16 @@ public:
     return true;
   }
 
+  [[deprecated("use get_type() instead")]]
   virtual bool
   is_steady() const
   {
     return Clock::is_steady;
+  }
+
+  virtual rcl_clock_type_t get_type() const
+  {
+    return Clock::is_steady ? RCL_STEADY_TIME : RCL_SYSTEM_TIME;
   }
 
   virtual void
@@ -164,6 +168,7 @@ public:
     return clock_->sleep_for(time_to_sleep);
   }
 
+  [[deprecated("use get_type() instead")]]
   virtual bool
   is_steady() const
   {
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -15,10 +15,10 @@
 #ifndef RCLCPP__RATE_HPP_
 #define RCLCPP__RATE_HPP_
 
-#include <chrono>
 #include <memory>
-#include <thread>
 
+#include "rclcpp/clock.hpp"
+#include "rclcpp/duration.hpp"
 #include "rclcpp/macros.hpp"
 #include "rclcpp/utilities.hpp"
 #include "rclcpp/visibility_control.hpp"
@@ -33,33 +33,32 @@ public:
 
   virtual ~RateBase() {}
   virtual bool sleep() = 0;
-  virtual bool is_steady() const = 0;
+  virtual rcl_clock_type_t get_type() const = 0;
   virtual void reset() = 0;
 };
 
-using std::chrono::duration;
-using std::chrono::duration_cast;
-using std::chrono::nanoseconds;
-
-template<class Clock = std::chrono::high_resolution_clock>
-class GenericRate : public RateBase
+class Rate : public RateBase
 {
 public:
-  RCLCPP_SMART_PTR_DEFINITIONS(GenericRate)
+  RCLCPP_SMART_PTR_DEFINITIONS(Rate)
 
-  explicit GenericRate(double rate)
-  : GenericRate<Clock>(
-      duration_cast<nanoseconds>(duration<double>(1.0 / rate)))
+  explicit Rate(
+    const double rate,
+    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME))
+  : Rate(
+      Duration::from_seconds(1.0 / rate), clock)
   {}
-  explicit GenericRate(std::chrono::nanoseconds period)
-  : period_(period), last_interval_(Clock::now())
+  explicit Rate(
+    const Duration & period,
+    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME))
+  : clock_(clock), period_(period), last_interval_(clock_->now())
   {}
 
   virtual bool
   sleep()
   {
     // Time coming into sleep
-    auto now = Clock::now();
+    auto now = clock_->now();
     // Time of next interval
     auto next_interval = last_interval_ + period_;
     // Detect backwards time flow
@@ -67,12 +66,10 @@ public:
       // Best thing to do is to set the next_interval to now + period
       next_interval = now + period_;
     }
-    // Calculate the time to sleep
-    auto time_to_sleep = next_interval - now;
     // Update the interval
     last_interval_ += period_;
     // If the time_to_sleep is negative or zero, don't sleep
-    if (time_to_sleep <= std::chrono::seconds(0)) {
+    if (next_interval <= now) {
       // If an entire cycle was missed then reset next interval.
       // This might happen if the loop took more than a cycle.
       // Or if time jumps forward.
@@ -82,38 +79,62 @@ public:
       // Either way do not sleep and return false
       return false;
     }
+    // Calculate the time to sleep
+    auto time_to_sleep = next_interval - now;
     // Sleep (will get interrupted by ctrl-c, may not sleep full time)
-    rclcpp::sleep_for(time_to_sleep);
+    clock_->sleep_for(time_to_sleep);
     return true;
   }
 
-  virtual bool
-  is_steady() const
+  virtual rcl_clock_type_t
+  get_type() const
   {
-    return Clock::is_steady;
+    return clock_->get_clock_type();
   }
 
   virtual void
   reset()
   {
-    last_interval_ = Clock::now();
+    last_interval_ = clock_->now();
   }
 
-  std::chrono::nanoseconds period() const
+  Duration period() const
   {
     return period_;
   }
 
 private:
-  RCLCPP_DISABLE_COPY(GenericRate)
+  RCLCPP_DISABLE_COPY(Rate)
 
-  std::chrono::nanoseconds period_;
-  using ClockDurationNano = std::chrono::duration<typename Clock::rep, std::nano>;
-  std::chrono::time_point<Clock, ClockDurationNano> last_interval_;
+  Clock::SharedPtr clock_;
+  Duration period_;
+  Time last_interval_;
 };
 
-using Rate = GenericRate<std::chrono::system_clock>;
-using WallRate = GenericRate<std::chrono::steady_clock>;
+class WallRate : public Rate
+{
+public:
+  explicit WallRate(const double rate)
+  : Rate(rate, std::make_shared<Clock>(RCL_STEADY_TIME))
+  {}
+
+  explicit WallRate(const Duration & period)
+  : Rate(period, std::make_shared<Clock>(RCL_STEADY_TIME))
+  {}
+};
+
+class ROSRate : public Rate
+
+{
+public:
+  explicit ROSRate(const double rate)
+  : Rate(rate, std::make_shared<Clock>(RCL_ROS_TIME))
+  {}
+
+  explicit ROSRate(const Duration & period)
+  : Rate(period, std::make_shared<Clock>(RCL_ROS_TIME))
+  {}
+};
 
 }  // namespace rclcpp
 
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -157,7 +157,7 @@ private:
   Time last_interval_;
 };
 
-class WallRate : public Rate
+class RCLCPP_PUBLIC WallRate : public Rate
 {
 public:
   explicit WallRate(const double rate);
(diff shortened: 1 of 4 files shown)


AlexeyMerzlyakov:
The PR contains only the `time.hpp` change. If the change will be agreed, I will fix/add new tests on it.

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -33,46 +33,43 @@ class RateBase
...
-
-template<class Clock = std::chrono::high_resolution_clock>
-class GenericRate : public RateBase
+class Rate : public RateBase
 {
 public:
-  RCLCPP_SMART_PTR_DEFINITIONS(GenericRate)
+  RCLCPP_SMART_PTR_DEFINITIONS(Rate)
 
-  explicit GenericRate(double rate)
```
fujitatomoya, line 50: I think we need to deprecate this for this moment with following message.


``[[deprecated("use rclcpp::Rate class instead of GenericRate")]]``

fujitatomoya (commented):
overall lgtm

rclcpp/include/rclcpp/rate.hpp
AlexeyMerzlyakov: Done in https://github.com/ros2/rclcpp/pull/2123/commits/c25fe6bdad6ee686fa411a89cd7ae140e0f3ac40 and https://github.com/ros2/rclcpp/pull/2123/commits/7323e9730a8f3e288177bb30b74777f34e33dfee. `GenericRate`'s API remained to be the same in order to make the compiler to inform developer only about class name change. Implementation was cleaned-out.

SteveMacenski (changes requested):

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -33,46 +34,59 @@ class RateBase
...
+    auto now = clock_->now();
     // Time of next interval
     auto next_interval = last_interval_ + period_;
     // Detect backwards time flow
     if (now < last_interval_) {
       // Best thing to do is to set the next_interval to now + period
       next_interval = now + period_;
     }
-    // Calculate the time to sleep
-    auto time_to_sleep = next_interval - now;
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: In vanilla branch, `time_to_sleep` is used to [avoid](https://github.com/ros2/rclcpp/blob/20e9cd17f6723337dc0915f4e2a966f9585c309c/rclcpp/include/rclcpp/rate.hpp#L75) negative values, and was initialized [just before](https://github.com/ros2/rclcpp/blob/20e9cd17f6723337dc0915f4e2a966f9585c309c/rclcpp/include/rclcpp/rate.hpp#L71) the usage. In this patch, `time_to_sleep` became to be of `rclcpp::Duration` type, which can not be negative, so this check was [transformed](https://github.com/AlexeyMerzlyakov/rclcpp/blob/7323e9730a8f3e288177bb30b74777f34e33dfee/rclcpp/include/rclcpp/rate.hpp#L89). `time_to_sleep` became to be required later, so it [was moved](https://github.com/AlexeyMerzlyakov/rclcpp/blob/7323e9730a8f3e288177bb30b74777f34e33dfee/rclcpp/include/rclcpp/rate.hpp#L100) to the place right before usage, for better code readability.

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -33,46 +34,59 @@ class RateBase
...
-class GenericRate : public RateBase
+class [[deprecated("use rclcpp::Rate class instead of GenericRate")]] GenericRate
 {
 public:
   RCLCPP_SMART_PTR_DEFINITIONS(GenericRate)
 
-  explicit GenericRate(double rate)
-  : GenericRate<Clock>(
-      duration_cast<nanoseconds>(duration<double>(1.0 / rate)))
+  explicit GenericRate(double) {}
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: The purpose to leave `GenericRate` class - is to warn developer that `Rate` should be used instead, from on now. All implementations were left empty to have only one one compiler deprecation warning message and thus to not cause a confusion of other inducted compiler errors (e.g. like "can not find such constructor or method"). It is supposed that developer will switch to `Rate` class after this message, so `GenericRate`'s implementations were emptied.
However, indeed, if warnings were disabled as errors on developer side, and he will refuse to follow the recommendations, this indeed might cause an undefined behavior.

> I'm not sure why this doesn't use `Rate` or `Rate` isn't implemented as `GenericRate` and we typedef a `Rate` from it or something if it doesn't make sense to have dual implementations

In vanilla version, `Rate` is a templated implementation of `GenericRate` (`using Rate = GenericRate<std::chrono::system_clock>;` or as you've mentioned, `typedef GenericRate<std::chrono::system_clock> Rate;` ). However, in current implementation, `Rate` concept now was changed. It is non-templated class, which can not be a template typedef of `GenericRate`. Moreover, if forcibly leave the implementation to use the `Rate` as `typedef` of `GenericRate` (as it now in vanilla), `GenericRate` as deprecated class, will cause compiler to warn when we will try to make the `Rate` as `typedef GenericRate<something>` .

Or did you mean to leave `GenericRate` untouched and have `Rate` as a completely new class, having dual-implementations? Wouldn't it cause even more confusion?

Finally, `GenericRate`'s implementation could be completely removed, which will cause additional compiler errors, but it never allow the developer to build their code with older API and thus to have an undefined behavior.

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -82,38 +96,61 @@ class GenericRate : public RateBase
       // Either way do not sleep and return false
       return false;
     }
+    // Calculate the time to sleep
+    auto time_to_sleep = next_interval - now;
     // Sleep (will get interrupted by ctrl-c, may not sleep full time)
-    rclcpp::sleep_for(time_to_sleep);
+    clock_->sleep_for(time_to_sleep);
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: The `rclcpp::Clock::sleep_for()` - is a [wrapper](https://github.com/ros2/rclcpp/blob/20e9cd17f6723337dc0915f4e2a966f9585c309c/rclcpp/src/rclcpp/clock.cpp#L179-L183) for `rclcpp::Clock::sleep_until()`. `sleep_until()` seems to have a time [clock handler](https://github.com/ros2/rclcpp/blob/20e9cd17f6723337dc0915f4e2a966f9585c309c/rclcpp/src/rclcpp/clock.cpp#L140), catching time jumps and clock source changes. It seems, that during this change waiting is being interrupted [[1](https://github.com/ros2/rclcpp/blob/20e9cd17f6723337dc0915f4e2a966f9585c309c/rclcpp/src/rclcpp/clock.cpp#L146)], [[2](https://github.com/ros2/rclcpp/blob/20e9cd17f6723337dc0915f4e2a966f9585c309c/rclcpp/src/rclcpp/clock.cpp#L167)].

So, we have two types of situations:
1. Time was jumped (backward or forward) during the waiting:
  - In this case this situation will be handled correctly by the `rclcpp::Clock::sleep_until()`.
  One thing it looks I've missed: `sleep()` routine needs return the value from `sleep_for()` instead of always `true`:
  ```
  - clock_->sleep_for(time_to_sleep);
  - return true;
  + return clock_->sleep_for(time_to_sleep);
  ```
2. Time was jumped between `sleep()` beginning (when we already got `now` variable) and `clock_->sleep_for()` call. Very rare situation, but anyway:
 - In this case, inside `sleep_for()` the time to sleep will be [calculated](https://github.com/ros2/rclcpp/blob/20e9cd17f6723337dc0915f4e2a966f9585c309c/rclcpp/src/rclcpp/clock.cpp#L182) in already new time after jump, and we will sleep a required `time_to_sleep`, but in new time. Not fully correct situaion, but later on, on next `sleep()` call, if time was jumped (back or forward) more than on one `period_`, it will be handled inside the `sleep()`: for [back](https://github.com/AlexeyMerzlyakov/rclcpp/blob/7323e9730a8f3e288177bb30b74777f34e33dfee/rclcpp/include/rclcpp/rate.hpp#L82) jump and for [forward](https://github.com/AlexeyMerzlyakov/rclcpp/blob/7323e9730a8f3e288177bb30b74777f34e33dfee/rclcpp/include/rclcpp/rate.hpp#L89) jump; and situation will be stabilized.

rclcpp/include/rclcpp/rate.hpp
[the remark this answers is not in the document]
SteveMacenski: I don't think that's the right philosophy to have on deprecation warnings. The warning is to let people know it will be depreciated soon in future releases, not that what you're currently using is already deprecated and will not work.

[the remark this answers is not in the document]
clalancette: > I don't think that's the right philosophy to have on deprecation warnings. The warning is to let people know it will be depreciated soon in future releases, not that what you're currently using is already deprecated and will not work.

Yes, agreed.  The old API needs to continue to work (with a compile-time deprecation warning).

rclcpp/include/rclcpp/rate.hpp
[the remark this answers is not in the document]
AlexeyMerzlyakov: > I'll do it in next PR update.

Done in the [latest commit](https://github.com/ros2/rclcpp/pull/2123/commits/30b95de7cc242e1981091f406fb9c8226d34f97f#diff-c6ecfbeee7b7384c9e72a8984cc2fc718415b3bb741c0ea41ffc1d9b0718fea7R164)

rclcpp/include/rclcpp/rate.hpp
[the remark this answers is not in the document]
AlexeyMerzlyakov: Okay, got it. I've restored the previous functionality of `GenericRate` class and API (including `is_steady()`) in the latest update

AlexeyMerzlyakov:
The build was failed due to outdated branch. Rebased over latest `rolling` and force-pushed branch.

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -42,19 +45,20 @@ using std::chrono::duration_cast;
 using std::chrono::nanoseconds;
 
 template<class Clock = std::chrono::high_resolution_clock>
-class GenericRate : public RateBase
+class [[deprecated("use rclcpp::Rate class instead of GenericRate")]] GenericRate
```
SteveMacenski, line 48: Shouldn't this derive from RateBase again?

SteveMacenski (approved):
Otherwise LGTM

rclcpp/include/rclcpp/rate.hpp
AlexeyMerzlyakov: Since of meaning of clock types has been updated (ROS-respecitve clock type added), new `rcl_clock_type_t get_type()` method [was introduced](https://github.com/ros2/rclcpp/pull/2123/files#diff-c6ecfbeee7b7384c9e72a8984cc2fc718415b3bb741c0ea41ffc1d9b0718fea7R39) in a `RateBase` basic class as a alternative/replacement for `bool is_steady()` one (I would like to completely remove `is_steady()` method from on now as unnecessary, but since we need to support old API for compativility for `rclcpp::Rate` and `rclcpp::WallRate` classes, `is_steady` was remain). `GenericRate` if being a child-class of `RateBase` now should implement `get_type()` method. But since `GenericRate` - is deprecated class, it might be overkill, I believe.

Therefore `GenericRate` was taken out of parenthood of `RateBase` remaining the same functionality as it previously has; while new `RateBase -> Rate -> WallRate/ROSRate` lineage was introduced instead.

clalancette:
Here is CI (just on Linux) to see how much we would have to change to get this in:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18580/)

clalancette:

clalancette:
So there were a *lot* of failures in CI.  That may be because this needs to be rebased, I'm not sure.  @AlexeyMerzlyakov @SteveMacenski can you rebase this onto the latest and we can try again?

wjwwood:
@Mergifyio rebase

wjwwood:
@Mergifyio rebase

wjwwood:
New CI after rebase: [Build Status](https://ci.ros2.org/job/ci_linux/18587/)

SteveMacenski:
@AlexeyMerzlyakov take a look. The policy is that deprecations in the core need to have PRs to resolve so that builds are green without deprecated API use. So anywhere that we have failures we should have complimentary PRs in the ROS 2 core so that everything turns over OK. Unfortunately due to that delay, this won't be able to get in before Iron's freeze

SteveMacenski:
Any update?

AlexeyMerzlyakov:
Waiting until https://github.com/ros2/system_tests/pull/516 (that covers failed tests) to be reviewed on the way to its merge. Then we need to restart testing again in this PR to make sure that all regressions were resolved.

rclcpp/include/rclcpp/rate.hpp
fujitatomoya: Sorry to be late to get back to you. What we need to deprecate is `RateBase::is_steady` since the application can get the clock type?

fujitatomoya (commented):
overall lgtm! just one minor comment.

rclcpp/include/rclcpp/rate.hpp
clalancette: The problem with removing `GenericRate` from the hierarchy is that this breaks certain applications.  Consider for example if a downstream node has `std::vector<RateBase>` in their code.  In that case, they will no longer be able to store a `GenericRate` in there, since it is no longer a child class of `RateBase`.

So I think we should add this back to the hierarchy (i.e. make this a child of `RateBase`), and implement `get_type()` as returning either `RCL_SYSTEM_TIME` (if `Clock::is_steady` is false), or `RCL_STEADY_TIME` (if `Clock::is_steady` is true).

AlexeyMerzlyakov: @fujitatomoya, OK. Thanks for reaching me. I've made `is_steady()` method to be deprecated for all classes in [this](https://github.com/ros2/rclcpp/pull/2123/commits/5f08e09bb048558689fb4e54e28b94bd1f9340ca) commit.

@clalancette, Got your point. This is indeed makes sense, so `RateBase` -> `GenericRate` relationship was restored in the [same](https://github.com/ros2/rclcpp/pull/2123/commits/5f08e09bb048558689fb4e54e28b94bd1f9340ca) commit.

AlexeyMerzlyakov:
For the clarity, I've prepared API `Rate*` change briefly summarized in the following document:
[Rate_API_change.pdf](https://github.com/ros2/rclcpp/files/12171335/Rate_API_change.pdf)

Hope, this might help to take an overall look on this change.

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -112,8 +120,109 @@ class GenericRate : public RateBase
...
+        last_interval_ = now + period_;
+      }
+      // Either way do not sleep and return false
+      return false;
+    }
+    // Calculate the time to sleep
+    auto time_to_sleep = next_interval - now;
+    // Sleep (will get interrupted by ctrl-c, may not sleep full time)
+    return clock_->sleep_for(time_to_sleep);
+  }
```
clalancette, line 169: Given that this class is not a template, I'd like to see all of this implementation moved into a `rate.cpp` file.  That will help with compile times of downstream projects.

The same goes for all of the other methods below.

clalancette (changes requested):
Thank you so much for putting together the document explaining the change.  It is really helpful.

I've left one change inline.

Besides that, the fact that the `Rate` and `WallRate` classes now require a context to be initialized is a big change from the previous implementation.  That doesn't necessarily mean we shouldn't do this, but I do want to have a conversation about it with the maintainers.  @fujitatomoya @alsora @wjwwood what do you think about this change in semantics?

fujitatomoya:
> Thank you so much for putting together the document explaining the change. It is really helpful.
> 
> I've left one change inline.
> 
> Besides that, the fact that the `Rate` and `WallRate` classes now require a context to be initialized is a big change from the previous implementation. That doesn't necessarily mean we shouldn't do this, but I do want to have a conversation about it with the maintainers. @fujitatomoya @alsora @wjwwood what do you think about this change in semantics?

that is true, after this PR, context must be initialized before using those classes. i think that would not be a problem for ROS 2 applications. and depending on ROS 2 `Clock` objects makes sense to me to support `ROSRate`.

alsora:
The PR looks good to me.
Before merging we should address the comment here (https://github.com/ros2/rclcpp/pull/2123#discussion_r1274959585) about moving stuff to the cpp file.

fujitatomoya:
@AlexeyMerzlyakov friendly ping. i will restart the CI all comments are resolved.

rclcpp/include/rclcpp/rate.hpp
AlexeyMerzlyakov: Moved all `Rate` implementations into source-file in https://github.com/ros2/rclcpp/pull/2123/commits/3d7b2232dca56ddd1021c174a4617714bb2f9538

AlexeyMerzlyakov:
Apologies for late respond. Yes, this definitely makes sense: moved all non-template implementations to separate source-file in the following commit: https://github.com/ros2/rclcpp/pull/2123/commits/3d7b2232dca56ddd1021c174a4617714bb2f9538, to save compile time of all dependent packages/projects.

@fujitatomoya, before CI-testing restart we need to ensure that https://github.com/ros2/system_tests/pull/516 is in local test container (if it possible to do this).

clalancette (commented):
This looks good to me!  Thank you for all of the iterations.  I'm going to run CI on it (along with https://github.com/ros2/system_tests/pull/516) on a number of platforms and see where we are.

fujitatomoya (approved):
with a couple of nitpicks, lgtm.

clalancette:
Also, this likely needs a rebase onto the latest before we can run CI on it.

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -112,8 +120,70 @@ class GenericRate : public RateBase
...
+class Rate : public RateBase
+{
+public:
+  RCLCPP_SMART_PTR_DEFINITIONS(Rate)
+
+  explicit Rate(
+    const double rate,
+    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME))
+  : Rate(
+      Duration::from_seconds(1.0 / rate), clock)
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: It was initially copied from `GenericRate` logic, but yes, this definitely makes sense. [Added](https://github.com/ros2/rclcpp/pull/2123/commits/8f19f28dfb72c50ab1d353026ffdf366777f327f) constructors argument checks at least for new `Rate` class. Do we need to add similar checks for obsolete `GenericRate` constructors, or better leave it as-is, making changes only in working API?

rclcpp/src/rclcpp/rate.cpp
```diff
@@ -0,0 +1,75 @@
+// Copyright 2014 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: Done in https://github.com/ros2/rclcpp/pull/2123/commits/8f19f28dfb72c50ab1d353026ffdf366777f327f

AlexeyMerzlyakov:
Force-pushed changes over the latest `rolling` commit. Also force-pushed the changes in https://github.com/ros2/system_tests/pull/516 over latest commit. So it should be ready for round 2 of CI testing.

fujitatomoya (approved):

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -112,8 +120,78 @@ class GenericRate : public RateBase
...
+  }
+  explicit Rate(
+    const Duration & period,
+    Clock::SharedPtr clock = std::make_shared<Clock>(RCL_SYSTEM_TIME))
+  : clock_(clock), period_(period), last_interval_(clock_->now())
+  {
+    if (period <= Duration(0, 0)) {
+      throw std::invalid_argument{"period must be greater than 0"};
+    }
+  }
```
clalancette, line 146: I think we should move these constructors into rate.cpp as well.

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -112,8 +120,78 @@ class GenericRate : public RateBase
...
+class ROSRate : public Rate
+{
+public:
+  explicit ROSRate(const double rate)
+  : Rate(rate, std::make_shared<Clock>(RCL_ROS_TIME))
+  {}
+
+  explicit ROSRate(const Duration & period)
+  : Rate(period, std::make_shared<Clock>(RCL_ROS_TIME))
+  {}
```
clalancette, line 193: And let's move these constructors to rate.cpp as well (sorry I'm being pedantic here, but I'm currently on a quest to reduce compile times in rclcpp and this is one way to help with that).

clalancette (changes requested):
I've left a couple more small things to fix, then I think this will be good.

As far as:

```
Do we need to add similar checks for obsolete GenericRate constructors, or better leave it as-is, making changes only in working API?
```

I'm fine with only adding the checks to the new API, and leaving `GenericRate` as-is.

rclcpp/include/rclcpp/rate.hpp
AlexeyMerzlyakov: Done in https://github.com/ros2/rclcpp/pull/2123/commits/a947972f50007dd3e0249f71f92017fcf22ca9e0

rclcpp/include/rclcpp/rate.hpp
AlexeyMerzlyakov: Yep, no problems. Done in last commit

clalancette:
Here's another round of CI to see where we are with this:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19401/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13930/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20122/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20123/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/463/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/29/)

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -112,8 +120,56 @@ class GenericRate : public RateBase
   std::chrono::time_point<Clock, ClockDurationNano> last_interval_;
 };
 
-using Rate = GenericRate<std::chrono::system_clock>;
-using WallRate = GenericRate<std::chrono::steady_clock>;
+class Rate : public RateBase
```
clalancette, line 123: Because this is no longer templated, on Windows I think we need to add the `PUBLIC` keyword:

```suggestion
class RCLCPP_PUBLIC Rate : public RateBase
```

clalancette: (if that doesn't work on Windows, then the other option is to add `RCLCPP_PUBLIC` to each of the methods in the class)

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -112,8 +120,56 @@ class GenericRate : public RateBase
...
+
+private:
+  RCLCPP_DISABLE_COPY(Rate)
+
+  Clock::SharedPtr clock_;
+  Duration period_;
+  Time last_interval_;
+};
+
+class WallRate : public Rate
```
clalancette, line 160:
```suggestion
class RCLCPP_PUBLIC WallRate : public Rate
```

rclcpp/include/rclcpp/rate.hpp
```diff
@@ -112,8 +120,56 @@ class GenericRate : public RateBase
...
+{
+public:
+  explicit WallRate(const double rate);
+  explicit WallRate(const Duration & period);
+};
+
+class ROSRate : public Rate
```
clalancette, line 167:
```suggestion
class RCLCPP_PUBLIC ROSRate : public Rate
```

clalancette (changes requested):

SteveMacenski (approved):
Beyond comments that were made, this looks good to me!

rclcpp/test/rclcpp/test_rate.cpp
```diff
@@ -99,23 +137,310 @@ TEST(TestRate, wall_rate_basics) {
...
+#else  // !defined(_WIN32)
+# pragma warning(pop)
+#endif
+    EXPECT_EQ(RCL_ROS_TIME, rate.get_type());
+  }
+}
+
+TEST(TestRate, incorrect_constuctor) {
+  // Constructor with 0-frequency
+  try {
```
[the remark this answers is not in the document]
AlexeyMerzlyakov: Yes, there is needed macro in `utils/rclcpp_gtest_macros.hpp`. Switched to it in the following commit. Thanks for notice!

AlexeyMerzlyakov:
Oops, seems like GitHub auto-commits broken the DCO signature test. Rebasing it with adding the signature and force-pushing the branch.

clalancette:
Let's see if Windows is happier with this:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20143/)

clalancette:
The good news is that it looks like we got beyond the compiling phase on Windows.  So I'm optimistic that this will be OK there.  I'm going to kick of the rest of CI here to parallelize things:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19422/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13950/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20143/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20145/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/465/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/30/)

clalancette:
So we are getting closer, but unfortunately Windows says:

```
non dll-interface class 'rclcpp::RateBase' used as base for dll-interface class 'rclcpp::Rate' (compiling source file C:\ci\ws\src\ros2\demos\action_tutorials\action_tutorials_cpp\src\fibonacci_action_server.cpp)
```

I'm honestly not sure what that means; that is a new one for me.  Maybe we need `RCLCPP_PUBLIC` on `RateBase`?  That's not a pattern I'm familiar with, but that may be what it means.

clalancette:
OK, I *think* my latest commit (bff4fd0) should fix things on Windows.  Let's run CI again and see if that is the case:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19424/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13952/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20147/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20148/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/466/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/31/)

AlexeyMerzlyakov:
@clalancette, thank you for combing-up Windows builds!
The latest Windows build fail seems really strange and related to some CI / cache issue:
```
03:59:06   * seven_zip_archive[iconv] action extractc:/temp/local-mode-cache/cache/cookbooks/windows/libraries/windows_helper.rb:74:in `cached_file': URI.unescape is obsolete (StructuredWarnings::StandardWarning)
```

However, Clang builds give linking problems, seem to be related to the latest commit
```
04:37:36 [  8%] Building CXX object CMakeFiles/rclcpp.dir/src/rclcpp/any_executable.cpp.o
04:37:36 In file included from /home/jenkins-agent/workspace/ci_linux_clang_libcxx/ws/src/ros2/rclcpp/rclcpp/src/rclcpp/any_executable.cpp:15:
04:37:36 In file included from /home/jenkins-agent/workspace/ci_linux_clang_libcxx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/any_executable.hpp:20:
04:37:36 In file included from /home/jenkins-agent/workspace/ci_linux_clang_libcxx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/callback_group.hpp:30:
04:37:36 In file included from /home/jenkins-agent/workspace/ci_linux_clang_libcxx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/timer.hpp:31:
04:37:36 /home/jenkins-agent/workspace/ci_linux_clang_libcxx/ws/src/ros2/rclcpp/rclcpp/include/rclcpp/rate.hpp:43:3: error: an attribute list cannot appear here
04:37:36   [[deprecated("use get_type() instead")]] virtual bool is_steady() const = 0;
```

Unfortunately, I checked only Linux builds when prepared the commit, so missed this point. If everything will fail, on next week I'll try to get out with Windows builds and check the problem as well

clalancette:
> The latest Windows build fail seems really strange and related to some CI / cache issue:

Yeah, it looks like the network was having problems there.  I'll kick them off again.

> However, Clang builds give linking problems (is it related to the latest commit?)

Yeah, most probably.  We'll have to take a closer look at building with clang.

clalancette:
I think f54ca7d should fix it on clang.  Let's see if that is happy across the board:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19425/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13953/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20151/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20152/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/467/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/33/)

clalancette (approved):
Thanks for all of the back-and-forth here @AlexeyMerzlyakov .  With the latest changes, this looks good to me.  I'm going to go ahead and merge, along with https://github.com/ros2/system_tests/pull/516

clalancette:
Just as an FYI, I released this into Rolling yesterday and I think there was some downstream fallout from this change.  In particular, look at:

* https://build.ros2.org/view/Rbin_uJ64/job/Rbin_uJ64__mavros__ubuntu_jammy_amd64__binary/233/console
* https://build.ros2.org/view/Rbin_uJ64/job/Rbin_uJ64__topic_tools__ubuntu_jammy_amd64__binary/91/console
* https://build.ros2.org/view/Rbin_uJ64/job/Rbin_uJ64__moveit_servo__ubuntu_jammy_amd64__binary/231/console

All of those failures revolve around `rclcpp::Rate`, which makes me suspect this.  I'm going to take a look, but @AlexeyMerzlyakov it would be helpful if you could look as well.

clalancette:
OK, I found it.  See #2301 which fixes it.

Issue #2100, opened 2023-02-07
Update client API to be able to remove pending requests for ROS2 foxy


aserbremen:
## Feature request

Update client API to be able to remove pending requests for ROS2 foxy (see pull request #1734)

#### Feature description

I am trying implement a ROS2 service client call according to 
https://github.com/ros2/examples/blob/rolling/rclcpp/services/minimal_client/main.cpp which removes pending requests if `rclcpp::spin_until_future_complete(node, future) != rclcpp::FutureReturnCode::SUCCESS`. 
I realized that `Client::remove_pending_request(const FutureAndRequestId &future)` is not implemented in ROS2 foxy which I am using at the moment. The minimal client example does not remove pending requests (https://github.com/ros2/examples/blob/foxy/rclcpp/services/minimal_client/main.cpp). What are the implications of not removing the pending request? 
I am not sure how these pull requests work for older versions of ROS2 and I was wondering if the functionality will be available for ROS2 foxy in the future. Thanks

GiorMM:
In the rolling version of rclcpp/include/rclcpp/client.hpp, line 695 has a comment stating that not **calling a certain function will cause the client to use more memory for each request that does not receive a reply from the server.**

In these two examples provided, this issue doesn't actually have any impact because both examples terminate the program directly after receiving a "service call failed" message.

It appears that this issue was not considered in foxy version, but both rolling and foxy use the private variable "pending_requests_" to control business related to requests.

If you don't mind changing the ABI, you can refer to my implementation here. #2126

PR #2126, opened 2023-03-14
foxy remove_pending_request implement

GiorMM:
Add remove_pending_request function and prune_pending_requests function like rolling version

No need to change the original way of using async_send_request

This may meet the requirement of #2100

### DIFF
--- a/rclcpp/include/rclcpp/client.hpp
+++ b/rclcpp/include/rclcpp/client.hpp
@@ -314,7 +314,7 @@ public:
     callback(future);
   }
 
-  /// Send a request to the service server.
+/// Send a request to the service server.
   /**
    * This method returns a `SharedFuture` instance
    * that can be passed to Executor::spin_until_future_complete() to
@@ -360,7 +360,7 @@ public:
     >::type * = nullptr
   >
   SharedFuture
-  async_send_request(SharedRequest request, CallbackT && cb, int64_t * sequence_number_ptr)
+  async_send_request(SharedRequest request, CallbackT && cb, int64_t * sequence_number_ptr = nullptr)
   {
     std::lock_guard<std::mutex> lock(pending_requests_mutex_);
     int64_t sequence_number;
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
@GiorMM thanks for the contribution.

I did not check the implementation but this requires changing API/ABI, that is something we can put back to foxy since it breaks ROS 2 user space. (at least I believe that the current policy.)

GiorMM:
This is my first PR and I have modified it twice for violating the writing rules，that's so embarrassed. And I can't pass the cplint and uncrustify test even with the original foxy version in my local host. I have no idea why would it be

Anyway, I am sorry for this will fix it later

GiorMM:
> @GiorMM thanks for the contribution.
> 
> I did not check the implementation but this requires changing API/ABI, that is something we can put back to foxy since it breaks ROS 2 user space. (at least I believe that the current policy.)

You are correct, this change may change the ABI, but I think it is worth it.

When in high request rate and the number of invalid requests increases, it will waste a significant amount of memory. This implementation is the best solution I can think of at present to minimize the required changes.

clalancette:
> You are correct, this change may change the ABI, but I think it is worth it.

Our usual practice is to maintain ABI in a released distribution.  We can break the ABI for very serious bugs, but I'm not sure that this qualifies.  Further, we are only a couple of months away from the Foxy End-of-Life (May 2023).  So I don't think we should break ABI here.

That said, if we can come up with a way to fix this that preserves ABI, we can take it in.

rclcpp/include/rclcpp/client.hpp
```diff
@@ -314,10 +314,40 @@ class Client : public ClientBase
...
+   * }
+   * ```
+   *
+   * \param[in] request request to be send.
+   * \param[in] sequence_number pointer to sequence_number of request.
+   * \return a SharedFuture instance.
+   */
   SharedFuture
-  async_send_request(SharedRequest request)
+  async_send_request(SharedRequest request, int64_t * sequence_number = nullptr)
```
fujitatomoya, line 348: adding the parameter will change the signature of method.

fujitatomoya (commented):
@GiorMM thanks for working on this. but this still requires change in ABI, so we are not able to take it.

PR #2132, opened 2023-03-19
extract the result response before the callback is issued.

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/2130

PR #2133, opened 2023-03-20
extract the result response before the callback is issued.

fujitatomoya:
humble backport of https://github.com/ros2/rclcpp/pull/2132

### DIFF
--- a/rclcpp_action/src/client.cpp
+++ b/rclcpp_action/src/client.cpp
@@ -319,14 +319,18 @@ ClientBase::handle_result_response(
   const rmw_request_id_t & response_header,
   std::shared_ptr<void> response)
 {
-  std::lock_guard<std::mutex> guard(pimpl_->result_requests_mutex);
-  const int64_t & sequence_number = response_header.sequence_number;
-  if (pimpl_->pending_result_responses.count(sequence_number) == 0) {
-    RCLCPP_ERROR(pimpl_->logger, "unknown result response, ignoring...");
-    return;
+  ResponseCallback response_callback;
+  {
+    std::lock_guard<std::mutex> guard(pimpl_->result_requests_mutex);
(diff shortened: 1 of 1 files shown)


fujitatomoya:
@iuhilnehc-ynos requesting review on this. since i borrowed your code, let me assign as co-author.

clalancette:
I'm really not sure if we should backport this one to humble or not.  It definitely changes behavior, in that a lock that was held across the callback before no longer is.

At the very least, we should do some additional testing in Rolling/Iron before we roll this change out to Humble.

fujitatomoya:
@clalancette 

behavior change is actually correct. only thing that i can think of is the following critical case.

- before, when user callback is executing, that element is in the `pending_result_response`. (this is guaranteed)
- after, when user callback is executing, that element is NOT in the `pending_result_response`.

I think get system lock during user callback is unnecessary, and we should avoid that situation. there are some places that we do avoid this situation because that is easy to make dead lock. in this particular case, i do not see any reason to keep this mutex lock... maybe i might be missing though :sweat_drops: 

> At the very least, we should do some additional testing in Rolling/Iron before we roll this change out to Humble.

I am not sure if we can add more test for this critical section, but i understand your concern.

at least, we can wait for a couple weeks to see https://github.com/ros2/rclcpp/pull/2132 could cause any issue? and then we can come back on this to see if we can backport this or not? How does that sound to you?

i mean, i am not pushing this backport fix asap since it is already there, but it would be nicer to avoid this deadlock eventually.

fujitatomoya:
btw, this PR can be backport to foxy. (see https://github.com/ros2/rclcpp/pull/2132#discussion_r1141226271)

iuhilnehc-ynos (approved):

fujitatomoya:
Ah, i just found that i did not run the CI against this PR. I will revert this and resubmit the PR again.

PR #2135, opened 2023-03-21
Fix unnecessary allocations in executor.cpp

cwecht:
std::function will allocate on every call of take_and_do_error_handling (at least with libstdc++ as in this implementation std::function's small buffer can only fit two pointers but most of the time, the lambdas will contain 3 pointers). This impose multiple heap allocations (and deallocations) on every callback.

This comes with the cost of increased binary size. But I suppose it's worth it as this change also enables better/easier inling and avoids indirect function calls through the std::function machinery.

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -558,13 +558,14 @@ Executor::execute_any_executable(AnyExecutable & any_exec)
   }
 }
 
+template<typename Taker, typename Handler>
 static
 void
 take_and_do_error_handling(
   const char * action_description,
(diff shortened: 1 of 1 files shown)


mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18314/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12836/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18980/)

fujitatomoya (approved):
lgtm

mjcarroll:
The windows test failures are known flakes.

PR #2140, opened 2023-03-24
add events-executor and timers-manager in rclcpp

alsora:
This PR introduces the events-executor based on the design here https://github.com/ros2/design/pull/305 and the already-existing open-source implementation in https://github.com/irobot-ros/events-executor.

FYI @clalancette @wjwwood @mjcarroll @mauropasse @bpwilcox

PR #2109, opened 2023-02-21
Avoid losing waitable handles while using MultiThreadedExecutor

Barry-Xu-2018:
Address #2012

This patch may not be the best fix.   
But I hope to find the best fix by discussing this patch.

PR #2142, opened 2023-03-28
Utilize rclcpp::WaitSet as part of the executors

mjcarroll:
Improve number of `rcl_wait_set` creations and deletions by making the default Single/Multithreaded executors work like the static single threaded executor in terms of entity collection rebuilding.

TODOS

- [x] The concepts of entity collector and notify waitable come from https://github.com/ros2/rclcpp/pull/2140, they need to be rectified with that pull request and iRobot copyright needs to be added to appropriate files
- [x] restore static-single threaded files (removed for ease of testing)
- [x] documentation/test coverage on some of the new structures


rclcpp/src/rclcpp/executors/executor_entities_collection.cpp
```diff
@@ -0,0 +1,355 @@
...
+  const ExecutorEntitiesCollection::TimerCollection & other,
+  TimerUpdatedFunc timer_added,
+  TimerUpdatedFunc timer_removed)
+{
+  for (auto it = this->timers.begin(); it != this->timers.end();)
+  {
+    if (other.count(it->first) == 0) {
+      auto timer = it->second.timer.lock();
+      if (timer)
+        timer_removed(timer);
```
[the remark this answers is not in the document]
alsora: the same applies to some of the other functions below.... I also wonder if we should have this implementation templated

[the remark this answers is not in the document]
mjcarroll: Templates are a good idea.  I think I avoided initially because it was accessing struct members, but I can give it another try.

alsora:
@mjcarroll I looked into the new entities collection and collector classes and they look great.

I think that there are very few changes needed in order to use them with the events-executor:
 1. add to `ExecutorEntitiesCollection` some public methods such as 
 ```
rclcpp::ClientBase::WeakPtr get_client(const rcl_client_t * client_handle);

```
 2. the `ExecutorNotifyWaitable` in the events-executor should have an `execute` function that refresh the events-executor collections. to do so, it will need to access some internal methods of the events-executor. this could be configurable through a lambda function passed during construction.

File: rclcpp/src/rclcpp/executors/executor_entities_collector.cpp [hunk 2]
```diff
@@ -0,0 +1,281 @@
...
+    has_executor.store(false);
+    collection.erase(iter);
+    this->notify_waitable_->remove_guard_condition(group_ptr->get_notify_guard_condition().get());
+  } else {
+    throw std::runtime_error("Callback group needs to be associated with executor.");
+  }
+}
+
+void
+ExecutorEntitiesCollector::add_automatically_associated_callback_groups()
```
alsora, line 230: this function could take a list of weak nodes as input.
when we call it after `add_node` or `add_callback_group` we don't need to loop through all nodes, but only on the new one.

mjcarroll: I think that we need to check for the condition where we have an existing node and an automatically-associated callback group is added to that node, correct?

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -138,95 +123,38 @@ Executor::~Executor()
...
-  std::lock_guard<std::mutex> guard{mutex_};
-  for (const auto & group_node_ptr : weak_groups_associated_with_executor_to_nodes_) {
-    groups.push_back(group_node_ptr.first);
-  }
-  for (auto const & group_node_ptr : weak_groups_to_nodes_associated_with_executor_) {
-    groups.push_back(group_node_ptr.first);
-  }
-  return groups;
+  std::lock_guard<std::mutex> lock(mutex_);
+  return this->collector_.get_all_callback_groups();
```
alsora, line 127: Should the various `Executor::get_all_callback_groups()` methods do a `this->collector_.update_collection()` ?

otherwise the following test would fail
```
auto executor = ...
auto node = ....
executor->add_node(node);
auto groups = executor->get_all_callback_groups();
```

here I would expect the `groups` to contain what I just added in the line before, but it won't happen with the current implementation.

I added the `update_collection` calls in the events-executor: https://github.com/ros2/rclcpp/pull/2155/files#diff-65295506f5cccb52eb107cb023966a999c0cc4076ccb92197b7c23d73ca7d767R360-R379

mjcarroll: I think it makes sense to do.  In the longer term, I wonder if these `get_` methods still make sense as part of the public executor API?  I imagine since the collector is now handling all of this, these are largely implementation details that can be eventually hidden.

In my early testing, we have to be cautious about manipulating the collection out from under the waitset, but I think now that I'm making a copy of the notify waitable for the duration of the wait, this shouldn't be an issue.

alsora: Yes, I don't see any reason why users would need to access the callback groups from the executor.

mjcarroll:
I believe I am ready for a full build here: 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18504/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13051/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19234/)

mjcarroll:
Up to rclcpp:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18522/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13068/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19252/)

mjcarroll:
Full up with a patch to tf2_ros https://github.com/ros2/geometry2/pull/595

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18525/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13072/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19255/)

mjcarroll:
I'm calling it for the night, but things look good locally, so this is a present for tomorrow me: 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18537/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19267/)

mjcarroll:
Another linux: [Build Status](https://ci.ros2.org/job/ci_linux/18542/)

rclcpp/src/rclcpp/callback_group.cpp
```diff
@@ -56,6 +56,16 @@ CallbackGroup::type() const
...
 }
 
+size_t
+CallbackGroup::size() const
+{
+  return subscription_ptrs_.size() +
+         service_ptrs_.size() +
+         client_ptrs_.size() +
+         timer_ptrs_.size() +
+         waitable_ptrs_.size();
```
wjwwood, line 66: nitpick: vertical aligning indents are generally avoided, however in this case since the length of `return ` will never change so the indent shouldn't either, so the main argument for it isn't valid here. I would have still used `()` to get them on a single two space indent.

not blocking

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -38,14 +41,23 @@
...
+/// Mask to indicate to the waitset to only add the subscription.
+/// The events and intraprocess waitable are already added via the callback group.
+static constexpr rclcpp::SubscriptionWaitSetMask kDefaultSubscriptionMask = {true, false, false};
+
 Executor::Executor(const rclcpp::ExecutorOptions & options)
 : spinning(false),
   interrupt_guard_condition_(std::make_shared<rclcpp::GuardCondition>(options.context)),
   shutdown_guard_condition_(std::make_shared<rclcpp::GuardCondition>(options.context)),
-  memory_strategy_(options.memory_strategy)
+  notify_waitable_(std::make_shared<rclcpp::executors::ExecutorNotifyWaitable>(
```
wjwwood, line 54: I see this is being used with the `collector_` and the `current_notify_waitable_`, but I always like to ask if we really need this to be a shared pointer? Is it possible to avoid that? If not that's ok. I don't have enough context to say easily.

wjwwood: I suppose it's nicer than having a non-owning pointer for `current_notify_waitable_` and passing a reference to `collector_`?

wjwwood (approved):
I don't see any issues. I made a few comments, but mostly about inconsequential things. lgtm

rclcpp/src/rclcpp/executor.cpp
mjcarroll: I think at some point that the `ExecutorNotifyWaitable` could also be derived from for other executor implementations, so keeping a pointer to the base was the idea here.  It may be that we don't need that anymore as this implementation does everything we need for the waitset and events executors (cc @alsora)

mjcarroll:
Do it again:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18559/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13104/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19295/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18574/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13118/)
* RHEL [Build Status](https://ci.ros2.org/job/ci_linux-rhel/403/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19311/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19312/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13130/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19325/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19326/)

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -38,14 +41,25 @@
 
 using namespace std::chrono_literals;
 
-using rclcpp::exceptions::throw_from_rcl_error;
 using rclcpp::Executor;
+class rclcpp::ExecutorImplementation {};
```
alsora, line 45: how is this going to be used?

mjcarroll: We weren't comfortable merging the `rclcpp::WaitSet` implementation with the few outstanding test failures on Windows, this gives a little bit of latitude to backport without breaking ABI into Iron.

mjcarroll:

mjcarroll:

mjcarroll:
@alsora looks like your new test is making me fail the pr builder now :(

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18887/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13423/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19644/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19001/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13525/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19755/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19176/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13746/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19928/)

mjcarroll:
Lets do _everything_

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19239/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13793/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19977/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/19978/)
* RHEL [Build Status](https://ci.ros2.org/job/ci_linux-rhel/451/)

schornakj:
@mjcarroll @clalancette what else needs to happen before this PR can be merged in? This fixes #2163, and it would be excellent to get that resolved since it involves ROS occasionally crashing due to relatively mundane interaction (destroying a subscriber, in this case).

clalancette:
> @mjcarroll @clalancette what else needs to happen before this PR can be merged in? This fixes #2163, and it would be excellent to get that resolved since it involves ROS occasionally crashing due to relatively mundane interaction (destroying a subscriber, in this case).

There is probably a bug still in here, as every time we run CI on it we get slightly different test failures.  That bug needs to be debugged and fixed before we can merge.

mjcarroll:
More specifically, it's a bug that only appears on Windows, which adds to the pain and suffering of debugging it and I haven't had the bandwidth to revisit in depth.

uavster:
@mjcarroll @clalancette Any updates on this one? We are consistently hit by #2163.

mjcarroll:
https://github.com/ros2/rcl/pull/1112 + https://github.com/ros2/rclcpp/pull/2344 landed, which fixes a segfault with the WaitSet.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19697/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14203/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20427/)

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19784/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14286/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20508/)

clalancette:
All right, one more fix that I think should fix the failing Windows test.  Here is new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19899/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14382/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20605/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20606/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/522/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/41/)

clalancette:
All right, style fixed in the latest.  Here is another round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19908/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14387/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20615/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20616/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/524/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/42/)

wjwwood:
Just as a data point, there's a new test being added in https://github.com/ros2/rclcpp/pull/2371 (and I improved in https://github.com/cellumation/rclcpp/pull/2) which is related to https://github.com/ros2/rclcpp/issues/2242 / https://github.com/ros2/rclcpp/pull/2250, and is related to how waitables and guard conditions a re handled in the executors.

Before this pr's changes only the static single-threaded executor and the events executor was failing the test, but after this pr all of them are failing it.

The way in which they fail differs between the executors, and I will be looking into why (whether it is flaws in the executor design or in the test), but @clalancette suggested I look at this change too and I wanted to let you guys know what I found.

This is the relevant part of the test:

https://github.com/cellumation/rclcpp/blob/b4b71d93e8069c85d767199d36cde61a8deb55ee/rclcpp/test/rclcpp/executors/test_executors.cpp#L681-L737

- create two waitables in the same mutually exclusive callback group
- add callback group to the executor
- trigger both waitables
- spin until a future is set by the first waitable
- assert the first waitable was executed but not the second
- manually set the callback group's `can_be_taken_from` to false
- spin again, looking for the second waitable's future, see it never happens
- assert that the first waitable has been called but not the second one, still
- reset callback group's `can_be_taken_from` to true
- spin until future for second waitable is completed
- assert that both were called once

The events executor seems to call the callback when it shouldn't (see the second assert).

The static single-threaded executor seems to never execute the second waitable.

I haven't yet figured out how the single and multi-threaded executors fail yet with this patch.

rclcpp/include/rclcpp/wait_set_policies/detail/storage_policy_common.hpp
```diff
@@ -338,17 +333,17 @@ class StoragePolicyCommon
...
         &rcl_wait_set_,
-        client_ptr_pair.second->get_client_handle().get(),
+        client->get_client_handle().get(),
         nullptr);
       if (RCL_RET_OK != ret) {
         rclcpp::exceptions::throw_from_rcl_error(ret);
       }
     }
 
+
```
wjwwood, line 343: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -117,164 +125,122 @@ StaticSingleThreadedExecutor::spin_some_impl(std::chrono::nanoseconds max_durati
...
+    auto waitable = entry.entity.lock();
+    if (!waitable) {
+      continue;
+    }
+    if (!waitable->is_ready(&rcl_wait_set)) {
+      continue;
+    }
+
+    auto data = waitable->take_data();
+    waitable->execute(data);
```
wjwwood, line 239: Can `take_data()` fail? If so, does it throw or return `nullptr`, should anything be done here to deal with that?

mjcarroll: I don't think it can fail, but it can return `nullptr`, which we check for in `execute` (at least in the intraprocess implementation).

mjcarroll:
Big Build: 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19948/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14420/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20653/)
* Windows (Debug) [Build Status](http://ci.ros2.org/job/ci_windows/20654/)
* RHEL [Build Status](https://ci.ros2.org/job/ci_linux-rhel/538/)
* Clang [Build Status](https://ci.ros2.org/job/ci_linux_clang_libcxx/48/)

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -745,227 +593,133 @@ Executor::execute_client(
...
+  current_collection_.guard_conditions.update(
+    collection.guard_conditions,
+    [this](auto guard_condition) {wait_set_.add_guard_condition(guard_condition);},
+    [this](auto guard_condition) {wait_set_.remove_guard_condition(guard_condition);});
+
+  current_collection_.waitables.update(
+    collection.waitables,
+    [this](auto waitable) {
+      wait_set_.add_waitable(waitable);
+    },
```
wjwwood, line 654: nitpick: Why not? (like the rest)

```suggestion
    [this](auto waitable) {wait_set_.add_waitable(waitable);},
```

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -745,227 +593,133 @@ Executor::execute_client(
...
+  if (notify_waitable_) {
+    current_notify_waitable_ = std::make_shared<rclcpp::executors::ExecutorNotifyWaitable>(
+      *notify_waitable_);
+    auto notify_waitable = std::static_pointer_cast<rclcpp::Waitable>(current_notify_waitable_);
+    collection.waitables.insert({notify_waitable.get(), {notify_waitable, {}}});
+  }
+
+  // We must remove expired entities here, so that we don't continue to use older entities.
+  // See https://github.com/ros2/rclcpp/issues/2180 for more information.
+  current_collection_.remove_expired_entities();
```
wjwwood, line 617: I'm interested in this line, what prevents the same problem occurring after this line but before the next ones? Does this function also lock them all to prevent them from expiring?

wjwwood (commented):
Looks pretty good to me, mostly comments and nitpicks, but my main concern surrounds the changes to spin some.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -677,39 +564,33 @@ class Executor
...
-  /// maps callback groups to guard conditions
-  WeakCallbackGroupsToGuardConditionsMap
-  weak_groups_to_guard_conditions_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+  /// Collector used to associate executable entities from nodes and guard conditions
+  rclcpp::executors::ExecutorEntitiesCollector collector_;
 
-  /// maps callback groups associated to nodes
-  WeakCallbackGroupsToNodesMap
-  weak_groups_associated_with_executor_to_nodes_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+  /// Waitset to be waited on.
```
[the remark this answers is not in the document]
mjcarroll: Updated

rclcpp/include/rclcpp/wait_set_policies/detail/storage_policy_common.hpp
```diff
@@ -204,13 +203,13 @@ class StoragePolicyCommon
...
-        clients_from_waitables += waitable.get_number_of_ready_clients();
-        services_from_waitables += waitable.get_number_of_ready_services();
-        events_from_waitables += waitable.get_number_of_ready_events();
+        const auto & waitable = waitable_entry.waitable;
+        subscriptions_from_waitables += waitable->get_number_of_ready_subscriptions();
+        guard_conditions_from_waitables += waitable->get_number_of_ready_guard_conditions();
+        timers_from_waitables += waitable->get_number_of_ready_timers();
+        clients_from_waitables += waitable->get_number_of_ready_clients();
+        services_from_waitables += waitable->get_number_of_ready_services();
+        events_from_waitables += waitable->get_number_of_ready_events();
```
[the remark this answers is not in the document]
mjcarroll: Reverted to 1-line change.

rclcpp/src/rclcpp/any_executable.cpp
```diff
@@ -16,11 +16,13 @@
 
 using rclcpp::AnyExecutable;
 
+RCLCPP_PUBLIC
```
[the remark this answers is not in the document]
mjcarroll: Nope removed

rclcpp/src/rclcpp/executor.cpp
mjcarroll: Updated

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -99,6 +99,19 @@ MultiThreadedExecutor::run(size_t this_thread_number)
 
     execute_any_executable(any_exec);
 
+    if (any_exec.callback_group &&
+      any_exec.callback_group->type() == CallbackGroupType::MutuallyExclusive &&
+      any_exec.callback_group->size() > 1)
```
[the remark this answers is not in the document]
mjcarroll: Since this guard condition is the thing that will wake up the WaitSet in another thread, I think it only makes sense to trigger in the case that there is actually work available.

In my mind, in the case that there is a callback group with only one thing in it, then triggering the guard condition doesn't really accomplish anything other than a spurious wakeup of the waitset.

rclcpp/src/rclcpp/executor.cpp
mjcarroll: The next section (syncronizing the new collection into the waitset) also additionally re-locks all of the entity weak_ptrs into shared_pts.

This specific line was introduced because the collection/current_collection maps are using the rcl handles as the key and there is a _good chance_ that handles end up getting re-used if you allocate/deallocate them in a tight loop.  The handles will then conflict and the collections won't get updated correctly.

After talking to @clalancette, I think we may be able to minimize some of the iterations over the collections here, but as a follow-on if possible.

My longer term thought would be to include some capabilities in the waitset for determining which entities are already members and allow some of this logic to be moved into the waitset itself, which should tidy these iterations up.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -745,227 +593,133 @@ Executor::execute_client(
...
+  this->collector_.update_collections();
+  auto callback_groups = this->collector_.get_all_callback_groups();
+  rclcpp::executors::build_entities_collection(callback_groups, collection);
+
+  // Make a copy of notify waitable so we can continue to mutate the original
+  // one outside of the execute loop.
+  // This prevents the collection of guard conditions in the waitable from changing
+  // while we are waiting on it.
+  if (notify_waitable_) {
+    current_notify_waitable_ = std::make_shared<rclcpp::executors::ExecutorNotifyWaitable>(
```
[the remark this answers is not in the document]
mjcarroll: My initial thought was no, but I think it will need to be protected somehow.  We already have a mutex inside of ExecutorNotifyWaitable to protect the internals, I will likely reuse that.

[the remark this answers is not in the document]
mjcarroll: Added protection in the copy constructor of `ExecutorNotifyWaitable` in 8ed094e8

sea-bass:
I'm also running into this segfault on `rclpy` on Humble, by the way. This is when using the [`rosbridge_suite`](https://github.com/RobotWebTools/rosbridge_suite) stack to advertise/unadvertise an action server. The stack trace is:

```
[rosbridge_websocket-2] ERROR:tornado.application:Exception in callback <function main.<locals>.<lambda> at 0x7f83e0230b80>
[rosbridge_websocket-2] Traceback (most recent call last):
[rosbridge_websocket-2]   File "/usr/lib/python3/dist-packages/tornado/ioloop.py", line 905, in _run
[rosbridge_websocket-2]     return self.callback()
[rosbridge_websocket-2]   File "/opt/underlay_ws/install/rosbridge_server/lib/rosbridge_server/rosbridge_websocket", line 340, in <lambda>
[rosbridge_websocket-2]     spin_callback = PeriodicCallback(lambda: executor.spin_once(timeout_sec=0.01), 1)
[rosbridge_websocket-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 705, in spin_once
[rosbridge_websocket-2]     handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
[rosbridge_websocket-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 691, in wait_for_ready_callbacks
[rosbridge_websocket-2]     return next(self._cb_iter)
[rosbridge_websocket-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 585, in _wait_for_ready_callbacks
[rosbridge_websocket-2]     waitable.add_to_wait_set(wait_set)
[rosbridge_websocket-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/action/server.py", line 499, in add_to_wait_set
[rosbridge_websocket-2]     self._handle.add_to_waitset(wait_set)
[rosbridge_websocket-2] rclpy._rclpy_pybind11.RCLError: Failed to add 'rcl_action_server_t' to wait set: action server pointer is invalid, at ./src/rcl_action/action_server.c:929
```

rclcpp/src/rclcpp/executors/executor_entities_collection.cpp
```diff
@@ -39,6 +39,30 @@ void ExecutorEntitiesCollection::clear()
...
+{
+  auto remove_entities = [](auto & collection) -> size_t {
+      size_t removed = 0;
+      for (auto it = collection.begin(); it != collection.end(); ) {
+        if (it->second.entity.expired()) {
+          ++removed;
+          it = collection.erase(it);
+        } else {
+          ++it;
+        }
```
jmachowinski, line 52: std::remove_if ?

`collection.erase(std::remove_if(collection.begin(), 
                              collection.end(),
                              [&removed](const auto & e) { 
                                   if(e.second.entity.expired())
                                   {
                                          removed++;
                                          return true;
                                    }
                                    return false;}), `

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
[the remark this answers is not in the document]
jmachowinski: Add a comment ?

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -105,9 +101,21 @@ StaticSingleThreadedExecutor::spin_some_impl(std::chrono::nanoseconds max_durati
...
+      this->collect_entities();
+    }
+
+    auto wait_result = wait_set_.wait(std::chrono::nanoseconds(0));
+    if (wait_result.kind() == WaitResultKind::Empty) {
+      RCUTILS_LOG_WARN_NAMED(
+        "rclcpp",
+        "empty wait set received in wait(). This should never happen.");
+      continue;
+    }
```
jmachowinski, line 115: There are 3 identical copies of this code here, perhaps put it into a function ?

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -117,164 +125,122 @@ StaticSingleThreadedExecutor::spin_some_impl(std::chrono::nanoseconds max_durati
...
+      if (!entity) {
+        continue;
       }
+      execute_subscription(entity);
+      if (spin_once) {
+        return true;
+      }
+      any_ready_executable = true;
     }
   }
```
jmachowinski, line 195: This loop is almost identical throughout the function, refactor in a lambda ?

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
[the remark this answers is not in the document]
mjcarroll: After thinking about this a little more, I'm going to remove the condition because it doesn't correctly account for group type changing or group membership changing during the course of the executable entity being run.

rclcpp/src/rclcpp/executor.cpp
wjwwood: > because the collection/current_collection maps are using the rcl handles as the key and there is a good chance that handles end up getting re-used if you allocate/deallocate them in a tight loop.

If they're being used as keys then they should be held as weak or shared pointers, not just raw pointer values, that way you can know if they have expired. I believe the control block associated with a shared pointer and connected weak pointers are separate from the pointer address and will not be subject to the reuse issue you're describing.

Given that, I don't see how this line actually addresses the issue, because like I mentioned above if between this line and the next ones a rcl handle is re-used you have the same problem no? The only way to prevent that is to try and lock the rcl handle, remove it if you cannot lock it, then keep the shared pointer until your done with the next steps. This line and the reasons that led to it are all very suspect to me, unless I misunderstand them still.

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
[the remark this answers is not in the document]
wjwwood: Sounds good, I'm still kind of confused about what this condition was checking though, because a `> 0` kind of makes sense, as in "no need to notify if the guard condition is empty", but `> 1` didn't really track for me.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -475,21 +314,44 @@ Executor::spin_some_impl(std::chrono::nanoseconds max_duration, bool exhaustive)
...
-      work_available = true;
-    } else {
-      if (!work_available || !exhaustive) {
+    if (work_in_queue > 0) {
+      // If there is work in the queue, then execute it
+      // This covers the case that there are things left in the queue from a
+      // previous spin.
+      if (get_next_ready_executable(any_exec)) {
+        execute_any_executable(any_exec);
+      }
```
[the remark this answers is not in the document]
wjwwood: I would just say we should avoid this particular change in this pull request.

We can continue to discuss it in the future, but I would argue that if this is the behavior of the static single threaded executor then it is subtly broken. Perhaps that doesn't come up often in practice, but when someone runs into this behavior I think that they'd be very surprised and have trouble understanding why it is happening.

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
wjwwood: Decent suggestion, but sometimes factoring this stuff out only ends up making it more difficult to understand. I leave it up to @mjcarroll to decide what is more readable.

wjwwood:
Ok, so based on my most recent work on the "missing events" test that @jmachowinski added originally (see: https://github.com/ros2/rclcpp/pull/2371#discussion_r1441172747), I don't think it should block this pr. I think that test is somewhat flawed, or at least tickling an under-defined part of the interface and in a way that's actually not that important. Instead we should fix the relationship between waitables and executors, specifically waitables with guard conditions and how they should "stay ready" (https://github.com/ros2/rclcpp/pull/2109 vs refactoring Waitable's interface and requirements).

I still don't care for the changes to `spin_some` and how it impacts even `spin_once`, and we should address that, but I don't think the new test cases related to the "missing events"/"double take events"/"rclcpp action issues" is blocking this change. In fact, I'd much rather take the changes in this pr before trying to add tests and fix those issues (currently having to test against this pr, rolling, and events executor which is a lot).

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
mjcarroll: Refactored into `collect_and_wait`

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
mjcarroll: Okay, I refactored this into a local lambda function.  I'm not 100% on it being more readable, take a look and see what you think.

rclcpp/src/rclcpp/executors/executor_entities_collection.cpp
mjcarroll: I was wondering why I didn't use this, but then when I went to implement, I remembered that collection is `std::pair`, and it ends up failing compilation because 

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/stl_pair.h:198:17: note: copy assignment operator is implicitly deleted because 'pair<const rcl_guard_condition_s *const, rclcpp::executors::CollectionEntry<rclcpp::GuardCondition>>' has a user-declared move constructor
      constexpr pair(pair&&) = default;         ///< Move constructor
```

jmachowinski: bummer :-(

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
jmachowinski: looks good to me, can I somehow mark things as resolved ? Gitlab has this functionality, but I can't find it on github....

mjcarroll: Do you see the "resolve conversation" button.  Maybe a permissions thing.

jmachowinski: Nope, no button...

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
jmachowinski: lgtm

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -559,18 +412,13 @@ Executor::execute_any_executable(AnyExecutable & any_exec)
     execute_client(any_exec.client);
   }
   if (any_exec.waitable) {
-    any_exec.waitable->execute(any_exec.data);
+    auto data = any_exec.waitable->take_data();
```
jmachowinski, line 415: This change breaks the multi threaded executor.

take_data must be called, in a sequence of is_ready, take_data under the protection of a mutex.

mjcarroll:
> I played around with the benchmarks, and can see a drop in performance for the StaticSingleThreadedExecutor. This should be investigated further...

What about `SingleThreadedExecutor`?  I haven't run the benchmarks in a little bit, but last time I did, it was much closer to the `StaticSingleThreadedExecutor`

rclcpp/src/rclcpp/executor.cpp
mjcarroll: Hmm, I don't think that we ever state that explicitly anywhere.  Ideally, `is_ready` should be idempotent in that you can call it multiple times on the same waitset and it always returns true if the waitable is ready, false otherwise.

`take_data`, I believe, should reset the state of the waitable so that it no longer returns true to `is_ready`, and should be done immediately before executing the waitable.

Previously, I was calling `take_data` when getting the list of all of the executables out of the waitset, but this could cause a condition where `take_data` was called without ever executing on it.

jmachowinski: >  Ideally, is_ready should be idempotent in that you can call it multiple times on the same waitset and it always returns true if the waitable is ready, false otherwise.

This holds, and is not affected by where we call take_data

> take_data, I believe, should reset the state of the waitable so that it no longer returns true to is_ready, and should be done immediately before executing the waitable.

Calling directly before the execution would be fine. You just need to make sure that you are holding the mutex, that prevents someone else from calling is_ready at the same time. 

Currently you can run into this scenario:
- T1 is_ready is called, element added to list of ready executables
- T1 Mutex is released
- T2 Second thread waits, calls is_ready, and adds the element
- T1 Execution thread comes by, calls take_data, nothing ready after this
- T2 Calls take_data, but nothing is there.

jmachowinski:
> What about `SingleThreadedExecutor`? I haven't run the benchmarks in a little bit, but last time I did, it was much closer to the `StaticSingleThreadedExecutor`

In the rclcpp benchmark_executors, the SingleThreadedExecutor got faster, but the StaticSingleThreadedExecutor got a huge performance decrease.  
Note, I have extended benchmarks in my callback executor tree, should I put them in a separate MR ?

In the irobot performance benchmark latencies for the SingleThreadedExecutor got better, but the ones of the StaticSingleThreadedExecutor increased. In general the irobot benchmark seems a bit more 'all over the place' on consecutive runs.

rclcpp/src/rclcpp/executor.cpp
wjwwood: This results in a wasted call to `take_data`, but to be clear it's ok to call `take_data` when nothing is ready, it's just not optimal and so it should be avoided when possible.

There are other reasons that something can happen between `is_ready` and `take_data` which will result in `take_data` having nothing to return. For example, if a subscription is set up with the lifespan QoS then `is_ready` could return true, then the message that make it ready expires and is removed, then `take_data` is called but the underlying DDS tells it that there is nothing to take. And that scenario is totally fine, but a bit of a waste of resources, which should be avoided if possible but isn't a bug or anything (in my opinion).

The multi-threaded issue you're describing above should probably be avoided, if possible, but I don't think it breaks anything exactly.

I could see that there should be a `if data == nullptr` check that avoids calling execute if there was no data to take, avoiding a similar check in every execute implementation.

jmachowinski: > This results in a wasted call to `take_data`, but to be clear it's ok to call `take_data` when nothing is ready, it's just not optimal and so it should be avoided when possible.
> 
Doing this will crash the you executable if using actions:
https://github.com/ros2/rclcpp/blob/99f0fc938b2d0c441787113a3920c16d3614a197/rclcpp_action/src/client.cpp#L592

mjcarroll: Discussed this with @wjwwood this afternoon and realized that this change was actually an artifact of the "ready executables queue" that was added.  I'm now removing that so this change will be un-done.

mjcarroll:
Merged from master, all tests passing locally, so running through CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20271/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14686/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21001/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20279/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14691/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21009/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20288/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14699/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21018/)

wjwwood (approved):
lgtm

only small nitpicks

mjcarroll:
@jmachowinski I'm chasing down one or two remaining test flakes here.  Do you think you could run benchmarks again when you get a moment to see where we stand?

I did see a potential change in the `StaticSingleThreadedExecutor` implementation that would have an impact on latency that I have reverted.

jmachowinski:
> @jmachowinski I'm chasing down one or two remaining test flakes here. Do you think you could run benchmarks again when you get a moment to see where we stand?

I opened a PR with the benchmarks,  here are the results of the latest versions
Rolling:
```
PerformanceTestExecutor/single_thread_executor_spin_some                                        357087 ns       357166 ns         1971
PerformanceTestExecutor/static_single_thread_executor_spin_some                                 125202 ns       125281 ns         5557
PerformanceTestExecutor/multi_thread_executor_spin_some                                         355054 ns       355125 ns         1972
PerformanceTestExecutor/single_thread_executor_wait_for_work                                     53089 ns        53046 ns        10000
PerformanceTestExecutor/multi_thread_executor_wait_for_work                                      48950 ns        48913 ns        13867
PerformanceTestExecutor/single_thread_executor_wait_for_work_rebuild                             58704 ns        58394 ns        12457
PerformanceTestExecutor/multi_thread_executor_wait_for_work_rebuild                              57292 ns        56982 ns        12222
CascadedPerformanceTestExecutor/single_thread_executor_spin                                     521077 ns         3258 ns        10000
CascadedPerformanceTestExecutor/static_single_thread_executor_spin                              336623 ns         3311 ns        10000
CascadedPerformanceTestExecutor/multi_thread_executor_spin                                      558918 ns         4023 ns        10000
PerformanceTestExecutorMultipleCallbackGroups/single_thread_executor_spin_some                  583123 ns       583151 ns         1149
PerformanceTestExecutorMultipleCallbackGroups/static_single_thread_executor_spin_some           123904 ns       123933 ns         5679
PerformanceTestExecutorMultipleCallbackGroups/multi_thread_executor_spin_some                   582915 ns       582636 ns         1173

``` 

This PR:
```
PerformanceTestExecutor/single_thread_executor_spin_some                                        155469 ns       155540 ns         4497
PerformanceTestExecutor/static_single_thread_executor_spin_some                                 138196 ns       138283 ns         5008
PerformanceTestExecutor/multi_thread_executor_spin_some                                         155705 ns       155782 ns         4451
PerformanceTestExecutor/single_thread_executor_wait_for_work                                     36104 ns        36125 ns        19595
PerformanceTestExecutor/multi_thread_executor_wait_for_work                                      37539 ns        37582 ns        18369
PerformanceTestExecutor/single_thread_executor_wait_for_work_rebuild                            153240 ns       153030 ns         4675
PerformanceTestExecutor/multi_thread_executor_wait_for_work_rebuild                             149238 ns       149019 ns         4630
CascadedPerformanceTestExecutor/single_thread_executor_spin                                     434639 ns         4315 ns        10000
CascadedPerformanceTestExecutor/static_single_thread_executor_spin                              435907 ns         4081 ns        10000
CascadedPerformanceTestExecutor/multi_thread_executor_spin                                     1880370 ns         4560 ns        10000
PerformanceTestExecutorMultipleCallbackGroups/single_thread_executor_spin_some                  159868 ns       159918 ns         4361
PerformanceTestExecutorMultipleCallbackGroups/static_single_thread_executor_spin_some           143552 ns       143645 ns         4850
PerformanceTestExecutorMultipleCallbackGroups/multi_thread_executor_spin_some                   161186 ns       161255 ns         4395

``` 

The result of CascadedPerformanceTestExecutor/multi_thread_executor_spin is concerning, as this one is closest to a real world scenario.

wjwwood:
@jmachowinski do you know if you ran those benchmarks in release or debug?

I've been running the same benchmarks (from your pr https://github.com/ros2/rclcpp/pull/2430) and I've seen a lot of variation run-to-run on the `CascadedPerformanceTestExecutor/multi_thread_executor_spin` bench.

I ran it many times in loops and did so with both release and debug, and I see that this particular benchmark is roughly 5-10% slower with this pr (based on the cpu time). I'm not sure if that constitutes a blocking issue or not. While this may be the most "real world-like" benchmark, it's still a micro benchmark, so I'm not sure how big of an issue this will be in practice.

----

I also looked at it in a profiler, and it's kind of difficult to compare them, because the call graph has changed so much, but I don't see any obvious issues with the new branch. Most of the time is spent in look ups in the an unordered map, which is where we're doing work to "collect entities" in the callback groups, which is a necessary step. However, there might be room for optimization there.

----

That observation, and a few others, leads me to conclude this new branch is more efficient at some parts of the work and less efficient at others, and that makes sense because the approach is slightly different, but I didn't see anything that stood out as an obvious issue (like spending lots of time doing something that doesn't seem necessary).

Either way, I'll do a bit more investigation to see if there's an "obvious win" to bring this particular benchmark closer to rolling's performance.

jmachowinski:
>do you know if you ran those benchmarks in release or debug?

Pretty sure it was release, I just checked and rerun the benchmark, it was release

> I've been running the same benchmarks (from your pr https://github.com/ros2/rclcpp/pull/2430) and I've seen a lot of variation run-to-run on the CascadedPerformanceTestExecutor/multi_thread_executor_spin bench.

Did you disable cpu frequency scaling ?
`cpupower frequency-set -g performance`

Runs are pretty consistent on my machine. (https://pastebin.com/9hs04y1M)

> While this may be the most "real world-like" benchmark, it's still a micro benchmark, so I'm not sure how big of an issue this will be in practice.

I agree in general that one should look at micro benchmarks with care. However, we are in the code path, that will be run with multiple kHz. From testing with our stack, and the CBGExecutor I saw CPU drops of 8-10% for our most loaded node, which is a memory optimized version of rosbag2 in snapshot mode. Therefore I can confidently say, that optimizing this code path has an impact.

I am wondering, don't you have a reference system of your own, that you use for testing ?

> Most of the time is spent in look ups in the an unordered map, which is where we're doing work to "collect entities" in the callback groups, which is a necessary step.

I saw this too, while working on a new executor approach. You might want to have a look at https://github.com/ros2/rclcpp/pull/2405 In this approach, I replaced the maps with an O(1) lookup, that resulted in a significant improvement.

>That observation, and a few others, leads me to conclude this new branch is more efficient at some parts of the work and less efficient at others, and that makes sense because the approach is slightly different, but I didn't see anything that stood out as an obvious issue

Most looked fine and show improvements for the single threaded case. The two benchmarks that stand out are
Rolling HEAD
```
PerformanceTestExecutor/single_thread_executor_wait_for_work_rebuild                             68867 ns        68442 ns        10129
``` 
This Branch
```
PerformanceTestExecutor/single_thread_executor_wait_for_work_rebuild                            184134 ns       183815 ns         3820
``` 

I think it is fine to ignore this one, it just benchmarks the cases, that the collection needs to be rebuild. As this should only happen if you add / remove an entitiy, this will not happen often in a real load scenario.

Rolling HEAD
```
PerformanceTestExecutorMultipleCallbackGroups/multi_thread_executor_spin_some                   828275 ns       828338 ns          807
``` 
This Branch
```
CascadedPerformanceTestExecutor/multi_thread_executor_spin                                     2236833 ns         2749 ns        10000
``` 
This one is still concerning to me, this is a times 3 regression in the multithreaded spinning case. And I am confused by it. 
Wait for work (without rebuild) got faster, so this one should get faster as well, in fact we can see this in the single threaded case.

@mjcarroll Might there be a bug, that you rebuild the collection in the multithreaded case every time ?

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -99,6 +99,18 @@ MultiThreadedExecutor::run(size_t this_thread_number)
 
     execute_any_executable(any_exec);
 
+    if (any_exec.callback_group &&
+      any_exec.callback_group->type() == CallbackGroupType::MutuallyExclusive)
+    {
+      try {
+        interrupt_guard_condition_->trigger();
```
jmachowinski, line 106: @mjcarroll To answer my own question, this should be the culprit of our performance regression. This will trigger a full rebuild of the collection.

jmachowinski: > Yes, but in this case, my understanding is that it is the correct thing to do.

Yes and no. You don't need to recollect all entities, this is super expensive. You only need to make sure, that you don't add entities that belong to a blocked callback group to the rcl_wait_set_t. 

E.g. you should have a rclcpp::WaitSet per callback group, and be able to add them sequentially

> Previously, this was done in the Executor::execute_any_executable method unconditionally 

Yep, but previously this operation also was 3 times faster :-(

mjcarroll: > You only need to make sure, that you don't add entities that belong to a blocked callback group to the rcl_wait_set_t.

But you also have to remove them from the waitset.  If, for instance, imagine you had a mutually exclusive group with two threads.  If you had a timer with an extremely long task (multiples of times longer than the period), but didn't remove other masked entities from the waitset, then the waitset would continuously "wake" but without any work to do.

This is a bit of a contrived example (hopefully?) but in the current implementation, any entity that isn't eligible to execute should also not be waited on.

jmachowinski: > But you also have to remove them from the waitset.

Now I am confused, the rcl_wait_set_t or the rclcpp::Waitset ?

>  If, for instance, imagine you had a mutually exclusive group with two threads. If you had a timer with an extremely long task (multiples of times longer than the period), but didn't remove other masked entities from the waitset, then the waitset would continuously "wake" but without any work to do.

You seem to have misunderstood what I meant, or in my mind stuff is different structured ;-)
In my mind a rclcpp::Waitset is a collection of entities, that can be waited on. In the moment, that you actually want to perform the wait, you need to build an rcl_wait_set_t from it.

This is the point, were you can decide, which entities you actually want to add to the rcl_wait_set_t and really wait on. This would be the point were you don't add entities, that are in a blocked callback group.

However, if you go down that road, you will hit the problem, that your backmapping from rcl_wait_set_t to rclcpp::Waitset is not broken, as there is no one to one relationship between the entities in the rclcpp::Waitset and rcl_wait_set_t any more. I ran into the same problem in the CBG Executor. As a solution, I added a data ptr to the entities in rcl, and set them before the wait to a variant containing a reference to my rclcpp entitiy.

mjcarroll: > In my mind a rclcpp::Waitset is a collection of entities, that can be waited on. In the moment, that you actually want to perform the wait, you need to build an rcl_wait_set_t from it.

Almost.  Since building the `rcl_wait_set_t` is somewhat expensive (as we have discussed), it is only built from the `rclcpp::WaitSet` in the case that the to-be-waited-on entities change.  In the case that you had each entity in a separate callback group or all entities in `Reentrant` callback groups, this would make it such that the `rcl_wait_set_t` structure would never have to be rebuilt.

> This is the point, were you can decide, which entities you actually want to add to the rcl_wait_set_t and really wait on. This would be the point were you don't add entities, that are in a blocked callback group.

That's _sort of_ what happens, except that it's done by adding/removing entities from the rclcpp::WaitSet.  The rclcpp::WaitSet has no concept of callback groups.

> However, if you go down that road, you will hit the problem, 

Every road has it's own set of problems here :).  The bookkeeping for the rclcpp::WaitSet does currently seem to be a fairly expensive part, maybe I can work at reducing that some for better performance.

mjcarroll: Sorry, you are correct.  It prevents the `rcl_wait_set_t` from being resized, but it does have to be re-initialized before each wait.

wjwwood:
> Did you disable cpu frequency scaling ?
> cpupower frequency-set -g performance

That's not a thing on macos, but I did the equivalent. I'll have to see if there's something else that's the issue.

> This one is still concerning to me, this is a times 3 regression in the multithreaded spinning case. And I am confused by it.
Wait for work (without rebuild) got faster, so this one should get faster as well, in fact we can see this in the single threaded case.

Ok, I need to better understand why the total time is regressed so much, while the CPU time is not nearly as bad (that's what I was referring to with the 5-10% regression before).

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
wjwwood: > The rclcpp::WaitSet has no concept of callback groups.

I think this might be the most important point @mjcarroll made. It's up to the executor to manipulate the wait set as needed to control what is waited on according to constraints by the callback groups.

In the past, however, I had suggested a new concept for the wait set (at the C++ level and all the way down through rcl and rmw too) which I was calling a mask. That way you can have a wait set with a few things in it and then periodically you can wait on it with a mask to avoid waiting on a few of them (temporarily). That would allow for much better optimizations at the rmw layer, where this is most expensive. Resetting the contents of the rcl wait set is not that expensive, it's just lists of pointers. The expensive part is usually at the DDS layer where there is a somewhat complex activity of "attaching" and "detaching" conditions to their wait set object.

The other expensive part (which we may be able to avoid) is (re)building the collection of rclcpp things. We may still need to build or modify an existing wait set from the collection, but maybe we could save some time there. We still have to handle the asynchronous destruction of entities though, which is something we have to check about the collection each time we use it, which is part of the expensive operations (essentially locking the weak pointers to make sure they're still valid). But that is just speculation on my part, more likely we have to keep doing what we're doing, but make it more efficient instead.

wjwwood: To be clear, the mask idea is out-of-scope for this pr, but I think it strikes at the root of the issue. I still want to look at why the multi-threaded is so much slower, but after a cursory look, I don't think this line you commented on can be avoided or changed safely (maybe @mjcarroll will disagree).

jmachowinski: > And by the way, changing the wait set never happens with single threaded because a callback group's entities never have to be removed from the wait set since it's either executing or waiting, but never both. Which obviously isn't true for the multi-threaded executors.

I am fully aware of this, this is my main point the whole time. This also has an impact on the execution order of entities.

jmachowinski: > > The rclcpp::WaitSet has no concept of callback groups.
> 
> I think this might be the most important point @mjcarroll made. It's up to the executor to manipulate the wait set as needed to control what is waited on according to constraints by the callback groups.

Let's rephrase this. The rclcpp::WaitSet is the central piece of the Executor. Currently I would say, that neither the  rclcpp::Executor nor the rclcpp::WaitSet have a real concept of the callback groups. 
There is no API, to add or remove CallbackGroups in a efficient manor from the next wait. There is no way of getting a next ready executable, without iterating over X entries, that belong to a blocked callback group.

IMHO the API of rclcpp::WaitSet and rclcpp::WaitResult classes should reflect that callback groups exist, and should provide efficient ways to handle blocked callback groups. Without this, you will run into the same problems in the Multi-threaded case again and again.

And from my point of view this is a big design flaw of this PR.

wjwwood: > that neither the rclcpp::Executor nor the rclcpp::WaitSet have a real concept of the callback groups.

The executors definitely do.

> IMHO the API of rclcpp::WaitSet and rclcpp::WaitResult classes should reflect that callback groups exist, and should provide efficient ways to handle blocked callback groups. Without this, you will run into the same problems in the Multi-threaded case again and again.

I have to very much disagree with this. The wait set should not have any connection to callback groups in my opinion. It's a building block that should be useful even without callbacks being used at all, if the user wants to do that. In fact that was the use case for which it was originally designed (we had users that wanted to handle their own execution and did not want to use the executor pattern).

ros-discourse:

https://discourse.ros.org/t/revival-of-client-library-working-group/36406/15

wjwwood:
@mjcarroll I added [5769359](https://github.com/ros2/rclcpp/pull/2142/commits/57693595edd9a99a795d8f55f3c0d728817ac648) which contains the changes we discussed last week. I need to retest the multi-threaded executor performance still, but I think the tests are working. I need to check for flaky tests.

rclcpp/include/rclcpp/wait_result.hpp
```diff
@@ -138,33 +142,80 @@ class WaitResult final
...
+   *
+   * \param[in] start_index index at which to start searching for the next ready
+   *   timer in the wait result. If the start_index is out of bounds for the
+   *   list of timers in the wait result, then {nullptr, start_index} will be
+   *   returned. Defaults to 0.
+   * \return next ready timer pointer and its index in the wait result, or
+   *   {nullptr, start_index} if none was found.
+   */
+  std::pair<std::shared_ptr<rclcpp::TimerBase>, size_t>
+  peak_next_ready_timer(size_t start_index = 0)
```
mjcarroll, line 168: Nit: this should be `peek` not `peak`.

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
```diff
@@ -204,18 +204,19 @@ TYPED_TEST(TestAddCallbackGroupsToExecutor, add_callback_groups_after_add_node_t
...
-        ASSERT_EQ(executor.get_all_callback_groups().size(), 3u);
+        ASSERT_GT(executor.get_all_callback_groups().size(), 1u);
         executor.cancel();
       }
       timer_count++;
     };
   rclcpp::CallbackGroup::SharedPtr cb_grp = node->create_callback_group(
     rclcpp::CallbackGroupType::MutuallyExclusive);
   rclcpp::TimerBase::SharedPtr timer_ = node->create_wall_timer(
-    2s, timer_callback, cb_grp);
```
mjcarroll, line 218: Is this change needed?

rclcpp/test/rclcpp/executors/test_executors.cpp
```diff
@@ -461,16 +470,16 @@ TYPED_TEST(TestExecutors, spinSome)
...
 
   // Do some work until sufficient calls to the waitable occur, but keep going until either
   // count becomes too large, spin exits, or the 1 second timeout completes.
   auto start = std::chrono::steady_clock::now();
   while (
     my_waitable->get_count() <= 1 &&
-    !spin_exited &&
-    (std::chrono::steady_clock::now() - start < 1s))
+    !spin_exited.load() &&
+    std::chrono::steady_clock::now() - start < 10s)
```
mjcarroll, line 482: Is this change needed?

rclcpp/test/rclcpp/test_add_callback_groups_to_executor.cpp
wjwwood: No this one is not necessary, though I believe it shortens the runtime of the test. Let me have another look at it. I'll try to avoid unnecessary changes.

But I might make another pr for this, and I may even shorten it further.

I think further changes to this test could also be made, since it really doesn't need the extra timers (adding more timers/cbgs after spin, which is the purpose of the test, isn't dependent on adding more than one, and frankly could hide an issue where only one added cbg does not work).

rclcpp/test/rclcpp/executors/test_executors.cpp
wjwwood: I don't think so, I can revert it, however, this timeout need not be small, because it's not the purpose of this test to ensure this part happens within a second. If it doesn't it could imply an issue, but again, I don't think that's the point of this test and also it's a fragile way to test it.

rclcpp/include/rclcpp/wait_result.hpp
wjwwood: Oops, haha. Good catch.

wjwwood:
As part of this, I made a new pr to improve the `spin_some` tests: https://github.com/ros2/rclcpp/pull/2460

rclcpp/test/rclcpp/executors/test_executors.cpp
wjwwood: This was addressed in https://github.com/ros2/rclcpp/pull/2460.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20589/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14950/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21313/)

wjwwood:
@mjcarroll can you have a look at [836946e](https://github.com/ros2/rclcpp/pull/2142/commits/836946e23f2f1c5b6e71fc985b98c279b1c0d9a0) and [2a88295](https://github.com/ros2/rclcpp/pull/2142/commits/2a88295fbac788f90cdf955608f5a00c5ef774e7)?

new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20601/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14957/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21322/)

rclcpp/include/rclcpp/wait_result.hpp
```diff
@@ -276,11 +276,11 @@ class WaitResult final
       auto & wait_set = this->get_wait_set();
       auto rcl_wait_set = wait_set.get_rcl_wait_set();
       while (next_waitable_index_ < wait_set.size_of_waitables()) {
-        auto cur_waitable = wait_set.waitables(next_waitable_index_);
+        auto cur_waitable = wait_set.waitables(next_waitable_index_++);
```
jmachowinski, line 279: I personally consider these constructs bad style. I always need to think twice, if the returned value of x++ is the pre or post incremented value, and find code written this way therefore hard to read.

wjwwood:
I ran the test that failed on aarch64 10000 times on my laptop over night, and got no failures with `/home/william/ros2_ws/build/rclcpp/test/rclcpp/test_executors_timer_cancel_behavior --gtest_filter=*MultiThreadedExecutor.testBackTimerCancelThenResetBehavior --gtest_repeat=10000 --gtest_break_on_failure`:

```
Repeating all tests (iteration 10000) . . .

Note: Google Test filter = *MultiThreadedExecutor.testBackTimerCancelThenResetBehavior
[==========] Running 1 test from 1 test suite.
[----------] 1 test from TestTimerCancelBehavior/MultiThreadedExecutor, where TypeParam = rclcpp::executors::MultiThreadedExecutor
[ RUN      ] TestTimerCancelBehavior/MultiThreadedExecutor.testBackTimerCancelThenResetBehavior
[       OK ] TestTimerCancelBehavior/MultiThreadedExecutor.testBackTimerCancelThenResetBehavior (1867 ms)
[----------] 1 test from TestTimerCancelBehavior/MultiThreadedExecutor (1867 ms total)

[==========] 1 test from 1 test suite ran. (1875 ms total)
[  PASSED  ] 1 test.
```

I ran a few CI as well and never got it to fail: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/14972/)

So I am going to say it's extremely flaky and that we should merge this pr. Separately I will look at refactoring that test to make it more reliable. But I'd like to hear from others. @mjcarroll @clalancette

rclcpp/include/rclcpp/wait_result.hpp
wjwwood: Ok, I can rework that to not use the postfix operator, but it was the smallest change I could make to fix the problem (not iterate after `break`). I may do that in a small follow pr though if we decide to merge this, because I want to get some soak time on this change before the release.

mjcarroll:
> @mjcarroll can you have a look at [836946e](https://github.com/ros2/rclcpp/pull/2142/commits/836946e23f2f1c5b6e71fc985b98c279b1c0d9a0) and [2a88295](https://github.com/ros2/rclcpp/pull/2142/commits/2a88295fbac788f90cdf955608f5a00c5ef774e7)?

LGTM.

I also ran locally on a many core x86_64 machine with no failures.  I don't know what the odds are of this flake, but I think I'm comfortable merging here to get more soak time.

rclcpp/include/rclcpp/wait_result.hpp
wjwwood: I will do this in a follow up so we can merge.

christophfroehlich:
Thanks for pinging this here. It is already failing on main, I think this change here was released already.

dhood:
ok - although I have a new thought - maybe now the _intended_ behaviour is being enforced! The exception "subscription already associated with a wait set" might mean that it was _never_ appropriate to have it in two wait sets :sweat_smile:

fujitatomoya:
> it might not be possible to have a subscription in two wait sets after the changes in this PR?

No, we cannot add the subscription to different wait set after this change. there is a test case to expect the exception, https://github.com/ros2/rclcpp/blob/de666d2bf41806d150744d7c5dcd7a07ad3ffcd0/rclcpp/test/rclcpp/node_interfaces/test_node_topics.cpp#L165-L167

CC: @wjwwood @mjcarroll

wjwwood:
> maybe now the intended behaviour is being enforced! The exception "subscription already associated with a wait set" might mean that it was never appropriate to have it in two wait sets 😅

Sorry, just catching up on this, yes indeed that is the case. Technically the constraint is that you cannot _wait_ on one subscription from two different wait sets simultaneously, but we enforce that by not allowing a single subscription to be contained in/associated with more than one wait set at a time.

wjwwood:
> they're [manually](https://github.com/ros-controls/ros2_controllers/blob/9e97c00496c239bf4961dc043a3f461aae9b23b8/velocity_controllers/test/test_joint_group_velocity_controller.cpp#L35) building a waitset around a subscription, which maybe is a bit of a niche case, but wanted to flag it for triaging before the next sync is planned

If you want to use a subscription with a wait set manually (and that is totally ok to do), then you need to make sure the subscription is not part of a callback group that was added to an executor. You can do that by creating a callback group that has the `automatically_add_to_executor_with_node` flag set to false (https://github.com/ros2/rclcpp/blob/de666d2bf41806d150744d7c5dcd7a07ad3ffcd0/rclcpp/include/rclcpp/node.hpp#L151) and use it when creating the subscription. That will ensure that callback group is never added to an executor (and therefore a wait set) when the node is added.

dhood:
oo hadn't noticed that before; awesome! thanks for the pointers!

PR #2143, opened 2023-03-29
Create common structures for executors to use

mjcarroll:
Broken off of #2142 so that it can be used as a base for #2140.

This introduces a few main structures:

* ExecutorNotifyWaitable - an object used to collect node and callback group guard conditions from things associated with an executor.
* ExecutorEntitiesCollector - an object to collect entities from executor-associated nodes and callback groups
* ExecutorEntitiesCollection - a collection of entities that are associated with an executor via callback groups and nodes.

CC: @alsora @mauropasse

### DIFF
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -180,18 +180,6 @@ public:
     return _find_ptrs_if_impl<rclcpp::Waitable, Function>(func, waitable_ptrs_);
   }
 
-  /// Return if the node that created this callback group still exists
-  /**
-   * As nodes can share ownership of callback groups with an executor, this
-   * may be used to ensure that the executor doesn't operate on a callback
-   * group that has outlived it's creating node.
-   *
-   * \return true if the creating node still exists, otherwise false
-   */
-  RCLCPP_PUBLIC
-  bool
-  has_valid_node() const;
-
   RCLCPP_PUBLIC
   std::atomic_bool &
   can_be_taken_from();
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -89,6 +89,8 @@ public:
    * added to the executor in either case.
    *
    * \param[in] group_type The type of the callback group.
+   * \param[in] get_node_context Lambda to retrieve the node context when
+   *   checking that the creating node is valid and using the guard condition.
    * \param[in] automatically_add_to_executor_with_node A boolean that
    *   determines whether a callback group is automatically added to an executor
    *   with the node with which it is associated.
@@ -96,6 +98,7 @@ public:
   RCLCPP_PUBLIC
   explicit CallbackGroup(
     CallbackGroupType group_type,
+    std::function<rclcpp::Context::SharedPtr(void)> get_node_context,
     bool automatically_add_to_executor_with_node = true);
 
   /// Default destructor.
@@ -136,6 +139,18 @@ public:
   {
     return _find_ptrs_if_impl<rclcpp::Waitable, Function>(func, waitable_ptrs_);
   }
+
+  /// Return if the node that created this callback group still exists
+  /**
+   * As nodes can share ownership of callback groups with an executor, this
+   * may be used to ensures that the executor doesn't operate on a callback
+   * group that has outlived it's creating node.
+   *
+   * \return true if the creating node still exists, otherwise false
+   */
+  RCLCPP_PUBLIC
+  bool
+  has_valid_node();
 
   RCLCPP_PUBLIC
   std::atomic_bool &
@@ -178,10 +193,23 @@ public:
   bool
   automatically_add_to_executor_with_node() const;
 
-  /// Defer creating the notify guard condition and return it.
+  /// Retrieve the guard condition used to signal changes to this callback group.
+  /**
+   * \param[in] context_ptr context to use when creating the guard condition
+   * \return guard condition if it is valid, otherwise nullptr.
+   */
+  [[deprecated("Use get_notify_guard_condition() without arguments")]]
   RCLCPP_PUBLIC
   rclcpp::GuardCondition::SharedPtr
   get_notify_guard_condition(const rclcpp::Context::SharedPtr context_ptr);
+
+  /// Retrieve the guard condition used to signal changes to this callback group.
+  /**
+   * \return guard condition if it is valid, otherwise nullptr.
+   */
+  RCLCPP_PUBLIC
+  rclcpp::GuardCondition::SharedPtr
+  get_notify_guard_condition();
 
   /// Trigger the notify guard condition.
   RCLCPP_PUBLIC
@@ -233,6 +261,8 @@ protected:
   // defer the creation of the guard condition
   std::shared_ptr<rclcpp::GuardCondition> notify_guard_condition_ = nullptr;
   std::recursive_mutex notify_guard_condition_mutex_;
+
+  std::function<rclcpp::Context::SharedPtr(void)> get_context_;
 
 private:
   template<typename TypeT, typename Function>
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -89,6 +89,46 @@ public:
    * added to the executor in either case.
    *
    * \param[in] group_type The type of the callback group.
+   * \param[in] automatically_add_to_executor_with_node A boolean that
+   *   determines whether a callback group is automatically added to an executor
+   *   with the node with which it is associated.
+   */
+  [[deprecated("Use CallbackGroup constructor with context function argument")]]
+  RCLCPP_PUBLIC
+  explicit CallbackGroup(
+    CallbackGroupType group_type,
+    bool automatically_add_to_executor_with_node = true);
+
+  /// Constructor for CallbackGroup.
+  /**
+   * Callback Groups have a type, either 'Mutually Exclusive' or 'Reentrant'
+   * and when creating one the type must be specified.
+   *
+   * Callbacks in Reentrant Callback Groups must be able to:
+   *   - run at the same time as themselves (reentrant)
+   *   - run at the same time as other callbacks in their group
+   *   - run at the same time as other callbacks in other groups
+   *
+   * Callbacks in Mutually Exclusive Callback Groups:
+   *   - will not be run multiple times simultaneously (non-reentrant)
+   *   - will not be run at the same time as other callbacks in their group
+   *   - but must run at the same time as callbacks in other groups
+   *
+   * Additionally, callback groups have a property which determines whether or
+   * not they are added to an executor with their associated node automatically.
+   * When creating a callback group the automatically_add_to_executor_with_node
+   * argument determines this behavior, and if true it will cause the newly
+   * created callback group to be added to an executor with the node when the
+   * Executor::add_node method is used.
+   * If false, this callback group will not be added automatically and would
+   * have to be added to an executor manually using the
+   * Executor::add_callback_group method.
+   *
+   * Whether the node was added to the executor before creating the callback
+   * group, or after, is irrelevant; the callback group will be automatically
+   * added to the executor in either case.
+   *
+   * \param[in] group_type The type of the callback group.
    * \param[in] get_node_context Lambda to retrieve the node context when
    *   checking that the creating node is valid and using the guard condition.
    * \param[in] automatically_add_to_executor_with_node A boolean that
@@ -143,7 +183,7 @@ public:
   /// Return if the node that created this callback group still exists
   /**
    * As nodes can share ownership of callback groups with an executor, this
-   * may be used to ensures that the executor doesn't operate on a callback
+   * may be used to ensure that the executor doesn't operate on a callback
    * group that has outlived it's creating node.
    *
    * \return true if the creating node still exists, otherwise false
--- a/rclcpp/include/rclcpp/callback_group.hpp
+++ b/rclcpp/include/rclcpp/callback_group.hpp
@@ -190,7 +190,7 @@ public:
    */
   RCLCPP_PUBLIC
   bool
-  has_valid_node();
+  has_valid_node() const;
 
   RCLCPP_PUBLIC
   std::atomic_bool &
--- a/rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
+++ b/rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
@@ -1,0 +1,137 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
+#define RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
+
+#include <deque>
+#include <unordered_map>
+#include <vector>
+
+#include "rcpputils/thread_safety_annotations.hpp"
+
+#include <rclcpp/any_executable.hpp>
+#include <rclcpp/node_interfaces/node_base.hpp>
+#include <rclcpp/callback_group.hpp>
+#include <rclcpp/executors/executor_notify_waitable.hpp>
+#include <rclcpp/visibility_control.hpp>
+#include <rclcpp/wait_result.hpp>
+#include <rclcpp/wait_set.hpp>
+
+namespace rclcpp
+{
+namespace executors
+{
+
+struct ExecutorEntitiesCollection
+{
+  struct TimerEntry
+  {
+    rclcpp::TimerBase::WeakPtr timer;
+    rclcpp::CallbackGroup::WeakPtr callback_group;
+  };
+  using TimerCollection = std::unordered_map<const rcl_timer_t *, TimerEntry>;
+
+  struct SubscriptionEntry
+  {
+    rclcpp::SubscriptionBase::WeakPtr subscription;
+    rclcpp::CallbackGroup::WeakPtr callback_group;
+  };
+  using SubscriptionCollection = std::unordered_map<const rcl_subscription_t *, SubscriptionEntry>;
+
+  struct ClientEntry
+  {
+    rclcpp::ClientBase::WeakPtr client;
+    rclcpp::CallbackGroup::WeakPtr callback_group;
+  };
+  using ClientCollection = std::unordered_map<const rcl_client_t *, ClientEntry>;
+
+  struct ServiceEntry
+  {
+    rclcpp::ServiceBase::WeakPtr service;
+    rclcpp::CallbackGroup::WeakPtr callback_group;
+  };
+  using ServiceCollection = std::unordered_map<const rcl_service_t *, ServiceEntry>;
+
+  struct WaitableEntry
+  {
+    rclcpp::Waitable::WeakPtr waitable;
+    rclcpp::CallbackGroup::WeakPtr callback_group;
+  };
+  using WaitableCollection = std::unordered_map<const rclcpp::Waitable *, WaitableEntry>;
+
+  using GuardConditionCollection = std::unordered_map<const rcl_guard_condition_t *,
+      rclcpp::GuardCondition::WeakPtr>;
+
+  TimerCollection timers;
+  SubscriptionCollection subscriptions;
+  ClientCollection clients;
+  ServiceCollection services;
+  GuardConditionCollection guard_conditions;
+  WaitableCollection waitables;
+
+  void clear();
+
+  using TimerUpdatedFunc = std::function<void (rclcpp::TimerBase::SharedPtr)>;
+  void update_timers(
+    const ExecutorEntitiesCollection::TimerCollection & other,
+    TimerUpdatedFunc timer_added,
+    TimerUpdatedFunc timer_removed);
+
+  using SubscriptionUpdatedFunc = std::function<void (rclcpp::SubscriptionBase::SharedPtr)>;
+  void update_subscriptions(
+    const ExecutorEntitiesCollection::SubscriptionCollection & other,
+    SubscriptionUpdatedFunc subscription_added,
+    SubscriptionUpdatedFunc subscription_removed);
+
+  using ClientUpdatedFunc = std::function<void (rclcpp::ClientBase::SharedPtr)>;
+  void update_clients(
+    const ExecutorEntitiesCollection::ClientCollection & other,
+    ClientUpdatedFunc client_added,
+    ClientUpdatedFunc client_removed);
+
+  using ServiceUpdatedFunc = std::function<void (rclcpp::ServiceBase::SharedPtr)>;
+  void update_services(
+    const ExecutorEntitiesCollection::ServiceCollection & other,
+    ServiceUpdatedFunc service_added,
+    ServiceUpdatedFunc service_removed);
+
+  using GuardConditionUpdatedFunc = std::function<void (rclcpp::GuardCondition::SharedPtr)>;
+  void update_guard_conditions(
+    const ExecutorEntitiesCollection::GuardConditionCollection & other,
+    GuardConditionUpdatedFunc guard_condition_added,
+    GuardConditionUpdatedFunc guard_condition_removed);
+
+  using WaitableUpdatedFunc = std::function<void (rclcpp::Waitable::SharedPtr)>;
+  void update_waitables(
+    const ExecutorEntitiesCollection::WaitableCollection & other,
+    WaitableUpdatedFunc waitable_added,
+    WaitableUpdatedFunc waitable_removed);
+};
+
+void
+build_entities_collection(
+  const std::vector<rclcpp::CallbackGroup::WeakPtr> & callback_groups,
+  ExecutorEntitiesCollection & collection);
+
+std::deque<rclcpp::AnyExecutable>
+ready_executables(
+  const ExecutorEntitiesCollection & collection,
+  rclcpp::WaitResult<rclcpp::WaitSet> & wait_result
+);
+
+}  // namespace executors
+}  // namespace rclcpp
+
+#endif  // RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
--- a/rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
+++ b/rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
@@ -15,7 +15,6 @@
 #ifndef RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
 #define RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
 
-#include "rclcpp/subscription_base.hpp"
 #include <deque>
 #include <unordered_map>
 #include <vector>
@@ -33,18 +32,19 @@ namespace rclcpp
 namespace executors
 {
 
-template <typename EntityValueType>
-struct CollectionEntry {
+template<typename EntityValueType>
+struct CollectionEntry
+{
   typename EntityValueType::WeakPtr entity;
   rclcpp::CallbackGroup::WeakPtr callback_group;
 };
 
-template <typename CollectionType>
+template<typename CollectionType>
 void update_entities(
   const CollectionType & update_from,
   CollectionType update_to,
-  std::function<void (typename CollectionType::mapped_type::EntitySharedPtr) > on_added,
-  std::function<void (typename CollectionType::mapped_type::EntitySharedPtr) > on_removed
+  std::function<void(typename CollectionType::mapped_type::EntitySharedPtr)> on_added,
+  std::function<void(typename CollectionType::mapped_type::EntitySharedPtr)> on_removed
 )
 {
   for (auto it = update_to.begin(); it != update_to.end(); ) {
@@ -68,18 +68,19 @@ void update_entities(
     }
   }
 }
-template <typename EntityKeyType, typename EntityValueType>
-class EntityCollection:
-  public std::unordered_map<const EntityKeyType *, CollectionEntry<EntityValueType>>
+template<typename EntityKeyType, typename EntityValueType>
+class EntityCollection
+  : public std::unordered_map<const EntityKeyType *, CollectionEntry<EntityValueType>>
 {
 public:
   using Key = const EntityKeyType *;
   using EntityWeakPtr = typename EntityValueType::WeakPtr;
   using EntitySharedPtr = typename EntityValueType::SharedPtr;
 
-  void update(const EntityCollection<EntityKeyType, EntityValueType> & other,
-              std::function<void (EntitySharedPtr)> on_added,
-              std::function<void (EntitySharedPtr)> on_removed)
+  void update(
+    const EntityCollection<EntityKeyType, EntityValueType> & other,
+    std::function<void(EntitySharedPtr)> on_added,
+    std::function<void(EntitySharedPtr)> on_removed)
   {
     update_entities(*this, other, on_added, on_removed);
   }
--- a/rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
+++ b/rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
@@ -15,11 +15,10 @@
 #ifndef RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
 #define RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
 
+#include "rclcpp/subscription_base.hpp"
 #include <deque>
 #include <unordered_map>
 #include <vector>
-
-#include "rcpputils/thread_safety_annotations.hpp"
 
 #include <rclcpp/any_executable.hpp>
 #include <rclcpp/node_interfaces/node_base.hpp>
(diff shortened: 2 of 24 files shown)


mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18358/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12888/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19022/)

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
```diff
@@ -0,0 +1,287 @@
...
+#include "rclcpp/executors/executor_entities_collector.hpp"
+#include "rclcpp/executors/executor_notify_waitable.hpp"
+
+namespace rclcpp
+{
+namespace executors
+{
+
+ExecutorEntitiesCollector::ExecutorEntitiesCollector()
+: notify_waitable_(std::make_shared<ExecutorNotifyWaitable>())
```
alsora, line 26: the entities collector should take a function as input and pass it to the notify waitable constructor

rclcpp/test/rclcpp/executors/test_executor_notify_waitable.cpp
```diff
@@ -0,0 +1,95 @@
...
+  }
+};
+
+TEST_F(TestExecutorNotifyWaitable, construct_destruct) {
+  {
+    auto waitable = std::make_shared<rclcpp::executors::ExecutorNotifyWaitable>();
+    waitable.reset();
+  }
+  {
+    auto on_execute_callback = []() {};
```
alsora, line 51: we should add a test with a custom callback and verify that it's called during "execute"

mjcarroll: Still working on tests once I get my other PR rebased on here.

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
```diff
@@ -0,0 +1,286 @@
...
+    collection.erase(iter);
+    this->notify_waitable_->remove_guard_condition(group_ptr->get_notify_guard_condition().get());
+  } else {
+    throw std::runtime_error("Callback group needs to be associated with executor.");
+  }
+}
+
+void
+ExecutorEntitiesCollector::add_automatically_associated_callback_groups(
+  std::list<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr> nodes_to_check
```
alsora, line 233: this can be passed by reference (maybe const reference also?)

alsora (approved):

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,199 @@
...
+  manually_added_groups_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+
+  CallbackGroupCollection
+  automatically_added_groups_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+
+  /// nodes that are associated with the executor
+  std::list<rclcpp::node_interfaces::NodeBaseInterface::WeakPtr>
+  weak_nodes_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+
+  std::shared_ptr<ExecutorNotifyWaitable> notify_waitable_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
```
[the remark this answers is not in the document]
mjcarroll: Yes my mistake as I was rebasing/refactoring, I will restore it.

rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
```diff
@@ -0,0 +1,138 @@
...
+  typename EntityValueType::WeakPtr entity;
+  rclcpp::CallbackGroup::WeakPtr callback_group;
+};
+
+template<typename CollectionType>
+void update_entities(
+  const CollectionType & update_from,
+  CollectionType update_to,
+  std::function<void(typename CollectionType::mapped_type::EntitySharedPtr)> on_added,
+  std::function<void(typename CollectionType::mapped_type::EntitySharedPtr)> on_removed
```
alsora, line 47: there's no `mapped_type` definition, this should just be
```suggestion
  std::function<void(typename CollectionType::EntitySharedPtr)> on_removed
```

rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
```diff
@@ -0,0 +1,138 @@
...
+        on_removed(entity);
+      }
+      it = update_to.erase(it);
+    } else {
+      ++it;
+    }
+  }
+  for (auto it = update_from.begin(); it != update_from.end(); ++it) {
+    if (update_to.count(it->first) == 0) {
+      auto entity = it->entity.lock();
```
alsora, line 63:
```suggestion
      auto entity = it->second.entity.lock();
```

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
```diff
@@ -0,0 +1,286 @@
...
+#include "rclcpp/node_interfaces/node_base_interface.hpp"
+
+namespace rclcpp
+{
+namespace executors
+{
+
+ExecutorEntitiesCollector::ExecutorEntitiesCollector(
+  std::function<void(void)> on_notify_waitable_callback)
+: notify_waitable_(std::make_shared<ExecutorNotifyWaitable>(on_notify_waitable_callback))
```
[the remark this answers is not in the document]
alsora: In hindsight, it's actually better as it is right now for the events-executor because I should handle it separately.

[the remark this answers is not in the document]
mjcarroll: I am currently manually adding it as part of my `on_notify_waitable_callback` function.

rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
```diff
@@ -0,0 +1,138 @@
...
+      if (entity) {
+        on_added(entity);
+      }
+      update_to.insert(*it);
+    }
+  }
+}
```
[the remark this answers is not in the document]
mjcarroll: Fixed in my latest push.

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
mjcarroll: Done

rclcpp/src/rclcpp/executors/executor_notify_waitable.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+  if (guard_condition == nullptr) {
+    throw std::runtime_error("Attempting to remove null guard condition.");
+  }
+  std::lock_guard<std::mutex> lock(guard_condition_mutex_);
+  to_remove_.push_back(guard_condition);
+}
+
+size_t
+ExecutorNotifyWaitable::get_number_of_ready_guard_conditions()
```
alsora, line 102: I'm not sure I understand this.
Why is the `get_number_of_ready_guard_conditions` function responsible for updating the `notify_guard_conditions_` list ?

What's the advantage of having the 3 lists `to_add`, `to_remove` and `notify_guard_conditions_` ?

mjcarroll: The three lists are to prevent the `notify_guard_conditions_` from being mutated while being waited on in the wait set.

I placed it in the `get_number_of_ready_guard_conditions` because when the waitable is used by the executor, the following happens:

* `get_number_of_ready_guard_conditions` - to check how many conditions to add from this waitable, and is accounted for when resizing the waitset
* `add_to_wait_set` - allows the waitable to insert itself into the waitset (but no more than what was returned from `get_number_of_ready_guard_conditions`
* `is_ready` - allows the waitable to check it's entries in the waitset, here is where we need the corresponding list of guard conditions to check against
* `take_data` and `execute` - runs the contents of the user callback

So in this case, if `add_node` or `remove_node` are called from another thread while the executor is spinning, we hold those guard conditions in a temporary location to be added when the next update comes.

mjcarroll: The other option would be to queue node/callback_group updates and process those when we re-collect entities?

alsora: Ok.
The events-executor needs to register a callback when a guard condition is added and remove it when it is removed.
Outside of that, it doesn;t need to hold on the callbacks.

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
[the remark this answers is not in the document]
alsora: What do you think of having the executor own the notify waitable and pass it to the entities collector constructor?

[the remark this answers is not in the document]
mjcarroll: I'm good with that, I will make the change.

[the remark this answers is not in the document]
mjcarroll: pushed.  From here on, I will be extra sure to not rebase this branch since I know you are using it.  My feature branch for the actual executor implementation may be rebased, so don't depend on that one.

rclcpp/src/rclcpp/executors/executor_notify_waitable.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+  return nullptr;
+}
+
+void
+ExecutorNotifyWaitable::add_guard_condition(rclcpp::GuardCondition::WeakPtr guard_condition)
+{
+  std::lock_guard<std::mutex> lock(guard_condition_mutex_);
+  if (notify_guard_conditions_.count(guard_condition) == 0) {
+    notify_guard_conditions_.insert(guard_condition);
```
alsora, line 107: should we add a guard condition only if it's valid?

mjcarroll: Good call, done in 9dd48ce

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18398/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12944/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19079/)

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -89,13 +89,16 @@ class CallbackGroup
...
+   * \param[in] get_node_context Lambda to retrieve the node context when
+   *   checking that the creating node is valid and using the guard condition.
    * \param[in] automatically_add_to_executor_with_node A boolean that
    *   determines whether a callback group is automatically added to an executor
    *   with the node with which it is associated.
    */
   RCLCPP_PUBLIC
   explicit CallbackGroup(
     CallbackGroupType group_type,
+    std::function<rclcpp::Context::SharedPtr(void)> get_node_context,
```
clalancette, line 101: If we are worried about tick-tocking this class (as we seem to be for `get_notify_guard_condition` below), then don't we need to keep a deprecated version of the constructor that doesn't take in `get_node_context`?

rclcpp/src/rclcpp/callback_group.cpp
```diff
@@ -129,6 +138,31 @@ CallbackGroup::get_notify_guard_condition(const rclcpp::Context::SharedPtr conte
...
+      notify_guard_condition_ = nullptr;
+    }
+
+    if (!notify_guard_condition_) {
+      notify_guard_condition_ = std::make_shared<rclcpp::GuardCondition>(context_ptr);
+    }
+
+    return notify_guard_condition_;
+  } else {
+    throw std::runtime_error("Couldn't get guard condition from invalid context");
```
clalancette, line 161: Should we really throw an exception here?  That is, it sort of seem to me like failing to get the context might be a reasonable thing to happen, so should we just return nullptr and let the caller deal with it?  I'm not at all sure about this, so please take this as a question.

rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
```diff
@@ -0,0 +1,211 @@
...
+ *  \param[inout] update_to The collection representing the current iteration's state
+ *  \param[in] on_added Callback fired when a new entity is detected
+ *  \param[in] on_removed Callback fired when an entity is removed
+ */
+template<typename CollectionType>
+void update_entities(
+  const CollectionType & update_from,
+  CollectionType & update_to,
+  std::function<void(typename CollectionType::EntitySharedPtr)> on_added,
+  std::function<void(typename CollectionType::EntitySharedPtr)> on_removed
```
clalancette, line 71: Is it expected that the `on_added` or `on_removed` callbacks will often need to share in the ownership of the Entity?  If not, we might want to consider passing these as a `shared_ptr &` or `const shared_ptr &` to avoid copying the shared_ptr object.

(same comment about `update` below which calls this function)

rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
```diff
@@ -0,0 +1,211 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
+#define RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTION_HPP_
+
+#include <deque>
```
clalancette, line 18:
```suggestion
#include <deque>
#include <functional>
```

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+ *
+ */
+class ExecutorEntitiesCollector
+{
+public:
+  /// Constructor
+  /**
+   * \param[in] on_notify_waitable Callback to execute when one of the associated
+   *   nodes or callback groups trigger their guard condition, indicating that their
+   *   corresponding entities have changed.
```
clalancette, line 63: This whole doc-block is outdated, isn't it?  The parameter is called `notify_waitable`, and doesn't seem to be a callback.

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+   */
+  RCLCPP_PUBLIC
+  explicit ExecutorEntitiesCollector(
+    std::shared_ptr<ExecutorNotifyWaitable> notify_waitable);
+
+  /// Destructor
+  RCLCPP_PUBLIC
+  ~ExecutorEntitiesCollector();
+
+  bool has_pending();
```
clalancette, line 73: Should this be `const`?

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTOR_HPP_
+#define RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTOR_HPP_
+
+#include <list>
```
clalancette, line 18: I don't think `std::list` is used anywhere in this file.

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
```diff
@@ -0,0 +1,407 @@
...
+  }
+
+  this->pending_removed_nodes_.insert(node_ptr);
+}
+
+void
+ExecutorEntitiesCollector::add_callback_group(rclcpp::CallbackGroup::SharedPtr group_ptr)
+{
+  std::atomic_bool & has_executor = group_ptr->get_associated_with_executor_atomic();
+  if (has_executor.exchange(true)) {
```
alsora, line 131: @mjcarroll I think that this is a bug.
when you call `add_callback_group` you immediately set this to "true", but then you also add the group to the `pending_manually_added_groups_`.

If you later call `update_collections`, it will try to add again the callback group to the executor and fail, becuase the bool variable was already set.

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+  remove_node(rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr);
+
+  /// Check if a node is associated with this executor/collector.
+  /**
+   * \param[in] node_ptr a shared pointer that points to a node base interface
+   * \return true if the node is tracked in this collector, false otherwise
+   */
+  RCLCPP_PUBLIC
+  bool
+  has_node(const rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_ptr);
```
clalancette, line 101: This API doesn't seem to be implemented; should we just drop it?

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
alsora: The same is probably true also in the `add_node` function.

I think that it's probably correct to do the "exchange" here, but then we shouldn't do it again in `add_callback_group_to_collection`

mjcarroll: Fixed in 1b1a915

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
```diff
@@ -0,0 +1,403 @@
...
+}
+
+void
+ExecutorEntitiesCollector::add_callback_group_to_collection(
+  rclcpp::CallbackGroup::SharedPtr group_ptr,
+  CallbackGroupCollection & collection)
+{
+  auto iter = collection.insert(group_ptr);
+  if (iter.second == false) {
+    throw std::runtime_error("Callback group has already been added to this executor.");
```
alsora, line 266: could the same problem we discussed before (https://github.com/ros2/rclcpp/pull/2143#discussion_r1156370153) apply also here?

I'm still having issues when adding a node

mjcarroll: I don't think so?  This should on happen in the case that a group has been added twice to the same collector.

mjcarroll: I reproduced, looking into it.

clalancette (changes requested):
I did a first pass here.  A lot of the changes I requested are trivial, but there are a couple of deeper things to think about in there.

rclcpp/src/rclcpp/callback_group.cpp
```diff
@@ -129,6 +138,31 @@ CallbackGroup::get_notify_guard_condition(const rclcpp::Context::SharedPtr conte
   return notify_guard_condition_;
 }
 
+rclcpp::GuardCondition::SharedPtr
+CallbackGroup::get_notify_guard_condition()
+{
+  std::lock_guard<std::recursive_mutex> lock(notify_guard_condition_mutex_);
+  auto context_ptr = this->get_context_();
+  if (context_ptr && context_ptr->is_valid()) {
+    std::lock_guard<std::recursive_mutex> lock(notify_guard_condition_mutex_);
```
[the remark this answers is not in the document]
mjcarroll: Nope, there has been a good bit of back and forth between here and #2142 and it slipped in a merge.

rclcpp/include/rclcpp/callback_group.hpp
mjcarroll: I don't know that we have to really worry about tick-tocking as this object is only created via the node `create_callback_group` call.  I started to out of an abundance of caution, but in hindsight it may not really be necessary.

rclcpp/src/rclcpp/executors/executor_entities_collector.cpp
mjcarroll: [0a9c9a6](https://github.com/ros2/rclcpp/pull/2143/commits/0a9c9a6403efa6271401e6440b6b425c538e1bda)

rclcpp/include/rclcpp/node_interfaces/node_base_interface.hpp
```diff
@@ -152,9 +152,14 @@ class NodeBaseInterface
    */
   RCLCPP_PUBLIC
   virtual
-  rclcpp::GuardCondition &
+  rclcpp::GuardCondition::SharedPtr
```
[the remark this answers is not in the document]
mjcarroll: Changed this to `get_shared_notify_guard_condition` so that we can tick tock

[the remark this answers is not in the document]
mjcarroll: [0ae0bea](https://github.com/ros2/rclcpp/pull/2143/commits/0ae0bea1faa6640a659cdc03ffb0022d49d1ba10)

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -137,6 +140,18 @@ class CallbackGroup
     return _find_ptrs_if_impl<rclcpp::Waitable, Function>(func, waitable_ptrs_);
   }
 
+  /// Return if the node that created this callback group still exists
+  /**
+   * As nodes can share ownership of callback groups with an executor, this
+   * may be used to ensures that the executor doesn't operate on a callback
```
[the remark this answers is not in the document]
mjcarroll: 87f41bff

rclcpp/include/rclcpp/callback_group.hpp
mjcarroll: Updated to retain both patterns: 87f41bff

If a user constructs the callbackgroup without passing a node context function and also doesn't pass in a context to `get_notify_guard_condition` then an exception will be thrown.

rclcpp/src/rclcpp/callback_group.cpp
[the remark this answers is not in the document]
mjcarroll: 58093288

rclcpp/src/rclcpp/callback_group.cpp
mjcarroll: Nope, the correct behavior should be to return nullptr, addressed in 87f41bf

rclcpp/src/rclcpp/executors/executor_entities_collection.cpp
```diff
@@ -0,0 +1,204 @@
...
+void check_ready(
+  EntityCollectionType & collection,
+  std::deque<rclcpp::AnyExecutable> & executables,
+  size_t size_of_waited_entities,
+  typename EntityCollectionType::Key * waited_entities,
+  std::function<bool(rclcpp::AnyExecutable &,
+  typename EntityCollectionType::EntitySharedPtr &)> fill_executable)
+{
+  for (size_t ii = 0; ii < size_of_waited_entities; ++ii) {
+    if (waited_entities[ii]) {
```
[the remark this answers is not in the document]
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_notify_waitable.hpp
```diff
@@ -0,0 +1,129 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__EXECUTORS__EXECUTOR_NOTIFY_WAITABLE_HPP_
+#define RCLCPP__EXECUTORS__EXECUTOR_NOTIFY_WAITABLE_HPP_
+
+#include <list>
+#include <memory>
+#include <set>
```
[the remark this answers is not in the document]
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+   * it = remove_weak_node(it);
+   *
+   * \param[in] weak_node iterator to the weak node to be removed
+   * \return Valid updated iterator in the same collection
+   */
+  RCLCPP_PUBLIC
+  NodeCollection::iterator
+  remove_weak_node(NodeCollection::iterator weak_node);
+
+  /// Implementation of removing a callback gruop from the collector.
```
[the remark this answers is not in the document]
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+  get_manually_added_callback_groups();
+
+  /// Get automatically-added callback groups known to this entity collector
+  /**
+   * This will include callback groups that are associated with nodes added via add_node
+   * \return vector of automatically-added callback groups
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::CallbackGroup::WeakPtr>
+  get_automatically_added_callback_groups();
```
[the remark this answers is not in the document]
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+  get_all_callback_groups();
+
+  /// Get manually-added callback groups known to this entity collector
+  /**
+   * This will include callback groups that have been added via add_callback_group
+   * \return vector of manually-added callback groups
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::CallbackGroup::WeakPtr>
+  get_manually_added_callback_groups();
```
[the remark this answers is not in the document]
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+  remove_callback_group(rclcpp::CallbackGroup::SharedPtr group_ptr);
+
+  /// Get all callback groups known to this entity collector
+  /**
+   * This will include manually added and automatically added (node associated) groups
+   * \return vector of all callback groups
+   */
+  RCLCPP_PUBLIC
+  std::vector<rclcpp::CallbackGroup::WeakPtr>
+  get_all_callback_groups();
```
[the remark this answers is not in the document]
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
mjcarroll: dropped in [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
```diff
@@ -0,0 +1,287 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTOR_HPP_
+#define RCLCPP__EXECUTORS__EXECUTOR_ENTITIES_COLLECTOR_HPP_
+
+#include <list>
+#include <map>
+#include <memory>
```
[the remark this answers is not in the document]
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
mjcarroll: [debe396](https://github.com/ros2/rclcpp/pull/2143/commits/debe396b71850b3783c024be95fe50214e4bc8c0)

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
mjcarroll: Updated docs

rclcpp/include/rclcpp/executors/executor_entities_collector.hpp
mjcarroll: Made this and following `const`

rclcpp/include/rclcpp/executors/executor_entities_collection.hpp
mjcarroll: Yes, const ref makes sense

rclcpp/include/rclcpp/callback_group.hpp
```diff
@@ -137,6 +180,18 @@ class CallbackGroup
...
+  /**
+   * As nodes can share ownership of callback groups with an executor, this
+   * may be used to ensure that the executor doesn't operate on a callback
+   * group that has outlived it's creating node.
+   *
+   * \return true if the creating node still exists, otherwise false
+   */
+  RCLCPP_PUBLIC
+  bool
+  has_valid_node();
```
clalancette, line 193: Can this be a `const` method?

rclcpp/include/rclcpp/executors/executor_notify_waitable.hpp
```diff
@@ -0,0 +1,130 @@
...
+{
+
+/// Maintain a collection of guard conditions from associated nodes and callback groups
+/// to signal to the executor when associated entities have changed.
+class ExecutorNotifyWaitable : public rclcpp::Waitable
+{
+public:
+  RCLCPP_SMART_PTR_DEFINITIONS(ExecutorNotifyWaitable)
+
+
```
clalancette, line 38: Nit: unnecessary space.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
```diff
@@ -213,6 +214,13 @@ NodeBase::create_callback_group(
 rclcpp::CallbackGroup::SharedPtr
 NodeBase::get_default_callback_group()
 {
+  // Create the default callback group, if needed.
+  if (nullptr == default_callback_group_) {
+    using rclcpp::CallbackGroupType;
+    default_callback_group_ =
+      NodeBase::create_callback_group(CallbackGroupType::MutuallyExclusive);
+  }
```
clalancette, line 222: This change makes it so `get_default_callback_group` is no longer logically `const`.  What's the particular reason for moving this out of the constructor and into here?

clalancette (changes requested):
I've left just a handful of new comments.  Otherwise this is looking pretty good to me.

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mjcarroll: The issue here is with the lambda that I used.  Since the lambda takes a weak pointer to node to later lock and get context, we need a valid weak pointer at the time that the callback group is created.

If the callback group is created in the constructor, smart pointers aren't valid yet (you can't get a shared or weak pointer in the constrcutor).

If you have a clever way around this, I would be willing to give it a shot.

rclcpp/include/rclcpp/executors/executor_notify_waitable.hpp
mjcarroll: removed

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18498/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13046/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19229/)

alsora (approved):

rclcpp/src/rclcpp/node_interfaces/node_base.cpp
mjcarroll: After a bit of thinking, I decided to remove the `has_valid_node` API and add some todo's about it.  As I discovered over here (https://github.com/ros2/geometry2/pull/595), while we shouldn't destroy a node before it's callback groups, there isn't any programmatic way of enforcing it right now.  Using the `has_valid_node` to throw an exception will cause users' programs to except where it previously didn't.

I think in the long term, we need to be better about managing Node/CallbackGroup lifetimes.  While the Node "owns" the default callback group (via shared pointer), it may share ownership with the person who constructed the Node (default callback group is a parameter).  Additionally, further added CallbackGroups are only held as weak_ptrs, so the node doesn't really "own" them.  This probably requires more consideration.

The TLDR for here is that we can drop the `has_valid_node` API (and corresponding checks in the entities collector) and just keep a weak copy of the context in the lambda to provide all that is necessary to create the guard conditions.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18532/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13078/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19263/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18533/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13079/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19263/)

mjcarroll:
aarch flaked: :( [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/13082/)

rclcpp/test/rclcpp/executors/test_entities_collector.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+  node.reset();
+
+  /**
+   * TODO(mjcarroll): Assert this when we are enforcing that nodes must be destroyed
+   * after their created callback groups.
+  RCLCPP_EXPECT_THROW_EQ(
+    entities_collector.remove_callback_group(cb_group),
+    std::runtime_error("Node must not be deleted before its callback group(s)."));
+  */
```
wjwwood, line 282: How do we enforce this? Or when do you think this TODO could be resolved?

wjwwood (approved):
I did a pretty thorough review and I didn't really see any issues. I assume a lot of them have been ironed out by the tests and existing applications which exercise it well.

lgtm

clalancette:
Windows unfortunately has a bunch of new failures that probably need to be looked into.  They were not in the nightlies.

(for that matter, we should also probably run CI on Windows Debug to see what it has to say here)

mjcarroll:
> Windows unfortunately has a bunch of new failures that probably need to be looked into. They were not in the nightlies.

Yeah, I'm looking into them.

mjcarroll:
Windows release again: [Build Status](https://ci.ros2.org/job/ci_windows/19276/)

clalancette:
@mjcarroll Given our long discussion yesterday, where are we with this PR?  I think we came to the conclusion that we have a flaky test in `test_rclcpp`, is there anything else that we need here?

clalancette (approved):

mjcarroll:
> @mjcarroll Given our long discussion yesterday, where are we with this PR? I think we came to the conclusion that we have a flaky test in test_rclcpp, is there anything else that we need here?

The test in `test_rclcpp` is unrelated this pull request, it's the pull request using the `rclcpp::WaitSet` is where that test failure came in.

clalancette:

Issue #2157, opened 2023-04-04
MultiThreadedExecutor virtual run


guillaumeautran:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - From sources
- Version or commit hash:
  - Galactic and Humble
- DDS implementation:
  - rmw_cyclonedds_cpp
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
The multi-threaded executor declares the [void run(size_t this_thread_number);](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp#L77-L79) method as protected but does not declares it as a virtual method. This prevents a child class from overriding the run method and prevents improving that method in a case by case basis. 
For example, the current implementation is not exception safe which means that if an exception is thrown in the lower layer of rcl implementation (possibly at the rmw layer), the multithreaded executor thread might abnormally terminate since the exception is not handled. 

#### Expected behavior
The run method should be override'able by a child class. 

#### Actual behavior
The run method is not override'able by a child class.

PR #2158, opened 2023-04-04
Virtual run method for MultithreadedExecutor #2157

guillaumeautran:
Change the MultiThreadedExecutor::run method to be virtual in order to allow overriding of the functionality. This helps to add exception handling and other functionality not specific to the executor itself.

issue: #2157

### DIFF
--- a/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
@@ -75,7 +75,7 @@ public:
 
 protected:
   RCLCPP_PUBLIC
-  void
+  virtual void
   run(size_t this_thread_number);
 
 private:
--- a/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
+++ b/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
@@ -61,8 +61,10 @@ MultiThreadedExecutor::spin()
   {
     std::lock_guard wait_lock{wait_mutex_};
     for (; thread_id < number_of_threads_ - 1; ++thread_id) {
-      auto func = std::bind(&MultiThreadedExecutor::run, this, thread_id);
-      threads.emplace_back(func);
+      threads.emplace_back(
+        [&](size_t this_thread_number) -> void {
+          this->run(this_thread_number);
+        }, thread_id);
     }
   }
 


rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -75,7 +75,7 @@ class MultiThreadedExecutor : public rclcpp::Executor
 
 protected:
   RCLCPP_PUBLIC
-  void
+  virtual void
```
fujitatomoya, line 78: > The run method should be override'able by a child class.

Can we just not create `UserMultiThreadedExecutor` inherited from `MultiThreadedExecutor` to override `run` method and then call `UserMultiThreadedExecutor::spin()`?

i think adding `virtual` here is supposed to mean that, having sub-class instance pointed by super-class pointer, and handle super-class pointers in the program which actually calls sub-class `run` override method as super-class? i do not really think of the use case for this.

i might be mistaken...

guillaumeautran: No we cannot really do that because since the run method is not virtual, the `MultiThreadedExecutor::spin()` will just call the `MultiThreadedExecutor::run(...)` method (as opposed to the child class implementation). 

Making this method virtual allows for the child class to be executed correctly. 
Here is an example of a child class implementation:
```
class SafeMultiThreadedExecutor : public  rclcpp::executors::MultiThreadedExecutor                                                                                                                                 
{                                                                                                                                                                                                                  
public:                                                                                                                                                                                                            
  SafeMultiThreadedExecutor() : rclcpp::executors::MultiThreadedExecutor()                                                                                                                                         
  {}                                                                                                                                                                                                                                                                                                                                                                                                                               
protected:                                                                                                                                                                                                         
  void run(size_t this_thread_number) override                                                                                                                                                                     
  {                                                                                                                                                                                                                
    try {                                                                                                                                                                                                          
      rclcpp::executors::MultiThreadedExecutor::run(this_thread_number);                                                                                                                                           
    } catch (const std::exception & e) {                                                                                                                                                                           
      RCUTILS_LOG_ERROR_NAMED(                                                                                                                                                                                     
        "api_relay", "Exception in MultiThreadedExecutor (%ld): %s", this_thread_number, e.what());                                                                                                                
    }                                                                                                                                                                                                              
  }                                                                                                                                                                                                                
};             
```

Note that the `override` keyword tells the compiler that this implementation is expected to override the base class implementation. Without my patch here, the compiler will fail indicating that this class does not override the base class implementation. 

Take a look at the [spin()](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp#L68-L70) method. This is how it should be implemented...

alsora: I think that your example should rather be implemented by overriding `spin` and wrapping the `run` calls with the try-catch
https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L69

guillaumeautran: That is another approach but it does force me to replicate the entire spin method. The run invocation that is problematic in my case is [this one](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L64) That's because those calls are run in a separate thread context. So if an exception is thrown in those other thread, the application will terminate abnormally with no possibility to handle the exception. 

Obviously, the whole spin method can be copied but I was hoping for a more lightweight approach...

fujitatomoya: Ah, i see what you mean. instead of `virtual`, it cannot call sub-class `run` in `MultiThreadedExecutor::spin()`. for your example, even calling `SafeMultiThreadedExecutor::spin()` (which is calling `MultiThreadedExecutor::spin()`) that will call `MultiThreadedExecutor::run()`.

> That is another approach but it does force me to replicate the entire spin method.

true.

fujitatomoya:
@iuhilnehc-ynos @wjwwood @mjcarroll @clalancette any comments? (be advised that this is ABI break change.)

clalancette:
> (be advised that this is ABI break change.)

That's totally fine for Rolling, though we will not be able to backport it to Humble.

In terms of the change itself, I'd like to get @mjcarroll 's input here, since he has a good handle of what is going on in the executors right now.

iuhilnehc-ynos:
Sorry, I can't agree on this fix.

Please notice that the `MultiThreadedExecutor::spin` is calling the [MultiThreadedExecutor::run](https://github.com/ros2/rclcpp/blob/85c0af4fa06264a6b92a519197774c0e10b9606e/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L64) directly.

Making the `run` as a virtual method can't make all threads call the override `run` if users create a class inherited from `MultiThreadedExecutor` and only override the `run`. Don't you think it's weird for users?

I think the original expectation for this design is,
If users want to override the `spin` in a new class, they can use the `MultiThreadedExecutor::run` in their spin.
That's why `MultiThreadedExecutor::run` is designed in the protected section without virtual.

mjcarroll:
I agre with @iuhilnehc-ynos, I think that the original intention was only for users to override the `spin` method, as that is what is also available on the base `rclcpp::Executor` class.

If you want to create a new multithreadedexecutor with different behavior, it may make more sense to derive directly from the `rclcpp::Executor` class and implement it as you see fit.

If the issue is that you feel the current multithreadedexecutor doesn't properly handle exceptions, then that is a bug that can be addressed in the current implementation.

guillaumeautran:
Yes, the fact that the current MultiThreadedExecutor does not handle exceptions is really what is bothering me right now. I can look to see if that in itself can be fixed. 
I do agree that my current implementation is probably not working as I intend it to.

alsora:
I think that we shouldn't have executors to catch all exceptions.
That's a decision the users should take depending on their application

guillaumeautran:
I understand, but with the current implementation, the user has two options: 
1) don't catch any exceptions.
2) re-write much of the MultiThreadedExecutor (which is not [D.R.Y.](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)) 

Neither option is acceptable IMHO. 

Let me propose a potential implementation that will support both usecase...

guillaumeautran:
This is a potential improvement on my original approach. The run method is still virtual and can still have an override by a child class which is going to be also used by the child class. 

If this is still not acceptable, then I'd like to get some guidance on how to handle this simple case of wanting to catch exceptions.

fujitatomoya:
> Please notice that the MultiThreadedExecutor::spin is calling the [MultiThreadedExecutor::run](https://github.com/ros2/rclcpp/blob/85c0af4fa06264a6b92a519197774c0e10b9606e/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L64) directly.

@iuhilnehc-ynos I was thinking that how about changing into `this`? probably it should call the object's `run` when calling `spin`? just idea though. (maybe different topic...)

> If the issue is that you feel the current multithreadedexecutor doesn't properly handle exceptions, then that is a bug that can be addressed in the current implementation.

@mjcarroll totally agree on this, we should go directly solve the problem and requirement 👍 

I would want to have the capability to register the callback how to handle the exception happens in the executor. with that, probably it can give the flexibility how to handler both user generated exception and system exception?

guillaumeautran:
So, what's the guidance here? 

> I would want to have the capability to register the callback how to handle the exception happens in the executor. with that, probably it can give the flexibility how to handler both user generated exception and system exception?

Where do I start?

fujitatomoya:
Probably we can create the new feature request for that, but that can be managed with another issue.

> I was thinking that how about changing into this? probably it should call the object's run when calling spin?

Sorry to be late, but i just realized you just did it. i think it is worth to keep this open for more discussion 👍

PR #2164, opened 2023-04-11
Trigger the intraprocess guard condition with data

mjcarroll:
If the intraprocess buffer still has data after taking, re-trigger the guard condition to ensure that the executor will continue to service it, even if incoming publications stop.

I found this bug when working on #2142.  There is a situation where if an intraprocess publisher sends all of it's messages before the subscription starts servicing it, then the executor will fail to service the entire queue.

Note that in the current `rclcpp` implementation, this really only impacts the `StaticSingleThreadedExecutor`, as the `SingleThreadedExecutor` and `MultiThreadedExector` both automatically trigger the waitset at the end of an executable cycle.  I discovered this because the rclcpp::Waitset based executor also does not recompute work.  I added the test to run against all executors to prevent regression.

### DIFF
--- a/rclcpp/test/rclcpp/executors/test_executors.cpp
+++ b/rclcpp/test/rclcpp/executors/test_executors.cpp
@@ -685,10 +685,11 @@ TYPED_TEST(TestIntraprocessExecutors, testIntraprocessRetrigger) {
   // Fire a timer every 10ms up to 5 seconds waiting for subscriptions to be read.
   loops = 0;
   auto timer = this->node->create_wall_timer(
-    std::chrono::milliseconds(10), [this, &executor, &loops]() {
+    std::chrono::milliseconds(10), [this, &executor, &loops, &kNumMessages]() {
       loops++;
       if (kNumMessages == this->callback_count.load() ||
-          loops == 500) {
+      loops == 500)
+      {
         executor.cancel();
       }
     });
--- a/rclcpp/test/rclcpp/executors/test_executors.cpp
+++ b/rclcpp/test/rclcpp/executors/test_executors.cpp
@@ -593,3 +593,95 @@ TEST(TestExecutors, testSpinUntilFutureCompleteNodePtr) {
 
   rclcpp::shutdown();
 }
+
+template<typename T>
+class TestIntraprocessExecutors : public ::testing::Test
+{
+public:
+  static void SetUpTestCase()
+  {
+    rclcpp::init(0, nullptr);
+  }
+
+  static void TearDownTestCase()
+  {
+    rclcpp::shutdown();
+  }
+
+  void SetUp()
+  {
+    const auto test_info = ::testing::UnitTest::GetInstance()->current_test_info();
+    std::stringstream test_name;
+    test_name << test_info->test_case_name() << "_" << test_info->name();
+    node = std::make_shared<rclcpp::Node>("node", test_name.str());
+
+    callback_count = 0;
+
+    const std::string topic_name = std::string("topic_") + test_name.str();
+
+    rclcpp::PublisherOptions po;
+    po.use_intra_process_comm = rclcpp::IntraProcessSetting::Enable;
+    publisher = node->create_publisher<test_msgs::msg::Empty>(topic_name, rclcpp::QoS(1), po);
+
+    auto callback = [this](test_msgs::msg::Empty::ConstSharedPtr) {
+        this->callback_count.fetch_add(1);
+      };
+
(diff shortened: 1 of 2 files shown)


mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18499/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13047/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19230/)

rclcpp/test/rclcpp/executors/test_executors.cpp
```diff
@@ -593,3 +593,95 @@ TEST(TestExecutors, testSpinUntilFutureCompleteNodePtr) {
...
+
+  using ExecutorType = TypeParam;
+  ExecutorType executor;
+  executor.add_node(this->node);
+
+  EXPECT_EQ(0, this->callback_count.load());
+  this->publisher->publish(test_msgs::msg::Empty());
+
+  const std::chrono::milliseconds sleep_per_loop(10);
+  while (1u != this->callback_count.load()) {
```
clalancette, line 669: I think it would be good to have an escape hatch here, where we only wait 500 times (5 seconds) or something.  Otherwise if there is a bug, this will hang forever.

rclcpp/test/rclcpp/executors/test_executors.cpp
```diff
@@ -593,3 +593,95 @@ TEST(TestExecutors, testSpinUntilFutureCompleteNodePtr) {
...
+  EXPECT_EQ(1u, this->callback_count.load());
+
+  // reset counter
+  this->callback_count.store(0);
+
+  for (size_t ii = 0; ii < kNumMessages; ++ii) {
+    this->publisher->publish(test_msgs::msg::Empty());
+  }
+  auto timer = this->node->create_wall_timer(
+    std::chrono::milliseconds(250), [&executor]() {
```
clalancette, line 682: So this timeout seems short, particularly on Windows.  Conversely, this is a "fixed" timeout, so we will always wait this long, even if everything is delivered faster.  Can we make this wall timer execute at a faster rate (maybe 100Hz), and cancel the executor as soon as it gets `kNumMessages` or times out (after 5 seconds or something)?

clalancette (changes requested):

rclcpp/test/rclcpp/executors/test_executors.cpp
mjcarroll: Done.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18502/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13049/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19232/)

clalancette (approved):

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18512/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13057/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19240/)
*

fujitatomoya (commented):
@mjcarroll i think this is one of the critical bug, should we backport this to humble? i am inclined to backport since many application possibly meets this issue when they use `StaticSingleThreadedExecutor`? (no backporting is also fine, since this is under experimental.)

CC: @clalancette @alsora

mjcarroll:
I think it is a bug, but unclear how many people would actually run into it.  

If you have anything else that wakes up the executor, it will still service the intraprocess subscriptions unconditionally (which may be another bug, but seems to work).

In the case of the current `SingleThreadedExecutor` implementation, it actually fires the interrupt guard condition after each executable, guaranteeing that it will re-evaluate.  In the case of the `StaticSingleThreadedExecutor`, you would need to have an executor with only intraprocess subscriptions that weren't actively being published to and held data.

fujitatomoya:
yeah, that is we never know... 💭 
under the condition, i would say that that is unlikely to see, but still that is the bug to address.
so that we can avoid the possible problem for user application. besides, it does support ABI compatibility.

probably we can give it a couple of weeks to see if nothing happens on rolling, and then backport to humble.

any objections?

mjcarroll:
No objections! better safe than sorry.

fujitatomoya:
@mergifyio backport to humble

fujitatomoya:
> Backport to branch to failed

😢 

@mergifyio backport humble

PR #2165, opened 2023-04-11
Dynamic Subscription (REP-2011 Subset): Stubs for rclcpp

methylDragon:
This is a stubbed out version of: https://github.com/ros2/rclcpp/pull/2160

- New classes are defined (stubbed out)
- New virtual methods are set for SubscriptionBase
- SubscriptionBase is using an enum to govern its behavior and execute_subscription behavior now uses that enum

Note: I had to introduce (and stub out) the new dynamic typesupport classes in rclcpp since the virtual methods on SubscriptionBase use them

I added the classes, then stubbed them out in a separate commit so it's clear what's being removed.

Stuff removed are:
- Non-virtual constructors
- Templated class methods
- Normal class methods
- The function implementations
- Executor behavior surrounding dynamic subscriptions (inside a code branch in execute_subscription())

PR #2168, opened 2023-04-13
Implement deliver message kind

methylDragon:
Follow-up to: https://github.com/ros2/rclcpp/pull/2165

From further discussions, it was found that the `SubscriptionType` enum to change the behavior of the subscription was better thought of as a way to configure the "category" of message that is being delivered by a subscription (i.e. what is being passed in the subscription's handle function.)

Under that view, it is also possible to collapse the two dynamic message related enum options into one (and have the behavior change in the implementation of the take function.)

### DIFF
--- a/rclcpp/include/rclcpp/subscription_base.hpp
+++ b/rclcpp/include/rclcpp/subscription_base.hpp
@@ -63,10 +63,18 @@ namespace experimental
 class IntraProcessManager;
 }  // namespace experimental
 
-/// The kind of message that the subscription delivers in its callback.
+/// The kind of message that the subscription delivers in its callback, used by the executor
 /**
+ * This enum needs to exist because the callback handle is not accessible to the executor's scope.
+ *
  * "Kind" is used since what is being delivered is a category of messages, for example, there are
- * different ROS message types that can be delivered, but they're all ROS messages
+ * different ROS message types that can be delivered, but they're all ROS messages.
+ *
+ * As a concrete example, all of the following callbacks will be considered ROS_MESSAGE for
+ * DeliveredMessageKind:
+ *   - void callback(const std_msgs::msg::String &)
+ *   - void callback(const std::string &)  // type adaption
+ *   - void callback(std::unique_ptr<std_msgs::msg::String>)
  */
 enum class DeliveredMessageKind : uint8_t
 {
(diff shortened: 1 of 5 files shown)


rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -63,13 +63,17 @@ namespace experimental
...
 {
-  INVALID = 0,  // The subscription type is most likely uninitialized
-  ROS_MESSAGE = 1,  // take message as ROS message and handle as ROS message
-  SERIALIZED_MESSAGE = 2,  // take message as serialized and handle as serialized
-  DYNAMIC_MESSAGE_DIRECT = 3,  // take message as DynamicMessage and handle as DynamicMessage
-  DYNAMIC_MESSAGE_FROM_SERIALIZED = 4  // take message as serialized and handle as DynamicMessage
+  INVALID = 0,
+  ROS_MESSAGE = 1,  // The subscription delivers a ROS message to its callback
+  SERIALIZED_MESSAGE = 2,  // The subscription delivers a serialized message to its callback
+  DYNAMIC_MESSAGE = 3,  // The subscription delivers a dynamic message to its callback
```
wjwwood, line 76: I asked @methylDragon to change this, since I believe the difference between a dynamic message delivered from the middleware or by wrapping a serialized message from the middleware should be encapsulated in the subscription, that is to say, neither the executor nor the user creating the subscription should care how the dynamic message is delivered. That also lead to the discussion of the name of this enum and how it was a bit ambiguous, e.g. normal vs generic subscription or lifecycle vs normal, "type" was just overloaded too much.

Also, we discussed a use case where the user wanted the dynamic message to be delivered by wrapping a serialized message always, even if the middleware supported directly returning one, but then we concluded the user could do this themselves by asking for a serialize message in their callback and creating a dynamic message from that themselves. And if it becomes a common pattern we want to put into our API, it should be a separate option form this enum.

wjwwood (approved):

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18549/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13093/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19283/)

PR #2171, opened 2023-04-19
Fix a format-security warning when building with clang

clalancette:
In particular, you should never have a "bare" string in a printf-like call; that could potentially access uninitialized memory. Instead, make sure to format the string with %s.


clalancette:
CI for this is over in https://github.com/ros2/rcl/pull/1064#issuecomment-1515468520

alsora (approved):

fujitatomoya (approved):

wjwwood (approved):

Yadunund:
Do we backport this to `iron`?

clalancette:
> Do we backport this to `iron`?

I don't think it is necessary; this is only in a test.

PR #2173, opened 2023-04-26
Feature/available capacity of ipm

DensoADAS:
We were in need to ensure that no messages are lost while intraprocess communication (something like acknowledge for interprocess communication). Therefore we've added method to the publisher to check if the buffers of the subscription have remaining capacity.

- updated ipm buffer and manager
- updated publisher
- added unit tests for everything

Thank you for reviewing!

### DIFF
--- a/rclcpp/src/rclcpp/intra_process_manager.cpp
+++ b/rclcpp/src/rclcpp/intra_process_manager.cpp
@@ -225,5 +225,48 @@ IntraProcessManager::can_communicate(
   return true;
 }
 
+size_t
+IntraProcessManager::lowest_available_capacity(const uint64_t intra_process_publisher_id) const
+{
+  size_t capacity = std::numeric_limits<size_t>::max();
+
+  auto publisher_it = pub_to_subs_.find(intra_process_publisher_id);
+  if (publisher_it == pub_to_subs_.end()) {
+    // Publisher is either invalid or no longer exists.
+    RCLCPP_WARN(
+      rclcpp::get_logger("rclcpp"),
+      "Calling get_subscription_count for invalid or no longer existing publisher id");
+    return 0;
+  }
+
+  for (const auto sub_id : publisher_it->second.take_shared_subscriptions) {
+    capacity = std::min(capacity, available_capacity(sub_id));
+  }
+
+  for (const auto sub_id : publisher_it->second.take_ownership_subscriptions) {
+    capacity = std::min(capacity, available_capacity(sub_id));
+  }
+
+  return capacity;
+}
+
+size_t
+IntraProcessManager::available_capacity(const uint64_t intra_process_subscription_id) const
+{
+  auto subscription_it = subscriptions_.find(intra_process_subscription_id);
+  if (subscription_it != subscriptions_.end()) {
+    auto subscription = subscription_it->second.lock();
+    if (subscription) {
+      return subscription->available_capacity();
+    }
+  }
+
+  // Subscription is either invalid or no longer exists.
+  RCLCPP_WARN(
+    rclcpp::get_logger("rclcpp"),
+    "Calling available_capacity for invalid or no longer existing subscription id");
+
+  return 0;
+}
 }  // namespace experimental
 }  // namespace rclcpp
--- a/rclcpp/src/rclcpp/intra_process_manager.cpp
+++ b/rclcpp/src/rclcpp/intra_process_manager.cpp
@@ -236,7 +236,14 @@ IntraProcessManager::lowest_available_capacity(const uint64_t intra_process_publ
     RCLCPP_WARN(
       rclcpp::get_logger("rclcpp"),
       "Calling lowest_available_capacity for invalid or no longer existing publisher id");
-    return 0;
+    return 0u;
+  }
+
+  if (publisher_it->second.take_shared_subscriptions.empty() &&
+    publisher_it->second.take_ownership_subscriptions.empty())
+  {
+    // no subscriptions available
+    return 0u;
   }
 
   for (const auto sub_id : publisher_it->second.take_shared_subscriptions) {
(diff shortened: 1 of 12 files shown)


rclcpp/src/rclcpp/intra_process_manager.cpp
```diff
@@ -225,5 +225,48 @@ IntraProcessManager::can_communicate(
...
+IntraProcessManager::lowest_available_capacity(const uint64_t intra_process_publisher_id) const
+{
+  size_t capacity = std::numeric_limits<size_t>::max();
+
+  auto publisher_it = pub_to_subs_.find(intra_process_publisher_id);
+  if (publisher_it == pub_to_subs_.end()) {
+    // Publisher is either invalid or no longer exists.
+    RCLCPP_WARN(
+      rclcpp::get_logger("rclcpp"),
+      "Calling get_subscription_count for invalid or no longer existing publisher id");
```
alsora, line 238: minor: 
```suggestion
      "Calling lowest_available_capacity for invalid or no longer existing publisher id");
```

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -384,3 +384,22 @@ std::vector<rclcpp::NetworkFlowEndpoint> PublisherBase::get_network_flow_endpoin
 
   return network_flow_endpoint_vector;
 }
+
+size_t PublisherBase::lowest_available_ipm_capacity() const
+{
+  if (!intra_process_is_enabled_) {
+    return std::numeric_limits<size_t>::max();
```
alsora, line 391: should this be `0` if intra-process is disabled?

DensoADAS: the idea was if intraprocess is not enabled, the capacity should be high to avoid blocking (in the case you are waiting for free buffers as they will never become free)

rclcpp/src/rclcpp/intra_process_manager.cpp
```diff
@@ -225,5 +225,48 @@ IntraProcessManager::can_communicate(
...
+
+  for (const auto sub_id : publisher_it->second.take_shared_subscriptions) {
+    capacity = std::min(capacity, available_capacity(sub_id));
+  }
+
+  for (const auto sub_id : publisher_it->second.take_ownership_subscriptions) {
+    capacity = std::min(capacity, available_capacity(sub_id));
+  }
+
+  return capacity;
```
fujitatomoya, line 250: there could be possibility that this returns `std::numeric_limits<size_t>::max()`. is that expected behavior? Or if `publisher_it` can be found, there must be corresponding subscription id in the list?

DensoADAS: It's expected behavior. If not subscriptions are present, the buffer is "free" -> return max. value. To check for the number of (intra-process) subscriptions there is another method.

alsora:
@DensoADAS i think that both Fujita and me are concerned with this "behavior" https://github.com/ros2/rclcpp/pull/2173#discussion_r1177629372 

The API is doing essentially two different things depending on whether a subscription is found or not. 

This makes the API unclear to use.
If there are no subscriptions, the available IPM capacity is zero, so this shouldn't return max. 

It looks like there's a very specific use-case that you are trying to solve with this API, but I don't think that this is the correct implementation.

alsora (changes requested):
@DensoADAS i think that both Fujita and me are concerned with this "behavior" https://github.com/ros2/rclcpp/pull/2173#discussion_r1177629372 

The API is doing essentially two different things depending on whether a subscription is found or not. 

This makes the API unclear to use.
If there are no subscriptions, the available IPM capacity is zero, so this shouldn't return max. 

It looks like there's a very specific use-case that you are trying to solve with this API, but I don't think that this is the correct implementation.

DensoADAS:
> @DensoADAS i think that both Fujita and me are concerned with this "behavior" [#2173 (comment)](https://github.com/ros2/rclcpp/pull/2173#discussion_r1177629372)
> 
> The API is doing essentially two different things depending on whether a subscription is found or not.
> 
> This makes the API unclear to use. If there are no subscriptions, the available IPM capacity is zero, so this shouldn't return max.
> 
> It looks like there's a very specific use-case that you are trying to solve with this API, but I don't think that this is the correct implementation.

I've updated the PR to return 0 in both cases.

rclcpp/src/rclcpp/intra_process_manager.cpp
fujitatomoya: so that means that publisher actually can write the maximum number of data in the buffer? i believe that what it means to return the maximum number here as interface, and what user application expects.

rclcpp/src/rclcpp/intra_process_manager.cpp
```diff
@@ -225,5 +225,55 @@ IntraProcessManager::can_communicate(
...
+
+  if (publisher_it->second.take_shared_subscriptions.empty() &&
+    publisher_it->second.take_ownership_subscriptions.empty())
+  {
+    // no subscriptions available
+    return 0u;
+  }
+
+  for (const auto sub_id : publisher_it->second.take_shared_subscriptions) {
+    capacity = std::min(capacity, available_capacity(sub_id));
```
alsora, line 250: I think that after the change `max -> 0`, this is now broken.
the current implementation of `available_capacity` will return 0 if a subscription is not found, but this is not correct here.

Let's say that a publisher has 2 subscriptions; if one goes out of scope and the other has capacity 2, the API should return 2.

I recommend to also add unit-tests for these corner cases

DensoADAS: that's a good hint. I've made changes to handle subscriptions which went out of scope and added an unit test.

alsora (approved):

clalancette:
@alsora If you are happy with this PR, can you please run CI on it and merge it (assuming CI comes back green)?  Thanks.

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18766/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13289/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19507/)

fujitatomoya:
Just FYI. @DensoADAS some comments are not resolved yet, but i will leave this to @alsora

PR #2177, opened 2023-05-01
Fix race condition in events-executor

alsora:
The initial implementation of the events-executor contained a bug where the executor would end up in an inconsistent state and stop processing interrupt/shutdown notifications. 

Manually adding a node to the executor results in a) producing a notify waitable event and b) refreshing the executor collections.
The notify waitable uses a `notify_waitable_event_pushed_` that is set to true when an event is pushed and to false when it is processed. This is used to prevent multiple events to be pushed while we still have something in the queue.

The inconsistent state would happen if the event was processed before the collections were finished to be refreshed: the executor would pick up the event but be unable to process it, because its associated entity wouldn't be part of the entities collection yet. The event would thus be discarded and this would leave the `notify_waitable_event_pushed_` flag to true, preventing additional notify waitable events to be pushed.
The behavior is observable only under heavy load.

The race condition could only happen during the first time the notify waitable is notified.
This PR fixes the problem by immediately adding the notify waitable to the collection to prevent races.

FYI @mauropasse @mjcarroll


alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18631/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13161/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19361/)

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18632/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19362/)

rclcpp/test/rclcpp/executors/test_events_executor.cpp
```diff
@@ -479,6 +479,73 @@ TEST_F(TestEventsExecutor, destroy_entities)
...
+{
+  // rmw_connextdds doesn't support events-executor
+  if (std::string(rmw_get_implementation_identifier()).find("rmw_connextdds") == 0) {
+    GTEST_SKIP();
+  }
+
+  // Spawn some threads to do some heavy work
+  std::atomic<bool> should_cancel = false;
+  std::vector<std::thread> stress_threads;
+  for (size_t i = 0; i < 5 * std::thread::hardware_concurrency(); i++) {
```
alsora, line 505: with these stress threads, I'm able to reproduce the bug ~75% of the times on my laptop.

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18633/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13163/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19363/)

rclcpp/src/rclcpp/experimental/executors/events_executor/events_executor.cpp
```diff
@@ -61,6 +64,10 @@ EventsExecutor::EventsExecutor(
       this->refresh_current_collection_from_callback_groups();
     });
 
+  // Make sure that the notify waitable is immediately added to the collection
+  // to avoid missing events
+  this->add_notify_waitable_to_collection(current_entities_collection_->waitables);
```
alsora, line 69: This is the fix: we immediately add the notify waitable to the current collection in the executor's constructor

mjcarroll:
Description and proposed fix make sense. While it's likely unique to the events executor, I think that it's worth moving the test to run against all of the executors, so that we can mark it as an edge case for future implementors or regressions.

rclcpp/test/rclcpp/executors/test_events_executor.cpp
```diff
@@ -479,6 +479,73 @@ TEST_F(TestEventsExecutor, destroy_entities)
...
+      // This is just some arbitrary heavy work
+      size_t total = 0;
+      for (size_t k = 0; k < 549528914167; k++) {
+        if (should_cancel) {
+          break;
+        }
+        total += k * (i + 42);
+      }
+      // Do something with the total to avoid the thread's work being optimized away
+      std::cout<<"The dummy total is: "<< total<<std::endl;
```
mjcarroll, line 516: Can we do something other than printing?  Our test logs are noisy enough.

rclcpp/test/rclcpp/executors/test_events_executor.cpp
mjcarroll: Just for reference, what is the value of hardware_concurrency on your system?

rclcpp/test/rclcpp/executors/test_events_executor.cpp
alsora: Done, I declared `total` as `volatile` which will prevent optimizations from the compiler.

alsora:
I moved the test to the "test_executors" file and added a note that it was developed for the events-executor, but it should succeed with all executors.

mjcarroll:
Changes all look good, I think we have time to rerun CI on the new changes as rolling is still not officially open for merges for a little bit.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18643/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13173/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19373/)

rclcpp/test/rclcpp/executors/test_executors.cpp
```diff
@@ -729,6 +731,77 @@ TYPED_TEST(TestExecutors, testSpinUntilFutureCompleteInterrupted)
...
+  }
+
+  // Spawn some threads to do some heavy work
+  std::atomic<bool> should_cancel = false;
+  std::vector<std::thread> stress_threads;
+  for (size_t i = 0; i < 5 * std::thread::hardware_concurrency(); i++) {
+    stress_threads.emplace_back(
+      [&should_cancel, i]() {
+        // This is just some arbitrary heavy work
+        size_t total = 0;
```
fujitatomoya, line 767: How about disable code optimization for this variable?

```suggestion
        volatile size_t total = 0;
```

fujitatomoya (commented):
lgtm, as @mjcarroll questions, cpu intensive part could be adjusted.

rclcpp/test/rclcpp/executors/test_executors.cpp
alsora: Yes, I had changed it to volatile before, I think I made a mistake with the following commit

alsora:
The github CI is failing in an unrelated test rclcpp.TestNodeGraph.construct_from_node

Is this known to be flaky?

clalancette:
> Is this known to be flaky?

Not as far as I know.

rclcpp/test/rclcpp/executors/test_events_executor.cpp
alsora: It's 16.

Note that the failure is reproducible also without these worker threads, but it takes many more trials (IIRC the failure rate was around 1%)

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18666/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13194/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19395/)

alsora:
Sorry for the noise, but I had forgot the signoff and I had to fix some errors due to moving the test from one file to the other.

This is ready for a final review

mjcarroll (approved):

Yadunund:
Do we backport this to `iron`?

clalancette:
> Do we backport this to `iron`?

This falls under the experimental section (and tests), so I would say it is eligible for backporting now (or even after release).

I'll leave the final determination of whether it should be backported to @mjcarroll and @alsora .

alsora:
Yes, considering that this is a critical bug fix for events executor users, I think it should be back ported.
The same applies to https://github.com/ros2/rclcpp/pull/2187

With these two changes, this new version of the events executor passes all our internal tests.

clalancette:
@Mergifyio backport iron

Issue #1660, opened 2021-05-06
Parameter Descriptor simplification for generic types


gbalke:
#### Feature description
Simplify creation of parameter descriptors for generic types or add documentation to rosdocs (or both).

When trying to mimic behavior in ROS1 for dynamic reconfigure, I noticed there was no documentation on how to set up the parameter descriptors. The interface also ended up being quite strange/esoteric. After trying to find ROS documentation for some time I gave up and looked for sample code which I eventually found here: https://git.3dvisionlabs.com/3dvisionlabs/software/hemistereo/apps/stereo/ros2/-/blob/master/src/hemistereo/src/ros2_wrapper.cpp

I wanted to define a single floating point value with fixed step sizes and min/max values. This ended up with the descriptor declaration:
```c++
rcl_interfaces::msg::ParameterDescriptor param_descriptor = {};
param_descriptor.type = rcl_interfaces::msg::ParameterType::PARAMETER_DOUBLE;
// Generalized for lists?
param_descriptor.floating_point_range.resize(1);
param_descriptor.floating_point_range.at(0).from_value = 0.0f;
param_descriptor.floating_point_range.at(0).to_value = 1.0f;
param_descriptor.floating_point_range.at(0).step = 0.001f;
this->declare_parameter<double>("value", 0.5, param_descriptor);
```

This is quite strange and there is a lot of excess code.

#### Implementation considerations
While the reason for this being complicated likely comes from the generated messages, that doesn't mean the implementation needs to be messy. It may be reasonable to create convenient tools which allow for the definition of such a variable in much fewer steps:
```c++
rcl_interfaces::msg::FloatingPointDescription(args_for_default_description, min, max, step);
```
This could be a class structure that simply wraps the default constructor and mimics the message types? Regardless of if this is a good idea or not, I think documentation for generic type descriptors should be more available as dynamic_reconfigure's GUI interface is nowhere near as valuable without it.

wjwwood:
I think a convenience API for this makes total sense.

It is generalized for lists and convoluted mostly to make it fit into generated message data, and not storing list and non-list separately (wire efficiency).

I'd recommend a builder pattern (https://cpppatterns.com/patterns/builder.html) for the API, so that if we add more to the descriptor in the future we won't have to add/subtract/reorder arguments to a function.

It could live in either `rcl_interfaces` as an additional header, or it could live in rclcpp, since that is where it is used in this fashion.

CursedRock17:
[...]

PR #2179, opened 2023-05-02
Parameter Descriptor Simplification 

CursedRock17:
These changes are a direct response to issue #1660 which calls for the simplification of common parameter_descriptor objects.

- Firstly, this first commit attempted to initialize a builder class for the ParameterDescription class. The references were mainly [cpppatterns](https://cpppatterns.com/patterns/builder.html) and [refactoringguru](https://refactoring.guru/design-patterns/builder) to establish a correct definition of the builder class. 
- Target Intension: to solve redundancy in the creation of param_descriptor objects, specifically the float version which comes with a counter. The `ParameterDescription` class is meant to hold all of the standard information in [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg) and then build up classes around it that can alter that information.
> It may be reasonable to create convenient tools which allow for the definition of such a variable in much fewer steps:

and 

> I'd recommend a builder pattern (https://cpppatterns.com/patterns/builder.html) for the API, so that if we add more to the descriptor in the future we won't have to add/subtract/reorder arguments to a function.

were the main tasks assigned in the creation in this so, to follow builder protocol I created a general first class, then secondary classes that contain the details for a counter such as: `min, max, step`. Then I created a function to handle the setting of these functions, where there had to be an overloaded version meant to hold templates given in the example of `this->declare_parameter<double>("value", 0.5, param_descriptor);` I assumed this had to be done because a Node must be initiated through this method, or a general version with a `rclcpp::ParameterValue.` Additionally, this was a function because the builder design patterns aims to stray away from massive constructors, which originally concerned me, because this PR was beginning to turn into that. 

There are definetly improvements to be made to this PR, especially in the addition of more subclasses - at least an int version - but I wanted to get some early feedback.


rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+// Additional ROS libraries needed
+#include "node.hpp"
+#include "rcl_interfaces/msg/describe_parameters.hpp"
+#include "rclcpp/node.hpp"
+
+namespace rclcpp
+{
+
+// Implments ParameterDesription class with builder design pattern
```
alsora, line 17:
```suggestion
// Implements ParameterDesription class with builder design pattern
```

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+class ParameterDescription::FloatingPointDescription
+{
+public:
+    // Our Main build methods which will construct the base class
+    ParameterDescription build() const;
+
+    //Builder Method - Mains
+    FloatingPointDescription& SetName(std::string name);
+    FloatingPointDescription& SetType(std::uint8_t type);
```
alsora, line 50: I think that many of these methods should be in a base class

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+    FloatingPointDescription& SetDescription(std::string description);
```
alsora, line 51: This should be something like "set description text", because the whole object is a description

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+};
+
+class ParameterDescription::FloatingPointDescription
+{
+public:
+    // Our Main build methods which will construct the base class
+    ParameterDescription build() const;
+
+    //Builder Method - Mains
+    FloatingPointDescription& SetName(std::string name);
```
alsora, line 49: Use Const reference to pass the string

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+    FloatingPointDescription& SetType(std::uint8_t type);
+    FloatingPointDescription& SetDescription(std::string description);
+    FloatingPointDescription& SetAdditionalConstraints(std::string constraints);
+    FloatingPointDescription& SetReadOnly(bool read_only);
+    FloatingPointDescription& SetDynamicTyping(bool dynamic_typing);
+
+    // Need the current node in order to begin the configuraiton state forit via the declare_parameter function which setups up the Node
+    template<typename ParameterType>
+    FloatingPointDescription& DeclareParameter<ParameterType>(ParameterType default_value, std::shared_ptr<rclcpp::Node> required_node)
```
alsora, line 58: This shouldn't take a node, but rather the required node interfaces

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+{
+    required_node->declare_parameter<ParameterType>(m_name, default_value, parameter_descriptor);
+    return *this;
+}
+    // Extended build methods specific to this class
+    FloatingPointDescription& SetMin(float min);
+    FloatingPointDescription& SetMax(float max);
+    FloatingPointDescription& SetStep(float step);
+
+    // Simplification Method - Outside of the base clss that the user should be able to set up easily, they should then be able to call a function which setups up the specifics for this class (The floating point parameter description)
```
alsora, line 68:
```suggestion
    // Simplification Method - Outside of the base class that the user should be able to set up easily, they should then be able to call a function which setups the specifics for this class (The floating point parameter description)
```

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+    class FloatingPointDescription;
+
+    // Two constructors in case certain methods don't need to declare_parameters with a Type T in the form of a template
+    ParameterDescription(std::string name, std::uint8_t type, std::string description, std::string additional_constraints, bool read_only. bool dynamic_typing) : m_name{name}, parameter_descriptor.type{type}, m_description{description}, m_additional_constraints{additional_constraints}, read_only{read_only}, dyanmic_typing{m_dyanmic_typing}{};
+
+private:
+    // The main descriptor object we're meant to adjust
+    rcl_interfaces::msg::ParameterDescriptor parameter_descriptor = {};
+
+    // Copies all the information in ParameterDescriptor.msg - https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg
```
alsora, line 31: Why do we copy all the information? Shouldn't we just store an internal parameter description message and return it as output of a "build" method?

alsora:
I'm not convinced by the need for creating different classes for each type of constraints we may eventually have on the parameters.

Given that we only have 1 parameter description object, having 1 builder class should be sufficient.
We could have some alias/utilities to simplify the creation of some commonly used descriptions, but that's a different story

CursedRock17:
I may have failed to grasp the objective the original issue was trying to fix, I was just confused on the direction they were trying to take. Though, now that you've defined it a bit better I may understand, because beforehand it seemed convoluted to use a builder - which takes zero arguments but uses multiple method calls - to create a convience API. From what you've suggested, we should make a builder API with a bunch of small method calls and a large build that returns all the details, for just the ParameterDescription obejct. Then, create two smaller function calls for the Integers ranges highlighed in both the PR and [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg) that work as a convience API.

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
CursedRock17: If any implemented class inherits from the `rclcpp::Node` class shouldn't that class be able to fall into the `rclcpp::Node` parameter, or does this engage a bad practice we don't want to follow. The goal is to use the `Node` as a form of `this`, where the original issue suggests [this code on something like line 48](https://git.3dvisionlabs.com/3dvisionlabs/software/hemistereo/apps/stereo/ros2/-/blob/master/src/hemistereo/src/ros2_wrapper.cpp). Where `this` acts as the encasing class which inherits from `rclcpp::Node` seen on [line 20](https://git.3dvisionlabs.com/3dvisionlabs/software/hemistereo/apps/stereo/ros2/-/blob/master/src/hemistereo/includes/ros2_wrapper.hpp) of that file.

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,102 @@
...
+    FloatingPointDescription& DeclareParameter<ParameterType>(ParameterType default_value, std::shared_ptr<rclcpp::Node> required_node)
+{
+    required_node->declare_parameter<ParameterType>(m_name, default_value, parameter_descriptor);
+    return *this;
+}
+    // Extended build methods specific to this class
+    FloatingPointDescription& SetMin(float min);
+    FloatingPointDescription& SetMax(float max);
+    FloatingPointDescription& SetStep(float step);
+
```
CursedRock17, line 67: Before I roll out the next set of changes how would we like these three functions: `SetMin`, `SetMax`, and `SetStep` to act?  There are multiple aspects to consider when deciding between simplicity and modularity. 


Currently these three functions have no use, because the values they set are unused. In order to make them have purpose we would not only set their corresponding value, but also resize the type of range it's trying to target, if that range had not already been set. After we would then call `parameter_descriptor.floating_point_range.at(0).corresponding_function`. On the other hand we can opt for simplicity by removing the set functions and the values they alter all together, then in the functions that Set the Description Range we can use default values to remove possible arguments because the .msg field specifies bounds that we can have. On top of this discussion, should we also create Set functions for the int range, but with different names i.e `SetMinInteger` or should we aim to use overloading like `SetMin(float/int min)`

CursedRock17:
Now with the issue #2184, should both ranges in the functions `SetFloatingPointDescriptionRange` and `SetIntegerDescriptionRange` be available in the main class. Or would it be wise to extend the `DeclareParameter` function to check, during runtime, for the type and determine the validity of a range. Similarly we could use the `parameter_descriptor.type` to check as well.

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
alsora: Not everything that creates a parameter is derived from `rclcpp::Node`; for example `rclcpp_lifecycle::LifecycleNode` doesn't.

The correct approach is to pass the list of node interfaces that you are going to need

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
CursedRock17: I interpret this as just returning the entirety of the `parameter_descriptor` message that we are currently using to store all the values, which makes the build method appropriate.

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
CursedRock17: > Not everything that creates a parameter is derived from `rclcpp::Node`; for example `rclcpp_lifecycle::LifecycleNode` doesn't.
> 
Currently in this pull request I have set up two implementations for the `DeclareParameter` function for both of those sited nodes, after digging around the codebase I saw an implementation [here](https://github.com/alsora/rclcpp/blob/1ce4727ec83e3989122bd5e6a76bad844590dd81/rclcpp/include/rclcpp/parameter_client.hppL#113). Would it be better to follow in the form of template based programming like you had made in that pull request in order to increase viability of the function?

In our case an example would be :
```
template<typename ParameterType, typename NodeT> 

DeclareParameter(ParameterType default_value, std::shared_ptr<NodeT> required_node){
    // Implementation
}

```

CursedRock17:
How important are tests to a simple builder class object? I was debating on adding tests but didn't know what they were going to look like.

fujitatomoya (commented):
@CursedRock17 i think this is good enhancement, sorry we did not allocate time for the review. Do you plan to open another PR or find something missing in here? just checking.

CursedRock17:
I was planning on opening this up as an extension in the rcl_interfaces library as `rclpy` wouldn't be able to use this, but I think my closing of the branch was a little too quick. Should I instead reopen this one and do something similar for rclpy?

rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp
```diff
@@ -0,0 +1,77 @@
...
+  // Describes the instances in a parameter_description object
+  ParameterDescription & SetName(const std::string & name);
+  ParameterDescription & SetType(std::uint8_t type);
+  ParameterDescription & SetDescriptionText(const std::string & description);
+  ParameterDescription & SetAdditionalConstraints(const std::string & constraints);
+  ParameterDescription & SetReadOnly(bool read_only);
+  ParameterDescription & SetDynamicTyping(bool dynamic_typing);
+  ParameterDescription & SetFloatingPointDescriptionRange(float min = 0.0f, float max = 1.0f,
+    float step = 0.0f);
+  ParameterDescription & SetIntegerDescriptionRange(int min = 0, int max = 1, int step = 0);
```
alsora, line 54: minor: I don't like having default values for these `SetFloatingPointDescriptionRange` and `SetIntegerDescriptionRange` APIs.
I would expect users to always set them all

rclcpp/src/rclcpp/parameter_descriptor_wrapper.cpp
```diff
@@ -0,0 +1,93 @@
...
+  float step)
+{
+  parameter_descriptor.floating_point_range.resize(1);
+  parameter_descriptor.floating_point_range.at(0).from_value = min;
+  parameter_descriptor.floating_point_range.at(0).to_value = max;
+  parameter_descriptor.floating_point_range.at(0).step = step;
+  return *this;
+}
+
+ParameterDescription & ParameterDescription::SetIntegerDescriptionRange(int min, int max, int step)
```
alsora, line 84: should this method and `SetFloatingPointDescriptionRange` check the parameter type?
e.g. if the parameter is a string you shouldn't try to set these

alsora:
Overall looks good, left some minor comments.
We need unit-tests before proceeding

alsora (approved):
Looks good with green CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20688/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15007/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21382/)

fujitatomoya (commented):
before checking the implementation, i have a question. CC: @CursedRock17

rclcpp/test/rclcpp/test_parameter_service.cpp
```diff
@@ -119,3 +120,32 @@ TEST_F(TestParameterService, describe_parameters) {
     EXPECT_EQ(0u, parameter_descs.size());
   }
 }
+
+TEST_F(TestParameterService, parameter_descriptor) {
+  {
+    rclcpp::ParameterDescription param_description;
```
fujitatomoya, line 126: I was expecting we were going to have the class `ParameterDescription` templated with `ParameterType` or something else to support syntactic sugar like following? So that it would be easier to create descriptor for the user application? current code almost provides the same experience with `rcl_interfaces::msg::ParameterDescriptor`?

```suggestion
    rclcpp::ParameterDescription<PARAMETER_INTEGER>
    param_description(
      "int_parameter",
      "description",
      "constraints",
      false,
      start, end, step
      );
   node->declare_parameter(rclcpp::ParameterValue{1}, param_description.build());
```

above is just a pseudo cod. i might be mistaken on this, probably we want to get feedback from @wjwwood @gbalke .

alsora:
@CursedRock17 can you rebase your branch on top of Rolling? I think that it's too old and it's causing CI failures

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20710/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15023/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21398/)

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20774/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15072/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21454/)

CursedRock17:
I should've commented this earlier, but I didn't expect the changes to pass as the includes don't work. I needed to include the header file in `rclcpp/rclcpp.hpp` since this feature isn't included in any other file.

These are the  steps I followed to get the includes working, I don't think there's anything wrong now: 

- Add new file to `rclcpp/CMakeLists.txt`
- Add Independent file to `rclcpp/rclcpp.hpp`
- Declare the [Header Guards](https://github.com/ros2/rclcpp/blob/db7342a94b5f2a9e5c0be37f767ade19778b15ef/rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp#L15)
- Put everything into a [namespace](https://github.com/ros2/rclcpp/blob/db7342a94b5f2a9e5c0be37f767ade19778b15ef/rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp#L33) like `rclcpp`
- Add [Visibility Controls](https://github.com/ros2/rclcpp/blob/db7342a94b5f2a9e5c0be37f767ade19778b15ef/rclcpp/include/rclcpp/parameter_descriptor_wrapper.hpp#L43)

ahcorde:
Pulls: ros2/rclcpp#2179
Gist: https://gist.githubusercontent.com/ahcorde/bed35067b11ed1d48fc7b8153b5cfcbe/raw/c4da2bcd15992d6d51d192a2e8490cd208ac1fc3/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17648
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26351/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19999/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6050/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26091/)

ahcorde:
@fujitatomoya do you mind to take a look ?

Issue #1473, opened 2020-11-22
get_action_names_and_types not exposed in NodeGraph


ipa-hsd:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Client library (if applicable):
  - rclcpp

---
While working on the `action_bridge` in https://github.com/ros2/ros1_bridge/pull/256, I realized that `get_action_names_and_types` is not exposed in [NodeGraph](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node_interfaces/node_graph.hpp), even though actions are considered to "first-class citizens". I am not quite sure how is `rcl` "used" wihtin `rclcpp`, because the function is part of the [rcl_action API](https://github.com/ros2/rcl/blob/dd1923b5e6dd73d04c6226148e642de2fcdf84c3/rcl_action/src/rcl_action/graph.c#L206). 

Is there any reason for this? Are there plans on doing this in the future?

fujitatomoya:
i think that the followings need to be exposed to `rclcpp`, (`rclpy` does have them already)

- rcl_action_get_client_names_and_types_by_node
- rcl_action_get_server_names_and_types_by_node
- rcl_action_get_names_and_types

emersonknapp:
This came up in rosgraph working group - noticed the gap when trying to implement "Describe Node", that rclcpp didn't have the Action graph inspection APIs exposed anywhere. Pinging @skyegalaxy to bring to client library WG for discussion

PR #2188, opened 2023-05-07
Exposed rcl_actions to the NodeGraph class

CursedRock17:
This pull request is meant to address issue #1473 which needs rcl_action functions exposed to the NodeGraph class. As mentioned:

- rcl_action_get_client_names_and_types_by_node
- rcl_action_get_server_names_and_types_by_node
- rcl_action_get_names_and_types

were all meant to be exposed from [rcl](https://github.com/ros2/rcl/blob/a01d0951adbf346fe0acd4e4be24b896bc95bf50/rcl_action/src/rcl_action/graph.c#L135) to rclcpp, which based on the previous comment meant to make methods similar to rclpy. So each of the functions developed in this PR all follow a similar format of gaining access to the original rcl_action methods with a rcl_ret_t type and return a map of all the names and types.


fujitatomoya:
@CursedRock17 thanks for the PR, i will take a look. in the meantime, can you address DCO error? https://github.com/ros2/rclcpp/pull/2188/checks?check_run_id=13298073729

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
```diff
@@ -67,6 +67,10 @@ class NodeGraph : public NodeGraphInterface
   std::map<std::string, std::vector<std::string>>
   get_service_names_and_types() const override;
 
+  RCLCPP_PUBLIC
+  std::map<std::string, std::vector<std::string>>
+  get_action_names_and_types(rclcpp::Node action_node) const override;
```
fujitatomoya, line 72: `action_node` is never used? and `NodeGraph` possesses the `node_base_` in private scope?

```suggestion
  get_action_names_and_types() const override;
```

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
```diff
@@ -79,6 +83,20 @@ class NodeGraph : public NodeGraphInterface
...
+    rclcpp::Node node,
+    const std::string & node_name,
+    const std::string & namespace_) const override;
+
+  RCLCPP_PUBLIC
+  std::map<std::string, std::vector<std::string>>
+  get_action_server_names_and_types_by_node(
+    rclcpp::Node node,
+    const std::string & node_name,
+    const std::string & namespace_) const override;
```
fujitatomoya, line 98: same here

```suggestion
  get_action_server_names_and_types_by_node(
    const std::string & node_name,
    const std::string & namespace_) const override;
```

fujitatomoya, line 98: same here?

```suggestion
  std::map<std::string, std::vector<std::string>>
  get_action_server_names_and_types_by_node(
    const std::string & node_name,
    const std::string & namespace_) const override;
```

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -215,6 +256,66 @@ NodeGraph::get_client_names_and_types_by_node(
   return services_and_types;
 }
 
+std::map<std::string, std::vector<std::string>>
+NodeGraph::get_action_client_names_and_types_by_node(
+    rclcpp::Node node,
+    const std::string & node_name,
+    const std::string & namespace_) const
+{
+  rcl_names_and_types_t action_names_and_types = rcl_get_zero_initialized_names_and_types();
```
fujitatomoya, line 265: i think this needs to be finalized with `rcl_names_and_types_fini` before exiting from this function scope to avoid memory leak.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -215,6 +256,66 @@ NodeGraph::get_client_names_and_types_by_node(
...
+      actions_and_types[action_name].emplace_back(action_names_and_types.types[i].data[j]);
+    }
+  }
+
+  return actions_and_types;
+}
+
+std::map<std::string, std::vector<std::string>>
+NodeGraph::get_action_server_names_and_types_by_node(
+    rclcpp::Node node,
```
fujitatomoya, line 291: never used?

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -215,6 +256,66 @@ NodeGraph::get_client_names_and_types_by_node(
...
+    const std::string & node_name,
+    const std::string & namespace_) const
+{
+  rcl_names_and_types_t action_names_and_types = rcl_get_zero_initialized_names_and_types();
```
fujitatomoya, line 295: i think this needs to be finalized with rcl_names_and_types_fini before exiting from this function scope to avoid memory leak.

fujitatomoya (changes requested):
in addition to comments, it would be ideal to add the test cases to `rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp`.

rclcpp/src/rclcpp/node_interfaces/node_graph.cpp
```diff
@@ -96,6 +96,47 @@ NodeGraph::get_topic_names_and_types(bool no_demangle) const
...
+    std::string action_name = action_names_and_types.names.data[i];
+    for (size_t j = 0; j < action_names_and_types.types[i].size; ++j) {
+      actions_and_types[action_name].emplace_back(action_names_and_types.types[i].data[j]);
+    }
+  }
+
+  ret = rcl_names_and_types_fini(&action_names_and_types);
+  if (ret != RCL_RET_OK) {
+    // *INDENT-OFF*
+    throw std::runtime_error(
```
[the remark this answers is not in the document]
CursedRock17: I think we should stick with throw_from_rcl_error to keep consistency because there are multiple other instances where `throw_from_rcl_error` is utilized for checking the reliability of `rcl`

CursedRock17:
> in addition to comments, it would be ideal to add the test cases to `rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp`.

The most recent commit I pushed contained the tests for the `rcl_action_get_names_and_types` I followed format for other similar methods in the file. Was that the way that made testing the most efficient for those types of functions, if so, I can replicate with so more tests for the other new functions.

fujitatomoya:
@CursedRock17 thanks for iterating and being responsive. i will take a look tomorrow, in the meantime can you address DCO error?

fujitatomoya (changes requested):
Can you also add the test cases to check if it can actually get names and types? for example, https://github.com/ros2/rclcpp/blob/fce021b14967f691340a6e90b15bdaa37a2008bd/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp#L195

rclcpp/include/rclcpp/node_interfaces/node_graph.hpp
```diff
@@ -67,6 +67,10 @@ class NodeGraph : public NodeGraphInterface
   std::map<std::string, std::vector<std::string>>
   get_service_names_and_types() const override;
 
+  RCLCPP_PUBLIC
+  std::map<std::string, std::vector<std::string>>
+  get_action_names_and_types() const override;
```
fujitatomoya, line 72: I think this cannot be built since there is no virtual function exists. Could you update `NodeGraphInterface` accordingly?

fujitatomoya: see https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/887/console

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -708,3 +708,164 @@ TEST_F(TestNodeGraph, get_info_by_topic_endpoint_info_array_fini_error)
...
+{
+  auto mock = mocking_utils::patch_and_return(
+    "lib:rclcpp", rcl_action_get_names_and_types, RCL_RET_ERROR);
+    rclcpp::exceptions::RCLError);
+  auto mock_info_array_fini = mocking_utils::patch_and_return(
+    "lib:rclcpp", rcl_names_and_types_fini, RCL_RET_ERROR);
+  EXPECT_THROW(
+    node_graph()->get_action_names_and_types("action", false),
+    rclcpp::exceptions::RCLError);
+
```
fujitatomoya, line 727: Needs brackets at the end of this test case?

```suggestion
}

```

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -708,3 +708,164 @@ TEST_F(TestNodeGraph, get_info_by_topic_endpoint_info_array_fini_error)
...
+    "lib:rclcpp", rcl_names_and_types_fini, RCL_RET_ERROR);
+  EXPECT_THROW(
+    node_graph()->get_action_names_and_types("action", false),
+    rclcpp::exceptions::RCLError);
+}
+
+TEST_F(TestNodeGraph, get_action_client_names_and_types_by_node)
+{
+  auto client =
+    node()->create_client<test_msgs::srv::Empty>("node1_service");
```
fujitatomoya, line 740: i think we need to create `action` but `service` for this test case.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -708,3 +708,164 @@ TEST_F(TestNodeGraph, get_info_by_topic_endpoint_info_array_fini_error)
...
+    }
+  }
+
+  EXPECT_TRUE(services_of_node1.find("/ns/node1_service") != services_of_node1.end());
+  EXPECT_FALSE(services_of_node2.find("/ns/node1_service") != services_of_node2.end());
+}
+
+TEST_F(TestNodeGraph, get_action_client_names_and_types_by_node_rcl_error)
+{
+  auto client = node()->create_client<test_msgs::srv::Empty>("node1_service");
```
fujitatomoya, line 773: same here, this is creating `service` but `action`. `rclcpp_action::create_client` should be called instead.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -708,3 +708,164 @@ TEST_F(TestNodeGraph, get_info_by_topic_endpoint_info_array_fini_error)
...
+    "lib:rclcpp", rcl_names_and_types_fini, RCL_RET_ERROR);
+  RCLCPP_EXPECT_THROW_EQ(
+    node_graph()->get_action_client_names_and_types_by_node(node_name, node_namespace),
+    std::runtime_error(
+      "failed to get action service names and types by node: error not set"));
+}
+
+TEST_F(TestNodeGraph, get_action_client_names_and_types_by_node_rcl_fini_error)
+{
+  auto client = node()->create_client<test_msgs::srv::Empty>("node1_service");
```
fujitatomoya, line 787: same here.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
```diff
@@ -708,3 +708,164 @@ TEST_F(TestNodeGraph, get_info_by_topic_endpoint_info_array_fini_error)
...
+{
+  auto handle_goal = [](const GoalUUID &, std::shared_ptr<const test_msgs::msg::Empty::Goal>){
+    return rclcpp::GoalResponse::ACCEPT_AND_DEFER;
+  };
+  auto handle_cancel = [](std::shared_ptr<test_msgs::msg::Empty> {
+    return rclcpp::CancelResponse::ACCEPT;
+  };
+  auto handle_accepted = [](std::shared_ptr<test_msgs::msg::Empty>){};
+
+  const server = node()->create_server<test_msgs::msg::Empty(
```
fujitatomoya, line 805: same here, this is `service`.

fujitatomoya (changes requested):
@CursedRock17 can you check building and testing with your local environment? i cannot build or test this PR... thanks in advance.

CursedRock17:
> @CursedRock17 can you check building and testing with your local environment? i cannot build or test this PR... thanks in advance.

Yes I can test, I can do it on both my computer and Docker for Ubuntu 22.04 correct?

fujitatomoya:
> Docker for Ubuntu 22.04 correct?

yes, as long as the userland is Ubuntu 22.04, i believe there should be no problem for this PR verification.

rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp
CursedRock17: While this makes sense, I cannot include the rclcpp_action part of the library in the rclcpp section, otherwise a linking error is thrown because rclcpp_action uses rclcpp, so this would result in an infinite loop. We need to both, have these functions in the node_graph_interface section, and utilize the rclcpp_action commands.

CursedRock17: One thing that we could do is add specifically the tests to the `rclcpp_action` part of the `rclcpp` [library](https://github.com/ros2/rclcpp/tree/rolling/rclcpp_action/test) with a new file along the names of test_action_calls. Since rclcpp_action has to include rclcpp in its [cmake file](https://github.com/ros2/rclcpp/blob/rolling/rclcpp_action/CMakeLists.txt#L7) and already has the ability to include `rclcpp_action::create_server` there would be no linking issues. The problem is just the location of these files because the `rcl_action` wrapper interfaces are in the `rclcpp/nodeinterfaces` section of the codebase, which makes this location a bit awkward for naming purpose. The functions like `get_action_types_and_names` make the most sense where they are now, but it seems the tests have to be moved simply because including rclcpp_action in rclcpp isn't possible.

clalancette: This conversation is making me question whether we want to expose this as part of `rclcpp` at all.  That is, I wonder if it would make more sense to move all of this to `rclcpp_action` (possibly with a new `NodeActionGraph` class), which avoids this whole problem.  That said, I haven't looked at this implementation closely yet, so maybe this doesn't make much sense.

CursedRock17: I would have to agree with moving this entire implementation to rclcpp_action. The folder is already quite slim with files and seems to be a better place to place wrapped rcl_action functions. While keeping all three of these functions in the node_graph.cpp section would be simpler; the fact that we would need to move tests to a different part of the project entirely showcases a design flaw.

CursedRock17: I'm now less confident in moving the entire implementation, the NodeGraph monitors a node given as `rclcpp::node_interfaces::NodeBaseInterface` . If we're monitoring the one node, and that node can provide a range of topics, services, and actions, then we split this implementation into two pieces, would the user lose the ability to keep all of that data in one node? Or would we just have to pass that same `clcpp::node_interfaces::NodeBaseInterface` into the constructor of this proposed `NodeActionGraph`, in order to view all of this data (assuming the user is fine with creating both a `NodeGraph` and a `NodeActionGraph`

CursedRock17: I'm now less confident in moving the entire implementation of `NodeGraph` into `rclcpp_action` as `NodeActionGraph`. While we are only using it for action based activities, it's not action specific, simply just an rcl_action wrapper, we then have to move all of the other general overrides to this new class. This will add another ~500 lines that we probably could have avoided.

 Additionally, if the user has some long-running process using an action server/client, while they have a small publisher/subscriber maybe keeping track of the environment, they won't be able to track these processes from one `NodeGraph`. If we did move the action side of things over to a new class, they have to create both a `NodeGraph` and `NodeActionGraph`, so if the user were to call `get_graph_event` and [gain another user](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/node_interfaces/node_graph.cpp#L558) on the `NodeGraph`, the `NodeActionGraph` wouldn't see that change, thus won't be able to accurately track the processes running.

Issue #2073, opened 2022-12-29
[feature] easier user control over action goal id's


Aposhian:
#### Feature description
Currently `async_send_goal` simply calls `generate_goal_id`, which is just random bytes. It would be nice to have the following capabilities nicely supported:
- generate goal ID's that are well-formed UUIDs
- supply specific goal ID's (that come from other semantic contexts)

#### Implementation considerations
Here are a few ideas:
- provide an overloaded signature for `async_send_goal` that also includes a goal uuid parameter. Or as a default argument
- Add a goal UUID to the send goal options.
- Be able to set a "Generator" class/function on the action client class that generates UUIDs according to a certain scheme.

fujitatomoya:
> generate goal ID's that are well-formed UUIDs

this makes sense to me.If that is UUID, well-formed UUID would be useful especially for debugging or finding the UUID format via grep from the log.

> supply specific goal ID's (that come from other semantic contexts)

can you elaborate the use case for this extension? I am not really sure what could be beneficial for user application to generate and set the UUID for action goal...

Aposhian:
The benefit for setting a user-supplied goal ID is that if that goal ID carries semantics from somewhere else in the application.

Also, I found a simple workaround to support this sort of use case for now, simply by subclassing `rclcpp_action::Client<>` and then overriding `generate_goal_id`, since it is declared `virtual` in `rclcpp_action::ClientBase`. So that maybe is sufficient to support the "user supplied goal id" use case.

However, I think it would be good to just change the default implementation to generate UUIDv4 ids instead of random bytes.

CursedRock17:
When it comes to making the UUIDs more enforced, issue #1999 was solved to make the strings more compliant with RFC [4122](https://datatracker.ietf.org/doc/html/rfc4122). Was that the sort of thing we were looking to do make make these more set towards UUID V4, if so that part doesn't need to change. Otherwise, would we be aiming at making a method to be cryptographically ensured - the way boost does it - or more so based on a frequency timer that the RFC documentation goes through. To extent on this, @hidmic seems to have started some sort of method in the generate_goal_id method, was there any progress made in this?

As for allowing the user to provide input to the goal_id function, I created an override method which also takes a const GoalUUID& argument. This is a good idea because it will ensure that there are not duplicate UUIDs by some absurd offchance, if the user is using the same generator.

PR #2193, opened 2023-05-17
Beginning overloaded functions

CursedRock17:
This pull request is directly linked to issue #2073 which discusses the volubility of the the goal id in the async_send_goal. So far the thread has come to the conclusion that well-formed uuids should be implemented in accordance to [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt) which may have been solved in pull request #1999, by converting to the appropriate string. Otherwise we must follow a representation of bit shifting seen in many [examples](https://www.cryptosys.net/pki/Uuid.c.html).

Since the RFC issue may have already been addressed to formulate valuable UUIDs at version 4, I then added an overloaded function of `async_send_goal` which allows the user to add their own UUID. This is a double edged sword, depending on the way the user generates a UUID. Technically speaking, if the user has another UUID it could use a generator different from the one called in the method, so it is slightly dangerous to have both of these generators mix. This is only if the UUIDs end up being reserved already, but there so many combinations that this shouldn't happen, still a slight chance though. Conversely it can act very smoothly when the user continuesly uses their own generator for whatever reason they choose. In this case I feel that the unlikely chance of duplication in exchange for user comfortability justifies a overloaded method.


clalancette:
We discussed this a bit, and we are not sure that this is a feature we should have.  In particular, the GoalUUID is really an implementation detail of the actions, so supplying them from the outside doesn't seem to be something we should do.

Instead, if you really want a UUID that you have control of in your actions, then it makes more sense to add a UUID field to your action message.  That way you'll have complete control of it, and not mess with the internal details of actions.

All of that said, we are still open to hearing more about the intended use case.

CursedRock17:
I'm not opposed to anything that you brought for keeping this method out of the library, I was just seeing if that implementation would be beneficial. We can close this PR and the issue attached if it's just stale and doesn't need anymore discussion. 

One this that might still need to be brought up is [the goal id generation](https://github.com/ros2/rclcpp/blob/rolling/rclcpp_action/src/client.cpp#L380-#L391), where there were talks about doing something better for UUID generation. I don't know if [this pull request](https://github.com/ros2/rclcpp/pull/1999/files) was necessarily what they were looking for.

PR #2005, opened 2022-08-26
Modifies timers API to select autostart state

Voldivh:
Signed-off-by: Voldivh <eloyabmfcv@gmail.com>

This PR addresses the issue #1980. Basically it gives the timers API a variable that enables or disables the start of the timer at initialization. The variable `autostart` is defaulted to `True` in order to avoid any breaks in current use of the API. In order to start the timer if `autostart = false` you can simply call the method `timer->reset()`. This PR is dependant on modifications on this [PR at the RCL package](https://github.com/ros2/rcl/pull/1004) and should be merged after that one.

PR #2194, opened 2023-05-17
add start_canceled bool flag when creating a timer

alsora:
This can be used to make sure that a timer is canceled before being registered with a node, avoiding the risk of race conditions if an executor tried to immediately trigger it

Consider a situation where a user creates a timer but wants to activate it only at a later moment.

```
auto my_executor = std::make_shared<SingleThreadedExecutor>();
std::thread spin_thread([my_executor]() {
    my_executor->spin();
});
my_executor->add_node(my_node);
auto timer = rclcpp::create_timer(my_node, ....);
timer->cancel();
```

As soon as the timer object is created, the `rclcpp::create_timer` function would add it to the node timers interface, which would result in notifying the executor about this new entity.
In case of thread starvation and/or race conditions, the executor may try to trigger the timer before it has been cancelled.

This could result in problems or crashes if the users were expecting that timer to not be called at that time.

This PR adds a boolean flag to the `rclcpp::create_timer` functions in order to cancel the timer in a safe way before adding it to the node timers interface.

### DIFF
--- a/rclcpp/include/rclcpp/create_timer.hpp
+++ b/rclcpp/include/rclcpp/create_timer.hpp
@@ -90,7 +90,8 @@ create_timer(
   rclcpp::Clock::SharedPtr clock,
   rclcpp::Duration period,
   CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  bool start_canceled = false)
 {
   return create_timer(
     clock,
@@ -98,7 +99,8 @@ create_timer(
     std::forward<CallbackT>(callback),
     group,
     node_base.get(),
-    node_timers.get());
+    node_timers.get(),
+    start_canceled);
 }
 
 /// Create a timer with a given clock
@@ -109,7 +111,8 @@ create_timer(
   rclcpp::Clock::SharedPtr clock,
   rclcpp::Duration period,
   CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  bool start_canceled = false)
 {
   return create_timer(
     clock,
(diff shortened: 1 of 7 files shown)


rclcpp/include/rclcpp/create_timer.hpp
```diff
@@ -90,15 +90,17 @@ create_timer(
   rclcpp::Clock::SharedPtr clock,
   rclcpp::Duration period,
   CallbackT && callback,
-  rclcpp::CallbackGroup::SharedPtr group = nullptr)
+  rclcpp::CallbackGroup::SharedPtr group = nullptr,
+  bool start_canceled = false)
```
fujitatomoya, line 94: how about the following? just a suggestion.

```suggestion
  bool auto_start = true)
```

rclcpp/test/rclcpp/test_timer.cpp
```diff
@@ -319,6 +319,23 @@ TEST_P(TestTimer, test_failures_with_exceptions)
   }
 }
 
+/// Create a timer that starts as canceled
```
fujitatomoya, line 322: It would be probably better to add the test case that

1. create canceled `timer`.
2. reset `timer`. (activate)
3. check timer callback is fired.

fujitatomoya (commented):
additional test case would be nice to have.

tgroechel:
Is this a duplicate of https://github.com/ros2/rclcpp/pull/2005?

alsora:
Thanks for catching it!
Should we close this one and resurrect the other one? Or should I just apply the requested changes here?

tgroechel:
> Should we close this one and resurrect the other one?

From a quick glance, it looks like the linked issue is part of a somewhat connected "web" of timer related issues. I'll do a quick search+summary and update this comment linking the web together. 

My only initial concern with keeping the other issue is it seems* a bit stale / older and this one could likely be merged more quickly.

> Or should I just apply the requested changes here?

 With either choice (keeping this or the other), I agree with https://github.com/ros2/rclcpp/pull/2194#discussion_r1197195218 that `bool auto_start = true` is slightly easier to understand/intuitive as it is more semantically inline with/idiomatic of default behavior imo.

fujitatomoya:
@alsora how about closing this one and working on [these open PRs](https://github.com/ros2/rclcpp/pull/2194#issuecomment-1573733617)? sorry i should have told you before...

PR #2195, opened 2023-05-17
[WIP] Extend DynamicSubscription to support serialized messages

emersonknapp:


rclcpp/src/rclcpp/dynamic_subscription.cpp
```diff
@@ -41,10 +41,17 @@ void DynamicSubscription::handle_message(std::shared_ptr<void> &, const rclcpp::
 }
 
 void DynamicSubscription::handle_serialized_message(
-  const std::shared_ptr<rclcpp::SerializedMessage> &, const rclcpp::MessageInfo &)
+  const std::shared_ptr<rclcpp::SerializedMessage> & message,
+  const rclcpp::MessageInfo & message_info)
 {
-  throw rclcpp::exceptions::UnimplementedError(
-          "handle_serialized_message is not implemented for DynamicSubscription");
+  using PlainCallback = std::function<void (std::shared_ptr<SerializedMessage>)>;
```
wjwwood, line 47: Maybe, to avoid duplication of the signature, we should define these before the variant, and then use them in the variant as well as here.

rclcpp/include/rclcpp/dynamic_subscription.hpp
```diff
@@ -56,10 +63,7 @@ class DynamicSubscription : public rclcpp::SubscriptionBase
     rclcpp::dynamic_typesupport::DynamicMessageTypeSupport::SharedPtr type_support,
     const std::string & topic_name,
     const rclcpp::QoS & qos,
-    std::function<void(
-      rclcpp::dynamic_typesupport::DynamicMessage::SharedPtr,
-      const rosidl_runtime_c__type_description__TypeDescription &
```
wjwwood, line 61: Will be interesting to see if not having the description in the callback always will be annoying, but I like it not being in the argument list myself. Ideally it would be accessible through the `DynamicMessage`, but that might not make sense, not sure.

rclcpp/include/rclcpp/dynamic_subscription.hpp
```diff
@@ -159,11 +165,10 @@ class DynamicSubscription : public rclcpp::SubscriptionBase
 private:
   RCLCPP_DISABLE_COPY(DynamicSubscription)
 
+  static bool is_callback_serialized(const rclcpp::AnyDynamicSubscriptionCallback &);
```
wjwwood, line 168: I think this could/should be `constexpr`.

rclcpp/include/rclcpp/dynamic_subscription.hpp
methylDragon: Unfortunately the descriptions aren't baked into DynamicMessage, since they're technically distinct concepts (and DynamicMessageTypes can be created at runtime **without** descriptions.)

The only major tradeoff to me of not having the description in the callback is feature discoverability (which was the reason I decided to put it into all callback signatures in the first place.) But if that's a non-issue I am very much for including getters in the DynamicSubscription class.

methylDragon:
Thanks Emerson for helping with this! 🙇

I think other than the points William listed above, we just need to add getters to the DynamicSubscription to get the `DynamicMessageType` (probably as const or copy) and the type description.

rclcpp/src/rclcpp/dynamic_subscription.cpp
emersonknapp: Yeah that's a little less wordy, done

rclcpp/include/rclcpp/dynamic_subscription.hpp
emersonknapp: I was thinking, if you need the description in the callback, you can provide it yourself via closure, or member access in the case of class methods:

```
auto dynamic_typesupport = DynamicMessageTypeSupport::make_shared(type_description);
auto sub = create_dynamic_subscription(
  node, 
  dynamic_typesupport, 
  topic_name, 
  qos, 
  [type_description](shared_ptr<SerializedMessage> message) { 
    // has access to type_description 
  });
```

Providing the description to the callback as a parameter does allow for reuse of the same callback for several dynamic subscriptions to different types, I guess - but I don't see the real life use case for that.

emersonknapp: This seems the same to me as e.g. `topic_name` and `topic_type_namei` - which you also don't have access to in subscription callbacks, right?

methylDragon:
There are some further discussions about the C++ classes on the parent PR:
- https://github.com/ros2/rclcpp/pull/2176#issuecomment-1552190126
- https://github.com/ros2/rclcpp/pull/2176#issuecomment-1552220254

rclcpp/include/rclcpp/dynamic_subscription.hpp
wjwwood: Right, it's the same, but only if the `DynamicMessage` can be useful without the description. Currently you cannot list the fields with just it, so you need the description to reasonably use it. Whereas with a message and topic name, you can most often make use of the message without the topic name, so it's fine to ask the user to figure that out with captures or something.

Anyway @methylDragon has a bunch of notes on this.

Issue #1699, opened 2021-06-22
Add take_loaned() method in Subscription class


carlossvg:
Currently, there is no `rclcpp` API exposed in the subscription class to take a loaned message. The only way to use zero-copy on the subscription side is through an executor which uses `rcl_take_loaned_message` directly. See: https://github.com/ros2/rclcpp/blob/7d8b2690ff188672bfe3703d8b3c074b87303931/rclcpp/src/rclcpp/executor.cpp#L598-L633

We should have a `take_loaned()` method or similar to support those cases where the user wants to take a loaned message directly. For example, when using `rclcpp::waitset` and zero-copy. 

This is required for example to support zero-copy using a `rclcpp::waitset` in `performance_test`.

Once this method is implemented I think should update the zero-copy article to add the new APIs: https://design.ros2.org/articles/zero_copy.html

fred-apex-ai:
This would also be required for a node that avoids any copy and modifies input message in place and publishes a message of the same type as it receives.

PR #2197, opened 2023-05-20
Loaned_Message Method

CursedRock17:
This is a Pull Request meant to help issue #1699 by creating a new method called `take_loaned_message` to allow the user the ability to zero-copy passing of loaned_message. There currently wasn't many other options other than those described in the original issue, so this is a good start. I assumed that the solution will be to make a wrapper around the `rcl_take_loaned_message` function, to provide rclcpp the ability to handle these, and allowing a user to pass a loaned_message via reference or pointer.


rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -361,6 +361,22 @@ class Subscription : public SubscriptionBase
...
+  // Take the loaned message
+  /*
+   * \param loaned_message : message that the user wants to take
+   */
+  void
+  take_loaded_message(
+    void * loaned_message,
+    const rclcpp::MessageInfo & message_info) override
+  {
+      rcl_ret_t ret = rcl_take_loaned_message(
```
alsora, line 373: The return value shouldn't be ignored.
I recommend to apply the same logic hat is currently found in the executor

and then have the executor use this function rather than dealing with the `rcl_take_loaned_message` directly

CursedRock17: That's what I assumed the process should look like so we have a more extendable method for the user, while still following DRY principles.

CursedRock17:
On line [615](https://github.com/ros2/rclcpp/blob/7d8b2690ff188672bfe3703d8b3c074b87303931/rclcpp/src/rclcpp/executor.cpp#L615) of executor.cpp there's a TODO message that is based on LoanedMessage support. If my previous commits were a correct implementation, does this mean the `Subscription` class has method when LoanedMessage is extened to support subscriptions, or would that mean that the `LoanedMessage` class takes its own `Subscription`?

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -617,21 +617,9 @@ Executor::execute_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
...
-                nullptr);
-              if (RCL_RET_SUBSCRIPTION_TAKE_FAILED == ret) {
-                return false;
-              } else if (RCL_RET_OK != ret) {
-                rclcpp::exceptions::throw_from_rcl_error(ret);
-              }
-              return true;
-            },
-            [&]() {subscription->handle_loaned_message(loaned_msg, message_info);});
+            [&]() {subscription->take_loaned_message(loaned_msg, message_info);},
```
iuhilnehc-ynos, line 620:
```suggestion
            [&]() {return subscription->take_loaned_message(loaned_msg, message_info);},
```

rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -234,6 +234,13 @@ class SubscriptionBase : public std::enable_shared_from_this<SubscriptionBase>
   void
   handle_loaned_message(void * loaned_message, const rclcpp::MessageInfo & message_info) = 0;
 
+  RCLCPP_PUBLIC
+  virtual
+  bool
+  take_loaned_message(
+    void * loaned_message,
+    const rclcpp::MessageInfo & message_info) = 0;
```
iuhilnehc-ynos, line 242:
```suggestion
    rclcpp::MessageInfo & message_info) = 0;
```

and need doc which can refer to `SubscriptionBase::take_type_erased`

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -361,6 +361,29 @@ class Subscription : public SubscriptionBase
...
   }
 
+  // Take the loaned message
+  /*
+   * \param loaned_message : message that the user wants to take
+   */
+  bool
+  take_loaned_message(
+    void * loaned_message,
+    const rclcpp::MessageInfo & message_info) override
```
iuhilnehc-ynos, line 371:
```suggestion
    rclcpp::MessageInfo & message_info) override
```

`message_info` is lost in the doc.

iuhilnehc-ynos (commented):
tiny comments

rclcpp/include/rclcpp/subscription_base.hpp
iuhilnehc-ynos: Can we implement it in SubscriptionBase without a virtual method?

rclcpp/include/rclcpp/subscription.hpp
alsora: the doc is also missing information about the bool returned value and the fact that this function can throw

rclcpp/include/rclcpp/subscription_base.hpp
CursedRock17: When the method is extended into subscription_base.cpp and subscription_base.hpp we won't need to make it virtual unless we extend it into the subscription.hpp and cpp files, respectively, correct? Additionally I know that we use doxygen for the cpp documentation, is there an exact place for it be rewritten for these methods?

alsora (approved):
Looks good to me.
It wouldn't be bad to add unit-tests, although this code was already existing somewhere else

rclcpp/include/rclcpp/subscription_base.hpp
```diff
@@ -234,6 +234,18 @@ class SubscriptionBase : public std::enable_shared_from_this<SubscriptionBase>
   void
   handle_loaned_message(void * loaned_message, const rclcpp::MessageInfo & message_info) = 0;
 
+// Take the loaned message
+/*
+ * Similar to take_type_erased, but this is a zero copy method that doesn't rely on the pointers on the stack, and doesn't use messages[out]
```
fujitatomoya, line 239: IMO, This depends on middleware, it could be zero copy that that does not promise via interface. we can only say that incoming messages can be loaned from middleware to the user application.

rclcpp/src/rclcpp/subscription_base.cpp
```diff
@@ -301,6 +301,25 @@ SubscriptionBase::can_loan_messages() const
   return rcl_subscription_can_loan_messages(subscription_handle_.get());
 }
 
+bool
+SubscriptionBase::take_loaned_message(
```
fujitatomoya, line 305: I believe that it must return the loaned message to the middleware, unless that will be memory leak or probably middleware cannot reuse the memory for the publisher side. probably message type should be templated and `LoanedMessage` object should be returned, so that it can return the loaned memory to the middleware via dtor?

fujitatomoya (changes requested):
several [uncrustify failures](https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/893/), can you address them?

@wjwwood can you take a look at this? i might be missing something.

rclcpp/src/rclcpp/subscription_base.cpp
CursedRock17: sorry for the delay, so to clarify you think we should just using a `LoanedMessage` object instead of the `void * loaned_message`? This then allows us to keep the rmw middleware in the function - `take_loaned_message` but allow the `LoanedMessage loaned_message` to deallocate the message in it's dtor [here](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/loaned_message.hpp#L156).

Originally when I made this pull request I was trying to match the [design docs for loaned messages](http://design.ros2.org/articles/zero_copy.html) where the user could either:
> - Manage the memory for the message instance themselves, providing a reference to it, into which the middleware should fill the data.
> - Take one or more loaned messages from the middleware and return the loans later.

I assumed the rclcpp API would just act as a wrapper to rcl, and would just have this object as a placeholder, which intends zero-copy.

CursedRock17: Additionally while I think `rclcpp::LoanedMessage` is much better than a `void * loaned_message`, `LoanedMessage` is currently based on the `Publisher` side of things:
```
  LoanedMessage(
    const rclcpp::PublisherBase & pub,
    std::allocator<MessageT> allocator)
```
Also see [TODO](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/executor.cpp#L631-L632), therefore we would have to revamp the LoanedMessage class to support subscriptions. I wouldn't mind working on supporting subscriptions in that class, especially since I see multiple issues and PRs based on `LoanedMessage` and intra-process messaging. Should we implement `Subscription` in `LoanedMessage` first, or should we push what we have then move to that?

sgf201: @fujitatomoya The change here does not affect the return of loanedbuf to the middleware.It is just act as a wrapper to rcl,code logic is the same as before

sgf201 (approved):

rclcpp/src/rclcpp/subscription_base.cpp
CursedRock17: @sgf201, I don't think this PR is where it needs to be to be honest. I need to take a bit more time to refine these changes to a better level. I'm going to make this a draft until I have it at the point where is needs to be.

PR #2200, opened 2023-05-29
Added Missing Std exceptions to some files

Imeguras:
On Arch Linux the build was failing because some includes were missing this should fix the problem

PR #2204, opened 2023-06-02
Added missing stdexcept on some files

Imeguras:
A fixed, rolling oriented version of https://github.com/ros2/rclcpp/pull/2200

### DIFF
--- a/rclcpp/include/rclcpp/any_service_callback.hpp
+++ b/rclcpp/include/rclcpp/any_service_callback.hpp
@@ -156,7 +156,7 @@ public:
     const std::shared_ptr<rmw_request_id_t> & request_header,
     std::shared_ptr<typename ServiceT::Request> request)
   {
-    TRACEPOINT(callback_start, static_cast<const void *>(this), false);
+    TRACETOOLS_TRACEPOINT(callback_start, static_cast<const void *>(this), false);
     if (std::holds_alternative<std::monostate>(callback_)) {
       // TODO(ivanpauno): Remove the set method, and force the users of this class
       // to pass a callback at construnciton.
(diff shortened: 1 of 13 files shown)


Imeguras:
Note this PR is with humble in mind and i believe that some missing files in rolling may be critical as i've noticed a lot of issues when installing humble on arch and had to solve a lot of them by hand

rclcpp/src/rclcpp/logging_mutex.hpp
```diff
@@ -17,6 +17,7 @@
 
 #include <memory>
 #include <mutex>
+#include <stdexcept>
```
alsora, line 20: this is not used here, but rather in `logging_mutex.cpp`, so we should include it there

rclcpp/src/rclcpp/signal_handler.hpp
```diff
@@ -19,6 +19,7 @@
 #include <csignal>
 #include <mutex>
 #include <thread>
+#include <stdexcept>
```
alsora, line 22: why is this needed here?

rclcpp/src/rclcpp/logging_mutex.hpp
Imeguras: I placed it there to keep all the includes in the header file although i can certainly roll back

rclcpp/src/rclcpp/signal_handler.hpp
Imeguras: i've must've added it due to humble issues. which are not present in the rolling version

fujitatomoya: > i've must've added it due to humble issues

this is build error? can you share the compile error for this?

rclcpp/src/rclcpp/logging_mutex.hpp
clalancette: Yeah, we should only put includes where they are used.  If they are not directly used in here (and this one is not), let's not put it in.

rclcpp/src/rclcpp/signal_handler.hpp
clalancette: Same thing here.  Since we don't use `stdexcept` in this header file, let's not add it here.

clalancette (changes requested):
As one other general comment on this PR, we should keep the includes in alphabetical order.

PR #2215, opened 2023-06-16
Do not crash Executor when send_response fails due to client failure.

kghost:
Related to https://github.com/ros2/ros2/issues/1253

It is not sane that a faulty client can crash our service Executor, as discussed in the referred issue, if the client is not setup properly, send_response may return RCL_RET_TIMEOUT, we should not throw an error in this case.

### DIFF
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -486,6 +486,14 @@ public:
   {
     rcl_ret_t ret = rcl_send_response(get_service_handle().get(), &req_id, &response);
 
+    if (ret == RCL_RET_TIMEOUT) {
+      RCLCPP_WARN(
+        rclcpp::get_node_logger(node_handle_.get()).get_child("rclcpp"),
+        "failed to send response: %s",
+        rcl_get_error_string().str);
+      rcl_reset_error();
+      return;
+    }
     if (ret != RCL_RET_OK) {
       rclcpp::exceptions::throw_from_rcl_error(ret, "failed to send response");
     }
--- a/rclcpp/include/rclcpp/service.hpp
+++ b/rclcpp/include/rclcpp/service.hpp
@@ -489,7 +489,7 @@ public:
     if (ret == RCL_RET_TIMEOUT) {
       RCLCPP_WARN(
         rclcpp::get_node_logger(node_handle_.get()).get_child("rclcpp"),
-        "failed to send response: %s",
+        "failed to send response (timeout): %s",
         rcl_get_error_string().str);
       rcl_reset_error();
       return;


rclcpp/include/rclcpp/service.hpp
```diff
@@ -486,6 +486,14 @@ class Service
...
     rcl_ret_t ret = rcl_send_response(get_service_handle().get(), &req_id, &response);
 
+    if (ret == RCL_RET_TIMEOUT) {
+      RCLCPP_WARN(
+        rclcpp::get_node_logger(node_handle_.get()).get_child("rclcpp"),
+        "failed to send response: %s",
+        rcl_get_error_string().str);
+      rcl_reset_error();
+      return;
+    }
```
fujitatomoya, line 496: The reason that i did not merge https://github.com/ros2/rcl/pull/1048 is how client should behave. It is easy to drop the request on the service side, but how we can manage the client still waiting on the request? I believe that needs to be considered before pushing this change into client libraries. any thoughts?

fujitatomoya (commented):
@kghost thanks for creating PR on this issue.

rclcpp/include/rclcpp/service.hpp
kghost: This is another problem, and should be solved on the response subscription part.

Currently the service is suffering from a serious vulnerable, any network service should be robust enough to protect itself from malicous client attacks, or else itself can be a DoS target.

So I do consider this being a top priority issue rather than fixing the client part.

fujitatomoya: @kghost thanks for iterating.

actually i was concerned that pending request would be the new problem for client side. but i believe that client can manage the timeout using user callback to check the future, or `spin_until_future_complete`. and then, client can remove the pending request from the list.

i would like to hear more opinions. @ros2/team 

- `rclcpp` (or `rclpy` client) should not retry the sending the response since underlying RMW implementation has done that already. (since it returns `timeout`, RMW implementation is responsible to retry.)
- if it gets `timeout`, specific exception should be thrown. Not just printing `warning` message here? this is the one of the exception, so that user application should be able to receive that? (this requires all dependent packages need to be adjusted so to avid the original issue.)

> malicous client attacks, or else itself can be a DoS target.

Even we fix this problem with this PR, this will be still a security problem? once it comes to this kind of requirement, i would use security feature, https://docs.ros.org/en/rolling/Concepts/About-Security.html ?

clalancette: We talked about this briefly in the internal meeting.  The basic problem is that we have no way to report the failure of `send_response` to the user.  However, I also agree that we should not crash the executor in this case.  So I think the best we can do at the moment is to recognize the failure, print a warning to the console, and continue on.

rclcpp/include/rclcpp/service.hpp
```diff
@@ -486,6 +486,14 @@ class Service
   {
     rcl_ret_t ret = rcl_send_response(get_service_handle().get(), &req_id, &response);
 
+    if (ret == RCL_RET_TIMEOUT) {
+      RCLCPP_WARN(
+        rclcpp::get_node_logger(node_handle_.get()).get_child("rclcpp"),
+        "failed to send response: %s",
```
fujitatomoya, line 492: just a suggestion, it would be clear for user that timeout happened?

```suggestion
        "failed to send response because of timeout: %s",
```

fujitatomoya (approved):

fujitatomoya:
@kghost thank you for being patient on this.

I will be checking the following and then start CI all together.

- underlying tier 1 rmw implementation returns timeout when it detects the timeout? (unless, it will generate the exception)
- `rclpy` counterpart PR against this `rclcpp`.

fujitatomoya:
> underlying tier 1 rmw implementation returns timeout when it detects the timeout? (unless, it will generate the exception)

see https://github.com/ros2/rcl/pull/1048#issuecomment-1613837733

> rclpy counterpart PR against this rclcpp.

see https://github.com/ros2/rclpy/pull/1136

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19079/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13606/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19825/)

rclcpp/include/rclcpp/service.hpp
clalancette: Yes, agreed that it would be clearer.

clalancette (changes requested):
This needs the one change inline, and also likely needs to be rebased.  Once those are done, I'm happy to approve this.

milidam:
I am experiencing the same kind of crash using the fastrtps RMW implementation with ros2 humble.
However, the proposed patch does not avoid the crash, as the `RMW_RET_TIMEOUT` error code [returned by the fastrtps' `__rmw_send_response`](https://github.com/ros2/rmw_fastrtps/blob/humble/rmw_fastrtps_shared_cpp/src/rmw_response.cpp#L150) is changed to `RCL_RET_ERROR` in [`rcl_send_response`.](https://github.com/ros2/rcl/blob/humble/rcl/src/rcl/service.c#L315)
The same is done in rolling : https://github.com/ros2/rcl/blob/rolling/rcl/src/rcl/service.c#L403

Am I missing somthing? or are there any other patches required for it to work?

clalancette:
> Am I missing somthing? or are there any other patches required for it to work?

Yes, you also need https://github.com/ros2/rcl/pull/1048

fujitatomoya:
@kghost are you willing to work on this? see https://github.com/ros2/rclcpp/pull/2215#pullrequestreview-1516578021

kghost:
> @kghost are you willing to work on this? see https://github.com/ros2/rclcpp/pull/2215#pullrequestreview-1516578021

no problem!

fujitatomoya:
@kghost thank you very much for the effort, can you address DCO error? see https://github.com/ros2/rclcpp/pull/2215/checks?check_run_id=14906339355

fujitatomoya:
@kghost this also need to be rebased on rolling, see https://ci.ros2.org/job/ci_linux/19079/console CI failed on building `rclcpp`.

kghost:
@fujitatomoya Yes, it is already rebased on rolling branch.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19133/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13693/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19869/)

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19156/)

windows failure is unrelated.

kghost:
@clalancette This is ready to merge.

rclcpp/include/rclcpp/service.hpp
```diff
@@ -486,6 +486,14 @@ class Service
   {
     rcl_ret_t ret = rcl_send_response(get_service_handle().get(), &req_id, &response);
 
+    if (ret == RCL_RET_TIMEOUT) {
+      RCLCPP_WARN(
+        rclcpp::get_node_logger(node_handle_.get()).get_child("rclcpp"),
```
alsora, line 491: why not using `node_logger_` here?

rclcpp/include/rclcpp/service.hpp
```diff
@@ -486,6 +486,14 @@ class Service
...
+        "failed to send response (timeout): %s",
```
alsora, line 492: I suggest to add the service name in the log message

mauropasse:
Should changes in [`rclcpp/include/rclcpp/service.hpp`](https://github.com/ros2/rclcpp/pull/2215/files#diff-b8d1d4683b3d8d68d748a323b49d58eae6e699156197fc62a0785632f757b007) be also applied also here?:

https://github.com/ros2/rclcpp/blob/c0d72c3ee0ac806379b49ed10cf17364a4bd443a/rclcpp_action/src/server.cpp#L340
https://github.com/ros2/rclcpp/blob/c0d72c3ee0ac806379b49ed10cf17364a4bd443a/rclcpp_action/src/server.cpp#L482
https://github.com/ros2/rclcpp/blob/c0d72c3ee0ac806379b49ed10cf17364a4bd443a/rclcpp_action/src/server.cpp#L539
https://github.com/ros2/rclcpp/blob/c0d72c3ee0ac806379b49ed10cf17364a4bd443a/rclcpp_action/src/server.cpp#L672

Related to this topic, we're in conversations to set the timeout in DDS from an XML, so we can increase the timeout before returning `RCL_RET_TIMEOUT`. If the client is discovered before it, it will get the response from the service. Otherwise, it'll hang forever waiting a response (at least, that happens with actions).

rclcpp/include/rclcpp/service.hpp
clalancette: I think the problem with trying to do that here is that not all of the constructors have access to the service name.  So this may not be possible to do without changing all of the constructors, or exposing `service_name` in `rcl_service_t`.  Personally, I think this is nice but I wouldn't hold up this PR for that.

rclcpp/include/rclcpp/service.hpp
clalancette: I'm not sure that there is access to the node logger in this class right now.

clalancette (changes requested):
I think that @mauropasse makes a good point and we should do the same thing for the action side of things as well.  Once that is fixed I'm happy to approve this PR.

fujitatomoya:
> I think that @mauropasse makes a good point and we should do the same thing for the action side of things as well. Once that is fixed I'm happy to approve this PR.

@clalancette @mauropasse can we manage this with another issue before adding the implementation in this PR? i would like to check if that happens in detail.

fujitatomoya:
> > I think that @mauropasse makes a good point and we should do the same thing for the action side of things as well. Once that is fixed I'm happy to approve this PR.
> 
> @clalancette @mauropasse can we manage this with another issue before adding the implementation in this PR? i would like to check if that happens in detail.

https://github.com/ros2/ros2/issues/1462

fujitatomoya:
@clalancette @mauropasse friendly ping.

mauropasse:
> @fujitatomoya : i would like to check if that happens in detail.

What do you mean with "that pappens in detail"?

I know it happens with actions since that's where we observed this issue first, and come with the temporary solution of increasing the timeout.

With regards of applying the action side of these fixes in a different PR, I'm OK with that.

fujitatomoya:
@mauropasse i was not sure if this problem can be actually reproducible with actions, thanks for the information.

> With regards of applying the action side of these fixes in a different PR, I'm OK with that

👍 @clalancette what do you think?

clalancette:
> 👍 @clalancette what do you think?

Yeah, that's fine.  Let's keep this one just about the service server, and we can do the follow-up with https://github.com/ros2/ros2/issues/1462 for action servers.

I'll go ahead and re-review this.

rclcpp/include/rclcpp/service.hpp
fujitatomoya: @kghost can you address this one? and then i think we are good to go.

fujitatomoya:
@kghost since you are not responding, i did create an another PR for this to proceed. I keep the all commits that you made. it would be appreciated if you can review it. see https://github.com/ros2/rclcpp/pull/2276

fujitatomoya:
@clalancette @kghost i will go ahead to close this, instead can we proceed with https://github.com/ros2/rclcpp/pull/2276?

PR #2218, opened 2023-06-19
If executor has been cancelled should not spin

mauropasse:
This simple code would hang forever:
```
  ExecutorType executor; // Any executor
  auto executor_thread = std::thread([&](){ executor.spin();});
  executor.cancel();
  executor_thread.join();
```
If `executor.cancel()` gets called before `executor.spin()`, cancel won't have any effect and the executor will keep spinning.

In this PR I add an extra bool to check if the executor was cancelled before blocking in `spin()`

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -637,6 +637,9 @@ protected:
   /// Spinning state, used to prevent multi threaded calls to spin and to cancel blocking spins.
   std::atomic_bool spinning;
 
+  /// Flag specifying that the executor has been cancelled and should not spin
+  std::atomic_bool executor_canceled{false};
+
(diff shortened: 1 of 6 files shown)


alsora:
I'm not convinced by this approach.
The following code should result in the executor spinning
```
auto executor = std::make_shared<executor>();
executor->cancel();
.... do some stuff ...
executor->spin(); // This should spin, the cancel that we called before should not affect it
```

fujitatomoya (commented):
@mauropasse thanks for opening PR. I am with @alsora on this. executor should be able to spin after `cancel`.

mauropasse:
Makes sense, a `cancel()` call should only affect an spinning executor, so the user should make sure the executor is spinning before cancelling.

PR #2224, opened 2023-06-22
Add new node interface TypeDescriptionsInterface to provide GetTypeDescription service

emersonknapp:
Part of ros2/ros2#1159

<s>Depends on ros2/rcl#1052</s>

Linked with https://github.com/ros2/system_tests/pull/520

Defines and implements new node interface `TypeDescriptionsInterface`, which provides the `~/get_type_description` service from ros2/rcl#1052.

Uses parameter `start_type_description_service` to start the service - defaulting to true!

This will not be a trivial backport to Iron - I propose moving forward with this as the "correct thing to build", then the backport can dump most of it into `node.cpp` for Iron.

Issue #1142, opened 2020-05-28
Start using PIMPL extensively in user facing classes, to improve ABI stability


ivanpauno:
Feature request

#### Feature description

We have run many times in the issue of not being able to backport a fix because it would break ABI.
e.g.:

- https://github.com/ros2/rclcpp/pull/1119 is almost impossible to backport.
- https://github.com/ros2/rclcpp/pull/1056

#### Implementation considerations

We should evaluate when the effort matters and where it doesn't.

wjwwood:
I have some concern with the PIMPL pattern being used extensively, due to complexity and performance.

The complexity comes from just having so many more indirections within the code, especially in the case that you need to write a template method but access data via the impl class, which can cause some awkwards scenarios where you need to type erase data.

The performance is just chasing pointers can be expensive if done everywhere. I'd like to consider the "fast PIMPL" pattern, which is a bit odd, but it can result in lots of performance improvements in certain cases:

http://www.gotw.ca/gotw/028.htm

ivanpauno:
FastPIMPL looks good to me.

There's also a lot of places were we can take advantage of an indirection we are already doing and have ABI stability for free. e.g.:

```cpp
std::shared_ptr<Publisher<MessageT>>
create_publisher(...)
...
class Publisher : public PublisherBase
```

We are already returning a pointer to the object and never the object directly.
Because we're inheriting from [enable_shared_from_this](https://github.com/ros2/rclcpp/blob/bba9dce253d98a31a3069218504d26c4be9fb8b8/rclcpp/include/rclcpp/publisher_base.hpp#L55), we actually shouldn't construct a `Publisher` or `PublisherBase` object directly.

With some slight modifications to the current API, we can provide ABI stability:

1. Make `PublisherBase` constructors protected.
  Document that classes that inherit from `PublisherBase` must have private (or protected) constructors and a friend factory function that returns a `shared_ptr`.
2. Delete copy constructor/copy assignment of `PublisherBase` (or make them private).
3. Do 1. and 2. for `Publisher`.
4. Ensure we don't provide methods that when inlined can break ABI stability. i.e.: don't explicetly refer to members in methods that can be inlined, use getters instead.

1 2 3 will make impossible for an user to hold a `Publisher`/`PublisherBase` object directly.
They will have to hold always a `std::shared_ptr<PublisherBase>` ...

4 ensures we don't break ABI stability because of inlined methods.
[...]

PR #2228, opened 2023-06-27
Add a pimpl inside rclcpp::Node for future distro backports

emersonknapp:
This came up in some current Iron backports, that there is no good place to hide any new members. My understanding is that this is not used for the majority of features for execution performance (avoid one indirection), however a small amount is not an overall performance trend. This escape hatch should make for some better backports in future distros.


emersonknapp:
@wjwwood we discussed this briefly, maybe you have further thoughts

fujitatomoya (commented):
@emersonknapp do you happen to have an example implementation detailed feature in mind (or plan) to backport iron? generally it would be helpful if we can keep the ABI during backport with this, but i am not sure what kind of feature can rely on this concealment.

emersonknapp:
Yes, I will want to backport https://github.com/ros2/rclcpp/pull/2224 to Iron. But, there is no good place to put the new `NodeTypeDescriptions` interface object. I will have to make some global variable or something to store it.

This proposal is so that, after Jazzy, I could do that backport cleanly - the new feature in Rolling would change the ABI, but the backport could use the pimpl.

alsora:
Adding this PIMPL in the node class, will definitely allow to easily backport things like https://github.com/ros2/rclcpp/pull/2224, but IMO only a small subset of PRs are adding members to the node class.
If we do this, it would be an important precedent for adding hidden implementation pointers to a lot of other rclcpp classes and I don't think that it would be a good decision.

I think that the backporting effort should focus on bug fixes rather than on new features.

wjwwood:
In my opinion, the PIMPL should not be used except for backports, feature or bugfix, i.e. we should prefer to put new members in the class directly rather than in the pimpl.

> If we do this, it would be an important precedent for adding hidden implementation pointers to a lot of other rclcpp classes and I don't think that it would be a good decision.

I think this is necessary to support backports in an ABI way, which includes bugfixes and not just features. So based on that, I would personally like to see pimpl pointers in most/all the classes, but have them unused except for backporting. This adds some conceptual and memory overhead (about one pointer per class), but has real value. There have been several times in the last few years where we've said something along the lines of "if we had a pimpl we could fix this easily, but instead..." and the "instead..." is always something less ideal.

Anytime this topic comes up, I also think about about performance, and preferring to avoid the use of the pimpl except when required definitely helps, but there's also the "fast pimpl" pattern, which has less capacity for ABI breaks, but is faster so long as you don't exceed it: http://www.gotw.ca/gotw/028.htm

Also, @clalancette did we have a general tracking issue for adding pimpl's to classes in ROS 2? I know we talked about it on several occasions.

emersonknapp:
I agree with @wjwwood points above, additionally

> I think that the backporting effort should focus on bug fixes rather than on new features.

I disagree with this. While I think that bugfixes and stability should be the _first_ and _top_ priority, I think that ROS distros have frequently been too "stuck" in that users don't get the benefit of useful new features within the lifetime (up to 5 years!) of a distribution. There are many users who will not upgrade distros on their product unless they absolutely have to, but frequently come requesting minor features that exist in the mainline. However, this is potentially a pretty big conversation, and I think this PR is not the right place for it.

clalancette:
> Also, @clalancette did we have a general tracking issue for adding pimpl's to classes in ROS 2? I know we talked about it on several occasions.

Yeah, we've discussed it several times.  Ivan opened an issue a few years ago: https://github.com/ros2/rclcpp/issues/1142

clalancette:
> > I think that the backporting effort should focus on bug fixes rather than on new features.

> I disagree with this. ...

> However, this is potentially a pretty big conversation, and I think this PR is not the right place for it.

Yeah, agreed that this is a large conversation and out of scope here.  If you'd like to discuss it further, I'll suggest adding an item to the ROS 2 weekly agenda.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1593,6 +1593,9 @@ class Node : public std::enable_shared_from_this<Node>
   const rclcpp::NodeOptions node_options_;
   const std::string sub_namespace_;
   const std::string effective_namespace_;
+
+  class NodeImpl;
+  std::shared_ptr<NodeImpl> hidden_impl_;
```
wjwwood, line 1598: Could/should this be a `unique_ptr`?

emersonknapp: No because of the "sub-node" protected constructor, which is a copy constructor that wants to share all the data

emersonknapp:
Based on discussion today, we move forward with this. Maybe the pattern gets extended to the node components - but this is a definitely useful start. Conditions for pimpl (mentioned in docstring comment on its implementation):
* Left empty in Rolling, used only for backports
* Left unallocated in Rolling, so no new allocation is performed

rclcpp/include/rclcpp/node.hpp
clalancette: I'm resolving this as I think we agreed that this should be a `shared_ptr`.

clalancette, line 1598: That said, I have a slightly different comment to make.  I think we should add a small comment here as well, something along the lines of:

```suggestion
  // This member is meant to be a place to backport features into stable distributions,
  // and new features targeting Rolling should not use this.  See the comment in node.cpp for more information.
  std::shared_ptr<NodeImpl> hidden_impl_{nullptr};
```

clalancette (changes requested):
I've left one really minor comment, otherwise this looks good to me.  I'd like one more look from @wjwwood before we go ahead and merge it, though.

alsora (approved):

rclcpp/include/rclcpp/node.hpp
```diff
@@ -1593,6 +1593,11 @@ class Node : public std::enable_shared_from_this<Node>
   const rclcpp::NodeOptions node_options_;
   const std::string sub_namespace_;
   const std::string effective_namespace_;
+
+  class NodeImpl;
+  // This member is meant to be a place to backport features into stable distributions,
+  // and new features targeting Rolling should not use this.  See the comment in node.cpp for more information.
```
clalancette, line 1599: I suspect that this line, as-is, is going to violate our 100 character length limit.

clalancette (changes requested):

clalancette (approved):

emersonknapp:
Pulls: ros2/rclcpp#2228
Gist: https://gist.githubusercontent.com/emersonknapp/8942d3eb82c983a5829f6fbdad2e145e/raw/8b261ac93a516e824e6dfc1df08eab979bc8799a/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12366
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19139/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13698/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19872/)

clalancette:
New CI after the rebase (so we can hopefully merge this):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19331/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13859/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20049/)

PR #2232, opened 2023-07-05
Move always_false_v to detail namespace

nwn:
Since this is a common idiom, especially under this name, we should define the `always_false_v` template within a namespace to avoid conflict with other libraries and user code. This could either be `rclcpp::detail` if it's intended only for internal use or just `rclcpp` if it's intended as a public helper. In this PR, I've initially chosen the former.

### DIFF
--- a/rclcpp/include/rclcpp/any_subscription_callback.hpp
+++ b/rclcpp/include/rclcpp/any_subscription_callback.hpp
@@ -34,14 +34,14 @@
 #include "rclcpp/type_adapter.hpp"
 
 
-template<class>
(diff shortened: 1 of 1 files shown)


clalancette (approved):
This looks good to me, thanks!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19066/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13599/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19818/)

PR #2235, opened 2023-07-07
Remove deprecated header scope_exit.hpp

emersonknapp:
According to the comment it is planned for I-turtle


emersonknapp:
wrong `hub` command

PR #2238, opened 2023-07-17
Stop using constref signature of benchmark DoNotOptimize.

clalancette:
Newer versions of google benchmark (1.8.2 in my case) warn that the compiler may optimize away the DoNotOptimize calls when using the constref version.  Get away from that here by explicitly *not* calling the constref version, casting where necessary.

### DIFF
--- a/rclcpp_lifecycle/test/benchmark/benchmark_lifecycle_node.cpp
+++ b/rclcpp_lifecycle/test/benchmark/benchmark_lifecycle_node.cpp
@@ -103,6 +103,8 @@ BENCHMARK_F(BenchmarkLifecycleNode, get_current_state)(benchmark::State & state)
         std::string("Node's current state is: ") + std::to_string(lifecycle_state.id());
       state.SkipWithError(message.c_str());
     }
+    // Google benchmark 1.8.2 warns that the constref DoNotOptimize signature may be optimized away
+    // by the compiler.  Cast const away to ensure we don't get that problem (and warning).
     benchmark::DoNotOptimize(const_cast<rclcpp_lifecycle::State &>(lifecycle_state));
     benchmark::ClobberMemory();
   }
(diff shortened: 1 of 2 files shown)


rclcpp_lifecycle/test/benchmark/benchmark_lifecycle_node.cpp
```diff
@@ -162,18 +162,18 @@ BENCHMARK_F(BenchmarkLifecycleNode, transition_valid_state)(benchmark::State & s
...
     }
-    const auto & inactive =
+    const rclcpp_lifecycle::State & inactive =
       node->trigger_transition(lifecycle_msgs::msg::Transition::TRANSITION_DEACTIVATE);
     if (inactive.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE) {
       state.SkipWithError("Transition to inactive state failed");
     }
-    benchmark::DoNotOptimize(active);
-    benchmark::DoNotOptimize(inactive);
+    benchmark::DoNotOptimize(const_cast<rclcpp_lifecycle::State &>(active));
```
wjwwood, line 175: Might want a comment as to why you're doing this, if I saw a `const_cast` like this, I'd be very curious, or maybe just drop it if I was editing this code.

wjwwood (approved):
lgtm, other than a suggested comment.

rclcpp_lifecycle/test/benchmark/benchmark_lifecycle_node.cpp
clalancette: Yeah, that is an excellent point.  I'll go ahead and add in some comments.

clalancette: Fixed in 2af0448.

Issue #689, opened 2019-04-08
Add API for queering if intraprocess comm is enabled in pub/subs


ivanpauno:
#### Feature description
Add a method to `Publisher` and `Subscription` classes, which allows queering if intraprocess communication is enabled or not.

```cpp
bool
is_intra_process_comm_enabled() const;
```

#### Reasoning
In #674 an extra argument has been added to `create_publisher` and `create_subscription` methods in `Node` class, which allows choosing if intraprocess communication is enabled or not with more granularity. The argument default value is `NodeDefault` which enables it or not depending on the node settings. That was the previous behavior.
Now, It's no way of checking if a publisher or subscriber has the intraprocess comm activated.
Also, it doesn't exist any test which sets the intraprocess comm with pub/sub granularity. This will allow to easily add a test.

#### Implementation considerations
Both publisher and subscriber has a bool storing the intraprocess subscription state, so we have to just return it.
I think it's also a good idea to add a test with intraprocess comm being set with pub/sub granularity.

wjwwood:
Makes sense to me.

PR #2239, opened 2023-07-19
Check intraprocess_comm state

CursedRock17:
This is a pull request targeting issue #689 (* Warning this is an older issue and the codebase has changed since then *). The point of that issue was just to expose the boolean for the intraprocess_comm subscription state, which looks like this:
```
enum class IntraProcessSetting
{
  /// Explicitly enable intraprocess comm at publisher/subscription level.
  Enable,
  /// Explicitly disable intraprocess comm at publisher/subscription level.
  Disable,
  /// Take intraprocess configuration from the node.
  NodeDefault
};
```
but both the Publisher and Subscription have `protected` bools: `intra_process_is_enabled_` and `use_intra_process_` respectively, which showcase if state can be used for intraprocess communication.

While this code was just meant to address the problem in the issue, the `create_publisher` and `create_subscription` functions have changed in the 4 years it's been since the original issue that the issue this PR addresses, was meant to fix. They no longer have `IntraProcessSetting use_intra_process_comm = IntraProcessSetting::NodeDefault);` in their function defintions, which makes adding the tests a bit more difficult for the context of the problem since it's not the same. I think the functions are still perfectly useful in the base classes and for querying, just don't know how to go about testing now, since they aren't complex functions.


fujitatomoya (changes requested):
@CursedRock17 thanks for addressing this. could you add the test if those methods return the expected state?

you can refer to https://github.com/ros2/rclcpp/blob/bc435776a257fcf76c5b0124bec26f6824342e34/rclcpp/test/rclcpp/test_publisher.cpp#L299-L313, and add the following test. (pseudo code)

```cpp
TEST_F(TestPublisher, intra_process_comm_state) {
  initialize();
  // intra-process comm is explicitly disabled
  {
    rclcpp::PublisherOptionsWithAllocator<std::allocator<void>> options;
    options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;
    auto publisher = node->create_publisher<test_msgs::msg::Empty>("topic", 10, options);
    EXPECT_FALSE(publisher->is_intra_process_comm_enabled());
  }
  ... and so on
```

rclcpp/test/rclcpp/test_publisher.cpp
```diff
@@ -296,6 +296,20 @@ TEST_F(TestPublisher, basic_getters) {
   }
 }
 
+TEST_F(TestPublisher, intra_process_comm_state_publisher) {
+  initialize();
+  // intra-process comm is explicitly disabled
+  rclcpp::PublisherOptionsWithAllocator<std::allocator<void>> options;
+  options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;
+  auto publisher = node->create_publisher<test_msgs::msg::Empty>("topic", 10, options);
+  EXPECT_FALSE(publisher->is_intra_process_comm_enabled());
```
fujitatomoya, line 305: should it have the scope for each test to make sure the publisher and option is created as we expect? (the same comment goes to subscription.)

```suggestion
  {
    rclcpp::PublisherOptionsWithAllocator<std::allocator<void>> options;
    options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;
    auto publisher = node->create_publisher<test_msgs::msg::Empty>("topic", 10, options);
    EXPECT_FALSE(publisher->is_intra_process_comm_enabled());
  }
```

rclcpp/test/rclcpp/test_publisher.cpp
```diff
@@ -296,6 +296,20 @@ TEST_F(TestPublisher, basic_getters) {
   }
 }
 
+TEST_F(TestPublisher, intra_process_comm_state_publisher) {
```
fujitatomoya, line 299: what about adding the test with `IntraProcessSetting::NodeDefault`? (this goes to subscription as well.)

rclcpp/test/rclcpp/test_publisher.cpp
CursedRock17: Sure that seems to be the way many of the other tests are implemented, just didn't know the extent of that. I'll get that done.

rclcpp/test/rclcpp/test_publisher.cpp
CursedRock17: Sure it's possible to add a test, according the [default node options](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/node_options.hpp#L46) `rclcpp::NodeOptions` would return `false` for the intra-process state. The [init method](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/test/rclcpp/test_publisher.cpp#L47-L50) for the `node` is just sets up a basic, default `NodeOptions`. So I could add the one test, for the way it's setup now, and just expect false. Then, if wanted I could create another `initialize` method which would change the default. Personally I think the latter is unnecessary, but the former is still valid.  Thoughts?

fujitatomoya: yes, i think so. so that we can detect the test or default behavior change for this.

CursedRock17:
Friendly Ping, was that `NodeDefault` Test enough, or should I extend it further?

PR #2246, opened 2023-07-27
Add spin_all shortcut

tonynajjar:
Adds spin_all to `executors.hpp` to be accessible as such: 

```
  auto node = std::make_shared<MyNode>();
  rclcpp::spin_all(node, 1000ms);
```

Should I also add `spin_once`?

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -297,6 +297,21 @@ public:
   virtual void
   spin_some(std::chrono::nanoseconds max_duration = std::chrono::nanoseconds(0));
 
+  /// Add a node, complete all immediately available work exhaustively, and remove the node.
+  /**
+   * \param[in] node Shared pointer to the node to add.
+   */
+  RCLCPP_PUBLIC
+  void
+  spin_node_all(
+    rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node,
+    std::chrono::nanoseconds max_duration);
+
+  /// Convenience function which takes Node and forwards NodeBaseInterface.
+  RCLCPP_PUBLIC
+  void
+  spin_node_all(std::shared_ptr<rclcpp::Node> node, std::chrono::nanoseconds max_duration);
+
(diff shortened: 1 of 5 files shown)


mjcarroll:
Thanks, I think that it makes sense to have `spin_all` in the global free-function API.  

On the other hand, `spin_once` probably doesn't make much sense due to the way that executors work.  In the case of `spin_once`, the executor would be created, populated with available work, one step of work (eg a single subscription) would be executed, and then the executor would be destroyed by going out of scope.  In some cases, this will be fine, as checking for available work has no side effects, but we cannot guarantee that every `Waitable` has the same safety.  In those cases, the act of populating the internal executor queue could mark those waitables as not ready without ever actually servicing them.

fujitatomoya (commented):
lgtm, test would be ideal.

tonynajjar:
> lgtm, test would be ideal.

sure, but I don't see a specific test for `rclcpp::spin` and `rclcpp:spin_some`, can you point me to them or if they don't exist give some more guidance where that test should be?

tonynajjar:
> > lgtm, test would be ideal.
> 
> sure, but I don't see a specific test for `rclcpp::spin` and `rclcpp:spin_some` to extend on, can you point me to them or if they don't exist give some more guidance where that test should be?

I've added a couple of tests, please let me know if more are needed and which ones

fujitatomoya (approved):

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19305/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13835/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20025/)

mjcarroll:
@tonynajjar I think there are some issues with the tests you added on aarch64.  Do you think you can take a look?

tonynajjar:
@mjcarroll should be fixed, can you rerun CI?

mjcarroll:
`Rpr` failures are unrelated here.

Issue #2242, opened 2023-07-20
ActionClient Crash: Taking data from action client but nothing is ready


jmachowinski:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries
- Version or commit hash:
  - iron
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Start action server
- server shall accept goal, but not terminate it

Start action client
- send goal

wait

#### Expected behavior
no crash

#### Actual behavior
after some time we get the exception
`terminate called after throwing an instance of 'std::runtime_error'
    what():  Taking data from action client but nothing is ready
`

#### Additional information
Using a multi threaded spinner, but the default callback group, I guess this one is mutual exclusive

Yadunund:
Could you prepare a minimal example that one could build and run to reproduce the error? It would also provide clarity on how various callbacks are defined.

jmachowinski:
I tried to create a minimal example, and figured out, that the cause of this issue was most likely due to deadlock issues in the client or the server. I can not reproduce the issue as original described in the minimal example.

However I managed reproduce the error now with a slightly different setup.

https://github.com/jmachowinski/rclcppActionDeadlock branch nothingReady

start server
start the client
strg+c the client
directly restart the client

Server:
```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  goal_handle attempted invalid transition from state EXECUTING with event CANCELED, at ./src/rcl_action/goal_handle.c:95
Aborted (core dumped)

``` 

client:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Taking data from action client but nothing is ready
Aborted (core dumped)

``` 


start the client
start the server

client just runs

server directly crashes with 
```
 ./server 
terminate called after throwing an instance of 'std::runtime_error'
  what():  Taking data from action server but nothing is ready
Aborted (core dumped)

``` 

Note, you need to do this multiple times, about 1 time out of 10 the error will trigger.

jmachowinski:
Okay, I had a look at the code now and it it pretty obvious that a race condition is ongoing. 
In order to fix this, I would need some background information on 

```c++
rcl_action_server_wait_set_get_entities_ready
``` 
or repectitive

```c++
rcl_wait
``` 

From the documentation of this function, it is unclear to me if it may return multiple events to be ready at once.
[...]

PR #2248, opened 2023-07-28
fix: Fixed race condition in action server between is_ready and take_…

jmachowinski:
…data and execute

Some background information: is_ready and take_data are guaranteed to be called in sequence without interruption from another thread. while execute is running, another thread may also call is_ready.

The problem was, that goal_request_ready_, cancel_request_ready_, result_request_ready_ and goal_expired_ were accessed and written from is_ready and execute.

This commit fixed this by only using the mentioned variables in is_ready and take_data. execute now only accesses the given pointer and works on this.

Note, this fixes parts of #2242

### DIFF
--- a/rclcpp_action/include/rclcpp_action/server.hpp
+++ b/rclcpp_action/include/rclcpp_action/server.hpp
@@ -22,6 +22,7 @@
 #include <unordered_map>
 #include <utility>
 
+#include "action_msgs/srv/cancel_goal.hpp"
 #include "rcl/event_callback.h"
 #include "rcl_action/action_server.h"
 #include "rosidl_runtime_c/action_type_support_struct.h"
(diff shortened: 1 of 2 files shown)


clalancette:
Can you please target this to the `rolling` branch first?  Our development practice is to get everything in `rolling` first, then backport it to stable distributions as necessary.  Thanks.

jmachowinski:
I'll create a new merge request, should I leave this one just open ?

clalancette:
I'll close this one for now.  Given the nature of this change, and the fact that we must maintain API/ABI in stable distributions, I think the final form that we would put into `iron` is going to be significantly different.

jmachowinski:
API (I only changed private functions) stability would be given, I can modify it to be also ABI compatible.

Note, we as a company are relying on this fix, and would see it merged ASAP. Therefore just tell me what to do, to speed the process up.

Issue #2249, opened 2023-07-28
rclcpp needs specific version of rcl [humble]


stan-guer:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 22.04 
- Installation type:
  - binaries
- Version or commit hash:
  - ROS 2 Humble Hawksbill - Patch Release 4
- DDS implementation:
  - N/A
- Client library (if applicable):
  - rclcpp, rcl

#### Steps to reproduce issue

```
rocker --home --network host -- ros:humble-ros-base
apt update; apt install ros-humble-rclcpp # do not update ros-humble-rcl
# build any simple colcon package
```

#### Expected behavior

No errors

#### Actual behavior

Getting error:
```
/usr/bin/ld: /opt/ros/humble/lib/librclcpp.so: undefined reference to `rcl_clock_time_started'
```

#### Additional information
Ever since PR #2210 has been merged, rclcpp now requires `rcl_clock_time_started` to be available in the rcl library, which was added [here](https://github.com/ros2/rcl/pull/1021). Currently there's no version check for rcl and now that the Patch 4 with this change is out it's possible to update rclcpp without updating rcl:
```
Package: ros-humble-rclcpp
Version: 16.0.5-2jammy.20230721.204813
Priority: optional
Section: misc
Maintainer: Ivan Paunovic <ivanpauno@ekumenlabs.com>
Installed-Size: 3342 kB
Depends: libc6 (>= 2.34), libgcc-s1 (>= 3.3.1), libstdc++6 (>= 12), ros-humble-ament-index-cpp, ros-humble-builtin-interfaces, ros-humble-libstatistics-collector, ros-humble-rcl, ros-humble-rcl-interfaces, ros-humble-rcl-yaml-param-parser, ros-humble-rcpputils, ros-humble-rcutils, ros-humble-rmw, ros-humble-rosgraph-msgs, ros-humble-rosidl-runtime-cpp, ros-humble-rosidl-typesupport-c, ros-humble-rosidl-typesupport-cpp, ros-humble-statistics-msgs, ros-humble-tracetools, ros-humble-ros-workspace
```

I scoured the ros2 infrastructure stuff and can't find where dependency versions are defined or how they're even built [(using cpack or bloom)](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/scripts/release/build_binarydeb.py), so I hope it's the correct place for this issue. 

[...]

PR #2252, opened 2023-07-30
rclcpp: check rcl version [humble]

stan-guer:
Since commit f87d548f95db4165525bd94e4d96b56b3c5478cc, rclcpp requires rcl_clock_time_started from rcl which is only availble since rcl version 5.3.4 for humble

Fixes issue #2249 for humble


rclcpp/package.xml
```diff
@@ -30,7 +30,7 @@
   <build_export_depend>rosidl_typesupport_cpp</build_export_depend>
 
   <depend>libstatistics_collector</depend>
-  <depend>rcl</depend>
+  <depend version_gte="5.3.4">rcl</depend>
```
fujitatomoya, line 33: https://github.com/ros2/rclcpp/commit/52327dd3a3b0fc47446ce86ecaf16fdde4fc4a60 is rclcpp version, so this should be the correct version.

fujitatomoya (commented):
I have never seen the version dependencies in the `package.xml` with the same distribution before, but according to [REP-140](https://www.ros.org/reps/rep-0140.html#build-depend-multiple) this should be the way how to apply version dependencies?

@ros2/team

alsora:
At a quick look, the approach seems syntactically correct.
However if we do it for `rcl`, we should likely do it for a lot of other packages (and keep doing it in the future, I don't think that this is the first time that this happens).

Overall, I think that this is a larger change than it looks, and it's probably not justified without a larger commitment and discussion on how to apply it project-wise.

clalancette:
> However if we do it for `rcl`, we should likely do it for a lot of other packages (and keep doing it in the future, I don't think that this is the first time that this happens).

No, it is definitely not the first time.  That said, it is somewhat rare in stable distributions, as we don't often backport features that require changes across both `rcl` and `rclcpp`.

> Overall, I think that this is a larger change than it looks, and it's probably not justified without a larger commitment and discussion on how to apply it project-wise.

Yes and no (in my opinion).  On the one hand, people taking piece-wise ROS package updates is a definite anti-pattern.  On the other hand, this is a relatively light-weight way to enforce it in the packages.  I don't think we should do this across the distribution (that would be way too much metadata to update all the time), but I'm OK doing it on a case-by-case basis.

fujitatomoya:
> On the one hand, people taking piece-wise ROS package updates is a definite anti-pattern.

basically this could be one of the anti-pattern from distributor perspective, but i do not take this minority. Once it comes to production phase, developer would update the necessary packages only to avoid the instability from other packages in their own CI/CD. in case of this issue, that is `rclcpp`.

if we expect this is anti-pattern for user application, we need to add version dependencies to let the user know.

stan-guer:
> On the one hand, people taking piece-wise ROS package updates is a definite anti-pattern.

To shortly interject here: this was not done by a person, our CI/CD server ran `rosdep install ... --reinstall ...` and one of the steps decided to update rclcpp, since it's a c++ codebase. I understand though that putting versions everywhere where they're missing and maintaining that correctly is daunting. I'm not sure where this lands on the spectrum of stability vs continuous development of ROS2.

clalancette (approved):
This is OK with me.

@nuclearsandwich or @cottsay I just want to double-check with you.  If we put this in the package.xml, will `bloom` pick this up and generate the correct Debian/RPM package dependencies?

PR #2254, opened 2023-08-01
Add lifecycle subscriber to lifecycle nodes

carmiac:
This adds a LifecycleSubscriber class to lifecycle nodes.  In general, it is a mirror of LifecyclePublisher scheme, both for determining whether to process messages and in the factory methods.

There is a [linked PR](https://github.com/ros2/demos/pull/646) that provides a demo and test node.

### DIFF
--- a/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_subscriber.hpp
+++ b/rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_subscriber.hpp
@@ -1,0 +1,84 @@
+// Copyright 2023 Elroy Air, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP_LIFECYCLE__LIFECYCLE_SUBSCRIPTION_HPP_
+#define RCLCPP_LIFECYCLE__LIFECYCLE_SUBSCRIPTION_HPP_
+
+// #include <functional>
+// #include <memory>
+
+#include "rclcpp/subscription.hpp"
+
+namespace rclcpp_lifecycle
+{
+
+/// @brief Child class of rclcpp::Subscription that adds lifecycle functionality.
+/**
+ * This class is a child class of rclcpp::Subscription that adds a lifecycle
+ * check to the callback function. If the node is in an inactive state, the
+ * callback will not be called.
+ */
+template<
+  typename MessageT,
+  typename AllocatorT = std::allocator<void>,
+  typename SubscribedT = typename rclcpp::TypeAdapter<MessageT>::custom_type,
+  typename ROSMessageT = typename rclcpp::TypeAdapter<MessageT>::ros_message_type,
+  typename MessageMemoryStrategyT = rclcpp::message_memory_strategy::MessageMemoryStrategy<
+    ROSMessageT,
+    AllocatorT
+  >>
+class LifecycleSubscription : public SimpleManagedEntity,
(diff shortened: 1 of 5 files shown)


ros-discourse:

https://discourse.ros.org/t/lifecycle-node-improvements/32706/1

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_subscriber.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+  /// Check if we need to handle the message, and execute the callback if we do.
+  void handle_message(
+    std::shared_ptr<void> & message, const rclcpp::MessageInfo & message_info) override
+  {
+    if (!this->is_activated()) {
+      return;
+    }
+    rclcpp::Subscription<MessageT>::handle_message(message, message_info);
+  }
```
fujitatomoya, line 78: with current implementation, if the node is `deactivated`, all messages during that state will be ignored and dropped on the subscription because the message is already taken. i am not sure if this is the expected behavior, since application would expect that the subscription can keep the message up to the specified history depth?

CC: @ros2/team

clalancette: Yes, this is an extremely good point; at this level, we are fetching it out of the RMW but then dropping it on the floor and not calling the user callback.

This gets into a bit of the philosophy behind the lifecycle transitions.  I would expect that when a `LifecycleSubscription` is deactivated, then it is actually not part of the graph at all anymore, i.e. `handle_message` would never be called.  That releases the most resources in the system, though it does have a possible side-effect of a "thundering herd" problem if a bunch of them are activated at once.

All of that said, I wasn't really involved with any of the initial concepts with the lifecycle code, so I don't fully know what the expectation is.  Maybe @wjwwood could chime in here with some ideas?

alsora: I think that this approach provides very little improvement in terms of performance.
By the time you are dropping the message, you already wasted CPU cycles to go through all the ROS and network stack and potentially copying the message multiple times.

I'm not opposed to the change, as long as it's just a starting point from which we can later re-implement lifecycle subscriptions/services to effectively disconnect the entity.

fujitatomoya: > "In the inactive state, any data that arrives on managed topics will not be read and or processed. Data retention will be subject to the configured QoS policy for the topic."

for me, this design makes sense. that is the responsibility for QoS but LifecycleNode state. user can configure the behavior with QoS for LifecycleSubscription.

> By the time you are dropping the message, you already wasted CPU cycles to go through all the ROS and network stack and potentially copying the message multiple times.

agree.

> Specifically, the former is when subscriptions should be created and do discovery, and the latter is when you should start receiving data.

IMO, we need to add a new state for subscription. (the one is `created and discovery`, and the other is `start taking the data from RMW queue(or wait_set is added)`)

> I'd be OK with going forward with this implementation and a TODO that says implementing case 1 is future work.

i am good to go for now, with follow-up issue for this TODO.

dadaroce:
Any progress on this? looks like a really useful feature 😄

fujitatomoya:
AFAIK, no progress. @carmiac are you going to work on this?

i think what we need to do is,

- add lifecycle subscriber test in this PR.
- add `TODO` for `Hold onto the data that arrives before activation, and deliver that on activation`, see https://github.com/ros2/rclcpp/pull/2254#discussion_r1290505034
- and probably  we can add some explanation for https://docs.ros.org/en/rolling/Tutorials/Demos/Managed-Nodes.html?

Amronos:
> AFAIK, no progress. @carmiac are you going to work on this?
> 
> i think what we need to do is,
> 
>     * add lifecycle subscriber test in this PR.
> 
>     * add `TODO` for `Hold onto the data that arrives before activation, and deliver that on activation`, see [Add lifecycle subscriber to lifecycle nodes #2254 (comment)](https://github.com/ros2/rclcpp/pull/2254#discussion_r1290505034)
> 
>     * and probably  we can add some explanation for [docs.ros.org/en/rolling/Tutorials/Demos/Managed-Nodes.html](https://docs.ros.org/en/rolling/Tutorials/Demos/Managed-Nodes.html)?

Hey, can I somehow continue this work and complete what's left?

UPDATE: I have created #2715

Issue #2247, opened 2023-07-28
Prefer custom exceptions to regular runtime_errors


CursedRock17:
#### Feature description
There are a few TODOs in the repository that are marked similar to:
> TODO(jacobperron): use custom exception

They can found by grepping for that format, like [here](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/node_interfaces/node_waitables.cpp#L35) where they are often surrounded by bulky if/else statements that have already been fixed to look better like [this](https://github.com/ros2/rclcpp/blob/rolling/rclcpp_components/src/component_manager.cpp#L115-#L122). There are most likely more of these style statements in the code base, that just haven't been marked, for instance the NodeWaitables example was recognized when the original pull request was made in 2018, but these were the easiest examples to target and start with.

#### Implementation considerations
I guess the main question is what the original authors meant by "custom" because in the [component_manager example](https://github.com/ros2/rclcpp/blob/rolling/rclcpp_components/src/component_manager.cpp#L115-#L122) there was a `ComponentManagerException` class:
```
class ComponentManagerException : public std::runtime_error
{
public:
  explicit ComponentManagerException(const std::string & error_desc)
  : std::runtime_error(error_desc) {}
};
```
It seems that it's a needless wrapper for `std::runtime_error` unless there's a benefit in passing the description via const reference. Would each of these exception follow a similar  format, could a new general class be made, or can we just stick with runtime_error?

clalancette:
> It seems that it's a needless wrapper for `std::runtime_error` unless there's a benefit in passing the description via const reference.

My understanding is that it is not a needless wrapper.  In particular, what it allows you to do is write code that catches rclcpp ComponentManager exceptions, but not all `std::runtime_error` exceptions.  Something like:

```
try {
    component_manager->do_something();
} catch (const rclcpp::ComponentManagerException & cme) {
    rclcpp_component_manager_error_handling();
} catch (const std::runtime_error & e) {
    general_error_handling();
}
```

CursedRock17:
Ok, I have also seen other classes with their own, little Exception handler class. In that case, can I open a PR that deals with all the current cases:
> TODO(person) use custom exception

To get their own Exception classes. Then, if needed deal with other areas that could maybe also use this style?

clalancette:
> In that case, can I open a PR that deals with all the current cases:

[...]

PR #2256, opened 2023-08-01
Adding Missing Group Exceptions

CursedRock17:
Follow up of issue #2247 which is meant to address a range of TODOs specifically for `std::runtime_error` with the description of
```
throw std::runtime_error("Cannot create something, group not in node.");
```
All of the errors follow that same format, so we can create a general class in `exceptions.hpp` that just take a `const std::string &` of whatever `something` is and runs the error in a much cleaner format.

### DIFF
--- a/rclcpp/include/rclcpp/exceptions/exceptions.hpp
+++ b/rclcpp/include/rclcpp/exceptions/exceptions.hpp
@@ -222,7 +222,7 @@ public:
   : std::runtime_error("event already registered") {}
 };
 
-/// Thrown when a group is missing from the node, when it wants to utilize the group.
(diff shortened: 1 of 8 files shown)


fujitatomoya (approved):
lgtm. (probably we would want to apply the same fix to `rclpy` as well)

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19293/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13826/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20013/)

PR #2257, opened 2023-08-02
Change TimeSource associated clocks storage to unordered_set

### LINKED ISSUE #2253: Potential clock duplication in TimeSource
Hello! I was working on an rcl client library and was using `rclcpp` (and `rclpy`) as reference implementation, and when working on `TimeSource` I found a potential issue.

Specifically, when looking at the `TimeSource` implementation for both rclpy and rclcpp, I noticed that for each time source, its associated clocks are stored in a [vector](https://github.com/ros2/rclcpp/blob/22a954e1b00ea2ba810666baa92a3c66fe1a8654/rclcpp/src/rclcpp/time_source.cpp#L188).

Now apart from the premature-optimization-root-of-all-evil issue of having finds be linear in time [here](https://github.com/ros2/rclcpp/blob/22a954e1b00ea2ba810666baa92a3c66fe1a8654/rclcpp/src/rclcpp/time_source.cpp#L107), I believe the fact that clocks are naively appended in `attachClock` would allow duplicates in the [vector](https://github.com/ros2/rclcpp/blob/22a954e1b00ea2ba810666baa92a3c66fe1a8654/rclcpp/src/rclcpp/time_source.cpp#L98). Now the drawback is not the end of the world since it's just that the same clock would be updated more than once for each subscription callback but would it make sense to either:

* Do an `std::find` call in `attachClock` and print a warning + avoid appending if the clock already exists
* Make the attached clocks storage structure a `unordered_set`?

If any of these makes sense I'll be happy to open upstream PRs in both `rclpy` and `rclcpp` since they have the same implementation (rclpy is [here](https://github.com/ros2/rclpy/blob/913afa019b0d60e7255bd84247dbeab4735d96ea/rclpy/rclpy/time_source.py#L116))

luca-della-vedova:
Fixes #2253 

As mentioned in the attached issue, using a `vector` makes it possible for the same clock to be stored multiple times in a `TimeSource`, this PR changes the storage to be `unordered_set` so duplicates are automatically avoided at insertion and removal is constant in time.

I thought a bit about adding a regression test but it wasn't so immediate to me it would be a good idea, the test would have been adding the same clock twice, removing it once then making sure it is not updated anymore. I can add it if you folks think it's useful.
There is also no warning if a user adds a clock that was already added, but that can also be added if needed.

I'll open a matching PR in `rclpy` once this is approved

### DIFF
--- a/rclcpp/src/rclcpp/time_source.cpp
+++ b/rclcpp/src/rclcpp/time_source.cpp
@@ -14,6 +14,7 @@
 
 #include <memory>
 #include <string>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -95,7 +96,7 @@ public:
       }
     }
     std::lock_guard<std::mutex> guard(clock_list_lock_);
-    associated_clocks_.push_back(clock);
+    associated_clocks_.insert(clock);
     // Set the clock to zero unless there's a recently received message
     set_clock(last_time_msg_, ros_time_active_, clock);
   }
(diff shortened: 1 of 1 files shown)


alsora (approved):

clalancette (approved):
This looks good to me.  I'm going to go ahead and run CI on it, but we should wait to merge it until the rclpy equivalent PR is open.

fujitatomoya (approved):

PR #2261, opened 2023-08-02
Add lifecycle timers

carmiac:
This adds lifecycle timers which start when the node activates and cancel when the node deactivates.  There is a demo available at https://github.com/carmiac/demos/blob/lifecycle_timer/lifecycle/src/lifecycle_timer.cpp


ros-discourse:

https://discourse.ros.org/t/lifecycle-node-improvements/32706/1

clalancette (changes requested):
This generally looks good to me, but can you please add some unit tests as well?

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_node.hpp
```diff
@@ -280,6 +294,19 @@ class LifecycleNode : public node_interfaces::LifecycleNodeInterface,
     CallbackT callback,
     rclcpp::CallbackGroup::SharedPtr group = nullptr);
 
+  /// Create a lifecycle timer that uses the node clock to drive the callback.
```
alsora, line 297: we should also have an API to create a timer with a custom clock

clalancette: I'm not sure whether that makes sense.  Sure, the underlying https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/create_timer.hpp has support for that, but the `rclcpp::Node` already has a `Clock` that is accessed via `get_clock`.  Giving the ability to create a timer based off another clock means that this particular timer will (possibly) run at a different rate from the rest of the `Node`.  While I can imagine some uses for that, it is kind of strange.

Further, it doesn't look like we expose that kind of functionality from https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/node.hpp .  And I don't really want `LifecycleNode` to "get ahead" of `Node` (we already have the opposite problem).

To me, adding that in is an orthogonal discussion, and if we are going to do it we should do it for both `Node` and `LifecycleNode`.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_timer.hpp
```diff
@@ -0,0 +1,91 @@
...
+   * \param period The interval at which the timer fires
+   * \param callback The callback function to execute every interval
+   * \param context node context
+   */
+  LifecycleTimer(
+    rclcpp::Clock::SharedPtr clock,
+    std::chrono::nanoseconds period,
+    FunctorT && callback,
+    rclcpp::Context::SharedPtr context,
+    bool autostart = true)
```
fujitatomoya, line 57: by default, `autostart` true means time starts immediately. that means timer callback will be called after creation w/o activating the lifecycle node? i was thinking that timer should not be called until node becomes to active status.

rclcpp_lifecycle/include/rclcpp_lifecycle/lifecycle_timer.hpp
```diff
@@ -0,0 +1,91 @@
...
+      context,
+      false),
+    autostart_(autostart)
+  {
+  }
+
+  void on_activate() override
+  {
+    SimpleManagedEntity::on_activate();
+    if (autostart_) {
```
fujitatomoya, line 71: what is user application creates the `LifecycleTimer` with `autostart` false after `on_activate`? timer will not be activated even the node is active status. i think that would not be what user expects.

rclcpp_lifecycle/test/test_lifecycle_timer.cpp
```diff
@@ -0,0 +1,165 @@
...
+  node_->trigger_transition(
+    rclcpp_lifecycle::Transition(Transition::TRANSITION_CONFIGURE), ret);
+  ASSERT_EQ(success, ret);
+  ret = reset_key;
+  node_->trigger_transition(
+    rclcpp_lifecycle::Transition(Transition::TRANSITION_ACTIVATE), ret);
+  ASSERT_EQ(success, ret);
+  ret = reset_key;
+  EXPECT_TRUE(timer_->is_activated());
+  EXPECT_TRUE(wall_timer_->is_activated());
```
fujitatomoya, line 113: not only checking with `is_activated`, but also needs to check `timer_called_` and `wall_timer_called_`.

rclcpp_lifecycle/test/test_lifecycle_timer.cpp
```diff
@@ -0,0 +1,165 @@
...
+TEST_F(TestLifecycleTimer, timer_managed_by_node) {
+  // Test that the timers transition with the node.
+  auto success = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+  auto reset_key = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR;
+  auto ret = reset_key;
+
+  EXPECT_EQ(State::PRIMARY_STATE_UNCONFIGURED, node_->get_current_state().id());
+  node_->trigger_transition(
+    rclcpp_lifecycle::Transition(Transition::TRANSITION_CONFIGURE), ret);
+  ASSERT_EQ(success, ret);
```
fujitatomoya, line 106: i think we need to check if `timer_called_` and `wall_timer_called_` are false here.

rclcpp_lifecycle/test/test_lifecycle_timer.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+TEST_F(TestLifecycleTimer, timer_callback_only_when_active) {
+  // Test that the callback only gets called when the timer is active.
+  auto success = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+  auto reset_key = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR;
+  auto ret = reset_key;
+
+  // Timer should not have been called yet
+  EXPECT_EQ(timer_called_, false);
+  EXPECT_EQ(wall_timer_called_, false);
```
fujitatomoya, line 132: i think there probably are failing since the timer will autostart w/o state checking.

fujitatomoya (changes requested):

PR #2263, opened 2023-08-03
Add RCLCPP_*_FORMAT logging macros 

nwn:
Similar to the `RCLCPP_*_STREAM` family of macros, which enable the use of C++'s stream-based output, these new logging macros allow users to conveniently invoke the new C++20 [`std::format()`](https://en.cppreference.com/w/cpp/utility/format/format) function for generating formatted text.

It should be noted that `rclcpp` (as of [Iron](https://www.ros.org/reps/rep-2000.html#iron-irwini-may-2023-november-2024)) still has a minimum requirement of C++17, which predates `std::format`. Moreover, many compilers only recently implemented support for this feature. (See the entry for P0645R10 [here](https://en.cppreference.com/w/cpp/compiler_support/20#C.2B.2B20_library_features).)

To account for this, the relevant standard library header is only included conditionally if it exists. This should guarantee that the include doesn't introduce errors for compilers without or with partial C++20 support, unless the user actually invokes the `RCLCPP_*_FORMAT` macros.

These macros are not strictly necessary. As seen in the implementation, one can mechanically replace any call to a `RCLCPP_*_FORMAT_*` macro with the corresponding printf-style `RCLCPP_*_*` macro, passing in the variadic arguments `"%s", std::format(...).c_str()`. However, this is less readable and more error prone.

### DIFF
--- a/rclcpp/include/rclcpp/logging.hpp
+++ b/rclcpp/include/rclcpp/logging.hpp
@@ -457,6 +457,20 @@
 #define RCLCPP_DEBUG_STREAM_THROTTLE(...)
 /// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
 #define RCLCPP_DEBUG_STREAM_SKIPFIRST_THROTTLE(...)
+/// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
+#define RCLCPP_DEBUG_FMT(...)
+/// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
+#define RCLCPP_DEBUG_FMT_ONCE(...)
+/// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
+#define RCLCPP_DEBUG_FMT_EXPRESSION(...)
+/// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
+#define RCLCPP_DEBUG_FMT_FUNCTION(...)
+/// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
+#define RCLCPP_DEBUG_FMT_SKIPFIRST(...)
+/// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
+#define RCLCPP_DEBUG_FMT_THROTTLE(...)
+/// Empty logging macro due to the preprocessor definition of RCLCPP_LOG_MIN_SEVERITY.
+#define RCLCPP_DEBUG_FMT_SKIPFIRST_THROTTLE(...)
 
 #else
 /**
@@ -557,6 +571,54 @@
   RCLCPP_LOG_STREAM_SKIPFIRST_THROTTLE(RCUTILS_LOG_SEVERITY_DEBUG, logger, clock, duration, \
     stream_arg)
 
+/**
+ * \def RCLCPP_DEBUG_FMT
+ * \copydoc RCLCPP_LOG_FMT
+ */
+#define RCLCPP_DEBUG_FMT(logger, ...) RCLCPP_LOG_FMT(RCUTILS_LOG_SEVERITY_DEBUG, logger, __VA_ARGS__)
+
+/**
+ * \def RCLCPP_DEBUG_FMT_ONCE
+ * \copydoc RCLCPP_LOG_FMT_ONCE
+ */
+#define RCLCPP_DEBUG_FMT_ONCE(logger, ...) \
+  RCLCPP_LOG_FMT_ONCE(RCUTILS_LOG_SEVERITY_DEBUG, logger, __VA_ARGS__)
+
+/**
+ * \def RCLCPP_DEBUG_FMT_EXPRESSION
+ * \copydoc RCLCPP_LOG_FMT_EXPRESSION
+ */
+#define RCLCPP_DEBUG_FMT_EXPRESSION(logger, expression, ...) \
+  RCLCPP_LOG_FMT_EXPRESSION(RCUTILS_LOG_SEVERITY_DEBUG, logger, expression, __VA_ARGS__)
+
+/**
+ * \def RCLCPP_DEBUG_FMT_FUNCTION
+ * \copydoc RCLCPP_LOG_FMT_FUNCTION
+ */
+#define RCLCPP_DEBUG_FMT_FUNCTION(logger, function, ...) \
+  RCLCPP_LOG_FMT_FUNCTION(RCUTILS_LOG_SEVERITY_DEBUG, logger, function, __VA_ARGS__)
+
+/**
+ * \def RCLCPP_DEBUG_FMT_SKIPFIRST
+ * \copydoc RCLCPP_LOG_FMT_SKIPFIRST
+ */
+#define RCLCPP_DEBUG_FMT_SKIPFIRST(logger, ...) \
+  RCLCPP_LOG_FMT_SKIPFIRST(RCUTILS_LOG_SEVERITY_DEBUG, logger, __VA_ARGS__)
+
+/**
+ * \def RCLCPP_DEBUG_FMT_THROTTLE
+ * \copydoc RCLCPP_LOG_FMT_THROTTLE
+ */
+#define RCLCPP_DEBUG_FMT_THROTTLE(logger, clock, duration, ...) \
+  RCLCPP_LOG_FMT_THROTTLE(RCUTILS_LOG_SEVERITY_DEBUG, logger, clock, duration, __VA_ARGS__)
+
+/**
+ * \def RCLCPP_DEBUG_FMT_SKIPFIRST_THROTTLE
+ * \copydoc RCLCPP_LOG_FMT_SKIPFIRST_THROTTLE
+ */
+#define RCLCPP_DEBUG_FMT_SKIPFIRST_THROTTLE(logger, clock, duration, ...) \
+  RCLCPP_LOG_FMT_SKIPFIRST_THROTTLE(RCUTILS_LOG_SEVERITY_DEBUG, logger, clock, duration, __VA_ARGS__)
+
 #endif
 
 /** @name Logging macros for severity INFO.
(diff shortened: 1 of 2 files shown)


clalancette:
My 2 cents is that we shouldn't take this PR until we can at least reasonably compile it (that is, we switch to C++20).  Otherwise this is just dead code.

But I'd like to hear opinions from @wjwwood @fujitatomoya @alsora @iuhilnehc-ynos, and anyone else on @ros2/team

rclcpp/resource/logging.hpp.em
```diff
@@ -149,10 +160,12 @@ def get_rclcpp_suffix_from_features(features):
 @(''.join(['      ' + p + ', \\\n' for p in params if p != stream_arg]))@
 @[ end if]@
       (logger).get_name(), \
-@[ if 'stream' not in feature_combination]@
-      __VA_ARGS__); \
-@[ else]@
+@[ if 'stream' in feature_combination]@
       "%s", rclcpp_stream_ss_.str().c_str()); \
+@[ elif 'format' in feature_combination]@
+      "%s", std::format(__VA_ARGS__).c_str()); \
```
fujitatomoya, line 166: this could possibly introduce `std::format_error`?

fujitatomoya (commented):
> My 2 cents is that we shouldn't take this PR until we can at least reasonably compile it (that is, we switch to C++20)

:+1:

nwn:
> @clalancette: My 2 cents is that we shouldn't take this PR until we can at least reasonably compile it (that is, we switch to C++20). Otherwise this is just dead code.

That's fair. When you say "switch to C++20", does that mean upgrading the build system to be able to test the code in CI, or bumping the minimum language requirement to C++20? In either case, is there a timeline or idea for when that might happen? On Ubuntu at least, 24.04 (which I assume Jazzy will target) should ship with ~full C++20 support in GCC.

rclcpp/resource/logging.hpp.em
nwn: Yes, `std::format` could throw `std::format_error` when the given format string is invalid for the provided arguments. However, in C++23 with [P2216](http://wg21.link/p2216), this is replaced by compile-time checks.

clalancette:
> That's fair. When you say "switch to C++20", does that mean upgrading the build system to be able to test the code in CI, or bumping the minimum language requirement to C++20?

Both, essentially.

> In either case, is there a timeline or idea for when that might happen? On Ubuntu at least, 24.04 (which I assume Jazzy will target) should ship with ~full C++20 support in GCC.

Unfortunately I don't have a timeline for it.  While we almost certainly will target Ubuntu 24.04, there is always a bunch of work necessary to switch C++ versions, and there is no allocated time in the schedule for doing that right now.  If you are interested in trying to make it happen, I can probably come up with list of things that I think will need to happen in order to do it.

Grave:
> @clalancette: If you are interested in trying to make it happen, I can probably come up with list of things that I think will need to happen in order to do it.

At least I would be interested. Would it be possible to get such a list?

clalancette:
> At least I would be interested. Would it be possible to get such a list?

The biggest issue is that while we will be switching to Ubuntu 24.04, which probably has good C++20 support, we still have to support older platforms.  In particular, we will still be supporting RHEL-9 (which has gcc 11.3.1), and Windows via MSVC 2019.  I have no idea what the C++20 support is like in those compilers.  So the list looks something like:

1.  Try out C++20 mode in both MSVC 2019 and gcc 11.3.1, and see what doesn't work.
2.  Potentially upgrade MSVC 2019 to MSVC 2022.
3.  We can't update the compiler on RHEL-9, so if it doesn't support some C++20 features, we'll have to find other ways to workaround it.
4.  Change the top stanza of all of the CMakeLists.txt in all 350 of the core packages to reference C++20 instead of C++17.
5.  Compile the entire core with C++20, and fix all the new warnings and errors that will pop up, across all of the supported platforms.  This usually also involves updating at least some of our ~24 vendored dependencies.

oysstu:
> ...
>     3. We can't update the compiler on RHEL-9, so if it doesn't support some C++20 features, we'll have to find other ways to workaround it.
>  ...
All of the C++20 language features are in GCC11, except full support for modules. The C++20 library features on the other hand are missing a few things. The C++20 format library is for example not available until GCC13. Using [fmtlib](https://github.com/fmtlib/fmt) for this PR would enable backwards compatibility with more or less the same API as the forthcoming standard without fragmenting the userbase.

On a side-note, C++20 also introduces the [feature test ](https://en.cppreference.com/w/cpp/feature_test) macros, which can be helpful when maintaining backwards compatibility with older compilers.

nwn:
I'm hoping to revive this PR since it sounds like Lyrical will be targeting C++20 (according to [this post](https://discourse.openrobotics.org/t/ros-2-lyrical-c-version/52551)).

I've rebased and updated the branch to reflect the new way of defining the logging macros introduced in https://github.com/ros2/rclcpp/pull/2870.

FYI @fujitatomoya @clalancette

fujitatomoya (approved):
i think it is okay to support these macros from Lyrical C++20, implementation looks good to me with tests. i would like to have another approval before making any further action.

fujitatomoya:
@mjcarroll @ahcorde any objections?

oysstu:
How are compile times looking when including the format header? I tried looking for a comparison with libfmt, but haven't found  a good source for any benchmarks. Since this will be included in many compilation units, it's something to consider, possibly putting it in a separate header if necessary.

Libfmt has a larger focus on compile time performance than the standard libraries.

nwn:
> @oysstu: How are compile times looking when including the format header?

I ran a very basic benchmark on timings from including the headers, and it looks like the standard `<format>` is slightly slower than `fmtlib`'s `<fmt/format.h>`, but not by a huge amount. On my system, including `<cstdio>` has basically no impact, `<sstream>` adds ~368 ms, `<format>` adds ~660 ms, and `<fmt/format.h>` adds ~615 ms.

```
Benchmark 1: none: g++ -std=c++20 -O2 -o /dev/null main.cpp -DVARIANT=NONE
  Time (mean ± σ):     147.0 ms ±   2.4 ms    [User: 91.7 ms, System: 55.5 ms]
  Range (min … max):   144.2 ms … 151.8 ms    10 runs
 
Benchmark 2: cstdio: g++ -std=c++20 -O2 -o /dev/null main.cpp -DVARIANT=CSTDIO
  Time (mean ± σ):     147.8 ms ±   2.0 ms    [User: 92.3 ms, System: 55.6 ms]
  Range (min … max):   145.0 ms … 151.3 ms    10 runs
 
Benchmark 3: sstream: g++ -std=c++20 -O2 -o /dev/null main.cpp -DVARIANT=SSTREAM
  Time (mean ± σ):     515.4 ms ±  10.2 ms    [User: 422.8 ms, System: 92.5 ms]
  Range (min … max):   503.0 ms … 533.9 ms    10 runs
 
Benchmark 4: format: g++ -std=c++20 -O2 -o /dev/null main.cpp -DVARIANT=FORMAT
  Time (mean ± σ):     807.9 ms ±  12.5 ms    [User: 695.1 ms, System: 112.6 ms]
  Range (min … max):   791.8 ms … 837.3 ms    10 runs
 
Benchmark 5: fmtlib: g++ -std=c++20 -O2 -o /dev/null main.cpp -DVARIANT=FMTLIB -l fmt
  Time (mean ± σ):     762.3 ms ±   6.6 ms    [User: 648.8 ms, System: 112.1 ms]
  Range (min … max):   749.1 ms … 769.8 ms    10 runs
 
Summary
  none ran
    1.01 ± 0.02 times faster than cstdio
    3.51 ± 0.09 times faster than sstream
    5.19 ± 0.10 times faster than fmtlib
    5.50 ± 0.12 times faster than format
```

(There's definitely a difference in compile times when calling their respective `format()` functions, though.)

oysstu:
> > @oysstu: How are compile times looking when including the format header?
> 
> I ran a very basic benchmark on timings from including the headers, and it looks like the standard `<format>` is slightly slower than `fmtlib`'s `<fmt/format.h>`, but not by a huge amount. On my system, including `<cstdio>` has basically no impact, `<sstream>` adds ~368 ms, `<format>` adds ~660 ms, and `<fmt/format.h>` adds ~615 ms.
> 
> (There's definitely a difference in compile times when calling their respective `format()` functions, though.)

I believe the standard library is header-only, whereas libfmt is partially precompiled by default. The latter will not have to evaluate as many templates on use, so that's probably the main reason for the compile time difference on use.

Since the compile time increase is only incurred on use, I think that's okay. Pay for what you use and all that.

Spdlog uses libfmt by default, but offers a compile option to switch to stdlib. Something like that can also be implemented in the future if the need arises. The APIs are largely compatible since stdlib is based off the libfmt design. Formatters for custom types are in different namespaces though.

nwn:
Just an idea to limit the compile-time impact, since logging tends to be used in a lot of translation units, we could split `logging.hpp` into:
- `logging.hpp` as a top-level header that includes the following three
- `logging/core.hpp` with the basic and common macros that we always need
- `logging/stream.hpp` with the `*_STREAM_*` macros
- `logging/fmt.hpp` with the `*_FMT_*` macros

Then users would still have access to everything via the existing `logging.hpp` header, but can optimize to only include what they want with the `logging/*.hpp` headers.

nwn:
@mjcarroll @ahcorde I'd appreciate a review of this PR. It would be great if we can still get this in before the Lyrical code freeze on April 20.

rclcpp/CMakeLists.txt
```diff
@@ -193,7 +193,7 @@ foreach(interface_file ${interface_files})
 endforeach()
 
 add_library(${PROJECT_NAME} ${${PROJECT_NAME}_SRCS})
-target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_17)
+target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_20)
```
jmachowinski: This line needs removal, we export the minimum c++ version in a different way now.

jmachowinski (approved):

rclcpp/CMakeLists.txt
nwn: Done

jmachowinski:
Pulls: ros2/rclcpp#2263
Gist: https://gist.githubusercontent.com/jmachowinski/d9b4fda1e17b870c3934a0c285419863/raw/4bb5cf63baa3f3de6f895bbc2cadedba331d447c/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19027
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28795/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21611/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8636/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27667/)

fujitatomoya:
so this does not work with RHEL just yet... sadly...

something we can do is feature-test and make the FMT macros conditional?

wjwwood:
> so this does not work with RHEL just yet... sadly...

What's the issue with RHEL? The gcc version or something?

I'm going to restart CI on this just to see if anything has changed. It would be awesome to see this land. Perhaps conditional inclusion and/or an option to use fmt if already included before our header might make it easier to accept. Let's see what CI does.

wjwwood:
Pulls: ros2/rclcpp#2263
Gist: https://gist.githubusercontent.com/wjwwood/8eb7f8c0b4407f0c9c46c05890992b77/raw/4bb5cf63baa3f3de6f895bbc2cadedba331d447c/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20088
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30079/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22846/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9964/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28999/)

wjwwood:
I'm not sure how to clear/reset the Rpr job, which seems out-of-date or defunct. I naively tried closing/reopening the pr, but that didn't work. Likely pushing a new commit will fix it.

mjcarroll:
Passing full CI seems fine, I'm just going to merge it.

PR #2266, opened 2023-08-04
doc fix: call `canceled` only after goal state is in canceling.

fujitatomoya:
address part of https://github.com/ros2/rclcpp/issues/2242

### DIFF
--- a/rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
+++ b/rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
@@ -196,7 +196,7 @@ public:
 
   /// Indicate that a goal has been canceled.
   /**
-   * Only call this if the goal is executing or pending, but has been canceled.
+   * Only call this if the goal is canceling.
    * This is a terminal state, no more methods should be called on a goal handle after this is
    * called.
    *


fujitatomoya (commented):
@clalancette can you take a look? the doc is not exactly correct, and if that is still executing state (not accepted cancel yet), this will hit the assertion with invalid state transition.

rclcpp_action/include/rclcpp_action/server_goal_handle.hpp
```diff
@@ -196,7 +196,7 @@ class ServerGoalHandle : public ServerGoalHandleBase
 
   /// Indicate that a goal has been canceled.
   /**
-   * Only call this if the goal is executing or pending, but has been canceled.
+   * Only call this if the goal is canceling, but has been canceled.
```
clalancette, line 199: I'm not 100% sure of this, but I think this will be clearer like this:

```suggestion
   * Only call this if the goal is canceling.
```

fujitatomoya:
@clalancette i dont think we need CI for this?

clalancette:
> @clalancette i dont think we need CI for this?

No, the Rpr job is good enough in my opinion.  Going ahead and merging.

Issue #2243, opened 2023-07-20
ActionClient deadlocks if accessing goalHandle in feedback_callback


jmachowinski:
## Bug report

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries
- Version or commit hash:
  - iron
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp-action

#### Steps to reproduce issue
Create a ActionClient with a custom feedback_callback

wait for the fallback to be called

call get_status() on the passes GoalHandle

#### Expected behavior
It should not deadlock

#### Actual behavior
It deadlocks

#### Additional information
The function call_feedback_callback (https://github.com/ros2/rclcpp/blob/c0d72c3ee0ac806379b49ed10cf17364a4bd443a/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L163) holds the mutex of the handle while calling the
feedback_callback. Therefore using any method on the handle, that tries to also acquire
the mutex will deadlock the system.

CursedRock17:
What GoalHandle are you calling with the feedback_callback, I set up a minimal ActionClient and can't seem to find how you're getting a deadlock, for instance I have:
```
void feedback_callback(
GoalHandle::SharedPtr, const std::shared_ptr<const ParamType::Feedback> feedback){
// Do stuff to feedback
}
```
Then I just have a `send_goal` function on a timer in a ClientConstructor, where I bind the feedback_callback. There's a mutex on every function in the file you gave in additional information. Can you explain, with code, how you're using call_feedback_callback?

jmachowinski:
https://github.com/jmachowinski/rclcppActionDeadlock

This code demonstrates the issue. 

There are two ways to fix this problem, either release the handle_mutex_ before calling the callback (here https://github.com/ros2/rclcpp/blob/22a954e1b00ea2ba810666baa92a3c66fe1a8654/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L177),
or make the handle_mutex a std::recursive_mutex.

tgroechel:
[...]

PR #2267, opened 2023-08-07
fix(ClientGoalHandle): Made mutex recursive to prevent deadlocks

jmachowinski:
This prevents deadlocks in cases, were e.g. get_status() would be called on the handle in a callback of the handle.

Fixes https://github.com/ros2/rclcpp/issues/2243

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
+++ b/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
@@ -59,7 +59,7 @@ template<typename ActionT>
 std::shared_future<typename ClientGoalHandle<ActionT>::WrappedResult>
 ClientGoalHandle<ActionT>::async_get_result()
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   if (!is_result_aware_) {
     throw exceptions::UnawareGoalHandleError();
   }
@@ -70,7 +70,7 @@ template<typename ActionT>
 void
 ClientGoalHandle<ActionT>::set_result(const WrappedResult & wrapped_result)
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   status_ = static_cast<int8_t>(wrapped_result.code);
   result_promise_.set_value(wrapped_result);
   if (result_callback_) {
@@ -82,7 +82,7 @@ template<typename ActionT>
 void
 ClientGoalHandle<ActionT>::set_feedback_callback(FeedbackCallback callback)
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   feedback_callback_ = callback;
 }
 
@@ -90,7 +90,7 @@ template<typename ActionT>
 void
 ClientGoalHandle<ActionT>::set_result_callback(ResultCallback callback)
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   result_callback_ = callback;
 }
 
@@ -98,7 +98,7 @@ template<typename ActionT>
 int8_t
 ClientGoalHandle<ActionT>::get_status()
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   return status_;
 }
 
@@ -106,7 +106,7 @@ template<typename ActionT>
 void
 ClientGoalHandle<ActionT>::set_status(int8_t status)
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   status_ = status;
 }
 
@@ -114,7 +114,7 @@ template<typename ActionT>
 bool
 ClientGoalHandle<ActionT>::is_feedback_aware()
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   return feedback_callback_ != nullptr;
 }
 
@@ -122,7 +122,7 @@ template<typename ActionT>
 bool
 ClientGoalHandle<ActionT>::is_result_aware()
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   return is_result_aware_;
 }
 
@@ -130,7 +130,7 @@ template<typename ActionT>
 bool
 ClientGoalHandle<ActionT>::set_result_awareness(bool awareness)
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   bool previous = is_result_aware_;
   is_result_aware_ = awareness;
   return previous;
@@ -140,7 +140,7 @@ template<typename ActionT>
 void
 ClientGoalHandle<ActionT>::invalidate(const exceptions::UnawareGoalHandleError & ex)
 {
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   // Guard against multiple calls
   if (is_invalidated()) {
     return;
@@ -168,7 +168,7 @@ ClientGoalHandle<ActionT>::call_feedback_callback(
     RCLCPP_ERROR(rclcpp::get_logger("rclcpp_action"), "Sent feedback to wrong goal handle.");
     return;
   }
-  std::lock_guard<std::mutex> guard(handle_mutex_);
+  std::lock_guard<std::recursive_mutex> guard(handle_mutex_);
   if (nullptr == feedback_callback_) {
     // Normal, some feedback messages may arrive after the goal result.
     RCLCPP_DEBUG(rclcpp::get_logger("rclcpp_action"), "Received feedback but goal ignores it.");
--- a/rclcpp_action/test/test_client.cpp
+++ b/rclcpp_action/test/test_client.cpp
@@ -852,6 +852,86 @@ TEST_F(TestClientAgainstServer, async_cancel_some_goals_with_callback)
   EXPECT_EQ(rclcpp_action::GoalStatus::STATUS_CANCELED, goal_handle0->get_status());
 }
 
+TEST_F(TestClientAgainstServer, deadlock_in_callbacks)
+{
+  std::atomic<bool> feedback_callback_called = false;
+  std::atomic<bool> response_callback_called = false;
+  std::atomic<bool> result_callback_called = false;
+  std::atomic<bool> no_deadlock = false;
+
+  std::thread tr = std::thread(
+    [&]() {
+      auto action_client = rclcpp_action::create_client<ActionType>(client_node, action_name);
+      ASSERT_TRUE(action_client->wait_for_action_server(WAIT_FOR_SERVER_TIMEOUT));
+
+      ActionGoal goal;
+
+      using GoalHandle = rclcpp_action::ClientGoalHandle<ActionType>;
+      rclcpp_action::Client<ActionType>::SendGoalOptions ops;
+      ops.feedback_callback =
+      [&feedback_callback_called](const GoalHandle::SharedPtr handle,
+      ActionType::Feedback::ConstSharedPtr) {
+        // call functions on the handle that acquire the lock
+        handle->get_status();
+        handle->is_feedback_aware();
+        handle->is_result_aware();
+
+        feedback_callback_called = true;
+      };
+      ops.goal_response_callback = [&response_callback_called](
+        const GoalHandle::SharedPtr & handle) {
+        // call functions on the handle that acquire the lock
+        handle->get_status();
+        handle->is_feedback_aware();
+        handle->is_result_aware();
+
+        response_callback_called = true;
+      };
+      ops.result_callback = [&result_callback_called](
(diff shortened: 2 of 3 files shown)


clalancette (changes requested):
Can you please add a test showing this in use?  Otherwise this looks good to me.

rclcpp_action/test/test_client.cpp
```diff
@@ -852,6 +852,82 @@ TEST_F(TestClientAgainstServer, async_cancel_some_goals_with_callback)
...
+  while (std::chrono::system_clock::now() - curTime < std::chrono::milliseconds(500) &&
+    !test_done)
+  {
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
+
+  // in case of a failure, we don't want to block further tests
+  tr.detach();
+
+  EXPECT_TRUE(response_callback_called);
```
fujitatomoya, line 926: we need to check if the test is completed w/o timeout. in the case that callbacks does not get deadlock, but did not complete the thread as expected.

```suggestion
  EXPECT_TRUE(test_done);
  EXPECT_TRUE(response_callback_called);
```

rclcpp_action/test/test_client.cpp
```diff
@@ -852,6 +852,82 @@ TEST_F(TestClientAgainstServer, async_cancel_some_goals_with_callback)
...
+        result_callback_called = true;
+      };
+
+
+      goal.order = 6;
+      auto future_goal_handle = action_client->async_send_goal(goal, ops);
+      dual_spin_until_future_complete(future_goal_handle);
+      auto goal_handle = future_goal_handle.get();
+
+      EXPECT_TRUE(goal_handle);
```
fujitatomoya, line 902: if goal_handle is NULL, i think we should abort the test here.

```suggestion
      ASSERT_TRUE(goal_handle);
```

rclcpp_action/test/test_client.cpp
```diff
@@ -852,6 +852,82 @@ TEST_F(TestClientAgainstServer, async_cancel_some_goals_with_callback)
...
+      test_done = true;
+    });
+
+  auto curTime = std::chrono::system_clock::now();
+
+  while (std::chrono::system_clock::now() - curTime < std::chrono::milliseconds(500) &&
+    !test_done)
+  {
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
```
fujitatomoya, line 921:
```suggestion
  bool timeout = false;
  do {
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
    timeout = (std::chrono::system_clock::now() - curTime) > std::chrono::milliseconds(500);
  } while (!timeout && !test_done)
```

and the probably we can check later with,

```cpp
  EXPECT_FALSE(timeout);
```

to make sure timeout has not been detected.

fujitatomoya (commented):
overall, lgtm with a few comments.

rclcpp_action/test/test_client.cpp
jmachowinski: A timeout is implicitly detected if test_done is false. Is there any benefit in testing explicit for it ?

rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
```diff
@@ -168,7 +168,7 @@ ClientGoalHandle<ActionT>::call_feedback_callback(
     RCLCPP_ERROR(rclcpp::get_logger("rclcpp_action"), "Sent feedback to wrong goal handle.");
     return;
   }
-  std::lock_guard<std::mutex> guard(handle_mutex_);
```
sloretz, line 171: Alternative idea, do we need to hold the lock here? What if we hold the lock in a scope `{...}` only long enough to copy `feedback_callback_` to a local variable, then release the lock before calling the callback? I think that solves the issue and avoids (probably small) overhead of using a recursive mutex.

clalancette: I'm not sure that will work.

There are a couple of assumptions at work here:
1.  That the `shared_this` being passed in here is directly the same as `this`.  That is a weird API, but OK.  In that case, we can consider that everything that would need to be protected from concurrent access to `this` also applies to `shared_this`.
2.  That the `handle_mutex_` here is protecting all members of this object from concurrent modification/access.

Taken together, I think it means that we need to hold the mutex across the entire operation (since `feedback_callback_` is passed `shared_this`).  Does that make sense?  Or am I maybe missing something?

sloretz: > That the shared_this being passed in here is directly the same as this. That is a weird API, but OK. In that case, we can consider that everything that would need to be protected from concurrent access to this also applies to shared_this.

If `shared_this` isn't `this` then holding a lock on `this->handle_mutex_` isn't going to help as the feedback callback is given `shared_this`, not `this`. I think we can ignore this though because there is a check that `shared_this.get() == this` above.

> That the `handle_mutex_` here is protecting all members of this object from concurrent modification/access.
> [...] Taken together, I think it means that we need to hold the mutex across the entire operation [...]

IIUC holding the lock during the feedback's entire call would protect against a situation where the goal handle changed state between two feedback calls. Just getting an idea for what could change, it looks like all the public methods are:

```
  /// Get the unique ID for the goal.
  const GoalUUID &
  get_goal_id() const;

  /// Get the time when the goal was accepted.
  rclcpp::Time
  get_goal_stamp() const;

  /// Get the goal status code.
  int8_t
  get_status();

  /// Check if an action client has subscribed to feedback for the goal.
  bool
  is_feedback_aware();

  /// Check if an action client has requested the result for the goal.
  bool
  is_result_aware()
```

All of those methods only "read" data from the class, so the concern would be a "write" somewhere else at the same time as the feedback callback. "writes" could come from the `Client` [as it's a friend of this class](https://github.com/ros2/rclcpp/blob/c6edffafefa477068f3e00d4e65b83a720195eef/rclcpp_action/include/rclcpp_action/client_goal_handle.hpp#L107).

I'm choosing to ignore `get_goal_id()` and `get_goal_stamp()` as they're not even protected by locks. They assume `info_` is constant (side note: [I think this could be enforced by adding const here](https://github.com/ros2/rclcpp/blob/c6edffafefa477068f3e00d4e65b83a720195eef/rclcpp_action/include/rclcpp_action/client_goal_handle.hpp#L154)).

`get_status` [is protected by a lock](https://github.com/ros2/rclcpp/blob/c6edffafefa477068f3e00d4e65b83a720195eef/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L101). If we wrap the `feedback_callback_` in the lock then it's possible the feedback callback could call get status twice and get different results (executing one call, then complete the next).

`is_feedback_aware()` [is protected by a lock](https://github.com/ros2/rclcpp/blob/c6edffafefa477068f3e00d4e65b83a720195eef/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L117) and means "does this goal handle have a feedback callback"? If the callback is being executed, it already knows the answer and doesn't have a need to call it. I suppose it's possible the feedback callback could get set to a `nullptr` while this one is running and `is_feedback_aware()` might return `false` if this function called it, but I wasn't able to find any calls to `client_goal_handle->set_feedback_callback()`  anywhere (side note: if it's private and no one calls it, can we delete it?).

`is_result_aware()` [is protected by a lock](https://github.com/ros2/rclcpp/blob/c6edffafefa477068f3e00d4e65b83a720195eef/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L125) and means "does someone want the result of this goal"? It's possible  that someone could write a feedback callback that they gave access to the `Client` and only asked for the result if the feedback reports a certain value. If they were using  multithreaded executor and put the Client into a Reentrant callback group and feedback messages came in fast enough then two feedback callbacks could be executed simultaneously and both would call [client->async_get_result()](https://github.com/ros2/rclcpp/blob/c6edffafefa477068f3e00d4e65b83a720195eef/rclcpp_action/include/rclcpp_action/client.hpp#L495C3-L495C19). There can only be one result callback, [so the second's feedback callback would replace the first's](https://github.com/ros2/rclcpp/blob/c6edffafefa477068f3e00d4e65b83a720195eef/rclcpp_action/include/rclcpp_action/client.hpp#L508-L511).

Affecting decision making in the `feedback_callback` based on `is_result_aware()` seems like the only potential negative consequence to me, and it requires the user passing in the `Client` to the callback some other way.

All that to say, I'm fine with the current approach, but I don't see any significant problems with releasing the lock during the feedback callback either. All holding the lock seems to do is make sure two feedback callbacks on one goal handle can't be run in parallel regardless of the executor or callback group.

fujitatomoya: > side note: if it's private and no one calls it, can we delete it?

I was thinking the same thing, i guess we can just remove this. but originally this should have been exposed to user application to change the feedback callback? i was not sure about that.

jmachowinski: I think releasing the mutex is a bad idea.
If a reentrant cb group is used, not holding the mutex could lead to strange stuff like the handle->getStatus() function changing the result in the middle of the feedback callback.
I guess this is the case, why the call_feedback_callback function exists in the first place.

jmachowinski:
> > So, can we merge this ?
> 
> As far as I can tell, we are still discussing the solution.

Okay, I was under the impression the discussion came to the result to not release the mutex....

fujitatomoya (approved):

fujitatomoya:
@jmachowinski can you fix DCO error?

jmachowinski:
> @jmachowinski can you fix DCO error?

Done

rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
clalancette: @sloretz Thanks for the detailed analysis, that is very helpful.  Here are a few comments:

> IIUC holding the lock during the feedback's entire call would protect against a situation where the goal handle changed state between two feedback calls.

To be clear, we are already holding the lock across the `feedback_callback_` right now.  The difference in this PR is that we are switching that to a `recursive_mutex`.  But your overall point stands that we could consider dropping the mutex before making that call.

Considering your points, I agree that given the code today we could probably drop the lock without too many negative consequences.  But I think I still prefer to keep holding the lock for two reasons:

1.  @jmachowinski point about things changing while the callback is running.  We've run into these kinds of problems before in `rclcpp`, particularly around parameters, so we also hold a lock around that for similar reasons.
2.  In the future it is possible someone could add a method that did a "write", and it would be hard to figure out that it breaks our current assumptions.

So all of that is to say that I'm good with this PR as-is.  I'm going to do a review here, and then we can run CI on it.

rclcpp_action/test/test_client.cpp
```diff
@@ -852,6 +852,83 @@ TEST_F(TestClientAgainstServer, async_cancel_some_goals_with_callback)
...
+        handle->is_feedback_aware();
+        handle->is_result_aware();
+
+        response_callback_called = true;
+      };
+      ops.result_callback = [&result_callback_called](
+        const GoalHandle::WrappedResult &) {
+        result_callback_called = true;
+      };
+
```
clalancette, line 895: Nit: unnecessary blank line:

```suggestion
```

rclcpp_action/test/test_client.cpp
```diff
@@ -852,6 +852,83 @@ TEST_F(TestClientAgainstServer, async_cancel_some_goals_with_callback)
...
+      EXPECT_TRUE(result_future.valid());
+      auto result = result_future.get();
+
+      test_done = true;
+    });
+
+  auto curTime = std::chrono::system_clock::now();
+
+  while (std::chrono::system_clock::now() - curTime < std::chrono::milliseconds(500) &&
```
clalancette, line 917: 500 milliseconds seems pretty short, particularly on Windows where everything is slow.  I'll suggest increasing this to at least a couple of seconds:

```suggestion
  while (std::chrono::system_clock::now() - curTime < std::chrono::milliseconds(2000) &&
```

rclcpp_action/test/test_client.cpp
```diff
@@ -852,6 +852,83 @@ TEST_F(TestClientAgainstServer, async_cancel_some_goals_with_callback)
...
+    !test_done)
+  {
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
+
+  // in case of a failure, we don't want to block further tests
+  tr.detach();
```
clalancette, line 924: I can't say I'm a fan of this; we are just leaking the thread here.  Instead, what I'll suggest is that we augment `dual_spin_until_future_complete` to have a timeout parameter, and then we can just do a regular `join`.  If we are *still* worried about things hanging up with the join, then we can do a loop with a sleep and a call to `joinable` before attempting the `join`.

clalancette (changes requested):
I've left a few things in the test to clean up.  This otherwise looks good to me.

rclcpp_action/test/test_client.cpp
jmachowinski: This tests for a deadlock, if it locks, it will never join, hence the test unit test will just hang. I don't see an alternative to leaking the thread.

rclcpp_action/test/test_client.cpp
jmachowinski: removed

rclcpp_action/test/test_client.cpp
jmachowinski: I modified the code, to join in the case, that the test went through

clalancette (changes requested):
One more minor thing to fix, then I'll go ahead and run CI on this.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19482/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14002/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20199/)

clalancette (approved):
All right, CI looks happy and I think we've come to agreement here.  Thanks for your patience @jmachowinski .  Going ahead and merging this one.

PR #2274, opened 2023-08-15
Serialized Messages with Topic Statistics

CursedRock17:
This pull request was created to solve a [TODO](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/subscription.hpp#L328) that was added in pull request #1557, but there's no discussion on the topic. As per the title, the point was to allow the user to use topic statistics when handling a serialized message. I tried to follow the way the other two functions `handle_message` and `handle_loaned_message` handled their topic statistics, but assumed that one couldn't pass a serialized message through the statistics, so I set up a basic serialization process. We only need to make one copy of the message that we create and since `ROSMessageType` is a string in disguise, we can still get all the data from it. Additionally, we still just dispatch the message, before we get an information about the topic statistics, so there shouldn't be any timing issues.

The last time a topic statistics was added to one of them methods was PR #1927, where the author replicated a similar format and didn't add any tests. I don't necessarily know how the test would work, but I can add them if needed, and wanted this PR to be as concise as possible without adding extra messages.

### DIFF
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -326,12 +326,6 @@ public:
     const std::shared_ptr<rclcpp::SerializedMessage> & serialized_message,
     const rclcpp::MessageInfo & message_info) override
   {
-    if (matches_any_intra_process_publishers(&message_info.get_rmw_message_info().publisher_gid)) {
-      // In this case, the message will be delivered via intra process and
-      // we should ignore this copy of the message.
-      return;
-    }
-
     std::chrono::time_point<std::chrono::system_clock> now;
     if (subscription_topic_statistics_) {
       // get current time before executing callback to
@@ -341,15 +335,10 @@ public:
 
     any_callback_.dispatch(serialized_message, message_info);
 
-    ROSMessageType deserialized_message;
-    auto serializer = rclcpp::Serialization<ROSMessageType>();
-    serializer.deserialize_message(serialized_message.get(), &deserialized_message);
-
-
     if (subscription_topic_statistics_) {
       const auto nanos = std::chrono::time_point_cast<std::chrono::nanoseconds>(now);
       const auto time = rclcpp::Time(nanos.time_since_epoch().count());
-      subscription_topic_statistics_->handle_message(deserialized_message, time);
+      subscription_topic_statistics_->handle_message(message_info.get_rmw_message_info(), time);
     }
   }
 
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -44,7 +44,6 @@
 #include "rclcpp/subscription_base.hpp"
 #include "rclcpp/subscription_options.hpp"
 #include "rclcpp/subscription_traits.hpp"
-#include "rclcpp/serialization.hpp"
 #include "rclcpp/type_support_decl.hpp"
 #include "rclcpp/visibility_control.hpp"
 #include "rclcpp/waitable.hpp"
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -326,7 +326,12 @@ public:
     const std::shared_ptr<rclcpp::SerializedMessage> & serialized_message,
     const rclcpp::MessageInfo & message_info) override
   {
-    // TODO(wjwwood): enable topic statistics for serialized messages
+    if (matches_any_intra_process_publishers(&message_info.get_rmw_message_info().publisher_gid)) {
+      // In this case, the message will be delivered via intra process and
+      // we should ignore this copy of the message.
+      return;
+    }
+
     std::chrono::time_point<std::chrono::system_clock> now;
     if (subscription_topic_statistics_) {
       // get current time before executing callback to
--- a/rclcpp/include/rclcpp/subscription.hpp
+++ b/rclcpp/include/rclcpp/subscription.hpp
@@ -44,6 +44,7 @@
 #include "rclcpp/subscription_base.hpp"
 #include "rclcpp/subscription_options.hpp"
 #include "rclcpp/subscription_traits.hpp"
+#include "rclcpp/serialization.hpp"
 #include "rclcpp/type_support_decl.hpp"
 #include "rclcpp/visibility_control.hpp"
 #include "rclcpp/waitable.hpp"
@@ -326,7 +327,25 @@ public:
     const rclcpp::MessageInfo & message_info) override
   {
     // TODO(wjwwood): enable topic statistics for serialized messages
+    std::chrono::time_point<std::chrono::system_clock> now;
+    if (subscription_topic_statistics_) {
+      // get current time before executing callback to
+      // exclude callback duration from topic statistics result.
+      now = std::chrono::system_clock::now();
+    }
+
     any_callback_.dispatch(serialized_message, message_info);
+
+    ROSMessageType deserialized_message;
+    auto serializer = rclcpp::Serialization<ROSMessageType>();
+    serializer.deserialize_message(serialized_message.get(), &deserialized_message);
+
+
+    if (subscription_topic_statistics_) {
+      const auto nanos = std::chrono::time_point_cast<std::chrono::nanoseconds>(now);
+      const auto time = rclcpp::Time(nanos.time_since_epoch().count());
+      subscription_topic_statistics_->handle_message(deserialized_message, time);
+    }
   }
 
   void


MichaelOrlov:
@CursedRock17 Out of curiosity in what cases the user might need to use statistics when handling serialized messages on strictly typed subscriptions?

I would imagine it would be useful for `GenericSubscription`, but have a hard time imagining a use case with a regular subscription.

CursedRock17:
I just assumed it was a nice feature for the user to gain information on the message, similar to the other two methods. Also according to `GenericSubscription`'s documentation:
> Since the type is not known at compile time

and the Topic Statistics Concept page:
>Each statistic set is calculated in constant time and constant memory 

I kind of figured that Topic Statistics would be much more difficult to handle at this point, but if you feel this should be handled within a serialized message type class then I can see that.

Also could not find any information in PR #1072 / Meta-ticket [901](https://github.com/ros2/ros2/issues/901), when trying to digger deeper for rationale behind the TODO

MichaelOrlov:

MichaelOrlov (approved):
@CursedRock17 Thanks for the PR.  
While implementation requires extra serialization which could be counted as suboptimal, it is unavoidable in the current available topics statistics API and design.  

In the future to get rid from the extra deserialization need to change or extend the topic statistics API to accept `message_info` instead of the deserialized message. Inside the topic statistics package deserialized message is used only for getting timestamp from the header. 
If I am not mistaken we have all timestamps in the message info.

Testing is not easy to implement in this case. It is going to be an integration type of test. When we will need to setup topic statistics for the node and subscription and get serialized messages from that subscription, then verify numbers in the topic statistics results. This is a cross package with publisher subscriber type of integration test. And it isn't worth it IMO.
cc: @clalancette

clalancette:
> Testing is not easy to implement in this case. It is going to be an integration type of test. When we will need to setup topic statistics for the node and subscription and get serialized messages from that subscription, then verify numbers in the topic statistics results. This is a cross package with publisher subscriber type of integration test. And it isn't worth it IMO.

We do have https://github.com/ros2/system_tests/ for exactly this kind of integration test.  Whether it is worth it, I'm not sure of.

That said, I have serious concerns with this PR, as it will make subscriptions far more expensive.  I'll leave a comment in a review.

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -325,8 +326,31 @@ class Subscription : public SubscriptionBase
...
+      // get current time before executing callback to
+      // exclude callback duration from topic statistics result.
+      now = std::chrono::system_clock::now();
+    }
+
     any_callback_.dispatch(serialized_message, message_info);
+
+    ROSMessageType deserialized_message;
+    auto serializer = rclcpp::Serialization<ROSMessageType>();
+    serializer.deserialize_message(serialized_message.get(), &deserialized_message);
```
clalancette, line 346: I know this is unavoidable given the current APIs, but this is going to make serialized messages vastly more expensive.  Consider a large, multi-megabyte pointcloud; we are deserializing this for every single message now.  I don't think that this is a viable strategy, as one of the goals of serialized messages is to avoid exactly that kind of deserialization (which is what we do in the normal, non-serialized message path).

clalancette (changes requested):

rclcpp/include/rclcpp/subscription.hpp
MichaelOrlov: @clalancette From one hand, yes it will make getting serialized messages out of a regular strictly typed subscription more expensive due to the deserialization. But on another hand, think about in what exact cases?
When all of the following conditions are true
1. Topics statistics enabled!
2. The user getting serialized messages out of the strictly typed subscription which is odd and very unusual. In ROS 2 core packages we are not using such a scenario AFAIK and I can't imagine from the top of my head why it would be needed for the user. For reference in rosbag2 we are using `GenericSubscription` which is not affected by these changes.

CursedRock17: > The user getting serialized messages out of the strictly typed subscription which is odd and very unusual.

If rosbag2 implements using `GenericSubscription` and this really is as uncommon as you say, then maybe this should be down in the `handle_serialized_message` method in that [class](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/generic_subscription.cpp#L50-L55). We most likely don't want to be writing features for very explicit scenarios when we can cover most sections with a valid general solution.

That being said, while I won't have to include any of the intraprocess stuff, I am interested in passing these serialized messages to topic statistics, while staying serialized. Would we do something similar to the passing chain of `SerializedMessage` to the [`dispatch` call](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/any_subscription_callback.hpp) in any_subscription_callback, except we would do it with [`SubscriptionTopicStatistics`](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp#L107-L115) and `handle_message`. Something along the lines of : 
```
virutal void handle_message(
    const rclcpp::SerializedMessage & received_message,
    const rclcpp::Time now_nanoseconds) const
  {
    for (const auto & collector : subscriber_statistics_collectors_) {
      collector->OnMessageReceived(received_message, now_nanoseconds.nanoseconds());
    }
  }
```
From there we would also have to implement a `SubscriptionTopicStatistics` object within the `GenericSubscription` class with a `typename T` of rclcpp::Serialized message, assuming it would still work with it.

MichaelOrlov: @CursedRock17 As regards to the 
> If rosbag2 implements using GenericSubscription and this really is as uncommon as you say, then maybe this should be down in the handle_serialized_message method in that [class](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/generic_subscription.cpp#L50-L55).

In this case, I would **not** approve this motion since in the current design and implementation it will really affect recording performance when libstatistics is enabled.

As regards your request and proposal 
> I am interested in passing these serialized messages to topic statistics, while staying serialized. 

And what are lib topic statistics supposed to do with those serialized messages?   
Deserialize inside? - IMO it will be the same performance burden but on the downstream layer and it doesn't make sense at all.

As I have said above if will extend or change API for the `libstatistic_collector` it will make sense to pass `message_info` structure instead of the real message.

fujitatomoya: using `message_info` does help here. and i think that is the way it should be? enabling statistics requires deserialization for message packet will be huge burden for user application. system should honor the behavior requested by the application to keep the message payload serialized.

fujitatomoya (changes requested):

rclcpp/include/rclcpp/subscription.hpp
CursedRock17: So is this something we bring to the [`libstatistic_collector`](www.github.com/ros-tooling/libstatistics_collector) library now, or do we wait for for necessary changes to be made? The library is quite low on issues and pull requests as of now, but I wouldn't want to release a bunch of breaking changes just for a more explicit addition to rclcpp.

MichaelOrlov: @CursedRock17 We have a lack of resources for adding a new feature in `libstatistics_collector` if you have time and are willing to help it will be very appreciated.  
I think the right way to handle it will be to create a new issue and PR in `libstatistic_collector` repo for adding a new API with `message_info` as a parameter and then deprecate the old one.

CursedRock17: I would definitely be willing to work on an issue and pull request for the library, but where else could `libstatistic_collector` get us out of this feature, for example could we then use it for the other `handle_x_message` functions after we deprecate the old `const T & recieved_message`?

MichaelOrlov: As regards
> could we then use it for the other handle_x_message functions after we deprecate the old const T & recieved_message

I am a bit confused with the question. Why not? The `message_info` seems available in all handlers.

CursedRock17: Oh I knew we could use `message_info` just wanted some clarification of proceeding that way, it shouldn't change the time/space complexity of those functions either, because we're passing both `message_info` and `received_message` by const ref. I'm going to go ahead and add both an issue and pr to the library.

fujitatomoya: according to this discussion, what about using `rmw_message_info_t` for statistics collector? that said, we can remove the dependency on `message` from here (agnostic from `CallbackMessageT`). that seems to be reasonable change to me, and clear design.

CursedRock17: Just for clarification, do you mean the TopicStatistic class within the rclcpp or the TopicStatistic in `libcollector` library. I brought up the type utilized within the issue I opened for that library. I'm not entirely familiar with the rmw_message_info object, could we just make changes in this library without having to change the `libcollector` library's method `OnMessageReceived`, because if so we should just make that change here in rclcpp.

fujitatomoya: > do you mean the TopicStatistic class within the rclcpp or the TopicStatistic in libcollector library. 

`libstatisticscollector`.

> could we just make changes in this library without having to change the libcollector library's method OnMessageReceived

libcollector library's method OnMessageReceived is the one which takes `CallbackMessageT`. that is the reason that we need to deserialize the message here. probably we need to change that anyway to take `rmw_message_info_t` like structure. (i am saying this since libcollector library is not dependent on rmw, should not be, i think)

CursedRock17: Perfect, yea I agree with what you're saying. I linked the issue I made in `libcollector`, just to make sure it's maintainers were ok with us changing to `rmw_message_info`. Also like you said `CallbackMessageT`and all the templating for those function should be deprecated since we'll already know the type. So once they agree in the issue and I make the PR, I'll come back here and change our `handle_x_message` functions.

CursedRock17:
With the closing of [https://github.com/ros-tooling/libstatistics_collector/issues/168](issue 168) in `libstatistics_collector` we can now just use `rmw_message_info_t` as our argument for `OnMessageReceived`. The issue is [`handle_message`](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp#L107-L115) uses a custom typed `topic_statistics_collector::TopicStatisticsCollector`, therefore `OnMessageReceived` uses a typed message. 

Now obviously we are going to get rid of these types as they will cause warnings, but is this PR the appropriate place? Should we push this PR, then create a brand new pull request, maybe called "Removing Topic Statistics Types" or should we just rename this pull request and do it all in here?

fujitatomoya:
> Now obviously we are going to get rid of these types as they will cause warnings, but is this PR the appropriate place?

IMO, either is fine. i would say it is up to you.

personally i would create PR dedicated issue or title, so it would be easier to understand and review. but this one is obviously related to this PR, after all either way you go, i am good to go with you.

CC: @MichaelOrlov

CursedRock17:
> personally i would create PR dedicated issue or title

I'm certainly on board with this, though if we do this I'm just going to comment out the logic of `handle_message` as it would have to keep the `DeserializedMessage` logic that we want to get rid of. So if that's alright I can pursue that route, especially if the PR removing all the warnings shouldn't be too rough since there's some discussion here.

MichaelOrlov:
@CursedRock17 I am a bit lost in your suggestions and don't quite understand why you want to comment out only 
```
subscription_topic_statistics_->handle_message(deserialized_message, time);
```

The proper way to follow up on our resolved https://github.com/ros-tooling/libstatistics_collector/issues/168 would be to create and merge a separate PR first for making `rclcpp::topic_statistics::SubscriptionTopicStatistics` as non templated to force it using `libstatistics_collector::TopicStatisticsCollector`, `libstatistics_collector::ReceivedMessageAgeCollector`  and `libstatistics_collector::ReceivedMessagePeriodCollector` type aliases as underlying types and using `rclcpp::MessageInfo` in its [handle_message](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp#L107-L115)
It will require bigger and unrelated changes to the current PR. 
Perhaps we can name those PR something like `Use message_info in rclcpp::topic_statistics::SubscriptionTopicStatistics instead of typed message`

Then we can easily make changes in the `rclcpp::subscription::handle_serialized_message(..)` per this PR.

CursedRock17:
@MichaelOrlov I was thinking of about merging this one first, to be clear you want to make a new PR with the correct untyped TopicStatistics first, merge that new one, then rebase this current PR on top of that? I would much rather do it this way if possible, it seems cleaner.

MichaelOrlov:
@CursedRock17 Yes, this is correct.

MichaelOrlov:
@CursedRock17 I have created a follow up PR [Use message_info in SubscriptionTopicStatistics instead of typed message](https://github.com/ros2/rclcpp/pull/2337)

CursedRock17:
@MichaelOrlov sounds great 👍, I've run a slight review

MichaelOrlov:
@CursedRock17 Th #2337 has been merged. We can easily finish this PR now, I think.

CursedRock17:
That last commit should do it, do we need to do any sort of rebasing for the CI, or will it work with the latest branch?

clalancette:
> That last commit should do it, do we need to do any sort of rebasing for the CI, or will it work with the latest branch?

You are definitely going to need to rebase, there's been some changes that have gone in to rcl in the meantime.

CursedRock17:
I did the rebase, it seems now the current build check failure, is related to `rcl`, relating to pull request #2344, which makes me believe this is fine. Also, the way we handle the message is synonymous to the other two `handle_x_message` methods.

fujitatomoya (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19713/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14223/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20442/)

fujitatomoya:
@clalancette since you left some comments before, can we have review again?

rclcpp/include/rclcpp/subscription.hpp
```diff
@@ -44,6 +44,7 @@
 #include "rclcpp/subscription_base.hpp"
 #include "rclcpp/subscription_options.hpp"
 #include "rclcpp/subscription_traits.hpp"
+#include "rclcpp/serialization.hpp"
```
MichaelOrlov, line 47: @CursedRock17 We probably don't need to include `rclcpp/serialization.hpp` anymore. Since we are not going to do deserialization.

CursedRock17: Good catch.

clalancette (approved):

clalancette:
Oh, the latest commit needs to have a `Signed-off-by` line, then we can re-run one last CI on this.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19720/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14228/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20447/)

fujitatomoya:
@CursedRock17 thank you for the contribution.

Issue #2273, opened 2023-08-15
Can't enable set_logger_levels service for LifecycleNode


bijoua29:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries
- Version or commit hash:
  - latest rolling as of 2023-08-14
- DDS implementation:
  - RTI Connext
- Client library (if applicable):
  - rclcpp lifecycle_node

#### Steps to reproduce issue

The enable_logger_service for NodeOptions doesn't seem to be enabled for LifecycleNode
```
node_options.enable_logger_service(true);
```
Passing node_options to creation of LifecycleNode doesn't enable the logger service

#### Expected behavior
The set_logger_levels service should be available when doing a 'ros2 node info' on the lifecycle node

#### Actual behavior
The set_logger_levels service is not seen when doing a 'ros2 node info' on the lifecycle node

#### Additional information

I notice creating a non-lifecyle node i.e. a standard node,  has the following code in the constructor

```
if (options.enable_logger_service()) {
    node_logging_->create_logger_services(node_services_);
}
```
However this doesn't exist in the constructor for the lifecycle node.

bijoua29:
Hmm. Ok. Is it as simple as adding 

```
if (options.enable_logger_service()) {
    node_logging_->create_logger_services(node_services_);
}
```

in the LifecycleNode constructor or am I missing something?

fujitatomoya:
implementation would be straight-forward to add the services, but test is also ideal. i think i can work on this.

bijoua29:
> implementation would be straight-forward to add the services, but test is also ideal. i think i can work on this.

Thanks @fujitatomoya

clalancette:
This was solved by #2277, so I'm going to close this out.

PR #2277, opened 2023-08-17
add logger level service to lifecycle node.

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/2273

### DIFF
--- a/rclcpp_lifecycle/src/lifecycle_node.cpp
+++ b/rclcpp_lifecycle/src/lifecycle_node.cpp
@@ -144,6 +144,10 @@ LifecycleNode::LifecycleNode(
       &LifecycleNodeInterface::on_deactivate, this,
       std::placeholders::_1));
   register_on_error(std::bind(&LifecycleNodeInterface::on_error, this, std::placeholders::_1));
+
+  if (options.enable_logger_service()) {
+    node_logging_->create_logger_services(node_services_);
+  }
 }
 
 LifecycleNode::~LifecycleNode()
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
@iuhilnehc-ynos i am hesitating to add whole test for `LifecycleNode` because it has the same object `node_logging_` with `Node`. we can add the test under `rclcpp_lifecycle/test` to make sure `LifecycleNode` creates the logging level services as `Node` does. but i do not want to add unnecessary test that could increase the test time. what do you think?

iuhilnehc-ynos:
> i do not want to add unnecessary test that could increase the test time. what do you think?

Yeah, it seems unnecessary for `lifecycle_node` to add the whole test similar to what the `node` did.
I think we can just add a very simple test to check whether the two services(set_logger_levels and get_logger_levels) are enabled (if  node.option(`enable_logger_service`) is true ) or not (if using the default node.option), something like

https://github.com/ros2/rclcpp/blob/a7a9b78feebff603e5c7c5506ce31b89916bde36/rclcpp_lifecycle/test/test_lifecycle_node.cpp#L670-L674

fujitatomoya:
@iuhilnehc-ynos requesting final review. i just added the test to check if the service is available or not as you suggested with https://github.com/ros2/rclcpp/pull/2277#issuecomment-1681803846

iuhilnehc-ynos (approved):

fujitatomoya:
@iuhilnehc-ynos thanks for the review.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19396/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13925/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20117/)

fujitatomoya:
@clalancette I will go ahead to merge this with @iuhilnehc-ynos 's approval.

btw, do we want to backport this to iron? originally this feature is introduced on iron and we can keep the ABI. personally i think it would be nicer for user, since `LifecycleNode` can have the same feature. what do you think?

fujitatomoya:
@Mergifyio backport to iron

Issue #2265, opened 2023-08-04
Action Client: Feeback callback called after cancel


jmachowinski:
## Bug report

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries
- Version or commit hash:
  - iron
- DDS implementation:
  - cyclone
- Client library (if applicable):
  - rclcpp_action

#### Steps to reproduce issue
Use mutlithreaded spinners

Set up a server that sends a lot of feedback

Set up a client
Cancel the goal

#### Expected behavior
After async_cancel goal the feedback callback should not be called any more

#### Actual behavior
The feedback callback is called after cancel

#### Additional information
This leads to segfaults for us, as we delete the objects associated with the 
ClientGoalHandle once its cancelled.

Also node, we can't check the status of the goal in the callback, as it deadlocks (see https://github.com/ros2/rclcpp/issues/2243). 

#### Implementation considerations
For a solution I would propose to to check the status of the handle in https://github.com/ros2/rclcpp/blob/ea29c142af320bff06f957ec9172caf3a4aff41a/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L163
and not call the feedback callback if the status of the is not STATUS_EXECUTING

CursedRock17:
So I set this up and can see what you're saying, but just as a clarifier where are you cancelling the goal in the code. I have an example set up just like #2243, where I cancel the goal before sending as well as after sending and it just throws an exception and terminates
```
exception of type rclcpp_action::exceptions::UnknownGoalHandleError: Goal handle is not known to this client.
```
which is the expected behavior. The real question is how much should this be down to the user to not: cancel the goal, then try to send it?

fujitatomoya:
@jmachowinski i am not really sure what the problem here, could you elaborate a bit?

> After async_cancel goal the feedback callback should not be called any more

what if the cancel request is rejected? we cannot assume cancel request is always accepted by server.

> The feedback callback is called after cancel

i think server publishes the feedback until the cancel request has been accepted on server side.

jmachowinski:
@CursedRock17 I got a huge system were multiple goals are send to one server, and either the client or the server cancels  them depending on various conditions. 

The crashes I am seeing seem to originate from two race conditions:
- Feedback, status goal cancel and result are all transported on different channels within rcl. Therefore there is no guaranteed order in which they arrive. 
- There is a race condition between the drop of the shared pointer the the goal handle and the callbacks.

[...]

PR #2281, opened 2023-08-21
Callback after cancel

jmachowinski:
fixes https://github.com/ros2/rclcpp/issues/2265

Note, this commit fixes a bug, that makes the current version of the tutorial 
not work any more, as the code now acts according to its documentation.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -409,8 +409,17 @@ public:
    * If the goal is rejected by an action server, then the future is set to a `nullptr`.
    *
    * The returned goal handle is used to monitor the status of the goal and get the final result.
-   * It is valid as long as you hold a reference to the shared pointer or until the
-   * rclcpp_action::Client is destroyed at which point the goal status will become UNKNOWN.
+   *
+   * You will receive callbacks for the goal, as long as you hold a reference to the shared pointer,
+   * or rclcpp_action::Client is destroyed. Dropping the shared pointer will not cancel the goal. In order
+   * to cancel it, you must explicitly call async_cancel_goal.
+   *
+   * WARNING this method has inconsistent behaviour and a memory leak bug.
+   * If you set the result callback, the handle will be self referencing, and you will receive callbacks
+   * even though you do not hold a reference to the shared pointer. In this case, the self reference will
+   * be deleted if the result callback was received. If there is no result callback, there will be a memory leak.
+   *
+   * To prevent the memory leak, you may call stop_callbacks() explicit. This will delete the self reference.
    *
    * \param[in] goal The goal request.
    * \param[in] options Options for sending the goal request. Contains references to callbacks for
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -562,6 +562,35 @@ public:
     return async_cancel(cancel_request, cancel_callback);
   }
 
+  /// Stops the callbacks for the goal in a thread safe way
+  /**
+   * This will NOT cancel the goal, it will only stop the callbacks.
+   *
+   * After the call to this function, it is guaranteed that there
+   * will be no more callbacks from the goal. This is not guaranteed
+   * if multiple threads are involved, and the goal_handle is just
+   * dropped.
+   *
+   * \param[in] goal_handle The goal were the callbacks shall be stopped
+   */
+  void stop_callbacks(typename GoalHandle::SharedPtr goal_handle)
+  {
+    goal_handle->set_feedback_callback(typename GoalHandle::FeedbackCallback());
+    goal_handle->set_result_callback(typename GoalHandle::ResultCallback());
+
+    std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+    const GoalUUID & goal_id = goal_handle->get_goal_id();
+    if (goal_handles_.count(goal_id) == 0) {
+      // someone else already deleted the entry
+      // e.g. the result callback
+      RCLCPP_DEBUG(
+        this->get_logger(),
+        "Given goal is unknown. Ignoring...");
+      return;
+    }
+    goal_handles_.erase(goal_id);
+  }
+
   /// Asynchronously request all goals at or before a specified time be canceled.
   /**
    * \param[in] stamp The timestamp for the cancel goal request.
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -448,7 +448,7 @@ public:
         std::shared_ptr<GoalHandle> goal_handle(
           new GoalHandle(goal_info, options.feedback_callback, options.result_callback));
         {
-          std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+          std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
           goal_handles_[goal_handle->get_goal_id()] = goal_handle;
         }
         promise->set_value(goal_handle);
@@ -466,7 +466,7 @@ public:
     // To prevent the list from growing out of control, forget about any goals
     // with no more user references
     {
-      std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+      std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
       auto goal_handle_it = goal_handles_.begin();
       while (goal_handle_it != goal_handles_.end()) {
         if (!goal_handle_it->second.lock()) {
@@ -496,7 +496,7 @@ public:
     typename GoalHandle::SharedPtr goal_handle,
     ResultCallback result_callback = nullptr)
   {
-    std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+    std::lock_guard<std::recursive_mutex> lock(goal_handles_mutex_);
     if (goal_handles_.count(goal_handle->get_goal_id()) == 0) {
       throw exceptions::UnknownGoalHandleError();
     }
@@ -531,7 +531,7 @@ public:
     typename GoalHandle::SharedPtr goal_handle,
     CancelCallback cancel_callback = nullptr)
   {
-    std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+    std::lock_guard<std::recursive_mutex> lock(goal_handles_mutex_);
     if (goal_handles_.count(goal_handle->get_goal_id()) == 0) {
       throw exceptions::UnknownGoalHandleError();
     }
@@ -578,9 +578,10 @@ public:
     goal_handle->set_feedback_callback(typename GoalHandle::FeedbackCallback());
     goal_handle->set_result_callback(typename GoalHandle::ResultCallback());
 
-    std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+    std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
     const GoalUUID & goal_id = goal_handle->get_goal_id();
-    if (goal_handles_.count(goal_id) == 0) {
+    auto it = goal_handles_.find(goal_id);
+    if (goal_handles_.end() == it) {
       // someone else already deleted the entry
       // e.g. the result callback
       RCLCPP_DEBUG(
@@ -588,7 +589,34 @@ public:
         "Given goal is unknown. Ignoring...");
       return;
     }
-    goal_handles_.erase(goal_id);
+    goal_handles_.erase(it);
+  }
+
+  /// Stops the callbacks for the goal in a thread safe way
+  /**
+   * For futher information see stop_callbacks(typename GoalHandle::SharedPtr goal_handle)
+   */
+  void stop_callbacks(const GoalUUID & goal_id)
+  {
+    typename GoalHandle::SharedPtr goal_handle;
+    {
+      std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
+      auto it = goal_handles_.find(goal_id);
+      if (goal_handles_.end() == it) {
+        // someone else already deleted the entry
+        // e.g. the result callback
+        RCLCPP_DEBUG(
+          this->get_logger(),
+          "Given goal is unknown. Ignoring...");
+        return;
+      }
+
+      goal_handle = it->lock();
+    }
+
+    if (goal_handle) {
+      stop_callbacks(goal_handle);
+    }
   }
 
   /// Asynchronously request all goals at or before a specified time be canceled.
@@ -619,7 +647,7 @@ public:
   virtual
   ~Client()
   {
-    std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+    std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
     auto it = goal_handles_.begin();
     while (it != goal_handles_.end()) {
       typename GoalHandle::SharedPtr goal_handle = it->second.lock();
@@ -666,7 +694,7 @@ private:
   void
   handle_feedback_message(std::shared_ptr<void> message) override
   {
-    std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+    std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
     using FeedbackMessage = typename ActionT::Impl::FeedbackMessage;
     typename FeedbackMessage::SharedPtr feedback_message =
       std::static_pointer_cast<FeedbackMessage>(message);
@@ -703,7 +731,7 @@ private:
   void
   handle_status_message(std::shared_ptr<void> message) override
   {
-    std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+    std::lock_guard<std::recursive_mutex> guard(goal_handles_mutex_);
     using GoalStatusMessage = typename ActionT::Impl::GoalStatusMessage;
     auto status_message = std::static_pointer_cast<GoalStatusMessage>(message);
     for (const GoalStatus & status : status_message->status_list) {
@@ -752,7 +780,7 @@ private:
           wrapped_result.goal_id = goal_handle->get_goal_id();
           wrapped_result.code = static_cast<ResultCode>(result_response->status);
           goal_handle->set_result(wrapped_result);
-          std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+          std::lock_guard<std::recursive_mutex> lock(goal_handles_mutex_);
           goal_handles_.erase(goal_handle->get_goal_id());
         });
     } catch (rclcpp::exceptions::RCLError & ex) {
@@ -784,7 +812,7 @@ private:
   }
 
   std::map<GoalUUID, typename GoalHandle::WeakPtr> goal_handles_;
-  std::mutex goal_handles_mutex_;
+  std::recursive_mutex goal_handles_mutex_;
 };
 }  // namespace rclcpp_action
 
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -405,18 +405,19 @@ public:
 
   /// Send an action goal and asynchronously get the result.
   /**
-   * If the goal is accepted by an action server, the returned future is set to a `ClientGoalHandle`.
+   * If the goal is accepted by an action server, the returned future is set to a `GoalHandle::SharedPtr`.
    * If the goal is rejected by an action server, then the future is set to a `nullptr`.
    *
-   * The returned goal handle is used to monitor the status of the goal and get the final result.
+   * The goal handle in the future is used to monitor the status of the goal and get the final result.
    *
-   * You will receive callbacks for the goal, as long as you hold a reference to the shared pointer,
-   * or rclcpp_action::Client is destroyed. Dropping the shared pointer will not cancel the goal. In order
-   * to cancel it, you must explicitly call async_cancel_goal.
+   * If callbacks were set in @param options, you will receive callbacks, as long as you hold a reference
+   * to the shared pointer contained in the returned future, or rclcpp_action::Client is destroyed. Dropping
+   * the shared pointer to the goal handle will not cancel the goal. In order to cancel it, you must explicitly
+   * call async_cancel_goal.
    *
    * WARNING this method has inconsistent behaviour and a memory leak bug.
-   * If you set the result callback, the handle will be self referencing, and you will receive callbacks
-   * even though you do not hold a reference to the shared pointer. In this case, the self reference will
+   * If you set the result callback in @param options, the handle will be self referencing, and you will receive
+   * callbacks even though you do not hold a reference to the shared pointer. In this case, the self reference will
    * be deleted if the result callback was received. If there is no result callback, there will be a memory leak.
    *
    * To prevent the memory leak, you may call stop_callbacks() explicit. This will delete the self reference.
--- a/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
+++ b/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp
@@ -75,6 +75,7 @@ ClientGoalHandle<ActionT>::set_result(const WrappedResult & wrapped_result)
   result_promise_.set_value(wrapped_result);
   if (result_callback_) {
     result_callback_(wrapped_result);
+    result_callback_ = ResultCallback();
   }
 }
 


rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -562,6 +562,18 @@ class Client : public ClientBase
     return async_cancel(cancel_request, cancel_callback);
   }
 
+  void drop_goal_handle(typename GoalHandle::SharedPtr goal_handle)
```
fujitatomoya, line 565: besides https://github.com/ros2/rclcpp/issues/2265#issuecomment-1685839599, i really do not understand why user needs this API to drop the goal handle. user application can drop the `std::shared_future<typename GoalHandle::SharedPtr>` anytime, that tells the system that user is not interested in that anymore.

> Now the handle is dropped, and the object the callbacks are pointing too is invalid.

Why? in that case, `if (!goal_handle) {` comes true? and which is also expected behavior that user does not need it anymore? I am not sure if that gets to the `calling invalid callback`.

```c++
    typename GoalHandle::SharedPtr goal_handle = goal_handles_[goal_id].lock();
    // Forget about the goal if there are no more user references
    if (!goal_handle) {
      RCLCPP_DEBUG(
        this->get_logger(),
        "Dropping weak reference to goal handle during feedback callback");
      goal_handles_.erase(goal_id);
      return;
    }
    auto feedback = std::make_shared<Feedback>();
    *feedback = feedback_message->feedback;
    goal_handle->call_feedback_callback(goal_handle, feedback);
```

fujitatomoya: refer to https://github.com/ros2/rclcpp/issues/2265#issuecomment-1685839599

> We jump back to Thread1. This thread has a internal copy of the shared pointer to the handle, and will call the callback.

as mentioned above, at this moment, `rclcpp` has the shared pointer to the managed object `goal_handle` so this will be valid.

https://github.com/ros2/rclcpp/blob/fd229d72ff498faf6583c1cc1f0f2f131575b2cf/rclcpp_action/include/rclcpp_action/client.hpp#L662

are you saying the `feedback_callback_` will be NULL?

https://github.com/ros2/rclcpp/blob/fd229d72ff498faf6583c1cc1f0f2f131575b2cf/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L171-L177

jmachowinski:
```c++
     class Foo {
         GoalHandle::SharedPtr handle;
         void callback();

         //calls async_cancel_goal()
         void cancel_goal();
     };
     std::unique_ptr<Foo> bar = std::make_unique<Foo>();
``` 
handle->feedback_callback points to the method 'callback' of the instance of Foo.
The moment you reset the bar ptr. You can get the race and a call into no longer valid memory.
E.g. the deleted instance of Foo.

jmachowinski: @fujitatomoya Did my last comment answer you question, do you understand the issue now ?

fujitatomoya: No, not really.

> are you saying the feedback_callback_ will be NULL?

so the answer is yes with your sample?

or some of `Foo` object will be destroyed, and after that using that object in the feedback callback would be the problem? but that because user application intentionally frees the memory?

i would like to hear 2nd opinion here.

jmachowinski: > are you saying the feedback_callback_ will be NULL?

No, feedback_callback_ contains a value, pointing to &Foo::callback 

> or some of Foo object will be destroyed, and after that using that object in the feedback callback would be the problem? 

Yes

> but that because user application intentionally frees the memory?

Yes, but as you can see, we are in a valid code path, we candled the goal, and afterwards destroyed the object.
And this is exactly the reason, why we need the drop_goal_handle function.

@fujitatomoya can you ping someone for this ?
@kohrt I don't know how to explain it any better, can you give it a try ?

Barry-Xu-2018: @jmachowinski @fujitatomoya 

I understand this change.   

At first, I describe the first problem.
During checking code of action client, I found the current implementation has the problem. 

According to the below declaration, action client only want to keep weak pointer to GoalHandle. (That is, users are responsible for holding the returned shared pointer of GoalHandle)
https://github.com/ros2/rclcpp/blob/ba175922d3008704c67b10fe82a2f7c0cbee2be4/rclcpp_action/include/rclcpp_action/client.hpp#L757

But actually, action client holds the shared pointer of GoalHandle at here
https://github.com/ros2/rclcpp/blob/ba175922d3008704c67b10fe82a2f7c0cbee2be4/rclcpp_action/include/rclcpp_action/client.hpp#L702-L733

The codes should be changed to use weak pointer of GoalHandle here.    

Current example code have the problem (Not hold returned GoalHandle). This is mentioned by your discussion (https://github.com/ros2/rclcpp/issues/2265#issuecomment-1668389518). I check test codes, returned GoalHandle are all held. So I believe the original design intended for users to hold the returned GoalHandle.

The above modification can partially address issue #2265, but it cannot completely solve the problem.    

The cause is

```c++
     class Foo {
         GoalHandle::SharedPtr handle;
         void callback();  // set to call_feedback_callback of GoalHanle
     };
     std::unique_ptr<Foo> bar = std::make_unique<Foo>();
     bar.reset();
```

While calling handle_feedback_message, the object bar is destroyed after `goal_handle = goal_handles_[goal_id].lock();`
Calling goal_handle->call_feedback_callback() still lead to access destroyed resource.

```c++
  void
  handle_feedback_message(std::shared_ptr<void> message) override
  {
    std::lock_guard<std::mutex> guard(goal_handles_mutex_);
    ....
    typename GoalHandle::SharedPtr goal_handle = goal_handles_[goal_id].lock();
    // Forget about the goal if there are no more user references
    if (!goal_handle) {
      RCLCPP_DEBUG(
        this->get_logger(),
        "Dropping weak reference to goal handle during feedback callback");
      goal_handles_.erase(goal_id);
      return;
    }
    auto feedback = std::make_shared<Feedback>();
    *feedback = feedback_message->feedback;
    goal_handle->call_feedback_callback(goal_handle, feedback);
  }
```

So add a new function `drop_goal_handle()` to directly remove GoalHanle in goal_handles_can. This operation avoid this special case.  `drop_goal_handle()` is called in destructor function of Foo.

fujitatomoya: @Barry-Xu-2018 thanks for the explanation, i was asking that via https://github.com/ros2/rclcpp/pull/2281/files#r1304621676. but a bit confused...

now it is clear for me.

clalancette: So I took a look at this issue as well.  Thanks to the excellent explanations by @jmachowinski and @Barry-Xu-2018 , I was able to understand the issue.

While this solution will fix the problem, I think we should explore the solution space further before we commit to this one.  The problem in my mind with this solution is that it adds to the already complex rclcpp_action API, and it will be extremely easy for users to miss calling this in the appropriate places.  In the latter case, users will likely run just fine for a long time without it, until they change to a multi-threaded executor or happen to hit the race.  So I think doing something more "automatic" might be nice, if we can get there.

So with that said, here are some ideas (just throwing them out there, I'm not 100% certain of the feasibility of each of them):

1.  This solution, which has the downsides I've mentioned above but definitely fixes the issue.
2.  Change the parameter type of the feedback, result, and response callbacks from a `GoalHandle::SharedPtr` to something different.  This "something different" would act like a shared/weak pointer, but with an important difference; it would hold a (blocking) mutex while altering the reference count.  That would ensure that the user code could *not* drop the reference while another thread was in the middle of using it.  This has some downsides in that it is easier for the user to create a deadlock, but it is a much more automatic solution that doesn't require a new API.
3.  Since the fundamental problem here is that user code is destroying associated data while an object is still active, we could provide some kind of callback to the user code telling it when we are about to use it.  Then the user code could take appropriate steps.  I'm not a huge fan of this solution, as it has most of the downsides of 1, while also being hard to use.  I just threw it in here as something I thought about while mulling this over.

Thoughts on any of the above, or other thoughts on how we could make this more automatic?

fujitatomoya: @jmachowinski i am with @clalancette on this, probably we can explore the ideas a bit.

the followings are just notes, i will share if i can find any better solution.

> This has some downsides in that it is easier for the user to create a deadlock

yeah, this has reported before, that is the reason we release the mutex lock during the callback...

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -708,23 +720,40 @@ class Client : public ClientBase
     }
     using GoalResultRequest = typename ActionT::Impl::GetResultService::Request;
     auto goal_result_request = std::make_shared<GoalResultRequest>();
+    const GoalUUID goal_id = goal_handle->get_goal_id();
```
fujitatomoya, line 723: @jmachowinski @clalancette aside from above discussion, this fixes the another issue. (using weak pointer instead of shared pointer) probably we can proceed this one to address in a part of this issue?

rclcpp_action/include/rclcpp_action/client.hpp
jmachowinski: The only 'automatic' solution I can think of would be to pass a weak ptr to the client to the goal handle. In the destructor the goal handle can than try to aquire the client and deregister itself using the mutex. Additional we can make the mutex recursive to try to prevent deadlocks. @clalancette @fujitatomoya what do you think about this approach ?

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -562,6 +562,18 @@ class Client : public ClientBase
     return async_cancel(cancel_request, cancel_callback);
   }
 
+  void drop_goal_handle(typename GoalHandle::SharedPtr goal_handle)
+  {
+    std::lock_guard<std::mutex> guard(goal_handles_mutex_);
+    const GoalUUID & goal_id = goal_handle->get_goal_id();
+    if (goal_handles_.count(goal_id) == 0) {
+      // someone else already deleted the entry
```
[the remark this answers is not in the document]
fujitatomoya: agree. we could use `RCLCPP_WARN` instead, since this is not expected by user application most likely but proceed.

rclcpp_action/include/rclcpp_action/client.hpp
jmachowinski: @fujitatomoya @clalancette  I openend https://github.com/ros2/rclcpp/pull/2296
to split out the first bug fix.

rclcpp_action/include/rclcpp_action/client.hpp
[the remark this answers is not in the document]
jmachowinski: I added a debug. From my point of view, a warning could spam. The reason is, that I can drop actions, that shortly before this call, we could receive a result callback that gets processed in a second thread. The result callback would also remove the handle from the goal_handle_ map.

rclcpp_action/include/rclcpp_action/client.hpp
clalancette: > The only 'automatic' solution I can think of would be to pass a weak ptr to the client to the goal handle. In the destructor the goal handle can than try to aquire the client and deregister itself using the mutex. Additional we can make the mutex recursive to try to prevent deadlocks. @clalancette @fujitatomoya what do you think about this approach ?

I think it is certainly worth a try to see if that solves the issue.

jmachowinski: Implemented a first version, feedback would be welcome.

fujitatomoya: @jmachowinski thanks for the fix. i think this works.

CC: @clalancette @iuhilnehc-ynos

jmachowinski: @clalancette @iuhilnehc-ynos ping

iuhilnehc-ynos: From the point of my view, it seems that the issue is not being fixed.

The destructor of `ClientGoalHandle` can't be called if the `ClientGoalHandle` is acquired(locked) in the callback.
which means if the user instance is destroyed at the moment, the callback on user side can be called, and then it could be crashed.
(Notice that the reference count of shared_ptr<ClientGoalHandle> with value(2) can not make the ClientGoalHandle destructor called when user destroy the instance which has the `shared_ptr<ClientGoalHandle>`.)

jmachowinski: @iuhilnehc-ynos  You are right, this won't work. 

Anyone an idea of how to solve this, or is the drop_goal_handle the way to go ?

jmachowinski: What about taking the handle_mutex_ in the destructor of the ClientGoalHandle and use code like this :
```c++
      typename GoalHandle::SharedPtr goal_handle = goal_handles_[goal_id].lock();
      if (!goal_handle) {
        return;
      }
      // as long as we have this mutex the destructor of the handle can not run through
      // therefore using a raw ptr will be safe here.
      goal_handle->handle_mutex_.lock();
      
      auto *raw_goal_handle_ptr = goal_handle.get();
      
      // from now in any user code dropping the share_ptr, will trigger the destructor of the ClientGoalHandle
      goal_handle.reset();
      if (!goal_handles_[goal_id].use_count() == 0) {
      {
        // last share_ptr pointing to the handle was deleted while we aquired the raw_goal_handle_ptr
        // note, the mutex was deleted, and we therefore do not call unlock on it.
        return;
      }

      // do something
      
      
      raw_goal_handle_ptr->handle_mutex_.unlock();
```

jmachowinski: Hm, I had a second thought about the solution above, the problem is, that one would need to reinstanciate the shared_ptr for passing it to the callbacl, and the we are back to the race condition. I'm out of ideas....

fujitatomoya: @jmachowinski thanks for consideration.

actually i did not see https://github.com/ros2/rclcpp/pull/2281/files#r1338111143 racy condition, but that could happen. (@iuhilnehc-ynos thanks.)

to be honest, i am not sure either way we go. probably we can keep the issue open until we come up  with feasible solution? any thoughts?

jmachowinski: As this is a fix for a real world problem, that is affecting at least one party (us ;-) ) I would vote for picking a solution within a bounded time frame and then merging it.

alsora:
Some more thoughts on this PR (also discussed in the last Client Library WG meeting).

 - I think that the expected behavior should be that the action client keeps the goals alive unless explicitly instructed to drop them.
 - As @wjwwood mentioned, it's ok that this behavior is different from what we do when creating ROS entities (e.g. if you create a subscription, you are responsible for keeping the subscription shared_ptr in scope)

This means that the documentation is wrong, but the demo code is correct.

Having to hold the future even if you don't do anything with it is a very strange user experience, since `std::future` is a standard C++ class with some expected usage and behavior, we shouldn't overload it with ROS-specific expectations.

I proposed to fix the bug by:
 1. Have the client class store `shared_ptr` rather than `weak_ptr` here https://github.com/ros2/rclcpp/blob/rolling/rclcpp_action/include/rclcpp_action/client.hpp#L757
 2. do not capture the `goal_handle` as a shared_ptr in the lambdas, but rather only as a weak_ptr

This should ensure proper destruction of things

jmachowinski:
> * I think that the expected behavior should be that the action client keeps the goals alive unless explicitly instructed to drop them.

Note, this will break the code of users not setting the callback. 

Just to make sure that we are all on the same page, the current behavior is:
If you give a callback to the result:
- The goal will keep itself alive, as the share_ptr is caputred in the lambda to the result callback

If you don't set a result callback:
- It works as documented, and the goal will terminate the second you drop the shared_ptr from the future

Whatever we decide, we will break the code of someone...

alsora:
> Whatever we decide, we will break the code of someone...

This is true, that's why I think we should rather focus on what we think is the correct behavior.

- **Current rolling code:**
   - inconsistent behavior depending on whether you set a result callback or not described in https://github.com/ros2/rclcpp/pull/2281#issuecomment-2031709370
   - has the bug described in https://github.com/ros2/rclcpp/issues/2265
   - does not respect the documentation
   - the demo examples are ok
- **Proposal 1: require the users to keep the `std::shared_future` (and goal handle) in scope**
   - this is the current state of this PR f574591c068d02af8945e34b83e4b770286549e9
   - abuses the meaning of a `std::future`, which not only is used for waiting, but also to keep the goal alive
   - fixes the bug
   - respects the documentation
   - breaks the demo examples (where the future is ignored)
   - the [[no-discard]] attribute can produce a warning (in most compilers) if users don't hold the future, but this can be easily ignored. If ignored, goals will be unexpectedly cancelled under the users feet.
- **Proposal 2: have the action client automatically keep goals alive and use a dedicated API to drop them**
   - proposed in the Client Library WG meeting and summarized in https://github.com/ros2/rclcpp/pull/2281#issuecomment-2030340944
   - fixes the bug
   - does not respect the documentation
   - the demo examples are ok
   - breaks the code for users that were relying on dropping the future / goal handle to terminate the goal. These users now need to use the dedicated API 

I think that this is a good summary of the the current state and the two proposals

My vote is for "proposal 2"

@clalancette @fujitatomoya @wjwwood @mjcarroll

mjcarroll:
I think @ros2/team is worth pinging here.

jmachowinski:
I vote for 1, as users at least get a compile warning.

Additional I would propose, that we directly return the goal handle by the async_send_goal method, as this would also give us a way to drop the goal, before it was accepted. This is currently a second design flaw in the API.

clalancette:
> > I think that the expected behavior should be that the action client keeps the goals alive unless explicitly instructed to drop them.

> Note, this will break the code of users not setting the callback.

Can you explain more of your thinking around this?

Today, if users are not setting the result callback, then either:
1. They are holding onto the `std::future`, and thus keeping it alive, or
2. They are forgetting to hold onto the handle, and getting UB (though they may get lucky and this may work by accident)

If we change it so that the action client keeps the goals alive, then in case 1., they are unnecessarily holding the `std::future`, and may delay it from being destroyed, but things should still work.  We will fix case 2 for them, in that we'll hold onto the handle.  In both cases, I don't think we make things worse for users who are not setting the result callback.

But it is entirely possible I'm missing something here.

jmachowinski:
> Can you explain more of your thinking around this?

Users that don't set the callback might rely on the fact, that the goal gets canceled the second they drop the handle.
If we change this to proposal 2, their code will silently break, as goals do not terminate any more after they drop the handle.

clalancette:
> Users that don't set the callback might rely on the fact, that the goal gets canceled the second they drop the handle.
> If we change this to proposal 2, their code will silently break, as goals do not terminate any more after the handle gets dropped.

Ah, right.  Thanks.

So I have a proposal, and then an opinion.

The proposal is option 3, where we add in a completely new API for sending goals (`async_send_goal2`, for lack of a better idea).  This API would implement behavior 2 as specified by @alsora , i.e. it would hold onto the handle and have a dedicated API to drop it.  Then we leave `async_send_goal` as it is, with a deprecation warning.  In that case, then, we would have a viable path to telling users that we are going to break their semantics, and we have a "new" way for users to change that they have to explicitly opt into.  It is entirely possible that it is difficult to implement both of these behaviors internally, but I think it at least deserves consideration.

My opinion is this: I like option 3 the best, as it is clear to users that changes to these APIs are coming.  If we don't want to do option 3 today, then I think we should go for option 1, as it is the least surprising for users.  In the future we can always decide to go for option 3 if we think that is better.

Thoughts?

jmachowinski:
I don't think we can implement option 3 until the feature freeze next week. I would also go more into the direction of a whole API redesign, introducing two new client classes, that solve the common 'I called wait on the future and deadlocked' issue.

ros-discourse:

https://discourse.ros.org/t/client-library-wg-meeting-april-5-2024/36998/1

fujitatomoya:
I second proposal-2. (from https://github.com/ros2/rclcpp/pull/2281#issuecomment-2032407192) I am okay with current implementation.

(note, i do not think i can make it tomorrow meeting, depends on traffic.)

fujitatomoya (approved):
still under discussion but lgtm with green CI atm.

jmachowinski:
Summary of the discussion:
We'll add 
```c++
async_send_goal_unowned()
``` 
This version will work as described in the documentation.
We will keep async_send_goal with the current broken behavior, but update the documentation.
Open for discussion : Will we add a [[deprecated, "Deprecated, as the behavior bla bla bla"]] prior to jazzy.

My vote is for adding the deprecated prior to jazzy, as users might then recognize a bug they were experiencing.

fujitatomoya:
> Open for discussion : Will we add a [[deprecated, "Deprecated, as the behavior bla bla bla"]] prior to jazzy.

just checking. so the decision is, to keep `async_send_goal` at this moment with broken behavior, but eventually we will remove `async_send_goal`. and will be replaced with `async_send_goal_unowned`?
Or do we keep both of them?

alsora:
> Or do we keep both of them?

Jazzy will have both of them.
The future is less set in stone =)

jmachowinski:
While implementing the proposed solution I rechecked the code and figured out, that the problem is different than described by me in this thread before. 

The current behavior of the action client library is : 

async_send_goal
- Send a goal, that is fire & forget.  It will be executed regardless if you hold the handle or not.
- The given handle only controls, if you will receive callbacks. 
- The need to keep the handle is inconsistent. 
  - If you only set the feedback callback, you will need to hold the handle, or you will receive no feedback.
  - As soon as you set the result callback the handle is self referencing.

Using the current implementation of async_send_goal will create a memory leak, as the goal handle will not be deleted.
- This can be patched for nominal cases, as we could delete the result callback after the result callback was called.

I am undecided on what to do with the async_send_goal function, still migrate to a new version owned ?

I would propose, to add a new function 
```c++
void stop_callbacks(typename GoalHandle::SharedPtr goal_handle)
``` 
instead of
```c++
void drop_goal_handle(typename GoalHandle::SharedPtr goal_handle)
``` 

I would also update the documentation.

Opinions? @clalancette @fujitatomoya @alsora @mjcarroll @wjwwood

jmachowinski:
@alsora updated docs

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20837/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15141/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21524/)

alsora:
Thank you!
Looks good to me with green CI

alsora (approved):

jmachowinski:
rebased to rolling.

CI seems to hang, can someone restart it ?

wjwwood:
I don't see where it's hung, but I restarted CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20851/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15154/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21537/)

jmachowinski:
Ready for merge ?

alsora (approved):

mjcarroll (approved):

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -562,6 +572,63 @@ class Client : public ClientBase
     return async_cancel(cancel_request, cancel_callback);
   }
 
+  /// Stops the callbacks for the goal in a thread safe way
+  /**
+   * This will NOT cancel the goal, it will only stop the callbacks.
+   *
+   * After the call to this function, it is guaranteed that there
+   * will be no more callbacks from the goal. This is not guaranteed
+   * if multiple threads are involved, and the goal_handle is just
```
fujitatomoya, line 581: this `This is not guaranteed if multiple threads are involved` doc section, why is it not guaranteed if multi-threaded are involved?

fujitatomoya (commented):
@jmachowinski lgtm anyway, but one question for doc section.

rclcpp_action/include/rclcpp_action/client.hpp
jmachowinski: Race conditions

jmachowinski: If you drop the handle without calling this function you have no guarantee that no further callback will happen.

rclcpp_action/include/rclcpp_action/client.hpp
```diff
@@ -562,6 +572,63 @@ class Client : public ClientBase
...
+      if (goal_handles_.end() == it) {
+        // someone else already deleted the entry
+        // e.g. the result callback
+        RCLCPP_DEBUG(
+          this->get_logger(),
+          "Given goal is unknown. Ignoring...");
+        return;
+      }
+
+      goal_handle = it->lock();
```
boale-hexagon, line 624: Should it be `it->second.lock()` instead?

```
../../examples/ros_examples/example_ros_node/src/example_action_client_node.cpp:106:3:   required from here
/nix/store/mmimqyq2k2mlar712s1jg1jq8zz4ris1-rclcpp-action-28.1.9/include/rclcpp_action/rclcpp_action/client.hpp:624:25: error: ‘struct std::pair<const std::array<unsigned char, 16>, std::weak_ptr<rclcpp_action::ClientGoalHandle<example_ros_interface_library::action::ExampleAction> > >’ has no member named ‘lock’
  624 |       goal_handle = it->lock();
      |                     ~~~~^~~~
```

fujitatomoya: see https://github.com/ros2/rclcpp/pull/2958

PR #2290, opened 2023-08-29
Correct the position of a comment.

Manchey:


CursedRock17:
Yea that looks right, can you sign off the DCO check?

Manchey:
> Yea that looks right, can you sign off the DCO check?

Done, thanks for review!

rclcpp/src/rclcpp/client.cpp
```diff
@@ -125,7 +125,6 @@ bool
 ClientBase::wait_for_service_nanoseconds(std::chrono::nanoseconds timeout)
 {
   auto start = std::chrono::steady_clock::now();
-  // make an event to reuse, rather than create a new one each time
   auto node_ptr = node_graph_.lock();
```
fujitatomoya, line 128: i think current place would be fine to have this comment. it actually starts from here getting shared pointer reference via `lock` to reuse the event before accessing it, and that is right below this. i do not think that we necessarily replace the comment for this. but i am also okay to take with other maintainer's approval.

btw, `rclcpp_action` has the same comment, so if we take this PR, we need to address the following as well.

https://github.com/ros2/rclcpp/blob/fd58bddb05ee430f7124f5feff0add683db6f86f/rclcpp_action/src/client.cpp#L166

Manchey: Thanks for your review and explanation! 
It's just kind of confusing for me when seeing _"make an event"_ in the comment and then in the next few lines do something unrelated:
https://github.com/ros2/rclcpp/blob/fd58bddb05ee430f7124f5feff0add683db6f86f/rclcpp/src/rclcpp/client.cpp#L133-L140
Is it necessary to get the `node_graph_` pointer before check `service_is_ready`? Maybe we can check `service_is_ready()` first and then get the `node_graph_` shared pointer?

clalancette: I think it does make sense to move the comment; most of the stuff before this is preliminaries.  So I'm OK with this patch as-is.

Regarding:

> Is it necessary to get the node_graph_ pointer before check service_is_ready? Maybe we can check service_is_ready() first and then get the node_graph_ shared pointer?

We could consider that change, though it does change semantics.  As it stands, if the node goes out of scope for some reason, we'll discover that first and throw the exception.  Only after that we will consider whether the service is ready or whether the check is non-blocking.  If we move the locking of the node pointer to after service_is_ready, then we could return true before checking if the node is still valid.

Personally, I'm inclined to leave the logic as-is and just move the comment.  But maybe others have a different opinion.

CursedRock17:
You should probably just do one more check to make sure there are no other comments that are related, that need to be moved. Once you're done with that can you  
`git rebase HEAD~2 -i` squash that newer commit into the top, then force-push-with-lease like you did for the DCO check. Technically it's correct to make these changes, and I've seen PRs that just fix 1 word typos, but we probably want the commit tree to be as small as possible.

fujitatomoya (approved):

clalancette:
> `git rebase HEAD~2 -i` squash that newer commit into the top, then force-push-with-lease like you did for the DCO check. Technically it's correct to make these changes, and I've seen PRs that just fix 1 word typos, but we probably want the commit tree to be as small as possible.

It's really not necessary as long as every commit has the DCO.  We "Squash and merge" as a matter of course anyway.

clalancette (approved):

clalancette:
Since this is just moving a comment, and the Rpr job passed, I'm going to go ahead and merge it without further CI.

PR #2296, opened 2023-09-05
!fix(Client): Do not hold goal_handle ptr if result cb is set.

jmachowinski:
The capture of the lambda held an copy of the shared_ptr to the goal handle. This lead to the behavior that the goal callbacks would be called, even though the user dropped the handle.

Note, this will break the current tutorial, as it restores the documented behavior that a goal will be invalidated as soon as one drops the handle.

### DIFF
--- a/rclcpp_action/include/rclcpp_action/client.hpp
+++ b/rclcpp_action/include/rclcpp_action/client.hpp
@@ -708,23 +708,40 @@ private:
     }
     using GoalResultRequest = typename ActionT::Impl::GetResultService::Request;
     auto goal_result_request = std::make_shared<GoalResultRequest>();
+    const GoalUUID goal_id = goal_handle->get_goal_id();
     goal_result_request->goal_id.uuid = goal_handle->get_goal_id();
     try {
       this->send_result_request(
         std::static_pointer_cast<void>(goal_result_request),
-        [goal_handle, this](std::shared_ptr<void> response) mutable
+        [goal_id, this](std::shared_ptr<void> response) mutable
         {
+          std::lock_guard<std::mutex> lock(goal_handles_mutex_);
+          if (goal_handles_.count(goal_id) == 0) {
+            RCLCPP_DEBUG(
+              this->get_logger(),
+              "Received result for unknown goal. Ignoring...");
+            return;
+          }
+          typename GoalHandle::SharedPtr goal_handle_ = goal_handles_[goal_id].lock();
+          // Forget about the goal if there are no more user references
+          if (!goal_handle_) {
+            RCLCPP_DEBUG(
+              this->get_logger(),
+              "Dropping weak reference to goal handle during result callback");
+            goal_handles_.erase(goal_id);
+            return;
+          }
+
           // Wrap the response in a struct with the fields a user cares about
           WrappedResult wrapped_result;
           using GoalResultResponse = typename ActionT::Impl::GetResultService::Response;
           auto result_response = std::static_pointer_cast<GoalResultResponse>(response);
           wrapped_result.result = std::make_shared<typename ActionT::Result>();
           *wrapped_result.result = result_response->result;
-          wrapped_result.goal_id = goal_handle->get_goal_id();
+          wrapped_result.goal_id = goal_id;
           wrapped_result.code = static_cast<ResultCode>(result_response->status);
-          goal_handle->set_result(wrapped_result);
-          std::lock_guard<std::mutex> lock(goal_handles_mutex_);
-          goal_handles_.erase(goal_handle->get_goal_id());
+          goal_handle_->set_result(wrapped_result);
+          goal_handles_.erase(goal_id);
         });
     } catch (rclcpp::exceptions::RCLError & ex) {
       // This will cause an exception when the user tries to access the result


fujitatomoya:
@iuhilnehc-ynos can you do 2nd review? i think we can backport this to iron and humble.

iuhilnehc-ynos:
> This lead to the behavior that the goal callbacks would be called, even though the user dropped the handle.

From my point of view, this PR doesn't totally fix the issue that the user's callback could be called if the user dropped the handle.

[goal_handle->set_result(wrapped_result);](https://github.com/ros2/rclcpp/blob/5870033007e27fd9f7940ba9bd5e83b47754a783/rclcpp_action/include/rclcpp_action/client.hpp#L743) -> [result_callback_(wrapped_result);](https://github.com/ros2/rclcpp/blob/bf1396b272782e007df7e0f89cb4c6d94452c848/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L77) if the user drop the goal_handle after [typename GoalHandle::SharedPtr goal_handle_ = goal_handles_[goal_id].lock();](https://github.com/ros2/rclcpp/blob/5870033007e27fd9f7940ba9bd5e83b47754a783/rclcpp_action/include/rclcpp_action/client.hpp#L725) is not nullptr.

I think there is a method to fix the issue without this PR.
`GoalHandle::SharedPtr` should be only managed on the user side, not in the https://github.com/ros2/rclcpp/blob/6815022909086487f2b8dc14c0540b00d4cf61a0/rclcpp_action/include/rclcpp_action/client.hpp#L757 since it could be held inside the `rclcpp_action`. The key is to use an extra level(object) of indirect in `rclcpp_action`, which member methods can be called in the destructor of `GoalHandle`.

jmachowinski:
> > This lead to the behavior that the goal callbacks would be called, even though the user dropped the handle.
> 
> From my point of view, this PR doesn't totally fix the issue that the user's callback could be called if the user dropped the handle.
> 
> [goal_handle->set_result(wrapped_result);](https://github.com/ros2/rclcpp/blob/5870033007e27fd9f7940ba9bd5e83b47754a783/rclcpp_action/include/rclcpp_action/client.hpp#L743) -> [result_callback_(wrapped_result);](https://github.com/ros2/rclcpp/blob/bf1396b272782e007df7e0f89cb4c6d94452c848/rclcpp_action/include/rclcpp_action/client_goal_handle_impl.hpp#L77) if the user drop the goal_handle after [typename GoalHandle::SharedPtr goal_handle_ = goal_handles_[goal_id].lock();](https://github.com/ros2/rclcpp/blob/5870033007e27fd9f7940ba9bd5e83b47754a783/rclcpp_action/include/rclcpp_action/client.hpp#L725) is not nullptr.
> 
This is correct, see https://github.com/ros2/rclcpp/pull/2281

Anyhow this fixes commit fixes a part of the issue and is obviously a bug. Therefore I would like to see it merged, before we continue the lengthy discussion on how to fix the complete issue for real.

jmachowinski:
> But there still exists the bug after applying this PR. What if further discussion doesn't need this PR for this issue, why should we merge it first?

This mostly comes down to it fixes 99% of the issue and is a step in the correct direction. This also restores the documented behaviour, that the user needs to hold the goal handle, in order for the action to succeed.
But if the consensus is that we need to get it perfect before merge, I am also fine with this.
 
> 
> I agree with clalancette on the `make this more automatic` mentioned at [#2281 (comment)](https://github.com/ros2/rclcpp/pull/2281#discussion_r1307354723), rather than calling `drop_goal_handle()` manually.

Can you then comment on https://github.com/ros2/rclcpp/pull/2281#discussion_r1307819954 ?

If I know the direction, this should go, I can implement it, and get it off my table ;-P

fujitatomoya:
@iuhilnehc-ynos 
CC: @jmachowinski 

that was my suggestion to separate this PR to merge, i think this can be handled with another problem. actually i was thinking that we should backport this to iron and humble, for doing so separating PR would be much better to fix the specific problem w/o breaking ABI.

> I agree with clalancette on the make this more automatic mentioned at https://github.com/ros2/rclcpp/pull/2281#discussion_r1307354723, rather than calling drop_goal_handle() manually.

yeah, i think we all are on the same page. let's post the idea on https://github.com/ros2/rclcpp/pull/2281#discussion_r1307819954 instead.

fujitatomoya:
that also sounds good to me 👍

iuhilnehc-ynos:
@fujitatomoya 

OK, Let's improve it step by step.

@jmachowinski 

It seems we can merge this PR into `rolling`. Before running CI, could you address https://github.com/ros2/demos/blob/bd49f4b64ee7f65616dccbe37e96fb64a9cad6ee/action_tutorials/action_tutorials_cpp/src/fibonacci_action_client.cpp#L76 and [the document](https://github.com/ros2/rclcpp/blob/6815022909086487f2b8dc14c0540b00d4cf61a0/rclcpp_action/include/rclcpp_action/client.hpp) as well?

jmachowinski:
@iuhilnehc-ynos 
> It seems we can merge this PR into `rolling`. Before running CI, could you address https://github.com/ros2/demos/blob/bd49f4b64ee7f65616dccbe37e96fb64a9cad6ee/action_tutorials/action_tutorials_cpp/src/fibonacci_action_client.cpp#L76

Done

> and [the document](https://github.com/ros2/rclcpp/blob/6815022909086487f2b8dc14c0540b00d4cf61a0/rclcpp_action/include/rclcpp_action/client.hpp#L411) as well?

I don't see anything that is wrong with the documentation.

I was wondering, if we should add [[nodiscard]] to async_send_goal

clalancette:
The fact that we had to open up https://github.com/ros2/demos/pull/649 means we are breaking user semantics.  Because of that, I don't think we should merge this in until we know the full shape of the solution.

Put another way, we *can* break API/semantics to fix bugs (though I will always push for a "deprecation" cycle if at all possible).  But what I don't think we should do is to break API/semantics for a partial fix, and then have to do it again once we have the full fix.  If we are going to have to break API/semantics, we should do it all in one go and let the community know.

jmachowinski:
Closing in favor for https://github.com/ros2/rclcpp/pull/2281

PR #2299, opened 2023-09-06
Cleanup flaky timers_manager tests.

clalancette:
The timers_manager tests were relying too heavily on specific timings; this caused them to be flaky on the buildfarm, particularly on Windows.

Here, we increase the timeouts, and remove one test which just relies too heavily on specific timeouts.  This should make this test much less flaky on the buildfarm.

@alsora @mjcarroll FYI

### DIFF
--- a/rclcpp/src/rclcpp/experimental/timers_manager.cpp
+++ b/rclcpp/src/rclcpp/experimental/timers_manager.cpp
@@ -28,8 +28,8 @@ using rclcpp::experimental::TimersManager;
 TimersManager::TimersManager(
   std::shared_ptr<rclcpp::Context> context,
   std::function<void(const rclcpp::TimerBase *)> on_ready_callback)
-: context_(context),
-  on_ready_callback_(on_ready_callback)
+: on_ready_callback_(on_ready_callback),
+  context_(context)
 {
 }
 
--- a/rclcpp/src/rclcpp/experimental/timers_manager.cpp
+++ b/rclcpp/src/rclcpp/experimental/timers_manager.cpp
@@ -28,9 +28,9 @@ using rclcpp::experimental::TimersManager;
 TimersManager::TimersManager(
   std::shared_ptr<rclcpp::Context> context,
   std::function<void(const rclcpp::TimerBase *)> on_ready_callback)
+: context_(context),
+  on_ready_callback_(on_ready_callback)
 {
-  context_ = context;
-  on_ready_callback_ = on_ready_callback;
 }
 
 TimersManager::~TimersManager()
--- a/rclcpp/test/rclcpp/test_timers_manager.cpp
+++ b/rclcpp/test/rclcpp/test_timers_manager.cpp
@@ -65,8 +65,10 @@ TEST_F(TestTimersManager, empty_manager)
 TEST_F(TestTimersManager, add_run_remove_timer)
 {
   size_t t_runs = 0;
+  std::chrono::milliseconds timer_period(10);
+
   auto t = TimerT::make_shared(
-    10ms,
+    timer_period,
     [&t_runs]() {
       t_runs++;
     },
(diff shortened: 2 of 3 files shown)


alsora:
I would argue that unit-test should not use wall-time, but always some sort of simulated/artificial time to have a deterministic execution.
Is there any plan to do that in the ROS 2 testing stack? Or at least interest (we have an internal solution for that and we may open-source it).

alsora (approved):

rclcpp/test/rclcpp/test_timers_manager.cpp
```diff
@@ -296,12 +237,11 @@ TEST_F(TestTimersManager, timers_thread)
 
   // Run timers thread for a while
   timers_manager->start();
-  std::this_thread::sleep_for(5ms);
+  std::this_thread::sleep_for(50ms);
   timers_manager->stop();
 
   EXPECT_LT(1u, t1_runs);
   EXPECT_LT(1u, t2_runs);
-  EXPECT_EQ(t1_runs, t2_runs);
```
alsora, line 304: what about
```
EXPECT_LE(t1_runs - t2_runs, 1u);
```

clalancette: Yeah, that's fair.  Fixed in f9de6eb6810e69f96e64672c3267ac7ee9360165 (note I had to change the types of `t1_runs` and `t2_runs`, because they now need to be able to be promoted to signed integers for the subtraction).

clalancette:
> Is there any plan to do that in the ROS 2 testing stack?

There are no current plans, no.

> Or at least interest (we have an internal solution for that and we may open-source it).

I will point out that we could do this today by setting `use_sim_time` and controlling the `/clock` topic from the tests.  The problem with that approach in general is that it is slightly different from the "normal" mechanisms, so it wouldn't necessarily be testing the same things.  Whether we actually want to do that in the unit tests is probably highly test-specific.

All of that is to say that I have some interest in it, but we'd have to carefully examine which tests we would switch over to that mechanism.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19467/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13989/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20189/)

alsora (approved):

rclcpp/test/rclcpp/test_timers_manager.cpp
```diff
@@ -296,12 +238,12 @@ TEST_F(TestTimersManager, timers_thread)
...
+  EXPECT_LE(std::abs(t1_runs - t2_runs), 1u);
```
fujitatomoya, line 246: technically `1u` needs to be `1` since this changes `*_runs` into integer?

rclcpp/src/rclcpp/experimental/timers_manager.cpp
```diff
@@ -28,9 +28,9 @@ using rclcpp::experimental::TimersManager;
 TimersManager::TimersManager(
   std::shared_ptr<rclcpp::Context> context,
   std::function<void(const rclcpp::TimerBase *)> on_ready_callback)
+: context_(context),
+  on_ready_callback_(on_ready_callback)
```
fujitatomoya, line 32: i think we should remove the following?

https://github.com/ros2/rclcpp/blob/5e152d77d8144c074894de9a5bc7025aac7f5813/rclcpp/include/rclcpp/experimental/timers_manager.hpp#L530

rclcpp/test/rclcpp/test_timers_manager.cpp
clalancette: Correct.  Fixed in 37edae0.

rclcpp/src/rclcpp/experimental/timers_manager.cpp
clalancette: Sure.  I don't think it matters too much either way, but we are sure it is always going to get set on construction via the initialization variables, so we can remove it.  I've done that in 37edae0.

clalancette:
OK, I've fixed things from review, and also fixed the build warning on Linux.  Let's try CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19481/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14001/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20198/)

fujitatomoya (approved):

PR #2301, opened 2023-09-08
Fix the return type of Rate::period.

clalancette:
In a recent commit (bc435776a257fcf76c5b0124bec26f6824342e34), we reworked how the Rate class worked so it could be used with ROS time as well.  Unfortunately, we also accidentally broke the API of it by changing the return type of Rate::period to a Duration instead of a
std::chrono::nanoseconds .  Put this back to a std::chrono::nanoseconds; if we want to change it to a Duration, we'll have to add a new method and deprecate this one.

@AlexeyMerzlyakov FYI

### DIFF
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -163,7 +163,7 @@ public:
   reset();
 
   RCLCPP_PUBLIC
-  Duration
+  std::chrono::nanoseconds
   period() const;
 
 private:
--- a/rclcpp/src/rclcpp/rate.cpp
+++ b/rclcpp/src/rclcpp/rate.cpp
@@ -14,6 +14,7 @@
 
 #include "rclcpp/rate.hpp"
 
+#include <chrono>
 #include <stdexcept>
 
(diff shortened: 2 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19487/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14007/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20205/)

ahcorde (approved):

SteveMacenski (approved):

fujitatomoya (approved):
Ah 🤕 right, thanks for the fix.

emersonknapp (approved):

AlexeyMerzlyakov:
Yes, the solution works, as we could implicitly convert the return parameter from `std::chrono` -> to `rclcpp::Duration` back using its constructor, if it necessary. New method might be required in the future for performance reasons - to avoid duration->chrono->duration double conversion, but anyway it works for now. @clalancette, thank you for the quick solution on it!

PR #2071, opened 2022-12-23
API url info fix

jrutgeer:
Change README to point at the ROS Index for API 
Cfr. PR https://github.com/ros2/rclcpp/pull/2058
@clalancette

PR #1620, opened 2021-04-05
Change index.ros.org -> docs.ros.org.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

PR #2302, opened 2023-09-08
Update API docs links in package READMEs

christophebedard:
The root README was updated to point to index.ros.org/p/rclcpp (#1620, #2071), but the individual package READMEs, which are displayed by index.ros.org, still link to docs.ros2.org.

### DIFF
--- a/rclcpp_lifecycle/README.md
+++ b/rclcpp_lifecycle/README.md
@@ -2,7 +2,8 @@
 
 Package containing a prototype for lifecycle implementation.
 
-Visit the [rclcpp_lifecycle API documentation](http://docs.ros2.org/latest/api/rclcpp_lifecycle/) for a complete list of its main components and features. For more information about LifeCycle in ROS 2, see the [design document](http://design.ros2.org/articles/node_lifecycle.html).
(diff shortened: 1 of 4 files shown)


clalancette (approved):
Nice catch!  This looks great to me.

rclcpp_lifecycle/README.md
```diff
@@ -2,7 +2,7 @@
 
 Package containing a prototype for lifecycle implementation.
 
-Visit the [rclcpp_lifecycle API documentation](http://docs.ros2.org/latest/api/rclcpp_lifecycle/) for a complete list of its main components and features. For more information about LifeCycle in ROS 2, see the [design document](http://design.ros2.org/articles/node_lifecycle.html).
```
christophebedard, line 5: Oopsies, I removed the link to the lifecycle nodes design document. Re-adding it.

clalancette:
Rpr failed, but that's only because we consider any test failure a failure.  And for some reason the `test_executor` test failed.  Given that this is an update to the README only, I'm going to take that as good enough and merge this as-is.  Thanks for the update!

PR #2304, opened 2023-09-12
Adding API to copy all parameters from one node to another

SteveMacenski:
Per discussion in slack CC @wjwwood 

Templated to support rclcpp::LifecycleNode, nav2_utils::LifecycleNode, and rclcpp::Node interchangably. Unit tests cover the spanning set (e.g. new params added, ensuring old params remain untouched, has nested params). Its overall pretty straight forward.


rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace rclcpp
+{
+
+/**
+ * @brief A method to copy all parameters from one node (parent) to another (child).
+ * May throw parameter exceptions in error conditions
+ * @param parent Node to copy parameters from
+ * @param child Node to copy parameters to
+ */
```
wjwwood, line 31: style nitpick:

```suggestion
/// Copy all parameters from one source node to another destination node.
/**
 * May throw parameter exceptions in error conditions
 * \param source Node to copy parameters from
 * \param destination Node to copy parameters to
 */
```

We use `///` for briefs, and the `\` escape key. Also, I think source and destination are probably better terms. Parent/child implies an on going relationship between the nodes, at least to me, which isn't the case.

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,47 @@
...
+template<typename NodeT1, typename NodeT2>
+void copy_all_parameters(const NodeT1 & parent, const NodeT2 & child)
```
wjwwood, line 33:
```suggestion
void
copy_all_parameters(const NodeT1 & source, const NodeT2 & destination)
```

As mentioned elsewhere, I think source/destination makes more sense.

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <string>
+#include "rcl_interfaces/srv/list_parameters.hpp"
+#include "rclcpp/parameter.hpp"
+
+namespace rclcpp
+{
+
+/**
+ * @brief A method to copy all parameters from one node (parent) to another (child).
+ * May throw parameter exceptions in error conditions
```
wjwwood, line 28: It would be good to enumerate the errors, or probably better to just mention which other methods one could look at to enumerate them, e.g. list/get/set parameters.

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * @param child Node to copy parameters to
+ */
+template<typename NodeT1, typename NodeT2>
+void copy_all_parameters(const NodeT1 & parent, const NodeT2 & child)
+{
+  using Parameters = std::vector<rclcpp::Parameter>;
+  std::vector<std::string> param_names = parent->list_parameters({}, 0).names;
+  Parameters params = parent->get_parameters(param_names);
+  for (Parameters::const_iterator iter = params.begin(); iter != params.end(); ++iter) {
+    if (!child->has_parameter(iter->get_name())) {
```
wjwwood, line 39: This is some implicit behavior that's not well covered in the documentation but there is a case of overlapping parameters in the test.

It would be good to better document what happens in this case, because I was surprised that existing parameters were not overwritten. I don't think one way is obviously better than the other (overwrite vs skip on existing parameter), but it's something to revisit and document the decision in the docblock.

rclcpp/test/rclcpp/test_copy_all_parameters.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "rclcpp/copy_all_parameters.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+class RclCppFixture
+{
+public:
+  RclCppFixture() {rclcpp::init(0, nullptr);}
+  ~RclCppFixture() {rclcpp::shutdown();}
+};
+RclCppFixture g_rclcppfixture;
```
wjwwood, line 25: Is this a pattern copied from somewhere else?

The only pattern I'm familiar with is ones like this:

https://github.com/ros2/rclcpp/blob/ea31f94eb453acf3f34b37b85e38284abc514d57/rclcpp/test/rclcpp/test_node.cpp#L41-L65

Using `TEST_F`.

If we keep the current code, then I think (according to our style) it should be named `RclcppFixture` or maybe `RCLCPPFixture`, but `RclCppFixture` stands out as strange in the code base I think.

wjwwood (changes requested):

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: I actually just went through and found that it cannot throw. Since the nodes were already accepted by one node, this logic cannot throw since its already validated to exist via a valid name and parameter type. The only way it could throw is if a param is attempted to declare if it already exists, if we were to change some of this logic from review comments below. We can follow up down there, but as long as we check if it has a param before declaring it, we're all good

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: I added to the doxygen this behavior. I also added a bool for options about what to do with this - that way its dealer's choice

rclcpp/test/rclcpp/test_copy_all_parameters.cpp
SteveMacenski: Its a pattern we use in Nav2 tests alot. I can't speak to the commonality of it, I've been doing it for so long I don't remember its origin. It works well as a global to setup the init when the test starts and shut it down once all the tests in a file end. I think its nicer than the `TestNode` example, but its not my sandbox. 

I don't feel strongly about it either way. I just updated the naming though to be in line. Happy to adjust should it be requested.

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,51 @@
...
+/// Copy all parameters from one source node to another destination node.
+/**
+ * \param source Node to copy parameters from
+ * \param destination Node to copy parameters to
+ * \param override Default false. Whether to override destination parameters
+ * if both the source and destination contain the same parameter.
+ */
+template<typename NodeT1, typename NodeT2>
+void
+copy_all_parameters(const NodeT1 & source, const NodeT2 & destination, const bool override = false)
```
alsora, line 35: can we replace the template with a `rclcpp::node_interfaces::NodeParametersInterface`?

alsora: `override` is a C++ keyword; I would prefer using a differently named variable

alsora:
Nodes should be independent from one another (e.g. to allow deploy them in different processes and to allow having generic main.cpp), what's the use-case for this API?

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: We could, though I much prefer the templating since it doesn't force the user to include nasty code into their application like `rclcpp::copy_all_parameters(node1->get_node_parameters_interface(), node2->get_node_parameters_interface())` which I think is a far less good user experience. 

I can change override if you like, what would you suggest?

SteveMacenski:
> what's the use-case for this API?

We have some situations where we have a node as a server with some plugins or behavior tree or other application that involves user-provided algorithms and systems. We have an internal node we pass onto those systems so that we protect users from making poor choices that deadlock the main server or cause other issues. This is not an issue for a company's internal systems where you control everything and can expect an engineer to have a relatively sophisticated understanding of the overall use of resources, but matters more for user customization of base frameworks. 

For our example need, we have the BT navigator where we give a client node to the BT's blackboard for nodes to use to create interfaces and callback groups. We'd like those BT nodes to also be able to get the parameters of the BT navigator even though its a separate node technically

rclcpp/include/rclcpp/copy_all_parameters.hpp
alsora: > I can change override if you like, what would you suggest?

IMO `override_params` is already better (I'm fine with everything that isn't just `override`).

> I much prefer the templating since it doesn't force the user to include nasty code into their application

As far as I know, passing node interfaces is the recommended approach, as it better scales to all the different "node-like" classes users may want to implement. Ultimately that's what you care about, i.e. that the object passed as argument has a parameter interface.
This also makes it a lot clearer how users should use the function: at a first read I would expect that `const NodeT1 & source` would expect a `rclcpp::Node`, but it actually expect some sort of pointer (e.g. `shared_ptr<rclcpp::Node>`) as it uses the arrow operator.

SteveMacenski: This isn't in a cpp file that would require explicit instantiation of the types, this would still work with any node-like object. 

If the naming would help changing to `source_node_ptr` or something, that could be done as well! Let me know, I just don't want to piecemeal update the PR 5x with minor changes that don't resolve your concerns. (1) node names and (2) override rename so far

alsora: This would work with the node-like classes that you mentioned (`rclcpp::LifecycleNode`, `nav2_utils::LifecycleNode`, and `rclcpp::Node`), but not with any other custom node-like class that, although it contains a node parameters interface (and thus it should be able to use this function) does not expose the methods used here (`get_parameters`, `set_parameters`, etc.

As a dummy example, I may want to use it with this custom node
```
class MyNode
{
public:
    MyNode();
    rclcpp::node_interfaces::NodeParametersInterface::SharedPtr node_parameters_;
}
```

If directly using parameters interface is not a viable solution, then we should at least use the new `NodeInterface` class (https://github.com/ros2/rclcpp/pull/2041).
This is compatible with both node-like classes and directly passing interfaces.

SteveMacenski: Mhm, this is a good point -- though perhaps a dumb question, couldn't you just literally send it that `node_parameters_`? That is a node like object by itself with those exact APIs defined. I think 

`rclcpp::copy_all_parameters(MyNode1.node_parameters_, MyNode2.node_parameters_)` would work all the same?

SteveMacenski: I generally resist having to make users have to `get_XYZ_base_interface()` in their application code. Its so dirty looking and really mucks things up. I really don't believe that the average ROS user should even need to know about the internal structure of the node enough to be aware of those interfaces to write applications. I know that's not the case right now, today, but I figure I gotta start somewhere :smile:

alsora: With the current implementation of `copy_all_parameters` it's not possible to directly pass a `rclcpp::node_interfaces::NodeParametersInterface` because it doesn't have a method named `set_parameter` (it has `set_parameterS` though).
Changing the implementation of the function (to use `set_parameters` instead of `set_parameter` would IMO not be the best change, as you are relying on node interfaces and nodes-like objects to have the same exact APIs (which is not enforced anywhere).

I think that using the `NodeInterfaces` class introduced in the PR above is likely the best of both worlds.
It will still allow users to directly pass a `rclcpp::Node` or similar object or to pass directly the node interfaces if a node is not available.

SteveMacenski: Are you suggesting something like:

```
template<typename NodeT1, typename NodeT2>
void
copy_all_parameters(const NodeT1 & source, const NodeT2 & destination, const bool override = false)
{
  using Parameters = std::vector<rclcpp::Parameter>;
  auto source_params = NodeInterfaces<NodeParametersInterface>(source).get_node_parameters_interface();
  auto dest_params = NodeInterfaces<NodeParametersInterface>(destination).get_node_parameters_interface();

  std::vector<std::string> param_names = source_params->list_parameters({}, 0).names;
  Parameters params = source_params->get_parameters(param_names);
  for (Parameters::const_iterator iter = params.begin(); iter != params.end(); ++iter) {
    if (!dest_params->has_parameter(iter->get_name())) {
      dest_params->declare_parameter(iter->get_name(), iter->get_parameter_value());
    } else if (override) {
      dest_params->set_parameter(*iter);
    }
  }
}
```

I can see the potential issue with a simple `NodeT` option if I'm just calling direct APIs from it if some non-standard-node doesn't implement the same API as all current nodes. I can get behind that idea; we shouldn't assume _too_ too much about the nature of the Node where reasonable accommodation can be made. 

Separate semi-rant: I went through that PR with `NodeInterfaces` and can understand the value in providing little to no assumptions about the Node-like object - though it feels **dramatically** over-engineered. This is the kind of obscure complexity creep that leaves a sour taste in my mouth when I look over some of the ROS 2 core codebases. I don't think this level of trying to accommodate all possible strange situations via these complex solutions is merited and really draws back on potential contributions to support extremely obscure needs. This solution **still** assumes that every node object implements the `get_node_XYZ_interface()`!  And the XYZ is still hardcoded in the software's interface definitions, so its not like this actually makes it more modular! 

```
    get_from_node_like(NodeT & node_like) \
    { \
      return node_like.get_node_ ## NodeInterfaceName ## _interface(); \
    } \
```

If we place this really minimal and reasonable constraints on all Node-like objects you'd like to be able to interact with the ROS Node-centric API that we implement the `get_node_XYZ_interface()` methods, then we can trivially do

 ```
template<typename NodeT1, typename NodeT2>
void
copy_all_parameters(const NodeT1 & source, const NodeT2 & destination, const bool override = false)
{
  using Parameters = std::vector<rclcpp::Parameter>;
  auto source_params = source->get_node_parameters_interface();
  auto dest_params = destination->get_node_parameters_interface();

  std::vector<std::string> param_names = source_params->list_parameters({}, 0).names;
  Parameters params = source_params->get_parameters(param_names);
  for (Parameters::const_iterator iter = params.begin(); iter != params.end(); ++iter) {
    if (!dest_params->has_parameter(iter->get_name())) {
      dest_params->declare_parameter(iter->get_name(), iter->get_parameter_value());
    } else if (override) {
      dest_params->set_parameter(*iter);
    }
  }
}
```

which is no different really in terms of the core assumptions made without all this casting nonsense that makes the code far less clear in intent. All we're really doing is moving the location of the assumed name and adding in layers and layers of nearly pointless templating! 

This seems like insanity for insanity's sake!

Comment (SteveMacenski) [reply rclcpp/include/rclcpp/copy_all_parameters.hpp:35]: For the scope of this PR - I suppose we just need to come to the agreement about whether to use the NodeInterfaces object or I can just access the interface `getter` to meet the generalization need you mention and I agree with

File: rclcpp/test/rclcpp/test_copy_all_parameters.cpp (hunk 5 above)
Comment (wjwwood) [reply rclcpp/test/rclcpp/test_copy_all_parameters.cpp]: I don't have a problem with it for any strong reason, only that I could imagine it might not play well with gtest if you were to, for example, have it loop tests or something. It seems like a good idea to use the gtest mechanism for setup/teardown if we can.

Why is it nicer, in your opinion?

Comment (wjwwood) [reply rclcpp/test/rclcpp/test_copy_all_parameters.cpp]: Also, it would be more consistent with the other tests. So I would say it would be better to use the gtest style.

Either way, the name should be changed.

File: rclcpp/include/rclcpp/copy_all_parameters.hpp (hunk 6 above)
Comment (wjwwood) [reply rclcpp/include/rclcpp/copy_all_parameters.hpp:35]: I mean we added `NodeInterfaces` for this reason, so it would make sense to use it imo. But we added it to make things easier/better, so if the benefit isn't there for this use case then maybe we should discuss it more.

Mostly I'd say use it just so we have consistency (or at least we're moving in the direction of consistency we want).

Comment (wjwwood) [reply rclcpp/include/rclcpp/copy_all_parameters.hpp:35]: I forgot about it when I reviewed this originally.

**Comment by wjwwood** [comment]
> Nodes should be independent from one another (e.g. to allow deploy them in different processes and to allow having generic main.cpp), what's the use-case for this API?

That's definitely the recommendation, but it's not the only way to use them.

But that brings up a good point. It would make sense to be able to do this with remote nodes (source, destination, or both) too. But if there's no one asking for that, then maybe we should wait on that version.

File: rclcpp/include/rclcpp/copy_all_parameters.hpp [hunk 7]
```diff
@@ -0,0 +1,51 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__COPY_ALL_PARAMETERS_HPP_
+#define RCLCPP__COPY_ALL_PARAMETERS_HPP_
+
+#include <vector>
+#include <string>
+#include "rcl_interfaces/srv/list_parameters.hpp"
+#include "rclcpp/parameter.hpp"
```
wjwwood, line 21: I'm a little surprised the linter didn't complain about this, but I think the style guide says these should be separated into sections:

```suggestion
#include <vector>
#include <string>

#include "rcl_interfaces/srv/list_parameters.hpp"

#include "rclcpp/parameter.hpp"
```

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

rclcpp/include/rclcpp/copy_all_parameters.hpp
wjwwood, line 35: New thread to separate it from the other discussion (can be resolved separately).

Use something other than `override`, maybe `override_existing_parameters`?

wjwwood: > Are you suggesting something like:

Sorry still catching up on the convo, but I think the suggestion is:

```suggestion
copy_all_parameters(NodeInterfaces<NodeParametersInterface> source, NodeInterfaces<NodeParametersInterface> destination, const bool override = false)
```

And everything else is the same?

Well except you can also do things like `auto npi = source.get<NodeParametersInterface>()` instead of `auto npi = source.get_node_parameters_interface()`, which I think is nicer.

Have a look at the test cases for more examples:

https://github.com/ros2/rclcpp/blob/ea31f94eb453acf3f34b37b85e38284abc514d57/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp#L48-L190

wjwwood: Sorry, I guess it's more complex than that, since you're actually doing more than using individual node interfaces, instead you're using methods which may only exist on the `Node` type itself (and some others like the `LifecycleNode` types).

Ideally, these methods on Node would instead be free functions, sort of like how we have free function versions of `create_publisher()`, which is what `Node::create_publisher()` calls. But that might not be the case.

However, if you can limit yourself to what's available on the `NodeParametersInterface` then it will be strictly more portable/flexible. I'd even be in favor of factoring out any syntactic sugar you need from methods on the `Node` class into free functions if that's not already the case, e.g. for things like `declare_parameter`, or `list_parameter`, etc. And it is all syntactic sugar, since they all end up going through `NodeParametersInterface` eventually anyway.

Even more ideally, we'd have only free functions and C++ would implement the "Unified Call Syntax" proposal (https://brevzin.github.io/c++/2019/04/13/ufcs-history/ just one resource, lots on the internet about it), but I won't hold my breathe :(

rclcpp/test/rclcpp/test_copy_all_parameters.cpp
SteveMacenski: It should work for that fine. This simply sets up init at the start of the file's execution and takes it down once the file is completed. If you were to be looping it, each of those executions are separate. 

> Why is it nicer, in your opinion?

Look at the amount of code :smile: Its like a little standard preamble (this is actually automatically generated via my `maketest` script) 

OK, I can change it, no problem.

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: >  it would make sense to use it imo.

@wjwwood its been in the project for 9 months and I just grepped the codebase and it has 0 uses outside of its implementation and unit tests. I would resist making this tiny utility the thing that brings it into production. A larger effort should be made to make other things portable before we start nit-picking new contributions to do so (eg have consistency, like changing tests to use current codebase styling).

Also: 

> copy_all_parameters(NodeInterfaces<NodeParametersInterface> source, NodeInterfaces<NodeParametersInterface> destination, const bool override = false)

looks nasty as heck, unless it automatically casts from the standard node types? Some remark in the original PR makes me think it may (?). Either way, that's an icky looking API that would be quite confusing to try to understand as an external user of the API. That's creating some real insider baseball. 

> you're using methods which may only exist on the Node type itself (

I'd update that one API to use parameter**S** to make it all use the interface API methods that are available. That should be possible, I just didn't update that one line for only illustration purposes :smile:

SteveMacenski: Yeah, will do :+1: 

I'm just not taking action to update the PR until I have my set of action items so I can do them all at once.

SteveMacenski: @wjwwood @alsora thoughts?

alsora: > unless it automatically casts from the standard node types

Yes, it casts automatically from node types (see example in the unit-tests https://github.com/ros2/rclcpp/blob/rolling/rclcpp/test/rclcpp/node_interfaces/test_node_interfaces.cpp#L59-L67)

> I'd update that one API to use parameterS

IMO that's not the correct approach as nothing requires that APIs in the node and in the underlying interface match.

 ----- 

This definitely evolved into a much larger discussion than what the PR was trying to address.
I would be ok to go with what is IMO a short-term approach (either using the parameter**S** API or using templates) only if we (ROS maintainers) prioritize fixing any lingering issue with the `NodeInterfaces` class.

Having APIs that take nodes as arguments is something that makes it harder to adhere to the best-practices to optimize perfomrance and have really modular ROS 2 applications (examples are passing nodes around or creating nodes within nodes).

If we want ROS to succeed in a variety of use-cases, we must make sure that it's both easy to use for beginners as well as possible to tune for advanced users.
As it stands today, there's still lot of work needed in both directions.

SteveMacenski: > and in the underlying interface match.

I think the entire system would break if we didn't assume that the ROS node abstract interfaces were implemented. The `NodeInterfaces` method assumes the exact same things as I am as I pointed out above. The `NodeInterfaces` direction equally relies on the parameters interface existing and gettable via `get_node_XYZ_interface()`. They're functionally identical but without the seemingly nutty levels of indirection. If we look at it from a requirements direction they both assume the following, no more, no less (if I use the `node_parameters->set_parameters` API:
- That there is some method called `get_node_parameter_interfaces()` that returns a ROS parameter interface
- That the interface contains the same get / set implementations (with **s**) 

If we can't agree on that, I don't see how we can assume anything about a node to accomplish anything in rclcpp. That is already the lowest common denominator. I'm not sure we can get any lower.

SteveMacenski: To be clear, I'm suggesting:
- I'll get the param interface via `get_node_parameter_interface()`  for each node
- Use the methods that the node parameters interface provides to me **only**, not `node->api()`, but `node->get_node_parameter_interface()->api()` 

Is that OK? I don't see how that's any different from the `NodeInterfaces` direction other than creating obscurity

SteveMacenski:
@wjwwood @alsora this should meet your requests :smile:

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,57 @@
...
+void
+copy_all_parameters(
+  const NodeT1 & source, const NodeT2 & destination, const bool override_existing_params = false)
+{
+  using Parameters = std::vector<rclcpp::Parameter>;
+  auto source_params = source->get_node_parameters_interface();
+  auto dest_params = destination->get_node_parameters_interface();
+
+  std::vector<std::string> param_names = source_params->list_parameters({}, 0).names;
+  Parameters params = source_params->get_parameters(param_names);
```
fujitatomoya, line 45: if there is racy condition during this process, possibly it generates the exceptions `rclcpp::exceptions::ParameterUninitializedException` and `rclcpp::exceptions::ParameterNotDeclaredException`. which i guess it is expected behavior? at least, we would want to add this comments in doc section?

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,57 @@
...
+  for (Parameters::const_iterator iter = params.begin(); iter != params.end(); ++iter) {
+    if (!dest_params->has_parameter(iter->get_name())) {
+      dest_params->declare_parameter(iter->get_name(), iter->get_parameter_value());
+    } else if (override_existing_params) {
+      dest_params->set_parameters_atomically({*iter});
```
fujitatomoya, line 50: i think this would fail in some cases,

- if the parameter is `read_only`.
- if the parameter value type is different.
- if the value being set does not complys with the descriptor.
- user `on_set_parameters_callbacks` rejects to set the parameter.

after all, with this function, we are not sure what parameters are successfully set to the `destination`. i think that could be complication for user application.

SteveMacenski: I would like to simply remove the `override_existing_params` option then and have it be documented that existing parameters remain untouched. Is that reasonable?

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: That seems reasonable, I can do that :+1:

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,57 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__COPY_ALL_PARAMETERS_HPP_
+#define RCLCPP__COPY_ALL_PARAMETERS_HPP_
+
+#include <vector>
+#include <string>
```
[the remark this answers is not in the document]
SteveMacenski: I haven't heard of that requirement, but I can do that :+1:

rclcpp/include/rclcpp/copy_all_parameters.hpp
fujitatomoya: i think that your use case is to copy all parameters from `Node` with parameters to `Node` w/o any parameters? in that case, `copy_parameter_declarations` will work. but any other cases, user cannot know how many, what parameter declarations are copied via this method, i think we can describe the current behavior in the doc section.

probably we can consider `override` option for future enhancements.

CC: @wjwwood what do you think?

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,57 @@
...
+{
+  using Parameters = std::vector<rclcpp::Parameter>;
+  auto source_params = source->get_node_parameters_interface();
+  auto dest_params = destination->get_node_parameters_interface();
+
+  std::vector<std::string> param_names = source_params->list_parameters({}, 0).names;
+  Parameters params = source_params->get_parameters(param_names);
+  for (Parameters::const_iterator iter = params.begin(); iter != params.end(); ++iter) {
+    if (!dest_params->has_parameter(iter->get_name())) {
+      dest_params->declare_parameter(iter->get_name(), iter->get_parameter_value());
```
fujitatomoya, line 48: I think we need to copy the descriptions as well with using `describe_parameters`. instead, the parameter behavior in the destination would be different from original one. (test would be ideal too.)

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: How about `copy_unique_parameters()`? That's a little more niche and implies that it won't copy over shared parameters. If i were in the reviewer's chair, that would meet the _intent_ of my comment but create a specialized case so specialized I'd question the value to `rclcpp` 

I re-read your comment and wondering if you're basically just asking me to error check? 

- If read only --> will that throw? If so, I can catch
- Different values will definitely throw, I can catch
- I'm not really sure what the actionable item is on "if the value being set does not comply with the descriptor". If the parameter is of the same name and type, is that really a realistic concern? If so, how could that be done without machine learning to read the two descriptions to tell if they're meaning the same thing :laughing: 
- Rejection: if the set params API returns to me a status, I can check that

If my reading of this is correct and its just error checking, how yeah, I can do that. I misread it the first go around

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: Can do :+1:

rclcpp/include/rclcpp/copy_all_parameters.hpp
fujitatomoya: > I re-read your comment and wondering if you're basically just asking me to error check?

basically yes. these cases happen if it overwrites the parameters via `set_parameter`.
besides, my major concern is user cannot know what parameters are copied successfully or failed, it just works best-effort, and then user needs to check what parameter copied to the other side after all. i am not sure if the application actually rely on this method.

> If the parameter is of the same name and type, is that really a realistic concern?

if the destination parameter is static typed parameter in the description, it cannot copy(set) the different typed value to it. on the other hand, if that is dynamic typed parameter, it can copy and store the different typed value.

https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L263-L306

SteveMacenski:
@fujitatomoya try this on for size

alsora (approved):

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,82 @@
...
+  rclcpp::Logger logger = destination->get_node_logging_interface()->get_logger();
+
+  std::vector<std::string> param_names = source_params->list_parameters({}, 0).names;
+  Parameters params = source_params->get_parameters(param_names);
+  Descriptions descriptions = source_params->describe_parameters(param_names);
+
+  for (unsigned int idx = 0; idx != params.size(); idx++) {
+    if (!dest_params->has_parameter(params[idx].get_name())) {
+      dest_params->declare_parameter(
+        params[idx].get_name(), params[idx].get_parameter_value(), descriptions[idx]);
```
fujitatomoya, line 58: if source node has done,

declared `dynamic_typing` parameter with (`dynamic_typing` requires default value but not static type)

https://github.com/ros2/rclcpp/blob/9284d7cefa9e37ce3b6926a44e6177d6e4a4b62a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp#L81-L93

and set parameter into some value, then this will generate the exception

https://github.com/ros2/rclcpp/blob/9284d7cefa9e37ce3b6926a44e6177d6e4a4b62a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L614-L618

because it tries to set the static type for `dynamic_typing` enabled descriptor?

rclcpp/include/rclcpp/copy_all_parameters.hpp
```diff
@@ -0,0 +1,82 @@
...
+    } else if (override_existing_params) {
+      try {
+        rcl_interfaces::msg::SetParametersResult result =
+          dest_params->set_parameters_atomically({params[idx]});
```
fujitatomoya, line 62: It only tries to store the parameter value, but descriptor. (e.g `undeclare_parameter`, then call `declare_parameter`)
I think that is okay but probably we would want to add this behavior in doc section as well to explicitly explain how this works?

(comment to share) at this moment, there wont be undeclared parameters, so we do not expect `ParameterNotDeclaredException`, unless there is racy condition.

fujitatomoya (commented):
@SteveMacenski thanks for iterating.

overall looks good to me, but a couple of minor comments. could you check?

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: There is no set parameter API for adding a descriptor. I think undeclaring and redeclaring is a bridge too far, because you're then wiping away the option for there to be type mismatches because you've set them totlaly differently.

In my mind, `copy_` in the function names means copy, but change the type and other features of existing ones where as its not a value copy. I suppose it could be renamed to `copy_all_parameter_values()` instead?

rclcpp/include/rclcpp/copy_all_parameters.hpp
SteveMacenski: @fujitatomoya that method is never called anywhere in the string of actions that this PR currently uses, its impossible for that to be a concern. We do not declare and nothing we do depends on that function that could throw that exception :-)

fujitatomoya: @SteveMacenski ah sorry, my bad. i was the one confused :sweat_smile: always calling `declare_parameter` with just a default_value.

rclcpp/include/rclcpp/copy_all_parameters.hpp
fujitatomoya: agree, i think that is overkill too.

> I suppose it could be renamed to copy_all_parameter_values() instead?

sounds good to me!

appreciate for the quick response!

SteveMacenski:
@fujitatomoya if this isn't done, I'll have to drive down to San Jose so we can pair program for an hour until its done :stuck_out_tongue:

fujitatomoya (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19571/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20295/)

SteveMacenski:
@fujitatomoya this good to go? The test failures here are not related to this PR

```
23:37:10 The following tests FAILED:
23:37:10 	 22 - test_client (Failed)
23:37:10 	 61 - test_publisher (Failed)
23:37:10 	 78 - test_service (Failed)
23:37:10 	 80 - test_subscription (Failed)
23:37:10 	116 - test_rosout_subscription (Failed)
```

clalancette:
> @fujitatomoya this good to go? The test failures here are not related to this PR

I'll point out that we have no failures on the nightlies, so either they come from this PR, or this PR needs to be rebased.  Either way they need to be solved, and CI re-run, before we can merge.

SteveMacenski:
@clalancette rebased

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19609/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14117/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20333/)

SteveMacenski:
Aarch failure seems unrelated:

```
18:18:26 Caused: hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@21d03d09:JNLP4-connect connection from ip-10-0-2-42.ec2.internal/10.0.2.42:48068": Remote call on JNLP4-connect connection from ip-10-0-2-42.ec2.internal/10.0.2.42:48068 failed. The channel is closing down or has closed down
```

fujitatomoya:
@SteveMacenski i see many unrelated changes in this PR after rebase. is this good to go? CI is green though.

SteveMacenski:
The rebase did something odd - those are just the changed pulled in. I will not be using the git global config rebase-pull mechanism again and just rebase normally. Please squash merge

PR #2308, opened 2023-09-12
Restoring components architecture document

mjcarroll:
Noticed this got left behind in a branch.

For context, this was originally written in 2019 and didn't get reviewed at the time.  I expect that there will also be updates required to bring the document up to date with the current implementation.

CC: @wjwwood

### DIFF
--- a/rclcpp_components/doc/architecture.rst
+++ b/rclcpp_components/doc/architecture.rst
@@ -1,0 +1,263 @@
+Architecture of ``rclcpp_components``
+=====================================
+
+``rclcpp_components`` provides tools and defines conventions for building ``rclcpp`` nodes into dynamically loadable components.
+Building composable nodes makes it easier to compose them with other composable nodes at runtime.
+For example, it allows you to build your nodes one time but then at runtime decide whether or not to run them in their own process or combine them into a single process.
+
+It also provides some generic tools for working with composable nodes, including a composable node container, which can load and execute multiple nodes in a variety of scenarios.
+
+Terminology
+-----------
+
+- Composable Node:
+  - An ``rclcpp`` based node that is setup so that it can be composed into a single process with other composable nodes at either runtime or programming time.
+- Node Component:
+  - Same thing as Composable Node, can be used interchangeably.
+- Composable Node Container:
+  - A program that can load and host composable nodes, it is often general purpose and has not a priori knowledge about the nodes it is instantiating and executing.
+- pluginlib:
+  - ROS package that provides mechanisms to create and register plugins based on user defined interface classes, and then dynamically lookup and load plugins at runtime.
+- Stand-alone Node Executable:
+  - Convenience executable that can be used to run a composable node as a single process or have it loaded dynamically in a remote composable node container.
+
+Theory of Operation
+-------------------
+
+The lowest level series of steps require to make a composable node are as follows:
+
+- Create a shared library in which one or more of your composable nodes are compiled.
+  - The nodes should also be registered as a ``pluginlib``-style plugin.
+  - The library should be created with CMake's ``add_library``
+  - The library should either explicitly be set to ``SHARED``.
+- Create or add to a ``plugindescription.xml`` file and register it so that ``pluginlib`` can find the node.
+- Create a stand-alone executable for each node.
+  - Which can either be used to run the node in its own process or to load it into an existing process and serving as a proxy while it is in the remote process.
+
+There are both C++ preprocessor and CMake macros within this package to assist with each of the above steps.
+Generally they can be applied in the following way:
+
+- Write files containing nodes, e.g. ``src/a.cpp``, ``src/b.cpp``, etc., and in each use the ``RCLCPP_COMPONENTS_REGISTER_NODE()`` preprocessor macro in the source code.
+- In CMake, create the shared library, e.g. ``add_library(my_library SHARED src/a.cpp src/b.cpp)``.
+  - Do any extra CMake steps for the library, e.g. ``target_link_libraries`` or ``target_include_directories``.
+- Describe each node in the library with the CMake using ``rclcpp_components_describe_node_component``.
+  - e.g. ``rclcpp_components_describe_node_component(my_library NAME MyNodeA TYPE my_cpp_namespace::MyNodeA DESCRIPTION "...")``
+- Register the library with ``pluginlib`` by calling ``rclcpp_components_register_node_component_library``.
+  - e.g. ``rclcpp_components_register_node_component_library(my_library GENERATE_PLUGIN_DESCRIPTION)``
+- Optionally, create a stand-alone executable for each node using ``rclcpp_components_create_executable_for_node``.
+  - e.g. ``rclcpp_components_create_executable_for_node(my_node_a_target NODE_COMPONENT_NAME MyNodeA)``
+  - An assumption is that package name is equal to ``PROJECT_NAME``, but can be overridden with the ``PACKAGE_NAME <package name>`` option.
+
+Using the ``GENERATE_PLUGIN_DESCRIPTION`` option with the ``rclcpp_components_register_node_component_library()`` macro will cause this function to make a plugin description XML file for you using the information from the ``rclcpp_components_describe_node_component`` calls.
+However, you can avoid this macro and the ``rclcpp_components_describe_node_component()`` macro calls by creating your own plugin description XML file and passing it with the ``PLUGIN_DESCRIPTION_FILE <file path>`` option.
+
+So, the required steps are registering your nodes in the source code, creating a shared library, and registering with ``pluginlib``.
+However, you have some choice in how the ``pluginlib`` description file is created and how registration occurs, plus creating the node executables is completely optional.
+
+There are more options available to you, and if you do not wish to use the convenience macros in CMake you can see in the details that follow how each are implemented and do any of the steps manually yourself.
+
+Writing the Code for a Node Component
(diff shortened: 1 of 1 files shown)


rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+
+Terminology
+-----------
+
+- Composable Node:
+  - An ``rclcpp`` based node that is setup so that it can be composed into a single process with other composable nodes at either runtime or programming time.
+- Node Component:
+  - Same thing as Composable Node, can be used interchangeably.
+- Composable Node Container:
+  - A program that can load and host composable nodes, it is often general purpose and has not a priori knowledge about the nodes it is instantiating and executing.
```
clalancette, line 18:
```suggestion
  - A program that can load and host composable nodes, which is often general purpose and has no a priori knowledge about the nodes it is instantiating and executing.
```

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+- pluginlib:
+  - ROS package that provides mechanisms to create and register plugins based on user defined interface classes, and then dynamically lookup and load plugins at runtime.
+- Stand-alone Node Executable:
+  - Convenience executable that can be used to run a composable node as a single process or have it loaded dynamically in a remote composable node container.
+
+Theory of Operation
+-------------------
+
+The lowest level series of steps require to make a composable node are as follows:
```
clalancette, line 27:
```suggestion
The series of steps require to make a composable node are as follows:
```

(at least, I'm not sure what "lowest level" is trying to convey here)

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+- Composable Node:
+  - An ``rclcpp`` based node that is setup so that it can be composed into a single process with other composable nodes at either runtime or programming time.
+- Node Component:
+  - Same thing as Composable Node, can be used interchangeably.
+- Composable Node Container:
+  - A program that can load and host composable nodes, it is often general purpose and has not a priori knowledge about the nodes it is instantiating and executing.
+- pluginlib:
+  - ROS package that provides mechanisms to create and register plugins based on user defined interface classes, and then dynamically lookup and load plugins at runtime.
+- Stand-alone Node Executable:
+  - Convenience executable that can be used to run a composable node as a single process or have it loaded dynamically in a remote composable node container.
```
clalancette, line 22: I don't really understand the latter part of this sentence (I'm not sure what a "remote composable node container" is).  Could you maybe explain a little further or reword it?

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+
+Theory of Operation
+-------------------
+
+The lowest level series of steps require to make a composable node are as follows:
+
+- Create a shared library in which one or more of your composable nodes are compiled.
+  - The nodes should also be registered as a ``pluginlib``-style plugin.
+  - The library should be created with CMake's ``add_library``
+  - The library should either explicitly be set to ``SHARED``.
```
clalancette, line 32:
```suggestion
  - The library should either explicitly be set to ``SHARED``, or be using ``ament_cmake_ros``.
```

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+- Create or add to a ``plugindescription.xml`` file and register it so that ``pluginlib`` can find the node.
+- Create a stand-alone executable for each node.
+  - Which can either be used to run the node in its own process or to load it into an existing process and serving as a proxy while it is in the remote process.
```
clalancette, line 35: Is the latter part of this sentence true?  Do we keep around a proxy process for composable nodes?

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+- Write files containing nodes, e.g. ``src/a.cpp``, ``src/b.cpp``, etc., and in each use the ``RCLCPP_COMPONENTS_REGISTER_NODE()`` preprocessor macro in the source code.
+- In CMake, create the shared library, e.g. ``add_library(my_library SHARED src/a.cpp src/b.cpp)``.
+  - Do any extra CMake steps for the library, e.g. ``target_link_libraries`` or ``target_include_directories``.
+- Describe each node in the library with the CMake using ``rclcpp_components_describe_node_component``.
+  - e.g. ``rclcpp_components_describe_node_component(my_library NAME MyNodeA TYPE my_cpp_namespace::MyNodeA DESCRIPTION "...")``
+- Register the library with ``pluginlib`` by calling ``rclcpp_components_register_node_component_library``.
+  - e.g. ``rclcpp_components_register_node_component_library(my_library GENERATE_PLUGIN_DESCRIPTION)``
+- Optionally, create a stand-alone executable for each node using ``rclcpp_components_create_executable_for_node``.
+  - e.g. ``rclcpp_components_create_executable_for_node(my_node_a_target NODE_COMPONENT_NAME MyNodeA)``
+  - An assumption is that package name is equal to ``PROJECT_NAME``, but can be overridden with the ``PACKAGE_NAME <package name>`` option.
```
clalancette, line 49: I think this set of steps is from an older version.  My understanding is that nowadays, we only need to call `rclcpp_components_register_nodes`, though I could be wrong.

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+
+    #define RCLCPP_COMPONENTS_REGISTER_NODE(NodeClass) \
+      CLASS_LOADER_REGISTER_CLASS(NodeFactoryTemplate<NodeClass>, rclcpp_components::NodeFactory)
+
+Where the base class ``rclcpp_components::NodeFactory`` is the actual plugin interface and the ``rclcpp_components::NodeFactoryTemplate`` class is a templated "adapter" class which will provide the ``rclcpp_components::NodeFactory`` interface given a sub-class of ``rclcpp::Node``.
+
+The NodeFactory and Related Interfaces
+++++++++++++++++++++++++++++++++++++++
+
+The ``rclcpp_components::NodeFactory`` interface is dual-purposed, first it works around a limitation of instantiating classes via a general purpose class loading interface which is that you cannot call constructors on derived classes because you only know the base class and are unaware of any different signatures for constructors in the derived class.
```
clalancette, line 100:
```suggestion
The ``rclcpp_components::NodeFactory`` interface is dual-purposed.
First, it works around a limitation of instantiating classes via a general purpose class loading interface which is that you cannot call constructors on derived classes because you only know the base class and are unaware of any different signatures for constructors in the derived class.
```

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+The ability to get access to the original Node instance shared pointer is just for debugging and special cases.
+Most of the time the program loading the instance of the node will not be aware of the original type of the node instance, whether it be a ``rclcpp::Node`` or something like ``my_ns::MyNode`` which inherits from ``rclcpp::Node`` or even something else completely.
+
+Finally, the ``rclcpp_components::NodeFactoryTemplate<NodeT>`` convenience class will create an implementation of ``NodeFactory`` based on the ``NodeT`` type.
+This template class will work so long as the class given for ``NodeT`` provides the ``get_node_base_interface()`` method that exists on the ``rclcpp::Node`` class.
+
+Building the Shared Library which Contains Composable Nodes
+-----------------------------------------------------------
+
+Typically this is done by just calling the standard ``add_library`` macro in CMake, but any thing that results in generating a CMake target which represents a shared library is sufficient.
```
clalancette, line 166:
```suggestion
Typically this is done by just calling the standard ``add_library`` macro in CMake, but anything that results in generating a CMake target which represents a shared library is sufficient.
```

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+
+Once you've created the target you may manipulate the target with standard CMake macros like ``target_include_directories`` and ``target_link_libraries``.
+
+Creating the Plugin Description File
```
clalancette, line 170: Just taking https://github.com/ros2/demos/tree/rolling/composition as an example, I don't see any pluginlib XML description file, either in the sources nor in the `install` directory.  But maybe I'm missing something.

rclcpp_components/doc/architecture.rst
```diff
@@ -0,0 +1,263 @@
...
+- node namespace
+  - overrides the node namespace hard coded into the source code (i.e. like ``__ns:=<new ns>`` on the command line)
+- remappings
+  - remappings that apply to this node for things like topics and services
+- initial parameter values
+  - initial values for required parameters or to be used instead of the defaults for optional parameters
+
+Essentially, where to find the node (package name plus node name) and anything that can be influenced with command line arguments to nodes normally.
+
+Compose Nodes with a Configuration File
```
clalancette, line 244: Is this currently implemented in the container?  It doesn't seem like it to me, but I could be wrong.

clalancette (changes requested):
I've left a bunch of feedback and questions.

rclcpp_components/doc/architecture.rst
mjcarroll: We do not, removing.

mjcarroll: Actually, many things in this section aren't relevant, because we aren't using `pluginlib` registration like this anymore.  Let me rework it.

PR #2312, opened 2023-09-15
Fix exception message for uninitialized parameter

luca-della-vedova:
A small nit but to my understanding declaring a statically typed parameter with an uninitialized value is allowed, what is not allowed is having an uninitialized type, as the exception `InvalidParameterTypeException` also explains.
Updated the exception message to reflect this.

### DIFF
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -528,7 +528,7 @@ declare_parameter_helper(
     if (rclcpp::PARAMETER_NOT_SET == type) {
       throw rclcpp::exceptions::InvalidParameterTypeException{
               name,
-              "cannot declare a statically typed parameter with an uninitialized value"
+              "cannot declare a statically typed parameter with an uninitialized type"
       };
     }
     parameter_descriptor.type = static_cast<uint8_t>(type);


fujitatomoya (approved):
lgtm

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19524/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14042/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20243/)

iuhilnehc-ynos:
It seems not good to me without a test.

BTW: 
https://github.com/ros2/rclcpp/blob/a7a9b78feebff603e5c7c5506ce31b89916bde36/rclcpp/test/rclcpp/test_node.cpp#L3436-L3439
It's the case to test the exception. I think `parameter with an uninitialized value` is good.

fujitatomoya:
> It's the case to test the exception. I think parameter with an uninitialized value is good.

👍

luca-della-vedova:
I'm not sure I follow, is the request to add a new test with a parameter named `parameter_with_uninitialized_value` or to change the name of the existing parameter in the test?

iuhilnehc-ynos:
Sorry to make you confused. I mean the original exception message seems good to me.
I am not sure why we need this PR.

fujitatomoya:
@luca-della-vedova let me add more explanation here.

i jumped on the fix and though this change aligns with the `InvalidParameterTypeException`. but as @iuhilnehc-ynos says, it actually checks if that is the `static` typing parameter, it can deduce the type from default value, and if default value is not provided, `InvalidParameterTypeException` will be generated here.

So after all, i believe `uninitialized value` is the correct message for user application, because it does not provide the default value to the static typed parameter. sorry i was mistaken too.

@iuhilnehc-ynos my explanation above makes sense to you? and thanks for pointing that out.

luca-della-vedova:
I agree that usually the type is inferred from the default value, however there is an API for `declare_parameter` that takes as an input the parameter type and leaves the value uninitialized [here](https://github.com/ros2/rclcpp/blob/c42745c5baea249ec6159f1100a9a0d745510872/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L599-L604).
From what I understand if a user was to use it, the exception wouldn't be thrown, but technically in that case they wouldn't have initialized the value, just the type.
For this reason I thought that the exception message should warn users if they haven't set a type.

PR #1253, opened 2020-07-31
Increase timeouts for connext for long tests

brawner:
Destroying nodes takes upwards of 3 seconds, which can cause many unit tests to take inordinate amount of time (#1250) with rmw_connext_cpp. Currently only `test_executors` and `test_allocator_memory_strategy` timeout, so the timeout is being increased here.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

Issue #1250, opened 2020-07-31
[rmw_connext_cpp] Destructing node, specifically a default waitable, is slow (> 3s) for rmw_connext_cpp


brawner:
The rclcpp unit tests take significantly longer for rmw_connext_cpp than other middleware implementations. 

Compare ConnextDDS unit tests with rclcpp (11 minutes)
http://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/32/testReport/rclcpp/

vs FastRTPS (1 min 30 s):
http://build.ros2.org/view/Rci/job/Rci__nightly-fastrtps_ubuntu_focal_amd64/34/testReport/rclcpp/

I've narrowed this down to the destruction of the rclcpp::Node, a call to destruct a waitable takes the vast majority of this time. I believe this is happening in AllocatorMemoryStrategy, but since it uses shared_ptrs, the actual destruct may be elsewhere.
If I put steady_clock measurements around [this line](https://github.com/ros2/rclcpp/blob/3b8c334d5dd5fddaf3fec03b34e6767a2fde3a78/rclcpp/include/rclcpp/strategies/allocator_memory_strategy.hpp#L90), or even if I iterate through the vector and call reset explicitly, there is one waitable that will consume significant amounts of time to delete. This isn't for all waitables, just one default one that I can't figure out where it's coming from.

clalancette:
This is essentially a duplicate of https://github.com/ros2/rmw_connext/issues/325 .  I'm going to close this out as a duplicate of that one, copying in your investigation here so it isn't lost.

PR #2313, opened 2023-09-20
Removing Old Connext Tests

CursedRock17:
So this is a PR in response to some TODOs:
> TODO(brawner) remove when destroying Node for Connext is resolved

This pull requests just removes these older, stale tests which don't have a purpose. The discussion for this takes place in issue #1250, the pull request that these special tests were added was PR #1253. Then, it seems that [this pull request](https://github.com/ros2/rmw_connext/pull/473) in the old rmw_connext repository was responsible for fixing why these tests were added. 

I figured that since that pull request solved the issue, the TODO requirement was met and there is no reason to have special cases on the tests. I assume we should still keep the tests because they're generalized and still provide value.

### DIFF
--- a/rclcpp/test/rclcpp/CMakeLists.txt
+++ b/rclcpp/test/rclcpp/CMakeLists.txt
@@ -34,13 +34,7 @@ if(TARGET test_exceptions)
   target_link_libraries(test_exceptions ${PROJECT_NAME} mimick)
 endif()
 
-# Increasing timeout because connext can take a long time to destroy nodes
-# TODO(brawner) remove when destroying Node for Connext is resolved. See:
-# https://github.com/ros2/rclcpp/issues/1250
-ament_add_gtest(
-  test_allocator_memory_strategy
-  strategies/test_allocator_memory_strategy.cpp
-  TIMEOUT 360)
+ament_add_gtest(test_allocator_memory_strategy strategies/test_allocator_memory_strategy.cpp)
 if(TARGET test_allocator_memory_strategy)
   ament_target_dependencies(test_allocator_memory_strategy
     "rcl"
@@ -669,13 +663,10 @@ if(TARGET test_interface_traits)
   target_link_libraries(test_interface_traits ${PROJECT_NAME})
 endif()
 
-# TODO(brawner) remove when destroying Node for Connext is resolved. See:
-# https://github.com/ros2/rclcpp/issues/1250
 ament_add_gtest(
   test_executors
   executors/test_executors.cpp
-  APPEND_LIBRARY_DIRS "${append_library_dirs}"
-  TIMEOUT 180)
+  APPEND_LIBRARY_DIRS "${append_library_dirs}")
 if(TARGET test_executors)
   ament_target_dependencies(test_executors
     "rcl"
--- a/rclcpp/test/rclcpp/CMakeLists.txt
+++ b/rclcpp/test/rclcpp/CMakeLists.txt
@@ -666,7 +666,8 @@ endif()
 ament_add_gtest(
   test_executors
   executors/test_executors.cpp
-  APPEND_LIBRARY_DIRS "${append_library_dirs}")
+  APPEND_LIBRARY_DIRS "${append_library_dirs}"
+  TIMEOUT 180)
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Yeah, this seems reasonable.  I'm going to run a few different kinds of CI on this to check.

However, I'll note that if we start to see these tests become flaky, we'll probably go ahead and revert it.

alsora (approved):

rclcpp/test/rclcpp/CMakeLists.txt
```diff
@@ -669,13 +663,10 @@ if(TARGET test_interface_traits)
   target_link_libraries(test_interface_traits ${PROJECT_NAME})
 endif()
 
-# TODO(brawner) remove when destroying Node for Connext is resolved. See:
-# https://github.com/ros2/rclcpp/issues/1250
 ament_add_gtest(
   test_executors
   executors/test_executors.cpp
-  APPEND_LIBRARY_DIRS "${append_library_dirs}"
-  TIMEOUT 180)
```
clalancette, line 678: So it looks like this test on Windows is taking longer than 60 seconds.  So I think we need to put this back to 180 seconds.

clalancette (changes requested):

rclcpp/test/rclcpp/CMakeLists.txt
clalancette: > 1. Is this the case for just this test, or does the other one need to be reverted too?

It looks like just for this `test_executors`, you can check the jobs above.

> 2. Is the problem only on windows?

It looks like it.  It's not really that surprising, Windows is much slower than Linux in all respects.

> 3. Do we need to fix the reasons why this test might be going wrong in relation to Connect?

Actually, this test isn't run on a per-RMW basis, only with the default one.  So the comment is bogus either way.

All of that said, it might be worthwhile taking a look at that particular test and seeing why it is taking over minute to complete.  If we can cut that down to something like 30 seconds, then we can probably lower this (and make CI faster to boot).

fujitatomoya (approved):

rclcpp/test/rclcpp/CMakeLists.txt
CursedRock17: I'm almost positive I found the problem [here](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/test/rclcpp/executors/test_executors.cpp#L637). It seems that the test that's taking forever is `testRaceConditionAddNode` taking about 20000 - 27000 ms, everything else compared to Linux, is just longer on Windows. 

This implementation uses `std::thread::hardware_concurrency`, so based on the number of cores that each system has, [cppreference here](https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency). It's the only place this is used in that test, other than in `multi_threaded_executor`'s test, but that's not an issue. My concern is how this works with the CI, since this is an actual hardware thing, can we find the number of cores we use, or is this just a Windows Software problem?

clalancette (approved):
Seems OK to me.  I'll kick off CI on it again to see what it looks like.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19557/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14071/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20279/)
* Linux repeated [Build Status](http://ci.ros2.org/job/ci_linux/19558/)
* Windows repeated [Build Status](http://ci.ros2.org/job/ci_windows/20280/)

clalancette:
I rebased this onto the latest.  Here's another try at CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19574/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14086/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20298/)
* Linux repeated [Build Status](http://ci.ros2.org/job/ci_linux/19575/)
* Windows repeated [Build Status](http://ci.ros2.org/job/ci_windows/20299/)

PR #2316, opened 2023-09-26
Update SignalHandler get_global_signal_handler to avoid complex types in static memory

tfoote:
This was flagged by msan as a problem. 

There's a description of why this is a potential problem here: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19557/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14070/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20278/)

### DIFF
--- a/rclcpp/src/rclcpp/signal_handler.cpp
+++ b/rclcpp/src/rclcpp/signal_handler.cpp
@@ -113,8 +113,8 @@ SignalHandler::get_logger()
 SignalHandler &
 SignalHandler::get_global_signal_handler()
 {
-  static SignalHandler * signal_handler = new SignalHandler();
-  return *signal_handler;
+  static SignalHandler & signal_handler = *new SignalHandler();
+  return signal_handler;
 }
 
 bool
--- a/rclcpp/src/rclcpp/signal_handler.cpp
+++ b/rclcpp/src/rclcpp/signal_handler.cpp
@@ -113,8 +113,8 @@ SignalHandler::get_logger()
 SignalHandler &
 SignalHandler::get_global_signal_handler()
 {
-  static SignalHandler signal_handler;
-  return signal_handler;
+  static SignalHandler* signal_handler = new SignalHandler();
+  return *signal_handler;
 }
 
 bool
--- a/rclcpp/src/rclcpp/signal_handler.cpp
+++ b/rclcpp/src/rclcpp/signal_handler.cpp
@@ -113,7 +113,7 @@ SignalHandler::get_logger()
 SignalHandler &
 SignalHandler::get_global_signal_handler()
 {
-  static SignalHandler* signal_handler = new SignalHandler();
+  static SignalHandler * signal_handler = new SignalHandler();
   return *signal_handler;
 }
 


rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -113,8 +113,8 @@ SignalHandler::get_logger()
 SignalHandler &
 SignalHandler::get_global_signal_handler()
 {
-  static SignalHandler signal_handler;
-  return signal_handler;
+  static SignalHandler * signal_handler = new SignalHandler();
```
fujitatomoya, line 116: I am not sure if this is the same behavior. this creates the new `SignalHandler` object every time it calls `SignalHandler::get_global_signal_handler()`. originally this was Singleton design, so it expects to return the same object within this class.

probably we need to do,

- having `static SignalHandler * signal_handler` in private member variable with `SignalHandler`.

and,

```suggestion
  if (signal_handler == NULL) {
    signal_handler = new SignalHandler();
    return signal_handler;
  } else {
    return signal_handler;
  }
```

wjwwood: I think the `static ...` statement will only ever be run once, e.g. `static int i = 0;` will only result in `i` being set once.

That being said, I'm not sure this is the right approach. I'll need to understand the root issue better.

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -113,8 +113,8 @@ SignalHandler::get_logger()
...
+  return *signal_handler;
```
wjwwood, line 117: Actually this is likely the right approach, but I'd prefer the use of a reference, e.g.:

```suggestion
  static SignalHandler & signal_handler = *new SignalHandler();
  return signal_handler;
```

rclcpp/src/rclcpp/signal_handler.cpp
tfoote: Yeah, that was my first worry too. However, I double checked this with a simple /.a.out program. It seemed safe from multiple allocations.

rclcpp/src/rclcpp/signal_handler.cpp
tfoote: Yeah, I saw versions like this too in my research of best practices.

tfoote:
Not sure why the PR job isn't getting triggered. I'll try close and reopen.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19563/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14076/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20287/)

rclcpp/src/rclcpp/signal_handler.cpp
fujitatomoya: > I think the static ... statement will only ever be run once

I see, and that makes sense. thank you very much for the explanation.

fujitatomoya (approved):
lgtm with green CI

rclcpp/src/rclcpp/signal_handler.cpp
```diff
@@ -113,7 +113,7 @@ SignalHandler::get_logger()
 SignalHandler &
 SignalHandler::get_global_signal_handler()
 {
-  static SignalHandler signal_handler;
+  static SignalHandler & signal_handler = *new SignalHandler();
```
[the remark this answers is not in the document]
tfoote: I believe that memory debuggers are aware of this pattern. They actually won't report this as a memory leak because there's always a valid reference to it during execution until final termination due to the statically allocated reference. As such for all times during the program's execution this memory is reachable.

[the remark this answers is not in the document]
wjwwood: There's no good time to delete it, and though I don't love it either, it is the recommended solution in the google c++ style guide. 🤷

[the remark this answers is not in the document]
wjwwood: The alternative would be to ensure this class is trivially destructible which isn't possible given that it needs to store things like the `rclcpp::Logger` which itself stores a string, a `std::thread`, etc... So I think this is the best solution unfortunately.

I will point out it is a recommended style, not a correctness thing. What's there today is valid and (as far as we can see) correct code, but the issue is if a bug works its way in, this pattern (non-trivial static objects) makes diagnosing harder, hence the recommendation. So I think it's a valid thing to change.

tfoote:
Thanks for the thoughts. I will push this in as it's the best solution we're aware of.

tfoote:
@mergifyio backport humble iron

Issue #2314, opened 2023-09-21
Creating a Subscriber with TopicStatistics Applies Custom QoS to /statistics Publisher


JWhitleyWork:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries
- Version or commit hash:
  - Humble, Iron
- DDS implementation:
  - Any
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
```
rclcpp::SubscriptionOptions opts{};
opts.topic_stats_options.state = rclcpp::TopicStatisticsState::Enable;

rclcpp::QoS bad_qos{};

this->creare_subscription<DummyMsg>("/some_topic", bad_qos.history(rclcpp::HistoryPolicy::Unknown), cb_func, opts);
```

#### Expected behavior
The `/statistics` publisher has a stable, well-defined QoS profile.

#### Actual behavior
The QoS profile passed to `create_subscription` is applied to both the subscription and the `/statistics` publisher. Code reference in [humble](https://github.com/ros2/rclcpp/blob/humble/rclcpp/include/rclcpp/create_subscription.hpp#L92) and [iron](https://github.com/ros2/rclcpp/blob/iron/rclcpp/include/rclcpp/create_subscription.hpp#L92).

fujitatomoya:
@JWhitleyWork that makes sense to me. adding `QoS` to `TopicStatisticsOptions` struct as `statistics` publisher QoS configuration.

CursedRock17:
Since that PR was merged this issue should be complete right?

clalancette:
Yeah, I'm going to close it.  Thanks!

PR #2323, opened 2023-10-02
Adding QoS to subscription options

CursedRock17:
This is meant to address issue #2314 which adds a custom QoS to the `TopicStatisticsOptions` in `SubscriptionOptionsBase`. This will allow the `statistics` to always have a stable `QoS` even if the subscription may not.


rclcpp/include/rclcpp/subscription_options.hpp
```diff
@@ -77,6 +77,10 @@ struct SubscriptionOptionsBase
     // Topic statistics publication period in ms. Defaults to one second.
     // Only values greater than zero are allowed.
     std::chrono::milliseconds publish_period{std::chrono::seconds(1)};
+
+    // An optional QoS which can provide topic_statistics with a stable QoS separate from
+    // the subscription's current QoS settings which could be unstable.
+    rclcpp::QoS statistics_qos_configuration{10};
```
fujitatomoya, line 83: a couple of things need to be added with this.

- use this QoS setting to statistics publisher.

https://github.com/ros2/rclcpp/blob/7c1143dc1502d5dda99a3dfa17124a72dbaea90c/rclcpp/include/rclcpp/create_subscription.hpp#L92

- How about using `SystemDefaultsQoS` (or some other QoS profile) instead of using hard coded number?

https://github.com/ros2/rclcpp/blob/7c1143dc1502d5dda99a3dfa17124a72dbaea90c/rclcpp/include/rclcpp/qos.hpp#L482-L502

fujitatomoya (commented):
@CursedRock17 thanks for taking care of this 👍  i do have a few comments.

fujitatomoya (commented):
@CursedRock17 

i think test would be ideal, can we add the test for,

https://github.com/ros2/rclcpp/blob/7c1143dc1502d5dda99a3dfa17124a72dbaea90c/rclcpp/test/rclcpp/test_subscription_options.cpp#L57

to check system default QoS is set to the `options.topic_stats_options.statistics_qos_configuration`?

CursedRock17:
That sounds correct to do, I'll go ahead and implement it. There shouldn't be anywhere else where QoS should be added.

CursedRock17:
I added the tests, but I have two concerns:
 - Should the name `statistics_qos_configuration` be shortened to just `qos_configuration`, Since you already have to reference the `StatisticsTopicCollector`.
 - Should we use a custom/different QoS compared to `BestAvailableQoS`. I just assumed that the other test subjects in their, were quite random, so that it wouldn't matter too much.

fujitatomoya (approved):

fujitatomoya:
> Should the name statistics_qos_configuration be shortened to just qos_configuration, Since you already have to reference the StatisticsTopicCollector

i am okay with either. `statistics_qos_configuration` is bit redundant, so it could be just `qos` or `statistics_qos`?

> Should we use a custom/different QoS compared to BestAvailableQoS. I just assumed that the other test subjects in their, were quite random, so that it wouldn't matter too much.

i do not think we need it. that would be the test cases for `rclcpp::QoS` class.

fujitatomoya:
@iuhilnehc-ynos requesting 2nd review on this.

iuhilnehc-ynos (approved):

fujitatomoya:
@iuhilnehc-ynos thanks for the review 👍 (i though it would next week, since you are on vacation 😅 thank you very much your contribution.) I will start CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19602/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20325/)

CursedRock17:
Is there a reason that the Windows CI was aborted, I assume it didn't fail entirely if both Linux versions succeeded. I would assume that all of the tests failing due to timing out wasn't part of the run.

clalancette:
> Is there a reason that the Windows CI was aborted, I assume it didn't fail entirely if both Linux versions succeeded. I would assume that all of the tests failing due to timing out wasn't part of the run.

It aborted because it ran for 15 hours and didn't show signs of finishing.  However, this isn't the first time we've seen this, so it may not be because of this PR.  I'll run it again to see what happens.

CursedRock17:
With a green CI and completed reviews, is this good to go?

ros-discourse:

https://discourse.ros.org/t/ros-2-jazzy-jalisco-released/37862/1

Issue #2307, opened 2023-09-12
ROSRate to be useless without chain to clock pointer


AlexeyMerzlyakov:
In #2123 there was added new interface: optional pointer to the `rclcpp::Clock` in `Rate` class, which could make it to operate with any type of clock; and derivative `WallRate` and `ROSRate` classes operating with clocks of `RCL_STEADY_TIME` and `RCL_ROS_TIME` types respectively.

However, without of having back the pointer to the `clock` created inside `ROSRate` class, the rate became to be useless having no respect to ROS simulated time and publishing to the `/clock` topic. Indeed, the subscription to `/clock` is being implemented in `rclcpp::TimeSource` class that correcting all attached clocks each time when new message from clock topic is arrived. Thus, to make `ROSRate` to operate in the same way as `Rate` with given custom clocks, we need to return back `rclcpp::Clock` pointer from `ROSRate` and then do something with it (e.g. attach the given clock to the `TimeSource` operating with selected node where `use_sim_time` parameter applied).

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - ROS2 rolling built from sources
- Version or commit hash:
  - ea31f94eb453acf3f34b37b85e38284abc514d57
- DDS implementation:
  - Checked on Fast-RTPS (default)
- Client library (if applicable):
  - rclcpp + rcl

#### Steps to reproduce issue
Steps to reproduce:
1. Build attached [test_ws.zip](https://github.com/ros2/rclcpp/files/12586071/test_ws.zip) workspace
2. Setup on local test workspace `$ source test_ws/install/setup.bash`
   and run launch-script, that will bring-up testing node along with Gazebo simulation:
```
$ ros2 launch test_pkg test_node.launch.py
```
Default outputs for 2Hz rate are `0.5` seconds per each duration (measured in real and ROS time):
```
[test_node-3] [INFO] [1694521064.567825994] [test_node]: Custom ROS clock ratetime Real: 0.502854 ROS: 0.502000
```

3. In Gazebo classic simulator open left pane -> Physics -> decrease "real time update rate" from `1000` -> to `500`
4. Observe what's changed

#### Expected behavior
When using `custom_rate` with custom clock:
```cpp
rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
[...]

PR #2326, opened 2023-10-04
Remove useless ROSRate class

AlexeyMerzlyakov:
This is the part of #2307 discussion to completely remove `ROSRate` inherited class, as we can not utilize its `clock` well without the subscription to `/clock` topic (which is being done by `rclcpp::TimeSource` attached to the node)

### DIFF
--- a/rclcpp/include/rclcpp/rate.hpp
+++ b/rclcpp/include/rclcpp/rate.hpp
@@ -184,16 +184,6 @@ public:
   explicit WallRate(const Duration & period);
 };
 
-class ROSRate : public Rate
-{
-public:
-  RCLCPP_PUBLIC
-  explicit ROSRate(const double rate);
-
-  RCLCPP_PUBLIC
-  explicit ROSRate(const Duration & period);
-};
-
 }  // namespace rclcpp
 
 #endif  // RCLCPP__RATE_HPP_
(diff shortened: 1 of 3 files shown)


AlexeyMerzlyakov:
On CI side, `test_executor` was crashed with SIGSEGV code. This might seem to be really strange (flaky situation?), since I've tested RCLCPP both locally and on jammy Docker container (made `rclcpp` workspace there and compiled/tested it over rest rolling binaries). There were no testing failures in both cases.

sloretz:
@mjcarroll FYI from waffle I assigned you for review!

mjcarroll:
@AlexeyMerzlyakov the `test_executor` has some known flakiness in the `Rpr` jobs.  I'm going to run this through CI, and if that comes back green, this should be good to go.

We additionally discussed that dropping this API doesn't require a deprecation cycle as it is extremely unlikely that it is used, given that it was introduced after the `iron` release.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19654/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20378/)

fujitatomoya (approved):
lgtm.

since https://github.com/ros2/rclcpp/pull/2123 merged a month ago, and only available on rolling, no released distribution is affected. we can just go ahead to remove these class.

note: so it always used system time instead of ros time. my bad 😅 sorry i did not see this during review (now, ask myself why ❓ 💦 ), thank you very much for the issue and PR.

clalancette (approved):

Issue #2327, opened 2023-10-05
:man_farmer: `TestRate.wall_rate_basics`, `TestRate.ros_rate_basics` and `TestRate.rate_basics` are flaky in `nightly_win_rep` CI


Blast545:
## Bug report

**Required Info:**

- Operating System:
  - Buildfarm, Windows Repeated jobs
- Installation type:
  - Source
- Version or commit hash:
  - Rolling
- DDS implementation:
  - N/A
- Client library (if applicable):
  - rclcpp

#### Information on the issue
Run a nightly_win_rep job, and these tests will sometime fail with a 44% chance

```
sql_run.sh calculate_flakiness_jobs.sql "rclcpp.TestRate.wall_rate_basics" "40 days"
job_name|last_fail|first_fail|build_count|failure_count|failure_percentage
nightly_win_rep|2023-10-03|2023-09-01|39|17|43.59
nightly_win_deb|2023-09-09|2023-09-09|38|1|2.63
```

List of builds where it has happened:
<details>

```
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_rep|3098|2023-09-01 04:00:01.454000|windows-container-0be6c6fc
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_rep|3101|2023-09-04 04:00:01.562000|windows-container-b2798132
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_rep|3104|2023-09-07 04:00:01.204000|windows-container-b2798132
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_rep|3105|2023-09-08 04:00:01.718000|windows-container-8f45de3b
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_deb|2845|2023-09-09 04:00:01.402000|windows-container-b2798132
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_rep|3106|2023-09-09 04:00:01.413000|windows-container-61f1fae3
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_rep|3107|2023-09-10 04:00:01.560000|windows-container-b2798132
rclcpp.TestRate.wall_rate_basics|rclcpp|nightly_win_rep|3109|2023-09-11 23:00:01.434000|windows-container-7abc1e20
[...]

PR #2329, opened 2023-10-06
address rate related flaky tests.

fujitatomoya:
https://ci.ros2.org/view/nightly/job/nightly_win_rep/3129/testReport/projectroot.test/rclcpp/test_rate/

> C:\ci\ws\src\ros2\rclcpp\rclcpp\test\rclcpp\test_rate.cpp(61): error: Value of: r.sleep()

we did not really change the test here.

https://github.com/ros2/rclcpp/blob/rolling/rclcpp/test/rclcpp/test_rate.cpp#L61

originally if it did not miss the cycle, no matter it actually sleeps or not, it will return true.
i think this is intention, since time goes by during calling `sleep` method. if that is racy, it actually does not sleep in here.

> unknown file: error: C++ exception with description "context is already initialized" thrown in the test body.

i think this is just a secondary problem because `ASSERT_TRUE(r.sleep());` will not get to `rclcpp::shutdown();`.
i believe we should use texture to avoid these possible confusions...

### DIFF
--- a/rclcpp/src/rclcpp/rate.cpp
+++ b/rclcpp/src/rclcpp/rate.cpp
@@ -67,7 +67,8 @@ Rate::sleep()
   // Calculate the time to sleep
   auto time_to_sleep = next_interval - now;
   // Sleep (will get interrupted by ctrl-c, may not sleep full time)
-  return clock_->sleep_for(time_to_sleep);
+  clock_->sleep_for(time_to_sleep);
+  return true;
 }
 
 bool
--- a/rclcpp/test/rclcpp/test_rate.cpp
+++ b/rclcpp/test/rclcpp/test_rate.cpp
@@ -40,7 +40,7 @@ public:
  */
 TEST_F(TestRate, rate_basics) {
   auto period = std::chrono::milliseconds(1000);
-  auto offset = std::chrono::milliseconds(500);
+  auto offset = std::chrono::milliseconds(300);
   auto epsilon = std::chrono::milliseconds(100);
   double overrun_ratio = 1.5;
 
@@ -95,7 +95,7 @@ TEST_F(TestRate, rate_basics) {
 
 TEST_F(TestRate, wall_rate_basics) {
   auto period = std::chrono::milliseconds(100);
-  auto offset = std::chrono::milliseconds(50);
+  auto offset = std::chrono::milliseconds(30);
   auto epsilon = std::chrono::milliseconds(1);
   double overrun_ratio = 1.5;
 
@@ -155,7 +155,7 @@ TEST_F(TestRate, wall_rate_basics) {
  */
 TEST_F(TestRate, generic_rate) {
   auto period = std::chrono::milliseconds(100);
-  auto offset = std::chrono::milliseconds(50);
+  auto offset = std::chrono::milliseconds(30);
   auto epsilon = std::chrono::milliseconds(1);
   double overrun_ratio = 1.5;
 
--- a/rclcpp/test/rclcpp/test_rate.cpp
+++ b/rclcpp/test/rclcpp/test_rate.cpp
@@ -40,7 +40,7 @@ public:
  */
 TEST_F(TestRate, rate_basics) {
   auto period = std::chrono::milliseconds(1000);
-  auto offset = std::chrono::milliseconds(300);
+  auto offset = std::chrono::milliseconds(500);
   auto epsilon = std::chrono::milliseconds(100);
   double overrun_ratio = 1.5;
 
@@ -95,7 +95,7 @@ TEST_F(TestRate, rate_basics) {
 
 TEST_F(TestRate, wall_rate_basics) {
   auto period = std::chrono::milliseconds(100);
-  auto offset = std::chrono::milliseconds(30);
+  auto offset = std::chrono::milliseconds(50);
   auto epsilon = std::chrono::milliseconds(1);
   double overrun_ratio = 1.5;
 
@@ -155,7 +155,7 @@ TEST_F(TestRate, wall_rate_basics) {
  */
 TEST_F(TestRate, generic_rate) {
   auto period = std::chrono::milliseconds(100);
-  auto offset = std::chrono::milliseconds(30);
+  auto offset = std::chrono::milliseconds(50);
   auto epsilon = std::chrono::milliseconds(1);
   double overrun_ratio = 1.5;
 
--- a/rclcpp/test/rclcpp/test_rate.cpp
+++ b/rclcpp/test/rclcpp/test_rate.cpp
@@ -21,12 +21,24 @@
 
 #include "../utils/rclcpp_gtest_macros.hpp"
 
+class TestRate : public ::testing::Test
+{
+public:
+  void SetUp()
+  {
+    rclcpp::init(0, nullptr);
+  }
+
+  void TearDown()
+  {
+    rclcpp::shutdown();
+  }
+};
+
 /*
    Basic tests for the Rate and WallRate classes.
  */
-TEST(TestRate, rate_basics) {
-  rclcpp::init(0, nullptr);
-
+TEST_F(TestRate, rate_basics) {
   auto period = std::chrono::milliseconds(1000);
   auto offset = std::chrono::milliseconds(500);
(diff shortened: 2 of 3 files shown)


fujitatomoya:
address https://github.com/ros2/rclcpp/issues/2327

AlexeyMerzlyakov:
The initial problem is that the test is being failed each time at the second `r.sleep()` call after waiting for some time less than rate period.
From the logs it seem that the failure case took exactly 2 seconds:
consoleText1_marked
which means that `Rate::sleep()` implementation in both cases called `Clock::sleep_for(1 second)` (`sleep_until()` wrapper) and waited until the end of required period. However for some reasons, in second call it was returned false-state, which from looking in the `Clock::wait_until()` code might be related to situation when the context is not valid or time source was changed during the sleep.

Now, testing it on Linux, but it extremely stable: 0 fails after ~700 runs in a row, so this requires some debugging on Windows platform. Unfortunately, I have no Windows setup yet, and according to Murphy's law, at the next week will be OOO without online connection :disappointed: Anyway, I am trying to make Windows local setup until the end of day to ge more clues on this.

And, yes, the problem with `context is already initialized` exception - is inducted problem after first assertion failure appeared.

AlexeyMerzlyakov (commented):
The idea to make `TestRate` basic class is right to avoid inducted problems in next testcases after `ASSERT_FAIL` was previously triggered.

I am not sure about the ignoring the `sleep_for` return value in Rate, since the failure during the clock sleep should be transmitted back to the end-developer, I believe. However, as a countermeasure to fix the stability of Windows builds (until the problem cause won't be found), it should be suitable.

rclcpp/test/rclcpp/test_rate.cpp
```diff
@@ -21,12 +21,24 @@
...
+
 /*
    Basic tests for the Rate, WallRate and ROSRate classes.
  */
-TEST(TestRate, rate_basics) {
-  rclcpp::init(0, nullptr);
-
+TEST_F(TestRate, rate_basics) {
   auto period = std::chrono::milliseconds(1000);
   auto offset = std::chrono::milliseconds(500);
```
fujitatomoya, line 43: > It seems like this is performance problem: when while sleeping 500ms (with rate period 1000ms), we are actually sleeping more than 1000ms.

according to @AlexeyMerzlyakov, these are too racy for windows platform. probably we should extend this like,

```suggestion
  auto period = std::chrono::milliseconds(2000);
  auto offset = std::chrono::milliseconds(1000);
```

any thoughts?

rclcpp/src/rclcpp/rate.cpp
```diff
@@ -67,7 +67,8 @@ Rate::sleep()
   // Calculate the time to sleep
   auto time_to_sleep = next_interval - now;
   // Sleep (will get interrupted by ctrl-c, may not sleep full time)
-  return clock_->sleep_for(time_to_sleep);
+  clock_->sleep_for(time_to_sleep);
+  return true;
```
fujitatomoya, line 71: I still think that this is the way we could take. but irrelevant to this flaky test reason.

since time goes by during calling `sleep_for` method. if that is racy, it actually does not sleep in here.

fujitatomoya (commented):
@AlexeyMerzlyakov thank you very much for the analysis.

rclcpp/test/rclcpp/test_rate.cpp
AlexeyMerzlyakov: After some analysis, I've discovered that the problem appeared due to being late inside `rclcpp::sleep_for()` waiting cycle: condition variable [sleeps](https://github.com/ros2/rclcpp/blob/7f7ffcf6ed1143b4d4a9e0cf248e190731b2e5f8/rclcpp/src/rclcpp/context.cpp#L499) for a required period of time twice, which determines the multiple increase of sleeping time (from 500ms -> to 1000ms).

So, I'd like to change the `offset` value only, to be equal to `30%` from `period` time in the `rate_basics`, `wall_rate_basics` and `generic_rate` testcases. This will allow even if `rclcpp::sleep_for()` is missed its time, it will sleep only 60% of period, and we won't miss rate timings.

Local testing shown promising failure rate: only 1 failure of 100 runs in a row. This failure rate was related to measured delta after `GenericRate::sleep()` (which uses `rclcpp::sleep_for()` in its sleep implementation) was not in between `(period-epsilon ... 1.5*period)` range. This to be the same `rclcpp::sleep_for()` problem, but to deal with it, we need to change its main cycle mechanism. Anyway, the failure rate after chaning `offset` to be 30% from `period` seems very promising to apply in this test.

rclcpp/src/rclcpp/rate.cpp
AlexeyMerzlyakov: Okay, let's do this, since in original code,the return value of `rclcpp::sleep_for()` [was ignored](https://github.com/ros2/rclcpp/blob/769a9d0439afade4f526acdc9588a8c15c88f330/rclcpp/include/rclcpp/rate.hpp#L86), we'd better to follow this logic in new code, as well

rclcpp/test/rclcpp/test_rate.cpp
fujitatomoya: @AlexeyMerzlyakov thanks for the analysis, addressed in https://github.com/ros2/rclcpp/pull/2329/commits/1097ba03c5c372bca9483dc66388adb7585a12e3.

btw,

> rclcpp::sleep_for() waiting cycle: condition variable [sleeps](https://github.com/ros2/rclcpp/blob/7f7ffcf6ed1143b4d4a9e0cf248e190731b2e5f8/rclcpp/src/rclcpp/context.cpp#L499) for a required period of time twice, which determines the multiple increase of sleeping time (from 500ms -> to 1000ms).

is this because of system tick resolution problem? i am not sure why this would happen.

fujitatomoya:
CC: @Crola1702

rclcpp/test/rclcpp/test_rate.cpp
fujitatomoya: see https://github.com/ros2/rclcpp/pull/2329/commits/bff0d84978f6fbbad0cde62662db46e218bd5ad5

and reverted https://github.com/ros2/rclcpp/pull/2329/commits/8dee8b3bd7358a09804576d26341b6c3f9444ce9

AlexeyMerzlyakov: Great! The proposed in https://github.com/ros2/rclcpp/commit/bff0d84978f6fbbad0cde62662db46e218bd5ad5 solution is used to sleep with remained amount of time on each iteration instead of fixed. This seems to be correct: even if we missed first cycle, the program will sleep-up remaining period of time.

Run-time verification (on `test_rate`) on Windows confirms that the cycle iteration sequence is being converged for `Context::sleep_for()`: maximum number of iterations for waiting cycle was `2` while average number of iterations during 50 runs - is `1.03`. On the Linux there were no cases with more than `1` cycle iteration were detected for 50 test runs.

fujitatomoya:
Note: we need to backport this to all distributions.

AlexeyMerzlyakov (approved):
The local testing on Linux and Windows shown no regressions for `colcon test --packages-select rclcpp`; and pretty stable results of `test_rate.exe` on local Windows: 
- vanilla rclcpp: 14% test failure rate (21 test failures out of 150 runs in a row)
- after patch applied: <1% test failure rate (3 test failures out of 350 runs in a row)
On Windows, `test_rate` sometimes still fail due to program slept more than required of time, even after one `Context::sleep_for()` cycle iteration. This number was noted when I've made the stress-CPU testing in parallel. So, it might be considered as flaky performance results. In any case, it is need to be tested on Windows CI, to confirm the change.

As from my side: I am happy with this change

fujitatomoya:
CI failure https://ci.ros2.org/job/ci_linux/19683/testReport/junit/ros2lifecycle.ros2lifecycle.test/test_cli/test_cli/ is known problem from https://github.com/ros2/ros2cli/issues/861

this PR is ready to go merge with another review.

clalancette (approved):

fujitatomoya:
@Mergifyio backport iron humble

PR #2335, opened 2023-10-11
Disable the loaned messages inside the executor.

clalancette:
They are currently unsafe to use; see the comment in the commit for more information.

This was originally brought up by https://github.com/ros2/rmw_cyclonedds/issues/469 , but we also had another report of it from Nav2 (parts reproduced here for context):

From @SteveMacenski 
```
*    It looks like Fast-DDS returns the same shared pointer memory for multiple subscriber callbacks in a row
*    We take the shared_ptr message from the callback and store it for use. So later on, the community member is identifying that Fast-DDS sets the same underlying memory in msg to the new message, even though we have stored the shared pointer in our code and just reusing it, wiping away our program's state
```

From @wjwwood:
```
This is probably a bug in rclcpp: https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/executor.cpp#L626-L662
```

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -623,12 +623,7 @@ Executor::execute_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
     // Deliver ROS message
     case rclcpp::DeliveredMessageKind::ROS_MESSAGE:
       {
-        // TODO(clalancette): The loaned message interface is currently not safe to use.
-        // If a user takes a reference to the shared_ptr, it can get freed from underneath them
-        // via rcl_return_loaned_message_from_subscription().  The correct solution is to return
-        // the loaned message in a custom deleter, but that needs to be carefully handled with
-        // locking.  Disable the entire interface for now until we sort through the issues.
-        if (false && subscription->can_loan_messages()) {
+        if (subscription->can_loan_messages()) {
           // This is the case where a loaned message is taken from the middleware via
           // inter-process communication, given to the user for their callback,
           // and then returned.
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -623,7 +623,12 @@ Executor::execute_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
     // Deliver ROS message
     case rclcpp::DeliveredMessageKind::ROS_MESSAGE:
       {
-        if (subscription->can_loan_messages()) {
+        // TODO(clalancette): The loaned message interface is currently not safe to use.
+        // If a user takes a reference to the shared_ptr, it can get freed from underneath them
+        // via rcl_return_loaned_message_from_subscription().  The correct solution is to return
+        // the loaned message in a custom deleter, but that needs to be carefully handled with
+        // locking.  Disable the entire interface for now until we sort through the issues.
+        if (false && subscription->can_loan_messages()) {
           // This is the case where a loaned message is taken from the middleware via
           // inter-process communication, given to the user for their callback,
           // and then returned.
@@ -669,6 +674,11 @@ Executor::execute_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
             subscription->get_topic_name(),
             [&]() {return subscription->take_type_erased(message.get(), message_info);},
             [&]() {subscription->handle_message(message, message_info);});
+          // TODO(clalancette): In the case that the user is using the MessageMemoryPool,
+          // and they take a shared_ptr reference to the message in the callback, this can
+          // inadvertently return the message to the pool when the user is still using it.
+          // This is a bug that needs to be fixed in the pool, and we should probably have
+          // a custom deleter for the message that actually does the return_message().
           subscription->return_message(message);
         }
         break;
--- a/rclcpp/src/rclcpp/subscription_base.cpp
+++ b/rclcpp/src/rclcpp/subscription_base.cpp
@@ -298,7 +298,20 @@ SubscriptionBase::setup_intra_process(
 bool
 SubscriptionBase::can_loan_messages() const
 {
-  return rcl_subscription_can_loan_messages(subscription_handle_.get());
+  bool retval = rcl_subscription_can_loan_messages(subscription_handle_.get());
+  if (retval) {
+    // TODO(clalancette): The loaned message interface is currently not safe to use with
+    // shared_ptr callbacks.  If a user takes a copy of the shared_ptr, it can get freed from
+    // underneath them via rcl_return_loaned_message_from_subscription().  The correct solution is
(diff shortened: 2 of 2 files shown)


fujitatomoya (commented):
@clalancette i am not sure if you had the issue, but i guess this one is known constraint originally.

https://github.com/ros2/rcl/blob/e37d7f8532c287df0fd8d3727677f5dde20dd5f3/rcl/include/rcl/subscription.h#L641-L644

which i do agree that is still unsafe, because user still would copy the message in the user callback. (i think that is your concern.)

(i think you put this as draft because of this.) i am not sure if we can forcibly disable this feature on subscription, since it will disable entire zero copy feature with ROS 2. if we go that way, i feel like we should never have enabled `can_loan_message` on the subscription... 😓

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -623,7 +623,12 @@ Executor::execute_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
     // Deliver ROS message
     case rclcpp::DeliveredMessageKind::ROS_MESSAGE:
       {
-        if (subscription->can_loan_messages()) {
+        // TODO(clalancette): The loaned message interface is currently not safe to use.
+        // If a user takes a reference to the shared_ptr, it can get freed from underneath them
```
wjwwood, line 627: Technically it's not a reference to the shared pointer that's the issue. That wouldn't be safe with or without loaning. It's if they copy the shared pointer.

wjwwood (approved):
Lgtm, as we discussed it offline. 

With one comment about how the todo could be worded more precisely.

wjwwood:
The users can still take loaned messages from subscriptions manually but it will effectively disable zero copy for callbacks until we make more changes.

wjwwood:
Also we should cross reference the existing issue and add the context we got from the Nav2 users here so that future us knows what is going on 😅

clalancette:
> which i do agree that is still unsafe, because user still would copy the message in the user callback. (i think that is your concern.)

Yes, that is the major concern.  The original issue mentioned here is https://github.com/ros2/rmw_cyclonedds/issues/469 , though we had another report on it today.

> (i think you put this as draft because of this.) i am not sure if we can forcibly disable this feature on subscription, since it will disable entire zero copy feature with ROS 2. if we go that way, i feel like we should never have enabled `can_loan_message` on the subscription... 😓

The problem is that right now, shared_ptr subscriptions are super dangerous to use.  Users who take a reference to the shared_ptr inside the callback (rightly) expect that the object will persist, but that is not the case with this automatic loaning.  So I kind of think we *have* to make this change for now.

That said, we can talk about options for reenabling it in the future.  That's why this draft is just disabling the feature, not removing it.

clalancette:
> Also we should cross reference the existing issue and add the context we got from the Nav2 users here so that future us knows what is going on 😅

Good point.  I'll modify the initial comment to have that context.

fujitatomoya:
@clalancette @wjwwood thanks for the explanation. i am not happy with this 💦  but so true...

how about this approach? cz forcibly and statically hard-code disabling can be surprising (current user who understands this constrain) and leaves us no choice instead of rebuilding.

```c
diff --git a/rcl/src/rcl/subscription.c b/rcl/src/rcl/subscription.c
index 1ca6edd..9b24852 100644
--- a/rcl/src/rcl/subscription.c
+++ b/rcl/src/rcl/subscription.c
@@ -233,7 +233,7 @@ rcl_subscription_get_default_options()
   default_options.rmw_subscription_options = rmw_get_default_subscription_options();

   // Load disable flag to LoanedMessage via environmental variable.
-  bool disable_loaned_message = false;
+  bool disable_loaned_message = true;
   rcl_ret_t ret = rcl_get_disable_loaned_message(&disable_loaned_message);
   if (ret == RCL_RET_OK) {
     default_options.disable_loaned_message = disable_loaned_message;
@@ -241,7 +241,7 @@ rcl_subscription_get_default_options()
     RCUTILS_SAFE_FWRITE_TO_STDERR("Failed to get disable_loaned_message: ");
     RCUTILS_SAFE_FWRITE_TO_STDERR(rcl_get_error_string().str);
     rcl_reset_error();
-    default_options.disable_loaned_message = false;
+    default_options.disable_loaned_message = true;
   }

   return default_options;
```

that said, if user want to use loaned interfaces on the subscription, user must set `export ROS_DISABLE_LOANED_MESSAGES=0`, originally this environmental variable is meant to `disable` but this can be option?

fujitatomoya:
@MiguelCompany @Barry-Xu-2018 @iuhilnehc-ynos you might want to keep eyes on this.

SteveMacenski (approved):

clalancette:
So the latest commit undoes the work here. Instead, all this does now is to print a warning if loaned messages are enabled.  This now needs to go in with https://github.com/ros2/rcl/pull/1110 , which is where we actually do the disabling.

clalancette:
Here is full CI on the pair, just to make sure things work as we expect:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19650/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14158/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20374/)

SteveMacenski (approved):
While I think we should outright disable it if its not safe (for the time being until it is safe), I think this is also reasonable if the performance hit on disabling it always is super significant.

clalancette:
> While I think we should outright disable it if its not safe (for the time being until it is safe), I think this is also reasonable if the performance hit on disabling it always is super significant.

I ended up doing it this way because there *are* ways to use it safely.  In particular if you use the default message strategy and only use ConstRef callback types, it is safe to use.  That is a common use case, so users who really know what they are doing could opt-in to this to get slightly better performance.

But since there are a bunch of other scenarios that users could use (including using the shared_ptr callback type), it should be disabled by default (at least, until we fix it).

fujitatomoya (approved):

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-10-12/34178/1

wjwwood (approved):

clalancette:
CI for this is in https://github.com/ros2/rcl/pull/1110#issuecomment-1791116916

clalancette:
@Mergifyio backport humble iron

ros-discourse:

https://discourse.ros.org/t/preparing-for-humble-sync-and-patch-release-2023-11-17/34628/3

ros-discourse:

https://discourse.ros.org/t/important-announcement-patch-release-3-and-new-packages-for-iron-irwini-2023-11-20/34742/1

ros-discourse:

https://discourse.ros.org/t/preparing-for-iron-patch-and-sync-2023-11-17/34609/2

PR #2336, opened 2023-10-12
Support users holding onto shared pointers in the message memory pool

clalancette:
Before this commit, the MessageMemoryPool would actually reuse messages in the pool, even if the user had taken additional shared_ptr copies.

This commit fixes things so that we properly handle that situation.  In particular, we only return shared_ptrs to the free pool once their use_count has dropped to 1 (meaning only this class still has a reference to it).  We also add in locks, since in a multi-threaded scenario we need to protect against multiple threads accessing the pool at the same time.

With this in place, things work as expected when users hold shared_ptr copies.  We also add in a test for this situation.

One note about performance: this update preserves the "no-allocations-at-runtime" aspect of the MessagePool.  However, there are some tradeoffs with CPU time here, particularly with very large message pools.  This could probably be optimized further to do less work when trying to add items back to the free_list, but I view that as a further enhancement.


rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
```diff
@@ -68,16 +71,25 @@ class MessagePoolMemoryStrategy
...
+    std::lock_guard<std::mutex> lock(pool_mutex_);
+    if (free_list_.size() == 0) {
+      for (size_t i = 0; i < Size; ++i) {
+        if (pool_[i].use_count() == 1) {
+          free_list_.push_back(i);
+          break;
+        }
+      }
+      if (free_list_.size() == 0) {
+        throw std::runtime_error("No more free slots in the pool!");
```
SteveMacenski, line 83: It would be nice if this was an specialized exception that could be captured for fixing when appears

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
```diff
@@ -87,24 +99,68 @@ class MessagePoolMemoryStrategy
...
+        }
       }
     }
-    throw std::runtime_error("Unrecognized message ptr in return_message.");
   }
 
 protected:
-  struct PoolMember
+  template<size_t N>
+  class CyclicSizeTArray
```
wjwwood, line 124: nitpick: is this just a circular buffer? Seems like we could call it that, or something like `CircularBuffer<T, N>`? "Cyclic<T>Array" feels a bit like trying to mention a character from another franchise without soliciting a copy-right strike, lol. I don't feel strongly about this, just struck me as weird.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
wjwwood, line 83: nitpick: don't use `!`, just state the error, no exclamation.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
```diff
@@ -87,24 +99,68 @@ class MessagePoolMemoryStrategy
...
-        return;
+    (void)msg;
+
+    // What we really want to do here is to figure out whether the user has taken an additional
+    // reference to the message, and only add it to the free list if that is *not* the case.
+    // However, we can't really do that for the currently passed-in msg; it can have an arbitrary
+    // reference count due to the mechanisms of rclcpp.  Instead, we look at all the rest of the
+    // pointers, and add the ones that the user has released into the free pool.
+    // We do the same thing in borrow_message(), so if the user has a pool of size 1
+    // (or only one free slot), we'll always find it.
```
wjwwood, line 110: I'm not convinced by this strategy, specifically I don't think we should be calling `return_message` when the given message isn't eligible for return. I'd go so far as to say, if `msg.use_count()` is not equal to `1` we throw here. The implication of this is that we should rely on the destructor being called when use_count reaches zero, which we're not doing here. I assume that's to do with not wanting to propagate the custom destructor? Or is it something else?

wjwwood (commented):
I have doubts about this approach, maybe we could discuss it more?

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
clalancette: > nitpick: is this just a circular buffer? Seems like we could call it that, or something like `CircularBuffer<T, N>`? "CyclicArray"

Yeah, it's a circular buffer.  The name `CyclicArray` came about because I was trying to show that it had no allocations at runtime, like `std::array`.  Maybe `CircularArray` instead?  I don't really have a strong opinion.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
clalancette: Fixed in 0eaf242566acb5b2ff53ed4aec810ab4df2cb4b0 .

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
clalancette: > I assume that's to do with not wanting to propagate the custom destructor? Or is it something else?

It has to do with the interface between the Executor class and the message strategies.

In particular, in [Executor::execute_subscription](https://github.com/ros2/rclcpp/blob/7f7ffcf6ed1143b4d4a9e0cf248e190731b2e5f8/rclcpp/src/rclcpp/executor.cpp#L615) you have this sequence:

```
          std::shared_ptr<void> message = subscription->create_message();
          take_and_do_error_handling(
            "taking a message from topic",
            subscription->get_topic_name(),
            [&]() {return subscription->take_type_erased(message.get(), message_info);},
            [&]() {subscription->handle_message(message, message_info);});
          subscription->return_message(message);
```

And the thing is, you *have* to do that sequence for the default message strategy of allocation.  That is, you need to "borrow" (allocate), pass that to the user callback, and then "return" (deallocate).  For a user callback with ConstRef type, this does exactly what we expect.  For a user callback with a shared_ptr type, the user may have incremented the reference count, in which case the "return" won't actually deallocate, but will instead decrement the reference count.  But either way, you have to do that decrement, otherwise you'll have a memory leak.

So with that interface in mind, you now come to this message pool, which has to fit into the same interface.  The "borrow" doesn't do an allocation, but instead just pulls the message from a preallocated pool.  The "return" really wants to return the message to the pool, but there are several internal rclcpp references to the shared_ptr at that time; one from the pool itself, one from `Executor::execute_subscription`, and one from https://github.com/ros2/rclcpp/blob/7f7ffcf6ed1143b4d4a9e0cf248e190731b2e5f8/rclcpp/include/rclcpp/subscription.hpp#L371 .

I originally implemented this with a check that returns the current message to the pool if `use_count() == 3`, and that works.  But it seemed really fragile to rely on that implementation detail (though we could mitigate that fragility with some tests).  That's why I ended up with this strategy of not checking the current message, but checking everything else in the pool instead (since by that point, all of the internal rclcpp references should have been dropped).

Does this make sense?

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
fujitatomoya: out of curiosity, can we just use sized `std::queue` thing? it would be easier to implement this class?

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
fujitatomoya: how about letting the user application know this situation at least once with warning message? i think it would be better to do that for user and debugging purpose instead of being silent?

clalancette: > how about letting the user application know this situation at least once with warning message?

Can you please clarify which situation you think we should warn about?  Thanks.

fujitatomoya: I was thinking something like this would be better? instead of being silent.

```suggestion
    if (msg.use_count() >= 3) {
      RCLCPP_WARN_ONCE(
        rclcpp::get_logger("rclcpp"),
        "User application still keeps the reference to the message.",
        "The message memory cannot be reused until user frees the reference to the message.");
```

clalancette: > I was thinking something like this would be better? instead of being silent.

Oh, right.  Well, it is a bit tricky.  In some sense, if the user holds onto a shared_ptr, they should know what they are doing and eventually release it.  It is kind of a limitation/feature of C++.  So my vote would not be to warn in this situation.

clalancette:
I talked to @wjwwood about this in some detail.

In particular, we talked about the interface between Executor::execute_subscription and the memory pools.  The current interface is actually quite odd, in that we get a reference to a `shared_ptr` via `create_message`, but then have to "return" it with `return_message`.  But the thing is that this is already a shared_ptr; it should be enough at this point for the executor to drop the reference and let the shared_ptr machinery do the right thing (this right thing might include a custom deleter).

So what I'm going to do here is attempt to make it so that `return_message` doesn't actually do anything.  Instead, what I'm going to do is to try to make the pool a pool of raw memory pointers.  During `borrow_message`, we'll actually allocate a `shared_ptr` to wrap that raw memory.  And then we don't need to do any additional work in `return_message`; when the executor drops the reference, it will either be deleted (if the user has not taken a reference), or it will decrease the use_count (in which case it will be deleted later on when the user drops the reference).

This same idea works for the default allocation strategy as well; I'll likely open a separate PR for that later.  And beyond that, a similar strategy should also work for the loaned message interface.

clalancette:
OK, I've implemented the latest idea in this PR, and it seems to work fine.  @wjwwood I'd appreciate a look at it to see what you think.

clalancette:
Since I'm actually pretty happy with this implementation now, I'm going to mark it as "Ready for Review" and run CI on it as well.

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19678/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14184/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20403/)

fujitatomoya (approved):
no further comments. lgtm with green CI.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
wjwwood: I'm not sure @fujitatomoya, maybe? I'd have to look at how the functions in `std::queue` work. My concern was mostly about naming tbh. It does feel like it could be generalized, e.g. not specific to `size_t`.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
```diff
@@ -68,16 +79,24 @@ class MessagePoolMemoryStrategy
...
-    pool_[current_index].used = true;
-    return pool_[current_index].msg_ptr_;
+    size_t current_index = free_list_.pop_front();
+
+    return std::shared_ptr<MessageT>(
+      pool_[current_index], [this](MessageT * p) {
+        for (size_t i = 0; i < Size; ++i) {
+          if (pool_[i] == p) {
+            p->~MessageT();
+            new(p) MessageT();
```
wjwwood, line 94: It feels weird to call the constructor here. Imo, it should either:

- call the constructor on borrow, or
- never call the constructor after the first time

wjwwood (changes requested):
One major comment, but otherwise lgtm.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
clalancette: Both of these are problematic, I think.

> * call the constructor on borrow, or

We could do this, but then we'd have to be more careful in the destructor of `MessagePoolMemoryStrategy`.  In particular, you can't call `delete` on an object that has already had the destructor run (at least, valgrind gives me warnings when I do this).  So we'd have to do additional tracking to not try to `delete` these (and even that doesn't work, since then we'd leak the memory).

> * never call the constructor after the first time

This I don't think we can do because it will break semantics of the messages.  In particular, assume you had a `std::vector` in one of the messages (which is what we generate for unbounded arrays).  So during the first borrow, the user code pushes 2 items onto the `std::vector`, then returns it to the pool.  Now the next borrow that uses that slot gets an object that already has 2 items in the vector.  Yes, this would be a weird usage, but I think we should always return freshly-initialized objects to the user code.

The current code basically always ensures that we have a freshly-initialized object ready to hand out in the pool.  And to me, that seems to be the easiest to reason about.  Thoughts?

wjwwood: Despite the need to be careful with the destructor of the pool, I think the first option I suggested is better. Consider also that the constructor of the messages might throw (`std::bad_alloc` for example), and you'd be doing that from the destructor of the shared pointer, rather than from the borrow call.

For the second option, I agree we'd probably want to return clean messages, but for example in the case of using a borrowed message with take, it always resets all the fields anyway, so it might be a small performance improvement to avoid the destructor/constructor in some cases (users do this implicitly with a message on the stack in a class with a publish loop sometimes). Maybe the answer is to destruct then construct in the borrow function with an optional flag to skip that step (with understanding that the user will be responsible for dealing with stale data in the message).

Either way, I really think at least construction should be done in the borrow call.

clalancette: > Either way, I really think at least construction should be done in the borrow call.

All right, that's fair.  In 7901ef3, I ended up revamping this so we only do allocation/deallocation in the `MessagePoolMemoryStrategy` constructor and destructor, respectively.  We then do placement-new construction when we hand out the `shared_ptr` from the pool, and manually run the destructor when the custom `shared_ptr` deleter is called.  I think this should satisfy all of the constraints here; please take another look.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19705/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14214/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20434/)

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
```diff
@@ -52,11 +59,28 @@ class MessagePoolMemoryStrategy
...
+  {
+    // The user may have held onto shared pointers after a borrow_message().  In that case,
+    // freeing the memory from the pool may lead to UB.  If we detect the situation where this
+    // class is being destroyed before the shared pointers, warn the user.
+
+    if (free_list_.size() != Size) {
+      RCLCPP_WARN(
+        rclcpp::get_logger("MessagePool"),
+        "User code is holding onto shared pointers from the message pool; this will leak memory");
+    }
```
wjwwood, line 79: This is fine for now (or forever if no one cares to improve it), but I just wanted to note that we could probably do something so that the destructor of the shared pointers will free the memory if the pool no longer exists. Maybe that could be a todo or something. This improvement would also let us optimize a bit and maybe allocate/free the memory all at once, maybe by having all the shared pointers that are borrowed also have a shared pointer to the arena of memory they're allocated from, that way after the last one is destroyed the actual memory for the pool would also be deleted (assuming the memory pool had let its handle on it go too).

Either way, it might be worth mentioning this limitation / expected pre condition to destroying the pool in the pool's docstring.

wjwwood: I can't suggest on the docstring because it's not part of the diff, but something like:

> All borrowed message must be returned before deleting the memory pool otherwise memory will be leaked.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
```diff
@@ -68,16 +92,25 @@ class MessagePoolMemoryStrategy
...
+      [this](MessageT * p) {
+        std::lock_guard<std::mutex> lock(pool_mutex_);
+        for (size_t i = 0; i < Size; ++i) {
+          if (pool_[i] == p) {
+            p->~MessageT();
+            free_list_.push_back(i);
+            break;
+          }
+        }
+      });
```
wjwwood, line 113: If the pool is destroyed before the borrowed shared pointers, then this code is broken too.

So either we need to just say this is undefined behavior (destroying the pool before all borrows are returned) or we need to improve this code.

If we do the first, then your warning about leaking memory is not necessary anymore (imo) since it's broken no matter what.

To fix this code, you'd need to make this pool an `enable_shared_from_this` and then capture a weak_ptr to it in this lambda, to avoid a circular reference. And then this code would need to just run the destructor and not check the pool or free list if the weak_ptr was unable to `lock()`.

Or alternatively you could capture the `pool_` itself (and the `free_list_`) and make it a shared pointer, and move the free'ing logic into its destructor.

wjwwood (changes requested):
This is looking a lot better, but I found an issue in the destructing logic that needs some kind of addressing.

rclcpp/include/rclcpp/strategies/message_pool_memory_strategy.hpp
clalancette: > Or alternatively you could capture the `pool_` itself (and the `free_list_`) and make it a shared pointer, and move the free'ing logic into its destructor.

Yeah, this is all a really good point.  I ended up going this direction, where all of the member variables are `shared_ptr`, and hence can outlive this class.  See ec86e41 .  I verified with this code that holding the references past when `MessagePoolMemoryStrategy` is destroyed does the right thing (at least, valgrind gave me no errors in that situation).

Please take another look.

wjwwood (approved):
lgtm, thanks for iterating :)

Issue #2202, opened 2023-06-02
Discussion: add ignore-local-endpoints functionality


alsora:
ROS 2 currently has a problem with double-delivery of messages when combining intra-process and inter-process communication.
Consider for example a scenario where you have two processes:
 - a process contains publisher A and subscriber X
 - a process contains subscriber Y

When publisher A publishes a message, we would expect the message to be delivered to subscriber X via intra-process comm and to subscriber Y via inter-process comm.

However, what actually happens is a double delivery: the inter-process publication will indeed reach also subscriber X, which will then discard the message, but still wasting a non-negligible amount of resources.

In order to get around this problem, it's necessary to tell publisher A that it doesn't need to send inter-process messages to subscriber X which is in the same process and it's going to be serviced by intra-process comm.

ROS already has the concept of "ignore local endpoints", however this is applied to the whole process.
This can be problematic if in a process there are both entities with intra-process comm enabled and disabled.

We are working with eProsima to allow a more fine-grained control of the "ignore local endpoints" functionality, such that each ROS 2 entity can indicate that to the RMW.
This will allow to solve the problem of the double delivery.

In order to implement that, it looks like it's necessary to make some changes to the rclcpp logic that decides whether to publish inter or intra process.

Right now we check for 
```
bool inter_process_publish_needed = get_subscription_count() > get_intra_process_subscription_count()
```

However, if publisher A is "ignoring local endpoints", this means that `get_subscription_count()` will only return the non-local ones.

We think that the logic here should be changed into something like

```
bool inter_process_publish_needed = get_non_local_subscription_count() > 0;
```

fujitatomoya:
@alsora just checking my understanding.

So if we just rely on RMW `LoanedMessage` for intra-process communication w/o rclcpp one, this should be no problem? since that can be managed by RMW implementation? (i understand the performance is not as same with `rclcpp intra-process communication`)

thanks,

MiguelCompany:
@alsora @fujitatomoya 

I prepared a set of PRs for the following:

> We think that the logic here should be changed into something like
> 
> ```
> bool inter_process_publish_needed = get_non_local_subscription_count() > 0;
> ```

- [ ] RMW interface
  - [ ] rmw: https://github.com/ros2/rmw/pull/358
  - [ ] rmw_implementation: https://github.com/ros2/rmw_implementation/pull/221
- [ ] RMW implementations
[...]

PR #2338, opened 2023-10-20
Feature/count non local subscriptions

MiguelCompany:
Adds a method in `PublisherBase` that returns the number of matched subscriptions that are not in the same context as the publisher.

It also makes inter-process publishing depend on this new method.

Depends on https://github.com/ros2/rcl/pull/1111

Part of https://github.com/ros2/rclcpp/issues/2202

full repos file [here](https://gist.github.com/MiguelCompany/c92eb293f97ad82ced4aa1f4b0c60d5e/raw/fe4f53f423377908cf2cebfd7bc666c3d67a356f/count_non_local_subscriptions_rolling.repos)

### DIFF
--- a/rclcpp/include/rclcpp/publisher_base.hpp
+++ b/rclcpp/include/rclcpp/publisher_base.hpp
@@ -133,6 +133,12 @@ public:
   size_t
   get_subscription_count() const;
 
+  /// Get non local subscription count
+  /** \return The number of non local subscriptions. */
+  RCLCPP_PUBLIC
+  size_t
+  get_non_local_subscription_count() const;
+
   /// Get intraprocess subscription count
   /** \return The number of intraprocess subscriptions. */
   RCLCPP_PUBLIC
--- a/rclcpp/src/rclcpp/publisher_base.cpp
+++ b/rclcpp/src/rclcpp/publisher_base.cpp
@@ -254,6 +254,33 @@ PublisherBase::get_subscription_count() const
 }
 
 size_t
+PublisherBase::get_non_local_subscription_count() const
+{
+  size_t inter_process_non_local_subscription_count = 0;
+
+  rcl_ret_t status = rcl_publisher_get_non_local_subscription_count(
+    publisher_handle_.get(),
+    &inter_process_non_local_subscription_count);
+
+  if (RCL_RET_PUBLISHER_INVALID == status) {
+    rcl_reset_error();  /* next call will reset error message if not context */
+    if (rcl_publisher_is_valid_except_context(publisher_handle_.get())) {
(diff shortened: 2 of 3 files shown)


alsora (approved):

alsora:
Can we add a couple of unit-tests?

rclcpp/include/rclcpp/publisher_base.hpp
```diff
@@ -133,6 +133,12 @@ class PublisherBase : public std::enable_shared_from_this<PublisherBase>
   size_t
   get_subscription_count() const;
 
+  /// Get non local subscription count
```
fujitatomoya, line 136: I think this means that `Not in the same context` for the ROS 2 user (precisely for DDS-RTPS, GUID prefix is not the same), that would be nicer to add the description explains?

rclcpp/src/rclcpp/publisher_base.cpp
```diff
@@ -253,6 +253,33 @@ PublisherBase::get_subscription_count() const
...
+      if (nullptr != context && !rcl_context_is_valid(context)) {
+        /* publisher is invalid due to context being shutdown */
+        return 0;
+      }
+    }
+  }
+  if (RCL_RET_OK != status) {
+    rclcpp::exceptions::throw_from_rcl_error(
+      status,
+      "failed to get get non local subscription count");
```
fujitatomoya, line 278:
```suggestion
      "failed to get non local subscription count");
```

fujitatomoya (changes requested):

sloretz:
@MiguelCompany Friendly ping. Would you be willing to address @fujitatomoya's feedback?

mauropasse:
I've been testing the new feature enabled by:
```
bool inter_process_publish_needed = get_non_local_subscription_count() > 0;
```      
This not only fixes the "double delivery" issue (local subscription receiving inter & inter process message), but also reduces CPU at startup and destruction of subscriptions.

When creating a subscription, there is a short time gap in which the intra-process subscription has not yet been created, so messages come through inter-process (serialization/deserialization/etc) causing high CPU utillization. 
The same happens when the subscription is destroyed, since first the intra-process subscription is destroyed, it'll receive messages through the DDS until the whole subscription is destroyed.

This is particularly noticeable for `/parameter_events` built in subscriptions: in a big system with many nodes using parameters, these subscriptions might receive thousands of inter-process messages at startup.

mauropasse:
Correction about my statement above:
> This not only fixes the "double delivery" issue..

This PR _doesn't_ fix the double delivery issue. It _does_ avoids getting inter-process messages at creation/destruction of subscriptions.

Is there anything missing to avoid the double delivery issue @MiguelCompany ? Or is that's a separate effort?

PR #2343, opened 2023-10-27
Added optional TimerInfo to timer callback

jmachowinski:
This allows us to get the correct time of the timer callback as node->now() can
return a later timestamp than 'expected' due to race conditions.

This commit depends on https://github.com/ros2/rcl/pull/1113


jmachowinski:
@clalancette : ping

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -317,6 +339,7 @@ class GenericTimer : public TimerBase
   RCLCPP_DISABLE_COPY(GenericTimer)
 
   FunctorT callback_;
+  rcl_timer_call_info_t timer_call_info_;
```
jmachowinski, line 342: @atsogias I don't like this, this introduces a potential race condition. I'll push a fix for this.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -709,9 +709,9 @@ Executor::execute_subscription(rclcpp::SubscriptionBase::SharedPtr subscription)
 }
 
 void
-Executor::execute_timer(rclcpp::TimerBase::SharedPtr timer)
+Executor::execute_timer(rclcpp::TimerBase::SharedPtr timer, const std::shared_ptr<void> & dataPtr)
```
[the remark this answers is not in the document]
jmachowinski: The return from take_data() is void as it's the standard in the rest of the code. Therefore this pointer is also void.

jmachowinski:
@wjwwood  @mjcarroll  @fujitatomoya Ping, I would like to see this merged for jazzy

fujitatomoya (commented):
@jmachowinski thank you for the PR, i had a few comments.

besides, test would be ideal.

rclcpp_action/src/client.cpp
```diff
@@ -31,6 +32,63 @@
 namespace rclcpp_action
 {
 
+
+struct ClientBaseData
```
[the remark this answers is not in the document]
jmachowinski: Nope, that was not supposed to be in this PR....

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -300,11 +318,26 @@ class GenericTimer : public TimerBase
...
+    typename std::enable_if<
+      rclcpp::function_traits::same_arguments<CallbackT, TimerInfoCallbackType>::value
+    >::type * = nullptr
+  >
+  void
+  execute_callback_delegate(const rcl_timer_call_info_t & timer_call_info_)
+  {
+    const TimerInfo info{Time{timer_call_info_.expected_call_time, clock_->get_clock_type()},
+      Time{timer_call_info_.actual_call_time, clock_->get_clock_type()}};
+    callback_(info);
```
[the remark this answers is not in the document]
jmachowinski: We ran into the following problem :
System was running in Simulation mode, so it used the clock topic.
We have a timer processing odometry data at the same frequency as the clock update (100 Hz).

Also the Process was using the Multithreaded Executor. Therefore the reception of the next clock update is concurrent, as the timer and clock are in different callback groups.

On the start of each timer callback, we call node->now() to get the current time.
The problem we ran into was, that due to concurrency, It could happen, that the time jumped between the call to the timer callback and us calling node->now().
As a result of this, we got two times the same timestamp returned from node->now() in consecutive timer callbacks.

jmachowinski:
@fujitatomoya Thanks for the review, I'll update the PR accordingly. 

> besides, test would be ideal.

Hm, I'm not really sure what to test here, the only thing I can think of would be that the info contains somewhat expected values.

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -198,16 +208,23 @@ class TimerBase
   set_on_reset_callback(rcl_event_callback_t callback, const void * user_data);
 };
 
+struct TimerInfo
+{
+  Time expected_call_time;
+  Time actual_call_time;
+};
```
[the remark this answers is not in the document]
jmachowinski: removed

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -231,8 +231,13 @@ StaticSingleThreadedExecutor::execute_ready_executables(bool spin_once)
     if (i < entities_collector_->get_number_of_timers()) {
       if (wait_set_.timers[i] && entities_collector_->get_timer(i)->is_ready()) {
         auto timer = entities_collector_->get_timer(i);
-        timer->call();
-        execute_timer(std::move(timer));
+        auto data = timer->call();
+        if (!data) {
+          throw std::runtime_error(
+                  "StaticSingleThreadedExecutor::execute_ready_executables() Error,"
+                  "timer is ready, but call returned false");
```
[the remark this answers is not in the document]
jmachowinski: updated

rclcpp_action/src/client.cpp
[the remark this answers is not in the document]
jmachowinski: removed from commit

jmachowinski:
@fujitatomoya  Added a simple test. And updated the commits.

clalancette:
I haven't looked in this in too much detail, but I have to say that the use of (effectively) a `void *` pointer here is extremely suspicious.  It doesn't give us any type checking at all.  Can you give us more detail on exactly what this change is fixing?

jmachowinski:
> I haven't looked in this in too much detail, but I have to say that the use of (effectively) a `void *` pointer here is extremely suspicious. It doesn't give us any type checking at all. Can you give us more detail on exactly what this change is fixing?

https://github.com/ros2/rclcpp/pull/2343#discussion_r1495498430

Here is the concrete example, of what we are trying to fix here.

The pointer returned by take_data is void, as you need to assign it to the AnyExecutable here
https://github.com/ros2/rclcpp/blob/10252e9f66ac87f3903f301b64320d32457f0658/rclcpp/include/rclcpp/any_executable.hpp#L50
in order to pass it on inside of the Executor.

rclcpp/include/rclcpp/timer.hpp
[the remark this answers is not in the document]
jmachowinski: Second nice thing, that occurred to me while writing the tests, you can detect timer overruns using this information.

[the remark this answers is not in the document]
fujitatomoya: i see, thanks for the explanation. so especially simulation time, we can get to know the when the callback is actually fired with this PR.

jmachowinski:
> This is standard for the executor/waitable model. The waitable is supposed to return a data handle that can be associated when the actual `execute` call happens. This is because a single waitable can, by design, have multiple ready executables internally

@mjcarroll Do you think the approach I took is viable, or should I look into something more typesafe ?
Note, this would mean, that we need to replace data AnyExecutable by a std::variant,

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -101,16 +108,19 @@ class TimerBase
    * The multithreaded executor takes advantage of this to avoid scheduling
    * the callback multiple times.
    *
-   * \return `true` if the callback should be executed, `false` if the timer was canceled.
+   * \return a shared_ptr if the callback should be executed, std::nullopt if the timer was canceled.
    */
   RCLCPP_PUBLIC
-  virtual bool
+  virtual std::optional<std::shared_ptr<void>>
```
mjcarroll, line 114: I'm a little confused as to why you did `std::optional<std::shared_ptr<void>>` here.  Wouldn't `std::shared_ptr<void>` be sufficient?

jmachowinski: Yes a std::shared_ptr containing a nullptr would work, but in my opinion the semantics are nicer if using an optional.

rclcpp/test/rclcpp/test_timer.cpp
```diff
@@ -364,3 +386,77 @@ TEST_P(TestTimer, test_timer_without_autostart)
     std::chrono::nanoseconds::max().count());
   EXPECT_FALSE(timer_without_autostart->is_canceled());
 }
+
+/// Timer testing bring up and teardown
+class TestTimerWithInfo : public ::testing::TestWithParam<TimerType>
```
fujitatomoya, line 391: what is this class for? never be used for anywhere?

jmachowinski: good catch, removed...

fujitatomoya (approved):

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20588/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14949/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21312/)

rclcpp/include/rclcpp/timer.hpp
alsora: I think that having a `std::shared_ptr<void>` is better for consistency.

Other pieces of code added in this PR (like the `struct ExecutorEvent`) contain a `std::shared_ptr<void> data;` that could be a nullptr, thus we already need to check that pointer before using it.
Having this API using also an optional, seems an unnecessary additional layer

rclcpp/include/rclcpp/timer.hpp
```diff
@@ -256,27 +267,28 @@ class GenericTimer : public TimerBase
    * \sa rclcpp::TimerBase::call
    * \throws std::runtime_error if it failed to notify timer that callback will occurr
    */
-  bool
+  std::optional<std::shared_ptr<void>>
   call() override
   {
-    rcl_ret_t ret = rcl_timer_call(timer_handle_.get());
+    rcl_timer_call_info_t timer_call_info_;
```
alsora, line 273: what about creating the shared pointer here and populating it in the line below, to avoid a copy?
```
auto timer_call_info_ = std::make_shared<rcl_timer_call_info_t>();
rcl_ret_t ret = rcl_timer_call_with_info(timer_handle_.get(), timer_call_info_.get());
```

jmachowinski: I don't know this is a tradeoff thing. The copy is 128 Bytes. But we could avoid a possible not needed memory allocation.

rclcpp/include/rclcpp/timer.hpp
jmachowinski: changes it to std::shared_ptr<void>

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20659/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14990/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21362/)

rclcpp/include/rclcpp/timer.hpp
alsora: I would expect the memory allocation to be needed most of the times, unless someone is cancelling and reactivating the timer at a similar rate as its period

jmachowinski:
reworked patch after merge of https://github.com/ros2/rclcpp/pull/2142

fujitatomoya:
CI (without https://github.com/ros2/examples/pull/375)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20669/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14996/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21370/)

clalancette:
I'm going to suggest we pause on these types of changes for the moment.  We have regressions right now that we need to track down, possibly related to either https://github.com/ros2/rcl/pull/1135 or https://github.com/ros2/rclcpp/pull/2142.

See * Linux [Build Status](http://ci.ros2.org/job/ci_linux/20668/)

fujitatomoya:
@jmachowinski probably we want to rebase this before restarting CI since some fixes are merged in a week?

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20761/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15060/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21440/)

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20762/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15061/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21441/)

rclcpp/include/rclcpp/timer.hpp
jmachowinski: changed

jmachowinski:
Can I somehow tell the auto CI, that this PR and https://github.com/ros2/examples/pull/375 belong together, and I should build using both ?

mjcarroll:
The `Rpr` job isn't aware of cross-repo changes, but we can submit any arbitrary `ros2.repos` file to the other CI.  Is it just those two branches that need to be done together?

jmachowinski:
> The `Rpr` job isn't aware of cross-repo changes, but we can submit any arbitrary `ros2.repos` file to the other CI. Is it just those two branches that need to be done together?

Yes, this and examples.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20824/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15128/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21511/)

wjwwood:
Had a follow up to this one: https://github.com/ros2/rclcpp/pull/2501

Issue #2362, opened 2023-11-08
Component manager load node service queue overflows and fails to launch some nodes


xmfcx:
## Bug report

**Required Info:**

- Operating System:
  - `Ubuntu 22.04`
- Installation type: 
  - binaries
- Version or commit hash:
  - `humble`
- DDS implementation:
  - `CycloneDDS`
- Client library (if applicable):
  - `rclcpp_components`

#### Steps to reproduce issue

```py
import launch
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode
from launch_ros.actions import LoadComposableNodes


def generate_launch_description():
    container = ComposableNodeContainer(
            name='my_container',
            package='rclcpp_components',
            executable='component_container',
            namespace='',
            output='log',
    )

    loaders = []
    for i in range(30):
      
        node = ComposableNode(
            package='composition',
            plugin='composition::Talker',
            name='talker_' + str(i),
        )
        
        loaders.append(
          LoadComposableNodes(
                composable_node_descriptions=[node],
                target_container="my_container",
            )
        )
    
    return launch.LaunchDescription([container, *loaders])
```

- Save this launch file to a file called `composition_overload.py`.
- Source ROS 2 humble: `source /opt/ros/humble/setup.bash`
- Run the launch file: `ros2 launch composition_overload.py`
- Run `ros2 component list`

#### Expected behavior

To see all talkers loaded into the container (30 for this case):

```console
$ ros2 component list
/my_container
  1  /talker_0
  2  /talker_1
  3  /talker_2
  4  /talker_3
  5  /talker_4
... (omitted for clarity)
 25  /talker_24
 26  /talker_25
 27  /talker_26
 28  /talker_27
 29  /talker_28
 30  /talker_29
```

#### Actual behavior

About 2/3 is loaded.

```console
$ ros2 component list
/my_container
  1  /talker_0
  2  /talker_1
  3  /talker_2
  4  /talker_3
  5  /talker_4
  6  /talker_11
  7  /talker_12
  8  /talker_28
  9  /talker_20
  10  /talker_21
  11  /talker_22
  12  /talker_23
  13  /talker_24
  14  /talker_19
  15  /talker_25
  16  /talker_26
  17  /talker_27
  18  /talker_29
```

#### Additional information

[...]

PR #2363, opened 2023-11-09
fix(rclcpp_components): increase the service queue sizes in component_container

xmfcx:
fixes: https://github.com/ros2/rclcpp/issues/2362

### DIFF
--- a/rclcpp_components/src/component_manager.cpp
+++ b/rclcpp_components/src/component_manager.cpp
@@ -40,11 +40,11 @@ ComponentManager::ComponentManager(
   loadNode_srv_ = create_service<LoadNode>(
     "~/_container/load_node",
     std::bind(&ComponentManager::on_load_node, this, _1, _2, _3),
-    rclcpp::QoS(200).get_rmw_qos_profile());
+    rclcpp::QoS(200));
   unloadNode_srv_ = create_service<UnloadNode>(
     "~/_container/unload_node",
     std::bind(&ComponentManager::on_unload_node, this, _1, _2, _3),
-    rclcpp::QoS(200).get_rmw_qos_profile());
+    rclcpp::QoS(200));
   listNodes_srv_ = create_service<ListNodes>(
     "~/_container/list_nodes",
     std::bind(&ComponentManager::on_list_nodes, this, _1, _2, _3));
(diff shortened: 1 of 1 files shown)


rclcpp_components/src/component_manager.cpp
```diff
@@ -39,10 +39,12 @@ ComponentManager::ComponentManager(
 {
   loadNode_srv_ = create_service<LoadNode>(
     "~/_container/load_node",
-    std::bind(&ComponentManager::on_load_node, this, _1, _2, _3));
+    std::bind(&ComponentManager::on_load_node, this, _1, _2, _3),
+    rclcpp::QoS(200));
```
esteve, line 43: Should this be instead `rclcpp::ServiceQoS().keep_last(200)` ?

xmfcx: I think they are both same:

https://github.com/ros2/rclcpp/blob/0f331f90a99da40f7b7a69b4f55b30b7245d294f/rclcpp/include/rclcpp/qos.hpp#L137-L146

I could also do: `rclcpp::QoS(rclcpp::KeepLast(history_depth))`, is there any preference for using the more explicit calls?

esteve: If the defaults for services ever change (e.g. reliability, livelines, etc.), using `rclcpp::ServicesQoS` would avoid any issues in the future, so I'd prefer `rclcpp::ServicesQoS().keep_last(200)`

clalancette: Yeah, agreed with that.

xmfcx: Applied: https://github.com/ros2/rclcpp/pull/2363/commits/9bf1b0923e38561433c599661753232df4d8c4b2

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19761/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14263/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20487/)

esteve (approved):
@xmfcx thanks for addressing our feedback

esteve:
@clalancette would it be possible to backport this change to Humble and Iron? It shouldn't change existing behavior.

clalancette:
> @clalancette would it be possible to backport this change to Humble and Iron? It shouldn't change existing behavior.

We can consider doing this, yes.  However, I would like to let it sit in Rolling a little bit before we do that.

PR #2373, opened 2023-11-21
[WIP] Adding template for thread priorization in components at run-time

SteveMacenski:
Howdy,

Per a conversation with @JanStaschulat at ROSCon / in Nav2 https://github.com/ros-planning/navigation2/issues/3974: I think it would be dandy to be able to set increase thread prioritizations for soft-realtime in the Composable Node containers. Both to be able to set at launch-time instead of setting on a per-node basis in a stack. In Nav2, there are a couple of nodes we have set able to do this recently, but it would be great if there was a generic way that users could select prioritization more broadly and easily at run-time. 

This PR implements a prototype of this behavior that I would like you feedback on to resolve necessary problems seen. I added a `setSoftRealTimePriority` to the ComponentManager that is used to set the prioritization when necessary. By default, I have the priority set to 49 after discussions with Bosch, but we could also make this configurable. Though, there's something to be said about simplicity and the suggestions in the RT workshop had the options of HIGH and LOW, so I think this might be OK to start off with. 

The way that I see this workflow occurring is the following for **Isolated Component Manager**:
- Users set in the `parameters` field of [`LoadComposableNodes`](https://github.com/ros2/launch_ros/blob/rolling/launch_ros/launch_ros/actions/load_composable_nodes.py#L301-L308) or from the parameters in [`CompositionNode`](https://github.com/ros2/launch_ros/blob/rolling/launch_ros/launch_ros/descriptions/composable_node.py#L167-L169) `use_soft_realtime_priority` to its value
- That is communicated in the `parameters` field of `LoadNode` service to the component container and sent to `override_parameters` in `NodeOptions`
- When we create the thread for our executor, we check the parameter overrides for this value. If it exists & is `true` then we set the prioritization 
- This way, we can control which nodes are added to the higher priority threads and which are not individually

Q: I believe all of the launch stuff (linked above) is setup already to be able to handle this workflow. Please I would appreciate making sure that's true since the launch_ros package is a little difficult to parse at times. 

The way that I see this workflow occurring is the following for **Normal Component Manager**:
- We declare `use_soft_realtime_priority` in the Component Manager constructor (`false`, for off default) 
- If true, we also set the priority
- This way, we can control the single threaded & multithreaded containers to have thread prioritization as well for its nodes
- Technically speaking, you could also set this to `true` in the Isolated case to increase the prioritization of the Isolated's load/unload/list services. While that's an odd choice, that's a perfectly valid option 

I have no questions on this one, I think this is good.


Zard-C:
hi @SteveMacenski, glad to see this draft, I think that `sched_setscheduler` api is only for linux system, but for MacOS or Windows, they may have their own apis to set thread priority. Thus another 2 platforms should be taking considered.
I think maybe we could visit the std::thread via the wrapper, and use `std::thread::native_handle()`  to get the native thread handle(which depends on platforms), and use specific api to set thread's real-time attributes.
I am currently working on real time tuning on linux and qnx, If you have better ideas, feel free to let me know. 😸

JanStaschulat:
> hi @SteveMacenski, glad to see this draft, I think that `sched_setscheduler` api is only for linux system, but for MacOS or Windows, they may have their own apis to set thread priority. Thus another 2 platforms should be taking considered.
> 
This [utility function](https://github.com/ros2/examples/blob/1d97c4fc7445554f6f85f63305d424fc017212a0/rclcpp/executors/cbg_executor/src/examples_rclcpp_cbg_executor/utilities.hpp#L73) supports thread configuration for Linux, Windows Mac. A similar version could be used here as well.

Zard-C:
> > hi @SteveMacenski, glad to see this draft, I think that `sched_setscheduler` api is only for linux system, but for MacOS or Windows, they may have their own apis to set thread priority. Thus another 2 platforms should be taking considered.
> 
> This [utility function](https://github.com/ros2/examples/blob/1d97c4fc7445554f6f85f63305d424fc017212a0/rclcpp/executors/cbg_executor/src/examples_rclcpp_cbg_executor/utilities.hpp#L73) supports thread configuration for Linux, Windows Mac. A similar version could be used here as well.

I took a look at this function, it supports QNX as well, very impressive 👍

SteveMacenski:
> This [utility function](https://github.com/ros2/examples/blob/1d97c4fc7445554f6f85f63305d424fc017212a0/rclcpp/executors/cbg_executor/src/examples_rclcpp_cbg_executor/utilities.hpp#L73) supports thread configuration for Linux, Windows Mac. A similar version could be used here as well.

Agreed for the Isolated case. How do we handle this for the normal case where this is simply called while inside of a thread, not in creating one (e.g. we call `setSoftRealTimePriority` in the `ComponentManager` constructor)? 

If that function can handle that and/or be made to handle that, then happy to try to find a home for it as part of this effort.

Zard-C:
> Agreed for the Isolated case. How do we handle this for the normal case where this is simply called while inside of a thread, not in creating one (e.g. we call `setSoftRealTimePriority` in the `ComponentManager` constructor)?
> 
On my thought,  the way that  to get the native handle of current thread, is also depends on platform:
For linux:

```cpp
#include <pthread.h>

void threadFunction() {
    pthread_t nativeHandle = pthread_self();
    // use nativeHandle to set attributes ...
}

```
For windows:

```cpp
#include <windows.h>

void threadFunction() {
    HANDLE pseudoHandle = GetCurrentThread();
    HANDLE actualHandle;
    DuplicateHandle(GetCurrentProcess(), pseudoHandle, GetCurrentProcess(), &actualHandle, 0, FALSE, DUPLICATE_SAME_ACCESS);
    // use the acutalHandle;
}
```

rclcpp_components/include/rclcpp_components/component_manager.hpp
```diff
@@ -256,6 +256,20 @@ class ComponentManager : public rclcpp::Node
     on_list_nodes(request_header, request, response);
   }
 
+  /// Set a manager's executor thread priority level, can throw if priority is not able to be set
+  void setSoftRealTimePriority()
+  {
+    sched_param sch;
+    sch.sched_priority = 49;
+    if (sched_setscheduler(0, SCHED_FIFO, &sch) == -1) {
```
Zard-C, line 264: I thought that calling this would increase the process's priority and change scheduling strategy, do we need to set all thread's attribute or just the thread which calls this function?

JanStaschulat: The parameter `0` in `sched_setscheduler` means take the current thread. Because `setSoftRealTimePriority` is called in the newly created thread, this function will only set the priority of the newly created thread.

Zard-C: > Agreed for the Isolated case. How do we handle this for the normal case where this is simply called while inside of a thread, not in creating one (e.g. we call `setSoftRealTimePriority` in the `ComponentManager` constructor)?
> 
On my thought, the way that to set the current thread's attrbutes takes 2 steps:

1. get the native handle of the current thread
2. set the priority and scheduling strategy of current thread(with native handle)

both needs platform-specific apis.

JanStaschulat:
> > This [utility function](https://github.com/ros2/examples/blob/1d97c4fc7445554f6f85f63305d424fc017212a0/rclcpp/executors/cbg_executor/src/examples_rclcpp_cbg_executor/utilities.hpp#L73) supports thread configuration for Linux, Windows Mac. A similar version could be used here as well.
> 
> Agreed for the Isolated case. How do we handle this for the normal case where this is simply called while inside of a thread, not in creating one (e.g. we call `setSoftRealTimePriority` in the `ComponentManager` constructor)?
> 
> If that function can handle that and/or be made to handle that, then happy to try to find a home for it as part of this effort.

[`int sched_setscheduler(pid_t pid, int policy, const struct sched_param *param);`](https://man7.org/linux/man-pages/man2/sched_setscheduler.2.html)
If pid equals zero, the scheduling policy and parameters of the calling thread will be set.

This is, however, not documented for [`pthread_setschedparam`](https://man7.org/linux/man-pages/man3/pthread_setschedparam.3.html).

rclcpp_components/include/rclcpp_components/component_manager.hpp
Zard-C: Thanks @JanStaschulat , you are right.
https://man7.org/linux/man-pages/man2/sched_setscheduler.2.html

SteveMacenski:
Sure, but your method takes in a `T native_handle` that needs to implement a number of things to a number of `OS`'s. Can you just send that as `0`, I wouldn't have thought so.

Zard-C:
> Sure, but your method takes in a `T native_handle` that needs to implement a number of things to a number of `OS`'s. Can you just send that as `0`, I wouldn't have thought so.

For windows, change 
[windows](https://github.com/ros2/examples/blob/1d97c4fc7445554f6f85f63305d424fc017212a0/rclcpp/executors/cbg_executor/src/examples_rclcpp_cbg_executor/utilities.hpp#L77)'s `native_handler to [GetCurrentThread()](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentthread)

For Mac, QNX, and linux,
change `native_handle` to `pthread_self()`
[pthread_self(macos)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/pthread_self.3.html)
[pthread_self(qnx)](https://www.qnx.com/developers/docs/6.3.0SP3/neutrino/lib_ref/p/pthread_self.html#id7)
[pthread_self(linux)](https://man7.org/linux/man-pages/man3/pthread_self.3.html)

Thus we don't need take `T handler` as parameter

PR #2374, opened 2023-11-22
Switch to target_link_libraries.

clalancette:
That way we can hide more of the implementation by using the PRIVATE keyword.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19866/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14354/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20577/)

rclcpp_components/cmake/rclcpp_components_register_node.cmake
```diff
@@ -70,10 +70,11 @@ macro(rclcpp_components_register_node target)
...
     INPUT ${PROJECT_BINARY_DIR}/rclcpp_components/node_main_configured_${node}.cpp.in)
   add_executable(${node} ${PROJECT_BINARY_DIR}/rclcpp_components/node_main_${node}.cpp)
-  ament_target_dependencies(${node}
-    "rclcpp"
-    "class_loader"
-    "rclcpp_components")
+  target_link_libraries(${node}
+    class_loader::class_loader
+    rclcpp::rclcpp
+    rclcpp_components::component
```
mjcarroll, line 76: I think that `component` is recursively dependent on these other ones.  I imagine that you aren't looking to audit all of the dependencies in this PR, but maybe something that we should check/address in the future.

clalancette: While that's true, it is also the case that the node_main.cpp also directly depends on all of these dependencies:

https://github.com/ros2/rclcpp/blob/d08d9cf5ffd5ecb80c1ef4085573c411eecbd737/rclcpp_components/src/node_main.cpp.in#L19-L22

So I think having all of these dependencies here is appropriate.

mjcarroll (approved):

PR #2375, opened 2023-11-28
[events executor] - Fix Behavior with Timer Cancel

mwcondino:
## Background
It appears that with the experimental events executor, some usages of the rclcpp API cause broken behavior. Specifically, if multiple timers are configured in a node, and during runtime one of the timers is cancelled via the `TimerBase::cancel` method, this will result in all other timers failing to execute.

A minimal working example of the bug is found here --
[example_package.tar.gz](https://github.com/ros2/rclcpp/files/13494070/example_package.tar.gz)

To reproduce, build the package using `colcon build`, then run `USE_EVENTS_EXECUTOR=1 ros2 run example_package timer_node`. It is observable that both timers get cancelled after the first timer cancels itself:
```
root@753c2035d766:/app/src# USE_EVENTS_EXECUTOR=1 ros2 run example_package timer_node
[INFO] [1701211973.238745755] [timer_node]: Using EventsExecutor
[INFO] [1701211974.239147118] [timer_node]: Timer 1!
[INFO] [1701211974.239268496] [timer_node]: Timer 2!
[INFO] [1701211975.239086741] [timer_node]: Timer 1!
[INFO] [1701211975.239251090] [timer_node]: Timer 2!
[INFO] [1701211976.239241288] [timer_node]: Timer 1!
[INFO] [1701211976.239417539] [timer_node]: Timer 2!
[INFO] [1701211977.239107783] [timer_node]: Timer 1!
[INFO] [1701211977.239260069] [timer_node]: Timer 2!
[INFO] [1701211978.239079416] [timer_node]: Timer 1!
[INFO] [1701211978.239232915] [timer_node]: Timer 2!
[INFO] [1701211979.239045079] [timer_node]: Timer 1!
[INFO] [1701211979.239198617] [timer_node]: Timer cancelling itself!
[INFO] [1701211979.239225057] [timer_node]: Timer 2!
```

Myself and @gusbrigantino  have tested this on latest `rolling` as well as `iron`.

## Solution
The root cause is that there’s an edgecase within the `TimersManager` class which causes the thread to hang indefinitely if the ‘head’ timer in an internal data structure gets cancelled. As a result, the method will block until timers are reset (the `timers_updated_` flag is set, which only occurs if the timers are reset. The underlying issue is that the `TimersManager` class is not properly checking for if a timer to be executed has actually been cancelled.

The implemented fix adds logic to the comparison used for the heap such that cancelled timers are always at the bottom of the heap. Furthermore,  `get_head_timeout_unsafe` is modified to return whether or not the current timer is invalid. This bool is then used to break out of the loop in `run_timers`.

### DIFF
--- a/rclcpp/include/rclcpp/experimental/timers_manager.hpp
+++ b/rclcpp/include/rclcpp/experimental/timers_manager.hpp
@@ -495,14 +495,7 @@ private:
     static bool timer_greater(TimerPtr a, TimerPtr b)
     {
       // TODO(alsora): this can cause an error if timers are using different clocks
-      bool both_canceled_or_active = a->is_canceled() == b->is_canceled();
-      if (both_canceled_or_active) {
-        return a->time_until_trigger() > b->time_until_trigger();
-      }
-      else if (a->is_canceled()) {
-        return true;
-      }
-      return false;
+      return a->time_until_trigger() > b->time_until_trigger();
     }
 
     std::vector<TimerPtr> owned_heap_;
--- a/rclcpp/include/rclcpp/experimental/timers_manager.hpp
+++ b/rclcpp/include/rclcpp/experimental/timers_manager.hpp
@@ -22,6 +22,7 @@
 #include <functional>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <thread>
 #include <utility>
 #include <vector>
@@ -171,13 +172,14 @@ public:
    * @brief Get the amount of time before the next timer triggers.
    * This function is thread safe.
    *
-   * @return std::chrono::nanoseconds to wait,
+   * @return std::optional<std::chrono::nanoseconds> to wait,
    * the returned value could be negative if the timer is already expired
    * or std::chrono::nanoseconds::max() if there are no timers stored in the object.
+   * If the head timer was cancelled, then this will return a nullopt.
    * @throws std::runtime_error if the timers thread was already running.
    */
   RCLCPP_PUBLIC
-  std::chrono::nanoseconds get_head_timeout();
+  std::optional<std::chrono::nanoseconds> get_head_timeout();
 
 private:
   RCLCPP_DISABLE_COPY(TimersManager)
@@ -268,6 +270,15 @@ public:
     bool empty() const
     {
       return weak_heap_.empty();
+    }
+
+    /**
+     * @brief Returns the size of the heap.
+     * @return the number of valid timers in the heap.
+     */
+    size_t size() const
+    {
+      return weak_heap_.size();
     }
 
     /**
@@ -511,12 +522,13 @@ private:
    * @brief Get the amount of time before the next timer triggers.
    * This function is not thread safe, acquire a mutex before calling it.
    *
-   * @return std::chrono::nanoseconds to wait,
+   * @return std::optional<std::chrono::nanoseconds> to wait,
    * the returned value could be negative if the timer is already expired
    * or std::chrono::nanoseconds::max() if the heap is empty.
+   * If the head timer was cancelled, then this will return a nullopt.
    * This function is not thread safe, acquire the timers_mutex_ before calling it.
    */
-  std::chrono::nanoseconds get_head_timeout_unsafe(bool& head_was_canceled);
+  std::optional<std::chrono::nanoseconds> get_head_timeout_unsafe();
 
   /**
    * @brief Executes all the timers currently ready when the function is invoked
--- a/rclcpp/include/rclcpp/experimental/timers_manager.hpp
+++ b/rclcpp/include/rclcpp/experimental/timers_manager.hpp
@@ -273,15 +273,6 @@ public:
     }
 
     /**
-     * @brief Returns the size of the heap.
-     * @return the number of valid timers in the heap.
-     */
-    size_t size() const
-    {
-      return weak_heap_.size();
-    }
-
-    /**
      * @brief This function restores the current object as a valid heap
      * and it returns a locked version of it.
      * Timers that went out of scope are removed from the container.
--- a/rclcpp/include/rclcpp/experimental/timers_manager.hpp
+++ b/rclcpp/include/rclcpp/experimental/timers_manager.hpp
@@ -25,7 +25,6 @@
 #include <thread>
 #include <utility>
 #include <vector>
-
 #include "rclcpp/context.hpp"
 #include "rclcpp/timer.hpp"
 
@@ -496,7 +495,14 @@ private:
     static bool timer_greater(TimerPtr a, TimerPtr b)
     {
       // TODO(alsora): this can cause an error if timers are using different clocks
-      return a->time_until_trigger() > b->time_until_trigger();
+      bool both_canceled_or_active = a->is_canceled() == b->is_canceled();
+      if (both_canceled_or_active) {
+        return a->time_until_trigger() > b->time_until_trigger();
+      }
+      else if (a->is_canceled()) {
+        return true;
+      }
+      return false;
     }
 
     std::vector<TimerPtr> owned_heap_;
@@ -517,7 +523,7 @@ private:
    * or std::chrono::nanoseconds::max() if the heap is empty.
    * This function is not thread safe, acquire the timers_mutex_ before calling it.
    */
-  std::chrono::nanoseconds get_head_timeout_unsafe();
+  std::chrono::nanoseconds get_head_timeout_unsafe(bool& head_was_canceled);
 
   /**
    * @brief Executes all the timers currently ready when the function is invoked
--- a/rclcpp/src/rclcpp/experimental/timers_manager.cpp
+++ b/rclcpp/src/rclcpp/experimental/timers_manager.cpp
@@ -268,11 +268,10 @@ void TimersManager::run_timers()
       TimersHeap locked_heap = weak_timers_heap_.validate_and_lock();
       locked_heap.heapify();
       weak_timers_heap_.store(locked_heap);
-    }
-    else if (time_to_sleep.value() != std::chrono::nanoseconds::zero()) {
-        // If time_to_sleep is zero, we immediately execute. Otherwise, wait
-        // until timeout or notification that timers have been updated
-        timers_cv_.wait_for(lock, time_to_sleep.value(), [this]() {return timers_updated_;});
+    } else if (time_to_sleep.value() != std::chrono::nanoseconds::zero()) {
+      // If time_to_sleep is zero, we immediately execute. Otherwise, wait
+      // until timeout or notification that timers have been updated
+      timers_cv_.wait_for(lock, time_to_sleep.value(), [this]() {return timers_updated_;});
     }
 
     // Reset timers updated flag
--- a/rclcpp/src/rclcpp/experimental/timers_manager.cpp
+++ b/rclcpp/src/rclcpp/experimental/timers_manager.cpp
@@ -291,6 +291,12 @@ void TimersManager::run_timers()
       else {
         // Wait until notification that timers have been updated
         timers_cv_.wait(lock, [this]() {return timers_updated_;});
+
+        // Re-heap in case ordering changed due to a cancelled timer
+        // re-activating.
+        TimersHeap locked_heap = weak_timers_heap_.validate_and_lock();
+        locked_heap.heapify();
+        weak_timers_heap_.store(locked_heap);
       }
     }
 
--- a/rclcpp/src/rclcpp/experimental/timers_manager.cpp
+++ b/rclcpp/src/rclcpp/experimental/timers_manager.cpp
@@ -246,28 +246,48 @@ void TimersManager::run_timers()
 
     std::optional<std::chrono::nanoseconds> time_to_sleep = get_head_timeout_unsafe();
 
-    if (!time_to_sleep.has_value()) {
-        // If head was cancelled, and there is more than 1 timer, we need to re-heapify. This
-        // is a case where the next up timer should really be used, so the next loop iteration
-        // will pick up the correct head.
-        if (weak_timers_heap_.size() > 1) {
-          TimersHeap locked_heap = weak_timers_heap_.validate_and_lock();
-          locked_heap.heapify();
-          weak_timers_heap_.store(locked_heap);
+    // Head timer is not cancelled
+    if (time_to_sleep.has_value()) {
+      // No need to wait if a timer is already available
+      if (time_to_sleep.value() > std::chrono::nanoseconds::zero()) {
+        // Heap is not empty
+        if (time_to_sleep.value() != std::chrono::nanoseconds::max()) {
+          // Check if head timer is in cancelled_timers_
+          TimerPtr head_timer = weak_timers_heap_.front().lock();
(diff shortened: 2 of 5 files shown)


mwcondino:
Looks like the build for `rolling` may be unhappy? I'm seeing a similar build failure for the latest merge into `rolling` - [link](https://github.com/ros2/rclcpp/pull/2374/commits)

mjcarroll:
> Looks like the build for rolling may be unhappy? I'm seeing a similar build failure for the latest merge into rolling - [link](https://github.com/ros2/rclcpp/pull/2374/commits)

You can probably disregard `Rpr` jobs right now, we will run full CI on it.

mjcarroll:
Is there any way that you can add your reproducible example as a test?

mwcondino:
> Is there any way that you can add your reproducible example as a test?

Sure, I can try to get it in the form of a unit test.

rclcpp/src/rclcpp/experimental/timers_manager.cpp
```diff
@@ -191,6 +192,7 @@ std::chrono::nanoseconds TimersManager::get_head_timeout_unsafe()
     }
     head_timer = locked_heap.front();
   }
+  head_was_cancelled = head_timer->is_canceled();
```
alsora, line 195: do we need an extra variable? what does `time_until_trigger` returns when the timer is cancelled?

rclcpp/include/rclcpp/experimental/timers_manager.hpp
```diff
@@ -496,7 +495,14 @@ class TimersManager
     static bool timer_greater(TimerPtr a, TimerPtr b)
     {
       // TODO(alsora): this can cause an error if timers are using different clocks
-      return a->time_until_trigger() > b->time_until_trigger();
```
alsora, line 499: I see, this seems a bug I introduced when porting the code to rclcpp.
in our internal version, we added a new function `next_call_time` which takes into account cancellations
https://github.com/irobot-ros/rclcpp/blob/galactic-events-executor/rclcpp/src/rclcpp/timer.cpp#L145-L164

rclcpp/src/rclcpp/experimental/timers_manager.cpp
mwcondino: Yes, I found it was necessary. `time_until_trigger` actually returns `std::chrono::nanoseconds::max()` when the timer is cancelled, which I think is how the bug manifests -- that then forces the call to `timers_cv_.wait(...)`

rclcpp/include/rclcpp/experimental/timers_manager.hpp
mwcondino: I think even if `next_call_time` was used, the bug would still manifest -- that would cause a situation where we wait on `timers_cv_.wait(lock, [this]() {return timers_updated_;});`, which won't return until a timer is manually reset.

An alternative fix that we considered was modifiying `rcl` and the timer object to support a cancellation callback, which would then be used similarly to how `set_on_reset_callback` is used in `TimersManager`. If we could trigger a callback upon the timer being reset, then I think that would allow us to break the condition and then re-order the heap, likely fixing the issue.

alsora: Ok, good point.
I think we should go ahead with the current fix for now. Adding a new callback for timers cancelled is doable, but it's a larger effort.
Any thoughts @mauropasse ?

mwcondino:
OK I ended up adding two tests:
1. `check_one_timer_cancel_doesnt_affect_other_timers` in `test_timers_manager.cpp` validates the behavior in the timers_manager object. Interestingly enough, this test **does not fail** on latest `iron`, which I did not expect given the example package's behavior
2. `testOneTimerCancelledWithExecutorSpin` in `test_executors.cpp` validates the behavior when using a `rclcpp::Node` that adds and then cancels it's own timers. This example uses `rclcpp::spin`, and **does** reproduce the behavior. On this branch, this test should be passing.

I can't figure out why there would be a difference in behavior between the two tests - as far as I can tell, when `spin` is called on the `EventsExecutor`, it makes a call to `start` it's own `timers_manager` object, so I would think that the behavior should be the same as when the `timers_manager` object is manually started and then the same sequence of waiting + cancelling is followed.

rclcpp/include/rclcpp/experimental/timers_manager.hpp
gusbrigantino: We ended up reverting this diff here in the `timer_greater` function because we discovered it was no longer necessary for the bug fix

rclcpp/src/rclcpp/experimental/timers_manager.cpp
alsora: It looks like the `get_head_timeout_unsafe` is essentially trying to indicate 3 different scenarios:
 - it should indicate when the heap is empty, and it currently does that by returning `std::chrono::nanoseconds::max()`
 - it should indicate when the front timer in the heap is cancelled, and it currently does that by setting the `head_was_cancelled` bool variable
 - if the front timer is good, then it should indicate the time until trigger

I think that this behavior is not very intuitive.

Moreover, what happens if there's only 1 timer which is cancelled? Will the heap keep polling it?

rclcpp/src/rclcpp/experimental/timers_manager.cpp
```diff
@@ -242,14 +244,22 @@ void TimersManager::run_timers()
...
+    std::chrono::nanoseconds time_to_sleep = get_head_timeout_unsafe(head_was_cancelled);
 
     // No need to wait if a timer is already available
     if (time_to_sleep > std::chrono::nanoseconds::zero()) {
       if (time_to_sleep != std::chrono::nanoseconds::max()) {
         // Wait until timeout or notification that timers have been updated
         timers_cv_.wait_for(lock, time_to_sleep, [this]() {return timers_updated_;});
-      } else {
+      }
+      else if (head_was_cancelled) {
```
alsora, line 256: What happens if there's only 1 timer which is cancelled? Will the heap keep entering here at every iteration, without ever sleeping?

rclcpp/src/rclcpp/experimental/timers_manager.cpp
mwcondino: Great point! Refactored to use `std::optional` as the return value.

rclcpp/src/rclcpp/experimental/timers_manager.cpp
alsora: I don't think that the fix works.
What if there are more than 1 timers but they are all cancelled?

I think that the correct approach would likely require to:
a) remove cancelled timers from the heap
b) have a notification to put them back in when they are re-activated

mwcondino: @alsora sorry for the delay here - had a busy couple of weeks.

@gusbrigantino and I refactored the code, and we came up with an implementation of `run_timers` that does not require storing cancelled timers and removing them from the heap. Rather, the approach is to try to re-heap if the head timer was cancelled, then re-organize some logic from that point on. If there are non-cancelled timers in the heap, then the re-heap will force them to be at the front and thus head is now valid and the logic remains the same. However, if all timers in the heap are cancelled, the behavior now matches the previous behavior with no timers in the heap -- we wait until `timers_updated` is `true`.

We added a few more test cases to validate a couple of different scenarios.

Please let us know if that makes sense and works for you!

rclcpp/include/rclcpp/experimental/timers_manager.hpp
```diff
@@ -271,6 +272,15 @@ class TimersManager
...
 
+    /**
+     * @brief Returns the size of the heap.
+     * @return the number of valid timers in the heap.
+     */
+    size_t size() const
+    {
+      return weak_heap_.size();
+    }
+
```
fujitatomoya, line 283: this method is not used currently, so should be removed from this PR?

mwcondino: Good call, removed

mwcondino:
@alsora @mjcarroll could someone please review whenever convenient? Thanks!

fujitatomoya:
@mwcondino FYI, besides unresolved comments, there are several failures you can find https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/1272/testReport/

mwcondino:
> @mwcondino FYI, besides unresolved comments, there are several failures you can find https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/1272/testReport/

Thank

> @mwcondino FYI, besides unresolved comments, there are several failures you can find https://build.ros2.org/job/Rpr__rclcpp__ubuntu_jammy_amd64/1272/testReport/

Thank you, appreciate the heads up! I am new to contributing to ROS repos, so I was unsure of which failures were important (I think the main build is still messed up, but that seems unrelated to this PR).

Linting should be fixed now.

rclcpp/src/rclcpp/experimental/timers_manager.cpp
```diff
@@ -242,17 +244,34 @@ void TimersManager::run_timers()
...
+    std::optional<std::chrono::nanoseconds> time_to_sleep = get_head_timeout_unsafe();
+
+    // If head timer was cancelled, try to reheap and get a new head.
+    // This avoids an edge condition where head timer is cancelled, but other
+    // valid timers remain in the heap.
+    if (!time_to_sleep.has_value()) {
+      // Re-heap to (possibly) move cancelled timer from head of heap. If
+      // entire heap is cancelled, this will still result in a nullopt.
+      TimersHeap locked_heap = weak_timers_heap_.validate_and_lock();
+      locked_heap.heapify();
```
alsora, line 256: The queue is sorted according to the `return a->time_until_trigger() > b->time_until_trigger();` comparison.
What does `time_until_trigger` returns for a cancelled timer?

alsora:
CI run 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20074/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14523/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20793/)

rclcpp/src/rclcpp/experimental/timers_manager.cpp
mwcondino: Upon a cancelled timer, `time_until_trigger` returns `std::chrono::nanoseconds::max()`, so resorting should force cancelled timers to be at the bottom of the heap. In a previous iteration of the PR, we had updated the comparison function for the heap to explicitly check for canceled timers and ensure they were on the bottom of the heap, but that was removed since it's obsolete if `std::chrono::nanoseconds::max()` is always returned for a cancelled timer.

alsora (approved):
The changes look good to me.
I would like @mauropasse to take a look as well before merging it.

mauropasse:
The fix and unit tests look good to me!

fujitatomoya:
windows are failing https://ci.ros2.org/job/ci_windows/20793/, i think that is because of timer accuracy. (e.g https://ci.ros2.org/job/ci_windows/20793/testReport/rclcpp/TestTimerCancelBehavior_SingleThreadedExecutor/testHeadTimerCancelThenResetBehavior/ timer does not get called as expected.)

mwcondino:
> windows are failing https://ci.ros2.org/job/ci_windows/20793/, i think that is because of timer accuracy. (e.g https://ci.ros2.org/job/ci_windows/20793/testReport/rclcpp/TestTimerCancelBehavior_SingleThreadedExecutor/testHeadTimerCancelThenResetBehavior/ timer does not get called as expected.)

Thanks for the callout -  I relaxed the timing constraints, hopefully that will fix things

clalancette:
> Thanks for the callout - I relaxed the timing constraints, hopefully that will fix things

They should be a *lot* more relaxed than that, otherwise we are going to end up with flaky tests.

In general, on Windows, you should expect things to take 4x as long as on Linux.  We shouldn't even consider having timing tests with constraints of less than 200ms or so.

alsora:
New build after changes in unit-tests:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20089/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14536/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20809/)


It's probably for a separate discussion, but I really feel unease with having so many unit-tests using wall-time clock.
IMO unit-tests should always use a simulated time, to be robust to latency that happens in the server.

clalancette:
> It's probably for a separate discussion, but I really feel unease with having so many unit-tests using wall-time clock.

I agree to an extent.  We probably should have more tests that use some kind of simulated time.  The main reason we don't is that those are harder tests to write.

But I also think that we are always going to have to have tests that use system time.  They show problems that a simulated time won't, and they are what most of our users are using.

mwcondino:
Updated the test to use a simulated clock. For now, that is at .25 realtime, but we can update as necessary if the tests are still flaky and require a slower rate.

mwcondino:
@alsora could you please trigger a full CI run whenever convenient?

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20129/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14568/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20841/)

mwcondino:
Looks like there's still some windows failures, but they appear unrelated to this PR (unless I'm missing something):
[projectroot.test_n_nodes__rmw_fastrtps_cpp](https://ci.ros2.org/job/ci_windows/20841/testReport/junit/(root)/projectroot/test_n_nodes__rmw_fastrtps_cpp/)
[test_rclcpp.TestNNodes.test_10_nodes](https://ci.ros2.org/job/ci_windows/20841/testReport/junit/test_rclcpp/TestNNodes/test_10_nodes/)
[test_rclcpp.TestNNodesAfterShutdown.test_10_nodes](https://ci.ros2.org/job/ci_windows/20841/testReport/junit/test_rclcpp/TestNNodesAfterShutdown/test_10_nodes/)

clalancette:
> Looks like there's still some windows failures, but they appear unrelated to this PR (unless I'm missing something):

While I generally agree with you (tests will only be using the events executor if they are explicitly configured to do so), it is kind of weird that these are failing.  We don't see those failures in the nightlies: https://ci.ros2.org/view/nightly/job/nightly_win_rel/ .

I think we should re-run Windows, and if it comes back green this time we can go ahead with it.

mwcondino:
> I think we should re-run Windows, and if it comes back green this time we can go ahead with it.

Sounds good - is it possible for me to kick off the jobs? I feel bad bugging folks to start the CI, but I can't see any buttons on https://ci.ros2.org/job/ci_windows/ to start things, and I'm wondering if I'm missing anything (admittedly I have not used Jenkins in the past so this could definiely be a noob question)

alsora:
Currently only maintainers can trigger a job, and don't hesitate to ping if we forget to do it!

Windowns-only job:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20889/)

alsora:
Windows CI is green, indicating that the previous failure is unrelated.
Merging this PR

PR #2376, opened 2023-11-30
[fix] Init EventHandlerBase::wait_set_event_index_

aosmw:
This prevents a call to is_ready() using an uninitalised index to access wait_set->events that may occur before the event has been setup in add_to_wait_set().

### DIFF
--- a/rclcpp/include/rclcpp/event_handler.hpp
+++ b/rclcpp/include/rclcpp/event_handler.hpp
@@ -230,7 +230,7 @@ protected:
   std::function<void(size_t)> on_new_event_callback_{nullptr};
 
   rcl_event_t event_handle_;
-  size_t wait_set_event_index_;
+  size_t wait_set_event_index_ = 0;
 };
 
 using QOSEventHandlerBase [[deprecated("Use rclcpp::EventHandlerBase")]] = EventHandlerBase;


aosmw:
I ran into this while developing a rclcpp lifecycle node with a number of services, publishers and subscribers with ASAN enabled.  
It triggers a segfault when I use `ros2 lifecycle /mynode set configure`  

I have made the change for rolling, but will also backport it to humble where I am really using it.

In humble the change will be performed in QOSEventBase which is in rclcpp/src/rclcpp/qos_event.cpp

A more thorough validation could be done in the is_ready function to ensure that the pointer to `wait_set != nullptr && wait_set_event_index_ < wait_set->size_of_events`

rclcpp/include/rclcpp/event_handler.hpp
```diff
@@ -230,7 +230,7 @@ class EventHandlerBase : public Waitable
   std::function<void(size_t)> on_new_event_callback_{nullptr};
 
   rcl_event_t event_handle_;
-  size_t wait_set_event_index_;
+  size_t wait_set_event_index_ = 0;
```
clalancette, line 233: I'm honestly not sure about this.  While this definitely makes the index a known value (rather than something random), the only place that this is used is to [dereference the wait_set->events later](https://github.com/ros2/rclcpp/blob/26f6efa840cbac047a93dc21d055cfe9cde7270f/rclcpp/src/rclcpp/event_handler.cpp#L71).  Thus, a value of 0 here may also be wrong.

Can you come up with a reproducible example of what you are seeing?  I think that would help inform what we should do about this bug.

clalancette (changes requested):

rclcpp/include/rclcpp/event_handler.hpp
aosmw: So I guess this is another reason not to do work(Create publishers/subscribers) in callbacks, which I knew, but thought I might be able to get away with it... `rclcpp_lifecycle::on_configure()` being called rarely and me not too concerned about setup overhead during this time.

I was led into creating a publisher in a service callback, because the convenient interface to create a control_toolbox::PidROS requires a shared_ptr to a node.  To my knowledge you cannot use "shared_from_this()" in a constructor.  Can you?  So I could not do `m_pidAccelerator = std::make_shared<control_toolbox::PidROS>(shared_from_this(), "/dbw/accel");` in the constructor of my containing Lifecycle node.

I can however workaround the problem (of uninitialized memory that this PR or something like it would fix/hide)  
by using the more verbose form of control_toolbox::PidROS in my Lifecycle Node constructor.  ie
```
m_pidAccelerator = std::make_shared<control_toolbox::PidROS>(                                                                                                                                   
    this->get_node_base_interface(),                                                                                                                                                              
    this->get_node_logging_interface(),                                                                                                                                                           
    this->get_node_parameters_interface(),                                                                                                                                                        
    this->get_node_topics_interface(),                                                                                                                                                            
    "/dbw/accel",                                                                                                                                                                                 
    /*prefix_is_for_params*/ false); 
 ```

I currently use humble so am referring to QOSEventHandler which changes to just EventHandler in rolling.

I haven't yet been able to follow the point where the QOSEventHandlerBase is known about but not yet configured enough for is_ready() to work properly with the existing code.  

My hunch is that when constructed in a service callback, something in control_toolbox::PidROS is creating a QOSEventHandler and extending the executor's `entities_collector_->get_number_of_waitables()`  so that is_ready() is able to be called on it, BEFORE, QOSEventHandlerBase::add_to_wait_set is able to be called to initialise the wait_set_event_index_ properly.  

I think that is_ready() should return false in this case and not crash.
Initialising wait_set_event_index_ to a large sentinel (0xDEADBEEF ... how many events could their possibly be??)  and testing it for sanity in is_ready() would prevent this crash. 

Other alternatives might a be an explicit `bool added_to_wait_set` initialised to false that is set true after a successful QOSEventHandlerBase::add_to_wait_set

Its a bit of a head stretch to come up with a test to exercise this.  I will need to study the test infrastructure to see If I can do it.

Comment (aosmw) [reply rclcpp/include/rclcpp/event_handler.hpp:233]: Further to how I was led into it... for the searchers

I was hitting runtime terminations, during construction with the following message when running under gdb
```
terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
```

For others who hit this.  You CAN use the convenient interface to control_toolbox::PidROS to if you inherit from  std::enable_shared_from_this properly, which I wasn't.  

```
 class DbwNode :
     public rclcpp_lifecycle::LifecycleNode,
     public std::enable_shared_from_this<DbwNode>    <--------------- THIS is required
 {
 public:                                                                                                                                                                                          
    DbwNode() = delete;
    DbwNode(const std::string& name, const rclcpp::NodeOptions& options); 
    virtual ~DbwNode() = default;
...
private:
    std::shared_ptr<control_toolbox::PidROS> m_pidAccelerator;
}    

Then you can do this in the constructor.
m_pidAccelerator = std::make_shared<control_toolbox::PidROS>(shared_from_this(), "/dbw/accel");

clalancette: I think the best thing to move forward with this PR is to either have a simple reproducer that shows the problem, or to write a test that shows the problem and how this fix changes it.  Once we have that we'll have a better idea of exactly the problem, and how to fix it.  Thanks!

PR #2382, opened 2023-12-06
feat(MultiThreadedExecutor): Added ability to handle exceptions from threads

jmachowinski:
This commit adds external exception handling for the worker threads, allowing application code to implement custom exception handling.

### DIFF
--- a/rclcpp/include/rclcpp/executor.hpp
+++ b/rclcpp/include/rclcpp/executor.hpp
@@ -82,6 +82,18 @@ public:
   virtual void
   spin() = 0;
 
+  /**
+   * \sa rclcpp::Executor:spin() for more details
+   * \throws std::runtime_error when spin() called while already spinning
+   * @param exception_handler will be called for every exception in the processing threads
+   *
+   * The exception_handler can be called from multiple threads at the same time.
+   * The exception_handler shall rethrow the exception it if wants to terminate the program.
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  spin(const std::function<void(const std::exception & e)> & exception_handler) = 0;
+
   /// Add a callback group to an executor.
   /**
    * An executor can have zero or more callback groups which provide work during `spin` functions.
@@ -469,6 +481,18 @@ protected:
   RCLCPP_PUBLIC
   void
   execute_any_executable(AnyExecutable & any_exec);
+
+  /// Find the next available executable and do the work associated with it.
+  /**
+   * \param[in] any_exec Union structure that can hold any executable type (timer, subscription,
+   * service, client).
+   * \throws std::runtime_error if there is an issue triggering the guard condition
+   */
+  RCLCPP_PUBLIC
+  void
+  execute_any_executable(
+    AnyExecutable & any_exec,
+    const std::function<void(const std::exception & e)> & exception_handler);
 
   /// Run subscription executable.
   /**
--- a/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
@@ -69,6 +69,18 @@ public:
   void
   spin() override;
 
+  /**
+   * \sa rclcpp::Executor:spin() for more details
+   * \throws std::runtime_error when spin() called while already spinning
+   * @param exception_handler will be called for every exception in the processing threads
+   *
+   * The exception_handler can be called from multiple threads at the same time.
+   * The exception_handler shall rethrow the exception it if wants to terminate the program.
+   */
+  RCLCPP_PUBLIC
+  void
+  spin(const std::function<void(const std::exception & e)> & exception_handler) override;
+
   RCLCPP_PUBLIC
   size_t
   get_number_of_threads();
@@ -76,7 +88,9 @@ public:
 protected:
   RCLCPP_PUBLIC
   void
-  run(size_t this_thread_number);
+  run(
+    size_t this_thread_number,
+    const std::function<void(const std::exception & e)> & exception_handler);
 
 private:
   RCLCPP_DISABLE_COPY(MultiThreadedExecutor)
--- a/rclcpp/include/rclcpp/experimental/executors/events_executor/events_executor.hpp
+++ b/rclcpp/include/rclcpp/experimental/executors/events_executor/events_executor.hpp
@@ -92,6 +92,18 @@ public:
   void
   spin() override;
 
+  /**
+   * \sa rclcpp::Executor:spin() for more details
+   * \throws std::runtime_error when spin() called while already spinning
+   * @param exception_handler will be called for every exception in the processing threads
+   *
+   * The exception_handler can be called from multiple threads at the same time.
+   * The exception_handler shall rethrow the exception it if wants to terminate the program.
+   */
+  RCLCPP_PUBLIC
+  void
+  spin(const std::function<void(const std::exception & e)> & exception_handler) override;
+
   /// Events executor implementation of spin some
   /**
    * This non-blocking function will execute the timers and events
--- a/rclcpp/include/rclcpp/experimental/timers_manager.hpp
+++ b/rclcpp/include/rclcpp/experimental/timers_manager.hpp
@@ -125,9 +125,14 @@ public:
   /**
    * @brief Starts a thread that takes care of executing the timers stored in this object.
    * Function will throw an error if the timers thread was already running.
+   *
+   * @param exception_handler if valid, the execution of the timer will be done in a try catch block,
+   *                          and any occurring exception will be passed to the given handler
    */
   RCLCPP_PUBLIC
-  void start();
+  void start(
+    const std::function<void(const std::exception & e)> & exception_handler = std::function<void(
+      const std::exception & e)>());
 
   /**
    * @brief Stops the timers thread.
@@ -512,6 +517,11 @@ private:
   void run_timers();
 
   /**
+   * @brief calls run_timers with a try catch block.
+   */
+  void run_timers(const std::function<void(const std::exception & e)> & exception_handler);
+
+  /**
    * @brief Get the amount of time before the next timer triggers.
    * This function is not thread safe, acquire a mutex before calling it.
    *
@@ -528,7 +538,7 @@ private:
    * while keeping the heap correctly sorted.
    * This function is not thread safe, acquire the timers_mutex_ before calling it.
    */
-  void execute_ready_timers_unsafe();
+  void execute_ready_timers_unsafe(std::function<void(const std::exception & e)> exception_handler);
 
   // Callback to be called when timer is ready
   std::function<void(const rclcpp::TimerBase *,
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -420,6 +420,10 @@ Executor::execute_any_executable(
     return;
   }
 
+  assert(
+    (void("cannot execute an AnyExecutable without a valid callback group"),
+    any_exec.callback_group));
+
   if (any_exec.timer) {
     TRACETOOLS_TRACEPOINT(
       rclcpp_executor_execute,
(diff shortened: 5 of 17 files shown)


jmachowinski:
@wjwwood  @mjcarroll  @fujitatomoya Ping, I would like to see this merged for jazzy

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -69,14 +69,26 @@ class MultiThreadedExecutor : public rclcpp::Executor
...
+   * \sa rclcpp::Executor:spin() for more details
+   * \throws std::runtime_error when spin() called while already spinning
+   * @param exception_handler will be called for every exception in the processing threads
+   *
+   * The exception_handler can be called from multiple threads at the same time.
+   * The exception_handler shall rethrow the exception it if wants to terminate the program.
+   */
+  RCLCPP_PUBLIC
+  void
+  spin(std::function<void(const std::exception & e)> exception_handler);
```
fujitatomoya, line 82: why this is dedicated to `MultiThreadedExecutor`? i think user application can find useful with other Executors as well.

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
```diff
@@ -51,6 +51,13 @@ MultiThreadedExecutor::~MultiThreadedExecutor() {}
 
 void
 MultiThreadedExecutor::spin()
+{
+  spin(std::function<void(const std::exception &)>());
+}
+
+void
+
+MultiThreadedExecutor::spin(std::function<void(const std::exception & e)> exception_handler)
```
fujitatomoya, line 60: how about adding the test case for this to make sure the callback suppresses the exception?

fujitatomoya (commented):
@jmachowinski thanks for the PR. i think this is useful! a couple of minor comments. (if we merge this, i think it would be nice to update the demos/example aligned with this.)

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
alsora: In a single thread executor, the user is the one that is creating the thread where we spin, so they can take care of handling the exception without any new API.
This is not the case for the threads spawned internally by an executor, such as in this case

fujitatomoya: yeah i understand that, but user can use the same manner for any executor? i just though that would be useful to write the application.

jmachowinski: I don't care either way...

rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp
jmachowinski: Added test case

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
jmachowinski: @fujitatomoya @alsora can you come to an agreement, so that we can move forward with this PR ?

alsora: I'm ok with adding it to all executors.

jmachowinski: Added spin API with exception handler to Executor class

fujitatomoya (approved):
lgtm with green CI.

@alsora friendly ping.

rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
```diff
@@ -96,4 +96,48 @@ TEST_F(TestMultiThreadedExecutor, timer_over_take) {
   auto timer = node->create_wall_timer(PERIOD_MS, timer_callback, cbg);
   executor.add_node(node);
   executor.spin();
+
+}
+
+TEST_F(TestMultiThreadedExecutor, catch_exception) {
```
alsora, line 102: can this test be moved to the "common executors tests" ?

rclcpp/include/rclcpp/experimental/executors/events_executor/events_executor.hpp
```diff
@@ -92,6 +92,8 @@ class EventsExecutor : public rclcpp::Executor
   void
   spin() override;
 
+  using Executor::spin;
```
alsora, line 95: to be precise, the events executor should have a dedicated implementation.
This line here https://github.com/ros2/rclcpp/blob/070d71e2c68bd8009c7f96c3c110f2cb2390dde3/rclcpp/src/rclcpp/experimental/executors/events_executor/events_executor.cpp#L114 will start a new thread

jmachowinski: fair point...

rclcpp/test/rclcpp/executors/test_multi_threaded_executor.cpp
jmachowinski: I'll look into it...

jmachowinski: done

rclcpp/include/rclcpp/experimental/executors/events_executor/events_executor.hpp
jmachowinski: done

fujitatomoya:
@alsora can you check and resolve your comments above?

fujitatomoya:
@jmachowinski can you rebase to rolling? i would like to make sure CI is green with current HEAD especially at this release moment.

alsora:
Looks good to me

alsora (approved):

jmachowinski:
> @jmachowinski can you rebase to rolling? i would like to make sure CI is green with current HEAD especially at this release moment.
> 
rebased, note this currently fails the pipline, as it uncovered the bug reported in https://github.com/ros2/rclcpp/issues/2474

fujitatomoya:
@jmachowinski this is ready for CI with https://github.com/ros2/demos/pull/670? just chekcing.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20841/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15144/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21527/)

jmachowinski:
rebased to rolling

I don't know why the CI failed, it builds for me using this file:
https://github.com/jmachowinski/build_desc/raw/exeception_handling/ros2.repos

wjwwood:
I need to make some changes before merging this, but I cannot push to this branch, so I'll be opening a new pr. Please don't merge this one.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -82,6 +82,18 @@ class Executor
...
+   * \sa rclcpp::Executor:spin() for more details
+   * \throws std::runtime_error when spin() called while already spinning
+   * @param exception_handler will be called for every exception in the processing threads
+   *
+   * The exception_handler can be called from multiple threads at the same time.
+   * The exception_handler shall rethrow the exception it if wants to terminate the program.
+   */
+  RCLCPP_PUBLIC
+  virtual void
+  spin(const std::function<void(const std::exception & e)> & exception_handler) = 0;
```
wjwwood, line 95: I would have liked to have had a non-virtual `spin(... exception_handler = nullptr)`, that called a virtual and protected `spin_impl(... exception_handler) = 0`, so that we didn't need two signatures. It would be easier to document and wouldn't require every executor to implement two functions, one of which is pretty much always be boilerplate calling the other.

However, I don't know if I have time to do that before the freeze, so I won't block this on it, but I wanted to make the comment somewhere.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -82,6 +82,18 @@ class Executor
   virtual void
   spin() = 0;
 
+  /**
+   * \sa rclcpp::Executor:spin() for more details
+   * \throws std::runtime_error when spin() called while already spinning
+   * @param exception_handler will be called for every exception in the processing threads
```
wjwwood, line 88: We don't use `\` everywhere (some other `@` have slipped through reviews), but we should tend towards the desired state when possible, which is `\` everywhere. I'll fix these up in my version of this pr.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -82,6 +82,18 @@ class Executor
   virtual void
   spin() = 0;
 
+  /**
+   * \sa rclcpp::Executor:spin() for more details
```
wjwwood, line 86: This needs a "brief", with a `/// ...` line before the block comment.

rclcpp/include/rclcpp/experimental/timers_manager.hpp
```diff
@@ -125,9 +125,14 @@ class TimersManager
   /**
    * @brief Starts a thread that takes care of executing the timers stored in this object.
    * Function will throw an error if the timers thread was already running.
+   *
+   * @param exception_handler if valid, the execution of the timer will be done in a try catch block,
```
wjwwood, line 129: I'll leave this `@`, because this file uses that style, but the whole file should be updated (it mostly got a review pass for being in the `experimental` namespace.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -395,6 +395,69 @@ Executor::execute_any_executable(AnyExecutable & any_exec)
   any_exec.callback_group->can_be_taken_from().store(true);
 }
 
+template<typename Function>
+void execute_guarded(
+  const Function & fun,
```
wjwwood, line 400: nitpick: use full variable names

```suggestion
  const Function & function,
```

rclcpp/src/rclcpp/executors/static_single_threaded_executor.cpp
```diff
@@ -43,6 +43,24 @@ StaticSingleThreadedExecutor::spin()
...
+  RCPPUTILS_SCOPE_EXIT(this->spinning.store(false); );
+
+  // This is essentially the contents of the rclcpp::Executor::wait_for_work method,
+  // except we need to keep the wait result to reproduce the StaticSingleThreadedExecutor
+  // behavior.
+  while (rclcpp::ok(this->context_) && spinning.load()) {
+    this->spin_once_impl(std::chrono::nanoseconds(-1), exception_handler);
+  }
+}
+
```
wjwwood, line 63: We should not copy-paste like this, we should have one call the other...

rclcpp_lifecycle/test/benchmark/benchmark_lifecycle_client.cpp
```diff
@@ -203,7 +203,7 @@ class BenchmarkLifecycleClient : public performance_test_fixture::PerformanceTes
     executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
     executor->add_node(lifecycle_node->get_node_base_interface());
     executor->add_node(lifecycle_client->get_node_base_interface());
-    spinner_ = std::thread(&rclcpp::executors::SingleThreadedExecutor::spin, executor);
+    spinner_ = std::thread([this]() {executor->spin();});
```
wjwwood, line 206: Why is this change needed?

wjwwood (changes requested):
I am going to fix up this pr and run CI on it as an option to merge, but I really think that it would be preferable to have an exception handler callback as an optional part of the ExecutorOptions instead, passing it during construction and/or setting it after construction with a new method, rather than modifying all of these spin functions and causing down stream changes.

Please don't merge this version of the pr, see: https://github.com/ros2/rclcpp/pull/2505

fujitatomoya:
> but I really think that it would be preferable to have an exception handler callback as an optional part of the ExecutorOptions instead

+1 for this.

wjwwood:
Another concern I've got after trying to shape this up for merge is that there's no way to indicate if an exception should cause the executor to stop or not. You can re-throw, but doing that in a thread will trigger termination, which by default will head towards `std::abort`, but can be overridden see: https://en.cppreference.com/w/cpp/error/set_terminate

Instead, I'd like to have seen either a return value from the exception handler that can tell the executor to stop or throw from the main thread or something, and/or pass the executor into the exception handler so that it can cancel it if desired.

rclcpp_lifecycle/test/benchmark/benchmark_lifecycle_client.cpp
wjwwood: I see it doesn't compile... this is the kind of change that suggests we should rethink the approach. The old style of calling it should still work I think. I'll leave it, but it's another point against this approach in my mind.

wjwwood:
Closing in favor of the updated version https://github.com/ros2/rclcpp/pull/2505

Issue #2383, opened 2023-12-07
Clock::sleep_for is not time shifts resistant


AlexeyMerzlyakov:
The `clock->sleep_for()` is not resistant to time shifts. This is related especially for backwards time jumps: when the time was changed to the past, the routine will hang for a long time (unless initially calculated due-time will be reached).

This I believe, is incorrect behavior for `sleep_for()`, which should sleep for a given period of time independently of what was changed in the system (e.g. time was corrected by OS time server, or something else).

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - ROS2 Rolling built from sources
- Version or commit hash:
  - fff009a75100f2afd8ef1c3863620bf5ebe67708
- DDS implementation:
  - FastDDS, CycloneDDS
- Client library (if applicable):
  - rclcpp/rcl

#### Steps to reproduce issue
The simple repro-testcase is attached to current ticket: [test_pkg.zip](https://github.com/ros2/rclcpp/files/13597086/test_pkg.zip).
It creates the node, clock and attaches the given clock to the node.
Then, `clock->sleep_for(2s)` is being called in a cycle, while execution time of each sleep are being calculated during run-time:
```
    rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
    // Attaching clock to TimeSource, that subscribing to /clock and operating through this node
    rclcpp::TimeSource time_source(shared_from_this());
    time_source.attachClock(clock);

    while (rclcpp::ok()) {
      std::chrono::time_point<std::chrono::steady_clock> t1, t2;
      rclcpp::Time t1_ros, t2_ros;
      t1 = std::chrono::steady_clock::now();
      t1_ros = this->get_clock()->now();

      clock->sleep_for(2s);

      t2 = std::chrono::steady_clock::now();
      t2_ros = this->get_clock()->now();
      const auto dt = std::chrono::duration<float>(t2 - t1).count();
      const auto dt_ros = (t2_ros - t1_ros).seconds();
      RCLCPP_INFO(get_logger(), "Sleep time Real: %f ROS: %f\n", dt, dt_ros);
    }
```
When the clock is being shifted, `sleep_for()` will produce incorrect timings (for the forward clock shifts), or sleep for an indefinite periods of time (for backwards clock shifts).

To run the testcase, create testing workspace `mkdir -p ~/test_ws/src`, unpack there `test_pkg.zip` and build it there.

[...]

PR #2386, opened 2023-12-12
Clock::sleep_for() to be time shifts resistant

AlexeyMerzlyakov:
This PR addresses the #2383 issue.

Currently, `Clock::sleep_for()` is a wrapper for `Clock::sleep_until()` routine. Due-time once being calculated won't be changed if system or ROS-clock was shifted from one time (date) to another.

The suggested patch - is the re-implementation attempt of `Clock::sleep_for()` to have a resistance to time shifts.
The implementation considerations are:

 - For the cases when clock has system, steady or inactive ROS time, `sleep_for()` implementation could consider the sleep for a necessary amount of time, similar how does `rclcpp::sleep_for()` do it. For all of these cases, the speed of time flow will be normal, and be the same as how steady time flows. So, for these cases, we are using `chrono::steady_clock`, which will sleep for a necessary amount of time independently of system or ROS time changes.
 - For the case when clock has active ROS time (that is subscribed `/clock` topic for time corrections), `sleep_for()` should check the clock the same way as `sleep_until()` does it, but also respecting backwards time shifts in a clock jump callbacks. If updated ROS-time was jumped backwards than previous noted time, we are correcting until-point to fit actual sleep timings.

Overall implementation scheme looks like following:
```
bool
Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
{
  ...
  if (this_clock_type == RCL_STEADY_TIME || this_clock_type == RCL_SYSTEM_TIME)
  {
    // Sleeping using std::chrono::steady_clock for time calculation
  } else if (this_clock_type == RCL_ROS_TIME) {
    if (!ros_time_is_active()) {
      // RCL_ROS_TIME is not active. Time speed is normal.
      // Sleeping using std::chrono::steady_clock for time calculation
    } else {
      // RCL_ROS_TIME is active. Need to sleep in respect to time jumps in the past.
    }
  }
}
```

### DIFF
--- a/rclcpp/src/rclcpp/clock.cpp
+++ b/rclcpp/src/rclcpp/clock.cpp
@@ -179,7 +179,102 @@ Clock::sleep_until(Time until, Context::SharedPtr context)
 bool
 Clock::sleep_for(Duration rel_time, Context::SharedPtr context)
 {
-  return sleep_until(now() + rel_time, context);
+  if (!context || !context->is_valid()) {
+    throw std::runtime_error("context cannot be slept with because it's invalid");
+  }
+  const auto this_clock_type = get_clock_type();
+  bool time_source_changed = false;
+
+  std::condition_variable cv;
+
+  // Wake this thread if the context is shutdown
+  rclcpp::OnShutdownCallbackHandle shutdown_cb_handle = context->add_on_shutdown_callback(
+    [&cv]() {
+      cv.notify_one();
+    });
+  // No longer need the shutdown callback when this function exits
+  auto callback_remover = rcpputils::scope_exit(
+    [context, &shutdown_cb_handle]() {
+      context->remove_on_shutdown_callback(shutdown_cb_handle);
+    });
+
+  if (this_clock_type == RCL_STEADY_TIME || this_clock_type == RCL_SYSTEM_TIME) {
+    const std::chrono::nanoseconds rel_time_chrono =
+      std::chrono::nanoseconds(rel_time.nanoseconds());
+    std::chrono::nanoseconds time_left = rel_time_chrono;
+    const std::chrono::steady_clock::time_point start = std::chrono::steady_clock::now();
+    do {
+      {
+        std::unique_lock lock(impl_->clock_mutex_);
+        auto now = std::chrono::steady_clock::now() - start;
+        // this will release the lock while waiting
+        cv.wait_for(lock, time_left);
+        time_left -= std::chrono::steady_clock::now() - start;
+      }
+    } while (time_left > std::chrono::nanoseconds::zero() && context->is_valid());
+  } else if (this_clock_type == RCL_ROS_TIME) {
+    Time curr = now();
+    Time until = curr + rel_time;
+
+    // Install jump handler for any amount of time change, for two purposes:
+    // - if ROS time is active, check if time reached on each new clock sample
+    // - Trigger via on_clock_change to detect if time source changes, to invalidate sleep
+    rcl_jump_threshold_t threshold;
+    threshold.on_clock_change = true;
+    // 0 is disable, so -1 and 1 are smallest possible time changes
+    threshold.min_backward.nanoseconds = -1;
+    threshold.min_forward.nanoseconds = 1;
+    auto clock_handler = create_jump_callback(
+      nullptr,
+      [&cv, &time_source_changed, &curr, &until, this](
+        const rcl_time_jump_t & jump)
+      {
+        if (jump.clock_change != RCL_ROS_TIME_NO_CHANGE) {
+          time_source_changed = true;
+        }
+        if (now() < curr) {
+          // Time jumped in the past: correcting until value on time shift
+          const Duration delta = curr - now();
+          until -= delta;
+        }
+
+        cv.notify_one();
+      },
+      threshold);
+
+    if (!ros_time_is_active()) {
+      // RCL_ROS_TIME is not active. Time flaws as steady one.
+      const std::chrono::nanoseconds rel_time_chrono =
+        std::chrono::nanoseconds(rel_time.nanoseconds());
+      std::chrono::nanoseconds time_left = rel_time_chrono;
+      const std::chrono::steady_clock::time_point start = std::chrono::steady_clock::now();
+      do {
+        {
+          std::unique_lock lock(impl_->clock_mutex_);
+          auto now = std::chrono::steady_clock::now() - start;
+          // this will release the lock while waiting
+          cv.wait_for(lock, time_left);
+          time_left -= std::chrono::steady_clock::now() - start;
+        }
+      } while (
+        time_left > std::chrono::nanoseconds::zero() && context->is_valid() &&
+        !time_source_changed);
+    } else {
+      // RCL_ROS_TIME is active. Need to waiting in respect to time jumps in the past.
+      std::unique_lock lock(impl_->clock_mutex_);
+      while (curr < until && context->is_valid() && !time_source_changed) {
+        cv.wait(lock);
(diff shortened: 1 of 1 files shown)


rclcpp/src/rclcpp/clock.cpp
```diff
@@ -179,7 +179,102 @@ Clock::sleep_until(Time until, Context::SharedPtr context)
...
+      [&cv, &time_source_changed, &curr, &until, this](
+        const rcl_time_jump_t & jump)
+      {
+        if (jump.clock_change != RCL_ROS_TIME_NO_CHANGE) {
+          time_source_changed = true;
+        }
+        if (now() < curr) {
+          // Time jumped in the past: correcting until value on time shift
+          const Duration delta = curr - now();
+          until -= delta;
```
sloretz, line 238: Thank you for the PR! It looks like this is keeping track of time already slept when ROS time is active and jumps backwards, and then using that to credit the amount of time slept.

This doesn't seem correct to me. ROS time being active indicates something artificial messing with time, such as simulation or replaying a bag. Say a bag was playing and a node was waiting 60 seconds using sleep_for(), but someone kept replaying 5 seconds in the middle while debugging some other part of the system. If those 5 seconds were credited towards the "sleep_for" time then from the perspective of the simulated time the sleep period would be completed faster than 60 simulated seconds.

 I think treating backwards time jumps as if no time had elapsed is the correct behavior in that case.

sloretz (changes requested):

sloretz:
@ros2/team for other's thoughts 🧇

rclcpp/src/rclcpp/clock.cpp
AlexeyMerzlyakov: Thank you for the review and reasonable notice. Yes, indeed from this point of view, I would like agree that we will miss this use-case, and thus vanilla `clock->sleep_for()` behavior seems to be correct.

However, developing a little bit further your example with replaying rosbag and some node that we are using for debugging. Imagine that this node is additionally publishing something using `sleep_for` _while_ we are waiting 60 seconds, e.g.:
```
while (ok) {
  pub_->publish(some_msg);
  this->get_clock()->sleep_for(100ms);
}
```

Vanilla implementation of `clock->sleep_for()` might cause the problem to debug this node, as once ROS-time was jumped backward during the sleep, `this->get_clock()->sleep_for(100ms)` would not complete withing expected period of time and no new messages will be arrived, when (as mentioned above) someone will re-play back rosbag file for debugging.

From one point of view, when `sleep_for()` API is used "out of" debugging, we should not react to time shifts to the past. However, from another point of view, when `sleep_for()` is used in debugging code, we will get stuck, and get non-reproducible program behavior.
So, I am still have some doubts, what is better to do here: leave `Clock::sleep_for()` to not consider time shifts (in our case: for ROS-time shifts, but I believe the logic of your reasoning could also be moved to System and non-activated ROS clocks), so we should remain `sleep_for()` as a wrapper for `sleep_until()`;
or `sleep_for()` should be time shifts - resistant routine?
I think, this some kind of fundamental agreement question of how we want this should be work. And I could miss something else?

fujitatomoya: @sloretz @AlexeyMerzlyakov 

> Say a bag was playing and a node was waiting 60 seconds using sleep_for(), but someone kept replaying 5 seconds in the middle while debugging some other part of the system. If those 5 seconds were credited towards the "sleep_for" time then from the perspective of the simulated time the sleep period would be completed faster than 60 simulated seconds.

i am not sure why this is the problem. i think this `5 seconds` has been already elapsed for the `sleep_for` before the time jumps backward, so that `until` should be deducted by this `5 seconds` based on the new `now()`? i think that this behavior is correct for sleep`_for`, because it is not sleep`_until`. i may be mistaken, please let me know if i got anything wrong. thanks in advance.

methylDragon: I tried looking at the C++ standard for this, and [their description of `sleep_for` is ambiguous](https://en.cppreference.com/w/cpp/thread/sleep_for)...

I quote (emphasis mine):
> The standard recommends that a steady clock is used to measure the duration. If an implementation uses a system clock instead, the wait time __may__ also be sensitive to clock adjustments.

So I think it's up to us, since a sim-time clock isn't a steady clock and makes no guarantees.

Now, in my opinion, a time jump backward in time counts as sleeping for **negative time** (because if time moving forward counts towards time slept for, why shouldn't time moving backwards count as anything other than (negative) time slept for?

So I lean towards @sloretz's opinion (in fact, I think time should be debited instead of being treated as no time in the case of a time jump backwards).

AlexeyMerzlyakov: > i am not sure why this is the problem. i think this 5 seconds has been already elapsed for the sleep_for before the time jumps backward, so that until should be deducted by this 5 seconds based on the new now()?

Current implementation of `Clock::sleep_for()` using once calculated `until` time at its beginning, and then calling `Clock::sleep_until(until)`. If ROS or system time was moved backwards during the sleep, `until` value won't be deducted.

-----

> The standard recommends that a steady clock is used to measure the duration. If an implementation uses a system clock instead, the wait time **may** also be sensitive to clock adjustments.

but:

> This probably does not matter, since the POSIX.1  specification  for  clock_set‐time(2) says that discontinuous changes in CLOCK_REALTIME should not affect nanosleep()

So, what is the decision here: should we follow POSIX-way to leave all sleep-for-s to be time change independent, or we'd better to leave `sleep_for()` implementation as-is for active ROS clocks? (following the same logic, we'd also leave the system clock (and inactive ROS clocks as well) the same way. In this case if developer wants to have time-shift resistance `sleep_for`, the `Clock` object should be chosen of `RCL_STEADY_TIME` type intentionally).

wjwwood: In my opinion, having sleep_for be a wrapper around sleep_until is reasonable since we are selecting the clock to use for each sleep call, and so the idiosyncrasies of each particular clock will apply. The idea of time being a discretely incrementing resource and therefore a resource that can contribute towards a sleep_for even though the "start + duration < now" (according to the clock), seems wrong to me.

Put another way, I think a test like this should be totally reasonable:

```c++
auto sleep_duration = 1s;
auto start = ros_time_clock.now();
ros_time_clock.sleep_for(sleep_duration);
auto now = ros_time_clock.now();
ASSERT(now >= start + sleep_duration);
```

And not just for a ROS Time based clock, but any clock. Though I think there are some sleep_for like calls on Windows that are allowed to be shorter than the specified duration, but I think that's actually insane, personally. I don't believe we rely on those at all, except maybe through another library.

In the given example of a `while(ok) { ... sleep_for(...); }` loop, I think it's totally appropriate for that sleep_for to never exit if ROS time is looping within its start time and theoretical end time. That might be surprising to some, but I think it's the most reasonable behavior, especially if you consider the same situation, but rather than looping time in a bag you instead pause time. I don't see a good reason why iterating in a short loop should produce different behavior from being paused, personally.

Developers using ROS time need to be aware of these issues and use a steady clock in loops like these if they don't want that behavior.

AlexeyMerzlyakov: Thanks @wjwwood @fujitatomoya @sloretz @methylDragon for the opinions! I think this dotted all the i's, so the ticket could be closed: current `Clock::sleep_for()` behavior is as we expect it to be in ROS2; and it proper for the reasons listed above.

AlexeyMerzlyakov:
Closed for the reasons discussed in https://github.com/ros2/rclcpp/pull/2386#discussion_r1434316678 thread

PR #2209, opened 2023-06-12
make type support helper supported for service

iuhilnehc-ynos:
to extend the type support for service
required by https://github.com/ros2/rclcpp/pull/2358

PR #2392, opened 2023-12-15
:man_farmer:  Revert "make type support helper supported for service (#2209)"

Blast545:
This reverts commit d9b2744057b8fd230f2c71c739fcdf7e27219d85.

The PR I'm proposing to revert seems to be breaking the cppcheck, causing a timeout in `nightly_win_deb`.

I'm opening this PR in draft to test with CI if this was actually the cause of the problem.

Reference failure we'll try to replicate: https://ci.ros2.org/view/nightly/job/nightly_win_deb/2942/ 

CI test, CI_Windows, debug, packages-up-to rclcpp_lifecycle
[Build Status](https://ci.ros2.org/job/ci_windows/20691/)

### DIFF
--- a/rclcpp/include/rclcpp/generic_publisher.hpp
+++ b/rclcpp/include/rclcpp/generic_publisher.hpp
@@ -77,7 +77,7 @@ public:
   : rclcpp::PublisherBase(
       node_base,
       topic_name,
-      *rclcpp::get_message_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib),
+      *rclcpp::get_typesupport_handle(topic_type, "rosidl_typesupport_cpp", *ts_lib),
       options.template to_rcl_publisher_options<rclcpp::SerializedMessage>(qos),
       // NOTE(methylDragon): Passing these args separately is necessary for event binding
       options.event_callbacks,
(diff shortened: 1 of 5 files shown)


clalancette:
Do we still need this PR?  It looks like the nightly_win_deb is generally passing without it: https://ci.ros2.org/view/nightly/job/nightly_win_deb/

That said, if we run into further problems here, my first thought would be to increase the cppcheck timeout.  We've done that elsewhere.

Blast545:
I was exactly right now looking into the results of the PR. I never had any intention to merge this one, just wanted to check if the last PR to the repo had any influence in the appearance ratio of the issue.

Given the recent history of the job, it seems the problem has stopped appearing while it happened consecutively 4 times last week. After a manual check, I realized it has happened as well in the past. I'm starting to think that the capabilities of the agent might be the actual cause behind the yellows last week.

Do you think increasing the timeout from 300s to something like 350s would be OK? It's a lot of time for just one linter.

clalancette:
> Do you think increasing the timeout from 300s to something like 350s would be OK? It's a lot of time for just one linter.

Yeah, I think that is OK for now (though I might suggest 360 seconds, just so it is an even number of minutes).

I agree it is a long time, but Windows is the only platform where cppcheck still works.  Eventually maybe we should just disable it, but that is a larger discussion.

Blast545:
Roger that, will open a PR to increase that timeout and close this PR.

PR #2400, opened 2024-01-03
Stop storing the context in the guard condition.

clalancette:
Stop storing the context in the guard condition.
    
This was creating a circular reference between GuardCondition
and Context, so that Context would never be cleaned up.
Since we never really need the GuardCondition to know
about its own Context, remove that part of the circular
reference.
    
While we are in here, we also change the get_context()
lambda to a straight weak_ptr; there is no reason for the
indirection since the context for the guard condition
cannot change at runtime.
    
We also remove the deprecated version of the
get_notify_guard_condition().  That's because there is
no way to properly implement it in the new scheme, and
it seems to be unused outside of rclcpp.
    
Finally, we add in a test that guarantees the use_count
is what we expect when inside and leaving a scope, ensuring
that contexts will properly be destroyed.
    
Signed-off-by: Chris Lalancette <clalancette@gmail.com>

@mjcarroll @wjwwood FYI

### DIFF
--- a/rclcpp/src/rclcpp/guard_condition.cpp
+++ b/rclcpp/src/rclcpp/guard_condition.cpp
@@ -23,16 +23,17 @@ namespace rclcpp
 {
 
 GuardCondition::GuardCondition(
-  rclcpp::Context::SharedPtr context,
+  const rclcpp::Context::SharedPtr & context,
   rcl_guard_condition_options_t guard_condition_options)
-: context_(context), rcl_guard_condition_{rcl_get_zero_initialized_guard_condition()}
+: rcl_guard_condition_{rcl_get_zero_initialized_guard_condition()}
 {
-  if (!context_) {
+  if (!context) {
     throw std::invalid_argument("context argument unexpectedly nullptr");
   }
+
   rcl_ret_t ret = rcl_guard_condition_init(
     &this->rcl_guard_condition_,
-    context_->get_rcl_context().get(),
+    context->get_rcl_context().get(),
     guard_condition_options);
   if (RCL_RET_OK != ret) {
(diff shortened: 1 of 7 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20023/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14487/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20746/)

mjcarroll (approved):
LGTM.  I know that there _was_ a reason for using a `std::function` here, rather than passing the Context itself, but I'm struggling to recall at this point.  @wjwwood do you remember?

clalancette:
I'm going to go ahead and merge this one in.  If there is any follow-up, I'm happy to do it in another PR.

Crola1702:
Hi! Do you think this PR fixes https://github.com/osrf/buildfarm-tools/issues/21#issuecomment-1883185244?

rclcpp/src/rclcpp/guard_condition.cpp
```diff
@@ -23,16 +23,17 @@ namespace rclcpp
...
 {
-  if (!context_) {
+  if (!context) {
     throw std::invalid_argument("context argument unexpectedly nullptr");
   }
+
   rcl_ret_t ret = rcl_guard_condition_init(
     &this->rcl_guard_condition_,
-    context_->get_rcl_context().get(),
+    context->get_rcl_context().get(),
```
wjwwood, line 36: Wait wasn't the point of holding the context this right here? It's not ok if the context goes out of scope before the guard condition. I'd have to understand the original issue better, but I wanted to bring this up as I saw it.

PR #2405, opened 2024-01-16
DRAFT CallbackGroup centric executor

jmachowinski:
This is a draft for a new executor, that saves ready executables per callback group. 
In doing so, this executor should have the same behavior as the SingleThreaded executor in the 
single and multithreaded use case. This hopefully fixes some ongoing bugs.

As a second feature, this executor features a O(1) backward mapping from rcl_events to rclcpp events.
This should hopefully result in better performance, but only measurements will tell...

Note, this is a DRAFT. It does not event compile yet, but it would be nice to get some feedback on the approach and obvious problems that may appear.

@mjcarroll @wjwwood @clalancette @fujitatomoya Can you have a short look at this ?


alsora:
> Note, there is something fishy going on, as the CbgExecutor receives more messages than the normal one

The likely explanation is that publishers/timers are running faster (thanks to the improved performance of the new executor) so you are publishing more messages.

I would be interested in seeing a performance comparison with the `EventsExecutor`, which is considerably faster than the other executors you tested with.
IMO if this proposed executor does not have better performance than the `EventsExecutor`, we should not add it to rclcpp.
We are already looking to unify the existing executors because there's no reason to have so many, so a new executor should be added only if it's superior in terms of performance.

Moreover, with a new executor proposal, we should have a detailed performance analysis and design document accompanying it.
The new executor should initially either be implemented in a separate repository or in the `experimental` namespace/directory.

NOTE: I didn't look at the code yet.
I'll try to do a review this week.

jmachowinski:
> The likely explanation is that publishers/timers are running faster (thanks to the improved performance of the new executor) so you are publishing more messages.
> 
That would be nice, but I think something is broken with the shutdown.

> I would be interested in seeing a performance comparison with the `EventsExecutor`, which is considerably faster than the other executors you tested with. 

Define 'faster' :-). The latency of the EventsExecutor is way better.

> IMO if this proposed executor does not have better performance than the `EventsExecutor`, we should not add it to rclcpp. 

The main point of this executor is that it has similar behavior in single and multithreaded mode. Note, this is not the case for the current executors. That it is fast, and shows a reduction of CPU load is a nice addon.

> Moreover, with a new executor proposal, we should have a detailed performance analysis and design document accompanying it. The new executor should initially either be implemented in a separate repository or in the `experimental` namespace/directory.

I changes the design 4 times during the refinement of the approach... This started out as a request for feedback for an idea, and meanwhile evolved into a full blown reasonable stable executor. I would first like to discuss it, and if I can get the changes needed for it into rclcpp and rcl, before putting any work into design documents...

> 
> NOTE: I didn't look at the code yet. I'll try to do a review this week.

That would be great, thanks

alsora:
> The main point of this executor is that it has similar behavior in single and multithreaded mode. Note, this is not the case for the current executors. That it is fast, and shows a reduction of CPU load is a nice addon.

This makes sense.
However I think that the current landscape of executors is already quite confusing for a user.

Usually the path forward for a new executor is the following
 - create a dedicated repository for this new executor (outside of rclcpp)
 - create design documents
 - gather community interest
 - eventually push to rclcpp if there's enough interest

This PR can stay open (maybe as a draft) to get initial feedbacks from the maintainers, but I would prefer to start with a design document to fully understand the theory of operation before diving into the code.

If the purpose of this executor is to "fix" a bug in the current single and multithreaded executors, I would rather ask to replace them and/or apply the changes to those executors themselves rather than creating a new one.

On the other hand, if "having a similar behavior in single and multithreaded mode" isn't really a bug-fix, but rather an alternative approach (similar to how the events executor processes events in a different order than the single threaded executor), we could consider adding this new executor after the community has showed enough interest in it (e.g. contributions, discourse discussions, etc)

> Define 'faster' :-). The latency of the EventsExecutor is way better.

I used "faster" on purpose as a very generic term. We showed how the events executor had better performance than the standard executor across a large number of tests and metrics (CPU usage, latency, throughput, accuracy of timers, and slightly even RAM usage).
If a similar case could be made for a new executor, I would recommend  the process described before with the goal to replace the events-executor (which is still marked as "experimental").

We currently have 4 executors in the repo. I really don't want to go to 5 unless there's a very strong reason.
We had plans to reduce the number of executors (fully deprecate the `StaticSingleThreadedExecutor`, move the `EventsExecutor` out of `::experimental` and make it the default, eventually remove the `SingleThreadedExecutor`), so it would be great if new executor proposals fit into this schema of "reducing the complexity for users".

jmachowinski:
> This PR can stay open (maybe as a draft) to get initial feedbacks from the maintainers, but I would prefer to start with a design document to fully understand the theory of operation before diving into the code.

I'll try to make some time to come up with a document.

> If the purpose of this executor is to "fix" a bug in the current single and multithreaded executors, I would rather ask to replace them and/or apply the changes to those executors themselves rather than creating a new one.
 
Replacing Static, Single and Multi threaded Executor with this one would be feasible, but also a bit drastic. My idea was to add it as a possible replacement for transition and testing.
From my point of view this is a needed bugfix, as the MultiThreadedExecutor caused a lot of headaches for us in the past year. But it is also a new approach from the internal working of things.

I had a short look at the EventsExecutor and would say, if we merge it with this approach, we will get something really nice. The EventsExecutor has the solution to how avoid the waitsets, which take ages to build. On the other side, the cbg executor has the logic, to schedule ready events in callback groups with priorized order by multiple threads.

jmachowinski:
@alsora I overworked the description, to contain a proposal and a detailed description of the issue that is should fix.

I would be fine, with having this executor outside of rclcpp, but as mentioned above, I will need to adopt some APIs and structures slightly in rcl and rclcpp for this.

As for my idea on how to combine this and the EventsExecutor approach, its basically 'take out the wait for work stuff from the spin function and fill the event queues from callbacks from RWM'. The housekeeping might be challenging though.

mjcarroll:
Many ideas here, thanks for doing the leg-work and sorry for taking so long to begin thinking about this.

One note, and something that @wjwwood keeps reminding me about our current implementation

> This will clear the global internal event queue (within AllocatorMemoryStrategy) and repopulate it with ready events.

This isn't so much a _queue_ as a _set_.  Basically, everything has a boolean state of "am I ready or not" rather than "how many things do I have ready".  In the case that you had a subscription that received 1 message during the wait and one that received 100 messages, both will be processed equally (one callback per subscription).

jmachowinski:
closing, final implementation can be found in https://github.com/cellumation/cm_executors

Issue #1029, opened 2020-03-20
Programmatically loading parameters from yaml files into nodes dynamically


ddengster:
## Feature request
Programmatically loading parameters from yaml files into nodes dynamically

#### Feature description
A simple c++ function (possibly put into rclcpp::Node?) for loading parameters from yaml files after the node has been started. (ie. not via supplying a yaml file in the `ros2 run` command)

#### Implementation considerations
Not sure if the function should be under rclcpp::Node.

There were several ros1 repos using this functionality of loading nodes dynamically and supplying yaml file. However, the spec on parameters changed with ros2.

Some upcoming repos that could use this functionality:
[ros2_controls loading controllers at runtime](https://github.com/ros-controls/ros2_control/issues/60)
ros2 version gazebo_ros_controls (blocked)

Reference code doing the same thing as this request that you could use:
https://github.com/ros-controls/ros2_control/pull/64/files

clalancette:
For what it is worth, it should be possible to do this pretty easily right now.  You'd just need a small program that opened up a YAML file (maybe using yaml-cpp), formatted that YAML into a https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/srv/SetParameters.srv (one per node), and then called the `set_parameters` service on the node(s).

I guess we could potentially make this easier with an API on rclcpp, but I'm not sure.

threeal:
Actually there is a `rcl_parse_yaml_file()` under [rcl_yaml_param_parser](https://github.com/ros2/rcl/blob/master/rcl_yaml_param_parser/include/rcl_yaml_param_parser/parser.h) that could be used to parse parameters from yaml config. It returns `rcl_params_t` though, but the good news, there is also `rclcpp::parameter_map_from()` that could be used to convert from `rcl_params_t` to `rclcpp::Parameter`.

So consider the parsing problem is solved, and it just need to overwrite or create a new parameter from the parsed `rclcpp::Parameter`.

CursedRock17:
Were the multiple `load_parameters` functions from pull request #1596 which can take from both a `parameter_map` and a yaml file enough, or can I work on this?

fujitatomoya:
@CursedRock17 i think we already can do this with, (really easy)

1. creating node with `namespace` and `node_name`.
2. create `rclcpp::AsyncParametersClient` with "`namespace/node_name`"
3. and then call `param_client->load_parameters(parameters_filepath)`

we can even do this with remote node.

[...]

PR #2406, opened 2024-01-17
Load params

CursedRock17:
This is a pull request meant to address #1029 and may start as a draft depending on the issue actually needing to be addressed. It adds the `load_parameters` function to `NodeParametersInterface` and all of its children; a way to add parameters during runtime through a yaml file.

### DIFF
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -586,6 +586,14 @@ public:
   void
   undeclare_parameter(const std::string & name);
 
+  /// Load a list of parameters from a yaml parameter file..
+  /**
+   * \param[in] yaml_name The name of the yaml file that needs to be loaded.
+   */
+  RCLCPP_PUBLIC
+  void
+  load_parameters(const std::string & yaml_filepath);
+
   /// Return true if a given parameter is declared.
   /**
    * \param[in] name The name of the parameter to check for being declared.
--- a/rclcpp/include/rclcpp/node.hpp
+++ b/rclcpp/include/rclcpp/node.hpp
@@ -591,8 +591,8 @@ public:
    * \param[in] yaml_name The name of the yaml file that needs to be loaded.
    */
   RCLCPP_PUBLIC
-  void
-  load_parameters(const std::string & yaml_filepath);
+  std::vector<rcl_interfaces::msg::SetParametersResult>
+  load_parameters(const std::string & yaml_filepath, const std::string & node_name_);
 
   /// Return true if a given parameter is declared.
   /**
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
@@ -137,8 +137,9 @@ public:
   undeclare_parameter(const std::string & name) override;
 
   RCLCPP_PUBLIC
-  void
-  load_parameters(const std::string & yaml_filepath) override;
+  std::vector<rcl_interfaces::msg::SetParametersResult>
+  load_parameters(
+  const std::string & yaml_filepath, const std::string & node_name_) override;
 
   RCLCPP_PUBLIC
   bool
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters.hpp
@@ -137,6 +137,10 @@ public:
   undeclare_parameter(const std::string & name) override;
 
   RCLCPP_PUBLIC
+  void
+  load_parameters(const std::string & yaml_filepath) override;
+
+  RCLCPP_PUBLIC
   bool
   has_parameter(const std::string & name) const override;
 
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
@@ -115,6 +115,15 @@ public:
   void
   undeclare_parameter(const std::string & name) = 0;
 
+  /// Load a list parameters from a yaml parameter file
+  /**
+   * \sa rclcpp::Node::load_parameters
+   */
+  RCLCPP_PUBLIC
+  virtual
+  void
+  load_parameters(const std::string & yaml_filepath) = 0;
+
   /// Return true if the parameter has been declared, otherwise false.
   /**
    * \sa rclcpp::Node::has_parameter
--- a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
+++ b/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp
@@ -121,8 +121,9 @@ public:
    */
   RCLCPP_PUBLIC
   virtual
-  void
-  load_parameters(const std::string & yaml_filepath) = 0;
+  std::vector<rcl_interfaces::msg::SetParametersResult>
+  load_parameters(
+  const std::string & yaml_filepath, const std::string & node_name_) = 0;
 
   /// Return true if the parameter has been declared, otherwise false.
   /**
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -373,6 +373,12 @@ Node::undeclare_parameter(const std::string & name)
   this->node_parameters_->undeclare_parameter(name);
 }
 
+void
+Node::load_parameters(const std::string & yaml_filepath)
+{
+  this->node_parameters_->load_parameters(yaml_filepath);
+}
+
 bool
 Node::has_parameter(const std::string & name) const
 {
--- a/rclcpp/src/rclcpp/node.cpp
+++ b/rclcpp/src/rclcpp/node.cpp
@@ -373,10 +373,10 @@ Node::undeclare_parameter(const std::string & name)
   this->node_parameters_->undeclare_parameter(name);
 }
 
-void
-Node::load_parameters(const std::string & yaml_filepath)
+std::vector<rcl_interfaces::msg::SetParametersResult>
+Node::load_parameters(const std::string & yaml_filepath, const std::string & node_name_)
 {
-  this->node_parameters_->load_parameters(yaml_filepath);
+  return this->node_parameters_->load_parameters(yaml_filepath, node_name_);
 }
 
 bool
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -657,6 +657,14 @@ NodeParameters::undeclare_parameter(const std::string & name)
   parameters_.erase(parameter_info);
 }
 
+void
+NodeParameters::load_parameters(const std::string & yaml_filepath)
+{
+  auto parameters_client = std::make_shared<rclcpp::AsyncParametersClient>(node_, "package_name");
+  parameters_client->load_parameters(yaml_filepath);
+
+}
+
 bool
 NodeParameters::has_parameter(const std::string & name) const
 {
--- a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
+++ b/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp
@@ -14,6 +14,7 @@
 
 #include "rclcpp/node_interfaces/node_parameters.hpp"
 
+#include <cwctype>
 #include <rcl_yaml_param_parser/parser.h>
 
 #include <array>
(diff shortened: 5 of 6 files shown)


wjwwood (changes requested):
We had a look at this in our issue triage meeting and the consensus was that while "loading" (more specifically setting and declaring if yet undeclared) parameters based on a yaml file is useful, however we don't think we should continue to expand the methods in the Node class or the NodeParametersInterface class and we'd prefer to see a free-function or similar that uses a Node and a yaml file to perform the operations.

To this point, your pull request doesn't add these methods to the `LifecycleNode` class, but if it was a free-function that took a `rclcpp::NodeInterface<...>` template argument (see https://github.com/ros2/rclcpp/blob/a29f58edcbe3a0347d71d0fec3e2232d699ea918/rclcpp/include/rclcpp/node_interfaces/node_interfaces.hpp#L68-L146), then this free function could operate on `Node` as well as `LifecycleNode` or any other kinds of "node-like" objects in the future. This is not only more reusable, but it helps us keep the size of the Node class manageable and therefore helps with maintainability.

rclcpp/include/rclcpp/node.hpp
```diff
@@ -586,6 +586,14 @@ class Node : public std::enable_shared_from_this<Node>
   void
   undeclare_parameter(const std::string & name);
 
+  /// Load a list of parameters from a yaml parameter file.
+  /**
+   * \param[in] yaml_name The name of the yaml file that needs to be loaded.
+   */
+  RCLCPP_PUBLIC
+  std::vector<rcl_interfaces::msg::SetParametersResult>
+  load_parameters(const std::string & yaml_filepath, const std::string & node_name_);
```
fujitatomoya, line 595: aside the on-going discussion on how to implement, do we need to have 2nd argument `node_name_`? what about `load_parameters` takes `yaml file` and internally calls `get_fully_qualified_name()`. if `load_parameters` belong to the Node class, i think it should load the parameters from yaml file in the context of the `Node`?

fujitatomoya (commented):
@wjwwood 

> we don't think we should continue to expand the methods in the Node class or the NodeParametersInterface class

i understand the concern and benefits for having this as free functions.
on the other hand, i also feel like `load_parameters` method can belong to `NodeParameterInterface`.

```cpp
  // pseudo code

  // list parameters belong to the node
  auto parameter_list = node->list_parameters({}, 0);

  // load parameters to the node
  results = rclcpp::load_parameters(node, yaml_file);
```

i do not have strong opinion, but it would be nice if we can provide the consistent way to manage parameters?

it would be probably nice to refactor the existed methods to use free function like `load_parameters` above? what do you think?

CursedRock17:
> it would be nice if we can provide the consistent way to manage parameters?

I would have to agree with this statement, I feel that too much separation will create inconsistency thus a worse DX. Furthermore, filing a free function doesn't seem clear right now, as right now in testing I just stuck it in `copy_all_parameter_values.hpp` and was planning on possibly renaming the file to make it more discoverable. That function declaration would look something like this:
```
template<typename NodeT>
load_parameters(
  const std::string & yaml_filepath, 
  rclcpp::node_interfaces::NodeInterfaces<NodeT> node_interface)
```



Though the benefits of a free function would be its variability to take various NodeInterfaces, if we have a limited amount of nodes that we care about; for example if we only care about `Node` and `LifecycleNode` opposed to caring about 6 different nodes or whatnot.

clalancette:
> I would have to agree with this statement, I feel that too much separation will create inconsistency thus a worse DX.

Looking at this holistically, I think my other problem with this is that our developer experience around parameters is already poor.  We have too many node APIs that deal with parameters.  Right now we have (all from https://github.com/ros2/rclcpp/blob/c10764f4329f8c1494b68ff60ce55a3c74c8de9e/rclcpp/include/rclcpp/node.hpp#L374-L1254):

1.  declare_parameter that takes a ParameterValue
2.  declare_parameter that takes a ParameterType
3.  declare_parameter that takes a templated default type
4.  declare_parameter that is templated on just the type
5.  declare_parameters that takes a std::map of <std::string, ParameterT>
6.  declare_parameters that takes a std::map of <std::string, std::pair<ParameterT, ParameterDescriptor>>
7.  undeclare_parameter
8.  has_parameter
9.  set_parameter
10. set_parameters
11. set_parameters_atomically
12. get_parameter
13. get_parameter, returning a bool if it was set
14. get_parameter, returning a bool if set and coercing the output to the templated type
15. get_parameter_or, which gets a parameter or a default value and returns a bool
16. get_parameter_or, which returns a templated type
17. get_parameters, which takes a vector of names and returns a vector of Parameter structures
18. get_parameters, which, given a prefix returns all parameters with that prefix
19. describe_parameter, which gets the ParameterDescriptor for a parameter
20. describe_parameters, which gets the ParameterDescriptros for a vector of parameter names
21. get_parameter_types, which gets the enum type for a vector of parameter names
22. list_parameters, which getst all parameters up to a depth
23. add_pre_set_parameters_callback, which registers a callback to be called before parameters are set
24. add_on_set_parameters_callback, which registers a callback to accept/reject parameters before they are set
25. add_post_set_parameters_callback, which registers a callback to be called after parameters are set
26. remove_pre_set_parameters_callback, which unregisters a pre callback
27. remove_on_set_parameters_callback, which unregisters an accept callback
28. remove_post_set_parameters_callback, which unregisters a post callback

And this list ignores the ParametersClient, ParameterEvents, ParameterEventsFilter, etc, that also revolve around parameters.

I think we need to think long and hard before we add to this mess, because I don't think it is sustainable to keep adding APIs here.  Instead I'm leaning towards just documenting how to do this, since I think this is completely achievable with the current APIs.

alsora:
+1 on documenting and/or add utility functions (that take node interfaces as input) as opposed to add APIs to the node class itself

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
```diff
@@ -15,14 +15,18 @@
 #ifndef RCLCPP__COPY_ALL_PARAMETER_VALUES_HPP_
 #define RCLCPP__COPY_ALL_PARAMETER_VALUES_HPP_
 
+#include <rcl_yaml_param_parser/parser.h>
+
```
fujitatomoya, line 19: Do we need this? This is included in `rclcpp/parameter_map.hpp` line 29?

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
```diff
@@ -77,6 +81,28 @@ copy_all_parameter_values(
...
 }
 
+/// Load a list of parameters from a yaml parameter file.
+/**
+ * \param[in] yaml_name The name of the yaml file that needs to be loaded.
+ * \param[in] node_interfaces The list of variadic NodeInterfaces taken as a template to set parameters for
+ */
+template<typename NodeT>
+std::vector<rcl_interfaces::msg::SetParametersResult>
+load_parameters(
```
fujitatomoya, line 91: this file name should be changed into something else like `parameter_interfaces.hpp`? `copy_all_parameter_values.hpp` sounds dedicated to `rclcpp::copy_all_parameter_values`?

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
```diff
@@ -77,6 +81,28 @@ copy_all_parameter_values(
   }
 }
 
+/// Load a list of parameters from a yaml parameter file.
+/**
+ * \param[in] yaml_name The name of the yaml file that needs to be loaded.
+ * \param[in] node_interfaces The list of variadic NodeInterfaces taken as a template to set parameters for
+ */
```
fujitatomoya, line 88: git does some weird thing? i think we can ignore this.

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
```diff
@@ -77,6 +81,28 @@ copy_all_parameter_values(
   }
 }
 
+/// Load a list of parameters from a yaml parameter file.
+/**
+ * \param[in] yaml_name The name of the yaml file that needs to be loaded.
+ * \param[in] node_interfaces The list of variadic NodeInterfaces taken as a template to set parameters for
```
fujitatomoya, line 87:
```suggestion
 * \param[in] yaml_name The name of the yaml file that needs to be loaded.
 * \param[in] node_interface The list of variadic NodeInterfaces taken as a template to set parameters for.
 * \returns an instance of a parameter map.
 * \throws InvalidParametersException if no valid parameters found.
```

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
```diff
@@ -77,6 +81,28 @@ copy_all_parameter_values(
...
+ */
+template<typename NodeT>
+std::vector<rcl_interfaces::msg::SetParametersResult>
+load_parameters(
+  const std::string & yaml_filepath, NodeT node_interface)
```
alsora, line 92: I would prefer if we didn't use a template here, but rather the `NodeInterfaces` class or directly passing the two required interface (`node_base_interface` and `node_parameters_interface`)

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
```diff
@@ -77,6 +81,28 @@ copy_all_parameter_values(
...
+{
+  rclcpp::ParameterMap parameter_map =
+    rclcpp::parameter_map_from_yaml_file(yaml_filepath, node_interface->get_name());
```
alsora, line 95: shouldn't this be `get_fully_qualified_name` ?

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
fujitatomoya: that would be better for compile time.

rclcpp/test/rclcpp/test_copy_all_parameter_values.cpp
```diff
@@ -86,3 +86,19 @@ TEST_F(TestNode, TestParamCopyingExceptions)
   node2->declare_parameter("Foo1", rclcpp::ParameterValue(0.123));
   EXPECT_NO_THROW(rclcpp::copy_all_parameter_values(node1, node2, override));
 }
+
+TEST_F(TestNode, TestFileImport)
+{
+  const uint64_t expected_param_count = 4;
+  auto load_node = std::make_shared<rclcpp::Node>(
```
fujitatomoya, line 93: we can add `LifecycleNode` test case here?

rclcpp/include/rclcpp/copy_all_parameter_values.hpp
CursedRock17: Yea I agree, I just didn't want to change it too fast. `parameter_interfaces.hpp` sounds good to me.

rclcpp/test/rclcpp/test_copy_all_parameter_values.cpp
CursedRock17: Would that not make a circular testing structure, as `LifecycleNode` in `lifecycle_rclcpp` needs to include `rclcpp`.

rclcpp/include/rclcpp/parameter_interfaces.hpp
```diff
@@ -77,6 +83,36 @@ copy_all_parameter_values(
...
 
+/// Load a list of parameters from a yaml parameter file.
+/**
+ * \param[in] yaml_name The name of the yaml file that needs to be loaded.
+ * \param[in] node_interface The list of variadic NodeInterfaces taken as a template to set parameters for.
+ * \returns an instance of a parameter map.
+ * \throws InvalidParametersException if no valid parameters found.
+ */
+std::vector<rcl_interfaces::msg::SetParametersResult>
+load_parameters(
```
clalancette, line 94: Just to be clear; further to my comments in https://github.com/ros2/rclcpp/pull/2406#issuecomment-1924589468 , this PR is currently a NAK from me.

clalancette (changes requested):

CursedRock17:
>  NAK 

Does this mean Negative Acknowledgement? I'm not familiar with the acronym.

If it does, I can see that point and would assume that in the context of:
> +1 on documenting and/or add utility functions (that take node interfaces as input

you're mainly on board with just documentation. Though, how detrimental would a utility function - which uses `NodeInterfaces` - be to organization of the parameter API.  Especially if it's separated from `Nodes` and `ParameterClients` then it shouldn't get in the way of any restructoring. Furthermore, if we're just documenting this function, where would we put something like this? Would it be just denoted near [`AsyncParametersClient::load_parameters`](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/parameter_client.cpp#L290)


Then, as a sidenote, would it be worth creating an issue for the long list of parameter APIs, if 
> We have too many node APIs that deal with parameters.

It sounds like, some of them may not be necessary, as suggested with this message. That issue would just highlight the process for removing certain methods.

tomkimsour:
Hi, is there any update on where this should be put ? How about the documentation ? Should it be only added to ros2 documentation or to some rclcpp functions as well

PR #2412, opened 2024-01-24
Add queue for events executor that allows user specified priorities

nightduck:
This allows the user to assign arbitrary fixed-priorities to events according to a user-defined function. It can be used like so:

    auto extract_priority = [](const rclcpp::experimental::executors::ExecutorEvent & event) {
         //Compute priority by examining event.entity_key
         return priority;
      };
    EventsExecutor executor(std::make_unique<PriorityEventsQueue>(extract_priority));

The PriorityEventsQueue allows arbitrary scheduling policies to be built on the EventsExecutor. For instance, rate-monotonic scheduling can be done by examining the deadline parameter in an Executable's QoS settings and prioritizing smaller values. Adding these deadlines to the release time makes an earliest-deadline-first scheduler.

pinging @alsora 
I pitched this idea to you at ROSCon last year.

### DIFF
--- a/rclcpp/include/rclcpp/experimental/executors/events_executor/priority_events_queue.hpp
+++ b/rclcpp/include/rclcpp/experimental/executors/events_executor/priority_events_queue.hpp
@@ -1,0 +1,155 @@
+// Copyright 2023 iRobot Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLCPP__EXPERIMENTAL__EXECUTORS__EVENTS_EXECUTOR__SIMPLE_EVENTS_QUEUE_HPP_
+#define RCLCPP__EXPERIMENTAL__EXECUTORS__EVENTS_EXECUTOR__SIMPLE_EVENTS_QUEUE_HPP_
+
+#include <condition_variable>
+#include <functional>
+#include <mutex>
+#include <queue>
+#include <utility>
+
+#include "rclcpp/experimental/executors/events_executor/events_queue.hpp"
+
+namespace rclcpp
+{
+namespace experimental
+{
+namespace executors
+{
+
+struct PriorityEvent {
+  int priority;
+  rclcpp::experimental::executors::ExecutorEvent event;
+};
+
+/**
+ * @brief This class implements an EventsQueue as a simple wrapper around a std::priority_queue.
+ * It does not perform any checks about the size of queue, which can grow
+ * unbounded without being pruned.
+ */
+class PriorityEventsQueue : public EventsQueue
+{
+public:
+  RCLCPP_PUBLIC
+  PriorityEventsQueue() {
+    // Default callback to extract priority from event
+    extract_priority_ = [](const rclcpp::experimental::executors::ExecutorEvent & event) {
+      return 0;
+    };
+  }
+
+  RCLCPP_PUBLIC
+  explicit PriorityEventsQueue(
+    std::function<int(const rclcpp::experimental::executors::ExecutorEvent &)> extract_priority)
+  : extract_priority_(extract_priority) {}
+
+  RCLCPP_PUBLIC
+  ~PriorityEventsQueue() override = default;
+
+  /**
+   * @brief enqueue event into the queue
+   * Thread safe
+   * @param event The event to enqueue into the queue
+   */
+  RCLCPP_PUBLIC
+  void
+  enqueue(const rclcpp::experimental::executors::ExecutorEvent & event) override
+  {
+    int priority = extract_priority_(event);
+    rclcpp::experimental::executors::PriorityEvent single_event = {priority, event};
+    single_event.event.num_events = 1;
+    {
+      std::unique_lock<std::mutex> lock(mutex_);
+      for (size_t ev = 0; ev < event.num_events; ev++) {
+        event_queue_.push(single_event);
+      }
+    }
+    events_queue_cv_.notify_one();
+  }
+
+  /**
+   * @brief waits for an event until timeout, gets a single event
+   * Thread safe
+   * @return true if event, false if timeout
+   */
+  RCLCPP_PUBLIC
+  bool
+  dequeue(
+    rclcpp::experimental::executors::ExecutorEvent & event,
+    std::chrono::nanoseconds timeout = std::chrono::nanoseconds::max()) override
+  {
+    std::unique_lock<std::mutex> lock(mutex_);
+
+    // Initialize to true because it's only needed if we have a valid timeout
+    bool has_data = true;
+    if (timeout != std::chrono::nanoseconds::max()) {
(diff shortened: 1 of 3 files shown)


alsora:
Hi @nightduck, 
thank you for the contribution.

Before we start with the review, could you please rebase your branch on top of `rolling` ? Currently there's a lot of conflicts and unrelated changes in the branch

nightduck:
I'm unable to reproduce the failed TestNodeGraph unit test on my machine, so I can't debug it.

rclcpp/include/rclcpp/experimental/executors/events_executor/priority_events_queue.hpp
```diff
@@ -0,0 +1,169 @@
...
+    extract_priority_ = [](const rclcpp::experimental::executors::ExecutorEvent & event) {
+        (void)(event);
+        return 0;
+      };
+  }
+
+  RCLCPP_PUBLIC
+  explicit PriorityEventsQueue(
+    std::function<int(const rclcpp::experimental::executors::ExecutorEvent &)> extract_priority)
+  : extract_priority_(extract_priority) {}
```
alsora, line 69: here we should check whether the passed `extract_priority` function is callable, and throw an exception if it's not

rclcpp/test/rclcpp/executors/test_priority_events_executor.cpp
```diff
@@ -0,0 +1,163 @@
...
+      (void)msg;
+      msg_received_high = true;
+      EXPECT_EQ(subscriptions_executed, 0);
+      subscriptions_executed++;
+    });
+
+  auto publisher = node->create_publisher<test_msgs::msg::Empty>("topic", qos);
+
+  // Create executor
+  auto extract_priority = [](const rclcpp::experimental::executors::ExecutorEvent & event) {
```
alsora, line 102: It looks like for the sake of this unit-test, this function can be greatly simplified to
```
if (event.type != rclcpp::experimental::executors::ExecutorEventType::SUBSCRIPTION_EVENT) {
    return 0L;
}
auto subscription = static_cast<const rcl_subscription_t *>(event.entity_key);
return rcl_subscription_get_options(subscription)->qos.deadline.sec;
```

alsora:
Overall the changes look good to me.
It's a relatively simple change that allows user to implement important prioritization logic.

I have a question for the other ROS 2 core maintainers (e.g. @wjwwood @clalancette @fujitatomoya etc).
Should new queue implementations be part of rclcpp or should they live in separate repositories?
Besides this one, we have a lock-free queue implementation (find it here https://github.com/irobot-ros/events-executor/tree/main/irobot_lock_free_events_queue) and we discussed about implementing a queue that would mimic the "waitset" behavior.

clalancette:
> I have a question for the other ROS 2 core maintainers (e.g. @wjwwood @clalancette @fujitatomoya etc). Should new queue implementations be part of rclcpp or should they live in separate repositories?

So that basically depends.

If it is core functionality that really has to be in the core, it should be there.  If it is an optional feature but has widespread appeal and usecases, it can also be in the core.

But if those aren't true, and/or we don't know the answer to those, then we should always start outside of the core.  That allows it to a) iterate on its own schedule, without the constraints of being in the core, b) reduces the burden on the core maintainers, c) keeps our build and test times manageable, and d) makes sure this functionality isn't release-critical.  We spend a lot of time making these systems extensible, so we should lean on that as much as possible.

nightduck:
This queue can certainly exist in its own package for now. But I've got a DAG scheduler in the pipeline that builds on it. And that requires additions/modifications to Node, SubscriptionBase, and TimerBase. If these are also left in their own package, I could work around it with subclassing, but it's clunky.

Once I can motivate a usecase and wider adoption for this work that would justify mainline-ing the priority queue and DAG scheduling executor, what would that process look like? Is there precedent for absorbing functionality from a child class into a parent class?

alsora:
>I've got a DAG scheduler in the pipeline that builds on it. And that requires additions/modifications to Node, SubscriptionBase, and TimerBase.

Whether these changes can be included right now really depends on how generally useful they might be and how invasive/large they are.
If their use-case is niche and restricted to the new executor, we should first to ensure that there's community interest for them in order to justify having them in the core.

The process could be:
 - create  a new repository that depends on rclcpp and extends it with the new features.
 - make sure to include a design document and motivations for the changes
 - gather community interest with discussions on discourse, on github, working groups or other avenues
 - once you have community interest we can merge the changes in rclcpp

This process is not set in stone, and it really depends on the type of changes you are looking to push.


Having said that, I would like that the core had sufficient tools to enable all different sort of use-cases through generic APIs, allowing the implementation of such tools in separate repositories/packages.
However, we need to keep the bloat of the ros 2 core under control.
For example, rather than adding new APIs, we should consider how to rework the existing ones to better serve a wider scope.

fujitatomoya (commented):
@nightduck thank you for the PR. this looks useful in general, but we would deal with the priority with cpu binding, scheduling policy and dedicated thread so on for products if necessary. that could be overkill for some cases, so this could be useful in general, i think.

after all, i agree with https://github.com/ros2/rclcpp/pull/2412#issuecomment-1931131490, probably we would need to see the feedback from community as 1st step.

rclcpp/test/rclcpp/executors/test_priority_events_executor.cpp
```diff
@@ -0,0 +1,143 @@
...
+    rclcpp::init(0, nullptr);
+  }
+
+  void TearDown()
+  {
+    rclcpp::shutdown();
+  }
+};
+
+TEST_F(TestPriorityEventsExecutor, priority_subs)
```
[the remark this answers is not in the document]
nightduck: By default, ROS2 gives preference to subscriptions that are registered first. In this test, the subscriptions are registered in reverse priority order. However, I can look into writing a separate unit test specifically for the `PriorityEventsQueue`

rclcpp/include/rclcpp/experimental/executors/events_executor/priority_events_queue.hpp
```diff
@@ -0,0 +1,173 @@
+// Copyright 2023 iRobot Corporation.
```
jmachowinski, line 1: hu ?

rclcpp/include/rclcpp/experimental/executors/events_executor/priority_events_queue.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+struct PriorityEvent
+{
+  int priority;
+  rclcpp::experimental::executors::ExecutorEvent event;
+};
+
+struct ComparePriorities : public std::binary_function<PriorityEvent, PriorityEvent, bool>
+{
+  _GLIBCXX14_CONSTEXPR
```
jmachowinski, line 43: I am pretty sure, this will either not compile on windows, or not give the wanted result.

jmachowinski (commented):
Looks interesting, this will fail though as soon as multi threading and locked callback groups get involved.

It should also be noted, that the single threaded executors in ros2 aren't priority first, but a mix between FIFO and priority. The scheduling is priority first for events that are discovered at the same time. The rest is FIFO.

rclcpp/include/rclcpp/experimental/executors/events_executor/priority_events_queue.hpp
nightduck: Whoops, casual plagiarism. (Used the simple_events_queue as a template)

nightduck:
Thanks everyone for your input and suggestions. But per @clalancette 's advice, I think I'm going to spin this off as a separate package and we can revisit this conversation next year.

ros-discourse:

https://discourse.ros.org/t/the-ros-2-c-executors/38296/4

PR #2420, opened 2024-01-30
Ensure waitables handle guard condition retriggering

wjwwood:
An emerging responsibility of Waitables that wasn't explicitly declared before is that they should be ensuring guard conditions should stay ready between waits so long as the condition for them being triggered is still true.

Some waitables will not need this, e.g. waitables that are used to only wake up an executor once, but don't have events tied to them don't need to be retriggered. But other waitables which use guard conditions to notify the executor of work to be done need to keep triggering those guard conditions on subsequent waits, so long as the work associated with them has not been completed.

This is still a WIP. I'll add more cross references and details here in the future.

### DIFF
--- a/rclcpp/include/rclcpp/event_handler.hpp
+++ b/rclcpp/include/rclcpp/event_handler.hpp
@@ -122,7 +122,7 @@ public:
   /// Check if the Waitable is ready.
   RCLCPP_PUBLIC
   bool
-  is_ready(rcl_wait_set_t * wait_set) override;
+  is_ready(const rcl_wait_set_t * wait_set) override;
 
   /// Set a callback to be called when each new event instance occurs.
   /**
@@ -294,7 +294,7 @@ public:
 
   /// Execute any entities of the Waitable that are ready.
   void
-  execute(std::shared_ptr<void> & data) override
+  execute(const std::shared_ptr<void> & data) override
   {
     if (!data) {
       throw std::runtime_error("'data' is empty");
--- a/rclcpp/include/rclcpp/event_handler.hpp
+++ b/rclcpp/include/rclcpp/event_handler.hpp
@@ -106,12 +106,6 @@ public:
     Event,
   };
 
-  // TODO(wjwwood): is this ok? do events continue to stay ready until they are
-  // taken/checked?
-  RCLCPP_PUBLIC
-  void
-  dummy() override {};
-
   RCLCPP_PUBLIC
   virtual ~EventHandlerBase();
 
--- a/rclcpp/include/rclcpp/event_handler.hpp
+++ b/rclcpp/include/rclcpp/event_handler.hpp
@@ -106,6 +106,12 @@ public:
     Event,
   };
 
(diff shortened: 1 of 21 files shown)


jmachowinski:
lgtm, execpt for the dummy stuff ;-)

jmachowinski:
@wjwwood Looking at you const changes to the execute API, it occurred to me, that it might be way better to use a std::unique_ptr here.

I got two motivations for this proposal:
- Performance.
Shared_ptr are relative expensive to copy around, as they trigger a cache lock. And we are in a hot code path, that is executed with a real high frequency.
- Cleaner semantics.
The waitable should drop the ownership of the taken data, and should have no way of changing it, after the take_data.

mjcarroll:
I especially agree on the clearer semantics in this case.

wjwwood:
What was there before, non-const-ref to shared_ptr, is definitely not what we want to convey in the APIs semantics, i.e. that it might reset/reseat the given pointer (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam).

jmachowinski:
> I will point out that the performance shouldn't be an issue since we're not copying the shared ptr in the interface I'm touching

The data pointer needs to be copied at least one time into the AnyExecutable.

mjcarroll:
@mjcarroll to run CI on this.

mjcarroll:
I think that this is actually ready for review.  I'm going to run CI to at least see where we are at on it.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20542/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14932/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21290/)

wjwwood:
I retargeted this onto https://github.com/ros2/rclcpp/pull/2142 temporarily until it is merged, then I'll retarget it to `rolling`.

wjwwood:
I'd like to merge https://github.com/ros2/rclcpp/pull/2467 before pushing this one for review.

wjwwood:
Now that https://github.com/ros2/rclcpp/pull/2467 is merged, I need to fix this pr up, which I will do asap.

jmachowinski:
@wjwwood I rebase this on head an cleaned it up a bit : 
https://github.com/jmachowinski/rclcpp/tree/ensure_waitables_handle_guard_condition_retriggering

Note, I removed the SubscriptionIntraProcess::add_to_wait_set as its already handled in the base class.

wjwwood:
Thanks, but I don't see any changes yet. Still has conflicts, did you push? Either way I was already working on the fix up, sorry for the duplication of effort.

wjwwood:
Trying to close and reopen the pr to fix the processing of new commits...

wjwwood:
It won't let me reopen it because the current target branch has been deleted, and I cannot edit it to change the target branch while closed, so I made a new pr: https://github.com/ros2/rclcpp/pull/2483

PR #2421, opened 2024-02-01
Split test_executors up into smaller chunks.

clalancette:
The original reason is that on Windows Debug, we were failing to compile because the compilation unit was too large.  But also this file was way too large (1200 lines), so it makes sense to split it up.

This is followup from #2375, which was the straw that broke the camel's back.  @alsora @mwcondino FYI.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20202/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14622/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20916/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20917/)

alsora (approved):

mjcarroll (approved):

mwcondino:
@clalancette  thanks for the heads up, sorry that previous PR broke things!

clalancette:
> @clalancette thanks for the heads up, sorry that previous PR broke things!

No worries.  Windows Debug isn't a "gating" change for us, so for anything that breaks we just do follow-up fixes.

PR #2428, opened 2024-02-18
docs: fix doxygen syntax for NodeInterfaces

ottojo:
No idea what doxygen does here (parser bug?), but the very extensive doxygen comments from the `NodeInterfaces` class doesn't make it into the HTML output currently, (see: https://docs.ros.org/en/rolling/p/rclcpp/generated/classrclcpp_1_1node__interfaces_1_1NodeInterfaces.html) and this fixes it...


mjcarroll (approved):

fujitatomoya (approved):
lgtm

fujitatomoya:
this is only doc section fix, so i think we do not need to run CI.

PR #2434, opened 2024-02-23
Rule of five: implement move operators

Timple:
I think @fujitatomoya  had a point: https://github.com/ros2/rclcpp/pull/2425#discussion_r1491699756

So this is the alternative to: https://github.com/ros2/rclcpp/pull/2425


clalancette:
@Timple It is preferred to update the old PR; that way the context of why we made the decision is preferred.  In #2425 , you can update both the patch (by force-pushing), and the issue title (the edit button).

So with that in mind, I'm going to close this one.

PR #2438, opened 2024-02-26
Switch from rcpputils::fs to std::filesystem

christophebedard:
Part of https://github.com/ros2/rcpputils/issues/164

I split the PR into two commits: internal changes and API changes.

### DIFF
--- a/rclcpp/include/rclcpp/logger.hpp
+++ b/rclcpp/include/rclcpp/logger.hpp
@@ -15,6 +15,7 @@
 #ifndef RCLCPP__LOGGER_HPP_
 #define RCLCPP__LOGGER_HPP_
 
+#include <filesystem>
 #include <memory>
 #include <string>
 #include <utility>
@@ -23,7 +24,6 @@
 
 #include "rcl/node.h"
 #include "rcutils/logging.h"
-#include "rcpputils/filesystem_helper.hpp"
 
 /**
  * \def RCLCPP_LOGGING_ENABLED
@@ -86,7 +86,7 @@ get_node_logger(const rcl_node_t * node);
  * \throws rclcpp::exceptions::RCLError if an unexpected error occurs.
  */
 RCLCPP_PUBLIC
-rcpputils::fs::path
+std::filesystem::path
 get_logging_directory();
 
 class Logger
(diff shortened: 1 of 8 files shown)


alsora (approved):

rclcpp/include/rclcpp/logger.hpp
```diff
@@ -86,7 +86,7 @@ get_node_logger(const rcl_node_t * node);
  * \throws rclcpp::exceptions::RCLError if an unexpected error occurs.
  */
 RCLCPP_PUBLIC
-rcpputils::fs::path
+std::filesystem::path
 get_logging_directory();
```
fujitatomoya, line 90: so your concern is, this breaks the followings, right? i mean basically we need to provide deprecation period for the user space, but we can not define both of the functions here.

```cpp
// Compile Error Case-1
rcpputils::fs::path my_path = rclcpp::get_logging_directory();
// Compile Error Case-2
auto your_path = rclcpp::get_logging_directory();
bool directory_flag = your_path.is_directory(); // std::filesystem::path does not have this method.
```

just idea, i am not even sure if this is actually feasible.

- keep this function `rclcpp` as implemented, just change into `std::filesystem:path`.
- add copy constructor and copy assignment operator to `rcpputils::fs::path` to take `std::filesystem::path`.
- (as planned) add deprecation to `rcpputils::fs::path`, so that user can see that `rcpputils::fs::path` will be removed in next release?

again, this might not be doable idea...

> I can't find any non-test calls to rclcpp::get_logging_directory() in the ROS 2 core.

the other part of me, asking myself if we need the deprecation period for this. but that is not something we cannot guarantee.

mjcarroll: I think that should work?  It's a relatively cheap copy-constructor (basically string to string)

christophebedard: I'll give it a shot and will probably add a small test for it. I'm just wondering if it's too ambiguous if someone uses `auto`.

> the other part of me, asking myself if we need the deprecation period for this. but that is not something we cannot guarantee.

Yeah it doesn't have a lot of users: https://github.com/search?q=rclcpp%3A%3Aget_logging_directory%28%29&type=code. Let's see if we can properly deprecate it first.

christophebedard: It might be too ambiguous, since both `rcpputils::fs::path` and `std::filesystem::path` have a conversion constructor for `std::string`.

christophebedard:
Superseded by #2579

Issue #997, opened 2020-02-18
LifecycleNode: Take shutdown transition when exiting


wieset:
#### Feature description

When a lifecycle node is shut down using CTRL-C, it would be helpful to automatically call the `shutdown` transition. I would actually consider this "expected behavior".

Currently, the user needs to take care of calling the `shutdown` transition before exiting, for example through installing their own signal handler. Otherwise, the node is destructed from whichever state it's currently in, potentially leaving the system in an undefined or unsafe state.

#### Implementation considerations

If the node is currently in another transitioning state, the shutdown request needs to be deferred. If the node is already finalized, the request should be ignored.

claireyywang:
@wieset Would you like to propose a PR? We'd be happy to review it.

Rezenders:
Any updates on this?
I tired the solution on: https://answers.ros.org/question/343744/ros2-lifecycle-node-how-to-shut-down-cleanly/
But it doesn't work

fujitatomoya:
@Rezenders it would be appreciated if you can try https://github.com/ros2/rclcpp/pull/2450

fujitatomoya:
it has been 4 years, but finally we can close this. https://github.com/ros2/rclcpp/pull/2450 is merged.

fujitatomoya:
Just a follow-up comment.

at this moment, that is a responsibility for user application to manage the state control including shutdown, see more information for https://github.com/ros2/rclcpp/issues/2520#issuecomment-2159092031.

PR #2450, opened 2024-03-14
call shutdown in LifecycleNode dtor to avoid leaving the device in un…

fujitatomoya:
…known state.

address https://github.com/ros2/rclcpp/issues/997

### DIFF
--- a/rclcpp_lifecycle/src/lifecycle_node.cpp
+++ b/rclcpp_lifecycle/src/lifecycle_node.cpp
@@ -152,6 +152,22 @@ LifecycleNode::LifecycleNode(
 
 LifecycleNode::~LifecycleNode()
 {
+  // shutdown if necessary to avoid leaving the device in unknown state
+  if (LifecycleNode::get_current_state().id() !=
+    lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED)
+  {
+    auto ret = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR;
+    auto finalized = LifecycleNode::shutdown(ret);
+    if (finalized.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED ||
+      ret != rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS)
+    {
+      RCLCPP_WARN(
+        rclcpp::get_logger("rclcpp_lifecycle"),
+        "Shutdown error in destruction of LifecycleNode: final state(%s)",
+        finalized.label().c_str());
+    }
+  }
+
   // release sub-interfaces in an order that allows them to consult with node_base during tear-down
   node_waitables_.reset();
   node_time_source_.reset();
--- a/rclcpp_lifecycle/test/test_lifecycle_node.cpp
+++ b/rclcpp_lifecycle/test/test_lifecycle_node.cpp
@@ -513,6 +513,80 @@ TEST_F(TestDefaultStateMachine, shutdown_from_each_primary_state) {
   }
 }
 
+TEST_F(TestDefaultStateMachine, test_shutdown_on_dtor) {
+  auto success = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+  auto reset_key = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR;
+
+  bool shutdown_cb_called = false;
+  auto on_shutdown_callback =
+    [&shutdown_cb_called](const rclcpp_lifecycle::State &) ->
+    rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn {
+      shutdown_cb_called = true;
+      return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+    };
+
+  // PRIMARY_STATE_UNCONFIGURED to shutdown via dtor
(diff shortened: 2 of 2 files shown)


fujitatomoya:
@Barry-Xu-2018 @alsora could you take a look as 1st review?

rclcpp_lifecycle/src/lifecycle_node.cpp
```diff
@@ -152,6 +152,22 @@ LifecycleNode::LifecycleNode(
...
+  {
+    auto ret = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR;
+    auto finalized = LifecycleNode::shutdown(ret);
+    if (finalized.id() != lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED ||
+      ret != rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS)
+    {
+      RCLCPP_WARN(
+        rclcpp::get_logger("rclcpp_lifecycle"),
+        "Shutdown error in destruction of LifecycleNode: final state(%s)",
+        finalized.label().c_str());
```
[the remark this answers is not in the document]
Barry-Xu-2018: Agree. Aside from providing warning messages, there aren't any other operations available here.

Barry-Xu-2018:
I agree with adding tests for destructing LifecycleNode under different states of lifecycle node. However, the code added in this PR indeed is difficult to check by tests.   

LGTM

rclcpp_lifecycle/test/test_lifecycle_node.cpp
```diff
@@ -447,6 +447,72 @@ TEST_F(TestDefaultStateMachine, bad_mood) {
   EXPECT_EQ(1u, test_node->number_of_callbacks);
 }
 
+
+TEST_F(TestDefaultStateMachine, shutdown_from_each_primary_state) {
```
alsora, line 451: this test doesn't seem to exercise the new feature, as it calls `shutdown` before letting the node go out of scope.
should we have an additional test where we verify that the shutdown callback is always invoked when the destructor is invoked if the node wasn't in the correct state?
for example the shutdown callback could set a global flag that is checked by the test

fujitatomoya: i think that is hard, `LifecycleNode` derived class `EmptyLifecycleNode` will be destructed before `LifecycleNode`, that means it will not call the `on_shutdown` callback via `LifecycleNode` destruction.

i tried as following.

```cpp
class EmptyLifecycleNode : public rclcpp_lifecycle::LifecycleNode
{
public:
  explicit EmptyLifecycleNode(const std::string & node_name)
  : rclcpp_lifecycle::LifecycleNode(node_name)
  {
    shutdown_cb_called = false;
  }

  inline static bool shutdown_cb_called;

protected:
  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
  on_shutdown(const rclcpp_lifecycle::State &)
  {
    EXPECT_EQ(State::TRANSITION_STATE_SHUTTINGDOWN, get_current_state().id());
    shutdown_cb_called = true;
    return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
  }
};
```

```cpp
TEST_F(TestDefaultStateMachine, test_shutdown_on_dror) {
  auto success = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
  auto reset_key = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR;

  // PRIMARY_STATE_INACTIVE to shutdown
  {
    auto ret = reset_key;
    auto test_node = std::make_shared<EmptyLifecycleNode>("testnode");
    EXPECT_EQ(State::PRIMARY_STATE_UNCONFIGURED, test_node->get_current_state().id());
    auto configured = test_node->configure(ret);
    EXPECT_EQ(success, ret);
    EXPECT_EQ(configured.id(), State::PRIMARY_STATE_INACTIVE);
    EXPECT_FALSE(EmptyLifecycleNode::shutdown_cb_called);
  }
  EXPECT_TRUE(EmptyLifecycleNode::shutdown_cb_called); // this will be false
}
```

fujitatomoya: @alsora what do you think? i think current test codes can verify the situation that destructor can meet to call `shutdown`.

fujitatomoya: Ah, sorry maybe we can use `register_on_shutdown` and free func with global variable to make sure that shutdown is called via destructor. (i guess that originally @Barry-Xu-2018 said.)

fujitatomoya: addressed in https://github.com/ros2/rclcpp/pull/2450/commits/73bb3e5840501de4f08e7b6635589138b74f78c1

fujitatomoya:
@alsora requesting another review! thanks in advance.

alsora (approved):

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20769/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15067/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21449/)

alsora:
Running Linux CI again.
The previous failure seems unrelated.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20772/)

fujitatomoya:
@Mergifyio backport iron humble

Issue #2452, opened 2024-03-18
:farmer: Flaky test `test_executors_timer_cancel_behavior` 


Crola1702:
## Bug report

**Required Info:**

- Operating System:
  - Windows
- Installation type:
  - Source
- Version or commit hash:
  - Rolling
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
1. Run a build in a windows job (preferably [windows_repeated](https://ci.ros2.org/view/nightly/job/nightly_win_rep/))
2. See `rclcpp.test_executors_timer_cancel_behavior` test fail

#### Expected behavior

Test should pass

#### Actual behavior

`rclcpp.test_executors_timer_cancel_behavior` test is failing

#### Additional information

Reference build: https://ci.ros2.org/view/nightly/job/nightly_win_rep/3297/

Test regression:
* [projectroot.test.rclcpp.test_executors_timer_cancel_behavior](https://ci.ros2.org/view/nightly/job/nightly_win_rep/3297/testReport/junit/projectroot.test/rclcpp/test_executors_timer_cancel_behavior/)

Log output:
```
[ RUN      ] TestTimerCancelBehavior/MultiThreadedExecutor.testBothTimerCancelThenResetT2Behavior
C:\ci\ws\src\ros2\rclcpp\rclcpp\test\rclcpp\executors\test_executors_timer_cancel_behavior.cpp(398): error: Expected: (std::abs(t1_runs_initial - t2_runs_initial)) <= (1), actual: 2 vs 1

[  FAILED  ] TestTimerCancelBehavior/MultiThreadedExecutor.testBothTimerCancelThenResetT2Behavior, where TypeParam = rclcpp::executors::MultiThreadedExecutor (2526 ms)
```

> **Note:** sometimes (26% of the times) it fails with this output. Other builds ([example](https://ci.ros2.org/view/nightly/job/nightly_win_rep/3295/testReport/junit/projectroot.test/rclcpp/test_executors_timer_cancel_behavior/)) contain a silent fail

---

Flakiness report:

job_name|last_fail|first_fail|build_count|failure_count|failure_percentage
-- | -- | -- | -- | -- | --
nightly_win_rep|2024-03-18|2024-03-04|16|15|93.75
nightly_win_rel|2024-03-17|2024-03-03|15|5|33.33
nightly_win_deb|2024-03-16|2024-03-09|15|3|20.0

fujitatomoya:
[...]

PR #2453, opened 2024-03-18
relax the test simulation rate for timer canceling tests.

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/2452

### DIFF
--- a/rclcpp/test/rclcpp/executors/test_executors_timer_cancel_behavior.cpp
+++ b/rclcpp/test/rclcpp/executors/test_executors_timer_cancel_behavior.cpp
@@ -102,7 +102,7 @@ private:
 class ClockPublisher : public rclcpp::Node
 {
 public:
-  explicit ClockPublisher(float simulated_clock_step = .001f, float realtime_update_rate = 0.25f)
+  explicit ClockPublisher(float simulated_clock_step = .001f, float realtime_update_rate = 0.10f)
   : Node("clock_publisher"),
     ros_update_duration_(0, 0),
     realtime_clock_step_(0, 0),


rclcpp/test/rclcpp/executors/test_executors_timer_cancel_behavior.cpp
```diff
@@ -102,7 +102,7 @@ class TimerNode : public rclcpp::Node
 class ClockPublisher : public rclcpp::Node
 {
 public:
-  explicit ClockPublisher(float simulated_clock_step = .001f, float realtime_update_rate = 0.25f)
+  explicit ClockPublisher(float simulated_clock_step = .001f, float realtime_update_rate = 0.10f)
```
fujitatomoya, line 105: this relaxes simulation rate from one forth to one tenth, should give enough relief for the timing sensitive tests.

alsora (approved):

clalancette (approved):

Crola1702:
Just running a repeated build to see if it reduces the flakiness: [Build Status](https://ci.ros2.org/job/ci_windows/21246/)

Crola1702:
Yeah, it seems right! Thanks for pushing this!

Crola1702:
Hold on. Just noticed that the CI builds are using *sequential-executor*, these kind of tests are mostly failing with parallel executor, so the issue is not completely fixed:
* [nightly_win_deb-3037](https://ci.ros2.org/view/nightly/job/nightly_win_deb/3037/)
* [nightly_win_rep-3298](https://ci.ros2.org/view/nightly/job/nightly_win_rep/3298/)

fujitatomoya:
@Crola1702 really appreciate for checking the CI.

@clalancette @alsora 

i believe i misunderstood the current test behavior.

i was expecting that

https://github.com/ros2/rclcpp/blob/1c350d0d7fb9c7158e0a39057112486ddbd38e9a/rclcpp/test/rclcpp/executors/test_executors_timer_cancel_behavior.cpp#L332

uses simulation clock to mitigate the racy system clock condition especially for windows to avoid ***the problem*** such as https://github.com/ros2/rclcpp/issues/2452

but it is not implemented as designed, since it uses system clock for the entire test. (that said nobody uses simulation clock.)

https://github.com/ros2/rclcpp/blob/1c350d0d7fb9c7158e0a39057112486ddbd38e9a/rclcpp/test/rclcpp/executors/test_executors_timer_cancel_behavior.cpp#L39-L48

https://github.com/ros2/rclcpp/blob/1c350d0d7fb9c7158e0a39057112486ddbd38e9a/rclcpp/test/rclcpp/executors/test_executors_timer_cancel_behavior.cpp#L202-L203

so actually what i have done here is making the situation worse, to give more time window for rasy timer condition?

i guess we already have made the mistake in https://github.com/ros2/rclcpp/pull/2375, precisely this specific commit https://github.com/ros2/rclcpp/pull/2375/commits/666e9d5cce3dfd7ddee4f7bc0ffd02b30b9ebfbb?

i would want to do the following after the confirmation of above my understanding.

- revert https://github.com/ros2/rclcpp/pull/2453 (i do not think we need this change.)
- change `use_sim_time` into `true`.

alsora:
@fujitatomoya good catch; the timers manager in rclcpp doesn't support ROS sim_time as it uses the std library wait primitives.

```
timers_cv_.wait_for(lock, time_to_sleep.value(), [this]() {return timers_updated_;});
```

Our internal implementation of ROS overrides the stdlibrary with the ROS sim time, so we didn't catch the problem.

I would suggest for now to disable the events executor from this test while we look for a solution.

fujitatomoya:
> P.S. does ROS provide a "sim-time aware" condition variable?

as far as i know, i do not think we have that... @clalancette @mjcarroll @wjwwood 

i will summarize
- creating another issue for `TimerManager` simulation clock support
- reverting https://github.com/ros2/rclcpp/pull/2453
- quick patch and removing `EventExecutor` from the timer test with simulation clock later.

@Crola1702 again, thanks for catching this, i will get back to you soon.

PR #2454, opened 2024-03-19
[documentation] Explain when `MultiThreadedExecutor` is running multiple threads

Cakem1x:
This is a tiny PR that adds a class docstring for the `MultiThreadedExecutor`. It includes a note about which methods support multithreaded execution.

Please let me know if this is not the right way to suggest additions to the documentation.
Also, please double check if the note is actually true or can be improved.

**My context:**
I had a unit test with two nodes that communicate with each other and I use the `MultiThreadedExecutor` to spin both of these nodes. One node publishes a TF, the other one regularly waits for that TF.

In the individual test cases, I have been using `multi_threaded_executor_.spin_until_future_complete(some_future);`. This leads to a flaky test that sometimes succeeds and sometimes fails. The reason for that seems to be that the `MultiThreadedExecutor` is only using a single thread when `spin_until_future_complete` is used.
After changing my test setup to use `multi_threaded_executor_.spin();` in a separate thread and just do `some_future.wait_for(...)` in the test cases, everything is working fine.

### DIFF
--- a/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
+++ b/rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
@@ -32,6 +32,12 @@ namespace rclcpp
 namespace executors
 {
 
+/**
+ * The Multi-threaded executor uses multiple threads to do work while spinning.
+ *
+ * Note, however, that it will only use multiple threads via its blocking `spin()` method.
+ * When using other methods, like e.g. `rclcpp::Executor:spin_until_future_complete(some_future)`, only the main thread will be used to do work.
+ */
 class MultiThreadedExecutor : public rclcpp::Executor
 {
 public:


rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -32,6 +32,12 @@ namespace rclcpp
 namespace executors
 {
 
+/**
+ * The Multi-threaded executor uses multiple threads to do work while spinning.
+ *
+ * Note, however, that it will only use multiple threads via its blocking `spin()` method.
+ * When using other methods, like e.g. `rclcpp::Executor:spin_until_future_complete(some_future)`, only the main thread will be used to do work.
+ */
```
fujitatomoya, line 40: how about adding the following on top of `MultiThreadedExecutor::spin`?

```cpp
  /// Multiple-threaded implementation of spin.
  /**
   * This function will block until work comes in, and try to execute all work with
   * configured multiple threads, and then repeat the process until canceled.
   * It may be interrupt by a call to rclcpp::Executor::cancel() or by ctrl-c
   * if the associated context is configured to shutdown on SIGINT.
   * \throws std::runtime_error when spin() called while already spinning
   */
```

i think `rclcpp::Executor:spin_until_future_complete(some_future)` explains it good enough.

Cakem1x: Thanks for reviewing! I am happy to add your suggestion to the spin method as part of this PR. :+1:

However, I don't think it addresses the point of my PR. I don't want to push this too hard. It's fine if you think it's unnecessary / out of place. I just want to clarify what exactly my point was:
While it explains that the `spin()` method will use multiple threads, it does not clarify that **only**(?) the `spin()` method will work with multiple threads. My (naive) expectation was that **all** spin-related methods I call on the `MultiThreadedExecutor` would run with multiple threads. Including methods inherited from the `rclcpp::Executor` (maybe except some, like `spin_once`, which explicitly only work on a single item).
Did I miss something that could have changed my expectation?

The documentation of [`rclcpp::Executor:spin_until_future_complete(some_future)`](https://docs.ros2.org/latest/api/rclcpp/classrclcpp_1_1Executor.html#a484f01fbbbdc2604f349586d2b2f6830) does not talk about multiple threads ("`Spin (blocking) until the future is complete, it times out waiting, or rclcpp is interrupted."`). I think that's fine, since it is probably nice if the base class does not need to know about its specializations' properties. So adding it there felt out of place, which is why I added the comment to the `MultiThreadedExecutor` class.

In the end, I ended up inspecting the implementation of the `MultiThreadedExecutor` and its base class `rclcpp::Executor` to see that the multi threading stuff is implemented only in the overridden `spin()` method and the methods the `Executor` implements do **not** use `virtual void spin()` (and are not overridden by the `MultiThreadedExecutor`).

I have found that another method, [spin_some](https://docs.ros2.org/latest/api/rclcpp/classrclcpp_1_1Executor.html#ae821cd4cf1040fe2da521fb19a458fa1), actually has a note about the default implementation only being suitable for a single-threaded model of execution. Which was what tipped me off while looking for the source of my unexpected behavior ("maybe other methods of the `rclcpp::Executor` are also only suitable for single-threaded use. Does the `MultiThreadedExecutor` have a specialized implementation? (no)").

fujitatomoya: thanks for the explanation.

i see what you mean here. i was thinking that under the class `MultiThreadedExecutor` only implements `spin` so it would be obvious for the user. but that is the problem here.

> maybe other methods of the rclcpp::Executor are also only suitable for single-threaded use

probably we would want to add more documentation for `rclcpp::Executor` here about single-threaded? because that is the behavior for `rclcpp::Executor`, and then user can see that is inherited from `rclcpp::Executor`.

i am not inclined to add docs like `only the main thread will be used to do work` to `MultiThreadedExecutor`. if we have `MultiThreadedXXXExecutor`, we need to add that every time. instead, i think we can add the docs where that behavior is actually implemented?

Cakem1x: TL;DR: I will add a note to `rclcpp::Executor`'s `spin_until_future_complete` method. This would have saved past-me from the mistakes I made. :)

> i am not inclined to add docs like only the main thread will be used to do work to MultiThreadedExecutor. if we have MultiThreadedXXXExecutor, we need to add that every time. instead, i think we can add the docs where that behavior is actually implemented?
True, good point, that would be annoying.

> class MultiThreadedExecutor only implements spin so it would be obvious for the user. but that is the problem here.
You're right, I was jumping to conclusions. I just saw a method that sounded like it would do what I wanted and I tried to use it without much consideration.
But by just looking at the headers/docs, I do not think it is obvious that `spin_until_future_complete` does not use a virtual method that might be implemented by the `MultiThreadedExecutor` such that you would get multi-threaded execution.
Of course, after seeing that only `spin()` is overwritten, and considering how spin() behaves, it should be fairly obvious.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -347,6 +347,7 @@ class Executor
 
   /// Spin (blocking) until the future is complete, it times out waiting, or rclcpp is interrupted.
   /**
+   * Handles one work item after the other, hence, it does not support parallel processing.
```
fujitatomoya, line 350: I would prefer to add `can do` instead of `cannot do` here. how about the following?

```suggestion
   * The implementation is suitable for a single-thread model of execution.
```

fujitatomoya:
@Cakem1x btw, can you also address DCO error?

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -61,8 +61,12 @@ class MultiThreadedExecutor : public rclcpp::Executor
   RCLCPP_PUBLIC
   virtual ~MultiThreadedExecutor();
 
+  /// Multi-threaded implementation of spin.
```
Zard-C, line 64: Multi-thread 😸

rclcpp/include/rclcpp/executors/multi_threaded_executor.hpp
```diff
@@ -61,8 +61,12 @@ class MultiThreadedExecutor : public rclcpp::Executor
...
   /**
-   * \sa rclcpp::Executor:spin() for more details
+   * This function will block until work comes in, and try to execute all work with
```
alsora, line 66: This is not correct.
This function won't block until works come in, but rather it will block until interrupted.

alsora (changes requested):
I think that the new documentation is misleading.

PR #2464, opened 2024-03-22
Do not generate the exception when action service response timeout.

fujitatomoya:
part of https://github.com/ros2/ros2/issues/1462

### DIFF
--- a/rclcpp_action/src/server.cpp
+++ b/rclcpp_action/src/server.cpp
@@ -343,6 +343,14 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
       response_pair.second.get());
   }
 
+  if (ret == RCL_RET_TIMEOUT) {
+    RCLCPP_WARN(
+      pimpl_->logger_,
+      "Failed to send goal response %s (timeout): %s",
+      to_string(uuid).c_str(), rcl_get_error_string().str);
+    rcl_reset_error();
+    return;
+  }
   if (RCL_RET_OK != ret) {
     rclcpp::exceptions::throw_from_rcl_error(ret);
   }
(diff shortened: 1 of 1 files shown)


fujitatomoya:
this replaces https://github.com/ros2/rclcpp/pull/2463

rclcpp_action/src/server.cpp
```diff
@@ -343,6 +343,14 @@ ServerBase::execute_goal_request_received(std::shared_ptr<void> & data)
       response_pair.second.get());
   }
 
+  if (ret == RCL_RET_TIMEOUT) {
+    RCLCPP_WARN(
+      pimpl_->logger_,
+      "Failed to send goal response %s (timeout): %s",
+      to_string(uuid).c_str(), rcl_get_error_string().str);
+    rcl_reset_error();
+    return;
```
fujitatomoya, line 352: I think returning here is the right thing to do. we do not need to process further since there is no client available.

Barry-Xu-2018: When returning RCL_RET_OK (Most cases return this), call the check function twice.  
How about combining our checks.
```
if (RCL_RET_OK != ret) {
   if (ret == RCL_RET_TIMEOUT) {
       RCLCPP_WARN(
      pimpl_->logger_,
      "Failed to send goal response %s (timeout): %s",
      to_string(uuid).c_str(), rcl_get_error_string().str);
    rcl_reset_error();
    return;
   } else {
   }
}
```

rclcpp_action/src/server.cpp
```diff
@@ -483,6 +491,15 @@ ServerBase::execute_cancel_request_received(std::shared_ptr<void> & data)
...
   }
 
+  if (ret == RCL_RET_TIMEOUT) {
+    GoalUUID uuid = request->goal_info.goal_id.uuid;
+    RCLCPP_WARN(
+      pimpl_->logger_,
+      "Failed to send cancel response %s (timeout): %s",
+      to_string(uuid).c_str(), rcl_get_error_string().str);
+    rcl_reset_error();
+    return;
```
jmachowinski, line 501: Wouldn't it make sense to change the method signature to something else than void ?

fujitatomoya: hmm, probably. so that ServerBase can what needs to be done based on the return code. i am not gonna do that in this PR, but that can be follow-up issue. (i am considering to backport this iron and humble.) thanks!

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21314/)

alsora (approved):

fujitatomoya:
@Mergifyio backport humble iron

PR #2466, opened 2024-03-25
API Changes for Multihtreaded Events Executor

jmachowinski:
These are the API changes needed in order to implement an external Multihtreaded Events Executor

@mjcarroll @alsora @fujitatomoya


jmachowinski:
Writing tests was a good suggestion, I screwed up the clock interface, it's not thread safe, I will redesign it...

rclcpp/test/benchmark/benchmark_executor.cpp
```diff
@@ -58,65 +58,277 @@ class PerformanceTestExecutor : public PerformanceTest
...
+}
 
-  callback_count = 0;
-  reset_heap_counters();
+BENCHMARK_F(PerformanceTestExecutor, multi_thread_executor_spin_some)(benchmark::State & st)
+{
+  executor_spin_some<rclcpp::executors::MultiThreadedExecutor>(st);
+}
 
-  for (auto _ : st) {
```
Zard-C, line 80: maybe 
```cpp
for ([[maybe_unused]] auto _ : st) {
    st.PauseTiming();
``` 
is more readable

jmachowinski: I agree, but this is not the style used in the rest of the code.

jmachowinski:
Redesigned the clock sleep interruption. Can the reviewers have a good lock at the mutex/lock/wait construct, its hard to get it right without data races....

rclcpp/src/rclcpp/clock.cpp
```diff
@@ -79,8 +82,20 @@ Clock::now() const
...
+  {
+    std::unique_lock lock(impl_->wait_mutex_);
+    impl_->stop_sleeping_ = true;
+  }
+  impl_->cv_.notify_one();
+}
+
 bool
-Clock::sleep_until(Time until, Context::SharedPtr context)
+Clock::sleep_until(
```
alsora, line 96: is it possible for multiple threads to call `sleep_until` on the same clock object?

jmachowinski: Yes one could do that. The cancel mechanism will only wake up one thread currently.

jmachowinski: I just checked the documentation of clock. It does not explicitly state that it is safe for usage from multiple threads, therefore I would not assume that it needs to handle this case.

jmachowinski:
I'll need to add a way to terminate the clock and all sleeps as soon as the shutdown call was done.

rclcpp/test/rclcpp/test_clock.cpp
```diff
@@ -0,0 +1,180 @@
...
+  node->set_parameter({"use_sim_time", true});
+  auto clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
+  rclcpp::TimeSource time_source(node);
+  time_source.attachClock(clock);
+
+  EXPECT_TRUE(clock->ros_time_is_active());
+
+  test_wakeup_after_sleep(clock);
+  test_wakeup_before_sleep(clock);
+}
```
fujitatomoya, line 142: can we use `::testing::TestWithParam<ClockType>` to bundle these based on the clock types?

```cpp
enum class ClockType
{
  RCL_STEADY_TIME,
  RCL_SYSTEM_TIME,
  RCL_ROS_TIME,
};

/// Timer testing bring up and teardown
class TestClockWakeup : public ::testing::TestWithParam<ClockType>
{
...
TEST_P(TestClockWakeup, test_name_xxx) {
...
```

fujitatomoya (commented):
overall lgtm, a couple of minor comments.

rclcpp/test/rclcpp/test_clock.cpp
jmachowinski: Done

rclcpp/include/rclcpp/clock.hpp
```diff
@@ -193,6 +193,11 @@ class Clock
   bool
   ros_time_is_active();
 
+  RCLCPP_PUBLIC
+  void
+  cancel_sleep_or_wait();
+
+
```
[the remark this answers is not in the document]
jmachowinski: done

jmachowinski:
rebased to rolling

mjcarroll:
Waiting for this to get in before running CI: https://github.com/ros2/rosidl_typesupport_fastrtps/pull/117

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20807/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15111/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21490/)

jmachowinski:
rebased and (hopefully) fixed windows compile error

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20813/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15117/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21497/)

jmachowinski:
Dropped the test refactoring.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20834/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15138/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21521/)

mjcarroll:
@jmachowinski this needs a merge from `rolling` because of the added API

jmachowinski:
rebased. test_executors_timer_cancel_behavior is flaky on my machine. I hope this gets fixed by #2500

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20835/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15139/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21522/)

fujitatomoya (approved):
lgtm with green CI

mjcarroll:
Looks like a failure in test_actions cli, which is unrelated to these changes, is this test known-flaky? https://ci.ros2.org/job/ci_linux/20835/testReport/junit/ros2action.ros2action.test/test_cli/test_cli/

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -470,6 +470,7 @@ class Executor
   void
   execute_any_executable(AnyExecutable & any_exec);
 
+public:
```
wjwwood, line 473: Why is this necessary?

rclcpp/include/rclcpp/executors.hpp
```diff
@@ -79,10 +79,11 @@ using rclcpp::executors::SingleThreadedExecutor;
  *   If the time spent inside the blocking loop exceeds this timeout, return a `TIMEOUT` return code.
  * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
  */
-template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
+template<typename FutureT, typename Executor, typename TimeRepT = int64_t,
+  typename TimeT = std::milli>
 rclcpp::FutureReturnCode
 spin_node_until_future_complete(
-  rclcpp::Executor & executor,
+  Executor & executor,
```
wjwwood, line 86: Why does this need to be a template argument? Are we actually passing an executor here that doesn't use `rclcpp::Executor` as a base?

rclcpp/test/rclcpp/test_clock.cpp
```diff
@@ -0,0 +1,236 @@
...
+using namespace std::chrono_literals;
+
+enum class ClockType
+{
+  RCL_STEADY_TIME,
+  RCL_SYSTEM_TIME,
+  RCL_ROS_TIME,
+};
+
+
```
wjwwood, line 36: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

rclcpp/test/rclcpp/test_clock.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "../utils/rclcpp_gtest_macros.hpp"
+
+using namespace std::chrono_literals;
+
+enum class ClockType
+{
+  RCL_STEADY_TIME,
+  RCL_SYSTEM_TIME,
+  RCL_ROS_TIME,
+};
```
wjwwood, line 34: Is this used?

wjwwood:
I merged https://github.com/ros2/rclcpp/pull/2500, so we can see if that addressed what you hoped it did @jmachowinski. Also, I have some questions about the changes to `executor.hpp` and `executors.hpp`, which I'd like to understand before giving the 👍.

rclcpp/include/rclcpp/executor.hpp
jmachowinski: Same as below I don't derive from rclcpp::Executor

rclcpp/include/rclcpp/executors.hpp
jmachowinski: Yes, I don't derive from rclcpp:: Executor

wjwwood: Is that something we actually need to support? I'd like to understand the use case here a little bit more. I'm suspicious of doing this. It prevents us from clearly defining the API of an executor and instead requires us to use these template functions that kind of duck-type object that was passed in.

Instead, I'd prefer to settle on changes to the base class to make it easier to inherit from, rather than make that optional.

wjwwood:
I'm highly suspect of the changes to the Executor API to make it so you don't need to inherit from `rclcpp::Executor`. I don't think we should go this route until it's explained better why that's necessary.

rclcpp/include/rclcpp/executors.hpp
wjwwood: For example, if the data in the `rclcpp::Executor` base class is not desirable, we should make a new base for that called like `rclcpp::ExecutorInterface` which defines the required API.

alsora: +1 on using an executor base class

fujitatomoya: makes sense, we want to use base class to avoid taking this path if possible.

jmachowinski: I agree, a base class is more desirable. Big question though, can we make this happen until monday ?

jmachowinski: As to why I need / want this change: 

rclcpp::Executor currently does stuff in the constructor that is not desirable for me, like registering callbacks etc.
I also hat name collisions with my variables and the ones already present.

Some methods of rclcpp::Executor are not virtual, therefore I will get a wrong behavior, if someone would call them on my Executor.
E.g. : 
https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L247

https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L276

https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L281

https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L311

https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L318

https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L424

https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L433

To me the proposed changes looked the least intrusive.

jmachowinski:
rebased to rolling

rclcpp/include/rclcpp/executors.hpp
wjwwood: > I agree, a base class is more desirable. Big question though, can we make this happen until monday ?

The thing is that if we merge this just because Monday is coming up, then we have to come back and undo what we merged through a deprecation cycle, which is very undesirable.

> https://github.com/ros2/rclcpp/blob/90e451154cc23c4c231ca55dc103358f38cfd999/rclcpp/include/rclcpp/executor.hpp#L247

This one is a template function, so that cannot be virtual and you shouldn't need to override that, only the function that it calls.

For the other functions, I'd prefer to see those get marked `virtual` than to make the executor base class not a requirement.

jmachowinski: > This one is a template function, so that cannot be virtual and you shouldn't need to override that, only the function that it calls.

I am aware of that, but the function it calls is also not virtual.

Actually, why is this a template function in the first place ?

rclcpp/include/rclcpp/executor.hpp
jmachowinski: removed

rclcpp/test/rclcpp/test_clock.cpp
jmachowinski: removed

rclcpp/test/rclcpp/test_clock.cpp
jmachowinski: nope, removed

jmachowinski:
I reworked this PR. Now everything is rclcpp::Executor is made virtual. 

I changed two function, that were template function, without any reason to normal functions.

I reworked the template function spin_until_future_complete, so that it can be fully overwritten by a subclass.

I added an constructor, that will not touch anything in the system.

@wjwwood can you do a review ?

rclcpp/include/rclcpp/executors.hpp
wjwwood: > Actually, why is this a template function in the first place ?

It's the easiest way to handle a wide array of input chrono types for the timeout, at least as far as I know. Perhaps there's a better way to do that now, but at the time I believe that's why it was made that way.

wjwwood (commented):
Pretty close, had a few small style things and a suggestion to scope down the changes.

Also, I think there might still be some style issues? please run `ament_uncrustify path/to/rclcpp --reformat` if you haven't already (saw some `if(` that should be `if (`).

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -242,42 +242,30 @@ class Executor
...
-    std::shared_ptr<NodeT> node,
-    std::chrono::duration<RepT, T> timeout = std::chrono::duration<RepT, T>(-1))
-  {
-    return spin_node_once_nanoseconds(
-      node->get_node_base_interface(),
-      std::chrono::duration_cast<std::chrono::nanoseconds>(timeout)
-    );
-  }
+    const std::shared_ptr<rclcpp::Node> & node,
+    std::chrono::nanoseconds timeout = std::chrono::nanoseconds(-1));
```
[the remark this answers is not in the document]
jmachowinski: I don't **really** need to, but all other functions in the executor have the same signature and compile just fine. Any std::chono::duration type should auto convert to std::chrono::nanoseconds.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -429,10 +376,19 @@ class Executor
...
+  virtual bool
   is_spinning();
 
 protected:
+  // constructor that will not setup any internals.
+  /**
+  * This constructor is intended to be used by any derived executor,
+  * that explicitly does not want to use the default implementation provided
+  * by this class. This constructor is guaranteed, to not modify the system
+  * state.
```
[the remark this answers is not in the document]
jmachowinski: The other constructor registers shutdown callbacks.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -447,6 +403,10 @@ class Executor
     rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node,
     std::chrono::nanoseconds timeout);
 
+  virtual FutureReturnCode spin_until_future_complete_impl(
```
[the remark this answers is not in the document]
jmachowinski: We got two colliding styles here, there is also
spin_once -> spin_once_impl and spin_some -> spin_some_impl

jmachowinski:
> Pretty close, had a few small style things and a suggestion to scope down the changes.
> 
> Also, I think there might still be some style issues? please run `ament_uncrustify path/to/rclcpp --reformat` if you haven't already (saw some `if(` that should be `if (`).

I did that, uncrustify doesn't fix this. BTW, why are we not using clang-format ? In my experience it works way better then the uncrustify / cpplint construct.

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
jmachowinski: Added docs

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -366,52 +354,11 @@ class Executor
...
-      if (now >= end_time) {
-        return FutureReturnCode::TIMEOUT;
-      }
-      // Subtract the elapsed time from the original timeout.
-      timeout_left = std::chrono::duration_cast<std::chrono::nanoseconds>(end_time - now);
-    }
-
-    // The future did not complete before ok() returned false, return INTERRUPTED.
-    return FutureReturnCode::INTERRUPTED;
+    );
```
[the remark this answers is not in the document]
jmachowinski: I would like to keep this change, as it allows me to use a second thread to wait for the future, instead of the current construct.  But if you have a hard opinion on dropping it, I am open for it.

wjwwood:
> I did that, uncrustify doesn't fix this (?).

I did this and if found the issues (and auto-formatted the fix when I used `--reformat`):

```
william@markook-ubuntu-2204:~/ros2_ws/src/ros2/rclcpp (git:wjwwood/executor_features:163f727)
$ ament_uncrustify .
...
Code style divergence in file 'rclcpp/test/rclcpp/test_clock.cpp':

--- rclcpp/test/rclcpp/test_clock.cpp
+++ rclcpp/test/rclcpp/test_clock.cpp.uncrustify
@@ -116,4 +116,4 @@
-        Clocks,
-        TestClockWakeup,
-        ::testing::Values(
-            RCL_SYSTEM_TIME, RCL_ROS_TIME, RCL_STEADY_TIME
+  Clocks,
+  TestClockWakeup,
+  ::testing::Values(
+    RCL_SYSTEM_TIME, RCL_ROS_TIME, RCL_STEADY_TIME
@@ -185,2 +185,3 @@
-  for(size_t nr = 0; nr < thread_finished.size(); nr++) {
-    threads.push_back(std::thread(
+  for (size_t nr = 0; nr < thread_finished.size(); nr++) {
+    threads.push_back(
+      std::thread(
@@ -189 +190 @@
-        // make sure the thread starts sleeping late
+          // make sure the thread starts sleeping late
@@ -192 +193 @@
-      }));
+        }));
@@ -198 +199 @@
-  for(const bool & finished : thread_finished) {
+  for (const bool & finished : thread_finished) {
@@ -209,2 +210,2 @@
-    for(const bool finished : thread_finished) {
-      if(!finished) {
+    for (const bool finished : thread_finished) {
+      if (!finished) {
@@ -219 +220 @@
-  for(const bool finished : thread_finished) {
+  for (const bool finished : thread_finished) {
@@ -223 +224 @@
-  for(auto & thread : threads) {
+  for (auto & thread : threads) {
...
```

Are you running it correctly?

clalancette:
> BTW, why are we not using clang-format ? In my experience it works way better then the uncrustify / cpplint construct.

We've had many long, detailed conversations about this, some as recently as a month ago.  While I won't rehash all of the arguments, the short of it is:

1.  It is not possible to configure `clang-format` to meet our preferred style.
2.  Even if we were to change our preferred style to meet what `clang-format` wants, we have literally hundreds of packages in the core that would need to be changed.  And they would all have to be changed at once.  And it would mean that we could no longer cleanly backport from Rolling to stable branches.

We have many real problems in the code, this doesn't seem to be one of them (in my opinion).

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
wjwwood: Arguably the others taking only `nanoseconds` is a probably a bug, see: https://stackoverflow.com/a/22362867

It seems like in C++20 we could have a function that takes `std::chrono::duration<auto>`, but we don't have that right now.

I'll remove this change in my fix up.

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
wjwwood: > We got two colliding styles here, there is also
> spin_once -> spin_once_impl and spin_some -> spin_some_impl

Yeah, I'll leave it as-is, but I think this is related to the templated functions that take duration vs just taking nanoseconds, I still believe the templated approach is better, and in that case the narrowing to the non-template protected methods are truly just to cast it down to nanoseconds, so that makes sense in those cases. Likely if I had noticed I would have suggesting the newer functions follow that pattern, but it's not that important now.

wjwwood:
Closing in favor of https://github.com/ros2/rclcpp/pull/2506

wjwwood:
> better then the uncrustify / cpplint construct.

Also, for what it's worth, I think `clang-tidy` might overlap a bit with `cpplint`, but it wouldn't completely replace it. most projects that use `cpplint` also use `clang-tidy` or something similar along side it.

jmachowinski:
> 
> Are you running it correctly?

Here is a paste of my system, it is clearly not working, but why ?

https://pastebin.com/gGbPLdaA

clalancette:
> Here is a paste of my system, it is clearly not working, but why ?

Oh, you know, we've recently updated to a new version of uncrustify (0.78.1), that does things slightly differently than the old way.  What version of uncrustify are you using?

jmachowinski:
> Oh, you know, we've recently updated to a new version of uncrustify (0.78.1), that does things slightly differently than the old way. What version of uncrustify are you using?

Recent installation, uncrustify_vendor installs 0.78.1f
```
$ uncrustify --version
Uncrustify-0.78.1_f
``` 

I even patched ament_uncrustify to check if it picked up the right version, and this one also reports Uncrustify-0.78.1_f

I have an Uncrustify-0.72.0_f installed by apt, but even if I uninstall this version, my system still reports no style divergence.

PR #2468, opened 2024-03-27
fix flakiness in TestTimersManager unit-test

alsora:
fix flakiness in TestTimersManager check_one_timer_cancel_doesnt_affect_other_timers

the previous version of the test was relying on the assumption that a timer with 1ms period gets called at least 6 times if the main thread waits 15ms. this is true most of the times, but it's not guaranteed, especially when running the test on windows CI servers. the new version of the test makes no assumptions on how much time it takes for the timers manager to invoke the timers, but rather focuses on ensuring that they are called the right amount of times, which is what's important for the purpose of the test

we need to ensure that if a timers manager has N timers and 1 gets cancelled, all the other ones keep getting invoked.

### DIFF
--- a/rclcpp/test/rclcpp/test_timers_manager.cpp
+++ b/rclcpp/test/rclcpp/test_timers_manager.cpp
@@ -399,7 +399,13 @@ TEST_F(TestTimersManager, check_one_timer_cancel_doesnt_affect_other_timers)
   timers_manager->start();
 
   // Wait for t1 to be canceled
+  auto loop_start_time = std::chrono::high_resolution_clock::now();
   while (!t1->is_canceled()) {
+    auto now = std::chrono::high_resolution_clock::now();
+    if (now - loop_start_time >= std::chrono::seconds(30)) {
+      FAIL() << "timeout waiting for t1 to be canceled";
+      break;
+    }
     std::this_thread::sleep_for(3ms);
   }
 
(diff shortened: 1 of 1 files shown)


alsora:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20642/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14978/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21351/)

rclcpp/test/rclcpp/test_timers_manager.cpp
```diff
@@ -397,11 +398,30 @@ TEST_F(TestTimersManager, check_one_timer_cancel_doesnt_affect_other_timers)
   // Start timers thread
   timers_manager->start();
 
-  std::this_thread::sleep_for(15ms);
+  // Wait for t1 to be canceled
+  while (!t1->is_canceled()) {
```
clalancette, line 402: Can we put an overall timeout on this, something like 30 seconds?  It just ensures that if there is a bug, this test won't hang our CI forever.

alsora:
FYI @mwcondino (as the author of the test)
Can you please take a look and validate my changes

rclcpp/test/rclcpp/test_timers_manager.cpp
```diff
@@ -397,11 +398,30 @@ TEST_F(TestTimersManager, check_one_timer_cancel_doesnt_affect_other_timers)
...
+  }
+
+  EXPECT_TRUE(t1->is_canceled());
+  EXPECT_FALSE(t2->is_canceled());
+  EXPECT_EQ(t1_runs, cancel_iter);
+
+  // Verify that t2 is still being invoked
+  const size_t start_t2_runs = t2_runs;
+  const size_t num_t2_extra_runs = 6;
+  while (t2_runs < start_t2_runs + num_t2_extra_runs) {
```
clalancette, line 413: Same here.

clalancette (changes requested):
The idea is great, thanks for looking into this.

I suggested two things inline to make it more robust, but  I'm otherwise happy to see this fixed.

rclcpp/test/rclcpp/test_timers_manager.cpp
alsora: Done in e06bd91a8979b6218fd8bca530bf38750dec1524

alsora:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20643/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14979/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21352/)

clalancette (approved):

mjcarroll (approved):

mwcondino:
> FYI @mwcondino (as the author of the test) Can you please take a look and validate my changes

Changes look good, thank you for the improvements @alsora!

Issue #2451, opened 2024-03-14
`spin_some` easily deadlocks action clients


asasine:
## Bug report
When running an `rclcpp_action::Client<T>` in using `spin_some()` and a `rclcpp::Rate`, if the server you connect to publishes feedback at a faster rate than your `rclcpp::Rate`, your client waits for the goal to be accepted forever, effectively becoming deadlocked.

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Binaries
- Version or commit hash:
  - Galactic
- DDS implementation:
  - Cyclone DDS
- Client library (if applicable):
  - rclcpp

Note: based on my understanding of the root cause (see `Additional information`), I believe this bug exists in rolling too, but I've only tested on Galactic as that's what I have access to at the moment.

#### Steps to reproduce issue
The below server and client reproduce 100% of the time on my machine when running over loopback. Compilation settings shouldn't matter. Once built, they can be run with `ros2 run ...` or directly with `./path/to/binary`. I'm using `action_tutorials_interfaces/action/Fibonacci`, but this should repro with any action interface.

<details>
<summary>Server</summary>

```cpp
#include <action_tutorials_interfaces/action/fibonacci.hpp>
#include <rclcpp/rclcpp.hpp>
#include <rclcpp_action/rclcpp_action.hpp>

int main(int argc, char *argv[])
{
  rclcpp::init(argc, argv);
  auto node = rclcpp::Node::make_shared("fast_feedback_server");
  using Action = action_tutorials_interfaces::action::Fibonacci;
  std::shared_ptr<rclcpp_action::ServerGoalHandle<Action>> goal_handle = nullptr;

  auto server = rclcpp_action::create_server<Action>(
      node, "fast_feedback",
      [&](const rclcpp_action::GoalUUID &, std::shared_ptr<const Action::Goal>)
      { return rclcpp_action::GoalResponse::ACCEPT_AND_EXECUTE; },
      [&](const std::shared_ptr<rclcpp_action::ServerGoalHandle<Action>>)
      {
        RCLCPP_INFO(node->get_logger(), "Cancellation requested, accepting.");
        goal_handle = nullptr;
        return rclcpp_action::CancelResponse::ACCEPT;
      },
      [&](const std::shared_ptr<rclcpp_action::ServerGoalHandle<Action>> gh)
      {
        RCLCPP_INFO(node->get_logger(), "Goal was accepted");
        goal_handle = gh;
      });

[...]

Issue #1679, opened 2021-05-19
Action client feedback callback does not reliably trigger


schornakj:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04 LTS
- Installation type:
  - Binaries
- Version or commit hash:
  - Foxy, Rolling
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - rclcpp

#### Expected behavior

My packages' integration tests should run with the same behavior both locally and in CI.

#### Actual behavior

When running in an environment with a comparatively small number of CPU cores (for example, 2 cores for Travis and GH Actions runners), my tests sometimes fail because action feedback that I require to pass the test was never received by my action client.

To test this, I'm using the `fibonacci_action_server` node from the `action_tutorials_cpp` package, which includes an action server of type `action_tutorials_interfaces::action::Fibonacci`.

My test node sends a goal to the Fibonacci action with `order = 5`. I expect the action server to publish 4 feedback messages. In the log output from my tests, I see that the server does publish 4 feedback messages, but the feedback callback registered with my action client is only called for the first 3 messages. Since my test needs to receive 4 feedback messages to succeed, it times out waiting for the 4th message and fails.

```
taskset -c 0-1 python3 -u /opt/ros/rolling/share/ament_cmake_test/cmake/run_test.py /tmp/asd \
  --package-name rclcpp_multicore_ci_test \
  --env TEST_LAUNCH_DIR=src/rclcpp_multicore_ci_test/test TEST_EXECUTABLE=build/rclcpp_multicore_ci_test/test_fixture_node \
  --command src/rclcpp_multicore_ci_test/test/test_nodes.py
```

```
[INFO] [fibonacci_action_server-1]: process started with pid [2053673]
[INFO] [test_node-2]: process started with pid [2053675]
[test_node-2] [==========] Running 1 test from 1 test suite.
[test_node-2] [----------] Global test environment set-up.
[test_node-2] [----------] 1 test from TaskSequenceInterface
[test_node-2] [ RUN      ] TaskSequenceInterface.setup_plan_exec_through_ros
[fibonacci_action_server-1] 1621445668.874821 [0] fibonacci_: using network interface wlp0s20f3 (udp/192.168.0.4) selected arbitrarily from: wlp0s20f3, docker0
[...]

PR #2471, opened 2024-04-01
rclcpp_action: take and execute service entities in priority.

fujitatomoya:
address https://github.com/ros2/rclcpp/issues/2451 and https://github.com/ros2/rclcpp/issues/1679

Note: more tests need to be done before CI.

### DIFF
--- a/rclcpp_action/src/client.cpp
+++ b/rclcpp_action/src/client.cpp
@@ -550,21 +550,8 @@ ClientBase::clear_on_ready_callback()
 std::shared_ptr<void>
 ClientBase::take_data()
 {
-  if (pimpl_->is_feedback_ready) {
-    std::shared_ptr<void> feedback_message = this->create_feedback_message();
-    rcl_ret_t ret = rcl_action_take_feedback(
-      pimpl_->client_handle.get(), feedback_message.get());
-    return std::static_pointer_cast<void>(
-      std::make_shared<std::tuple<rcl_ret_t, std::shared_ptr<void>>>(
-        ret, feedback_message));
-  } else if (pimpl_->is_status_ready) {
-    std::shared_ptr<void> status_message = this->create_status_message();
-    rcl_ret_t ret = rcl_action_take_status(
-      pimpl_->client_handle.get(), status_message.get());
-    return std::static_pointer_cast<void>(
-      std::make_shared<std::tuple<rcl_ret_t, std::shared_ptr<void>>>(
-        ret, status_message));
-  } else if (pimpl_->is_goal_response_ready) {
+  // Take the service entities 1st to apply to the client
+  if (pimpl_->is_goal_response_ready) {
     rmw_request_id_t response_header;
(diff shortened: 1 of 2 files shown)


fujitatomoya:
- IMO, we do not guarantee all feedback messages must be delivered to the client. (if the goal accept response is not yet delivered, that would be dropped. but eventually, feedback messages will be received on the client once goal handle is being tracked.)
- Service responses should be handled with priority than `feedback` and `status` topics because it changes internal action state in the client.

rclcpp_action/src/client.cpp
```diff
@@ -550,21 +550,8 @@ ClientBase::clear_on_ready_callback()
...
-        ret, feedback_message));
-  } else if (pimpl_->is_status_ready) {
-    std::shared_ptr<void> status_message = this->create_status_message();
-    rcl_ret_t ret = rcl_action_take_status(
-      pimpl_->client_handle.get(), status_message.get());
-    return std::static_pointer_cast<void>(
-      std::make_shared<std::tuple<rcl_ret_t, std::shared_ptr<void>>>(
-        ret, status_message));
-  } else if (pimpl_->is_goal_response_ready) {
+  // Take the service entities 1st to apply to the client
```
alsora, line 553: The change looks good, but I'm not sure I understand the meaning of this comment.
What's a "service entity"?

alsora (approved):

rclcpp_action/src/client.cpp
fujitatomoya: terms might be wrong. but it was supposed to mean it should take the action service executables 1st before action subscriptions.

fujitatomoya:
@alsora thanks for the review.

this will change the action client behavior internally so i need to do more verification locally, and i would like to have more feedback for this before further. please keep it open for a while.

rclcpp_action/src/client.cpp
alsora: I would suggest a more verbose comment, for example:

```
/*
The order of these statements is important when using waitset-based executors because this method
will be called only once per executor iteration, regardless of how many events are ready.
We first check whether the state of the action goal has changed (i.e. whether we received a goal response,
a result response or a cancellation response).
This ensures that secondary events (e.g. feedback messages) won't prevent the goal state from advancing.
*/
```

jmachowinski:
Uhm, this implementation is ontop of a racy state...

We should merge https://github.com/ros2/rclcpp/pull/2250 first.

jmachowinski:
Regarding the issue itself. From my point of view, the design of the action is broken, as it uses multiple communication channels to communicate, thus creating this kind of problems. 

Btw you can create the same bug by spamming feedback, and make an action never terminate.

fujitatomoya:
@jmachowinski thanks for the comment. this one is not in the merge queue yet, i need to take some more time to verify this, since this changes the action behavior, i am expecting that userspace (tests) would break.

fujitatomoya:
https://github.com/ros2/rclcpp/pull/2612 replaces this one.

PR #1957, opened 2022-06-24
Revert "Revert "Add support for spin_until_timeout (#1821)""

wjwwood:
Reverts ros2/rclcpp#1956

An un-revert of https://github.com/ros2/rclcpp/pull/1874, see https://github.com/ros2/rclcpp/pull/1874#issuecomment-1165811614.

PR #2475, opened 2024-04-02
Add support for spin_until_timeout (take 2)

christophebedard:
Reverts #1956, so it un-reverts #1874 (with some updates), see https://github.com/ros2/rclcpp/pull/1874#issuecomment-1165811614

Supersedes #1957

This goes along with the following PRs:

- https://github.com/ros2/ros2_tracing/pull/2
- TBD


rclcpp/include/rclcpp/executor.hpp
```diff
@@ -350,6 +351,51 @@ class Executor
...
+        std::is_same_v<bool, RetT>,
+        "Conditional callable has to return boolean type");
+      return spin_until_complete_impl(condition, timeout);
+    } else {
+      auto check_future = [&condition]() {
+          return condition.wait_for(std::chrono::seconds(0)) ==
+                 std::future_status::ready;
+        };
+      return spin_until_complete_impl(check_future, timeout);
+    }
```
wjwwood, line 384: I know that this was discussed last time we tried to merge all of these prs (last release), and I messed that up somewhat due to merging it incorrectly (sorry about that), but since we're here again, I do want to (re)ask a fundamental question (https://github.com/ros2/rclcpp/pull/1874#discussion_r835586538) here, do we need to deprecate the `spin_until_future_complete()` API and have this logic in the new API?

Just looking at it again now, it seems like the pros and cons of deprecating it are:

- Pro: one fewer methods to document and test
- Con: many downstream changes to only the function name being called
- Con: complicates the interface of the new `spin_until_complete()` by making it take various types, and therefore requires you read the docstring to know what the `ConditionT` can be
- Con: uses template deduction when overloading could work (e.g. two `spin_until_complete()` methods that take different args), which is harder for text editors and other tools to handle

Am I missing anything?

If not, this seems to be a poor tradeoff and I'd argue that keeping the syntactic sugar function `spin_until_future_complete()` and having it just call the new `spin_until_complete()` is far simpler than deprecating it and complicating this function.

@hliberacki @SteveMacenski @clalancette @ivanpauno @alsora @fujitatomoya CC you all since you were all part of the original discussion a year ago.

wjwwood (commented):
I'm sorry to just now be commenting on this change (even though it's the second iteration), but I wanted to revoice some concerns about the way that the change is being made.

rclcpp/include/rclcpp/executor.hpp
SteveMacenski: > Con: complicates the interface of the new spin_until_complete() by making it take various types, and therefore requires you read the docstring to know what the ConditionT can be

At least it can then take various types :smile: 

> Con: uses template deduction when overloading could work (e.g. two spin_until_complete() methods that take different args), which is harder for text editors and other tools to handle

I'd be OK with that. I just want the spin APIs to be independent of specifically futures to allow for spinning other types of things in the future -- for example with a simpler `SimpleServiceClient`-esk object to spin the object which itself contains some future information internally that a user shouldn't have to worry about for most normal cases.  

If that's done via overloading or templating, I'm fine with either. The main objective I had was renaming the method to be more general and take up less space to specify what its arguments tell you is the case: `spin_until_future_complete(future, timeout)` is less natural than `spin_until_complete(future, timeout)`

christophebedard: I do agree that the type of `condition` is confusing.

I personally do not have a strong opinion; I was just trying to finally take care of this big set of PRs.

fujitatomoya: > keeping the syntactic sugar function spin_until_future_complete() and having it just call the new spin_until_complete() is far simpler than deprecating it and complicating this function.

actually this is reasonable, i am inclined to be away from deprecating. they can live together.

clalancette: > I'd be OK with that. I just want the spin APIs to be independent of specifically futures to allow for spinning other types of things in the future -- for example with a simpler `SimpleServiceClient`-esk object to spin the object which itself contains some future information internally that a user shouldn't have to worry about for most normal cases.

Let me ask a question; does this PR implement the functionality to spin on "something else" other than a future?

If it does, then I suggest what we do is to modify this PR to reimplement `spin_until_future_complete` in terms of that function.

If it doesn't, then I suggest we just close this.  Without a second user, we are almost certain to get the API wrong and then have to add another variant.

In either case, I think we should *not* deprecate or remove `spin_until_future_complete`.  That is well-named (even if narrow), and there is no reason to cause tons of work downstream to fix that.  That also means that I think we should close all of the downstream PRs from this, since everything downstream will still compile and work as before.

wjwwood: > spin_until_future_complete(future, timeout) is less natural than spin_until_complete(future, timeout)

Yeah, I guess I don't agree on that one, but it is subjective. But either way, it's pretty disruptive to change it to be a single name. And I have a hard time arguing (to myself) that it's inaccurately named or that it's misleading or something like that, which would be a more compelling reason to change it despite the disruption.

So, I'd prefer to avoid the rename myself.

> Let me ask a question; does this PR implement the functionality to spin on "something else" other than a future?

Yes, it just expects a callable object and calls it each time the executor wakes up, returning if the callable returns true or it timesout, otherwise it continues to spin.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -570,6 +572,55 @@ class Executor
...
+    if (timeout_ns > std::chrono::nanoseconds::zero()) {
+      end_time += timeout_ns;
+    }
+    std::chrono::nanoseconds timeout_left = timeout_ns;
+
+    // Preliminary check, finish if conditon is done already.
+    if (condition()) {
+      return FutureReturnCode::SUCCESS;
+    }
+
```
jmachowinski, line 594: This implementation is racy. If the condition switches after this line of code, and an infinite timeout is given, we got a dead lock.

rclcpp/include/rclcpp/executor.hpp
```diff
@@ -350,6 +351,51 @@ class Executor
...
+   *   `-1` is block forever, `0` is non-blocking.
+   *   If the time spent inside the blocking loop exceeds this timeout, return a TIMEOUT return
+   *   code.
+   * \return The return code, one of `SUCCESS`, `INTERRUPTED`, or `TIMEOUT`.
+   */
+  template<typename ConditionT, typename DurationT = std::chrono::milliseconds>
+  FutureReturnCode
+  spin_until_complete(
+    const ConditionT & condition,
+    DurationT timeout = DurationT(-1))
```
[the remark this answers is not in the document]
christophebedard: Any suggestion? We could add a more generic version of `FutureReturnCode` (what name?), but it would be functionally the same.

rclcpp/include/rclcpp/executor.hpp
christophebedard: Alright, in conclusion:

1. Do add `spin_until_complete`, but with a callable, _not_ templated to accept a callable or a future
1. Do not deprecate `spin_until_future_complete`
1. Do implement `spin_until_future_complete` using `spin_until_complete`
1. Do the same for `rclpy`
1. Do close downstream PRs; no need to rename calls in downstream packages from `spin_until_future_complete` to `spin_until_complete`

Can you confirm?

clalancette: That seems like a good plan to me.

rclcpp/include/rclcpp/executor.hpp
jmachowinski: Currently the function is named spin_until_complete. Therefore I would expect, I can pass it a std::future or any condition, and it will do the correct thing, and protect me from this kind of races. 

This is clearly not the case. I think it is not even possible with the current API. 
To solve this, we would also need to pass a std::unique_lock, like done here:
https://en.cppreference.com/w/cpp/thread/condition_variable/wait

christophebedard: Wait, I don't see how this can cause an _unwanted_ infinite wait/spin.

It is assumed that `condition()` is initially `false`. If it is `true` from the get-go, then it just returns without spinning. Otherwise, it spins once before checking if `condition()` is `true`, otherwise it spins once again, and returns if it times out. If `condition()` never changes to `true` and `timeout` is negative, then yes it will spin infinitely, but that's intentional.

Am I missing something?

christophebedard: That still feels like user error to me. @wjwwood thoughts?

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -283,6 +283,37 @@ TEST_F(TestExecutor, spin_some_elapsed) {
...
+    std::chrono::milliseconds(0), [&]() {
+      timer_called = true;
+    });
+  dummy.add_node(node);
+  // Wait for the wall timer to have expired.
+  dummy.spin_for(std::chrono::milliseconds(0));
+
+  ASSERT_TRUE(timer_called);
+}
+
```
jmachowinski, line 301: I am missing a test, that shows, that the spin terminated within a time span X after the condition became ready.

jmachowinski (changes requested):

rclcpp/include/rclcpp/executor.hpp
jmachowinski: I am more on the side of bad documentation of the function, and bad API, as you can pass in anything here.

I am pushing this point, as this is a public function, that is even promoted in the tutorials (the future version). But using it is not trivial if you want to get deterministic results. Whatever you pass in needs to explicitly wake up the wait after it changes the result of the condition to true. And doing this requires knowledge of rather internal stuff.

christophebedard: FYI we will not be changing downstream calls to use this new API (https://github.com/ros2/rclcpp/pull/2475#discussion_r1550080593), so I've updated https://github.com/ros2/ros2_documentation/pull/4252 to only mention this change and not change any tutorials.

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
wjwwood: That's a good step, but I was also thinking, since the Condition must return a bool, should this function not return one? The reason for a complex type for futures is that it is multi-state, but that doesn't seem to be true for the condition.

rclcpp/include/rclcpp/executor.hpp
wjwwood: I think it's a documentation issue, it's expected that if you ask for it to wait until a condition is complete, then you need your condition to change before waking the wait set. This has all the same considerations as passing a predicate to a condition variable, except that a condition variable can spurious wake up, so even if you do something wrong (e.g. have the predicate change after interrupting, resulting in a data race), it might not result in a deadlock.

As to it being bad API, I don't see how we could enforce that people write their condition function such that it works correctly.

The only thing we could do (that I can think of) is to restrict the API to only take futures that are produced and managed by rclcpp itself (that way we can ensure they're coded correctly), but that's even more restrictive than what is there now.

Or were you thinking of something else?

Is it sufficient to document that a condition function (a.k.a. a predicate) must change its answer before waking the wait set?

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -488,6 +517,37 @@ TEST_F(TestExecutor, spin_until_future_complete_future_already_complete) {
...
+  auto condition = []() {return true;};
+  EXPECT_EQ(
+    rclcpp::FutureReturnCode::SUCCESS,
+    dummy.spin_until_complete(condition, std::chrono::milliseconds(1)));
+}
+
+TEST_F(TestExecutor, spin_until_complete_returns_after_condition) {
+  DummyExecutor dummy;
+  auto node = std::make_shared<rclcpp::Node>("node", "ns");
+  bool spin_called = false;
```
wjwwood, line 531: nitpick: isn't this more like?

```suggestion
  bool timer_called = false;
```

rclcpp/test/rclcpp/test_executor.cpp
```diff
@@ -488,6 +517,37 @@ TEST_F(TestExecutor, spin_until_future_complete_future_already_complete) {
...
+  const auto condition_delay = std::chrono::milliseconds(10);
+  const auto start = std::chrono::steady_clock::now();
+  auto condition = [&]() {return std::chrono::steady_clock::now() - start > condition_delay;};
+  EXPECT_EQ(
+    rclcpp::FutureReturnCode::SUCCESS,
+    dummy.spin_until_complete(condition, std::chrono::seconds(1)));
+  EXPECT_TRUE(spin_called);
+  const auto end_delay = std::chrono::steady_clock::now() - start;
+  EXPECT_GE(end_delay, condition_delay);
+  EXPECT_LT(end_delay, std::chrono::milliseconds(500));
```
wjwwood, line 548: This check doesn't look great to me, we make no guarantee that functions like `spin_until_complete()` start-up and shutdown within 500ms, discounting the actual blocking/executing time. I would just not check this. The test is still useful without it because it's checking that the condition eventually gets checked for true even though it doesn't directly wake the wait set.

wjwwood: I mean they should be that fast, but also this isn't a performance test, and on a very slow computer or on a non-real-time OS, this kind of thing cannot be assumed and avoid flakes in rare cases. Increasing the time from 500ms will only ever asymptotically reduce the chances of a flake, but not avoid them completely.

wjwwood:
I had a few questions/suggestions/comments, but generally the approach lgtm. I think once you've addressed them to your satisfaction @christophebedard, then it should be good to go.

rclcpp/include/rclcpp/executor.hpp
jmachowinski: I see tow ways:

change the function to
```c++
  spin_until_complete(
    const std::function<bool(void)> & condition,
    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1),
     std::chrono::duration<TimeRepT, TimeT> condition_check_interval = std::chrono::millisecons(10))
``` 
And have the condition checked in this interval.

You could create an object like this
```c++
class WaitCondition
{
    WaitCondition();

    void stop_spin();
};
```

And change the function to
```c++
  spin_until_complete(
    const std::function<bool(WaitCondition &cond)> & condition,
    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
```  

Now, that I tried to use this interface I am confused of how it is intended to work. @christophebedard  can you provide an example of a function, you would pass into this interface ?

rclcpp/test/rclcpp/test_executor.cpp
christophebedard: makes sense, renamed

rclcpp/test/rclcpp/test_executor.cpp
christophebedard: Totally agree. I removed this line. I wasn't sure if you also wanted me to remove the `EXPECT_GE(end_delay, condition_delay);` line above.

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
christophebedard: Can `spin_until_complete` not be `INTERRUPTED` even for non-future conditions? It just means it stopped spinning because `rclcpp::ok()` returned `false`.

rclcpp/include/rclcpp/executor.hpp
jmachowinski: Regarding my question
Is the condition actively changed by an external event / callback ? 
It is supposed to be changed by a second thread in some way ?

rclcpp/test/rclcpp/test_executor.cpp
wjwwood: I think that one is reasonable, since the call needs to take longer than the condition's sleep, but also I wouldn't expect it to fail. I think it's fine as-is now that you removed the last check.

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
wjwwood: That's true, I suppose.

With that in mind, the name `rclcpp::ConditionWaitReturnCode` is kind of awkward, since it's the first time "Wait" is used in this context (the function isn't called "spin_and_wait_for_condition()" or something like that), but I'm not sure of an alternative, maybe the explicit, but verbose, `rclcpp::SpinUntilCompleteReturnCode`?

Another route would be to copy `std::future_status` (https://en.cppreference.com/w/cpp/thread/future_status), and call it something like `rclcpp::condition_status`, but that (over)uses the term `condition` in my opinion, and we might want that later, so I'd prefer something more namespaced, like `rclcpp::SpinUntilCompleteStatus`.

I'll leave it up to you, but I like the `rclcpp::SpinUntilCompleteReturnCode` option, since it matches our prexisting `rclcpp::FutureReturnCode`.

rclcpp/include/rclcpp/executor.hpp
wjwwood: >  Is the condition actively changed by an external event / callback ?

Yes, that's the implication I think. You could either have a condition that's affected by some callback in the executor or you could have a condition that is completely decoupled with the tasks in the executor, but that could lead to a dead lock as you suggested. If you wanted to do the latter it should be probably be done using a guard condition that you add to the executor to ensure the condition's state is checked after it has been changed by triggering the guard condition after changing the condition.

>  It is supposed to be changed by a second thread in some way ?

As I said, yes, you could but you'd need to ensure the wait set is notified after the condition had changed, which you could do from a second thread using a guard condition.

> can you provide an example of a function, you would pass into this interface ?

I think the simplest example is the implementation of the syntactic sugar function `spin_until_future_complete()`, something like:

```c++
  template<typename FutureT, typename TimeRepT = int64_t, typename TimeT = std::milli>
  FutureReturnCode
  spin_until_future_complete(
    const FutureT & future,
    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1))
  {
    const auto condition = [&future]() {
        return future.wait_for(std::chrono::seconds(0)) == std::future_status::ready;
      };
    return spin_until_complete(condition, timeout);
  }
```

But even in this case (and we should document that), the future given to this function should have it's promise set before notifying the wait set (and it must notify the wait set). The obvious use case for this is using a future created by rclcpp itself, e.g. `client->send_request_async(...)` which returns a future which is completed in an executor callback. But there are others, like completing a future from a callback in a test to spin until a callback was called by the executor.

rclcpp/include/rclcpp/executor.hpp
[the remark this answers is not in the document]
christophebedard: Sounds good, I changed it to `rclcpp::SpinUntilCompleteReturnCode`.

rclcpp/include/rclcpp/executor.hpp
wjwwood: That would be really helpful, I think. I can't think of way to misuse that API right now, definitely safer.

I'm wondering if that would an acceptable API for the people who pushed for this change in the first place? @SteveMacenski I think you were one of the folks asking for the change (but I'd be happy to hear from others too), what do you think about an API like the above where you get an object that can stop the executor rather than passing a callback function?

The future API was really similar (except a little less safe since the user needed to make sure not to fulfill the promise before changing their data), so I'm afraid it will suffer the same inconveniences.

Personally I like it better.

Also, normally I'd say "yes and ..." ask for it to be a follow up pr, but I think we should probably not have this new `spin_until_complete()` API if we choose to use this newer proposal.

SteveMacenski: > what do you think about an API like the above where you get an object that can stop the executor rather than passing a callback function?

I have 2 general goals:
- Make the APIs easier to use / understand without exposing the technical details under the hood for the Jr Eng / researcher / student that shouldn't be required to learn the details, and 
- Lessen the strict requirement that these are specifically futures -- which also has the benefit of making programs lines shorter since `spin_until_future_complete()` takes up alot of space in line writing application code with this in there day-in day-out. It seems like a minor thing, but with the amount of actions I touch, it would actually be a QoL increase. But I also had some ideas for other non-future things I'd like to spin, but the details of that old dream are escaping me for the moment. 

So, I wouldn't support this change if it requires users in their application code to "own" the construction and management of some new Condition object in order to pass to the spin function from the output of a service / action client -- that's making my first goal even more cumbersome with a further level of indirection to understand. 

If however that's handled under the hood for users as part of the rclcpp API so its hidden from them, no issue. We can have multiple implementations of `spin_until_complete` which takes in different types of inputs, that's largely the purpose of this API change -- decouple from futures and allow for other options.

wjwwood: Ok, with that in mind, why don't we move forward with this change and consider adding a version like you described @jmachowinski soon (after the API freeze though) for the next release. We can also consider if making this "wait condition" style api the recommended way and even deprecating the callback later.

We just don't have time to hash out a new approach right now, and while still error prone, as you pointed out, the callback function isn't really any more hazardous than what we have now, and it satisfies @SteveMacenski's concerns.

wjwwood (approved):
lgtm, now that we've resolved the discussion about the API for now.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20846/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15149/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21532/)

christophebedard:
Should we merge this now, or wait until the `rclpy` PR (https://github.com/ros2/rclpy/pull/1268) has been reviewed and is ready to go too?

rclcpp/include/rclcpp/executor.hpp
jmachowinski: @wjwwood to quote your here :
"The thing is that if we merge this just because Monday is coming up, then we have to come back and undo what we merged through a deprecation cycle, which is very undesirable."

I still haven't seen any example of how this API would be used in a real word application.
@SteveMacenski you want this API for "Make the APIs easier to use / understand without exposing the technical details under the hood for the Jr Eng / researcher / student that shouldn't be required to learn the details, and"

The currently proposed API does not reach this goal, as you need to be aware that you need to wake the wait, and to do this, you need to either have a known side effect of another API (which can change by any internal implementation change) or explicit manage a guard condition.

wjwwood: That is true, and I stand by what I said, we don't want to merge things that we think will be obsolete immediately.

However, I believe that the situation where you're pulling my quote from (https://github.com/ros2/rclcpp/pull/2466#discussion_r1564553671) and this one are quite different. First, it's not clear to me that what you've proposed here would actually replace the desire for the new API in this pr, so we may or may not deprecate what this pr adds even with the approach you suggested implemented and merged later. Whereas in the other situation, it's more of an A or B style design question. If we choose A now, we'll have a mess switching direction to B in the future, and it would never make sense to have both. Therefore it's more important to act when we're sure it's the right approach, unlike here, where adding a new function is by comparison low risk.

wjwwood:
We should really merge them together, because I don't think it's good if we have rclpy and rclcpp deviate from one another.

wjwwood:
My CI above did not include your changes for rclpy unfortunately. We'll have to run separate CI for that.

rclcpp/include/rclcpp/executor.hpp
jmachowinski: I still believe, that the currently proposed API is a to big pitfall for end users.
E.g.
```c++
spin_until_complete([end = std::chrono::system_clock::now() + std::duration::Seconds(1)] ()
{
return end < std::chrono::system_clock::now();
});
```  
will not work as expected. I also don't really see how it will make the live of a user easier in its current state.

I also proposed 
```c++
spin_until_complete(
    const std::function<bool(void)> & condition,
    std::chrono::duration<TimeRepT, TimeT> timeout = std::chrono::duration<TimeRepT, TimeT>(-1),
     std::chrono::duration<TimeRepT, TimeT> condition_check_interval = std::chrono::millisecons(10))
``` 
This approach, which is trivial to implements, and would make the situation way better.

wjwwood: > I also don't really see how the API will make the live of a user easier in its current state.

I also don't link the callback approach is really that helpful (but I think it can result in less code), for that you should talk with @SteveMacenski. But I still think it's no better or worse than using futures.

> this approach, which is trivial to implements, and would make the situation way better.

Having the check interval (essentially polling) is bandaid, really the condition should always be attached to some activity that causes the spin to interrupt in an event driven way. If you want something like the check interval, you can just use the timeout argument and loop the function, or you could use an empty timer, etc. etc.

Most of the time this is used to wait for a Service Client to receive a response to its request, or something similar, and for that this function will work fine, and in fact the future-based version of this worked fine.

---

@christophebedard @fujitatomoya can you guys comment on whether or not to delay this until after the release or not?

christophebedard: >> I also don't really see how the API will make the live of a user easier in its current state.
>
> I also don't think the callback approach is really that helpful (but I think it can result in less code), for that you should talk with @SteveMacenski. But I still think it's no better or worse than using futures.

Just to be clear, I picked up this set of PRs mainly to finally push them across the finish line. I personally do not have a need for this API in itself, so I'd also like to hear what Steve (or any other potential user) has to say here.

> @christophebedard @fujitatomoya can you guys comment on whether or not to delay this until after the release or not?

At this point I'd prefer to just push it to make sure to get it right.

roncapat:
Any news on this?

russkel:
This seems like a low hanging fruit addition?

jmachowinski:
@alsora Can you add this to the agenda of the next client workgroup meeting ?

jmachowinski:
@russkel Do you have a use case for this API ? 

We discussed this in the client workgroup meeting and basically don't see a valid use case for this, that is not flawed / dangerous.

PR #2476, opened 2024-04-02
Fix callback group logic in executor

### LINKED ISSUE #2474: Bug in Executor::get_next_ready_executable ?
Hey

https://github.com/ros2/rclcpp/blob/f9c4894f96ea083fc5acfcea4f1ea83850759e63/rclcpp/src/rclcpp/executor.cpp#L692

I am highly confused by this code block:
```c++
        auto callback_group = entity_iter->second.callback_group.lock();
        if (callback_group && !callback_group->can_be_taken_from()) {
          continue;
        }
``` 

Shouldn't it be ?
```c++
        auto callback_group = entity_iter->second.callback_group.lock();
        if (!callback_group || !callback_group->can_be_taken_from()) {
          continue;
        }
```

mjcarroll:
Fixes #2474

CC: @jmachowinski

### DIFF
--- a/rclcpp/src/rclcpp/executor.cpp
+++ b/rclcpp/src/rclcpp/executor.cpp
@@ -689,7 +689,8 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
       auto entity_iter = current_collection_.timers.find(timer->get_timer_handle().get());
       if (entity_iter != current_collection_.timers.end()) {
         auto callback_group = entity_iter->second.callback_group.lock();
-        if (callback_group && !callback_group->can_be_taken_from()) {
+        if (!callback_group || !callback_group->can_be_taken_from()) {
+          current_timer_index++;
           continue;
         }
         // At this point the timer is either ready for execution or was perhaps
@@ -699,6 +700,7 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
         wait_result_->clear_timer_with_index(current_timer_index);
         // Check that the timer should be called still, i.e. it wasn't canceled.
         if (!timer->call()) {
+          current_timer_index++;
           continue;
         }
         any_executable.timer = timer;
@@ -715,7 +717,7 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
         subscription->get_subscription_handle().get());
       if (entity_iter != current_collection_.subscriptions.end()) {
         auto callback_group = entity_iter->second.callback_group.lock();
-        if (callback_group && !callback_group->can_be_taken_from()) {
+        if (!callback_group || !callback_group->can_be_taken_from()) {
           continue;
         }
         any_executable.subscription = subscription;
@@ -731,7 +733,7 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
       auto entity_iter = current_collection_.services.find(service->get_service_handle().get());
       if (entity_iter != current_collection_.services.end()) {
         auto callback_group = entity_iter->second.callback_group.lock();
-        if (callback_group && !callback_group->can_be_taken_from()) {
+        if (!callback_group || !callback_group->can_be_taken_from()) {
           continue;
         }
         any_executable.service = service;
@@ -747,7 +749,7 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
       auto entity_iter = current_collection_.clients.find(client->get_client_handle().get());
       if (entity_iter != current_collection_.clients.end()) {
         auto callback_group = entity_iter->second.callback_group.lock();
-        if (callback_group && !callback_group->can_be_taken_from()) {
+        if (!callback_group || !callback_group->can_be_taken_from()) {
           continue;
         }
         any_executable.client = client;
@@ -763,7 +765,7 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
       auto entity_iter = current_collection_.waitables.find(waitable.get());
       if (entity_iter != current_collection_.waitables.end()) {
         auto callback_group = entity_iter->second.callback_group.lock();
-        if (callback_group && !callback_group->can_be_taken_from()) {
+        if (!callback_group || !callback_group->can_be_taken_from()) {
           continue;
         }
         any_executable.waitable = waitable;
@@ -781,7 +783,6 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
       any_executable.callback_group->can_be_taken_from().store(false);
     }
   }
-
 
   return valid_executable;
 }
--- a/rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
+++ b/rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
@@ -1,0 +1,139 @@
+// Copyright 2024 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * This test checks that when callback groups go out of scope, that their associated executable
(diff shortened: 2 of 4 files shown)


alsora (approved):
Is it easy to create a unit-test that would fail without this PR and succeeds now?

jmachowinski (approved):

fujitatomoya (commented):
@mjcarroll do we need to apply the same fix for, (where is irrelevant from https://github.com/ros2/rclcpp/pull/2142)

https://github.com/ros2/rclcpp/blob/f9c4894f96ea083fc5acfcea4f1ea83850759e63/rclcpp/src/rclcpp/executors/executor_entities_collection.cpp#L156

`callback_group` is null means that entity is not valid. AFAIS, `rclcpp` does not use  `ready_executables` but user would?

CC: @jmachowinski

jmachowinski:
> Is it easy to create a unit-test that would fail without this PR and succeeds now?
Should be straight forward.
A custom executor, 
Add a cbg
call collect, 
then delete the cbg
Call get next executable.
This failed for me, I got a any executable with an empty cbg

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
```diff
@@ -0,0 +1,132 @@
+// Copyright 2017 Open Source Robotics Foundation, Inc.
```
jmachowinski, line 1: C&P ?

jmachowinski (approved):

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  rclcpp::AnyExecutable next() {
+    rclcpp::AnyExecutable ret;
+    this->get_next_ready_executable(ret);
+    return ret;
+  }
+};
+
+
+TEST(TestCallbackGroup, ValidCbg)
```
wjwwood, line 62: nitpick, we typically use snakecase for the test name and I'd appreciate it if we try to use the full name for things (helps with translation for people who don't speak english and pattern matching), e.g.:

```suggestion
TEST(TestCallbackGroup, valid_callback_group)
```

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
```diff
@@ -0,0 +1,140 @@
...
+  EXPECT_EQ(nullptr, next_executable.client);
+  EXPECT_EQ(nullptr, next_executable.service);
+  EXPECT_EQ(nullptr, next_executable.subscription);
+  EXPECT_EQ(nullptr, next_executable.waitable);
+  EXPECT_EQ(nullptr, next_executable.data);
+
+  rclcpp::shutdown();
+}
+
+TEST(TestCallbackGroup, InvalidCbg)
```
wjwwood, line 98: nitpick:

```suggestion
TEST(TestCallbackGroup, invalid_callback_group)
```

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+  rclcpp::init(0, nullptr);
+
+  // Create a timer associated with a callback group
+  auto node = std::make_shared<rclcpp::Node>("node");
+
+  // Add the callback group to the executor
+  auto executor = CustomExecutor();
+  executor.add_node(node);
+  executor.spin_all(std::chrono::milliseconds(10));
```
wjwwood, line 108: Just curious why is this needed?

wjwwood (commented):
changes lgtm, had a question about the tests

Also a general comment on the test, I'd love to see the use of `spin()` instead of `spin_all()`, or even `spin_once()` in a loop. And I've found that using a custom waitable that can be explicitly triggered rather than a timer can avoid unintentional race conditions and sleeps. Though we've been copying that custom waitable around a lot. It might be a good idea to create a new entity type that is like a `TriggeredCallback` class which would inherit from `Waitable` and would differ from a `GuardCondition` in that it has a callback to be called when the executor executes it, rather than when it is triggered (which the `GuardCondition` class already has). It would definitely be useful in tests and may even be useful for users, allowing them to manually schedule a callback to be run in the executor on demand.

If others agree I could work on that.

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
alsora: According to gtest docs, `PascalCase` should be used

> TEST() arguments go from general to specific. The first argument is the name of the test suite, and the second argument is the test’s name within the test suite. Both names must be valid C++ identifiers, and they should not contain any underscores (_)

from https://google.github.io/googletest/primer.html#simple-tests

jmachowinski:
> And I've found that using a custom waitable that can be explicitly triggered rather than a timer can avoid unintentional race conditions and sleeps

In this special case, the timer is not racy, as we just need it to be ready. Also I would like to keep the timer, as it uncovered a bug causing an endless loop.

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
mjcarroll: This was to clear out all of the node startup transients.

rclcpp/src/rclcpp/executor.cpp
```diff
@@ -689,7 +689,7 @@ Executor::get_next_ready_executable(AnyExecutable & any_executable)
       auto entity_iter = current_collection_.timers.find(timer->get_timer_handle().get());
       if (entity_iter != current_collection_.timers.end()) {
         auto callback_group = entity_iter->second.callback_group.lock();
-        if (callback_group && !callback_group->can_be_taken_from()) {
+        if (!callback_group || !callback_group->can_be_taken_from()) {
```
jmachowinski, line 692: current_timer_index++;

jmachowinski: Also below

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
```diff
@@ -0,0 +1,140 @@
...
+  node.reset();
+  timer.reset();
+
+  // Since the callback group has been reset, this should not be allowed to
+  // be a valid executable (timer and cbg should both be nullptr).
+  // In the regression, timer == next_executable.timer whil
+  // next_executable.callback_group == nullptr, which was incorrect.
+  executor.wait();
+  auto next_executable = executor.next();
+  EXPECT_NE(nullptr, next_executable.timer);
```
jmachowinski, line 130: EXPECT_EQ

mjcarroll:
@jmachowinski I have incorporated your changes from: https://github.com/cellumation/rclcpp/tree/fix_cbg_condition

They have been squashed and I marekd you with co-authored-b

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20760/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15059/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21439/)

rclcpp/test/rclcpp/executors/test_executors_callback_group_behavior.cpp
wjwwood: That's interesting, I never knew that, looking into it, it seems like it doesn't matter now, but could in the future: https://stackoverflow.com/a/34543433

I think we should refactor them all to avoid this condition, but obviously not in this pr. @clalancette

clalancette: Wow, I had no idea.  We have literally thousands of gtests in ROS 2 using underscore.

That said, I actually think we should stick with our current snake-case style, even though it is slightly wrong.  To me it seems better to be consistent, and if we want to fix this (pretty theoretical) issue, we can do it all at once.

jmachowinski:
@mjcarroll I picked the test over to https://github.com/ros2/rclcpp/pull/2494

2494 also fixes the tests fails that came up in the CI.

I think we can close this PR in favor for 2494

wjwwood:
Ok, let's go ahead and close this then. We can reopen it if needed, and I think the feed back was addressed in the new pr as far as I can tell.

For future reference, it's usually better to make prs to other prs rather than replacing them, though sometimes that doesn't work out due to needing to rebase. Or like me you mess up the pr or branch some how and have to close it 🙃.

